PR #256, opened 2020-08-07
tls: introduce mode and sni to cert matching behavior

hbagdi:
This patch adds a `mode` property to TLSConfig which determines the TLS
behavior for each listener.

This patch implements one part of the [TLSConfig proposal](https://docs.google.com/document/d/15fkzMrhN_7tA-i2mHKwZpqcjN1o2Pe9Am9Qt828x1lo/edit#).


apis/v1alpha1/gateway_types.go
```diff
@@ -249,6 +249,39 @@ type Listener struct {
...
 	// ignored otherwise.
 	//
+	// The association of SNIs to Certificate defined in TLSConfig is
+	// defined based on the value of HostnameMatchType for this listener:
+	// - "Domain": Certificate should be used used for the domain and its
+	//   first-level subdomains.
+	// - "Exact": Certificate should be used for the domain only.
+	// - "Any": Certificate in TLSConfig is the default certificate to use.
+	//
+	// The GatewayClass must use the most specific available certificate.
```
robscott, line 259: Maybe GatewayClass be Gateway or implementation here?

apis/v1alpha1/gateway_types.go
```diff
@@ -249,6 +249,39 @@ type Listener struct {
...
+	// - hostname:
+	//     name: foo.example.com
+	//     match: Exact
+	//   port: 443
+	//   protocol: HTTPS
+	//   tls:
+	//     certificateRef:
+	//       resource: secrets
+	//       group: core
+	//       name: foo-example-com-cert
```
robscott, line 280: I like this example but am scared that it could easily become stale as we change the API. Maybe we can simply say the certificate associated with the longest matching hostname should be chosen here?

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -60,7 +80,7 @@ type TLSConfig struct {
 	// values.
 	//
 	// +optional
-	MinimumVersion *string `json:"minimumVersion" protobuf:"bytes,2,opt,name=minimumVersion"`
+	MinimumVersion *string `json:"minimumVersion" protobuf:"bytes,3,opt,name=minimumVersion"`
```
robscott, line 83: I believe this was removed by https://github.com/kubernetes-sigs/service-apis/pull/257

apis/v1alpha1/gateway_types.go
hbagdi: We follow GatewayClass in a lot of places and that's what I borrowed from here. Should I change it in this case or keep as is?

apis/v1alpha1/tlsconfig_types.go
hbagdi: Yes, will fix in the next rebase.

apis/v1alpha1/gateway_types.go
hbagdi: Fair point. I'll remove this and summarize in text in the next rebase.

hbagdi:
Resolved the two comments. PTAL.

apis/v1alpha1/gateway_types.go
```diff
@@ -249,6 +249,16 @@ type Listener struct {
 	// is required if the Protocol field is "HTTPS" or "TLS" and
 	// ignored otherwise.
 	//
+	// The association of SNIs to Certificate defined in TLSConfig is
+	// defined based on the value of HostnameMatchType for this listener:
+	// - "Domain": Certificate should be used used for the domain and its
```
Miciah, line 254: "used" is doubled.

apis/v1alpha1/generated.proto
```diff
@@ -1201,20 +1211,28 @@ message TCPRouteStatus {
...
-  // CertificateRefs is a list of references to Kubernetes objects that each
-  // contain an identity certificate.  The host name in a TLS SNI client hello
-  // message is used for certificate matching and route host name selection.
-  // The SNI server_name must match a route host name for the Gateway to route
-  // the TLS request.  If an entry in this list omits or specifies the empty
+  // Mode defines the TLS behavior for the TLS session initiated by the client.
+  // There are two possible modes:
+  // - Terminate: The TLS session between the downstream client
+  //   and the Gateway is terminated at the Gateway.
+  // - Passthrough: The TLS session NOT terminated by the Gateway. This
```
danehans, line 1218: The godocs should states that all other TLSConfig fields are ignored when mode = Passthrough

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -25,20 +37,28 @@ package v1alpha1
...
-	// CertificateRefs is a list of references to Kubernetes objects that each
-	// contain an identity certificate.  The host name in a TLS SNI client hello
-	// message is used for certificate matching and route host name selection.
-	// The SNI server_name must match a route host name for the Gateway to route
-	// the TLS request.  If an entry in this list omits or specifies the empty
+	// Mode defines the TLS behavior for the TLS session initiated by the client.
+	// There are two possible modes:
+	// - Terminate: The TLS session between the downstream client
+	//   and the Gateway is terminated at the Gateway.
+	// - Passthrough: The TLS session NOT terminated by the Gateway. This
```
danehans, line 44: The godocs should states that all other TLSConfig fields are ignored when mode = Passthrough

danehans:
xref to add backend reencrypt mode at the xRoute level: https://github.com/kubernetes-sigs/service-apis/pull/81/files.

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -25,20 +37,29 @@ package v1alpha1
...
+	//   implies that the Gateway can't decipher the TLS stream except for
+	//   the ClientHello message of the TLS protocol.
+	//   CertificateRef field is ignored in this mode.
+	Mode TLSMode `json:"mode,omitempty" protobuf:"bytes,1,opt,name=mode"`
+
+	// CertificateRef is the reference to Kubernetes object that
+	// contain an identity certificate.
```
jpeach, line 51: Perhaps, "that contains a TLS server certificate and private key"?

apis/v1alpha1/gateway_types.go
```diff
@@ -263,6 +263,16 @@ type Listener struct {
 	// is required if the Protocol field is "HTTPS" or "TLS" and
 	// ignored otherwise.
 	//
+	// The association of SNIs to Certificate defined in TLSConfig is
+	// defined based on the value of HostnameMatchType for this listener:
```
jpeach, line 267: "on the vale of the Hostname field for this Listener"

apis/v1alpha1/gateway_types.go
```diff
@@ -263,6 +263,16 @@ type Listener struct {
...
+	// - "Domain": Certificate should be used for the domain and its
+	//   first-level subdomains.
+	// - "Exact": Certificate should be used for the domain only.
+	// - "Any": Certificate in TLSConfig is the default certificate to use.
+	//
+	// The GatewayClass MUST use the longest matching SNI out of all
+	// available certificate for any TLS handshake.
```
jpeach, line 274: I'm not sure in general about this paragraph. The way I had thought about this was that (in this PR) there is a single CertificateRef, and that contains a single server certificate. The `Hostname` field defines how to match the requested SNI servername, so the proxy should just offer up the certificate it has and hope for the best.

This paragraph seems like it is written to capture the case where the proxy may have multiple certificates to choose from. However, if that is the case, I'd prefer to simply specify that the proxy should serve the certificate that is the most specific match (e.g. it's better to match the SNI name exactly than to serve a wildcard certificate).

hbagdi: > This paragraph

Do you mean the stanza above this which defines how HostnameMatchType and certificates relate or specifically for this line?

jpeach: Yup. Do we need to specify name matching rules for certificates, or is it enough to just say "serve the certificate resolved by CertificateRef`? Maybe that depends on the contents of that resource?

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -25,20 +37,29 @@ package v1alpha1
...
 type TLSConfig struct {
-	// CertificateRefs is a list of references to Kubernetes objects that each
-	// contain an identity certificate.  The host name in a TLS SNI client hello
-	// message is used for certificate matching and route host name selection.
-	// The SNI server_name must match a route host name for the Gateway to route
-	// the TLS request.  If an entry in this list omits or specifies the empty
+	// Mode defines the TLS behavior for the TLS session initiated by the client.
+	// There are two possible modes:
+	// - Terminate: The TLS session between the downstream client
+	//   and the Gateway is terminated at the Gateway.
```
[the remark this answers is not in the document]
hbagdi: We can consider it as part of another patch.

I'd still avoid calling anything "re-encrypt". Most users are mature enough to understand that downstream and upstream TLS are separate things and can be tweaked independently. And the API design should make it even hard to mix the two concerns, which is what we are attempting to do with these TLS patches.

apis/v1alpha1/gateway_types.go
hbagdi: The reason I think we should define it is because with compatible listeners, we get into muddy waters here.
We can skip defining the behavior if we do not want to enforce the behavior at all.

hbagdi: Ping @jpeach.

apis/v1alpha1/tlsconfig_types.go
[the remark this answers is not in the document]
hbagdi: @robscott, Is it okay to resolve this thread for the PR?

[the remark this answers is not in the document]
robscott: Makes sense, thanks for the explanation!

apis/v1alpha1/gateway_types.go
robscott: This specific use feels more tied to implementation than class, but this does not need to block the PR.

robscott:
Thanks for your work on this! This LGTM but will hold for final approval from @jpeach since there's still one thread open.

/lgtm
/hold

apis/v1alpha1/gateway_types.go
jpeach: Let's just merge and see how it goes.

hbagdi:
@robscott Rebased, can you take another look?
/assign @robscott

hbagdi:
Rebased again to resolve the conflict.
ping @bowei

robscott:
@hbagdi this still LGTM, just needs a rebase.

/approve

hbagdi:
Rebase again. @robscott

robscott:
Thanks!

/lgtm

PR #257, opened 2020-08-07
tls: drop tls.MinimumVersion in favor of tls.Options

hbagdi:
Drop tls.MinimumVersion as a defined property and instead use
tls.Options for defining tls.MinimumVersion.


robscott:
Since `tls.options` is just a `map[string]string` it seems like it would be hard to standardize on common options like this. Are there other bits of common TLS config that would also end up in options?

apis/v1alpha1/generated.proto
```diff
@@ -1196,18 +1196,6 @@ message TLSConfig {
...
-  // the TLS_* constants above. Note: MinimumVersion is not strongly
-  // typed to allow implementation-specific versions to be used without
-  // requiring updates to the API types. String must be of the form
-  // "<protocol><major>_<minor>".
-  //
-  // Support: Core for TLS1_{1,2,3}. Implementation-specific for all other
-  // values.
-  //
-  // +optional
-  optional string minimumVersion = 2;
```
yiyangy, line 1209: This changes the support of "minimumVersion" from core to implementation-specific. Is this intended?

hbagdi:
To answer both @robscott and @yiyangy, for now, I think all TLS protocol specific will be folded inside `Options`. This indeed moves MinimumVersion from Core to ImplementationSpecific, we briefly discussed this in one of our meetings. We can introduce this back into core.

bowei:
I'm OK with this, given the diversity of how TLS is specified.


/lgtm
/approve

PR #258, opened 2020-08-07
httproute: drop default HTTPRouteHost

hbagdi:
This patch removes default HTTPRouteHost and instead recommends the
following two ways to implement default routing behavior:
1. For a catch-all route for a hostname or a subset of hostname, use a
`PathMatchType` of `Prefix` with `/` path.
2. For a catch-all route across all hostnames on a given listener, combine (1) with
a listener with `HostnameMatchType` of `Any`.

This should fix https://github.com/kubernetes-sigs/service-apis/issues/255.


robscott:
Thanks! Had to double check that we never had examples referencing this field - must indicate that it was never that useful...

/lgtm

hbagdi:
@bowei @robscott I rebased, another `/lgtm` please?

PR #260, opened 2020-08-07
Update default PathMatchType in HTTPRoute to be prefix.

### LINKED ISSUE #254: Default PathType match should be prefix
The current default behavior for PathType matching is `/` + `Exact`. I think it would be preferable to default to `Prefix` as the path type matching so by default a rule matches all requests. This would make it easier to initially set up a Gateway + Route combination that routed requests to all paths to a Service. I believe this would also more closely match user expectations when interacting with this API.

yiyangy:
This patch changes the default `PathMatchType` in `HTTPRouteMatch` from `Exact` to `Prefix`. 

This should fix #254 .

### DIFF
--- a/apis/v1alpha1/generated.proto
+++ b/apis/v1alpha1/generated.proto
@@ -642,11 +642,11 @@ message HTTPRouteMatch {
   // Support: core (Exact, Prefix)
   // Support: custom (RegularExpression, ImplementationSpecific)
   //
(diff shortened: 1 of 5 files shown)


yiyangy:
I signed it

robscott:
Thanks!

/lgtm
/assign @bowei

PR #263, opened 2020-08-13
Clarifying how AllowedGatewayNamespaces field should be interpreted

robscott:
This commit renames allowedGatewayNamespaceSelector to allowedGatewayNamespaces. This field becomes something that is enforced entirely by a validating webhook. The corresponding condition has also been removed.

/cc @jpeach @hbagdi @yiyangy 
/assign @bowei


apis/v1alpha1/gatewayclass_types.go
```diff
@@ -64,19 +64,12 @@ type GatewayClassSpec struct {
 	Controller string `json:"controller" protobuf:"bytes,1,opt,name=controller"`
 
 	// AllowedGatewayNamespaceSelector is a selector of namespaces that Gateways
-	// can use this GatewayClass from. This is a standard Kubernetes
-	// LabelSelector, a label query over a set of resources. The result of
-	// matchLabels and matchExpressions are ANDed. Controllers must not support
-	// Gateways in namespaces outside this selector.
+	// of this class can be created in. Gateway validation will prevent Gateways
+	// specifying this class from being created outside of these namespaces.
```
hbagdi, line 68: Validation webhooks are cool but can we ALWAYS assume that validation webhook will always be setup if these CRDs are in use?

robscott: I agree that this is not something we can guarantee, but I think it's our least bad option here. Requiring controllers to enforce this gets really messy and leaves open the possibility that a change here could invalidate/break all Gateways using the class. I do think that we'll want to rely on a validating webhook for more than just this use case though, so it would be good for us to make it easy to install and use.

jpeach: Perhaps there's 2 levels of capability here:

1. web hook (may be absent) must enforce the allowed namespaces
2. controllers must ignore gateways that are not in the allowed namespaces

robscott: I'd originally planned for 2 as well, but as I've thought about it more I've realize it's quite dangerous. That's especially true for any implementations where provisioning Gateways takes time. If this field changed and invalidated a number of Gateways, implementations would have to drop all of them with this interpretation. That would make a change on GatewayClass potentially quite destructive and seems to go against the guiding principle of not breaking things that are working. 

That's where this proposal comes from - move this to a validation problem and something that's only enforced on create. We still gain the security/authorization capabilities here, but we don't need to deal with the potential of a GatewayClass change resulting in removing/destroying a large number of Gateways.

hbagdi: I do get the motivation here but this is a divergence from the behavior of this API in other places, which is not necessarily a bad thing.

If we accept this change, my question then is: Why is changing label selector or `onlySameNamespace` not considered to be equally destructive?

robscott: That's a great point. My main concern is that for some implementations provisioning Gateways is much more expensive than provisioning Routes. As an example, recreating a Gateway may result in a new IP being assigned. I think this specific selector is the only one that affects the validity of Gateways. Everything else affects which Routes a Gateway selects, a resource that seems much simpler for all variations to provision. 

With that said, I do not like the divergence of behaviors here. Ideally both would be treated equally. Unfortunately I can't find a great way to provide the same kind of meaningful create-only validation here for RouteNamespaces. Here's one option that could work, but it relies very heavily on cross-resource validation and would have to deal with quite a few potential race conditions:

On Gateway create:
- If RouteNamespaces are specified, ensure they are at least as restrictive as GatewayClass field.
- If not specified, copy values from GatewayClass field.

On Gateway update:
- Ensure RouteNamespaces are at least as restrictive as GatewayClass field.

On GatewayClass update:
- Ensure RouteNamespaces are at least as restrictive for Gateways of this class. If not, require Gateways to be individually updated first.

To try to match that behavior, we could prevent GatewayClass updates that would change allowedGatewayNamespaces to a value that would invalidate any existing Gateways. Of course changing labels of individual namespaces could still bypass this validation.

All of the above feels relatively fragile, and I'm concerned about the potential race conditions here as well as the significant dependency on webhooks. It could potentially make things feel a bit more consistent here though.

hbagdi: Thanks for that write up, that helped me understand the differences in how the gateway and route selector will behave operationally.

So, the problem we are trying to solve here is:

> My main concern is that for some implementations provisioning Gateways is much more expensive than provisioning Routes. As an example, recreating a Gateway may result in a new IP being assigned. I think this specific selector is the only one that affects the validity of Gateways. Everything else affects which Routes a Gateway selects, a resource that seems much simpler for all variations to provision.

which is a problem for all vendors.

Another problem with the change in this PR is that there is no way for a controller to determine if a Gateway that is already created is valid or not:

1. A Gateway passes validation and is running happily, a user updates GatewayClass, making the Gateway invalid.
2. A Gateway is created in an invalid namespace (no webhook or a data-race between gw and gwc create/update operations). The controller starts up and works happily (as per the change in this PR, the controller doesn't need to bother to check if it violates `allowedGatewayNamespaceSelector` in GatewayClass or not. ).

I think this PR is in the right direction and we should explicitly call the following out in docs:
- `allowedGatewayNamespaceSelector` is a field only for validation (this PR)
- If a user is able to bypass validations and all the guardrails and is able to provision a Gateway that is created in namespaces outside `allowedGatewayNamespaceSelector`, the gateway will work.
- In case a Gateway violates `allowedGatewayNamespaceSelector`, the status of the Gateway should reflect so. Meaning, the gateway will continue to serve traffic but it will also have an orphaned or some such condition set to true.

I think that this is similar to Persistent volumes and namespaces. If namespaces in k8s are deleted, PVs that are provisioned in GKE/EKS are not deleted.
In the current case, a namespace deletion will certainly delete the Gateway resource unless we use finalizers.

jpeach: > In case a Gateway violates allowedGatewayNamespaceSelector, the status of the Gateway should reflect so. Meaning, the gateway will continue to serve traffic but it will also have an orphaned or some such condition set to true.

In this case, does the controller still update the violating Gateway? If so. it sounds like we could ship a controller that just monitors the allowed namespaces and updates status on gateways.

robscott: Those are all great points. I agree that adding a condition to Gateway status would be ideal if it is violating `allowedGatewayNamespaceSelector`. Unfortunately the specifics of that implementation make me hesitant to require this condition to be set. Maybe for the sake of this PR we can provide the condition and say that it _may_ be set on a Gateway if it is violating the `allowedGatewayNamespaceSelector`. 

Requiring implementations to set this condition could be problematic if a controller does not have visibility to all Gateways in all namespaces. On the other hand, developing a common controller in addition to a common webhook would result in significant deployment overhead and additional potential failure modes. I think I'd prefer to revisit requiring this condition to be set after our v1alpha1 release. As part of that discussion we could further explore if a shared implementation makes sense.

hbagdi: > Requiring implementations to set this condition could be problematic if a controller does not have visibility to all Gateways in all namespaces. On the other hand, developing a common controller in addition to a common webhook would result in significant deployment overhead and additional potential failure modes.

+1. However, I do feel that sharing code across implementations for these cases will help in further having a better user experience and also lower the development and maintenance cost for all of us.


> Unfortunately the specifics of that implementation make me hesitant to require this condition to be set. Maybe for the sake of this PR we can provide the condition and say that it may be set on a Gateway if it is violating the allowedGatewayNamespaceSelector.

Can you elaborate? I want to understand the concern here.

> Requiring implementations to set this condition could be problematic if a controller does not have visibility to all Gateways in all namespaces.

That shouldn't be the case. If a Gateway resource is created in a cluster, a controller must have access to it, else how would that resource even be satisfied?
If a Gateway is created and no controller is aware of it, isn't that equivalent to that resource not being present in the first place?

robscott: >If a Gateway is created and no controller is aware of it, isn't that equivalent to that resource not being present in the first place?

Let's say a controller is deployed with access to Gateways in namespaces labeled `env=dev` and it implements a GatewayClass that also limits where Gateways of this class can be created to the same subset of namespaces. What happens if a Gateway referencing that class is created in a namespace outside that subset of namespaces? The controller wouldn't know and therefore wouldn't be able to add `ConditionForbiddenNamespaceForClass` to the status for that Gateway. 

If we require controller implementations to set this condition, it means that controllers would need to have access to Gateways in all namespaces. That leaves a shared/common implementation here which could be helpful but of course requires us to test and maintain one more component as well as requiring users of this API to deploy and maintain one more component. 

All of this for a condition that provides relatively minimal value. I think we're all in agreement on this:
>If a user is able to bypass validations and all the guardrails and is able to provision a Gateway that is created in namespaces outside allowedGatewayNamespaceSelector, the gateway will work.

And if that's the case, this condition would really just indicate that "this Gateway shouldn't really be here but it will keep on working anyway".  Maybe instead we should explore adding a condition on Gateway that indicates that a controller is implementing the Gateway.

hbagdi: > Let's say a controller is deployed with access to Gateways in namespaces labeled env=dev and it implements a GatewayClass that also limits where Gateways of this class can be created to the same subset of namespaces. What happens if a Gateway referencing that class is created in a namespace outside that subset of namespaces? The controller wouldn't know and therefore wouldn't be able to add ConditionForbiddenNamespaceForClass to the status for that Gateway.

Correct. But in this case, the Gateway won't even function because the controller can't reach it. The Gateway resource will exist but nothing should happen in the infrastructure.

The point I'm trying to make is, if a controller is controlling/implementing a Gateway and it sees that the Gateway violates `allowedGatewayNamespaceSelector` then the controller MUST add in the ConditionForbiddenNamespaceForClass to the status. This of course doesn't apply to Gateway that is not managed by any controller.

robscott:
We should update GatewayClass docs to reference this discussion as well as other relevant ones.

/close

robscott:
Closed the wrong PR during triage...

/reopen

apis/v1alpha1/gatewayclass_types.go
robscott: Got it, so this condition is meant to be more of a warning to the user? It won't actually have any affect right? It feels like it might get a bit confusing for users if that condition is only set if the controller can see the Gateway. This does seems like a bit of an edge case though.

robscott: I'm not completely convinced a condition here is necessary. I think it would be impossible to guarantee that each controller set the condition since they may not be able to access Gateways in all namespaces. That inconsistency could make the condition less valuable. With that said, I'm interested in what others think here.

/cc @jpeach @danehans

hbagdi: I cant totally see that.
Maybe an event makes sense here? My point is to simply indicate to the user that a Gateway is invalid.

We can track this as a separate issue and this should not be a blocker to get this PR in.

robscott:
This PR has been updated to work better with https://github.com/kubernetes-sigs/service-apis/pull/285.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -64,23 +64,13 @@ type GatewayClassSpec struct {
...
-	//
-	// An empty selector (default) indicates that Gateways can use this
-	// GatewayClass from any namespace.
-	//
-	// When a Gateway attempts to use this class from a namespace that is not
-	// allowed by this selector, the controller implementing the GatewayClass
-	// may add a new "ForbiddenNamespaceForClass" condition to the Gateway
-	// status. Adding this condition is considered optional since not all
-	// controllers will have access to all namespaces.
+	// of this class can be created in. Gateway validation will prevent Gateways
```
hbagdi, line 67: Can we change validation to validation webhook to make it more specific?
Other than that this lgtm.

howardjohn: Isn't it frowned upon to have validation cross-resource since it can invalidate objects already persisted, cause ordering issues, etc? Or is that just an Istio practice not a general Kubernetes practice?

robscott: Thanks for catching that @hbagdi! I'll change the wording to include "webhook". I completely agree that any kind of cross resource validation is less than ideal and generally something that should be avoided. Unfortunately I haven't been able to come up with a better alternative here. I think this is the least bad option available, but I'm open to other options.

This PR intentionally changes the behavior of this field to restrict creation of Gateways, not updates. It intentionally will not invalidate existing Gateways if this field changes. Before this PR, changes to this field could have quite destructive results by invalidating all Gateways of a class. Essentially this shifts the burden of cross resource validation from controllers to a webhook and moves it from validation required on every Gateway or GatewayClass change to something that happens once at Gateway creation.

I think the distinction of this validation only applying on create and the expectation that this field on GatewayClass will not change very frequently makes the cross resource validation an acceptable compromise here. Unfortunately I don't know of any way to accomplish this with anything other than OPA which would end up having similar issues.

howardjohn: Should we reach out to some of the API machinery folks to see if there are some recommended practices here? The Istio community has already come the conclusion we will never support cross resource validation webhooks, so it seems unlikely we can support this part of the spec (which is fine if that's what the project wants, we are just one implementor).

I guess what I am trying to get at is cross resource validation a "bad practice" or a "really really bad idea". If it's the former and there are no alternatives it seems reasonable. If it's the latter, which is my personal opinion, then even if there are not alternatives we still should not do it.

robscott: You're completely right, asked in Slack and will share any answers I get here. If I don't get anything I'll reach out to someone directly tomorrow.

docs-src/security-model.md
```diff
@@ -72,14 +72,12 @@ readable by most roles, so instead we'll focus on write access for this model.
...
 
-* **allowedGatewayNamespaceSelector**: This field is a selector of namespaces
-  that Gateways can use this GatewayClass from. This is a standard Kubernetes
-  LabelSelector, a label query over a set of resources. The result of
-  matchLabels and matchExpressions are ANDed. Controllers must not support
-  Gateways in namespaces outside this selector. An empty selector (default)
-  indicates that Gateways can use this GatewayClass from any namespace. This
-  field is intentionally not a pointer because the nil behavior (no namespaces)
-  is undesirable here.
+* **allowedGatewayNamespaceSelector**: AllowedGatewayNamespaceSelector is a
```
hbagdi, line 75: Out of date. This should be `allowedGatewayNamespaces`.

docs-src/security-model.md
```diff
@@ -72,14 +72,12 @@ readable by most roles, so instead we'll focus on write access for this model.
...
+  selector of namespaces that Gateways of this class can be created in. Gateway
+  validation will prevent Gateways specifying this class from being created
```
hbagdi, line 77: Again, specifying  validation webhook would be more clearer here as our users do often confuse validation with openapi schema validation.

robscott:
Thanks for the great feedback on this one @hbagdi and @howardjohn. I asked about cross-resource validation in Slack (sig-auth). Although there was no clear yes/no, there were some good reasons to avoid it, including:

> admission that checks other objects to try to enforce uniqueness of an attribute is always racy with concurrent requests
> instead of synchronous protection like that, I'd encourage having good reconciliation and status reporting of conflicts which can handle all cases

With that background, @bowei suggested a bit of an intermediate approach. We rely on implementations to ensure that the Gateways they are supporting are in valid namespaces. Importantly, we only require that this validation occur for new Gateways as they are being provisioned - it does not _have_ to be enforced for existing Gateways. 

Not every implementation will have sufficient state to make this distinction though, and if they don't they should deprovision the Gateway. The theory here is that implementations that have a heavy cost of provisioning will also be most likely to have some kind of way to differentiate between Gateways that have already been provisioned and Gateways that have not been. If implementations are unable to make this distinction, it's likely very quick for them to reprovision gateways and thus invalidating them immediately is not quite as bad a thing to do.

I'll be the first to say that I don't love any of the approaches here so far, but this may be the least bad compromise available. Also interested in feedback from others like @jpeach or @danehans if they have thoughts.

apis/v1alpha1/gatewayclass_types.go
danehans: @robscott any updates regarding alternatives to cross resource validation?

robscott: I addressed this in my [comment below](https://github.com/kubernetes-sigs/service-apis/pull/263#issuecomment-690821018) but should have followed up here as well. I [asked in Slack](https://kubernetes.slack.com/archives/C0EN96KUY/p1599699017096700) and the general theme of the answer was not that we can't but that we probably shouldn't. Here are a couple of the tips from Jordan:

>admission that checks other objects to try to enforce uniqueness of an attribute is always racy with concurrent requests

> instead of synchronous protection like that, I'd encourage having good reconciliation and status reporting of conflicts which can handle all cases

With that in mind, I've [proposed yet another variation](https://github.com/kubernetes-sigs/service-apis/pull/263#issuecomment-690821018) below that is a bit of a hybrid approach.

docs-src/security-model.md
danehans: +1

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -63,23 +63,25 @@ type GatewayClassSpec struct {
 	Controller string `json:"controller"`
 
 	// AllowedGatewayNamespaceSelector is a selector of namespaces that Gateways
```
danehans, line 65: s/AllowedGatewayNamespaceSelector/AllowedGatewayNamespaces/

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -63,23 +63,25 @@ type GatewayClassSpec struct {
...
-	// can use this GatewayClass from. This is a standard Kubernetes
-	// LabelSelector. Controllers must not support Gateways in namespaces
-	// outside this selector.
+	// of this class can be created in. Implementations must not support
+	// Gateways when they are created in namespaces prohibitied by this field.
```
danehans, line 67: s/prohibitied/not specified/

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -63,23 +63,25 @@ type GatewayClassSpec struct {
...
 	//
-	// An empty selector (default) indicates that Gateways can use this
-	// GatewayClass from any namespace.
+	// This selector may be changed in a way that invalidates existing Gateways.
+	// Implementations may choose to immediately terminate related
+	// infrastructure. If implementations can differentiate between provisioned
+	// and pending Gateways, they may choose to continue to support any Gateways
+	// that were provisioned before this field changed. This differentiation can
+	// not be based on the Gateway status or other fields that users can
```
danehans, line 74: @robscott why can't the differentiation between provisioned and pending be based on gateway status?

robscott: It's a security issue. This field has no value if you can always bypass it by manually modifying the Gateway status.

hbagdi:
This is horribly ugly but I can live with this given other options.
We can also revisit after v1alpha1 with feedback from a wider group.

Only one minor nit:

> Implementations must not support
Gateways when they are created in namespaces prohibitied by this field.

Is it okay to change must to should? Depending on how one implements in-cluster proxies for this API, for in-cluster proxies, barring status, there might be no data to decide if a Gateway is new (and thus not support) or if it is an existing one or a new one.

docs-src/security-model.md
robscott: Updated, thanks!

docs-src/security-model.md
robscott: Updated these docs to reflect the new proposal

apis/v1alpha1/gatewayclass_types.go

apis/v1alpha1/gatewayclass_types.go

apis/v1alpha1/gatewayclass_types.go
robscott: Added some clarification here, thanks!

robscott:
OK, yet another variation here... As I was typing a response to @hbagdi's last question I think I came up with a not entirely terrible set of options. This may not actually be too difficult for every implementation to support since RBAC grants access to status subresources separately. (You could allow a user to create/update a Gateway but not modify its status). Here are the two options I'm considering:

1. Rely on the existing `Gateway.Status` fields to determine if a Gateway has been provisioned, communicate that RBAC status write access should be granted sparingly.

2. Introduce a new `ProvisionedGateways` list on `GatewayClass.Status` to track all Gateways that have currently been provisioned using the GatewayClass.

Although 2 introduces more overhead, it may also provide more value as an easy way to see where a GatewayClass is actively being used. A significant downside of 1 is that Gateways are a namespaced resource. This means that any admins within a namespace would be able to modify Gateway status and bypass this validation. Since this is intended as a way to prevent namespaces from being able to use a GatewayClass, this is a significant downside. On the other hand, GatewayClass is a cluster-scoped resource, it would be much more difficult for users to gain access to modify GatewayClass status. For those reasons, I'm leaning towards the first approach and have updated the PR to reflect that.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,24 +62,18 @@ type GatewayClassSpec struct {
...
+	// Gateways have already been provisioned when this field changes, they will
+	// continue to be considered valid and supported. For the purposes of this
+	// field, a Gateway must only be considered provisioned if it appears in the
+	// ProvisionedGateways list in GatewayClassStatus.
 	//
 	// Support: Core
 	//
 	// +optional
-	AllowedGatewayNamespaceSelector metav1.LabelSelector `json:"allowedGatewayNamespaceSelector"`
+	AllowedGatewayNamespaces metav1.LabelSelector `json:"allowedGatewayNamespaces"`
```
danehans, line 76: Add `omitempty` to json tag if this is optional.

danehans: @robscott since `AllowedGatewayNamespaces` is optional, is the default behavior to allow all/no namespaces?

danehans:
@robscott I'm in favor of option 2 in https://github.com/kubernetes-sigs/service-apis/pull/263#issuecomment-693196822. The benefits are worth the additional overhead IMO. Other than https://github.com/kubernetes-sigs/service-apis/pull/263#pullrequestreview-491702436, LGTM.

/cc @Miciah

docs-src/security-model.md
```diff
@@ -72,14 +72,14 @@ readable by most roles, so instead we'll focus on write access for this model.
...
-  field is intentionally not a pointer because the nil behavior (no namespaces)
-  is undesirable here.
+* **allowedGatewayNamespaces**: AllowedGatewayNamespaces is a selector of
+  namespaces that Gateways of this class can be created in. This selector may be
+  changed in a way that invalidates existing Gateways. Implementations may
+  choose to immediately terminate related infrastructure. If implementations can
+  differentiate between provisioned and pending Gateways, they may choose to
+  continue to support any Gateways that were provisioned before this field
+  changed. To ensure this field is not bypassed, this differentiation should not
+  be based on the Gateway status or other fields that users can manipulate.
```
Miciah, line 82: It would help to state on what this differentiation _should_ be based, namely the _`GatewayClass`_ status (per the godoc for `AllowedGatewayNamespaces`: "a `Gateway` must only be considered provisioned if it appears in the `ProvisionedGateways` list in `GatewayClassStatus`").

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,24 +62,18 @@ type GatewayClassSpec struct {
...
-	// may add a new "ForbiddenNamespaceForClass" condition to the Gateway
-	// status. Adding this condition is considered optional since not all
-	// controllers will have access to all namespaces.
+	// AllowedGatewayNamespaces is a selector of namespaces that Gateways of
+	// this class can be created in. Implementations must not support Gateways
+	// when they are created in namespaces not specified by this field. If
+	// Gateways have already been provisioned when this field changes, they will
+	// continue to be considered valid and supported. For the purposes of this
+	// field, a Gateway must only be considered provisioned if it appears in the
+	// ProvisionedGateways list in GatewayClassStatus.
```
Miciah, line 71: Namespaced-based access control is weird in the Kubernetes context (and it seems like I'm not alone in this opinion).  It feels like a re-invention of RBAC, which brings risk of increased confusion and mistakes.  So here's a possible alternative:  Instead of specifying namespaces that may use the `GatewayClass`, could we use RBAC to define a role that grants verb `use` to resource `gatewayclasses` with the `GatewayClass`'s name and bind that role to service accounts in namespaces that should be able to use the `GatewayClass`?

robscott: That's a really creative way to use RBAC. Of course we couldn't grand this ability directly to users since controllers would have no idea which user was responsible for the Gateway creation. I think if we could find a way to meaningfully grant `use` directly to users this would be a really compelling option, but I'm hesitant to introduce this requirement for service accounts in a namespace. If I'm understanding how this approach would work, it would require:

- Users/Controllers to create ServiceAccounts in each namespace that needed to use a GatewayClass
- Cluster admins to create custom ClusterRoles to grant access to each individual GatewayClass
- Cluster admins to create custom ClusterRoleBindings to bind to each of those Service Accounts (in each namespace) to each ClusterRole/GatewayClass they should have access to
- Controllers to have read access to and watch ServiceAccounts in all namespaces as well as ClusterRoles and ClusterRoleBindings
- Controllers to be able to parse/understand RBAC roles + react to changes in them
- Users to understand their RBAC config enough to know where a GatewayClass could be used

There's a decent chance I'm misunderstand the proposal and/or complexity of the approach, definitely let me know if I'm off base here.

I would definitely prefer a more native approach here but unfortunately I haven't thought of anything that is easier to use or understand than my current proposal. I don't think it's a complete reinvention of RBAC though. Users still need to be able to create a Gateway in a namespace for any of this to apply. This is just an extra level of granularity on top of that.

Although the API supports it, users won't have to use it. By default, all GatewayClasses will be available for use in all namespaces. If they already have something like OPA in their cluster, that may be a better way for certain organizations to enforce this kind of policy. They could also experiment with their own validating webhook for custom validation if they want to. 

With that said, I think this approach provides a reasonable compromise in terms of usability and simplicity and I hope that users will actually find it helpful. With that said, this is still just an alpha - we can always pull this out or change it if a better alternative comes around.

apis/v1alpha1/gatewayclass_types.go
robscott: I think I should update this to be an explicit default of `{}`. I'd thought that's what optional without omitempty would do, but even if it does, that's not as clear as actually setting a default value here. The intent is for the default to allow a GatewayClass to be used everywhere - I imagine that will be the most common use case.

robscott: Actually, I thought about it more and opted to make it empty and just define the unspecified state like you recommended.

docs-src/security-model.md
robscott: Thanks for catching this, updated this to match the godocs.

apis/v1alpha1/gatewayclass_types.go
Miciah: > Of course we couldn't grand this ability directly to users since controllers would have no idea which user was responsible for the Gateway creation. [...] If I'm understanding how this approach would work, it would require:
> 
> * Users/Controllers to create ServiceAccounts in each namespace that needed to use a GatewayClass

You could use the `default` service account of the namespace.

> * Cluster admins to create custom ClusterRoles to grant access to each individual GatewayClass
> * Cluster admins to create custom ClusterRoleBindings to bind to each of those Service Accounts (in each namespace) to each ClusterRole/GatewayClass they should have access to

If you wanted to restrict access to specific gatewayclasses, yes.

> * Controllers to have read access to and watch ServiceAccounts in all namespaces as well as ClusterRoles and ClusterRoleBindings
> * Controllers to be able to parse/understand RBAC roles + react to changes in them

An admission controller could perform the SAR.  I think it would be fine to enforce RBAC only on Gateway creation (similar to what you're proposing).

> * Users to understand their RBAC config enough to know where a GatewayClass could be used

Better to need to learn 1 RBAC system than 2.  \*grin\*

(Also, why isn't `kubectl who-can` a standard command?)

> There's a decent chance I'm misunderstand the proposal and/or complexity of the approach, definitely let me know if I'm off base here.

No worries; there's a decent chance I'm misunderstanding my own proposal too.

> I would definitely prefer a more native approach here but unfortunately I haven't thought of anything that is easier to use or understand than my current proposal. I don't think it's a complete reinvention of RBAC though. Users still need to be able to create a Gateway in a namespace for any of this to apply. This is just an extra level of granularity on top of that.
> 
> Although the API supports it, users won't have to use it. By default, all GatewayClasses will be available for use in all namespaces. If they already have something like OPA in their cluster, that may be a better way for certain organizations to enforce this kind of policy. They could also experiment with their own validating webhook for custom validation if they want to.

The default RBAC should allow use of all GatewayClasses, something roughly like the following:
<details>

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: allow-all-gatewayclasses
rules:
- apiGroups:
  - networking.x-k8s.io
  resources:
  - gatewayclasses
  verbs:
  - use
  # name unspecified to allow use of all names
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: allow-all-gatewayclasses
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: allow-all-gatewayclasses
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:authenticated
```
</details>
For clusters for which this suffices, no futzing with RBAC is needed.  If the cluster administrator needs to specify finer-grained policy, ey can delete the subject from the above binding and define finer-grained policies to restrict access to certain gatewayclasses to certain service accounts:
<details>

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: allow-boring-gatewayclasses
rules:
- apiGroups:
  - networking.x-k8s.io
  resources:
  - gatewayclasses
  verbs:
  - use
  resourceNames:
  - boring-gatewayclass-1
  - boring-gatewayclass-2
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: allow-fancy-gatewayclasses
rules:
- apiGroups:
  - networking.x-k8s.io
  resources:
  - gatewayclasses
  verbs:
  - use
  resourceNames:
  - fancy-gatewayclass
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: allow-fancy-gatewayclasses
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: allow-fancy-gatewayclasses
subjects:
- kind: ServiceAccount
  name: default
  namespace: fancy-project
```
</details>
Does that make sense?

robscott: Thanks for the detailed explanation, this is a very interesting idea.

> Better to need to learn 1 RBAC system than 2.

😆 this is a great point - also completely agree that kubectl who-can would be an awesome standard feature of kubectl instead of just a plugin.

Admission controller + `SubjectAccessReview` like you're suggesting would make this something controllers would generally not have to worry about. As a brief sidebar, Kubernetes RBAC is effectively a form of cross-resource validation... Not particularly different than the original webhook idea here (although admittedly closer to standard).

I generally like everything in the proposal until it gets to the subject of the ClusterRoleBinding. I imagine a very common mistake users would make would be to assign these roles to specific users/groups. If we could make that work, I think this approach would be nearly perfect. Unfortunately I can't think of a way to make that work, and so it feels like we'd be stuck with a relatively unintuitive/non-standard subject to bind to.  

I just can't get past the idea that anyone familiar with RBAC is going to want to grant the ability to `use` GatewayClasses to specific users or groups and we simply can't offer that. Maybe worse than that, we also can't prevent it so users will be able to grant this ability to users and groups, may even try something like `kubectl who-can` or `kubectl auth can-i`, see that their user should clearly be able to use the GatewayClass, but then get denied by the admission controller. I think offering something that feels so similar to standard RBAC but has this one significant and confusing difference may result in user frustration as their RBAC config for GatewayClasses just doesn't work.

bowei: Namespaces are an access boundary in some general sense, it has been used for network policy, so it's not quite as big of a stretch. It does have an advantage of being easy to understand :-)

I would like to investigate the RBAC proposal but maybe not for this iteration of the API. It feels like the proposal would need to be properly fleshed out.

robscott: @Miciah Although I'm interested in exploring RBAC further I think it may be a bit too much to figure out between now and an alpha launch. Would it be reasonable to revisit this post v1alpha1?

Miciah: Yeah, we could do that.  What would be helpful in further exploring the idea?  Further example RBAC definitions?  A proof-of-concept admission controller?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,24 +62,24 @@ type GatewayClassSpec struct {
...
+	// that it has been provisioned but the Gateway does not appear in the
+	// ProvisionedGateways list on GatewayClass it must not be supported.
 	//
-	// When a Gateway attempts to use this class from a namespace that is not
-	// allowed by this selector, the controller implementing the GatewayClass
-	// may add a new "ForbiddenNamespaceForClass" condition to the Gateway
-	// status. Adding this condition is considered optional since not all
-	// controllers will have access to all namespaces.
+	// When this field is unspecified or an empty selector, Gateways will be
+	// able to use this GatewayClass in any namespace.
```
hbagdi, line 77:
```suggestion
	// able to use this GatewayClass in any namespace (default).
```

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -168,6 +168,12 @@ type GatewayClassStatus struct {
 	// +optional
 	// +kubebuilder:default={{type: "InvalidParameters", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ProvisionedGateways is a list of Gateways that have been provisioned
+	// using this class. Implementations must add any Gateways of this class to
+	// this list once they have been provisioned and remove Gateways as soon as
+	// they are deleted or deprovisioned.
+	ProvisionedGateways []GatewayObjectReference `json:"provisionedGateways"`
```
hbagdi, line 176: Questions: Is is possible for different controllers to race here?

hbagdi (commented):
LGTM

apis/v1alpha1/gatewayclass_types.go
robscott: That's a great question - I think there are a few things that could be helpful here:

1. Any examples of this pattern being used elsewhere. Are there any other common use cases that involve binding to a namespace service account instead of users and groups directly?  
2. Some feedback from sig-auth to ensure this kind of approach would be within the intended scope of RBAC.
3. Further exploration of any ways we could accomplish this by binding the capability directly to users or groups. 
4. For bonus points: some feedback from end users on the potential UX of an approach like this.

I think there's a lot of value in this RBAC approach, especially if we can figure out 3.

apis/v1alpha1/gatewayclass_types.go
robscott: I don't think so. Each GatewayClass should only be managed by a single controller.

apis/v1alpha1/gatewayclass_types.go
robscott: Thanks - updated the wording here to be more clear about the default behavior.

PR #253, opened 2020-08-06
Hoist HTTP match types into type-specific structs.

jpeach:
Update the route match struct to hoist the path and header match
specifications into their own structs. This improves API clarity and
consistency by using the same matching syntax for paths and headers.

Update the matching defaults at various levels so that there is always a
default prefix path match on "/". This reduces the amount of YAML needed
for simple cases and removes the validation burden from controllers.

Signed-off-by: James Peach <jpeach@vmware.com>

PR #264, opened 2020-08-13
httproute: validation for matches

hbagdi:
This commit adds a validation for minimum 1 item for the `matches` array
and also adds in comment for configuring the default route.

This is a follow up on https://github.com/kubernetes-sigs/service-apis/pull/241#issuecomment-673139234.

### DIFF
--- a/apis/v1alpha1/generated.proto
+++ b/apis/v1alpha1/generated.proto
@@ -718,7 +718,10 @@ message HTTPRouteRule {
   // - path prefix of `/v2/foo`
   // Please see doc for HTTPRouteMatch on how to specify multiple
   // match conditions that should be ANDed together.
-  // +optional
+  //
+  // Default catch-all route can be configured using path `/`.
+  //
+  // +kubebuilder:validation:MinItems=1
   repeated HTTPRouteMatch matches = 1;
 
   // Filter defines what filters are applied to the request.
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -96,7 +96,10 @@ type HTTPRouteRule struct {
 	// - path prefix of `/v2/foo`
 	// Please see doc for HTTPRouteMatch on how to specify multiple
 	// match conditions that should be ANDed together.
-	// +optional
+	//
+	// Default catch-all route can be configured using path `/`.
+	//
+	// +kubebuilder:validation:MinItems=1
 	Matches []HTTPRouteMatch `json:"matches" protobuf:"bytes,1,rep,name=matches"`
 	// Filter defines what filters are applied to the request.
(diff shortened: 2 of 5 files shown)


hbagdi:
@bowei @robscott @jpeach I added in validation rather than documenting that empty matches is invalid.

jpeach:
> @bowei @robscott @jpeach I added in validation rather than documenting that empty matches is invalid.

In office hours discussion of #253, there was a proposal to default the matches array to a prefix path match of the "/". I was planning to include that when I base it today.

hbagdi:
> In office hours discussion of #253, there was a proposal to default the matches array to a prefix path match of the "/". I was planning to include that when I base it today.

Wasn't that proposal deemed problematic as defaulting an array seemed like it could create problems. I might have misunderstood @bowei's point here.

Another reason to not add a default `/` prefix match is it makes it easier for collisions to happen when multiple routes for a listener have the default match pointing to potentially different services.

jpeach:
> > In office hours discussion of #253, there was a proposal to default the matches array to a prefix path match of the "/". I was planning to include that when I base it today.
> 
> 
> 
> Wasn't that proposal deemed problematic as defaulting an array seemed like it could create problems. I might have misunderstood @bowei's point here.
> 
> 
> 
> Another reason to not add a default `/` prefix match is it makes it easier for collisions to happen when multiple routes for a listener have the default match pointing to potentially different services.

It doesn't seem to me that defaulting the matches array introduces any new problems around conflicts, though I do agree that it could make it easier to accidentally conflict.

robscott:
I think a default of matching all paths makes sense and provides the least friction for simple use cases. @hbagdi raises a good point about the increased potential for conflicts, but the only alternative I can think of is a route rule that doesn't match anything which also seems confusing. We'll already have to solve conflicts between equivalent route rules at different levels, so although a default `/` could make that more common, it represents logic we'd already have to have. I know we'd also been concerned that defaulting an array could get messy, but as long as it works reliably it seems like a reasonable option to me.

jpeach:
> I think a default of matching all paths makes sense and provides the least friction for simple use cases. @hbagdi raises a good point about the increased potential for conflicts, but the only alternative I can think of is a route rule that doesn't match anything which also seems confusing. 

One thing I like about the defaulting is that it shows up in the returned resource yaml so it's clear what's going on.

> We'll already have to solve conflicts between equivalent route rules at different levels, so although a default `/` could make that more common, it represents logic we'd already have to have. I know we'd also been concerned that defaulting an array could get messy, but as long as it works reliably it seems like a reasonable option to me.

There's an example here that works like I expected it to.

hbagdi:
In that case, let's solve this problem in the conflict resolution work. Closing this.

PR #267, opened 2020-08-20
Fix protobuf build ordering.

jpeach:
Generating protobuf bindings needs to run after generating deep copy code
since it needs valid Go code to process. It's also the slowest generator,
so it deserves to go last.

Signed-off-by: James Peach <jpeach@vmware.com>


robscott:
Thanks!

/lgtm

Issue #196, opened 2020-05-19
Service-level configuration


hbagdi:
## Current Status

We are collecting issues and use cases that should be associated with core.Service.
Once we have a good understanding of problems, we will start discussing potential solutions, ultimately ending in a KEP.  Please feel free to add your comments to this thread and we will try to incorporate those here.

## Summary

There are service-level configuration properties that cannot be specified anywhere in k8s or service-apis resources currently.
The community seems to have a consensus that `core.Service` is an overloaded resource with features concerning different areas and the more fields should **not** be added to further bloat the Service abstraction.

While these features are laid out with Ingress or Gateway in mind, some of these configurations are also valid in other areas such as clients of a Service running within the cluster perimeter or outside, Service Mesh deployments, etc.

### Load balancing

Some examples of load balancing features are (not exhaustive):

- Algorithm to use for load-balancing (L4/L7) and their properties
- Retry behavior: If a request fails, it should be retried(L4/L7) a number of times, with delays, and to different/same endpoints
- Cluster or region aware load-balancing of traffic. This also ties with Multicluster Service effort

`core.Service` has some aspects of load balancing using fields like `SessionAffinity`, and `ExternalTrafficPolicy`. The existing fields are (intentionally?) not exhaustive but also don’t provide any extension mechanisms.

Related issues:

- https://github.com/kubernetes-sigs/service-apis/issues/98
- https://github.com/kubernetes-sigs/service-apis/pull/184

### Traffic management

Service upgrades are performed in a gradual way to avoid surprises. Canarying and Mirroring traffic are some of the most common ways this is performed. Ingress and mesh vendors provide such features using CRDs. Service-apis project has also seen a lot of interest in this area. While this spans multiple core.Service resources, it might be worth to explore how an end-user should think of Services and endpoints when it comes to upgrades.

Related issues:
- https://github.com/kubernetes-sigs/service-apis/issues/58

### Health checking

While kubernetes performs health checking of pods that are associated with a Service, it is common for proxies and Load Balancer to perform health checking on their part as well.
There have been some implementations which derive the health checking info from the liveliness or readiness probes but those seem incorrect as the reasons/concerns for kubelet and the proxy are different.

Feature requests:
- Active health-checking (L4/L7)
- Passive health-checking a.k.a. circuit-breaking
- Timeout policies (overlaps with #L4-connection)

Related issues:

- https://github.com/kubernetes-sigs/service-apis/issues/97
[...]

PR #268, opened 2020-08-20
Adding BackendPolicy for Service-level config

robscott:
This is roughly based on the [Service-level config doc](https://docs.google.com/document/d/1Kz2X7zKfaSGW9YTlzqFeFuTCj5uDxdB9D363nMkB6xk/edit) we've been discussing. It also incorporates the excellent work from @danehans in https://github.com/kubernetes-sigs/service-apis/pull/184. This PR still needs examples and documentation but I'm opening it early so we can get some more feedback on the concept. This is meant to be a starting point, I think there's still lots of room for improvement here.

/cc @danehans @hbagdi @jpeach


apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,232 @@
...
+	// Retries define how retries should be handled for these backends.
+	//
+	// Support: Extended
+	//
+	// +optional
+	Retries []RetryPolicy `json:"retries" protobuf:"bytes,3,rep,name=retries"`
+
+	// TLS is the TLS configuration for these backends.
+	//
+	// Support: Core
```
yiyangy, line 72: Shall TLS be extended feature?

danehans:
@robscott this is a PR for a `BackendPolicy`, i.e. a policy for a backend. I think the docs should be updated to define what a "backend" is.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,232 @@
...
+// BackendPolicySpec defines desired policy for a Backend.
+type BackendPolicySpec struct {
+	// BackendRefs define which backends this policy should be applied to. If
+	// more than one BackendPolicy targets the same Backend, precedence will be
+	// given to the oldest BackendPolicy.
+	//
+	// Support: Core
+	//
+	// +optional
+	BackendRefs []BackendRef `json:"backendRefs" protobuf:"bytes,1,opt,name=backendRefs"`
```
howardjohn, line 54: I am not sure a list of references is a very common case. I guess it doesn't hurt to have a list here vs copying the spec over. But a "global defaults" is a common request we get for these type of options

hbagdi:
This might of interest to Knative folks @mattmoor @tcnghia @ZhiminXiang.

apis/v1alpha1/backendpolicy_types.go
hbagdi: TLS on upstream connection is fairly common.
Do you have concerns about specifics of TLS in different implementations?

robscott:
I've pushed an update that dramatically reduces the scope of this first version of TLSPolicy. This now only includes TLS Config. This is one relatively simple thing that we know we'll want to be associated with a backend whereas the other fields I'd initially added had significantly more potential for variation among implementations and could also potentially fit on routes instead.

apis/v1alpha1/backendpolicy_types.go
robscott: Now that the scope of this PR has been reduced to TLS config a global default might not be as helpful. Definitely agree that it could be helpful to have global defaults for some potential future additions to BackendPolicy. Defining these defaults could also potentially be accomplished with defaults for GatewayClass params. 

I agree that multiple BackendRefs are not a very common use case, but it seemed simple enough to support and added some extra flexibility for users.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// TLS is the TLS configuration for these backends.
+	//
+	// Support: Core
+	//
+	// +optional
+	TLS *TLSConfig `json:"tls,omitempty" protobuf:"bytes,4,opt,name=tls"`
```
hbagdi, line 63: TLSConfig struct is currently being used for downstream TLS settings for the Gateway.

Using the same struct for Upstream can be confusing, such as, what does CertificateRef mean in this context? Is it the client certificate-key pair that the Gateway should use to authenticate with the upstream Service or is it the CACertificate to use to verify upstream's TLS certificate?

We should ask what settings of upstream TLS we want to make configurable and then create a new struct (and probably rename the (`TLSConfig` struct).

Another question: How/Where is the protocol the Gateway should use to talk to upstream defined?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Port is the port of the referent. If unspecified, this policy applies to
+	// all ports on the backend.
+	// +optional
+	Port *int32 `json:"port,omitempty" protobuf:"varint,4,opt,name=port"`
+}
+
+// RetryPolicy describes a policy for retrying a connection until the connection
+// is either refused or times out.
+type RetryPolicy struct {
```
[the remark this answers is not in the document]
robscott: Thanks for catching this!

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,232 @@
...
+
+// BackendPolicyList contains a list of BackendPolicy
+type BackendPolicyList struct {
+	metav1.TypeMeta `json:",inline" protobuf:"bytes,1,opt,name=typeMeta"`
+	metav1.ListMeta `json:"metadata,omitempty" protobuf:"bytes,2,opt,name=metadata"`
+	Items           []BackendPolicy `json:"items" protobuf:"bytes,3,rep,name=items"`
+}
+
+// BackendPolicySpec defines desired policy for a Backend.
+type BackendPolicySpec struct {
```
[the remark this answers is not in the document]
robscott: Updated to add that statement, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: I think it is probably better to make this extended. I know not every implementation supports this and those that do have varying levels of support for TLS config.

apis/v1alpha1/backendpolicy_types.go
robscott: Thanks for catching this, it really didn't make much sense to share the type here. I've split this out into a new BackendTLSConfig that supports references for client cert, key, and trusted CA. Those seemed to be fairly commonly supported but interested in feedback here and/or other fields that should be included.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// Support: Extended
+	//
+	// +optional
+	TrustedCertificateRef *CertificateObjectReference `json:"trustedCertificateRef,omitempty" protobuf:"bytes,3,rep,name=trustedCertificateRef"`
+
+	// Options are a list of key/value pairs to give extended options to the
+	// provider.
+	//
+	// Support: Implementation-specific.
+	Options map[string]string `json:"options" protobuf:"bytes,4,rep,name=options"`
```
hbagdi, line 160: We need to mark this as `+optional`.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// that may be used to connect to these backends. If an entry in this list
+	// omits or specifies the empty string for both the group and the resource,
+	// the resource defaults to "secrets". An implementation may support other
+	// resources (for example, resource "mycertificates" in group
+	// "networking.acme.io").
+	//
+	// Support: Extended
+	//
+	// +optional
+	ClientCertificateKeyRef *CertificateObjectReference `json:"clientCertificateKeyRef,omitempty" protobuf:"bytes,2,rep,name=clientCertificateKeyRef"`
```
hbagdi, line 145: Having a separate object Reference for certificate and key seems wrong.

Client certificate work similar to Server certificates, meaning, a user can put the key and certificate in the same secret (or whatever CRD) and then refer both of them at the same time.

We should follow the same pattern we have in TLSConfig here.

hbagdi: Kubernetes secrets have a tls secret type which can be reused here.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// TrustedCertificateRef is a reference to a certificate for a trusted
+	// certificate authority. This may be useful for backends using self-signed
+	// certificates.
+	//
+	// Support: Extended
+	//
+	// +optional
+	TrustedCertificateRef *CertificateObjectReference `json:"trustedCertificateRef,omitempty" protobuf:"bytes,3,rep,name=trustedCertificateRef"`
```
hbagdi, line 154: This should be an array as there can be multiple CA certs, especially during certificate rotation periods.

Is `CACertifcateRefs` a better name here (which indicates that the user needs to put in a reference to a CA certificate?

Since this is an extended feature, we should have a standardized way (key name, content formatting etc) of defining certificates using secrets or config maps for this feature.

We don't need to do this for client cert-key because we can directly rely on TLS secrets.

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: I initially transitioned to `CACertificateRefs` but in the last meeting I think @jpeach mentioned that we could include a CA Bundle inside a TLS secret. That is a configuration currently [supported by nginx-ingress](https://kubernetes.github.io/ingress-nginx/examples/auth/client-certs/), though does not seem to be as broadly supported or used by other implementations. 

A potential benefit of having everything in the same resource is that there are less Secrets to maintain. A downside is that some use cases may involve reusing a CA bundle but not client certs, something that would be a bit more difficult with everything in a single Secret.

Allowing `ca.crt` to be specified at this level does not prevent us from adding support for a separate field to reference a CA bundle in the future. Given that, I think this simple approach is likely the best starting point.

danehans:
protobuf tags should be removed from fields since https://github.com/kubernetes-sigs/service-apis/pull/283 merged.

danehans:
Can you add add a `BackendPolicy` example in /example?

apis/v1alpha1/backendpolicy_types.go
hbagdi: I don't agree here.

First, CA certificates are usually managed separately from client certificates. They have different life-cycles and clubbing them together often leads to unnecessary duplication of CA certificates.

Further, the doc states:

> 	// If a Secret is referenced, it must be of type "kubernetes.io/tls" and
	// contain tls.crt and tls.key data fields that contain the certificate and
	// private key to use for TLS. It may optionally include a certificate
	// authority bundle under a ca.crt data field



Is that even valid?
The `kubectl` document says `kubectl create secret tls tls-secret --cert=path/to/tls.cert --key=path/to/tls.key` is how you create a tls secret and it doesn't allow you to specify any additional field. You can edit a secret or manually (using stringData) create a secret of type `kubernetes.io/tls` that allows you to include `ca.crt`.

The [nginx document](https://kubernetes.github.io/ingress-nginx/examples/auth/client-certs/) asks you to create a generic secret instead of a TLS secret:
```
kubectl create secret generic ca-secret --from-file=tls.crt=server.crt --from-file=tls.key=server.key --from-file=ca.crt=ca.crt
```

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,205 @@
...
+type BackendPolicyConditionType string
+
+const (
+	// ConditionNoSuchBackend indicates that one or more of the the specified
+	// Backends does not exist.
+	ConditionNoSuchBackend BackendPolicyConditionType = "NoSuchBackend"
+)
+
+// BackendPolicyCondition is an error status for a given BackendPolicy.
+type BackendPolicyCondition struct {
```
hbagdi, line 166: With #302 under way, can we update this to use metav1.Condition type?

apis/v1alpha1/backendpolicy_types.go
howardjohn: Just throwing some additional info our there, Istio supports the two options nginx-ingress supports (single secret with ca.crt or split secrets) and doesn't care if its `generic` or `kubernetes.io/tls` due to the issue mentioned ^ with `kubectl create`

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Split this back up into 2 refs, thanks for the feedback!

robscott:
@danehans Thanks for the review! I've removed the protobuf tags and added a BackendPolicy example.

examples/backendpolicy.yaml
```diff
@@ -0,0 +1,15 @@
+kind: BackendPolicy
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: my-app
+spec:
```
yiyangy, line 5: Shall a backendRefs be added to this example?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+// BackendPolicySpec defines desired policy for a backend.
+type BackendPolicySpec struct {
+	// BackendRefs define which backends this policy should be applied to. If
+	// more than one BackendPolicy targets the same backend, precedence will be
+	// given to the oldest BackendPolicy.
+	//
+	// Support: Core
+	//
+	// +optional
+	BackendRefs []BackendRef `json:"backendRefs"`
```
yiyangy, line 57: Just to clarify, an BackendPolicy object only apply to backends in the same namespace, correct?

hbagdi: A very important question. My understanding here is yes, namespace restrictions apply. Is that the case, Rob?

docs-src/concepts.md
```diff
@@ -104,6 +104,13 @@ Route objects define protocol-specific rules for mapping requests from a Gateway
 `HTTPRoute` and `TCPRoute` are currently the only defined Route objects. Additional protocol-specific Route
 objects may be added in the future.
 
+### BackendPolicy
+
+BackendPolicy provides a way to configure backends. For the purpose of this API,
+a backend is any resource that a route can forward traffic to. A common example
+of a backend is a Service. Configuration at this level is currently limited to
+TLS, but will expand in the future to support more advanced policies.
```
hbagdi, line 112: Rob, I think it is important to include what can go into BackendPolicy. What I mean to say is, we should communicate that configuration that relates to the Gateway-Upstream hop should go in here. Linking #196 here might be a good idea as well.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+type BackendPolicyList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BackendPolicy `json:"items"`
+}
+
+// BackendPolicySpec defines desired policy for a backend.
+type BackendPolicySpec struct {
+	// BackendRefs define which backends this policy should be applied to. If
+	// more than one BackendPolicy targets the same backend, precedence will be
```
hbagdi, line 51: > precedence will be

precedence must be

apis/v1alpha1/backendpolicy_types.go
hbagdi: > // BackendRef identifies an API object within a known namespace that defaults
// group to core and resource to services if unspecified

This is what we are looking for I think, @yiyangy.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Port is the port of the referent. If unspecified, this policy applies to
+	// all ports on the backend.
+	// +optional
+	Port *int32 `json:"port,omitempty"`
+}
+
+// BackendTLSConfig describes TLS configuration for a backend.
+type BackendTLSConfig struct {
+	// ClientCertificateRef is a reference to a TLS client certificate that may
```
hbagdi, line 122: s/certificate/certificate-key pair/

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+	//
+	// +optional
+	ClientCertificateRef *CertificateObjectReference `json:"clientCertificateRef,omitempty"`
+
+	// CertificateAuthorityRef is a reference to a resource that includes
+	// trusted CA certificates for the associated backends. If an entry in this
+	// list omits or specifies the empty string for both the group and the
+	// resource, the resource defaults to "secrets". An implementation may
+	// support other resources (for example, resource "mycertificates" in group
+	// "networking.acme.io").
```
hbagdi, line 143: If we are supporting secrets here, let's add in:
- certificates should be present inside `data["ca.crt"]` key
- they should be PEM encoded
- Multiple certificates can be specified (concat with a newline)

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+	// Support: Implementation-specific.
+	// +optional
+	Options map[string]string `json:"options,omitempty"`
+}
+
+// BackendPolicyStatus defines the observed state of BackendPolicy.
+type BackendPolicyStatus struct {
+	// Conditions describe the current conditions of the BackendPolicy.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
hbagdi, line 162: We can do this in future, but should we structure status subresource in a way that allows for reporting status at a per backendRef granularity?

apis/v1alpha1/backendpolicy_types.go
danehans: +1 that a BackendPolicy object should only apply to backends in the same namespace.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+// this API, a backend is defined as any resource that a route can forward
+// traffic to. A common example of a backend is a Service. Configuration that is
+// implementation specific may be represented with similar implementation
+// specific custom resources.
+type BackendPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   BackendPolicySpec   `json:"spec,omitempty"`
+	Status BackendPolicyStatus `json:"status,omitempty"`
```
danehans, line 36: xref'ing kubebuilder tag for status defaulting: https://github.com/kubernetes-sigs/service-apis/pull/324/files?file-filters%5B%5D=.go&file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.yaml#diff-2927d10c91558c2273bc3a8ea6464b85R34

apis/v1alpha1/backendpolicy_types.go
danehans: @robscott should `BackendRefs` be required instead of optional? What is the purpose of `BackendPolicySpec` without specifying `BackendRefs`?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+	//
+	// +optional
+	BackendRefs []BackendRef `json:"backendRefs"`
+
+	// TLS is the TLS configuration for these backends.
+	//
+	// Support: Extended
+	//
+	// +optional
+	TLS *BackendTLSConfig `json:"tls,omitempty"`
```
danehans, line 64: @robscott TLS configuration is currently supported by the `Gateway` API and [PR 281](https://github.com/kubernetes-sigs/service-apis/pull/281) proposes adding TLS configuration to the `HTTPRoute` API. Can you help me understand why TLS configuration is required in the `BackendPolicy` API too? If it's required, how does `BackendPolicy` work with the TLS configuration of the other two APIs? cc: @hbagdi

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// BackendPolicyStatus defines the observed state of BackendPolicy.
+type BackendPolicyStatus struct {
+	// Conditions describe the current conditions of the BackendPolicy.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// BackendPolicyConditionType is a type of condition associated with a
```
danehans, line 165: Should `BackendPolicy` conditions follow the Gateway condition pattern being established in https://github.com/kubernetes-sigs/service-apis/pull/324?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Name is the name of the referent.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Port is the port of the referent. If unspecified, this policy applies to
+	// all ports on the backend.
+	// +optional
+	Port *int32 `json:"port,omitempty"`
```
danehans, line 117: Should port min/max validation be specified?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// BackendTLSConfig describes TLS configuration for a backend.
+type BackendTLSConfig struct {
```
danehans, line 121: @robscott what is the reason for being able to specify an empty `BackendTLSConfig`? If no reason exists, should the client cert ref be required?

apis/v1alpha1/backendpolicy_types.go
hbagdi: Daneyon, should we resort to a per backend status instead?
https://github.com/kubernetes-sigs/service-apis/pull/268#discussion_r487252302

apis/v1alpha1/backendpolicy_types.go
hbagdi: TLS in Gateway and in #281 is about downstream TLS i.e. client <-> Gateway.
Backend policy is about upstream TLS i.e. Gateway <-> Upstream Service.

apis/v1alpha1/backendpolicy_types.go
robscott: That's an interesting idea. When would this config be invalid for one backend and not another?

apis/v1alpha1/backendpolicy_types.go
robscott: That's a good question. I think it made sense as an optional field on GatewayTLSConfig but I can't think of any reason it needs to be optional here, will update.

apis/v1alpha1/backendpolicy_types.go
robscott: I could be wrong here, but I think this may the wrong direction. When would a timeout be valid for one backend and not another? Maybe we need a way of indicating that a backendRef does not resolve to a resource? What might be more significant here is if more than one GatewayClass/controller is using a backendRef and a field on that backendRef is invalid for them. Each controller may need a way to indicate that separately. Not quite sure if we've found a great way to solve this yet, but it feels like a similar problem to Route status...

apis/v1alpha1/backendpolicy_types.go
robscott: We likely need a TLS page/section in the docs to describe all the different places it can be configured.

apis/v1alpha1/backendpolicy_types.go
robscott: Discussed a few more potential ways we might want to group conditions in another comment: https://github.com/kubernetes-sigs/service-apis/pull/268#discussion_r489191518

examples/backendpolicy.yaml
robscott: Added, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated to make this required and clarify that this is for same-namespace backends only - thanks for the feedback!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated to make this required, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: AppProtocol may be a per-backend issue. Problems could appear on route status instead?

hbagdi: Good point. I was thinking more about the happy path where user wants feedback on whether this BackendPolicy was successfully applied to which services.

hbagdi:
I'm not certain here but a thought: should this be named BackendNetworkPolicy? My (tiny) concern is that BackendPolicy might be a tad generic and we might be creating a kitchen sink.

docs-src/concepts.md
robscott: Good call - updated to make this more clear.

danehans:
> I'm not certain here but a thought: should this be named BackendNetworkPolicy? My (tiny) concern is that BackendPolicy might be a tad generic and we might be creating a kitchen sink.

@hbagdi I'm reluctant to use `Network` in the type's name due to the existing k8s `NetworkPolicy` type. I feel like the schema is modeling connection-level attributes. Maybe this is a nit, but if we want to narrow the scope of the name, what do you think about `BackendConnectionPolicy`?

apis/v1alpha1/backendpolicy_types.go
robscott: Since there are so many potential ways to slice this up, I think the simplest for now is to have per Backend conditions at this level and then place implementation specific status on Routes (which already have per-Gateway status). I've updated the comments to suggest that approach here.

apis/v1alpha1/backendpolicy_types.go
robscott: I don't think we can have meaningful defaults here since we'd have to have a new "unknown" default value for each Backend. Although that might be ideal, I don't think we can accomplish it with kubebuilder tags.

apis/v1alpha1/backendpolicy_types.go
robscott: Created https://github.com/kubernetes-sigs/service-apis/issues/331 to track this.

apis/v1alpha1/backendpolicy_types.go
robscott: Since there are so many potential ways to slice this up, I think the simplest for now is to have per Backend conditions at this level and then place implementation specific status on Routes (which already have per-Gateway status). I've updated the comments to suggest that approach here.

robscott:
@hbagdi @danehans I think everything has been resolved now except for naming, let me know if I've missed anything. As far as naming, I'm open to a less generic name than BackendPolicy but concerned about how painful it might become to `kubectl get` a long resource name. I agree that `BackendNetworkPolicy` makes me think of something more closely related to the `NetworkPolicy` resource. I think `BackendConnectionPolicy` feels like a very accurate name, it just may become annoying to interact with due to its length. I created a quick Slack poll to see what others think and/or if anyone has other ideas for a name here.

danehans:
xref [slack thread](https://kubernetes.slack.com/archives/CR0H13KGA/p1600751302007100) from https://github.com/kubernetes-sigs/service-apis/pull/268#issuecomment-696514015.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,182 @@
...
+
+// BackendPolicySpec defines desired policy for a backend.
+type BackendPolicySpec struct {
+	// BackendRefs define which backends this policy should be applied to. This
+	// policy can only apply to backends within the same namespace. If more than
+	// one BackendPolicy targets the same backend, precedence must be given to
+	// the oldest BackendPolicy.
+	//
+	// Support: Core
+	// +kubebuilder:validation:MaxItems=16
```
danehans, line 56: @robscott should this field also include `// +kubebuilder:validation:MinItems=1`?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,182 @@
...
+	//
+	// fooRef:
+	//   group: acme.io
+	//   resource: foos
+	//   name: myfoo
+	//
+	// +optional
+	// +kubebuilder:default=core
+	// +kubebuilder:validation:MaxLength=253
+	Group string `json:"group"`
```
danehans, line 87: Since defaulting is being used, `+optional` can be removed and add `omitempty` json tag.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,182 @@
...
+	//
+	// fooRef:
+	//   group: acme.io
+	//   resource: foos
+	//   name: myfoo
+	//
+	// +optional
+	// +kubebuilder:default=services
+	// +kubebuilder:validation:MaxLength=253
+	Resource string `json:"resource"`
```
danehans, line 107: Same as https://github.com/kubernetes-sigs/service-apis/pull/268/files#r492946054.

apis/v1alpha1/backendpolicy_types.go
robscott: I don't think so. I can imagine some organizations may want to predefine policies before backends exist. Although I don't think it's a common use case, I'd rather not arbitrarily prevent it.

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

apis/v1alpha1/backendpolicy_types.go
hbagdi: > implementation specific status on Routes

Umm, to be clear, the policy is suppose to target a resource like core.Service and not a Route, right?
The whole point of #196 is to associate such configuration, in _most_ cases, to upstream Services and not with Routes.


(I'm fine with the status being what it is here but want to clarify how we are thinking about using backendPolicy)

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,180 @@
...
+	// CertificateAuthorityRef is a reference to a resource that includes
+	// trusted CA certificates for the associated backends. If an entry in this
+	// list omits or specifies the empty string for both the group and the
+	// resource, the resource defaults to "secrets". An implementation may
+	// support other resources (for example, resource "mycertificates" in group
+	// "networking.acme.io").
+	//
+	// Certificates should be specified within the "ca.crt" data field and by
+	// PEM encoded. Multiple certificates can be specified, concatenated by new
+	// lines.
```
hbagdi, line 148:
```suggestion
	// Certificates must be PEM encoded and specified within the "ca.crt" data field of
	// the Secret. Multiple certificates can be specified, concatenated by new
	// lines.
```

hbagdi (approved):
LGTM

apis/v1alpha1/backendpolicy_types.go
robscott: Yep, absolutely. This was just intended as a way to avoid a large number of dimensions in BackendPolicyStatus. As an example, we might end up with a top level set of conditions here and then conditions per backend + conditions per route targeting this backend. This at least temporarily shifts that last one to the Route. I tried to clean up the comments here so it's a bit clearer.

apis/v1alpha1/backendpolicy_types.go
robscott: Updated, thanks!

hbagdi:
LGTM.
I'm not adding `/lgtm` because Daneyon might have additional comments. 
Let's do another PR for the naming issue if needed.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,180 @@
...
+	// the oldest BackendPolicy.
+	//
+	// Support: Core
+	// +kubebuilder:validation:MaxItems=16
+	BackendRefs []BackendRef `json:"backendRefs"`
+
+	// TLS is the TLS configuration for these backends.
+	//
+	// Support: Extended
+	TLS *BackendTLSConfig `json:"tls"`
```
hbagdi, line 62: Why is this a pointer if this is not optional?
Should this be optional? I think we discussed this before and decided against it because it is the only policy we have currently. Does this mean we will make it optional in future if we have something like health-checking in here?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// BackendPolicyStatus defines the observed state of BackendPolicy. Conditions
+// that are related to a specific Route or Gateway should be placed on the
+// Route(s) using backends configured by this BackendPolicy.
+type BackendPolicyStatus struct {
+	// Conditions describe the current conditions of the BackendPolicy.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
```
hbagdi, line 170: Should we have the limit of 5 conditions like we did in the API cleanup PR?

apis/v1alpha1/backendpolicy_types.go
robscott: Good point, this really should be optional, updated.

apis/v1alpha1/backendpolicy_types.go
robscott: I think I'd rather save that for the API cleanup PR so we can have consistent limits everywhere.

apis/v1alpha1/backendpolicy_types.go
hbagdi: This needs `omitempty` now.

robscott: Updated, thanks!

Issue #269, opened 2020-08-21
Reconsider whether route actions are optional.


jpeach:
In most (all?) of the route types, the action field is optional. It's not clear to me what a proxy should do if the action is omitted. Should it drop the traffic on the floor? Or should the action not really be optional?

jpeach:
Applies to TCP, UDP and HTTP route types.

hbagdi:
@robscott @jpeach We have gone back and forth on this one. #312 does some renaming and keeps the forward optional. I think we have found use cases like path/host redirection or HTTP redirection where it makes sense to have the forwarding target as empty.
I might be stating the opposite here, in which case, let's discuss more on this issue.
If not, we might able to close this issue.

robscott:
I think we can close this issue now that #312 is in, feel free to reopen if we need more discussion here.

/close

PR #273, opened 2020-08-24
httproute: make action required

hbagdi:
This commit makes action inside httproute required.
This follows
TCPRoute's action is already a required field.

In future, this condition can be relaxed if there is a need.

See #269

### DIFF
--- a/apis/v1alpha1/generated.proto
+++ b/apis/v1alpha1/generated.proto
@@ -725,7 +725,6 @@ message HTTPRouteRule {
   optional HTTPRouteFilter filter = 2;
 
   // Action defines what happens to the request.
-  // +optional
   optional HTTPRouteAction action = 3;
 }
 
@@ -1173,8 +1172,6 @@ message TCPRouteRule {
   optional TCPRouteMatch match = 1;
 
   // Action defines what happens to the connection.
-  //
-  // +optional
   optional TCPRouteAction action = 2;
 }
 
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -104,7 +104,6 @@ type HTTPRouteRule struct {
 	// +optional
 	Filter *HTTPRouteFilter `json:"filter" protobuf:"bytes,2,opt,name=filter"`
 	// Action defines what happens to the request.
-	// +optional
 	Action *HTTPRouteAction `json:"action" protobuf:"bytes,3,opt,name=action"`
 }
 
--- a/apis/v1alpha1/tcproute_types.go
+++ b/apis/v1alpha1/tcproute_types.go
@@ -49,8 +49,6 @@ type TCPRouteRule struct {
 	// +optional
 	Match *TCPRouteMatch `json:"match" protobuf:"bytes,1,opt,name=match"`
 	// Action defines what happens to the connection.
-	//
-	// +optional
 	Action *TCPRouteAction `json:"action" protobuf:"bytes,2,opt,name=action"`
 }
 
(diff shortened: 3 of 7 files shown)


robscott:
Thanks for the work on this!

/lgtm
/assign @bowei

hbagdi:
@robscott and I had an out of band discussion on this and we decided that it would be best to not add this validation.
There is the common use case of HTTP->HTTPS redirects, for which it doesn't make sense to have an action. (The HTTPS upgrade could be part of the listener on Gateway itself as well, but that's not the point here).
There can also be other use-cases for custom implementations which will render an action useless.

@jpeach Does that sound good to you? If so, I'll close this PR and the linked issue.

robscott:
@hbagdi I actually had a slightly different conclusion here. My thought was actually that every route should require an action. that may be a forwardTo, redirectTo (URL), or an extension ref of some kind. I'd actually thought that HTTP->HTTPS redirects might make the most sense on a Gateway listener. Here's the rough notes I had as I was trying to understand the difference between actions and filters:

**Actions**
* All routes must have a single action 
* Actions may include
  * forwardTo
  * redirectTo (URL only, HTTPS redirect may be best represented on a Gateway listener)
  * extensionRef

**Filters**
* A filter may be applied before or after an action takes place
* Filters can perform actions on a request or response
* Routes may have 0-many filters
* Filters may include
  * Request authentication or authorization
  * Modify headers
  * Mirror traffic

Will add this to agenda for office hours.

jpeach:
> @hbagdi I actually had a slightly different conclusion here. My thought was actually that every route should require an action. that may be a forwardTo, redirectTo (URL), or an extension ref of some kind. I'd actually thought that HTTP->HTTPS redirects might make the most sense on a Gateway listener. Here's the rough notes I had as I was trying to understand the difference between actions and filters:

I'm not sure that there is an agreed-upon distinction between filters and actions. Maybe we don't need to draw get distinction, however, if we drop actions as a general concept and just describe processing in terms of filters and forwarding then the difficulty of drawing a distinction can be avoided. For example:

```Go
type HTTPRouteRule struct {
	Filters []HTTPRouteFilter `json:"filters" protobuf:"bytes,2,rep,name=filters"`
        Forward *HTTPForwardingTarget `json:"forward" protobuf:"bytes,3,opt,name=forward"`
}

type HTTPRouteAction struct {
        To []ForwardToTarget `json:"to" protobuf:"bytes,1,rep,name=to"`
	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef" protobuf:"bytes,2,opt,name=extensionRef"`
}
```

In general, I think it makes sense to have proxy configurations that don't do any forwarding in cases where the proxy itself handles the response (e.g. generates a redirect or a response body).

hbagdi:
We have discussed the above comments and we are in agreement with what James is proposing in the above comment. The next step here is for me to close this PR and open another one with that change. I'll be following up on it in a couple of days.

jpeach:
> We have discussed the above comments and we are in agreement with what James is proposing in the above comment. The next step here is for me to close this PR and open another one with that change. I'll be following up on it in a couple of days.

@hbagdi Are you planning to align all the route types with this style?

hbagdi:
> @hbagdi Are you planning to align all the route types with this style?

Yes, but with 1 PR at a time. I've on my todo to fix all the routes to be as homogenous as HTTPRoute where possible (and also remove filters from places where we don't know how they will be used).

hbagdi:
#273 implements the above proposed naming, meaning this PR is no longer required.
/close

PR #275, opened 2020-08-24
remove auto-generated controller from repo

hbagdi:


hbagdi:
@bowei We will need to get https://github.com/kubernetes/test-infra/pull/18969 merged in before we can merge this.

robscott:
Thanks for taking this on! Just had a few high level notes that didn't directly belong on any individual line:

- We likely still need go fmt and vet, especially if we also start adding a webhook implementation
- We still need to be able to generate informers and listers (pkg/client)
- I think you may be able to get rid of `make deploy`

hbagdi:
> We likely still need go fmt and vet, especially if we also start adding a webhook implementation

Added back in.


> We still need to be able to generate informers and listers (pkg/client)

I might be missing something but it doesn't seem like we have a Make workflow for this. The diff in this PR doesn't change anything for `pkg/client`.

> I think you may be able to get rid of make deploy

Removed.

hbagdi:
@robscott @bowei Do you know what is ` Context retired without replacement. ` in the build job status?

robscott:
@hbagdi I believe that job was removed with https://github.com/kubernetes/test-infra/pull/18969.

tools/tools.go
```diff
@@ -24,5 +24,6 @@ import (
 	_ "github.com/ahmetb/gen-crd-api-reference-docs"
 	_ "k8s.io/code-generator/cmd/go-to-protobuf"
 	_ "k8s.io/code-generator/cmd/go-to-protobuf/protoc-gen-gogo"
+	_ "sigs.k8s.io/controller-runtime/pkg/scheme"
```
robscott, line 27: Why is this being added here?

go.mod
```diff
@@ -4,10 +4,8 @@ go 1.13
 
 require (
 	github.com/ahmetb/gen-crd-api-reference-docs v0.2.0
-	github.com/go-logr/logr v0.1.0
 	github.com/gogo/protobuf v1.3.1
-	github.com/onsi/ginkgo v1.13.0
-	github.com/onsi/gomega v1.10.1
+	github.com/onsi/ginkgo v1.13.0 // indirect
```
robscott, line 8: Where are we using ginkgo?

tools/tools.go
hbagdi: I had to add this to make the CI pass.
I think this is why: https://github.com/kubernetes-sigs/service-apis/blob/master/hack/update-proto.sh#L27. I'm not sure.

go.mod
hbagdi: I think because of the controller-runtime dependency:

```shell
$ go mod why github.com/onsi/ginkgo                    
# github.com/onsi/ginkgo
sigs.k8s.io/service-apis/tools
sigs.k8s.io/controller-runtime/pkg/scheme
sigs.k8s.io/controller-runtime/pkg/scheme.test
github.com/onsi/ginkgo
```

robscott:
Thanks!

/lgtm

PR #249, opened 2020-07-31
httproute: implement HTTPRouteFilter

hbagdi:
This PR implements the "Actions" part of the [the advanced routing](https://docs.google.com/document/d/1SkTb6ECuiQiayvumdsZ3sxDcXD6cYs-DKB_C81Ajvh4/edit#) proposal.

Traffic mirroring is intentionally left out of this PR for brevity and will be added in a follow up PR once we settle on this patch.

Fixes #55

PR #185, opened 2020-05-07
Adds Traffic Mirroring Support to HTTPRoute

danehans:
Adds the `MirrorTo` attribute to an `HTTPRouteAction` for sending a copy of the request to a separate backend object. Note that `MirrorTo` does not expect to receive a response from the backend.

Partially fixes #58 

/assign @bowei 
/cc @robscott @Miciah

PR #276, opened 2020-08-25
 httproute: implement RequestMirror filter 

hbagdi:


hbagdi:
# Note for reviewers

Please review the last commit only. The other commit is the commit in #249

hbagdi:
This overrides #185.

apis/v1alpha1/httproute_types.go
```diff
@@ -277,9 +277,13 @@ const (
...
-	// +optional
-	FilterHTTPRequesttHeader = "RequestHeader"
-	// TODO(hbagdi): add FilterHTTPRequestMirroring filter
+	FilterHTTPReqeustHeader = "RequestHeader"
+
+	// FilterTypeHTTPRequestMirror can be used to mirror requests to a
+	// different backend. The responses from this backend MUST be ignored
+	// by the Gateway.
+	// Support: extended
+	FilteTyperHTTPRequestMirror = "RequestMirror"
```
yiyangy, line 286: typo: s/FilteTyperHTTPRequestMirror/FilterTypeHTTPRequestMirror

apis/v1alpha1/httproute_types.go
```diff
@@ -332,6 +336,8 @@ type HTTPRouteFilter struct {
 	// Filter-specific configuration definitions for core and extended filters
 
 	RequestHeader *HTTPRequestHeaderConfig `json:"requestHeader" protobuf:"bytes,3,opt,name=requestHeader"`
+
+	RequestMirror *HTTPRequestMirrorFilter `json:"requestMirror" protobuf:"bytes,4,opt,name=requestMirror"`
```
robscott, line 340: Nit: RequestMirror and RequestHeader both make sense in isolation but seem a bit confusing when side by side like this. Maybe it would be clearer to have field names like `modifyHeaders` (or `modifyRequestHeaders`) and `mirrorRequest`.

apis/v1alpha1/httproute_types.go
```diff
@@ -377,6 +383,35 @@ type HTTPRequestHeaderConfig struct {
...
+// RequestMirror filter.
+type HTTPRequestMirrorFilter struct {
+	// TargetRef is an object reference to forward matched requests to.
+	// The resource may be "services" (omit or use the empty string for the
+	// group), or an implementation may support other resources (for
+	// example, resource "myroutetargets" in group "networking.acme.io").
+	// Omitting or specifying the empty string for both the resource and
+	// group indicates that the resource is "services".  If the referent
+	// cannot be found, the "InvalidRoutes" status condition on any Gateway
+	// that includes the HTTPRoute will be true.
```
robscott, line 396: This condition makes sense for a ForwardTo but may be a bit extreme for this extended MirrorTo functionality. Maybe there's a way to add a condition to the Route itself for this? Any ideas @jpeach?

apis/v1alpha1/httproute_types.go
```diff
@@ -277,9 +277,13 @@ const (
 	// FilterHTTPReqeustHeader can be used to add or remove an HTTP
 	// header from an HTTP request before it is sent to the upstream target.
 	// Support: core
-	// +optional
-	FilterHTTPRequesttHeader = "RequestHeader"
-	// TODO(hbagdi): add FilterHTTPRequestMirroring filter
+	FilterHTTPReqeustHeader = "RequestHeader"
```
robscott, line 280:
```suggestion
	FilterHTTPRequestHeader = "RequestHeader"
```

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/httproute_types.go
jpeach: Yeh, it probably makes sense to expose more detail about routes on the route itself, but it's also problematic because different controllers resolving the same routes can get different results. I don't have a concrete suggestion for this yet.

apis/v1alpha1/httproute_types.go
hbagdi: Great point. Can I change this in a follow up PR? (I've taken a note of this)

robscott: 👍 I'm fine with this as a follow up. Mind creating an issue to track that?

apis/v1alpha1/httproute_types.go
```diff
@@ -379,6 +385,35 @@ type HTTPRequestHeaderFilter struct {
 	// TODO
 }
 
+// HTTPRequestMirrorFilter defines configuration for the
+// RequestMirror filter.
+type HTTPRequestMirrorFilter struct {
```
howardjohn, line 390: May be wider scoped than this PR, since the same applies to HTTPRequestHeaderFilter, but how can an implementation extend this? For example, we currently support mirroring a percentage of traffic. There are no extension points here. Would we define our own filter?

robscott: That's a great question. I think for now the answer would have to be to define a custom filter for that use case. Although a percentage does seem straightforward enough to add here, I'm hesitant to add at this point. So far, at least for v1alpha1, we've been trying to limit fields to common functionality that is supported by most implementations. This is likely worth revisiting in the future though. Mind creating a follow up issue to track this?

howardjohn: https://github.com/kubernetes-sigs/service-apis/issues/306

apis/v1alpha1/httproute_types.go
hbagdi: https://github.com/kubernetes-sigs/service-apis/issues/307

apis/v1alpha1/httproute_types.go
hbagdi: I agree here.
We have to be careful about adding features in the core filters because we want most (or rather all) implementations to support it.

A meta note since we are here, we also need to figure out a way to track all custom filters that are implemented by vendors so that we can find out about convergence and react accordingly (by promoting some fields/filters into core/extended).

robscott:
Thanks @hbagdi!

/lgtm
/approve

PR #277, opened 2020-08-25
implement TLSRoute for TLS support

hbagdi:


apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Requests will be matched against the Host field in the following order:
+	// 1. If SNI is precise, the request matches this rule if
+	//    the SNI in ClientHello is equal to one of the defined SNIs.
+	// 2. If SNI is a wildcard, then the request matches this rule if
+	//    the SNI is to equal to the suffix
+	//    (removing the first label) of the wildcard rule.
+	//
+	// Support: core
+	//
+	SNIs []string `json:"snis,omitempty" protobuf:"bytes,1,rep,name=snis"`
```
jpeach, line 108: We had feedback on the API doc that there are multiple semantically-equivalent ways to match hostnames . Can we use `HostnameMatch` here?

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,131 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TLSRouteSpec defines the desired state of TLSRoute
+type TLSRouteSpec struct {
```
jpeach, line 23: You can match on the SNI server name at the Listener and use a TCPRoute, so how can we explain when users should do that and when they should use TLSRoute? Maybe we need to address that in the concepts doc rather than in this PR though.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// Rules are a list of TLS matchers and actions.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	Rules []TLSRouteRule `json:"rules" protobuf:"bytes,1,rep,name=rules"`
```
jpeach, line 27: Did you consider following the HTTPRoute Host/Rules pattern here?

https://github.com/kubernetes-sigs/service-apis/pull/238#issuecomment-666008041

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,131 @@
...
+type TLSRoute struct {
+	metav1.TypeMeta   `json:",inline" protobuf:"bytes,1,opt,name=typeMeta"`
+	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,2,opt,name=metadata"`
+
+	Spec   TLSRouteSpec   `json:"spec,omitempty" protobuf:"bytes,3,opt,name=spec"`
+	Status TLSRouteStatus `json:"status,omitempty" protobuf:"bytes,4,opt,name=status"`
+}
+
+// TLSRouteRule is the configuration for a given rule.
+type TLSRouteRule struct {
```
jpeach, line 49: Shall we have this (and TCPRoute too I guess) follow the HTTPRoute pattern by making the `Matched` field a slice and adding a `Filter` field?

apis/v1alpha1/tlsroute_types.go
hbagdi: I did consider that feedback and I think Dan has a great point.
I decided to follow HTTPRoute in this case as this is also a route.

I initially started with changing HTTPRoute to include HostnameMatch in a different PR but that feels more verbose than it needs to be. We should instead change listener in G/W to mirror what we have in HTTPRoute.

apis/v1alpha1/tlsroute_types.go
hbagdi: That's a great point. I actually didn't think about that case because I assumed that most users would have a Any match for TLS.
I think SNI match here and Host header match in HTTPRoute can have the same semantics here.
What change should we do to this PR?

apis/v1alpha1/tlsroute_types.go
hbagdi: Making Match a slice is a good update. I'll update it here and send another PR to update it for TCPRoute.

Adding a filter seems pre-mature. I'd like to know use-cases and and an actual need for it before we add it. If we can do it after v1alpha1, that would be better.

apis/v1alpha1/tlsroute_types.go
hbagdi: I did consider it but decided against it because: 

> Disadvantages are that the binding to the SNI requirement is less obvious, which could make the API harder to reason about and prone to misunderstanding. Maybe

apis/v1alpha1/tlsroute_types.go
hbagdi: > send another PR to update it for TCPRoute.

So, TCPRoute has a match but it only has an extensionRef, which can have plural matches if needed. The counter argument is that we should have a uniformity in our APIs.

jpeach: IMHO we should adjust everything to have:
```Go
Matches []XXXRouteMatch `json:"matches" protobuf:"bytes,1,rep,name=matches"`
```

apis/v1alpha1/tlsroute_types.go
jpeach: Ok let's keep this for now and table the wider discussion.

apis/v1alpha1/tlsroute_types.go
jpeach: Ok, I can sort of see that, but I realized that we no longer have anything that we can have a virtual host concept on in the TCP or TLS routes. That's awkward because a virtualhost seems like a natural boundary to want to attach an extension (e.g. this host needs some custom protocol on top of UDP).

Anyway, we don't need to hold up this PR discussing this problem :)

apis/v1alpha1/tlsroute_types.go
hbagdi: I agree: https://github.com/kubernetes-sigs/service-apis/pull/282

hbagdi: The 'Matches` part of this thread seems to be resolved now.

Regarding a `Filter` field, do we have any use cases that require `Filter`? I'd like to add more fields to the route incrementally (in a backwards compatible way) after this PR is merged in.

hbagdi:
The clientgen is somehow messed up. I'm not sure what exactly is going wrong. It seems to be related to some caching but I've not been able to get to the bottom of it yet.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,134 @@
...
+	Spec   TLSRouteSpec   `json:"spec,omitempty"`
+	Status TLSRouteStatus `json:"status,omitempty"`
+}
+
+// TLSRouteRule is the configuration for a given rule.
+type TLSRouteRule struct {
+	// Matches define conditions used for matching the rule against
+	// incoming TLS handshake.
+	// Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied. +optional
```
jpeach, line 53: `+optional` needs to be on a newline

apis/v1alpha1/tlsroute_types.go
jpeach: Yeh if we don't have any obvious filters now, let's table that discussion.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// SNIs defines a set of SNI names that should match against the
+	// SNI attribute of TLS CLientHello message in TLS handshake.
+	//
+	// SNI can be "precise" which is a domain name without the terminating
+	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is
+	// a domain name prefixed with a single wildcard label (e.g. "*.example.com").
+	// The wildcard character '*' must appear by itself as the first DNS
+	// label and matches only a single label.
+	// You cannot have a wildcard label by itself (e.g. Host == "*").
+	// Requests will be matched against the Host field in the following order:
```
jpeach, line 102: Nit: need a newline here so this formats as a markdown numbered list.

apis/v1alpha1/tlsroute_types.go
jpeach: Perhaps we could add some editorial in the doc comment here. Something like "TLSRoute is similar to TCPRoute but can be configured to match against TLS-specific metadata. This allows more flexibility in matching the streams that a Listener is configured to accept."

???

bowei:
/approve

pending the comments from @jpeach -- it looks good overall

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,144 @@
...
+	//
+	// 1. If SNI is precise, the request matches this rule if
+	//    the SNI in ClientHello is equal to one of the defined SNIs.
+	// 2. If SNI is a wildcard, then the request matches this rule if
+	//    the SNI is to equal to the suffix
+	//    (removing the first label) of the wildcard rule.
+	//
+	// Support: core
+	//
```
robscott, line 120: Nit: we should define min and max lengths here. Maybe 1 and 10 respectively?

hbagdi:
@jpeach Please take another look.
/assign @jpeach

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,144 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TLSRouteSpec defines the desired state of TLSRoute
+
```
robscott, line 23: Nit: I think gofmt might not light this extra line.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -0,0 +1,144 @@
...
+// +kubebuilder:subresource:status
+
+// TLSRoute is the Schema for the TLSRoute resource.
+// TLSRoute is similar to TCPRoute but can be configured to match against
+// TLS-specific metadata.
+// This allows more flexibility in matching streams for in a given TLS listener.
+//
+// If you need to forward traffic to a single target for a TLS listener, you
+// could chose to use a TCPRoute with a TLS listener.
+
```
robscott, line 47: Nit: I think gofmt might not light this extra line.

robscott (commented):
Thanks for your work on this! Just a few minor nits.

apis/v1alpha1/tlsroute_types.go
hbagdi: Can we do this in another PR all over the project as part of #290?
This requires defining a new type (aliasing string) and then defining the length restrictions on that type.

I've added the following for the array length validation:

robscott: Oh good catch, I was actually meaning min and max lengths of the list, though both string and list would be good. Mind adding a kubebuilder:validation:MaxItems line here?

robscott: I'm fine with waiting for the string length restriction

hbagdi: > Mind adding a kubebuilder:validation:MaxItems line here?

Updated.

robscott:
Thanks!

/lgtm

PR #279, opened 2020-08-28
Add linguist attribute to generated files.

jpeach:
Add the "linguist-generated" attribute to generated files so that the
Github UI shows changes in these files folded by default.

Signed-off-by: James Peach <jpeach@vmware.com>


robscott:
This is great, had no idea it was possible. Thanks for cleaning this up!

/lgtm
/assign @bowei

PR #281, opened 2020-08-28
tls: configure downstream TLS certificate in Routes

### LINKED ISSUE #103: Allow creation of domain names and TLS information without interacting with cluster operator
As an application developer, I want to be able to add arbitrary domain names, including TLS information if required, to a gateway without having access to modify Gateway objects.

As an application developer, I want to be able to expose my service in a shared cluster, using a shared load balancer, using my own custom domain, with TLS information either autogenerated or supplied by me.

I do not want to have to interact in any other way than the API with the cluster operator team.

Contrast this use case with #102.

**Why is this needed**:
We've had some requests come in to Contour for similar functionality, because the HTTPProxy model does not work for them.

This issue is an attempt to capture something like the requirement, although I'm not sure if I got it right.

hbagdi:
This patch is part of the [TLS proposal](https://docs.google.com/document/d/15fkzMrhN_7tA-i2mHKwZpqcjN1o2Pe9Am9Qt828x1lo).

Fixes #103


hbagdi:
## Notes for reviewers

Please review the last commit only. This PR builds on top of #256.

/cc @robscott @bowei @jpeach @youngnick

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -25,20 +42,37 @@ package v1alpha1
...
 	// Support: Core (Kubernetes Secrets)
 	// Support: Implementation-specific (Other resource types)
 	//
 	// +required
-	CertificateRefs []CertificateObjectReference `json:"certificateRefs,omitempty" protobuf:"bytes,1,rep,name=certificateRefs"`
+	CertificateRef CertificateObjectReference `json:"certificateRef,omitempty" protobuf:"bytes,2,opt,name=certificateRef"`
+
+	// AllowedInRoute dictates if TLS certificate can be configured via Routes
+	// or not. If set to true, a TLS certificate for a hostname defined in
+	// Route takes takes precedence over the certificate defined in Gateway.
```
yiyangy, line 71: nit: s/takes takes precedence/take precedence

apis/v1alpha1/httproute_types.go
```diff
@@ -76,7 +85,24 @@ type HTTPRouteHost struct {
 	// Support: custom
 	//
 	// +optional
-	ExtensionRef *RouteHostExtensionObjectReference `json:"extensionRef" protobuf:"bytes,3,opt,name=extensionRef"`
+	ExtensionRef *RouteHostExtensionObjectReference `json:"extensionRef" protobuf:"bytes,4,opt,name=extensionRef"`
+}
+
+// TLSinRouteConfig describes a TLS configuration defined at the Route level.
+type TLSInRouteConfig struct {
```
yiyangy, line 92: Does it make sense to define this in tlsconfig_types.go ?

robscott: That's a good point. If we end up having separate TLSConfig types per resource maybe we can leave this here? For BackendTLSConfig I'm proposing keeping it in the `backendpolicy_types.go` file. On that note, would `RouteTLSConfig` or `HTTPRouteTLSConfig` be a better name here? I'm not to set on a name here, just think we should match behavior between this and https://github.com/kubernetes-sigs/service-apis/pull/268

apis/v1alpha1/httproute_types.go
```diff
@@ -76,7 +85,24 @@ type HTTPRouteHost struct {
...
+	// CertificateRef is the reference to Kubernetes object that
+	// contain a TLS certificate and private key.
+	// This certificate MUST be used for TLS handshakes for the domain
+	// this TLSConfig is associated with.
```
robscott, line 96: Nit: TLSConfig does not match the updated type name here.

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -65,6 +65,14 @@ type TLSConfig struct {
 	//
 	// +required
 	CertificateRef CertificateObjectReference `json:"certificateRef,omitempty" protobuf:"bytes,2,opt,name=certificateRef"`
+
+	// AllowedInRoute dictates if TLS certificate can be configured via Routes
```
robscott, line 69: Nit: I think this should either by "if a TLS certificate" or "if TLS certificates"

apis/v1alpha1/httproute_types.go
```diff
@@ -59,10 +59,19 @@ type HTTPRouteHost struct {
 	// +optional
 	Hostnames []string `json:"hostnames,omitempty" protobuf:"bytes,1,opt,name=hostnames"`
 
+	// TLS defines the TLS certificate to use for Hostnames defined in this
+	// Route. The certificate defined here take effect only if AllowedInRoute
```
robscott, line 63:
```suggestion
	// Route. The certificate defined here takes effect only if the AllowedInRoute
```

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/httproute_types.go
hbagdi: I did consider moving the type into TLSConfig but had similar though process as Rob on this one.
Regarding name of the struct itself, I think RouteTLSConfig makes sense here since we are referring to downstream TLS configuration and I expect that to be same across multiple types of Routes (HTTPRoute or TLSRoute). Thoughts?

jpeach: I'm hugely +1 on establishing a naming pattern for the different TLS config types. I could live with `TLSInRouteConfig` and `TLSInListenerConfig`, etc. Updating that in a separate PR would be best though.

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -65,6 +65,14 @@ type TLSConfig struct {
 	//
 	// +required
 	CertificateRef CertificateObjectReference `json:"certificateRef,omitempty"`
+
+	// AllowedInRoute dictates if TLS certificates can be configured via Routes
+	// or not. If set to true, a TLS certificate for a hostname defined in
+	// Route takes takes precedence over the certificate defined in Gateway.
+	// +kubebuilder:default=false
+	// +optional
+	AllowedInRoute bool `json:"allowedInRoute,omitempty"`
```
jpeach, line 74: Is there a more specific name that evokes a better idea of what this field does? Perhaps `AllowRouteCertificates` or `ImportRouteCertificates`? Or maybe this needs to express that TLS config in general can be imported (merged?) from the routes?

Here's a YAML example that shows the problem:
```yaml
tls:
  mode: Terminate
  allowedInRoute: true
  certificateRef:
    name: foo
```

apis/v1alpha1/httproute_types.go
```diff
@@ -59,6 +59,15 @@ type HTTPRouteHost struct {
...
 	Hostnames []string `json:"hostnames,omitempty"`
 
+	// TLS defines the TLS certificate to use for Hostnames defined in this
+	// Route. The certificate defined here takes effect only if the AllowedInRoute
+	// field is set to true in the associated Gateway resource.
+	//
+	// Support: Core
+	//
+	// +optional
+	TLS *TLSInRouteConfig `json:"tls" protobuf:"bytes,2,opt,name=tls"`
```
howardjohn, line 69: Sorry I don't have access to the doc to get more context - shouldn't this equally apply to tcp route as well?

apis/v1alpha1/tlsconfig_types.go
danehans: nit: s/Route takes takes precedence/Route takes precedence/

danehans:
@hbagdi can you allow access to the TLS proposal in https://github.com/kubernetes-sigs/service-apis/pull/281#issue-475549551?

hbagdi:
I messed up the link to the Google doc.
This is the correct link:
 https://docs.google.com/document/d/15fkzMrhN_7tA-i2mHKwZpqcjN1o2Pe9Am9Qt828x1lo

Also updated the link in the PR description. Sorry for the trouble.

apis/v1alpha1/httproute_types.go
hbagdi: Renamed this to RouteTLSConfig.

apis/v1alpha1/tlsconfig_types.go
hbagdi: As discussed in the office hours, renamed the field to `AllowOverrideViaRoute`.

apis/v1alpha1/httproute_types.go
hbagdi: It doesn't really make sense at the TCPRoute level because TCPRouteMatch doesn't have any SNI based matching semantics.
This might make sense in TLSRoute though.

apis/v1alpha1/tlsconfig_types.go
```diff
@@ -65,6 +65,18 @@ type TLSConfig struct {
...
+	// AllowOverrideViaRoute dictates if TLS certificates can be configured
+	// via Routes or not. If set to true, a TLS certificate for a hostname
+	// defined in  Route takes precedence over the certificate defined in
+	// Gateway.
+	// CertificateRef must be defined even if this is set to true as it will
+	// be used as the default certificate for the listener.
+	//
+	// +kubebuilder:default=false
+	// +optional
+	AllowOverrideViaRoute bool `json:"allowOverrideViaRoute,omitempty"`
```
robscott, line 78: Not sure if this is actually better, but would `AllowRouteOverride` work here?

hbagdi:
I've updated the fieldnames to AllowOverrideViaRoute and RouteTLSConfig.
I've also updated language on requiring certificate at the listener level.
PTAL

cc @jpeach @yiyangy  @danehans @robscott

apis/v1alpha1/httproute_types.go
```diff
@@ -58,6 +58,16 @@ type HTTPRouteHost struct {
 	// +optional
 	Hostnames []string `json:"hostnames,omitempty"`
 
+	// TLS defines the TLS certificate to use for Hostnames defined in this
+	// Route. The certificate defined here takes effect only if the
```
robscott, line 62: Maybe a small change since the certificate itself isn't actually defined here.
```suggestion
	// Route. This configuration only takes effect if the
```

apis/v1alpha1/httproute_types.go
```diff
@@ -78,6 +88,23 @@ type HTTPRouteHost struct {
 	ExtensionRef *RouteHostExtensionObjectReference `json:"extensionRef"`
 }
 
+// RouteTLSConfig describes a TLS configuration defined at the Route level.
+type RouteTLSConfig struct {
+	// CertificateRef is the reference to Kubernetes object that
```
robscott, line 93:
```suggestion
	// CertificateRef refers to a Kubernetes object that
```

apis/v1alpha1/httproute_types.go
```diff
@@ -78,6 +88,23 @@ type HTTPRouteHost struct {
...
+	// contain a TLS certificate and private key.
```
robscott, line 94:
```suggestion
	// contains a TLS certificate and private key.
```

apis/v1alpha1/httproute_types.go
```diff
@@ -78,6 +88,23 @@ type HTTPRouteHost struct {
...
+	// This certificate MUST be used for TLS handshakes for the domain
+	// this RouteTLSConfig is associated with.
```
robscott, line 96: What if this domain is used by other routes that provide other certificates? This could potentially be interpreted as: "any TLS handshake involving the domain defined here should use this cert".

apis/v1alpha1/httproute_types.go
```diff
@@ -58,6 +58,16 @@ type HTTPRouteHost struct {
...
 
+	// TLS defines the TLS certificate to use for Hostnames defined in this
+	// Route. The certificate defined here takes effect only if the
+	// AllowOverrideViaRoute field is set to true in the associated Gateway
+	// resource.
+	//
+	// Support: Core
+	//
+	// +optional
+	TLS *RouteTLSConfig `json:"tls" protobuf:"bytes,2,opt,name=tls"`
```
danehans, line 69: Why not use `TLSConfig`? I know that `Mode` is not applicable here, but can't that be addressed in docs?

apis/v1alpha1/httproute_types.go
```diff
@@ -78,6 +88,23 @@ type HTTPRouteHost struct {
...
 }
 
+// RouteTLSConfig describes a TLS configuration defined at the Route level.
+type RouteTLSConfig struct {
+	// CertificateRef is the reference to Kubernetes object that
+	// contain a TLS certificate and private key.
+	// This certificate MUST be used for TLS handshakes for the domain
+	// this RouteTLSConfig is associated with.
+	// If an entry in this list omits or specifies the empty
+	// string for both the group and the resource, the resource defaults to "secrets".
```
robscott, line 98: This could use some clarification. The default group is "core" and the default resource is "secrets". The current wording makes it seem like both need to be left empty for the default to apply.

apis/v1alpha1/httproute_types.go
```diff
@@ -78,6 +88,23 @@ type HTTPRouteHost struct {
...
+	// An implementation may support other resources (for example, resource
+	// "mycertificates" in group "networking.acme.io").
+	// Support: Core (Kubernetes Secrets)
+	// Support: Implementation-specific (Other resource types)
+	//
+	// +required
+	CertificateRef CertificateObjectReference `json:"certificateRef,omitempty" protobuf:"bytes,2,opt,name=certificateRef"`
```
robscott, line 105: Proto tags need to be removed - I missed this in a couple of my PRs as well.

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/httproute_types.go
robscott: One of the bits of feedback from Tim was to avoid recycling types quite so much, especially if not all fields applied in all cases. I think it's useful to ensure our TLSConfig types are consistent but I'd rather not have extra fields in the API that can't actually be used.

apis/v1alpha1/tlsconfig_types.go
hbagdi: Sure, that sounds pleasant to my ears as well. Will update shortly.

apis/v1alpha1/httproute_types.go
hbagdi: I initially did that but decided against for several reasons:
- The generated CRD will have the fields and it can and probably will confuse users if the fields are present.
- Implementers that come in later could misinterpret things.
- We might add additional fields at the Gateway-level which will automatically appear here. I think it would be better for us to make a conscious decisions to introduce fields at this level instead.
- Based on the previous discussions, it seems that providing ability to configure certificate is the most asked for requirement. It is unlikely that we will want to expand TLS at this level much more.

apis/v1alpha1/httproute_types.go
hbagdi: > What if this domain is used by other routes that provide other certificates?

That is a conflict. Route can't be selected at the time of TLS session so only one route should define a TLS certificate for a hostname. Or all routes should point to the same certificate.
We have the same problem in Ingress v1.

apis/v1alpha1/httproute_types.go
hbagdi: I copied this over from https://github.com/kubernetes-sigs/service-apis/blob/master/apis/v1alpha1/tlsconfig_types.go#L59-L60. Can we change this outside this PR?

apis/v1alpha1/httproute_types.go
robscott: 👍 That's a great point, that seems like something we should also document on the Route hostnames field.

apis/v1alpha1/tlsconfig_types.go
danehans: What if this field was changed from a bool to a struct that provides additional override flexibility? Something like:

```
// RouteOverrideType defines a schema for overriding gateway certificates.
// +kubebuilder:validation:Enum=AllowAll,DenyAll,AllowNamespaces
// +kubebuilder:default=DenyAll
type RouteOverrideType string

const (
	// AllowAll allows cert overrides for all routes.
	AllowAll RouteOverrideType = "AllowAll"

	// Deny denies cert overrides for all routes.
	DenyAll RouteOverrideType = "DenyAll"

	// AllowNamespaces allows cert overrides for routes in specified namespaces.
	AllowNamespaces RouteOverrideType = "AllowNamespaces"
)

// CertOverridePolicy defines a schema for overriding gateway certificates.
type CertOverridePolicy struct {
	// Type is the type of cert override policy for this gateway to use.
	//
	// Support: Core
	//
	// +optional
	// +kubebuilder:default=DenyAll
	Type RouteOverrideType `json:"type"`

        // namespaces is the list of namespaces where routes can override gateway certs.
	Namespaces []string `json:"namespaces"`

type TLSConfig struct {
...
CertOverridePolicy CertOverridePolicy `json:"certOverridePolicy"`
...
}

File: apis/v1alpha1/httproute_types.go (hunk 13 above)
Comment (danehans) [reply apis/v1alpha1/httproute_types.go]: Got it, thanks for the additional input.

File: apis/v1alpha1/tlsconfig_types.go (hunk 8 above)
Comment (hbagdi) [reply apis/v1alpha1/tlsconfig_types.go]: We can make this more flexible if there is a need for this although we are going to sacrifice on simplicity quite a bit.
Since TLS is a sensitive piece of configuration, we should aim to keep things a little dumb, where possible, to minimize the number of bugs in implementations and end-user interpretation of the API.

I see the reasoning for adding what you are proposing but I'm concerned about the complexity here.
I'm totally open to being convinced otherwise here!

File: apis/v1alpha1/httproute_types.go (hunk 12 above)
Comment (hbagdi) [reply apis/v1alpha1/httproute_types.go:105]: Updated. PTAL.

File: apis/v1alpha1/tlsconfig_types.go (hunk 8 above)
Comment (hbagdi) [reply apis/v1alpha1/tlsconfig_types.go]: I've updated the PR with some parts of what @danehans has proposed. Please take a look.

File: apis/v1alpha1/httproute_types.go [hunk 16]
```diff
@@ -58,6 +58,26 @@ type HTTPRouteHost struct {
 	// +optional
 	Hostnames []string `json:"hostnames,omitempty"`
 
+	// TLS defines the TLS certificate to use for Hostnames defined in this
+	// Route. This configuration only takes effect if the AllowRouteOverride
+	// field is set to true in the associated Gateway resource.
+	//
+	// Collisions can happen if multiple HTTPRoutes define a TLS certificate
+	// for the same hostname. The behavior in such a case is implementation
```
robscott, line 66: Does this behavior need to be implementation specific? Could we rely on some combination of most specific hostname match + oldest resource wins? If there's a conflict within a route maybe we could then say that the first matching rule in that route wins?

hbagdi: > Could we rely on some combination of most specific hostname match + oldest resource wins?

This comment applies when the hostnames are equal so most specific hostname doesn't matter here. The oldest resource wins makes sense here.

We have had users who have an edge case where the oldest resource has an expired certificate and they expected the controller to fallback to the next valid certificate in that case. While this is technically possible, it is more complicated than I'd like the controller to be. Is that fair?

> If there's a conflict within a route maybe we could then say that the first matching rule in that route wins?

SGTM.

apis/v1alpha1/gateway_types.go
```diff
@@ -297,6 +297,35 @@ const (
...
+	// Certificate dictates if TLS certificates can be configured
+	// via Routes. If set to 'AllowAll', a TLS certificate for a hostname
+	// defined in a Route takes precedence over the certificate defined in
+	// Gateway.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:default=DenyAll
+	Certificate TLSRouteOverrideType `json:"certificate"`
```
danehans, line 326: @hbagdi here is an [example](https://github.com/projectcontour/contour-operator/pull/24/files) of using kubebuilder defaulting for required custom types.

apis/v1alpha1/gateway_types.go
```diff
@@ -297,6 +297,35 @@ const (
...
 
+// TLSRouteOverrideType type defines the level of allowance for Routes
+// to override a specific TLS setting.
+// +kubebuilder:validation:Enum=AllowAll;DenyAll
+// +kubebuilder:default=DenyAll
+type TLSRouteOverrideType string
+
+const (
+	// AllowAll allows the parameter to be configured from all routes.
+	AllowAll TLSRouteOverrideType = "AllowAll"
```
bowei, line 308: Not sure if the "All" makes sense any more

Maybe `Allow` or `Deny`

apis/v1alpha1/gateway_types.go
bowei: Does this need to be pointer type (optional struct)

apis/v1alpha1/gateway_types.go
```diff
@@ -330,6 +359,17 @@ type GatewayTLSConfig struct {
...
+
+	// AllowRouteOverride dictates if TLS settings can be configured
+	// via Routes or not.
+	//
+	// CertificateRef must be defined even if allowRouteOverride.certificate is
+	// set to 'AllowAll' as it will be used as the default certificate for the
+	// listener.
+	//
+	// +optional
+	AllowRouteOverride TLSOverridePolicy `json:"allowRouteOverride"`
```
bowei, line 371: Hmm we should think about a better name:

```
AllowRouteOverride: Allow
```

how about?

```
RoutePolicy: AllowOverride

or

RouteOverride: Allow
```

apis/v1alpha1/gateway_types.go
danehans: @bowei the field does not need to be a pointer since we're using defaulting. @hbagdi if we don't want to require that users specify a value for this field, then add the `omitempty` json tag.

danehans: `// +optional` should also be removed since defaulting will set set the value.

apis/v1alpha1/gateway_types.go
danehans: I'm in favor of either of ^ that @bowei suggests.

danehans:
@hbagdi can you add an example to `/examples` as part of this PR?

danehans:
@hbagdi you may also want to reference https://github.com/kubernetes-sigs/service-apis/issues/290 for field defaulting/validation behavior.

apis/v1alpha1/gateway_types.go
robscott: Since this would be a per-attribute struct, I think I prefer `RouteOverride` as the name here. That would translate into something like this:

```
routeOverride:
  certificate: Allow
```

And in the future that might be expanded with more fields and/or more potential values such as `AllowNamespaceList`.

hbagdi:
I've updated the descriptions and field names and also added an example (the example is based on the new flattened route which is not merged yet).

PTAL @bowei @robscott @danehans @jpeach @yiyangy @howardjohn

robscott:
Thanks @hbagdi! This LGTM, will approve it and leave the final LGTM for someone else.

/approve

apis/v1alpha1/httproute_types.go
```diff
@@ -77,6 +98,23 @@ type HTTPRouteHost struct {
...
+	// this RouteTLSConfig is associated with.
+	// If an entry in this list omits or specifies the empty
+	// string for both the group and the resource, the resource defaults to "secrets".
+	// An implementation may support other resources (for example, resource
+	// "mycertificates" in group "networking.acme.io").
+	// Support: Core (Kubernetes Secrets)
+	// Support: Implementation-specific (Other resource types)
+	//
+	// +required
+	CertificateRef CertificateObjectReference `json:"certificateRef,omitempty"`
```
danehans, line 115: @hbagdi `omitempty` should be removed when the field is required and defaulting is not being used.

hbagdi:
@danehans Fixed, PTAL.

PR #282, opened 2020-08-28
tcproute: pluralize match and make action required

hbagdi:
- Pluralizing match to match what we have in HTTPRoute
- An optional action doesn't really make sense here. We can relax this
in future in a non-breaking fashion if there comes a need


hbagdi:
See https://github.com/kubernetes-sigs/service-apis/pull/277#discussion_r476107045

apis/v1alpha1/tcproute_types.go
```diff
@@ -51,13 +51,15 @@ type TCPRoute struct {
...
+	// Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied.
 	//
 	// +optional
-	Match *TCPRouteMatch `json:"match" protobuf:"bytes,1,opt,name=match"`
+	Matches []TCPRouteMatch `json:"matches" protobuf:"bytes,1,rep,name=matches"`
+
 	// Action defines what happens to the connection.
-	//
-	// +optional
```
robscott, line 60: Instead of just removing `+optional` we probably want to make this non-nil and required.

apis/v1alpha1/tcproute_types.go
```diff
@@ -51,13 +51,15 @@ type TCPRoute struct {
...
 type TCPRouteRule struct {
-	// Match defines which connections match this rule.
+	// Matches define conditions used for matching the rule against
+	// incoming TCP connections.
+	// Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied.
 	//
 	// +optional
-	Match *TCPRouteMatch `json:"match" protobuf:"bytes,1,opt,name=match"`
+	Matches []TCPRouteMatch `json:"matches" protobuf:"bytes,1,rep,name=matches"`
```
robscott, line 60: Although I like pluralizing matches for HTTPRoute it's less clear what the use case is here. I believe the only kind of match we support at this level is an `extensionRef`. Since it's always easy to move from singular to plural in the future I think I'd personally rather leave this as singular until we have a use case for multiple matches here.

hbagdi: The background for this change is in the following conversation:
https://github.com/kubernetes-sigs/service-apis/pull/277#discussion_r476107045

This change is more about API consistency than anything else.

jpeach: Yeh, I'd prefer to be consistent with the plurality even though the use case is multiple extension matches.

hbagdi: @robscott Do you agree/disagree here?
(Pinging to move the PR forward to merge or close)

robscott: 👍 I can get behind that for the sake of consistency

hbagdi: Okay, resolving this thread on this PR then.

hbagdi:
Rebased.
/assign @robscott @bowei

jpeach:
@hbagdi need rebase 😂

robscott:
Thanks for the work on this @hbagdi, looks like you just need one more rebase. Approving now so any lgtm + rebase will get it in.

/approve

hbagdi:
@jpeach @robscott Rebased again.

robscott:
Thanks!

/lgtm

PR #283, opened 2020-08-30
Remove generated protobufs

howardjohn:
There have been concerns about merge conflicts, etc by protobuf. This removes them from the repo. Consumers can integrated by controller-manager or the generated client


robscott:
Thanks for the work on this!
/cc @hbagdi

Makefile
```diff
@@ -43,38 +43,6 @@ generate:
 manifests:
 	$(MAKE) -f kubebuilder.mk manifests
 
-# Generate protobufs
```
hbagdi, line 46: Please remove `make proto` from line 39 and update the corresponding comments as well.

hbagdi: We can also remove PROTO_IMG definition from this file.

apis/v1alpha1/generated.pb.go
```diff
@@ -1,11639 +0,0 @@
-/*
```
hbagdi, line 1: While this file is removed, it will be present in local branches of existing PRs and could be added by mistake by contributors.
Reviewers can ensure that this doesn't happen but we can automate this by failing the verify job if these deleted files are present. Not a blocker for this PR to get in.

jpeach:
@howardjohn I guess we still want to support the protobuf field tags? Do you know of a way to validate that they are correct?

howardjohn:
> @howardjohn I guess we still want to support the protobuf field tags? Do you know of a way to validate that they are correct?

What is the benefit if we also do not have the generated proto? I don't think we had them before we introduced protobuf, they are automatically added.

jpeach:
> > @howardjohn I guess we still want to support the protobuf field tags? Do you know of a way to validate that they are correct?
> 
> What is the benefit if we also do not have the generated proto? I don't think we had them before we introduced protobuf, they are automatically added.

I'm not familiar with how protobuf works in CRDs, but it sounds like you are saying that we can safely remove the protobuf field tags? And not break anyone that wants to use the API w/ protobufs?

howardjohn:
@jpeach if we merge this PR, anyone using the api via protobufs will be broken regardless of if the tags are there. The tags are necessary but not sufficient to use is my understanding, at least

hbagdi:
The conflicts this merge will create on existing PRs is not going to be fun but there is no alternative. This change will reduce the conflicts that arise in future PRs and that should pay off the near term overhead.
/lgtm

PR #284, opened 2020-08-31
Add back kustomize support

howardjohn:
This was removed in
https://github.com/kubernetes-sigs/service-apis/pull/274.

This breaks 1 line installation of CRDs used by a few demos already like
`kubectl apply -k 'github.com/kubernetes-sigs/service-apis/config/crd'`

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -1,0 +1,5 @@
+resources:
+- bases/networking.x-k8s.io_gatewayclasses.yaml
+- bases/networking.x-k8s.io_gateways.yaml
+- bases/networking.x-k8s.io_httproutes.yaml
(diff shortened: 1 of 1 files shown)


howardjohn:
cc @hbagdi

hbagdi:
/ok-to-test

We should automate this if possible but not a blocker for this PR to get in.
/lgtm

config/crd/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+resources:
```
jpeach, line 1: Is the objectmeta required here?
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
```

jpeach: Oh I see; it wasn't needed in the original, so it should be fine :)

PR #285, opened 2020-09-01
Adding Gateway selection to Routes

robscott:
This allows Routes to limit which Gateways can use them. This will replace the less flexible AllowedRouteNamespaces field on GatewayClass.

This is a follow up to https://github.com/kubernetes-sigs/service-apis/pull/263 where I had not been able to come up with a good solution to limit the blast radius of AllowedRouteNamespaces. As I've thought about it more, I think this approach could provide a simpler API with more flexibility and of course a smaller blast radius for potential changes.

I've also taken the opportunity to move some of the related fields that had been shared between TCPRoute and HTTPRoute into this new types file.

/cc @hbagdi @jpeach @yiyangy @danehans


hbagdi:
This seems like a large change and turns the table over when it comes route selection.
The purpose of `AllowedRouteNamespaces` was to limit route selection from namespaces and this is something a gateway owner or cluster owner would have control over.

With this change, we are completely changing the control mechanism and instead asking the service owner to define the selection of gateway (which is already possible with labels).

robscott:
>This seems like a large change and turns the table over when it comes route selection. The purpose of AllowedRouteNamespaces was to limit route selection from namespaces and this is something a gateway owner or cluster owner would have control over.

>With this change, we are completely changing the control mechanism and instead asking the service owner to define the selection of gateway (which is already possible with labels).

Although this seems like a large change it may not be quite as significant as it initially seems. I think the most common use case here will involve Gateways referencing routes in the same namespace. Nothing changes there. Everything continues to work as is.

The most significant change here has to do with cross-namespace route references. Previously the GatewayClass admin would have to allow routes to be referenced in a certain set of namespaces by users of that GatewayClass. Then Gateway admins would be able to select Routes in multiple namespaces. This removes the GatewayClass requirement and replaces it with route level config.

Essentially this adds a requirement for cross-namespace Route references that both the Gateway and Route have agreed to that. I think that's a much more flexible way to approach this. It doesn't require new GatewayClasses for subsets of allowed route namespaces, and it requires explicit buy in/acknowledgement from the route owner that this route can be referenced from another namespace.

And of course the original motivation is still significant here. There's no more potential for a change to AllowedRouteNamespaces to invalidate a huge number of routes referenced by Gateways. The blast radius has transitioned from all resources sharing a GatewayClass to anything that might be referencing an individual route.

hbagdi:
I think we are mixing two different things here:
- One one hand, we want to remove AllowedRouteNamespaces because of blast radius concerns, which are completely valid
- On other hand, we are introducing a feature at the route-level to perform some sort of "handshake"  between the route and gateway, which is a net new feature and doesn't really relate with the problem at hand.

I think both are valid but we should not mix them up. To me, both of these are orthogonal.


I'm not able to come up with a stance on if we should ship this feature or not.
Have we had any request for this feature?

robscott:
> I think we are mixing two different things here:
> * One one hand, we want to remove AllowedRouteNamespaces because of blast radius concerns, which are completely valid
> * On other hand, we are introducing a feature at the route-level to perform some sort of "handshake" between the route and gateway, which is a net new feature and doesn't really relate with the problem at hand.
>
> I think both are valid but we should not mix them up. To me, both of these are orthogonal.

Although I get that they seem disconnected, I think they're more closely related than they seem on the surface. In my opinion we simply cannot ship Gateways that can select routes across multiple namespaces without some kind of mechanism to limit that selection. The very nature of using a selector to target Routes makes this even riskier than say a list of object refs. It becomes remarkably easy to target all Routes in the cluster from any Gateway anywhere. It's a security nightmare. Namespaces are meant to be a significant security boundary in Kubernetes and allowing selection of resources across that boundary is not something that should be taken lightly. There's a reason that Services can only select endpoints within their namespace. We could of course just not ship multi-namespace route selection at all, but I think that it's still a desirable enough feature to work towards.

So that leaves us with how we can restrict this feature so it is both reasonably safe to support and not overly painful to use. My initial attempt with the `AllowedRouteNamespaces` field on GatewayClass had significant problems, including:

* large blast radius for changes.
* potential for large number of GatewayClass resources to achieve desired isolation.

The approach I'm proposing in this PR, although very different than the original approach, has the same goal. It is trying to provide both safety and simplicity. By default you don't have to do anything - you can select routes in the same namespace. If you want to cross the namespace boundary, you have to coordinate with the owner of Routes you are trying to target in another namespace. In single-tenant clusters, Route owners can simply set their routes to be accessible by any Gateway. This is the best approach I can think of that allows us to safely support multi-namespace Route selection, but I'm open to alternatives.

apis/v1alpha1/gateway_types.go
```diff
@@ -398,6 +398,39 @@ type RouteBindingSelector struct {
...
+	// An empty selector (default) indicates that Routes in any namespace can be
+	// selected.
+	//
+	// The OnlySameNamespace field takes precedence over this field. This
+	// selector will only take effect when OnlySameNamespace is false.
+	//
+	// Support: Core
+	//
+	// +optional
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector"`
```
hbagdi, line 418: Should this be named `namespace`?
We have had feedback that we shouldn't have `selector` in the name.

danehans:
@robscott @hbagdi I can see both of your perspectives and I too am unsure of the best approach. What if the route namespace selection stayed with with `GatewayClass`, but only consisted of `OnlySameNamespace` for v1alpha1? I feel like `OnlySameNamespace` addresses a majority of the use cases, prevents the blast radius attributed to label-based namespace selection, maintains the existing security model, and simplifies the user experience. Thoughts?

apis/v1alpha1/gateway_types.go
robscott: I think this may be out of scope for this PR since this is just moving the fields from GatewayClass to Gateway types files. I do agree the name may be a bit too repetitive (`routes.routeNamespaces.namespaceSelector`). I would actually lean the other way though - that we might want to change the name to `routes.namespaces.selector`. Maybe we need a follow up issue for this?

robscott:
@danehans That's an interesting idea, it would definitely be an easy option to fall back to if this approach doesn't end up being viable. Unfortunately I think the blast radius concerns are still significant. Changing `OnlySameNamespace` from false to true could result in a very destructive change for all Gateways using that class.

apis/v1alpha1/gateway_types.go
hbagdi: https://github.com/kubernetes-sigs/service-apis/issues/311

danehans:
> Unfortunately I think the blast radius concerns are still significant. Changing OnlySameNamespace from false to true could result in a very destructive change for all Gateways using that class.

@robscott I agree that `OnlySameNamespace` still has a significant blast radius, but it's also a very straight forward yes/no are routes allowed in all ns's or just the gateway ns. Maybe adding docs to `OnlySameNamespace` such as "changing this value should be done with extreme caution...". IMO this is a reasonable solution for v1alpha1 and we iterate based on future feedback and  use cases. @hbagdi @bowei @jpeach any thoughts here?

apis/v1alpha1/tcproute_types.go
```diff
@@ -23,6 +23,11 @@ import (
 type TCPRouteSpec struct {
 	// Rules are a list of TCP matchers and actions.
 	Rules []TCPRouteRule `json:"rules"`
+
+	// Gateways defines which Gateways can use this Route.
+	// +kubebuilder:default={allow: "SameNamespace"}
+	// +optional
+	Gateways *RouteGateways `json:"gateways,omitempty"`
```
hbagdi, line 30: Please add this to UDPRoute and  TLSRoute.

hbagdi:
After thinking for a bit and playing out scenarios of how users will use this, I think I've a better understanding here.

There are multiple changes going on here.

The first one, GatewayClass no longer has a route selection mechanism.
This makes all the sense to me. I never really understood the usefulness of route selection at a class level especially when the same selection was already present at the Gateway level. For a human who controls GatewayClass, they want control over where/who can use the GatewayClass, meaning who can spawn valid Gateways. I think we are getting to an okay place with that regard in #263. Route selection for a specific Gateway is a concern of the Gateway owner.

Given the multi-tier route selection we have in the mainline branch, I really doubt if implementers and users can understand it correctly, and most will just disable route selection at one of the levels (GatewayClass being most likely candidate).

The second changes, related to the first, is for routes to have the ability of selecting Gateways:

> The approach I'm proposing in this PR, although very different than the original approach, has the same goal. It is trying to provide both safety and simplicity. By default you don't have to do anything - you can select routes in the same namespace. If you want to cross the namespace boundary, you have to coordinate with the owner of Routes you are trying to target in another namespace.

This is an important point. Making a Gateway-Route binding explicit gives humans operators a knob for use-cases where they wish to have strict control over the self-service experience of the traffic ingress stack. This is also important for service-owners who want some control over where the traffic comes from (up until now, I've always thought "Gateway owners care which route to select and expose", which is really half the story).

I do not think we are sacrificing any use-cases with this change and it solves the problems of blast radius and security implications as well.
Another positive aspect here is that we have sane defaults and most users who don't require this level of complexity, don't even have to learn about it.

After writing this comment, I realized that I was saying the same thing that Rob has been trying to communicate for so long. :man_facepalming:

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks for catching this!

robscott:
@danehans Thanks for the feedback here. This is definitely a significant change/addition but I think that it's worthwhile. Although I'm all for limiting the scope of v1alpha1, one of the things I'd like to get mostly right is how our resources interact with each other. We can always add fields to resources, but ideally our initial API would get the security and resource model mostly right the first time. Changes at that level will be more painful/significant in future releases.

As much as I originally proposed the concept of AllowedRouteNamespaces on Gateways, it never felt like an ideal solution. Whenever we have to write a comment along the lines of "only change this value with extreme caution" (as I originally proposed) our API might have a problem. 

I think this proposal is fundamentally better than an attribute on GatewayClass. It completely removes the blast radius issues, provides finer grained control, gives that control to the appropriate personas, and manages to achieve that with little to no additional overhead for the standard use case (single-namespace Gateways and Routes). 

I think @hbagdi's summary above is excellent. This is one of those things where once it clicked for me, the previous approach just felt fundamentally broken. With that said, I'm very open to additional feedback/thoughts on this approach.

docs-src/security-model.md
```diff
@@ -68,9 +68,10 @@ readable by most roles, so instead we'll focus on write access for this model.
 | Application Admins | No | In Specified Namespaces | In Specified Namespaces |
 | Application Developers | No | No | In Specified Namespaces |
 
-## Namespace Restrictions
-The extra configuration options are not possible to control with RBAC. Instead,
-they will be controlled with configuration fields on GatewayClasses:
+## GatewayClass Namespace Limitations 
+Kubernetes RBAC only allows us to grant access to create Gateways as a whole,
+not limit creation to specific GatewayClasses. A field on GatewayClass allows
```
danehans, line 73: @robscott you're referencing the `allowedGatewayNamespaceSelector` field, correct?

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/onsi/ginkgo v1.13.0 // indirect
 	golang.org/x/net v0.0.0-20200904194848-62affa334b73 // indirect
 	golang.org/x/tools v0.0.0-20200904185747-39188db58858 // indirect
+	k8s.io/api v0.19.0
```
danehans, line 11: @robscott why is a dep included in this PR?

docs-src/security-model.md
robscott: Thanks for catching this! This whole section was out of date - I've updated it here but will also need to reconcile that whenever https://github.com/kubernetes-sigs/service-apis/pull/263 merges.

go.mod
robscott: Removed, thanks!

bowei:
Thanks for this PR -- funny that we have come full circle back to the original handshake idea ... well better and more nuanced :-)

I like this -- from many of the discussions, it was clear that there are many use cases on both sides of control.

docs-src/security-model.md
```diff
@@ -68,32 +68,35 @@ readable by most roles, so instead we'll focus on write access for this model.
...
-  indicates that Gateways can use this GatewayClass from any namespace. This
-  field is intentionally not a pointer because the nil behavior (no namespaces)
-  is undesirable here.
-
-* **allowedRouteNamespaces**: AllowedRouteNamespaces indicates in which
-  namespaces Routes can be selected for Gateways of this class. This is
-  restricted to the namespace of the Gateway by default.
-  
+## Limiting Namespaces Where a GatewayClass Can Be Used
+Kubernetes RBAC only allows us to grant access to create Gateways as a whole,
```
hbagdi, line 72: nit: `only allows for granting`

hbagdi:
Mostly LGTM.
One question: Let's say I've a gateway in a `gateway` namespace that selects route across the entire cluster. An app owner creates a route in her dedicated namespace, and doesn't set the field `gateways.allow` to `All` on the route. Now, things won't work, but how is the user suppose to know what to do? Is there a status that needs to be set here or do we rely on users to read the docs?

apis/v1alpha1/route_types.go
```diff
@@ -19,6 +19,38 @@ import (
...
+// GatewayAllowType specifies which Gateways should be allowed to use a Route.
+type GatewayAllowType string
+
+const (
+	// GatewayAllowAll indicates that all Gateways will be able to use this
+	// route.
+	GatewayAllowAll GatewayAllowType = "All"
+	// GatewayAllowList indicates that only Gateways that have been specified
+	// in GatewayRefs will be able to use this route.
+	GatewayAllowList GatewayAllowType = "List"
```
bowei, line 31: FromList (list is commonly used as a verb in k8s)

Seems pretty intuitive:

```
allow: All
allow: FromList
allow: SameNamespace 
```

bowei (commented):
looks good, just a nit on the naming

bowei:
(lgtm) pingback for re-review

docs-src/security-model.md
```diff
@@ -81,19 +82,23 @@ they will be controlled with configuration fields on GatewayClasses:
...
-  restricted to the namespace of the Gateway by default.
-  
+## Route Namespaces
+Service APIs allow Gateways to select Routes across multiple Namespaces.
+Although this can be remarkably powerful, this capability needs to be used
+carefully. Gateways include a `RouteNamespaces` field that allows selecting
+multiple namespaces with a label selector. By default this is limited to limit
+Gateways to select Routes within the same namespace. Additionally, Routes
+include a `Gateways` field that allows them to restrict which Gateways use them.
+By default this is restricted to Gateways in the same namespace.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

apis/v1alpha1/tcproute_types.go
```diff
@@ -23,6 +23,11 @@ import (
 type TCPRouteSpec struct {
 	// Rules are a list of TCP matchers and actions.
 	Rules []TCPRouteRule `json:"rules"`
+
+	// Gateways defines which Gateways can use this Route.
```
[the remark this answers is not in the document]
robscott: I think this could be interesting but I'd rather wait to make this change until the other larger restructuring gets in since that will directly affect where this field goes.

apis/v1alpha1/route_gateway_types.go
```diff
@@ -0,0 +1,106 @@
...
+	// +kubebuilder:default=SameNamespace
+	// +optional
+	Allow GatewayAllowType `json:"allow,omitempty"`
+	// GatewayRefs may be specified when Allow is set to "List". In that case,
+	// only Gateways referenced in this list will be allowed to use this route.
+	// +optional
+	GatewayRefs []GatewayObjectReference `json:"gatewayRefs,omitempty"`
+}
+
+// GatewayObjectReference identifies a Gateway object.
```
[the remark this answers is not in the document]
robscott: Updated to use the newer LocalObjectReference and remove this type entirely.

apis/v1alpha1/route_types.go

robscott:
@hbagdi That's a great catch, thanks! For now I've added a note saying that the Admitted condition should be set to false in that case. I'm open to how @jpeach would like to handle this but I think that could probably be further refined in a follow up PR.

docs-src/security-model.md
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
```diff
@@ -423,6 +423,39 @@ type RouteBindingSelector struct {
...
```
danehans, line 443: @roberthbailey this field needs the `omitempty` tag if it's optional.

apis/v1alpha1/gateway_types.go
```diff
@@ -423,6 +423,39 @@ type RouteBindingSelector struct {
...
+	// within the same Namespace as the Gateway should be selected.
+	//
+	// This field takes precedence over the NamespaceSelector field. That
+	// selector should only take effect when this field is set to false.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:default=true
+	OnlySameNamespace bool `json:"onlySameNamespace"`
```
danehans, line 456: Sine this field has a default, I suggest removing `optional` and adding the `omitempty` tag.

apis/v1alpha1/gateway_types.go
roberthbailey: I think you meant @robscott.

apis/v1alpha1/httproute_types.go
```diff
@@ -23,6 +23,11 @@ import (
 type HTTPRouteSpec struct {
 	// Hosts is a list of Host definitions.
 	Hosts []HTTPRouteHost `json:"hosts,omitempty"`
+
+	// Gateways defines which Gateways can use this Route.
+	// +kubebuilder:default={allow: "SameNamespace"}
+	// +optional
+	Gateways *RouteGateways `json:"gateways,omitempty"`
```
danehans, line 30: Since this field is being defaulted, you can remove `// +optional`.

apis/v1alpha1/route_types.go
```diff
@@ -19,16 +19,39 @@ import (
...
+// that Route.
+type RouteGateways struct {
+	// Allow indicates which Gateways will be allowed to use this route.
+	// Possible values are:
+	// * All: Gateways in any namespace can use this route.
+	// * FromList: Only Gateways specified in GatewayRefs may use this route.
+	// * SameNamespace: Only Gateways in the same namespace may use this route.
+	// +kubebuilder:validation:Enum=All;List;SameNamespace
+	// +kubebuilder:default=SameNamespace
+	// +optional
```
danehans, line 49: Same as https://github.com/kubernetes-sigs/service-apis/pull/285/files#r492915253.

apis/v1alpha1/route_types.go
```diff
@@ -19,16 +19,39 @@ import (
...
+// field results in preventing the selection of a Route by a Gateway, an
+// "Admitted" condition with a status of false must be set for the Gateway on
+// that Route.
+type RouteGateways struct {
+	// Allow indicates which Gateways will be allowed to use this route.
+	// Possible values are:
+	// * All: Gateways in any namespace can use this route.
+	// * FromList: Only Gateways specified in GatewayRefs may use this route.
+	// * SameNamespace: Only Gateways in the same namespace may use this route.
+	// +kubebuilder:validation:Enum=All;List;SameNamespace
```
danehans, line 47: s/List/FromList/

apis/v1alpha1/route_types.go
danehans: Consider naming the filed `Allowed` instead of `Allow`.

apis/v1alpha1/route_types.go
```diff
@@ -19,16 +19,39 @@ import (
...
+	// +kubebuilder:default=SameNamespace
+	// +optional
+	Allow GatewayAllowType `json:"allow,omitempty"`
+	// GatewayRefs may be specified when Allow is set to "List". In that case,
```
danehans, line 51: See https://github.com/kubernetes-sigs/service-apis/pull/285/files#r492916616.

apis/v1alpha1/tcproute_types.go
```diff
@@ -23,6 +23,11 @@ import (
...
+	// +kubebuilder:default={allow: "SameNamespace"}
+	// +optional
```
danehans, line 29: Same as https://github.com/kubernetes-sigs/service-apis/pull/285/files#r492915253.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -25,6 +25,11 @@ type TLSRouteSpec struct {
 	// +required
 	// +kubebuilder:validation:MinItems=1
 	Rules []TLSRouteRule `json:"rules"`
+
+	// Gateways defines which Gateways can use this Route.
+	// +kubebuilder:default={allow: "SameNamespace"}
+	// +optional
+	Gateways *RouteGateways `json:"gateways,omitempty"`
```
danehans, line 32: Same as https://github.com/kubernetes-sigs/service-apis/pull/285/files#r492919270.

apis/v1alpha1/udproute_types.go
```diff
@@ -23,6 +23,11 @@ import (
 type UDPRouteSpec struct {
 	// Rules are a list of UDP matchers and actions.
 	Rules []UDPRouteRule `json:"rules"`
+
+	// Gateways defines which Gateways can use this Route.
+	// +kubebuilder:default={allow: "SameNamespace"}
+	// +optional
+	Gateways *RouteGateways `json:"gateways,omitempty"`
```
danehans, line 30: Same as https://github.com/kubernetes-sigs/service-apis/pull/285/files#r492920007.

docs-src/security-model.md
```diff
@@ -68,32 +68,36 @@ readable by most roles, so instead we'll focus on write access for this model.
...
+a set of resources. Controllers must not provision new Gateways in namespaces
+outside this selector. The default empty selector indicates that Gateways can
+use this GatewayClass from any namespace.
+
+## Route Namespaces
+Service APIs allow Gateways to select Routes across multiple Namespaces.
+Although this can be remarkably powerful, this capability needs to be used
+carefully. Gateways include a `RouteNamespaces` field that allows selecting
+multiple namespaces with a label selector. By default this is limited to limit
+Gateways to select Routes within the same namespace. Additionally, Routes
```
danehans, line 88: > By default this is limited to limit Gateways to select Routes within the same namespace.

@robscott this sentence is a tough read. Any thoughts for improvements? Maybe:
>By default, this is limited to Routes in the same namespace as the Gateway.`

apis/v1alpha1/route_types.go
robscott: Updated, thanks! I don't mind `Allowed` but have a slight preference for `Allow` since it's a couple less characters and `allow: All` reads more naturally to me than `allowed: All`.

apis/v1alpha1/udproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/route_types.go
robscott: Updated, thanks!

apis/v1alpha1/route_types.go
robscott: Updated, thanks!

apis/v1alpha1/httproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

docs-src/security-model.md
robscott: Wow, that was not great, thanks for catching that! Updated the wording to match your suggestion.

apis/v1alpha1/route_types.go
hbagdi: Re-opening this thread, the diff still shows `List` for me.

apis/v1alpha1/gateway_types.go
```diff
@@ -423,6 +423,39 @@ type RouteBindingSelector struct {
...
+	// limited to the same Namespace as the Gateway. When true, only Routes
+	// within the same Namespace as the Gateway should be selected.
+	//
+	// This field takes precedence over the NamespaceSelector field. That
+	// selector should only take effect when this field is set to false.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:default=true
```
hbagdi, line 455: `true` by default is not the most friendly or good API design. Do we want to consider thinking through this again in a separate issue?

docs-src/security-model.md
hbagdi: Rob, maybe you didn't push a commit up. I'm still seeing the old wording.

hbagdi:
(wish Github had threads for comments)

> @hbagdi That's a great catch, thanks! For now I've added a note saying that the Admitted condition should be set to false in that case. 

That won't work unless that is Admitted:false is a default condition added to the resource. A Gateway can't definitively say that a Route is not Admitted because:
- The route might be Admitted by another Gateway (and Gateways are not aware of each other)
- It might be intentional that the user doesn't want to target that Route at this point

apis/v1alpha1/gateway_types.go
robscott: That's a good point. I'd prefer to solve this in a separate PR/issue. The only change that's happening to this type in this PR is a move from the gatewayclass_types file to gateway_types file because it's no longer used in GatewayClass.

docs-src/security-model.md
robscott: 🤦 I absolutely missed pushing that fix up. I'll leave it open until you or someone else confirms I've actually resolved it this time :)

apis/v1alpha1/route_types.go
robscott: 🤦 Fix should actually be pushed now...

robscott:
@hbagdi Thanks for catching my failure to actually push up that commit! As far as the condition, I think it's actually safe here because the Admitted condition exists per Gateway. Here's part of the godoc for `status.Gateways`:

```
Gateways is a list of the Gateways that are associated with the route, and the status of the route with respect to each of these Gateways.
```

I did think about the possibility of a user not wanting to target a Route but I think if a Gateway owner doesn't want to target a Route they should have a more refined selector. Having a status condition on the Route can help the Route owner know that a Gateway wants to use it but is currently prevented from doing so due to configuration on that Route. I'm hopeful that this will help debug any potential issues here.

hbagdi:
> I did think about the possibility of a user not wanting to target a Route but I think if a Gateway owner doesn't want to target a Route they should have a more refined selector. Having a status condition on the Route can help the Route owner know that a Gateway wants to use it but is currently prevented from doing so due to configuration on that Route.

That seems reasonable. There might be cases where multiple Gateways select a Route but route doesn't allow all but one gateway. In this case, there will be one condition that will be useful to the user and others will not be (sorta expected errors). If this gets out of hand or if users want a cleaner status, they should be using label selectors.

Consider this thread resolved!

apis/v1alpha1/gateway_types.go
hbagdi: #343

apis/v1alpha1/route_types.go
```diff
@@ -19,16 +19,39 @@ import (
...
+	// Possible values are:
+	// * All: Gateways in any namespace can use this route.
+	// * FromList: Only Gateways specified in GatewayRefs may use this route.
+	// * SameNamespace: Only Gateways in the same namespace may use this route.
+	// +kubebuilder:validation:Enum=All;FromList;SameNamespace
+	// +kubebuilder:default=SameNamespace
+	Allow GatewayAllowType `json:"allow,omitempty"`
+	// GatewayRefs may be specified when Allow is set to "FromList". In that
+	// case, only Gateways referenced in this list will be allowed to use this
+	// route.
```
hbagdi, line 52:
```suggestion
	// GatewayRefs must be specified when Allow is set to "FromList". In that
	// case, only Gateways referenced in this list will be allowed to use this
	// route. This field is ignored for other values of "Allow".
```

apis/v1alpha1/route_types.go
```diff
@@ -19,16 +19,39 @@ import (
...
+const (
+	// GatewayAllowAll indicates that all Gateways will be able to use this
+	// route.
+	GatewayAllowAll GatewayAllowType = "All"
+	// GatewayAllowFromList indicates that only Gateways that have been
+	// specified in GatewayRefs will be able to use this route.
+	GatewayAllowFromList GatewayAllowType = "FromList"
+	// GatewayAllowSameNamespace indicates that only Gateways within the same
+	// namespace will be able to use this route.
+	GatewayAllowSameNamespace GatewayAllowType = "SameNamespace"
```
hbagdi, line 34: nit: As noted in the pre-alpha feedback, we should organize these constants from least specific to most specific. We c an argue either ways on which of SameNamespace and FromList is most specific. Accept/reject this change as you see fit, no biggie.

```suggestion
	// GatewayAllowAll indicates that all Gateways will be able to use this
	// route.
	GatewayAllowAll GatewayAllowType = "All"
	// GatewayAllowSameNamespace indicates that only Gateways within the same
	// namespace will be able to use this route.
	GatewayAllowSameNamespace GatewayAllowType = "SameNamespace"
	// GatewayAllowFromList indicates that only Gateways that have been
	// specified in GatewayRefs will be able to use this route.
	GatewayAllowFromList GatewayAllowType = "FromList"
```

apis/v1alpha1/httproute_types.go
```diff
@@ -23,6 +23,10 @@ import (
 type HTTPRouteSpec struct {
 	// Hosts is a list of Host definitions.
 	Hosts []HTTPRouteHost `json:"hosts,omitempty"`
+
+	// Gateways defines which Gateways can use this Route.
+	// +kubebuilder:default={allow: "SameNamespace"}
+	Gateways *RouteGateways `json:"gateways,omitempty"`
```
hbagdi, line 29: This field is not a pointer in other routes but is a pointer here.
Most likely an oversight, if not why pointer?

hbagdi (commented):
Almost ready, thanks for working on this. Huge improvement in UX!

apis/v1alpha1/route_types.go

apis/v1alpha1/httproute_types.go
robscott: Good catch, I think this was a leftover from when this was an optional field.

apis/v1alpha1/route_types.go
robscott: This is a tough one, I'd initially thought of the `FromList` option as primarily being used to allow Gateways from other namespaces to use a Route, so I'd thought of it as less restrictive than SameNamespace. Definitely depends on the use case though. One tiny advantage of the current order is that it also happens to be alphabetical, not that that really matters.

hbagdi: Let's skip this change for now.

config/crd/bases/networking.x-k8s.io_tlsroutes.yaml
```diff
@@ -32,6 +32,40 @@ spec:
...
+                        name:
+                          description: Name is the name of the referent.
+                          type: string
+                        resource:
+                          description: Resource is the API resource name of the referent.
+                          type: string
+                      required:
+                      - group
+                      - name
+                      - resource
```
hbagdi, line 65: 1. How is a user suppose to target a Gateway in another Namespace?
2. Do we really need a Group and Resource here?

robscott: 🤦 thanks for catching this! Should be fixed now.

apis/v1alpha1/gateway_types.go
```diff
@@ -423,6 +423,38 @@ type RouteBindingSelector struct {
...
+	// limited to the same Namespace as the Gateway. When true, only Routes
+	// within the same Namespace as the Gateway should be selected.
+	//
+	// This field takes precedence over the NamespaceSelector field. That
+	// selector should only take effect when this field is set to false.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:default=true
+	OnlySameNamespace *bool `json:"onlySameNamespace,omitempty"`
```
danehans, line 455: Since defaulting is being used, make `OnlySameNamespace` a value instead of a pointer.

robscott:
@danehans Just pushed and rebased. Caught one more issue where RouteStatus was using LocalObjectReference to refer to a Gateway and have updated that to use the GatewayReference type (includes namespace, excludes group and version).

robscott:
Sorry, just pushed one more tiny addition of adding a MaxLength for Name and Namespace in GatewayRef.

PR #286, opened 2020-09-01
Updating ParametersRef docs and examples to be cluster-scoped

robscott:
This matches the scoping of the GatewayClass and matches the intention of parameters being configured with implementation specific custom resources.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/144

/cc @jpeach @hbagdi

### DIFF
--- a/apis/v1alpha1/gatewayclass_types.go
+++ b/apis/v1alpha1/gatewayclass_types.go
@@ -97,9 +97,12 @@ type GatewayClassSpec struct {
 	// +kubebuilder:default={onlySameNamespace:true}
 	AllowedRouteNamespaces RouteNamespaces `json:"allowedRouteNamespaces,omitempty"`
 
-	// ParametersRef is a controller-specific resource containing the configuration
-	// parameters corresponding to this class. This is optional if the controller
-	// does not require any additional configuration.
+	// ParametersRef is a controller-specific resource containing the
+	// configuration parameters corresponding to this class. This is optional if
+	// the controller does not require any additional configuration.
+	//
+	// Parameters resources are implementation specific custom resources. These
+	// resources must be cluster-scoped.
 	//
 	// If the referent cannot be found, the GatewayClass's "InvalidParameters"
 	// status condition will be true.
@@ -144,8 +147,8 @@ type RouteNamespaces struct {
 	OnlySameNamespace bool `json:"onlySameNamespace"`
 }
 
-// GatewayClassParametersObjectReference identifies a parameters object for a
-// gateway class within a known namespace.
+// GatewayClassParametersObjectReference identifies a cluster-scoped parameters
+// resource for a GatewayClass.
 //
 // +k8s:deepcopy-gen=false
 type GatewayClassParametersObjectReference = LocalObjectReference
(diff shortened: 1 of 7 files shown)


hbagdi:
Question: If an implementation doesn't want to use a CustomResource for this purpose and wants to resort to a ConfigMap, that is not possible? If yes, why do we want to impose such a limitation?

robscott:
@hbagdi, that's a great question. I did not add enough context here. I think I gave a better summary as I was closing my previous PR related to this (https://github.com/kubernetes-sigs/service-apis/pull/183#issuecomment-673109175)

hbagdi:
> https://github.com/kubernetes-sigs/service-apis/pull/183#issuecomment-673109175

I somehow missed that activity in my notifications.

> Secrets are potentially interesting but I don't think I've seen any use cases for GatewayClass parameters that would benefit from Secrets.

You could have an operator running in the cluster that reconciles Gateways associated with a specific GatewayClass by spinning up infra inside and outside the k8s cluster. To do this, it requires a set of credentials, stored in Secrets. Now, one can argue that the secrets then should be part of the deployment of the operator and not the GatewayClass.

I think it is safer to start with cluster-scoped references and then re-evaluate this if the above hypothetical case becomes a reality.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -98,18 +98,15 @@ type GatewayClassSpec struct {
...
-	// or specifying the empty string for both the resource and group
-	// indicates that the resource is "configmaps".  If the referent cannot
-	// be found, the GatewayClass's "InvalidParameters" status condition
-	// will be true.
+	// ParametersRef is a controller-specific resource containing the
+	// configuration parameters corresponding to this class. This is optional if
+	// the controller does not require any additional configuration.
+	//
+	// Parameters resources will be implementation specific custom resources.
+	// These resources will also be cluster-scoped.
```
hbagdi, line 106: Can we use a more stronger language here? Like "These resources must be cluster-scoped"?

hbagdi (approved):
LGTM overall.
@jpeach can you have another look and then `/lgtm` this?

robscott:
>You could have an operator running in the cluster that reconciles Gateways associated with a specific GatewayClass by spinning up infra inside and outside the k8s cluster. To do this, it requires a set of credentials, stored in Secrets. Now, one can argue that the secrets then should be part of the deployment of the operator and not the GatewayClass.

>I think it is safer to start with cluster-scoped references and then re-evaluate this if the above hypothetical case becomes a reality.

I agree with all of that. If we want to support namespace-scoped params here in the future, I think the starting point is convincing the broader sig-network community that IngressClass params should be namespace-scoped. If we can build consensus there, I think it's a no brainer to add the same capability here.

apis/v1alpha1/gatewayclass_types.go

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -150,11 +147,18 @@ type RouteNamespaces struct {
...
-// GatewayClassParametersObjectReference identifies a parameters object for a
-// gateway class within a known namespace.
+// GatewayClassParametersObjectReference identifies a cluster-scoped parameters
+// resource for a GatewayClass.
 //
 // +k8s:deepcopy-gen=false
-type GatewayClassParametersObjectReference = ConfigMapsDefaultLocalObjectReference
+type GatewayClassParametersObjectReference struct {
+	// Group is the group of the referent.
+	Group string `json:"group" protobuf:"bytes,1,opt,name=group"`
```
danehans, line 156: @robscott please remove the `protobuf` tags from `Group`, `Resource` and `Name` fields. Otherwise, LGTM.

robscott: 🤦 woops, thanks @danehans!

danehans:
@robscott thanks for addressing https://github.com/kubernetes-sigs/service-apis/pull/286#discussion_r486641898, LGTM.

hbagdi:
@robscott Needs another rebase.
/lgtm

robscott:
@hbagdi just rebased this PR, please take another look when you have time.

hbagdi:
@robscott Needs more rebasing.

robscott:
@hbagdi Rebased once more, don't think there are any more PRs open that will conflict with this.

danehans:
LGTM

PR #288, opened 2020-09-02
Fixes Install Make Target

danehans:
- `Makefilke`: Adds support for go vet and fmt that was loist when the `controller` target was removed.
- `kubebuilder.mk `: Adds the `install` target required for `make install` from the [dev guide.](https://kubernetes-sigs.github.io/service-apis/devguide/)

Fixes Issue: https://github.com/kubernetes-sigs/service-apis/issues/287

/assign @hbagdi 
/cc @jpeach @robscott

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ TOP := $(dir $(firstword $(MAKEFILE_LIST)))
 # ROOT is the root of the mkdocs tree.
 ROOT := $(abspath $(TOP))
 
-all: generate controller verify
+all: generate vet fmt verify
 
 # Run generators for protos, Deepcopy funcs, CRDs, and docs.
 #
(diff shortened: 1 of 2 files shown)


danehans:
Requires PR https://github.com/kubernetes-sigs/service-apis/pull/284

Makefile
```diff
@@ -32,6 +32,14 @@ generate:
 	$(MAKE) docs
 	$(MAKE) proto
 
+# Run go fmt against code
+fmt:
+	$(MAKE) -f kubebuilder.mk fmt
+
+# Run go vet against code
+vet:
+	$(MAKE) -f kubebuilder.mk vet
```
robscott, line 41: Maybe we can remove these from kubebuilder.mk and just have them here instead?

PR #289, opened 2020-09-03
Make CertificateRefs field in TLSConfig optional instead of repeated.

yiyangy:
An implementation may not have any SSL certificate as Kubernetes resource but, for example, use named certificate specified via Options field.


hbagdi:
This will conflict with #256 but I'll take care of that.
Another reason to make it optional is that in Passthrough TLS mode, one doesn't need to configure a TLS certificate.

robscott:
Thanks for updating this! I thought that this would actually change the CRD but apparently the +required wasn't actually making any difference here.

/lgtm
/assign @bowei

PR #298, opened 2020-09-03
Update Gateway API Type Order

### LINKED ISSUE #291: API types should appear in order they are referenced
This is a small thing but will simplify API reviews in the future. For example, Gateway should be followed immediately by Listener.

eth0xFEED:
fixes #291


robscott:
Thanks for the work on this! I just had one minor nit here. I think the preference for reviewing types files like this is depth first instead of breadth first. So as an example here, I think we'd want to read the following starting order:
- Gateway 
- GatewaySpec
- Listener
- HostnameMatch
- ProtocolType

Sorry I didn't explain that better in #291

eth0xFEED:
@robscott thanks for the example order, I will update and rebase !

Issue #295, opened 2020-09-03
Makefiles should be merged and cleaned up


robscott:
We currently have 2 makefiles, the standard `Makefile` and `kubebuilder.mk`. Although I think the original motivation was to keep consistent with kubebuilder, we've long since diverged from that pattern. This separation has resulted in a number of bugs that I think could've been easier to catch if we had a single, simpler makefile without all the cross references.

jpeach:
Fixed in #301

PR #301, opened 2020-09-08
Fix code generation targets.

### LINKED ISSUE #296: Code generation was accidentally removed from makefile
`make generate` should update generated code - it doesn't anymore. This immediate issue will be fixed by 
 https://github.com/kubernetes-sigs/service-apis/pull/263. Ideally a full fix would also tackle https://github.com/kubernetes-sigs/service-apis/issues/295.

jpeach:
* Remove the kubebuilder generated makefile, which is really just adding
  complication at this point.

* Move controller-gen directly into the `generate` target, and enable
  deepcopy and CRD code generation.

* Remove the codegen hack script, which wasn't being used and doesn't
  seem to work well with Go moduled.

* Replace codegen verification with a dumber check that uses git to
  find files that were modified by `make generate`.

This fixes #296, #295.

/cc @robscott @hbagdi


robscott:
Thanks! This all works well for me.

/lgtm
/approve

PR #303, opened 2020-09-09
Adding Guidance for Conflict Resolution

robscott:
This provides a starting point for conflict resolution, based on the [original doc](https://docs.google.com/document/d/1ArLzUF_oJr3jV-fhXKfaEBsvRXW-6p6UuaB9FGuosIg/edit) that outlined some proposed solutions to potential conflicts. Importantly this does not attempt to solve all potential conflicts, just provides guidance for some common scenarios as well as the guiding principles we identified earlier.

/cc @jpeach @hbagdi @bowei


apis/v1alpha1/gateway_types.go
```diff
@@ -281,6 +283,20 @@ type Listener struct {
 	// are compatible with the application protocol specified in
 	// the Protocol field.
 	//
+	// Multiple Routes may match a request to this Listener. If that happens,
+	// precedence MUST be given in order of the following criteria:
```
jpeach, line 287: Nit: add a newline here so this formats as a markdown unordered list.

apis/v1alpha1/gateway_types.go
```diff
@@ -281,6 +283,20 @@ type Listener struct {
...
+	// * The match with the longest combination of hostname and path.
+	// * For conflicts within a Route, the match appearing first.
+	// * The oldest Route based on creation timestamp.
+	// * The Route appearing first in alphabetical order (namespace/name).
+	//
+	// All valid portions of Routes matching a Gateway selector should be
```
jpeach, line 293: suggest:
```
All valid portions of the Routes selected by this field should be ...
```

apis/v1alpha1/gateway_types.go
```diff
@@ -281,6 +283,20 @@ type Listener struct {
 	// are compatible with the application protocol specified in
 	// the Protocol field.
 	//
+	// Multiple Routes may match a request to this Listener. If that happens,
+	// precedence MUST be given in order of the following criteria:
+	// * The match with the longest combination of hostname and path.
```
jpeach, line 288: This one is HTTP-specific, maybe we should move it to that route?

apis/v1alpha1/gateway_types.go
```diff
@@ -281,6 +283,20 @@ type Listener struct {
 	// are compatible with the application protocol specified in
 	// the Protocol field.
 	//
+	// Multiple Routes may match a request to this Listener. If that happens,
```
jpeach, line 286: I'm not quite sure how to parse "Multiple Routes may match a request to this Listener". Does it mean that a given client request could be matched to different routes?

apis/v1alpha1/gateway_types.go
hbagdi: We should replace this with 'More specific matches should be given precedence over less specific ones.'.

apis/v1alpha1/gateway_types.go
```diff
@@ -281,6 +283,20 @@ type Listener struct {
...
+	// precedence MUST be given in order of the following criteria:
+	// * The match with the longest combination of hostname and path.
+	// * For conflicts within a Route, the match appearing first.
```
hbagdi, line 289: Is this technically feasible to do in most implementations?

robscott: Good point, it's not. It seems to be a select few where order is meaningful. I've removed this.

apis/v1alpha1/gateway_types.go
robscott: This was confusing, I've tried rewording it but still not quite happy with the end result, open to options. I'm trying to communicate that multiple rules may technically match a request but only one may be selected.

apis/v1alpha1/gateway_types.go
robscott: Good catch, thanks! I wasn't sure how to fit this into HTTPRoute so I added that as an example for a more generic statement like Harry suggested. Maybe in the future it would be good to define specificity per route as well, but right now the only route we have with native matching is HTTPRoute (TCP and UDP both rely on extensionRef).

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

docs-src/concepts.md
```diff
@@ -168,10 +168,18 @@ resources:
...
+In most cases, guidance for conflict resolution is provided along with the
+documentation for fields that could have conflicts. If a conflict does not have
+a prescribed resolution, the following guiding principles should be applied:
+
+* Prefer not to break things that are working.
+* Drop as little traffic as possible.
+* Provide a consistent experience when conflicts occur.
+* Make it clear which path has been chosen when a conflict has been identified.
+* More specific matches should be given precedence over less specific ones.
+* Oldest resource wins.
```
robscott, line 182: namespace/name

apis/v1alpha1/gateway_types.go
```diff
@@ -292,6 +294,26 @@ type Listener struct {
 	// are compatible with the application protocol specified in
 	// the Protocol field.
 	//
+	// Although a client request may technically match multiple route rules,
+	// only one may ultimately receive the request. Matching precedence MUST be
```
danehans, line 298: `only one may ultimately receive the request`, clarify `one` as a route or a rule.

apis/v1alpha1/gateway_types.go
```diff
@@ -292,6 +294,26 @@ type Listener struct {
...
+	// determined in order of the following criteria:
+	//
+	// * The most specific match. For example, the most specific HTTPRoute match
+	//   would be determined by the longest matching combination of hostname and
```
danehans, line 302: s/would be/is/ (present tense).

apis/v1alpha1/gateway_types.go
```diff
@@ -292,6 +294,26 @@ type Listener struct {
...
+	//   path.
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2020-09-08 01:02:03" would be given precedence
```
danehans, line 305: s/would be/is/

apis/v1alpha1/gateway_types.go
```diff
@@ -292,6 +294,26 @@ type Listener struct {
...
+	//   over a Route with a creation timestamp of "2020-09-08 01:02:04".
+	// * The Route appearing first in alphabetical order (namespace/name). For
+	//   example, foo/bar would be given precedence over foo/baz.
```
danehans, line 308: s/would be/is/

apis/v1alpha1/gateway_types.go
```diff
@@ -292,6 +294,26 @@ type Listener struct {
...
+	//
+	// All valid portions of the Routes selected by this field should be
```
danehans, line 310: s/the Routes/a Route/ to match the following sentence > `Invalid portions of a Route`.

docs-src/concepts.md
```diff
@@ -168,10 +168,18 @@ resources:
...
 between cluster ops and application developers) can result in conflicts in the
 configuration. For example, two application teams may inadvertently submit
-configuration for the same HTTP path. There are several different strategies for
-handling this:
-
-* TODO
+configuration for the same HTTP path.
+
+In most cases, guidance for conflict resolution is provided along with the
+documentation for fields that could have conflicts. If a conflict does not have
```
danehans, line 174: s/could have conflicts/may have conflicts/

apis/v1alpha1/gateway_types.go
yiyangy, line 308: What if multiple HTTPRouteHost within a HTTPRoute match the same hostname?

docs-src/concepts.md
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go
hbagdi: Then the match should be based on the most specific match.

danehans: > Then the match should be based on the most specific path/header match.

@hbagdi is this documented? If not where should this be documented? I reviewed the docs for `HTTPRouteHost` and `HTTPRouteMatch` and could not come to this conclusion.

danehans:
LGTM, although https://github.com/kubernetes-sigs/service-apis/pull/303#discussion_r488149164 should be documented somewhere.

apis/v1alpha1/gateway_types.go
yiyangy: I asked because if the precedence cannot be determined from the most specific match within a HTTPRoute, the second and third criteria do not apply here. But I suppose within a HTTPRoute, such conflict happens rarely.

hbagdi: Couple of updates:
- #337 to track @danehans's comment
- I'm not sure I completely understand you, @yiyangy. Today in our weekly sync, we discussed removing multiple hosts from HTTPRoute to simplify the resource and that might solve the problem you are trying to point out here. If you could share examples of what you mean here, that would be very helpful.

hbagdi: While reviewing, I found the following in gateway_types.go:
```
	// * The most specific match. For example, the most specific HTTPRoute match
	//   is determined by the longest matching combination of hostname and path.
```

@danehans Does that help?

docs-src/concepts.md
```diff
@@ -168,10 +168,21 @@ resources:
...
+configuration for the same HTTP path.
+
+In most cases, guidance for conflict resolution is provided along with the
+documentation for fields that may have a conflict. If a conflict does not have a
+prescribed resolution, the following guiding principles should be applied:
+
+* Prefer not to break things that are working.
+* Drop as little traffic as possible.
+* Provide a consistent experience when conflicts occur.
+* Make it clear which path has been chosen when a conflict has been identified.
```
hbagdi, line 180: For my understanding, is this about setting the appropriate status?

docs-src/concepts.md
```diff
@@ -168,10 +168,21 @@ resources:
...
+* More specific matches should be given precedence over less specific ones.
+* The resource with the oldest creation timestamp wins.
+* If everything else is equivalent, precedences should be given to the resource
```
hbagdi, line 183: This might be a nit but can we `s/equivalent/equivalent (including creation timestamp)`?
The reasoning here is that this is very unlikely and users and implementers should understand this clearly, else we risk everyone abusing prioritization based on such an alphabetical order.

apis/v1alpha1/gateway_types.go
robscott: I think this is going away with the latest proposal to have to make `HTTPRouteHost` `HTTPRouteSpec` (no more list of hosts per Route). That's covered by https://github.com/kubernetes-sigs/service-apis/pull/336. As long as we're OK with that approach I think we can avoid anything related to conflict resolution here.

docs-src/concepts.md
robscott: Updated to clarify that this should be done by setting conditions on status wherever possible.

docs-src/concepts.md
robscott: Good call, updated.

robscott:
@hbagdi I think I've resolved all the feedback now, PTAL.

apis/v1alpha1/gateway_types.go
```diff
@@ -108,8 +108,10 @@ type GatewaySpec struct {
 	Addresses []GatewayAddress `json:"addresses"`
 }
 
-// Listener embodies the concept of a logical endpoint where a
-// Gateway can accept network connections.
+// Listener embodies the concept of a logical endpoint where a Gateway can
+// accept network connections. Each listener in a Gateway must have a unique
+// combination of Hostname, Port, and Protocol. This will be enforced by a
+// validating webhook.
```
danehans, line 114: @robscott @bowei will this validating webhook be part of the service-apis repo?

apis/v1alpha1/gateway_types.go
danehans: > That's covered by #336. As long as we're OK with that approach I think we can avoid anything related to conflict resolution here.

ack

danehans:
LGTM

apis/v1alpha1/gateway_types.go
robscott: Yep, that's the plan. There are a number of validations that simply can't be accomplished with kubebuilder annotations at this point. I believe this would be OK post alpha API launch though.

PR #305, opened 2020-09-09
Standardize common route status fields.

### LINKED ISSUE #304: Standardize the status structure across route types
The `HTTPRouteStatus` is different from the status fields of other route types. It probably make sense for all the types to have the same status fields (or at least for all the types to have the `RouteGatewayStatus` field to hold the `Admitted` condition).

jpeach:
All the route types share a common need to expose when they are
successfully bound to a Gateway. To address this, we can hoist the
`HTTPRouteStatus` fields into a generic struct and inline it into
all the route types.

This fixes #304 

/cc @Miciah @hbagdi


apis/v1alpha1/route_types.go
```diff
@@ -0,0 +1,85 @@
...
+type RouteConditionType string
+
+const (
+	// ConditionRouteAdmitted indicates whether the route has been admitted
+	// or rejected by a Gateway, and why.
+	ConditionRouteAdmitted RouteConditionType = "Admitted"
+)
+
+// RouteCondition is a status condition for a given route.
+type RouteCondition struct {
```
robscott, line 45: Will this change to metav1.Condition once this merges? https://github.com/kubernetes-sigs/service-apis/pull/302

jpeach: Yes

apis/v1alpha1/route_types.go
```diff
@@ -0,0 +1,85 @@
...
+	GatewayRef GatewayObjectReference `json:"gatewayRef"`
+	// Conditions describes the status of the route with respect to the
+	// Gateway.  For example, the "Admitted" condition indicates whether the
+	// route has been admitted or rejected by the Gateway, and why.  Note
+	// that the route's availability is also subject to the Gateway's own
+	// status conditions and listener status.
+	Conditions []RouteCondition `json:"conditions,omitempty"`
+}
+
+// RouteStatus defines the observed state that is common to all route types.
```
hbagdi, line 76: Since there is no better place to add this, can we add in here that all Route types SHOULD use this in their status?
The existing doc comment suggests that this exists because things are common. I think we should instead make it clear that structure of the status must be the same for all types of routes.

hbagdi (commented):
Apart from one nitpick, this looks ready.
We will need to take care of PR ordering with the ongoing metav1.Condition refactor.

apis/v1alpha1/route_types.go
jpeach: I updates this to say that it's required for all routes. WDYT?

hbagdi: Now I can live with it. ;)

robscott:
Thanks!

/approve

PR #308, opened 2020-09-10
Fixing default value for GatewayClass conditions

robscott:
With the transition to the metav1.Condition type more values are required as part of a condition. The previous default value made the GatewayClass CRD invalid.

/cc @jpeach @hbagdi

### DIFF
--- a/apis/v1alpha1/gatewayclass_types.go
+++ b/apis/v1alpha1/gatewayclass_types.go
@@ -40,7 +40,7 @@ type GatewayClass struct {
 	// Spec for this GatewayClass.
 	Spec GatewayClassSpec `json:"spec,omitempty"`
 	// Status of the GatewayClass.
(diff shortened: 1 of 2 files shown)


danehans:
@robscott LGTM, but why is `GatewayClass` the only resource with default status conditions?

robscott:
@danehans That's a great question. I believe this came from the idea that conditions would always be present, but it clearly doesn't line up with what we're doing on other resources. I'd be open to removing this entirely in the future. Interested in what @jpeach is thinking about for conditions on the class.

jpeach:
@robscott should the message field really end with a period?

robscott:
@jpeach Thanks for catching that! It does look like most messages do not end with a period - removed it here.

jpeach:
> @robscott LGTM, but why is `GatewayClass` the only resource with default status conditions?

I agree that we should apply a reasonable default to other resources.

/lgtm

PR #309, opened 2020-09-10
tls: make CertificateRef and Options optional

hbagdi:
- For the "Passthrough" mode, certificateRef is not necessary as the
Gateway doesn't terminate the TLS session
- Options are meant as an extension point and should be optional.


robscott:
Thanks! Maybe @jpeach or @danehans can add a LGTM?

/approve

PR #312, opened 2020-09-10
httproute: rename Action to Forward

hbagdi:
/cc @jpeach @robscott


hbagdi:
Once this PR merges, I'll make the same change to other Route types as well.

robscott:
Thanks!

/lgtm
/approve

I think it's fine to combine the changes to all the other routes into a single follow up PR for simplicity.

robscott:
To clarify the quick approval here, I think this matches discussions we've had on previous calls. If that's not the case, let me know.

PR #317, opened 2020-09-11
simplify CRD organization

hbagdi:
This patch simplifies the organization of generated CRDs by moving all code into a single top-level `crd` directory.


hbagdi:
This might be a controversial change as it could possibly break downstream scripts.
The goal here is to remove unnecessary directories and keep navigation in the directory as dumb as possible.

jpeach:
@hbagdi Can we do `config/crd` instead?

hbagdi:
> Can we do config/crd instead?

I don't mind that either. Is that a standard pattern in the kube community?

jpeach:
> > Can we do config/crd instead?
> 
> I don't mind that either. Is that a standard pattern in the kube community?

"standard" might be a bit strong, but it's pretty common

hbagdi:
> "standard" might be a bit strong, but it's pretty common

In that case, should we leave it as is?

robscott:
I think `config/crd` comes from kubebuilder - cluster-api follows that pattern. Controllers like nginx-ingress or cert-manager both have a `deploy` directory, and external-dns uses a `kustomize` directory. Given that cluster-api is the most similar to this project, it probably makes sense to keep following that pattern. I think the value of the `bases` directory is that you can easily `kubectl apply -f config/crd/bases` whereas I think if you had `kustomization.yaml` in the same directory as the CRDs you might end up with an apply error. I'm not against finding a way to simplify things, but I think each part of the existing structure may have a purpose, as annoying as all the directories are.

hbagdi:
That makes sense. Thanks for the explanation @robscott!
I think I went overboard with simplifying the config directory here, let's keep things as they are.
/close

PR #319, opened 2020-09-11
use RouteStatus for TLSRoute

hbagdi:
8c25d9e was merged in while this resource was developed. This patch
ensures TLSRoute uses the same status structure as other routes.


robscott:
Thanks!

/lgtm
/approve

PR #320, opened 2020-09-11
remove filters from TCP/TLS/UDP routes

hbagdi:
We added filters to other routes because we share the same forwarding
type. This was an error due to type sharing.


apis/v1alpha1/forwarding_target_types.go
```diff
@@ -91,3 +91,55 @@ type TargetWeight int32
 //
 // +k8s:deepcopy-gen=false
 type ForwardToTargetObjectReference = ServicesDefaultLocalObjectReference
+
+// GenericForwardToTarget identifies a target object within a known namespace.
```
hbagdi, line 95: Any better naming suggestions are most welcome here!

Question, should we consider having a separate Forwarding target type for each Route type?

apis/v1alpha1/forwarding_target_types.go
```diff
@@ -91,3 +91,55 @@ type TargetWeight int32
...
+type GenericForwardToTarget struct {
+	// TargetRef is an object reference to forward matched requests to.
+	// The resource may be "services" (omit or use the empty string for the
+	// group), or an implementation may support other resources (for
+	// example, resource "myroutetargets" in group "networking.acme.io").
+	// Omitting or specifying the empty string for both the resource and
+	// group indicates that the resource is "services".  If the referent
```
robscott, line 102: I think we need better wording here to indicate that there are defaults for both group and resource separately.

apis/v1alpha1/forwarding_target_types.go
```diff
@@ -91,3 +91,55 @@ type TargetWeight int32
...
+	// Support: Core (Kubernetes Services)
+	// Support: Implementation-specific (Other resource types)
+	//
+	TargetRef ForwardToTargetObjectReference `json:"targetRef"`
+
+	// TargetPort specifies the destination port number to use for the TargetRef.
+	// If unspecified and TargetRef is a Service object consisting of a single
+	// port definition, that port will be used. If unspecified and TargetRef is
+	// a Service object consisting of multiple port definitions, an error is
+	// surfaced in status.
```
robscott, line 115: Do we know what error that will be?

apis/v1alpha1/forwarding_target_types.go
robscott: I think a shared type is fine until they diverge more. I've been trying to think of a better prefix than "Generic" here but haven't come up with anything.

robscott (commented):
This mostly LGTM, just a couple minor nits on things that aren't exactly new, just issues copied from the existing forwardToTarget. If you'd rather handle them in a follow up PR that's also fine.

hbagdi:
>  just a couple minor nits on things that aren't exactly new, just issues copied from the existing forwardToTarget. If you'd rather handle them in a follow up PR that's also fine.

robscott:
Thanks!

/lgtm
/approve

hbagdi:
Rebase. @robscott PTAL.

PR #324, opened 2020-09-11
Update Gateway condition types.

jpeach:
Update the condition types and reasons for the Gateway condition type.
Gateways should now expose their state using the `Scheduled` and `Ready`
conditions, providing additional information in the reason.

See conditions vocabulary doc: https://docs.google.com/document/d/17WGAL5jjjGHsmDCeMstZ3L5fFp6QGPHrIplQ9LPZ91k

/cc @robscott @danehans @hbagdi

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ vet:
 
 # Install CRD's and example resources to a pre-existing cluster.
 .PHONY: install
-install: manifests crd example
+install: crd example
 
 # Install the CRD's to a pre-existing cluster.
 .PHONY: crd
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -29,7 +29,9 @@ type Gateway struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   GatewaySpec   `json:"spec,omitempty"`
+	Spec GatewaySpec `json:"spec,omitempty"`
+
+	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
 	Status GatewayStatus `json:"status,omitempty"`
 }
 
@@ -505,14 +507,30 @@ type GatewayStatus struct {
 	//
 	// These addresses should all be of type "IPAddress".
 	//
-	// +required
+	// +optional
 	Addresses []GatewayAddress `json:"addresses"`
 
 	// Conditions describe the current conditions of the Gateway.
-	// +optional
+	//
+	// Implementations should prefer to express Gateway conditions
+	// using the `GatewayConditionType` and `GatewayConditionReason`
+	// constants so that operators and tools can converge on a common
+	// vocabulary to describe Gateway state.
+	//
+	// Known condition types are:
+	//
+	// * "Scheduled"
+	// * "Ready"
+	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:default={{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
 	// Listeners provide status for each unique listener port defined in the Spec.
+	//
 	// +optional
 	Listeners []ListenerStatus `json:"listeners,omitempty"`
 }
@@ -522,30 +540,88 @@ type GatewayStatus struct {
 // field.
 type GatewayConditionType string
 
+// GatewayConditionReason defines the set of reasons that explain
+// why a particular Gateway condition type has been raised.
+type GatewayConditionReason string
+
 const (
-	// ConditionForbiddenNamespaceForClass indicates that this Gateway is in
-	// a namespace forbidden by the GatewayClass.
-	ConditionForbiddenNamespaceForClass GatewayConditionType = "ForbiddenNamespaceForClass"
+	// GatewayConditionScheduled indicates whether the controller
+	// managing the Gateway has scheduled the Gateway to the
+	// underlying network infrastructure.
+	//
+	// Possible reasons for this condition to be false are:
+	//
+	// * "NotReconciled"
+	// * "NamespaceForbidden"
+	// * "NoSuchGatewayClass"
+	// * "NoResources"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.`
+	GatewayConditionScheduled GatewayConditionType = "Scheduled"
 
-	// ConditionGatewayNotScheduled indicates that the Gateway has not been
-	// scheduled.
-	ConditionGatewayNotScheduled GatewayConditionType = "GatewayNotScheduled"
+	// GatewayReasonNotReconciled is used when the Gateway is
+	// not scheduled because it recently been created and no
+	// controller has reconciled it yet.
+	GatewayReasonNotReconciled GatewayConditionReason = "NotReconciled"
 
-	// ConditionListenersNotReady indicates that at least one of the specified
-	// listeners is not ready. If this condition has a status of True, a more
-	// detailed ListenerCondition should be present in the corresponding
-	// ListenerStatus.
-	ConditionListenersNotReady GatewayConditionType = "ListenersNotReady"
+	// GatewayReasonNamespaceForbidden is used when the Gateway is
+	// not scheduled because the controller's configuration
+	// forbids it from scheduling objects from that namespace.
+	GatewayReasonNamespaceForbidden GatewayConditionReason = "NamespaceForbidden"
 
-	// ConditionInvalidListeners indicates that at least one of the specified
-	// listeners is invalid. If this condition has a status of True, a more
-	// detailed ListenerCondition should be present in the corresponding
-	// ListenerStatus.
-	ConditionInvalidListeners GatewayConditionType = "InvalidListeners"
+	// GatewayReasonNoSuchGatewayClass is used when the Gateway is
+	// not scheduled because there is no controller that recognizes
+	// the GatewayClassName. This reason should only be set by
+	// a controller that has cluster-wide visibility of all the
+	// installed GatewayClasses.
+	GatewayReasonNoSuchGatewayClass GatewayConditionReason = "NoSuchGatewayClass"
 
-	// ConditionInvalidAddress indicates one or more of the
-	// Gateway's Addresses is invalid or could not be assigned.
-	ConditionInvalidAddress GatewayConditionType = "InvalidAddress"
+	// GatewayReasonNoResources is used when the Gateway is
+	// not scheduled because no infrastructure resources are
+	// available for this Gateway.
+	GatewayReasonNoResources GatewayConditionReason = "NoResources"
+)
+
+const (
+	// GatewayConditionReady indicates whether the Gateway is able
+	// to serve traffic. Note that this does not indicate that the
+	// Gateway configuration is current or even complete (e.g. the
+	// controller may still not have reconciled the latest version,
+	// or some parts of the configuration could be missing).
+	//
+	// If both the "ListenersNotValid" and "ListenersNotReady"
+	// reasons are true, the Gateway controller should prefer the
+	// "ListenersNotValid" reason.
+	//
+	// Possible reasons for this condition to be false are:
+	//
+	// * "ListenersNotValid"
+	// * "ListenersNotReady"
+	// * "AddressNotAssigned"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.`
+	GatewayConditionReady GatewayConditionType = "Ready"
+
+	// GatewayReasonListenersNotValid is used when one or more
+	// Listeners have an invalid or unsupported configuration
+	// and cannot be configured on the Gateway.
+	GatewayReasonListenersNotValid GatewayConditionReason = "ListenersNotValid"
+
+	// GatewayReasonListenersNotReady is used when one or more
+	// Listeners are not ready to serve traffic.
+	GatewayReasonListenersNotReady GatewayConditionReason = "ListenersNotReady"
+
+	// GatewayReasonAddressNotAssigned is used when the requested
+	// address has not been assigned to the Gateway. This reason
+	// can be used to express a range of circumstances, including
+	// (but not limited to) IPAM address exhaustion, invalid
+	// or unsupported address requests, or a named address not
+	// being found.
+	GatewayReasonAddressNotAssigned GatewayConditionReason = "AddressNotAssigned"
 )
 
 // ListenerStatus is the status associated with a Listener port.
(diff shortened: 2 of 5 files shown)


apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	//
+	// Known condition types are:
+	//
+	// * "Scheduled"
+	// * "Ready"
+	//
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
```
robscott, line 506: Nice details here, we should specify these in more places. It looks like the `+patchMergeKey` and `+patchStrategy` labels are not being translated into anything meaningful for the CRD spec though? Maybe we should only rely on the `+listType` and `+listMapKey` type.

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// +kubebuilder:default={{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
```
robscott, line 507: I think "NotReconciled" may be a better reason for the default GatewayClass condition as well.

Makefile
```diff
@@ -45,7 +45,7 @@ vet:
 
 # Install CRD's and example resources to a pre-existing cluster.
 .PHONY: install
-install: manifests crd example
+install: crd example
```
robscott, line 48: Good catch, I missed this when we removed it earlier.

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	GatewayReasonNoResources GatewayConditionReason = "NoResources"
+)
 
-	// ConditionInvalidAddress indicates one or more of the
-	// Gateway's Addresses is invalid or could not be assigned.
-	ConditionInvalidAddress GatewayConditionType = "InvalidAddress"
+const (
+	// GatewayConditionReady indicates whether the Gateway is able
+	// to serve traffic. Note that this does not indicate that the
+	// Gateway configuration is current or even complete.
```
robscott, line 578: > this does not indicate that the Gateway configuration is current or even complete.

What does that mean? Is this meant to indicate that a Gateway is ready to receive traffic but may not represent the most up to date configuration here? Or that a Gateway is ready but the configuration here is invalid?

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// GatewayReasonNoSuchGatewayClass is used when the Gateway is
+	// not scheduled because there is no controller that recognizes
+	// the gateway class name. This reason should only be set by
+	// a controller that has cluster-wide visibility of all the
+	// installed gateway classes.
+	GatewayReasonNoSuchGatewayClass GatewayConditionReason = "NoSuchGatewayClass"
+
+	// GatewayReasonNoResources is used when the Gateway is
+	// not scheduled because no infrastructure resources are
+	// available for this Gateway.
```
robscott, line 571:
```suggestion
	// not scheduled because no infrastructure resources
	// available for this Gateway.
```

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+
+	// GatewayReasonNamespaceForbidden is used when the Gateway is
+	// not scheduled because, while the Gateway is visible to
+	// the managing controller, the controller's configuration
+	// forbids is from scheduling objects from that namespace.
+	GatewayReasonNamespaceForbidden GatewayConditionReason = "NamespaceForbidden"
+
+	// GatewayReasonNoSuchGatewayClass is used when the Gateway is
+	// not scheduled because there is no controller that recognizes
+	// the gateway class name. This reason should only be set by
```
robscott, line 564:
```suggestion
	// the GatewayClassName. This reason should only be set by
```

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// a controller that has cluster-wide visibility of all the
+	// installed gateway classes.
```
robscott, line 566:
```suggestion
	// installed GatewayClasses.
```

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// controller has reconciled it yet.
+	GatewayReasonNotReconciled GatewayConditionReason = "NotReconciled"
+
+	// GatewayReasonNamespaceForbidden is used when the Gateway is
+	// not scheduled because, while the Gateway is visible to
+	// the managing controller, the controller's configuration
+	// forbids is from scheduling objects from that namespace.
+	GatewayReasonNamespaceForbidden GatewayConditionReason = "NamespaceForbidden"
+
+	// GatewayReasonNoSuchGatewayClass is used when the Gateway is
```
robscott, line 562: What happens if multiple controllers are trying to write to this condition? Any chance we'd end up with some kind of race here?

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	GatewayConditionScheduled GatewayConditionType = "Scheduled"
+
+	// GatewayReasonNotReconciled is used when the Gateway is
+	// not scheduled because it recently been created and no
+	// controller has reconciled it yet.
+	GatewayReasonNotReconciled GatewayConditionReason = "NotReconciled"
+
+	// GatewayReasonNamespaceForbidden is used when the Gateway is
+	// not scheduled because, while the Gateway is visible to
+	// the managing controller, the controller's configuration
```
robscott, line 558: I think it might be simpler to either not mention the Gateway being visible to the managing controller or save that for a second sentence.
```suggestion
	// not scheduled because the controller's configuration
```

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// forbids is from scheduling objects from that namespace.
```
robscott, line 559:
```suggestion
	// forbids it from scheduling objects from that namespace.
```

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// Valid reasons for this condition to be false are:
+	//
+	// * "NotReconciled"
+	// * "NamespaceForbidden"
+	// * "NoSuchGatewayClass"
+	// * "NoResources"
+	GatewayConditionScheduled GatewayConditionType = "Scheduled"
+
+	// GatewayReasonNotReconciled is used when the Gateway is
+	// not scheduled because it recently been created and no
```
robscott, line 552:
```suggestion
	// not scheduled because it has recently been created and no
```

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+const (
+	// GatewayConditionReady indicates whether the Gateway is able
+	// to serve traffic. Note that this does not indicate that the
+	// Gateway configuration is current or even complete.
+	//
+	// If both the "ListenersNotValid" and "ListenersNotReady"
+	// reasons are true, the Gateway controller should prefer the
+	// "ListenersNotValid" reason.
+	//
+	// Valid reasons for this condition to be false are:
```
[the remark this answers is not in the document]
jpeach: > Is this intended to be a complete list of valid reasons? 

Just the valid reasons defined in the common API.

> Could a controller set reasons beyond these values?

Yes a controller can use its own reasons, though I would discourage that personally.

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
-	// Gateway's Addresses is invalid or could not be assigned.
-	ConditionInvalidAddress GatewayConditionType = "InvalidAddress"
+const (
+	// GatewayConditionReady indicates whether the Gateway is able
+	// to serve traffic. Note that this does not indicate that the
+	// Gateway configuration is current or even complete.
+	//
+	// If both the "ListenersNotValid" and "ListenersNotReady"
+	// reasons are true, the Gateway controller should prefer the
+	// "ListenersNotValid" reason.
```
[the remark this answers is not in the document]
jpeach: > Maybe we should define an order these should be assigned in and the implementation should choose the earliest applicable reason in that list. As an example, ListenersNotValid likely (though not always) also means ListenersNotReady and AddressNotAssigned.

Yeh that's a possibility. In general the reasons are carefully chosen to subsume each other. You can't be ready unless you are also valid for example. Not sure there is a clear benefit to defining a priority order though.

apis/v1alpha1/gateway_types.go
jpeach: > Nice details here, we should specify these in more places. It looks like the `+patchMergeKey` and `+patchStrategy` labels are not being translated into anything meaningful for the CRD spec though? Maybe we should only rely on the `+listType` and `+listMapKey` type.

Huh, maybe we need a different CRD generator?

apis/v1alpha1/gateway_types.go
jpeach: You can be ready but not current (if the spec changed), but also partial in some senses. For example, if you can't resolve routes or you can resolve the CA secrets for some of the routes.

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
-	ConditionForbiddenNamespaceForClass GatewayConditionType = "ForbiddenNamespaceForClass"
+// GatewayConditionReason defines the set of reasons that explain
+// why a particular Gateway condition type has been raised.
+type GatewayConditionReason string
 
+const (
 	// ConditionGatewayNotScheduled indicates that the Gateway has not been
 	// scheduled.
-	ConditionGatewayNotScheduled GatewayConditionType = "GatewayNotScheduled"
+	ConditionGatewayNotScheduled GatewayConditionType = "NotScheduled"
```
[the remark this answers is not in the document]
danehans: +1, according to https://github.com/kubernetes-sigs/service-apis/pull/324/files#diff-2927d10c91558c2273bc3a8ea6464b85R493-R501, the types should be "Scheduled" and "Ready".

apis/v1alpha1/gateway_types.go
danehans: Are field tags required to specify the patch strategy? See https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1@v0.19.0?tab=doc#Condition and https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch

apis/v1alpha1/gateway_types.go
hbagdi: Also, any reason for defining the default twice? Once here and then on line 32 where the GatewayStatus struct is used?

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
...
+	// GatewayReasonNotReconciled is used when the Gateway is
+	// not scheduled because it recently been created and no
+	// controller has reconciled it yet.
+	GatewayReasonNotReconciled GatewayConditionReason = "NotReconciled"
+
+	// GatewayReasonNamespaceForbidden is used when the Gateway is
+	// not scheduled because, while the Gateway is visible to
+	// the managing controller, the controller's configuration
+	// forbids is from scheduling objects from that namespace.
+	GatewayReasonNamespaceForbidden GatewayConditionReason = "NamespaceForbidden"
```
hbagdi, line 560: > the controller's configuration forbids it from scheduling objects from that namespace

Is the controller configuration that is forbidding it or the GatewayClass spec?

apis/v1alpha1/gateway_types.go
robscott: I had to do the same thing for GatewayClass. If someone created a GatewayClass with a `{}` status, I believe the default here would be required. Unfortunately the defaulting logic is pretty basic here so it involves a bit of repetition.

apis/v1alpha1/gateway_types.go
```diff
@@ -483,47 +485,125 @@ type GatewayStatus struct {
 	//
 	// These addresses should all be of type "IPAddress".
 	//
-	// +required
+	// +optional
```
danehans, line 488: Instead of `Addresses` being optional, is their value in keeping it required and adding an "Unknown" `AddressType`?

apis/v1alpha1/gateway_types.go
jpeach: This isn't special. If multiple controllers want to write this condition, they will race and interfere with each other.

apis/v1alpha1/gateway_types.go
jpeach: The former. There is a designated reason for the latter, and controllers with a GatewayClass should not be touching Gateways of different classes.

apis/v1alpha1/gateway_types.go
jpeach: I don't think that's necessary, but I'd be happy to argue my case in a separate issue :D

apis/v1alpha1/gateway_types.go
robscott: That's kinda what I'd been thinking as well. Although it seems unlikely, it seems like one poorly behaving controller here could cause any other good controllers to spend an awful lot of time updating this condition. Unless we can find a way to set this condition safely I'd be tempted to remove it for now.

apis/v1alpha1/gateway_types.go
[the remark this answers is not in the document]
jpeach: Added text to make this explicit.

apis/v1alpha1/gateway_types.go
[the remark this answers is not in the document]
jpeach: I don't think this is needed. Let's see how implementations handle it.

apis/v1alpha1/gateway_types.go
jpeach: https://github.com/kubernetes-sigs/service-apis/issues/342

apis/v1alpha1/gateway_types.go
jpeach: I don't think that I see the problem. It's already possible for badly behaved controllers to whack each others fields in bad ways. A well-behaved controller would never set this reason on a condition or even touch a gateway that didn't match it class. I think that the service-apis project should provide a controller that sets this.

robscott: That's an interesting idea, do you think there are other conditions a common/shared controller would be responsible for setting? I'm hesitant to add one more component to maintain here, but if it seems like it might be worth it if it helped avoid potential race conditions like this.

jpeach: > That's an interesting idea, do you think there are other conditions a common/shared controller would be responsible for setting?

I think the other thing that has been discussed is a validating admission controller.

>  I'm hesitant to add one more component to maintain here, but if it seems like it might be worth it if it helped avoid potential race conditions like this.

I'm not yet convinced there is a real race here. This reason is intended for a dedicated controller, since actual ingress controllers should only even be touching gateways of their own class. I updated the description to add some more discouragement.

robscott:
Thanks!

/approve

PR #325, opened 2020-09-11
Sort gateway types in the order of introduction

### LINKED ISSUE #291: API types should appear in order they are referenced

eth0xFEED:
fixes #291


robscott:
Thanks!

/lgtm
/approve

PR #233, opened 2020-07-10
Expands on GatewayClass.spec.controller

bowei:
Adds more details on how to interpret the controller field, give
an example around how versioning will work.

Fixes: https://github.com/kubernetes-sigs/service-apis/issues/229

PR #327, opened 2020-09-11
Expands on GatewayClass.spec.controller

hbagdi:
Adds more details on how to interpret the controller field, give
an example around how versioning will work.

Fixes: https://github.com/kubernetes-sigs/service-apis/issues/229

Co-authored-by: Harry Bagdi <harrybagdi@gmail.com>

### DIFF
--- a/docs-src/concepts.md
+++ b/docs-src/concepts.md
@@ -343,6 +343,30 @@ status:
     Message: "foobar" is an FooBar.
 ```
 
+### GatewayClass controller selection
+
+The `GatewayClass.spec.controller` field is used to determine whether
+or not a given `GatewayClass` is managed by the controller.
+This format of this field is opaque and specific to a particular controller.
+Which GatewayClass is selected by a given controller field depends on how
+various controller(s) in the cluster interpret this field.
+
+It is RECOMMENDED that controller authors/deployments make their
+selection unique by using a domain / path combination under their
+administrative control (e.g. controller managing of all `controller`s
+starting with `acme.io` is the owner of the `acme.io` domain) to avoid
+conflicts.
+
+Controller versioning can be done by encoding the version of a
+controller into the path portion. An example scheme could be (similar
+to container URIs):
+
+```text
+acme.io/gateway/v1   // Use version 1
+acme.io/gateway/v2   // Use version 2
+acme.io/gateway      // Use the default version
+```
+
 ## Gateway
 
 A `Gateway` is 1:1 with the life cycle of the configuration of
--- a/docs/concepts/index.html
+++ b/docs/concepts/index.html
@@ -375,6 +375,13 @@
   
 </li>
         
+          <li class="md-nav__item">
+  <a href="#gatewayclass-controller-selection" class="md-nav__link">
+    GatewayClass controller selection
+  </a>
+  
+</li>
+        
       </ul>
     </nav>
   
@@ -717,6 +724,13 @@
           <li class="md-nav__item">
   <a href="#gatewayclass-status" class="md-nav__link">
     GatewayClass status
+  </a>
+  
+</li>
+        
+          <li class="md-nav__item">
+  <a href="#gatewayclass-controller-selection" class="md-nav__link">
+    GatewayClass controller selection
   </a>
   
 </li>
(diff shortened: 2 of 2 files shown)


hbagdi:
/assign @robscott 
I took @bowei 's PR and udpated it to get it in faster.

Close #233

docs-src/concepts.md
```diff
@@ -343,6 +343,30 @@ status:
...
+starting with `acme.io` is the owner of the `acme.io` domain) to avoid
+conflicts.
+
+Controller versioning can be done by encoding the version of a
+controller into the path portion. An example scheme could be (similar
+to container URIs):
+
+```text
+acme.io/gateway:1.14   // Use version 1.14
+acme.io/gateway:1.15   // Use version 1.15
```
robscott, line 366: I believe these would be invalid names if we were using the same validation as Ingress. Each path segment there must conform relatively strict validation: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L177.

hbagdi: Do we do validations on the Ingress class name?

robscott: Yeah our validations are quite strict on Ingress and really and k8s resources. It won't be quite as straightforward to do that here until we have a webhook. Until then, I'm hesitant to suggest patterns that would be invalid for IngressClass - ideally we'd want the same controller names to be valid for both. As a slight variation of this, maybe we could suggest `acme.io/gateway/v1` and `acme.io/gateway/v2`? I believe that would also be a valid IngressClass controller name. (That validation relies on the [IsDomainPrefixedPath](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L131) validation function).

jpeach: +1 on showing a validation safe example. Best to mention that future spec releases will validate this field to the a HTTP URL.

hbagdi: TIL and makes sense. Updated, PTAL.

docs-src/concepts.md
```diff
@@ -343,6 +343,30 @@ status:
...
+administrative control (e.g. controller managing of all `controller`s
+starting with `acme.io` is the owner of the `acme.io` domain) to avoid
+conflicts.
+
+Controller versioning can be done by encoding the version of a
+controller into the path portion. An example scheme could be (similar
+to container URIs):
+
+```text
+acme.io/gateway/v1.14   // Use version 1.14
```
robscott, line 365: Even the `.` is invalid according to the upstream Kubernetes validation. Each segment of the path must be an DNS 1123 label as defined here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L177

robscott: My suggestion would be to just use v1 and v2 for simplicity in this example. In practice you could also express this as `v1-14`, though that's not ideal.

hbagdi: I read the following:

```
// IsDomainPrefixedPath checks if the given string is a domain-prefixed path
// (e.g. acme.io/foo). All characters before the first "/" must be a valid
// subdomain as defined by RFC 1123. All characters trailing the first "/" must
// be valid HTTP Path characters as defined by RFC 3986.
func IsDomainPrefixedPath(fldPath *field.Path, dpPath string) field.ErrorList {
```
and updated the doc as such.

We shouldn't do such a validation just for the sake of it and rather relax validation to include dot in our case, because that's what makes most sense for implementers and users.

robscott: I don't mind the idea of more relaxed validation here in the future, I'd just prefer to intentionally discuss/choose that separately. For now I think it's safest to suggest a value that would be safe for both IngressClass and GatewayClass. For example `acme.io/gateway/v1` and `acme.io/gateway/v2` communicate the same use case and are fully valid for both APIs.

hbagdi: I see. Updated for now. This is a minor detail but still would like to track it in #330.

danehans:
LGTM

robscott:
Thanks!

/lgtm
/approve

docs-src/concepts.md
robscott: For the record, I was wrong here: https://github.com/kubernetes-sigs/gateway-api/issues/330#issuecomment-844588025.

PR #328, opened 2020-09-14
Deprecate configmap as default on extension refs

### LINKED ISSUE #292: ConfigMap defaults for extension refs are confusing
I'm not sure what a config map extension would look like, but I don't think it's a great default here. It doesn't feel quite as intuitive or obvious as our other defaults. Defaulting to Secret refs for certificates or Services for forwarding generally seems to line up with the most common use of those refs. I don't think that's true for config maps and extension refs.

eth0xFEED:
fixes #292


apis/v1alpha1/gatewayclass_types.go
```diff
@@ -103,11 +103,10 @@ type GatewayClassSpec struct {
...
 	// Valid resources for reference are up to the Controller. Examples
-	// include "configmaps" (omit or specify the empty string for the group
-	// to indicate the core API group) or a custom resource (CRD).  Omitting
-	// or specifying the empty string for both the resource and group
-	// indicates that the resource is "configmaps".  If the referent cannot
-	// be found, the GatewayClass's "InvalidParameters" status condition
+	// include "configmaps"  or a custom resource (CRD).
+	//
+	// If the referent cannot be found,
+	// the GatewayClass's "InvalidParameters" status condition
```
robscott, line 109: Nit: the line length here feels a bit off. I generally try to wrap lines at 80 chars.

apis/v1alpha1/httproute_types.go
```diff
@@ -64,13 +64,16 @@ type HTTPRouteHost struct {
...
+	// "host" block.  The resource may be "configmaps"  or an
+	// implementation-defined resource
 	// (for example, resource "myroutehosts" in group "networking.acme.io").
-	// Omitting or specifying the empty string for both the resource and
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the HTTPRoute will be true.
+	//
+	// Valid resources for reference are up to the Controller. Examples
+	// include "configmaps"  or a custom resource (CRD).
```
robscott, line 72: Nit: This seems to repeat the comment block above (`The resource may be "configmaps" or an implementation-defined resource (for example, resource "myroutehosts" in group "networking.acme.io"`). I think either are fine, we should just remove one.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -62,10 +62,9 @@ type ServicesDefaultLocalObjectReference struct {
 	Name string `json:"name"`
 }
 
-// ConfigMapsDefaultLocalObjectReference identifies an API object within a
-// known namespace that defaults group to core and resource to configmaps
-// if unspecified.
-type ConfigMapsDefaultLocalObjectReference struct {
+// LocalObjectReference identifies an API object within a
+// known namespace that defaults group to core
```
robscott, line 66:
```suggestion
// LocalObjectReference identifies an API object within a known namespace.
```

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -85,23 +84,16 @@ type ConfigMapsDefaultLocalObjectReference struct {
 	// +optional
 	// +kubebuilder:default=core
```
robscott, line 85: I think we should also remove the core default here.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -85,23 +84,16 @@ type ConfigMapsDefaultLocalObjectReference struct {
 	// +optional
```
robscott, line 84: This shouldn't be optional anymore.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -85,23 +84,16 @@ type ConfigMapsDefaultLocalObjectReference struct {
...
-	// Otherwise, if the configmaps resource is not desired, specify the desired
-	// group:
+	// specify the desired resource and group
 	//
 	// fooRef:
 	//   group: acme.io
 	//   resource: foos
 	//   name: myfoo
 	//
 	// +optional
```
robscott, line 96: This shouldn't be optional anymore.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -85,23 +84,16 @@ type ConfigMapsDefaultLocalObjectReference struct {
...
 	// +kubebuilder:default=core
 	Group string `json:"group"`
-	// Resource is the API resource name of the referent. Omitting the value
-	// or specifying the empty string indicates the configmaps resource. For
-	// example, use the following to specify a configmaps resource:
-	//
-	// fooRef:
-	//   name: myconfigmap
+	// Resource is the API resource name of the referent.
```
robscott, line 87: I think this is the only comment line we need here.

apis/v1alpha1/httproute_types.go
```diff
@@ -64,13 +64,16 @@ type HTTPRouteHost struct {
 	Rules []HTTPRouteRule `json:"rules"`
 
 	// ExtensionRef is an optional, implementation-specific extension to the
-	// "host" block.  The resource may be "configmaps" (omit or specify the
-	// empty string for the group) or an implementation-defined resource
+	// "host" block.  The resource may be "configmaps"  or an
+	// implementation-defined resource
```
robscott, line 68: Nit: The line length is also a bit off here

robscott (commented):
Thanks for your work on this!

eth0xFEED:
thanks for the review and tips @robscott , I think i've resolved as well as made the requested changes. One area where some uncertainty arose was WRT to the desired end state on local_object_reference_types.go . Unsure if the reduced comment strings should still include the group/resource example ... please advise if i have not guessed correctly :)

robscott:
Thanks for the updates @eth0xFEED! This mostly LGTM. I think I'd prefer removing the example from the comments in this case. I'd also suggest removing the `+kubebuilder:validation:Required` (sorry I missed this earlier). It looks like this specific annotation is entirely unnecessary since fields are required by default.

robscott:
Thanks!

/lgtm
/approve

PR #329, opened 2020-09-15
Unrevert client generation

howardjohn:
This was reverted in a16ebea. This PR copies the update-codegen.sh
script from prior to that commit without modification, adds it to `make
generate`, and runs it.


robscott:
Thanks!

/lgtm
/approve

PR #336, opened 2020-09-17
API Validation and Cleanup

robscott:
- Adds min and max length everywhere possible
- Removes +required since it is inferred
- Updates default fields to no longer be optional
- Updates Routes to use HTTP forward to style and shared type
- Fixes all examples to be valid again

Fixes https://github.com/kubernetes-sigs/service-apis/issues/290

/cc @hbagdi @danehans


apis/v1alpha1/forwarding_target_types.go
```diff
@@ -139,7 +167,8 @@ type GenericForwardToTarget struct {
 	// Support: Core (HTTPRoute)
 	// Support: Extended (TCPRoute)
 	//
-	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
```
howardjohn, line 171: I am not sure we should have a max here, the sum could be greater than 100. At least in theory we can have 
```
weight=1
weight=999
```

Unless we are intentionally restricting the granularity

apis/v1alpha1/gateway_types.go
```diff
@@ -85,8 +85,8 @@ type GatewaySpec struct {
 	//
 	// Support: Core
 	//
-	// +required
 	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
```
howardjohn, line 89: Is there any defined spec for how 1 gateway with 2 listeners compares to 2 gateways with 1 listener each?

apis/v1alpha1/gateway_types.go
```diff
@@ -533,12 +528,12 @@ type ListenerStatus struct {
...
 	// status of all such Listeners.
 	//
-	// +required
+	// +kubebuilder:validation:MaxLength=250
 	Port string `json:"port"`
 
 	// Conditions describe the current condition of this listener.
 	//
-	// +required
+	// +kubebuilder:validation:MaxItems=16
```
howardjohn, line 536: is this a common constraint on conditions?

apis/v1alpha1/httproute_types.go
```diff
@@ -56,7 +56,8 @@ type HTTPRouteHost struct {
 	// Support: Core
 	//
 	// +optional
-	Hostnames []string `json:"hostnames,omitempty"`
+	// +kubebuilder:validation:MaxItems=16
```
howardjohn, line 59: same question I had about 1:N vs N:1 Gateways here

howardjohn: Basically, if we need more than 16 can we easily just split off another httproute and get the same behavior

apis/v1alpha1/httproute_types.go
```diff
@@ -377,10 +378,9 @@ type HTTPRequestHeaderFilter struct {
 	//   GET /foo HTTP/1.1
 	//   My-Header2: DEF
 	//
-	// Support: extended?
+	// Support: Extended
+	// +kubebuilder:validation:MaxItems=16
```
howardjohn, line 382: Why is remove limited to 16 but add is not?

howardjohn (commented):
Is `16` some standard number for list constraints? I am a bit worried we may artificial cut off some use cases by restricting it this low.

apis/v1alpha1/httproute_types.go
robscott: Yeah, absolutely. This is a rather arbitrary limit and I'm happy to increase it if there are use cases that will run into it. Just trying to enforce reasonable upper limits wherever we can.

apis/v1alpha1/httproute_types.go
robscott: Sadly I can't find a way to limit the number of keys in a map with kubebuilder tags (or CRD validation in general).

apis/v1alpha1/gateway_types.go
robscott: I don't think so. Just need some kind of upper limit here. Given that we have ~5 possible conditions that can be set here, it seemed unlikely that we would exceed 16.

apis/v1alpha1/forwarding_target_types.go
robscott: Is 1000 a sufficient level of granularity here? I'd initially thought 100 would be sufficient, but you're right that we can have a max variation 100:1, and maybe some users will want to split off a tinier fraction than that?

apis/v1alpha1/forwarding_target_types.go
```diff
@@ -62,9 +64,8 @@ type HTTPForwardToTarget struct {
 	// Support: Core (HTTPRoute)
 	// Support: Extended (TCPRoute)
 	//
-	// +optional
 	// +kubebuilder:default=1
-	Weight TargetWeight `json:"weight"`
+	Weight int32 `json:"weight,omitempty"`
```
robscott, line 68: Min: 1
Max: 10000

apis/v1alpha1/forwarding_target_types.go
```diff
@@ -79,12 +80,37 @@ type HTTPForwardToTarget struct {
...
+	// "myroutetargets" in group "networking.acme.io"). Omitting or specifying
+	// the empty string for both the resource and group indicates that the
+	// resource is "services". If the referent cannot be found, the
+	// "InvalidRoutes" status condition on any Gateway that includes the Route
+	// will be true.
+	//
+	// Support: core
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 97: min: 1
max: 4

apis/v1alpha1/forwarding_target_types.go
robscott: Max: 10000

apis/v1alpha1/gateway_types.go
robscott, line 89: MaxItems: 128

apis/v1alpha1/gateway_types.go
```diff
@@ -235,6 +228,7 @@ type HostnameMatch struct {
 // * "Any"
 //
 // +kubebuilder:validation:Enum=Domain;Exact;Any
+// +kubebuilder:validation:MaxLength=30
```
robscott, line 231: Remove this

apis/v1alpha1/gateway_types.go
```diff
@@ -483,15 +476,17 @@ type GatewayStatus struct {
...
 	Addresses []GatewayAddress `json:"addresses"`
 
 	// Conditions describe the current conditions of the Gateway.
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
 	// Listeners provide status for each unique listener port defined in the Spec.
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 489: Match above

apis/v1alpha1/gateway_types.go
```diff
@@ -483,15 +476,17 @@ type GatewayStatus struct {
 	//
 	// These addresses should all be of type "IPAddress".
 	//
-	// +required
+	// +kubebuilder:validation:MaxItems=16
 	Addresses []GatewayAddress `json:"addresses"`
 
 	// Conditions describe the current conditions of the Gateway.
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 484: MaxItems=8

danehans (commented):
Why is `apis/v1alpha1/forwarding_target_types.go` removed in this PR?

apis/v1alpha1/gateway_types.go
```diff
@@ -205,9 +198,8 @@ type HostnameMatch struct {
 	// Match specifies how the hostname provided by the client should be
 	// matched against the given value.
 	//
-	// +optional
 	// +kubebuilder:default=Exact
```
danehans, line 201: @robscott I question whether "Exact" is the correct default due to the `Name` field details:
```
<SNIP>
// This field is required for the "Domain" and "Exact" match types.
//
// +optional
Name string `json:"name"`
```

`mach: Exact` will be defaulted which requires `Name` but name is optional for this match type. Maybe the default should be "Any". Thoughts? Let me know if this issue should be addressed separately.

apis/v1alpha1/gateway_types.go
```diff
@@ -377,9 +372,8 @@ type RouteBindingSelector struct {
 	//
 	// Support: Core
 	//
-	// +optional
 	// +kubebuilder:default={onlySameNamespace:true}
-	RouteNamespaces RouteNamespaces `json:"routeNamespaces,omitempty"`
+	RouteNamespaces *RouteNamespaces `json:"routeNamespaces,omitempty"`
```
danehans, line 376: `RouteNamespaces` should continue to be a value.

apis/v1alpha1/gateway_types.go
```diff
@@ -429,14 +423,13 @@ type GatewayAddress struct {
 	//
 	// Support: Extended
 	//
-	// +optional
 	// +kubebuilder:default=IPAddress
-	Type AddressType `json:"type"`
+	Type *AddressType `json:"type,omitempty"`
```
danehans, line 427: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492273748

apis/v1alpha1/gateway_types.go
```diff
@@ -483,15 +476,17 @@ type GatewayStatus struct {
 	//
 	// These addresses should all be of type "IPAddress".
 	//
-	// +required
+	// +kubebuilder:validation:MaxItems=16
```
danehans, line 479: Since `Addresses` is required, should an "Unknown" `AddressType` be supported for controllers that are unable to assign an address during reconciliation?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -93,9 +93,8 @@ type GatewayClassSpec struct {
 	//
 	// Support: Core
 	//
-	// +optional
 	// +kubebuilder:default={onlySameNamespace:true}
-	AllowedRouteNamespaces RouteNamespaces `json:"allowedRouteNamespaces,omitempty"`
+	AllowedRouteNamespaces *RouteNamespaces `json:"allowedRouteNamespaces,omitempty"`
```
danehans, line 97: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492273748

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -142,9 +141,8 @@ type RouteNamespaces struct {
 	//
 	// Support: Core
 	//
-	// +optional
 	// +kubebuilder:default=true
-	OnlySameNamespace bool `json:"onlySameNamespace"`
+	OnlySameNamespace *bool `json:"onlySameNamespace,omitempty"`
```
danehans, line 145: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492273748

apis/v1alpha1/httproute_types.go
danehans: @robscott should a maximum # of `Rules` and/or `ExtensionRef` also be established at this time?

apis/v1alpha1/httproute_types.go
```diff
@@ -131,11 +154,14 @@ type HTTPRouteRule struct {
 	// Support: core
 	//
 	// +optional
```
danehans, line 156: // +kubebuilder:validation:MinItems=0

apis/v1alpha1/gateway_types.go
```diff
@@ -105,6 +105,7 @@ type GatewaySpec struct {
 	// Support: Core
 	//
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
danehans, line 108: // +kubebuilder:validation:MinItems=0 and omitempty like other optional lists in this PR.

apis/v1alpha1/gateway_types.go
```diff
@@ -132,11 +132,8 @@ type Listener struct {
...
 	// Support: Core
 	//
-	// +required
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=65536
-	// +kubebuilder:validation:ExclusiveMinimum=true
-	// +kubebuilder:validation:ExclusiveMaximum=true
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
 	Port int32 `json:"port,omitempty"`
```
danehans, line 137: Remove `omitempty` if this field is required and a default is not being set.

apis/v1alpha1/gateway_types.go
```diff
@@ -155,8 +152,6 @@ type Listener struct {
 	//   applied at both the TLS and HTTP protocol layers.
 	//
 	// Support: Core
-	//
-	// +required
 	Protocol ProtocolType `json:"protocol,omitempty"`
```
danehans, line 155: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492290720

apis/v1alpha1/gateway_types.go
```diff
@@ -223,6 +215,7 @@ type HostnameMatch struct {
 	// This field is required for the "Domain" and "Exact" match types.
 	//
 	// +optional
+	// +kubebuilder:validation:MaxLength=253
 	Name string `json:"name"`
```
danehans, line 219: Add `omitempty` tag if this field is optional.

apis/v1alpha1/gateway_types.go
danehans: // +kubebuilder:validation:MinItems=0 to match other optional lists.

apis/v1alpha1/gateway_types.go
danehans: // +kubebuilder:validation:MinItems=0 to match other optional lists.

apis/v1alpha1/httproute_types.go
```diff
@@ -112,9 +135,9 @@ type HTTPRouteRule struct {
 	// path match on "/", which has the effect of matching every
 	// HTTP request.
 	//
-	// +optional
 	// +kubebuilder:default={{path:{ type: "Prefix", value: "/"}}}
-	Matches []HTTPRouteMatch `json:"matches"`
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 139: // +kubebuilder:validation:MinItems=1 to match other required lists.

apis/v1alpha1/httproute_types.go
```diff
@@ -247,9 +268,8 @@ type HTTPRouteMatch struct {
 	// Path specifies a HTTP request path matcher. If this field is not
 	// specified, a default prefix match on the "/" path is provided.
 	//
-	// +optional
 	// +kubebuilder:default={type: "Prefix", value: "/"}
-	Path *HTTPPathMatch `json:"path"`
+	Path *HTTPPathMatch `json:"path,omitempty"`
```
danehans, line 272: `Path` should be a value instead of a pointer when the field is being defaulted.

apis/v1alpha1/httproute_types.go
```diff
@@ -268,7 +288,7 @@ type HTTPRouteMatch struct {
 	// Support: custom
 	//
 	// +optional
-	ExtensionRef *RouteMatchExtensionObjectReference `json:"extensionRef"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef"`
```
danehans, line 291: Add `omitempty` tag if field is optional.

apis/v1alpha1/httproute_types.go
```diff
@@ -319,8 +339,7 @@ type HTTPRouteFilter struct {
 	// types to extend the core API with implementation-specific behavior.
 	//
 	// +unionDiscriminator
-	// +kubebuilder:validation:Required
-	// +required
+	// +kubebuilder:validation:MaxLength=100
```
danehans, line 342: Should `// +kubebuilder:validation:MinLength=1` be set?

apis/v1alpha1/httproute_types.go
```diff
@@ -331,7 +350,7 @@ type HTTPRouteFilter struct {
 	// group indicates that the resource is "configmaps".
 	// ExtensionRef MUST NOT be used for core and extended filters.
 	// +optional
-	ExtensionRef *RouteFilterExtensionObjectReference `json:"extensionRef"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef"`
```
danehans, line 353: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492305973.

apis/v1alpha1/httproute_types.go
```diff
@@ -331,7 +350,7 @@ type HTTPRouteFilter struct {
...
 
 	// Filter-specific configuration definitions for core and extended filters
```
danehans, line 356: `RequestHeader` and `RequestMirror` appear to be optional (i.e. pointers). If so, set +optional and omitempty for each.

apis/v1alpha1/httproute_types.go
danehans: `// +kubebuilder:validation:MinItems=1`  to match other requried lists

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
+	//
+	// Support: Extended
 	//
-	TargetRef ForwardToTargetObjectReference `json:"targetRef" protobuf:"bytes,1,opt,name=targetRef"`
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
```
danehans, line 428: Should `LocalObjectReference` be a pointer since it's optional?

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	// Support: Extended
+	// +kubebuilder:validation:MaxItems=16
 	Remove []string `json:"remove"`
-
-	// TODO
 }
 
-// HTTPRequestMirrorFilter defines configuration for the
-// RequestMirror filter.
+// HTTPRequestMirrorFilter defines configuration for the RequestMirror filter.
```
danehans, line 404: Currently all `HTTPRequestMirrorFilter` fields are optional. Does a validation mechanism exists to ensure at least 1 field contains a value? An empty `HTTPRequestMirrorFilter` doesn't make sense, but I understand why each field is optional.

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this
+	// Route to true.
 	//
-	// Support: Core (Kubernetes Services)
-	// Support: Implementation-specific (Other resource types)
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 416: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	ServiceName *string `json:"serviceName,omitempty"`
```
danehans, line 417: // +kubebuilder:validation:Minimum=1

apis/v1alpha1/httproute_types.go
danehans, line 416: // +kubebuilder:validation:Minimum=1

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	// "InvalidRoutes" status condition on any Gateway that includes this
+	// Route to true.
 	//
-	// Support: core
+	// Support: Core
 	//
-	// +kubebuilder:validation:MinItems=1
-	To []HTTPForwardToTarget `json:"to"`
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 457: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492313616

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
 	//
-	// Support: custom
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
```
danehans, line 469: Field should be a pointer it's optional and no defaulting.

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
+	Weight int32 `json:"weight,omitempty"`
+
+	// Filters defined at this-level should be executed if and only if the
+	// request is being forwarded to the backend defined here.
+	//
+	// Conformance: Filtering support, including core filters, is NOT guaranteed
+	// at this-level. Use Filters in HTTPRouteRule for portable filters across
+	// implementations.
+	//
+	// Support: Customf
```
danehans, line 505: s/Customf/Custom/

apis/v1alpha1/httproute_types.go
```diff
@@ -377,70 +396,116 @@ type HTTPRequestHeaderFilter struct {
...
 	//
 	// +optional
```
danehans, line 507: Should the list have a min/max length to match other optional lists?

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -35,9 +35,9 @@ type ServicesDefaultLocalObjectReference struct {
 	//   resource: foos
 	//   name: myfoo
 	//
-	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 38: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +53,25 @@ type ServicesDefaultLocalObjectReference struct {
 	//   resource: foos
 	//   name: myfoo
 	//
-	// +optional
 	// +kubebuilder:default=services
-	Resource string `json:"resource"`
+	// +kubebuilder:validation:MaxLength=253
+	Resource string `json:"resource,omitempty"`
```
danehans, line 58: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492315609

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +53,25 @@ type ServicesDefaultLocalObjectReference struct {
...
 	// Name is the name of the referent.
 	//
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 61: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r492315609

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +53,25 @@ type ServicesDefaultLocalObjectReference struct {
...
 	Name string `json:"name"`
 }
 
 // LocalObjectReference identifies an API object within a known namespace.
 type LocalObjectReference struct {
 	// Group is the API group name of the referent.
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 68: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +53,25 @@ type ServicesDefaultLocalObjectReference struct {
...
 	Group string `json:"group"`
 	// Resource is the API resource name of the referent.
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 71: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +53,25 @@ type ServicesDefaultLocalObjectReference struct {
...
 	Resource string `json:"resource"`
 	// Name is the name of the referent.
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 74: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -92,9 +95,9 @@ type SecretsDefaultLocalObjectReference struct {
 	//   resource: foos
 	//   name: myfoo
 	//
-	// +optional
 	// +kubebuilder:default=core
-	Group string `json:"group"`
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 99: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -110,11 +113,10 @@ type SecretsDefaultLocalObjectReference struct {
 	//   resource: foos
 	//   name: myfoo
 	//
-	// +optional
 	// +kubebuilder:default=secrets
-	Resource string `json:"resource"`
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 117: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -110,11 +113,10 @@ type SecretsDefaultLocalObjectReference struct {
...
+	Resource string `json:"resource,omitempty"`
 	// Name is the name of the referent.
-	//
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 120: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/route_types.go
```diff
@@ -25,12 +25,63 @@ type GatewayObjectReference struct {
 	// +optional
 	Namespace string `json:"namespace,omitempty"`
```
danehans, line 26: Should min/max length be established for `Name` and `Namespace` fields?

apis/v1alpha1/route_types.go
```diff
@@ -25,12 +25,63 @@ type GatewayObjectReference struct {
...
+	// to. When specified, this takes the place of BackendRef. If both
+	// BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this
+	// Route to true.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 43: // +kubebuilder:validation:MinLength=1

apis/v1alpha1/tcproute_types.go
```diff
@@ -51,42 +51,14 @@ type TCPRouteRule struct {
 	// if **any** one of the matches is satisfied.
 	//
 	// +optional
-	Matches []TCPRouteMatch `json:"matches"`
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 54: // +kubebuilder:validation:MinItems=0

apis/v1alpha1/tcproute_types.go
```diff
@@ -104,7 +76,7 @@ type TCPRouteMatch struct {
 	// Support: custom
 	//
 	// +optional
-	ExtensionRef *RouteMatchExtensionObjectReference `json:"extensionRef"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef"`
```
danehans, line 79: `omitempty` tag if field is optional.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -132,7 +103,7 @@ type TLSRouteMatch struct {
 	// Support: custom
 	//
 	// +optional
-	ExtensionRef *RouteMatchExtensionObjectReference `json:"extensionRef"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef"`
```
danehans, line 106: `omitempty` if field is optional.

apis/v1alpha1/udproute_types.go
```diff
@@ -97,7 +72,7 @@ type UDPRouteMatch struct {
 	// Support: custom
 	//
 	// +optional
-	ExtensionRef *RouteMatchExtensionObjectReference `json:"extensionRef"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef"`
```
danehans, line 75: `omitempty` if field is optional.

examples/basic-http.yaml
```diff
@@ -35,32 +35,23 @@ metadata:
...
+    - headers:
+        type: Exact
+        values:
+          x-forwarded-proto: https
+          magic: foo
+      path:
+        type: Prefix
+        value: /some/thing
+    forwardTo:
+    - serviceName: my-service2
```
danehans, line 57: Newline

hbagdi:
Rob, is it a lot of work to remove Flattening Route form this PR?
That might help with quality and speed of reviews and getting this in faster.

robscott:
@hbagdi That's definitely my intent. For now they're two separate commits in the same PR because I already have so many others that will require rebasing. I'm trying to push for those to get in first and then work on the clean up and then flattening the route just so I don't have endless rebasing.

apis/v1alpha1/forwarding_target_types.go
hbagdi: Question: Is it worth defining a `Port` type with the min, max and default and then use that everywhere?
I understand that you still need to specify the kubebuilder annotation in the place where it is used but having a type would be sort of a self-documenting thing through the API.

apis/v1alpha1/forwarding_target_types.go
```diff
@@ -79,12 +80,37 @@ type HTTPForwardToTarget struct {
 	Filters []HTTPRouteFilter `json:"filters"`
```
hbagdi: Can we start with an arbitrary limit of 8 or 16 filters?
We can certainly bump it up as needed in future. The reasoning here is that this limit acts as a warning or alert to us that this feature is being used a lot and in certain ways that we might not have anticipated and it might be helpful to take action --- let the community be creative or course correct.

apis/v1alpha1/forwarding_target_types.go
hbagdi: With a ratio of 10000:1, one can get a split of 0.01%, which seems to be enough for most use cases I know of. In case this is not sufficient, we can again bump in future.

apis/v1alpha1/gateway_types.go
hbagdi: Rob, I thought through this a bit more and I think start with a lower limit of 32 or 64.
If you really think about this, 64 listeners are a lot from a controller's perspective. I really doubt even large use cases will run into this limit because if you have a lot of listeners, operators like to split them out into multiple Gateways to avoid a very large blast radius.

apis/v1alpha1/gateway_types.go
```diff
@@ -85,8 +85,8 @@ type GatewaySpec struct {
 	//
 	// Support: Core
 	//
-	// +required
```
hbagdi, line 88: Github won't let me comment on portions which are not diffed.
Above in this file, we have:
```
GatewayClassName string `json:"gatewayClassName"`
```
Does it make sense to have min and max length on that field?

apis/v1alpha1/gateway_types.go
hbagdi: Since ProtocolType is a free-string to allow for arbitrary protocols. let us add a min and max length validation here.

apis/v1alpha1/gateway_types.go
hbagdi: @jpeach Do you recall why we settled on Exact in the past? It probably warrants its own issue.

apis/v1alpha1/gateway_types.go
```diff
@@ -347,6 +341,7 @@ type GatewayTLSConfig struct {
 // TLSModeType type defines behavior of gateway with TLS protocol.
 // +kubebuilder:validation:Enum=Terminate;Passthrough
 // +kubebuilder:default=Terminate
+// +kubebuilder:validation:MaxLength=30
 type TLSModeType string
```
hbagdi, line 345: nit: 30 characters is excessive (unless we were naming in Java) for a ModeType. How about 16? Is that too less?

apis/v1alpha1/gateway_types.go
hbagdi: > Since Addresses is required, should an "Unknown" AddressType be supported for controllers that are unable to assign an address during reconciliation?

Shouldn't that be a Condition instead?

apis/v1alpha1/gateway_types.go
hbagdi: We discussed in last office hours that we will limit to 8 conditions. Excessive conditions harm UX.

apis/v1alpha1/httproute_types.go
hbagdi: What value do we get out of min=0 validation?

apis/v1alpha1/httproute_types.go
hbagdi: I'm wondering about this in other places as well. The field is required so there should be something in the field. Does the validation allow for empty string like `"  "`?

apis/v1alpha1/tlsroute_types.go
```diff
@@ -22,7 +22,6 @@ import (
 // TLSRouteSpec defines the desired state of TLSRoute
 type TLSRouteSpec struct {
 	// Rules are a list of TLS matchers and actions.
-	// +required
 	// +kubebuilder:validation:MinItems=1
 	Rules []TLSRouteRule `json:"rules"`
```
hbagdi, line 26: Add a max limit of 16 like we have done in HTTPRoute.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -59,39 +58,9 @@ type TLSRouteRule struct {
 	// if **any** one of the matches is satisfied.
 	// +optional
 	Matches []TLSRouteMatch `json:"matches"`
```
hbagdi, line 60: Add a max limit of 16 like we have done in HTTPRoute.

apis/v1alpha1/udproute_types.go
```diff
@@ -49,37 +49,10 @@ type UDPRouteRule struct {
 	//
 	// +optional
 	Matches []UDPRouteMatch `json:"matches"`
```
hbagdi, line 51: Add a max limit of 16.

apis/v1alpha1/httproute_types.go
robscott: Add minLength=1 validation, we likely need that on more strings as well. Unfortunately I can't find a way to handle/trim whitespace with kubebuilder validation tags.

apis/v1alpha1/udproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/forwarding_target_types.go
robscott: Updated, thanks!

apis/v1alpha1/httproute_types.go
robscott: Updated to add a max # of rules as well (16).

apis/v1alpha1/httproute_types.go
robscott: @danehans I don't think we want to have a minimum here? Wouldn't it be valid to only add headers and not remove headers with this filter?

apis/v1alpha1/forwarding_target_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: That's a good point. I've lowered this to 64.

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go
robscott: Updated to match. Wouldn't `MinItems=0` be default? Do we need to specify it everywhere?

apis/v1alpha1/gateway_types.go
robscott: Updated max items to 8. Wouldn't `MinItems=0` be default? Do we need to specify it everywhere?

apis/v1alpha1/gateway_types.go
robscott: https://github.com/kubernetes-sigs/service-apis/issues/348

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go
robscott: I'm hesitant to add a minimum here since status will likely need to be populated at some point after a Gateway was created. As far as if a controller is unable to assign an address during reconciliation, I think that would be better communicated with some kind of Condition, maybe `Scheduled=False` with a reason of `NoResources`? Does that seem reasonable? If not, maybe we should continue this discussion in a follow up issue.

apis/v1alpha1/gatewayclass_types.go
robscott: Field is removed now, thanks!

apis/v1alpha1/gatewayclass_types.go

apis/v1alpha1/httproute_types.go
robscott: Agree, don't think we need to add min=0 for lengths. I do see value for ints.

apis/v1alpha1/gateway_types.go
robscott: Added omitempty, thanks!

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/gateway_types.go

apis/v1alpha1/httproute_types.go
robscott: Do we need a minItems here if we already have a default?

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: Added minLength=1 validation, we likely need that on more strings as well. Unfortunately I can't find a way to handle/trim whitespace with kubebuilder validation tags.

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: Sadly this will be one more thing that we'll need to rely on a validating webhook for.

robscott: https://github.com/kubernetes-sigs/service-apis/issues/349

apis/v1alpha1/httproute_types.go
robscott: Why would we add a minLength validation of 1 on an optional field?

apis/v1alpha1/httproute_types.go
robscott: Why would we add a minLength validation of 1 on an optional field?

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/route_types.go

apis/v1alpha1/route_types.go
robscott: Does a min length make sense for an optional field?

apis/v1alpha1/tcproute_types.go
robscott: Do we need to specify minItems=0 if that's already the default?

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/udproute_types.go
robscott: Updated, thanks!

examples/basic-http.yaml
robscott: Updated, thanks!

apis/v1alpha1/forwarding_target_types.go
robscott: Set to 16 in both cases as a starting point. Seems pretty generous.

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: Good call, removed in favor of enum validation.

apis/v1alpha1/gateway_types.go
```diff
@@ -53,6 +53,8 @@ type GatewayList struct {
 type GatewaySpec struct {
 	// GatewayClassName used for this Gateway. This is the name of a
 	// GatewayClass resource.
+	// +kubebuilder:validation:MaxLength=1
```
jpeach, line 56:
```suggestion
	// +kubebuilder:validation:MinLength=1
```

apis/v1alpha1/gateway_types.go
```diff
@@ -302,6 +297,8 @@ const (
 // * "UDP"
 //
 // +kubebuilder:validation:Enum=HTTP;HTTPS;TLS;TCP;UDP
+// +kubebuilder:validation:MaxLength=1
+// +kubebuilder:validation:MaxLength=100
```
jpeach, line 301: Length validation are redundant due to the enum.

jpeach (changes requested):

apis/v1alpha1/gateway_types.go
robscott: Good call, I think we'll have to add this back in whenever we add more flexible protocol support but that's not in scope here. https://github.com/kubernetes-sigs/service-apis/issues/203

apis/v1alpha1/gateway_types.go

robscott:
If anyone would prefer to review route flattening separately, I've created another PR that is a subset of this one: https://github.com/kubernetes-sigs/service-apis/pull/353

danehans:
> If anyone would prefer to review route flattening separately, I've created another PR that is a subset of this one: #353

If route flattening has been removed from this PR, can you update the [description](https://github.com/kubernetes-sigs/service-apis/pull/336#issue-488405963).

apis/v1alpha1/gateway_types.go
```diff
@@ -702,12 +698,12 @@ type ListenerStatus struct {
 	// shares the same port value, this message reports the combined
 	// status of all such Listeners.
 	//
-	// +required
+	// +kubebuilder:validation:MaxLength=250
```
danehans, line 701: Why is `Port` a string for status but an int32 for spec?

apis/v1alpha1/httproute_types.go
```diff
@@ -19,18 +19,34 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// +genclient
```
danehans, line 22: According to the kubebuilder migration [guide](https://book.kubebuilder.io/migration/guide.html), `genclient` is no longer needed.

apis/v1alpha1/httproute_types.go
danehans: Probably not.

apis/v1alpha1/httproute_types.go
```diff
@@ -173,11 +197,14 @@ type HTTPRouteRule struct {
...
 	// +optional
-	Filters []HTTPRouteFilter `json:"filters"`
+	// +kubebuilder:validation:MaxItems=16
+	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
-	// Forward defines the upstream target(s) where the request should be sent.
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	Forward *HTTPForwardingTarget `json:"forward"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 205: Should the same be applied to `MinItems=0` here since [this](https://github.com/kubernetes-sigs/service-apis/pull/336/files#r495361672) consensus was achieved?

apis/v1alpha1/httproute_types.go
danehans: My bad, I missed that this is an optional field.

apis/v1alpha1/httproute_types.go
```diff
@@ -421,70 +442,117 @@ type HTTPRequestHeaderFilter struct {
...
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
 	//
 	// Support: Core
 	//
 	// +optional
-	TargetPort *TargetPort `json:"targetPort,omitempty"`
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 486: Should a minimum be set if this field is optional?

apis/v1alpha1/httproute_types.go
```diff
@@ -421,70 +442,117 @@ type HTTPRequestHeaderFilter struct {
...
+	// to true.
 	//
-	// Support: custom
+	// Support: Extended
 	//
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef"`
+	BackendRef *LocalObjectReference `json:"backendRef,omitempty"`
+
+	// Port specifies the destination port number to use for the backend. If
```
danehans, line 517: This could be interpreted as only being applicable to `BackendRef`. To avoid any potential confusion, it may be a good idea to be explicit by referencing the `backendRef` and `serviceName fields.

apis/v1alpha1/httproute_types.go
```diff
@@ -421,70 +442,117 @@ type HTTPRequestHeaderFilter struct {
...
+	// unspecified and a Service object consisting of a single port definition
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 527: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r496862195

apis/v1alpha1/httproute_types.go
```diff
@@ -421,70 +442,117 @@ type HTTPRequestHeaderFilter struct {
...
+	// Weight specifies the proportion of traffic forwarded to a backend,
+	// computed as weight/(sum of all weights in this ForwardTo list). Weight is
+	// not a percentage and the sum of weights does not need to equal 100. If
+	// only one backend is specified, 100% of the traffic is forwarded to that
+	// backend. If unspecified, weight defaults to 1.
+	//
+	// Support: Core (HTTPRoute)
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 540: Is `kubebuilder:validation:Minimum` needed since defaulting is being used?

apis/v1alpha1/local_object_reference_types.go
danehans: I think I may have been wrong here. I'm not sure it makes sense to enforce min on a field that is being defaulted.

apis/v1alpha1/local_object_reference_types.go
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r496873587

apis/v1alpha1/local_object_reference_types.go
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r496873587

apis/v1alpha1/local_object_reference_types.go
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r496873587

apis/v1alpha1/route_types.go
danehans: It does not, my bad.

apis/v1alpha1/route_types.go
```diff
@@ -57,13 +57,69 @@ type RouteGateways struct {
...
+	ServiceName *string `json:"serviceName,omitempty"`
+
+	// BackendRef is a reference to a backend to forward matched requests to. If
```
danehans, line 84: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.yaml&hide-deleted-files=true#r496867788

apis/v1alpha1/route_types.go
```diff
@@ -57,13 +57,69 @@ type RouteGateways struct {
...
+	// unspecified and a Service object consisting of a single port definition
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 105: Should min be removed since this field is optional?

apis/v1alpha1/route_types.go
```diff
@@ -57,13 +57,69 @@ type RouteGateways struct {
...
+	// Weight specifies the proportion of traffic forwarded to a backend,
+	// computed as weight/(sum of all weights in this ForwardTo list). Weight is
+	// not a percentage and the sum of weights does not need to equal 100. If
+	// only one backend is specified, 100% of the traffic is forwarded to that
+	// backend. If unspecified, weight defaults to 1.
+	//
+	// Support: Extended
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 118: Is min needed if defaulting is being used?

apis/v1alpha1/tcproute_types.go
```diff
@@ -55,42 +55,14 @@ type TCPRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the TCPRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef,omitempty"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 63: Should min be removed since this field is optional?

apis/v1alpha1/tlsroute_types.go
```diff
@@ -62,40 +62,14 @@ type TLSRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the TLSRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef,omitempty"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 70: Is min needed for optional fields?

apis/v1alpha1/udproute_types.go
```diff
@@ -52,38 +52,14 @@ type UDPRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the UDPRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 60: Is min needed for optional fields?

examples/basic-http.yaml
```diff
@@ -35,32 +35,23 @@ metadata:
...
-      action:
-        forwardTo:
-          - targetRef:
-              name: my-service
-              group: core
-              resource: services
+  hostnames:
+  - "foo.com"
+  rules:
+  - matches:
```
danehans, line 41: This is out of scope for this PR, but I think we should consider renaming `matches` to `match`. There are other areas in the APIs where a singular field name is being used for a list.

examples/basic-udp.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
   name: my-gateway
   namespace: default
 spec:
-  class: acme-lb
+  gatewayClassName: acme-lb
```
danehans, line 18: Out of scope for this PR, but PTAL at https://github.com/kubernetes-sigs/service-apis/issues/356 when you have a moment.

examples/default-match-http.yaml
```diff
@@ -39,35 +39,22 @@ metadata:
   labels:
     app: default-match
 spec:
-  hosts:
-  - hostnames:
+  hostnames:
```
danehans, line 42: Wow... what a difference it made to flatten hosts!

danehans:
A few nits, otherwise LGTM.

apis/v1alpha1/gateway_types.go
robscott: I don't think we've defined that very well in the spec yet. It seems like something that would be fairly implementation specific though. Although we'd initially discussed the possibility of merging Gateways that seems to no longer be necessary with this listener abstraction.

apis/v1alpha1/tcproute_types.go
robscott: Removed, thanks!

apis/v1alpha1/udproute_types.go
robscott: Removed, thanks!

apis/v1alpha1/httproute_types.go
robscott: Removed, thanks!

apis/v1alpha1/gateway_types.go
robscott: Created a follow up issue for this one: https://github.com/kubernetes-sigs/service-apis/issues/357

apis/v1alpha1/httproute_types.go
robscott: I think it's still valid here, we want to ensure we have non-negative and non-zero values.

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: I think it's still valid here, we want to ensure we have non-negative and non-zero values.

apis/v1alpha1/route_types.go

apis/v1alpha1/route_types.go
robscott: I think it's still valid here, we want to ensure we have non-negative and non-zero values.

apis/v1alpha1/route_types.go
robscott: I think it's still valid here, we want to ensure we have non-negative and non-zero values.

apis/v1alpha1/tlsroute_types.go
robscott: Removed, thanks!

apis/v1alpha1/httproute_types.go
robscott: I tried to do this and it resulted in a huge diff. I think it broke our clientgen. This may need a follow up issue.

apis/v1alpha1/httproute_types.go
robscott: I think it's still valid here, we want to ensure we have non-negative and non-zero values.

robscott:
Now that https://github.com/kubernetes-sigs/service-apis/pull/353 is in, I've updated this PR to be purely focused on API cleanup again. Thanks to everyone for the great feedback, I think I've resolved everything now. PTAL.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -167,6 +167,7 @@ type BackendTLSConfig struct {
 type BackendPolicyStatus struct {
 	// Conditions describe the current conditions of the BackendPolicy.
 	// +optional
+	// +kubebuilder:validation:MaxItems=8
```
bowei, line 170: What is the forward/backward compatibility issues with being conservative here? Is this something that can be changed easily w/ a backward compatible change? I remember changing validation can be hard to get right.

apis/v1alpha1/gateway_types.go
```diff
@@ -53,6 +53,8 @@ type GatewayList struct {
 type GatewaySpec struct {
 	// GatewayClassName used for this Gateway. This is the name of a
 	// GatewayClass resource.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=100
```
bowei, line 57: is 100 a common limit for this?

DNS can be 253
Maybe 255 or so

bowei (commented):
I think we should be careful with setting some of the list limits to small numbers where we don't have a good reason for the limit other than preventing abuse.

apis/v1alpha1/backendpolicy_types.go
robscott: Validation is definitely difficult to get right, but I think it's generally easier to loosen it than tighten it. I know there's definitely precedent for loosening validation in k/k (including the Ingress API). The guidance I remember from Tim was to start with relatively strict validation and loosen it if/when we need to.

apis/v1alpha1/gateway_types.go
robscott: Thanks for catching this! This should have been 253 because it's referring to the name of a Kubernetes resource. Most Kubernetes resources use the standard `IsDNS1123Subdomain` validation for names which includes a [limit of 253 chars](https://github.com/kubernetes/kubernetes/blob/4373d9aadd7b98aace0ed8b90ff45b255d4b1e88/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L202). Should be updated now.

apis/v1alpha1/httproute_types.go
danehans: OK, let's leave it as-is and I created https://github.com/kubernetes-sigs/service-apis/issues/361 to track the issue.

apis/v1alpha1/gateway_types.go
```diff
@@ -482,7 +476,7 @@ type RouteBindingSelector struct {
 	//
 	// Support: Core
 	//
-	// +required
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 479: Should `Resource` have a MinLength=1 since it's required and not being defaulted?

apis/v1alpha1/gateway_types.go
```diff
@@ -524,14 +518,13 @@ type GatewayAddress struct {
...
-	// +optional
 	// +kubebuilder:default=IPAddress
-	Type AddressType `json:"type"`
+	Type AddressType `json:"type,omitempty"`
 
 	// Value. Examples: "1.2.3.4", "128::1", "my-ip-address". Validity of the
 	// values will depend on `Type` and support by the controller.
 	//
-	// +required
+	// +kubebuilder:validation:MaxLength=253
```
danehans, line 527: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r497814413.

apis/v1alpha1/gateway_types.go
danehans, line 701: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r497814413.

apis/v1alpha1/gateway_types.go
```diff
@@ -702,12 +698,12 @@ type ListenerStatus struct {
...
 	// status of all such Listeners.
 	//
-	// +required
+	// +kubebuilder:validation:MaxLength=250
 	Port string `json:"port"`
 
 	// Conditions describe the current condition of this listener.
 	//
-	// +required
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 706: Should Listener conditions have the [same attributes](https://github.com/kubernetes-sigs/service-apis/pull/336/files#diff-2927d10c91558c2273bc3a8ea6464b85R592-R595) as Gateway conditions?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -118,7 +118,8 @@ const (
 type GatewayClassStatus struct {
 	// Conditions is the current status from the controller for
 	// this GatewayClass.
-	// +optional
+	//
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 122: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files#r497816693

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -35,9 +35,10 @@ type ServicesDefaultLocalObjectReference struct {
 	//   resource: foos
 	//   name: myfoo
 	//
-	// +optional
+	// +kubebuilder:validation:MinLength=1
```
danehans, line 38: Remove min since the field is being defaulted.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -53,22 +54,30 @@ type ServicesDefaultLocalObjectReference struct {
...
```
danehans, line 57: Remove min since the field is being defaulted.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -110,11 +120,12 @@ type SecretsDefaultLocalObjectReference struct {
...
```
danehans, line 123: min can be removed since the field is being defaulted.

apis/v1alpha1/local_object_reference_types.go
```diff
@@ -92,9 +101,10 @@ type SecretsDefaultLocalObjectReference struct {
...
```
danehans, line 104: min can be removed sine the field is being defaulted.

apis/v1alpha1/route_types.go
```diff
@@ -140,6 +142,7 @@ type RouteGatewayStatus struct {
 	// route has been admitted or rejected by the Gateway, and why.  Note
 	// that the route's availability is also subject to the Gateway's own
 	// status conditions and listener status.
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 145: Same as https://github.com/kubernetes-sigs/service-apis/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.html&file-filters%5B%5D=.md&file-filters%5B%5D=.yaml&hide-deleted-files=true#r497816693

apis/v1alpha1/udproute_types.go
```diff
@@ -52,14 +37,30 @@ type UDPRouteRule struct {
 	// Matches defines which packets match this rule.
 	//
 	// +optional
+	// +kubebuilder:validation:MaxItems=8
 	Matches []UDPRouteMatch `json:"matches"`
```
danehans, line 41: Add `omitempty` tag if the field is optional.

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go
robscott: I think it's fine for conditions to be required here since this entry should only exist if there are conditions to list.

apis/v1alpha1/gatewayclass_types.go
robscott: I'm not sure what to change here. I don't think this should be optional here given the default value. Are there other differences I'm missing?

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/local_object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha1/route_types.go
robscott: I think this should also be a required field because `RouteGatewayStatus` would only exist if there were a condition to set.

apis/v1alpha1/udproute_types.go
robscott: Updated, thanks!

PR #338, opened 2020-09-17
Make HTTPRouteFilter types optional

howardjohn:
I am not really sure if unionDiscriminator even works, but per
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190325-unions.md#go-tags
the individual items should be optional. Otherwise, you need to specify
all of them which is not quite right.

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -335,9 +335,11 @@ type HTTPRouteFilter struct {
 
 	// Filter-specific configuration definitions for core and extended filters
 
-	RequestHeader *HTTPRequestHeaderFilter `json:"requestHeader"`
+	// +optional
+	RequestHeader *HTTPRequestHeaderFilter `json:"requestHeader,omitempty"`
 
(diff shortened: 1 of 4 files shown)


apis/v1alpha1/httproute_types.go
```diff
@@ -335,8 +335,10 @@ type HTTPRouteFilter struct {
 
 	// Filter-specific configuration definitions for core and extended filters
 
+	// +optional
 	RequestHeader *HTTPRequestHeaderFilter `json:"requestHeader"`
```
hbagdi, line 339: Does this also need `omitempty` here?

hbagdi (commented):
/lgtm

apis/v1alpha1/httproute_types.go
howardjohn: Probably so. I wasn't sure how nil was handled - it will output `null` without omitempty (https://play.golang.org/p/Ag3r0tr1-ox)

robscott:
Thanks!

/lgtm
/approve

PR #341, opened 2020-09-22
Renames AllowedGatewayNamespaceSelector Field

danehans:
Renames `AllowedGatewayNamespaceSelector` to `GatewayNamespaceSelector` for brevity and consistency with other selector-based fields. This PR also adds the `omitempty` tag to the field since it's optional and makes minor changes to the field's godocs.

/hold until https://github.com/kubernetes-sigs/service-apis/pull/285 merges and then update `docs-src/security-model.md` to reference new field name.


robscott:
This is an interesting approach. In #285 I tried to shorten the name by cutting off `Selector`. The result was `AllowedGatewayNamespaces` instead of `GatewayNamespaceSelector`. I liked the clarity of `Allowed` in the name more than `Selector`, but I think this is a fairly subjective thing. Now that we're removing Route auth from GatewayClass entirely, I wonder if we could get away with something more generic like `NamespaceSelector` or `AllowedNamespaces`.

hbagdi:
> Thanks for filing this! In #285 I tried to shorten the name by cutting off Selector.

This has been brought up before but we never acted on it. Is cutting off `Selector` from all places a good change that we should pursue in the entire API?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,13 +62,13 @@ type GatewayClassSpec struct {
...
-	// outside this selector.
+	// GatewayNamespaceSelector is a selector of namespaces. Gateways that
+	// run in matched namespaces can use this GatewayClass. This is a standard
+	// Kubernetes LabelSelector. Controllers must not support Gateways in
+	// namespaces outside this selector.
 	//
-	// An empty selector (default) indicates that Gateways can use this
-	// GatewayClass from any namespace.
+	// An empty selector (default) indicates that this GatewayClass is available
+	// to use by Gateways in any namespace.
```
sgreene570, line 71: `available to use` => `available for use` ?

danehans:
Closed in favor of the updated naming implemented in https://github.com/kubernetes-sigs/service-apis/pull/285.

PR #344, opened 2020-09-23
remove unnecessary deps

hbagdi:


jpeach:
Why are these unnecessary?

hbagdi:
> Why are these unnecessary?

More of an experiment. I _think_ we don't need to track them in `go.mod`. I blew `go.mod` away and then re-init it, and this is what it came to. And things continue to work as expected as well.

jpeach:
> > Why are these unnecessary?
> 
> More of an experiment. I _think_ we don't need to track them in `go.mod`. I blew `go.mod` away and then re-init it, and this is what it came to. And things continue to work as expected as well.

Probably because you removed tools.go as well. This is needed to force venturing for things that exec with `go run`.

jpeach:
xref #352

hbagdi:
This doesn't work as James pointed out.

PR #351, opened 2020-09-27
Add Gateway extensions field.

### LINKED ISSUE #272: ListenerExtensionObjectReference is not used
**What happened**:

ListenerExtensionObjectReference is not used anywhere. Looks like it was dropped as part of #193 .

There should be a way to specify extended semantics for Listeners or Gateways.

jpeach:
Add an extensions array to the GatewaySpec. This allows implementations
to support additional policy and configuration for Gateways.

This fixes #272


hbagdi:
James, I'm trying to understand the need for this field. What sort of use cases require this?

robscott:
Also interested in potential use cases here. Although we already have `extensionRef` scattered throughout the API, I'm hesitant to add yet another extension point. Even though this is an alpha API, it will always be easier to add a field in the future than to remove one.

jpeach:
> James, I'm trying to understand the need for this field. What sort of use cases require this?

This would be for config that applies to a proxy globally, e.g.
- network access control
- logging policy and format

robscott:
@jpeach Could those concepts be configured with GatewayClass parameters instead?

jpeach:
> @jpeach Could those concepts be configured with GatewayClass parameters instead?

That would lead to a proliferation of classes. IIUC, classes are more the domain of the infrastructure provider, but gateways would be configured by cluster operators. I've been thinking of classes as fairly static and congruent with the lifetime of the cluster.

hbagdi:
Would these rather be part of the core Service resource that represents a Gateway (for in-cluster proxies)?

bowei:
ExtensionRef might make less sense on a whole object level. ExtensionRef really is needed to decorate subtrees.  And yes, inconsistency is ugly, this is definitely something we should try to rationalize going forward. There you have other ways to do things:

* name mapped decorator (a la Service)
* annotations

jpeach:
Discussed in the WG; consensus was that we would prefer to document extension techniques outlined [here](https://github.com/kubernetes-sigs/service-apis/pull/351#issuecomment-701541809) than to add explicit extension refs.

jpeach:
Closing, since WG would prefer to use other gateway extension techniques.

PR #352, opened 2020-09-27
Fix codegen script to work with Go modules.

### LINKED ISSUE #339: Client codegen doesn't work with Go modules
**What happened**:

Run `.hack/update-codegen.sh`. Notice that it puts the generated code in `~/go/src/sigs.k8s.io/service-apis`, not my actual git repository.

It should update the generated code in my git workspace, wherever that is. Note that deepcopy functions are also generated by `controller-gen` which does work with Go modules. Unfortunately `controller-gen` doesn't generate typed clients or listers, etc.

jpeach:
To make the codegen script work wil Go modules, we first need to add
the generator module dependencies by referencing them in tools.go.
There doesn't seem to be a c0.19.0 release, so I bumped everything
up to v0.19.2.

The generator always writes to the $GOPATH, so faking one up in the script
is necessary for developers who don't check out the repository in the
traditional Go source hierarchy.

Finally, switch to "go run" so as to not pollute the $GOBIN directory,
format the command arguments consistently and apply shellcheck
suggestions.

This fixes #339


hack/update-codegen.sh
```diff
@@ -18,42 +18,64 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
+readonly SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE}")"/.. && pwd)"
 
-go install k8s.io/code-generator/cmd/{client-gen,lister-gen,informer-gen,deepcopy-gen,register-gen}
+readonly GO111MODULE="on"
+readonly GOFLAGS="-mod=vendor"
```
howardjohn, line 24: Why do we need the vendor?

jpeach: It's about getting consistent behavior. We don't want this script to update `go.mod`, so I think that either `mod=vendor` or `mod=readonly` would probably be OK.

> By default, if the go version in go.mod is 1.14 or higher and a vendor directory is present, the go command will act as if -mod=vendor were used. Otherwise, the go command will act as if -mod=mod were used.

https://golang.org/ref/mod#build-commands

Makefile
```diff
@@ -66,7 +66,8 @@ uninstall:
 # Run static analysis.
 .PHONY: verify
 verify:
-	hack/verify-all.sh
+	go mod vendor
```
bowei, line 69: It's usually bad form to have the Makefile create potentially version controller files.
Can we wrap this using docker so the files downloaded are temporary?

hack/update-codegen.sh
```diff
@@ -18,42 +18,64 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
+readonly SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE}")"/.. && pwd)"
```
bowei, line 21: interesting -- is this to get around missing readlink on OS X

Makefile
bowei: Or we should gitignore and make clean vendor?

hack/update-codegen.sh
jpeach: `-mod=readonly` seems to work fine.

Makefile
jpeach: Using `-mod=readonly` removes the need to populate vendor.

hack/update-codegen.sh
jpeach: I just need an absolute path for making the symlink later. Not sure how robust it is compared to using readlink but it has always worked for me :)

robscott:
Thanks for your work on this! Have verified that it works locally for me.

/approve

hbagdi:
@jpeach Could you rebase this one so that we can get it in?

robscott:
@hbagdi I think this PR has already been rebased, just wanted someone else to verify/LGTM it.

hbagdi:
Sorry, read the timeline in the UI wrong.

I tested it by blowing away the generated code and it seems to work for me too.

/lgtm

PR #353, opened 2020-09-28
Flattening Routes

robscott:
This is a subset of #336.

/cc @hbagdi @danehans @jpeach


apis/v1alpha1/httproute_types.go
```diff
@@ -19,18 +19,34 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// +genclient
```
danehans, line 22: See https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r496850613

apis/v1alpha1/httproute_types.go
```diff
@@ -173,11 +190,14 @@ type HTTPRouteRule struct {
...
 	// +optional
-	Filters []HTTPRouteFilter `json:"filters"`
+	// +kubebuilder:validation:MaxItems=16
+	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
-	// Forward defines the upstream target(s) where the request should be sent.
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	Forward *HTTPForwardingTarget `json:"forward"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 198: Is min required for optional fields?

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// BackendRef is a local object reference to mirror matched requests to. If
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
+	//
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
```
danehans, line 474: Should `LocalObjectReference` be a pointer since it's optional?

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
 	//
 	// Support: Core
 	//
 	// +optional
-	TargetPort *TargetPort `json:"targetPort,omitempty"`
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 486: Same as https://github.com/kubernetes-sigs/service-apis/pull/353/files#r496895810

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
 	//
-	// Support: custom
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
```
danehans, line 515: Same as https://github.com/kubernetes-sigs/service-apis/pull/353/files#r496896418

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+
+	// Port specifies the destination port number to use for the backend. If
```
danehans, line 517: Same as https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r496867788

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
+	//
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
+
+	// Port specifies the destination port number to use for the backend. If
```
danehans, line 476: Same as https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r496867788

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
+	//
+	// Support: Core
 	//
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef"`
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 527: Same as https://github.com/kubernetes-sigs/service-apis/pull/353/files#r496895810

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// +kubebuilder:validation:Maximum=65535
+	Port *int32 `json:"port,omitempty"`
+
+	// Weight specifies the proportion of traffic forwarded to a backend,
```
danehans, line 531: The same as https://github.com/kubernetes-sigs/service-apis/pull/353/files#r496897901 could be said for weight.

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	// computed as weight/(sum of all weights in this ForwardTo list). Weight is
+	// not a percentage and the sum of weights does not need to equal 100. If
+	// only one backend is specified, 100% of the traffic is forwarded to that
+	// backend. If unspecified, weight defaults to 1.
+	//
+	// Support: Core (HTTPRoute)
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 540: Is min needed when defaulting is being used?

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +447,111 @@ type HTTPRequestHeaderFilter struct {
...
+	Weight int32 `json:"weight,omitempty"`
+
+	// Filters defined at this-level should be executed if and only if the
+	// request is being forwarded to the backend defined here.
+	//
+	// Conformance: Filtering support, including core filters, is NOT guaranteed
+	// at this-level. Use Filters in HTTPRouteRule for portable filters across
+	// implementations.
+	//
+	// Support: Customf
```
danehans, line 551: s/Customf/Custom/

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,60 @@ type GatewayReference struct {
...
+	// BackendRef is a reference to a backend to forward matched requests to. If
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
+	//
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
```
danehans, line 91: Pointer?

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,60 @@ type GatewayReference struct {
...
+	// unspecified and a Service object consisting of a single port definition
+	// is the backend, that port will be used. If unspecified and the backend is
+	// a Service object consisting of multiple port definitions, controllers
+	// must set the "InvalidRoutes" status condition on any Gateway that
+	// includes this Route to true.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 103: Is min needed for optional fields?

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,60 @@ type GatewayReference struct {
...
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
+	//
+	// Support: Extended
+	//
+	// +optional
+	BackendRef LocalObjectReference `json:"backendRef,omitempty"`
+
+	// Port specifies the destination port number to use for the backend. If
```
danehans, line 93: Same as https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r496867788

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,60 @@ type GatewayReference struct {
...
+	// includes this Route to true.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port *int32 `json:"port,omitempty"`
+
+	// Weight specifies the proportion of traffic forwarded to a backend,
```
danehans, line 107: Same as https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r496867788

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,60 @@ type GatewayReference struct {
...
+	// computed as weight/(sum of all weights in this ForwardTo list). Weight is
+	// not a percentage and the sum of weights does not need to equal 100. If
+	// only one backend is specified, 100% of the traffic is forwarded to that
+	// backend. If unspecified, weight defaults to 1.
+	//
+	// Support: Extended
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
```
danehans, line 116: Is min needed when defaulting is being used?

apis/v1alpha1/tcproute_types.go
```diff
@@ -55,42 +55,14 @@ type TCPRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the TCPRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef,omitempty"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 63: Is min=0 needed for optional fields?

apis/v1alpha1/tlsroute_types.go
```diff
@@ -63,39 +63,12 @@ type TLSRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the TLSRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef,omitempty"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 69: Is min=0 needed for optional fields?

apis/v1alpha1/udproute_types.go
```diff
@@ -53,37 +53,12 @@ type UDPRouteRule struct {
...
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the UDPRoute will be true.
-	//
-	// Support: custom
-	//
+	// ForwardTo defines the backend(s) where matching requests should be sent.
 	// +optional
-	ExtensionRef *RouteActionExtensionObjectReference `json:"extensionRef"`
+	// +kubebuilder:validation:MinItems=0
```
danehans, line 59: Is min=0 needed for optional fields?

danehans:
A few nits, otherwise LGTM.

apis/v1alpha1/httproute_types.go
robscott: I agree that we don't need to specify `MinItems` or `MinLength` for optional fields, but a `Minimum` is still important here so we ensure that if users do set this value it's not 0 or negative.

apis/v1alpha1/httproute_types.go
robscott: I agree that we don't need to specify MinItems or MinLength for optional fields, but a Minimum is still important here so we ensure that if users do set this value it's not 0 or negative.

apis/v1alpha1/httproute_types.go
robscott: I think so - you could still set a negative value here.

apis/v1alpha1/route_types.go
robscott: I agree that we don't need to specify MinItems or MinLength for optional fields, but a Minimum is still important here so we ensure that if users do set this value it's not 0 or negative.

apis/v1alpha1/route_types.go
robscott: I agree that we don't need to specify MinItems or MinLength for optional fields, but a Minimum is still important here so we ensure that if users do set this value it's not 0 or negative.

apis/v1alpha1/udproute_types.go
robscott: Good call, removed.

apis/v1alpha1/tlsroute_types.go
robscott: Good call, removed.

apis/v1alpha1/tcproute_types.go
robscott: Good call, removed.

apis/v1alpha1/route_types.go

apis/v1alpha1/route_types.go

apis/v1alpha1/route_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go

apis/v1alpha1/httproute_types.go
robscott: Will save that for #336 if that works for you. This isn't actually an addition, just moving the type to the top.

apis/v1alpha1/httproute_types.go

robscott:
Thanks for the feedback @danehans! I think I've responded to everything now, PTAL.

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +446,114 @@ type HTTPRequestHeaderFilter struct {
...
-	// To references referenced object(s) where the request should be sent. The
-	// resource may be "services" (omit or use the empty string for the
-	// group), or an implementation may support other resources (for
-	// example, resource "myroutetargets" in group "networking.acme.io").
-	// Omitting or specifying the empty string for both the resource and
-	// group indicates that the resource is "services".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the HTTPRoute will be true.
+// HTTPRouteForwardTo defines how a HTTPRoute should forward a request.
+type HTTPRouteForwardTo struct {
```
hbagdi, line 492: In HTTPRouteForwardTo and RouteForwardTo, 4 fields are common, does it make sense to perform type embedding here?

In HTTPRouteForwardTo, HTTPRequestMirrorFilter, 3 fields are common but this duplication could be okay if we de-duplicate the first case.

robscott: That a great idea. If it's OK, I think I'd prefer to do that in a follow up just to keep the momentum around this.

hbagdi: https://github.com/kubernetes-sigs/service-apis/issues/358 to track.

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +446,114 @@ type HTTPRequestHeaderFilter struct {
...
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the HTTPRoute will be true.
+	// BackendRef is a reference to a backend to forward matched requests to. If
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence. If the referent cannot be found, controllers must set the
+	// "InvalidRoutes" status condition on any Gateway that includes this Route
+	// to true.
 	//
-	// Support: custom
+	// Support: Extended
```
hbagdi, line 512: BackendRef is implementation specific so should this not be Custom?

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +446,114 @@ type HTTPRequestHeaderFilter struct {
...
+	Port *int32 `json:"port,omitempty"`
+
+	// Weight specifies the proportion of traffic forwarded to the backend
+	// referenced by the ServiceName or BackendRef field. computed as
+	// weight/(sum of all weights in this ForwardTo list). Weight is not a
+	// percentage and the sum of weights does not need to equal 100. If only one
+	// backend is specified, 100% of the traffic is forwarded to that backend.
+	// If unspecified, weight defaults to 1.
+	//
+	// Support: Core (HTTPRoute)
```
hbagdi, line 539: What does HTTPRoute mean here?

apis/v1alpha1/httproute_types.go
```diff
@@ -427,64 +446,114 @@ type HTTPRequestHeaderFilter struct {
...
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10000
+	Weight int32 `json:"weight,omitempty"`
```
hbagdi, line 544: I'm not sure but `omitempty` shouldn't be present here?
Maybe this belongs to the clean up PR.

apis/v1alpha1/route_types.go
```diff
@@ -64,6 +64,62 @@ type GatewayReference struct {
...
+	// percentage and the sum of weights does not need to equal 100. If only one
+	// backend is specified, 100% of the traffic is forwarded to that backend.
+	// If unspecified, weight defaults to 1.
+	//
+	// Support: Extended
+	//
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10000
+	Weight int32 `json:"weight,omitempty"`
```
hbagdi, line 120: Same as https://github.com/kubernetes-sigs/service-apis/pull/353/files#r497065424

apis/v1alpha1/httproute_types.go
robscott: I think this meets our current definition of "Extended":

> features are those that are portable but not universally supported across implementations. Those implementations that support the feature will have the same behavior and semantics.

Since this is just an object ref, it is inherently portable, even though each object referenced may not be supported by each implementation. At least that's my interpretation here, I may be misunderstanding the distinction here though.

apis/v1alpha1/route_types.go
robscott: Unfortunately every field with a default applied needs to be `omitempty` or `optional`. In this case `omitempty` seemed to make more sense here. Just tried this out to verify and got `missing required field "weight" in io.x-k8s.networking.v1alpha1.HTTPRoute.spec.rules.forwardTo`.

apis/v1alpha1/httproute_types.go
robscott: Unfortunately every field with a default applied needs to be `omitempty` or `optional`. In this case `omitempty` seemed to make more sense here. Just tried this out to verify and got `missing required field "weight" in io.x-k8s.networking.v1alpha1.HTTPRoute.spec.rules.forwardTo`.

apis/v1alpha1/httproute_types.go
hbagdi: > even though each object referenced may not be supported by each implementation

This is the reason I'm confused. To be included in Extended, shouldn't the behavior be same across even a subset of implementations?
In this case, unless there are a set of Backend-like CRDs that service-apis agrees upon and advises implementers to implement, this adds little value.

apis/v1alpha1/httproute_types.go
hbagdi: One day, I'll get the hang of this. I need to sit down and get to the bottom of all the different possibilities here.
Apologies for the noise.

robscott: No worries, I've been super confused as well on the appropriate combination of omitempty, optional, etc.

apis/v1alpha1/httproute_types.go
robscott: 👍 that makes sense, switched to Custom.

hbagdi:
This PR changes quite a bit. Merging this in and then we can do iterative improvements with smaller diffs.

Also, with HTTPRoute being flattened, it should be possible to write a CLI tool to convert an Ingress resource into HTTPRoute that works for 70-80% cases.

/lgtm

PR #360, opened 2020-09-30
Changing Resource to Kind, Removing Defaults for Group and Kind

robscott:
This fixes https://github.com/kubernetes-sigs/service-apis/issues/293 by replacing all uses of `resource` fields with `kind` fields to match the upstream pattern.

I've also gone one step further to remove all defaults in object references. I'm not 100% sold on removing the defaults forever, but prefer that direction now for the following reasons:

**1. Less potential confusion when looking at YAML manifests.** 

Although it's certainly simpler to just refer to a certificate by name, this might make it appear like there was an actual `certificate` resource we were referring to. I don't think it's entirely obvious what resource this is referring to:
```
certificateRef:
  name: foo
```

On the other hand, this approach is more verbose but also leaves very little room for misunderstanding.
```
certificateRef:
  name: foo
  kind: Secret
  group: core
```

**2. It's a lot easier to add defaults in the future than to remove them.**

Since this seems like something we're not entirely sure about, I'd rather leave it out of the v1alpha1 API. If we get a lot of feedback that defaults would be nice and/or we start to see patterns emerging for what groups and kinds are used consistently, we can add defaults back.

On the other hand, removing defaults would be a backwards incompatible change. That's not the end of the world for an alpha API, but I'd still prefer to avoid them wherever possible.


robscott:
One further note, this PR is based on #336. I'm going to hold this PR until that gets in. Anyone reviewing this PR should focus on the second commit until 336 gets in.

/hold

robscott:
Removing hold now that #336 has merged in.

/hold cancel

hbagdi:
> I've also gone one step further to remove all defaults in object references. It's always going to be easier to add defaults back in the future, but if we release an API with them in, removing them would become a backwards incompatible change. These defaults were not always obvious or clear and it dramatically simplifies our API types to remove them for now. In the future if we want to provide defaults, Tim's recommendation was to follow the pattern we've already followed with Routes - add something like serviceName as an alternative to a full ObjectRef.

- Having no defaults makes the API harder to use for the end user (and introduced another area for typos and errors). In my opinion, there are areas like Certificates and Services where having a default is an obvious choice for majority of use cases.
- I don't 100% agree with the reasoning that removing Go API types for defaulting has a dramatic effect on our API surface.


I don't feel too strongly against this but I'd like to better understand why not having defaults is the right thing to do here.

robscott:
@hbagdi those are good questions. I realized that my initial explanation in the PR description was not great, so I've updated that above. To answer your specific comments/questions:

> Having no defaults makes the API harder to use for the end user (and introduced another area for typos and errors). In my opinion, there are areas like Certificates and Services where having a default is an obvious choice for majority of use cases.

Although this would require a bit more YAML, my theory is that the extra verbosity will actually make the API easier to understand. This feels like a relatively common practice throughout k8s APIs. Are there any common APIs that allow you to leave out kind or group in object references?

> I don't 100% agree with the reasoning that removing Go API types for defaulting has a dramatic effect on our API surface.

You're completely right that it doesn't make a difference to users of the API, but it does make it slightly easier for maintainers and implementers of this API to have a few less types to support.

hbagdi:
I see that reasoning and now I can reason either ways about this. 

It is funny how we are in the phase of development where we are all removing features as much as possible. That's a good thing to do nearly always!

hbagdi:
Add lgtm label because we have consensus and overall looks good. Please let me know if we should have more eyes before a lgtm label in such cases in future.
/lgtm

PR #362, opened 2020-09-30
Updating RouteNamespaces to use RouteGateways pattern

robscott:
This fixes https://github.com/kubernetes-sigs/service-apis/issues/343 by replacing the boolean with a pattern similar to what we're using for RouteGateways. 

This PR is based on #336. I'm going to hold this PR until that gets in. Anyone reviewing this PR should focus on the second commit until 336 gets in.

/hold
/cc @bowei @danehans @hbagdi @jpeach


apis/v1alpha1/gateway_types.go
```diff
@@ -435,7 +435,7 @@ type RouteBindingSelector struct {
 	//
 	// Support: Core
 	//
-	// +kubebuilder:default={onlySameNamespace:true}
+	// +kubebuilder:default={from:SameNamespace}
```
robscott, line 438: Should we remove this default here and instead rely on controllers to interpret the default?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -167,6 +167,7 @@ type BackendTLSConfig struct {
 type BackendPolicyStatus struct {
 	// Conditions describe the current conditions of the BackendPolicy.
 	// +optional
+	// +kubebuilder:validation:MaxItems=8
```
danehans, line 170: Same feedback as https://github.com/kubernetes-sigs/service-apis/pull/336#discussion_r497816693

apis/v1alpha1/gateway_types.go
danehans: I can see a use cases for allowing all ns's or same ns only as a default, so +1 on not setting a default.

apis/v1alpha1/gateway_types.go
```diff
@@ -480,36 +480,41 @@ type RouteBindingSelector struct {
...
 
-	// OnlySameNamespace is a boolean used to indicate if Route references are
-	// limited to the same Namespace as the Gateway. When true, only Routes
-	// within the same Namespace as the Gateway should be selected.
-	//
-	// This field takes precedence over the NamespaceSelector field. That
-	// selector should only take effect when this field is set to false.
+	// Selector must be specified when From is set to "Selector". In that case,
+	// only Routes in Namespaces matching this Selector will be selected by this
+	// Gateway. This field is ignored for other values of "From".
```
danehans, line 512: I think it's worth mentioning in the docs that an empty selector will select all ns's. I also like the idea of adding an `AllNamespaces` enum to make this even clearer to users.

robscott: Yeah I think that does make sense. I added an `AllNamespaces` option here for simpler config.

apis/v1alpha1/gateway_types.go
robscott: I think it could make sense to remove the default but I think we'd still want to define a consistent default behavior between controllers (likely still the `from:SameNamespace` equivalent).

robscott: Updated to remove the default now.

robscott:
@bowei @danehans As suggested today, I've removed the default here and updated the examples. PTAL.

robscott:
Removing hold now that #336 has merged.

/hold cancel

apis/v1alpha1/gateway_types.go
```diff
@@ -481,36 +479,44 @@ type RouteBindingSelector struct {
 	Resource string `json:"resource"`
 }
 
+// GatewaySelectType specifies where Routes should be selected by a Gateway.
+// +kubebuilder:validation:Enum=AllNamespaces;Selector;SameNamespace
+// +kubebuilder:default=SameNamespace
+type GatewaySelectType string
```
hbagdi, line 485: nit: I was a little confused. Should this be RouteSelectType instead?

robscott: Good call, should be fixed now.

examples/tls-in-route.yaml
```diff
@@ -32,8 +32,7 @@ spec:
     routes:
       resource: httproutes
       routeNamespaces:
-        namespaceSelector: {}
-        onlySameNamespace: false
+        from: "AllNamespaces"
```
hbagdi, line 35: Reading this feels repetitive.

Is the following an improvement?

```yaml
routeNamespaces:
  from: All
```

So, instead the constants would be 'All', 'Selector' and 'Same'.

hbagdi (approved):
LGTM other than one comment.

examples/tls-in-route.yaml
robscott: Good call, updated with the simplified types.

robscott: And now I actually updated the examples too.

hbagdi:
LGTM.
@danehans Can you take one final look at this one?

apis/v1alpha1/gateway_types.go
```diff
@@ -430,9 +430,7 @@ type RouteBindingSelector struct {
 	// default.
 	//
 	// Support: Core
-	//
-	// +kubebuilder:default={onlySameNamespace:true}
-	RouteNamespaces RouteNamespaces `json:"routeNamespaces,omitempty"`
+	RouteNamespaces RouteNamespaces `json:"routeNamespaces"`
```
jpeach, line 433: Do we need a default here?

PR #365, opened 2020-10-01
Fix the type of the ListenerStatus.Port.

### LINKED ISSUE #357: ListenerStatus.Port should probably be an int
@danehans has pointed out that it is confusing for `ListenerStatus.Port` to be a string instead of an int. Currently it has this description:

```
	// Port is the unique Listener port value for which this message
	// is reporting the status. If more than one Gateway Listener
	// shares the same port value, this message reports the combined
	// status of all such Listeners.
```

@jpeach or @hbagdi can you think of any reason not to make this an int?

Context: https://github.com/kubernetes-sigs/service-apis/pull/336/files#r496830370

jpeach:
This fixes #357

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -703,7 +703,7 @@ type ListenerStatus struct {
 	// status of all such Listeners.
 	//
 	// +required
-	Port string `json:"port"`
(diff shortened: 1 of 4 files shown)


robscott:
Thanks!

/lgtm
/approve

PR #366, opened 2020-10-01
deprecate the feedback page update issue link

### LINKED ISSUE #334: Update feedback page
This should primarily involve removing `TODO` and providing better guidance on when/how to file an issue. It would also be good to include a reference to our Slack channel.

eth0xFEED:
resolves #334


eth0xFEED:
@robscott It was not very clear how to add arbitrary header links we've already used the `repo_url` field of the mkdocs.yml configuration file. Do you think we need to create a custom template patch to get this in for resolving the intent of issue #334 ?

mkdocs.yml
```diff
@@ -5,6 +5,10 @@ site_dir: docs
 docs_dir: docs-src
 theme:
   name: material
+  icon:
+    repo: fontawesome/brands/github
+  features:
+    - navigation.tabs
```
robscott, line 11: What does this do?

robscott:
@eth0xFEED Thanks for this PR! It does look like something that is specific to the `repo_name` field and would require some customization to do anything more than that. I don't think it's particularly urgent to do anything beyond this PR for now.

mkdocs.yml
eth0xFEED: Unfortunately nothing though at one time i believed it would potentially help 🤦  
... sorry about that will fix up. 👍

robscott:
Thanks!

/lgtm
/approve

PR #367, opened 2020-10-01
Fixes Boilerplate Generation and Verification

danehans:
- Removes duplicate `boilerplate.go.txt` file and updates references to use `hack/boilderplate/boilerplate.go.txt`.
- Removes boilerplate from `.md` files based on [this](https://github.com/kubernetes-sigs/service-apis/pull/367#discussion_r498542402) feedback.
- Manually updates all non-generated `.go` files to comply with boilerplate verification.

__Notes:__
- According to `boilerplate.py` dates can be 2014, 2015, 2016, ..., CURRENT_YEAR, company holder names can be anything.
- According to `update-codegen.sh` the service-apis repo must be cloned in `$GOPATH/src/sigs.k8s.io` for generators to properly update boilerplate in generated code.

Fixes: https://github.com/kubernetes-sigs/service-apis/issues/345

/assign @robscott @bowei @hbagdi @jpeach 
/cc @Miciah


CONTRIBUTING.md
```diff
@@ -1,3 +1,19 @@
+<!--
```
robscott, line 1: I haven't seen many cases where this header is added to markdown files, is there a way to exclude md and/or only include go files?

danehans: @robscott I can remove the boilerplate for `.md` files. I added this since I noticed some `docs.src/*.md` files contain the boilerplate and others don't. Instead of adding the boilderplate to `.md` files, should it be removed? I'm fine either way as long as we have consistency.

robscott: Good to know, I didn't realize that we had that header in any markdown files. Taking a look at k/k and k/website it looks like upstream markdown files do not contain the boilerplate. I think it makes sense to follow that precedent here by removing the header from markdown files.

danehans:
@robscott commit `9c80167` addresses your [feedback](https://github.com/kubernetes-sigs/service-apis/pull/367#discussion_r498542402).

robscott:
Thanks!

/lgtm
/approve

PR #371, opened 2020-10-03
Update Listener condition types.

jpeach:
Update the condition types and reasons for the Listener condition type.

/cc @danehans @robscott

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -87,7 +87,7 @@ type GatewaySpec struct {
 	//
 	// If this field specifies multiple Listeners that have the same
 	// Port value but are not compatible, the GatewayClass must raise
-	// a "PortConflict" condition on the Gateway.
+	// a "Conflicted" condition in the Listener status.
 	//
 	// Support: Core
 	//
@@ -284,11 +284,11 @@ const (
 	HostnameMatchAny HostnameMatchType = "Any"
 )
 
-// ProtocolType defines the application protocol accepted by a
-// Listener. Implementations are not required to accept all the
-// defined protocols. If an implementation does not support a
-// specified protocol, it should raise a "ConditionUnsupportedProtocol"
-// condition for the affected Listener.
+// ProtocolType defines the application protocol accepted by a Listener.
+// Implementations are not required to accept all the defined protocols.
+// If an implementation does not support a specified protocol, it
+// should raise a "Detached" condition for the affected Listener with
+// a reason of "UnsupportedProtocol".
 //
 // Valid ProtocolType values are:
 //
@@ -591,6 +591,8 @@ type GatewayStatus struct {
 	// Listeners provide status for each unique listener port defined in the Spec.
 	//
 	// +optional
+	// +listType=map
+	// +listMapKey=port
 	// +kubebuilder:validation:MaxItems=64
 	Listeners []ListenerStatus `json:"listeners,omitempty"`
 }
@@ -618,7 +620,7 @@ const (
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
-	// interoperability.`
+	// interoperability.
 	GatewayConditionScheduled GatewayConditionType = "Scheduled"
 
 	// GatewayReasonNotReconciled is used when the Gateway is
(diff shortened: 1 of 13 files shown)


jpeach:
ref to original doc https://docs.google.com/document/d/17WGAL5jjjGHsmDCeMstZ3L5fFp6QGPHrIplQ9LPZ91k

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
...
 
+// ListenerConditionReason defined the set of reasons that explain
+// why a particular Listener condition type has been raised.
+type ListenerConditionReason string
+
+const (
+	// ListenerConditionConflicted indicates that the controller
+	// was unable to resolve conflicting specification requirements
+	// for this Listener. If a Listener is conflicted, its network
+	// port should not be configured on any network elements.
```
robscott, line 728: I think we need a bit more detail here. What if a conflict arises after something has already been provisioned? Our standard practice here has been to accept one valid config in a conflicting situation based on those guiding principles for conflict resolution.

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
 // field.
 type ListenerConditionType string
 
+// ListenerConditionReason defined the set of reasons that explain
```
robscott, line 720: Nit: "defines" to match the tense used elsewhere.

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
...
+	// For example, a Listener that specifies "UDP" as the protocol
+	// but a route selector that resolves "TCPRoute" objects.
+	ListenerReasonRouteConflict ListenerConditionReason = "RouteConflict"
+)
+
+const (
+	// ListenerConditionDetached indicates that, even though
+	// the listener is syntactically and semantically valid, the
+	// controller is not able to configure it on the underlying
+	// Gateway infrastructure.
```
robscott, line 765: What happens if the Gateway was already valid and something changes to arrive at this condition? This almost makes it sound like the Gateway would not be provisioned/updated at all.

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
...
+const (
+	// ListenerConditionResolvedRefs indicates whether the
+	// controller was able to resolve all the object references
+	// for the Listener.
+	//
+	// Possible reasons for this condition to be false are:
+	//
+	// * "DroppedRoutes"
+	// * "InvalidCertificateRef"
+	// * "InvalidRoutesRef"
```
robscott, line 809: Maybe we should provide some guidance as far as the order of when these should be applied if they are all true? (Assuming only one entry for this condition will exist at a given time).

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
...
+	//
+	// * "Invalid"
+	// * "Pending"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ListenerConditionReady ListenerConditionType = "Ready"
+
+	// ListenerReasonInvalid is used when the Listeners is
```
robscott, line 849: Nit: Listener should probably be singular here.

apis/v1alpha1/gateway_types.go
```diff
@@ -717,38 +717,140 @@ type ListenerStatus struct {
...
+	// syntactically or semantically invalid.
+	ListenerReasonInvalid ListenerConditionReason = "Invalid"
+
+	// ListenerReasonPending is used when the Listeners is not
```
robscott, line 853: Nit: Listener should probably be singular here.

apis/v1alpha1/httproute_types.go
```diff
@@ -325,9 +325,13 @@ type HTTPRouteMatch struct {
 	// string for the group) or an implementation-defined resource (for
 	// example, resource "myroutematchers" in group "networking.acme.io").
 	// Omitting or specifying the empty string for both the resource and
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the HTTPRoute will be true.
+	// group indicates that the resource is "configmaps".
+	//
+	// If the referent cannot be found, the route must be dropped
+	// from the Gateway. The controller should raise the "ResolvedRefs"
```
robscott, line 331: This seems like it might be a bit too harsh. If someone has a large route and adds an invalid extension ref to one of many matches, we may not want to drop the entire route.

apis/v1alpha1/httproute_types.go
```diff
@@ -453,9 +457,13 @@ type HTTPRequestMirrorFilter struct {
 	// ServiceName refers to the name of the Service to mirror matched requests
 	// to. When specified, this takes the place of BackendRef. If both
 	// BackendRef and ServiceName are specified, ServiceName will be given
-	// precedence. If the referent cannot be found, controllers must set the
-	// "InvalidRoutes" status condition on any Gateway that includes this
-	// Route to true.
+	// precedence.
+	//
+	// If the referent cannot be found, the route must be dropped
+	// from the Gateway. The controller should raise the "ResolvedRefs"
```
robscott, line 463: Same thoughts about this potentially being too harsh.

apis/v1alpha1/httproute_types.go
```diff
@@ -465,22 +473,33 @@ type HTTPRequestMirrorFilter struct {
 
 	// BackendRef is a local object reference to mirror matched requests to. If
 	// both BackendRef and ServiceName are specified, ServiceName will be given
-	// precedence. If the referent cannot be found, controllers must set the
-	// "InvalidRoutes" status condition on any Gateway that includes this Route
-	// to true.
+	// precedence.
+	//
+	// If the referent cannot be found, the route must be dropped
+	// from the Gateway. The controller should raise the "ResolvedRefs"
```
robscott, line 479: Especially in the case of mirroring, it seems appropriate to forego mirroring but leave the rest of the functionality intact.

apis/v1alpha1/httproute_types.go
```diff
@@ -465,22 +473,33 @@ type HTTPRequestMirrorFilter struct {
...
+	// of a single port definition is the backend, that port will
+	// be used.
+	//
+	// If the port is unspecified and the backend is a Service
+	// object consisting of multiple port definitions, the route
+	// must be dropped from the Gateway. The controller should
+	// raise the "ResolvedRefs" condition on the Gateway with the
+	// "DroppedRoutes" reason.  The gateway status for this route
+	// should be updated with a condition that describes the error
+	// more specifically.
```
robscott, line 502: I think this is probably the best we can do, but just want to make sure there's no better alternative here. For example, this could be quite bad if a user adds a new port to a Service before adding that port to route definitions, it could break all the things.

apis/v1alpha1/httproute_types.go
```diff
@@ -495,9 +514,13 @@ type HTTPRouteForwardTo struct {
...
-	// precedence. If the referent cannot be found, controllers must set the
-	// "InvalidRoutes" status condition on any Gateway that includes this
-	// Route to true.
+	// precedence.
+	//
+	// If the referent cannot be found, the route must be dropped
+	// from the Gateway. The controller should raise the "ResolvedRefs"
+	// condition on the Gateway with the "DroppedRoutes" reason.
+	// The gateway status for this route should be updated with a
+	// condition that describes the error more specifically.
```
robscott, line 523: If there were multiple forwardTos specified here, would it be safer to defer to any valid ones than to drop the route entirely?

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/gateway_types.go
jpeach: There's actually no way to express that a port has both working Listeners and conflicted Listeners. The `ListenerStatus` is a slice that is keyed by port so if you had a working configuration on the port, then an update introduced a conflict, I don't see how that can be expressed.

apis/v1alpha1/gateway_types.go
jpeach: This is a similar problem to the conflicts. Say you have a Listener with protocol "TCP" and the operator changes the protocol top "FOU". The gateway may want to reject that change, but there's no way to show that the protocol is still "TCP" while also expressing that the "FOU" protocol change hasn't been applied. If the conditions are all positive, then the operator would expect that the "FOU" protocol is satisfactorily configured.

I suppose that a controller could say the listener is Detached/UnsupportedProtocol and still leave the "TCP" protocol in place. I'm not sure how I feel about that, since it's requiring Spec to vary from Status in a semi-permanent way.

apis/v1alpha1/gateway_types.go
jpeach: > Assuming only one entry for this condition will exist at a given time

Yes, Condition semantics are that each type is always present, exactly once.

apis/v1alpha1/httproute_types.go
jpeach: Maybe. This introduces a new failure mode that we don't have any way to express - the route was accepted but some part of it was elided with unknown consequences.

Would we want to specify different error paths depending on the reference? For example, `.HTTPRouteSpec.ExtensionRef` seems like a good place to hang an external authorization configuration. If that fails to resolve, is it better to drop the route or program the route without authorization? Or should the controller make a decision based on the type and context of the reference?

apis/v1alpha1/gateway_types.go
robscott: OK, that's a good clarification. This simply indicates that this specific listener would be deprovisioned in the event of invalid configuration here?

jpeach:
To summarize a few points from a slack conversation between @robscott and myself:

* Dropping whole routes works against our principle for updates to break as little as possible
* Not every reference within the route seems to have the same importance, e.g. maybe it doesn't matter that much if mirroring isn't working
* Match references seem more critical than mirror references
* Failing to resolve some of multiple forwardTo targets could be ok in practice, though it's a grey failure (may end up costing more to resolve?)
* Sending all traffic to my canary might still be better than sending no traffic
* We can probably construct thought experiments for each reference where it does matter, so the API is in some sense oblivious to how bad these kinds of errors are
* Routes could be considered a well-defined failure domain, so operators can define them wisely in that case
* There is a benefit to a standard, consistent behavior for failing to resolve references
* If we don't drop the route, how do we expose the failure? Having the state silently not match the spec seems bad.

apis/v1alpha1/gateway_types.go
```diff
@@ -709,38 +711,140 @@ type ListenerStatus struct {
...
+	// * "RouteConflict"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ListenerConditionConflicted ListenerConditionType = "Conflicted"
+
+	// ListenerReasonHostnameConflict is used when the Listener
+	// violates the Hostname match constraints for that allow
+	// collapsing Listeners. For example, this reason would be
```
Miciah, line 737: I'm having difficulty parsing "for that allow".  Is this saying that there is a set of conditions that necessitate collapsing listeners and a set of conditions that prohibit collapsing listeners, and `HostnameConflict` indicates that both sets of conditions were met for the same group of listeners?

jpeach: Good catch! This should be either "violates the Hostname match constraints that allow collapsing" or "violates the Hostname match constraints for collapsing". Looks like I got stuck half-way :)

apis/v1alpha1/gateway_types.go
```diff
@@ -709,38 +711,140 @@ type ListenerStatus struct {
...
+	// * "InvalidRoutesRef"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ListenerConditionResolvedRefs ListenerConditionType = "ResolvedRefs"
+
+	// ListenerReasonDroppedRoutes indicates that not all of
+	// the routes selected by this Listener could not be
+	// configured. The specific reason why each route was dropped
```
Miciah, line 812: Too many nots.
```suggestion
	// ListenerReasonDroppedRoutes indicates that not all of
	// the routes selected by this Listener could be
	// configured. The specific reason why each route was dropped
```

apis/v1alpha1/gateway_types.go
```diff
@@ -709,38 +711,140 @@ type ListenerStatus struct {
...
+
+	// ListenerReasonInvalidCertificateRef is used when the
+	// Listener has a TLS configuration with a TLS CertificateRef
+	// that is invalid or cannot be resolved.
+	ListenerReasonInvalidCertificateRef ListenerConditionReason = "InvalidCertificateRef"
+
+	// ListenerReasonInvalidRoutesRef is used when the Listener's
+	// Routes selector is invalid or cannot be resolved. Note that
+	// it is not an error for this selector to not resolve any
+	// Routes and the ListenerConditionResolvedRefs should not be
```
Miciah, line 824: Either "the" should be deleted or "status condition" should be added.  Also, I suggest referring to the status condition as "ResolvedRefs" rather than by the Go symbol name as `kubectl explain` shows the former.
```suggestion
	// Routes, and the "ResolvedRefs" status condition should not be
```

apis/v1alpha1/gateway_types.go
```diff
@@ -709,38 +711,140 @@ type ListenerStatus struct {
...
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ListenerConditionResolvedRefs ListenerConditionType = "ResolvedRefs"
+
+	// ListenerReasonDroppedRoutes indicates that not all of
+	// the routes selected by this Listener could not be
+	// configured. The specific reason why each route was dropped
+	// should be indicated in the route's .Status.Conditions field.
+	ListenerReasonDroppedRoutes ListenerConditionType = "DroppedRoutes"
```
Miciah, line 814:
```suggestion
	ListenerReasonDroppedRoutes ListenerConditionReason = "DroppedRoutes"
```

robscott:
Thanks!

/approve

PR #372, opened 2020-10-04
Consistently apply list-map typing to Conditions slices.

### LINKED ISSUE #342: Investigate patch merge types annotations for conditions
Condition slices should get the following merge annotations:

```
	// +patchMergeKey=type
	// +patchStrategy=merge
	// +listType=map
	// +listMapKey=type
```

However, this doesn't seem to result in any changes to the CRD. Investigate and fix (if possible).

jpeach:
Patch merge strategy is not supported for CRDs, so we can drop those
markers. Apply the list-map markers to all Condition slices. Conditions
should never be optional, since a controller should ensure that all
Condition types are always present in the status field.

This fixes #342 

/cc @robscott @bowei @danehans


robscott:
Thanks!

/lgtm
/approve

PR #373, opened 2020-10-06
httproute: remove extension points in spec

hbagdi:


apis/v1alpha1/httproute_types.go
```diff
@@ -107,19 +107,6 @@ type HTTPRouteSpec struct {
...
-	// resource (for example, resource "myroutehosts" in group "networking.acme.io").
-	//
-	// If the referent cannot be found,
-	// the GatewayClass's "InvalidParameters" status condition
-	// will be true.
-	//
-	// Support: custom
-	//
-	// +optional
-	ExtensionRef *RouteHostExtensionObjectReference `json:"extensionRef,omitempty"`
```
hbagdi, line 122: This extension point seems extraneous and is not present in other route types. Should implementers choose to go for a specific Route type if this is needed?

apis/v1alpha1/httproute_types.go
```diff
@@ -319,20 +306,6 @@ type HTTPRouteMatch struct {
...
-	// example, resource "myroutematchers" in group "networking.acme.io").
-	// Omitting or specifying the empty string for both the resource and
-	// group indicates that the resource is "configmaps".  If the referent
-	// cannot be found, the "InvalidRoutes" status condition on any Gateway
-	// that includes the HTTPRoute will be true.
-	//
-	// Support: custom
-	//
-	// +optional
-	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
```
hbagdi, line 335: The reasoning for removing this extension point stems from the case that if we chose to keep this extension point, then the routing definition will be defined in HTTPRoute but the match will be defined in another CRD. This can get cumbersome for users to read and maintain.

On another note, if we remove this extension point, then does having extension points inside TCPRoute, UDPRoute and TLSRoute at the same level make sense?

apis/v1alpha1/httproute_types.go
robscott: I completely agree with removal of this one. As discussed last week, anyone wanting to add extension points that were 1:1 with the route like this could use annotations or name-mapping.

apis/v1alpha1/httproute_types.go
robscott: I'm not quite as sure about this removal since we have extension points at the same place on other routes and this would be harder to replicate through other methods like name-mapping or annotations. With that said, one could always create/support custom routes for different forms of matching.

hbagdi: I know this is a little bit of a contention point and I'm not sure about this either.
I've removed this from the scope of this PR right now.
Tracking this in https://github.com/kubernetes-sigs/service-apis/issues/359.

robscott:
Thanks!

/approve

hbagdi:
@danehans @jpeach Could you take a look at this one?

PR #374, opened 2020-10-07
document NamedAddress

rosenhouse:


robscott:
Thanks!

/ok-to-test
/approve

jpeach:
This is documented just below, do you think that summarizing here is needed too?

https://github.com/kubernetes-sigs/service-apis/blob/9c92ae4817fcf62a80285a246fd4913774600bd6/apis/v1alpha1/gateway_types.go#L536-L552

jpeach:
deeeeeeeeeeeeeeeeeeeeeeeebvvvvvvvvbv        sda---------------------=r

rosenhouse:
@jpeach The code comments you quoted do not appear in the docs page (try doing an in-page text search for `is an address selected by name` at [this url](https://kubernetes-sigs.github.io/service-apis/spec/).  I'm not clear on how I'd do that if I wanted to.

Also: I was trying to understand what an example of a  `NamedAddress` might look like ([slack thread](https://kubernetes.slack.com/archives/CR0H13KGA/p1602028630063400)).  As someone new to the project, having an extra note giving the example of "static" or "elastic" IP helped me better understand how this field is meant to be used.

jpeach:
> @jpeach The code comments you quoted do not appear in the docs page 

https://github.com/ahmetb/gen-crd-api-reference-docs/issues/21 

:(

jpeach:
This is a general problem with the online docs. We have a lot of API-relevant information attached to constants in the Go code which isn't exposed in the HTML docs. It doesn't seem practical to hoist all that up to the corresponding types, and in some cases the types aren't anchored all the way up to a CRD spec (e.g. we don't constrain condition types using the type system).

robscott:
@jpeach Should we restructure how we document our code to so the reference docs are more useful? We could either duplicate documentation or move the majority of the description/types to where they are actually used.

robscott:
@rosenhouse @jpeach Although I don't love duplicating documentation, I think it's worth it for the sake of understandable reference docs.

apis/v1alpha1/gateway_types.go
```diff
@@ -527,8 +527,10 @@ type GatewayAddress struct {
 // AddressType defines how a network address is represented as a text string.
 // Valid AddressType values are:
 //
-// * "IPAddress"
-// * "NamedAddress"
+// * "IPAddress": an IPv4 or IPv6 address
+// * "NamedAddress": an opaque identifier, controller-dependent
```
robscott, line 531: Nit: I think it's worth updating the text here to more closely reflect the descriptions below.

hbagdi:
@rosenhouse A friendly ping to get this PR moving forward.

rosenhouse:
@hbagdi I think I've got it cleaned up, please take a look again.

apis/v1alpha1/gateway_types.go
rosenhouse: done

PR #376, opened 2020-10-07
Decreasing min weight from 1 to 0

robscott:
This was suggested by @mark-church as a way to simplify blue green deployments or rollbacks. Users may want to start/end with a weight of 0 to simplify the workflow. 

/cc @danehans @hbagdi @jpeach

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -557,12 +557,13 @@ type HTTPRouteForwardTo struct {
 	// weight/(sum of all weights in this ForwardTo list). Weight is not a
 	// percentage and the sum of weights does not need to equal 100. If only one
 	// backend is specified, 100% of the traffic is forwarded to that backend.
+	// If weight is set to 0, no traffic should be forwarded for this entry.
 	// If unspecified, weight defaults to 1.
 	//
 	// Support: Core
 	//
 	// +kubebuilder:default=1
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=10000
 	Weight int32 `json:"weight,omitempty"`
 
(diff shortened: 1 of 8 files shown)


howardjohn (approved):
LGTM

apis/v1alpha1/httproute_types.go
```diff
@@ -528,7 +528,7 @@ type HTTPRouteForwardTo struct {
 	// Support: Core
 	//
 	// +kubebuilder:default=1
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
```
hbagdi, line 531: Do we need a doc line to indicate that the target is ignored if weight is zero or is that self-explanatory?

robscott: I thought about that briefly but thought it was self explanatory, don't mind adding it in though if that helps clear up anything.

danehans: I agree with @hbagdi, let's be explicit by having a sentence explaining `weight: 0` ignores the target.

jpeach: What happens when the total weight is 0?

robscott: Thanks for the feedback! I've updated this to add the following sentence: 

> If weight is set to 0, no traffic should be forwarded for this entry.

I thought about adding one for if the total weight is 0, but every way I tried to say it sounded redundant:

> If all weights in this list are set to 0, no traffic should be forwarded for this forwardTo.

Since this field is optional, implementations already have to support a lack of a `forwardTo` target here, so I don't think dropping the min weight to 0 here significantly changes how this API would be implemented.

jpeach: Yup, I agree the comment here is clear enough

PR #378, opened 2020-10-09
Change type alias to creating a new type

bowei:
Fixes the documentation generation issue.

### DIFF
--- a/apis/v1alpha1/gatewayclass_types.go
+++ b/apis/v1alpha1/gatewayclass_types.go
@@ -96,14 +96,8 @@ type GatewayClassSpec struct {
 	// Support: Custom
 	//
 	// +optional
-	ParametersRef *GatewayClassParametersObjectReference `json:"parametersRef,omitempty"`
+	ParametersRef *LocalObjectReference `json:"parametersRef,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


hbagdi (approved):

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -103,7 +103,7 @@ type GatewayClassSpec struct {
 // resource for a GatewayClass.
 //
 // +k8s:deepcopy-gen=false
-type GatewayClassParametersObjectReference = LocalObjectReference
+type GatewayClassParametersObjectReference LocalObjectReference
```
robscott, line 106: Is there any point in leaving these aliases in place or should we just use `LocalObjectReference` directly? I think we're already doing that for a lot of references in the API.

jpeach: Yeh, just using `LocalObjectReference` directly makes it easier to read IMHO

bowei: Don't really have a preference, I can change it.

bowei:
Updated

Issue #190, opened 2020-05-12
Regex support in HTTP path and header matching


hbagdi:
Regular expression support for route matching is a common feature request.

A little bit of history:

Ingress v1beta1 had regex support for posix regex and also had strict(er) path validation.
This was seen as a limitation because some implementation didn't support posix regex but instead supported PCRE regex.

We need to clarify follow in our API:
1. Should regex support be in the "core", "extended" or "custom" part of the API?
2. [Currently](https://github.com/kubernetes-sigs/service-apis/blob/96e09e85cc0ced3fb6f41dbd16d885a42970e11b/api/v1alpha1/httproute_types.go#L90-L91), we have a `RegularExpression` and `ImplementationSpecific` as two different path types. Does it make sense to fold Regex into Implementation specific? Or should there be `PCRERegex` and `POSIXRegex` as two different path types, there by providing a little more portability between implementations?

hbagdi:
@jpeach has also suggest to add a "contains" HeaderMatchType, which could be supported by the core API itself and be easily portable across implementations.

hbagdi:
cc @aledbf

robscott:
@hbagdi are you still able to take this one on?

PR #379, opened 2020-10-09
Clean up match type and add regex header match

hbagdi:
ImplementationSpecific and RegularExpression are kept separate for
end-user documentation purposes. There is some expectation of
portability for some regexes across implementation in future.


hbagdi:
Fixes #190

robscott:
Thanks!

/approve
/cc @danehans or @jpeach for LGTM

PR #381, opened 2020-10-09
remove dependency on kustomize binary

hbagdi:
kustomize is already included in kubectl.


robscott:
Thanks!

/lgtm
/approve

Issue #170, opened 2020-04-30
Define or discuss lifetime coupling between Gateways and their GatewayClass


jpeach:
When a GatewayClass is deleted, the spec does not address what happens to any Gateways that refer to that class.

I can think of 3 reasonable options:
1. Do nothing, orphaning the Gateways
2. Prevent the deletion of the GatewayClass while the class still owns Gateways
3. Delete all the owned Gateways as a side-effect

(1) is how many Ingress controllers work today, but the practical outcome is implementation-specific (cloud, on-prem and in-cluster implementations are all going to do something different)

(2) seems good from a UX perspective.

(3) seems like an explicable and consistent behavior, but potentially catastrophic UX

Maybe it's not necessary for the spec to define a single behavior as required. If so, then the spec should probably add some language to make it clear that implements can (and will) vary.

bowei:
I think this can only be recommendations otherwise we may be imposing requirements for gateway controllers to watch gateways that are not explicitly managed (i.e. linked via gatewayclass field):

- API guarantee: no specific spec for behavior, user should not rely on behavior after gatewayclass deletion.
- Recommendation for controller authors: document behavior, use finalizers mechanism to manage lifetimes, fail gracefully to avoid disrupting customer traffic (fail static)

akutz:
Hi @bowei,

> I think this can only be recommendations 

I think that is what @jpeach and I were suggesting anyway. I originally pinged James and @youngnick to solicit their advice based on their experience with `IngressClass`. James indicated this was a new-sh concern, and so at best I think we're just trying to define the best practice for the situation and document it so that readers:

* are aware of the issue
* are presented with some recommendations

hbagdi:
#385 takes a stab at this. PTAL.

hbagdi:
#385 was merged in.
Please re-open if this doesn't solve the problem.
/close

PR #385, opened 2020-10-09
add recommendation to add finalizer to GatewayClass

hbagdi:
There is nothing that prevents a user to delete a GatewayClass resource
when there are Gateway resources that refer to it.

This patch adds in recommendation to use finalizers to prevent deletion
of GatewayClass resources that are in use.

This is a secondary safety-net. The first one is based on our personas
and RBAC where for a large cluster, only a small subset of users should
have delete permissions on GatewayClass.

See #170

### DIFF
--- a/apis/v1alpha1/gatewayclass_types.go
+++ b/apis/v1alpha1/gatewayclass_types.go
@@ -114,6 +114,10 @@ const (
 	// validity of the Parameters set for a given controller. This
 	// will initially start off as "Unknown".
 	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+
+	// GatewayClassFinalizerGatewaysExist should be added as a finalizer to the
+	// GatewayClass whenever there are provisioned Gateways using a GatewayClass.
+	GatewayClassFinalizerGatewaysExist = "gateway-exists-finalizer.networking.x-k8s.io"
 )
 
 // GatewayClassStatus is the current status for the GatewayClass.
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -27,6 +27,11 @@ import (
 
 // Gateway represents an instantiation of a service-traffic handling
 // infrastructure by binding Listeners to a set of IP addresses.
+//
+// Implementations should add the `gateway-exists-finalizer.networking.x-k8s.io`
+// finalizer on the associated GatewayClass whenever Gateway(s) is running.
+// This ensures that a GatewayClass associated with a Gateway(s) is not
+// deleted while in use.
 type Gateway struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/config/crd/bases/networking.x-k8s.io_gateways.yaml
+++ b/config/crd/bases/networking.x-k8s.io_gateways.yaml
@@ -23,7 +23,7 @@ spec:
     name: v1alpha1
     schema:
       openAPIV3Schema:
-        description: Gateway represents an instantiation of a service-traffic handling infrastructure by binding Listeners to a set of IP addresses.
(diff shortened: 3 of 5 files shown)


hbagdi:
There is still another problem: A Gateway is considered provisioned and will continue to function normally if it is included in ProvisionedGateways, even if it violates the `AllowedGatewayNamespaces` restrictions of a GatewayClass. In case the GatewayClass itself is deleted, there is no selector to validate the Gateway and there is no ProvisionedGateways field to validate if a Gateway was previously a valid Gateway or not.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -114,6 +114,10 @@ const (
 	// validity of the Parameters set for a given controller. This
 	// will initially start off as "Unknown".
 	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+
+	// FinalizerGatewaysExist should be added as a finalizer to the GatewayClass
+	// whenever there are provisioned Gateways using a GatewayClass.
+	FinalizerGatewaysExist = "Gateways using GatewayClass exist"
```
robscott, line 120: Nit: To match the var name above should this be `GatewayClassFinalizerGatewaysExist`? I know that is pretty long though...

robscott: The example in the [k8s docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers) seems to suggest that a finalizer name should be more structured. Maybe something like `gateway-finalizer.networking.x-k8s.io`?

hbagdi: > GatewayClassFinalizerGatewaysExist

The reason I chose FinalizerName was to catagorize all the finalizers together (makes lookup easier). Granted that we only have a single finalizer in this package currently.

> The example in the k8s docs seems to suggest that a finalizer name should be more structured. Maybe something like gateway-finalizer.networking.x-k8s.io?

Updated.

robscott: Wouldn't finalizers be applied to a specific resource though? Seems like they could be used similarly to conditions which are also resource specific. If finalizers were globally applicable it might make more sense to specify them in a separate file.

hbagdi: I see. That is a good point. I've updated this to be `GatewayClassFinalizerGatewaysExist`. Not the best name but considering that this won't be used very often, it should be okay? I'm open to better names.

hbagdi:
> There is still another problem: A Gateway is considered provisioned and will continue to function normally if it is included in ProvisionedGateways, even if it violates the AllowedGatewayNamespaces restrictions of a GatewayClass. In case the GatewayClass itself is deleted, there is no selector to validate the Gateway and there is no ProvisionedGateways field to validate if a Gateway was previously a valid Gateway or not.

@robscott Any guidance on how to go about solving this problem? Should it be another Github issue? And do we need to solve it in v1alpha1?

danehans:
> In case the GatewayClass itself is deleted, there is no selector to validate the Gateway and there is no ProvisionedGateways field to validate if a Gateway was previously a valid Gateway or not.

Instead of `GatewayClass` containing the list of provisioned gateways, maybe it's better to mark a route as "admitted" after it's successfully bound to a gateway. An implementation will not mark a new route "admitted" if it violates `AllowedGatewayNamespaces`. An implementation will not remove "admitted" routes even if `AllowedGatewayNamespaces` is updated in a manner that violates the routes namespace. For example:
```
HTTPRoute
...
const (
	// RouteAdmitted means the route is able to service requests to hostnames.
	RouteAdmitted RouteConditionType = "Admitted"

	// TODO: add other route condition types
)

// RouteGatewayStatus describes the status of a route with respect to an associated Gateway.
type RouteGatewayStatus struct {
	// Type is the type of the condition. Currently only Admitted.
	Type RouteConditionType

	...
}
```

@robscott @hbagdi @jpeach thoughts?

hbagdi:
I think since we have decided to remove `AllowedGatewayNamespaces` for the time being until we find a better solution, we can punt on this problem. The ProvisionedGateway field is also going away with this change.

hbagdi:
@robscott @danehans With #402 coming in, I've updated this PR. There is now some guidance on the Gateway resource itself for adding the finalizer to GatewayClass. PTAL.

robscott:
Thanks!

/approve

Issue #335, opened 2020-09-17
Add FAQs to documentation


robscott:
Some starting ideas:

- What implementations support Service APIs?

- Are Service APIs going to replace the Ingress API?

- How do I get involved?

PR #386, opened 2020-10-12
Update the FAQ.

jpeach:
This updates #335

### DIFF
--- a/docs-src/faq.md
+++ b/docs-src/faq.md
@@ -1,6 +1,72 @@
 # Frequently Asked Questions (FAQ)
 
-*   Q: Will there be a default controller implementation (in this repo)?<br>
+*   **Q: How can I get involved with Service API?<br>**
+    A: The [community](/community) page keeps track of how to get
+    involved with the project.
+
+*   **Q: Will Service API replace the Ingress API?<br>**
+    A: No. The Ingress API is GA since Kubernetes 1.19. There are no
+    plans to deprecate this API and we expect most Ingress controllers
+    to support it indefinitely.
+
+*   **Q: What are the differences between Ingress and Service API?<br>**
+    A: Ingress primarily targets exposing HTTP applications with a
+    simple, declarative syntax. Service API exposes a more general API
+    for proxying that can be used for more protocols than just HTTP,
+    and models more infrastucture components to provide better
+    deployment and management options for cluster operators.
+
+*   **Q: Will there be a default controller implementation (in this repo)?<br>**
     A: There is no current plan to have an "official" or "default"
     implementation. You will see the controller code in this repo be
     used for testing the support libraries.
+
+*   **Q: How can I expose custom capabilities through Service API?<br>**
+    A: There is a lot of diversity in the networking and proxying
+    ecosystem, and many products will have features that are not directly
+    supported in the API.  However, there are a few mechanisms available
+    for extending the API with implementation-specific capabilities:
+
+    * Decorate Service API objects with implementation-specific objects. A
+      policy or configuration object could match the Service API object either
+      by name or by using an explicit object reference.
+
+        For example, given a `Gateway` object with name "inbound",
+        creating a `AccessPolicy` object that also has the name "inbound"
+        could cause an implementation to attach a specified access
+        control policy. This is an example of matching the object by name.
+
+        Service API uses this decorator pattern with the
+        [`BackendPolicy`](/spec/#networking.x-k8s.io/v1alpha1.BackendPolicy)
+        resource, which modifies how a `Gateway` should forward traffic
+        to a backend target (commonly a Kubernetes `Service`). This is
+        an example of using explicit object references.
+
+    * Use implementation-specific values for string fields. In many
+      places, the fields of Service API resources have the type
+      "string". This allows an implementation to support custom values
+      for those fields in addition to any values specified in the API.
+
+    * Use implementation-specific annotations. For some kinds of
+      configuration, implementations may choose to support custom
+      annotations on Servce API objects. This approach continues
+      a proud tradition of extending Ingress objects.
+
+    * Use API-defined extension points. Some Service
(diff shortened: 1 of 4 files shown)


docs-src/faq.md
```diff
@@ -1,6 +1,59 @@
 # Frequently Asked Questions (FAQ)
 
-*   Q: Will there be a default controller implementation (in this repo)?<br>
+*   **Q: Will Service API replace the Ingress API?<br>**
+    A: No. The Ingress API is GA since Kubernetes 1.19. There are no
+    plans to deprecate this API and we expect most ingress controllers
```
robscott, line 5:
```suggestion
    plans to deprecate this API and we expect most Ingress controllers
```

docs-src/faq.md
```diff
@@ -1,6 +1,59 @@
 # Frequently Asked Questions (FAQ)
 
-*   Q: Will there be a default controller implementation (in this repo)?<br>
+*   **Q: Will Service API replace the Ingress API?<br>**
```
robscott, line 3: I think this leads to some natural follow up questions:

* What is the difference between this API and the Ingress API?
* Why would I use this API instead of the Ingress API?
* Why would I use the Ingress API instead of this one?

docs-src/faq.md
```diff
@@ -1,6 +1,59 @@
...
+      annotations on Servce API objects. This approach continues
+      a proud tradition of extending Ingress objects.
+
+    * Use API-defined extension references. In a few cases, Service
+      API objects allow references to arbitrary other Kubernetes
+      objects. These are cases where a single object may need to
+      refer to some implementation-defined values multiple times
+      or in a way that is context-dependent.
+
+*  **Q: Where can I find Service API releases?<br>**
```
robscott, line 54: Another follow up question from here:

- What does an alpha release mean? Will there be breaking changes in the next release?

docs-src/faq.md
```diff
@@ -1,6 +1,59 @@
...
+      for those fields in addition to any values specified in the API.
+
+    * Use implementation-specific annotations. For some kinds of
+      configuration, implementations may choose to support custom
+      annotations on Servce API objects. This approach continues
+      a proud tradition of extending Ingress objects.
+
+    * Use API-defined extension references. In a few cases, Service
+      API objects allow references to arbitrary other Kubernetes
+      objects. These are cases where a single object may need to
```
robscott, line 50:
```suggestion
      objects. There are cases where a single object may need to
```

robscott (commented):
Thanks for your work on this!

jpeach:
@robscott I partially addressed the extra FAQ entries you suggested; we can add more in follow-up PRs.

docs-src/faq.md
```diff
@@ -1,6 +1,74 @@
...
+    for proxying that can be used for more protocols than just HTTP,
+    and models the proxying infrastucture to provide better deployment
+    and management options for cluster operators.
+
+*   **Q: Will there be a default controller implementation (in this repo)?<br>**
     A: There is no current plan to have an "official" or "default"
     implementation. You will see the controller code in this repo be
     used for testing the support libraries.
+
+*   **Q: How can I expose custom capabilities through Service API?<br>**
```
robscott, line 24: Now that https://github.com/kubernetes-sigs/service-apis/pull/388 has been merged, I think some/most of the content in the answer to this question can likely be added to the documentation Harry added.

SantoDE:
You probably talked about it 100x times already, but I still wanna ask: I know the official name of all of that is Service API, but aren't we primarily focusing on the "Gateway API" part of now? That being said, maybe it could make sense to "compare" the Gateway API with the Ingress API part for now? Or am I totally off-track?

docs-src/faq.md
```diff
@@ -1,6 +1,74 @@
...
+*   **Q: Will Service API replace the Ingress API?<br>**
+    A: No. The Ingress API is GA since Kubernetes 1.19. There are no
+    plans to deprecate this API and we expect most Ingress controllers
+    to support it indefinitely.
+
+*   **Q: What are the differences between Ingress and Service API?<br>**
+    A: Ingress primarily targets exposing HTTP applications with a
+    simple, declarative syntax. Service API exposes a more general API
+    for proxying that can be used for more protocols than just HTTP,
+    and models the proxying infrastucture to provide better deployment
```
danehans, line 16: I question whether we should use the word "proxy" as it may be interpreted by the reader as an implementation detail or requirement.

docs-src/faq.md
jpeach: Updated with a link, but IMHO the general technique suggestions don't belong in the API concepts doc.

jpeach:
> You probably talked about it 100x times already, but I still wanna ask: I know the official name of all of that is Service API, but aren't we primarily focusing on the "Gateway API" part of now? That being said, maybe it could make sense to "compare" the Gateway API with the Ingress API part for now? Or am I totally off-track?

AFAIK, the official name is still "Service API", though I do agree that it is becoming more of a misnomer 🤷

robscott:
Thanks! Looks like something went wrong in the verify step, but otherwise LGTM. 

/approve

jpeach:
> Thanks! Looks like something went wrong in the verify step, but otherwise LGTM. 

Yeh, looks like a golint problem 🤷‍♂️

SantoDE:
> > You probably talked about it 100x times already, but I still wanna ask: I know the official name of all of that is Service API, but aren't we primarily focusing on the "Gateway API" part of now? That being said, maybe it could make sense to "compare" the Gateway API with the Ingress API part for now? Or am I totally off-track?
> 
> AFAIK, the official name is still "Service API", though I do agree that it is becoming more of a misnomer 🤷

Alright. Thanks. Maybe we can keep it in our head and address that at a later point in time :)

robscott:
@jpeach I think you may need to rebase this on master.

PR #388, opened 2020-10-12
document extension concept

hbagdi:


docs-src/concepts.md
```diff
@@ -230,7 +230,24 @@ be differences in behavior from the spec.
...
 ### Extension points
 
-TODO
+A number of extension points are provided in the API to provide flexibility in
+addressing the large number of use-cases that cannot be addressed by a general
+purpose API.
+
+Here is a summary of extension points in the API:
+- **Custom Routes**: Implementers can chose to create custom Route resources
+  for protocols that are not currently supported in the API.
```
robscott, line 239: Since I think this intended to be our extension point of last resort, maybe we should list it last with some kind of caveat like "if none of the above extension points are sufficient, you can create a custom route resource..."

docs-src/concepts.md
```diff
@@ -230,7 +230,24 @@ be differences in behavior from the spec.
 
 ### Extension points
 
-TODO
+A number of extension points are provided in the API to provide flexibility in
```
robscott, line 233: This PR has a lot of overlap with the additions to the FAQ in #386 by @jpeach, specifically the "How can I expose custom capabilities through Service API" question. I think that question/answer may be a bit too large for FAQ, maybe we can combine the content here?

robscott (commented):
Thanks for your work on this!

docs-src/concepts.md
hbagdi: Totally makes sense, updated.

docs-src/concepts.md
robscott: @jpeach what do you think makes the most sense here? Should we merge this one first and then add your content from FAQs in a follow up?

jpeach:
/lgtm

let's land this and I'll rebase the FAQ entry to link to it

robscott:
Thanks!

/approve

PR #389, opened 2020-10-12
fix indents and sections headers in concepts

hbagdi:
Concepts docs has exploded with content that could use some
re-organization.

This is small patch in a series of fixes to make this document
understandable and easy to navigate.

### DIFF
--- a/docs-src/concepts.md
+++ b/docs-src/concepts.md
@@ -232,7 +232,9 @@ be differences in behavior from the spec.
 
 TODO
 
-## GatewayClass
+## API Resources
+
+### GatewayClass
 
 `GatewayClass` ([source code][gatewayclass-src]) is cluster-scoped resource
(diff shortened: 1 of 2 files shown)


jpeach:
s/indendts/indents/, in the commit. Suggest that the commit fixes the section headers, not just indents

docs-src/concepts.md
```diff
@@ -424,7 +426,7 @@ status entries corresponding to their name. Both `GatewayStatus` and
 This is a list that includes a type of condition, the status of that condition,
 and the last time this condition changed.
 
-#### Listeners
+##### Listeners
```
jpeach, line 429: Why does `Listeners` nest inside `Gateway Status`. I would expect it to be at the same level since it's a spec field.

hbagdi:
Updated. PTAL.

robscott:
Thanks!

/approve

Issue #331, opened 2020-09-17
Add documentation covering TLS configuration


robscott:
This documentation should include diagrams and examples. It should describe the purpose of TLS configuration on each resource and the server vs client roles here.

robscott:
@hbagdi are you able to take this one on?

danehans:
Closed by https://github.com/kubernetes-sigs/service-apis/pull/391

PR #391, opened 2020-10-12
add document elaborating on how to use TLS and examples

hbagdi:


hbagdi:
This fixes #331.

docs-src/tls.md
```diff
@@ -0,0 +1,328 @@
...
+## Client/Server and TLS
+
+![overview](tls-overview.svg)
+
+For Gateways, there are two connections involved:
+
+- **Downstream**: This is the connection between the client and the Gateway.
+- **Upstream**: This is the connection between the Gateway and Service. The
+  Service in question can be a core Kubernetes Service or any other network
+  endpoint as well.
```
robscott, line 16: This is really hard to describe well, I've never found a great way to say it, but maybe something like this?
```suggestion
- **Upstream**: This is the connection between the Gateway and backend resources
   specified by routes. These backend resources will usually be Services.
```

robscott (commented):
Thanks for taking this on! This is going to be a huge addition to our docs.

docs-src/tls.md
```diff
@@ -0,0 +1,328 @@
...
+    hostname:
+      name: foo.example.com
+      match: Exact
+    tls:
+      certificateRef:
+        kind: Secret
+        group: core
+        name: foo-example-com-cert
+      routeOverride:
+        certificate: Deny
```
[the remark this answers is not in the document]
hbagdi: I considered this but decided against it because of two reasons:
- makes it clear that TLS config is ONLY in Listener
- for anybody trying to learn the TLS section of this API, it helps with keeping all bits in front. I've already remoed route selection section. 

That being said, I can reason either ways and don't feel strongly about it.

docs-src/tls.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Examples
+
+#### TLS in listener
+
+In this example, Gateway is configured to serve `foo.example.com` and
+`bar.example.com` domains. The certificate for these domains are specified
+in Gateway.
+
+```yaml
```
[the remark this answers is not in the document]
hbagdi: That is a fair concern but I deleted the YAML files from the `examples` directory to ensure that I'm not increasing any duplication. I think it is better to include text with examples than having examples in a directory full of YAML files. I, as a user, mostly get lost in that.

Is there any upside of having these in YAML files and then referencing them? I guess one upside is that users can grab the URL and `kubectl apply` it.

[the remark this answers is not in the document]
robscott: I think the biggest upside is that we'd have a way to automatically verify that our examples weren't stale.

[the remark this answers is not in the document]
hbagdi: Good point. Let me try to see how to get this to work in a separate PR (and hold this until that). I've no experience with mkdocs.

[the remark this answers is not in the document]
hbagdi: https://github.com/kubernetes-sigs/service-apis/pull/393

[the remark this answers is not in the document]
hbagdi: I've updated the PR with YAMLs inside `examples` and using file embedding approach.

#394 fixes a bug and should go in before this.

docs-src/concepts.md
```diff
@@ -1,9 +1,6 @@
 # API Concepts
 
-This document is a deep dive into the reasoning and design for the API. The
-content of this document started from this [API sketch][api-sketch].
-
-[api-sketch]: https://docs.google.com/document/d/1BxYbDovMwnEqe8lj8JwHo8YxHAt3oC7ezhlFsG_tyag/preview
+This document is a deep dive into the reasoning and design for the API.
```
danehans, line 3: s/the API/Service APIs/

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+Downstream TLS settings are configured using Listeners at the Gateway level.
+
+### Listener and TLS 
+
+Listeners expose the TLS setting on a per domain or sub-domain basis.
+TLS settings of a listener are applied to all domains that satisfy the
+`hostname.match` criteria.
+
+In the following example, a Gateway is configured to serve a TLS certificate
+for any request that is served a Gateway.
```
Miciah, line 35:
```suggestion
for any request that is served by the Gateway.
```

docs-src/concepts.md
```diff
@@ -128,19 +125,10 @@ A typical client/gateway API request flow for a gateway implemented using a reve
...
-* In most cases, users that are configuring TLS will naturally also have access
-  to Gateways.
-* In other cases, TLS config could be implemented with a controller watching
-  Routes and adding generated certs to corresponding Gateways.
-* This does not solve the use case for users wanting to provide their own certs
-  for Routes while not having access to a Gateway resource. This seems like it
-  would be a rare edge case and is not worth supporting at this point. The
-  security model outlined a potential approach to enable this in the future, but
-  there does not seem to be a sufficient reason to work towards that now.
+TLS configuration is tied to Gateway listeners. Additionally, TLS certificates
```
danehans, line 128: COnsider the following minor change:

s/TLS configuration is tied to Gateway listeners./TLS is configured on Gateway listeners./

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+      name: default-cert
+    routeOverride:
+      certificate: Deny
+```
+
+If `hostname.match` is set to `Exact`, then the TLS settings apply to only the
+specific hostname that is set in `hostname.name`. This is the recommended way
+of configuring TLS settings in Service APIs as it does not involve any
+redirection and should suffice in the majority of use-cases.
+Please take a look at the examples below for various possibilities.
```
Miciah, line 57:
```suggestion
Please take a look at the examples below for various alternatives.
```

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Route and TLS
+
+If `listener.tls.routeOverride.certificate` is set to `Allow`, TLS certificates
```
Miciah, line 61:
```suggestion
If `listeners[].tls.routeOverride.certificate` is set to `Allow`, TLS certificates
```

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+can be configured at the Route-level. This feature is primarily meant for a
+cluster with a self-service model where Service owners bring their own TLS
+certificates. This feature also mirrors the behavior of TLS as defined in the
+Ingress v1 resource. One should use this feature only when the Gateway Admin
+wishes to delegate TLS to Service or App owners.
```
Miciah, line 66: I don't think "Gateway Admin" and "App owner" are defined anywhere.  Is defining "Gateway Admin" covered by https://github.com/kubernetes-sigs/service-apis/issues/95?  I guess we can come back and update this text if needed depending on what role names are defined in https://github.com/kubernetes-sigs/service-apis/issues/95.

docs-src/concepts.md
```diff
@@ -128,19 +125,10 @@ A typical client/gateway API request flow for a gateway implemented using a reve
...
+can be configured at the Route-level for certain self-service use cases.
+
+Please refer to the [TLS document](tls.md) for a deep dive on TLS.
```
danehans, line 131: Instead of TLS document, maybe the document should be called TLS Details, etc..

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+
+When using this feature, please note that the TLS certificate to serve is chosen
+before an HTTPRoute is selected. This is because the TLS handshake is completed
+before an HTTP request is sent from the client.
```
Miciah, line 70: This makes me wonder:  Does the same restriction apply if `spec.listeners[].routes.kind` is set to `TLSRoute`?  For HTTPRoutes, the restriction derives directly from the interaction between TLS and HTTP, but TLSRoutes aren't subject to the same limitation.  If this hasn't been decided already, I would suggest that we apply the same restriction to TLSRoutes (and perhaps document it here), even if it isn't strictly necessary, in order to provide a consistent API and clear semantics for implementers.

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+
+[TLS Certificate in Route](#tls-certificate-in-route) provides an example
+on how this feature can be used.
```
Miciah, line 73:
```suggestion
[TLS Certificate in Route](#tls-certificate-in-route) provides an example
of how this feature can be used.
```
Using "of" feels more correct to me, but this might be a regional difference.

docs-src/tls.md
danehans, line 35: Either drop ` that is served a Gateway` or reword.

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Examples
+
+#### TLS in listener
+
+In this example, Gateway is configured to serve `foo.example.com` and
+`bar.example.com` domains. The certificate for these domains are specified
+in Gateway.
```
Miciah, line 81:
```suggestion
In this example, the Gateway is configured to serve the `foo.example.com` and
`bar.example.com` domains. The certificate for these domains is specified
in the Gateway.
```

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+Listeners expose the TLS setting on a per domain or sub-domain basis.
+TLS settings of a listener are applied to all domains that satisfy the
+`hostname.match` criteria.
+
+In the following example, a Gateway is configured to serve a TLS certificate
+for any request that is served a Gateway.
+
+```yaml
+listeners:
+- protocol: HTTPS # Other possible value is `TLS`
```
danehans, line 39: This section is titled "Listener and TLS" but the protocol is `HTTPS` with "# Other possible value is `TLS`. Either rename this section to "Listener and HTTPS" or provide background of the `protocol` field before diving into this example.

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+{% include 'tls-basic.yaml' %}
+```
+
+#### Wildcard TLS listeners
+
+In this example, the Gateway is configured with wildcard certificate for
+`*.example.com` and a different certificate for `foo.example.com`.
+Since a specific match takes priority, the Gateway will serve the
+`foo-example-com-cert` for requests to `foo.example.com` and the
+`wildcard-example-com-cert` for all other requests.
```
Miciah, line 93:
```suggestion
In this example, the Gateway is configured with a wildcard certificate for
`*.example.com` and a different certificate for `foo.example.com`.
Since a specific match takes priority, the Gateway will serve
`foo-example-com-cert` for requests to `foo.example.com` and
`wildcard-example-com-cert` for all other requests.
```
or
```suggestion
In this example, the Gateway is configured with a wildcard certificate for
`*.example.com` and a different certificate for `foo.example.com`.
Since a specific match takes priority, the Gateway will serve
the certificate in `foo-example-com-cert` for requests to `foo.example.com` and
the certificate in `wildcard-example-com-cert` for all other requests.
```

docs-src/tls.md
danehans, line 57: s/take a look at/review/

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+{% include 'wildcard-tls-gateway.yaml' %}
+```
+
+#### TLS Certificate in Route
+
+In this example, Gateway is configured with a default certificate that will be
+served for all hostnames. In addition, `tls.routeOverride.certificate` is set to
+`Allow`, meaning Routes can specify TLS certificates for any domains. Next,
+there are two HTTPRoute resources which define certificates for
+`foo.example.com` and `bar.example.com`.
```
Miciah, line 105:
```suggestion
In this example, the Gateway is configured with a default certificate that will be
served for all hostnames. In addition, `tls.routeOverride.certificate` is set to
`Allow`, meaning routes can specify TLS certificates for any domains. Next,
there are two HTTPRoute resources which specify certificates for
`foo.example.com` and `bar.example.com`.
```
I think we should use lower-cased "routes" for the general concept, and only capitalize when referring to a specific kind (such as HTTPRoute).

examples/tls-cert-in-route.yaml
```diff
@@ -1,22 +1,16 @@
...
-    name: acme-lb
-    group: acme.io
-    kind: Parameters
----
 kind: Gateway
 apiVersion: networking.x-k8s.io/v1alpha1
 metadata:
-  name: my-gateway
-  namespace: default
+  name: cert-in-route--gateway
```
Miciah, line 4: Is the `--` significant, or should that be a single `-`?

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+If `hostname.match` is set to `Exact`, then the TLS settings apply to only the
+specific hostname that is set in `hostname.name`. This is the recommended way
+of configuring TLS settings in Service APIs as it does not involve any
+redirection and should suffice in the majority of use-cases.
+Please take a look at the examples below for various possibilities.
+
+### Route and TLS
+
+If `listener.tls.routeOverride.certificate` is set to `Allow`, TLS certificates
+can be configured at the Route-level. This feature is primarily meant for a
```
danehans, line 62: s/at the Route-level/on Routes that are bound to the Gateway/

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+- **downstream**: This is the connection between the client and the Gateway.
+- **upstream**: This is the connection between the Gateway and backend resources
+   specified by routes. These backend resources will usually be Services.
+  
+With Service APIs, TLS configuration of downstream and
+upstream connections is managed independently.
+
+Please note that in case of `Passthrough` TLS mode, no TLS settings take
+effect as the TLS session from the client is NOT terminated at the Gateway.
+The rest of the document assumes that TLS is being terminated at the Gateway.
```
Miciah, line 22: Perhaps state that "Terminate" is the default.  I noticed that some examples specify `mode: Terminate` and some omit it (but get `Terminate` because it is the default).

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+## Upstream TLS
+
+Upstream TLS configuration applies to the connection between the Gateway
+and Service.
+
+There is only one way to configure upstream TLS: using the `BackendPolicy`
+resource.
+
+Please note that the TLS configuration is related to the Service or backend
+resource and not related to a specific Route resource.
```
Miciah, line 120:
```suggestion
resource and not related to a specific route resource.
```

examples/upstream-tls.yaml
```diff
@@ -0,0 +1,51 @@
...
+spec:
+  hostnames:
+  - "foo.com"
+  rules:
+  - matches:
+    - path:
+        type: Prefix
+        value: /
+    forwardTo:
+    - serviceName: my-service1
```
Miciah, line 50:
```suggestion
    - serviceName: my-service
```
Typo, I think?

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+Route.
+
+```yaml
+{% include 'upstream-tls.yaml' %}
+```
+
+## Extensions
+
+Both upstream and downstream TLS configs provide an `options` map to add
+additional TLS settings for implementation specific features.
```
Miciah, line 136:
```suggestion
additional TLS settings for implementation-specific features.
```

docs-src/tls.md
danehans: Although someone can infer `routeOverride` will override the gateway listener cert with the route cert, it might be a good idea to explicitly call this out here.

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+      group: core
+      name: default-cert
+    routeOverride:
+      certificate: Deny
+```
+
+If `hostname.match` is set to `Exact`, then the TLS settings apply to only the
+specific hostname that is set in `hostname.name`. This is the recommended way
+of configuring TLS settings in Service APIs as it does not involve any
+redirection and should suffice in the majority of use-cases.
```
Miciah, line 56: What does "redirection" refer to here?

docs-src/tls.md
```diff
@@ -0,0 +1,138 @@
...
+
+If `listener.tls.routeOverride.certificate` is set to `Allow`, TLS certificates
+can be configured at the Route-level. This feature is primarily meant for a
+cluster with a self-service model where Service owners bring their own TLS
+certificates. This feature also mirrors the behavior of TLS as defined in the
+Ingress v1 resource. One should use this feature only when the Gateway Admin
+wishes to delegate TLS to Service or App owners.
+
+When using this feature, please note that the TLS certificate to serve is chosen
+before an HTTPRoute is selected. This is because the TLS handshake is completed
```
danehans, line 69: I'm a bit confused by "When using this feature, please note that the TLS certificate to serve is chosen
before an HTTPRoute is selected." If `listener.tls.routeOverride.certificate` is set to `Allow`, then the cert is configured at the route-level.

danehans: Reading further, I think I have a better understanding of what the above sentence is trying to convey. It might be helpful to introduce the `BackendPolicy` here for clarification.

examples/upstream-tls.yaml
```diff
@@ -0,0 +1,51 @@
...
+  backendRefs:
+  - name: my-service
+    group: core
+    kind: Service
+    port: 443
+  tls:
+    clientCertificateRef:
+      name: gateway-client-cert
+      group: core
+      kind: Secret
```
danehans, line 15: It sounds like `clientCertificateRef` is moving to `Gateway`, so either file an issue to track the need to update this doc or update this PR based on the upcoming changes.

danehans:
@hbagdi Great job with this doc. I have a few nits.

bowei:
LGTM (holding for others comments)

docs-src/tls.md
hbagdi: Redirection in the sense that TLS configuration is directly defined in the Gateway and doesn't require a Cluster Operator to check all routes for TLS definitions.

docs-src/tls.md
hbagdi: It does apply to TLSRoute. Is this the right place to document it though? Should it not be in API docs rather?

docs-src/concepts.md
```diff
@@ -128,19 +125,10 @@ A typical client/gateway API request flow for a gateway implemented using a reve
...
-  to Gateways.
-* In other cases, TLS config could be implemented with a controller watching
-  Routes and adding generated certs to corresponding Gateways.
-* This does not solve the use case for users wanting to provide their own certs
-  for Routes while not having access to a Gateway resource. This seems like it
-  would be a rare edge case and is not worth supporting at this point. The
-  security model outlined a potential approach to enable this in the future, but
-  there does not seem to be a sufficient reason to work towards that now.
+TLS is configured on Gateway listeners. Additionally, TLS certificates
+can be configured at the Route-level for certain self-service use cases.
```
jpeach, line 129:
```suggestion
can be configured on Route objects for certain self-service use cases.
```

docs-src/tls.md
```diff
@@ -0,0 +1,144 @@
...
+### Listener and TLS 
+
+Listeners expose the TLS setting on a per domain or sub-domain basis.
+TLS settings of a listener are applied to all domains that satisfy the
+`hostname.match` criteria.
+
+In the following example, the Gateway serves the TLS certificate
+defined in `default-cert` Secret resource for all requests.
+Although, the example refers to HTTPS protocol, one can also use the same
+feature for TLS-only protocol alongwith TLSRoutes.
```
jpeach, line 38:
```suggestion
feature for TLS-only protocol along with TLSRoutes.
```

docs-src/tls.md
Miciah: Ah, as in cross-referencing resources?  And does "This" in "This is the recommended way" refer to setting `tls.routeOverride.certificate: Deny`?  So if I understand correctly, this statement could be rephrased as, "Specifying `tls.routeOverride.certificate: Deny` is recommended because it centralizes TLS configuration within the Gateway specification and should suffice for the majority of use-cases."

hbagdi: Updated.

robscott:
@hbagdi It looks like you've addressed most feedback now, is there anything still left? This generally LGTM.

hbagdi:
> @hbagdi It looks like you've addressed most feedback now, is there anything still left? This generally LGTM.

That's my understanding as well. Other than [one comment](https://github.com/kubernetes-sigs/service-apis/pull/391#discussion_r504989542) from @Miciah , I'm seeking an approval so that this PR can get merged.

robscott:
Thanks! Will let @danehans, @jpeach, or @Miciah add the final LGTM here.

/approve

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+Downstream TLS settings are configured using Listeners at the Gateway level.
+
+### Listener and TLS 
+
+Listeners expose the TLS setting on a per domain or sub-domain basis.
+TLS settings of a listener are applied to all domains that satisfy the
+`hostname.match` criteria.
+
+In the following example, the Gateway serves the TLS certificate
+defined in `default-cert` Secret resource for all requests.
```
Miciah, line 36:
```suggestion
defined in the `default-cert` Secret resource for all requests.
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+
+Specifying `tls.routeOverride.certificate: Deny` is recommended because it
+centralizes TLS configuration within the Gateway specification and should
+suffice for the majority of use-cases. Please take a look at the examples below
+for various alternatives.
+
+### Route and TLS
+
+If `listeners[].tls.routeOverride.certificate` is set to `Allow`, TLS certificates
+can be configured on Routes that are bound to the Gateway. This feature is
```
Miciah, line 67:
```suggestion
can be configured on routes that are bound to the Gateway. This feature is
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+primarily meant for a cluster with a self-service model where Application developers
+bring their own TLS certificates. This feature also mirrors the behavior of
+TLS as defined in the Ingress v1 resource. One should use this feature only
+when the Cluster Operator wishes to delegate TLS to Service or Application Developer
+the certificate defined in the Route will override any certificate defined in
+the Gateway
```
Miciah, line 73: Looks like this sentence got corrupted?
```suggestion
TLS as defined in the Ingress v1 resource. One should use this feature only
when the Cluster Operator wishes to delegate TLS configuration to the Application Developer.
With this feature, the certificate defined in the route overrides any certificate defined in
the Gateway.
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+
+Please note that the TLS configuration is related to the Service or backend
+resource and not related to a specific route resource.
+
+### Example
+
+The following example shows how upstream TLS can be configured. We have
+omitted downstream TLS configuration for simplicity. As noted before, it
+doesn't matter how downstream TLS is configured for the specific Listener or
+Route.
```
Miciah, line 134:
```suggestion
doesn't matter how downstream TLS is configured for the specific listener or
route.
```

docs-src/concepts.md
```diff
@@ -128,19 +125,10 @@ A typical client/gateway API request flow for a gateway implemented using a reve
 
 ## TLS Configuration
 
-TLS configuration is tied to Gateway listeners. Although adding the option to
-configure TLS on other resources was considered, ultimately TLS configuration on
-Gateway listeners was deemed sufficient for the following reasons:
+TLS is configured on Gateway listeners. Additionally, TLS certificates
+can be configured on Route objects for certain self-service use cases.
```
Miciah, line 129:
```suggestion
can be configured on route objects for certain self-service use cases.
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+
+If `hostname.match` is set to `Exact`, then the TLS settings apply to only the
+specific hostname that is set in `hostname.name`.
+
+Specifying `tls.routeOverride.certificate: Deny` is recommended because it
+centralizes TLS configuration within the Gateway specification and should
+suffice for the majority of use-cases. Please take a look at the examples below
+for various alternatives.
+
+### Route and TLS
```
Miciah, line 64:
```suggestion
### Routes and TLS
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+upstream connections is managed independently.
+
+Please note that in case of `Passthrough` TLS mode, no TLS settings take
+effect as the TLS session from the client is NOT terminated at the Gateway.
+The rest of the document assumes that TLS is being terminated at the Gateway,
+which is the default setting.
+
+## Downstream TLS
+
+Downstream TLS settings are configured using Listeners at the Gateway level.
```
Miciah, line 27:
```suggestion
Downstream TLS settings are configured using listeners at the Gateway level.
```

docs-src/tls.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Listener and TLS
```
Miciah, line 29:
```suggestion
### Listeners and TLS 
```

Miciah:
/lgtm
Let's address https://github.com/kubernetes-sigs/service-apis/pull/391/files#r505915278 in a follow-up (I agree, the API godoc can cover this).

PR #394, opened 2020-10-13
default to Same namespace in route selection

hbagdi:


apis/v1alpha1/gateway_types.go
```diff
@@ -430,6 +430,8 @@ type RouteBindingSelector struct {
 	// default.
 	//
 	// Support: Core
+	//
+	// +optional
```
robscott, line 434: I think an optional field should be `omitempty`

apis/v1alpha1/gateway_types.go
```diff
@@ -500,6 +502,7 @@ type RouteNamespaces struct {
 	// * Same: Only Routes in the same namespace may be used by this Gateway.
 	//
 	// Support: Core
+	// +kubebuilder:default=Same
```
robscott, line 505: I'm not sure how valuable a default is at this level. I can see the value of a default defined on the `RouteNamespaces` field, but if someone is going to go out of their way to define `RouteNamespaces` they'll probably want to provide a specific value here. I think the only way this default would be used is if someone specified `routeNamespaces: {}`.

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/gateway_types.go
hbagdi: But it has a default, so it will technically never be empty.

danehans:
Kubebuilder slack thread [xref](https://kubernetes.slack.com/archives/CAR30FCJZ/p1602786471038600) on the recommended approach for specifying fields that are optional for a user but required for persisting the object.

hbagdi:
@robscott @danehans 

So, I tried adding defaults at the parent level by adding instead:
```diff
+       // +kubebuilder:default={from: "Same"}
        RouteNamespaces RouteNamespaces `json:"routeNamespaces"`
```

[CRD](https://gist.githubusercontent.com/hbagdi/c16a6607a91805391779784126b32cf7/raw/ef3cada99f884e2d049a56444d8725dca7aea075/gateway-crd.yaml) is generated.
And with the following gateway:
```yaml
kind: Gateway
metadata:
  name: my-gateway-2
  namespace: default
spec:
  gatewayClassName: acme-lb
  listeners:  # Use GatewayClass defaults for listener definition.
  - protocol: HTTP
    port: 80
    routes:
      kind: HTTPRoute
```

gives this error:

```
error: error validating "examples/basic-http.yaml": error validating data: ValidationError(Gateway.spec.listeners[0].routes): missing required field "routeNamespaces" in io.x-k8s.networking.v1alpha1.Gateway.spec.listeners.routes; if you choose to ignore these errors, turn validation off with --validate=false
```

What am I missing here?

jpeach:
@hbagdi AFAICT, `+optional` is needed when setting defaults. For example:

```
diff --git apis/v1alpha1/gateway_types.go apis/v1alpha1/gateway_types.go
index 1af2085..26e7147 100644
--- apis/v1alpha1/gateway_types.go
+++ apis/v1alpha1/gateway_types.go
@@ -430,6 +430,9 @@ type RouteBindingSelector struct {
        // default.
        //
        // Support: Core
+       //
+       // +optional
+       // +kubebuilder:default={from: "Same"}
        Namespaces RouteNamespaces `json:"namespaces"`
        // Selector specifies a set of route labels used for selecting
        // routes to associate with the Gateway. If RouteSelector is defined,
```

Then create a route that would use the default:
```
diff --git examples/basic-http.yaml examples/basic-http.yaml
index cd47a69..8e05276 100644
--- examples/basic-http.yaml
+++ examples/basic-http.yaml
@@ -24,8 +24,6 @@ spec:
       selector:
         matchLabels:
           app: foo
-      namespaces:
-        from: "All"
```

and it gets set
```
$ k get gateway my-gateway -o yaml
kind: Gateway
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"networking.x-k8s.io/v1alpha1","kind":"Gateway","metadata":{"annotations":{},"name":"my-gateway","namespace":"default"},"spec":{"gatewayClassName":"acme-lb","listeners":[{"port":80,"protocol":"HTTP","routes":{"kind":"HTTPRoute","selector":{"matchLabels":{"app":"foo"}}}}]}}
  creationTimestamp: "2020-10-20T02:47:07Z"
  generation: 1
    manager: kubectl-client-side-apply
    operation: Update
    time: "2020-10-20T02:47:07Z"
  name: my-gateway
  namespace: default
  resourceVersion: "2757"
  selfLink: /apis/networking.x-k8s.io/v1alpha1/namespaces/default/gateways/my-gateway
  uid: 16133815-9387-4a4e-9886-589635b13992
spec:
  gatewayClassName: acme-lb
  listeners:
  - hostname:
      match: Any
    port: 80
    protocol: HTTP
    routes:
      group: networking.x-k8s.io
      kind: HTTPRoute
      namespaces:
        from: Same
      selector:
        matchLabels:
          app: foo
```

The required field check takes precedence over the defaulting.

robscott:
@hbagdi @jpeach This is not intuitive, but the standard we've ended up settling on is `omitempty`. I think this ends up behaving similarly to the `+optional` tag but it's a bit less confusing. Here's an example of how we're already using defaults without `+optional`.

https://github.com/kubernetes-sigs/service-apis/blob/master/apis/v1alpha1/gateway_types.go#L140-L141

hbagdi:
@robscott @jpeach 
I've updated to use `omitempty` and changed the RouteNamespaces to a pointer.
This works as expected as well. Thanks a lot for the guidance here.

PTAL and let me know if we need further changes here.

robscott:
Thanks!

/lgtm
/approve

Issue #363, opened 2020-09-30
Simplify Gateway Route Selection


robscott:
This is a follow up to https://github.com/kubernetes-sigs/service-apis/pull/362.

A listener currently looks like this:
```
  - protocol: HTTP
    port: 80
    routes:
      resource: httproutes
      routeNamespaces:
        from: "Selector"
        selector: {}
      routeSelector:
        matchLabels:
          app: foo
```

I'm interested in the possibility of simplifying that to:

```
  - protocol: HTTP
    port: 80
    routes:
      resource: httproutes
      namespaces:
        from: "Selector"
        selector: {}
      selector:
        matchLabels:
          app: foo
```

The two selectors being so close to each other feels potentially confusing. On the other hand, the standard case becomes much simpler:

```
  - protocol: HTTP
    port: 80
    routes:
      resource: httproutes
      selector:
        matchLabels:
          app: foo
```

hbagdi:
This is now fixed.
/close

PR #397, opened 2020-10-13
gateway: simplify route selection naming

hbagdi:
See #363


apis/v1alpha1/gateway_types.go
```diff
@@ -425,21 +425,21 @@ const (
...
+	// Selector specifies a set of route labels used for selecting
 	// routes to associate with the Gateway. If RouteSelector is defined,
 	// only routes matching the RouteSelector are associated with the Gateway.
 	// An empty RouteSelector matches all routes.
 	//
 	// Support: Core
 	//
 	// +optional
-	RouteSelector metav1.LabelSelector `json:"routeSelector,omitempty"`
+	Selector metav1.LabelSelector `json:"selector,omitempty"`
```
robscott, line 442: One potential negative here is that indenting `selector` one level could have a significant impact (would affect namespaces and not routes). I think this update is still a net improvement though, just wanted to mention it in case there were any other ideas.

robscott (commented):
Thanks for your work on this!

/approve

apis/v1alpha1/gateway_types.go
hbagdi: Yes, I realized that too. we could switch the ordering to help a little with that so instead of:

```yaml
      namespaces:
        from: "All"
      selector:
        matchLabels:
          "app": "foo"
```

do:

```yaml
      selector:
        matchLabels:
          "app": "foo"
      namespaces:
        from: "All"
```

That doesn't solve the problem but protects us a tiny bit. Most of this YAMLs will proliferate as is.

robscott: 👍 that's a great idea

hbagdi: Updated all the yamls with this change as well.

robscott: Thanks!

Issue #390, opened 2020-10-12
Future proof ProtocolType


hbagdi:
**What would you like to be added**:

Future proof ProtocolType to minimize collisions of type definitions in future.
Priori art is available with [AppProtocol](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/discovery/types.go#L138-L145).

**Why is this needed**:

Protocols in Gateway.Spec.Listeners are open-ended strings in the API. A user can define a ProtocolType in a custom implementation, service-apis then introduces the same type as a supported core/extended type and there can be collisions.

PR #398, opened 2020-10-13
future proof open-ended ProtocolType

hbagdi:
See #390


hbagdi:
This has largely been borrowed from [AppProtocol](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/discovery/types.go#L138-L145).

howardjohn (approved):

apis/v1alpha1/gateway_types.go
```diff
@@ -298,6 +298,12 @@ const (
 // * "TCP"
 // * "UDP"
 //
+// Implementations can define their own protocols if one is not already defined.
+// Un-prefixed names are reserved for IANA standard service names (as per
+// RFC-6335 and http://www.iana.org/assignments/service-names).
+// Non-standard protocols should use prefixed names such as
+// mycompany.com/my-custom-protocol.
```
robscott, line 305: Do we need to clarify the support level here? Assuming this is custom?

hbagdi: Updated.

jpeach: I think that this needs a broader discussion and concept. The protocols we currently specify are spread across multiple OSI layers (e.g. we have both "UDP" and "HTTP"), and can be protocol families (e.g. "HTTPS" probably means HTTP/1.1 and HTTP/2, possibly HTTP/3).

The IANA service-names registry is just UDP/TCP port registrations. There's no "h2" or "GRPC" for example.

robscott: @jpeach Those are good points, do you know of any better lists we could use here?

danehans: Consider splitting `ProtocolType` into `AppProtocolType` that contains application layer protocols (i.e. HTTP) and `TransportProtocolType` that contains transport layer protocols (i.e. TCP).

robscott:
@thockin Do you have any ideas on how we could expand the list of protocols we support on Gateway listeners?

apis/v1alpha1/gateway_types.go
robscott: At least short term, what if we remove the RFC reference entirely and simplify the options to:
- Protocols we've listed and officially support
- Custom protocols can be specified with domain prefixes

We can always expand our officially supported list in the future but this feels like a reasonable compromise for now.

hbagdi: @robscott I like that approach. I've updated the PR to reflect that.

apis/v1alpha1/gateway_types.go
```diff
@@ -295,15 +295,11 @@ const (
...
-// * "TLS"
-// * "TCP"
-// * "UDP"
-//
-// +kubebuilder:validation:Enum=HTTP;HTTPS;TLS;TCP;UDP
+// Implementations can define their own protocols if one is not already defined.
+// Un-prefixed names are reserved for protocols defined in this package.
+// Non-standard protocols should use prefixed names such as
+// mycompany.com/my-custom-protocol.
+// Any protocol defined by implementations will fall under custom conformance.
```
robscott, line 302: I think we probably still want to retain a list of core protocol types that we support. Maybe just replace `Valid ProtocolType values...` with `Core ProtocolType values...`. Then you can add the paragraph about custom protocols below that.

hbagdi: Reworded this a little bit. PTAL.

robscott:
Thanks! Interested in any final feedback from @danehans or @jpeach here.

/approve

danehans:
/lgtm

I created https://github.com/kubernetes-sigs/service-apis/issues/433 to capture the need to define "Extended" protocol types.

PR #399, opened 2020-10-14
docs: clarify that gateway merging is not allowed

hbagdi:
Ingress v1 resource merging is very popular. Although, HTTPRoute is
meant as an equivalent of Ingress resource, this patch explicitly calls
out that two Gateway resources cannot be merged..


apis/v1alpha1/gateway_types.go
```diff
@@ -77,6 +77,10 @@ type GatewaySpec struct {
 	// As a special case, each group of compatible listeners
 	// may contain exactly one Listener with a match type of "Any".
 	//
+	// Listeners across Gateway resources cannot be collapsed i.e. Gateway
```
howardjohn, line 80: Doesn't L84 directly contradict with this? Or am I reading this wrong

hbagdi: No. L84 refers to listeners within the same Gateway resource.

hbagdi:
Discussion from weekly meeting:
We seem to have consensus on the change itself but we are not sure if it is worded correctly. 
I, personally, don't like the wording and arrangement so alternates are most welcome!

apis/v1alpha1/gateway_types.go
Miciah: On L67, L84, and L93, should "GatewayClass" be changed to "Gateway" or perhaps "implementation"?  You are saying the listeners within the _Gateway_ can be collapsed, so it seems odd to say that the _GatewayClass_ collapses the listeners.

apis/v1alpha1/gateway_types.go
```diff
@@ -77,6 +77,10 @@ type GatewaySpec struct {
...
+	// merging is not allowed. Each Gateway resource is a distinct network
+	// endpoint.
```
Miciah, line 82: Since you are soliciting suggestions on the wording:
```suggestion
	// Listeners in different Gateways cannot be collapsed.  Each Gateway
	// resource represents a distinct logical network endpoint.
```
And where we talk about collapsing listeners above, we could state more explicitly that they can be collapsed only _within a single gateway_:
```go
	// In general, each listener in this array must have a unique port.
	// However, an implementation may collapse listeners within the same
	// Gateway definition that have the same port into a single listener,
	// as long as the listeners are "compatible", as defined below.
```

jpeach:
Maybe I missed the discussion, but IIRC the intention was always that a controller should be allowed to merge gateways down to a single infrastructure deployment. Can you explain the rationale behind this change?

mattmoor:
+1 to elaborating on this.  It seems odd that the specification is prescriptive about implementation details 🤔

Miciah:
> Can you explain the rationale behind this change?

I think part of it was because I kept asking, "How does feature X work when you have gateway merging?"  * grin *.

On a more serious note, the semantics of the API are confusing or underspecified when gateway merging is allowed.  For example, what if you define one gateway that selects routes of kind TCPRoute and another gateway that selects routes of kind HTTPRoute?  How gateway merging affect RBAC and traffic splitting (when we add those)?  The semantics around merged gateways or rules for when gateways could be merged were becoming too complicated and proved to be too easy to forget to take into account.  

Ultimately, recent changes to allow routes to specify TLS configuration pretty much solved the main use-case for gateway merging, namely virtual host configuration (a route can specify a host as well as a TLS certificate for that host, if the gateway is configured to allow it).  Is there a reason gateway merging is still needed?

howardjohn:
How is this expected to work with large companies? We have users with 10k domains, which cannot be in a single Gateway. They also cannot just make a wildcard listener and search for Routes across the cluster otherwise there is no security - they need to be able to create a listener for foo.domain.com and look for routes in namespace `foo`.

jpeach:
> > Can you explain the rationale behind this change?
> 
> I think part of it was because I kept asking, "How does feature X work when you have gateway merging?" * grin *.
> 
> On a more serious note, the semantics of the API are confusing or underspecified when gateway merging is allowed. For example, what if you define one gateway that selects routes of kind TCPRoute and another gateway that selects routes of kind HTTPRoute?

Do you mean when disjoint kinds are selected for the same listener port on different gateways?

>  How does gateway merging affect RBAC and traffic splitting (when we add those)?

I'm not sure that it does at all. Gateways are merged onto some underlying infrastructure the is opaque to RBAC.

>  The semantics around merged gateways or rules for when gateways could be merged were becoming too complicated and proved to be too easy to forget to take into account.

It might be complex, but if an implementation wants to tackle it, I think the feedback would be interesting. I'm also interested in user feedback on whether such implementations meet their needs in useful ways.

youngnick:
I also thought when we originally discussed this that the way that a collapsing controller would work is kind of like they merge all the Gateway specs into one big one, and then apply the same validation rules you would apply for a single, large Gateway spec to the resultant config. If there are issues, they show up in the status of all related Gateways.

I can see how that dilutes the model of "one Gateway == one deployment", and it's very likely that in the storm of change I've missed in the last little while, that the reasons for this have gone away. However, even in the case @Miciah mentioned - with HTTPRoutes able to take domain names and certs, there still exist some cases where people want *not* to allow arbitrary certs to be added to infrastructure.

I'll make sure I'm more up to date on the spec and come to the next office hours to discuss.

robscott:
I meant to follow up here earlier, we had a great discussion about this at the latest office hours. I think the comments here and the discussion there led us to the consensus that we want to allow for Gateway merging but not be too prescriptive about how  or if an implementation supports it. With that said, I think we should rework this documentation to reflect that. Something like:

"Implementations may support merging Gateways. This functionality is considered custom and is not portable across implementations of the API."

^ That's far from perfect, but maybe a starting point.

hbagdi:
Although I'm the one who proposed this addition, I'd like to change my stance here.
I did some more thinking after the WG meetings last week and read through our documentation once again and realized that we are not at all prescriptive about how implementation translate the resources into the underlying infrastructure. I feel it is best to leave any mention of merging out at this point and re-visit this in future after we have some experience reports. 

> we want to allow for Gateway merging but not be too prescriptive about how or if an implementation supports it

That fuzziness in the definition of merging has the potential to introduce more confusion. When we have the option of  allowing all types of implementation and interpretations, I prefer that over explicitly referring to a poorly defined interpretation. 

I'd like to close this PR and conclude the topic if the above is agreeable to all.

Miciah:
We should be explicit about whether or not merging is permitted.  If it is permitted, let's add a statement to that effect.  

In addition (and I think this is consistent Rob's statements in https://github.com/kubernetes-sigs/service-apis/pull/399#issuecomment-716976850), if the consensus is that the API should not be prescriptive as to _whether_ gateway/listener merging is done at all, it would make sense not to be too prescriptive about _how_ merging is to be done (that is, the conditions under which gateways/listeners are merged should be implementation-defined).  

The relevant text on the master branch currently reads as follows: 
```go
	// Each Listener in this array must have a unique Port field,
	// however a GatewayClass may collapse compatible Listener
	// definitions into a single implementation-defined acceptor
	// configuration even if their Port fields would otherwise conflict.
	//
	// Listeners are compatible if all of the following conditions are true:
	//
	// 1. all their Protocol fields are "HTTP", or all their Protocol fields are "HTTPS" or TLS"
	// 2. their Hostname fields are specified with a match type other than "Any"
	// 3. their Hostname fields are not an exact match for any other Listener
	//
	// As a special case, each group of compatible listeners
	// may contain exactly one Listener with a match type of "Any".
```
I suggest that we add some "may" statements about merging and reframe the conditions as suggestions.  Here is a suggested rewording: 
```go
	// In general, each Listener in this array must specify a unique Port
	// value.  However, an implementation may group Listeners by Port and
	// then collapse each group of Listeners into a single Listener if the
	// implementation determines that the Listeners in the group are
	// "compatible". An implementation may also group together and collapse
	// compatible Listeners belonging to different Gateways.
	//
	// For example, an implementation may consider Listeners to be
	// compatible with each other if all of the following conditions are
	// met:
	//
	// 1. Either each Listener within the group specifies the "HTTP"
	//    Protocol or each Listener in the group specifies either the
	//    "HTTPS" or "TLS" Protocol.
	//
	// 2. Each Listener within the group specifies a Hostname that is unique
	//    within the group.
	//
	// 3. As a special case, one Listener within a group may omit Hostname,
	//    in which case this Listener matches when no other Listener
	//    matches.
```
I think that provides reasonable guidance for implementers.  Users will need to refer to implementation documentation to determine what exact behavior to expect.

youngnick:
+1 to @Miciah's solution here.

Miciah:
https://github.com/kubernetes-sigs/service-apis/pull/437 updates the wording around collapsing listeners based on the wording that I proposed in https://github.com/kubernetes-sigs/service-apis/pull/399#issuecomment-718268785.

hbagdi:
Closing this and we can continue this discussion on #437.
Please re-open if needed.
/close

PR #402, opened 2020-10-15
Removing allowedGatewayNamespaces from GatewayClass

robscott:
Although this initially seemed promising, we could never get it quite right. In light of that, removing this for v1alpha1 and going back to the drawing board for a better solution in the next release.

More context for this in https://github.com/kubernetes-sigs/service-apis/issues/375.


danehans:
@robscott please create an issue to track the need for adding an opa/gatekeeper example and link to `docs-src/security-model.md`.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -126,12 +107,6 @@ type GatewayClassStatus struct {
 	// +kubebuilder:validation:MaxItems=8
 	// +kubebuilder:default={{type: "InvalidParameters", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
hbagdi, line 109: Can't really comment on the right diff but I think we now need to remove the Condition `GatewayReasonNamespaceForbidden` in gateway_types.go.

jpeach: Perhaps we should generalize that to `GatewayReasonForbidden`?

hbagdi: I'm fine with that too but I would prefer to leave the condition out for now since the workflow of forbidden gateway is not clear yet.

jpeach: IMHO, either option (renaming or removing) is OK for now, but we should do one or the other :)

robscott: Updated to just remove the condition out, we can always add one back in later.

hbagdi:
/lgtm

/hold
because of one outstanding comment: https://github.com/kubernetes-sigs/service-apis/pull/402#issuecomment-709493139

robscott:
@hbagdi Thanks for catching that! I created https://github.com/kubernetes-sigs/service-apis/issues/404 to track that.

Issue #396, opened 2020-10-13
Upstream protocol cannot be defined


hbagdi:
**What happened**:

I was trying to define a set of manifests to use HTTPS between the Gateway and backend:
```yaml
kind: BackendPolicy
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: my-app
spec:
  backendRefs:
  - name: my-service
    group: core
    kind: Service
    port: 8443
  tls:
    clientCertificateRef: {}
    certificateAuthorityRef:
      name: my-cluster-ca
      group: core
      kind: Secret
    options: {}
---
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  ports:
  - name: http
    port: 8000
    targetPort: 8000
  - name: https
    port: 8443
    targetPort: 8443
  selector:
    app: my-service
---
kind: HTTPRoute
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: my-service-route
spec:
  hostnames:
  - "foo.com"
  rules:
  - matches:
    - path:
        type: Prefix
        value: /
    forwardTo:
    - serviceName: my-service1
      port: 8443
```

There is no place to define the protocol the Gateway should use to talk to the upstream. The above info is also not sufficient to determine it. This also applies to protocols like TCP/TLS.

**What you expected to happen**:

There should be a way to define upstream protocol to use.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

danehans:
As discussed during the community meeting, `AppProtocol` will be the long-term answer to this issue. I'm in favor of providing a temporary `Service` annotation until then. My 2nd preference is adding a `protocol` field to `BackendPolicy`.

jpeach:
> As discussed during the community meeting, `AppProtocol` will be the long-term answer to this issue. I'm in favor of providing a temporary `Service` annotation until then. My 2nd preference is adding a `protocol` field to `BackendPolicy`.

In Contour-land, the common cases are

| name | description |
| - | - |
| tls | HTTP/1.1 over TLS |
| h2 | HTTP/2 over TLS (also gRPC) |
| h2c | HTTP/2 over TCP (also gRPC) |

[...]

PR #407, opened 2020-10-22
define app-protocol annotation for k8s < 1.18

hbagdi:
See #396


hbagdi:
This patch takes a stab at #396 where it seems that having an annotation could be a reasonable stop-gap solution.
This is meant as a conversation starter and I'm open to discussing and changing details involved here.
I do understand the ugliness of this solution, alternate ideas are very welcome!

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,23 @@ const (
...
+	// discouraged when AppProtocol field is available. AppProtocol field, when
+	// populated, takes precedence over this annotation.
+	//
+	// The annotation value must be a comma separated with each value containing
+	// a port number followed by a forward slash and the protocol.
+	// Examples:
+	//
+	// - `networking.x-k8s.io/app-protocol: 8000/http,8443/https`
+	// - `networking.x-k8s.io/app-protocol: 4000/tcp,5000/tls`
+	AnnotationAppProtocol = "networking.x-k8s.io/app-protocol"
```
hbagdi, line 52: One limitation here is that if one defines the same port for TCP, UDP, SCTP, one cannot define different protocols for them.
In my experience, this is not common enough but I could be wrong here.

apis/v1alpha1/httproute_types.go
```diff
@@ -511,6 +511,14 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In absence of `AppProtocol` field
+	// (Kubernetes versions < 1.18), `networking.x-k8s.io/app-protocol` annotation
+	// may be used to define the protocol. The annotation serves as a stop-gap
+	// solution for the `AppProtocol` field and it's use is highly discouraged
+	// when AppProtocol field is available. AppProtocol field, when populated,
+	// takes precedence over the annotation.
```
hbagdi, line 520: Another step forward would be asking controllers to ignore this annotation for Kubernetes version > `x` for some value of `x`.

apis/v1alpha1/route_types.go
robscott: To clarify, are you referring to a Service that is exposing all 3 protocols on the same port number?

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,23 @@ const (
...
+	// solution for the Kubernetes versions < 1.18, where the `AppProtocol` field
+	// on the `Service` resource is not available. Using this annotation is highly
+	// discouraged when AppProtocol field is available. AppProtocol field, when
+	// populated, takes precedence over this annotation.
+	//
+	// The annotation value must be a comma separated with each value containing
+	// a port number followed by a forward slash and the protocol.
+	// Examples:
+	//
+	// - `networking.x-k8s.io/app-protocol: 8000/http,8443/https`
```
robscott, line 50: We've defined our protocols as all uppercase. I know that's not ideal but it does match what's done in upstream. The suggestions for app protocol are noticeably lowercase here, contrasting with what we have elsewhere for HTTP. I'm not necessarily suggesting a change here, just calling out a difference.

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,23 @@ const (
...
+	// - `networking.x-k8s.io/app-protocol: 4000/tcp,5000/tls`
```
robscott, line 51: Is `/` used elsewhere for this kind of differentiation? If we used BackendPolicy here we could simply have a single annotation mapping to a protocol. Although BackendPolicy is not useful for much else right now, it will stay that way unless we start adding config there when it makes sense.

apis/v1alpha1/route_types.go
```diff
@@ -80,6 +97,14 @@ type RouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In absence of `AppProtocol` field
+	// (Kubernetes versions < 1.18), `networking.x-k8s.io/app-protocol` annotation
+	// may be used to define the protocol. The annotation serves as a stop-gap
+	// solution for the `AppProtocol` field and it's use is highly discouraged
+	// when AppProtocol field is available. AppProtocol field, when populated,
```
robscott, line 105: Nit:
```suggestion
	// when AppProtocol field is available. The AppProtocol field, when populated,
```

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,23 @@ const (
 	GatewayAllowSameNamespace GatewayAllowType = "SameNamespace"
 )
 
+const (
+	// AnnotationAppProtocol defines the protocol a Gateway should use for
+	// communication with a Kubernetes Service. The annotation serves as a stop-gap
+	// solution for the Kubernetes versions < 1.18, where the `AppProtocol` field
+	// on the `Service` resource is not available. Using this annotation is highly
+	// discouraged when AppProtocol field is available. AppProtocol field, when
```
robscott, line 43: Nit: 
```suggestion
	// discouraged when AppProtocol field is available. The AppProtocol field, when
```

apis/v1alpha1/httproute_types.go
```diff
@@ -511,6 +511,14 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In absence of `AppProtocol` field
+	// (Kubernetes versions < 1.18), `networking.x-k8s.io/app-protocol` annotation
+	// may be used to define the protocol. The annotation serves as a stop-gap
+	// solution for the `AppProtocol` field and it's use is highly discouraged
+	// when AppProtocol field is available. AppProtocol field, when populated,
```
robscott, line 519: Nit:
```suggestion
	// when AppProtocol field is available. The AppProtocol field, when populated,
```

apis/v1alpha1/httproute_types.go
robscott: If we move this to BackendPolicy we may want this to stick around for other potential backend resources.

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/route_types.go
jpeach: The `AppProtocol` field uses `/` ...

> Un-prefixed names are reserved for IANA standard service names (as per RFC-6335 and http://www.iana.org/assignments/service-names). Non-standard protocols should use prefixed names such as mycompany.com/my-custom-protocol.

apis/v1alpha1/route_types.go
robscott: As @jpeach pointed out below, this lowercase approach matches the format of the AppProtocol field introduced in Kubernetes.

robscott: For reference, we're using the IsQualifiedName function to validate AppProtocol in upstream (we are not validating against the full service names list but we are requiring that values be lowercase).

apis/v1alpha1/route_types.go
hbagdi: Correct. This is a problem if there is more than one protocol exposed on the same port number.

apis/v1alpha1/route_types.go
hbagdi: Although AppProtocol uses `/` to differentiate, this is only tangentially related.
I initially went with `:` but then changed it to `/` to avoid parsing and reading issues in YAML.
As noted earlier, I don't have a preference here as long as we can encode the necessary data into a string.

> BackendPolicy is not useful for much else right now, it will stay that way unless we start adding config there when it makes sense.

BackendPolicy makes sense, especially for cases where the upstream resource is NOT a k8s Service.
We didn't have enough consensus (I might have read the feelings wrong) so I decided to write up more of a stop-gap solution.

apis/v1alpha1/route_types.go
hbagdi: Would it help if we document that the protocol is case-insensitive in here and examples are updated to be uppercase?

hbagdi:
I pushed to origin instead of my fork by mistake. Branch protection rules are causing all sorts of trouble so I've opened up #421 with the same diff.
Apologies for the trouble here.
/close

apis/v1alpha1/route_types.go
robscott: I think an annotation on BackendPolicy may be the right balance for now. It allows us to easily transition that annotation into a field if this ends up being widely used.

PR #408, opened 2020-10-22
Updates Filter Naming and Docs

danehans:
Updates `HTTPRoute` filter godocs, naming and field attributes.

__Note:__ The PR consists of 2 commits as I could not settle on a naming scheme for the core/extended filters. In addition to reviewing the PR, please provide your choice of naming.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/307

/assign @hbagdi @robscott 
/cc @jpeach @bowei


danehans:
@robscott if/when this PR merges, I was considering creating a docs section that covers filters. Let me know what you think.

apis/v1alpha1/httproute_types.go
```diff
@@ -328,79 +328,94 @@ type HTTPRouteMatch struct {
...
+	// RequestMirror defines a schema for a filter that mirrors requests.
+	//
+	// Support: Extended
+	//
 	// +optional
-	RequestHeader *HTTPRequestHeaderFilter `json:"requestHeader,omitempty"`
+	RequestMirror *HTTPRequestMirrorFilter `json:"requestMirror,omitempty"`
 
+	// ExtensionRef is an optional, implementation-specific extension to the
+	// "filter" behavior.  The resource may be "configmap" or an
```
hbagdi, line 382: Recently, we have intentionally removed references to ConfigMaps in extension points.
Our recent stance has been that there is nothing stopping users from using ConfigMap but we don't want to encourage this pattern.

hbagdi: This was already present in the code before this PR so this could be considered as a small scope creep for this PR.

examples/http-filter-mod-req-header.yaml
```diff
@@ -0,0 +1,52 @@
+kind: GatewayClass
```
hbagdi, line 1: What is the purpose of using `mod` in the filename?

apis/v1alpha1/httproute_types.go
```diff
@@ -328,79 +328,94 @@ type HTTPRouteMatch struct {
...
+// HTTPRouteFilterType identifies a type of HTTPRoute filter.
+// +kubebuilder:validation:Enum=RequestHeaderModifier;RequestMirror;Custom
+type HTTPRouteFilterType string
+
+const (
+	// RequestHeaderModifierHTTPRouteFilter can be used to add or remove an HTTP
+	// header from an HTTP request before it is sent to the upstream target.
+	//
+	// Support: Core
+	RequestHeaderModifierHTTPRouteFilter HTTPRouteFilterType = "RequestHeaderModifier"
```
hbagdi, line 402: In most places in this API, each `FooType` resource is named as `FooBar`.

Should we name this as `HTTPRouteFilterRequestHeaderModifier`?

This has a benefit of grouping the related constants in Godoc and also helps with code-completion like tools.

apis/v1alpha1/httproute_types.go
```diff
@@ -328,79 +328,94 @@ type HTTPRouteMatch struct {
...
 	// +optional
-	RequestMirror *HTTPRequestMirrorFilter `json:"requestMirror,omitempty"`
+	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
 }
 
-// HTTPRequestHeaderFilter defines configuration for the
-// RequestHeader filter.
+// HTTPRouteFilterType identifies a type of HTTPRoute filter.
+// +kubebuilder:validation:Enum=RequestHeaderModifier;RequestMirror;Custom
+type HTTPRouteFilterType string
```
hbagdi, line 395: Question: should this instead be called RouteFilterType?

Two reasons:
1. We could keep doors open for filters on other protocols. Unlikely and I'm myself not convinced if we will ever get there.
2. Some implementations might chose to also use these filters in resources other than HTTPRoute.

apis/v1alpha1/httproute_types.go
```diff
@@ -328,79 +328,94 @@ type HTTPRouteMatch struct {
...
+	// different backend. The responses from this backend MUST be ignored by
+	// the Gateway.
+	//
+	// Support: Extended
+	RequestMirrorHTTPRouteFilter HTTPRouteFilterType = "RequestMirror"
+
+	// CustomHTTPRouteFilter should be used for configuring custom HTTP filters.
+	//
+	// Support: Implementation-specific
+	CustomHTTPRouteFilter HTTPRouteFilterType = "Custom"
```
hbagdi, line 414: We usually resort to `ImplementationSpecific` literal string in most places in the API, any reason for using `Custom` here?

I like the conciseness of `Custom` over `ImplementationSpecific` but we should also be consistent. Should we consider changing to `Custom` in all places?

hbagdi (commented):
Nice cleanup and additions. Thanks @danehans!

apis/v1alpha1/httproute_types.go
hbagdi: This also relates to https://github.com/kubernetes-sigs/service-apis/issues/413

apis/v1alpha1/httproute_types.go
danehans: Although filters may be common across the protocol stack, they will have different meaning so I figured they should be protocol-specific. The following is an example of the request header modifier for `TCPRoute`:
```
type TCPRouteRule struct {
...
	// Filters define the filters that are applied to requests that match
	// this rule.
	Filters []TCPRouteFilter `json:"filters,omitempty"`
}

// TCPRouteFilterType identifies a type of TCPRoute filter.
type TCPRouteFilterType string

const (
	// RequestHeaderModifierTCPRouteFilter can be used to modify the header of
        // a TCP segment before it is sent to the upstream target.
	//
	// Support: Core
	RequestHeaderModifierTCPRouteFilter TCPRouteFilterType = "RequestHeaderModifier"
...
)
```

Thoughts?

apis/v1alpha1/httproute_types.go
danehans: I'll change to `ImplementationSpecific`. I went with `Custom` for brevity, but we can readdress across the API types in a separate PR.

examples/http-filter-mod-req-header.yaml
danehans: I used `mod-req-header` in the filename because I plan on having several filter examples. I'll rename the example to `http-filter.yaml` and can rename when I submit the subsequent PR that provides additional filter examples.

danehans:
Fixes https://github.com/kubernetes-sigs/service-apis/issues/413

apis/v1alpha1/httproute_types.go
hbagdi: I don't see any harm in doing that. It leads to having more types in the API, but then I could also argue that separating out types is better in this case. Let's leave this be.

robscott (commented):
Thanks for your work on this!

robscott:
Thanks!

/approve

Issue #206, opened 2020-06-04
Discuss domain match vs. wildcard match for listener hostnames


jpeach:
In #193, we define a "Domain" match type to match a set of hostnames to a single domain name. The semantics of matching the "example.com" domain are the same as matching a TLS wildcard of "*.example.com" (i.e. a single leftmost domain label is wild).

In #193 there was some discussion that "wildcard" terminology was easier to understand. So we could consider changing this:
```yaml
hostname:
  match: Domain
  name: example.com
```

to this:
```yaml
hostname:
  match: Wildcard
  name: *.example.com
```

bowei:
Couple things to clarify:

* Does domain match suffix or only a single level (a la TLS SAN)?
* Do people care about the more esoteric wildcard expressions e.g. `a*z.example.com`. Are there compelling enough users to support this?

I do like the simplicity of Domain to describe and understand.

jpeach:
> Couple things to clarify:
> 
> * Does domain match suffix or only a single level (a la TLS SAN)?

A single level.

> * Do people care about the more esoteric wildcard expressions e.g. `a*z.example.com`. Are there compelling enough users to support this?

We could also introduce these sorts of matches with separate Wildcard or Regex match types (might be proxy-dependent implementations).

hbagdi:
> A single level.

James, in our conversation, I think you mentioned this to support subdomains as well. This is certainly a feature that users want.

> Do people care about the more esoteric wildcard expressions

I've seen people do this with headers and paths but (I daresay) never with host.

jpeach:
> > A single level.
> 
> James, in our conversation, I think you mentioned this to support subdomains as well. This is certainly a feature that users want.

If you have more than 1 level of wildcard, how do you select the right server certificate (since that is only allowed to match a single level)?

hbagdi:
> If you have more than 1 level of wildcard, how do you select the right server certificate (since that is only allowed to match a single level)?

I think our minds are on different tracks because of bad naming.
What I meant was (and maybe @bowei too) was that is it possible to do `*.subdomain.example.com` or not? And the answer to that is yes.

jpeach:
> On 9 Jun 2020, at 12:32 pm, Harry <notifications@github.com> wrote:
> 
> 
> If you have more than 1 level of wildcard, how do you select the right server certificate (since that is only allowed to match a single level)?
> 
> I think our minds are on different tracks because of bad naming.
[...]

Issue #410, opened 2020-10-22
Gateway and HTTPRoute hostname matching should be more similar


robscott:
On Gateway we support the following HostnameMatchTypes:

- Domain
- Exact
- Any

On HTTPRoute we rely on wildcard for similar functionality (`*`, `*.foo.com`, and `foo.com` are all valid options).

robscott:
I think there's a lot of overlap with this issue and #403. I think the pattern we've used for hostname matching in HTTPRoute is easy to understand and has the added benefit of being similar in nature to the Ingress host config. I propose replacing Gateway hostname config with a copy of the hostnames field on HTTPRoute. This field would behave in the same way as the field on HTTPRoute, if no hostnames were specified, all requests would match.

Although the current approach on Gateway is inherently more extensible, all of the existing match types can easily be communicated with wildcards. I haven't been able to think of any potential extension points where we might want to add more match types here, so the simpler structure seems like it would be preferable here.

jpeach:
In general, I think that the HostnameMatchTypes is more explicit and hence easier to understand. For example, you can't easily know that `foo.*.example.com` is not a valid hostname match otherwise. Making the semantics explicit with an enum seems to me to be more in the spirit of the Kubernetes API.

robscott:
`foo.*.example.com` would also be invalid for HTTPRoute:

>The wildcard character '*' must appear by itself as the first DNS label and matches only a single label

I think the wildcard pattern in general is fairly widely used and understood at this point. Here's what I think a transition would look like:

### Any
**Current**
```
hostname:
  match: Any
```

**Proposed**
No config needed


### Exact
**Current**
```
hostname:
  match: Exact
  name: foo.com
```

**Proposed**
```
hostname: foo.com
```

### Domain
**Current**
```
hostname:
  match: Exact
  name: foo.com
```

**Proposed**
```
hostname: *.foo.com
```


I'm definitely a fan of enums as I've introduced plenty of them here, but in this case I think it may add more complexity than necessary.

robscott:
I've opened #416 to explore what this approach would actually look like.

PR #416, opened 2020-10-24
Updating Gateway hostname to more closely match HTTPRoute hostname

### LINKED ISSUE #403: Gateway listeners default for HostnameMatch
**What happened**:
I was trying to use a Gateway for protocols without a concept of "hostname".
The HostnameMatch field in Gateway defaults to `match: any` which is ambiguous. 

One should be able to use protocols that don't have the concept of hostnames.

```
  listeners:
  - hostname:
      match: Any
    port: 8080
    protocol: TCP
    routes:
      group: networking.x-k8s.io
      kind: TCPRoute
      routeNamespaces:
        from: All
      routeSelector:
        matchLabels:
          app: foo
```

robscott:
In issues #403 and #410 I'd proposed this approach as a way to both simplify defaults and more closely match HTTPRoute configuration. I thought it would actually be helpful to actually try this out and see how it would actually work and if it would actually make sense. If this does end up making sense, it will fix #403 and #410. 

/cc @jpeach @hbagdi


apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,27 @@ type GatewaySpec struct {
...
-	// If a match type other than "Any" is supplied, it MUST
-	// be compatible with the specified Protocol field.
+	// 1. IPs are not allowed.
+	// 2. The `:` delimiter is not respected because ports are not allowed.
+	//
+	// Hostname can be "precise" which is a domain name without the terminating
+	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
+	// domain name prefixed with a single wildcard label (e.g. "*.example.com").
+	// The wildcard character '*' must appear by itself as the first DNS label
+	// and matches only a single label.
```
jpeach, line 142: What does a hostname of`*` mean?

hbagdi:
This makes host matching consistent across Gateway and HTTPRoute (and potentially other rouetes which have the concept of hot).
+1

apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,27 @@ type GatewaySpec struct {
...
 	//
 	// Support: Core
 	//
-	// +kubebuilder:default={match: "Any"}
-	Hostname HostnameMatch `json:"hostname,omitempty"`
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
+	Hostname *string `json:"hostname,omitempty"`
```
hbagdi, line 148: Did you consider reusing `HTTPRouteHostname` type for this field or maybe refactoring to a common `Hostname` type?

apis/v1alpha1/gateway_types.go
robscott: Good call, I've updated the spec to clarify that this is equivalent to the hostname field being empty or unspecified.

apis/v1alpha1/gateway_types.go
robscott: Good call, updated to share a generic Hostname type between both Gateway and HTTPRoute.

apis/v1alpha1/gateway_types.go
jpeach: IIRC,`*` is explicitly not allowed in HTTPRoute?

robscott: Yeah, I'm not sure why we made that distinction there. It is really not necessary here either, but since our minLength is 1, we can't prevent `*` from being set. It seemed safest just to treat that as one would expect, identical to the empty case. HTTPRoute is a bit different because we do have a list of hostnames, but I'm not sure that we need to explicitly prevent `*` from being set there. For reference, the comment on HTTPRoute states:

```
You cannot have a wildcard label by itself (e.g. Host == "*")
```

apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,27 @@ type GatewaySpec struct {
...
-	// according to the given HostnameMatchType to select
-	// the Routes from this Listener.
+	// Hostname is the fully qualified domain name of a
+	// network host, as defined by RFC 3986. Note the following deviations from
+	// the "host" part of the URI as defined in the RFC:
 	//
-	// If a match type other than "Any" is supplied, it MUST
-	// be compatible with the specified Protocol field.
+	// 1. IPs are not allowed.
+	// 2. The `:` delimiter is not respected because ports are not allowed.
```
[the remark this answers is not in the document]
robscott: Yeah that's fair. I think the answer is simply that both protocol and port are already defined elsewhere on the Gateway and we want to keep that consistency for protocols where hostname doesn't make sense.

apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,27 @@ type GatewaySpec struct {
...
-	// Incoming requests that include a hostname are matched
-	// according to the given HostnameMatchType to select
-	// the Routes from this Listener.
+	// Hostname is the fully qualified domain name of a
+	// network host, as defined by RFC 3986. Note the following deviations from
+	// the "host" part of the URI as defined in the RFC:
 	//
-	// If a match type other than "Any" is supplied, it MUST
-	// be compatible with the specified Protocol field.
+	// 1. IPs are not allowed.
```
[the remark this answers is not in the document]

apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,27 @@ type GatewaySpec struct {
 // combination of Hostname, Port, and Protocol. This will be enforced by a
 // validating webhook.
 type Listener struct {
-	// Hostname specifies to match the virtual hostname for
-	// protocol types that define this concept.
+	// Hostname specifies the virtual hostname to match for protocol types that
+	// define this concept. When unspecified, all hostnames are matched.
```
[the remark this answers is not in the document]
robscott: Good call, there were lots of places to update here. I'm not quite sure I got them all right though, PTAL.

apis/v1alpha1/gateway_types.go
```diff
@@ -125,20 +125,26 @@ type GatewaySpec struct {
...
+	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
+	// domain name prefixed with a single wildcard label (e.g. "*.example.com").
+	// The wildcard character '*' must appear by itself as the first DNS label
+	// and matches only a single label.
 	//
 	// Support: Core
 	//
-	// +kubebuilder:default={match: "Any"}
-	Hostname HostnameMatch `json:"hostname,omitempty"`
+	// +optional
```
hbagdi, line 146: I remember Tim had a comment in pre-alpha review that we should explicitly add in a line for protocols that don't care about or don't have the concept of hostname.

We currently do have:
```
When unspecified or "*", all hostnames are matched
```

Should we add in something along the lines of "This field can be omitted for protocols that don't require hostname based matching".

apis/v1alpha1/gateway_types.go
```diff
@@ -731,10 +671,10 @@ const (
 	// interoperability.
 	ListenerConditionConflicted ListenerConditionType = "Conflicted"
 
-	// ListenerReasonHostnameConflict is used when the Listener
-	// violates the Hostname match constraints that allow collapsing
-	// Listeners. For example, this reason would be used when multiple
-	// Listeners on the same port use the "Any" hostname match type.
+	// ListenerReasonHostnameConflict is used when the Listener conflicts with
+	// hostnames in other Gateway Listeners. For example, this reason would be
```
hbagdi, line 675: nit: `s/other Gateway Listeners/other Listeners` because Listener, although not a resource, is a well known construct. This made me think and confuse about  listeners on other Gateway. (that's probably because merging is a topic on our minds right now).

apis/v1alpha1/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha1/gateway_types.go

hbagdi (approved):
LGTM. Deferring the final lgtm to @jpeach.

danehans:
@robscott this PR needs rebasing.

robscott:
@danehans rebased now, thanks for catching that!

hbagdi:
Fixes #206

PR #419, opened 2020-10-26
Installing golint if it doesn't exist

### LINKED ISSUE #418: verify-golint is failing in CI
It looks like golint is not present in the latest CI images and therefore our verify-golint script is failing. I've looked through the last few CI images and haven't figured out when it was removed yet. It seems like we should install golint if it does not exist locally.

/assign

robscott:
This should fix #418.


robscott:
Seems to have worked - from that passing run:
```
golint not found, installing
```

PR #420, opened 2020-10-27
Fixing verify-golint by using mod=readonly go flag

robscott:
This change has been verified to work in https://github.com/kubernetes-sigs/service-apis/pull/416. 

/cc @jpeach @hbagdi

### DIFF
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -18,11 +18,12 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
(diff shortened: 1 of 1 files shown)


jpeach:
Is there an explanation of what this fixes? Best to include that in the commit message .

robscott:
@jpeach Good call, I've updated the commit message to be more descriptive and the PR title to match. I'm still not quite sure I 100% understand why the tests were still failing after the initial fix. Obviously this passed for the CI run on the previous PR, but any PR that modified go files seemed to fail even after the update. My theory is that the changes to go.mod and go.sum resulted in something else failing, this ensures those files won't be changed. This time I've also tested this fix on #416 to ensure the change also fixes PRs that change go files.

PR #421, opened 2020-10-27
define app-protocol annotation for k8s < 1.18

hbagdi:
This patch provides a stop-gap solution for AppProtocol field for k8s < 1.18.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/396


apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,23 @@ const (
...
+	// on the `Service` resource is not available. Using this annotation is highly
+	// discouraged when AppProtocol field is available. The AppProtocol field, when
+	// populated, takes precedence over this annotation.
+	//
+	// The annotation value must be a comma separated with each value containing
+	// a port number followed by a forward slash and the protocol.
+	// Examples:
+	//
+	// - `networking.x-k8s.io/app-protocol: 8000/http,8443/https`
+	// - `networking.x-k8s.io/app-protocol: 4000/tcp,5000/tls`
```
robscott, line 51: I think my thoughts from the previous PR still apply here. I think this would be cleaner on BackendPolicy where the field can simply refer to a single protocol, not a port/protocol combination like we'd need here.

hbagdi: Could you elaborate a little on how this field in BackendPolicy work in k8s > 1.18? Will that field continue to exist forever? If yes, will it take precedence over AppProtocol?

robscott: To clarify, I'm still suggesting an annotation, just on BackendPolicy. Initially I think the guidance should be that AppProtocol defined on the Service should be given precedence over this annotation. Maybe in the future we'll see more use cases here and we can graduate the annotation to a field. Or maybe most of our users will already be on k8s 1.18 and we won't need this annotation for other resources. Starting on BackendPolicy just seems like it gives us the most flexibility + the simplest config.

hbagdi: > To clarify, I'm still suggesting an annotation, just on BackendPolicy.

Ah, I was understanding this all wrong.
Having an annotation on BackendPolicy that works similar to a field on the Service resource seems odd. It sits somewhere in the middle of "having a property in the BackendPolicy resource" and "having an annotation on the backend resource (i.e. Service in this case).

I do see the flexibility that it would bring, especially for non-Service backends. But then we need to make a decision if we expect AppProtocol to be the place for protocol (and then by extension, other resources which can be backends should also have a similar field) or is the protocol something that stays in BackendPolicy and that applies for all resource types.

I think we need to agree on the problem and the scope here. Are we providing a stop-gap solution for AppProtocol or do we see this as a larger problem and want to fix it properly?

hbagdi: > To clarify, I'm still suggesting an annotation, just on BackendPolicy.

Ah, I was understanding this all wrong.
Having an annotation on BackendPolicy that works similar to a field on the Service resource seems odd. It sits somewhere in the middle of "having a property in the BackendPolicy resource" and "having an annotation on the backend resource (i.e. Service in this case).

I do see the flexibility that it would bring, especially for non-Service backends. But then we need to make a decision if we expect AppProtocol to be the place for protocol (and then by extension, other resources which can be backends should also have a similar field) or is the protocol something that stays in BackendPolicy and that applies for all resource types.

I think we need to agree on the problem and the scope here. Are we providing a stop-gap solution for AppProtocol or do we see this as a larger problem and want to fix it properly?

apis/v1alpha1/httproute_types.go
```diff
@@ -511,6 +511,14 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In absence of `AppProtocol` field
+	// (Kubernetes versions < 1.18), `networking.x-k8s.io/app-protocol` annotation
+	// may be used to define the protocol. The annotation serves as a stop-gap
+	// solution for the `AppProtocol` field and it's use is highly discouraged
+	// when AppProtocol field is available. The AppProtocol field, when populated,
+	// takes precedence over the annotation.
```
Miciah, line 520:
```suggestion
	// The protocol to use is defined using the `AppProtocol` field (introduced
	// in Kubernetes 1.18) in the Service resource. In the absence of the `AppProtocol` field
	// (Kubernetes versions < 1.18), the `networking.x-k8s.io/app-protocol` annotation
	// may be used to define the protocol. The annotation serves as a stop-gap
	// solution for the `AppProtocol` field, and its use is highly discouraged
	// when the AppProtocol field is available. The AppProtocol field, when populated,
	// takes precedence over the annotation.
```

danehans:
Due to the requirements of structured annotation values, I'm in favor of adding an "app-protocol" similar field to `BackendPolicy`.

hbagdi:
> Due to the requirements of structured annotation values

Here is an alternate solution: we don't require the annotation to be structured. The users need to specify the protocol value only. The gateway will use the protocol defined in the annotation whenever it forwards traffic to this service. This does have obvious limitations but IME, it will work for 80-90% of use-cases.

hbagdi:
@robscott @danehans @Miciah I've updated the PR based on our discussion in the last meeting. The annotation should be present on BackendPolicy resource now. PTAL.

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,24 @@ const (
 	GatewayAllowSameNamespace GatewayAllowType = "SameNamespace"
 )
 
+const (
+	// AnnotationAppProtocol defines the protocol a Gateway should use for
+	// communication with a Kubernetes Service.
+	// This annotation must be applied to the `BackendPolicy` resource and will
+	// apply to all Service resources that are selected by
+	// `BackendPolicy.Spec.BackendRefs`. The annotation serves as a stop-gap
```
robscott, line 43: It's likely worth clarifying that BackendRefs can select Service ports directly. So this annotation would apply to Service ports selected by BackendRefs.

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In absence of `AppProtocol` field
```
robscott, line 504:
```suggestion
	// in Kubernetes 1.18) in the Service resource. In the absence of the `AppProtocol` field
```

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
...
+	// (Kubernetes versions < 1.18), `networking.x-k8s.io/app-protocol` annotation
```
robscott, line 505: Since you just specified the Kubernetes version above, I don't think it's needed again.
```suggestion
	// a `networking.x-k8s.io/app-protocol` annotation
```

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
...
+	// on the BackendPolicy resource may be used to define the protocol.
+	// The annotation serves as a stop-gap solution for the `AppProtocol` field
+	// and it's use is highly discouraged when AppProtocol field is available.
```
robscott, line 508:
```suggestion
	// and its use is highly discouraged when the AppProtocol field is available.
```

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,24 @@ const (
...
+	// solution for the Kubernetes versions < 1.18, where the `AppProtocol`
+	// field on the `Service` resource is not available. Using this annotation
+	// is highly discouraged when AppProtocol field is available. The AppProtocol
+	// field, when populated, takes precedence over this annotation.
+	//
+	// Examples:
+	//
+	// - `networking.x-k8s.io/app-protocol: https`
+	// - `networking.x-k8s.io/app-protocol: tls`
```
robscott, line 52: Do we need a more stringent spec here that matches the `AppProtocol` description as far as protocol names + format supported?

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/route_types.go
hbagdi: Given that we are using this annotation as a stop-gap, should we avoid doing it?
Or is this an API that we think is a stop-gap but will eventually stay around for a long-time?

Defining an unclear behavior gives future ourselves another reason to remove it, but then it also has the potential to confuse users. I don't think users would be confused here though.

apis/v1alpha1/route_types.go
hbagdi: I did some minor rephrasing here. PTAL.

apis/v1alpha1/route_types.go
robscott: I think it would be good just to say that any values set for this annotation should also pass validation for the `AppProtocol` field (or something along those lines). I agree that does seem relatively obvious, but I just want to be as clear as possible with the spec here.

hbagdi: Added in, PTAL. Thanks!

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
+	// in Kubernetes 1.18) in the Service resource. In the absence of the `AppProtocol` field
+	// a `networking.x-k8s.io/app-protocol` annotation
+	// on the BackendPolicy resource may be used to define the protocol.
+	// The annotation serves as a stop-gap solution for the `AppProtocol` field
+	// and its use is highly discouraged when the AppProtocol field is available.
```
robscott, line 508:
```suggestion
	// If the AppProtocol field is available, this annotation should not be used.
```

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using `AppProtocol` field (introduced
```
robscott, line 503: Tiny nit, I think we should avoid backticks here. I haven't seen them used widely in k8s godocs.

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,16 @@ type HTTPRouteForwardTo struct {
...
+	// in Kubernetes 1.18) in the Service resource. In the absence of the `AppProtocol` field
+	// a `networking.x-k8s.io/app-protocol` annotation
+	// on the BackendPolicy resource may be used to define the protocol.
```
robscott, line 506: Tiny nit, the line length here is a bit inconsistent.

apis/v1alpha1/route_types.go
```diff
@@ -35,6 +35,26 @@ const (
...
+	// field on the `Service` resource is not available. Using this annotation
+	// is highly discouraged when AppProtocol field is available. The AppProtocol
+	// field, when populated, takes precedence over this annotation.
+	// The value of the annotation must be validated same as `ProtocolType`
+	// in this package. If invalid, the annotation is ignored.
+	//
+	// Examples:
+	//
+	// - `networking.x-k8s.io/app-protocol: HTTPS`
+	// - `networking.x-k8s.io/app-protocol: TLS`
```
robscott, line 54: I know what I'd said here earlier was rather confusing, but I think it actually makes the most sense to match validation on the AppProtocol field, not our protocol field on Gateways. The key part of that validation is that the value must pass [isQualifiedName](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L42). That's certainly more flexible than our `ProtocolType` validation here. I think that's important for something like AppProtocol. So maybe we should say something like this:

```
The value of this annotation must be also be a valid value for the AppProtocol field.
```

docs/tls/index.html
```diff
@@ -884,6 +884,8 @@ <h3 id="example">Example</h3>
 apiVersion: networking.x-k8s.io/v1alpha1
 metadata:
   name: my-app
+  annotations:
+    networking.x-k8s.io/app-protocol: https
```
robscott, line 888: Nit: This doesn't match your example in the godocs above.

robscott (commented):
Thanks for your work on this! Mostly LGTM, just a few minor nits.

docs/tls/index.html
hbagdi: Good catch. Updated.

hbagdi:
@robscott I think I've incorporated all the changes and replicated them wherever applicable. PTAL.

apis/v1alpha1/httproute_types.go
```diff
@@ -500,6 +500,15 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// The protocol to use is defined using AppProtocol field (introduced in
+	// Kubernetes 1.18) in the Service resource. In the absence of the
+	// AppProtocol field a `networking.x-k8s.io/app-protocol` annotation on the
```
robscott, line 505: Sorry, one more nit. I think all that's really needed is a `the` in front of `AppProtocol`, but I also experimented with a slightly different way to phrase the first two sentences, not 100% sure it's better.

```suggestion
	// The protocol to use should be specified with the AppProtocol field on Service
	// resources. This field was introduced in Kubernetes 1.18. If using an earlier version
	// of Kubernetes, a `networking.x-k8s.io/app-protocol` annotation on the
```

robscott:
Thanks!

/approve

Issue #414, opened 2020-10-22
Require port to be specified whenever a Service is targeted


robscott:
From prealpha review - the following comment should be removed. 

```
// If the port is unspecified and a Service object consisting
// of a single port definition is the backend, that port will
// be used.
```

Port should always be specified when a Service is the target resource. Leaving this in would risk significant breaking changes if a second port was added to a Service.

robscott:
For HTTPRoute we may want to default port to 80.

hbagdi:
What are the downsides of always requiring a port?

It helps with debugging where a Gateway and Route resource tell you all you need to know about the network path of a request.

robscott:
I can't think of many downsides other than one more field to specify in manifests. I think requiring a port would be a good improvement to the API, especially if we could have a sensible default of 80 on HTTPRoute.

hbagdi:
I see. I think I understand what you mean. Let me push up a patch for this now and we can take it from there.

Issue #346, opened 2020-09-24
Validate examples against CRD schema


howardjohn:
Currently our examples fall out of date pretty quickly. We should validate these against the CRD schema. As a side effect, this will allow us to see awkward validation logic in the PR.

I could not find any tools to do this (aside from a real cluster). I tried to write one: https://github.com/istio/istio/blob/df7c20c2a261fa08dd93c9e09e65dd973a30c454/pkg/config/crd/validator.go#L66 and got really close but am stuck on one issue: https://kubernetes.slack.com/archives/C0EG7JC6T/p1600914427006300?thread_ts=1600454679.014900&cid=C0EG7JC6T

hbagdi:
@howardjohn Just following up, were you able to make any progress towards this?

howardjohn:
No, I am stuck on this by https://github.com/kubernetes/kubernetes/issues/95407

howardjohn:
/reopen

I finally got a local CRD validator working 100%. I tested it against all our examples and it fails on all the invalid ones and success on all valid ones. If we are interested I can send a PR to use it.

Either we can maintain the tool in tree here or can import from my repo once I push it there, either one works for me...

hbagdi:
@robscott @youngnick This is exactly what I was referring to when I was saying we should have a common validation package that all implementations can run before processing any objects.

hbagdi:
@howardjohn It makes a lot of sense to have a shared package for this. Would love to hear thoughts from others.

youngnick:
I commented on #926, but in short, I agree.

robscott:
Agree that we should work on a common validation package that merges the logic @howardjohn has for CRD validation + our existing webhook validation code.

robscott:
This was fixed by https://github.com/kubernetes-sigs/gateway-api/pull/428

PR #422, opened 2020-10-27
make port a required field

hbagdi:
See #414


robscott:
Thanks!

/approve

robscott:
@hbagdi It looks like this will require some updates to our examples to actually specify ports.

hbagdi:
> @hbagdi It looks like this will require some updates to our examples to actually specify ports.

Ah, the examples as always. We really need CI to catch that. Waiting for #346.
Thanks for catching that.

/hold

hbagdi:
Updated.
/hold cancel

jpeach:
Making the port required is a good idea, but I don't think we should set a default.

Do arbitrary apps really listen on port 80 by default? I would be pretty surprised. Even if there was a predominance of ports 80 listeners, that requires running as root in the container (for the low-numbered port) and we should prefer to not encourage that :)

howardjohn:
> Do arbitrary apps really listen on port 80 by default? I would be pretty surprised. Even if there was a predominance of ports 80 listeners, that requires running as root in the container (for the low-numbered port) and we should prefer to not encourage that :)

I cannot give any real numbers other than general feeling but it seems `servicePort=80,targetPort>1024` is pretty common from what I have seen, so I don't think this encourages root?

robscott:
Yeah I think it's good to keep in mind that this is the port the Service is listening on, not the container itself. It's probably fairly likely that an HTTP service would listen on 80 and then target some other port that the containers were listening on.

jpeach:
> Yeah I think it's good to keep in mind that this is the port the Service is listening on, not the container itself.

Good point, I had forgotten that 👍 

>  It's probably fairly likely that an HTTP service would listen on 80 and then target some other port that the containers were listening on.

Maybe. I'm still unconvinced that this is an improvement. If the default is wrong you'll have to debug it (high cost), and any benefit is very small.

robscott:
>Maybe. I'm still unconvinced that this is an improvement. If the default is wrong you'll have to debug it (high cost), and any benefit is very small.

👍 that's fair. Always easier to add a default in the future than to remove one. I'm fine with leaving that out of this change.

hbagdi:
I agree with the thoughts here. Updated to remove the defaulting logic.

apis/v1alpha1/httproute_types.go
```diff
@@ -535,24 +520,11 @@ type HTTPRouteForwardTo struct {
...
-	// should be updated with a condition that describes the error
-	// more specifically.
-	//
 	// Support: Core
 	//
-	// +optional
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
-	Port *int32 `json:"port,omitempty"`
+	Port int32 `json:"port,omitempty"`
```
jpeach, line 527:
```suggestion
	Port int32 `json:"port"`
```

hbagdi: Updated.

robscott:
Thanks!

/lgtm

PR #423, opened 2020-10-27
define a new Go type Port for consolidation 

hbagdi:
- Defines a new type Port
- Re-use the type across all instances of port
- Change the type from int32 to uint32

### Note

I spotted this while working on #422.
Please review only the last commit here. To avoid conflicts, this PR builds on top of #422 and should be merged after that lands in.

### DIFF
--- a/apis/v1alpha1/route_types.go
+++ b/apis/v1alpha1/route_types.go
@@ -55,6 +55,12 @@ type RouteGateways struct {
 	GatewayRefs []GatewayReference `json:"gatewayRefs,omitempty"`
 }
 
+// PortNumber defines a network port.
+//
+// +kubebuilder:validation:Minimum=1
+// +kubebuilder:validation:Maximum=65535
+type PortNumber int32
+
 // GatewayReference identifies a Gateway in a specified namespace.
 type GatewayReference struct {
 	// Name is the name of the referent.
@@ -106,10 +112,7 @@ type RouteForwardTo struct {
 	// backend referenced by the ServiceName or BackendRef field.
 	//
 	// Support: Core
-	//
-	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=65535
-	Port int32 `json:"port"`
+	Port PortNumber `json:"port"`
 
 	// Weight specifies the proportion of traffic forwarded to the backend
 	// referenced by the ServiceName or BackendRef field. computed as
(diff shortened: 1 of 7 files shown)


apis/v1alpha1/route_types.go
```diff
@@ -55,6 +55,12 @@ type RouteGateways struct {
 	GatewayRefs []GatewayReference `json:"gatewayRefs,omitempty"`
 }
 
+// Port defines a network port.
+//
+// +kubebuilder:validation:Minimum=1
+// +kubebuilder:validation:Maximum=65535
+type Port uint32
```
jpeach, line 62: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Do not use unsigned integers, due to inconsistent support across languages and libraries. Just validate that the integer is non-negative if that's the case.

The validation keeps it in range, so it should be OK to leave this as `int32`.

jpeach, line 62: Suggest that we call this type `PortNumber` for clarity.

hbagdi: `PortNumber` feels a little more verbose than it needs to be. What is the reasoning behind the suggestion?

robscott: In k/k our [Port types include a ton of fields](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3702), including a name and protocol. I think using `PortNumber` here may make the simplicity of this type more clear.

hbagdi: This is a type alias and not even a struct. Do we still need to be explicit even then?

robscott: It's a small detail, I'd just imagine that anyone browsing through the types might see `Port` in a `forwardTo` and wonder if that included all the things that `ServicePort` did or if it was just a port number. Obviously it's not difficult to figure that out, just takes an extra step.

hbagdi: I see. I've updated this to be `PortNumber`. Like you said, it is a small detail and I don't feel too strongly either ways.

robscott:
Thanks!

/lgtm
/approve

robscott:
Sorry, just had to double check that a port was intentionally optional.

/hold cancel

apis/v1alpha1/route_types.go
jpeach: Yeh, it's clarity ... `PortNumber` is immediately obvious and evocative of its meaning.

PR #425, opened 2020-10-27
multiple cleanups in docs

hbagdi:
- consolidate personas in a single doc
- add a new guides page which will refer to getting started guides and
provide cookbooks; removed cookbookmd and userguide.md as part of this
change
- re-wrote the intro page to better guide new users
- nav corrections and updates based on above changes


docs-src/security-model.md
```diff
@@ -22,25 +22,17 @@ security model:
...
+There are 3 primary roles in the API:
+
+- Infrastructure provider
+- Cluster Operator
+- Application Developer
+
+There could be a fourth a role of Application Admin in some use cases.
+
+Please refer to [roles and personas](concepts.md#roles-and-personas) section
+in API concepts for details.
```
robscott, line 35: Thanks for cleaning up the duplication between this and the API concepts pages. I'm actually wondering if we should switch these around though. This PR adds all the detail around roles to the API concepts page and leaves a short summary here - maybe the short summary should be in API concepts and the detailed overview should be here. Either way, it's better to not have the full detail in both places, so thanks for fixing that.

robscott (commented):
These are some awesome improvements, thanks for taking them on!

docs-src/security-model.md
hbagdi: SGTM. I've updated to switch this around.

docs-src/concepts.md
```diff
@@ -2,38 +2,18 @@
...
-design, refer to the [Security Model](security-model.md).
+## Roles and personas.
+
+There are 3 primary roles in the API:
+
+- Infrastructure provider
+- Cluster Operator
+- Application Developer
+
+There could be a fourth a role of Application Admin in some use cases.
```
robscott, line 13:
```suggestion
There could be a fourth role of Application Admin in some use cases.
```

docs-src/concepts.md
```diff
@@ -2,38 +2,18 @@
...
+
+Please refer to [roles and personas](security-model.md#roles-and-personas) section
```
robscott, line 15:
```suggestion
Please refer to the [roles and personas](security-model.md#roles-and-personas) section
```

docs-src/concepts.md
```diff
@@ -2,38 +2,18 @@
...
+in Security model for details.
```
robscott, line 16:
```suggestion
in the Security model for details.
```

robscott:
Some great improvements here, just had a few tiny nits.

/approve

robscott:
Thanks!

/lgtm

PR #426, opened 2020-10-27
add shortName to Gateway and GatewayClass

hbagdi:


apis/v1alpha1/gateway_types.go
```diff
@@ -22,6 +22,7 @@ import (
 
 // +genclient
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName="gw"
```
robscott, line 25: This also matches the [Istio short name](https://github.com/istio/istio/blob/0cbbae1d79b44ada06313fb29b32584b7af745aa/manifests/charts/base/files/gen-istio-cluster.yaml#L1453). If we were going to keep the same resource name it would probably make sense to differentiate short names, but I also can't think of a great alternative to `gw`. 

/cc @howardjohn

robscott: Maybe one of the following?

- gtw
- gwy
- gtwy

hbagdi: I had `gw` and `gtwy` as shortnames but remove `gtwy` to simplify. Can certainly add that back in.

hbagdi: What if we keep `gtwy` and `gw` both as shortnames and in our documentation, we stick to using `gtwy`?
This has the added benefit where the succulent `gw` shorthand can be used in clusters which don't have Istio installed without breaking any workflow for clusters which have Istio installed already.

Doesn't the first installed CRD wins and takes the shorthand reservation?

howardjohn: This still breaks Istio I think. Just installing the APIs already breaks `kubectl get gateway`. service-apis seems to always win, I don't know why. Maybe it is "last one wins" and I never tried service-apis first though

bowei: If we can have multitple short names, let's have both "gw" and "gtw".
Accounting for other CRDs that exist is going to be an impossible task, especially for a common word.

For users that only install service APIs, having the most natural (gw) makes it easier to use.

bowei: Is there a link to the precedence? I thought it starts rejecting due to name clash instead of auto resolving in a particular direction.

howardjohn: https://github.com/kubernetes/kubernetes/issues/94860

If we can fix kubectl to do something like warning, rejecting, etc I think its ok. But breaking Istio users feels not great to me, given we expect there to be a large overlap of Istio users and servce-api users. We are already breaking `kubectl get gateway`, it would be great to not force users to type `kubectl get gateways.networking.istio.io` every time

hbagdi: > Accounting for other CRDs that exist is going to be an impossible task, especially for a common word.

That's exactly my thinking as well. 

> That might be a bug...I distinctly remember kubectl rejecting me

AFAIK, one resource wins.

> If we can have multitple short names, let's have both "gw" and "gtw".

Yes, we can. Should it be `gw` and `gtw` or `gtwy`?

robscott: > Accounting for other CRDs that exist is going to be an impossible task, especially for a common word

> breaking Istio users feels not great to me, given we expect there to be a large overlap of Istio users and servce-api users. We are already breaking kubectl get gateway, it would be great to not force users to type kubectl get gateways.networking.istio.io every time

Although I agree that we can't guarantee that we'll have a completely unique name here, Istio seems like a pretty large project to conflict with. I think it's in our best interest to try to cooperate on this front. IE if both projects claim all the same resource names and short names it will simply depend on install order/whatever determines precedence.  I would rather not try to claim the short name that Istio is already using for this resource. If Istio has been installed in a cluster before Service APIs we risk providing a bad experience for users of our Gateway API if we're not using a unique short name.

With that said, my vote would be to use the `gtwy` short name for this project. We can also work with sig-cli to figure out a better long term solution, but until then, I'd rather not intentionally conflict with the `gw` name.

hbagdi:
@bowei @robscott @howardjohn I've updated this PR to use `gtw` and `gtwy` as shortnames. PTAL.
I'd love to get this in v1alpha1 as well if feasible.

robscott:
Thanks!

/approve

PR #427, opened 2020-10-27
write a very simple getting started guide

hbagdi:
## Note

This PR builds on top of #425. Please review the last commit only.

### DIFF
--- a/docs-src/getting-started.md
+++ b/docs-src/getting-started.md
@@ -1,0 +1,61 @@
+# Getting started with Services APIs
+
+## Installing CRDs
+
+This project provides a collection of Custom Resource Definitions (CRDs) that can
+be installed into any Kubernetes (>= 1.16) cluster.
+
+To install the CRDs, please execute:
+
+```
+kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
+| kubectl apply -f -
+```
+
+## Install an implementation
+
+Multiple projects implement the APIs defined by this project.
+You will need to either install an implementation or verify that one is already
+setup for your cluster.
+
+## Sample Gateway
+
+Once you have the CRDs and an implementation installed, you are ready to
+use Service APIs.
+
+In this example, we are installing three resources:
+
+- An `acme-lb` GatewayClass which is being managed by a `acme.io/gateway-controller`
+  controller running in the cluster. Typically, a GatewayClass is provided by
+  the implementation and must be installed in the cluster.
+- A Gateway which is of type `acme-lb`:
+    - This gateway has a single HTTP listener on port 80 which selects HTTPRoutes
+      from all namespaces which have the label `app: foo` on them.
+
+- Finally, we have an HTTPRoute resource which is attached to the above Gateway
+  and has two rules:
+    - All requests with path beginning with `/bar` are forwarded to my-service1
+      Kubernetes Service.
+    - All requests with path beginning with `/some/thing` AND have an HTTP header
+      `magic: foo` are forwarded to my-service2 Kubernetes Service.
+
+With this configuration, you now have a Gateway resource which is forwarding
+traffic to two Kubernetes Services based on HTTP request metadata.
+
+```
+{% include 'basic-http.yaml' %}
+```
+
+For more advanced examples, please read the other [guides](guides.md).
+
+## Uninstalling the CRDs
+
+To uninstall the CRDs and all resources created with them, run the following
+command. Note that this will remove all GatewayClass and Gateway resources in
+your cluster. If you have been using these resources for any other purpose do
+not uninstall these CRDs.
+
+```
+kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
+| kubectl delete -f -
+```
--- a/docs-src/guides.md
+++ b/docs-src/guides.md
@@ -3,4 +3,5 @@
 Guides demonstrate and provide examples of how to use the API.
 Please checkout one of the following guides:
 
+- [Getting started](getting-started.md)
 - [Configuring TLS](tls.md)
--- a/docs/404.html
+++ b/docs/404.html
@@ -262,6 +262,18 @@
 
 
   <li class="md-nav__item">
+    <a href="/getting-started/" title="Getting started" class="md-nav__link">
+      Getting started
+    </a>
+  </li>
+
+        
+          
+          
+          
+
+
+  <li class="md-nav__item">
     <a href="/tls/" title="TLS" class="md-nav__link">
       TLS
     </a>
--- a/docs/community/index.html
+++ b/docs/community/index.html
@@ -266,6 +266,18 @@
 
 
   <li class="md-nav__item">
+    <a href="../getting-started/" title="Getting started" class="md-nav__link">
+      Getting started
+    </a>
+  </li>
+
+        
+          
+          
+          
+
+
+  <li class="md-nav__item">
     <a href="../tls/" title="TLS" class="md-nav__link">
       TLS
     </a>
--- a/docs/concepts/index.html
+++ b/docs/concepts/index.html
@@ -537,6 +537,18 @@
 
 
   <li class="md-nav__item">
+    <a href="../getting-started/" title="Getting started" class="md-nav__link">
+      Getting started
+    </a>
+  </li>
+
+        
+          
+          
+          
+
+
+  <li class="md-nav__item">
     <a href="../tls/" title="TLS" class="md-nav__link">
       TLS
     </a>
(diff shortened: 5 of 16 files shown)


docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
...
+## Installing CRDs
+
+Service APIs is a collection of Kubernetes Custom Resource Definitions that need
+to be installed into any Kubernetes cluster.
+
+To install the CRDs, please execute:
+
+```
+kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
+| kubectl apply -f -
```
robscott, line 12: Is there a good way to do this without kustomize? It seems like we're not really using kustomize for anything. (This does not need to block this PR).

docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+## Install an implementation
+
+Service APIs is a set of specification that multiple implementations confirm to.
```
robscott, line 17: This doesn't feel quite right. Maybe "Multiple projects implement the APIs defined by this project."

docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
...
+setup for your cluster.
+
+## Sample Gateway
+
+Once you have the CRDs and an implementation installed, you are ready to
+use Service APIs.
+
+In this example, we are installing three resources:
+
+- An `acme-lb` GatewayClass which is being managed by a `acme.io/gateway-controller` controller running in the cluster.
```
robscott, line 28: Nit: This line feels too long.

docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
...
+- A Gateway which is of type `acme-lb`:
+    - This gateway has a single HTTP listener on port 80 which selects HTTPRoutes
+      from all namespaces which have the label `app: foo` on them.
+
+- Finally, we have an HTTPRoute resource which is attached to the above Gateway
+  and has two rules:
+    - All requests with Path beginning with `/bar` are forwarded to my-service1
```
robscott, line 35: Nit: I'm not sure path needs to be capitalized here. I think it could use an `a` before it though.

docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
...
+      Kubernetes Service.
+    - All requests with Path beginning with `/some/thing` AND have an HTTP header
```
robscott, line 37: Same comment as above

docs-src/getting-started.md
```diff
@@ -0,0 +1,55 @@
+# Getting started with Services APIs
+
+## Installing CRDs
+
+Service APIs is a collection of Kubernetes Custom Resource Definitions that need
+to be installed into any Kubernetes cluster.
```
robscott, line 6: Maybe "This project provides a collection of Custom Resource Definitions that can be installed into any Kubernetes 1.16+ cluster."

robscott (commented):
Thanks for your work on this!

docs-src/getting-started.md
hbagdi: We do this in our Makefile too: https://github.com/kubernetes-sigs/service-apis/blob/master/Makefile#L54

Do note that the user doesn't need to install `kustomize`. It is included as a plugin inside `kubectl` itself. Users need to make sure that their `kubectl` version is not an ancient one, which again is expected.

hbagdi:
Depends on #425 
/hold

hbagdi:
`golint` went missing so I rebased on master to ensure that I pull in the fixes we put in but not luck.
Is it the CI acting up?

robscott:
@hbagdi Our verify output is super confusing, in this case it looks like another run of `make generate` is needed:

```
Uncommitted changes in generated sources:
 M docs/getting-started/index.html
```

hbagdi:
Rebased and this should be good to go.
/hold cancel

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
+| kubectl apply -f -
+```
+
+## Install an implementation
+
+Multiple projects implement the APIs defined by this project.
+You will need to either install an implementation or verify if one is already
```
robscott, line 18: Nit
```suggestion
You will need to either install an implementation or verify that one is already
```

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Sample Gateway
+
+Once you have the CRDs and an implementation installed, you are ready to
+use Service APIs.
+
+In this example, we are installing three resources:
+
+- An `acme-lb` GatewayClass which is being managed by a `acme.io/gateway-controller`
+  controller running in the cluster.
```
robscott, line 29: Maybe should add a note that users may want to use a GatewayClass provided by the implementation installed in their cluster. If one is not provided, they should ensure that the controller name used here matches a controller name supported by an implementation installed in their cluster.

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+    - All requests with path beginning with `/bar` are forwarded to my-service1
+      Kubernetes Service.
+    - All requests with path beginning with `/some/thing` AND have an HTTP header
+      `magic: foo` are forwarded to my-service2 Kubernetes Service.
+
+With this configuration, you now have a Gateway reource which is forwarding
+traffic to two Kubernetes Services based on HTTP request metadata.
+
+```
+{% include 'basic-http.yaml' %}
```
robscott, line 45: I don't have a good solution for this, and this doesn't need to block the PR, but one downside of embedding examples is that the examples could actually begin to skew from the docs. Maybe we should find a way to document which examples are being used in which docs?

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+For more advanced examples, please checkout other [guides](guides.md).
```
robscott, line 48: Tiny nit:
```suggestion
For more advanced examples, please read the other [guides](guides.md).
```

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
+# Getting started with Services APIs
+
+## Installing CRDs
+
+This project provides a collection of Custom Resource Definitions (CRDs)that can
```
robscott, line 5: Nit:
```suggestion
This project provides a collection of Custom Resource Definitions (CRDs) that can
```

robscott (commented):
Thanks for the work on this! Mostly LGTM, just a few nits.

docs-src/getting-started.md
hbagdi: I thought about this problem too and arrived at the same solution as you.
Essentially adding a link to the guide file as a yaml comment.
Let's continue this in https://github.com/kubernetes-sigs/service-apis/issues/446.

docs-src/getting-started.md
hbagdi: > If one is not provided

Skipped this part. We shouldn't be encouraging implementers not providing a GatewayClass right now.

hbagdi:
@robscott Updated. PTAL.

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+      from all namespaces which have the label `app: foo` on them.
+
+- Finally, we have an HTTPRoute resource which is attached to the above Gateway
+  and has two rules:
+    - All requests with path beginning with `/bar` are forwarded to my-service1
+      Kubernetes Service.
+    - All requests with path beginning with `/some/thing` AND have an HTTP header
+      `magic: foo` are forwarded to my-service2 Kubernetes Service.
+
+With this configuration, you now have a Gateway reource which is forwarding
```
robscott, line 42:
```suggestion
With this configuration, you now have a Gateway resource which is forwarding
```

docs-src/getting-started.md
```diff
@@ -0,0 +1,56 @@
...
+traffic to two Kubernetes Services based on HTTP request metadata.
+
+```
+{% include 'basic-http.yaml' %}
+```
+
+For more advanced examples, please read the other [guides](guides.md).
+
+## Uninstalling the CRDs
+
```
robscott, line 52: Maybe add a bit of an explanation here, something like:

```
To uninstall the CRDs and all resources created with them, run the following command. Note that this will remove all GatewayClass and Gateway resources in your cluster. If you have been using these resources for any other purpose, do not uninstall these CRDs.
```

robscott (commented):
Thanks for your work on this! Just a couple small things but otherwise LGTM.

youngnick (commented):
This is looking great.

hbagdi:
@robscott Updated. PTAL.

robscott:
Thanks!

/lgtm
/approve

PR #428, opened 2020-10-27
Adding a way to verify examples with kind, fixing BackendPolicy example

robscott:
Fixes https://github.com/kubernetes-sigs/service-apis/issues/346.

/cc @hbagdi @jpeach

### DIFF
--- a/examples/backendpolicy.yaml
+++ b/examples/backendpolicy.yaml
@@ -8,10 +8,6 @@ spec:
       group: core
       kind: Service
   tls:
-    clientCertificateRef: 
-      name: my-app-cert
-      group: core
-      kind: Secret
     certificateAuthorityRef: 
       name: my-app-ca
       group: core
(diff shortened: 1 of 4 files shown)


robscott:
I believe tests will continue to fail on this PR until https://github.com/kubernetes/test-infra/pull/19722 gets merged in.

robscott:
@jpeach @hbagdi The test-infra PR that was blocking this has been merged in, tests are passing now. Would love to get this in ASAP to prevent any further breaking changes to examples.

hbagdi:
Sweet. Thanks Rob!
Question: What value does `hack/verify-examples.sh` add if we are adding a check using kind itself? Should that be removed?
cc @jpeach

robscott:
@hbagdi That's a great question - I think we can probably remove it but I'd rather save that for a follow up PR.

hbagdi (approved):
LGTM.
Deferring approval to @jpeach or @danehans.

robscott:
@danehans or @hbagdi - just fixed a couple examples that were missing a port that was now required. Please take another look and LGTM when you have a chance.

PR #429, opened 2020-10-28
Updates NamedAddressType and IPAddressType Godocs

### LINKED ISSUE #409: Clarify what happens when a NamedAddress is requested but not supported
The `NamedAddress` field is listed as `ImplementationSpecific`, what should an implementation do when this is requested in a manifest but can't actually be supported?

danehans:
- Updates `NamedAddressType` and `IPAddressType` in the `Gateway` API type to provide additional information when a a requested address is unsupported by an implementation.
- Introduces the "UnsupportedAddress" listener status condition reason constant.

/assign @robscott @jpeach 

Fixes #409

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -550,14 +550,18 @@ const (
 	// form. IPv6 addresses must be in a standard IPv6 text
 	// representation (see RFC 5952).
 	//
-	// Implementations should accept any address representation
-	// accepted by the inet_pton(3) API.
+	// If the requested address is unsupported, the controller
+	// should raise the "Detached" listener status condition on
+	// the Gateway with the "UnsupportedAddress" reason.
 	//
 	// Support: Extended.
 	IPAddressType AddressType = "IPAddress"
 
-	// NamedAddressType is an address selected by name. The interpretation of
-	// the name is dependent on the controller.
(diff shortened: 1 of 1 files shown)


apis/v1alpha1/gateway_types.go
```diff
@@ -551,13 +551,19 @@ const (
 	// representation (see RFC 5952).
 	//
 	// Implementations should accept any address representation
-	// accepted by the inet_pton(3) API.
+	// accepted by the inet_pton(3) API. If the requested address
```
bowei, line 554: we should probably lay this out from a different reference than linux manpages -- i noticed the bsd manpages have a slightly different description that doesn't include ipv6

bowei (commented):
doc comment

apis/v1alpha1/gateway_types.go
hbagdi: I'm not familiar with that RFC.
I can't seem to figure out what is it that we are trying to communicate here. Daneyon, could you give an example of what we mean by `any address representation`?

danehans:
xref https://github.com/kubernetes-sigs/service-apis/issues/443

robscott:
Thanks!

/approve

Will leave final LGTM for @hbagdi or @bowei

PR #436, opened 2020-10-29
Cleanup from prealpha feedback

robscott:
- Add an upper limit of 100 to the number of Gateways that will be stored in RouteGatewayStatus.
- Clarify support level of filters defined in ForwardTo.
- Increase max weight to 1 million.
- Clarify how 1 backend with a weight of 0 should be handled.
- Replace "undefined" with "unspecified"
- resource: fooroutes -> kind: FooRoute

Fixes https://github.com/kubernetes-sigs/service-apis/issues/435

/cc @danehans @hbagdi @jpeach

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -168,14 +168,14 @@ type HTTPRouteRule struct {
 	// Filters define the filters that are applied to requests that match
 	// this rule.
 	//
-	// The effects of ordering of multiple behaviors are currently undefined.
+	// The effects of ordering of multiple behaviors are currently unspecified.
 	// This can change in the future based on feedback during the alpha stage.
 	//
 	// Conformance-levels at this level are defined based on the type of filter:
 	// - ALL core filters MUST be supported by all implementations.
 	// - Implementers are encouraged to support extended filters.
 	// - Implementation-specific custom filters have no API guarantees across implementations.
-	// Specifying a core filter multiple times has undefined or custom conformance.
+	// Specifying a core filter multiple times has unspecified or custom conformance.
 	//
 	// Support: core
 	//
@@ -386,20 +386,23 @@ const (
 	// HTTPRouteFilterRequestHeaderModifier can be used to add or remove an HTTP
 	// header from an HTTP request before it is sent to the upstream target.
 	//
-	// Support: Core
+	// Support in HTTPRouteRule: Core
+	// Support in HTTPRouteForwardTo: Extended
 	HTTPRouteFilterRequestHeaderModifier HTTPRouteFilterType = "RequestHeaderModifier"
 
 	// HTTPRouteFilterRequestMirror can be used to mirror HTTP requests to a
 	// different backend. The responses from this backend MUST be ignored by
 	// the Gateway.
 	//
-	// Support: Extended
+	// Support in HTTPRouteRule: Extended
+	// Support in HTTPRouteForwardTo: Extended
 	HTTPRouteFilterRequestMirror HTTPRouteFilterType = "RequestMirror"
 
 	// HTTPRouteFilterExtensionRef should be used for configuring custom
 	// HTTP filters.
 	//
-	// Support: Implementation-specific
+	// Support in HTTPRouteRule: Custom
+	// Support in HTTPRouteForwardTo: Custom
 	HTTPRouteFilterExtensionRef HTTPRouteFilterType = "ExtensionRef"
 )
 
@@ -528,29 +531,30 @@ type HTTPRouteForwardTo struct {
 	//
 	Port PortNumber `json:"port"`
 
-	// Weight specifies the proportion of traffic forwarded to the backend
-	// referenced by the ServiceName or BackendRef field. computed as
-	// weight/(sum of all weights in this ForwardTo list). Weight is not a
-	// percentage and the sum of weights does not need to equal 100. If only one
-	// backend is specified, 100% of the traffic is forwarded to that backend.
-	// If weight is set to 0, no traffic should be forwarded for this entry.
-	// If unspecified, weight defaults to 1.
+	// Weight specifies the proportion of HTTP requests forwarded to the backend
+	// referenced by the ServiceName or BackendRef field. This is computed as
+	// weight/(sum of all weights in this ForwardTo list). For non-zero values,
+	// there may be some epsilon from the exact proportion defined here
+	// depending on the precision an implementation supports. Weight is not a
+	// percentage and the sum of weights does not need to equal 100.
+	//
+	// If only one backend is specified and it has a weight greater than 0, 100%
+	// of the traffic is forwarded to that backend. If weight is set to 0, no
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
 	//
 	// Support: Core
 	//
 	// +kubebuilder:default=1
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=10000
+	// +kubebuilder:validation:Maximum=1000000
 	Weight int32 `json:"weight,omitempty"`
 
 	// Filters defined at this-level should be executed if and only if the
 	// request is being forwarded to the backend defined here.
 	//
-	// Conformance: Filtering support, including core filters, is NOT guaranteed
-	// at this-level. Use Filters in HTTPRouteRule for portable filters across
-	// implementations.
-	//
-	// Support: Custom
+	// Support: Custom (For broader support of filters, use the Filters field
+	// in HTTPRouteRule.)
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
--- a/apis/v1alpha1/route_types.go
+++ b/apis/v1alpha1/route_types.go
@@ -114,19 +114,23 @@ type RouteForwardTo struct {
 	// Support: Core
 	Port PortNumber `json:"port"`
 
-	// Weight specifies the proportion of traffic forwarded to the backend
-	// referenced by the ServiceName or BackendRef field. computed as
-	// weight/(sum of all weights in this ForwardTo list). Weight is not a
-	// percentage and the sum of weights does not need to equal 100. If only one
-	// backend is specified, 100% of the traffic is forwarded to that backend.
-	// If weight is set to 0, no traffic should be forwarded for this entry.
-	// If unspecified, weight defaults to 1.
+	// Weight specifies the proportion of HTTP requests forwarded to the backend
+	// referenced by the ServiceName or BackendRef field. This is computed as
+	// weight/(sum of all weights in this ForwardTo list). For non-zero values,
+	// there may be some epsilon from the exact proportion defined here
+	// depending on the precision an implementation supports. Weight is not a
+	// percentage and the sum of weights does not need to equal 100.
+	//
+	// If only one backend is specified and it has a weight greater than 0, 100%
+	// of the traffic is forwarded to that backend. If weight is set to 0, no
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
 	//
 	// Support: Extended
 	//
 	// +kubebuilder:default=1
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=10000
+	// +kubebuilder:validation:Maximum=1000000
 	Weight int32 `json:"weight,omitempty"`
 }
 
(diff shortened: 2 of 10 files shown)


apis/v1alpha1/httproute_types.go
```diff
@@ -535,26 +535,29 @@ type HTTPRouteForwardTo struct {
...
+	// of the traffic is forwarded to that backend. If weight is set to 0, no
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
 	//
 	// Support: Core
 	//
 	// +kubebuilder:default=1
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=10000
+	// +kubebuilder:validation:Maximum=1000000000
```
hbagdi, line 551: Not all implementations will be able to support arbitrary weights.
Kong, for example, has an upper limit of `65535`.

apis/v1alpha1/httproute_types.go
```diff
@@ -535,26 +535,29 @@ type HTTPRouteForwardTo struct {
...
 	// Filters defined at this-level should be executed if and only if the
 	// request is being forwarded to the backend defined here.
 	//
-	// Conformance: Filtering support, including core filters, is NOT guaranteed
-	// at this-level. Use Filters in HTTPRouteRule for portable filters across
-	// implementations.
+	// Conformance: Support for filters defined at this level is NOT guaranteed.
+	// Labels implying "Core" or "Extended" support for a specific filter type
+	// do NOT apply at this level. For broader support of filters, use the
+	// Filters field defined one level higher as part of HTTPRouteRule.
```
hbagdi, line 560: Should we say that the conformance level here is extended or custom?

apis/v1alpha1/route_types.go
```diff
@@ -115,18 +115,20 @@ type RouteForwardTo struct {
...
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
 	//
 	// Support: Extended
 	//
 	// +kubebuilder:default=1
 	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=10000
+	// +kubebuilder:validation:Maximum=1000000000
 	Weight int32 `json:"weight,omitempty"`
```
hbagdi, line 132: Probably scope-creep for this PR: We should DRY this out and have a `type Weight int32` in this package.

apis/v1alpha1/route_types.go
```diff
@@ -166,5 +168,11 @@ type RouteStatus struct {
 	// manages the Gateway should add an entry to this list when the
 	// controller first sees the route and should update the entry as
 	// appropriate when the route is modified.
+	//
+	// A maximum of 100 Gateways will be represented in this list. If this list
+	// is full, there may be additional Gateways using this Route that are not
+	// included in the list.
+	//
+	// +kubebuilder:validation:MaxItems=100
```
hbagdi, line 176: Can we stick to powers of 2 here? How about 64 or 128?

apis/v1alpha1/httproute_types.go
bowei: Harry -- does it make sense to scale in the controller? will that work?
If it may be scaled, we should document.

apis/v1alpha1/httproute_types.go
bowei: Rob -- I think we should say filters need to be documented conformance "in the context" of where it is used. Therefore, there is a separate conformance level for:

pre backend selection vs post backend selection.

We should modify the docs in filter to reflect this distinction.

This is due to the reuse of types. We could make them different types but with the same content, not clear long-term if that is even a bad idea.

apis/v1alpha1/route_types.go
bowei: Is there a strong reason to align to 2^i? Are we sticking these in cache lines or something :^)

apis/v1alpha1/httproute_types.go
robscott: I think the key here is that this is calculated as `weight/(sum of all weights in this ForwardTo list)`. Not all/or many implementations will be able to provide this level of granularity, but they should be able to round this result into something the implementation can support. As far as documenting the scaling, are the current docs sufficient?

```
// Weight specifies the proportion of traffic forwarded to the backend
// referenced by the ServiceName or BackendRef field. This is computed as
// weight/(sum of all weights in this ForwardTo list).
```

On the other hand, I don't think this is a very important change, so if it ends up causing issues for some implementations I'm fine with pulling it.

apis/v1alpha1/httproute_types.go
robscott: I think it would add unnecessary overhead to define filter types twice. Maybe we should simply provide two conformance levels for each filter:
```
Support in HTTPRouteRule: Core
Support in HTTPRouteFilter: Extended
```

robscott: I like that idea, would this be an understandable way to communicate that? 
 
```
Support in HTTPRouteRule: Core
Support in HTTPRouteFilter: Extended
```

bowei: lgtm

apis/v1alpha1/route_types.go
robscott: https://github.com/kubernetes-sigs/service-apis/issues/440

apis/v1alpha1/httproute_types.go
bowei: Couple of things:

* We should state proportion of HTTP requests to be specific
* Implementations can implement this in many different ways, so there may be some \epsilon from the exact proportion (w/ exception of the weight==0 case). This also means that `weight` should not be expected to be passed through w/out change unless the semantics of the underlying controller matches.

apis/v1alpha1/route_types.go
robscott: Although it's an arbitrary limit, I think 100 might be easier for users to understand than 128. 64 does feel a bit too low here.

apis/v1alpha1/httproute_types.go
robscott: 👍 Updated spec to include that.

apis/v1alpha1/httproute_types.go
robscott: 👍 Updated spec to include that.

apis/v1alpha1/route_types.go
hbagdi: > Is there a strong reason to align to 2^i? Are we sticking these in cache lines or something :^)

There is no reason that I know of. I think it is only about consistency and having limits that users can easily remember and interpret.

apis/v1alpha1/httproute_types.go
```diff
@@ -392,20 +392,23 @@ const (
 	// HTTPRouteFilterRequestHeaderModifier can be used to add or remove an HTTP
 	// header from an HTTP request before it is sent to the upstream target.
 	//
-	// Support: Core
+	// Support in HTTPRouteRule: Core
+	// Support in HTTPRouteFilter: Extended
```
robscott, line 396: HTTPRouteForwardTo

apis/v1alpha1/route_types.go
robscott: I think powers of 2 makes a lot of sense for numbers below a certain threshold, maybe that's 100? At least to me, using 128 would feel more out of place/strange than 100 here.

apis/v1alpha1/httproute_types.go
robscott: Updated to 1 million - seemed like a less extreme upper limit but also low enough to ensure there will be no issues with a sum across multiple ForwardTo statements.

apis/v1alpha1/route_types.go
hbagdi: I don't feel strongly either ways.

hbagdi (approved):

apis/v1alpha1/httproute_types.go
```diff
@@ -392,20 +392,23 @@ const (
 	// HTTPRouteFilterRequestHeaderModifier can be used to add or remove an HTTP
 	// header from an HTTP request before it is sent to the upstream target.
 	//
-	// Support: Core
+	// Support in HTTPRouteRule: Core
+	// Support in HTTPRouteForwardTo: Extended
```
danehans, line 396: Why is this filter considered "Extended" in `HTTPRouteForwardTo` and not `HTTPRouteRule`?

apis/v1alpha1/httproute_types.go
```diff
@@ -534,27 +537,31 @@ type HTTPRouteForwardTo struct {
...
-	// weight/(sum of all weights in this ForwardTo list). Weight is not a
-	// percentage and the sum of weights does not need to equal 100. If only one
-	// backend is specified, 100% of the traffic is forwarded to that backend.
-	// If weight is set to 0, no traffic should be forwarded for this entry.
-	// If unspecified, weight defaults to 1.
+	// Weight specifies the proportion of HTTP requests forwarded to the backend
+	// referenced by the ServiceName or BackendRef field. This is computed as
+	// weight/(sum of all weights in this ForwardTo list). For non-zero values,
+	// there may be some epsilon from the exact proportion defined here
+	// depending on the precision an implementation supports. Weight is not a
```
danehans, line 544: Should we consider a range that is considered core support and a range that is considered extended or implementation-specific?

apis/v1alpha1/httproute_types.go
```diff
@@ -534,27 +537,31 @@ type HTTPRouteForwardTo struct {
...
+	// +kubebuilder:validation:Maximum=1000000
 	Weight int32 `json:"weight,omitempty"`
 
 	// Filters defined at this-level should be executed if and only if the
 	// request is being forwarded to the backend defined here.
 	//
-	// Conformance: Filtering support, including core filters, is NOT guaranteed
-	// at this-level. Use Filters in HTTPRouteRule for portable filters across
-	// implementations.
+	// Conformance: Support for filters defined at this level is NOT guaranteed.
```
danehans, line 562: I find it confusing that "Conformance" is defined for `Filters` but nowhere else in the APIs. How should the combination of "Conformance" and "Support" be determined. Should "Conformance" be removed and update "Support: Custom" as follows:
```
	// Support: Custom (For broader support of filters, use the Filters field in HTTPRouteRule)
```

apis/v1alpha1/route_types.go
```diff
@@ -114,19 +114,23 @@ type RouteForwardTo struct {
...
+	// referenced by the ServiceName or BackendRef field. This is computed as
+	// weight/(sum of all weights in this ForwardTo list). For non-zero values,
+	// there may be some epsilon from the exact proportion defined here
+	// depending on the precision an implementation supports. Weight is not a
+	// percentage and the sum of weights does not need to equal 100.
+	//
+	// If only one backend is specified and it has a weight greater than 0, 100%
+	// of the traffic is forwarded to that backend. If weight is set to 0, no
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
```
danehans, line 127: Same as https://github.com/kubernetes-sigs/service-apis/pull/436/files#r516161976

apis/v1alpha1/httproute_types.go
robscott: Yeah this is confusing. I chose "extended" based on our current definitions in the API Concepts doc:

```
* **CORE** features will be portable and we expect that there is a
  reasonable roadmap for ALL implementations towards support of APIs
  in this category.
* **EXTENDED** features are those that are portable but not
  universally supported across implementations. Those implementations
  that support the feature will have the same behavior and
  semantics. It is expected that some number of EXTENDED features will
  eventually migrate into the CORE. EXTENDED features will be part of
  the API types and schema.
* **CUSTOM** features are those that are not portable and are
  vendor-specific. CUSTOM features will not have API types and schema
  except via generic extension points.
```

In this case I think something like a request header modifier would be implemented the same way by all implementations that could support the feature on ForwardTo. That makes it portable and thus not custom. On the other hand, I don't think there is a reasonable roadmap for ALL implementations to support filters at this level, so it also can't be core. That leaves extended.

apis/v1alpha1/httproute_types.go
robscott: I think it's reasonable to assume that every implementation can do some basic calculations to transform a ratio into a weight that matches what is desired as closely as possible. We know that at least one implementation is limited to ~65k points of precision here. That could be a starting point, but what if another implementation was limited to 10k, or maybe just 100? I think the best we can do here is to say that this will be used to calculate a proportion that implementations can translate into weights. All implementations will be able to support all valid inputs, not all will be able to provide the same level of precision though. I don't know, open to other suggestions here.

apis/v1alpha1/route_types.go
robscott: Responded above https://github.com/kubernetes-sigs/service-apis/pull/436#discussion_r516211996.

apis/v1alpha1/httproute_types.go

PR #448, opened 2020-11-04
define HTTP route tiebreaker semantics

eth0xFEED:
possibly resolves https://github.com/kubernetes-sigs/service-apis/issues/337

- PathMatchType specific matching behavior
- Tie breaker on hostname path docs

@robscott  @hbagdi still some doubts here please let me know if we are close or missed here :) thanks !


apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,20 @@ type HTTPRouteRule struct {
 	// path match on "/", which has the effect of matching every
 	// HTTP request.
 	//
+	//
+	// Although a client request may technically match multiple HTTP route rules,
+	// only one rule may ultimately receive the request. Matching precedence
+	// MUST be determined in order of the following criteria:
+	//
+	// * The most specific match is determined by the longest matching combination
+	//   of hostname and path. In the event of a tie, the larger number of header
```
robscott, line 176: Nit: The second sentence could potentially be a second bullet point.

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,20 @@ type HTTPRouteRule struct {
...
+	//   matches will be used to differntiate.
+	// * The oldest Route based on creation timestamp. For example, a Route with
```
robscott, line 178: I'm not sure exactly how to structure this, but both the point above and below start with something saying something like "if still equivalent". Maybe all bullet points should start with that? I could also be fine with none as long as it was consistent.

apis/v1alpha1/httproute_types.go
```diff
@@ -203,6 +217,17 @@ type HTTPRouteRule struct {
...
+// For Prefix and Exact paths:
+// Every Path p_i must be syntactically valid:
+// 	- Must begin with the '/' character
+// 	- Must not contain consecutive '/' characters (e.g. /foo///, //).
+// 	- For prefix paths, a trailing '/' character in the Path is ignored,
+//      e.g. /abc and /abc/ specify the same match.
+//
+// If there is more than one potential match:
+// 	- Exact match is preferred to a Prefix match.
+//  - For multiple prefix matches, the longest Path p_i will be the matching path.
```
robscott, line 229: Nit: Indentation looks inconsistent here.

robscott (commented):
Thanks for your work on this! Mostly LGTM, just a few nits.

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,20 @@ type HTTPRouteRule struct {
...
 	// HTTP request.
 	//
+	//
+	// Although a client request may technically match multiple HTTP route rules,
+	// only one rule may ultimately receive the request. Matching precedence
+	// MUST be determined in order of the following criteria:
+	//
+	// * The most specific match is determined by the longest matching combination
+	//   of hostname and path. In the event of a tie, the larger number of header
+	//   matches will be used to differntiate.
```
Miciah, line 177:
```suggestion
	//   matches will be used to differentiate.
```

apis/v1alpha1/httproute_types.go
hbagdi: > larger number of header

What does this exactly mean?

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,20 @@ type HTTPRouteRule struct {
...
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
```
hbagdi, line 180: We discussed this before when we were working on conflict resolution semantics. It is not really possible to implement this in many implementations. It works for other resources and cases but not for matching inside a proxy.

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,20 @@ type HTTPRouteRule struct {
...
+	// * If everything else is equivalent, the Route appearing first in
+	//   alphabetical order (namespace/name) should be given precedence. For
+	//   example, foo/bar is given precedence over foo/baz.
```
hbagdi, line 183: Same comment as above.

apis/v1alpha1/httproute_types.go
```diff
@@ -203,6 +217,17 @@ type HTTPRouteRule struct {
...
+// Every Path p_i must be syntactically valid:
+// 	- Must begin with the '/' character
+// 	- Must not contain consecutive '/' characters (e.g. /foo///, //).
+// 	- For prefix paths, a trailing '/' character in the Path is ignored,
+//      e.g. /abc and /abc/ specify the same match.
+//
+// If there is more than one potential match:
+// 	- Exact match is preferred to a Prefix match.
+//  - For multiple prefix matches, the longest Path p_i will be the matching path.
+//
```
hbagdi, line 230: This can be made more specific.
See: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20190125-ingress-api-group.md#path-matching-semantics

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,17 @@ type HTTPRouteRule struct {
 	// path match on "/", which has the effect of matching every
 	// HTTP request.
 	//
+	//
+	// A client request may match multiple HTTP route rules,
+	// to differentiate we define matching precedence where matches
+	// MUST be determined in order of the following criteria, continuing on ties:
+	// * longest matching combination of hostname and path.
```
robscott, line 174: we can probably split this up so longest hostname match is considered first, then path.

apis/v1alpha1/httproute_types.go
eth0xFEED: i think ive moved into a better language form in the latest commit marking resolve please nit if this is in error :) !

apis/v1alpha1/httproute_types.go
eth0xFEED: resolved with discussion in office hours

apis/v1alpha1/httproute_types.go
```diff
@@ -203,6 +215,17 @@ type HTTPRouteRule struct {
 // * "RegularExpression"
 // * "ImplementationSpecific"
 //
+// For Prefix and Exact paths:
+// Every Path p_i must be syntactically valid:
```
robscott, line 219:
```suggestion
// Prefix and Exact paths must be syntactically valid:
```

apis/v1alpha1/httproute_types.go
```diff
@@ -203,6 +215,17 @@ type HTTPRouteRule struct {
...
+// 	- Must begin with the '/' character
+// 	- Must not contain consecutive '/' characters (e.g. /foo///, //).
+// 	- For prefix paths, a trailing '/' character in the Path is ignored,
+//      e.g. /abc and /abc/ specify the same match.
+//
+// If there is more than one potential match:
+//  - Exact match is preferred to a Prefix match.
+//  - For multiple prefix matches, the longest Path p_i will be the matching path.
```
robscott, line 227: I think this is already covered above. It would probably be best to remove this so we don't risk the two sections becoming out of sync.

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,18 @@ type HTTPRouteRule struct {
 	// path match on "/", which has the effect of matching every
 	// HTTP request.
 	//
+	//
+	// A client request may match multiple HTTP route rules,
+	// to differentiate we define matching precedence where matches
+	// MUST be determined in order of the following criteria, continuing on ties:
```
robscott, line 173:
```suggestion
	// A client request may match multiple HTTP route rules. Matching precedence
	// MUST be determined in order of the following criteria, continuing on ties:
```

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,18 @@ type HTTPRouteRule struct {
...
+	// * longest match hostname
+	// * longest match path
```
robscott, line 175:
```suggestion
	// * longest matching hostname
	// * longest matching path
```

apis/v1alpha1/httproute_types.go
```diff
@@ -167,6 +167,18 @@ type HTTPRouteRule struct {
...
+	// * largest number of header matches
```
robscott, line 176: The list items above all start with lowercase characters. The list items below all start with uppercase characters, specifically `The` and end with `.`. I think I prefer the style below, maybe update the above to match that?

robscott (commented):
Thanks for the work on this! Just a few small nits but mostly LGTM.

robscott:
Thanks!

/lgtm
/approve

PR #449, opened 2020-11-05
Various v1alpha1 API Clean-Up

### LINKED ISSUE #415: Add a way for TCPRoute, TLSRoute, and UDPRoute to match all requests
From the prealpha review: We have not clarified what happens when no matches are specified. It may make sense to use an enum here that supported the following options:

- All
- ExtensionRefs
- None

This could get confusing because we support multiple match structs here. Alternatively we may want to specify that if no matches are specified, all requests should match.

danehans:
- Renames `apis/v1alpha1/route_types.go` to `apis/v1alpha1/types.go` since the defined types are referenced by more than just `xRoute` types, i.e. `Gateway`. 
- Moves `Hostname` from `apis/v1alpha1/gateway_types.go` to `apis/v1alpha1/types.go`.
- Updates `xMatch` godocs  for the various API types.

/assign @robscott @hbagdi @bowei 
/cc @jpeach @Miciah 

Fixes #415

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -174,7 +174,9 @@ type HTTPRouteRule struct {
 	// Conformance-levels at this level are defined based on the type of filter:
 	// - ALL core filters MUST be supported by all implementations.
 	// - Implementers are encouraged to support extended filters.
-	// - Implementation-specific custom filters have no API guarantees across implementations.
+	// - Implementation-specific custom filters have no API guarantees across
+	//   implementations.
+	//
 	// Specifying a core filter multiple times has unspecified or custom conformance.
 	//
 	// Support: core
@@ -184,6 +186,10 @@ type HTTPRouteRule struct {
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// ForwardTo defines the backend(s) where matching requests should be sent.
+	// If unspecified, the rule performs no forwarding. If unspecified and no
+	// filters are specified that would result in a response being sent, a 503
+	// error code is returned.
+	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=4
 	ForwardTo []HTTPRouteForwardTo `json:"forwardTo,omitempty"`
@@ -304,12 +310,12 @@ type HTTPRouteMatch struct {
 	Headers *HTTPHeaderMatch `json:"headers"`
 
 	// ExtensionRef is an optional, implementation-specific extension to the
-	// "match" behavior.  For example, resource "myroutematcher" in group
-	// "networking.acme.io". If the referent cannot be found, the route must
-	// be dropped from the Gateway. The controller should raise the
-	// "ResolvedRefs" condition on the Gateway with the "DroppedRoutes"
-	// reason. The gateway status for this route should be updated with a
-	// condition that describes the error more specifically.
+	// "match" behavior. For example, resource "myroutematcher" in group
+	// "networking.acme.io". If the referent cannot be found, the rule is not
+	// included in the route. The controller should raise the "ResolvedRefs"
+	// condition on the Gateway with the "DegradedRoutes" reason. The gateway
+	// status for this route should be updated with a condition that describes
+	// the error more specifically.
 	//
 	// Support: custom
 	//
@@ -457,11 +463,10 @@ type HTTPRequestMirrorFilter struct {
 	// BackendRef and ServiceName are specified, ServiceName will be given
 	// precedence.
 	//
-	// If the referent cannot be found, the route must be dropped
-	// from the Gateway. The controller should raise the "ResolvedRefs"
-	// condition on the Gateway with the "DroppedRoutes" reason.
-	// The gateway status for this route should be updated with a
-	// condition that describes the error more specifically.
+	// If the referent cannot be found, the rule is not included in the route.
+	// The controller should raise the "ResolvedRefs" condition on the Gateway
+	// with the "DegradedRoutes" reason. The gateway status for this route should
+	// be updated with a condition that describes the error more specifically.
 	//
 	// Support: Core
 	//
@@ -473,11 +478,10 @@ type HTTPRequestMirrorFilter struct {
 	// both BackendRef and ServiceName are specified, ServiceName will be given
 	// precedence.
 	//
-	// If the referent cannot be found, the route must be dropped
-	// from the Gateway. The controller should raise the "ResolvedRefs"
-	// condition on the Gateway with the "DroppedRoutes" reason.
-	// The gateway status for this route should be updated with a
-	// condition that describes the error more specifically.
+	// If the referent cannot be found, the rule is not included in the route.
+	// The controller should raise the "ResolvedRefs" condition on the Gateway
+	// with the "DegradedRoutes" reason. The gateway status for this route should
+	// be updated with a condition that describes the error more specifically.
 	//
 	// Support: Custom
 	//
--- a/apis/v1alpha1/tcproute_types.go
+++ b/apis/v1alpha1/tcproute_types.go
@@ -36,9 +36,13 @@ type TCPRoute struct {
 // TCPRouteSpec defines the desired state of TCPRoute
 type TCPRouteSpec struct {
 	// Rules are a list of TCP matchers and actions.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
 	Rules []TCPRouteRule `json:"rules"`
 
 	// Gateways defines which Gateways can use this Route.
+	//
 	// +kubebuilder:default={allow: "SameNamespace"}
 	Gateways RouteGateways `json:"gateways,omitempty"`
 }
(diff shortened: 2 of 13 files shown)


apis/v1alpha1/tcproute_types.go
```diff
@@ -51,15 +54,18 @@ type TCPRouteStatus struct {
...
+	// If unspecified, all requests from the associated gateway TCP
+	// listener will match.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
-	// ForwardTo defines the backend(s) where matching requests should be sent.
+	// ForwardTo defines the backend(s) where matching requests should be
+	// sent. If unspecified, matching traffic will not be forwarded.
```
hbagdi, line 67: This is more of a general behavior question (not specific to TCP):
What should be the behavior in this case?
Should the request instead be sent to the forwarding target of the next match (in priority order)?
Or should the request be dropped i.e. 5xx or similar?

apis/v1alpha1/tcproute_types.go
```diff
@@ -69,16 +75,11 @@ type TCPRouteRule struct {
...
-	// Omitting or specifying the empty string for both the resource and
-	// group indicates that the resource is "configmaps".
-	//
-	// If the referent cannot be found, the route must be dropped
-	// from the Gateway. The controller should raise the "ResolvedRefs"
-	// condition on the Gateway with the "DroppedRoutes" reason.
-	// The gateway status for this route should be updated with a
+	// "match" behavior.  For example, resource "mytcproutematcher" in group
+	// "networking.acme.io". If the referent cannot be found, the route must
+	// be dropped from the Gateway. The controller should raise the
```
hbagdi, line 80: This could be a scope creep for the PR but I've a question:

If the referent can't be found, what should be dropped?
- The entire Route
- The entire Rule
- The specific Match

Dropping the entire route seem a little draconian and also violates a few of our [guiding principles](https://kubernetes-sigs.github.io/service-apis/concepts/#conflicts).

apis/v1alpha1/tlsroute_types.go
```diff
@@ -76,38 +78,36 @@ type TLSRouteRule struct {
...
-	// 2. If SNI is a wildcard, then the request matches this rule if
-	//    the SNI is to equal to the suffix
-	//    (removing the first label) of the wildcard rule.
+	// 1. If SNI is precise, the request matches this rule if the SNI in
+	//    ClientHello is equal to one of the defined SNIs.
+	// 2. If SNI is a wildcard, then the request matches this rule if the
+	//    SNI is to equal to the suffix (removing the first label) of the
+	//    wildcard rule.
+	// 3. If SNIs is unspecified, all requests associated to the gateway TLS
+	//    listener will match.
```
hbagdi, line 98:
```suggestion
	// 3. If SNIs is unspecified, all requests associated to the gateway TLS
	//    listener will match. This can be used to define a default backend for a TLS listener.
```

apis/v1alpha1/tcproute_types.go
danehans: @hbagdi this was copied from httpRoute [L306-L317](https://github.com/kubernetes-sigs/service-apis/blob/master/apis/v1alpha1/httproute_types.go#L306-L317) from https://github.com/kubernetes-sigs/service-apis/pull/408. I recall discussing this topic but I'm a foggy on the details. I thought we settled on dropping the route if rules can't be properly merged by the implementation. cc: @robscott @bowei

apis/v1alpha1/tcproute_types.go
danehans: The language I propose above is consistent with `HTTPRoute`. I can see a potential use case to expose a "honeypot" endpoint where the matched requests get dropped.

danehans:
Commit `1889dad` addresses the above feedback from @hbagdi.

apis/v1alpha1/httproute_types.go
```diff
@@ -174,16 +174,23 @@ type HTTPRouteRule struct {
 	// Conformance-levels at this level are defined based on the type of filter:
 	// - ALL core filters MUST be supported by all implementations.
 	// - Implementers are encouraged to support extended filters.
-	// - Implementation-specific custom filters have no API guarantees across implementations.
+	// - Implementation-specific custom filters have no API guarantees across
+	//   implementations.
+	//
 	// Specifying a core filter multiple times has unspecified or custom conformance.
 	//
+	// If unspecified, no filters will be applied to matched requests.
```
robscott, line 182: I'm not sure this is necessary. It may also prevent future creative implementations where controllers might choose to apply a default set of filters for a given GatewayClass or something? No idea how useful that would be, but I think it would be safer to remove this instead of accidentally preventing that use case.

apis/v1alpha1/httproute_types.go
```diff
@@ -174,16 +174,23 @@ type HTTPRouteRule struct {
...
+	//
 	// Support: core
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// ForwardTo defines the backend(s) where matching requests should be sent.
+	// If unspecified, the rule performs no forwarding. If unspecified and filters
+	// is unspecified, a 503 error code is returned.
```
robscott, line 192: I think this needs to be clarified. We'd still return a 503 if filters were specified that didn't actually result in a response being sent.

apis/v1alpha1/tcproute_types.go
```diff
@@ -51,15 +54,19 @@ type TCPRouteStatus struct {
...
+	// listener will match.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
-	// ForwardTo defines the backend(s) where matching requests should be sent.
+	// ForwardTo defines the backend(s) where matching requests should be
+	// sent. If unspecified, matching requests will be dropped and a 503
+	// error code returned.
```
robscott, line 68: Maybe ForwardTo should be required on these routes? Admittedly doesn't make sense where we have filters, but I'm not sure if we'd ever add something like that here. Not sure on this one...

apis/v1alpha1/tcproute_types.go
robscott: It seems like removing the match could be sufficient here. I'd hate for someone to add a new filter to a match in a large route and have the whole thing be dropped.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -76,38 +79,37 @@ type TLSRouteRule struct {
...
-	//    the SNI in ClientHello is equal to one of the defined SNIs.
-	// 2. If SNI is a wildcard, then the request matches this rule if
-	//    the SNI is to equal to the suffix
-	//    (removing the first label) of the wildcard rule.
+	// 1. If SNI is precise, the request matches this rule if the SNI in
+	//    ClientHello is equal to one of the defined SNIs.
+	// 2. If SNI is a wildcard, then the request matches this rule if the
+	//    SNI is to equal to the suffix (removing the first label) of the
+	//    wildcard rule.
+	// 3. If SNIs is unspecified, all requests associated to the gateway TLS
```
robscott, line 98: Nit: I think I prefer "with" instead of "to" here.

robscott (commented):
Thanks for the work on this! Lots of great improvements here. One suggestion - I think I'd prefer `shared_types.go` as the filename instead of just `types.go`.

apis/v1alpha1/tcproute_types.go
danehans: I'll update all route types to indicate that an unresolved ref will cause the matching rule to not be included in the route and that a status condition should be surfaced.

apis/v1alpha1/tcproute_types.go
danehans: I'll make the `ForwardTo` required for all route types other than `HTTPRoute`. We can readdress post v1alpha1.

apis/v1alpha1/tcproute_types.go
hbagdi: Un-resolving this thread. I still see the following in the diff:
```
	// "networking.acme.io". If the referent cannot be found, the route must
	// be dropped from the Gateway. The controller should raise the
```

danehans:
@robscott commit `5dc703a` updates the PR based on your comments, PTAL.

apis/v1alpha1/httproute_types.go
```diff
@@ -184,6 +186,11 @@ type HTTPRouteRule struct {
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// ForwardTo defines the backend(s) where matching requests should be sent.
+	// If unspecified, the rule performs no forwarding. If unspecified and filters
+	// is unspecified, a 503 error code is returned. If unspecified and filters
+	// are specified that result in a response not being sent, a 503 error code
+	// is returned.
```
robscott, line 192: Nit: Maybe the last two sentences could be combined here, something like this:

> If unspecified and no filters are specified that would result in a response being sent, a 503 error code is returned.

apis/v1alpha1/tcproute_types.go
```diff
@@ -36,9 +36,12 @@ type TCPRoute struct {
 // TCPRouteSpec defines the desired state of TCPRoute
 type TCPRouteSpec struct {
 	// Rules are a list of TCP matchers and actions.
+	//
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 40: Should we have `MinItems=1` here as well? What good is a TCPRoute without any rules? I understand matches being optional, but aren't rules the only way to forward traffic anywhere?

robscott (commented):
Thanks for you work on this!

danehans:
Commit `7854137` fixes merge conflicts and commit `7c194ea` addresses feedback from @robscott.

robscott:
Thanks!

/approve

Issue #333, opened 2020-09-17
Add examples to documentation


robscott:
At a minimum, these examples should cover:

- [x] a simple GatewayClass, Gateway, HTTPRoute*
- [x] a Gateway that targets routes in multiple namespaces*
- [x] an example of TLS configuration
- [ ] working with other route types like TCPRoute or TLSRoute

We should try to embed existing examples from the repo instead of creating new ones.

`*` = must haves for v1alpha1 release

hbagdi:
#427 accomplishes part of this.

robscott:
#427 also potentially accomplishes 'a Gateway that targets routes in multiple namespaces' as well

hbagdi:
> #427 also potentially accomplishes 'a Gateway that targets routes in multiple namespaces' as well

@robscott Do we need a separate guide from $427 for the multiple namespace point or can that box be checked off as well?

robscott:
I'm on the fence. I think it's probably good enough to no longer block alpha, but we could probably use an example that actually showed different routes in different namespaces being targeted by the same gateway.

hbagdi:
I was under the same confusion so errored on the side of having a new guide #452. PTAL.

robscott:
I've dropped this down to priority/important-soon now that #452 has merged and all of our must haves for v1alpha1 are in.

PR #452, opened 2020-11-10
write a guide for routes in multiple namespaces

hbagdi:
A guide which illustrates how routes from multiple namespaces are targeted by a single Gateway resource.

See #333

### DIFF
--- a/docs-src/multiple-ns.md
+++ b/docs-src/multiple-ns.md
@@ -1,0 +1,37 @@
+# Routes in multiple namespaces
+
+With the Service APIs, a single Gateway can target routes across multiple
+namespaces. 
+
+This guide assumes that you have installed Service APIs CRDs and a conformant
+controller.
+
+In the following example:
+
+- `acme-lb` GatewayClass: The GatewayClass responsible for satisfying Gateway
+  and Route resources.
+- `multi-ns-gateaway` Gateway: The Gateway is configured with a single listener
+  on port 80 which selects routes from **all namespaces** which have the label
+  `product: baz`. Notice how the `routes.namespaces.from` field in the listener
+  is set to `All`.
+- `service-apis-example-ns1` and `service-apis-example-ns2` Namespaces: These
+  are two namespaces in which we route resources are installed.
+- `http-app-1` and `http-app-2` HTTPRoutes: These are two resources that are
+  installed in separate namespaces. These routes will be picked up by
+  `multi-ns-gateway` resource for the following reasons:
+    - Both have the `product: baz` label on them.
+    - `http-app-1` HTTPRoute has `spec.gateways.allow` set to `All`.  The route
+      owner here has opted to allow **all** Gateways in the cluster to bind to
+      this Route.
+    - `http-app-2` HTTPRoute has `spec.gateways.allow` set to `FromList` and
+      contains a reference to the `multi-ns-gateway` in `default` namespace. In
+      this case, no other Gateway can bind to this route.  explictly set this
+      field so that Gateways from other namespaces can target the route.
+
+```
+{% include 'routes-in-multiple-namespaces.yaml' %}
+```
+
+Please note that this guide illustrates this feature for HTTPRoute resource
+only as an example. The same can be accomplished with other route types as
+well.
--- a/docs-src/multiple-ns.md
+++ b/docs-src/multiple-ns.md
@@ -11,22 +11,23 @@ In the following example:
 - `acme-lb` GatewayClass: The GatewayClass responsible for satisfying Gateway
   and Route resources.
 - `multi-ns-gateaway` Gateway: The Gateway is configured with a single listener
-  on port 80 which selects routes from **all namespaces** which have the label
-  `product: baz`. Notice how the `routes.namespaces.from` field in the listener
-  is set to `All`.
+  on port 80 which selects routes that have the label `product: baz` in any
+  namespace.  Notice how the `routes.namespaces.from` field in the listener is
+  set to `All`.
 - `service-apis-example-ns1` and `service-apis-example-ns2` Namespaces: These
-  are two namespaces in which we route resources are installed.
+  are the namespaces in which route resources are instantiated.
 - `http-app-1` and `http-app-2` HTTPRoutes: These are two resources that are
-  installed in separate namespaces. These routes will be picked up by
-  `multi-ns-gateway` resource for the following reasons:
+  installed in separate namespaces. These routes will be bound to Gateway
+  `multi-ns-gateway` for the following reasons:
     - Both have the `product: baz` label on them.
     - `http-app-1` HTTPRoute has `spec.gateways.allow` set to `All`.  The route
-      owner here has opted to allow **all** Gateways in the cluster to bind to
-      this Route.
+      owner has opted to allow **all** Gateways in the cluster to bind to this
+      Route.
     - `http-app-2` HTTPRoute has `spec.gateways.allow` set to `FromList` and
-      contains a reference to the `multi-ns-gateway` in `default` namespace. In
-      this case, no other Gateway can bind to this route.  explictly set this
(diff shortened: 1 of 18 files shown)


examples/basic-http.yaml
```diff
@@ -25,7 +24,7 @@ spec:
         matchLabels:
           app: foo
       namespaces:
-        from: "All"
+        from: "SameNamespace"
```
hbagdi, line 27: This was simplified because the route and gateway are being installed in the same namespace in this example.

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,32 @@
...
+
+This guide assumes that you have installed Service APIs CRDs and a conformant
+controller.
+
+In the following example:
+
+- `acme-lb` GatewayClass: The GatewayClass responsible for satisfying Gateway
+  and Route resources.
+- `multi-ns-gateaway` Gateway: The Gateway is configured with a single listener
+  on port 80 which selects routes from **All namespaces** which have the label
```
robscott, line 14: Nit: "All" -> "all"

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,32 @@
...
+  is set to `All`.
+- `foo-ns` and `bar-ns` Namespaces: These are two namespaces in which we route
+  resources are installed.
+- `http-app-1` and `http-app-2` HTTPRoutes: These are two resources that are
+  installed in separate namespaces. These routes will be picked up by
+  `multi-ns-gateway` resource for two reasons:
+    - Both have the `product: baz` label on them.
+    - Both have `spec.gateways.allow` set to `All`. The route owner needs to
+      explictly set this field so that Gateways from other namespaces can target
+      the route.
```
robscott, line 25: Can we use this example to show both approaches? One with `allow: All` and one with `allow: FromList`. Unfortunately that would also require creating a namespace for the Gateway.

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,32 @@
...
+
+In the following example:
+
+- `acme-lb` GatewayClass: The GatewayClass responsible for satisfying Gateway
+  and Route resources.
+- `multi-ns-gateaway` Gateway: The Gateway is configured with a single listener
+  on port 80 which selects routes from **All namespaces** which have the label
+  `product: baz`. Notice how the `routes.namespaces.from` field in the listener
+  is set to `All`.
+- `foo-ns` and `bar-ns` Namespaces: These are two namespaces in which we route
```
robscott, line 17: I'm not exactly sure where this belongs, but it would probably be good to call attention to the fact that this example creates namespaces and explicitly installs resources in them. If a user is trying to keep all their examples in a single namespace, that will not work here. Similarly, if they happen to be using namespaces with these names that could be problematic. Especially if they happen to cleanup demo resources with `kubectl delete -f`. 

Maybe to be extra safe resources like namespaces should have a clear association with this project? Something like `service-apis-example-ns1`, `service-apis-example-ns2` is overly verbose but also unlikely to conflict with any existing namespaces. Maybe there's a less verbose option that has a similar effect of clearly being tied to an example from this project?

robscott (commented):
Thanks for your work on this! One small note - this does not seem to have a reference in the Guides intro like the TLS one, I may have just missed it though.

docs-src/multiple-ns.md
hbagdi: Opted to change the name to not conflict with anything that already exists.
I'm not explicitly calling it out because it seems self-explanatory. Is that okay?

docs-src/multiple-ns.md
hbagdi: > Unfortunately that would also require creating a namespace for the Gateway.

I'll put the Gateway in the default namespace.

hbagdi:
> this does not seem to have a reference in the Guides intro like the TLS one

Good catch. Added.

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+- `http-app-1` and `http-app-2` HTTPRoutes: These are two resources that are
+  installed in separate namespaces. These routes will be picked up by
+  `multi-ns-gateway` resource for the following reasons:
+    - Both have the `product: baz` label on them.
+    - `http-app-1` HTTPRoute has `spec.gateways.allow` set to `All`.  The route
+      owner here has opted to allow **all** Gateways in the cluster to bind to
+      this Route.
+    - `http-app-2` HTTPRoute has `spec.gateways.allow` set to `FromList` and
+      contains a reference to the `multi-ns-gateway` in `default` namespace. In
+      this case, no other Gateway can bind to this route.  explictly set this
```
robscott, line 28:
```suggestion
      this case, only Gateways in the `default` namespace can bind to this route.
```

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+      field so that Gateways from other namespaces can target the route.
```
robscott, line 29: Maybe something like "Additional namespaces may be added to this list to allow Gateways in other namespaces to reference this namespace."

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+This guide assumes that you have installed Service APIs CRDs and a conformant
+controller.
+
+In the following example:
+
+- `acme-lb` GatewayClass: The GatewayClass responsible for satisfying Gateway
+  and Route resources.
+- `multi-ns-gateaway` Gateway: The Gateway is configured with a single listener
+  on port 80 which selects routes from **all namespaces** which have the label
+  `product: baz`. Notice how the `routes.namespaces.from` field in the listener
```
robscott, line 15: This is a bit confusing - it almost sounds like the namespaces need to have those labels, not the routes. 

```suggestion
  on port 80 which selects routes that have the label `product: baz` in any namespace. 
  Notice how the `routes.namespaces.from` field in the listener
```

robscott (commented):
Thanks for your work on this!

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+  is set to `All`.
+- `service-apis-example-ns1` and `service-apis-example-ns2` Namespaces: These
+  are two namespaces in which we route resources are installed.
```
danehans, line 18: s/These are two namespaces in which we route resources are installed/These are the namespaces in which route resources are instantiated.

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+- `http-app-1` and `http-app-2` HTTPRoutes: These are two resources that are
+  installed in separate namespaces. These routes will be picked up by
+  `multi-ns-gateway` resource for the following reasons:
```
danehans, line 21: s/These routes will be picked up by `multi-ns-gateway` resource for the following reasons/These routes will be bound to Gateway `multi-ns-gateway` for the following reasons/

docs-src/multiple-ns.md
```diff
@@ -0,0 +1,37 @@
...
+    - Both have the `product: baz` label on them.
+    - `http-app-1` HTTPRoute has `spec.gateways.allow` set to `All`.  The route
+      owner here has opted to allow **all** Gateways in the cluster to bind to
+      this Route.
```
danehans, line 25: The route owner ~~here~~ has opted to allow **all** Gateways in the cluster to bind to this Route.

danehans:
Great doc contribution @hbagdi! I have a few nits, otherwise LGTM.

docs-src/multiple-ns.md
hbagdi: This was worded completely wrong. Updated to rephrase and include your suggestion.

hbagdi:
@robscott @danehans I've incorporated your feedback. PTAL.

robscott:
Thanks!

/lgtm
/approve

PR #457, opened 2020-11-12
clarify HTTP Header name and value matching criteria

### LINKED ISSUE #412: Clarify what case-insensitive means for header matching
From prealpha review:

>What does this mean in the "RegularExpression" context? I can see two interpretations here:
> 1. that "^hOsT$" and "^HoSt$" will both match "Host"
> 2. that the gateway e.g. lowercases the header name, and expects the regexp to match against a lowercased text

hbagdi:
See #412

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -214,8 +214,8 @@ const (
 	PathMatchImplementationSpecific PathMatchType = "ImplementationSpecific"
 )
 
-// HeaderMatchType specifies the semantics of how HTTP headers should be compared.
-// Valid HeaderMatchType values are:
+// HeaderMatchType specifies the semantics of how HTTP header values should be
+// compared. Valid HeaderMatchType values are:
 //
 // * "Exact"
 // * "RegularExpression"
@@ -226,9 +226,6 @@ type HeaderMatchType string
 
 // HeaderMatchType constants.
 const (
-	// HeaderMatchTypeExact matches HTTP request-header fields.
-	// Field name matches are case-insensitive while field value matches
-	// are case-sensitive.
 	HeaderMatchExact                  HeaderMatchType = "Exact"
 	HeaderMatchRegularExpression      HeaderMatchType = "RegularExpression"
 	HeaderMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
@@ -255,7 +252,8 @@ type HTTPPathMatch struct {
 	Value string `json:"value"`
 }
 
-// HTTPHeaderMatch describes how to select a HTTP route by matching HTTP request headers.
+// HTTPHeaderMatch describes how to select a HTTP route by matching HTTP request
+// headers.
 type HTTPHeaderMatch struct {
 	// Type specifies how to match against the value of the header.
 	//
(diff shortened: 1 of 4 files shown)


robscott:
Fixes #412

robscott (commented):
Thanks for the work on this! As I'm reading through this, I think lines 230-231 are a bit confusing and out of place - maybe we can add documentation covering field value case sensitivity for each match type there instead.

apis/v1alpha1/httproute_types.go
```diff
@@ -274,8 +274,9 @@ type HTTPHeaderMatch struct {
 	// It MUST contain at least one entry.
 	//
 	// The HTTP header field name to match is the map key, and the
-	// value of the HTTP header is the map value. HTTP header field
-	// names MUST be matched case-insensitively.
+	// value of the HTTP header is the map value. HTTP header field names MUST
+	// be always matched case-insensitively, no matter which HeaderMatchType is
+	// used.
```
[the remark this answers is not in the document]
hbagdi: I tried to rework this PR a little bit to include the following bits:
- Names are always case-insensitive
- Values are always case-sensitive
- HeaderMatchType applies to values only 

PTAL

apis/v1alpha1/httproute_types.go
```diff
@@ -267,15 +264,18 @@ type HTTPHeaderMatch struct {
...
 	Type HeaderMatchType `json:"type,omitempty"`
 
 	// Values is a map of HTTP Headers to be matched.
 	// It MUST contain at least one entry.
 	//
 	// The HTTP header field name to match is the map key, and the
-	// value of the HTTP header is the map value. HTTP header field
-	// names MUST be matched case-insensitively.
+	// value of the HTTP header is the map value. HTTP header field names MUST
+	// be always matched case-insensitively.
```
robscott, line 278: Nit: I think `always` is unnecessary here.

apis/v1alpha1/httproute_types.go
```diff
@@ -267,15 +264,18 @@ type HTTPHeaderMatch struct {
 	// Please read the implementation's documentation to determine the supported
 	// dialect.
 	//
+	// HTTP Header names MUST be always matched case-insensitively (RFC 2616 -
```
robscott, line 267: "case-insensitively" sounds a bit off to me. Maybe something like this would work: "HTTP Header name matching MUST be case-insensitive"

apis/v1alpha1/httproute_types.go
robscott: Similar comment as above about some kind of alternative wording for `case-insensitively`.

apis/v1alpha1/httproute_types.go
```diff
@@ -267,15 +265,17 @@ type HTTPHeaderMatch struct {
...
 	// +kubebuilder:default=Exact
 	Type HeaderMatchType `json:"type,omitempty"`
 
 	// Values is a map of HTTP Headers to be matched.
 	// It MUST contain at least one entry.
 	//
 	// The HTTP header field name to match is the map key, and the
-	// value of the HTTP header is the map value. HTTP header field
-	// names MUST be matched case-insensitively.
+	// value of the HTTP header is the map value. HTTP header field names matching
```
robscott, line 277: This is _super_ nitpicky, but I think it should be `name` instead of `names` here.

robscott:
Thanks!

/lgtm
/approve

Issue #434, opened 2020-10-29
ListenerStatus should be more clearly tied to Listeners


robscott:
From the prealpha feedback:

> Given that TCP and UDP port spaces are independent, and it seems that most (if not all) load balancing solutions distinguish the two visually everywhere, in the form similar to "80/tcp", "53/udp":

> Is there a reason to meld port spaces of various families (TCP, UDP) into one in ListenerStatus?

Some possible solutions:
- Add a name field to listeners to uniquely identify them, use that in ListenerStatus
- Add protocol (and maybe hostname) to ListenerStatus

/cc @jpeach

PR #458, opened 2020-11-12
add protocol and optional hostname to listener status

hbagdi:
Listeners can be uniquely identified with tuple of port, protocol and an
optional hostname.
This patch adds protocol and hostname to ListenerStatus to tie each
ListenerStatus with a specific Listener on the Gateway.

See #434


robscott (commented):
Thanks!

/approve

hbagdi:
I've removed the part which allowed implementations to skip Hostname in ListenerStatus.

robscott:
Thanks!

/lgtm

PR #459, opened 2020-11-12
docs: split out general API guidelines from concepts

hbagdi:
This is another patch in the continued work to improve our documentation experience.


mkdocs.yml
```diff
@@ -14,6 +14,7 @@ nav:
       API concepts: concepts.md
       API resources: resources.md
       Security Model: security-model.md
+      API guidelines: guidelines.md
```
robscott, line 17: Nit: Your page title is "Implementation Guidelines". I think that may make the most sense here since this part of the docs is primarily targeted towards implementers.

hbagdi: I had a split mind when writing this and it showed up in the patch as well.
I've changed this to `Implementation Guidelines`.

robscott:
Thanks!

/lgtm
/approve

fyi @mark-church, this was the split @hbagdi was talking about. I think it makes sense to merge this in first and then iterate with any tweaks/additions/reorgs you suggest after.

PR #461, opened 2020-11-12
added a simple gateway guide

mark-church:
This adds the simplest possible self-contained example. The purpose of this example is to demonstrate the fastest path to doing something useful. It is a Gateway with all the default settings with a single HTTPRoute. The HTTPRoute matches all traffic and sends it to a single foo-svc. I debated whether or not to include GatewayClass in this example but opted to not, and assume that it already exists so we can show an even simpler developer workflow (without having to worry about the GatewayClass).

I'm open to changing the name of resources so we can be consistent (like the GatewayClass we use in our examples). The image source for this is at https://docs.google.com/presentation/d/1neBkFDTZ__vRoDXIWvAcxk2Pb7-evdBT6ykw_frf9QQ/edit?usp=sharing


mark-church:
Also just realized that the API spec is probably outdated so I need to fix that ...

danehans:
@mark-church are you able to fix the CI error in time for the v1alpha1 release?

docs-src/simple-gateway.md
```diff
@@ -0,0 +1,40 @@
...
+      selector:
+        matchLabels:
+          gateway: prod-web
+```
+
+The Gateway represents the instantation of a logical load balancer. It's templated from a hypothetical `l7-gw-class` GatewayClass. The Gateway listens for HTTP traffic on port 80. This particular GatewayClass automatically assigns an IP address which will be shown in the Gateway.status after it has been deployed. 
+
+To match traffic that is sent to the Gateway listener an HTTPRoute can be bound to the Gateway. This is done using label selection from the Gateway to labels on the HTTPRoute.  The following HTTPRoute resource defines the mapping of a Gateway to a backend Service that traffic routes to. Because there are no host routes or paths specified, this HTTPRoute will match all traffic that arrives at port 80 of the load balancer and send it to the `foo-svc` Pods.
+
+```yaml
```
robscott, line 28: I think this and the above example should ideally come from the examples folder so we can automatically validate them and ensure they don't become stale. Some of the other guides like [TLS](https://github.com/kubernetes-sigs/service-apis/blob/master/docs-src/tls.md) show this pattern being used.

docs-src/simple-gateway.md
```diff
@@ -0,0 +1,40 @@
+# Deploying a simple Gateway
+
+The simplest possible deployment is a Gateway and Route resources which are deployed together and have the same owner. This represents the same kind of model used for Ingress. A Gateway is deployed and an HTTPRoute is deployed which matches all traffic and sends it to a single Service named foo-svc.
```
robscott, line 3: A small nit - we usually try to wrap markdown files with 80 char line lengths where possible. There are a lot of tools that do this automatically for you, I use the [rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) for VSCode.

mkdocs.yml
```diff
@@ -21,6 +21,7 @@ nav:
   - Guides:
       - Introduction: guides.md
       - Getting started: getting-started.md
+      - Simple Gateway: simple-gateway.md
```
robscott, line 24: In addition to adding a reference here, it would be great to add this to the Guides page: https://github.com/kubernetes-sigs/service-apis/blob/master/docs-src/guides.md

robscott (commented):
Thanks for your work on this - this will be a great addition!

docs-src/simple-gateway.md
```diff
@@ -0,0 +1,40 @@
...
+
+The simplest possible deployment is a Gateway and Route resources which are deployed together and have the same owner. This represents the same kind of model used for Ingress. A Gateway is deployed and an HTTPRoute is deployed which matches all traffic and sends it to a single Service named foo-svc. 
+
+![Simple Gateway](single-service-gateway.png)
+
+```yaml
+kind: Gateway
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: prod-web
```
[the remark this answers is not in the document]
mark-church: Changing this to `prod-web-gw` so it is clear that the name and label can be different and are not connected. However, I think it is a valuable precedent to set that it helps users understand better which gateway they are using if the match label and gateway name are similar.

robscott:
Thanks!

/lgtm
/approve

PR #462, opened 2020-11-12
added an example for HTTPRoute traffic matching

mark-church:
Adds a self-contained example for deploying different types of traffic matching. This illustrates host, path, header, and default matching within a single HTTPRoute resource.


mark-church:
Also just realized that the API spec is probably outdated so I need to fix that ...

mark-church:
This one probably won't get finished in the next day so I would not block on it

mark-church:
This one is ready for review. It's goal is to introduce two concepts:

- The fact that a Gateway can bind to multiple Routes and that those Routes will merge together
- To showcase the core traffic matching functionality of HTTPRoute for host, path, and header matches

This is the scenario for the guide:
image

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
+# HTTP routing
+
+HTTPRoute resources allow you to match on HTTP traffic and direct it Kubernetes
```
robscott, line 3:
```suggestion
HTTPRoute resources allow you to match on HTTP traffic and direct it to Kubernetes
```

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+backends. This guide shows how the HTTPRoute matches on host, header, and path
+fields and forwards traffic to different Kubernetes Services.
+
+The following diagram describes a required traffic flow across three different
+Services:
+
+- Traffic to `foo.example.com/login` is forwarded to `foo-svc` 
+- Traffic to `bar.example.com/*` with a `Env: canary` header is forwarded
```
robscott, line 11: Nit for both this and the corresponding graphic + yaml, I think I'd prefer the lowercase `env` here.

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+
+The dotted lines show the Gateway resources deployed to configure this routing
+behavior. There are two HTTPRoute resources that create routing rules on the
+same `prod-web` Gateway. This illustrates how [more than one Route can bind to a
+Gateway](api-overview.md/#gateway) which allows Routes to merge on a Gateway as
+long as they don't conflict. 
+
+The following `prod-web` Gateway is defined from the `acme-lb` GatewayClass.
+`prod-web` listens for HTTP traffic on port 80 and will bind to all Routes in
+the `default` Namespace that have the matching labels  `gateway: prod-web-gw`.
```
robscott, line 26:
```suggestion
the `default` Namespace that have the matching `gateway: prod-web-gw` label.
```

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+Route labels allow the Route to determine which Gateways recieve their routing
```
robscott, line 27: Although technically true, this feels a bit backwards to me. Gateways select Routes with specific labels, Route owners can explicitly allow from specific Gateways using the `gateways` field on a route. I think there's a few different ways users could choose to make this connection, and we probably need a doc just for this discussion. For example:

1. Routes are labeled with `app` and `env` labels, Gateway selects based on app and/or env it is intended to serve traffic for.
2. Gateways select Routes with a specific label like `prod-web-gw`, Routes add label to indicate the Gateway they'd like to be served by.
3. Gateways select all routes in the namespace, Routes only allow selection from a specific Gateway or set of Gateways.

I don't really like 3, but it is technically possible. I think I prefer 1 over 2 as it more closely matches the intent of the API, but I can see the use case for 2. So with all that said, I'm not quite sure how to phrase this sentence, I'm just hesitant to frame it so concretely since this is one of several possible models for connecting Gateways and Routes.

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+rules.  
+
+```yaml 
+{% include 'http-routing/gateway.yaml' %}  
+```
+
+An HTTPRoute can match against a [single set of
+hostname](https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteSpec).
```
robscott, line 35:
```suggestion
hostnames](https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteSpec).
```

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+These hostnames are matched before any other matching within the HTTPRoute takes
+place. Since `foo.example.com` and `bar.example.com` are separate hosts with
+different routing requirements, each is deployed as its own HTTPRoute -
+`foo-route` and `bar-route`.
+
+The following `foo-route` will match any traffic for `foo.example.com` and apply
+its routing rules to forward the traffic to the correct backend. Since there is
+only one match specified, only `foo.example.com/login*` traffic will be
```
robscott, line 43: It looks like we have left the matching rather ambiguous here, but in the Ingress spec, a prefix match for `login` would match requests to `/login` or `/login/foo` but not `/loginfoo`. I think that's likely worth carrying over to this spec, or at least assuming for now.

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml 
+{% include 'http-routing/foo-httproute.yaml' %}  
+```
+
+Similarly, the `bar-route` HTTPRoute matches traffic for `bar.example.com`. All
+traffic for this hostname will be evaluated against the routing rules. The most
+specific match will take precedence which means that any traffic with the `Env:
+canary` header will be forwarded to `bar-svc-canary` and if the header is
+missing then it'll be forwarded to `bar-svc`.
```
robscott, line 55:
```suggestion
missing or not set to `canary` then it'll be forwarded to `bar-svc`.
```

docs-src/http-routing.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml 
+{% include 'http-routing/bar-httproute.yaml' %}  
+```
+
+Any traffic that does not correspond to any of the hostnames, or does not match
+the routing rules within an HTTPRoute will be forwarded to the default backend
+for this Gateway.
```
robscott, line 63: We don't have a default backend for a Gateway in the spec right now. I think it might be simplest to leave this sentence out for now.

robscott (commented):
Thanks for your work on this!

docs-src/http-routing.md
mark-church: Since matching Routes with Gateways (or Gateways with Routes) has so much flexibility I agree we need a section (probably in api-overview) that talks about conventions for connecting the two. 

Personally, I find it really strange not to be more explicit around which gateway is being used, but this is another thing we need to get feedback from users from on how they think about connecting Gateways and Routes. I'll tone down this sentence to make it a little less suggestive about which is the right convention.

docs-src/http-routing.md
mark-church: changed this to /login/*

docs-src/http-routing.md
mark-church: removed this sentence and also from the graphic

mark-church:
Fixed issued brought up in comments and new image that leaves out the default route:
image

examples/http-routing/gateway.yaml
```diff
@@ -0,0 +1,15 @@
+kind: Gateway
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: prod-web
+  namespace: default
```
robscott, line 5: One tiny nit - I think we've traditionally avoided hardcoding a namespace in our examples. This makes it easier for users to deploy examples into specific test/demo namespaces if they want.

robscott (commented):
Thanks for the work on this! Mostly LGTM, just had one tiny nit.

mark-church:
Removed namespace references from YAMLs and also updated guides.md which was missing the http-routing.md page. Should hopefully be good to go now.

robscott:
Thanks!

/lgtm
/approve

PR #463, opened 2020-11-13
Reorganizes Docs and Adds HTTPRoute Resource Doc

danehans:
- `apis/v1alpha1/gateway_types.go` and `apis/v1alpha1/httproute_types.go`: Fixes minor godoc errors.
- `docs-src/httproute.md`: Adds docs for detailing the `HTTPRoute` API type.
- All other changes are related to reorganizing documentation and content generation.

/assign @robscott @bowei @hbagdi 

xref: https://github.com/kubernetes-sigs/service-apis/issues/332


docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+Possible values for `allow` are:
+
+- `All`: Gateways in any namespace can use this route.
+- `FromList`: Only Gateways specified in `gatewayRefs` may use this route.
+- `SameNamespace` (default): Only Gateways in the same namespace may use this route.
+	
+If `allow` results in preventing the selection of an HTTPRoute by a Gateway, an “Admitted: false” condition must be set
+on the Gateway for this Route.
```
robscott, line 43: Nit: Instead of "must" maybe we should use "will". This kind of doc seems like it should be more targeted towards users than implementers.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+spec:
+  hostnames:
+    - my.filter.com
+  rules:
+    ...
+      filters:
+      - type: RequestHeaderModifier
+        requestHeaderModifier:
+          add:
+            my-header: foo
```
robscott, line 172: Tangential, but seeing this example makes me wonder if we should have split this into 2 filter types - one to add headers and one to remove headers just for the sake of slightly more concise config.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+    - my.filter.com
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /
+      forwardTo:
+      - serviceName: my-filter-svc1
+        port: 80
+```
```
robscott, line 208: I think any complete example like this should probably come from the examples dir so we can ensure it doesn't become stale. Alternatively you could choose to highlight just the small part of this config relevant to `forwardTo`.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+      - path:
+          type: Prefix
+          value: /
+      forwardTo:
+      - port: 8080
+        backendRef:
+          name: my-custom-resource
+          group: acme.io
+          kind: CustomBackend
+```
```
robscott, line 231: Same comment about potentially stale examples.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+          type: Prefix
+          value: /
+      forwardTo:
+      - serviceName: my-filter-svc1
+        port: 80
+        weight: 50
+      - serviceName: my-filter-svc2
+        port: 80
+        weight: 50
+```
```
robscott, line 255: Also same comment as above. One additional note - it may be preferable to use weights that don't some to 100 to avoid weight being interpreted as a percentage.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+
+Reference the [forwardTo][forwardto] API documentation for additional details of `weight` and other fields.
+
+## Status
+
+Status defines the observed state of HTTPRoute.
+
+### RouteStatus
+
+RouteStatus defines the observed state that is required across all route types.
```
robscott, line 265: I'm not sure how much value this has. We already have the API spec defined elsewhere in the docs so I don't think we need to cover every field unless we're adding new/more detailed or user friendly content.

docs-src/httproute.md
```diff
@@ -0,0 +1,303 @@
...
+
+[httproute]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRoute
+[gateways]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.RouteGateways
+[httprouterule]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule
+[hostname]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.Hostname
+[tls-config]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.RouteTLSConfig
+[rfc-3986]: https://tools.ietf.org/html/rfc3986
+[matches]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteMatch
+[filters]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteFilter
+[forwardto]: https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteForwardTo
```
robscott, line 303: I actually thought this was going to link to the ForwardTo section of this doc. I think it's probably better to link to the API spec like you're doing here, just not what I expected. Maybe we should make these links `target=_blank` since they're going to the spec?

robscott (commented):
Thanks for all your work on this! I realized part of the way through that I was commenting on things you only moved, not actually changed. I can move that to a follow up issue if you'd prefer.

docs-src/gateway.md
```diff
@@ -0,0 +1,47 @@
+# Gateway
+
+A `Gateway` is 1:1 with the life cycle of the configuration of
+infrastructure. When a user creates a `Gateway`, some load balancing
+infrastructure is provisioned or configured
+(see below for details) by the `GatewayClass` controller. `Gateway` is the
+resource that triggers actions in this API. Other resources in this API are
+configuration snippets until a Gateway has been created to link the resources
+together.
```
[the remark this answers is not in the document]
danehans: I didn't update any of the existing content, but simply moved it as part of the refactor.

docs-src/gateway.md
```diff
@@ -0,0 +1,47 @@
...
+
+The `Gateway` spec defines the following:
+
+*   The `GatewayClass` used to instantiate this Gateway.
+*   The Listener bindings, which define addresses and ports, protocol termination,
```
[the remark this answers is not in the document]
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/463#discussion_r524514788

docs-src/gateway.md
```diff
@@ -0,0 +1,47 @@
...
+    and TLS settings. The Listener configuration requested by a Gateway definition can
+    be incompatible with a given `GatewayClass` (e.g. port/protocol combination
+    is not supported).
```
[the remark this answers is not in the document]
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/463#discussion_r524514788

docs-src/gateway.md
```diff
@@ -0,0 +1,47 @@
...
+*   The Routes, which describe how traffic is processed and forwarded.
```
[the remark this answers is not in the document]
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/463#discussion_r524514788

docs-src/gateway.md
```diff
@@ -0,0 +1,47 @@
...
+* Use cloud APIs to create an LB instance.
+* Spawn a new instance of a software LB (in this or another cluster).
+* Add a configuration stanza to an already instantiated LB to handle the new
+  routes.
+* Program the SDN to implement the configuration.
+* Something else we haven’t thought of yet...
+
+The API does not specify which one of these actions will be taken. Note that a
+GatewayClass controller that manages in-cluster proxy processes MAY restrict
+Gateway configuration scope, e.g. only be served in the same namespace.
```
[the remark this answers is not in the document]
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/463#discussion_r524514788.

docs-src/httproute.md
danehans: I thought it would be useful to include an example of what an HTTPRoute status looks like. The API docs for status do not provide an example. I can remove if you feel strongly about this comment.

docs-src/httproute.md
danehans: > Maybe we should make these links target=_blank since they're going to the spec?

I'm not following what you mean by `target=_blank`. Can you provide an example?

danehans:
@robscott I addressed your comments either above or in commit `bea1950`, PTAL.

docs-src/httproute.md
robscott: Not important and maybe not even a good idea, would just default the link to open in a new tab/window depending on the browser.

robscott:
Thanks!

/lgtm
/approve

PR #466, opened 2020-11-17
Update GatewayClass Condition flow to add `Admitted` Condition type.

### LINKED ISSUE #465: Change GatewayClass Condition to `Accepted`

Currently, GatewayClass' `status` stanza has only one Condition type explicitly laid out in the type files - `InvalidParameters`. This makes sense as a coordination mechanism for controllers that actually use the `spec.parametersRef` field, but is confusing for ones that don't.

**Suggested change**:
We change the standard condition for GatewayClass to `Accepted`, with `InvalidParameters` as a sample *reason* for it to be false. This condition then will be the default, with its status set to `Unknown` until a controller sees the object, at which time the controller *must* update the Condition `status` to `true` or `false`.

I've got a PR for this change ready to go.

youngnick:
Fixes #465

### DIFF
--- a/apis/v1alpha1/gatewayclass_types.go
+++ b/apis/v1alpha1/gatewayclass_types.go
@@ -85,10 +85,25 @@ type GatewayClassSpec struct {
 type GatewayClassConditionType string
 
 const (
-	// GatewayClassConditionStatusInvalidParameters indicates the
-	// validity of the Parameters set for a given controller. This
-	// will initially start off as "Unknown".
-	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+	// GatewayClassConditionStatusAdmitted indicates that the GatewayClass
+	// has been admitted by the controller set in the `spec.controller`
+	// field.
+	//
+	// It defaults to False, and MUST be set by a controller when it sees
+	// a GatewayClass using its controller string.
+	// The status of this condition MUST be set to true if the controller will support
+	// provisioning Gateways using this class. Otherwise, this status MUST be set to false.
+	// If the status is set to false, the controller SHOULD set a Message and Reason as an
+	// explanation.
+	// GatewayClassNotAdmittedInvalidParameters is provided as an example Reason.
+	GatewayClassConditionStatusAdmitted GatewayClassConditionType = "Admitted"
+
+	// GatewayClassNotAdmittedInvalidParameters should be used as a Reason on the Admitted
+	// condition if the parametersRef field is invalid, with more detail in the message.
+	GatewayClassNotAdmittedInvalidParameters string = "InvalidParameters"
+
+	// GatewayClasssNotAdmittedWaiting is the default Reason on a new GatewayClass.
+	GatewayClasssNotAdmittedWaiting string = "Waiting"
 
 	// GatewayClassFinalizerGatewaysExist should be added as a finalizer to the
 	// GatewayClass whenever there are provisioned Gateways using a GatewayClass.
@@ -103,7 +118,7 @@ type GatewayClassStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{type: "InvalidParameters", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	// +kubebuilder:default={{type: "Admitted", status: "False", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
--- a/config/crd/bases/networking.x-k8s.io_gatewayclasses.yaml
+++ b/config/crd/bases/networking.x-k8s.io_gatewayclasses.yaml
@@ -83,8 +83,8 @@ spec:
                 - lastTransitionTime: "1970-01-01T00:00:00Z"
                   message: Waiting for controller
                   reason: Waiting
-                  status: Unknown
-                  type: InvalidParameters
+                  status: "False"
+                  type: Admitted
                 description: Conditions is the current status from the controller for this GatewayClass.
                 items:
                   description: "Condition contains details for one aspect of the current state of this API Resource. --- This struct is intended for direct use as an array at the field path .status.conditions.  For example, type FooStatus struct{     // Represents the observations of a foo's current state.     // Known .status.conditions.type are: \"Available\", \"Progressing\", and \"Degraded\"     // +patchMergeKey=type     // +patchStrategy=merge     // +listType=map     // +listMapKey=type     Conditions []metav1.Condition `json:\"conditions,omitempty\" patchStrategy:\"merge\" patchMergeKey:\"type\" protobuf:\"bytes,1,rep,name=conditions\"` \n     // other fields }"


apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,21 @@ type GatewayClassSpec struct {
...
-	// will initially start off as "Unknown".
-	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+	// GatewayClassConditionStatusAccepted indicates that the GatewayClass
+	// has been accepted by the controller set in the `spec.controller`
+	// field.
+	//
+	// It defaults to Unknown, and MUST be set by a controller when it sees
+	// a GatewayClass using its controller string.
+	// It must be set to `status: true` if the controller is using the class,
+	// and `status:false` if the controller is not using the class for some reason.
```
robscott, line 95: This feels a bit ambiguous. I think some might interpret "using the class" as "the controller is using the class to render at least one Gateway". Put another way, if no Gateway is trying to use this class, is the class actually used? If I'm interpreting this correctly, I think something like this might be a bit clearer:

> The status of this condition MUST be set to true if the controller will support provisioning Gateways using this class. Otherwise, this status MUST be set to false.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,21 @@ type GatewayClassSpec struct {
...
+	// In either case, the controller SHOULD set a Message and Reason as an explanation.
+	// GatewayClassNotAcceptedInvalidParameters is provided as an example Reason.
```
robscott, line 97: Do we need a message and reason if status == true?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,21 @@ type GatewayClassSpec struct {
...
+	GatewayClassConditionStatusAccepted GatewayClassConditionType = "Accepted"
+
+	// GatewayClassNotAcceptedInvalidParameters should be used as a Reason on the Accepted
+	// condition if the parametersRef field is invalid, with more detail in the message.
+	GatewayClassNotAcceptedInvalidParameters string = "InvalidParameters"
```
robscott, line 102: We should likely add a Waiting reason here to correspond with our default.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -103,7 +114,7 @@ type GatewayClassStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{type: "InvalidParameters", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	// +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
```
robscott, line 117: I think it may be clearer to start off with a false state.
```suggestion
	// +kubebuilder:default={{type: "Accepted", status: "false", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
```

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,21 @@ type GatewayClassSpec struct {
 type GatewayClassConditionType string
 
 const (
-	// GatewayClassConditionStatusInvalidParameters indicates the
-	// validity of the Parameters set for a given controller. This
-	// will initially start off as "Unknown".
-	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+	// GatewayClassConditionStatusAccepted indicates that the GatewayClass
```
robscott, line 88: To follow up with the suggestion from @jpeach, it might be worth standardizing on naming between this and routes since they use `Admitted` for a similar purpose. I don't have a strong preference between the words but might slightly prefer using `Admitted` here to minimize the changes necessary.

robscott (commented):
Thanks for your work on this!

apis/v1alpha1/gatewayclass_types.go
youngnick: I wasn't sure, that makes sense to add.

apis/v1alpha1/gatewayclass_types.go
youngnick: I thought that the defaulting behavior for conditions is supposed to be `Unknown` until the relevant controller has seen it? In Contour's case, it's possible for us to set `False`, so the distinction between `Unknown` and `False` is relevant.

apis/v1alpha1/gatewayclass_types.go
youngnick: That makes sense. `Admitted` it is.

apis/v1alpha1/gatewayclass_types.go
robscott: I think "Unknown" made a lot more sense when the condition was "InvalidParameters". Now that we have switched to "Admitted" I think it's reasonable to start in a state where "Admitted" is false with a "Waiting" reason. Controllers should switch it to true or change the reason it is false if/when they process the class. I think `admitted == false` means that there is no controller that will support provisioning gateways of this class. That seems like a reasonable default state until a controller picks this up and changes it.

apis/v1alpha1/gatewayclass_types.go
robscott: I think it's also fine + potentially simpler to say "If the status is set to false, the controller SHOULD set a Message and Reason as an explanation". I'm not sure we need any kind of message reason for the true case.

apis/v1alpha1/gatewayclass_types.go
youngnick: Yeah, that makes sense. I'll change this up.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,25 @@ type GatewayClassSpec struct {
...
-	// GatewayClassConditionStatusInvalidParameters indicates the
-	// validity of the Parameters set for a given controller. This
-	// will initially start off as "Unknown".
-	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+	// GatewayClassConditionStatusAdmitted indicates that the GatewayClass
+	// has been accepted by the controller set in the `spec.controller`
+	// field.
+	//
+	// It defaults to Unknown, and MUST be set by a controller when it sees
+	// a GatewayClass using its controller string.
```
robscott, line 93: This should be updated to match the new default value.

youngnick: 🤦

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -85,10 +85,25 @@ type GatewayClassSpec struct {
 type GatewayClassConditionType string
 
 const (
-	// GatewayClassConditionStatusInvalidParameters indicates the
-	// validity of the Parameters set for a given controller. This
-	// will initially start off as "Unknown".
-	GatewayClassConditionStatusInvalidParameters GatewayClassConditionType = "InvalidParameters"
+	// GatewayClassConditionStatusAdmitted indicates that the GatewayClass
+	// has been accepted by the controller set in the `spec.controller`
```
robscott, line 89:
```suggestion
	// has been admitted by the controller set in the `spec.controller`
```

youngnick: Yeah, better not to be confusing.

robscott: Quick sanity check here, this has been marked resolved but I'm still seeing "accepted" in the comments? Maybe my copy is out of date or an update didn't get pushed?

youngnick:
I think that's everything from @robscott? @hbagdi, what do you think?

robscott:
Thanks!

/lgtm
/approve

PR #471, opened 2020-11-20
fix GatewayClassStatus default value

### LINKED ISSUE #470: Fix the GatewayClass default Condition
**What happened**:

Create a GatewayClass object and have the following default condition:

```json
{
  "conditions": [
    {
      "lastTransitionTime": "1970-01-01T00:00:00Z",
      "message": "Waiting for controller",
      "reason": "Waiting",
      "status": "Unknown",
      "type": "InvalidParameters"
    }
  ]
}

```

According to #466

```json
{
  "conditions": [
    {
      "lastTransitionTime": "1970-01-01T00:00:00Z",
      "message": "Waiting for controller",
      "reason": "Waiting",
      "status": "False",
      "type": "Admitted"
    }
  ]
}
```

```bash
kubectl apply -f config/crd/bases/
kubectl apply -f examples/basic-http.yaml
kubectl get -o json gatewayclasses/acme-lb | jq .status
```

jbdoumenjou:
Fixes #470 

Maybe it would be better to remove one of the default values but I keep both to be homogeneous with other objects.
Update documentation accordingly.


jbdoumenjou:
I've signed the CLA

robscott:
Thanks for this PR! This generally LGTM, but I'm a bit concerned about the huge diff in the generated `docs/spec` page. Did anything go wrong when running `make generate`? I thought our CI should catch a regression like this but that doesn't seem to be working.

jbdoumenjou:
@robscott,
you were right I fixed, thank you. :smile:

robscott:
Thanks!

/approve

PR #473, opened 2020-11-24
Adding sig-network leads and service-apis maintainers as approvers

robscott:
This mirrors the [structure that cluster-api uses](https://github.com/kubernetes-sigs/cluster-api/blob/master/OWNERS). I've left Bowei in temporarily just in case all these aliases don't end up working as expected.

/cc @danehans @hbagdi @jpeach 
/assign @bowei


hbagdi (approved):
Verified against the referenced org manifests and checked for misspellings in usernames. All checks out.
/lgtm

PR #474, opened 2020-11-24
Adding a PR Template

robscott:
This is mostly a copy of the [Kubernetes PR template](https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) with some of the less relevant bits removed. All I really wanted was some kind of release note automation, so if the rest of this feels like it adds too much overhead, I'm fine with simplifying this.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,52 @@
+<!--  Thanks for sending a pull request! Here are some tips for you:
+
+1. If this is your first time contributing to Service APIs, please read our
+   developer guide (https://kubernetes-sigs.github.io/service-apis/devguide/)
+   and our community page (https://kubernetes-sigs.github.io/service-apis/community/).
+2. If this is your first time contributing to a Kubernetes project, please read
+   our contributor guidelines:
+   https://git.k8s.io/community/contributors/guide/first-contribution.md#your-first-contribution
+3. Please label this pull request according to what type of issue you are
+   addressing, especially if this is a release targeted pull request. For
+   reference on required PR/issue labels, read here:
+   https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-kind-label
+4. If you want *faster* PR reviews, read how:
+   https://git.k8s.io/community/contributors/guide/pull-requests.md#best-practices-for-faster-reviews
+5. If the PR is unfinished, see how to mark it:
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,52 @@
...
+-->
+Fixes #
+
+**Does this PR introduce a user-facing change?**:
+<!--
+If no, just write "NONE" in the release-note block below.
+If yes, a release note is required:
+Enter your extended release note in the block below. If the PR requires
+additional action from users switching to the new release, include the string
+"action required".
```
hbagdi, line 48: I don't think the `action required` bit needed for this project. If you agree, then let's remove to simplify.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,52 @@
...
+1. If this is your first time, please read our contributor guidelines:
+   https://git.k8s.io/community/contributors/guide/first-contribution.md#your-first-contribution
+2. Please label this pull request according to what type of issue you are
+   addressing, especially if this is a release targeted pull request. For
+   reference on required PR/issue labels, read here:
+   https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-kind-label
+3. If you want *faster* PR reviews, read how:
+   https://git.k8s.io/community/contributors/guide/pull-requests.md#best-practices-for-faster-reviews
+4. If the PR is unfinished, see how to mark it:
+   https://git.k8s.io/community/contributors/guide/pull-requests.md#marking-unfinished-pull-requests
```
hbagdi, line 12: I don't want to make this more verbose but what do you think about adding a link to https://kubernetes-sigs.github.io/service-apis/community/ here?

hbagdi (commented):
+1, I've left some minor suggestions, but otherwise lgtm.

.github/PULL_REQUEST_TEMPLATE.md
robscott: Good call, removed that.

.github/PULL_REQUEST_TEMPLATE.md
robscott: Good call, added.

robscott:
@hbagdi I think I've resolved everything, PTAL.

PR #475, opened 2020-11-24
Add Set field to HTTPRequestHeaderFilter

nak3:
This patch adds `Set` field to `HTTPRequestHeaderFilter`.

Currently `Add`(append) and `Remove` are available in
HTTPRequestHeaderFilter. But it is not possible to overwrite a
header.


howardjohn (commented):
LGTM

apis/v1alpha1/httproute_types.go
```diff
@@ -432,6 +432,24 @@ const (
...
+	// Config:
+	//   set: {"my-header": "XYZ"}
+	//
+	// Output:
+	//   GET /foo HTTP/1.1
+	//   my-header: XYZ
+	//
+	// Support: Extended
+	// +optional
+	Set map[string]string `json:"set,omitempty"`
```
hbagdi, line 451: +1 to this change.
One concern. Can most implementations differentiate between addition and override?

robscott: From what I can tell, adding headers is pretty widely supported among other implementations, but setting/overwriting headers is certainly trickier. HAProxy has a very clear differentiation that would match this config closely (add-header and set-header). The distinction is less clear with nginx (add_header and proxy_set_header but defined in different modules), and I think you might need a custom filter to overwrite a header with envoy. 

With all that said, I think this is a good addition for the implementations that can support it, and the `extended` support level feels appropriate.

robscott:
Thanks for this addition @nak3! We're still trying to figure out what we'd like to include in updates to the v1alpha1 API vs new work on the v1alpha2 API. We should be able to discuss this more at our community meeting next week. Until then, I'd like to hold this change to ensure we're not adding more than we want to to v1alpha1.

/hold

howardjohn:
Envoy can do this easily, see `append` flag:
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto.html?highlight=headervalueoption#config-core-v3-headervalueoption

FWIW its part of Istio API as well.

robscott:
We discussed this a bit further yesterday and this seemed like a reasonable change. We also got https://github.com/kubernetes-sigs/service-apis/pull/492 merged in yesterday which clarifies that we can make additive changes to an existing API. With that I think this is mostly good to go.

/hold cancel
/retest

robscott:
I'm going to approve this and wait for a final LGTM from @hbagdi or @jpeach.

/approve

apis/v1alpha1/httproute_types.go
```diff
@@ -432,6 +432,24 @@ const (
 // HTTPRequestHeaderFilter defines configuration for the RequestHeaderModifier
 // filter.
 type HTTPRequestHeaderFilter struct {
+	// Set overwrites the request with the given header (name, value)
+	// before the action.
+	//
+	// Input:
+	//   GET /foo HTTP/1.1
+	//   My-Header: ABC
```
hbagdi, line 440: Can we lower case this?
Header field names are case-insensitive but keeping this example consistent throughout the documentation would be ideal.

apis/v1alpha1/httproute_types.go
```diff
@@ -432,6 +432,24 @@ const (
...
+	//
+	// Output:
+	//   GET /foo HTTP/1.1
+	//   my-header: XYZ
+	//
+	// Support: Extended
+	// +optional
+	Set map[string]string `json:"set,omitempty"`
+
 	// Add adds the given header (name, value) to the request
```
hbagdi, line 453: Since now we have `Add` and `Set`, we should clarify here what happens when a header already exists in the `Add` case.

hbagdi (commented):
This is very close. I've a couple of small documentation improvements.

apis/v1alpha1/httproute_types.go
```diff
@@ -456,16 +476,16 @@ type HTTPRequestHeaderFilter struct {
 	//
 	// Input:
 	//   GET /foo HTTP/1.1
-	//   My-Header1: ABC
-	//   My-Header2: DEF
-	//   My-Header2: GHI
```
nak3, line 461: I believe this is a typo. Fixed by `s/My-Header2/My-Heade3` in this PR.

apis/v1alpha1/httproute_types.go
```diff
@@ -432,19 +432,39 @@ const (
...
+	// Input:
+	//   GET /foo HTTP/1.1
 	//   my-header: foo
 	//
+	// Config:
+	//   add: {"my-header": "bar"}
+	//
+	// Output:
+	//   GET /foo HTTP/1.1
+	//   my-header: foo bar
```
hbagdi, line 466: This seems incorrect. Header values are allowed to contain whitespaces.
The usual textual representation for multiple headers is:

```
	//   GET /foo HTTP/1.1
	//   my-header: foo
	//   my-header: bar
```

apis/v1alpha1/httproute_types.go
```diff
@@ -432,19 +432,39 @@ const (
...
-	//   add: {"my-header": "foo"}
+	//   set: {"my-header": "bar"}
 	//
 	// Output:
 	//   GET /foo HTTP/1.1
+	//   my-header: bar
+	//
+	// Support: Extended
+	// +optional
+	Set map[string]string `json:"set,omitempty"`
```
jpeach, line 451: Is the order of operations defined here? set and add will yield different results depending on order, so we should nail that down.

apis/v1alpha1/httproute_types.go
nak3: Thank you! Updated.

apis/v1alpha1/httproute_types.go
nak3: I think it needs another discussion, so perhaps opening another PR is better? The issue seems to be tracked by https://github.com/kubernetes-sigs/service-apis/issues/480.

hbagdi:
/lgtm

@jpeach How does https://github.com/kubernetes-sigs/service-apis/pull/475#discussion_r542945413 sound to you? We had discussed in the office hours to track the ordering bug separately.
/hold
Please cancel the hold if that is resolved.

There is another minor thing where if the PR was based on the latest master then the PR description could add release notes. This is minor enough that we can ignore.

nak3:
@jpeach Friendly ping. Could you please check above comment?

Issue #447, opened 2020-11-04
Develop a plan for repository and docs structure post v1alpha1


robscott:
Although we're currently very focused on a v1alpha1 release, we also need to figure out how we'll update our project structure after the API release. We need to answer the following questions:

- How can we ensure that all new work targets v1alpha2 by default?
- Should we create branches for each API release?
- How can we structure the docs to cover multiple API versions?

hbagdi:
Removing this from v1alpha1 milestone because this is intended for post v1alpha1.

robscott:
One thing that we’ll need to figure out pretty quickly - what do we want to do with changes that come in now that v1alpha1 is launched?

As I see it, we have 2 main options:
1. Create a v1alpha2 API directory and branch immediately, make the majority of changes excluding bug fixes there.
2. Continue to iterate on v1alpha1, release new minor versions of service-apis with updates to v1alpha1. We would avoid a new API version if/until we wanted to make breaking changes.

/cc @bowei @danehans @hbagdi @jpeach

robscott:
I'm personally leaning towards option 1 since I think it could be confusing to have lots of variations to the v1alpha1 API.

hbagdi:
+1 to option 1. It will be much more easy to maintain and ship changes that way.
One concern I have is too many versions. CRD versions are kind of heavy weight and having a lot of them can easily confuse users and maintainers alike.

PR #476, opened 2020-11-24
de-duplicate types by embedding RouteForwardTo into HTTPRouteForwardTo

### LINKED ISSUE #440: Reuse as much as possible between HTTPRouteForwardTo and RouteForwardTo
There's a lot of overlap in these two types, and so far they've both been entirely duplicated. I believe the only difference is that `HTTPRouteForwardTo` includes filters.

hbagdi:
This is a breaking change for the Go API.

Fix #440


robscott:
I think we should try to figure out what to do with breaking changes. Although less significant than breaking changes to the k8s APIs, I think a breaking change to the Go API is still important to get right. Maybe this is something that we need to move into v1alpha2?
/hold

hbagdi:
> Maybe this is something that we need to move into v1alpha2?

I had assumed that this would be v1alpha2 and I had also assumed that all v1alpha2 work will be done in the `master` branch, since that's our development branch. The problem here is that our wesbite is also driven off master. This probably needs to hold until #447 is completed.

thockin:
I'm OK with de-duplicating as long as the WHOLE type is applicable in
every context.

Any time we re-use a struct and add a comment like "field X only
apples when this type is used under a Foo" or "field Y must be 0 when
used in a Bar", I die a little inside.

hbagdi:
> Any time we re-use a struct and add a comment like "field X only
apples when this type is used under a Foo" or "field Y must be 0 when
used in a Bar", I die a little inside.

100% agree with this.
The only field I can think of that could end up in such a situation here is `Weight`. I'd love to know what others think about it.

hbagdi:
As discussed in community meeting on Feb 3rd, this is no longer blocked because upstream k8s allows for breaking changes in the Go API (minimize as much as possible).
I think this is a reasonable refactor but I'm open to hearing concerns around not merging this.

I can update the PR if we want to move this forward.

/hold cancel

hbagdi:
We discussed this PR in office hours today and decided not to merge it in.
There is nothing against this PR but there are no stronger drivers for it either.
HTTPRoute is a little special than other ones and having a little duplication here doesn't harm much.

If you are reading this and feel strongly about this, please feel free to re-open.

/close

PR #477, opened 2020-11-24
enable search on docs site

hbagdi:
Search functionality was previously disabled in 8b26365 to prevent merge
conflicts during initial development phase.

This patch re-enables search for a better experience.


robscott:
I'm a little hesitant about this one. Especially since our docs include API spec now. This would mean _any_ change to API types or docs (99% of PRs) would conflict with each other. I agree that search would be helpful, but I'm hesitant to add that kind of overhead to the dev workflow.

robscott:
As I recall, the issue was that all docs content was stored in the single line index.json file. That meant any change to docs would automatically conflict.

hbagdi:
> As I recall, the issue was that all docs content was stored in the single line index.json file. That meant any change to docs would automatically conflict.

Indeed. This certainly increases the overhead. I do find search useful but we can chose to hold this until we settle on #447.

PR #483, opened 2020-11-26
Fixing bad link in README.md

kingdonb:
https://kubernetes-sigs.github.io/service-apis/concepts renders as `404 - Not found`

"api-overview" is the first section in Concepts, so I assume it was intended to link here instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,6 @@ Participation in the Kubernetes community is governed by the
 [slack]: https://kubernetes.slack.com/messages/sig-network-service-apis
 [guides]: https://kubernetes-sigs.github.io/service-apis/guides
 [spec]: https://kubernetes-sigs.github.io/service-apis/spec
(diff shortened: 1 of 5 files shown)


kingdonb:
I started reading from the README and branched out until I was satisfied I'd read all the pages that a new person would read. (I am new here, so this was easy.)

My reading is complete for now, I found some other errors besides the bad link

robscott:
@kingdonb Thanks for the PR! Unfortunately it looks like you'll also have to run `make docs` for these updates to take effect.

kingdonb:
Thanks for the tip! Ran `make docs` and rebased, I took a look at the generated docs and it looks like that indented list is better now.

hbagdi:
@kingdonb This looks good. Do you mind squashing all your commits into a single commit? It should be good to merge then.

kingdonb:
Done! 👍

PR #494, opened 2020-12-06
Fix broken doc link

beriberikix:
**What type of PR is this?**
/kind documentation

Looks like there was a big doc change and this link wasn't updated.

### DIFF
--- a/docs-src/faq.md
+++ b/docs-src/faq.md
@@ -53,7 +53,7 @@
       a proud tradition of extending Ingress objects.
 
     * Use API-defined extension points. Some Service
-      API objects have explicit [extension points](/concepts/#extension-points)
+      API objects have explicit [extension points](/api-overview/#extension-points)
       for implementations to use.
 
 *  **Q: Where can I find Service API releases?<br>**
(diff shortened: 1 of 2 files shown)


beriberikix:
Thanks @jpeach for sharing how to generate docs with `make generate`! Also tested with `mkdocs serve`.

robscott:
Thanks!

/ok-to-test

hbagdi (approved):
A quick grep reveals that there is no other `/concepts` link in our docs.
Thanks for these fixes, @beriberikix.

/lgtm

beriberikix:
So while this worked when testing locally, something broke when deployed. Currently it points to:

`https://kubernetes-sigs.github.io/api-overview/#extension-points` 

when it should be:

`https://kubernetes-sigs.github.io/service-apis/api-overview/#extension-points`. 

I can't see any obvious issues - any ideas?

hbagdi:
@beriberikix I think this should fix it: https://github.com/kubernetes-sigs/service-apis/pull/497/files

PR #496, opened 2020-12-07
docs: add guide on using TCP route

### LINKED ISSUE #333: Add examples to documentation

### LINKED ISSUE #488: RFE: add layer-3 (TCPRoute) examples
I've been reading through the service API documentation, and it's mostly excellent! Thank you, especially, for writing so many examples. It really makes it a lot easier to grok the shape of the API without poring through autogenerated documentation.

However, I don't see any examples for layer-3 Gateways. 

Some basic examples for how a layer-3 load balancer might be configured.

I know that one of the goals of this API is to replace the Services API. Thus, an example for each existing Service type (ExternalName, ClusterIP, NodePort, and LoadBalancer) would be really nice.

Thanks!

hbagdi:
**What type of PR is this?**

/kind documentation

Adds a guide to demonstrate the use of TCPRoute

Fixes #488
Fixes #333

```release-note
NONE
```

### DIFF
--- a/docs-src/tcp.md
+++ b/docs-src/tcp.md
@@ -1,0 +1,26 @@
+Service APIs is designed to work with multiple protocols.
+[TCPRoute](spec/#networking.x-k8s.io/v1alpha1.TCPRoute) is one such route which
+allows for managing TCP traffic.
+
+In this example, we have one Gateway resource and two TCPRoute resources that
+distribute the traffic with the following rules:
+
+- All TCP streams on port 8080 of the Gateway are forwarded to port 6000 of
+  `my-foo-service` Kubernetes Service.
+- All TCP streams on port 8090 of the Gateway are forwarded to port 6000 of
+  `my-bar-service` Kubernetes Service.
+
+Please note the following:
+
+- The `protocol` of listeners on the Gateway is `TCP`.
+- Each listener selects exactly one TCPRoute. This is important since the routing
+  decision is performed based on destination port only. If more metadata is used
+  for routing decisions, then one may associate multiple TCPRoutes to a single
+  Gateway listener. Implementations can support such use-cases by adding a custom
+  resource to specify advanced routing properties and then referencing it in
+  `spec.rules[].matches[].extensionRef`. Conflicts due to routing colisions should
+  be resolved as per the [conflict resolution](guidelines.md#conflicts) guidelines.
+
+```
+{% include 'basic-tcp.yaml' %}
+```
--- a/examples/basic-tcp.yaml
+++ b/examples/basic-tcp.yaml
@@ -12,10 +12,10 @@ spec:
 kind: Gateway
 apiVersion: networking.x-k8s.io/v1alpha1
 metadata:
-  name: my-gateway
+  name: my-tcp-gateway
 spec:
   gatewayClassName: acme-lb
-  listeners:  # Use GatewayClass defaults for listener definition.
+  listeners:
   - protocol: TCP
     port: 8080
     routes:
@@ -23,8 +23,13 @@ spec:
       selector:
         matchLabels:
           "app": "foo"
-      namespaces:
-        from: "All"
+  - protocol: TCP
+    port: 8090
+    routes:
+      kind: TCPRoute
+      selector:
+        matchLabels:
+          "app": "bar"
 ---
 kind: TCPRoute
 apiVersion: networking.x-k8s.io/v1alpha1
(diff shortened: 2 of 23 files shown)


docs-src/tcp.md
```diff
@@ -0,0 +1,23 @@
...
+- All TCP streams on port 8090 of the Gateway are forwarded to port 6000 of
+  `my-bar-service` Kubernetes Service.
+
+Please note the following:
+
+- The `protocol` of listeners on the Gateway is `TCP`.
+- Each listener selects exactly one TCPRoute. This is important since routing
+  decision is performed based on destination port only. If more metadata is used
+  for routing decisions, then one may associate multiple TCPRoutes to a single
+  Gateway listener.
```
robscott, line 19: I think it would be helpful to describe briefly what this may look like. Maybe something like:

```
Implementations may support custom resources for more advanced matching logic. This could be used to associate multiple TCPRoutes with a single Gateway listener.
```

docs-src/tcp.md
```diff
@@ -0,0 +1,23 @@
...
+
+- All TCP streams on port 8080 of the Gateway are forwarded to port 6000 of
+  `my-foo-service` Kubernetes Service.
+- All TCP streams on port 8090 of the Gateway are forwarded to port 6000 of
+  `my-bar-service` Kubernetes Service.
+
+Please note the following:
+
+- The `protocol` of listeners on the Gateway is `TCP`.
+- Each listener selects exactly one TCPRoute. This is important since routing
```
robscott, line 16:
```suggestion
- Each listener selects exactly one TCPRoute. This is important since the routing
```

docs-src/tcp.md
```diff
@@ -0,0 +1,23 @@
+Service APIs is designed to work with multiple protocols (unlike the HTTP only
+Ingress/v1 resource). [TCPRoute](spec/#networking.x-k8s.io/v1alpha1.TCPRoute) is
+one such route which allows for managing TCP traffic.
```
robscott, line 3: I don't think we necessarily need the comparison to Ingress here. Though this is a great reminder that we should probably have an entire page devoted to the differences between these APIs and Ingress and/or some examples of how Ingress config could be translated.

examples/basic-tcp.yaml
```diff
@@ -37,5 +42,18 @@ metadata:
...
   - forwardTo:
-    - serviceName: my-service
+    - serviceName: my-foo-service
+      port: 6000
+---
+kind: TCPRoute
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: tcp-app-2
+  namespace: default
```
robscott, line 52: I think we should probably remove any hardcoded namespace references throughout our examples. Leaving these in makes it harder to deploy an example in a test namespace.

robscott (commented):
Thanks for your work on this! It will be great to have a TCPRoute guide in our docs.

examples/basic-tcp.yaml
hbagdi: I had this exact same thought but left it in because other examples had it.
Tracking this in https://github.com/kubernetes-sigs/service-apis/issues/498 to avoid scope creep here.

docs-src/tcp.md
hbagdi: Removed the mention of Ingress.

docs-src/tcp.md
hbagdi: Worded this a little differently. PTAL.

hbagdi:
ping @robscott for another review.

docs-src/tcp.md
```diff
@@ -0,0 +1,25 @@
...
+
+Please note the following:
+
+- The `protocol` of listeners on the Gateway is `TCP`.
+- Each listener selects exactly one TCPRoute. This is important since the routing
+  decision is performed based on destination port only. If more metadata is used
+  for routing decisions, then one may associate multiple TCPRoutes to a single
+  Gateway listener. Implementations can support such use-cases by adding a custom
+  resource to specify advanced routing properties and then referencing it in
+  `spec.rules[].matches[].extensionRef`.
```
robscott, line 21: Not absolutely required, but this might be a good time to reference our conflict resolution principles.

robscott:
Thanks! Adding a temporary hold in case you want to add a reference to conflict resolution principles. If not feel free to remove hold and get this in.

/lgtm
/hold

docs-src/tcp.md
hbagdi: Added. PTAL.

robscott:
Thanks!

/lgtm
/hold cancel

PR #501, opened 2020-12-10
Clarifying that unreleased changes may be reverted

robscott:
**What type of PR is this?**
/kind documentation

This clarifies that unreleased changes may be reverted. This is important to note since we continue to actively develop on our master branch.

```release-note
NONE
```


robscott:
To clarify a bit further here, I'm not trying to suggest that we should merge things with the intent of reverting them later. Instead I'm trying to ensure that no one relies on our latest commits to any branch. This would be a similar approach to Kubernetes in that nothing should be relied on outside of a published release.

PR #504, opened 2020-12-14
docs: pin to CRD manifests to v0.1.0

### LINKED ISSUE #502: Improve the CRD Installation Experience
An easier way to install Service APIs CRDs, including a specific version. The current user experience is:

```
kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
| kubectl apply -f -
```

**Why is this needed**:
Currently, the installation experience has the following limitations:

1. kustomize is required
2. Only the master branch is referenced.

Should a single yaml manifest get rendered so users can simply kubectl https://kubernetes-sigs.github.io/service-apis/release/${VERSION}/service-apis-crds.yaml. This could be accomplished with something like:
```
kustomize build config/crd > examples/crd/${VERSION}/service-apis-crds.yaml
```

If more trickery is needed or we need to keep the CRDs separate, we can use an installer script similar to Istio. Thoughts?

/cc @robscott @bowei @hbagdi

hbagdi:
**What type of PR is this?**

/kind documentation
Pins the manifests being pulled to v0.1.0. This ensures that any problematic changes in the `master` branch do not affect end users immediately.

Fixes #502

```release-note
NONE
```

### DIFF
--- a/docs-src/getting-started.md
+++ b/docs-src/getting-started.md
@@ -8,7 +8,7 @@ be installed into any Kubernetes (>= 1.16) cluster.
 To install the CRDs, please execute:
 
 ```
-kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=master" \
+kubectl kustomize "github.com/kubernetes-sigs/service-apis/config/crd?ref=v0.1.0" \
 | kubectl apply -f -
 ```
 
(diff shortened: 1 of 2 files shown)


robscott:
Thanks!

/lgtm

PR #509, opened 2020-12-20
Make an editorial pass over the API documentation.

jpeach:
**What type of PR is this?**

/kind cleanup
/kind documentation

Especially now that enum constants are rendered in the HTML API
documentation, the godoc conventions make it harder for a reader
to understand the intentions of the documentation.

Make an editorial pass that improves readability without any semantic
changes.

n/a

```release-note
NONE
```


apis/v1alpha1/tlsroute_types.go
```diff
@@ -24,12 +24,12 @@ import (
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 
-// TLSRoute is the Schema for the TLSRoute resource. TLSRoute is similar to
-// TCPRoute but can be configured to match against TLS-specific metadata.
+// The TLSRoute TLSRoute resource is similar to
```
youngnick, line 27: Nit: Doubled `TLSRoute`?

youngnick:
Looks like you'll need to revert some changes to pass go linting though, godoc requires constant comments to be phrased `CONSTANTNAME does something`.

jpeach:
> On 21 Dec 2020, at 10:02 am, Nick Young <notifications@github.com> wrote:
> 
> 
> Looks like you'll need to revert some changes to pass go linting though, godoc requires constant comments to be phrased CONSTANTNAME does something.

Yup, I pinged slack on this. The go convention conflicts with making the docs useful for their intended purpose (people generating YAML). Specifically, the constant names are completely useless to anyone reading the docs to figure out how to interpret the API, since those names aren't visible in the API.


We could nobble the golint error if we switch to golangci-lint.

robscott:
Thanks for the work on this @jpeach! There are a lot of great improvements here, I'm not quite sure what to do as far as doc links though. The links added in this PR would look somewhat out of place everywhere but the API spec docs. This is likely much more complicated, but it seems like this is the kind of issue that would ideally be solved by improving the docs/spec generation here. If that could recognize the names of types it was processing and add links automatically we'd have a readable spec + go types that pass linting.

jpeach:
> Thanks for the work on this @jpeach! There are a lot of great improvements here, I'm not quite sure what to do as far as doc links though. The links added in this PR would look somewhat out of place everywhere but the API spec docs. This is likely much more complicated, but it seems like this is the kind of issue that would ideally be solved by improving the docs/spec generation here. If that could recognize the names of types it was processing and add links automatically we'd have a readable spec + go types that pass linting.

Yeah IIUC godoc automatically links type names that it finds in the doc text. But the godoc currently has a ton of ugliness, so I don't really see the markdown links as a blocker here. My perspective is that most people will read and learn the API through the HTML docs.

https://github.com/ahmetb/gen-crd-api-reference-docs/issues/30

robscott:
@jpeach that is a good point, the annotations we already have make things like https://pkg.go.dev/sigs.k8s.io/service-apis/apis/v1alpha1 rather ugly. This would represent a further shift away from readable godocs towards more readable HTML docs. I'm not sure that's necessarily a bad thing, but interested in what others think here.

/cc @hbagdi @danehans @bowei

hbagdi:
> My perspective is that most people will read and learn the API through the HTML docs.

I think we have control over this part of user-experience. Most doc generators have some quirks that make them harder to use. Whenever I'm browsing upstream k8s APIs, I use the HTML browser to figure out the types I need to lookup and then go and read the godoc version.

Regarding the markdown in Godoc, I'd rather prefer to not have it in there but I can live with a little bit of it sprinkled around.

apis/v1alpha1/gateway_types.go
```diff
@@ -466,54 +468,45 @@ type RouteNamespaces struct {
...
 
-	// NamedAddressType is an opaque identifier. The interpretation
-	// of the name is dependent on the controller. If a NamedAddress is
-	// requested but unsupported by an implementation, the controller
-	// should raise the "Detached" listener status condition on the Gateway
-	// with the "UnsupportedAddress" reason.
+	// An opaque identifier that represents a specific IP address. The interpretation
+	// of the name is dependent on the controller. For example,
+	// a "NamedAddress" might be a cloud-dependent ID for a "static"
+	// or "elastic" IP.
```
hbagdi, line 509: I think it is best to leave the idiomatic Godoc style "<Identifier> is ..." in here for the sake of consistency across the entire project.

jpeach: The problem with the Go idiom here is that the API isn't the constant t name but the constant value (ie. the value is what you need in the YAML). So using the constant name here is actually misleading readers of the documentation. It's also confusing because the constant names done appear anywhere in the HTML rendering.

robscott: This is an interesting one. I took a look at the Kubernetes reference docs and they lack consistency here but the godoc style Harry is suggesting above does seem to be more common ([ConfigMap](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#configmap-v1-core) is an example of that). Your proposed change does certainly read better in the reference docs, we just need to determine how heavily we want to prioritize that over other formats.

apis/v1alpha1/gateway_types.go
```diff
@@ -466,54 +468,45 @@ type RouteNamespaces struct {
...
 	//
 	// +kubebuilder:default=IPAddress
 	Type AddressType `json:"type,omitempty"`
 
-	// Value. Examples: "1.2.3.4", "128::1", "my-ip-address". Validity of the
-	// values will depend on `Type` and support by the controller.
+	// Value of the address.
+	// The validity of the  values will depend on the type and support by the controller.
+	//
+	// Examples: "1.2.3.4", "128::1", "my-ip-address".
```
robscott, line 481: Nit: The indentation here feels a bit off to me. I generally try to wrap lines at 80 chars - instead this seems to add a new line for each sentence.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -80,30 +80,38 @@ type GatewayClassSpec struct {
...
 
+// GatewayClassConditionReason defines the set of reasons that explain why
+// a particular GatewayClass condition type has been raised.
+type GatewayClassConditionReason string
+
 const (
-	// GatewayClassConditionStatusAdmitted indicates that the GatewayClass
-	// has been admitted by the controller set in the `spec.controller`
+	// This condition indicates whether the GatewayClass
+	// has been admitted by the controller requested in the `spec.controller`
```
robscott, line 93: Nit: Indentation feels a bit off here.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -80,30 +80,38 @@ type GatewayClassSpec struct {
...
+	// This condition defaults to False, and MUST be set by a controller when it sees
 	// a GatewayClass using its controller string.
 	// The status of this condition MUST be set to true if the controller will support
 	// provisioning Gateways using this class. Otherwise, this status MUST be set to false.
 	// If the status is set to false, the controller SHOULD set a Message and Reason as an
 	// explanation.
-	// GatewayClassNotAdmittedInvalidParameters is provided as an example Reason.
+	//
+	// Controllers should prefer to use the values of
+	// [GatewayClassConditionReason](#networking.x-k8s.io/v1alpha1.GatewayClassConditionReason")
```
robscott, line 104: I'd prefer not to include markdown links in our godocs at this point, or at least get the rest of this PR in before figuring out what to do here. I think the ideal would be to update our docs generation tool to recognize these types references instead of adding manual links like this.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -115,6 +123,10 @@ type GatewayClassStatus struct {
 	// Conditions is the current status from the controller for
 	// this GatewayClass.
 	//
+	// Controllers should prefer to publish Conditions using values of
+	// [GatewayClassConditionType](#networking.x-k8s.io/v1alpha1.GatewayClassConditionType")
```
robscott, line 127: Same comment as above - I'd personally prefer to avoid manually adding markdown links manually here.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -38,7 +38,7 @@ type TLSRoute struct {
 	Status TLSRouteStatus `json:"status,omitempty"`
 }
 
-// TLSRouteSpec defines the desired state of TLSRoute
+// TLSRouteSpec defines the desired state of TLSRoute resource.
```
robscott, line 41: Maybe this helps?
```suggestion
// TLSRouteSpec defines the desired state of a TLSRoute resource.
```

robscott (commented):
Thanks for your work on this @jpeach! Lots of great cleanup + godoc improvements here. I'd personally prefer to separate out the markdown link additions/discussion so we can get the rest of this cleanup in sooner.

apis/v1alpha1/gateway_types.go
jpeach: Here's an example of the current docs:

Screen Shot 2021-01-17 at 11 40 11 am

Looking at "NamedAddress", which is a designated enum value that API clients have to provide, it is described in terms of a "NamedAddressType", a term which is never defined in the documentation and is utterly irrelevant to anyone who doesn't have the Go code in front of them.

In terms of the API, the fact that this constant is available in Go doesn't really matter. What API consumers need to know is the *value* of the constant.

apis/v1alpha1/tlsroute_types.go
jpeach: fixed

apis/v1alpha1/gateway_types.go
jpeach: Pretty sure I addressed all the cases of this.

apis/v1alpha1/gatewayclass_types.go
jpeach: Dropped.

apis/v1alpha1/gateway_types.go
```diff
@@ -134,7 +134,7 @@ type GatewaySpec struct {
 // validating webhook.
 type Listener struct {
 	// Hostname specifies the virtual hostname to match for protocol types that
-	// define this concept. When unspecified, "", or "*", all hostnames are
+	// define this concept. When unspecified, "", or "&ast;", all hostnames are
```
robscott, line 137: This change feels strange, is it required for the reference docs?

apis/v1alpha1/gateway_types.go
```diff
@@ -147,8 +147,8 @@ type Listener struct {
 	//
 	// Hostname can be "precise" which is a domain name without the terminating
 	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
-	// domain name prefixed with a single wildcard label (e.g. "*.example.com").
-	// The wildcard character '*' must appear by itself as the first DNS label
+	// domain name prefixed with a single wildcard label (e.g. "&ast;.example.com").
+	// The wildcard character '&ast;' must appear by itself as the first DNS label
```
robscott, line 151: Same comment as above.

apis/v1alpha1/gateway_types.go
```diff
@@ -325,7 +313,9 @@ type GatewayTLSConfig struct {
 	// string for both the group and the resource, the resource defaults to "secrets".
 	// An implementation may support other resources (for example, resource
 	// "mycertificates" in group "networking.acme.io").
+	//
 	// Support: Core (Kubernetes Secrets)
+	//
```
robscott, line 318: What do these added line breaks do?

apis/v1alpha1/gateway_types.go
```diff
@@ -466,56 +448,47 @@ type RouteNamespaces struct {
...
 	//
 	// Support: Extended
 	//
 	// +kubebuilder:default=IPAddress
 	Type AddressType `json:"type,omitempty"`
 
-	// Value. Examples: "1.2.3.4", "128::1", "my-ip-address". Validity of the
-	// values will depend on `Type` and support by the controller.
+	// Value of the address. The validity of the values will depend
+	// on the type and support by the controller.
```
robscott, line 459: Nit: I think the backticks and capitalization above made it more clear that "type" was referring to an API field and not a characteristic of the controller.

apis/v1alpha1/gateway_types.go
```diff
@@ -466,56 +448,47 @@ type RouteNamespaces struct {
...
 
 const (
-	// IPAddressType a textual representation of a numeric IP
-	// address. IPv4 addresses must be in dotted-decimal
-	// form. IPv6 addresses must be in a standard IPv6 text
-	// representation (see RFC 5952).
-	//
-	// If the requested address is unsupported, the controller
-	// should raise the "Detached" listener status condition on
-	// the Gateway with the "UnsupportedAddress" reason.
```
robscott, line 507: This seems like an important removal, what's the reason for it?

apis/v1alpha1/gateway_types.go
```diff
@@ -466,56 +448,47 @@ type RouteNamespaces struct {
...
+	// A textual representation of a numeric IP address. IPv4
+	// addresses must be in dotted-decimal form. IPv6 addresses
+	// must be in a standard IPv6 text representation
+	// (see [RFC 5952](https://tools.ietf.org/html/rfc5952)).
```
robscott, line 481: I'm not sold on markdown links in Go comments. I admittedly don't mind this as much as relative links to parts of reference docs that would be useless elsewhere.

apis/v1alpha1/gateway_types.go
```diff
@@ -466,56 +448,47 @@ type RouteNamespaces struct {
...
 	//
-	// Support: Extended.
+	// Support: Extended
 	IPAddressType AddressType = "IPAddress"
 
-	// NamedAddressType is an opaque identifier. The interpretation
-	// of the name is dependent on the controller. If a NamedAddress is
-	// requested but unsupported by an implementation, the controller
-	// should raise the "Detached" listener status condition on the Gateway
-	// with the "UnsupportedAddress" reason.
```
robscott, line 516: Why remove the note about controller status condition here?

apis/v1alpha1/gateway_types.go
```diff
@@ -584,26 +557,25 @@ const (
...
+	// is not scheduled because there is no controller that recognizes
 	// the GatewayClassName. This reason should only be set by
 	// a controller that has cluster-wide visibility of all the
 	// installed GatewayClasses.
 	GatewayReasonNoSuchGatewayClass GatewayConditionReason = "NoSuchGatewayClass"
 
-	// GatewayReasonNoResources is used when the Gateway is
-	// not scheduled because no infrastructure resources are
-	// available for this Gateway.
+	// This reason is used with with the "Scheduled" condition when
```
robscott, line 571:
```suggestion
	// This reason is used with the "Scheduled" condition when
```

apis/v1alpha1/gateway_types.go
```diff
@@ -584,26 +557,25 @@ const (
...
+	// the Gateway is not scheduled because insufficient infrastructure
+	// resources are available.
 	GatewayReasonNoResources GatewayConditionReason = "NoResources"
 )
 
 const (
-	// GatewayConditionReady indicates whether the Gateway is able
+	// This condition is true when the Gateway is epected to be able
```
robscott, line 578:
```suggestion
	// This condition is true when the Gateway is expected to be able
```

apis/v1alpha1/gateway_types.go
```diff
@@ -770,27 +741,28 @@ const (
 	// interoperability.
 	ListenerConditionResolvedRefs ListenerConditionType = "ResolvedRefs"
 
-	// ListenerReasonDegradedRoutes indicates that not all of the routes
-	// selected by this Listener could be configured. The specific reason
-	// for the degraded route should be indicated in the route's
-	// .Status.Conditions field.
+	// This reason is used with the "ResolvedRefs" condition
```
robscott, line 744: Is there any way we can make this linking more obvious/automatic/structured in the reference docs?

apis/v1alpha1/httproute_types.go
```diff
@@ -64,11 +64,12 @@ type HTTPRouteSpec struct {
 	//
 	// Hostname can be "precise" which is a domain name without the terminating
 	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is
-	// a domain name prefixed with a single wildcard label (e.g. "*.example.com").
-	// The wildcard character '*' must appear by itself as the first DNS
+	// a domain name prefixed with a single wildcard label (e.g. "&ast;.example.com").
+	// The wildcard character '&ast;' must appear by itself as the first DNS
```
robscott, line 68: Just trying to capture all changes from "*" in case we miss any.

apis/v1alpha1/httproute_types.go
```diff
@@ -64,11 +64,12 @@ type HTTPRouteSpec struct {
...
 	// label and matches only a single label.
-	// You cannot have a wildcard label by itself (e.g. Host == "*").
+	// You cannot have a wildcard label by itself (e.g. Host == "&ast;").
```
robscott, line 70: Just trying to capture all changes from "*" in case we miss any.

apis/v1alpha1/httproute_types.go
```diff
@@ -119,7 +120,9 @@ type RouteTLSConfig struct {
 	// string for both the group and kind, the resource defaults to "secrets".
 	// An implementation may support other resources (for example, resource
 	// "mycertificates" in group "networking.acme.io").
+	//
 	// Support: Core (Kubernetes Secrets)
+	//
```
robscott, line 125: It would be good to understand the best practices around whitespace/formatting here and try to document them wherever possible.

hack/verify-golint.sh
```diff
@@ -25,6 +25,6 @@ cd "${KUBE_ROOT}"
 
 # See configuration file in ${KUBE_ROOT}/.golangci.yml.
 
-docker run --rm -v $(pwd):/app -w /app "golangci/golangci-lint:$VERSION" golangci-lint run
+docker run --rm -v $(pwd):/app -w /app "golangci/golangci-lint:$VERSION" golangci-lint run "$@"
```
robscott, line 28: This change feels a bit out of place in this PR - mind creating a separate one just for this?

robscott (commented):
Thanks for the work on this! A few more questions and nits, but lots of good improvements here. Still not sure about the transition away from the more traditional `Type is ...` structure. Added something to agenda tomorrow to try to get some consensus on which structure we want to standardize on throughout and how much we want to prioritize the readability of reference docs.

apis/v1alpha1/httproute_types.go
jpeach: This is ultimately markdown, so a newline is needed to separate the support tag into its own paragraph. Otherwise it runs on from the descriptive text.

hack/verify-golint.sh
jpeach: yeh we can drop this

apis/v1alpha1/gateway_types.go
jpeach: In markdown the `*` triggers italicizing, so it never appears in the rendered API reference at all. This is the only workaround I could find. Happy to concede that it sucks :)

apis/v1alpha1/gateway_types.go
jpeach: Screen Shot 2021-01-20 at 3 54 59 pm

Fixes this run-on paragraph.

apis/v1alpha1/gateway_types.go
jpeach: it's already documented on `AddressType`

apis/v1alpha1/gateway_types.go
jpeach: We can manually link it. Could try using separate type aliases for the reasons for each condition type, but might be awkward/overkill.

apis/v1alpha1/gateway_types.go
robscott: Oh yep, that makes a lot of sense. Any chance we can update the generator to escape potential formatting like this? This looks awfully weird in a Go file and feels like something that could be misconstrued by someone referring to go types as reference.

apis/v1alpha1/httproute_types.go
hbagdi: What happens if we do `*` (code escape in markdown)?

danehans:
IMO we should prioritize documentation for godocs. Is it possible for this PR to focus on changes that are applicable to godoc and html? The more controversial changes can be discussed further in a separate PR.

hbagdi:
I think the minor intrusive changes are okay but the change of `*` to `&ast;` is taking a step too further as it makes godocs incomprehensible.

robscott:
Hey @jpeach, we discussed this at the community meeting today - here's a quick recap:

- We want to ensure that godocs remain readable and accurate
- We love the API reference docs and want to find a way to update the generator to escape special characters
- If we can't find a way to update the generator, we may need to evaluate using https://pkg.go.dev/sigs.k8s.io/service-apis/apis/v1alpha1 temporarily until we can find a better solution

I think there's general agreement that the best case scenario involves finding a way to update the generator so we can have both readable godocs and great reference docs. Happy to help figure that out, just not sure where to even start.

hack/verify-golint.sh
jpeach: dropped

apis/v1alpha1/httproute_types.go
jpeach: dropped

apis/v1alpha1/gateway_types.go
jpeach: We din't like manual linking and there's no reasonable way to automatically link AFAICT.

apis/v1alpha1/gateway_types.go
jpeach: That would be find if it was used consistently. We can make a separate pass to add that.

apis/v1alpha1/gateway_types.go
jpeach: fixed.

apis/v1alpha1/httproute_types.go
jpeach: Hard to tell from the diff but that paragraph is not removed; it's now generic to `AddressType` because any kind of address could be unsupported (at least in an API types sense).

apis/v1alpha1/gateway_types.go
jpeach: fixed-ish

apis/v1alpha1/gateway_types.go
jpeach: fixed-ish

apis/v1alpha1/gateway_types.go
jpeach: seems a bit mean to force people to google this themselves.

apis/v1alpha1/gatewayclass_types.go
jpeach: That would be clear if it was done consistently. Maybe we can make a separate pass to do that everywhere.

hack/api-docs/generate.sh
```diff
@@ -50,4 +50,4 @@ fi
 gendoc::build
 gendoc::exec \
     -api-dir sigs.k8s.io/gateway-api/apis/v1alpha1 \
-    -out-file "$1"
+    -out-file /dev/stdout | sed '-es|&quot;*\*|\&quot;\&ast;|g' > "$1"
```
robscott, line 53: With https://github.com/kubernetes-sigs/gateway-api/pull/549 merged I don't think this and the associated changes to escaping asterisks are necessary anymore. I think it's still worth looking into ways to improve our docs generation, but what we have now is probably good enough for the v0.2.0 release.

robscott (commented):
Thanks for all your work on this! This LGTM, just needs one more rebase.

hbagdi:
#552 merged instead
/close

PR #514, opened 2021-01-04
More detailed and holistic documentation for Gateway<->Route binding

mark-church:
**What type of PR is this?**
/kind documentation

In our Alpha tests of the Gateway controller on GKE commonly the first questions/confusions we are getting from field teams are about how Routes bind to Gateways. Since this is a brand new concept it has not been intuitive to new users of how this works so I wanted to document Gateway and Route binding more clearly.

I chose not to break apart the explanation to the dedicated HTTPRoute and Gateway resource pages. Instead Gateway-Route binding is handled in the API Overview page 1) because it is a fundamental concept to understanding how Service APIs works and 2) because it lies at the intersection of both resource types and splitting the content across multiple pages would only make it more confusing.

I also wrapped all the content in this page to 80 chars because some of the original page was overflowing. The only new content is under `### Gateway and Route binding`.

@robscott @bowei @yiyangy @hbagdi @danehans


docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+
+![schema](images/gateway-route-binding.png)
+
+There is a lot of flexibility in how Routes can bind to Gateways to achieve
+different organizational policies and scopes of responsibility. These are 
+different relationships that Gateways and Routes can have:
+
+- **One-to-one** - A Gateway and Route may be deployed and used by a single 
+  owner and have a one-to-one relationship. Team C is an example of this. This 
+  is also how the Ingress resource works.
```
hbagdi, line 127: This statement has the potential for confusion. Some of the Ingress controller take multiple (mostly all) Ingress resource of a specific class and configure a single proxy. I understand that cloud LBs don't do this, it would be best to remove this.

docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+- **One-to-many** - A Gateway can have many Routes bound to it that are owned by
+  different teams from across different Namespaces. Teams A and B are an example 
+  of this.
+- **Many-to-one** - Routes can also be bound to more than one Gateway, allowing 
+  a single Route to control application exposure simultaneously across different
+  IPs, load balancers, or networks.
+
+*In summary, Gateways select Routes and Routes also select Gateways. When both 
+select each other, and there are no conflicts between existing bound Routes,
+then a Route will bind to a Gateway. When Routes are bound to a Gateway, it
```
hbagdi, line 137: > there are no conflicts between existing bound Routes

I think we should remove the conflict part. Conflict resource is not trivial, especially with HTTP paths and I expect implementers to follow our guidelines on a best-effort basis.

docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+means their collective routing rules are configured on the underlying load
+balancers or proxies that are managed by that Gateway. Thus, a Gateway is a
+symbolic representation of a networking data plane that can be configured
```
hbagdi, line 140: nit: maybe `s/symbolic/logical`

docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+through Routes.*
+
+#### Route Selection
+
+The primary way in which Gateways select Routes (effectively allowing them to
+bind) is via their Route resource labels. Routes are actually bound to specific
```
hbagdi, line 146: > via their Route resource labels

This is a confusing sentence.

Maybe something along the lines of  "A Gateway selects routes based on metadata of the Route resources - specifically, the namespace, kind and labels."

docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+listeners within the Gateway so each listener has a `listener.routes` field
+which selects Routes by different criteria. There are three criteria considered
+for a Gateway to select Routes:
+
+- **Route label** - A Gateway selects Routes via labels that exist on the 
+resource (similar to how Services select Pods via Pod labels).
+- **Route type** - A Gateway listener can only select a single type of Route
```
hbagdi, line 153: nit: How about "Route Kind"?
We have chosen to use `kind` in all of our documentation and API as that's familiar to most users.

docs-src/api-overview.md
```diff
@@ -79,6 +96,128 @@ BackendPolicy. For more information on what may be configured with this resource
...
+to the Route label and Namespace selectors on the Gateway. 
+- `SameNamespace` only allows this Route to bind with Gateways from the 
+same Namespace.
+- `FromList` allows an explicit list of Gateways to be specifiied that a 
+Route will bind with. `gateways.gatewayRefs` is only supported with this option. 
+
+The following Route can only bind with  `foo-gateway` in the `foo-namespace`.
+
+```yaml
+kind: HTTPRoute
```
hbagdi, line 198: A minor comment: How about we add a namespace to this route? That will further clarify that the list can contain references to Gateways from a different namespace.

jpeach:
@mark-church WDYT about submitting the wrapping change as a separate PR that we can land immediately?

docs-src/api-overview.md
mark-church: yeah good point - I removed the comparison with Ingress

docs-src/api-overview.md
mark-church: good idea - I added that

docs-src/api-overview.md
mark-church: makes sense - I've changed these headings to Label, Kind, and Namespace

docs-src/api-overview.md
```diff
@@ -79,6 +96,132 @@ BackendPolicy. For more information on what may be configured with this resource
...
+  - routes:
+      kind: HTTPRoute
+      selector:
+        matchLabels:
+          gateway: prod-web-gw 
+      namespaces:
+        from: All
+```
+
+#### Gateway Selection
```
robscott, line 181:
```suggestion
#### Limiting which Gateways can select a Route
```

docs-src/api-overview.md
```diff
@@ -79,6 +96,132 @@ BackendPolicy. For more information on what may be configured with this resource
...
+    as this Gateway will be selected. 
+    - `All` will select Routes from all Namespaces.
+    - `Selector` means that Routes from a subset of Namespaces selected by a 
+    Namespace label selector will be selected by this Gateway. When `Selector` 
+    is used then the `listeners.routes.namespaces.selector` field can be used to 
+    specify label selectors. This field is not supported with `All` or 
+    `SameNamespace`.
+
+The below Gateway will select all HTTPRoute resources with the `gateway:
+prod-web-gw` across all Namespaces in the cluster.
```
robscott, line 165: I think using the `gateway` label in this example could add further confusion. Maybe instead we could use a common k8s label like `app: web` and/or `traffic: external`.

docs-src/api-overview.md
```diff
@@ -79,6 +96,132 @@ BackendPolicy. For more information on what may be configured with this resource
...
+selectors that do not match any existing Route then nothing will bind to it even
+if a Route's `spec.gateways.allow = All`. Similarly, if a Route references a
+specific Gateway, but that Gateway is not selecting the Route's Namespace, then
+they will not bind. A binding will only take place if both resources select each
+other.
+
+It may not always be apparent from the resource specifications which Gateways
+and Routes are bound, but binding can be determined from the resource status.
+The [Route status](httproute.md#routestatus) will list all of the Gateways that
+a Route is bound to and any relevant conditions for the binding.
```
robscott, line 223: This is a tiny nitpick that may not be worth adding, but technically it's the first 100 Gateways that select a Route. (https://github.com/kubernetes-sigs/service-apis/blob/master/apis/v1alpha1/shared_types.go#L198-L202)

robscott (commented):
Thanks so much for your work on this! This will be a massive improvement to one of our most fundamental and complex concepts in this API.

GnatorX (commented):
Just small nitpicks

docs-src/api-overview.md
```diff
@@ -79,6 +96,132 @@ BackendPolicy. For more information on what may be configured with this resource
...
+- **One-to-one** - A Gateway and Route may be deployed and used by a single 
+  owner and have a one-to-one relationship. Team C is an example of this. 
+- **One-to-many** - A Gateway can have many Routes bound to it that are owned by
+  different teams from across different Namespaces. Teams A and B are an example 
+  of this.
+- **Many-to-one** - Routes can also be bound to more than one Gateway, allowing 
+  a single Route to control application exposure simultaneously across different
+  IPs, load balancers, or networks.
+
+*In summary, Gateways select Routes and Routes also select Gateways. When both
```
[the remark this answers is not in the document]
GnatorX: For my own interest, since I am still new to looking at service API. Is right to think that a Gateway defines a behavior it offers and Routes will "choose" which Gateway, with appropriate behavior, the Route logic should apply to? For example, a Gateway that allows public ingress traffic will label itself as public-gw and if you want a network path that can route public traffic to your Service you will label the Route as such so that the Gateway will pick it up?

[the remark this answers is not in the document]
mark-church: @GnatorX that is correct, a Gateway implements behavior which is defined by a GatewayClass. The characteristics of the GatewayClass define how the Route is exposed (on what infrastructure, where in the network, with what capabilities etc).

@robscott I agree that the way we show Route-to-Gateway selection by using labels that represent one or more Gateways is indirect. However, even though the Gateways define the labels, and Route owners may have no awareness of how these labels map to actual Gateways, the act of putting a label on the Route to influence which Gateways a Route will bind to, is an act of the Route owner selecting the Gateways.

I still think that each persona things of themselves as selecting or limiting Route binding. Gateway owners want to have the power to select or limit which Routes can bind. Route owners also want to have the power to select or limit which Gateways their Route will bind to. I don't think an app owner is comfortable just throwing their Route to the wind and not caring or knowing where their app will be exposed. The app owner wants to influence where and how their app will be exposed on networks, particularly if they have multiple Gateways available for them to choose from. 

Is there any reason why you don't think it makes sense to think of this as both Route->Gateway selection and Gateway->Route selection?

docs-src/api-overview.md
mark-church: I'll leave this detail to the HTTPRoute page since this is still the concepts page.

docs-src/api-overview.md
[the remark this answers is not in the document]
mark-church: @robscott coming back to this, I agree that technically speaking Routes do not select _specific_ Gateways, however I do believe the mental model of Route owners is that they are making an explicit decision that influences which Gateways expose their Routes.

To say that it's the Gateway limiting which Routes select, feels like we are explaining it in a more complicated way than the average user has to know about.

docs-src/api-overview.md
mark-church: There's two ways of going about this ...
- Binding via Gateway symbolic reference like `expose=prod-web-gw`. This feels like it would be useful for most kinds of deployments. Using a single key for Gateway selection does not work so well if Routes must be exposed by multiple Gateways though because you can't use the key more than once.
- Binding via Gateway characteristics like `network=internet && stage=prod`. This feels like it would be useful if you had many GatewayClasses and so you wanted to make it easier to pick one based on it's characteristics (rather than intuitively knowing what the characteristics of the class are intuitively). I'm not sure how necessary this is unless you have more than 5 Gateways available (which I feel is rare btw idk).
- Binding via Gateway symbolic reference like `prod-web-gw=true` or `prod-internal-gw=true` would allow a Route to be exposed by multiple Gateways.

robscott: @mark-church Thanks for the summary of potential approaches here! I've added something to the agenda for the community meeting tomorrow. I think it will be helpful to have some broader discussion around this.

mark-church: changed this to `expose=prod-web-gw` for the simplest approach that does not overload use of the term gateway

docs-src/api-overview.md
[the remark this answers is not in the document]
mark-church: Changed this to:

>In summary, Gateways select Routes and Routes allow or deny the selection. When both 
select each other then a Route will bind to a Gateway.

docs-src/api-overview.md
mark-church: Changing this to Route Exposure

docs-src/api-overview.md
```diff
@@ -96,6 +96,131 @@ BackendPolicy. For more information on what may be configured with this resource
...
+- **One-to-one** - A Gateway and Route may be deployed and used by a single 
+  owner and have a one-to-one relationship. Team C is an example of this. 
+- **One-to-many** - A Gateway can have many Routes bound to it that are owned by
+  different teams from across different Namespaces. Teams A and B are an example 
+  of this.
+- **Many-to-one** - Routes can also be bound to more than one Gateway, allowing 
+  a single Route to control application exposure simultaneously across different
+  IPs, load balancers, or networks.
+
+*In summary, Gateways select Routes and Routes control their exposure. When a Gateway selects a Route that allows itself to be exposed, then the Route will bind to the Gateway. When Routes are bound to a
```
robscott, line 135: Nit: The line wrapping is all so perfect in this PR except for here.

robscott (commented):
One tiny nit, otherwise LGTM. Thanks for the work on this!

hbagdi (approved):

docs-src/api-overview.md
mark-church: corrected this, hopefully good to go now!

robscott:
Thanks!

/lgtm
/approve
/hold cancel

PR #518, opened 2021-01-06
Fixes TLS Mode Defaulting

danehans:
**What type of PR is this?**
/kind bug

Properly sets the "Terminate" default for `gateway.spec.tls.mode`.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/517

Yes, sets the default TLS mode to "Terminate".

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -310,21 +310,30 @@ type GatewayTLSConfig struct {
 	// Mode defines the TLS behavior for the TLS session initiated by the client.
 	// There are two possible modes:
 	// - Terminate: The TLS session between the downstream client
-	//   and the Gateway is terminated at the Gateway.
+	//   and the Gateway is terminated at the Gateway. This mode requires
+	//   certificateRef to be set.
 	// - Passthrough: The TLS session is NOT terminated by the Gateway. This
 	//   implies that the Gateway can't decipher the TLS stream except for
 	//   the ClientHello message of the TLS protocol.
 	//   CertificateRef field is ignored in this mode.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:default=Terminate
 	Mode TLSModeType `json:"mode,omitempty"`
 
-	// CertificateRef is the reference to Kubernetes object that
-	// contain a TLS certificate and private key.
-	// This certificate MUST be used for TLS handshakes for the domain
-	// this GatewayTLSConfig is associated with.
-	// If an entry in this list omits or specifies the empty
-	// string for both the group and the resource, the resource defaults to "secrets".
-	// An implementation may support other resources (for example, resource
+	// CertificateRef is the reference to Kubernetes object that contain a
+	// TLS certificate and private key. This certificate MUST be used for
+	// TLS handshakes for the domain this GatewayTLSConfig is associated with.
+	//
+	// This field is required when mode is set to "Terminate" (default) and
+	// optional otherwise.
+	//
+	// If an entry in this list omits or specifies the empty string for both
+	// the group and the resource, the resource defaults to "secrets". An
+	// implementation may support other resources (for example, resource
 	// "mycertificates" in group "networking.acme.io").
+	//
 	// Support: Core (Kubernetes Secrets)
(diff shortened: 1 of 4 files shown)


robscott:
Thanks! This LGTM other than all the boilerplate year changes. That should be resolved once https://github.com/kubernetes-sigs/service-apis/pull/519 is in.

hbagdi (approved):
LGTM. This will need a rebase once #519 is merged.

hbagdi:
/hold

Just to make sure this doesn't get merged.

hbagdi:
@danehans Can you rebase to resolve the conflict?
/hold cancel

robscott:
@danehans Thanks for making this PR! After a rebase this LGTM. One small nit - I think this is a user-facing change since we're adding a default value.

danehans:
@robscott I updated the PR description to include user-facing changes.

apis/v1alpha1/gateway_types.go
```diff
@@ -315,16 +315,25 @@ type GatewayTLSConfig struct {
...
+	// +kubebuilder:default=Terminate
 	Mode TLSModeType `json:"mode,omitempty"`
 
 	// CertificateRef is the reference to Kubernetes object that
 	// contain a TLS certificate and private key.
 	// This certificate MUST be used for TLS handshakes for the domain
 	// this GatewayTLSConfig is associated with.
+	//
+	// This field is required when mode is set to "Terminate" and optional
+	// otherwise.
```
hbagdi, line 330: I was about lgtm this when I recalled why this was *not* the case earlier:
We wanted the user to explicitly set the condition to `Terminate` and make them supply the Certificate in that case.
In this case, if this default kicks in, and a CertificateRef is absent, then how do we handle this case?
We could add validation in the webhook for this case.

Do we still want this to default though?

robscott: That's a very good point - thanks for catching this! I think that approach makes sense, maybe instead of setting a default here we should clarify that both a CertificateRef and mode must be set as part of TLS config?

danehans: @hbagdi @robscott if ^ is the case, why not default `Mode` to "Passthrough" so `CertificateRef` can continue to be optional?

hbagdi: Because `Passthrough` is the rare case. It only makes sense with TCPRoute and TLSRoute (sni sniffing).

danehans: If `Mode` is optional, then a user can supply a `CertificateRef ` without setting `Mode`, resulting in an unusable TLS config. IMO, if "Terminate" is the common use-case, then it should be the default and the details documented. @hbagdi @robscott @bowei thoughts?

hbagdi: The default of `Terminate` makes sense here. What should be the error condition that should be set when CertificateRef is not supplied?

robscott: Just to make sure I'm understanding this correctly, this default will not come into play unless someone specifies TLS config. It will essentially allow a shortcut where most users would only need to specify CertificateRef here. Does that sound right? If so, I think this is a very reasonable default.

@hbagdi That's a great question. I think our existing `InvalidCertificateRef` condition would mostly apply here, even if the initial intent was slightly different.

hbagdi: > Just to make sure I'm understanding this correctly, this default will not come into play unless someone specifies TLS config. It will essentially allow a shortcut where most users would only need to specify CertificateRef here. Does that sound right?

That's correct. I pulled the patch down and tested it myself.
We actually missed specifying `mode` even in our example: https://github.com/kubernetes-sigs/service-apis/blob/27c4d2dcd3c70a65a1d512fc7a21d2ca4403d218/examples/tls-basic.yaml#L23-L29

> @hbagdi That's a great question. I think our existing InvalidCertificateRef condition would mostly apply here, even if the initial intent was slightly different.

Sounds fair. If the overhead is not too much, I'd advocate for a new AbsentCertificateRef or a generic AbsentRef condition.

hbagdi:
LGTM.
Leaving final approval to Rob.

PR #519, opened 2021-01-06
Removing year from boilerplate

robscott:
**What type of PR is this?**
/kind cleanup

This will prevent unnecessary changes and confusion each time we enter a new year. This matches what other similar projects have already done. This should unblock https://github.com/kubernetes-sigs/service-apis/pull/518 and https://github.com/kubernetes-sigs/service-apis/pull/513. More context in https://github.com/kubernetes-sigs/service-apis/pull/513#discussion_r551608574.

```release-note
NONE
```

/cc @danehans @jpeach


hbagdi:
/lgtm cancel

So it seems that we will need to remove the year check from `boilerplate.py`.
https://github.com/kubernetes-sigs/service-apis/blob/master/hack/boilerplate/boilerplate.py#L107

robscott:
@hbagdi Should be fixed now - had to remove year from all our license headers, not just the generated ones.

hbagdi:
> @hbagdi Should be fixed now - had to remove year from all our license headers, not just the generated ones.

But is that legally accurate? The common practice is that source code files should have the year in which they were first authored. I'm not a lawyer and I don't want to make this more annoying than it already is.

danehans:
`error: unable to recognize "examples/basic-udp.yaml": no matches for kind "UDPRoute" in version "networking.x-k8s.io/v1alpha1"`

I've seen this test flake locally. If it becomes more common, we may want to sleep a few seconds between installing the CRDs and installing the examples.

/retest

jpeach:
> > @hbagdi Should be fixed now - had to remove year from all our license headers, not just the generated ones.
> 
> But is that legally accurate? The common practice is that source code files should have the year in which they were first authored. I'm not a lawyer and I don't want to make this more annoying than it already is.

It's not necessary

https://www.apache.org/legal/src-headers.html
https://www.gnu.org/licenses/gpl-howto.en.html

robscott:
Quick update: I've copied the approach from Kubernetes upstream directly with the same generatego template and boilerplate.py they use to ignore generated files. Given that this will update our approach to match upstream, I think it's safe to remove the hold on this PR.

/hold cancel

hack/boilerplate/boilerplate.py
```diff
@@ -183,19 +204,25 @@ def get_files(extensions):
...
+    regexs["date"] = re.compile(get_dates())
     # strip // +build \n\n build constraints
     regexs["go_build_constraints"] = re.compile(
         r"^(// \+build.*\n)+\n", re.MULTILINE)
-    # strip #!.* from shell scripts
+    # strip #!.* from scripts
     regexs["shebang"] = re.compile(r"^(#!.*\n)\n*", re.MULTILINE)
+    # Search for generated files
+    regexs["generated"] = re.compile('DO NOT EDIT')
     return regexs
```
hbagdi, line 228: Rob, a lot of changes in this file seem unrelated. Did you copy this _almost_ verbatim from somewhere specific in upstream? If yes, should we include a comment at the top of the file to point to the original source for future-selves?

robscott: Good call, I've added a comment at the top of this file with a link to the specific upstream file this was copied from: https://github.com/kubernetes/kubernetes/blob/04c2b1fbdc1289c9a72eda87cf7072346e60d241/hack/boilerplate/boilerplate.py

robscott:
@hbagdi Thanks for the suggestion! I've added a link to the upstream file - PTAL

PR #525, opened 2021-01-12
Clarifying what should happen when Gateway hostname is ""

robscott:
**What type of PR is this?**
/kind cleanup

We want this field to be optional as a pointer because it will be irrelevant for many protocols. That leaves us with three potential states, and this comment only covered two of them. Thanks to @yiyangy for finding this one!

```release-note
Clarifying that an empty Gateway hostname should be treated the same as an unspecified value.
```

Note: This is based on #519 since anything else will fail tests or result in an excessively large diff. Holding this PR until that merges. Please look at the second commit of this PR for the changes that are unique here.

/hold


hbagdi:
A rebase should get this merged in.

robscott:
Thanks!

/hold cancel

PR #526, opened 2021-01-12
disable verbose logging for verify target

### LINKED ISSUE #522: Make it more obvious why pull-service-apis-verify failed
An easy way to see why this test failed.

**Why is this needed**:
When tests fail, it is rather complex to understand which verify script actually failed and why. Any improvements would be greatly appreciated.

hbagdi:
While verbose output is usually helpful, in the current output in CI, it
is so verbose that it is hard to make out the exact cause of a failure.

**What type of PR is this?**

/kind cleanup

Improve developer experience of debugging CI failures.

Fixes #522

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ uninstall:
 # Run static analysis.
 .PHONY: verify
 verify:
-	hack/verify-all.sh -v
+	hack/verify-all.sh
 
 # Build the documentation.
 .PHONY: docs


hbagdi:
This certainly results in loss of information but it makes it easy to determine which script failed.

robscott:
Thanks!

/lgtm

PR #528, opened 2021-01-13
Updates RouteStatus Godocs

danehans:
**What type of PR is this?**
/kind documentation

Updates godocs for a `RouteGatewayStatus`.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/313

```release-note
No (only godoc updates).
```


apis/v1alpha1/shared_types.go
```diff
@@ -182,6 +182,7 @@ type RouteGatewayStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:default={{type: "Admitted", status: "False", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}
```
robscott, line 185: Wouldn't this RouteGatewayStatus object only exist if it was added by a controller to begin with?

hbagdi: That sounds right. Do we have a bug in the API then? There should be a way to reflect this default status.

robscott: This specific object only gets added when a Route binds to a Gateway. That is logic controllers will be responsible for. I think the current empty default state here for the parent list is correct - no Gateways are using this Route. When that changes, it will because a controller is connecting the two, and at that point it becomes the controller's responsibility to set this status for the Gateway(s) it is managing. On that note, the associated godoc here could really use some more detail around what we're expecting of controllers here.

hbagdi: My last comment was ambiguous. What I mean is, when a Route is created and no controller has admitted it, should there not be a default condition on the route which communicates the same to the user?
We can educate the users to think of a Route without any conditions equivalent to a Route not yet admitted. I think that might trip some users.

robscott: If I'm understanding the structure correctly, the current default status for all routes would be:

```yaml
status:
  gateways: []
```

That seems like a reasonable starting point. Although it's not completely obvious, I think that many will be able to interpret that as "no Gateways are using this". 

I think the only way we could make this more clear would be to add a separate "conditions" list that could have a default value that indicated the route was not in use. Unfortunately with any conditions applied to the route as a whole it becomes rather difficult to determine which controller is responsible for updating them, since 0-multiple controllers may be processing the same route at any moment in time. 

If we did add that default condition, we could state that the first controller to add a gateway to status would be responsible for removing the condition. It becomes messier on removal though, especially if multiple controllers are removing Gateways at the same time.

hbagdi: Hmm, now I recall that we had a similar discussion before. Maybe we should explicitly document that no status means the route has not been admitted. Far from perfect, but it is something.

robscott: Agree with that, I think that's probably the best option we have here. What do you think @danehans?

youngnick: Swinging back in after having been away for a while, sorry!

For Conditions, it's really supposed to be that the default value is "Unknown", where that means "no controller has claimed this yet".

In this case, I think that we should both document that the empty slice means "not owned by anything", but default the Conditions slice to have one entry of `type: Admitted`, `status: Unknown`, with the current Reason of `To be updated by the owning controller` or something similar? That way, once a controller does add a RouteGatewayStatus, the `Admitted` condition will be there by default.

robscott: Welcome back! This is an interesting situation because it seems like the only entity that would add a Gateway to this list would be the owning controller. So a default would only apply when that controller was adding the Gateway to the list, and hopefully also setting a value inside conditions. If the controller did not set any value in conditions, is it more confusing for the slice to be empty or to have a "To be updated by the owning controller" condition despite actually being set by the owning controller? Neither option is great, but I'm not sure a default value here is very helpful. Maybe we need to provide clearer guidance that a condition must be set here? We could even add validation that ensured at least 1 condition was set when a Gateway was added here.

danehans: > Agree with that, I think that's probably the best option we have here. What do you think @danehans?

I think a reasonable next step is to update the godocs to explain that an empty `status.gateways` means the route has not been admitted by any gateway.

danehans:
@robscott @hbagdi @youngnick commit `486c90a` implements https://github.com/kubernetes-sigs/service-apis/pull/528#discussion_r569660410.

robscott:
Follow up from community meeting - we decided we should also make some changes to `RouteGatewayStatus` as part of this:

- We should add min length validation for conditions to ensure at least 1 is always specified
- We should update the godocs to note that the "Admitted" condition must always be specified by controllers

apis/v1alpha1/shared_types.go
```diff
@@ -181,24 +181,27 @@ type RouteGatewayStatus struct {
...
-	// included in the list.
-	//
+	// Gateways is a list of Gateways that are associated with the route,
+	// and the status of the route with respect to each Gateway. When a
+	// Gateway selects this route, the controller that manages the Gateway
+	// must add an entry to this list when the controller first sees the
+	// route and should update the entry as appropriate when the route is
+	// modified.
+	//
+	// A minimum of 1 and a maximum of 100 Gateways will be represented
```
robscott, line 199: I don't think we can require a minimum at this level. A Route may not be selected by any Gateways.

apis/v1alpha1/shared_types.go
```diff
@@ -181,24 +181,27 @@ type RouteGatewayStatus struct {
...
+	// in this list. If the list is full, additional Gateways may be
+	// using this Route that are not included in the list. An empty list
+	// means the route has not been admitted by any Gateway.
+	//
+	// +kubebuilder:validation:MinItems=1
```
robscott, line 204: Same comment about minimum.

apis/v1alpha1/shared_types.go
```diff
@@ -181,24 +181,27 @@ type RouteGatewayStatus struct {
 	//
```
robscott, line 181: > For example, the "Admitted" condition indicates whether the route has been admitted or rejected by the Gateway, and why.

Instead of the comment above, maybe we should say something like this::

> Controllers must include an "Admitted" condition to indicate whether the route has been admitted or rejected by the Gateway.

robscott (commented):
Thanks for the work on this @danehans!

apis/v1alpha1/shared_types.go
danehans: @robscott if that's the case, then I would expect `status.gateways` to be optional or `minItems=1` with a default that includes a nil gatewayRef and a default status condition. For example:
```
status:
  gateways:
  - conditions:
     - type: Admitted 
        status: False
        message: Waiting for gateway
        reason: waiting

File: apis/v1alpha1/shared_types.go (hunk 3 above)
Comment (danehans) [reply apis/v1alpha1/shared_types.go:204]: Same response as https://github.com/kubernetes-sigs/service-apis/pull/528#discussion_r571154934.

File: apis/v1alpha1/shared_types.go (hunk 4 above)
Comment (danehans) [reply apis/v1alpha1/shared_types.go:181]: > Controllers must include an "Admitted" condition to indicate whether the route has been admitted or rejected by the Gateway.

@robscott if we default the Admitted condition as described in https://github.com/kubernetes-sigs/service-apis/pull/528#discussion_r571154934, then I would expect the documentation to state:

> Controllers must update the "Admitted" condition to indicate whether the route has been admitted or rejected by the Gateway.

File: apis/v1alpha1/shared_types.go (hunk 2 above)
Comment (robscott) [reply apis/v1alpha1/shared_types.go:199]: I think the ideal empty state here is actually `gateways: []`. It does seem helpful to have some indication that no Gateways are using this route. With that said, I'm not sure if we could accomplish that though. Maybe a default `[]` combined with no `omitempty` would work? Or maybe this could be a "required" field with a min length of 0? 

I believe the example you have above would be invalid because GatewayRef is a required field. In my opinion, it does not make sense to make GatewayRef optional in a field that is all about describing the relationship of a route with a specific Gateway. If we did that, would a controller need to remove that when they were adding their own status? Would the controller need to add the empty status back if they were removing their entry in this list?

Comment (danehans) [reply apis/v1alpha1/shared_types.go:199]: > I believe the example you have above would be invalid because GatewayRef is a required field.

What if we default GatewayRef to "None". For example:
```
status:
  gateways:
  - gatewayRef:
     name: None
     namespace: None
    - conditions:
       - type: Admitted 
          status: False
          message: Waiting for gateway
          reason: waiting
```

an owning controller is responsible for updating the condition.

**Comment by danehans** [comment]
@robscott commit `5ce9b3a` resolves https://github.com/kubernetes-sigs/service-apis/pull/528#issuecomment-772801796, PTAL.

File: apis/v1alpha1/shared_types.go [hunk 5]
```diff
@@ -173,31 +173,34 @@ type RouteGatewayStatus struct {
...
-	// that the route's availability is also subject to the Gateway's own
-	// status conditions and listener status.
+	// Gateway. The "Admitted" condition must always be specified by controllers
+	// to indicate whether the route has been admitted or rejected by the Gateway,
+	// and why. Note that the route's availability is also subject to the Gateway's
+	// own status conditions and listener status.
 	//
 	// +listType=map
 	// +listMapKey=type
+	// +kubebuilder:validation:MinItems=1
```
robscott, line 185: I think this is technically a breaking change. Can we file a separate issue/PR to track this for v1alpha2? Admittedly this is super nitpicky and unlikely to actually break anyone at this point, but good to be as consistent as possible with backwards compatibility.

robscott (commented):
Thanks for the work on this @danehans! Despite what I said earlier, I think it would be safest to avoid adding a minLength until v1alpha2. Otherwise this LGTM.

apis/v1alpha1/shared_types.go
danehans: @robscott done, xref: https://github.com/kubernetes-sigs/service-apis/issues/542

robscott:
Thanks!

/lgtm

PR #532, opened 2021-01-19
Authorization HTTPFilter

### LINKED ISSUE #114: Pluggable access control

A standardized mechanism to call an external service to authorize request forwarding to a selected destination (backend).

**Why is this needed**:

## User Story

* As an application developer, I want to extract authentication and authorization requirements from an application to a common infrastructure component. The authentication and authorization function should have access to request headers and optionally the request body up to some limit.
* As a cluster operator, using an external call (akin in spirit to Kubernetes' ValidatingWebhooks) rather than a proxying service allows me to control the following policies:
   * Ability to separately measure latency/error rate
   * Ability to fail-open as well as fail-closed on error
   * Reduced complexity of authorization webhook

## Usage examples

* Cloud Foundry UAA for OAuth token verification would be one example, as would a SAML implementation.
* If TLS information is provided in the request, this might provide an implementation for #93 (enforce TLS validation for an application)
* Access control could also be used to implement a rate-limiting mechanism or load-shedding mechanism to prevent overloading backend services like a SQL database.
* Application services could also provide validation on other header fields; Knative event delivery (CloudEvents over HTTP) might want to be able to limit delivery to only certain event types (CE-Type header).

jmprusi:
**What type of PR is this?**
/kind design

This PR shows a possible new Filter design, called RequestAuthorization, that would be used to authorize incoming HTTP requests with an external authorization component.

Design doc: [Pluggable Access Control for HTTPRoute](https://docs.google.com/document/d/1pFsIPK9uOmnL9jRqRvcRceYIeeVY_TipQUVjXl-lkfY/edit)

Fixes #114

```release-note
Adds a new HTTPFilter: "RequestAuthorization", that allows users to authorize incoming HTTP request with an external authorization service.
```


apis/v1alpha1/httproute_types.go
```diff
@@ -532,6 +546,45 @@ type HTTPRequestMirrorFilter struct {
 	Port PortNumber `json:"port"`
 }
 
+// HTTPRequestAuthorization defines configuration for the RequestAuthorization filter.
+type HTTPRequestAuthorization struct {
+	// URL indicates the address of the service used to perform authorization checks.
+	//
+	// There are two protocols supported:
+	// - grpc: In this mode, the authorization server should receive a grpc request of type <service.auth.v3.CheckRequest>.
+	//   It is expected for the authorization server to implement the Envoy ExtAuthz GRPC protocol:
```
howardjohn, line 555: Sorry if I missed some discussion about this, but I am surprised that a generic API is referencing a concrete implementation (envoy). Would we expect, say, an nginx implementor to use the envoy auth?

apis/v1alpha1/httproute_types.go
```diff
@@ -532,6 +546,45 @@ type HTTPRequestMirrorFilter struct {
...
+	//   https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto
+	//
+	// - https: An HTTP request that will include all the Headers and Request Path, but no body content.
+	//   It is expected for the authorization server to implement the Envoy ExtAuthz HTTP protocol:
+	//   https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto
+	//
+	URL string `json:"url"`
```
howardjohn, line 562: How does configuration of how to reach this URl work? For example, if we need authentication, TLS, etc.

if it were a k8s service it would be "easy" since existing APIs defined here like BackendPolicy would apply. For a generic URL, I am not sure its easy.

apis/v1alpha1/httproute_types.go
```diff
@@ -532,6 +546,45 @@ type HTTPRequestMirrorFilter struct {
...
+	//
+	// +optional
+	// +kubebuilder:default= 100
+	TimeoutMs int `json:"timeoutMs,omitempty"`
+
+	// Metadata can be used to provide additional context information for the authorization
+	// endpoint to make a decision.
+	//
+	// +optional
+	Metadata *map[string]string `json:"metadata,omitempty"`
```
howardjohn, line 585: a pointer to a map feels a bit odd, is this used in other parts of the API?

apis/v1alpha1/httproute_types.go
jmprusi: @howardjohn hi!  I'm planning on presenting a design proposal at tomorrow's community meeting, where I will try to explain some of those points to discuss them. 

While this is a concrete implementation, it is an open protocol that can be implemented by "any" authorization server. I agree that the grpc implementation is trickier to support in different proxies and perhaps not possible... But the HTTP version is very "straightforward" to implement in Openresty, and I'm currently working on an implementation for HAProxy.

apis/v1alpha1/httproute_types.go
jmprusi: Yes, that's a good point. Thinking out loud, perhaps splitting the URL into a `service reference` and a `type` or `protocol` field to define the type of authorization (grpc/HTTP) could work?

howardjohn: That could potentially work, I think some sort of reference is probably more consistent with the rest of the API as well. FWIW we recently went through lots of debate with https://docs.google.com/document/d/1V4mCQCw7mlGp0zSQQXYoBdbKMDnkPOjeyUb85U07iSI/ about this exact topic in Istio.

apis/v1alpha1/httproute_types.go
jmprusi: That doesn't look ok, I've changed this type from `string` to a `map[string]string` and left it there. Let me fix it. Thanks.

apis/v1alpha1/httproute_types.go
jmprusi: Thanks, @howardjohn! I will review that document and try to find a better API design.

apis/v1alpha1/httproute_types.go
```diff
@@ -532,6 +546,45 @@ type HTTPRequestMirrorFilter struct {
...
+type HTTPRequestAuthorization struct {
+	// URL indicates the address of the service used to perform authorization checks.
+	//
+	// There are two protocols supported:
+	// - grpc: In this mode, the authorization server should receive a grpc request of type <service.auth.v3.CheckRequest>.
+	//   It is expected for the authorization server to implement the Envoy ExtAuthz GRPC protocol:
+	//   https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto
+	//
+	// - https: An HTTP request that will include all the Headers and Request Path, but no body content.
+	//   It is expected for the authorization server to implement the Envoy ExtAuthz HTTP protocol:
```
jpeach, line 559: Not all HTTP external auth mechanisms are compatible with the envoy way. There's TLS and Host header considerations to take into account as well.

apis/v1alpha1/httproute_types.go
```diff
@@ -421,6 +428,13 @@ const (
 	// Support in HTTPRouteForwardTo: Extended
 	HTTPRouteFilterRequestMirror HTTPRouteFilterType = "RequestMirror"
 
+	// HTTPRouteFilterRequestAuthorization can be used to authorize a request with an
+	// external authorization endpoint before it is send to the upstream target.
```
danehans, line 432: s/it is send/being sent/

robscott:
Thanks for the work on this @jmprusi! Based on our conversation yesterday at the community meeting, it seems like there's broad support for an approach like this if we can make it portable. Ideally this filter should be usable with more than Envoy ExtAuthz. It seems like external auth is generally configured with the same approximate concepts:

- Protocol
- URL
- Timeout
- FailureMode

That seems rather similar to what is already proposed here. I'm hesitant about specifying which underlying approach an implementation of this API should take, but if we can avoid that, I think this has the potential to be a valuable and portable filter.

jmprusi:
> Thanks for the work on this @jmprusi! Based on our conversation yesterday at the community meeting, it seems like there's broad support for an approach like this if we can make it portable. Ideally this filter should be usable with more than Envoy ExtAuthz. It seems like external auth is generally configured with the same approximate concepts:
> 
> * Protocol
> * URL
> * Timeout
> * FailureMode
> 
> That seems rather similar to what is already proposed here. I'm hesitant about specifying which underlying approach an implementation of this API should take, but if we can avoid that, I think this has the potential to be a valuable and portable filter.


I've updated the PR, but I think we should discuss a couple of fields:

- **Metadata**: This could be a useful field to add. It will allow the user to add "hardcoded" specific information either for the controller or the authorization service. This can be implemented in the form of a Header for nginx or as context metadata in envoy.
- **RequestBody**: Send the original request body to the authorization service or not. This can be configured in Envoy or Nginx.

Another interesting "common" field could be:
- **ForwardHeaders**: A list of headers that will be forwarded to the authorization endpoint from the original request, instead of the "default" behavior of sending all of them or a known subset.

apis/v1alpha1/httproute_types.go
jmprusi: As discussed in the community meeting, it's better to avoid tying this filter to a specific implementation.

apis/v1alpha1/httproute_types.go
youngnick: Just getting to review this one now, Contour added an ExtensionService to cover the "service reference" part of this (see https://projectcontour.io/docs/v1.12.0/config/api/#projectcontour.io/v1alpha1.ExtensionService), and then an AuthorizationServer struct that holds the ExtensionService reference. (https://projectcontour.io/docs/v1.12.0/config/api/#projectcontour.io/v1.AuthorizationServer). Both the ExtensionService and AuthorizationServer have places to put config.

I'm not suggesting doing the same thing, but I think the idea of a service reference and a type or protocol field would be great.

jmprusi: @youngnick It already has a protocol field, but for the service reference, we can follow the example of the other filters and replace URL with:

```go
ServiceName *string `json:"serviceName,omitempty"`
BackendRef *LocalObjectReference `json:"backendRef,omitempty"`
Port *PortNumber `json:"port,omitempty"`
```

Or add a new structure that allows us to cross the namespace boundary, with additional information on how to connect to the external service, like for example a TLS client certificate to use.

apis/v1alpha1/httproute_types.go
```diff
@@ -560,6 +574,65 @@ type HTTPRequestMirrorFilter struct {
...
+	// +kubebuilder:default=100
+	TimeoutMs int `json:"timeoutMs,omitempty"`
+
+	// Metadata can be used to provide additional context information for the authorization
+	// service to make a decision.
+	//
+	// Support: Extended
+	//
+	// +optional
+	Metadata map[string]string `json:"metadata,omitempty"`
```
jpeach, line 633: AFAIK, only the envoy GRPC protocol can pass arbitrary key value pairs here. How do you indicate missing or partial support for this field?

apis/v1alpha1/httproute_types.go
```diff
@@ -560,6 +574,65 @@ type HTTPRequestMirrorFilter struct {
...
+	URL string `json:"url"`
+
+	// Protocol is used to define the Protocol that the gateway will use to communicate with the authorization
+	// service. If not set, it will default to "http" or inferred from the scheme in URL.
+	// Examples: "grpc","http2"
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:default=http
```
jpeach, line 592: Can we add enum validation for allowed values?

apis/v1alpha1/httproute_types.go
```diff
@@ -560,6 +574,65 @@ type HTTPRequestMirrorFilter struct {
...
+	Protocol string `json:"protocol,omitempty"`
```
jpeach, line 593: For protocols over TLS, is there a way to tweak the server certificate validation?

apis/v1alpha1/httproute_types.go
jmprusi: I think this is for the implementor to decide; for example, in Nginx, this could be implemented as a custom HTTP header sent to the authorization service. Perhaps we should put some validation in place to make that easier, like a max length and allowed characters.

If the implementation doesn't support this, it can reject the route and set the proper status condition.

apis/v1alpha1/httproute_types.go
jmprusi: Yes, that would great to have, but I don't see any example of this in the API. I looked at the `HTTPRouteForwardTo` struct, and it doesn't seem to support any TLS settings. All the other TLS settings seem to be focused on the "listener side," not starting a connection.

Any suggestion on how can we add this to the filter?

evankanderson:
(Looks like this got rebased a few weeks ago)

hbagdi:
I think it is unlikely that we will merge this. A custom filter would be the path forward here.
@robscott @jpeach @youngnick Thoughts?

robscott:
I tend to agree with that. If there's a way we can provide some form of portability for authorization here that would be awesome, but we've got so many things on the go right now I'm not sure how soon we'll be able to tackle that. It's likely easiest to start as a custom filter.

hbagdi:
Let the community experiment for now. If there is convergence, we would love to get this into this API itself.
/close

PR #534, opened 2021-01-23
Add link to "Kubernetes API Versioning" by using correct name

nak3:

[Current doc for Releases/Versioning](https://kubernetes-sigs.github.io/service-apis/releases/#versioning) misses the link for `[Kubernetes API Versioning] (example: v1alpha1)`.
This is because the link is for `Kubernetes API Versioning` does not
match to the `Kubernetes API Version`.

This patch fixes it.

**What type of PR is this?**

/kind documentation

```release-note
NONE
```

### DIFF
--- a/docs-src/releases.md
+++ b/docs-src/releases.md
@@ -44,7 +44,7 @@ fixes. New API versions will be released with new minor or major versions.
 Our changelog and release notes will always include both the semantic version
 and API version(s) included in the release. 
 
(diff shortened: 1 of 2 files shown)


hbagdi:
Pulled down and verified that the patch works.
/ok-to-test
/lgtm
/approve

@nak3 Thanks for fixing this!

PR #538, opened 2021-02-04
Aligns API Fields with K8s Conventions

danehans:
**What type of PR is this?**

/kind bug
/kind api-change

This PR does the following:

- Updates API fields to be consistent with Kubernetes API [conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md).
- Makes `httproute.spec.rules` optional since all subfields are optional and sets a default value.
- Makes the port field of `RouteForwardTo` optional. When unspecified, the port of the request will be used to forward to a backend. __Note:__ All other `RouteForwardTo` fields are optional.

Fixes https://github.com/kubernetes-sigs/service-apis/issues/537

```release-note
- The port field of `RouteForwardTo` is now optional.
- `httproute.spec.rules` is now optional. When unspecified, `kubebuilder:default={{matches: {{path: {type: "Prefix", value: "/"}}}}}` will be used.
```

xref: https://github.com/kubernetes/enhancements/pull/2366#discussion_r567007118

/assign @robscott @hbagdi 
/cc @bowei @youngnick @jpeach @stevesloka


danehans:
@robscott @cmluciano does the validating webhook ensure that a `RouteForwardTo` includes a serviceName or backendRef value? Note that both fields are optional but one is required to forward the request to the appropriate object.

apis/v1alpha1/gateway_types.go
```diff
@@ -293,8 +301,9 @@ type TLSOverridePolicy struct {
 	//
 	// Support: Core
 	//
+	// +optional
 	// +kubebuilder:default=Deny
-	Certificate TLSRouteOverrideType `json:"certificate"`
+	Certificate TLSRouteOverrideType `json:"certificate,omitempty"`
```
danehans, line 306: +optional/omitempty was added since this field is being defaulted.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -33,15 +33,19 @@ import (
 // for creating Gateway resources.
 //
 // GatewayClass is a Cluster level resource.
-//
-// Support: Core.
```
danehans, line 37: Support level is removed since no other resources specify it. Fields should indicate the support-level and not the top-level resource.

apis/v1alpha1/httproute_types.go
```diff
@@ -104,9 +114,10 @@ type HTTPRouteSpec struct {
 
 	// Rules are a list of HTTP matchers, filters and actions.
 	//
-	// +kubebuilder:validation:MinItems=1
+	// +optional
```
danehans, line 117: Set to optional since defaulting is now being used.

apis/v1alpha1/shared_types.go
```diff
@@ -62,13 +62,18 @@ type RouteGateways struct {
...
 	// +kubebuilder:validation:Enum=All;FromList;SameNamespace
 	// +kubebuilder:default=SameNamespace
 	Allow GatewayAllowType `json:"allow,omitempty"`
+
 	// GatewayRefs must be specified when Allow is set to "FromList". In that
 	// case, only Gateways referenced in this list will be allowed to use this
 	// route. This field is ignored for other values of "Allow".
+	//
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
danehans, line 76: Is 16 an appropriate maximum number of gatewayRefs?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -31,16 +31,25 @@ import (
 // implementation specific may be represented with similar implementation
 // specific custom resources.
 type BackendPolicy struct {
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+	// +optional
```
jpeach, line 35: why optional here?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -31,16 +31,25 @@ import (
...
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   BackendPolicySpec   `json:"spec,omitempty"`
+	// Spec defines the desired state of BackendPolicy.
+	//
+	// +optional
```
jpeach, line 40: What's the utility of a resource with an empty spec?

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -102,26 +121,28 @@ type BackendTLSConfig struct {
 	// Support: Extended
 	//
 	// +optional
-	CertificateAuthorityRef *LocalObjectReference `json:"certificateAuthorityRef,omitempty"`
+	CertificateAuthorityRef LocalObjectReference `json:"certificateAuthorityRef,omitempty"`
```
jpeach, line 124: this is optional, so should still be a pointer field?

apis/v1alpha1/gateway_types.go
```diff
@@ -34,18 +34,25 @@ import (
 // This ensures that a GatewayClass associated with a Gateway(s) is not
 // deleted while in use.
 type Gateway struct {
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+	// +optional
```
jpeach, line 38: is object meta really optional?

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -77,7 +81,7 @@ type GatewayClassSpec struct {
 	// Support: Custom
 	//
 	// +optional
-	ParametersRef *LocalObjectReference `json:"parametersRef,omitempty"`
+	ParametersRef LocalObjectReference `json:"parametersRef,omitempty"`
```
jpeach, line 84: optional should be pointer?

apis/v1alpha1/httproute_types.go
```diff
@@ -318,13 +333,14 @@ type HTTPRouteMatch struct {
...
 	//
+	// +optional
 	// +kubebuilder:default={type: "Prefix", value: "/"}
 	Path HTTPPathMatch `json:"path,omitempty"`
 
 	// Headers specifies a HTTP request header matcher.
 	//
 	// +optional
-	Headers *HTTPHeaderMatch `json:"headers"`
+	Headers HTTPHeaderMatch `json:"headers,omitempty"`
```
jpeach, line 343: optional should be a pointer

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

jpeach (changes requested):
Couple of broad issues for this change

- probably meta, spec and status should not be optional; that seems quite surprising to me
- optional field should consistently be pointer fields, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -55,33 +64,43 @@ type BackendPolicySpec struct {
 	// the oldest BackendPolicy.
 	//
 	// Support: Core
+	//
+	// +kubebuilder:validation:MinItems=1
```
robscott, line 68: I'm not sure we want this change. I think we want it to be possible to specify a BackendPolicy without applying it to any specific backends. This could be helpful if you want to provide a standard set of policies that can be applied to different backends, even if not all of them apply to a backend at once.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -55,33 +64,43 @@ type BackendPolicySpec struct {
...
 	// +kubebuilder:validation:MaxItems=16
 	BackendRefs []BackendRef `json:"backendRefs"`
 
 	// TLS is the TLS configuration for these backends.
 	//
 	// Support: Extended
+	//
 	// +optional
-	TLS *BackendTLSConfig `json:"tls,omitempty"`
+	TLS BackendTLSConfig `json:"tls,omitempty"`
```
robscott, line 77: Although the API conventions do have an allowance for types with a built-in nil value, it seems like the overwhelming preference is to use pointers for all optional fields if I'm reading the API conventions correctly.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -55,33 +64,43 @@ type BackendPolicySpec struct {
...
 }
 
 // BackendRef identifies an API object within a known namespace that defaults
 // group to core and resource to services if unspecified.
```
robscott, line 81: This comment does not agree with the spec below. I think I'd previously assumed this meant these values would be inferred by a controller, but now that we're formally marking both Group and Kind required, I think we should remove this as well.

apis/v1alpha1/gatewayclass_types.go
robscott: I think there may be a place for this, especially for route types, but I'm not exactly sure what that should be. I think we want a world where both L4 and L7 implementations can support different parts of this API. I don't mind removing this here though since it doesn't seem like we've added it for other resources.

robscott (commented):
Thanks for the work on this @danehans! Some of these changes were rather surprising, but the vast majority seem to line up with API Conventions. I was especially surprised to see all fields with a default value marked as optional. Even stranger to see meta, spec, and status all marked as optional. In all cases, that seems to line up with upstream conventions. Still feels awfully strange though. Will take a closer look soon, but wanted to add some comments from my first run through.

apis/v1alpha1/backendpolicy_types.go
danehans: @jpeach I'm following k8s api conventions. When a field includes `omitempty`, it should be tagged +optional.

xref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required
xref: https://github.com/kubernetes/api/blob/master/apps/v1/types.go

apis/v1alpha1/backendpolicy_types.go
danehans: Same as https://github.com/kubernetes-sigs/service-apis/pull/538/files#r571116077.

A potential use case is to use api defaulting for spec. A resource with all defaults can be created as such:
```
kind: Gateway
metadata:
  name: example
```

and will result in:
```
kind: Gateway
metadata:
  name: example
spec:
  gatewayClassName: default
  listeners:
  - protocol: HTTP
    port: 80
    routes:
      kind: HTTPRoute
      selector:
        matchLabels:
          app-type: http-exposed
      namespaces:
        from: "Same"
```

This could simplify the user experience for the basic http use case.

__Note:__ All these defaults above do not exist today.

apis/v1alpha1/backendpolicy_types.go
danehans: @robscott if that's the case, then shouldn't this field be optional? I do like your idea for a nil backendRefs, but the docs and field requirements indicate otherwise.

apis/v1alpha1/backendpolicy_types.go
danehans: > preference is to use pointers for all optional fields

Not all optional fields.

> They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) __or have a built-in nil value (e.g. maps and slices).__

We are inconsistently using pointers for struct fields.

apis/v1alpha1/backendpolicy_types.go
robscott: This is similar to my comment on the other issue, but in some cases I think it's helpful to actually show an empty list instead of just omitting the field entirely. I'm curious what the behavior is now, I'm assuming a user can create a BackendPolicy with `[]` for this field, do they even need to specify that? I don't read anything in the docs that specifically requires a BackendRef to be set, but we have not made it clear enough that empty is a valid option here, that would definitely be good to update.

apis/v1alpha1/shared_types.go
robscott: Good catch, this does feel like an appropriate limit.

apis/v1alpha1/backendpolicy_types.go
robscott: Yeah, the conventions are a bit confusing here. I completely agree that they allow for non-pointer optional fields in these scenarios, but the points below seem to suggest we should be using pointers for structs:

> * structs are not omitted from encoder output even where omitempty is specified, which is messy;
> * having a pointer consistently imply optional is clearer for users of the Go language client, and any other clients that use corresponding types

apis/v1alpha1/backendpolicy_types.go
danehans: Agreed.

apis/v1alpha1/backendpolicy_types.go
danehans: xref https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L50-L52 for other native k8s types with optional struct fields.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -102,26 +121,28 @@ type BackendTLSConfig struct {
...
 // BackendPolicyStatus defines the observed state of BackendPolicy. Conditions
-// that are related to a specific Route or Gateway should be placed on the
+// that are related to a specific Route or Gateway must be placed on the
 // Route(s) using backends configured by this BackendPolicy.
 type BackendPolicyStatus struct {
 	// Conditions describe the current conditions of the BackendPolicy.
 	//
 	// +listType=map
 	// +listMapKey=type
+	// +kubebuilder:validation:MinItems=1
```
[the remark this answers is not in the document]
danehans: Similar to https://github.com/kubernetes-sigs/service-apis/pull/528/files#r571154934. For example:
```
status:
- conditions:
  - type: Admitted 
     status: False
     message: Waiting for controller
     reason: waiting
```

Otherwise we make conditions optional and the lack of conditions means the object has not been reconciled by a controller. I prefer to be explicit instead of requiring the user to infer.

apis/v1alpha1/gateway_types.go
danehans: Yes, according to k8s api conventions, any field with `omitempty` should be marked as `+optional`.

xref: https://github.com/kubernetes/api/blob/master/apps/v1/types.go

apis/v1alpha1/gatewayclass_types.go
danehans: Per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required:
> They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) or _have a built-in nil value (e.g. maps and slices)._

Take a look through other areas of Service APIs and k8s native APIs and you will see:
```
// +optional
SomeField SomeFieldCustomType `json:"someField,omitempty"`
```

apis/v1alpha1/httproute_types.go
danehans: > They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) __or have a built-in nil value (e.g. maps and slices).__

apis/v1alpha1/gatewayclass_types.go
danehans: > They are a pointer type in the Go definition (e.g. AwesomeFlag *SomeFlag) __or have a built-in nil value (e.g. maps and slices).__

apis/v1alpha1/shared_types.go
danehans: @robscott @bowei @jpeach @hbagdi can I get feedback on ^?

apis/v1alpha1/gatewayclass_types.go
jpeach: I don't think that `ParametersRef` has an OK nil value; the empty string fields create ambiguity that could easily be avoided.

apis/v1alpha1/gateway_types.go
jpeach: Are there any object that do this? I don't recall seeing it before.

apis/v1alpha1/backendpolicy_types.go
jpeach: I wonder whether this UX would be better in a kubectl plugin? I'm a bit skeptical that defaulting at this layer is the right approach.

apis/v1alpha1/backendpolicy_types.go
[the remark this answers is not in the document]
jpeach: Yeh, I agree that there will always be at least one condition.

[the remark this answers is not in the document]
robscott: So to clarify, we're saying that status itself is optional, but if it is specified it must include at least one condition? I think status on BackendPolicy is much less clear than other resources. Controllers have clear ownership/management of Gateways and status on Routes is tied directly to the Gateways. That makes it clear which controller is responsible for setting conditions in each case.

Unfortunately with BackendPolicy we don't have a good answer to that. I don't think we ever had a great answer here, but even the godocs for this status suggest that most conditions should be set on route and not here. Given that, I'd rather mark this as optional than require a condition.

apis/v1alpha1/backendpolicy_types.go
hbagdi: Is it that the `omitempty` tag here is wrong instead?
How can `metav1.ObjectMeta` possibly be empty for any valid k8s resource?

apis/v1alpha1/backendpolicy_types.go
hbagdi: From an implementer's perspective, checking for a nil pointer to a struct is far less error prone than checking each member in that struct.

apis/v1alpha1/backendpolicy_types.go
hbagdi: > I think it's helpful to actually show an empty list instead of just omitting the field entirely.

+1. I'm not versed in conventions but from a user's perspective, this is very helpful for some fields.

apis/v1alpha1/gateway_types.go
```diff
@@ -34,18 +34,25 @@ import (
...
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// Spec defines the desired state of Gateway.
+	//
+	// +optional
 	Spec GatewaySpec `json:"spec,omitempty"`
```
hbagdi, line 44: I think it should be the other way around where we remove the `omitempty` tag. An object without a `spec` is going to confuse users and even maintainers.

apis/v1alpha1/httproute_types.go
```diff
@@ -27,16 +27,24 @@ import (
...
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta `json:",inline"`
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   HTTPRouteSpec   `json:"spec,omitempty"`
+	// Spec defines the desired state of HTTPRoute.
+	//
+	// +optional
+	Spec HTTPRouteSpec `json:"spec,omitempty"`
```
hbagdi, line 37: Same comment as before. We should remove `omitempty` tag from Spec everywhere rather than adding an `+optional`.

apis/v1alpha1/shared_types.go
hbagdi: I think 16 is higher but I will not object against it strongly.
 
I can't think of cases where a single route is admitted to more than 16 Gateways.
Even if someone does find a use-case for this, between all kinds of selectors and lists we have in the API, the debugging experience, or even the general UX, of the API would be bad.

There could be some Service Mesh scenarios that would use this. My general preference is to start with lower limits.

apis/v1alpha1/backendpolicy_types.go
```diff
@@ -31,16 +31,25 @@ import (
...
-	Spec   BackendPolicySpec   `json:"spec,omitempty"`
+	// Spec defines the desired state of BackendPolicy.
+	//
+	// +optional
+	Spec BackendPolicySpec `json:"spec,omitempty"`
+
+	// Status defines the current state of BackendPolicy.
+	//
+	// +optional
+	// +kubebuilder:default={conditions: {{type: "Admitted", status: "False", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
```
robscott, line 46: I think we've already discussed this in another thread, but any form of condition on a BackendPolicy is going to be very hard to get right. This is a resource that may be used by zero-many controllers. I think this kind of default value makes sense for resources like Gateway that are clearly waiting for a single controller, but that's not the case here.

robscott (commented):
@danehans Thanks for all the work on this! There's a lot of great cleanup in here, and I'd love to get it into v0.2.0 if possible. This PR is pretty big though, and it's hard to truly capture all of it in a review. I'm hoping there are some ways to simplify this so it's easier to get in this week. Here are some areas I'd rather see shelved or punted to another PR.

1. I don't understand why Metadata, Spec, and Status are all optional. I recognize that all k8s APIs have these fields configured with `omitempty` and `+optional` tags, but none are pointers. I think mirroring upstream would be ideal here.
2. I think structs should be pointers when they are optional. As you've pointed out, the API conventions make an allowance for optional maps and slices that are not pointers, but I'm not seeing the same guidance for structs. In many cases I think it would be difficult to understand if a struct was truly empty, at least more difficult than simply checking if the value of the field was nil.
3. I think a default value for BackendPolicy status does not make sense. I'll add more comments specifically on that field.
4. In general, I'm uncomfortable with increases to min lengths in this scope because they represent breaking API changes. Although we've been open to breaking changes in our Go API for new minor releases, I don't think we should extend that to our Kubernetes API unless it's in the form of loosening validation.

apis/v1alpha1/backendpolicy_types.go
danehans: > I'm assuming a user can create a BackendPolicy with [] for this field, do they even need to specify that?

Other than `port`, all  `BackendRef ` fields are required so an empty `BackendRefs` is not permitted.

apis/v1alpha1/backendpolicy_types.go
danehans: @roberthbailey @hbagdi thank you for the feedback. If pointers are used for optional structs, I would like to be consistent across the API's. This means changing spec/status for all resources. For example:
```
Spec GatewaySpec `json:"spec,omitempty"`

// change to:
Spec *GatewaySpec `json:"spec,omitempty"`
```

By doing so, spec/status optionality will differ from most [k8s APIs](https://github.com/kubernetes/api).

apis/v1alpha1/backendpolicy_types.go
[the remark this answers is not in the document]
danehans: > So to clarify, we're saying that status itself is optional, but if it is specified it must include at least one condition?

@robscott yes, that is what the changes in this PR propose. I will revert these changes so the PR can proceed.

apis/v1alpha1/backendpolicy_types.go
robscott: I haven't seen any k8s resources using pointers for meta, spec, or status. I'm not sure why that is, I really don't even understand why they're optional, but I'd rather not diverge from the pattern in upstream for these fields.

apis/v1alpha1/gateway_types.go
danehans: I have not seen any object omit this field.

apis/v1alpha1/gateway_types.go
danehans: Are we OK diverging from how upstream defines spec/status/metadata optionality?

xref: https://github.com/kubernetes/api/blob/master/core/v1/types.go

apis/v1alpha1/gatewayclass_types.go
danehans: I will revert these changes so the PR can proceed.

apis/v1alpha1/httproute_types.go
danehans: Same response as https://github.com/kubernetes-sigs/service-apis/pull/538/files#r573947867.

apis/v1alpha1/httproute_types.go
danehans: I'll revert this change so the PR can proceed.

apis/v1alpha1/gateway_types.go
robscott: As strange as it is, I think it likely is appropriate to follow what upstream is doing here.

apis/v1alpha1/shared_types.go
danehans: @hbagdi thanks for the feedback. I reduced the maximum to 8. @howardjohn PTAL and let us know if you have issue with this maximum.

howardjohn: For a service mesh use case I would imagine its either a couple gateways at most, or 1 Gateway per service at which point this would be well beyond 16 and instead be 1000s which wouldn't work before or after this change. Probably the former would be a much more sensible way to model it.

So this seems reasonable

apis/v1alpha1/shared_types.go
```diff
@@ -181,24 +195,27 @@ type RouteGatewayStatus struct {
...
+	// must add an entry to this list when the controller first sees the
+	// route and should update the entry as appropriate when the route is
+	// modified.
+	//
+	// A minimum of 1 and a maximum of 100 Gateways will be represented
+	// in this list. If the list is full, additional Gateways may be
+	// using this Route that are not included in the list. An empty list
+	// means the route has not been admitted by any Gateway.
+	//
+	// +kubebuilder:validation:MinItems=1
```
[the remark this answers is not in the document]
danehans: ack

apis/v1alpha1/backendpolicy_types.go
danehans: ack

danehans:
> I don't understand why Metadata, Spec, and Status are all optional. I recognize that all k8s APIs have these fields configured with omitempty, but none of them have +optional tags. Although following API conventions strictly would result in a +optional tag here, I'm just not seeing it anywhere else. Because of that, I'd prefer to leave these fields as they originally were and try to get some clearer guidance on this from apimachinery. There has to be a reason for how this is done, I'm just not seeing it spelled out anywhere.

`+optional` is used in the apps group APIs: https://github.com/kubernetes/api/blob/master/apps/v1/types.go

In an effort to move the PR forward, I'll remove `+optional` from all spec/status/meta fields.

> I think structs should be pointers when they are optional. As you've pointed out, the API conventions make an allowance for optional maps and slices that are not pointers, but I'm not seeing the same guidance for structs. In many cases I think it would be difficult to understand if a struct was truly empty, at least more difficult than simply checking if the value of the field was nil.

I'll revert structs to pointers.

> I think a default value for BackendPolicy status does not make sense. I'll add more comments specifically on that field.

I'll remove the default.

> In general, I'm uncomfortable with increases to min lengths in this scope because they represent breaking API changes. Although we've been open to breaking changes in our Go API for new minor releases, I don't think we should extend that to our Kubernetes API unless it's in the form of loosening validation.

I'll remove minLength changes from this PR.

apis/v1alpha1/shared_types.go
danehans: I see `MaxItems=16` throughout the API's. I want to avoid having this PR cause a regression, so I'm going to remove this change.

robscott:
> `+optional` is used in the apps group APIs: https://github.com/kubernetes/api/blob/master/apps/v1/types.go
@danehans you're completely right on this one, sorry for the confusion! I edited my comment but was too late. I'm fine with adding `+optional` tags in to match upstream.

apis/v1alpha1/httproute_types.go
```diff
@@ -608,7 +630,7 @@ type HTTPRouteForwardTo struct {
 	// in HTTPRouteRule.)
 	//
 	// +optional
-	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:MaxItems=8
```
robscott, line 633: I think we should only tighten validation with new API versions. If we want to do this I think it should be part of v1alpha2.

apis/v1alpha1/shared_types.go
```diff
@@ -181,23 +194,25 @@ type RouteGatewayStatus struct {
 	//
 	// +listType=map
 	// +listMapKey=type
+	// +kubebuilder:validation:MinItems=1
```
robscott, line 197: Same comment here - I don't think we should tighten validation as part of this release. It does seem like we're compiling an increasing number of changes for v1alpha2 though.

apis/v1alpha1/shared_types.go
```diff
@@ -181,23 +194,25 @@ type RouteGatewayStatus struct {
...
+	// and the status of the route with respect to each Gateway. When a
+	// Gateway selects this route, the controller that manages the Gateway
+	// must add an entry to this list when the controller first sees the
+	// route and should update the entry as appropriate when the route is
+	// modified.
+	//
+	// A minimum of 1 and a maximum of 100 Gateways will be represented
+	// in this list. If the list is full, additional Gateways may be
+	// using this Route that are not included in the list. An empty list
+	// means the route has not been admitted by any Gateway.
```
robscott, line 215: I think this is likely covered better with https://github.com/kubernetes-sigs/service-apis/pull/528, can you pull this change out of this PR?

docs-src/guidelines.md
```diff
@@ -83,3 +83,12 @@ We expect there will be varying levels of conformance among the
...
 be differences in behavior from the spec.
+
+## API Conventions
+
+Service APIs follow Kubernetes API [conventions][1]. These conventions
+are intended to ease client development and ensure that configuration
+mechanisms can consistently be implemented across a diverse set of use
+cases.
+
+[1]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
```
robscott, line 94: I think the way we're treating fields with default values may be slightly unique here and worth a callout. If I'm understanding the approach correctly, fields with default values are considered optional but will not be pointers because they will never truly be empty?

robscott (commented):
Thanks for the work on this @danehans! Just a few small things but otherwise LGTM.

docs-src/guidelines.md
danehans: @robscott I've updated, PTAL.

apis/v1alpha1/shared_types.go
danehans: @robscott done.

apis/v1alpha1/shared_types.go
danehans: @robscott done.

apis/v1alpha1/httproute_types.go
danehans: @robscott done.

robscott:
Thanks!

/lgtm

PR #540, opened 2021-02-10
Cleaning up codegen

robscott:
**What type of PR is this?**
/kind cleanup

We were previously running codegen twice - once with controller-gen, and once with deepcopy-gen. This cleans that up a bit and ensures we're not duplicating work. This also removes some unnecessary annotations from GatewayClass. (They were useful for deepcopy-gen but not for controller-gen).

```release-note
NONE
```


howardjohn (approved):

hbagdi (approved):
I pulled this down, removed all the generated code from the repo and then ran `make generate` to make sure everything works as expected. It does. Thanks for the clean up Rob!
/approve

PR #541, opened 2021-02-10
Renaming Service APIs to Gateway API

### LINKED ISSUE #536: Change repo name?
Should we change the name of the repo to align with the primary resource names (gateway-apis?)

robscott:
**What type of PR is this?**
/kind cleanup

As discussed in #536, Gateway API more clearly represents this project now.

Fixes #536

```release-note
Renames Service APIs to Gateway API
```

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,10 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
-# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/service-apis/blob/master/OWNERS_ALIASES for a list of members for each alias. 
+# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/gateway-api/blob/master/OWNERS_ALIASES for a list of members for each alias.
 
 approvers:
   - sig-network-leads
-  - service-apis-admins
-  - service-apis-maintainers
+  - gateway-api-admins
+  - gateway-api-maintainers
 
 reviewers:
-  - service-apis-maintainers
+  - gateway-api-maintainers
--- a/docs-src/community.md
+++ b/docs-src/community.md
@@ -7,14 +7,14 @@ discussions around the APIs.
 
 Feedback and bug reports should be filed as [Github Issues][gh-issues] on this repo.
 
-[gh-issues]: https://github.com/kubernetes-sigs/service-apis/issues/new/choose
+[gh-issues]: https://github.com/kubernetes-sigs/gateway-api/issues/new/choose
 
 ## Communications
 
 Major discussions and notifications will be sent on the [SIG-NETWORK mailing
 list][signetg].
 
-We also have a [Slack channel (sig-network-service-apis)][slack] on k8s.io for day-to-day
+We also have a [Slack channel (sig-network-gateway-api)][slack] on k8s.io for day-to-day
 questions, discussions.
 
 [signetg]: https://groups.google.com/forum/#!forum/kubernetes-sig-network
@@ -22,9 +22,9 @@ questions, discussions.
 
 ## Meetings
 
-Meetings discussing the evolution of the service APIs will alternate times to
+Meetings discussing the evolution of the Gateway API will alternate times to
 accommodate participants from various time zones. This calendar includes all
-Service APIs meetings as well as any other SIG-Network meetings.
+Gateway API meetings as well as any other SIG-Network meetings.
 
 <iframe
   src="https://calendar.google.com/calendar/embed?src=88fe1l3qfn2b6r11k8um5am76c%40group.calendar.google.com"
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,6 +1,6 @@
 ---
 name: Bug Report
-about: Report a bug encountered with the service-apis project
+about: Report a bug encountered with the gateway-api project
 labels: kind/bug
 
 ---
--- a/.github/ISSUE_TEMPLATE/enhancement.md
+++ b/.github/ISSUE_TEMPLATE/enhancement.md
@@ -1,6 +1,6 @@
 ---
 name: Enhancement Request
-about: Suggest an enhancement to the service-apis project
+about: Suggest an enhancement to the gateway-api project
 labels: kind/feature
 
 ---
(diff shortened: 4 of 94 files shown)


robscott:
Adding a temporary hold on this until I figure out the appropriate order of operations with the related k/org work.

/hold

OWNERS
```diff
@@ -1,10 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
-# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/service-apis/blob/master/OWNERS_ALIASES for a list of members for each alias. 
+# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/gateway-api/blob/master/OWNERS_ALIASES for a list of members for each alias.
 
 approvers:
   - sig-network-leads
-  - service-apis-admins
-  - service-apis-maintainers
+  - gateway-api-admins
+  - gateway-api-maintainers
```
hbagdi, line 7: Should we do this at once? It could break our workflows if we don't get prow configurations fixed in the different repository.

OWNERS
```diff
@@ -1,10 +1,10 @@
 # See the OWNERS docs at https://go.k8s.io/owners
-# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/service-apis/blob/master/OWNERS_ALIASES for a list of members for each alias. 
+# See the OWNERS_ALIASES file at https://github.com/kubernetes-sigs/gateway-api/blob/master/OWNERS_ALIASES for a list of members for each alias.
```
hbagdi, line 2: @bowei Do you have sufficient privileges to do the rename? Just want to make sure we can change the repo name once this PR gets in.

OWNERS
hbagdi: I see https://github.com/kubernetes/org/pull/2491. Great job!

docs-src/community.md
```diff
@@ -24,7 +24,7 @@ questions, discussions.
 
 Meetings discussing the evolution of the service APIs will alternate times to
```
hbagdi, line 25: grep caught this left-over reference to Service APIs.

docs-src/community.md
```diff
@@ -7,14 +7,14 @@ discussions around the APIs.
 
 Feedback and bug reports should be filed as [Github Issues][gh-issues] on this repo.
 
-[gh-issues]: https://github.com/kubernetes-sigs/service-apis/issues/new/choose
+[gh-issues]: https://github.com/kubernetes-sigs/gateway-api/issues/new/choose
```
hbagdi, line 10: Do the PRs in org and community repo take care of title of Youtube recordings?

hbagdi (commented):
Great job, Rob!
This is a big lift and thanks for taking care of it.

hbagdi:
`make serve` fails:

```bash
make -f docs.mk serve                                                                                                                                                                                              
make[1]: Entering directory '/<dir>/service-apis'                                                                                                                                                            
docker run \                                                                                                                                                                                                       
        -it \                                                                                                                                                                                                      
        --sig-proxy=true \                                                                                                                                                                                         
        --mount type=bind,source=/<dir>/service-apis,target=/d \                                                                                                                                             
        -p 127.0.0.1:8000:8000 \                                                                                                                                                                                   
        --rm \                                                                                                                                                                                                     
        k8s.gcr.io/gateway-api-mkdocs:latest \                                                                                                                                                                     
        /bin/bash -c "cd /d && mkdocs serve -a 0.0.0.0:8000"                                                                                                                                                       
Unable to find image 'k8s.gcr.io/gateway-api-mkdocs:latest' locally                                                                                                                                                
docker: Error response from daemon: manifest for k8s.gcr.io/gateway-api-mkdocs:latest not found: manifest unknown: Failed to fetch "latest" from request "/v2/gateway-api-mkdocs/manifests/latest".                
See 'docker run --help'.                                                                                                                                                                                           
make[1]: *** [docs.mk:88: serve] Error 125                                                                                                                                                                         
make[1]: Leaving directory '/<dir>/service-apis'                                                                                                                                                             
make: *** [Makefile:87: serve] Error 2                                                                        
```

We need o create a new Docker repo under `k8s.gcr.io` or the name be.

docs-src/community.md
robscott: Updated, thanks!

docs-src/community.md
robscott: I think that's still setup relatively manually by an org admin, but I can check to be sure that actually happens.

OWNERS
robscott: This will be completed one https://github.com/kubernetes/org/pull/2491 merges in. The owner aliases we rely on within this repo for approvals are all tied to the local OWNERS_ALIASES file which we simply try to keep in sync with k/org. So even if the timing doesn't match up perfectly, everything should still work.

OWNERS
robscott: Thanks! This will be completed one kubernetes/org#2491 merges in. The owner aliases we rely on within this repo for approvals are all tied to the local OWNERS_ALIASES file which we simply try to keep in sync with k/org. So even if the timing doesn't match up perfectly, everything should still work.

robscott:
@hbagdi That's a confusing error but I don't think it's actually an issue here. Other than for the webhook, I don't think any of our images have been pushed anywhere yet. In the case of this image, it's manually built by `make docs` (and also `make generate` since it calls `make docs`). So anyone who's trying to run `make serve` for the first time without having run either of those commands first will run into that error. This would be true for anyone new to the project or of course now that we're changing the name, there's a slightly higher likelihood of someone running `make serve` before other `make` commands and getting this error.

hbagdi:
> @hbagdi That's a confusing error but I don't think it's actually an issue here. Other than for the webhook, I don't think any of our images have been pushed anywhere yet. In the case of this image, it's manually built by make docs (and also make generate since it calls make docs). So anyone who's trying to run make serve for the first time without having run either of those commands first will run into that error. This would be true for anyone new to the project or of course now that we're changing the name, there's a slightly higher likelihood of someone running make serve before other make commands and getting this error.

My bad. This is no big deal and is fine as is.


I looked through the large patch and I think everything matches up.
One thing that could be helpful is to include somewhere that this project was known as Service  APIs in the past.
Apart from that, this looks good to me.
I'll hold off on approval because one more set of eyes should look through this and I'm not sure about the timing.

PR #543, opened 2021-02-11
allow namespace-scoped parametersRef

### LINKED ISSUE #524: Inconsistency in ParametersRef Documentation
**What happened**:
I'm trying to implement Service APIs and I'm confused on what resources can be used for `gatewayclass.spec.parametersRef`.

That either cluster-scoped or namespace-scoped resources are allowed. The namespace **must** be known by the implementation when using a namespace-scoped resource.

N/A

N/A

hbagdi:
This patch adds a namespace field to the parametersRef reference.
This allows cluster-scoped GatewayClass resource to reference a
namespaced-scoped parameters resource.

This is in-line with upstream KEP 2365:
https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-network/2365.yaml

Why is it done the way it is done?
- Namespace field was not added to LocalObjectReference because that
type is referenced in a lot of places. We don't want to add in an
optional namespace field in all these places and increase security
issues with cross-namespace references.
- ObjectReference was not used because upstream discourages its use:
https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference. Instead, a new type was
introduced as per upstream's guidance.
- A new "Cluster" field was added as advised upstream:
https://github.com/kubernetes/enhancements/pull/2366#discussion_r567114299
- This patch also incorporates the docs update from https://github.com/kubernetes-sigs/service-apis/pull/523.

**What type of PR is this?**

/kind feature

This patch adds a namespace field to the parametersRef reference.
This allows cluster-scoped GatewayClass resource to reference a
namespaced-scoped parameters resource.

Fixes #524
Close #523

```release-note
Added a new namespace field to specify a namespace for Parameters reference in GatewayClass resource.
```


apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,20 +62,58 @@ type GatewayClassSpec struct {
...
+	// Name is the name of the referent.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Scope represents if the referent is a Cluster or Namespace scoped resource.
+	// This may be set to "Cluster" or "Namespace".
+	// +kubebuilder:validation:Enum=Cluster;Namespace
+	// +kubebuilder:default=Cluster
```
robscott, line 106: I think for defaulting to work we need an `omitempty` tag. If we have an `omitempty` tag, we should also mark the field as optional to match the API conventions. @danehans can confirm here, but that's how I remember this working.

robscott (commented):
Thanks for your work on this @hbagdi! This all matches up with the KEP well, just one tiny nit about defaulting but otherwise LGTM.

apis/v1alpha1/gatewayclass_types.go
hbagdi: I tested and this is indeed the case. Updated.

apis/v1alpha1/gatewayclass_types.go
```diff
@@ -62,20 +62,59 @@ type GatewayClassSpec struct {
...
+	// +omitempty
```
robscott, line 107: This should actually be `+optional`

hbagdi: Oops, updated.

robscott:
Thanks!

/lgtm
/approve

PR #544, opened 2021-02-11
define semantics for extended protocols

### LINKED ISSUE #433: Define Semantics for Extended Protocol Types
The Gateway API defines "Core" and "Implementation-specific" protocol types but does not address "Extended" protocols. Should "Extended" protocols contain a prefix, i.e. `extended/GRPC" to denote that implementations are not required to support this protocol type?

xref: https://github.com/kubernetes-sigs/service-apis/pull/398

hbagdi:
**What type of PR is this?**

/kind cleanup

Removes ambiguity around extended protocols.

Fixes #433

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -252,8 +252,8 @@ type Listener struct {
 // Implementations can define their own protocols if a core ProtocolType does not
 // exist. Such definitions must use prefixed name, such as
 // `mycompany.com/my-custom-protocol`. Un-prefixed names are reserved for core
-// protocols. Any protocol defined by implementations will fall under custom
-// conformance.
+// and extended protocols. Any protocol defined by implementations will fall
+// under custom conformance.
 type ProtocolType string
 
 const (
--- a/docs-src/spec.md
+++ b/docs-src/spec.md
@@ -3247,8 +3247,8 @@ a reason of &ldquo;UnsupportedProtocol&rdquo;.</p>
 <p>Implementations can define their own protocols if a core ProtocolType does not
 exist. Such definitions must use prefixed name, such as
 <code>mycompany.com/my-custom-protocol</code>. Un-prefixed names are reserved for core
-protocols. Any protocol defined by implementations will fall under custom
-conformance.</p>
+and extended protocols. Any protocol defined by implementations will fall
+under custom conformance.</p>
 </p>
 <table>
 <thead>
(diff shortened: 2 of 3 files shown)


apis/v1alpha1/gateway_types.go
```diff
@@ -252,8 +252,8 @@ type Listener struct {
 // Implementations can define their own protocols if a core ProtocolType does not
 // exist. Such definitions must use prefixed name, such as
 // `mycompany.com/my-custom-protocol`. Un-prefixed names are reserved for core
-// protocols. Any protocol defined by implementations will fall under custom
-// conformance.
+// and extended protocols. Any protocol defined by implementations will fall
```
robscott, line 255: What are extended protocols here? It looks like we haven't associated conformance levels with the protocols defined below.

hbagdi: Good question. The protocol themselves will probably not make sense to have conformance levels defined. I think conformance should be based on how the Protocol is being used, for example, for Route associated with the protocol. Like, there have been brief statements saying that TCP and UDP routes would be extended conformance in the beginning and then later be moved into core as we gain experience from implementers.

This begs the question: Why should protocols be tied to conformance ladder? I feel like I could be missing something.

cc @danehans @bowei

robscott: If I'm remembering this correctly, this all came from the idea that we wanted to allow users to specify protocols that we're not defined within this API. Previously we had strict validation that protocol must come from the list defined here. That really limited the extensibility of the API. I think what we're trying to communicate here is if the protocol is not defined in the API, a domain prefixed path should be used to ensure it does not conflict with any future protocols added to the API.

howardjohn: If I were to implement `http3` or similar, it feels weird to put `mycompany.com/http3`, even if it later becomes part of the API. This would just make users have to migrate from  `mycompany.com/http3` to `http3` if it does become core.

robscott: Yeah, that's a good point. Here's the context for the original decision around this: https://github.com/kubernetes-sigs/service-apis/pull/398#discussion_r510493532. It's hard to find a good list to act as a reference here so at this point we've just said we'll expand our list as additional commonly used protocols are supported. That will admittedly not scale well, so any alternative solutions are very welcome here.

robscott: A bit more discussion around this in https://github.com/kubernetes-sigs/service-apis/issues/433.

hbagdi: > If I were to implement http3 or similar, it feels weird to put mycompany.com/http3, even if it later becomes part of the API. This would just make users have to migrate from mycompany.com/http3 to http3 if it does become core.

What is really the problem with that? I see that as completely fine and migration to be gradual and painless too.

PR #549, opened 2021-02-12
Fixing API Reference Doc Generation

### LINKED ISSUE #539: Rendering of literal star characters in documentation
In reading your documentation, specifically that surrounding wildcard hostnames for a [PR on Traefik](https://github.com/traefik/traefik/issues/7874), I noticed that you appear to be suffering from markdown rendering literal `*` characters as markup generating `<em>` tags. On slack I was shown that this [had indeed been noticed](https://github.com/kubernetes-sigs/service-apis/pull/509) but the solution wasn't obvious to you. I believe if it is markdown, that wrapping the offending wildcard hostnames in backticks should be a higher precedence operation and prevent it from interpreting them as emphasis marks.

Per our discussion on slack I'm opening this bug report for tracking.

robscott:
**What type of PR is this?**
/kind cleanup

Previously reference docs were converted from godoc to HTML with gen-crd-api-reference-docs and then converted to markdown with mkdocs as part of generating the docs site. This takes out the second step and instead inserts the generated html for the reference docs directly into the generated html for the docs site.

I will be the first to say this is a rather ugly solution and I'm very open to others. I've tried lots of different approaches but it's quite difficult to solve this without removing the double conversion that was happening before. I tried using the `include` directive we use elsewhere with mkdocs but even that results in the html getting converted to markdown again. https://github.com/kubernetes-sigs/service-apis/issues/539 has some other options I tried as well with limited success.

Some notable downsides of this approach:
- We need to use backticks to escape `*` characters in godoc.
- Generated reference docs are not included in `make serve`.
- `docs-src/spec.md` does not include generated reference docs - need to either look at `docs-src/spec.html` or the website itself.

I'm very open to other solutions here, but so far this is the least bad solution I've found.

Fixes #539 

```release-note
NONE
```

/cc @jpeach @hbagdi @jberger


robscott:
Need to get tests running again now that repo has been renamed, will make a PR to test-infra shortly. 
/hold

robscott:
Waiting for https://github.com/kubernetes/test-infra/pull/20851 to merge in before cancelling hold on this.

docs.mk
```diff
@@ -55,7 +55,7 @@ images: .mkdocs.dockerfile.timestamp
 		--sig-proxy=true \
 		--rm \
 		$(MKDOCS_IMAGE) \
-		/bin/bash -c "cd /d && $(MKDOCS) build; find /d/docs -exec chown $(UID):$(GID) {} \;"
+		/bin/bash -c "cd /d && $(MKDOCS) build && sed -i -e '/<replace-with-generated-content>/{r docs-src/spec.html' -e 'd}' docs/spec/index.html; find /d/docs -exec chown $(UID):$(GID) {} \;"
```
bowei, line 58: (minor) might want to make it REPLACE_WITH_GENERATED_CONTENT to make it more clear what that string is doing

bowei:
seems ok to me

/lgtm

docs.mk
robscott: Good call, that is better - updated.

robscott:
@bowei Updated to use better replacement string. PTAL.

/hold cancel

youngnick:
I agree it's not the most elegant solution, but it's much, much better than what we had before.

/lgtm

PR #553, opened 2021-02-16
Adding notes referring to previous project name

robscott:
**What type of PR is this?**
/kind documentation

@hbagdi suggested this in https://github.com/kubernetes-sigs/gateway-api/pull/541 and I just realized I forgot to actually update that PR. This should make it a bit clearer that this is actually the project that used to be known as "Service APIs" for anyone that might be confused.

```release-note
NONE
```

/cc @hbagdi

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
 # Kubernetes Gateway API
 
-The Gateway API is a part of the [SIG Network][sn], and this repository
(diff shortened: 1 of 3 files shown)


hbagdi:
/lgtm

Thanks for taking care of this Rob!

PR #559, opened 2021-02-17
Adding Netlify support (Part 2)

robscott:
**What type of PR is this?**
/kind documentation

This is a follow up to #560 that completes the move to `gateway-api.sigs.k8s.io`. Because the diff is massive, let me try to summarize the key changes:

- Since Netlify can build our docs for us, there should be no need for us to retain the generated files in this repo. This removes generated docs files entirely.
- This moves `docs` to `site` and `docs-src` to `site-src` to allow us to keep `docs` for GitHub pages.
- This adds new redirect pages to `docs` so our GitHub pages will redirect to Netlify.
- This reenables search. We disabled that because we had so many conflicts with the index file. Since this removes the generated docs from tracking we should not have that problem anymore.

Examples of the redirect in action can be seen here: https://robscott.github.io/gateway-api
The Netlify deploy preview for this change can be seen here: https://deploy-preview-559--kubernetes-sigs-gateway-api.netlify.app/

```release-note
NONE
```


robscott:
This PR as it is written will completely break our GitHub pages. I may need to split this out into 2+ PRs, but I want to make sure that Netlify will work well without any generated docs checked into the repo. Adding a hold to this until we can verify that everything works.

/hold

.gitignore
```diff
@@ -49,3 +49,4 @@ Session.vim
 .vscode
 /www/test_out
 .*.timestamp
+/docs
```
hbagdi, line 52: Should this not be `docs` or `docs/`?
Either that or TIL a new `.gitignore` syntax. So many of them

hbagdi:
> If you do want to make docs locally, it should still work, you'll just need to have mkdocs installed.

Rob, how hard is it remove this requirement of having `mkdocs` installed on your workstation? Most maintainers and contributors have their workflows centered around Docker and I think what we have in mainline is much easier to work with for everyone.

Is that Netlify doesn't work well with Docker here?

Other than this one thing, other changes are looking good.

.gitignore
robscott: My understanding is that this ignores a docs directory in the root of the project but would not ignore something like `foo/docs` in the future. We already have a surprising amount of lines in our `.gitignore` with the `/` prefix so hopefully it's not too out of place.

robscott:
>Rob, how hard is it remove this requirement of having mkdocs installed on your workstation? Most maintainers and contributors have their workflows centered around Docker and I think what we have in mainline is much easier to work with for everyone.
>Is that Netlify doesn't work well with Docker here?

Yeah this is one downside of this approach. To clarify, no one will need to have mkdocs installed to contribute to gateway-api, they can simply just update markdown and see the deploy preview that Netlify automatically provides for their PR. That's always going to be the most accurate way to preview docs changes.

Netlify builds run in Docker containers and it does not look like they support docker in docker. We could potentially maintain our own build image separately and use that for docs previews, but that does not seem like it is worth the effort to me.

I think it will be relatively rare that users want to preview docs changes locally. In those instances, I don't think it's too much overhead to install mkdocs locally. Our current approach builds a docker image that includes mkdocs every time we run `make docs`. Although we could continue that approach, it's _much_ slower than just running mkdocs locally. Since this is no longer a required or common step, I think it's reasonable to pull out the docker parts of this.

robscott:
OK I think this is ready to go once kubernetes/k8s.io#1668 merges in. I've restructured things to use `site` and `site-src` instead of `docs` and `docs-src`. This enables us to continue using `docs` for GitHub pages which I'm using to point to our updated docs site (example here: https://robscott.github.io/gateway-api). 

Our other options here would be:

1. Add `index.md` and `404.md` to the root directory, but I got rather unhelpful build errors when trying this with GitHub pages. This would also require an update to our GitHub pages config.
2. Publish a new and separate `gh-pages` branch and switch our GitHub pages config to use this. This is interesting but might be a bit more difficult to manage.

The Netlify builds and deploy previews for this PR appear to be working well - example here: https://deploy-preview-559--kubernetes-sigs-gateway-api.netlify.app/

robscott:
Although the DNS has been updated, this PR should not be merged until the part 1 PR (#560) is merged in and we've confirmed that gateway-api.sigs.k8s.io is working.

hbagdi:
> I think it will be relatively rare that users want to preview docs changes locally. In those instances, I don't think it's too much overhead to install mkdocs locally. Our current approach builds a docker image that includes mkdocs every time we run make docs. Although we could continue that approach, it's much slower than just running mkdocs locally. Since this is no longer a required or common step, I think it's reasonable to pull out the docker parts of this.

Okay. Not a big deal. Let's continue with what we have right now.

docs/404.md
```diff
@@ -0,0 +1,6 @@
+---
+permalink: /404.html
+---
+# Gateway API
+
+Our docs have moved to [gateway-api.sigs.k8s.io](https://gateway-api.sigs.k8s.io).
```
hbagdi, line 6: Can/Should we auto-redirect?
https://superdevresources.com/redirects-jekyll-github-pages/

robscott: I had no idea that was possible, but that's a great idea, thanks! Going to experiment with some of these.

robscott: Good call, that is a lot better. Appears to be working well at https://robscott.github.io/gateway-api (GitHub pages on my fork is configured to use this branch).

robscott:
I've updated the PR description to update the limited scope of this PR now that #560 has merged in. Based on @hbagdi's great suggestion, GitHub pages will now redirect to Netlify. As a precaution, I'll leave a hold on this PR until @bowei is able to change our GitHub site link to https://gateway-api.sigs.k8s.io/. Otherwise I think this is ready to go.

robscott:
Link has been updated in sidebar, should be good to go now.

/hold cancel

PR #560, opened 2021-02-17
Adding Netlify support (Part 1)

robscott:
**What type of PR is this?**
/kind documentation

This adds initial Netlify support and removes parts of docs generation that are no longer necessary. Merging this in before https://github.com/kubernetes-sigs/gateway-api/pull/559 will allow us to have both Github pages and Netlify in place for a short time while we ensure that Netlify works well. This should not be a long term solution though as the goal is to fully transition to Netlify ASAP.

```release-note
NONE
```


runtime.txt
```diff
@@ -0,0 +1 @@
+3.7
```
hbagdi, line 1: Is this required?

docs/spec.html
```diff
@@ -1,4524 +0,0 @@
-<p>Packages:</p>
```
hbagdi, line 1: Since this file is deleted, the Github pages site will no longer have the spec.
That seems okay since this is only for a small transition period (I expect a day or so) but I want to make sure that we do this intentionally.

robscott: Sorry for the confusion here. What's being renamed here is actually the raw file generated by `gen-crd-api-reference-docs`. GitHub Pages actually serves `spec/index.html` when you visit `/spec` but serves this file if you visit `/spec.html` (https://kubernetes-sigs.github.io/gateway-api/spec.html). Netlify was actually serving `spec.html` in both cases which was a problem. This PR effectively moves `docs-src/spec.html` to `docs/genspec.html` to get around that. That results in Netlify also rendering `/spec` from `/spec/index.html`:

- https://deploy-preview-560--kubernetes-sigs-gateway-api.netlify.app/spec
- https://deploy-preview-560--kubernetes-sigs-gateway-api.netlify.app/genspec

The follow up PR will then exclude this dir from git and we won't have to deal with any of these generated doc files in PRs going forward.

runtime.txt
robscott: Good catch, this was not clearly tied to Netlify. We had to tell it to use Python 3 instead of 2, but this was only one of several ways to do that. I've moved this to an environment variable in `netlify.toml` which is probably much easier to understand.

docs/spec.html
robscott: Correction, as @nak3 pointed out in #561 even https://kubernetes-sigs.github.io/gateway-api/spec renders spec.html for GitHub pages, only https://kubernetes-sigs.github.io/gateway-api/spec/ works as expected right now. This rename should fix that.

PR #562, opened 2021-02-22
Allow empty string for Group field to indicate core API group

nak3:
When using core API, the apiGroup is `""` (empty string) is used in general.

(c.f. api-resources shows empty API group for core api.)
```
$ kubectl api-resources
NAME                              SHORTNAMES   APIGROUP                       NAMESPACED   KIND
secrets                                                                       true         Secret
```

However, current CRD does not allow empty strings and it seems that
"core" is expected to specify as per example.
It is different from other Kubernetes resources, such as Role, so
difficult to find it.

**What type of PR is this?**

/kind api-change

```release-note
Group in RouteBindingSelector, ParametersReference and LocalObjectReference allows empty string as "core" API group.
```

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -406,11 +406,11 @@ type RouteBindingSelector struct {
 	//
 	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
-	// Group is the group of the route resource to select.
+	// Group is the group of the route resource to select. Omitting the value
+	// indicates the networking.x-k8s.io API group.
 	// For example, use the following to select an HTTPRoute:
 	//
 	// routes:
-	//   group: networking.x-k8s.io
 	//   kind: HTTPRoute
 	//
 	// Otherwise, if an alternative API group is desired, specify the desired
@@ -423,8 +423,9 @@ type RouteBindingSelector struct {
 	// Support: Core
 	//
 	// +optional
+	// +kubebuilder:default=networking.x-k8s.io
 	// +kubebuilder:validation:MaxLength=253
-	Group *string `json:"group"`
+	Group *string `json:"group,omitempty"`
 	// Kind is the kind of the route resource to select.
 	//
 	// Kind MUST correspond to kinds of routes that are compatible with the
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -406,8 +406,8 @@ type RouteBindingSelector struct {
 	//
 	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
-	// Group is the group of the route resource to select. Omitting the value
-	// indicates the networking.x-k8s.io API group.
+	// Group is the group of the route resource to select. Omitting the value or specifying
+	// the empty string indicates the networking.x-k8s.io API group.
 	// For example, use the following to select an HTTPRoute:
 	//
 	// routes:
@@ -424,6 +424,7 @@ type RouteBindingSelector struct {
 	//
 	// +optional
 	// +kubebuilder:default=networking.x-k8s.io
+	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
 	Group *string `json:"group,omitempty"`
 	// Kind is the kind of the route resource to select.
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -406,11 +406,11 @@ type RouteBindingSelector struct {
 	//
 	// +optional
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
-	// Group is the group of the route resource to select. Omitting the value or specifying
-	// the empty string indicates the networking.x-k8s.io API group.
+	// Group is the group of the route resource to select.
 	// For example, use the following to select an HTTPRoute:
 	//
 	// routes:
+	//   group: networking.x-k8s.io
 	//   kind: HTTPRoute
 	//
 	// Otherwise, if an alternative API group is desired, specify the desired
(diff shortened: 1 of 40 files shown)


robscott:
Thanks @nak3! This LGTM, going to bring it up in community meeting tomorrow to make sure I'm not missing anything.

hbagdi:
Looks good overall.
There is one instance left: https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/tls.md#L48.

nak3:
Thank you! Updated tls.md.

apis/v1alpha1/gateway_types.go
```diff
@@ -424,7 +424,6 @@ type RouteBindingSelector struct {
 	//
 	// +optional
 	// +kubebuilder:default=networking.x-k8s.io
-	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
 	Group string `json:"group,omitempty"`
```
hbagdi, line 428: I was about to lgtm when I noticed that this field has `+optional`.
Why can't the user not specify a group at all? That would pass validation here, right?

Also, a related question, why is this not a pointer?

robscott:
Hey @nak3, we also discussed this at the community meeting today. This approach would conflict with our Group defaulting for the RouteSelector in Gateway. We are currently defaulting that to `networking.x-k8s.io` since that is what most Routes will belong to. I think that leaves us with three options:

1. Make "Group" a pointer in that specific case to maintain defaulting.
2. Remove the default entirely.
3. Continue to require "core" instead of "".

I don't love any of those options. Although 1 might be the least bad, it does feel rather messy and inconsistent with the rest of our API. Interested in what others think here as well.

robscott:
The more I'm thinking about this, the more I think we may want to remove the default entirely. For better or worse, "" has become synonymous with a reference to the core API group throughout Kubernetes resources. Although we could potentially differentiate between nil and "", that would recommend a significant difference between two values that can look nearly identical from the perspective of a yaml manifest.

hbagdi:
> The more I'm thinking about this, the more I think we may want to remove the default entirely.

Do you mean from the RouteSelector on Gateway or for a bigger scope?

I _feel_ defaulting `group` in RouteSelector is good since that will represent a large majority. I'm still split between empty or "core", I like "core" better because explicit APIs are usually better in terms of evolution and readability. We should document that empty group and "core" mean the same thing and then use "core" for defaulting (where needed). how does that sound? 

> Although we could potentially differentiate between nil and "", that would recommend a significant difference between two values that can look nearly identical from the perspective of a yaml manifest.

+1 to not differentiating between nil and "".  It only adds confusion and irritation.

bowei:
I think since there is some confusion around defaulting that we should not do it for the time being, if we cannot reject bad API group spec synchronously.

hbagdi:
With other enhancements getting in, we should make a decision on this one. Should empty group field imply `core` or should `core` be explicitly required? Or both?

/cc @robscott @jpeach

lavalamp:
"" is the canonical way to spell "core" and you'll confuse tools if you do something different.

robscott:
Thanks @lavalamp! Do you know if there's any reasonable way to both set a default group (networking.x-k8s.io in this case) and support "" as core for CRDs? @nak3 do you mind rebasing this PR?

lavalamp:
> Do you know if there's any reasonable way to both set a default group (networking.x-k8s.io in this case) and support "" as core for CRDs?

Yes, set "networking.x-k8s.io" as the default, the field is already a pointer, if a user says it should be "", that will not trigger the defaulting (but null / not present will).

(this comment is only about the technical possibility, and not about whether this is a good idea or not -- no opinion on that since I haven't looked carefully)

liggitt:
> Thanks @lavalamp! Do you know if there's any reasonable way to both set a default group (networking.x-k8s.io in this case) and support "" as core for CRDs? @nak3 do you mind rebasing this PR?

To distinguish between `absent` and `""`, the field has to be a string pointer. Absent will be `nil`, an explicit `""` will be preserved as `"".

It looks like the go field is already a `*string`, not `string`, so it looks like you can already accomplish what you want. You don't need to remove `omitempty` in order to serialize/preserve an explicit `""` value.

A default is only applied to a custom resource property when it is absent/null.

nak3:
Thank you! Alright, I rebased and updated the PR as:
- The `Group` in `RouteBindingSelector`:
  - keeps the default as it is.
  - drops MinLength to support empty string `""`.
- Other `Group`s which do not have any default:
  - drop MinLength to support empty string `""`.

robscott:
Thanks @nak3! This all makes sense to me. I'm not completely sold on leaving defaults in at all, but that should be resolved by GEP #724. At this point we're only planning to make changes to v1alpha2, v1alpha1 is effectively frozen, so if you revert the v1alpha1 changes the rest LGTM.

nak3:
> At this point we're only planning to make changes to v1alpha2, v1alpha1 is effectively frozen, so if you revert the v1alpha1 changes the rest LGTM.

Okay! I reverted the v1alpha1 changes. And so, I reverted the change in `examples` as well (because they are also v1alpha1).

robscott:
Thanks!

/lgtm
/approve

apis/v1alpha2/local_object_reference_types.go
```diff
@@ -21,7 +21,6 @@ package v1alpha2
 type LocalObjectReference struct {
 	// Group is the group of the referent.
 	//
-	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
 	Group string `json:"group"`
```
maelvls, line 25: Would it make sense to backport this change to `v1alpha1` too? Currently, the following is not allowed:

```yaml
kind: Gateway
spec:
  listeners:
  - tls:
      certificateRef:
        name: haproxy-tls
        kind: Secret
        group: "" # 🔥 Validation error, cannot be left empty.
```

which is quite surprising.

Related:
- https://github.com/jcmoraisjr/haproxy-ingress/issues/830

maelvls: Oops, I just noticed the above messages:

> At this point we're only planning to make changes to v1alpha2, v1alpha1 is effectively frozen, so if you revert the v1alpha1 changes the rest LGTM.

> Okay! I reverted the v1alpha1 changes. And so, I reverted the change in examples as well (because they are also v1alpha1).

Never mind my request!

Issue #563, opened 2021-02-22
HTTPRouteMatch always requires Path field when creating HTTPRoute by API


nak3:
**What happened**:

- When creating HTTPRoute with HTTPRouteMatch, HTTPPathMatch field always needs to be defined.

**What you expected to happen**:

- HTTPPathMatch should be pointer.

**How to reproduce it (as minimally and precisely as possible)**:

#### 1. Create HTTPRoute with direct API.

```go
  spec := gwv1alpha1HTTPRouteSpec{
              Rules: []gwv1alpha1.HTTPRouteRule{
                  {
                      Matches: []gwv1alpha1.HTTPRouteMatch{{
                              Headers: &gwv1alpha1.HTTPHeaderMatch{
                                      Type:   gwv1alpha1.HeaderMatchExact,
                                      Values: map[string]string{network.TagHeaderName: tagName},
                              },
                      },
                     ...
  _, err = clients.GatewayAPIClient.HTTPRoutes.Create(ctx, hr, metav1.CreateOptions{})
```

#### 2. Get the following error

```
    headers.go:146: Error creating Ingress "serving-tests/ingress-conformance-headers-tags-udukikgr": HTTPRoute.networking.x-k8s.io "ingress-conformance-headers-tags-udukikgr" is invalid: spec.rules.matches.path.value: Invalid value: "": spec.rules.matches.path.value in body should be at least 1 chars long
```

This is because `HTTPPathMatch` get empty values instead of `nil` like this:

```
Rules:[{Matches:[{Path:{Type: Value:} Headers:0xc0002f7720 ExtensionRef:<nil>}]
```

**Anything else we need to know?**:

- Creating resource by `kubectl apply` does not hit this issue because CRD has [the default value](https://github.com/kubernetes-sigs/gateway-api/blob/7417038f3c546877ee4b107a2baa4022502a1fef/config/crd/bases/networking.x-k8s.io_httproutes.yaml#L82-L87).

PR #564, opened 2021-02-22
Use pointer for HTTPPathMatch

nak3:

Currently `Path` field in `HTTPRoute` is defined by non-pointer. It is not
consistent with `HTTPHeaderMatch` and hits an issue like https://github.com/kubernetes-sigs/gateway-api/issues/563.

Hence this patch changes to use pointer for HTTPPathMatch.

**What type of PR is this?**

/kind cleanup

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/563

```release-note
NONE
```


apis/v1alpha1/httproute_types.go
```diff
@@ -337,7 +337,7 @@ type HTTPRouteMatch struct {
 	//
 	// +optional
 	// +kubebuilder:default={type: "Prefix", value: "/"}
-	Path HTTPPathMatch `json:"path,omitempty"`
+	Path *HTTPPathMatch `json:"path,omitempty"`
```
hbagdi, line 340: Interesting. `HTTPHeaderMatch` is a pointer because there is no default value for it. This was not a pointer because this couldn't have a `nil` value since it has a default.

robscott:
Thanks for this PR @nak3! We discussed this at today's community meeting and realized this was a bit more complicated than I initially thought. It seems like we'll probably need to change all references to optional structs in this API to pointers if this pattern applies broadly. Before we do that, we want to make sure we're completely understanding why this is happening for requests from client-go but not kubectl. If you (or anyone) are able to try creating the same resource with a yaml manifest, I'd like to confirm that is working for the same config client-go is failing with. 

This does feel like a significant issue and a gap in our testing, I just want to make sure that we're correctly understanding the scope and solution here.

robscott:
@danehans or @hbagdi did either of you get time to look into this? It does seem like we need to switch to a pointer here but I haven't had the time to verify yet myself.

robscott:
Thanks @nak3! I've created #576 to track follow up on this so we can go ahead and get this in.

/lgtm
/approve

PR #565, opened 2021-02-26
Add implementations documentation

### LINKED ISSUE #558: Docs Improvements
I wanted to list out some general docs improvements that would improve the site, for others to contribute, and also to keep track for myself. If anybody else has docs improvements that they would find valuable, feel free to pile on this issue and we can keep a running tracker.

- [ ] Make user guide Service names consistent (and consistent within concept docs too). There is foo-v1, foo-v2, foo, foo-canary, etc. Ideally, a small set of Deployment and Service resources should be able to be used to run every single one of the user guides so making the example names converge to only a few would allow this to happen.
- [ ] This needs discussion, but we should decide whether we want to specify an actual application Deployment within the Gateway docs so we can include end-to-end examples that contain real output from an app. I personally think this would be a good move because it would allow the user guides to have concrete steps in them rather than being theoretical examples as they are now. In a [previous version](https://github.com/GoogleCloudPlatform/gke-networking-recipes/tree/master/gateway#traffic-splitting) of the user guide I used [whereami](https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/tree/master/whereami) but I'm happy to use any app that highlights the user guides we have.
- [ ] The [multi-ns user guide](https://kubernetes-sigs.github.io/gateway-api/multiple-ns/) should be reformatted into a similar style as the other guide
- [ ] The [TLS in HTTPRoute user guide](https://kubernetes-sigs.github.io/gateway-api/tls/) has some good conceptual content that is maybe better suited for [Gateway concepts](https://kubernetes-sigs.github.io/gateway-api/gateway/). Perhaps we should move some of that content there and slim down this TLS example to a smaller and self-contained guide.
- [ ] A TLSRoute user guide and also TLSRoute API Types page.
- [x] Addition of a "Gateway API implementations" page (or maybe add this content to [getting started](https://kubernetes-sigs.github.io/gateway-api/getting-started/) ) that lists out the current and planned implementations for Gateway. Ideally a user can go to an implementations page which shows how to install or enable that Gateway Controller and corresponding GatewayClasses in a K8s cluster. Then the user can deploy a reference Gateway from one of these GatewayClasses and continue with any of the standard user-guides, which should all theoretically function if that GatewayClass and controller is compliant.
- [ ] Gateway and Route ownership patterns need to be documented. There are several design patterns that make sense:
    - Single-tenant Gateways - Gateway and Route in same Namespace / Gateway and Route deployed by same owner
    - Shared Gateways - Gateways shared by Routes which are owned by different users (across Namespaces and also within the same Namespace)
    - Controlling hostname ownership - This is related to the route merging and conflicts, but a very common question is how to control which route-owners can own which hostnames. This can be done by specifying multiple listeners for the same port with different hostnames. Each of these listeners is bound to a different set of Routes, effectively giving those Routes exclusive ownership over the hostname. Since this is a design pattern, this capability is not really obvious to users and so it should be documented.
- [ ] We probably need some more conceptual content and examples for the [Gateway concepts](https://kubernetes-sigs.github.io/gateway-api/gateway/) page.
- [ ] Route merging and conflicts are only described in the API specification reference. This should have a section in the concept (or HTTPRoute) portion of the docs.

Happy to take edits or additions to this list. I just wanted to record what's on the top of my head before I forget :)

kind/docs

shaneutt:
**What type of PR is this?**

/kind documentation

As per [k8s slack conversations][slack] and https://github.com/kubernetes-sigs/gateway-api/issues/558 this PR adds documentation to keep track of the downstream Gateway API implementations in other projects.

Fixes #558 

NONE

### DIFF
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -10,15 +10,13 @@ Implementors of Gateway API are encouraged to update this document with status i
 - [Google Cloud][2] (work in progress)
 - [Istio][3] (alpha)
 - [Kong][4] (work in progress)
-- [Knative][5] (work in progress)
-- [Traefik][6] (alpha)
+- [Traefik][5] (alpha)
 
 [1]:#contour
 [2]:#google-cloud-platform
 [3]:#istio
 [4]:#kong
-[5]:#knative
-[6]:#traefik
+[5]:#traefik
 
 ## Project References
 
@@ -71,14 +69,6 @@ You can subscribe to [kong/kubernetes-ingress-controller/issues/692][kong-1] to 
 [kong]:https://konghq.com
 [kic]:https://github.com/kong/kubernetes-ingress-controller
 [kong-1]:https://github.com/kong/kubernetes-ingress-controller/issues/692
-
-### Knative
-
-[Knative][knative] is an open source Kubernetes-based platform to deploy and manage modern serverless workloads.
-
-Knative is actively working towards an implementation of the Gateway API, status updates and documentation will be provided here as the work progresses.
-
-[knative]:https://knative.dev/
 
 ### Traefik
 
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -1,0 +1,95 @@
+# Implementations
+
+This document tracks downstream implementations of Gateway API and provides status and resource references for them.
+
+Implementors of Gateway API are encouraged to update this document with status information about their implementations, the versions they cover, and documentation to help users get started.
+
+## Implementation Status
+
+|                | [Contour][1]       | [GCP][2]           | [Istio][3]         | [Kong][4]         | [Knative][5]       | [Traefik][6]       |
+| :------------- | :----------------: | :----------------: | :----------------: | :---------------: | :----------------: | :----------------: |
+| [v0.1.x][v0.1] | :heavy_minus_sign: | :heavy_minus_sign: | :heavy_check_mark: |:heavy_minus_sign: | :heavy_minus_sign: | :heavy_check_mark: |
+| [v0.2.x][v0.2] | :heavy_check_mark: | :heavy_plus_sign:  | :heavy_plus_sign:  |:heavy_plus_sign:  | :heavy_minus_sign: | :heavy_plus_sign:  |
+
+- :heavy_check_mark: available
+- :heavy_plus_sign: in progress
+- :heavy_minus_sign: not planned
+
+[1]:#contour
+[2]:#google-cloud-platform
+[3]:#istio
+[4]:#kong
+[5]:#knative
+[6]:#traefik
+
+[v0.1]:https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.1.0
+[v0.2]:https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.2.0
+
+## Project References
+
+In this section you will find specific links to blog posts, documentation and other Gateway API references for specific implementations.
+
+### Contour
+
+[Contour][contour] is an open source ingress controller for Kubernetes.
+
+Contour currently implements the latest `v0.2.x` Gateway API Specification.
+
+See the [Contour Gateway API Guide][contour-1] for information on how to deploy and use Contours's Gateway implementation.
+
+[contour]:https://projectcontour.io
+[latest]:https://doc.traefik.io/traefik/routing/providers/kubernetes-gateway/
+[contour-1]:https://projectcontour.io/guides/gateway-api/
+
+### Google Cloud Platform
+
+The [Google Cloud Platform (GCP)][gcp] is a cloud computing platform and infrastructure provider.
+
+GCP is actively working towards an implementation of the Gateway API `v0.2.x` specification, and status updates and documentation will be provided here as the work progresses.
+
+[gcp]:https://cloud.google.com
+
+### Istio
+
+[Istio][istio] is an open source [service mesh][mesh] for Kubernetes clusters.
+
+Istio currently supports the previous `v0.1.x` Gateway API specification and is working towards supporting the latest `v0.2.x` release.
+
+Status updates and documentation for `v0.2.x` will be provided here as the work progresses.
+
+See the [Istio Gateway API Documentation][istio-1] for information on how to deploy and use Istio's Gateway implementation.
+
+[istio]:https://istio.io
+[mesh]:https://istio.io/latest/docs/concepts/what-is-istio/#what-is-a-service-mesh
+[istio-1]:https://istio.io/latest/docs/tasks/traffic-management/ingress/service-apis/
+
+### Kong
+
+[Kong][kong] is an open source API Gateway built for hybrid and multi-cloud environments.
+
+Kong is actively working towards an implementation of the Gateway API `v0.2.x` specification for it's [Kubernetes Ingress Controller][kic] and status updates and documentation will be provided here as the work progresses.
+
+You can subscribe to [kong/kubernetes-ingress-controller/issues/692][kong-1] to track the implementation progress and contribute.
+
+[kong]:https://konghq.com
+[kic]:https://github.com/kong/kubernetes-ingress-controller
+[kong-1]:https://github.com/kong/kubernetes-ingress-controller/issues/692
+
+### Knative
+
+[Knative][knative] is an open source Kubernetes-based platform to deploy and manage modern serverless workloads.
+
+Knative is actively working towards an implementation of the Gateway API, status updates and documentation will be provided here as the work progresses.
+
+[knative]:https://knative.dev/
+
+### Traefik
+
+[Traefik][traefik] is an open source cloud-native application proxy.
+
+Traefik currently supports the previous `v0.1.x` Gateway API specification, check the [Kubernetes Gateway Documentation][traefik-1] for information on how to deploy and use Traefik's Gateway implementation.
+
+Traefik is currently working on implementing `v0.2.x`, status updates and documentation will be provided here as the work progresses.
+
+[traefik]:https://traefik.io
+[traefik-1]:https://doc.traefik.io/traefik/routing/providers/kubernetes-gateway/
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -28,13 +28,16 @@ In this section you will find specific links to blog posts, documentation and ot
 
 [Contour][contour] is an open source ingress controller for Kubernetes.
 
-Contour currently implements the latest `v0.2.x` Gateway API Specification.
+Contour started implementing Gateway API `v0.1.0` but held on releasing that in favor of `v0.2.0` once it released.
 
 See the [Contour Gateway API Guide][contour-1] for information on how to deploy and use Contours's Gateway implementation.
+
+Note that not all of the `v0.2.0` specification is completed yet, [contributions welcome!][contour-2].
 
 [contour]:https://projectcontour.io
 [latest]:https://doc.traefik.io/traefik/routing/providers/kubernetes-gateway/
 [contour-1]:https://projectcontour.io/guides/gateway-api/
+[contour-2]:https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md
 
 ### Google Cloud Platform
 
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -39,13 +39,16 @@ Note that not all of the `v0.2.0` specification is completed yet, [contributions
 [contour-1]:https://projectcontour.io/guides/gateway-api/
 [contour-2]:https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md
 
-### Google Cloud Platform
+### Google Kubernetes Engine
 
-The [Google Cloud Platform (GCP)][gcp] is a cloud computing platform and infrastructure provider.
+The [Google Kubernetes Engine (GKE)][gke] is a managed Kubernetes platform offered by Google Cloud.
 
-GCP is actively working towards an implementation of the Gateway API `v0.2.x` specification, and status updates and documentation will be provided here as the work progresses.
+GKE is actively working towards an implementation of the Gateway API for orchestration of [Google Cloud Load Balancing][gclb].
 
-[gcp]:https://cloud.google.com
+The GKE Gateway Controller will become a managed component of GKE to automate and ease load balancing for containers.
+
+[gke]:https://cloud.google.com/kubernetes-engine
+[gclb]:https://cloud.google.com/load-balancing
 
 ### Istio
 
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -64,11 +64,12 @@ Istio is actively working towards an implementation of the Gateway API and you c
 
 Kong is actively working towards an implementation of the Gateway API `v0.2.x` specification for it's [Kubernetes Ingress Controller][kic] and status updates and documentation will be provided here as the work progresses.
 
-You can subscribe to [kong/kubernetes-ingress-controller/issues/692][kong-1] to track the implementation progress and contribute.
+You can subscribe to [kong/kubernetes-ingress-controller/issues/692][kong-1] to track the implementation progress and [contribute][kong-2]!
 
 [kong]:https://konghq.com
 [kic]:https://github.com/kong/kubernetes-ingress-controller
 [kong-1]:https://github.com/kong/kubernetes-ingress-controller/issues/692
+[kong-2]:https://github.com/Kong/kubernetes-ingress-controller/blob/main/CONTRIBUTING.md
 
 ### Traefik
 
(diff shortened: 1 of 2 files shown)


site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+[v0.1]:https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.1.0
+[v0.2]:https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.2.0
+
+## Project References
+
+In this section you will find specific links to blog posts, documentation and other Gateway API references for specific implementations.
+
+### Contour
+
+[Contour][contour] is an open source ingress controller for Kubernetes.
```
shaneutt, line 34: TODO: need to verify that this is how the contour maintainers want to be represented

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+Contour currently implements the latest `v0.2.x` Gateway API Specification.
```
shaneutt, line 36: I didn't see a `v0.1.0` implementation, but I could have missed it?

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+See the [Contour Gateway API Guide][contour-1] for information on how to deploy and use Contours's Gateway implementation.
```
shaneutt, line 38: We need to see if the Contour maintainers want to add more resources here, docs, blogs, e.t.c.

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+[contour]:https://projectcontour.io
+[latest]:https://doc.traefik.io/traefik/routing/providers/kubernetes-gateway/
+[contour-1]:https://projectcontour.io/guides/gateway-api/
+
+### Google Cloud Platform
+
+The [Google Cloud Platform (GCP)][gcp] is a cloud computing platform and infrastructure provider.
```
shaneutt, line 46: @robscott is that how you and your team want to be represented? Please do feel free to make direct edits to get it the way you want it.

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+GCP is actively working towards an implementation of the Gateway API `v0.2.x` specification, and status updates and documentation will be provided here as the work progresses.
```
shaneutt, line 48: @robscott I wasn't sure if you had any links to documentation, google docs, or blog posts you wanted to provide here yet?

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+[gcp]:https://cloud.google.com
+
+### Istio
+
+[Istio][istio] is an open source [service mesh][mesh] for Kubernetes clusters.
```
shaneutt, line 54: TODO: need to verify that this is how the Istio maintainers want to be represented

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+Istio currently supports the previous `v0.1.x` Gateway API specification and is working towards supporting the latest `v0.2.x` release.
+
+Status updates and documentation for `v0.2.x` will be provided here as the work progresses.
```
shaneutt, line 58: Need to verify this, as I was just guessing when I made the PR draft, I'll do some more digging before I take this out of draft and make it ready for review.

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+You can subscribe to [kong/kubernetes-ingress-controller/issues/692][kong-1] to track the implementation progress and contribute.
+
+[kong]:https://konghq.com
+[kic]:https://github.com/kong/kubernetes-ingress-controller
+[kong-1]:https://github.com/kong/kubernetes-ingress-controller/issues/692
+
+### Knative
+
+[Knative][knative] is an open source Kubernetes-based platform to deploy and manage modern serverless workloads.
```
shaneutt, line 80: TODO: again need to make sure the Knative maintainers are on board with this representation. Also I'm _not sure_ whether they're actively working on an implementation or planning to. I'll do more digging before taking this out of draft state.

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+[knative]:https://knative.dev/
+
+### Traefik
+
+[Traefik][traefik] is an open source cloud-native application proxy.
+
+Traefik currently supports the previous `v0.1.x` Gateway API specification, check the [Kubernetes Gateway Documentation][traefik-1] for information on how to deploy and use Traefik's Gateway implementation.
+
+Traefik is currently working on implementing `v0.2.x`, status updates and documentation will be provided here as the work progresses.
```
shaneutt, line 92: Similar to the above, need to check in with Traefik maintainers that they're happy with this representation.

shaneutt (commented):
When I first made this doc it felt sort of odd to do write ups for other projects (I represent [Kong](https://github.com/kong)), but I tried to pull information from previous conversations in SIG networking zoom syncs and the K8s slack channel.

@robscott I would appreciate your review on this prior to moving it out of draft phase, perhaps I should stick with yours and mine implementations only for now? Not really sure since I'm new to the project, let me know :vulcan_salute:

site-src/implementations.md
robscott: @danehans @jpeach @youngnick Does this look good to you?

site-src/implementations.md
SantoDE, line 92: + TCP :)

SantoDE: Apart from that Traefik is good. Thanks a lot!

I was about to do the same this weekend 🥳

SantoDE (commented):
See comments :)

site-src/implementations.md
robscott: Nothing public yet, but hope to have something out soon. I think this is good for now.

site-src/implementations.md
robscott: I think "Google Cloud" is the preferred term now. I'll let @mark-church weigh in on what we'd like to include here. I think we'll likely also want to add some GKE specific reference, but not quite sure how to integrate that.

site-src/implementations.md
howardjohn: we are importing the v0.2.x code now, but its in our `master` branch not a release yet and our docs aren't updated yet to point to the new "Gateway" name. Not sure what emoji that turns into :slightly_smiling_face:

site-src/implementations.md
robscott: @nak3 Do you have any insight you can provide here?

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+[kong]:https://konghq.com
+[kic]:https://github.com/kong/kubernetes-ingress-controller
+[kong-1]:https://github.com/kong/kubernetes-ingress-controller/issues/692
+
+### Knative
+
+[Knative][knative] is an open source Kubernetes-based platform to deploy and manage modern serverless workloads.
+
+Knative is actively working towards an implementation of the Gateway API, status updates and documentation will be provided here as the work progresses.
```
howardjohn, line 82: I think its a bit misleading? Knative is a user of the gateway API, not an implementation, afaik

site-src/implementations.md
stevesloka: We could also add some more info about the Contour Operator which helps to deploy things. @danehans

site-src/implementations.md
stevesloka: We started with v0.1.0 but never released it, so v0.2.0 is technically accurate. Also, I'm not sure that we're "fully implemented", it's very much a work in progress.

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+- :heavy_check_mark: available
+- :heavy_plus_sign: in progress
+- :heavy_minus_sign: not planned
+
+[1]:#contour
+[2]:#google-cloud-platform
+[3]:#istio
+[4]:#kong
+[5]:#knative
+[6]:#traefik
```
robscott, line 23: Although I like this table, our docs site rendering does not :(. Maybe we can just turn this into a list of implementations with approximate state in parentheses. Something like:

```
- [Contour][1] (alpha)
- [Google Cloud][2] (in progress)
...
```

site-src/implementations.md
```diff
@@ -0,0 +1,95 @@
...
+
+Knative is actively working towards an implementation of the Gateway API, status updates and documentation will be provided here as the work progresses.
+
+[knative]:https://knative.dev/
+
+### Traefik
+
+[Traefik][traefik] is an open source cloud-native application proxy.
+
+Traefik currently supports the previous `v0.1.x` Gateway API specification, check the [Kubernetes Gateway Documentation][traefik-1] for information on how to deploy and use Traefik's Gateway implementation.
```
SantoDE, line 90: We could add a hint about enabling and installing it with the helm chart. Not 100% sure this is the correct place tough :)

site-src/implementations.md
shaneutt:
```suggestion
Traefik is currently working on implementing TCP, status updates and documentation will be provided here as the work progresses.
```

shaneutt: Just double checking this is what you wanted? Otherwise please drop a suggestion and I'll be happy to commit it :bow:

site-src/implementations.md
shaneutt: Roger that, I'll find someone from Knative and work on getting this written more appropriately for their liking :+1:

site-src/implementations.md
shaneutt: Ok thanks for the heads up! @robscott had suggested maybe dropping the implementation version stuff so perhaps I'll just go with that and then individual maintainers can put whatever they think are important notes about their implementations in their sections?

site-src/implementations.md
shaneutt: Sounds good, _please_ do feel free to add github suggestions for what to change I'm happy to commit them I'm really seeing all of these sections as placeholders that eventually the maintainers will want to add their own writeup for once we have a loose format down anyhow :+1:

site-src/implementations.md
shaneutt:
```suggestion
The [Google Cloud][gcp] is a cloud computing platform and infrastructure provider.
```

site-src/implementations.md
shaneutt: :+1:

site-src/implementations.md
shaneutt: Understood, thanks for the heads up! I'll try to get this updated to better reflect the state of things.

site-src/implementations.md
shaneutt: Sure, I'll change it to that format instead, and perhaps we can wait until further down the road to see if people care to have a version implementation table, since it wouldn't seem to be all too important right now. :+1:

site-src/implementations.md
shaneutt: Sounds fine to me, is there just a link to that installation you want to list here?

site-src/implementations.md
mark-church: @shaneutt thanks for contributing this!

Here's a short statement we can use:

```
The [Google Kubernetes Engine (GKE)][gke] is a managed Kubernetes platform offered by Google Cloud. 

GKE is actively working towards an implementation of the Gateway API for orchestration of [Google Cloud Load Balancing][gclb]. The GKE Gateway Controller will become a managed component of GKE to automate and ease load balancing for containers.

[gke]:https://cloud.google.com/kubernetes-engine
[gclb]:https://cloud.google.com/load-balancingThe [Google Kubernetes Engine (GKE)][gke] is a managed Kubernetes platform offered by Google Cloud. 
```

site-src/implementations.md
mark-church: @howardjohn not sure if this should clarify that the current Gateway API is for Istio Ingress but not currently Istio sidecar mesh

site-src/implementations.md
nak3: Yeah, as @howardjohn also mentioned, Knative is an user of the gateway API but not an implementation. Knative is going to support the implemented Ingress like Istio, Kong, Contour etc.

site-src/implementations.md
nak3: Yes, It is a little bit weird that Knative is listed in the Implementation. Knative is going to support the Ingress which implemented gateway-api, so Knative is a kind of "user" rather than "implementation" as John Howard also mentioned below.

site-src/implementations.md
shaneutt: I think we should just add more context in a follow up, if that's OK with you? Since this PR is mainly about adding the base page and getting some general information down. Let me know what you think, and feel free to unresolve and let me know your preference.

site-src/implementations.md
shaneutt: Sounds good, thank you! I've updated to match your contents: 267f6fbfea56821cbaeb1241c1842333abfb5ac9

site-src/implementations.md
shaneutt: I've updated this to be simpler and then more context can be added later in future iterations, lmkwyt :+1: 

5f51da790e097b5030548838e5bd51abceadc3fb

site-src/implementations.md
shaneutt: Ok, for now I've opted to remove it from the doc for now and then we can add further context as needed in other iterations: 7781c9acbc944c0c05270bc322bd2ba24509a17c

site-src/implementations.md
shaneutt: I've removed Knative for now as per above comments: 7781c9acbc944c0c05270bc322bd2ba24509a17c

We can come back around to documenting it further in a later iteration :+1:

site-src/implementations.md
shaneutt: For the time being I'm going to pass just so we can get the base page down, but we can definitely add further context as follow up :+1:

site-src/implementations.md
shaneutt: Based on updates from @stevesloka below, I believe we should be all set now. Let me know if there's anything you still feel needs to be changed!

site-src/implementations.md
shaneutt: Based on below responses I think we should be OK with Istio, but let me know if we're not!

shaneutt:
At this point I think we've received enough maintainer feedback and updated for that feedback that this PR is ready to move from draft to ready. Please let me know any other considerations, or changes that you feel need to be made! :vulcan_salute:

hbagdi:
Let's see if we can get an approval from maintainers from each implementation:

Contour: @stevesloka
GKE: @robscott @bowei
Istio: @howardjohn 
Kong: Approved by mysefl and @shaneutt 
Traefik: @SantoDE

howardjohn (approved):

site-src/implementations.md
```diff
@@ -0,0 +1,83 @@
+# Implementations
+
+This document tracks downstream implementations of Gateway API and provides status and resource references for them.
+
+Implementors of Gateway API are encouraged to update this document with status information about their implementations, the versions they cover, and documentation to help users get started.
+
+## Implementation Status
+
+- [Contour][1] (alpha)
```
bowei, line 9: can we make this a table?

```
| Implementation | Status |
| ---- | ---- | 
etc
```

bowei: Also, if we can put the version implemented (instead of just Alpha)
that would be useful for users

bowei: e.g. Supports `v1alpha1`

bowei:
I put a hold -- let's collect ack from the impls

This looks great

site-src/implementations.md
shaneutt: When this was in draft it started out as a table, see:

https://github.com/kubernetes-sigs/gateway-api/pull/565#discussion_r583944465

However @robscott had particularly requested that I change it to the format you see here.

If Rob is OK with changing it back to a table but using the format you suggest above, that sounds good to me. Just want to check with him.

robscott:
Just following up on Harry's summary from above, looks like we have most approvals, but still need a couple more. We currently have approvals for:
- [x] GKE
- [x] Istio
- [x] Kong

We still need approval from:
- [] Contour: @stevesloka
- [] Traefik: @SantoDE

/assign @stevesloka @SantoDE

site-src/implementations.md
bowei: This is a minor thing and we can change in a later diff. My thinking was that table is easier to deal with more info (say which versions are supported)

mark-church: I also think a table would be a bit cleaner (though we don't have to block this PR). The original table had the vendors as columns instead of rows which doesn't leave much room for written content. The following format would allow us to take the existing written content we have in the page and surface it through a table.

```
| Implementation | Status |
| [Project/Product](link) | Explanation of progress towards support and/or support levels (which versions) | 
| [Project/Product](link) | Explanation of progress towards support and/or support levels (which versions) | 
etc
```

I think support is a gray area right now since we are early in the spec so it's prob too early for a clean cut version/checkmark box.

site-src/implementations.md
```diff
@@ -0,0 +1,83 @@
...
+- [Google Cloud][2] (work in progress)
```
mark-church, line 10: Should be:

```
- [Google Kubernetes Engine][2] (work in progress)
```

PR #572, opened 2021-03-03
api: Update Gateway.Spec.Listener[x].Routes.Selector to be a pointer

### LINKED ISSUE #571: Gateway.Spec.Listener[x].Routes.Selector should be a pointer
**What happened**:

When implementing GatewayAPI, I noticed the `selector` field on a Gateway.Spec.Listener[x].Routes.Selector (https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteBindingSelector) is currently not a pointer, but nice if it was to know if the field has been provided or not by a user. 

I’m using the [LabelSelectorAsSelector](https://pkg.go.dev/k8s.io/apimachinery@v0.20.4/pkg/apis/meta/v1#LabelSelectorAsSelector) method from api-machinery to then use [labels.Matches(…)](https://pkg.go.dev/k8s.io/apimachinery/pkg/labels), but `LabelsSelectorAsSelector` takes a pointer to the Selector. 

//cc @danehans

stevesloka:
**What type of PR is this?**
/kind bug
/kind api-change

The Gateway.Spec.Listener[x].Routes.Selector is not a pointer and should be.

Fixes #571

```release-note
Update Gateway.Spec.Listener[x].Routes.Selector to be a pointer.
```

//cc @danehans 

Signed-off-by: Steve Sloka <slokas@vmware.com>


stevesloka:
I'm struggling to get the docs to regen. My mac just won't get the right bits to work. Could we have an optional docker build step to run those in docker removing the need for local deps?

robscott:
@stevesloka No need for docs generation at all anymore, we've moved to Netlify. Deploy previews are available from "deploy/preview" presubmit (this one is https://deploy-preview-572--kubernetes-sigs-gateway-api.netlify.app/). 

/ok-to-test

stevesloka:
> No need for docs generation at all anymore

w00t! Thanks @robscott! =)

robscott:
Unfortunately k8s infra doesn't like `fixes` keywords in commit messages so this can't be merged until that is removed. (https://github.com/kubernetes/test-infra/issues/9360)

stevesloka:
Sorry, it's a habit @robscott. I updated the commit message to remove that.

hbagdi:
Like #564, this is another case of an optional struct should be a pointer. 
We discussed previously that we need to audit the entire code to ensure this is the case.
I think we should certainly do that but such small improvements are also good.
/lgtm

PR #575, opened 2021-03-08
Index page refresh and "why this matters" content

mark-church:
/kind documentation

As the Gateway API grows up, we also need documentation that is more approachable for new users (and not just controller implementers). This is a refresh of the index page that adds "why this matters" content. Here are some specific additions:

- Why a role-oriented API matters - this is designed to provide the reader with a very tangible example of how Gateway is different than Ingress. The existing graphic in the page is high-level and great to explain that Gateway is a collection of new API resources, but it does not explain the significance or relationship of these resources and why any of this should matter to anybody. The example is designed to introduce many of the core concepts (route binding, cross-namespace Gateways, traffic splitting etc) with links to those pages as a teaser. I think this is appropriate for the index page because it is very tangiblle.
- Why portability matters - a short description behind the philosophy of portability and extensibility that Gateway has. This effectively answers, why do we even need Yet Another API?
- Update of the getting started list - I just structured this into a cleaner list 
- Who is working on Gateway - this creates a reference to the implementers page so this PR has a dependency on https://github.com/kubernetes-sigs/gateway-api/pull/565

This is the image used for the example:
image


site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
+
+The Gateway API is a Kubernetes
+[SIG-Network](https://github.com/kubernetes/community/tree/master/sig-network)
+project being built by a variety of companies and users. Current and
+in-progress implementations include Contour, Google Kubernetes Engine (GKE),
+Istio, Kong, and Traefik. Check out the [implementations
+reference](implementations.md) to see the latest projects & products that
+support Gateway. If you are interested in contributing to or building an
+implementation using the Gateway API then don’t hesitate to [get
+involved!](#how-to-get-involved)
```
robscott, line 115: This has to match section header. I don't have a strong preference between the two, but since the section header is changing here, this should to.
```suggestion
involved!](#getting-involved)
```

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
-- [Go docs for the package](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1alpha1)
+- Start with the [API overview](api-overview.md) to understand the basics
+- Follow through the [user guides](guides.md) to see what use-cases Gateway 
+addresses
+- Check out the [Gateway controller implementations](implementations.md) to 
+get hands-on experience in your Kubernetes cluster
+- See the [API reference](spec.md) for a detailed breakdown of the API surface 
+area
+- If you’re ready to start developing with Gateway, check out the 
+[Go docs for the package](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1alpha1)
```
robscott, line 133: This is a huge improvement! I think this may be one of the most helpful parts of the page, I just hope it doesn't get missed due to it being so far down the page.

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
-
-- **Shared Gateways** - They allow the sharing of load balancers and VIPs by
-permitting independent Route resources to bind to the same Gateway. This allows
-teams to share infrastructure safely without requiring direct coordination.
-- **Typed backend references** - With typed backend references Routes can
-reference Kubernetes Services, but also any kind of Kubernetes resource that is
-designed to be a Gateway backend.
-- **Cross-Namespace references** - Routes across different Namespaces can bind
-to Gateways. This allows for shared networking infrastructure despite Namespace
-segmentation for workloads.
```
robscott, line 50: Although the diagram/text below shows this, I really liked highlighting these points. I think these are some of the more compelling parts of this API.

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
+with flexibility in how much of the Gateway API they want to support. In
+combination with various extension points within the API, this allows the core
+functionality of Gateway to be portable, while still surfacing
+implementation-specific capabilities.
+
+### Who is working on Gateway?
+
+The Gateway API is a Kubernetes
+[SIG-Network](https://github.com/kubernetes/community/tree/master/sig-network)
+project being built by a variety of companies and users. Current and
```
robscott, line 109: This is a tiny nit, but "a variety of companies and users" doesn't feel quite right. I can't think of a great alternative, but maybe we can just leave it as a "Kubernetes SIG-Network project"? Feel free to ignore this as well, maybe this is a non issue.
```suggestion
project. Current and
```

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
+applications through it. Centralized policies [such as
+TLS](tls.md#downstream-tls) can be enforced on the Gateway by the cluster
+operator. Meanwhile, the store and site teams run [in their own
+Namespaces](multiple-ns.md), but bind their Routes against the same shared
+Gateway, allowing them to independently control their [routing
+logic](http-routing.md). This separation of concerns allows the store team to
+manage their own [traffic splitting rollout](traffic-splitting.md) while
+leaving centralized policies and control to the cluster operators.
+
+![Gateway API Roles](./images/gateway-roles.png)
```
robscott, line 74: This is a great diagram. Combined with the earlier diagram above, this provides a great demonstration of what this API can do.

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
-- **Typed backend references** - With typed backend references Routes can
-reference Kubernetes Services, but also any kind of Kubernetes resource that is
-designed to be a Gateway backend.
-- **Cross-Namespace references** - Routes across different Namespaces can bind
-to Gateways. This allows for shared networking infrastructure despite Namespace
-segmentation for workloads.
-- **Classes** - GatewayClasses formalize types of load balancing implementations.
-These classes make it easy and explicit for users to understand what kind of
-capabilities are available as a resource model itself.
+### Why does a role-oriented API matter?
```
robscott, line 44: Both this section and the one below are great and well written, but they might be a bit too much content for the main page. Maybe a new "Why Gateway?" page or similar?

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
-- **More expressive** - They express more core functionality for things like
-header-based matching, traffic weighting, and other capabilities that were only
-possible in Ingress through custom means.
-- **More extensible** - They allow for custom resources to be linked at various
-layers of the API. This allows for more granular customization at the
-appropriate places within the API structure.
-- **Role oriented** - They are separated into different API resources that map
-to the common roles for running applications on Kubernetes.
-- **Generic** - This isn't an improvement but rather something
+Following are some of the design goals where Gateway aims to improve upon
```
robscott, line 23: Nit: I think it's still worth referring to this as "Gateway API" over Gateway in most docs just so we don't confuse users with the Gateway resource. That will likely still be somewhat confusing, but maybe slightly less so.

site-src/index.md
```diff
@@ -20,54 +20,119 @@ variety of networking use-cases.
...
-- **Generic** - This isn't an improvement but rather something
+Following are some of the design goals where Gateway aims to improve upon 
+current standards like Ingress.
+
+- **Role-oriented** - Gateway is composed of API resources which model 
+organizational roles that use and configure Kubernetes service networking. 
+A role-oriented API allows diverse teams to operate on shared network 
+infrastructure. Gateway is designed to balance the flexibility and control of 
+these roles in a way that reflects the culture and structure of the organization
+ using it.
```
robscott, line 31: Although this content is good, my personal preference would be to keep content like this a bit more brief on the main page.

robscott (commented):
Thanks for the work on this @mark-church! There's some great content in here.

site-src/index.md
mark-church: yeah good point. I'll update API references to "Gateway API"

site-src/index.md
mark-church: I slimed it down to about 50% of what it was before, lmk if that's small enough

site-src/index.md
mark-church: I brought this back and also cleaned it up a bit by converging it to 3 bullet points

site-src/index.md
```diff
@@ -80,3 +80,55 @@ contribution is too small - even opinions matter!
...
+
+
+## Why does a role-oriented API matter?
+
+Whether it’s roads, power, data centers, or Kubernetes clusters,
+infrastructure is built to be shared. However, shared infrastructure raises a
+common challenge - how to provide flexibility to users of the infrastructure
+while maintaining control by owners of the infrastructure? 
+
+The Gateway API accomplishes this through a role-orietned design for
```
pmallory, line 107: typo: orietned

robscott:
This is great, thanks for all the work on this!

/lgtm
/approve

PR #579, opened 2021-03-10
Add the pointer for optional fields.

### LINKED ISSUE #576: Pointers need to be used consistently for optional fields
As noted in #564 and #572 we have not consistently been using pointers throughout this API. We should be using pointers for all fields that are optional and not a map or slice. That includes if the field has a default value.

Hyzhou:
Signed-off-by: HongYu Zhou <misakazhou@tencent.com>

/kind cleanup

Fixed all the case about pointers need to be used consistently for optional fields.

Fixes #576 

NONE


robscott:
Thanks @Hyzhou! This matches my understanding of how we should be handling optional fields. Will leave final LGTM for someone else.

/approve
/cc @danehans @hbagdi

PR #581, opened 2021-03-10
Fix example as per discussion (should be port, not targetPort)

bowei:
/kind documentation

```release-note
NONE
```

### DIFF
--- a/examples/upstream-tls.yaml
+++ b/examples/upstream-tls.yaml
@@ -25,10 +25,8 @@ spec:
   ports:
   - name: http
     port: 80
-    targetPort: 8000
   - name: https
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
my 1c/nit: having targetPort in the Service actually shows that the port is the ServicePort. This makes it ambiguous.

robscott:
> having targetPort in the Service actually shows that the port is the ServicePort

I agree with this, but think this might be better to solve/document separately. Since this specific example is for upstream TLS, I think we should keep it as simple as possible.

robscott:
Thanks!

/lgtm

PR #584, opened 2021-03-14
Add default value "/" for HTTPPathMatch

nak3:

`HTTPPathMatch` has a default `prefix match on the "/" path`.
For example, if `match` was not configured, the default prefix match with value will be filled.

(This configuration works.)
```
spec:
  rules:
```

However, once we added `type` such as `Prefix` or `Exact`, the `value`
field must be filled by ourselves.

(This configuration does not work due to missing `value` field.)
```
spec:
  rules:
    matches:
    - path:
        type: Prefix
```

Hence this patch adds the default `/` value for `path.value`.

**What type of PR is this?**

/kind api-change

```release-note
HTTPPathMatch path.value sets default value "/" when type was configured but value was not.
```


robscott:
Thanks! This makes sense to me, assigning to @danehans to be sure.

/ok-to-test
/lgtm
/assign @danehans

apis/v1alpha1/httproute_types.go
```diff
@@ -280,7 +280,8 @@ type HTTPPathMatch struct {
 
 	// Value of the HTTP path to match against.
 	//
-	// +kubebuilder:validation:MinLength=1
+	// +optional
+	// +kubebuilder:default="/"
```
danehans, line 284: @nak3 since this field is now being defaulted, the `omitempty` json tag should be included.

danehans:
@nak3 thanks for catching this issue. Since `value` is being defaulted at higher layers of the API, it should be defaulted and be set to optional directly on the field.

/lgtm when https://github.com/kubernetes-sigs/gateway-api/pull/584#discussion_r596253461 is resolved.

apis/v1alpha1/httproute_types.go
nak3: Thank you! Updated.

apis/v1alpha1/httproute_types.go
```diff
@@ -280,8 +280,9 @@ type HTTPPathMatch struct {
 
 	// Value of the HTTP path to match against.
 	//
-	// +kubebuilder:validation:MinLength=1
-	Value string `json:"value"`
+	// +optional
+	// +kubebuilder:default="/"
+	Value string `json:"value,omitempty"`
```
robscott, line 285: Since this is now optional I think it also needs to be a pointer. (cc @danehans to be sure)

danehans: +1

nak3: Updated. Thank you!

robscott:
Thanks!

/lgtm
/approve

PR #587, opened 2021-03-18
Adding new AdmittedGateways field to Routes to improve kubectl output

### LINKED ISSUE #586: Add GATEWAYS field to kubectl get httproute output
Add a GATEWAYS field to the `kubectl get httproute` output. Thanks to the feedback from @joshrosso I realize it's challenging to easily understand which Gateways a Route is bound to or whether it's bound at all. This field would key off of the HTTPRoute status which indicates the Gateways that a Route is bound to: https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.RouteGatewayStatus

The below example shows HTTPRoutes, some which are bound to different Gateways, some bound to multiple Gateways, and one that's not bound to any Gateway (which is probably an accident that the deployer doesn't know yet). This would be really useful because it provides a quick snapshot of all the Gateways being used in a certain Namespace without having to dig into each Route individually. It also provides a very quick check to validate that a Route has been bound to a Gateway successfully after creation.

```
$ kubectl get httproute

NAME         HOSTNAMES                    GATEWAYS
route1       ["bar.example.com"]          prod-gw
route2       ["bar.example.com"]          dev-gw
route3       ["foo.example.com"]          dev-gw, staging-gw
route4       ["foo.example.com"]
```
cc @robscott @bowei @stevesloka @hbagdi

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This adds a new `AdmittedGateways []string` field in route status to list all Gateways that have admitted the route. This field is used in kubectl output to make it clearer which Gateways are bound to a given Route.

Fixes #586 

```release-note
Routes now include an admittedGateways field in status.
```


apis/v1alpha1/shared_types.go
```diff
@@ -211,6 +211,19 @@ type RouteStatus struct {
...
+	// namespace/name format. This field is primarily used to provide kubectl
+	// output.
+	//
+	// A maximum of 100 Gateways will be represented in this list. If this list
+	// is full, there may be additional Gateways using this Route that are not
+	// included in the list. An empty list means the route has not been admitted
+	// by any Gateway.
+	//
+	// +kubebuilder:validation:MaxItems=100
+	AdmittedGateways []string `json:"admittedGateways"`
```
robscott, line 226: I'm not tied to this name. I also considered `AdmittingGateways`, `ConnectedGateways`, or `ActiveGateways` if any of those seem better. Also open to other names.

stevesloka: I guess the question is what are we trying to convey? If the status of the object was `Admitted: false` because there's some sort of error, should it still show in this list? 

If that's the case then maybe `admittedGateways` isn't the right name (but was my first thought as well). =)

Possibly `selectedGateways` but that implies the opposite ordering (the route doesn't select the gateway).

apis/v1alpha1/shared_types.go
```diff
@@ -211,6 +211,19 @@ type RouteStatus struct {
 	//
 	// +kubebuilder:validation:MaxItems=100
 	Gateways []RouteGatewayStatus `json:"gateways"`
+
+	// AdmittedGateways is a list of Gateways that have admitted this route.
```
bowei, line 215: Should we make sure that there is one way language we talk about this relationship? Is it "admitted" or "bound" (both terms exist in the repo).

Ask: can we decide on one and change all references to be consistent?

SantoDE: I like bound more :)

apis/v1alpha1/shared_types.go
```diff
@@ -211,6 +211,19 @@ type RouteStatus struct {
...
+	// This list includes the namespace and name of each Gateway in a
+	// namespace/name format. This field is primarily used to provide kubectl
+	// output.
+	//
+	// A maximum of 100 Gateways will be represented in this list. If this list
```
bowei, line 220: Can we leave the #'s out of this comment? We should just say that this list may be truncated and also include an indicator for truncation:

`AdmittedGatewaysTruncated bool`

or something similar, maybe 

`AdmittedGatewaysCount int32`

that people can see total even though it was truncated.

apis/v1alpha1/shared_types.go
danehans: > Should we make sure that there is one way language we talk about this relationship?

+1, we currently use "Admitted" and should continue doing so IMO.

apis/v1alpha1/shared_types.go
```diff
@@ -211,6 +211,19 @@ type RouteStatus struct {
 	//
 	// +kubebuilder:validation:MaxItems=100
 	Gateways []RouteGatewayStatus `json:"gateways"`
```
danehans, line 213: Should `Gateways` still exist? It seems redundant.

robscott: The big difference is that this list of Gateways includes all the conditions/status for the associated Gateways. In this PR I'm proposing adding a simpler `[]string` for the purpose of populating kubectl output, but the full status/conditions would still be valuable.

danehans: @robscott does kubectl provide a way to iterate through the gateway ref list and output gateways of "Admitted=True"?

robscott: Even if we were able to get a fix into Kubernetes 1.22, the kubectl experience would be quite poor for anyone on an older k8s version since the current implementation always returns the first element in the matching list.

apis/v1alpha1/shared_types.go
robscott: Yeah I think the most useful thing we can show here for kubectl is the Gateways that are actively using this Route. I was thinking that if a user expects a Gateway to be in the list and it isn't they can dig into the full yaml output/status.

hbagdi:
@robscott Did we land on a decision on if we want to include this feature or hold it?

robscott:
@hbagdi Thanks for following up! This is on me to follow up with apimachinery to see what's possible/recommended for these kinds of things.

robscott:
Sent this out to sig-api-machinery to see if they have any insight. https://groups.google.com/g/kubernetes-sig-api-machinery/c/GxXWe6T8DoM

robscott:
I've created an [upstream PR](https://github.com/kubernetes/kubernetes/pull/101205) that should help with this. Of course that's probably only a realistic solution if we can also backport the fix to earlier k8s versions. For now I'm going to put a hold on this until there's more clarity upstream.

/hold

apis/v1alpha1/shared_types.go
youngnick: I agree that `Admitted` doesn't seem right, but I can't think of anything else that doesn't sound bad. `InUseByGateways` is probably accurate, but a bit of a mouthful.

apis/v1alpha1/shared_types.go
jpeach: So this is the names of the gateways that are included in the `Gateways` array and have an `Admitted` status of true? If so, it's redundant but I can see your point here.

The doc should specify the semantic described here. I don't think that the intention is to invent new semantics for this field; we just want to make existing semantics consumable by kubectl. We should avoid the word "bound" since that's not a term of art in the API.

robscott:
I think the correct approach is to try to solve this upstream, even though it will take more time. I'm closing this in favor of https://github.com/kubernetes/kubernetes/pull/101205 for now. If I can't get momentum upstream I'll revisit making changes like this.

/close

Issue #588, opened 2021-03-19
Add more columns to CRDs


howardjohn:
**What would you like to be added**:
We should add more columns to the CRDs to expose useful information in `kubectl`. For example, we can expose Admitted for most types:
```
$ k get gc
NAME    CONTROLLER                    ADMITTED
istio   istio.io/gateway-controller   True
```

I would also suggest adding AGE to align with most other k8s types.
**Why is this needed**:
Useful to expose info here.

howardjohn:
Related: https://github.com/kubernetes-sigs/gateway-api/issues/586

robscott:
Agree that this would be great to have. #592 adds a "Created At" column, but I think "Admitted" is going to be more difficult here. Unfortunately `additionalPrinterColumns` just does not handle status conditions (or other lists) well. We could follow the same approach I proposed in #587 and add a new field to status just to improve `kubectl get` output. Alternatively we could push to get better support built into upstream. Of course if this API itself ever does end up becoming built into k/k, many of these problems would go away on their own.

howardjohn:
You can do it with `.status.conditions[?(@.type=="Admitted")].status`

Also can do it directly like `kubectl get gatewayclass -o 'custom-columns=NAME:.metadata.name,HEALTHY:.status.conditions[?(@.type=="Admitted")].status'`

For the ones that are nested (listener status) I don't know how to handle though.

robscott:
@howardjohn Unfortunately I haven't been able to get that to work for `additionalPrinterColumns`. It seems that the implementation between `custom-columns` in kubectl and `additionalPrinterColumns` is quite different. For example, running this works:

```
kubectl get gateway -o'custom-columns=NAME:metadata.name,TEST:.spec.listeners[?(@.protocol=="HTTP")].routes.selector'
```

but unfortunately adding the same logic to `additionalPrinterColumns` just returns empty. I've seen this kind of similar pattern for anything other than a simple path lookup.

howardjohn:
Hm.. it seems like cert-manager is doing this: https://github.com/jetstack/cert-manager/pull/1194/files#diff-4cd218a9d6c5b00138ca643ec6b983015ba906cff00011340e798d668b9dfa39R27. Not sure what the gap is. The `additionalPrinterColumns` stuff is a bit frustrating

robscott:
[...]

PR #592, opened 2021-03-19
Adding gateway-api category and age column to CRDs

### LINKED ISSUE #589: Add categories to CRDs
Add something like:
```
    categories:
    - gateway-api
```

or similar to the CRD spec

**Why is this needed**:
This allows people to do `kubectl get gateway-api` and get all associated types, instead of needing to do `kubectl get gwc,gc,httproute,...`

robscott:
**What type of PR is this?**
/kind cleanup

This adds a gateway-api category to our CRDs as well as a new "age" column for "kubectl get" output.
Screen Shot 2021-03-19 at 3 06 04 PM

Fixes #589
Also inspired by #588 

```release-note
* Gateway API CRDs are not part of the "gateway-api" category. Running "kubectl get gateway-api" now returns all Gateway API resources.
* Each Gateway API resource now includes an "Age" column in "kubectl get" output.
```

### DIFF
--- a/config/crd/bases/networking.x-k8s.io_tlsroutes.yaml
+++ b/config/crd/bases/networking.x-k8s.io_tlsroutes.yaml
@@ -10,13 +10,19 @@ metadata:
 spec:
   group: networking.x-k8s.io
   names:
+    categories:
+    - gateway-api
     kind: TLSRoute
     listKind: TLSRouteList
     plural: tlsroutes
     singular: tlsroute
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1alpha1
     schema:
       openAPIV3Schema:
         description: "The TLSRoute resource is similar to TCPRoute, but can be configured to match against TLS-specific metadata. This allows more flexibility in matching streams for a given TLS listener. \n If you need to forward traffic to a single target for a TLS listener, you could choose to use a TCPRoute with a TLS listener."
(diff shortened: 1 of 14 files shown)


config/crd/bases/networking.x-k8s.io_tlsroutes.yaml
```diff
@@ -10,13 +10,19 @@ metadata:
...
+    - gateway-api
     kind: TLSRoute
     listKind: TLSRouteList
     plural: tlsroutes
     singular: tlsroute
   scope: Namespaced
   versions:
-  - name: v1alpha1
+  - additionalPrinterColumns:
+    - jsonPath: .metadata.creationTimestamp
```
howardjohn, line 22: I recommend:
```
    - description: 'CreationTimestamp is a timestamp representing the server time
        when this object was created. It is not guaranteed to be set in happens-before
        order across separate operations. Clients may not set this value. It is represented
        in RFC3339 form and is in UTC. Populated by the system. Read-only. Null for
        lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata'
      jsonPath: .metadata.creationTimestamp
      name: Age
      type: date
```

This will match Kubernetes core types. I also noticed when using Age it gives relative time - maybe that is a special kubectl feature?

howardjohn: oh nevermind, probably its from `type`

robscott: Woops, somehow I thought I was already setting type correctly. Thanks for catching this!

howardjohn (approved):

stevesloka (approved):
Nice work! 🎉

hbagdi:
@robscott Any reason for leaving `BackendPolicy` out?
Other than that, LGTM.

robscott:
Thanks for catching that @hbagdi! BackendPolicy is included now.

youngnick:
This is really great, nice work @robscott.

Issue #487, opened 2020-12-02
Implement a Validating Webhook


robscott:
Although we've discussed this in a variety of previous community meetings, we didn't have an issue to track work on this. We've labeled [issues that need to be covered](https://github.com/kubernetes-sigs/service-apis/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fwebhook) with this webhook with `area/webhook`. Those include:

- [Support for setting a default GatewayClass](https://github.com/kubernetes-sigs/service-apis/issues/262)
- [Enforce GatewayClass controller immutability](https://github.com/kubernetes-sigs/service-apis/issues/189)
- [Validate HTTPRequestMirrorFilter](https://github.com/kubernetes-sigs/service-apis/issues/349)

This is likely not an exhaustive list, but should provide a good starting point. For simplicity, implementation of this webhook should likely live within this repo.

cmluciano:
I would would like to take this on if no one has started yet

/assign @cmluciano

cmluciano:
I have the webhook ready to be reviewed at https://github.com/kubernetes-sigs/service-apis/pull/506 .

@bowei @robscott Do we currently have a process for docker image management ?

robscott:
Thanks for the work on this @cmluciano! We don't currently have anything set up for docker image management, but I believe a PR to https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io could fix that. I'm not very familiar with the process but happy to help however I can.

cmluciano:
> Thanks for the work on this @cmluciano! We don't currently have anything set up for docker image management, but I believe a PR to https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io could fix that. I'm not very familiar with the process but happy to help however I can.

Thanks! So it sounds like we might need to merge the PR I have open first and then we can add a new job there.

danehans:
Will this issue resolve the following Gateway validation use cases:

1. Is `gateway.spec.listeners[x].hostname` valid according to the spec of the field, i.e. not an IP address?
2. If `gateway.spec.listeners[x].protocol` is “HTTPS” or “TLS” that gateway.spec.listeners[x].tls is specified?
3. If `gateway.spec.listeners[x].tls.mode` is “Terminate”, that `gateway.spec.listeners[x].tls. certificateRef ` is specified.
4. If `gateway.spec.listeners[x].tls. certificateRef ` is specified, that the referenced object exists?
[...]

PR #597, opened 2021-03-30
Introduces a Validation Package

danehans:
**What type of PR is this?**

/kind feature

This PR adds a validation package. It's needed so implementors no longer need to create a Gateway API validation package. This approach provides validation consistency across implementations, including the future validating webhook. The package is modeled after the Kubernetes [networking APIs validation package](https://github.com/kubernetes/kubernetes/tree/release-1.19/pkg/apis/networking/validation).

Partially Fixes # https://github.com/kubernetes-sigs/gateway-api/issues/487

```release-note
NONE
```

/cc @stevesloka @youngnick @jpeach

### DIFF
--- a/apis/v1alpha1/validation/validation.go
+++ b/apis/v1alpha1/validation/validation.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"net"
+	"strings"
+
+	gatewayv1a1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateGateway validates gw according to the Gateway API specification.
+// For additional details of the Gateway spec, refer to:
+//  https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Gateway
+func ValidateGateway(gw *gatewayv1a1.Gateway) field.ErrorList {
+	return validateGatewaySpec(&gw.Spec, field.NewPath("spec"))
+}
+
+// validateGatewaySpec validates whether required fields of spec are set according to the
+// Gateway API specification.
+func validateGatewaySpec(spec *gatewayv1a1.GatewaySpec, path *field.Path) field.ErrorList {
+	// TODO [danehans]: Add additional validation of spec fields.
+	return validateGatewayListeners(spec.Listeners, path.Child("listeners"))
+}
+
+// validateGatewayListeners validates whether required fields of listeners are set according
+// to the Gateway API specification.
+func validateGatewayListeners(listeners []gatewayv1a1.Listener, path *field.Path) field.ErrorList {
+	// TODO [danehans]: Add additional validation of listener fields.
+	return validateListenerHostname(listeners, path)
+}
+
+// validateListenerHostname validates each listener hostname is not an IP address and is one
+// of the following:
+//  - A fully qualified domain name of a network host, as defined by RFC 3986.
+//  - A DNS subdomain as defined by RFC 1123.
+//  - A wildcard DNS subdomain as defined by RFC 1034 (section 4.3.3).
+func validateListenerHostname(listeners []gatewayv1a1.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, h := range listeners {
+		// When unspecified, “”, or *, all hostnames are matched.
+		if h.Hostname == nil || (*h.Hostname == "" || *h.Hostname == "*") {
+			continue
+		}
+		hostname := string(*h.Hostname)
+		if ip := net.ParseIP(hostname); ip != nil {
+			errs = append(errs, field.Invalid(path.Index(i).Child("hostname"), hostname, "must be a DNS hostname, not an IP address"))
+		}
+		if strings.Contains(hostname, "*") {
+			for _, msg := range validation.IsWildcardDNS1123Subdomain(hostname) {
+				errs = append(errs, field.Invalid(path.Index(i).Child("hostname"), hostname, msg))
+			}
+		} else {
+			for _, msg := range validation.IsDNS1123Subdomain(hostname) {
+				errs = append(errs, field.Invalid(path.Index(i).Child("hostname"), hostname, msg))
+			}
+		}
+	}
+	return errs
+}
--- a/apis/v1alpha1/validation/doc.go
+++ b/apis/v1alpha1/validation/doc.go
@@ -1,0 +1,19 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -36,6 +36,8 @@ linters-settings:
     ignore-words:
 
 issues:
+  exclude:
+    - Using the variable on range scope `tc` in function literal
```
danehans, line 40: Required due to:
```
$ ./hack/verify-golint.sh
apis/v1alpha1/validation/validation_test.go:149:4: Using the variable on range scope `tc` in function literal (scopelint)
			tc.mutate(gw)
			^
apis/v1alpha1/validation/validation_test.go:151:11: Using the variable on range scope `tc` in function literal (scopelint)
			if len(tc.expectErrsOnFields) != len(errs) {
			       ^
apis/v1alpha1/validation/validation_test.go:152:59: Using the variable on range scope `tc` in function literal (scopelint)
				t.Fatalf("Expected %d errors, got %d errors: %v", len(tc.expectErrsOnFields), len(errs), errs)
```

xref: https://github.com/kyoh86/scopelint/issues/4#issuecomment-525540030

hbagdi:
We should figure out a way how to integrate this with https://github.com/kubernetes-sigs/gateway-api/pull/506.
/cc @cmluciano

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+
+	gatewayv1a1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateGateway validates Gateway objects.
+func ValidateGateway(gw *gatewayv1a1.Gateway) field.ErrorList {
```
bowei, line 30: we should think about repetition here (validation.ValidateXZY)

validate.Gateway?

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,75 @@
...
+	return ValidateListenerHostname(listeners, fldPath)
+}
+
+// ValidateListenerHostname validates each listener hostname is not an IP address and is one
+// of the following:
+//  - A fully qualified domain name of a network host, as defined by RFC 3986.
+//  - A DNS subdomain as defined by RFC 1123.
+//  - A wildcard DNS subdomain as defined by RFC 1034 (section 4.3.3).
+func ValidateListenerHostname(listeners []gatewayv1a1.Listener, fldPath *field.Path) field.ErrorList {
+	errs := field.ErrorList{}
```
bowei, line 54: var errs []field.ErrorList

(nil list is empty)

danehans:
> We should figure out a way how to integrate this with #506.

@hbagdi that's what I discussed with @robscott and @jpeach in https://github.com/kubernetes-sigs/gateway-api/issues/487. As mentioned in the description, I would expect the webhook implementation to be a consumer of this package. cc: @cmluciano

apis/v1alpha1/validation/validation.go
danehans: @bowei agreed. I modeled the package off https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go, which follows a similar naming scheme.

danehans:
@bowei thanks for the review. Commit `32c1d66` resolves https://github.com/kubernetes-sigs/gateway-api/pull/597#discussion_r604217258 and https://github.com/kubernetes-sigs/gateway-api/pull/597#discussion_r604217614.

.golangci.yml
bowei: can you get rid of the warning with a copy? 
`tc := tc`

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ValidateListenerHostname validates each listener hostname is not an IP address and is one
+// of the following:
+//  - A fully qualified domain name of a network host, as defined by RFC 3986.
+//  - A DNS subdomain as defined by RFC 1123.
+//  - A wildcard DNS subdomain as defined by RFC 1034 (section 4.3.3).
+func ValidateListenerHostname(listeners []gatewayv1a1.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, h := range listeners {
```
jpeach, line 55: suggest early breaks to improve readability:

```Go
for i, h := range listeners {
        if h.Hostname == nil {
                continue
        }
        switch *h.Hostname {
        case  "": 
                // is this really valid?
        case "*":
                // valid wildcard
        default:
                // parse and check
        }
} 
```

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// ValidateGateway validates Gateway objects.
+func ValidateGateway(gw *gatewayv1a1.Gateway) field.ErrorList {
+	return ValidateGatewaySpec(&gw.Spec, field.NewPath("spec"))
+}
+
+// ValidateGatewaySpec validates whether required fields of spec are set according to the
+// Gateway API specification.
+func ValidateGatewaySpec(spec *gatewayv1a1.GatewaySpec, path *field.Path) field.ErrorList {
```
jpeach, line 36: Does this really need to be exported? Better for everyone to just use `ValidateGateway` IMHO.

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,75 @@
...
+	"net"
+	"strings"
+
+	gatewayv1a1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateGateway validates Gateway objects.
```
jpeach, line 29: Can we elaborate here and give some indication on what sorts of rules this enforces?

danehans:
@jpeach thanks for the review. Commit `a8bd76c` resolves most of your feedback.

apis/v1alpha1/validation/validation.go
danehans: I provided a link to the gateway spec and include details in the functions that are validating specific fields.

apis/v1alpha1/validation/validation.go
danehans: ack

apis/v1alpha1/validation/validation.go
danehans: I refactored the conditional logic that I think makes it easier to follow.

danehans:
As a follow-on from today's community meeting, I did a [quick implementation](https://github.com/danehans/service-apis/commit/ee30b103ef2b5c60124fbda1b9829c7c789c643e) of the validation pkg as a proof-of-concept for #506, PTAL.

apis/v1alpha1/validation/validation.go
```diff
@@ -0,0 +1,78 @@
...
+//  - A fully qualified domain name of a network host, as defined by RFC 3986.
+//  - A DNS subdomain as defined by RFC 1123.
+//  - A wildcard DNS subdomain as defined by RFC 1034 (section 4.3.3).
+func validateListenerHostname(listeners []gatewayv1a1.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, h := range listeners {
+		// When unspecified, “”, or *, all hostnames are matched.
+		if h.Hostname == nil || (*h.Hostname == "" || *h.Hostname == "*") {
+			continue
+		} else {
```
bowei, line 61: Do not need nested else if there are no additional statements following the block:

```
  if (blah) {
    continue // <-- control flow exits block
  } else {
  }
  // <-- nothing here
}
```

prefer, it's easier to understand

```
  if (blah) { continue }
  // no else
  stmt;
  stmt;

bowei:
Small changes and then this should be able to go in...

danehans:
@bowei thanks again for the review. Commit `30ca87d` resolves your https://github.com/kubernetes-sigs/gateway-api/pull/597#discussion_r606302946.

PR #603, opened 2021-04-09
Add forwarding shortcut to Gateway Listener

### LINKED ISSUE #529: Add forwarding shortcut to Gateway Listener
This is a follow up from what we discussed at today's community meeting. During that meeting, option B was favored due to its simplicity and clear purpose. This issue is closely related to https://github.com/kubernetes-sigs/service-apis/issues/174, but takes a slightly different approach.

---

Adding a forwarding shortcut to Gateway listeners would allow users to create simple Gateways without the need for Routes. This would be especially useful for L4 Gateways where TCP or UDP Routes provide rather minimal value. 

In all cases, we do not want to recreate advanced routing behavior at this level. This forwarding shortcut will not include weighting, delegation to additional routes, or traffic splitting. It is meant for the simplest of use cases. Anything beyond that should make use of route resources.

**Option A: DefaultBackend with Ingress-like behavior**
In addition to the L4 shortcut use case, this could also serve a similar purpose to the default backend that is part of the Ingress API. This could be used if no matching routes were found.

```yaml
listeners:
  - protocol: "TCP"
    port: 80
    defaultBackend:
      serviceName: "example"
      port: 8080
```

**Option B: BackendRef with exclusive behavior**
Alternatively, this field could be mutually exclusive with a route selector. When a backendRef was specified, the route selector would be ignored. This would dramatically simplify config for simple load balancers, most commonly L4 protocols.

```yaml
listeners:
  - protocol: "TCP"
    port: 80
    backendRef:
      serviceName: "example"
      port: 8080
```

**Protocol**
The upstream protocol would be determined by the app protocol or protocol defined by the backend, or if not specified at that level, the protocol defined by the listener.

Listener Protocol | Backend Port Protocol | Backend App Protocol | Selected Protocol
-- | -- | -- | --
HTTPS | HTTP |   | HTTP
HTTPS | HTTP | acme.io/custom | acme.io/custom
HTTP | HTTPS |   | HTTPS
TCP | HTTP |   | HTTP
TCP |   |   | TCP

stevesloka:
**What type of PR is this?**
/kind feature
/kind api-change

Adds a backendRef to a Gateway Listener to allow a direct forwarding shortcut to a serviceName/port.

Fixes #529 

```release-note
Add forwarding shortcut to Gateway Listener.
```

Signed-off-by: Steve Sloka <steve@stevesloka.com>


stevesloka:
I have a couple questions on this:

- What should the support on this be? I left as "Custom". 
- I only left a `serviceName`/ `port` in the spec, wasn't sure if a `LocalObjectReference` was desired. 
- What should the status be if the service is invalid? I added `DegradedListener` since `DegradedRoutes` didn't make sense.

apis/v1alpha1/gateway_types.go
```diff
@@ -234,6 +234,41 @@ type Listener struct {
...
+	//
+	// If the referent cannot be found, the controller should raise
+	// the "ResolvedRefs" condition on the Gateway with the "DegradedListeners"
+	// reason. The gateway status should be updated with a condition
+	// that describes the error more specifically.
+	//
+	// Support: Custom
+	//
+	// +optional
+	BackendRef *GatewayBackendRef `json:"backendRef,omitempty"`
```
howardjohn, line 248: This seems inconsistent with route's forward to, which is
```
forwardTo:
  service: ..
  backendRef: ...
  port: ...
```

Since now we have a backendRef with service and port inside of it?

We also have backendRef in backendpolicy which is different as well, which I worry may confuse users

stevesloka: I agree with you @howardjohn. I'll add that back since it felt weird to not have it there, however, it also seemed like this was a simple shortcut path proposal which might warrant not having it there.

hbagdi: I think not including `backendRef` is better. Let's start with the bare minimum and then include other resources as required.
Custom backends at this level does make me think if we are adding one more extension point in the Gateway resource than needed. This is really a shortcut for a specific problem and I say let's keep it short.

apis/v1alpha1/gateway_types.go
```diff
@@ -234,6 +234,41 @@ type Listener struct {
 	//
 	// Support: Core
 	Routes RouteBindingSelector `json:"routes"`
+
+	// BackendRef is a local object reference to route requests to.
```
hbagdi, line 238: Can we include some more contextual information here like:
- Setting this backendRef means traffic is directly passed on without any route selection or IOW relationship of this field with `routes` above
- Protocol selection: how is the protocol between the Gateway and this backend selected, this is mentioned in the issue.

apis/v1alpha1/gateway_types.go
robscott: I agree that we should target this as narrowly as possible. Seeing it here, I think `backendRef` could get confusing since it's so much narrower than backendRef elsewhere. I think we have a few options here:

1. Introduce `listener.forwardTo` or `listener.backend` that would include `serviceName`, `backendRef`, and `port` fields. This is the most consistent with the rest of the API and would enable a great deal of functionality, but at the cost of increased complexity and too many ways to do something. IE how long until we'd want to bubble up filters or other route concepts to listeners?
2. Update `listenener.routes.kind` to support `Service`. This gets rather confusing as it's no longer a direct reference and Services are definitely not Routes.
3. Tweak this approach to define `listener.service` with `name` and `port` fields. 

I think 3 is definitely the simplest and most straightforward option here. One concern I have is that it would exclude referencing multicluster Services at this level. Since this is intended to be a shortcut, that may not be a significant issue, but want to be sure. 

I've also seen some discussion over the past year of updating route selection to support a Service type. I don't think we'd want to do both, so we really need to be sure we don't want listeners to be able to select Services in the same way they can select backends.

/cc @JeremyOT @mark-church @akutz

mark-church: There's two different dimensions that we're discussing here:

- The ability for a Gateway to target multiple L4 resources vs a single L4 resource
- The ability for a Gateway to target just Services vs other CRs as well

There are valid use-cases across all of these, though I definitely do not see the use-cases all as the same priority:

1. Highest-value use-case and the one that started this design decision is the ability to map a single Service directly to a Gateway. I think that for Gateway to have any success at all for L4 this must be done, because otherwise the verbosity of creating an empty UDP/TCPRoute makes it feel too complex to the average user.
2. Another use-case mentioned was to map a single multi-cluster Service (think of a hypothetical CR that represents an aggregation of Services across clusters) to a Gateway.
3. There's another interesting case that is more relevant for service mesh use-cases which is L7 Service virtualization. This is where the Service FQDN is taken to generate an implicit host route for that Service. This allows a Service to automatically be routed at L7 without any additional configuration. This requires the mapping of multiple Services (using `listener.routes.kind=Service`) to a single Gateway where they are differentiated via their "virtualized" Service hostnames.

In all three cases we are actually doing the same thing: creating a Route shortcut. 

For #1 it creates an empty TCPRoute shortcut to a Service. For #2 it creates an empty TCPRoute shortcut to a CR. For #3 it creates an empty HTTPRoute shortcut (using the Service FQDN) to a Service. I originally thought of these as separate things but after thinking about it, this is all closely related and logically consistent.

I agree that we should not open up the possibilities too soon because if we move too fast without considering the consequences then we could start to create some inconsistencies. We should continue to prioritize use-case #1 but it would be great if it would not block #2 or #3 in the future. 

It feels like a `listener.forwardTo` which only includes `serviceName` and `port`, but not `backendRef` would be consistent but not open up too much. `listener.routes.kind=Service` could always be added in the future to support Service virtualization.

robscott: It seems like we have 2 primary concerns here:

1. A meaningful name that doesn't get confused with other similar fields in the existing API
2. Ensuring this does not take on the full scope of `forwardTo` and instead is reserved for use as a simple shortcut

So with that background, I don't think we should reuse either `backendRef` or `forwardTo` terms as they both have different meanings and larger scopes than what we're proposing here. Since this is essentially a route shortcut, what if we used `routeTo`? That would include `serviceName` and `port`, and it would be fairly straightforward to explain that the `routes` and `routeTo` fields were mutually exclusive fields.

apis/v1alpha1/gateway_types.go
robscott: Could we safely rely on either the protocol specified by the Gateway or the Service? If there are edge cases that wouldn't cover, maybe we'd need an optional `protocol` field in `routeTo`.

robscott: Thinking about this more, although we may not _need_ a protocol as part of this, I think it would be much easier to understand than any of the alternatives. If we didn't have protocol here, we'd need to read it from the Service spec. That wouldn't work if there were multiple protocols defined for the same port number so we'd likely need to fall back to the protocol defined on the Gateway listener. Adding a protocol field seems like it would be worth it to avoid the potential confusion with the alternative sources.

robscott:
Follow up from community call. I think there was general support for getting this into v0.3.0. 

There were some concerns that this might grant Gateway admins more access than they currently have. Most notably, this would mean that any Gateway admin has the ability to route traffic to a Service, something that previously required creating an xRoute resource. Although that would be problematic if we were crossing namespace boundaries, it seems fair to say that Gateway admins are highly trusted within the namespace(s) they can manage Gateways for. Since Gateway admins can already arbitrarily target Routes in the same namespace, this feels like a minimal change in the effective capabilities of Gateway admins.

Given the above, I'd propose moving forward with this PR and renaming the field to `routeTo`, initially supporting `serviceName` and `port` fields.

/cc @bowei @hbagdi @danehans

hbagdi:
> Given the above, I'd propose moving forward with this PR and renaming the field to routeTo, initially supporting serviceName and port fields.

+1 to this. Let's update and get this merged in!

stevesloka:
What support type so this get? I added it as "Extended" thinking that it could be adopted by folks but not required. 

// @robscott @hbagdi @mark-church @howardjohn

apis/v1alpha1/gateway_types.go
```diff
@@ -234,6 +234,39 @@ type Listener struct {
...
+	//
+	// Support: Extended
+	//
+	// +optional
+	RouteTo *ListenerRouteTo `json:"routeTo,omitempty"`
+}
+
+// ListenerRouteTo defines how a Listener can forward traffic to
+// a service name & port directly.
+type ListenerRouteTo struct {
```
robscott, line 254: I think it would also be good to include protocol here as a required field, and then conformance/support of this feature could be based on the protocol.

apis/v1alpha1/gateway_types.go
```diff
@@ -234,6 +234,39 @@ type Listener struct {
...
+
+	// RouteTo defines how to route requests directly to a backend without
+	// creating an xRoute resource.
+	//
+	// If the referent cannot be found, the controller should raise
+	// the "ResolvedRefs" condition on the Gateway with the "DegradedListeners"
+	// reason. The gateway status should be updated with a condition
+	// that describes the error more specifically.
+	//
+	// Support: Extended
```
robscott, line 246: I think if we base support off of protocol, we could consider say this has core support for core protocols (as defined slightly lower in this file).

apis/v1alpha1/gateway_types.go
```diff
@@ -234,6 +234,39 @@ type Listener struct {
 	//
 	// Support: Core
 	Routes RouteBindingSelector `json:"routes"`
+
+	// RouteTo defines how to route requests directly to a backend without
```
robscott, line 238: I think we need to define how this relates to the Routes field above. To make this work I think we should:

1. Make the Routes field optional
2. Add webhook validation to ensure that exactly one of Routes or RouteTo must be specified (can be a follow up)
3. Add guidance to the RouteTo field that if both Routes and RouteTo are specified, RouteTo should be given precedence (should not happen with webhook in place, but can't guarantee that)

robscott (commented):
Thanks for the work on this!

apis/v1alpha1/gateway_types.go
jpeach: The selector in `Routes` can already be used to select service resources. Not sure why we need a special case here?

robscott: That's a good question. I thought that would actually be more confusing as users might start to think of Service as a Route when it is really a backend. For example, what if a route selector matched multiple services? For routes we have pretty clear conflict resolution, that doesn't really translate to backends. Although you could say we could just round robin between all matching Services, I'm pretty sure that would result in more confusing behavior than it's worth. Instead, we want to intentionally forward to a specific backend.

What I like about this approach is that it keeps Services clearly defined as a "Backend" and distinct from routes. This functions as a routing shortcut going directly to a backend, but it's still clear that all 3 concepts are still in play.

robscott:
Good discussion around this issue in [Slack](https://kubernetes.slack.com/archives/CR0H13KGA/p1619386375146100). It looks like there's still plenty of hesitation around an approach like this, so I'm going to remove this from the milestone and lower the priority. We may also want to close this PR out too, but I'll leave that final decision for @stevesloka.

bowei:
Freezing this PR pending resolving the discussion.

PR #614, opened 2021-04-19
Making GatewayClass an optional resource

### LINKED ISSUE #567: Support namespace scoped implementations

Currently, someone deploying an implementation must have cluster scoped privileges to access GatewayClass. Often in multitenant clusters, teams are given access only to a set of namespaces without permissions to access cluster scoped resources.
Users in this scenario should be able to deploy an implementation of the API.

For example, they would deploy an nginx proxy controller, a LoadBalancer Service, a Gateway, and some routes.
This is really close to working, but they still need to create some GatewayClass.

## Potential options

1. Cluster admin creates generic GatewayClass with no params, say "in-cluster-proxy". NS admin creates a Gateway referencing it. They still need to configure parameters of the gateway (ie things that would typically exist in parametersRef of Gatewayclass), which they cannot do in GatewayClass. As a result, they need to externalize these configurations to somewhere else (annotations on the deployment/service/gateway, nginx-specific configmap, etc). This isn't great since now its entirely implementation specific how things are configured - but its also not too bad, as parametersRef is all implementation specific anyways. Still requires some cluster admin coordination

2. Same as (1), but maybe they don't even bother with a generic GatewayClass and just put some bogus value there (it is a required field)

3. Do not support deployments with only namespace permissions

robscott:
**What type of PR is this?**
/kind feature
/kind api-change

This has been a popular request for IngressClass as well and would make namespace-scoped implementations of the API possible.

Fixes #567

```release-note
GatewayClass resources are now optional to provide better support for namespace-scoped implementations.
```

/cc @howardjohn @hbagdi @bowei 
/assign @thockin

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -63,8 +63,13 @@ type GatewayList struct {
 // webhook, but there are many cases that will require asynchronous
 // signaling via the GatewayStatus block.
 type GatewaySpec struct {
-	// GatewayClassName used for this Gateway. This is the name of a
-	// GatewayClass resource.
+	// GatewayClassName used for this Gateway. This may refer to a GatewayClass
+	// resource by name.
+	//
+	// Implementations may also support using this field to specify a controller
+	// name directly without a corresponding GatewayClass resource. This is
+	// particularly useful for namespace-scoped implementations that may not
+	// have access to the cluster-scoped GatewayClass resource.
 	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
--- a/config/crd/bases/networking.x-k8s.io_gateways.yaml
+++ b/config/crd/bases/networking.x-k8s.io_gateways.yaml
@@ -86,8 +86,12 @@ spec:
                 maxItems: 16
                 type: array
               gatewayClassName:
-                description: GatewayClassName used for this Gateway. This is the name
-                  of a GatewayClass resource.
+                description: "GatewayClassName used for this Gateway. This may refer
+                  to a GatewayClass resource by name. \n Implementations may also
+                  support using this field to specify a controller name directly without
+                  a corresponding GatewayClass resource. This is particularly useful
+                  for namespace-scoped implementations that may not have access to
+                  the cluster-scoped GatewayClass resource."
                 maxLength: 253
                 minLength: 1
                 type: string


howardjohn (approved):

apis/v1alpha1/gateway_types.go
```diff
@@ -63,8 +63,13 @@ type GatewayList struct {
...
 type GatewaySpec struct {
-	// GatewayClassName used for this Gateway. This is the name of a
-	// GatewayClass resource.
+	// GatewayClassName used for this Gateway. This may refer to a GatewayClass
+	// resource by name.
+	//
+	// Implementations may also support using this field to specify a controller
+	// name directly without a corresponding GatewayClass resource. This is
+	// particularly useful for namespace-scoped implementations that may not
+	// have access to the cluster-scoped GatewayClass resource.
```
jpeach, line 72: What is a controller meant to do if it is referred to by name? For example, a cluster has controller A with gateway classes A1 and A2 configured. When a gateway is specified with a classname of A, there's no configuration that matches that, so what is the right behavior?

robscott: Yeah that's a good question. From my perspective, referring to a controller name directly would result in using the controller's default behavior. 

To take this a step further, I'd thought it might be worth supporting not just controller name, but controller name + suffix. Implementations could choose to ship with a set of preloaded "virtual" GatewayClasses that could be referenced by that suffix. For example, let's say that controller name is `acme.io/gateway` and that controller ships with predefined/virtual `xlb` and `ilb` concepts. A new Gateway could be created with `acme.io/gateway/xlb` as `GatewayClassName` in that case.

jpeach: I think that the critical part of the problem is

> namespace-scoped implementations that may not have access to the cluster-scoped GatewayClass resource.

Maybe some more context around what kind of implementation you are thinking of here would make the change clearer.

robscott: I think the [upstream Ingress issue](https://github.com/kubernetes/kubernetes/issues/99824) provides some compelling examples here. In that case, someone is trying to deploy either Traefik or ingress-nginx into a namespace but both fail because they don't have access to IngressClass. I think it's relatively common for users to want to be able to deploy an in-cluster proxy like this within a single namespace. In that case, they're not really concerned about custom behavior, just connecting a Gateway to a controller. 

With that said, I think there are some interesting extensions of this model like the virtual/predefined GatewayClass concept that could enable further capabilities when scoped within a single namespace. 

Although those 2 examples are from the upstream issue, I think the concept is broadly applicable. I'm hoping that whatever we try here could also make sense for Ingress. Although we could try to make namespace-scoped *Class resources here, I think that would result in significantly more complexity than this approach.

hbagdi:
Can we also update documentation of GatewayClass resource?

apis/v1alpha1/gateway_types.go
jpeach: I guess there's also a question of whether you can distinguish between a missing resource and a controller name/label?

hbagdi: > I guess there's also a question of whether you can distinguish between a missing resource and a controller name/label?

I think overloading two different semantics here will be problematic.
If we have two fields - `gatewayClassName` and `controller` and only one of the two are required, that is a more clear API for the users and implementers.

robscott: Yeah that's a good point. That would be significantly clearer. Since this same problem applies to Ingress, I think I was getting confused by the terminology in that API where a class defined by an annotation does not refer to an IngressClass resource. Of course that was never really a formal part of the API and developed a bit more organically. If we went forward with `controller` here, we could add validation to the webhook to ensure only one field was set at a time. Or if these really are mutually exclusive fields, maybe there's a more generic term we could use that would work for both use cases?

hbagdi: > we could add validation to the webhook to ensure only one field was set at a time

There is a way to do that validation using CRD itself but kubebuilder doesn't support it. -_-
But yes, that is something I was thinking too - two mutual exclusive fields.

> Or if these really are mutually exclusive fields, maybe there's a more generic term we could use that would work for both use cases?

Using a single field for two (or more) constants or variable values would be okay but as @jpeach pointed out, the semantics here are different.
A solution that comes to mind is using a single field for value and then a union discriminator field to figure out the meaning of the value. That's more complexity for no gain, unless we plan to have a third field, which I _feel_ is unlikely.

robscott:
I think this requires a bit more thought. I'm working on a doc that goes through some of our options here. For now I'm going to close this, but may reopen later depending on the outcome of a broader discussion.

/close

lukasrieger-dev:
Hi @robscott,
what's the current state of the broader discussion?
I'd still be interested in a solution to this issue.

robscott:
Hey @lukasrieger-dev, thanks for checking in on this. Unfortunately this lost steam. There were some concerns about the ambiguity of any possible optional GatewayClass solutions. Unfortunately I'm not sure how much time I'll have to push this forward in the next few months, but if anyone else has time to push this discussion forward, I'd be happy to provide feedback.

PR #616, opened 2021-04-20
Adding a controller field to GatewayReference in RouteGatewayStatus

### LINKED ISSUE #591: Handling Route status when multiple controllers are involved
Consider a route that selects many different gateways from many different controllers. This may cause issues in writing the route status.

At first glance, you could have each controller only update info about gateways they own. However, this makes it hard to handle removal of Gateways. For example, if I see existing status of `gateways: [a, b, c, d]`, and now I see the route matches gateways `a` and `b` owned by my controller, I cannot tell if `c` and `d` are gateways controlled by another controller, or if they are gateways that _used_ to be controlled by my controller but have since been removed.

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This adds controller to GatewayReference in RouteGatewayStatus. This will be especially helpful for cleaning up old entries in status.

Fixes #591 

```release-note
Route status now includes controller name for each Gateway.
```

/cc @hbagdi @howardjohn @stevesloka


stevesloka (approved):
lgtm

I think it would be up to the implementors to make this field unique by prefixing with the `controller` name on the GatewayClass with something else unique to their implementation.

config/crd/bases/networking.x-k8s.io_httproutes.yaml
```diff
@@ -71,6 +71,11 @@ spec:
                       description: GatewayReference identifies a Gateway in a specified
                         namespace.
                       properties:
+                        controller:
+                          description: Controller is the name of the controller implementing
+                            the Gateway.
+                          maxLength: 253
+                          type: string
```
hbagdi, line 78: Is this an intended side-effect?
This would require Route owners to also define the controller and not just the Gateway. It seems that we are breaking our abstraction layers here.

hbagdi (commented):
Two concerns:
- GWC is not immutable so someone could change the name of the controller. 🤷🏾 There is only so much we can do using CRDs.
- Sometimes multiple controllers of the same implementations could be running. I'm assuming that in that case, `controller` field will be different for each instance of the same implementation. @youngnick Any thoughts?

config/crd/bases/networking.x-k8s.io_httproutes.yaml
robscott: Good catch, I'd missed that addition here. I'd already moved on in my head to where we had a selector here and not a list. This is an optional field, but even still, it doesn't make sense here. I'll pull it out for this use case until we change the handshake entirely in v1alpha2.

robscott: Created a new `GatewayStatusReference` to differentiate these two references.

robscott:
Thanks for the great feedback on this @hbagdi and @stevesloka! I think I've resolved everything now - PTAL.

apis/v1alpha1/shared_types.go
```diff
@@ -194,6 +194,37 @@ type RouteGatewayStatus struct {
...
+	// GatewayClass.
+	//
+	// Example: "acme.io/gateway-controller".
+	//
+	// The format of this field is DOMAIN "/" PATH, where DOMAIN and PATH are
+	// valid Kubernetes names
+	// (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names).
+	//
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
```
hbagdi, line 224: Should we make this required?
Conformant implementations will set this field and clean up after themselves. But then this API allows for a mistake whereby an implementation might not set the controller field and then the user has a sub-optimal experience.

robscott: I want to make this required in v1alpha2, but I don't think we can require it in v0.3.0 or else it would be a breaking change. I'll make a tracking issue for this if/when this PR gets in.

hbagdi: Ah, I missed that. Makes sense.

apis/v1alpha1/shared_types.go
```diff
@@ -194,6 +194,37 @@ type RouteGatewayStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
+// GatewayStatusReference identifies a Gateway in a specified namespace. This
+// reference also includes a controller name to simplify cleaning up status
+// entries.
+type GatewayStatusReference struct {
```
hbagdi, line 200: From Go API user's perspective, I can confuse this as a reference of something inside GatewayStatus.
How about `RouteStatusGatewayReference`?

This is a little bit of a style and language thing so hard to concrete argue one way or the other.

robscott: Yeah good call, that is better. Updated to use that name.

robscott:
@hbagdi Thanks for the great feedback on this one, I think everything should be resolved now. PTAL.

PR #617, opened 2021-04-20
implement an admission webhook server

### LINKED ISSUE #349: HTTPRequestMirrorFilter needs webhook validation
We need to be able to ensure that only one filter is set at a time in this type.

hbagdi:
Co-authored-by: Christopher M. Luciano <cmluciano@isovalent.com>

**What type of PR is this?**

/kind feature

This PR introduces a new admission webhook server that will be used for advanced validation functionality.

Fixes #349

```release-note
NONE
```


hbagdi:
The scope of this PR is to get this merged into mainline and then work on documentation, otherwise the scope of this work keeps on increasing and it gets very hard to actually land in the mainline. I had to invest significant amount of time reworking major area over #506 to get this working again.


How can a reviewer test this out?

Bad HTTPRoute: https://gist.github.com/hbagdi/b7d7cfde79610133846edf2644530e6c, create a local file out of the gist.

```
make crd
kubectl apply -f admission_webhook.yaml
kubectl apply -f certificate_config.yaml 
# you might have to restart the admission-server pod sometimes, the ordering is not there yet
kubectl apply -f path/to/bad-httproute.yaml
# this should show the right error, if you get an error for TLS, make sure the jobs in gateway-api namespace are completed and restart the server pod
```

deploy/admission_webhook.yaml
```diff
@@ -0,0 +1,95 @@
...
+  template:
+    metadata:
+      name: gateway-api-admission-server
+      labels:
+        name: gateway-api-admission-server
+    spec:
+      containers:
+        - name: webhook
+          # TODO(hbagdi): Swap image name to the k8s official image
+          image: hbagdi/gateway-api-admission-webhook:latest
```
hbagdi, line 68: I'm working on getting us a new docker repo to host this.

robscott: Sorry, missed that this was before the Gateway API rename, thanks for following up with https://github.com/kubernetes/k8s.io/pull/1954!

deploy/admission_webhook.yaml
```diff
@@ -0,0 +1,95 @@
...
+          ports:
+            - containerPort: 8443
+              name: webhook
+          resources:
+            limits:
+              memory: 50Mi
+              cpu: 300m
+            requests:
+              memory: 50Mi
+              cpu: 300m
```
robscott, line 85: Have you seen any initial utilization numbers with this container in action yet?

pkg/admission/server.go
```diff
@@ -0,0 +1,168 @@
...
+	switch request.Resource {
+	case httpRouteGVR:
+		var hRoute v1alpha1.HTTPRoute
+		deserializer := codecs.UniversalDeserializer()
+		_, _, err := deserializer.Decode(request.Object.Raw, nil, &hRoute)
+		if err != nil {
+			return nil, err
+		}
+		// The admission hook is only configured for create & update, so we can
+		// nolint:exhaustive
```
robscott, line 147: Which line is this needed for?

pkg/admission/server.go
```diff
@@ -0,0 +1,168 @@
...
+
+	switch request.Resource {
+	case httpRouteGVR:
+		var hRoute v1alpha1.HTTPRoute
+		deserializer := codecs.UniversalDeserializer()
+		_, _, err := deserializer.Decode(request.Object.Raw, nil, &hRoute)
+		if err != nil {
+			return nil, err
+		}
+		// The admission hook is only configured for create & update, so we can
```
robscott, line 146: Nit: Sentence seems like it gets cut off early here

robscott (commented):
This is very cool, thanks for all the work on this @hbagdi and @cmluciano! Excited to see this getting so close.

robscott:
Tested this out in a new cluster and it worked flawlessly, nicely done!

deploy/admission_webhook.yaml
hbagdi: Is it cool to merge this in before we get the image repo set up?
It feels like this is going to be a chicken and egg problem which will require some iterations.

deploy/admission_webhook.yaml
hbagdi: Nope. I didn't even pay much attention to these for the PR. 
Since it is always better to start with some limits, these seem okay.
Do you think these are too high/low?

pkg/admission/server.go
hbagdi: Removed it. It was for a `switch` statement which I removed to simplify the code.

pkg/admission/server.go
hbagdi: Removed this too, part of the same switch statement that I removed.

hbagdi:
@cmluciano @robscott I simplified the code to make the behavior across Create and Update operation exactly same. I don't think we have a case (yet) that requires validation to be different for update and create. Does that feel right?

@bowei @danehans Since this is an initial PR for some code that is going to grow in future, can you also review this?

deploy/admission_webhook.yaml
robscott: I think this is reasonable but CPU may be a bit too high here, maybe 100 or 200 would be sufficient?

hbagdi: Fair, switched to 100.

robscott:
LGTM, adding a hold until we can get an official image published.

/hold

deploy/admission_webhook.yaml
hbagdi: @robscott What if I update the `image` to `TODO` and then work on another PR to get [cloudbuild](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md) setup for admission webhook?

robscott: I think that's reasonable, adding to agenda for community meeting

hbagdi:
@robscott I removed the docker image reference to `TODO`.
I considered adding in a comment or some documentation around 'this is still a work in progress' but couldn't find a suitable place for it.
Since we don't document the webhook server anywhere, I don't expect any users to run into it. Some contributors to this project might run into it and that would be a good place to have more conversations and seek contribution.

I'm open to adding more here if necessary.
If we can get another review and get this merged in, that would be helpful.

robscott:
Thanks @hbagdi! I think this is good to go no so we can unblock the follow up PRs. Removing the hold and will leave the final LGTM for @danehans.

/hold cancel

deploy/admission_webhook.yaml
```diff
@@ -0,0 +1,95 @@
...
+      name: gateway-api-admission-server
+  template:
+    metadata:
+      name: gateway-api-admission-server
+      labels:
+        name: gateway-api-admission-server
+    spec:
+      containers:
+        - name: webhook
+          # TODO(hbagdi): Swap image name to the k8s official image
```
danehans, line 67: @hbagdi I think it would be helpful include an xref the issue/pr in this TODO comment. Is it https://github.com/kubernetes-sigs/gateway-api/pull/621?

pkg/admission/server.go
```diff
@@ -0,0 +1,167 @@
...
+	_, err = w.Write(data)
+	if err != nil {
+		klog.Errorf("failed to write HTTP response: %v\n", err)
+		return
+	}
+	return
+}
+
+func handleValidation(request admission.AdmissionRequest) (
+	*admission.AdmissionResponse, error) {
```
danehans, line 123: Why a new line for the returned types?

pkg/admission/server_test.go
```diff
@@ -0,0 +1,309 @@
...
+	assert.Nil(err)
+	handler.ServeHTTP(res, req)
+	assert.Equal(http.StatusMethodNotAllowed, res.Code)
+	assert.Equal("invalid method GET, only POST requests are allowed\n",
+		res.Body.String())
+}
+
+func TestServeHTTPSubmissions(t *testing.T) {
+	for _, apiVersion := range []string{
+		"admission.k8s.io/v1beta1",
```
cmluciano, line 63: Do we still need to accept this if we use only setup [webhook v1 validation](https://github.com/kubernetes-sigs/gateway-api/pull/617/files#diff-267c12189c93f1a7a1a0163c9f10522d8f2fe0792223c211200f09f941d5c3ddR22-R23)

cmluciano (commented):
One comment about testing v1beta1 submissions but otherwise lgtm. Thank you for carrying this forward :D

pkg/admission/server_test.go
hbagdi: Updated to v1 only.

PR #620, opened 2021-04-21
Adding conflict resolution guidance for equivalent rules within a route

### LINKED ISSUE #613: Clarify route conflicts in the same resource
I ran into a conflict today thinking about implementation and thought it would be best to gather everyone else's thoughts on it. 

There is some logic written in the API which states in the event of conflicts occur, how to handle (e.g. The oldest Route based on creation timestamp, or the Route appearing first in alphabetical order). 

What if I have an `HTTPRoute` which has the same `matches` set, but with different `forwardTo`? We can't apply any timestamp logic since there isn't a resource to compare against.

What should a `GET` for `local.projectcontour.io/` respond with? What conditions should I set? 

**Example:**
```
kind: HTTPRoute
metadata:
  name: http-filter-1
  namespace: projectcontour-roots 
spec:
  hostnames:
    - local.projectcontour.io
  rules:
    - matches:  
      - path:
          type: Prefix
          value: /  
      forwardTo:
      - serviceName: insecure   #<--- serviceName different
        port: 80
    - matches:
      - path:
          type: Prefix
          value: / 
      forwardTo:
      - serviceName: rootapp   #<--- serviceName different
        port: 80
```

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This adds conflict resolution guidance for equivalent rules within a route.

Fixes #613 

```release-note
Adds conflict resolution guidance for equivalent rules within a route
```

/cc @stevesloka @youngnick @hbagdi

### DIFF
--- a/apis/v1alpha1/httproute_types.go
+++ b/apis/v1alpha1/httproute_types.go
@@ -179,12 +179,17 @@ type HTTPRouteRule struct {
 	//
 	// * The longest matching hostname.
 	// * The longest matching path.
-	// * The largest number of header matches
+	// * The largest number of header matches.
+	//
+	// If ties still exist across multiple Routes:
 	// * The oldest Route based on creation timestamp. For example, a Route with
 	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
 	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
 	// * The Route appearing first in alphabetical order (namespace/name) for
 	//   example, foo/bar is given precedence over foo/baz.
+	//
+	// If ties still exist within the Route that has been given precedence:
+	// * The first matching rule meeting the above criteria.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/httproute_types.go
```diff
@@ -179,13 +179,18 @@ type HTTPRouteRule struct {
...
+	//
+	// If ties still exist across multiple Routes:
 	// * The oldest Route based on creation timestamp. For example, a Route with
 	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
 	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
 	// * The Route appearing first in alphabetical order (namespace/name) for
 	//   example, foo/bar is given precedence over foo/baz.
 	//
+	// If ties still exist within the Route that has been given precedence:
+	// * The first matching rule meeting the above criteria.
```
hbagdi, line 192: I just realized that we are documenting the whole thing in the wrong place. Most of these rules will apply to non-HTTP cases as well.
Changing this shouldn't be part of this PR so no action required here.

hbagdi:
/lgtm


@youngnick had some thoughts here so let's wait for his review too.
/hold

apis/v1alpha1/httproute_types.go
robscott: Yeah, I'd realized that as well but we have such minimal matching semantics on the other routes applying this guidance there would turn it into a much larger change. I'll create a follow up issue or PR to take this on.

youngnick:
/lgtm

Yeah, I think the first-matching is slightly better than last-matching. Thanks for making this update @robscott!

PR #621, opened 2021-04-21
automate webhook server image build and publishing

hbagdi:
**What type of PR is this?**

This PR adds plumbing code and cloudbuild configuration to automatically build images of admission webhook server.
The other part of this is https://github.com/kubernetes/test-infra/pull/21881.
The prow PR  can be merged only after this PR is merged in.

None

```release-note
NONE
```


hbagdi:
Please only review the last commit. This PR will be rebased on top of mainline branch after #617 is merged in.

hbagdi:
I'm not really sure if the `cloudbuild.yaml` is correct or not. This is my first time writing it and I don't have any experience with it.

Regarding the rest of the changes:
- If reviewer can pull down this patch and run `make build` and check if all bits work as they should, that would help. Additionally, `export TAG=new-fancy-tag && make build` would help to ensure that the version gets propagated to the binary in the container.
- I looked at other k8s-sigs projects and almost everyone has an elaborate Makefile that takes care of multiple CPU architectures and sub-systems. I considered that approach but figured that starting small and failing early was the better path forward.
- I went ahead with the `make` approach instead of a native cloud-build approach because that gives us more flexibility. If there are more artifacts that need to be built (more components or CPU architectures) then we have all the control here. No external configuration will need to be updated.

cloudbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+    entrypoint: make
+    env:
+    - DOCKER_CLI_EXPERIMENTAL=enabled
+    - TAG=$_GIT_TAG
+    args:
+    - release
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
```
robscott, line 20: Do we need this?

Dockerfile
```diff
@@ -17,10 +17,13 @@ RUN mkdir -p /go/src/sig.k8s.io/gateway-api
...
-RUN cd cmd/admission/ && CGO_ENABLED=0 GOOS=linux go build -a -o gateway-api-webhook && chmod +x gateway-api-webhook
+ARG TAG
+ARG COMMIT
+RUN CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -a -o gateway-api-webhook \
+      -ldflags "-s -w -X main.VERSION=$TAG -X main.COMMIT=$COMMIT" ./cmd/admission
 
 FROM scratch
-COPY --from=build-env /go/src/sig.k8s.io/gateway-api/cmd/admission/gateway-api-webhook .
+COPY --from=build-env /go/src/sig.k8s.io/gateway-api/gateway-api-webhook .
 COPY --from=build-env /etc/passwd /etc/passwd
```
robscott, line 27: Do you know if/why we need this?

robscott (commented):
This looks great - I was able to test out `make build` locally and it worked great for me. Just had a few follow up comments/questions:

cloudbuild.yaml
hbagdi: I'm not sure but it seems so. it is present in other repos in the k8s-sigs org in Github.

cloudbuild.yaml
```diff
@@ -0,0 +1,23 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4'
```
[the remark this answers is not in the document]
hbagdi: I updated to use the newer image: `gcr.io/k8s-testimages/gcb-docker-gcloud:v20210331-c732583`.

Dockerfile
hbagdi: I think this is for the user `webhook` that is created in the build stage. This ensures that that the webhook user is present in the final image as well (to run the image in as non-root).

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN mkdir -p /go/src/sig.k8s.io/gateway-api
+WORKDIR /go/src/sig.k8s.io/gateway-api
+COPY  . .
+RUN useradd -u 10001 webhook
+ARG TAG
+ARG COMMIT
+RUN CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -a -o gateway-api-webhook \
+      -ldflags "-s -w -X main.VERSION=$TAG -X main.COMMIT=$COMMIT" ./cmd/admission
+
+FROM scratch
```
jpeach, line 25: drive by comment: maybe distroless is a better choice? Or whatever base image core Kubernetes is standardizing on?

hbagdi: Thanks for the nudge. I switched to `gcr.io/distroless/static:nonroot` instead which simplified the Dockerfile a bit too. PTAL.

Dockerfile
hbagdi: This has now been removed.

Dockerfile
hbagdi: I've borrowed the base image from Cluster API:https://github.com/kubernetes-sigs/cluster-api/blob/master/Dockerfile#L56

hbagdi:
@robscott @jpeach I've rebased this and is ready for review.

jpeach:
Found the Kubernetes docker image issue for reference https://github.com/kubernetes/kubernetes/issues/70249

robscott:
Thanks!

/lgtm

jpeach (approved):
Looks pretty reasonable to me. Suggest filing a separate issue to track the need to come back and enable multi-arch builds.

PR #626, opened 2021-04-23
Clarifying how conflicting matches should be resolved in non-HTTP Routes

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This is a follow up from https://github.com/kubernetes-sigs/gateway-api/pull/620. It takes the conflict resolution guidance we added to HTTPRoute and applies it to the other route types.

```release-note
Clarifies how conflicting matches should be resolved in non-HTTP routes
```

/cc @hbagdi @youngnick

### DIFF
--- a/apis/v1alpha1/tcproute_types.go
+++ b/apis/v1alpha1/tcproute_types.go
@@ -60,11 +60,32 @@ type TCPRouteStatus struct {
 
 // TCPRouteRule is the configuration for a given rule.
 type TCPRouteRule struct {
-	// Matches define conditions used for matching the rule against
-	// incoming TCP connections. Each match is independent, i.e. this
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TCP
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TCP
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified (i.e. empty),
+	// this Rule will match all requests for the associated Listener.
+	//
+	// Each client request MUST map to a maximum of one route rule. If a request
+	// matches multiple rules, matching precedence MUST be determined in order
+	// of the following criteria, continuing on ties:
+	//
+	// * The most specific match specified by ExtensionRef. Each implementation
+	//   that supports ExtensionRef may have different ways of determining the
+	//   specificity of the referenced extension.
+	//
+	// If ties still exist across multiple Routes, matching precedence MUST be
+	// determined in order of the following criteria, continuing on ties:
+	//
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
+	// * The Route appearing first in alphabetical order by
+	//   "<namespace>/<name>". For example, foo/bar is given precedence over
+	//   foo/baz.
+	//
+	// If ties still exist within the Route that has been given precedence,
+	// matching precedence MUST be granted to the first matching rule meeting
+	// the above criteria.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
--- a/apis/v1alpha1/tlsroute_types.go
+++ b/apis/v1alpha1/tlsroute_types.go
@@ -65,11 +65,35 @@ type TLSRouteStatus struct {
 
 // TLSRouteRule is the configuration for a given rule.
 type TLSRouteRule struct {
-	// Matches define conditions used for matching the rule against an
-	// incoming TLS handshake. Each match is independent, i.e. this
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TLS
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TLS
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified (i.e. empty),
+	// this Rule will match all requests for the associated Listener.
+	//
+	// Each client request MUST map to a maximum of one route rule. If a request
+	// matches multiple rules, matching precedence MUST be determined in order
+	// of the following criteria, continuing on ties:
+	//
+	// * The longest matching SNI.
+	// * The longest matching precise SNI (without a wildcard). This means that
+	//   "b.example.com" should be given precedence over "*.example.com".
+	// * The most specific match specified by ExtensionRef. Each implementation
+	//   that supports ExtensionRef may have different ways of determining the
+	//   specificity of the referenced extension.
+	//
+	// If ties still exist across multiple Routes, matching precedence MUST be
+	// determined in order of the following criteria, continuing on ties:
+	//
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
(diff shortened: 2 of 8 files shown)


apis/v1alpha1/tlsroute_types.go
```diff
@@ -65,11 +65,28 @@ type TLSRouteStatus struct {
...
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TLS
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TLS
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified, all requests
+	// from the associated gateway listener will match.
+	//
+	// A client request may match multiple route rules. Matching precedence MUST
+	// be determined in order of the following criteria, continuing on ties:
```
hbagdi, line 74: > A client request may match multiple route rules.

This can be interpreted as "it is okay for a request to match multiple route rules" but what we really want to say is that when a client matches multiple rules, use the following precedence order to break ties.

One way to put that:
Every client request should map to one route rule. In case a request matches multiple rules, the following precedence order is used to break ties:

robscott: Good catch, this was confusing. I've merged this idea with the existing wording and copied it across all route types - hopefully the updated version is a bit easier to understand.

apis/v1alpha1/tcproute_types.go
```diff
@@ -60,11 +60,28 @@ type TCPRouteStatus struct {
...
-	// incoming TCP connections. Each match is independent, i.e. this
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TCP
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TCP
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified, all requests
+	// from the associated gateway listener will match.
+	//
+	// Each client request should map to a maximum of one route rule. If a
```
hbagdi, line 68: nit: `s/should/must`?

youngnick: Strong +1 from me here, and on the other places where this phrase is used.

youngnick:
I think changing "should match a maximum of one route rule" to "MUST match a maximum of one route rule" is vital. But apart from that, this looks great.

apis/v1alpha1/tcproute_types.go
```diff
@@ -60,11 +60,28 @@ type TCPRouteStatus struct {
...
 // TCPRouteRule is the configuration for a given rule.
 type TCPRouteRule struct {
-	// Matches define conditions used for matching the rule against
-	// incoming TCP connections. Each match is independent, i.e. this
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TCP
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TCP
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified, all requests
```
bowei, line 65: For some reason I'm finding the language a little harder to understand, maybe:

```
If unspecified (i.e. empty), this Rule will match all requests for the associated Listener.
```
```

File: apis/v1alpha1/tcproute_types.go [hunk 4]
```diff
@@ -60,11 +60,28 @@ type TCPRouteStatus struct {
...
+	//
+	// Each client request should map to a maximum of one route rule. If a
+	// request matches multiple rules, matching precedence MUST be determined in
+	// order of the following criteria, continuing on ties:
+	//
+	// * The most specific match specified by ExtensionRef. Each implementation
+	//   that supports ExtensionRef may have different ways of determining the
+	//   specificity of the referenced extension.
+	//
+	// If ties still exist across multiple Routes:
```
bowei, line 76:
```
// If ties still exist ..., break ties in the following order
```

apis/v1alpha1/tcproute_types.go
```diff
@@ -60,11 +60,28 @@ type TCPRouteStatus struct {
...
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2020-09-08 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2020-09-08 01:02:04".
+	// * The Route appearing first in alphabetical order (namespace/name) for
```
bowei, line 80: `The Route appearing first in alphabetical order by "<namespace>/<name>" for`


(more explicit)

apis/v1alpha1/tlsroute_types.go
```diff
@@ -65,11 +65,29 @@ type TLSRouteStatus struct {
...
 // TLSRouteRule is the configuration for a given rule.
 type TLSRouteRule struct {
-	// Matches define conditions used for matching the rule against an
-	// incoming TLS handshake. Each match is independent, i.e. this
-	// rule will be matched if **any** one of the matches is satisfied.
-	// If unspecified, all requests from the associated gateway TLS
-	// listener will match.
+	// Matches define conditions used for matching the rule against incoming TLS
+	// connections. Each match is independent, i.e. this rule will be matched if
+	// **any** one of the matches is satisfied. If unspecified, all requests
```
bowei, line 70: See my comment above:

If unspecified (i.e. empty), this Rule will match all requests for the associated Listener.

apis/v1alpha1/tlsroute_types.go
```diff
@@ -65,11 +65,29 @@ type TLSRouteStatus struct {
...
+	// from the associated gateway listener will match.
+	//
+	// Each client request should map to a maximum of one route rule. If a
```
bowei, line 73: must

apis/v1alpha1/tlsroute_types.go
```diff
@@ -65,11 +65,29 @@ type TLSRouteStatus struct {
...
+	// request matches multiple rules, matching precedence MUST be determined in
+	// order of the following criteria, continuing on ties:
+	//
+	// * The longest matching SNI.
```
bowei, line 77: Do we need to comment on how to compare wildcard strings?

bowei:
Nits aside

LGTM

apis/v1alpha1/tlsroute_types.go
robscott: Good catch. I think "longest matching" would still be what we want even if wildcards were involved. We were missing guidance for what should be done if "*.example.com" and "b.example.com" both match - I added a new line item below this to clarify that precedence should be given to the longest precise match in the event of a tie.

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tlsroute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tcproute_types.go
robscott: Good call, updated with something similar.

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha1/tcproute_types.go
robscott: Updated, thanks!

robscott:
Thanks for the great feedback on this @bowei, @hbagdi, and @youngnick! I think I've resolved everything, PTAL.

PR #630, opened 2021-04-23
Adding ValidReferences condition to Gateway

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This condition replaces the NoSuchGatewayClass reason. That reason was always confusing because it was unclear which controller(s) would be responsible for setting it. Instead, this switches to an "Unknown" default state and expands the scope to include all references from Gateway.

Although I considered using "InvalidReferences" as the condition name here, I think it's preferable to avoid double negatives here and have this condition be present at all times.

```release-note
A new ValidReferences condition has been added to Gateway
The NoSuchGatewayClass reason has been removed from Gateway
```

/cc @jpeach @hbagdi

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -43,7 +43,7 @@ type Gateway struct {
 
 	// Status defines the current state of Gateway.
 	//
-	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"},{type: "ValidReferences", status: "Unknown", reason:"NotValidated", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
 	Status GatewayStatus `json:"status,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/gateway_types.go
```diff
@@ -649,6 +641,32 @@ const (
 	GatewayReasonAddressNotAssigned GatewayConditionReason = "AddressNotAssigned"
 )
 
+const (
+	// This condition is true when resources referenced by this Gateway exist.
+	// This covers any direct references to resources in the Gateway, including
+	// GatewayClassName and CertificateRef. This does not include selector
+	// references. This does not reflect the validity of the referenced
+	// resources, simply if the referenced resources exist.
```
hbagdi, line 649: > simply if the referenced resources exist

Is there a word missing here? Maybe `simply validates if the referenced resources exist`?

hbagdi:
> Although I considered using "InvalidReferences" as the condition name here, I think it's preferable to avoid double negatives here and have this condition be present at all times.

+1

PR looks good other than one tiny comment.
My knowledge and judgement of status conditions is not the strongest so I suggest another reviewer review this as well.

robscott:
Thanks for the review on this @hbagdi! I'd just recently chatted with @jpeach about this as well and I think we're all on board with deprecating `NoSuchGatewayClass` but not as sure about this new condition. I pulled the deprecation part out into a new PR and will close this in favor of it: https://github.com/kubernetes-sigs/gateway-api/pull/635.

/close

robscott:
And one last bit of clarification - I don't mind revisiting adding a new condition here, but for now I mainly just wanted to start the process of deprecating `NoSuchGatewayClass` due to the confusion around implementing it.

PR #631, opened 2021-04-23
Adding QueryParam matching to HTTPRoute

### LINKED ISSUE #259: Support Query Param Matching
It may be helpful to extend matching to support query parameters. Although not as common as header matching, this could be useful in some cases.

**Why is this needed**:
From what I can tell, some proxy implementations support query param matching but definitely not all. If we support this it will likely have to be an extended feature.

robscott:
**What type of PR is this?**
/kind feature
/kind api-change

This adds query param matching to HTTPRoute

Fixes #259 

```release-note
HTTPRoute now supports matching query parameters
```

/cc @bowei @danehans @hbagdi @jpeach @youngnick


apis/v1alpha1/httproute_types.go
```diff
@@ -326,6 +345,37 @@ type HTTPHeaderMatch struct {
...
+// HTTPQueryParamMatch describes how to select a HTTP route by matching HTTP
+// query parameters.
+type HTTPQueryParamMatch struct {
+	// Type specifies how to match against the value of the query parameter.
+	//
+	// Support: Extended (Exact, Presence)
+	//
+	// Support: Custom (RegularExpression, ImplementationSpecific)
+	//
+	// Since RegularExpression PathType has custom conformance, implementations
```
hbagdi, line 357: > RegularExpression PathType

copy-pasta error with `PathType`

apis/v1alpha1/httproute_types.go
```diff
@@ -270,6 +270,25 @@ const (
...
+//
+// +kubebuilder:validation:Enum=Exact;Presence;RegularExpression;ImplementationSpecific
+type QueryParamMatchType string
+
+// QueryParamMatchType constants.
+const (
+	QueryParamMatchExact                  HeaderMatchType = "Exact"
+	QueryParamMatchPresence               HeaderMatchType = "Presence"
+	QueryParamMatchRegularExpression      HeaderMatchType = "RegularExpression"
+	QueryParamMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
```
hbagdi, line 289: The types of all these should be `QueryParamMatchType`.

apis/v1alpha1/httproute_types.go
```diff
@@ -270,6 +270,25 @@ const (
...
+// * "RegularExpression"
+// * "ImplementationSpecific"
+//
+// +kubebuilder:validation:Enum=Exact;Presence;RegularExpression;ImplementationSpecific
+type QueryParamMatchType string
+
+// QueryParamMatchType constants.
+const (
+	QueryParamMatchExact                  HeaderMatchType = "Exact"
+	QueryParamMatchPresence               HeaderMatchType = "Presence"
```
hbagdi, line 287: I'm not very familiar with query parameter matching myself, is `Presence` a commonly supported or asked one?

apis/v1alpha1/httproute_types.go
```diff
@@ -326,6 +345,37 @@ type HTTPHeaderMatch struct {
...
+	// Support: Extended (Exact, Presence)
+	//
+	// Support: Custom (RegularExpression, ImplementationSpecific)
+	//
+	// Since RegularExpression PathType has custom conformance, implementations
+	// can support POSIX, PCRE or any other dialects of regular expressions.
+	// Please read the implementation's documentation to determine the supported
+	// dialect.
+	//
+	// HTTP query parameter matching MUST be case-sensitive.
```
hbagdi, line 362: Both key and value? If yes, let's be precise.

hbagdi (commented):
We should get more eyes on this before this lands in.
Overall, this looks good and is in the right direction. Thanks for your work on this!

apis/v1alpha1/httproute_types.go
howardjohn: https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--drain_listeners is one example in the wild where a `Presence` could be useful?

It does seem inconsistent with Header at least.

howardjohn: To be more clear, I mean there should probably be one common way to describe this in the API. So add Presence to Header or do something else, but they should be consistent.

I think you could express it with regex of "" possibly.

For a prior art, Istio has ~everything as the same StringMatch type: https://istio.io/latest/docs/reference/config/networking/virtual-service/#StringMatch

hbagdi: > So add Presence to Header or do something else, but they should be consistent.

But then Presence as PathType makes little sense.
I think it is okay to have different match types for different metadata pieces like query string, header, path, etc because they have different properties and uses.

howardjohn: I agree they don't need to be the exact same type, but they should be consistent where it makes sense? Presence in path definitely would be odd. I see headers and query params as pretty similar so I would envision them having the same options

hbagdi: > they should be consistent where it makes sense?

We need to dig deeper into what consistency means here.

I agree that they should be consistent for the end-user - kubectl user, meaning having a RegularExpression match on path/header/query param should do the same thing and not surprise the user.
For the implementers, I think having separate Go types and different subsets is okay and good because that makes this API (our go package) easy to use.

robscott: Yeah that makes sense, I'm fine with dropping Presence from this PR to keep it smaller in scope. We can add it later to both if there's interest.

howardjohn: I was only talking about the yaml API surface. Whether we use a single or different go type  I don't have a strong opinion, dedicated types seems reasonable.

apis/v1alpha1/httproute_types.go
robscott: 🤦  thanks for catching that!

apis/v1alpha1/httproute_types.go
robscott: Agreed, clarified that both key and value should be case-sensitive. This is based on the following guidance from the RFC (emphasis mine):

The scheme and host are case-insensitive and normally provided in lowercase; **all other components are compared in a case-sensitive manner**.

https://tools.ietf.org/html/rfc7230#page-19

apis/v1alpha1/httproute_types.go
robscott: Yep agree that we should still use separate Go types, kept those in place while dropping "Presence"

apis/v1alpha1/httproute_types.go
hbagdi: Case-sensitive is a source of confusion and conflict usually, although I'm not sure how much of that is a problem with query param matching.
Can we include the link to the RFC and mention the section which states the case-sensitiveness?

hbagdi:
Looks good to me.
We should merge with one or two more +1 given how core this is.

apis/v1alpha1/httproute_types.go
robscott: 🤦 thanks for catching that!

apis/v1alpha1/httproute_types.go
```diff
@@ -326,6 +343,38 @@ type HTTPHeaderMatch struct {
...
+	//
+	// HTTP query parameter matching MUST be case-sensitive for both keys and
+	// values.
+	//
+	// +optional
+	// +kubebuilder:default=Exact
+	Type *QueryParamMatchType `json:"type,omitempty"`
+
+	// Values is a map of HTTP query parameters to be matched. It MUST contain
+	// at least one entry.
```
youngnick, line 368: Nit: If this must contain at least one entry, should we set the validation minimum?

apis/v1alpha1/httproute_types.go
```diff
@@ -326,6 +343,38 @@ type HTTPHeaderMatch struct {
...
+	//
+	// The query parameter name to match is the map key, and the value of the
+	// query parameter is the map value.
+	//
+	// Multiple match values are ANDed together, meaning, a request must match
+	// all the specified query parameters to select the route.
```
youngnick, line 374: Nit: I feel like we should be clear that, even if you specify one of the ImplementationSpecific matches, we shouldn't allow regexes or other craziness in the query param *name*, just the value?

Something like:
```suggestion
	// all the specified query parameters to select the route.
	//
	// Note that the query param type matching MUST only be used for the query param value, not the query param name.
```

youngnick (commented):
/lgtm

with a couple of small doc nits, just to head off incompatible implementations before we end up there.

apis/v1alpha1/httproute_types.go
bowei: More direct would be (don't refer to param matching type to avoid indirection):
```
Query parameter key is must be an exact match (string comparison).
```

apis/v1alpha1/httproute_types.go
robscott: Added in #636

apis/v1alpha1/httproute_types.go
robscott: Unfortunately it does not look like it's possible to validate that with annotations. It does seem like something we should catch with the webhook though (for both this and header matching).

apis/v1alpha1/httproute_types.go
robscott: Good call, added in #636

apis/v1alpha1/httproute_types.go
```diff
@@ -270,6 +270,23 @@ const (
...
+// * "Exact"
+// * "RegularExpression"
+// * "ImplementationSpecific"
+//
+// +kubebuilder:validation:Enum=Exact;RegularExpression;ImplementationSpecific
+type QueryParamMatchType string
+
+// QueryParamMatchType constants.
+const (
+	QueryParamMatchExact                  QueryParamMatchType = "Exact"
```
jpeach, line 285: These need goidoc comments for the generated API reference.

PR #633, opened 2021-04-27
Enhance docs around Gateways, Routes and TLS

### LINKED ISSUE #612: Clarify how Gateways and Routes should interact with TLS
This is a followup to the great conversation in [Slack](https://kubernetes.slack.com/archives/CR0H13KGA/p1618300603046300). I think we reached a good consensus there, we just need to document it.

/assign @hbagdi

SantoDE:
**What type of PR is this?**

Enhance Docs with the relation between Gateways, Routes and TLS

Fixes #612

```release-note

```

### DIFF
--- a/site-src/guides/tls.md
+++ b/site-src/guides/tls.md
@@ -80,7 +80,7 @@ before an HTTP request is sent from the client.
 [TLS Certificate in Route](#tls-certificate-in-route) provides an example
 of how this feature can be used.
 
-Also, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependant on the protocol on the listener level. Listener Protocol `HTTPS` or `HTTP` allowes to bind against `HTTPRoute` as the TLS Termination is done at the listener level and thus, only http information are used for routing.
+Also, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependant on the protocol on the listener level. Listener Protocol `HTTPS` or `HTTP` allows to bind against `HTTPRoute` as the TLS Termination is done at the listener level and thus, only http information are used for routing.
 
 For `TLSRoute`, its allowed to bind against `TCP` and `TLS` depending on the mode. If the mode is `Terminate`, plain `TCP` routing is possible. If its `TLS`, its also possible to forward all TLS information for `TLS` Routing.
 
--- a/site-src/guides/tls.md
+++ b/site-src/guides/tls.md
@@ -85,7 +85,7 @@ before an HTTP request is sent from the client.
 [TLS Certificate in Route](#tls-certificate-in-route) provides an example
 of how this feature can be used.
 
-Also, as mentioned above, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependant on the protocol on the listener level. Listener Protocol `HTTPS` or `HTTP` allows to bind against `HTTPRoute` as the TLS Termination is done at the listener level and thus, only http information are used for routing.
+Also, as mentioned above, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependent on the protocol on the listener level. Listeners with `HTTPS` or `HTTP` protocols can use `HTTPRoute` as the TLS Termination is done at the listener level and thus, only HTTP information is used for routing.
 
 For `TLSRoute`, its allowed to bind against `TCP` and `TLS` depending on the mode. If the mode is `Terminate`, plain `TCP` routing is possible. If its `TLS`, its also possible to forward all TLS information for `TLS` Routing.
 
--- a/site-src/guides/tls.md
+++ b/site-src/guides/tls.md
@@ -17,6 +17,14 @@ For Gateways, there are two connections involved:
 With Gateway API, TLS configuration of downstream and
 upstream connections is managed independently.
 
+Depending on the Listener Protocol, different TLS modes and Route types are supported.
+
+Listener Protocol | TLS Mode | Route Type Supported
+--- | --- | ---
+TLS | Passthrough | TLSRoute
+TLS | Terminate | TCPRoute
+HTTPS | Terminate | HTTPRoute
+
 Please note that in case of `Passthrough` TLS mode, no TLS settings take
 effect as the TLS session from the client is NOT terminated at the Gateway.
 The rest of the document assumes that TLS is being terminated at the Gateway,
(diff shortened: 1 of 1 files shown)


SantoDE:
Ping @robscott, @stevesloka.

heres an initial draft. LMK if thats what you expected or what I should change. I got some more time :)

robscott:
Thanks for the work on this @SantoDE! I took some time to read through the original [Slack thread](https://kubernetes.slack.com/archives/CR0H13KGA/p1618300603046300) again and I think it could actually be summarized with a table. Since this is a TLS guide, I think we only need to cover the supported combinations that include TLS support, which also makes the table pretty simple:

| Listener Protocol | TLS Mode | Route Type Supported |
|-|-|-|
| TLS | Passthrough | TLSRoute |
| TLS | Terminate | TCPRoute |
| HTTPS | Terminate | HTTPRoute | 

This could fit in right before the "Please note that in case of Passthrough TLS mode" paragraph. I _think_ this kind of table along with a short intro/description may be all that we need here, but I may be missing some of the nuance. 

Adding some of the people who were on the original thread to get their feedback as well.

/cc @howardjohn @hbagdi @youngnick

SantoDE:
Yeah, thats a pretty neat idea. Let me change things up. Thanks @robscott

youngnick:
Love the table idea, the rest of the PR looks great too.

SantoDE:
@robscott I added the table with a little intro text, and based on @youngnick comment I kept the part below with more explanation. LMK what you think :)

site-src/guides/tls.md
```diff
@@ -17,6 +17,14 @@ For Gateways, there are two connections involved:
 With Gateway API, TLS configuration of downstream and
 upstream connections is managed independently.
 
+Depending on the Listener Protocol, different TLS modes and Route types are supported.
+
+Listener Protocol | TLS Mode | Route Type Supported
+--- | --- | ---
+TLS | Passthrough | TLSRoute
```
stevesloka, line 24: Should it also be possible to have protocol `TLS`, tls mode `Terminate`, with route type `TLSRoute`?

SantoDE: Given the Slack discussion, I understood it as "No". As TLSRoute will need TLS routing information which would be terminated? But maybe im wrong.

stevesloka: Nah sounds like I don't understand either. I think the bit I'm not grasping is what @howardjohn described in the thread in regards to the 3 protocols, `incoming protocol, routing protocol, and upstream protocol` and how those apply.

SantoDE: Good questions. Maybe @howardjohn @robscott @youngnick can help?

site-src/guides/tls.md
```diff
@@ -77,6 +85,12 @@ before an HTTP request is sent from the client.
 [TLS Certificate in Route](#tls-certificate-in-route) provides an example
 of how this feature can be used.
 
+Also, as mentioned above, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependant on the protocol on the listener level. Listener Protocol `HTTPS` or `HTTP` allows to bind against `HTTPRoute` as the TLS Termination is done at the listener level and thus, only http information are used for routing.
```
robscott, line 88: Nit: A couple small typos
```suggestion
Also, as mentioned above, the Route Kind (`HTTPRoute`, `TLSRoute`, `TCPRoute`) is dependent on the protocol on the listener level. Listeners with `HTTPS` or `HTTP` protocols can use `HTTPRoute` as the TLS Termination is done at the listener level and thus, only HTTP information is used for routing.
```

site-src/guides/tls.md
```diff
@@ -77,6 +85,12 @@ before an HTTP request is sent from the client.
...
+
+For `TLSRoute`, its allowed to bind against `TCP` and `TLS` depending on the mode. If the mode is `Terminate`, plain `TCP` routing is possible. If its `TLS`, its also possible to forward all TLS information for `TLS` Routing.
```
robscott, line 90:
```suggestion
Listeners with the `TLS` protocol must use `TLSRoute` when the mode is set to `Passthrough` and `TCPRoute` when the mode is `Terminate`.
```

site-src/guides/tls.md
```diff
@@ -77,6 +85,12 @@ before an HTTP request is sent from the client.
...
+
+For `TCPRoute`, only plain TCP routing (e.g. no SNI) is possible.
```
robscott, line 92: Listeners with the `TCP` protocol must use `TCPRoute`.

site-src/guides/tls.md
robscott: I think there are three protocols involved here:

Protocol to the Gateway - defined by the Listener
Protocol to be used for Routing - defined by the Route (selected by the Listener)
Protocol to be used upstream from the Gateway - defined by the upstream Service, usually the same as routing protocol

That's at least my interpretation of how the different pieces fit together. We likely need a follow up to better document these concepts.

site-src/guides/tls.md
robscott, line 88: Also one more tiny nit - do you mind wrapping these lines at 80 characters?

robscott (commented):
Thanks for the work on this @SantoDE! Mostly LGTM, just had some nits with the sentences at the end.

site-src/guides/tls.md
howardjohn: > I wonder then also how the AppProtocol field comes into play, meaning if you had a TCPProxy with TLS on the gateway and an AppProtocol of TLS would you terminate TLS at the gateway, then forward upstream TLS?

I would expect two valid configurations for AppProtocol=TLS personally:
* TLS=Passthrough, forward the original TLS
* TLS=Terminate, with BackendPolicy configuring to use TLS. This would terminate then originate a new TLS connection.

SantoDE: So should I change and list both modes?

robscott: I could be wrong, but I think this discussion is somewhat different than the kind of routes that are supported with each mode. So although I think this is important to document, I don't think we need to delay getting this PR in.

SantoDE: SGTM. Nitpick changes coming in soon

site-src/guides/tls.md
SantoDE: Done

SantoDE:
I hope I addressed your changes @robscott.

For the open point, I agree. I think we can do that after v0.3 tough.

robscott:
Thanks!

/lgtm
/approve

PR #645, opened 2021-04-30
Restructuring repo to support v1alpha2

robscott:
**What type of PR is this?**
/kind design
/kind api-change

This adds new sets of v1alpha2 APIs and docs. This should mark the end of changes to v1alpha1. Note that both the docs and API definitions are starting as an exact copy of the corresponding v1alpha1 files. I've also added a note to the readme to indicate that only v1alpha1 is supported at this point.

```release-note
NONE
```


robscott:
This is a pretty massive change that will likely require discussion. This is just one of several approaches we could take here. Adding a temporary hold on this until it seems like we have some consensus on how to approach this.

/hold

robscott:
Still need to fix some codegen things, but docs are working as I'd hoped. For example:

v1alpha1 spec shows by default: https://deploy-preview-645--kubernetes-sigs-gateway-api.netlify.app/references/spec/
v1alpha2 spec is available at /v1alpha2: https://deploy-preview-645--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/spec/

robscott:
I think I've got codegen working well now, so this should be ready to go. PTAL.

/cc @danehans @jpeach

danehans:
@robscott to see how the cluster api project handles api versioning and provide his findings to this PR.

robscott:
@danehans From what I can tell, cluster-api has taken the same general approach proposed here and used in upstream. The main branch represents current work towards the next version, and when needed, release-{version} branches are maintained for patches to older releases. Given that, I think it makes sense to proceed with this approach here as well.

bowei:
LGTM

hbagdi:
Hard to review but did a few `ls | shasum` to figure out if we missed any files and it doesn't seem like it. 
/lgtm

PR #649, opened 2021-05-04
tag latest releases in the staging repo

hbagdi:
**What type of PR is this?**

/kind feature
The latest tag in the staging repository gives an easy way to test the
latest and greatest image of the admission-webhook-server.
This is primarily intended for maintainers of this repository and for
implementers of this API who might want to target their CI systems
against this tag.

None

```release-note
NONE
```


hbagdi:
For clarity, this only tags `latest` for staging repository.
The production images that users of concrete implementation will rely on should be from the production repository.
Production images will be available in a different repository and the process to promote builds to production is documented here: https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/README.md#image-promoter

bowei:
Doesn't this replace the release? Do we need both build targets?

hbagdi:
> Doesn't this replace the release? Do we need both build targets?

If you mean the Makefile target, then yes. This patch renames the target so we won't have `release` target anymore.

bowei:
This looks ok but can you put in the commit comment why we are renaming release to release-staging (not obvious)

hbagdi:
> This looks ok but can you put in the commit comment why we are renaming release to release-staging (not obvious)

@bowei Updated the commit message.

robscott:
Thanks!

/lgtm
/approve

PR #651, opened 2021-05-06
add validation-webhook support for v1alpha2

hbagdi:
**What type of PR is this?**

/kind feature
**What this PR does / why we need it**:

This PR adds in validation support for v1alpha2.

**Which issue(s) this PR fixes**:

None

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #656, opened 2021-05-07
pkg/admission: Add Gateway validation to Admission Webhook

stevesloka:
**What type of PR is this?**
/kind feature

Adds Gateway validation to Admission Webhook

Updates #487 

```release-note
NONE
```
//cc @hbagdi

### DIFF
--- a/pkg/admission/server.go
+++ b/pkg/admission/server.go
@@ -51,6 +51,16 @@ var (
 		Version:  v1alpha2.SchemeGroupVersion.Version,
 		Resource: "httproutes",
 	}
+	v1a1gatewayGVR = meta.GroupVersionResource{
+		Group:    v1alpha1.SchemeGroupVersion.Group,
+		Version:  v1alpha1.SchemeGroupVersion.Version,
+		Resource: "gateways",
+	}
+	v1a2gatewayGVR = meta.GroupVersionResource{
+		Group:    v1alpha2.SchemeGroupVersion.Group,
+		Version:  v1alpha2.SchemeGroupVersion.Version,
+		Resource: "gateways",
+	}
 )
 
 func log500(w http.ResponseWriter, err error) {
@@ -157,7 +167,24 @@ func handleValidation(request admission.AdmissionRequest) (*admission.AdmissionR
 		if err != nil {
 			return nil, err
 		}
+
 		fieldErr = v1a2Validation.ValidateHTTPRoute(&hRoute)
+	case v1a1gatewayGVR:
+		var gateway v1alpha1.Gateway
+		_, _, err := deserializer.Decode(request.Object.Raw, nil, &gateway)
+		if err != nil {
+			return nil, err
+		}
+
+		fieldErr = v1a1Validation.ValidateGateway(&gateway)
+	case v1a2gatewayGVR:
+		var gateway v1alpha2.Gateway
+		_, _, err := deserializer.Decode(request.Object.Raw, nil, &gateway)
+		if err != nil {
+			return nil, err
+		}
+
+		fieldErr = v1a2Validation.ValidateGateway(&gateway)
 	default:
 		return nil, fmt.Errorf("unknown resource '%v'", request.Resource.Resource)
 	}
--- a/pkg/admission/server_test.go
+++ b/pkg/admission/server_test.go
@@ -59,6 +59,10 @@ func TestServeHTTPInvalidMethod(t *testing.T) {
 }
 
 func TestServeHTTPSubmissions(t *testing.T) {
+
+	const invalidHostname = "!@!.foo.com"
+	const lowercaseRFC1123ErrorMessage = "a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"
+
 	for _, apiVersion := range []string{
 		"admission.k8s.io/v1",
 		"admission.k8s.io/v1",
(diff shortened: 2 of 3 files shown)


pkg/admission/server.go
```diff
@@ -43,6 +43,11 @@ var (
 		Version:  v1alpha1.SchemeGroupVersion.Version,
 		Resource: "httproutes",
 	}
+	gatewayGVR = meta.GroupVersionResource{
+		Group:    v1alpha1.SchemeGroupVersion.Group,
+		Version:  v1alpha1.SchemeGroupVersion.Version,
```
hbagdi, line 48: #651 was merged in. Can we instead use v1alpha2?

This leads to a bigger question, do we want to implement validation for the latest version or add it for v1alpha2 as well?
I expect the transition from v1alpha1 to 2 will be slow but then do we continue to support alpha1 or give users more reasons to adopt alpha2 sooner.

cc @robscott @danehans

robscott: I think we should plan on supporting at least the current and trailing version of the API. That would mean v1alpha1 and v1alpha2 now, eventually transitioning to v1beta1 and v1alpha2 in the future.

stevesloka: I just rebased & added support for both versions.

pkg/admission/server.go
```diff
@@ -172,6 +182,36 @@ func handleValidation(request admission.AdmissionRequest) (*admission.AdmissionR
...
+			return nil, err
+		}
+
+		fieldErr := v1a2Validation.ValidateGateway(&gateway)
+		if fieldErr != nil {
+			message = fmt.Sprintf("%s", fieldErr.ToAggregate())
+			ok = false
+		} else {
+			ok = true
+		}
```
hbagdi, line 214: Please add `gateways` to `resource` in `deploy/admission_webhook.yaml`. Else, this validation will never be invoked.

pkg/admission/server.go
```diff
@@ -50,6 +50,16 @@ var (
...
+	v1a1gatewayGVR = meta.GroupVersionResource{
+		Group:    v1alpha1.SchemeGroupVersion.Group,
+		Version:  v1alpha1.SchemeGroupVersion.Version,
+		Resource: "gateways",
+	}
+	v1a2gatewayGVR = meta.GroupVersionResource{
+		Group:    v1alpha2.SchemeGroupVersion.Group,
+		Version:  v1alpha2.SchemeGroupVersion.Version,
+		Resource: "gateways",
+	}
```
hbagdi, line 62: (This is not in scope of this PR but I don't know where else to discuss this.)
Should these GVRs (and GVKs in future?) be defined in the `v1alphaX` package and be exported?
We don't have any duplication yet but I wanted to know if that is a good practice in general or is that advised against.

pkg/admission/server_test.go
```diff
@@ -120,6 +120,208 @@ func TestServeHTTPSubmissions(t *testing.T) {
...
+						"request": {
+							"uid": "7313cd05-eddc-4150-b88c-971a0d53b2ab",
+							"resource": {
+								"group": "networking.x-k8s.io",
+								"version": "v1alpha2",
+								"resource": "gateways"
+							},
+							"object": {
+   								"kind": "Gateway",
+   								"apiVersion": "networking.x-k8s.io/v1alpha1",
```
hbagdi, line 287:
```suggestion
   								"apiVersion": "networking.x-k8s.io/v1alpha2",
```

pkg/admission/server_test.go
```diff
@@ -120,6 +120,208 @@ func TestServeHTTPSubmissions(t *testing.T) {
...
```
hbagdi, line 238:
```suggestion
   								"apiVersion": "networking.x-k8s.io/v1alpha2",
```

pkg/admission/server_test.go
```diff
@@ -120,6 +120,208 @@ func TestServeHTTPSubmissions(t *testing.T) {
...
+						"operation": "CREATE"
+						}
+					}`),
+				wantRespCode: http.StatusOK,
+				wantSuccessResponse: admission.AdmissionResponse{
+					UID:     "7313cd05-eddc-4150-b88c-971a0d53b2ab",
+					Allowed: false,
+					Result: &metav1.Status{
+						Code:    400,
+						Message: "spec.listeners[0].hostname: Invalid value: \"!@!.foo.com\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')",
```
hbagdi, line 321: Can/Should we DRY out `a lowercase RFC...`?
Maybe define a string at the test package level and user that here and one other place we have?

hbagdi (commented):
Thanks for working on this!

pkg/admission/server.go
robscott: This is a good question. Although it could be useful, I don't think there's any upstream precedent for doing that. For example, comparing our [generated register file](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/zz_generated.register.go) with a corresponding [upstream register file](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/discovery/register.go) shows that we already have a nearly identical structure to upstream types files. Not saying we can't/shouldn't add to that though.

pkg/admission/server_test.go
stevesloka: 👍

pkg/admission/server_test.go
stevesloka: 👍

pkg/admission/server_test.go
stevesloka: Ok I pulled the hostname + error message out in a const. Is this what you were thinking @hbagdi?

pkg/admission/server.go
stevesloka: 👍

pkg/admission/server.go
stevesloka: I know in Contour we've duplicated the same to use them, so would be nice to have a central spot to reference. I'm not sure if that's a good practice or not.

pkg/admission/server_test.go
```diff
@@ -59,6 +59,10 @@ func TestServeHTTPInvalidMethod(t *testing.T) {
 }
 
 func TestServeHTTPSubmissions(t *testing.T) {
+
+	const invalidHostname = "!@!.foo.com"
+	const invalidHostnameErrorMsg = "spec.listeners[0].hostname: Invalid value: \"!@!.foo.com\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"
```
hbagdi, line 64: It might have been better to only extract `a lowercase ...` because the part before it has values that can change.

pkg/admission/server.go
```diff
@@ -157,7 +167,26 @@ func handleValidation(request admission.AdmissionRequest) (*admission.AdmissionR
...
+		deserializer := codecs.UniversalDeserializer()
+		_, _, err := deserializer.Decode(request.Object.Raw, nil, &gateway)
+		if err != nil {
+			return nil, err
+		}
+
+		fieldErr = v1a1Validation.ValidateGateway(&gateway)
+	case v1a2gatewayGVR:
+		var gateway v1alpha2.Gateway
+		deserializer := codecs.UniversalDeserializer()
```
hbagdi, line 183: @stevesloka I was about to approve but it seems like #662 was merged in. Could you rebase this on top of it and remove this line and line 174?

stevesloka: all rebased!

pkg/admission/server_test.go
stevesloka: ahh I think I follow you know, sorry, let me update with that.

hbagdi:
@stevesloka This is looking good. Can you please squash your commits and then we can merge this in?
The bot does a git merge which keeps the code review commit in the history which we don't need.
Thanks for working on this.

stevesloka:
No problem @hbagdi! All squashed up!

robscott:
Thanks!

/lgtm
/approve

robscott:
Also recently found out about the very handy `tide/merge-method-squash` label that can automate the squashing for us.

hbagdi:
> Also recently found out about the very handy tide/merge-method-squash label that can automate the squashing for us.

Thank you for sharing this!

PR #657, opened 2021-05-10
Switching Header and Query Param matching to use lists

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This matches the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) that prefer lists of named subobjects over maps.

```release-note
Header and Query Param matching are now represented with lists of named subobjects instead of maps.
```


apis/v1alpha2/httproute_types.go
```diff
@@ -406,15 +409,24 @@ type HTTPRouteMatch struct {
...
 
-	// Headers specifies a HTTP request header matcher.
+	// Headers specifies HTTP request header matchers. Multiple match values are
+	// ANDed together, meaning, a request must match all the specified headers
+	// to select the route.
+	//
+	// If multiple entries specify equivalent header names, the first entry with
+	// an equivalent name MUST be given precedence. Due to the
+	// case-insensitivity of header names, "foo" and "Foo" are considered
+	// equivalent.
```
hbagdi, line 419: Question for clarification:
```yaml
headers:
- name: foo
  value: bar
- name: Foo
   value: bar1
```

What happens if a request with header `foo: bar1` comes in? Is it matched or not?
I'm not sure if I clearly understand the meaning of "precedence" here.

robscott: Yeah, this could be clarified better. I was trying to say `foo: bar` should be given precedence here since it appears first in the list. That means that `foo: bar1` would not match (unless type was "Prefix").

stevesloka:
lgtm with @hbagdi's comments. =)

apis/v1alpha2/httproute_types.go
```diff
@@ -336,18 +319,38 @@ type HTTPHeaderMatch struct {
...
+	// +kubebuilder:validation:MaxLength=256
+	Name string `json:"Name"`
+
+	// Value is the value of HTTP Header to be matched.
 	//
-	// Multiple match values are ANDed together, meaning, a request
-	// must match all the specified headers to select the route.
-	Values map[string]string `json:"values"`
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=4096
```
hbagdi, line 333: 4096 should be good enough for most cases but someone might come in future wanting more. We can revisit then.
People use HTTP headers for all sorts of use-cases.

apis/v1alpha2/httproute_types.go
```diff
@@ -336,18 +319,38 @@ type HTTPHeaderMatch struct {
...
+	// Name is the name of the HTTP Header to be matched. Name matching MUST be
+	// case insensitive. (See
+	// https://tools.ietf.org/html/rfc7230#section-3.2).
 	//
-	// The HTTP header field name to match is the map key, and the
-	// value of the HTTP header is the map value. HTTP header field name matching
-	// MUST be case-insensitive.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	Name string `json:"Name"`
```
hbagdi, line 328: I don't feel strongly about this but is `Key` a better name here?
In my interpretation, API conventions don't explicitly call out that this field should be `Name`.

robscott: Yeah I went back and forth between the two. Ultimately I chose "name" due to the RFC using "header name": https://datatracker.ietf.org/doc/html/rfc7230#section-3.2. Admittedly the same argument could be used for "key" for query params, but I'd rather keep the 2 structures consistent in this API: https://datatracker.ietf.org/doc/html/rfc3986#section-3.4.

robscott: A quick search of the k8s codebase shows both `key` and `name` are used in these cases, though it looks like `name` is a bit more common. As an example, this pattern is used for Pod DNS config: https://github.com/kubernetes/kubernetes/blob/fbdc4f6acbc881eb26544d9d68ae520b7cbad3e7/pkg/apis/core/types.go#L3055

apis/v1alpha2/httproute_types.go
jpeach: This is in conflict with the general rule, which is "multiple match values are ANDed". So I'm not sure that we need to add the complexity of a special case for duplicates (as a quality of implementation issue, maybe validation could reject this case or a controller could emit a warning).

robscott: so in the above example no requests should match because the result of an AND here would always be false?

apis/v1alpha2/httproute_types.go
hbagdi: I don't feel strongly about it. Let's leave this as `name`. This thread will help in future if it comes up. Resolving now.

apis/v1alpha2/httproute_types.go
jpeach: > so in the above example no requests should match because the result of an AND here would always be false?

Yup, exactly.

hbagdi: That is hard to understand based on the current language. I've a suggestion to improve the language a little bit here.
I think having validation to prevent the user from providing two different values for the same header with type `Exact` is a good idea. We should open an issue to track it if you all

jpeach:
On May 13, 2021 at 3:37 PM, Harry ***@***.***> wrote:






@hbagdi commented on this pull request.

In apis/v1alpha2/httproute_types.go:

> @@ -406,15 +409,24 @@ type HTTPRouteMatch struct {
 	Path *HTTPPathMatch `json:"path,omitempty"`
 
-	// Headers specifies a HTTP request header matcher.
+	// Headers specifies HTTP request header matchers. Multiple match values are
+	// ANDed together, meaning, a request must match all the specified headers
+	// to select the route.
+	//
+	// If multiple entries specify equivalent header names, the first entry with
+	// an equivalent name MUST be given precedence. Due to the
+	// case-insensitivity of header names, "foo" and "Foo" are considered
+	// equivalent.


That is hard to understand based on the current language. I've a suggestion to improve the language a little bit here.
I think having validation to prevent the user from providing two different values for the same header with type Exact is a good idea. We should open an issue to track it if you all agree.

We can just be more explicit about  cane-insensitivity of HTTP headers (i.e. call it out in a separate sentence up front rather than an aside at the end). I agree that adding duplicate detection in the validation is good, but you will want to be careful with character sets (ie. USASCII, not UTF8).

apis/v1alpha2/httproute_types.go
robscott: So I spent awhile trying to find the right language here, and everything I typed out was either too obvious, too confusing, or both. I think it might be clearest to leave it as is:

- Name matching is case insensitive
- Matches are ANDed together

Given those rules, the above example could never result in a match. Does that seem reasonable?

apis/v1alpha2/httproute_types.go
```diff
@@ -366,21 +367,21 @@ type HTTPQueryParamMatch struct {
...
-	// all the specified query parameters to select the route.
-	//
-	// HTTP query parameter matching MUST be case-sensitive for both keys and
-	// values. (See https://tools.ietf.org/html/rfc7230#section-2.7.3).
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	Name string `json:"Name"`
+
+	// Value is the value of HTTP Header to be matched. Value matching
+	// matching MUST be case-sensitive. (See
```
bowei, line 379: Can we just say the value matches if it is the same (string). Calling out case-sensitive leads people to wonder about what other special cases there are (there aren't any).

apis/v1alpha2/httproute_types.go
```diff
@@ -366,21 +367,21 @@ type HTTPQueryParamMatch struct {
 	// +kubebuilder:default=Exact
 	Type *QueryParamMatchType `json:"type,omitempty"`
 
-	// Values is a map of HTTP query parameters to be matched. It MUST contain
-	// at least one entry.
-	//
-	// The query parameter name to match is the map key, and the value of the
-	// query parameter is the map value.
+	// Name is the name of the HTTP query param to be matched. Name matching
+	// matching MUST be case-sensitive. (See
```
bowei, line 371: see my comment about just saying the string must be equal.

apis/v1alpha2/httproute_types.go
bowei: that sounds fine to me -- a user that put "Foo=a" and "foo=b" and expecting something is likely a rare case

bowei: Although -- rob that would mean the implementation needs to detect this and generate a "match none" configuration?

Or we could explicitly say, having multiple entries is up to implementation to interpret -- less work on implementors, but more vague for user.

It feels like the first one is not too bad but requires a special case handling in impls.

bowei: BTW -- we should probably add a comment that duplicate "Foo=A", "foo=B" will match nothing to make that behavior very clear, as opposed to "derived from the postulates"

apis/v1alpha2/httproute_types.go
robscott: The rationale for calling out case-sensitive here is that we're calling out for related fields as case-insensitive. I thought it would be helpful to contrast the case sensitivity for value with the case insensitivity for name. Of course if we didn't have the case-insensitive comment above I don't think this would be worthwhile.

robscott: 🤦 Just realized I have a typo from copy-pasting here. Will fix. Now that I realize this is for query param where both are case-sensitive, maybe it doesn't matter quite as much. I think the contrast with the very similar header matching text still may make it worthwhile, but don't have to stick with that.

apis/v1alpha2/httproute_types.go
robscott: There's a lot of nuance here and it's difficult to get the wording right. For example, the following would be valid:

```
- name: foo
  type: RegularExpression
  value: [a-z]
- name: Foo
  type: RegularExpression
  value: [a-z0-9]
```

It may not actually make any sense to configure a Route that way, but a value like `z` would be valid. And similar ambiguity would likely also exist for ImplementationSpecific match types. So the only thing we can say for sure is that specifying different values with an exact match for equivalent header names would not result in a match. And once I took the time to write that out, it seemed obvious, so I just omitted it. Maybe providing the specific example you suggested would be a better example?

robscott: Added an example here that I think helps explain how this should be interpreted.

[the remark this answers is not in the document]
robscott: Tweaked this comment based on discussion above.

apis/v1alpha2/httproute_types.go
robscott: Fixed typo and removed case-sensitive statement since it is not present for header value either.

apis/v1alpha2/httproute_types.go
robscott: I've left this in and added a simple example to contrast with header name matching. I don't have to stick with it, but wanted to get some feedback to see if this was easier to understand when compared with the header match spec.

bowei: Isn't exact string match non-ambiguous and case-sensitive? Comments like these require people to dig around for the full behavior (I would if I read that in the spec).

apis/v1alpha2/httproute_types.go
bowei: Actually given the regex example, we should just say if you repeat items with the same name, it will be "implementation specific" how this is implemented. Some infra may not support AND'ing regex together -- and ANDing regex is not an operator that is easy to do with a regex.

robscott: Good point that ANDing a regex is simply not a practical expectation here. Instead of falling back to implementation specific handling of this, I think it may make the most sense simply to revert back to something resembling the original approach. Specifying that implementations must ignore subsequent matches for equivalent names and only implement the first one should be relatively easy to implement for everyone and results in a consistent experience.

apis/v1alpha2/httproute_types.go
robscott: Yep, that's fair, updated to say that.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
robscott: Ended up with something very similar to this suggestion, thanks!

apis/v1alpha2/httproute_types.go
robscott: Agree that we can likely raise this if needed. This was roughly based on the lowest default limit among relatively common implementations, but of course everything is configurable: https://stackoverflow.com/a/8623061.

apis/v1alpha2/httproute_types.go
bowei: +1 makes sense, this is both easy to implement and consistent for users.

hbagdi:
I found a bug in our verify script and opened up #670 to fix it.
/lgtm

hbagdi:
@robscott It seems the examples didn't pass validation, I'm not sure why they failed. Could you rebase on the latest HEAD on our dev branch?

robscott:
@hbagdi it was actually a great catch - I had `Name` as the json field name instead of `name`. Should be fixed now.

Issue #189, opened 2020-05-11
GatewayClass controller immutability is not enforced


robscott:
**What happened**:
There is no enforcement of the claim that the controller field on GatewayClass is immutable.

**What you expected to happen**:
Either it should not be possible to change the controller field after creation or we should remove the comment stating that it is immutable.

For reference, the only way I've found to enforce this would be a validating webhook. There is currently no support for an annotation to mark a field immutable, though it is [planned](https://github.com/kubernetes/enhancements/issues/1101).

/cc @lelenanam

PR #663, opened 2021-05-12
implement validation for GatewayClass resource

hbagdi:
This currently comprises only of an immutability check for
`spec.controller` field. This can expanded in future.

**What type of PR is this?**

/kind feature
This validation is a pre-requisite for #189. Once this is merged in, we can start enforcing this using a validation webhook.

None

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/validation/validation_test.go
+++ b/apis/v1alpha2/validation/validation_test.go
@@ -17,11 +17,13 @@ limitations under the License.
 package validation
 
 import (
+	"reflect"
 	"testing"
 
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	utilpointer "k8s.io/utils/pointer"
 )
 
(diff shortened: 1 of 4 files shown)


hbagdi:
@robscott @jpeach Should we be implementing this validation for v1alpha1 and v1alpha2 both?

jpeach:
> @robscott @jpeach Should we be implementing this validation for v1alpha1 and v1alpha2 both?

My preference would be just v1alpha2

robscott:
> Should we be implementing this validation for v1alpha1 and v1alpha2 both?

My personal preference would be to support the current and trailing API versions with this webhook. That would mean `v1alpha1` and `v1alpha2` now, eventually transitioning to `v1beta1` and `v1alpha2` in the future. Interested in what others think here as well.

/cc @bowei @danehans

robscott:
This LGTM, my personal preference would be to also add identical validation for v1alpha1 as part of this PR, but I can be convinced otherwise.

apis/v1alpha2/validation/validation_test.go
```diff
@@ -424,3 +426,76 @@ func portNumberPtr(p int) *gatewayv1a2.PortNumber {
...
+				},
+				newClass: &gatewayv1a2.GatewayClass{
+					Spec: gatewayv1a2.GatewayClassSpec{
+						Controller: "bar",
+					},
+				},
+			},
+			want: field.ErrorList{
+				{
+					Type:     field.ErrorTypeInvalid,
```
bowei, line 477: my preference in these kind of tests is to just test for the error Type and not do DeepEqual. DeepEqual makes it so that we have to change a lot of tests if the error message needs to be updated.

hbagdi: Can a smart replace tool not help us there?
I'm not a huge fan of the current approach but it helps when there are tests that have multiple errors of the same type.

hbagdi:
> This LGTM, my personal preference would be to also add identical validation for v1alpha1 as part of this PR, but I can be convinced otherwise.

I've added this for now as it was trivial. If at some point, it get cumbersome to maintain these, let's revisit.
We can de-duplicate by converting from older to newer API types, is that acceptable (not in scope for this PR)?

apis/v1alpha2/validation/validation_test.go
robscott: Yeah in this case I think it's useful to check for the error message since that is an important part of the output here.  If it gets more complex in the future we may need to revisit this, but for now I think it's good for the tests to cover the message.

robscott:
Thanks! 

/lgtm
/approve

PR #664, opened 2021-05-12
breaking change: make object.spec required on all resources

### LINKED ISSUE #661: GatewayClass spec.controller non-empty requirement is not enforced
The `GatewayClass` `spec.controller` field is documented as "cannot be empty", but there's no minimum length validation, and the `spec` field itself is optional. This means that it is possible to create a `GatewayClass` with an empty controller name.

hbagdi:
**What type of PR is this?**

/kind bug
/kind api-change

Makes `spec` field required for all API types as having objects without any `spec` section should be of no use in this API.

Fixes #661

```release-note
For all resources in the API, `.spec` section is now required.
```

/cc @jpeach @robscott @danehans

### DIFF
--- a/apis/v1alpha2/backendpolicy_types.go
+++ b/apis/v1alpha2/backendpolicy_types.go
@@ -37,7 +37,7 @@ type BackendPolicy struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	// Spec defines the desired state of BackendPolicy.
-	Spec BackendPolicySpec `json:"spec,omitempty"`
+	Spec BackendPolicySpec `json:"spec"`
 
 	// Status defines the current state of BackendPolicy.
 	Status BackendPolicyStatus `json:"status,omitempty"`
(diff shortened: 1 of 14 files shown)


howardjohn (approved):

robscott:
I'm not actually sure we wanted to do this. It's very strange, and I don't know the rationale, but 100% of upstream types have optional spec and status from what I can tell. For example: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3828

hbagdi:
@robscott I can open a revert PR and re-open this for discussion.

I tried with a few resources in core and they error out as if `spec` is required.
For example, this service errors out:
```yaml
kind: Service
apiVersion: v1
metadata:
  name: filter-lb
```

On the other hand, if we take Cluster API for reference, it doesn't remove `omitempty` as done here.
I'm not against reverting this patch but I also don't know of any other way to solve this problem. Like you said, it is very strange.

robscott:
Yeah I know k/k has more validation built in to it. I don't think we need to revert this right away, but we should probably have an issue to follow up and get some kind of clear answer from apimachinery as a blocker to a v1alpha2 release.

hbagdi:
@robscott Tracking this in https://github.com/kubernetes-sigs/gateway-api/issues/665, let's continue the conversation there.

PR #667, opened 2021-05-14
pkg/admission: validate GatewayClass updates

### LINKED ISSUE #189: GatewayClass controller immutability is not enforced

hbagdi:
**What type of PR is this?**

/kind feature
This patch introduces validation for updates to GatewayClass resources.

Fixes #189

```release-note
Immutability of `spec.controller` field in GatewayClass resource is now enforced using the validation webhook.
```

/cc @robscott @stevesloka @jpeach

### DIFF
--- a/pkg/admission/server_test.go
+++ b/pkg/admission/server_test.go
@@ -603,6 +603,262 @@ func TestServeHTTPSubmissions(t *testing.T) {
 				},
 			},
 			{
+				name: "v1a1 GatewayClass create events do not result in an error",
+				reqBody: dedent.Dedent(`{
+						"kind": "AdmissionReview",
+						"apiVersion": "` + apiVersion + `",
+						"request": {
+							"uid": "7313cd05-eddc-4150-b88c-971a0d53b2ab",
+							"resource": {
+								"group": "networking.x-k8s.io",
+								"version": "v1alpha1",
+								"resource": "gatewayclasses"
+							},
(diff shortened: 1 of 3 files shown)


stevesloka:
lgtm!

pkg/admission/server_test.go
```diff
@@ -602,6 +602,262 @@ func TestServeHTTPSubmissions(t *testing.T) {
 					},
 				},
 			},
+			{
+				name: "v1a1 GatewayClass create events do not result in an error",
```
robscott, line 606: Not a blocker on this PR, but it may make sense to split these tests out into different files or at least different sections per API version so that if/when we remove support for v1alpha1 it's easier to pull it out in one piece.

robscott (commented):
Thanks for the work on this!

/lgtm

pkg/admission/server_test.go
hbagdi: I had the same thought. I've this on my todo.

PR #668, opened 2021-05-17
api/v1alpha2: Refactor validation into separate files

stevesloka:
**What type of PR is this?**
/kind cleanup

Refactors the validation package into separate files to make it
easier to manage as the functionality grows.

// @hbagdi 

```release-note
NONE
```


hbagdi:
/ok-to-test
I had this on my mind too but I didn't do it because it goes against upstream convention.
For example, [core validation file](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go) is 6k+ LOC!
@robscott @bowei Do you know if there is a reason for that pattern? And do we need to follow that pattern?

I personally favor this approach but want to make sure that we are not going against norms/standards.

robscott:
Thanks for this PR @stevesloka! I think this is a good change. Although k/k definitely has some long validation files with a single `validation.go` per API, I don't think that's a requirement for us. There are some great [upstream guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#validation) for validation, but none of them refer to the structure of the code. I agree with @hbagdi that this is a better approach. Would add a LGTM but it looks like `hack/verify-golint.sh` is failing.

robscott:
Thanks!

/approve

PR #672, opened 2021-05-18
Adding hostname address type

robscott:
**What type of PR is this?**
/kind feature
/kind api-change

This corresponds with the hostname field in the [upstream LoadBalancerStatus struct](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3547-L3550).

```release-note
Gateway status now supports a new Hostname address type
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -502,7 +502,7 @@ type GatewayAddress struct {
 // should raise the "Detached" listener status condition on
 // the Gateway with the "UnsupportedAddress" reason.
 //
-// +kubebuilder:validation:Enum=IPAddress;NamedAddress
+// +kubebuilder:validation:Enum=IPAddress;Hostname;NamedAddress
 type AddressType string
 
 const (
@@ -514,12 +514,19 @@ const (
 	// Support: Extended
 	IPAddressType AddressType = "IPAddress"
 
-	// An opaque identifier that represents a specific IP address. The
-	// interpretation of the name is dependent on the controller. For
-	// example, a "NamedAddress" might be a cloud-dependent identifier
-	// for a static or elastic IP.
+	// A Hostname represents a DNS based ingress point. This is similar to the
+	// corresponding hostname field in Kubernetes load balancer status. For
+	// example, this concept may be used for cloud load balancers where a DNS
+	// name is used to expose a load balancer.
 	//
-	// Support: Implementation-specific
+	// Support: Extended
+	HostnameAddressType AddressType = "Hostname"
+
+	// A NamedAddress provides a way to reference a specific IP address by name.
+	// For example, this may be a name or other unique identifier that refers
+	// to a resource on a cloud provider such as a static IP.
+	//
+	// Support: Implementation-Specific
 	NamedAddressType AddressType = "NamedAddress"
 )
 
--- a/config/crd/bases/networking.x-k8s.io_gateways.yaml
+++ b/config/crd/bases/networking.x-k8s.io_gateways.yaml
@@ -732,6 +732,7 @@ spec:
                       description: "Type of the address. \n Support: Extended"
                       enum:
                       - IPAddress
+                      - Hostname
                       - NamedAddress
                       type: string
                     value:
@@ -1123,6 +1124,7 @@ spec:
                       description: "Type of the address. \n Support: Extended"
                       enum:
                       - IPAddress
+                      - Hostname
                       - NamedAddress
                       type: string
                     value:


apis/v1alpha2/gateway_types.go
```diff
@@ -525,6 +525,12 @@ const (
 	//
 	// Support: Implementation-specific
 	NamedAddressType AddressType = "NamedAddress"
+
+	// A hostname represents a DNS based ingress points. For example, this
+	// concept may be used for AWS load-balancers.
+	//
+	// Support: Implementation-specific
+	HostnameAddressType AddressType = "Hostname"
```
howardjohn, line 533: Is it worth calling out explicitly the different in hostname vs NamedAddress? I think I understand it now but only when looking at the two fields docs side by side

hbagdi: Can we switch the ordering here a bit?
I expect IPAdressType and HostnameAddressType to be the more common one. In that order, NamesAddress creates less confusion too.

robscott: Good call, I've tried to update the `NamedAddress` type to be a bit less confusing and moved it after hostname to more clearly match how frequently the types will be used. Hopefully that helps. Open to any other ways to clarify the difference between these types.

apis/v1alpha2/gateway_types.go
```diff
@@ -518,12 +518,20 @@ const (
...
+	// Support: Extended
+	HostnameAddressType AddressType = "Hostname"
+
 	// An opaque identifier that represents a specific IP address. The
-	// interpretation of the name is dependent on the controller. For
-	// example, a "NamedAddress" might be a cloud-dependent identifier
-	// for a static or elastic IP.
+	// interpretation of the name is dependent on the controller. For example, a
+	// "NamedAddress" might be a name or other cloud-specific identifier that
+	// refers to a static or elastic IP on a cloud provider.
```
hbagdi, line 532: Should we remove this line?
Reading up to `cloud-specific identifier` makes sense but as soon as I read `static or elastic IP`, I get confused with IPAddressType. I know that `refers to` is the important part but I think we should come up with some better examples or skip examples here. Is this type intended to be used for something like [AWS ARNs](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html)?

robscott: Yeah I'm not really sure how to describe this well. I think an ARN is a great example of how a reference could work, but I'm hesitant to add too many specific examples here. IE if we suggest that an AWS ARN could be used here, users may assume that it works with whatever implementation they're using which may not be the case. 

At the same time, I know we did not have the `static or elastic IP` example initially but that resulted in confusion as far as what this should be used to refer to. I agree that there's room for improvement here but haven't found a great way to do that yet without adding more confusion.

robscott: Took another shot at this, but not sure it's any better. PTAL.

apis/v1alpha2/gateway_types.go
```diff
@@ -518,12 +518,19 @@ const (
 	// Support: Extended
 	IPAddressType AddressType = "IPAddress"
 
-	// An opaque identifier that represents a specific IP address. The
-	// interpretation of the name is dependent on the controller. For
-	// example, a "NamedAddress" might be a cloud-dependent identifier
-	// for a static or elastic IP.
+	// A hostname represents a DNS based ingress point. This is similar to the
```
hbagdi, line 521:
```suggestion
	// Hostname represents a DNS based ingress point. This is similar to the
```
nit: Don't feel too strongly about this one, I like Go's convention more (although this is not following that strictly either).

apis/v1alpha2/gateway_types.go
```diff
@@ -518,12 +518,19 @@ const (
...
 	//
-	// Support: Implementation-specific
+	// Support: Extended
+	HostnameAddressType AddressType = "Hostname"
+
+	// A NamedAddress provides a way to reference a specific IP address by name.
+	// For example, this may be a name or other unique identifier that refers
+	// to a static IP address on a cloud provider.
+	//
+	// Support: Implementation-Specific
```
hbagdi, line 533: Only leaving a note (no changes expected): Support in these cases will be based on a combination of the controller AND the cluster provider (IaaS, whatever) supporting and agreeing on some out-of-band API. This is abstract enough that we don't need to be prescriptive about it but we also have to be careful of users not using this in place of the other two more concrete types (for any reason).

apis/v1alpha2/gateway_types.go
```diff
@@ -518,12 +518,19 @@ const (
...
+	// example, this concept may be used for cloud load balancers where a DNS
+	// name is used to expose a load balancer.
 	//
-	// Support: Implementation-specific
+	// Support: Extended
+	HostnameAddressType AddressType = "Hostname"
+
+	// A NamedAddress provides a way to reference a specific IP address by name.
+	// For example, this may be a name or other unique identifier that refers
+	// to a static IP address on a cloud provider.
```
hbagdi, line 531:
```suggestion
	// A NamedAddress provides a way to reference a network endpoint by name.
	// For example, this may be a unique identifier that refers to a a cloud
	// provider-specific resource.
```

Getting completely rid of IP here might get the creativity of users flowing better.

robscott: Update to make this a bit more generic, but wanted to leave some reference to a static IP as an example so there was at least one concrete use case referenced by the spec.

apis/v1alpha2/gateway_types.go
robscott: I think @jpeach may have changed the format of these intentionally to make our reference docs a bit more readable. I could be wrong on the history, but it does read slightly better here: https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.AddressType.

PR #674, opened 2021-05-20
Update documentation around hostnames for HTTPRoute and Listener

### LINKED ISSUE #654: Clarify Gateway.Listener.Hostname interaction with HTTPRoute
I had a question when implementing `Gateway.Listener.Hostname` and how it interacts with `HTTPRoute.Spec.Hostnames` on Slack (Thread here: https://kubernetes.slack.com/archives/CR0H13KGA/p1620328409212700) with @hbagdi, @jpeach, & @robscott. 

My understanding of how it works was incorrect and wanted to open this issue to clarify how it should work so I could help update the docs. 

- An HTTPRoute that binds to a listener on the gateway MUST have hostnames that are a subset of the hostname(s) defined on the listener.
- A listener with `unspecified`, `“”, or *as hostname, then a HTTPRoute can have **any** hostname.
- A listener with `test.foo.com` as hostname, then a HTTPRoute can **only have the hostname** `test.foo.com`. All others won't match the Gateway and should function similar to `[RouteGateways](https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.RouteGateways)`?
- A listener with `*.foo.com` as hostname, then a HTTPRoute can have `test.foo.com` or `*.foo.com` or `steve.foo.com`. A HTTPRoute with `bar.com` would not match the Gateway and should function similar to `[RouteGateways](https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.RouteGateways)`?

In the example above where the Gateway has `*.foo.com`. What if my HTTPRoute had `test.foo.com` as well as `steve.bar.com` where one hostname matches, but another doesn't. Should we route on only `test.foo.com` or make the Route invalid? (Or maybe do both?)

youngnick:
This commit updates documentation around hostnames to make the
interaction between Gateway Listener `Hostname` and HTTPRoute
`Hostnames` field to be more clear.

Closes #654

**What type of PR is this?**
/kind cleanup
/kind documentation

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -144,18 +144,14 @@ type Listener struct {
 	// matched. This field can be omitted for protocols that don't require
 	// hostname based matching.
 	//
-	// Hostname is the fully qualified domain name of a network host, as defined
-	// by RFC 3986. Note the following deviations from the "host" part of the
-	// URI as defined in the RFC:
-	//
-	// 1. IP literals are not allowed.
-	// 2. The `:` delimiter is not respected because ports are not allowed.
-	//
-	// Hostname can be "precise" which is a domain name without the terminating
-	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
-	// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS label
-	// and matches only a single label.
+	// For HTTPRoute objects, there is an interaction with the
+	// `spec.hostnames` array. When both listener and route specify hostnames,
+	// there must be an intersection between the values for a Route to be admitted.
+	// For example, a Gateway with `*.example.com` would admit a Route that included
+	// `foo.example.com` as a hostname, but not a Route that *only* included
+	// `foo.acme.io` as a hostname. A Route that included both `foo.example.com`
+	// and `foo.acme.io` would be admitted, but the `foo.acme.io` hostname would
+	// be silently ignored.
 	//
 	// Support: Core
 	//
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -70,19 +70,40 @@ type HTTPRouteSpec struct {
 	// HTTPRoute rules. If no hostname is specified, traffic is routed
 	// based on the HTTPRouteRules.
 	//
-	// Hostname can be "precise" which is a domain name without the terminating
-	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is
-	// a domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS
-	// label and matches only a single label.
-	// You cannot have a wildcard label by itself (e.g. Host == `*`).
 	// Requests will be matched against the Host field in the following order:
 	//
-	// 1. If Host is precise, the request matches this rule if
-	//    the HTTP Host header is equal to Host.
-	// 2. If Host is a wildcard, then the request matches this rule if
+	// 1. If Hostname is precise, the request matches this rule if
+	//    the HTTP Host header is equal to the Hostname.
+	// 2. If Hostname is a wildcard, then the request matches this rule if
 	//    the HTTP Host header is to equal to the suffix
 	//    (removing the first label) of the wildcard rule.
+	// 3. If Hostname is unspecified, empty, or `*`, then any request will match
+	//    this route.
+	//
+	// If a hostname is specified by the Listener that the HTTPRoute is bound
+	// to, at least one hostname specified here must match the Listener specified
+	// hostname as per the rules above. Other hostnames will not affect processing
+	// of the route in that case.
+	//
+	// If no hostname is specified by the Listener, then that value will be treated
+	// as '*', match any hostname, and so any hostname on this Route will match.
+	//
+	// If all hostnames do not match, then the HTTPRoute is not admitted, and
+	// the implementation must raise an 'Admitted' Condition with a status of
+	// `false` for that Listener.
+	//
+	// Examples:
+	// - A Listener with unspecified, empty, or `*` values for Hostname matches
+	//   any HTTPRoute hostname.
+	// - A HTTPRoute with unspecified, empty, or `*` values for Hostname matches
+	//   any Listener hostname.
+	// - A Listener with `test.foo.com` as the hostname matches *only*
+	//   `test.foo.com` or `*.foo.com`. Any other hostnames present must be ignored.
+	// - A Listener with `*.foo.com` as hostname, all hostnames in the HTTPRoute
+	//   must have any single label where the star is, and the rest of the hostname
+	//   must match exactly. So, `test.foo.com`, `*.foo.com` or `blog.foo.com` match.
+	//   `test.blog.foo.com`, `test.bar.com`, or `bar.com` do not. Hostnames that do
+	//   not match will be ignored.
 	//
 	// Support: Core
 	//
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -245,7 +245,22 @@ type RouteStatus struct {
 	Gateways []RouteGatewayStatus `json:"gateways"`
 }
 
-// Hostname is used to specify a hostname that should be matched.
+// Hostname is the fully qualified domain name of a network host, as defined
+// by RFC 3986. Note the following deviations from the "host" part of the
+// URI as defined in the RFC:
+//
+// 1. IP literals are not allowed.
+// 2. The `:` delimiter is not respected because ports are not allowed.
+//
+// Hostname can be "precise" which is a domain name without the terminating
+// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
+// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
+// The wildcard character `*` must appear by itself as the first DNS label
+// and matches only a single label.
+//
+// Note that as per RFC1035 and RFC1123, a *label* must consist of lower case
+// alphanumeric characters or '-', and must start and end with an alphanumeric
+// character. No other punctuation is allowed.
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
--- a/config/crd/bases/networking.x-k8s.io_gateways.yaml
+++ b/config/crd/bases/networking.x-k8s.io_gateways.yaml
@@ -788,17 +788,15 @@ spec:
                         for protocol types that define this concept. When unspecified,
                         \"\", or `*`, all hostnames are matched. This field can be
                         omitted for protocols that don't require hostname based matching.
-                        \n Hostname is the fully qualified domain name of a network
-                        host, as defined by RFC 3986. Note the following deviations
-                        from the \"host\" part of the URI as defined in the RFC: \n
-                        1. IP literals are not allowed. 2. The `:` delimiter is not
-                        respected because ports are not allowed. \n Hostname can be
-                        \"precise\" which is a domain name without the terminating
-                        dot of a network host (e.g. \"foo.example.com\") or \"wildcard\",
-                        which is a domain name prefixed with a single wildcard label
-                        (e.g. `*.example.com`). The wildcard character `*` must appear
-                        by itself as the first DNS label and matches only a single
-                        label. \n Support: Core"
+                        \n For HTTPRoute objects, there is an interaction with the
+                        `spec.hostnames` array. When both listener and route specify
+                        hostnames, there must be an intersection between the values
+                        for a Route to be admitted. For example, a Gateway with `*.example.com`
+                        would admit a Route that included `foo.example.com` as a hostname,
+                        but not a Route that *only* included `foo.acme.io` as a hostname.
+                        A Route that included both `foo.example.com` and `foo.acme.io`
+                        would be admitted, but the `foo.acme.io` hostname would be
+                        silently ignored. \n Support: Core"
                       maxLength: 253
                       minLength: 1
                       type: string
--- a/config/crd/bases/networking.x-k8s.io_httproutes.yaml
+++ b/config/crd/bases/networking.x-k8s.io_httproutes.yaml
@@ -1041,21 +1041,47 @@ spec:
                   not allowed. 2. The `:` delimiter is not respected because ports
                   are not allowed. \n Incoming requests are matched against the hostnames
                   before the HTTPRoute rules. If no hostname is specified, traffic
-                  is routed based on the HTTPRouteRules. \n Hostname can be \"precise\"
-                  which is a domain name without the terminating dot of a network
-                  host (e.g. \"foo.example.com\") or \"wildcard\", which is a domain
-                  name prefixed with a single wildcard label (e.g. `*.example.com`).
-                  The wildcard character `*` must appear by itself as the first DNS
-                  label and matches only a single label. You cannot have a wildcard
-                  label by itself (e.g. Host == `*`). Requests will be matched against
-                  the Host field in the following order: \n 1. If Host is precise,
-                  the request matches this rule if    the HTTP Host header is equal
-                  to Host. 2. If Host is a wildcard, then the request matches this
-                  rule if    the HTTP Host header is to equal to the suffix    (removing
-                  the first label) of the wildcard rule. \n Support: Core"
+                  is routed based on the HTTPRouteRules. \n Requests will be matched
+                  against the Host field in the following order: \n 1. If Hostname
+                  is precise, the request matches this rule if    the HTTP Host header
+                  is equal to the Hostname. 2. If Hostname is a wildcard, then the
+                  request matches this rule if    the HTTP Host header is to equal
+                  to the suffix    (removing the first label) of the wildcard rule.
+                  3. If Hostname is unspecified, empty, or `*`, then any request will
+                  match    this route. \n If a hostname is specified by the Listener
+                  that the HTTPRoute is bound to, at least one hostname specified
+                  here must match the Listener specified hostname as per the rules
+                  above. Other hostnames will not affect processing of the route in
+                  that case. \n If no hostname is specified by the Listener, then
+                  that value will be treated as '*', match any hostname, and so any
+                  hostname on this Route will match. \n If all hostnames do not match,
+                  then the HTTPRoute is not admitted, and the implementation must
+                  raise an 'Admitted' Condition with a status of `false` for that
+                  Listener. \n Examples: - A Listener with unspecified, empty, or
+                  `*` values for Hostname matches   any HTTPRoute hostname. - A HTTPRoute
+                  with unspecified, empty, or `*` values for Hostname matches   any
+                  Listener hostname. - A Listener with `test.foo.com` as the hostname
+                  matches *only*   `test.foo.com` or `*.foo.com`. Any other hostnames
+                  present must be ignored. - A Listener with `*.foo.com` as hostname,
+                  all hostnames in the HTTPRoute   must have any single label where
+                  the star is, and the rest of the hostname   must match exactly.
+                  So, `test.foo.com`, `*.foo.com` or `blog.foo.com` match.   `test.blog.foo.com`,
+                  `test.bar.com`, or `bar.com` do not. Hostnames that do   not match
+                  will be ignored. \n Support: Core"
                 items:
-                  description: Hostname is used to specify a hostname that should
-                    be matched.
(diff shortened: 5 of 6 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -144,18 +144,13 @@ type Listener struct {
...
-	// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS label
-	// and matches only a single label.
+	// For HTTPRoute objeccts, there is an interaction with the
+	// `spec.hostnames` array. In short, if a hostname is specified in the
+	// Listener, then all the HTTPRoute's hostnames must match the hostname
+	// specified here, including both precise and wildcard hostname types.
+	// See the HTTPRoute `spec.hostname` fields for information about the exact
+	// rules used, and see the `Hostname` type for definitions of `precise` and
+	// `wildcard` hostnames.
```
youngnick, line 153: I moved the rules for hostnames to the `Hostname` struct, but wanted to note the possible interaction here, as I felt that means we are covered for both users (cluster admin and application developer). I can see an argument for not having this here though.

apis/v1alpha2/shared_types.go
```diff
@@ -244,7 +244,22 @@ type RouteStatus struct {
 	Gateways []RouteGatewayStatus `json:"gateways"`
 }
 
-// Hostname is used to specify a hostname that should be matched.
+// Hostname is the fully qualified domain name of a network host, as defined
+// by RFC 3986. Note the following deviations from the "host" part of the
+// URI as defined in the RFC:
+//
+// 1. IP literals are not allowed.
```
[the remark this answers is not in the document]
jpeach: Per RFC 6066, the name in a SNI message is not permitted to be a IP address literal.

apis/v1alpha2/shared_types.go
```diff
@@ -244,7 +244,22 @@ type RouteStatus struct {
...
+// 2. The `:` delimiter is not respected because ports are not allowed.
```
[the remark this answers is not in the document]
jpeach: This is a perennial problem with ports. If you don't strip the port and allow the `foo:port` host header to go upstream, then you run the risk of the upstream not knowing it ought to accept the port. Probably worth discussing this topic in a separate issue.

apis/v1alpha2/shared_types.go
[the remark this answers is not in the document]
howardjohn: Hostname is used for more than just SNI though. The `Host` header allows IP literal?

apis/v1alpha2/httproute_types.go
```diff
@@ -70,20 +70,30 @@ type HTTPRouteSpec struct {
...
-	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is
-	// a domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS
-	// label and matches only a single label.
-	// You cannot have a wildcard label by itself (e.g. Host == `*`).
 	// Requests will be matched against the Host field in the following order:
 	//
-	// 1. If Host is precise, the request matches this rule if
+	// 1. If Hostname is precise, the request matches this rule if
 	//    the HTTP Host header is equal to Host.
```
jpeach, line 76: Suggest: "is equal to the Hostname"?

apis/v1alpha2/httproute_types.go
```diff
@@ -70,20 +70,30 @@ type HTTPRouteSpec struct {
...
+	// If a hostname is specified by the Listener that the HTTPRoute is bound
+	// to, all hostnames specified here must match specified hostname as per
+	// the rules above. If any hostname does not match, then the entire Route
+	// is not admitted, and the implementation must raise an 'Admitted'
+	// Condition with a status of `false` for that Listener.
+	//
+	// Examples:
+	// - A Listener with unspecified, empty, or `*` values for Hostname matches
+	//   any HTTPRoute hostname.
+	// - A Listener with `test.foo.com` as the hostname matches *only*
```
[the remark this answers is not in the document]
jpeach: IMHO no, because it's a 2-level match. A request would have to match the listener (and implicitly the TLS certificates attached to the listener) before it can match a hostname on any routes.

apis/v1alpha2/httproute_types.go
```diff
@@ -70,20 +70,30 @@ type HTTPRouteSpec struct {
...
+	//   `test.foo.com`. Any other hostnames present will invalidate the HTTPRoute.
+	// - A Listener with `*.foo.com` as hostname, all hostnames in the HTTPRoute
+	//   must have any single label where the star is, and the rest of the hostname
```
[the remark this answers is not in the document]
jpeach: IIRC in envoy, you would match the wildcard at the server_names filter chain match, then add specific virtual hosts below that.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
howardjohn: There is a distinction between the runtime matching and configuration matching.

If the listener says test.foo.com, I would agree at *runtime* we can only match test.foo.com. But I don't see why I cannot have a route that specifies `*.foo.com` be used.

For example, I create one route with `*.foo.com`, then select a few gateways with hostnames `a.foo.com`, `b.foo.com`, etc. 

This makes it consistent with the other direction as well - the listener and route hostname are a [bidirectional wildcard](https://github.com/istio/istio/blob/65bd2ac632cbd6ec8e3fd85bb9874f685aea0cfd/pkg/config/host/name.go#L36)

apis/v1alpha2/shared_types.go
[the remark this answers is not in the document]
jpeach: Yeh, but it would be confusing to have different kinds of hostname with different semantics. IIRC the idea is to have consisent hostname semantics across the API.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
howardjohn: server_names is SNI, not hostname. We can say that SNI must == Hostname, but that is a whole other can of worms. Even if we do say SNI must == Hostname, IIRC Envoy doesn't support that either?

apis/v1alpha2/shared_types.go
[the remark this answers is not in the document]
howardjohn: I don't know.. I see your point but it feels a bit like hiding the complexity of the realities of the internet behind a overly simplified API. A bit like Ingress. Would be interested in others opinions here.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
youngnick: Envoy can be made to support this matching - we do it at the moment in Contour. It's not straightforward though.

Same goes for making SNI == Hostname. This prevents domain fronting (which I'd argue is undesirable most of the time anyway), but is only implementable right now with a Lua addition, which Contour also does already. :)

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
jpeach: > There is a distinction between the runtime matching and configuration matching.
> 
> If the listener says test.foo.com, I would agree at _runtime_ we can only match test.foo.com. But I don't see why I cannot have a route that specifies `*.foo.com` be used.

Oh I see what you mean. Yeh that seems perfectly reasonable to me. The runtime effect would be that only `test.foo.com` is exposed.

apis/v1alpha2/shared_types.go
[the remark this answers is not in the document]
youngnick: Yes, for both of these bullet items, I'm moving the already-existing definitions from `Listener` to `Hostname`. So we should probably discuss any changes on a separate issues, as they have strong implications, whether we leave them how they are or change them.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
jpeach: Well, envoy can be configured to enforce SNI == Hostname.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
youngnick: So, you're thinking of the usecase where you have a service, and you want to be able to serve the site out of a bunch of domains? I can kind of see that, I guess.

But our experience when implementing the Ingress API (which does technically allow this pattern) was that it created a very large number of edge cases, especially because there's a *list* of hostnames. But that was because of the interaction between the hostnames under the route section and the hostnames under the TLS section.

I think the Istio matching code makes how the matches would work pretty clear.

tl;dr I'm okay with changing this if everyone else is.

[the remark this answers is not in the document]
howardjohn: > So, you're thinking of the usecase where you have a service, and you want to be able to serve the site out of a bunch of domains? I can kind of see that, I guess.

TBH that example is probably a bit contrived, a more common one is the route has `*` ("I don't care how this is exposed") and the gateway picks the domain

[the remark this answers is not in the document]
youngnick: I can see that one too, thanks.

apis/v1alpha2/gateway_types.go
```diff
@@ -144,18 +144,13 @@ type Listener struct {
...
-	//
-	// 1. IP literals are not allowed.
-	// 2. The `:` delimiter is not respected because ports are not allowed.
-	//
-	// Hostname can be "precise" which is a domain name without the terminating
-	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
-	// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS label
-	// and matches only a single label.
+	// For HTTPRoute objeccts, there is an interaction with the
```
jpeach, line 147:
```suggestion
	// For HTTPRoute objects, there is an interaction with the
```

apis/v1alpha2/httproute_types.go
```diff
@@ -70,20 +70,30 @@ type HTTPRouteSpec struct {
...
+	//    the HTTP Host header is equal to the Hostname.
+	// 2. If Hostname is a wildcard, then the request matches this rule if
 	//    the HTTP Host header is to equal to the suffix
 	//    (removing the first label) of the wildcard rule.
 	//
+	// If a hostname is specified by the Listener that the HTTPRoute is bound
+	// to, all hostnames specified here must match specified hostname as per
+	// the rules above. If any hostname does not match, then the entire Route
+	// is not admitted, and the implementation must raise an 'Admitted'
+	// Condition with a status of `false` for that Listener.
```
hbagdi, line 85: > the entire Route

This goes against our guidelines: https://gateway-api.sigs.k8s.io/concepts/guidelines/
Specifically these two:
- Prefer not to break things that are working.
- Drop as little traffic as possible.

Could we use the following guideline?
> Make it clear which path has been chosen when a conflict has been identified. Where possible, this should be communicated by setting appropriate status conditions on relevant resources.

So, set the status of the route to reflect that one of the hostname is not admitted but others are?

youngnick: We totally can do that, but the problem is then, how do we communicate that? Right now, we only have a single Condition, `Admitted` to cover all the errors in the Route, so I was using that.

I'd guess in this case, we should add another, negative-polarity (so it's only `status: true` if there's a problem) Condition that covers either the more general "Some config was ignored" case or the the specific "This hostname was ignored" case. 

If we add any Conditions though, we probably should add other Conditions to cover other partial error cases too, I think.

youngnick: So, to be clear, if we're going to make this more nuanced (which I think is a good idea), then we will need more Condition coverage, as `Admitted` as a boolean isn't capable of conveying "mostly okay, some problems".

The two options are:
- add a single new Condition called something like "PartialProblem", or - probably better - "Warning", which would normally be `status: false`, and would change to `status: true` in the case of some error that's not bad enough to mean you need to throw away the whole object (which `Admitted, status: false` implies). This doesn't seem to be how other objects have used Conditions though.
- add extra Conditions for each type of partial thing that could go wrong. In this case, we'd add an `InvalidHostname` Condition that can be `status: true` or `status: false`, without necessarily impacting the `Admitted` state. I like this one better because it's more like the `Node` conditions, which feel like the best use of Conditions to me.

I'll leave this for feedback from @hbagdi, @robscott, @jpeach, or other interested parties. It could also maybe use a document/more detailed discussion, I suppose.

robscott: I've always thought of hostnames here as being a matching mechanism more than a binding mechanism. Although those concepts are awfully similar, I think there's a distinction here. I've also thought that the hostnames list was interpreted with an OR, not AND.

Let's say a Route owner wants to match requests to foo.com and bar.com, if a Gateway is bound that only supports foo.com I don't think that's an error state. Imagine a separate Gateway binds to this route and supports bar.com. We don't want each set of Gateway conditions in the Route status to include warning messages that they each only support 1 of the 2 hostnames specified. Instead I think it would be more appropriate if they were able to indicate in status which hostnames they did support. 

If there are no Gateway listeners with foo.com or bar.com that select this route, I think that's also fine and more of a lack of selection than an error.

howardjohn: There may be O(10k)+ hostnames, I am not sure this is feasible.

jpeach: > I've always thought of hostnames here as being a matching mechanism more than a binding mechanism. 

Yep, I've described it like that before as well. Operationally, an org may end up encoding multiple hostnames into the Route, but only exposing a subset of those through the Gateway Listener. That's alright IMHO. I also think it's OK to not have conditions expressing things that don't happen (in this case, a potential route match that can't happen).

youngnick: For other cases where we have a lot of possible options, we already have an upper limit of 100 in some places, maybe we put that on a `matchedHostnames` field?

If we don't have *something* that tells you what hostnames your Route is selecting though, then a person will need to look at the Gateway and listener hostnames, and take a guess. That does not seem like a good user experience to me.

howardjohn: Oh I was thinking this was on the gateway, so aggregating all routes together. On the route it's *likely* less problematic. A route could select 100s of gateways but that seems pretty odd; if we set a limit it would probably handle most cases.

Are there other examples of status fields that are capped and thus are somewhat "best effort"?

robscott: > Are there other examples of status fields that are capped and thus are somewhat "best effort"?

Within Gateway API, we already limit the number of gateways in Route status to 100: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/shared_types.go#L238. Given the large limit there, we may want to have a smaller limit for number of matched hostnames per GW. 

I agree that a concept like `matchedHostnames` could be quite helpful here in many cases. I do have some questions though:

1. Does a matchedHostname simply represent a hostname from any listener in the GW matching the route?
2. Would this still be reasonable/sensible to compute if Listener was expanded to support multiple hostnames?
3. If the GW has at least one wildcard listener would the matchedHostname list be a direct copy of the route hostname list?
4. What happens if the route hostname list is also empty/unspecified?

jpeach: Let's not get ahead of ourselves here. Is there user experience that suggests we need to highlight this in the API? Do we really need to add more fields or just make the documentation clear to set the expectation?

robscott: > Let's not get ahead of ourselves here. Is there user experience that suggests we need to highlight this in the API? Do we really need to add more fields or just make the documentation clear to set the expectation?

I do think that this kind of field would be helpful, but yeah, this potential field in status is likely a large enough topic that it could move to a separate issue. No need for us to bundle a new status field with this PR. 

I think we're at a point where we agree on how this should be interpreted and the main question is around how we can make that easier for users to understand. The better documentation that will be added by this PR is a great step and a status field could also be helpful as a follow up.

youngnick: That seems fair. So, how to proceed with Harry's point in the beginning? Right now, with the existing status structure, I don't think there's any way to communicate partial validity. So we have two options:
- we keep what I've put here, and if there's a hostname problem, we throw the whole route out. It seems like noone wants that.
- We allow partial validity, and don't tell people when their config is partially invalid? Or do we put that as a Reason when `Admitted` is `status: true`?

robscott: I don't think this is partial validity, more partial matching. For example, a Gateway may be listening on "*" but only have DNS pointed to it for foo.example.com. If a Route accepts requests for both foo.example.com and bar.example.com, I don't think we need to add any kind of special status message. I think the real thing we care about here is if there is _no way_ that a request to a Gateway will be routed with the specific Route resource. In that case, I think `Admitted` should be set to false. That includes if there's no hostname intersection between route and listener. I don't think we need to be concerned about a Route accepting requests from other hostnames though. 

I think it's likely going to be relatively common for a Route to accept multiple hostnames, and for those hostnames to be served by different Gateways. I don't think we need any kind of warning state for that.

youngnick: That makes sense to me. I'll try and update this with something to that effect.

youngnick: Okay, this is updated now, I think this one is done?

apis/v1alpha2/gateway_types.go
```diff
@@ -144,18 +144,13 @@ type Listener struct {
...
-	// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
-	// The wildcard character `*` must appear by itself as the first DNS label
-	// and matches only a single label.
+	// For HTTPRoute objects, there is an interaction with the
+	// `spec.hostnames` array. In short, if a hostname is specified in the
+	// Listener, then all the HTTPRoute's hostnames must match the hostname
+	// specified here, including both precise and wildcard hostname types.
+	// See the HTTPRoute `spec.hostname` fields for information about the exact
+	// rules used, and see the `Hostname` type for definitions of `precise` and
+	// `wildcard` hostnames.
```
robscott, line 153: I think this section still needs to be updated to match the conversation below. Maybe some variation of this would help?

>When both listener and route specify hostnames, there must be an intersection between the values for a Route to be admitted. For example, a Gateway with *.example.com would admit a Route that included foo.example.com as a hostname, but not a Route that only included foo.acme.io as a hostname.

robscott (commented):
Thanks for the work on this @youngnick!

apis/v1alpha2/shared_types.go
```diff
@@ -245,7 +245,22 @@ type RouteStatus struct {
...
+//
+// Hostname can be "precise" which is a domain name without the terminating
+// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
+// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
+// The wildcard character `*` must appear by itself as the first DNS label
+// and matches only a single label.
+//
+// Note that as per RFC1035, a *label* is a string containing only characters
+// from the sets "a-z", "A-Z", and the literal "-". No other punctuation is
+// allowed.
```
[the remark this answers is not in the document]
youngnick: Thanks, updated.

apis/v1alpha2/gateway_types.go
youngnick: Thanks for this one too, updated.

apis/v1alpha2/httproute_types.go
```diff
@@ -70,20 +70,33 @@ type HTTPRouteSpec struct {
...
-	//    the HTTP Host header is equal to Host.
-	// 2. If Host is a wildcard, then the request matches this rule if
+	// 1. If Hostname is precise, the request matches this rule if
+	//    the HTTP Host header is equal to the Hostname.
+	// 2. If Hostname is a wildcard, then the request matches this rule if
 	//    the HTTP Host header is to equal to the suffix
 	//    (removing the first label) of the wildcard rule.
 	//
+	// If a hostname is specified by the Listener that the HTTPRoute is bound
+	// to, at least one hostname specified here must match the Listener specified
```
robscott, line 82: I think there's still a bit of nuance here. The absence of hostnames on Routes or Listeners is interpreted as "*", so this is only the case if hostnames are specified on both the Route and the Listener.

youngnick: Updated again. Hopefully this should make it all explicit. Thanks for careful reading, I think we're getting closer to something that's properly unambiguous.

robscott:
Thanks @youngnick! This LGTM, probably just needs another `make generate` to get tests to pass.

/approve

youngnick:
Dang, missed the last `make generate`, fixed now.

robscott:
Thanks!

/lgtm

PR #675, opened 2021-05-20
Add guidance on plural field names to Implementation Guide

### LINKED ISSUE #355: Rename Matches to Match
I would like to rename field `matches` to `match`.

**Why is this needed**:
It reads better and other examples in the APIs where a singular field name is used for a list.

__Current__
```
kind: HTTPRoute
metadata:
  name: http-app-1
  namespace: default
  labels:
    app: foo
spec:
  hostnames:
  - "foo.com"
  rules:
  - matches:
    - path:
        type: Prefix
        value: /bar
    forwardTo:
    - serviceName: my-service1
  - matches:
    - headers:
        type: Exact
        values:
          x-forwarded-proto: https
          magic: foo
      path:
        type: Prefix
        value: /some/thing
    forwardTo:
    - serviceName: my-service2
```

__Proposed:__
```
kind: HTTPRoute
metadata:
  name: http-app-1
  namespace: default
  labels:
    app: foo
spec:
  hostnames:
  - "foo.com"
  rules:
  - match:
    - path:
        type: Prefix
        value: /bar
    forwardTo:
    - serviceName: my-service1
  - match:
    - headers:
        type: Exact
        values:
          x-forwarded-proto: https
          magic: foo
      path:
        type: Prefix
        value: /some/thing
    forwardTo:
    - serviceName: my-service2
```

youngnick:

**What type of PR is this?**
/kind cleanup
/kind documentation

Fixes #355

```release-note
NONE
```

### DIFF
--- a/site-src/concepts/guidelines.md
+++ b/site-src/concepts/guidelines.md
@@ -104,11 +104,21 @@ mostly but not entirely portable.
 
 ## API Conventions
 
-Service APIs follow Kubernetes API [conventions][1]. These conventions
+Gateway API follows Kubernetes API [conventions][1]. These conventions
 are intended to ease client development and ensure that configuration
 mechanisms can consistently be implemented across a diverse set of use
-cases. One variation from the Kubernetes conventions is for fields with
-default values. These fields are considered optional but are not pointers
-since they will never be empty.
+cases. In addition to the Kubernetes API conventions, Gateway API has the
+following conventions:
+
+### List Names
+
+Another convention this project uses is for plural field names for lists
+in our CRDs. We use the following rules:
+
+- If the field name is a noun, use a plural value.
+- If the field name is a verb, use a singular value.
+
+So for example, in HTTPRoute, `hostnames` uses a plural, but `forwardTo` is singular,
+although they are both lists. 
 
 [1]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md


jpeach:
s/Implmentation/Implementation/ ;)

youngnick:
🤦 thanks

site-src/concepts/guidelines.md
```diff
@@ -111,4 +111,13 @@ cases. One variation from the Kubernetes conventions is for fields with
 default values. These fields are considered optional but are not pointers
 since they will never be empty.
 
+Another convention this project uses is for plural field names for lists
+in our CRDs. We use the following rules:
+
+- If the field name is a noun, use a plural value.
+- If the field name is a verb, use a singular value.
+
+So for example, in HTTPRoute, `hostnames` uses a plural, but `ForwardTo` is singular,
```
robscott, line 120: Nit: capitalization is inconsistent here. I think I'd lean towards switching to `forwardTo` instead of switching to `Hostnames` but don't have a strong preference.

site-src/concepts/guidelines.md
```diff
@@ -111,4 +111,13 @@ cases. One variation from the Kubernetes conventions is for fields with
 default values. These fields are considered optional but are not pointers
 since they will never be empty.
 
+Another convention this project uses is for plural field names for lists
```
robscott, line 114: 🤦  I'd missed updating the paragraph above which is unfortunately quite outdated. It looks like we need to fix the "Service APIs" reference and the variation described in that paragraph is no longer true. I think it might make sense to restructure all of the above to something like this:

```md
Gateway API follows Kubernetes API [conventions][1]. These conventions
are intended to ease client development and ensure that configuration
mechanisms can consistently be implemented across a diverse set of use
cases. In addition to the Kubernetes API conventions, Gateway API has the
following conventions:

### List Names
...
```

robscott (commented):
Thanks for the work on this @youngnick!

site-src/concepts/guidelines.md
youngnick: Hah, I missed that too. I'll update.

site-src/concepts/guidelines.md
```diff
@@ -104,11 +104,26 @@ mostly but not entirely portable.
...
 mechanisms can consistently be implemented across a diverse set of use
-cases. One variation from the Kubernetes conventions is for fields with
-default values. These fields are considered optional but are not pointers
+cases. In addition to the Kubernetes API conventions, Gateway API has the
+following conventions:
+
+### Fields with default values
+
+These fields are considered optional but are not pointers
 since they will never be empty.
```
robscott, line 116: Sorry for the confusion here, I think this particular point is no longer the case. We went back and forth on this before eventually settling on pointers for optional values, even if they did have a default value. For example: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/gateway_types.go#L324-L326

hbagdi: +1

hbagdi:
No additional comments from me.

site-src/concepts/guidelines.md
youngnick: Okay, I'll just remove this whole bit then, no worries.

robscott:
Thanks!

/lgtm

PR #677, opened 2021-05-20
Adds GatewayClass Name Validation

danehans:
/kind feature

```Requires the `gatewayclasses.metadata.name` to be valid as defined by RFC 1123.```

### DIFF
--- a/apis/v1alpha2/validation/validation.go
+++ b/apis/v1alpha2/validation/validation.go
@@ -22,6 +22,8 @@ import (
 
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by danehans** [comment]
@robscott I'm interested in validating the existence of the resource referenced by `gatewayclasses.spec.parametersRef`, but it's my understanding that you prefer not to validate resource references. Please confirm that my understanding is correct.

File: apis/v1alpha2/validation/validation.go [hunk 1]
```diff
@@ -34,6 +36,23 @@ var (
...
+// For additional details of the GatewayClass spec, refer to:
+// https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha2.GatewayClass
+func ValidateGatewayClass(gc *gatewayv1a2.GatewayClass) field.ErrorList {
+	return validateGatewayClassObjMeta(&gc.ObjectMeta, field.NewPath("metadata"))
+}
+
+// validateGatewayClassObjMeta validates whether required fields of metadata are set according
+// to the Gateway API specification.
+func validateGatewayClassObjMeta(meta *metav1.ObjectMeta, path *field.Path) field.ErrorList {
+	return apivalidation.ValidateObjectMeta(meta, false, ValidateGatewayClassName, path)
```
jpeach, line 53: suggest we just inline this call

apis/v1alpha2/validation/validation.go
```diff
@@ -34,6 +36,23 @@ var (
...
+}
```
hbagdi, line 54: Please also add this validation to `ValidateGatewayClassUpdate()` code path.

apis/v1alpha2/validation/validation.go
hbagdi: Is the API-server not already doing this kind of validation for custom resources?

apis/v1alpha2/validation/validation.go
robscott: Is a name change even possible?

apis/v1alpha2/validation/validation.go
robscott: I'm not exactly sure how this works, but it does appear that CRD validation already exists. Simply applying a GatewayClass with an invalid name without this validation gives the following error message:

```
The GatewayClass "aAcme-lb" is invalid: metadata.name: Invalid value: "aAcme-lb": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

The [main validation function](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/validation/validation.go#L47) does not appear to include this but it must happen further up the stack.

robscott:
>I'm interested in validating the existence of the resource referenced by gatewayclasses.spec.parametersRef, but it's my understanding that you prefer not to validate resource references

@danehans Yep, although it would be helpful in many cases, we've almost exclusively relied on status for this in upstream. Otherwise applying a yaml manifest with multiple resources could randomly fail if dependent resources didn't quite exist in time for that validation to run.

apis/v1alpha2/validation/validation.go
hbagdi: Probably not. The main motivation for this was to have most validation logic between create and update as common and only some corner cases are handled specially.

robscott:
@danehans Can we close this? I think the validation is already being handled for us.

danehans:
Closing since this validation is already being handled.

PR #681, opened 2021-05-25
httproute: use lists insteads of maps for HTTPRequestHeaderFilter

### LINKED ISSUE #658: Transition from maps to lists of named subobjects
**What happened**:
Maps were used several places throughout the API. Including, but likely not limited to:

* Query Param Matching (#657)
* Header Matching (#657)
* Header Modification (assignee: @hbagdi, #681)

This is also the case for TLS options, but I think it's likely appropriate in that case.

Lists of named subojects should be used to match the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps).

hbagdi:
**What type of PR is this?**

/kind api-change
/kind cleanup
This matches the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) that prefer lists of named subobjects over maps.

Fixes #658
See #657

```release-note
Headers in Request modifier filter are now represented with lists of named subobjects instead of maps.
```


apis/v1alpha2/httproute_types.go
```diff
@@ -530,6 +530,23 @@ const (
 	HTTPRouteFilterExtensionRef HTTPRouteFilterType = "ExtensionRef"
 )
 
+// HTTPHeader represents an HTTP Header name and value as defined by RFC 7230.
+type HTTPHeader struct {
+	// Name is the name of the HTTP query param to be matched. This must be an
+	// exact string match. (See
```
robscott, line 536: Why is this an exact string match? Below we mention that header names are case insensitive. I think you may have meant to refer to a different RFC: https://datatracker.ietf.org/doc/html/rfc2616#section-4.2.

apis/v1alpha2/httproute_types.go
```diff
@@ -530,6 +530,23 @@ const (
 	HTTPRouteFilterExtensionRef HTTPRouteFilterType = "ExtensionRef"
 )
 
+// HTTPHeader represents an HTTP Header name and value as defined by RFC 7230.
+type HTTPHeader struct {
+	// Name is the name of the HTTP query param to be matched. This must be an
```
jpeach, line 535:
```suggestion
// Name is the name of the HTTP header to be matched. This must be an
```

apis/v1alpha2/httproute_types.go
jpeach: I assume that an exact header match is case-insensitive?

robscott: Exact/Regex is how we match the value, the header name has always been exact (with the exception being case sensitivity).

hbagdi:
@jpeach @robscott I had made a copy pasta error. PTAL.

hbagdi:
One thing to note:
I tried to do struct embedding [here](https://github.com/kubernetes-sigs/gateway-api/blob/53e7802d328a867bd40c370b41a6eeb34cdbf940/apis/v1alpha2/httproute_types.go#L320-L337) but kubebuilder CRD generator didn't like that. I know upstream doesn't recommend embedding unless it absolutely makes sense, which I think is the case here. Should we report this problem to kubebuilder community or is this expected?

apis/v1alpha2/httproute_types.go
```diff
@@ -550,7 +572,7 @@ type HTTPRequestHeaderFilter struct {
 	// Support: Extended
 	//
 	// +optional
-	Set map[string]string `json:"set,omitempty"`
+	Set []HTTPHeader `json:"set,omitempty"`
 
 	// Add adds the given header (name, value) to the request
```
robscott, line 577: Tiny nit: I think the pluralization here and for `Add` should match what we have for `Remove`.
```suggestion
	// Add adds the given header(s) (name, value) to the request
```

robscott:
>I tried to do struct embedding here but kubebuilder CRD generator didn't like that. I know upstream doesn't recommend embedding unless it absolutely makes sense, which I think is the case here. Should we report this problem to kubebuilder community or is this expected?

I don't have a strong opinion on struct embedding here, but don't think it needs to block this PR getting in. What happened to the CRD, was the yaml just inaccurate/unexpected? If so, I think it's worth filing a bug on controller-tools repo with a reference to the [command we're using](https://github.com/kubernetes-sigs/gateway-api/blob/master/hack/update-codegen.sh#L49) to generate CRDs.

robscott:
@hbagdi This looks like it's close, let me know when it's ready for me to take another look.

hbagdi:
@robscott updated. PTAL.

robscott:
Thanks @hbagdi! This LGTM but looks like it needs another `make generate` run.

hbagdi:
Fixed, I blazed by a bit too fast.

robscott:
Thanks!

/lgtm

PR #682, opened 2021-05-25
tlsroute: make TLSRoute structure consistent with HTTPRoute

### LINKED ISSUE #566: TLSRoute SNI match inconsistency
TLS route:
```yaml
spec:
  rules:
  - matches:
    - snis: ["foo.com"]
    forwardTo:
    - serviceName: httpbin
      port: 443
```

HTTP route:
```yaml
spec:
  hostnames: ["foo.com"]
  rules:
  - forwardTo:
    - serviceName: httpbin
      port: 80
```

The inconsistency here feels a bit odd when compared to HTTPRoute.

Would it make sense to make a `hostnames` field in TLSRoute and remove the `snis` field in matches? Then the specs above would be identical (both looking like the "HTTP Route")

hbagdi:
**What type of PR is this?**

/kind cleanup
/kind api-change

This PR makes the structure of TLSRoute similar to HTTPRoute:
- SNIs field has been renamed to Hostnames
- The same field has been pulled up to the Spec struct instead of a match

Fixes #566

```release-note
- Breaking change: the structure of TLSRoute is now consistent with HTTPRoute. 
```

Putting a hold here as there is still work to do. I'd like to get consensus on moving this forward before putting more time in here.
/hold

/cc @howardjohn @bowei @jpeach @robscott


apis/v1alpha2/tlsroute_types.go
```diff
@@ -46,17 +46,44 @@ type TLSRoute struct {
...
+	//    wildcard rule.
+	// 3. If SNIs is unspecified, all requests associated with the gateway TLS
+	//    listener will match. This can be used to define a default backend
+	//    for a TLS listener.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	Hostnames []Hostname `json:"snis,omitempty"`
```
jpeach, line 80: Would you mind reverting the Go field name to match the JSON/YAML field name? Divergence on the names for fields has led me astray many times in the past :)

hbagdi: Updated the other way around since the discussion has to be around renaming this field to `hostnames` as well.

apis/v1alpha2/tlsroute_types.go
```diff
@@ -46,17 +46,44 @@ type TLSRoute struct {
...
+
+	// Hostnames defines a set of SNI names that should match against the
+	// SNI attribute of TLS ClientHello message in TLS handshake.
+	//
+	// SNI can be "precise" which is a domain name without the terminating
+	// dot of a network host (e.g. "foo.example.com") or "wildcard", which is
+	// a domain name prefixed with a single wildcard label (e.g. `*.example.com`).
+	// The wildcard character `*` must appear by itself as the first DNS label
+	// and matches only a single label. You cannot have a wildcard label by
+	// itself (e.g. Host == `*`).
```
howardjohn, line 63: Why not? I think right now in the spec the answer would be "Because you should just use TCPRoute instead in that case", but in the future could we not allow routing on other attributes of the TLS handshake?

I don't feel strongly about this *at all* - our implementation only supports routing by SNI today - just curious.

howardjohn: I see this is also copy+pasted so definitely not blocking this PR at all

howardjohn (approved):

robscott:
@hbagdi Looks like this needs a rebase. I'm not quite sure how I feel about the sni -> hostname rename here, @danehans or @bowei do you have any strong opinions on this?

robscott:
@hbagdi I've thought about this more and agree that this change makes sense. We discussed this briefly at community meeting today and it didn't sound like there was any opposition to this change, so I'm happy to LGTM once it's rebased.

hbagdi:
@jpeach @robscott I've rebased and updated the release notes. PTAL.

robscott:
Thanks!

/lgtm
/approve

PR #683, opened 2021-05-25
httproute: introduce RequestRedirect filter

### LINKED ISSUE #200: Add support for configurable HTTP redirects

On a per route basis, a user can configure HTTP redirects 301/302 (may be 307/308 too?) and the location of where to redirect to. It could be simply a redirect to HTTPS, in which case, redirect with the Location header matching the URL of the request being processed or it could be a redirect to a different location. The second part can get tricky because that can invovle path munging (we could consider that to be an extended/custom API).

**Why is this needed**:

This is a common requirement and most controllers already support this using annotations.

hbagdi:
**What type of PR is this?**

/kind feature

Introduces a new filter for configurable redirects.

Fixes #200 (funny :)

```release-note
HTTP redirects are now supported with the addition of new `RequestRedirect` filter.
```

This draft PR is to get the conversation started on what we would like to see here.
I opted to do a filter to not further add to the surface of the API and a filter seemed like a good place to add this functionality.

One use case that this proposal doesn't solve:
- Gateway owner wants to perform HTTP->HTTPS on all requests. Could this be a property of GatewayClass?

Would love to hear more thoughts here.
/cc @robscott @bowei @jpeach @danehans 
/hold


apis/v1alpha2/httproute_types.go
```diff
@@ -598,6 +614,24 @@ type HTTPRequestHeaderFilter struct {
...
 }
 
+// HTTPRequestRedirect defines configuration for the RequestRedirect filter.
+type HTTPRequestRedirect struct {
+	// AutoHTTPS if set to true redirects any incoming HTTP requests to HTTPS.
+	// The path and querystring of the request MUST be preserved in the Location
+	// header in the response.
+	AutoHTTPS bool `json:"auto_https,omitempty"`
+	// TODO if defaulting is possible here (not sure about defaulting in filter), default this?
+	// What status codes should be allowed? 301,302,307,308?
```
howardjohn, line 624: If we have bounded ones one option we should consider is status code number vs an enum? like MOVED_PERMANENTLY, etc.

apis/v1alpha2/httproute_types.go
```diff
@@ -598,6 +614,24 @@ type HTTPRequestHeaderFilter struct {
 	Remove []string `json:"remove,omitempty"`
 }
 
+// HTTPRequestRedirect defines configuration for the RequestRedirect filter.
+type HTTPRequestRedirect struct {
+	// AutoHTTPS if set to true redirects any incoming HTTP requests to HTTPS.
```
howardjohn, line 619: > Gateway owner wants to perform HTTP->HTTPS on all requests. Could this be a property of GatewayClass?

Would it be possible to define a route for `*` and set auto_https?

apis/v1alpha2/httproute_types.go
```diff
@@ -598,6 +614,24 @@ type HTTPRequestHeaderFilter struct {
...
+	// header in the response.
+	AutoHTTPS bool `json:"auto_https,omitempty"`
+	// TODO if defaulting is possible here (not sure about defaulting in filter), default this?
+	// What status codes should be allowed? 301,302,307,308?
+	// TODO check conformance on this field
+	// +kubebuilder:default=302
+	RedirectStatusCode *int `json:"redirect_status_code,omitempty"`
+	// Location denotes the value of location header.
+	// +optional
+	Location *string `json:"location,omitempty"`
```
howardjohn, line 630: I am a bit concerned with having the raw header here, rather than something a bit more abstract, like
```
    redirect:
      uri: /v1/bookRatings
      authority: newratings.default.svc.cluster.local
```

The reason is that I can no longer make it generic - I cannot just say redirect /v1 to /v2, I need to now also include http vs https, so now the route is aware of the TLS layer which is leaking from Gateway

apis/v1alpha2/httproute_types.go
howardjohn: Even better if our use case also supports ACME where we probably want to redirect eveything *but* /.wellknown/acme

hbagdi: > Even better if our use case also supports ACME where we probably want to redirect eveything but /.wellknown/acme

Solid point! 

> Would it be possible to define a route for * and set auto_https?

Gateway admin can but it has limitations. The most specific route matches a request and if that doesn't have the filter defined, then HTTP requests will be accepted.

apis/v1alpha2/httproute_types.go
jpeach: Does it makes sense to have arbitrary headers here (and do proxies support that)? If so, you can use the same name list structure as the other parts of the API.

apis/v1alpha2/httproute_types.go
hbagdi: My concern with enum is it is one more piece of configuration that the user needs to learn and remember.
We could  prevent errors via validation here if that is the concern.

apis/v1alpha2/httproute_types.go
hbagdi: >  I need to now also include http vs https

IME this is not an issue and users want to be as explicit as possible. But you do make a good point that the user now has to be aware of `authority` as well, which could be abstracted away in Gateway for some use cases. I'll update.

hbagdi:
Outstanding questions:
1. Should we have  the `StatusCode` field? If yes, what should be the conformance level for this? The support seems to be available for in-cluster implementations, do could providers support it?
2. Should path-munging be a feature of this filter? Paths are harder to generalize. Some users want to redirect with path substitution, meaning, `/v1/foo/bar` should be redirected to `/v2/foo/bar`, while in some cases, it is a more targeted redirect where `/foo/bar` has to be redirected to `/foobar`. The substitution part can get complex with a bunch of corner cases. Should we keep that out of scope for now and come back to it later?

hbagdi:
ping @robscott @bowei @jpeach @danehans @Miciah

apis/v1alpha2/httproute_types.go
robscott: Agree that I think numbers will likely be easier to understand here.

apis/v1alpha2/httproute_types.go
robscott: But in that case you could have a listener on 80 that only selected the auto redirect route and a listener on 443 that selected all the other routes? This seems like it could become a common pattern.

robscott:
> 1. Should we have the StatusCode field? If yes, what should be the conformance level for this? The support seems to be available for in-cluster implementations, do could providers support it?

It looks like 301 and 302 are fairly broadly supported, but it trails off quickly after that. (See [GCP](https://cloud.google.com/load-balancing/docs/https/traffic-management#response-codes), [AWS](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#redirect-actions), and [Azure](https://docs.microsoft.com/en-us/azure/application-gateway/redirect-overview) docs). Maybe 301 and 302 values would be considered core and the rest would be extended?

> 2. Should path-munging be a feature of this filter? Paths are harder to generalize. Some users want to redirect with path substitution, meaning, /v1/foo/bar should be redirected to /v2/foo/bar, while in some cases, it is a more targeted redirect where /foo/bar has to be redirected to /foobar. The substitution part can get complex with a bunch of corner cases. Should we keep that out of scope for now and come back to it later?

I think we should consider this out of scope for now.

hbagdi:
> Maybe 301 and 302 values would be considered core and the rest would be extended?

Having a conformance at that level would be very confusing for our users. How about we start with only 301 and 302 and have that included as core conformance?

Also, I'm assuming this filter would be under core conformance,. Thoughts?

robscott:
@hbagdi Agree with both of those - makes sense to start with just 301 and 302 and consider the filter a core feature.

hbagdi:
Updated the PR to take into account (I think) all the discussions we have had so far.

robscott:
Thanks for the work on this @hbagdi! This mostly LGTM, just a couple requests:

 1. Can you add a release note to the PR?
 2. Can you add an example with this PR? (Maybe http -> https redirect)

apis/v1alpha2/httproute_types.go
```diff
@@ -641,6 +657,27 @@ type HTTPRequestHeaderFilter struct {
 	Remove []string `json:"remove,omitempty"`
 }
 
+// HTTPRequestRedirect defines configuration for the RequestRedirect filter.
+type HTTPRequestRedirect struct {
+	// Protocol is the protocol to be used in the value of the `Location`
+	// header in the response.
+	// When empty, the protocol of the request is used.
+	// +kubebuilder:validation:Enum=HTTP;HTTPS
```
jpeach, line 665: Nit: prefer to leave a blank like between the comment and the codegen tags. Just makes is a litle easier to read.

apis/v1alpha2/httproute_types.go
```diff
@@ -641,6 +657,27 @@ type HTTPRequestHeaderFilter struct {
...
+	Protocol *string `json:"protocol,omitempty"`
+	// Hostname is the hostname to be used in the value of the `Location`
+	// header in the response.
+	// When empty, the hostname of the request is used.
+	Hostname *string `json:"hostname,omitempty"`
```
jpeach, line 670: Nit: even for pointer fields, I think it's good to tag with `+optional` just to make the semantics really explicit for the reader.

hbagdi:
@robscott @jpeach Addressed all comments, PTAL.

hbagdi:
I had put in a hold when this was a draft. Removing it now (which will merge this in).
/hold cancel

apis/v1alpha2/httproute_types.go
```diff
@@ -641,6 +657,35 @@ type HTTPRequestHeaderFilter struct {
...
+	// When empty, port (if specified) of the request is used.
+	//
+	// +optional
+	Port *int `json:"port,omitempty"`
+	// StatusCode is the HTTP status code to be used in response.
+	//
+	// +optional
+	// +kubebuilder:default=302
+	// +kubebuilder:validation=301;302
+	StatusCode *int `json:"redirect_status_code,omitempty"`
```
Miciah, line 686: Sorry to bring this up _after_ the merge, but should the json tag be lower camel-case?
```suggestion
	StatusCode *int `json:"statusCode,omitempty"`
```

Issue #578, opened 2021-03-09
HTTPRoute forwardTo ports not required


stevesloka:
**What happened**:

In implementing HTTPRoutes, I had a question on how to handle `forwardTo.Ports` if unspecified. I was unsure of which port to use since that is currently an optional field. 

The docs state: 

> If unspecified, the destination port in the request is used when forwarding to a backendRef or serviceName.

I asked a question on Slack (https://kubernetes.slack.com/archives/CR0H13KGA/p1615242865086900) and found out about this thread in Slack (https://kubernetes.slack.com/archives/CR0H13KGA/p1613674798102500?thread_ts=1613607883.098300&cid=CR0H13KGA) which already was asked. =)

My original question was this:

> The docs say that if not specified, it is to use the port the request came on. But in the event that we have two ports, an insecure & secure, is the assumption that my backend app will listen on both? Or only the https port (assuming I have the 301 redirect in place)?

Chatting with @danehans the thought was to revert the port from being optional back to being required. This would allow someone to specify which port to use when forwarding traffic to an upstream all the time.

To clarify how I understand, by making the port required we can assume the following:

1. If `serviceName` is provided, then `port` is used to forward to `name:port`
2. If `backedRef` is provided (and not `serviceName`), then `port` is used to forward to `backendRef:port`
3. If both `serviceName` + `backendRef` is supplied, then `serviceName:port` takes priority

My only question is that is it assumed the `port` from `forwardTo` would be used in conjunction with a `backendRef` (i.e. example 2 previously), or would that referenced object want to configure those bits in that custom resource? If the backedRef wants to handle all of that itself, then maybe we should leave this alone and just require controllers to validate that `port` is provided when a `serviceName` is used. 

// cc @danehans @robscott @howardjohn

danehans:
Notes from the community meeting:
- We need to decide if `port` should be required or option. As @howardjohn pointed out, how does an implementation know what port to use if unspecified?
- [This example](https://github.com/kubernetes-sigs/gateway-api/blob/master/examples/upstream-tls.yaml) should be updated to remove the service target ports and change the httproute port to 443.
- As @Miciah mentioned, the `port` docs should be updated to state that port should be the `port` of a Service when `serviceName` is specified.

cc: @stevesloka @robscott

hbagdi:
My proposal for path forward on this problem:
- Clarify the optionality of Port: #685.
- Open a new issue to add validation for Port present when ServiceName is in use.

[...]

PR #685, opened 2021-05-26
clarify requirement for port in forwardTo

hbagdi:
**What type of PR is this?**

/kind documentation

Clarify when port is required and when it is optional.

For #578

```release-note
NONE
```

/cc @robscott @stevesloka @danehans

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -685,10 +685,11 @@ type HTTPRouteForwardTo struct {
 
 	// Port specifies the destination port number to use for the
 	// backend referenced by the ServiceName or BackendRef field.
-	// If unspecified, the destination port in the request is used
-	// when forwarding to a backendRef or serviceName.
+	// Port is required when the request is being forwarded to a Kubernetes
+	// Service. Port is optional with BackendRef to enable use cases where
+	// the port should be read from the referrant resource.
 	//
-	// Support: Core
+	// Support: Core for Kubernetes Service
 	//
 	// +optional
 	Port *PortNumber `json:"port,omitempty"`
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -137,10 +137,11 @@ type RouteForwardTo struct {
 
 	// Port specifies the destination port number to use for the
 	// backend referenced by the ServiceName or BackendRef field.
-	// If unspecified, the destination port in the request is used
-	// when forwarding to a backendRef or serviceName.
+	// Port is required when the request is being forwarded to a Kubernetes
+	// Service. Port is optional with BackendRef to enable use cases where
+	// the port should be read from the referrant resource.
 	//
-	// Support: Core
+	// Support: Core for Kubernetes Service
 	//
 	// +optional
 	Port *PortNumber `json:"port,omitempty"`
--- a/config/crd/bases/networking.x-k8s.io_httproutes.yaml
+++ b/config/crd/bases/networking.x-k8s.io_httproutes.yaml
@@ -1486,9 +1486,11 @@ spec:
                           port:
                             description: "Port specifies the destination port number
                               to use for the backend referenced by the ServiceName
-                              or BackendRef field. If unspecified, the destination
-                              port in the request is used when forwarding to a backendRef
-                              or serviceName. \n Support: Core"
+                              or BackendRef field. Port is required when the request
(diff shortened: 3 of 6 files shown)


hbagdi:
Please see https://github.com/kubernetes-sigs/gateway-api/issues/578#issuecomment-848909030 for some additional context.

apis/v1alpha2/httproute_types.go
```diff
@@ -685,10 +685,11 @@ type HTTPRouteForwardTo struct {
 
 	// Port specifies the destination port number to use for the
 	// backend referenced by the ServiceName or BackendRef field.
-	// If unspecified, the destination port in the request is used
-	// when forwarding to a backendRef or serviceName.
+	// Port is required when the request is being forwarded to a Kubernetes
+	// Service. Port is optional with BackendRef to enable use cases where
+	// the port should be read from the referrant resource.
```
robscott, line 690: This seems like something we should enforce with webhook validation. What should we do if BackendRef points to a Service?

robscott (commented):
Thanks for the work on this!

apis/v1alpha2/httproute_types.go
hbagdi: > This seems like something we should enforce with webhook validation.

Absolutely, please see https://github.com/kubernetes-sigs/gateway-api/issues/578#issuecomment-848909030. With this patch and the validation bit, the user experience should be good.

> What should we do if BackendRef points to a Service?

I didn't consider that part because I wrongly assumed that was not allowed. Is there a good reason to allow such a thing in the first place? At this level of detail, I personally prefer having only one way to do something.

robscott: >I didn't consider that part because I wrongly assumed that was not allowed. Is there a good reason to allow such a thing in the first place?

One potential example - say we end up adding the ability to forward traffic to a Service in another namespace. We have 3 options:

1. Add a `backendRef.Namespace` field, this supports everything if Service can be referenced by BackendRef
2. Add a `serviceNamespace` field, this only supports Service
3. Add both a `serviceNamespace` field and `backendRef.Namespace` field

I'm pretty sure we want to avoid 3, and 2 could eventually lead to 3.

>At this level of detail, I personally prefer having only one way to do something.

This may be a good argument against the `serviceName` shortcut altogether. I need to think about it more.

hbagdi: This is devolving into a separate conversation on if we should have `serviceName` shortcut or not.
Let's take a look at the following two examples:

```yaml
forwardTo:
- serviceName: foo
  port: 80
```

```yaml
forwardTo:
- backendRef:
    kind: Service
    name: foo
  port: 80
```

The easy path is shorter but not significantly so. The two additional lines also help with creating awareness around this extension point. Given how many corner cases exist with this special case, I think it would be best if we drop `serviceName` special case from here and rest of the API as well.

@jpeach @bowei @robscott @danehans Would love to hear your take on this.

robscott: Would it be confusing if `backendRef.kind` defaulted to `Service`?  That would make the difference even smaller:

```yaml
forwardTo:
- serviceName: foo
  port: 80
```

```yaml
forwardTo:
- backendRef:
    name: foo
    port: 80
```

Given that we've required `port` to be specified when the backend is a Service, there's no longer a simple `serviceName` only short form anyway.

hbagdi: I'd rather take one extra line of YAML over solving the corner cases the shortcut is introducing.
Would love to hear from others.

youngnick: I agree with @hbagdi on this one. I don't see the `serviceName` shortcut adding enough value to make the effort worthwhile.

mark-church: If we are thinking about having `backendRef` default to Service then why not do the following (unless the earlier comment was a typo):

```yaml
forwardTo:
- backendName: foo
  port: 80
  kind: <defaults to Service>
```

This still provides the shortcut, provides only 1 way of doing things, and is also the minimal number of lines. I hate to be so focused on number of YAML lines, but `forwardTo` is also one of the most heavily used fields in the API so the verbosity will be multiplied.

hbagdi:
@youngnick @robscott @mark-church I've introduced a small GEP to arrive at a decision here: https://github.com/kubernetes-sigs/gateway-api/pull/719.
I'd love to hear your thoughts in that PR.

hbagdi:
/close
See #750

PR #686, opened 2021-05-26
use squash as default merge behavior

hbagdi:
**What type of PR is this?**

This automation makes tide use squash as  the default merge method so that PR authors can make incremental commits through the review process.

None

```release-note
NONE
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,7 @@
+---
+labels: 'tide/merge-method-squash'
+---
+
 <!--  Thanks for sending a pull request! Here are some tips for you:
 
 1. If this is your first time contributing to Gateway API, please read our


robscott:
I'm not sure how I feel about this. I think there are some cases where PRs are large enough in scope where multiple commits could be helpful. On the other hand, this label would likely be more frequently helpful than annoying. I'll defer to someone else to add the final LGTM on this, I don't think I have a strong enough opinion either way here.

hbagdi:
>  I think there are some cases where PRs are large enough in scope where multiple commits could be helpful.

100%. It is less common and so I thought having this label as the default behavior might help.
I don't feel strongly either way.

youngnick:
I think this is great, since most of the time the PR creator will need to squash down to one commit anyway.

bowei:
Does prow and tide support squash now? We had a discussion several years ago and we had to switch to merge from squash.

hbagdi:
> Does prow and tide support squash now? We had a discussion several years ago and we had to switch to merge from squash.

Seems so. https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/config.md

PR #691, opened 2021-06-07
Initial Conformance Tests for Gateway API

### LINKED ISSUE #20: Add Initial Conformance Test Framework
We want to ensure that controllers using these APIs are conformant. Having a good conformance test suite will help establish consistency across community implementations and help developers identify any missing pieces in their implementation. This should work similarly to the existing Kubernetes conformance tests and be accessible with the same workflows.

robscott:
**What type of PR is this?**
/kind feature

This is an alternative to https://github.com/kubernetes-sigs/gateway-api/pull/516. That PR uses standard go tests while this one builds on that and uses godog to be more similar to the ingress-controller-conformance test suite.

Fixes #20

```release-note
NONE
```


conformance/features/httpgateway/httpgateway.go
```diff
@@ -0,0 +1,87 @@
...
+	ctx.Step(`^a new random namespace$`, aNewRandomNamespace)
+	ctx.Step(`^the "([^"]*)" scenario$`, theScenario)
+	ctx.Step(`^Gateway.Status.Conditions "([^"]*)" condition should be set to "([^"]*)" within (\d+) minutes$`, conditionMustBeTrueWithin)
+
+	ctx.BeforeScenario(func(*godog.Scenario) {
+		state = tstate.New()
+	})
+
+	ctx.AfterScenario(func(*messages.Pickle, error) {
+		// delete namespace an all the content
```
howardjohn, line 47: this won't clean up GatewayClass(?)

conformance/features/httpgateway/httpgateway.go
```diff
@@ -0,0 +1,87 @@
...
+
+func theScenario(name string) error {
+	dp := utils.DynamicParams{
+		Path:      name + ".yaml",
+		Namespace: state.Namespace,
+	}
+	return utils.DynamicApply(dp)
+}
+
+func conditionMustBeTrueWithin(condName string, condValue string, minutes int) error {
```
howardjohn, line 70: Can we/do we want to have this expose more info? Like condition not found vs False, etc?

This exact case its probably not that important, just wondering in general

conformance/gatewayclass.yaml
```diff
@@ -0,0 +1,7 @@
+kind: GatewayClass
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: gateway-conformance
+spec:
+  controller: istio.io/gateway-controller
```
howardjohn, line 6: I assume this needs to be configurable?

conformance/features/httpgateway/httpgateway.go
```diff
@@ -0,0 +1,87 @@
...
+	tstate "sigs.k8s.io/gateway-api/conformance/state"
+	"sigs.k8s.io/gateway-api/conformance/utils"
+)
+
+var (
+	state *tstate.Scenario
+)
+
+// InitializeScenario configures the Feature to test
+func InitializeScenario(ctx *godog.ScenarioContext) {
```
howardjohn, line 37:
```
?       sigs.k8s.io/gateway-api/conformance/features/httpgateway        [no test files]
=== RUN   TestSuite
I0607 15:08:10.180706  827919 conformance_test.go:109] testFeature ===> ../scenarios/httpgateway/httpgateway.feature

No scenarios
No steps
6.333µs
--- PASS: TestSuite (0.00s)
```

Not sure what I am doing wrong, but this seems skipped?

conformance/scenarios/httpgateway/httpgateway.yaml
```diff
@@ -0,0 +1,45 @@
+kind: Gateway
```
howardjohn, line 1: Is it require, for conformance, that an implementation works with a minimal Gateway like this?

For Istio we have the GW declare which Service will be backing it. We have a default so it does technically work

conformance/scenarios/httpgateway/httpgateway.yaml
```diff
@@ -0,0 +1,45 @@
...
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: my-gateway
```
howardjohn, line 4: namespace may be important here. For in-cluster implementations this is presumably going to require a deployment/service in the same namespace. Unless their is some operator provisioning these in response to Gateways, then we need some way to provision these prior to running conformance test, which means we need to know the namespace, which I think is random?

conformance/scenarios/httpgateway/httpgateway_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"sigs.k8s.io/gateway-api/apis/v1alpha1"
+	"sigs.k8s.io/gateway-api/conformance/utils"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog"
+)
+
+func TestHttpGateway(t *testing.T) {
+	// Test Setup
```
howardjohn, line 34: This feels like a lot of boilerplate that will be in every tests, perhaps it could be abstracted?

conformance/scenarios/httpgateway/httpgateway_test.go
```diff
@@ -0,0 +1,141 @@
...
+	var err error
+	utils.KubeClient, utils.GWClient, err = utils.LoadClientset()
+	if err != nil {
+		klog.Fatalf("error loading Kubernetes clientsets: %v", err)
+	}
+	err = utils.DynamicApply(utils.DynamicParams{Path: "../../../config/crd/bases"})
```
howardjohn, line 40: applying CRDs seems like it may not be safe. Maybe fine for now and a flag in the future could be useful?

conformance/scenarios/httpgateway/httpgateway_test.go
```diff
@@ -0,0 +1,141 @@
...
+	dp := utils.DynamicParams{
+		Path:      "httpgateway.yaml",
+		Namespace: ns,
+	}
+	err = utils.DynamicApply(dp)
+	if err != nil {
+		t.Fatalf("error installing gateway-api examples: %v", err)
+	}
+
+	t.Run("Gateway.Status.Conditions Scheduled condition should be set to true within 3 minutes", func(t *testing.T) {
```
howardjohn, line 58: nit: 3 minutes in description, but its 1 minute in practice

conformance/features/httpgateway/httpgateway.go
robscott: Yep, with this specific iteration the goal is for the Gateway implementation to preinstall both the controller and associated GatewayClass before running these tests. I don't think we'd want to clean up either of those resources as part of this.

conformance/features/httpgateway/httpgateway.go
robscott: Yeah I experimented with this, it's not quite as straightforward as `t.Logf`, but using `klog` seems acceptable here.

conformance/gatewayclass.yaml
robscott: Yep, added some comments to make this configurable. Will eventually likely change this to a generic name and/or remove it altogether, but for now it's helpful when running tests.

conformance/features/httpgateway/httpgateway.go
robscott: Not doing anything wrong, the latest version I'd pushed didn't have that linked together yet. Should work well now.

conformance/scenarios/httpgateway/httpgateway.yaml
robscott: I think the simplest starting point for conformance is going to be to focus on these very minimal examples. This specific example seems like something we'd want everyone to support, but I haven't thought it through to carefully. If we run into issues here it can help us understand if there's problems with either the API, the implementations, or the tests.

conformance/scenarios/httpgateway/httpgateway.yaml
robscott: Yeah I've been thinking about that. For prior art here, ingress-controller-conformance automatically generated these resources based on the references in Ingress. I think we'll likely want to just embed the Service and Deployment references in the YAML manifests so we can also test for missing/invalid refs. (It may be simpler to just create a different test path/func for that purpose, not quite sure yet).

As far as namespaces, we'll also need some more advanced capabilities here when compared with ingress-controller-conformance. We'll need to be able to generate multiple namespaces and have a way to describe which resources belong in each of those namespaces. Not quite sure what the cleanest way to do that is, but it's something that there's no prior art for due to the mostly single namespace capabilities of ingress.

conformance/scenarios/httpgateway/httpgateway_test.go
robscott: Yeah this was all leftovers from #516, the idea was to refactor that out into some shared code, but I've just pulled this out for now so the two PRs are completely separate.

conformance/scenarios/httpgateway/httpgateway_test.go
robscott: Apparently I'm not doing that in the godog test setup. That feels more appropriate since we're relying on the cluster these tests run in to have a GatewayClass resource and corresponding controller preinstalled.

conformance/scenarios/httpgateway/httpgateway_test.go
robscott: Good catch, was a temporary workaround for the sake of debugging that I forgot to revert.

conformance/scenarios/httpgateway/httpgateway_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	tstate "sigs.k8s.io/gateway-api/conformance/state"
+	"sigs.k8s.io/gateway-api/conformance/utils"
+)
+
+var (
+	state *tstate.Scenario
+)
+
+func TestHTTPGateway(t *testing.T) {
```
robscott, line 35: @bowei @hbagdi Thanks for the feedback on these tests earlier this week! I ended up moving back to an approach that didn't rely on godog, shown here. The structure is highly similar to the features file the first file in this PR. These tests fully pass with Istio, but of course this is a tiny subset of what we'd want to include in a full conformance suite. 

Some of the key advantages:
- Meaningful output as tests are running
- No additional translation/abstraction layer
- Simpler debugging
- Less dependencies
- More flexibility when writing complex tests

Of course we also lose the following:
- Feature files that are easy to read
- Pretty output that comes with godog
- Consistency with ingress-controller-conformance

This still needs work in terms of cleanup and restructuring, but I'm interested in high level feedback on how these tests are actually structured. Does it feel simple enough to write and understand?

robscott:
Meant to mark this as "ready for review" earlier. These are by no means complete, but I don't want to go too far without some consensus on structure. This contains 2 commits, one that shows what this would look like with godog, and a second that transitions away from it. This largely follows the pattern of ingress-controller-conformance, recycling a number of helpers. That means this also shows what it would look like to move that project away from godog if we were interested in that.

robscott:
Adding a hold because although this is ready for high level review, it's definitely not polished enough to actually merge.

/hold

youngnick:
@robscott, I think this PR is probably a little stale, but that we should come back to it soon?

/remove-lifecycle stale

robscott:
Replacement PR in progress that will recycle at least parts of this.

PR #694, opened 2021-06-11
Updating v1alpha2 guidelines docs to match v1alpha1

robscott:
**What type of PR is this?**
/kind documentation

It looks like a recent update to docs only made it to v1alpha1. This fixes that.

```release-note
NONE
```


hbagdi:
Verified the diff against e5dae42 and this checks out.
We will have to pay attention to do the version being updated here. Is there any scope for automation here?
/lgtm

youngnick:
🤦 Thanks @robscott!

PR #697, opened 2021-06-27
Fix doc as user cannot omit port field in Gateway

nak3:
This patch makes a tiny change to remove `ports` that 
current doc says user can omit it from Gateway. 
Port in Gateway is not optional.

/kind cleanup
/kind documentation

```release-note
NONE
```


robscott:
Thanks!

/lgtm
/approve

PR #698, opened 2021-06-28
updated implementations page for GKE gateway controller

mark-church:
/kind documentation

This refreshes the GKE gateway controller section of the implementation page

### DIFF
--- a/site-src/references/implementations.md
+++ b/site-src/references/implementations.md
@@ -63,14 +63,18 @@ Gloo Edge 2.0 is an Istio-native, fully-featured Envoy based API gateway that br
 ### Google Kubernetes Engine
 
 [Google Kubernetes Engine (GKE)][gke] is a managed Kubernetes platform
-offered by Google Cloud.
+offered by Google Cloud. GKE's implementation of the Gateway API is through the 
+[GKE Gateway controller][gke-gateway] (currently in Preview) which provisions 
+Google Cloud Load Balancers for Pods in GKE clusters. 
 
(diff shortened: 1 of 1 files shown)


robscott:
Thanks for this PR! This technically will need to change for v1alpha2 as well, but I think the better way to handle that and other changes is with a symlink for files that are not version specific. So no need to block this PR.

/ok-to-test
/lgtm
/approve

site-src/references/implementations.md
```diff
@@ -65,12 +65,16 @@ Gloo Edge 2.0 is an Istio-native, fully-featured Envoy based API gateway that br
...
+GKE's implementation of the Gateway API is through the [GKE Gateway controller](gke-gateway) 
+(currently in Preview) which provisions [Google Cloud Load Balancers](gclb). 
+Deploy [private or public/internet Gateways](gke-gateway-deploy) or 
+[multi-cluster Gateways](gke-multi-cluster-gateways). 
 
 [gke]:https://cloud.google.com/kubernetes-engine
 [gclb]:https://cloud.google.com/load-balancing
 [gke-gateway]: https://cloud.google.com/kubernetes-engine/docs/concepts/gateway-api
+[gke-gateway-deploy]: https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-gateways
+[gke-multi-cluster-gateway]: https://cloud.google.com/kubernetes-engine/docs/how-to/deploying-multi-cluster-gateways
```
robscott, line 77: Testing out the links on the deploy preview makes it look like all three of these should not have a space after the colon.

PR #670, opened 2021-05-17
hack: install examples recursively

hbagdi:
This was found during manual testing for #657:

```
$ kubectl apply -f examples | wc
     38      76    1989
$ kubectl apply --recursive -f examples | wc
     47      94    2424
```

**What type of PR is this?**

/kind cleanup
**What this PR does / why we need it**:
To fix a bug in our verify script.
**Which issue(s) this PR fixes**:

None

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #700, opened 2021-06-29
recursively install examples

hbagdi:
This was found while working on #683.
This was assumed to be fixed with #670 but as it turns out we have two
kubectl apply in our shell script.

**What type of PR is this?**

/kind cleanup
Fixes `make install ` to install v1alpha2 examples as well.

None

```release-note
NONE
```


robscott:
Thanks!

/lgtm
/approve

PR #705, opened 2021-07-04
Use RouteForwardTo for HTTP routes.

jpeach:
Embed the RouteForwardTo struct in HTTPRoute rather than maintaining a
context-specific copy. Although this reduces code duplication, it means
that balancing weights have the same conformance level for all types of
routes, which may not be desirable.

/kind cleanup
/kind api-change

```release-note
All type of routes now require weighted balancing as a core conformance requirement.
```


jpeach:
I don't specifically recall the discussion about the difference in conformance level for weights across the different route types, and it wasn't raised in #353. We can discuss here if there are any concerns.

hbagdi:
xref https://github.com/kubernetes-sigs/gateway-api/pull/476

jpeach:
> xref #476

Nice! If we don't dedupe, then it will diverge eventually.

jpeach:
Closing based on discussion in #476

PR #706, opened 2021-07-08
Adding symlinks to avoid unnecessary duplication in v1alpha2 docs

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

There are a number of docs files that should have no difference between v1alpha1 and v1alpha2. To simplify updating them, this adds symlinks to a single copy of each file.

```release-note
NONE
```


robscott:
Deploy preview looks promising: https://deploy-preview-706--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/contributing/devguide/.

hbagdi:
Thought: Should we do this the other way around where v1alpha1 points to v1alpha2? At some point, we would like to update v1alpha2 and v1alpha1 will be in a "frozen" state.
I don't feel strongly about this so feel free to reject it.

hbagdi:
> implementations.md

If I were to be pedantic, this would be debatable as different versions would have different implementation progress.
I think this makes sense as is.
Leaving a note here for posterity.

jpeach:
> Thought: Should we do this the other way around where v1alpha1 points to v1alpha2? At some point, we would like to update v1alpha2 and v1alpha1 will be in a "frozen" state.

Yeh I think this is preferable. You always update the latest, and earlier versions can either link to a later version or can have their own snapshot.

robscott:
Thanks for the feedback @jpeach and @hbagdi, agree that the other direction makes more sense for symlinks, updated to match.  

>>implementations.md

> If I were to be pedantic, this would be debatable as different versions would have different implementation progress.
I think this makes sense as is.

I agree that this could go either way, but I think it will be easier to maintain if we have a single page that track which versions each implementation supports instead of trying to update different pages for each version.

PR #707, opened 2021-07-08
Moving v1alpha2 to the gateway.networking.k8s.io API Group

### LINKED ISSUE #690: Change API group from x-k8s.io to k8s.io
After some discussion with @thockin, @liggitt, and an earlier [Slack thread](https://kubernetes.slack.com/archives/C09QYUH5W/p1618570463278100) with @aojea, it sounds like we should be moving to the k8s.io API group as part of the v1alpha2 release.

As I understand it, `x-k8s.io` is intended to represent experimental Kubernetes APIs while `k8s.io` is intended to represent official Kubernetes APIs that are subject to all the standard API reviews as APIs within k/k. Given that, `x-k8s.io` was likely the correct choice for v1alpha1, but no longer correct as we're approaching a release where we want to achieve some form of stability, including:

- conversion support with future API versions
- hopefully no breaking changes in future releases

Unfortunately changing to a new API group will result in the CRDs being distinct. This means that both v1alpha1 and v1alpha2 could be represented as different CRDs (and `kubectl get httproute` could refer to either without fully specifying). Although that's not great, I don't think we want to be an experimental API long term, and this v1alpha2 release feels like an appropriate time to make the transition.

We can't simply go from `networking.x-k8s.io` to `networking.k8s.io` because that API is already being served from k/k. Some alternatives I've thought of:

- routing.k8s.io
- gateway.networking.k8s.io
- routing.networking.k8s.io

Any other ideas would be very welcome. I think this discussion is also highly relevant to network-policy team, so CCing them here as well:

/cc @rikatz @jayunit100

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

Fixes #690

```release-note
The API Group has changed from networking.x-k8s.io to gateway.networking.k8s.io for v1alpha2 and beyond.
```


hbagdi:
Great work!
Can you please update `admission_webhook.yaml` to include the new group? With the current patch, admission webhook won't be triggered.

I can't think of any other thing that might stop working but I expect some amount of whack-a-mole with this one.

robscott:
Thanks for catching that @hbagdi! Should be updated now.

hbagdi:
@robscott There is a merge conflict because the other PR got merged in.
/lgtm

robscott:
@hbagdi thanks for the reviews! Just rebased this after the latest conflict.

hack/update-codegen.sh
```diff
@@ -52,30 +52,42 @@ go run sigs.k8s.io/controller-tools/cmd/controller-gen \
...
-         --listers-package "${OUTPUT_PKG}/listers" \
-         --output-package "${OUTPUT_PKG}/informers" \
-         ${COMMON_FLAGS}
-
-echo "Generating register at ${FQ_APIS}"
-go run k8s.io/code-generator/cmd/register-gen \
-        --input-dirs "${FQ_APIS}" \
-        --output-package "${FQ_APIS}" \
-        ${COMMON_FLAGS}
+# TODO(robscott): Change this once v1alpha2 has received formal API approval.
```
hbagdi, line 55: @robscott Can we please create a Github issue to track this and possibly reference this here in a follow up PR?

hbagdi:
/lgtm

@robscott Merging this in but please create an issue to track formal API approval.

hack/update-codegen.sh
robscott: Good call, thanks for catching that! https://github.com/kubernetes-sigs/gateway-api/issues/716

config/crd/bases/gateway.networking.k8s.io_backendpolicies.yaml
```diff
@@ -0,0 +1,226 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    api-approved.kubernetes.io: unapproved
```
liggitt, line 7: is there an issue tracking getting this approved?

robscott: https://github.com/kubernetes-sigs/gateway-api/issues/716 + https://github.com/kubernetes/enhancements/pull/2830

Issue #709, opened 2021-07-09
GEP: Enable cross namespace references from Routes


robscott:
**What would you like to be added**:
This GEP attempts to tackle both cross namespace forwarding and route inclusion (proposed earlier by @stevesloka). These are closely related concepts that can likely be solved with the ReferencePolicy concept originally proposed by @youngnick.

**Why is this needed**:
Both cross namespace forwarding and route inclusion are common feature requests. ReferencePolicy provides a way to safely accomplish this. This is closely related to #411, #582, and #634. 

Relevant docs include:
- [Gateway API Reference Policy](https://docs.google.com/document/d/18MoabVA-fr5XL9cYdf6cxclqRwFpOvHUXV_UYzSiooY/edit)
- [Selection Policy Proposal](https://docs.google.com/document/d/1S9t4YiDBwe1X7q915zKO0meZ8O_UPa8bzBLWBY8_XdM/edit?usp=sharing)
- [Route Inclusion Proposal](https://docs.google.com/document/d/1-0mgRRAY784OgGQ1_LCOshpLLbeAtIr4eXd0YVYK4RY/edit#heading=h.8cfxzle5tmqb)
- [Cross Namespace Forwarding Proposal](https://docs.google.com/document/d/1_B1G9JcNw3skNYLtdK7lTTzOeyz5w2hpa84cKA_MGKk/edit)

Note: Although the Gateway -> Route relationship is closely related to this discussion, I've decided to tackle that in a later GEP.

youngnick:
I'll take this one.

/assign

Issue #695, opened 2021-06-18
Canary deployments with Gateway API between namespaces


arapulido:
The Gateway API spec seems to expect each team to work on a single namespace, which feels a bit too rigid for certain use cases. 

In my case, I would like to be able to do canary deployments that involve services that are not directly getting external traffic. This is a lot easier to do if you can do a different deployment in a second namespace, something like this:

![image](https://user-images.githubusercontent.com/925437/122544568-3bc7fd00-d02d-11eb-9431-c67970de9713.png)

```yaml
kind: HTTPRoute
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: httproute1
  namespace: ns1
  labels:
    app: ecommerce
spec:
  gateways:
    allow: "All"
  hostnames:
    - "www.example.org"
  rules:
    - matches:
        - path:
            type: Prefix
            value: /
      forwardTo:
        - serviceName: service1
          port: 80
          weight: 50
```

```yaml
kind: HTTPRoute
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: httproute2
  namespace: ns2
  labels:
    app: ecommerce
spec:
  gateways:
    allow: "All"
  hostnames:
    - "www.example.org"
  rules:
    - matches:
        - path:
            type: Prefix
            value: /
      forwardTo:
        - serviceName: service1
          port: 80
          weight: 50
```

This is not currently possible with the current spec and something that will be useful to have (to put an example, the NGINX Ingress controller allows to do this).

Similar, I am wondering if this other way to implement the same thing is allowed by the spec:

![image](https://user-images.githubusercontent.com/925437/122545212-de807b80-d02d-11eb-91f5-b6454bcdec6b.png)

Basically, having a single HTTPRoute object, but that points to an ExternalService, pointing to the service in the second namespace

```yaml
kind: HTTPRoute
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: httproute-external
  namespace: ns1
  labels:
    app: ecommerce
spec:
  gateways:
    allow: "All"
  hostnames:
    - "www.example.org"
  rules:
    - matches:
        - path:
            type: Prefix
            value: /
      forwardTo:
        - serviceName: service1
          port: 80
          weight: 50
        - serviceName: service1-external
          port: 80
          weight: 50
```

With `service1-external` looking like this:

```yaml
kind: Service
apiVersion: v1
metadata:
  name: service1-external
spec:
[...]

Issue #634, opened 2021-04-27
Request Filtering Between Gateways and Namespaced Routes


Freyert:
**What would you like to be added**:
A method for `Gateway` administrators to use routing rules to filter which traffic can be forwarded to downstream routes.

As a `Gateway` administrator managing an HTTP load balancer I would like to delegate particular path prefixes to namespaced `HTTPRoutes` to prevent namespace administrators from writing rules that conflict with the rules from other namespaces.

For example, I should be able to delegate `/cats` to the `cats` namespace and `/dogs` to the `dogs` namespace. Then the administrators for each of those namespaces can safely write any rule they want, but with the expectation that the path
rules are prefixed with `/cats` or `/dogs`. `/cats/food` and `/dogs/food` would never collide  because the `Gateway` administrator had already pre-filtered the traffic.

**Why is this needed**:
Currently the API allows request filtering for `HTTPRoutes` and `TCPRoutes`. `Gateway` administrators can delegate traffic routing down to those routes which should be owned by individual **namespace** administrators (Think Hub and Spoke).

This immediately raises the issue of how can a central administrator prevent multiple namespaces from writing conflicting routing rules. Current documentation states that if namespace `A` and namespace `B` create the same routes the conflict will be resolved via timestamp ([HTTPRouteRule](https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule)).

howardjohn:
I didn't put too much thought into this, but would it makes sense to have something like:
Gateway my-gw in ns `gw`: select routes in same namespace only
HTTPRoute my-route in ns `gw`: match /cats, forwardTo _HTTPRoute_ cats in `cats` namespace, match /dogs, forwardTo _HTTPRoute_ dogs in `dogs` namespace
HTTPRoute dogs in ns `dogs`: ...
HTTPRoute cats in ns `cats`: ...

This introduces a new concept of forwardTo another route. I am pretty sure this was in some of the early designs?

robscott:
Yep this concept was in the early designs and something I wanted to revisit for v1alpha2. It really covers 2 related ideas - Routes forwarding to other namespaces, and Routes forwarding to other Routes. I think both are worth exploration as part of v1alpha2. There's also a bit more context for this in the related Slack [thread](https://kubernetes.slack.com/archives/CR0H13KGA/p1619468613155200).

robscott:
[...]

PR #711, opened 2021-07-10
Adding GEP-709: Cross namespace references from Routes

robscott:
**What type of PR is this?**
/kind gep

This adds GEP #709 as the culmination of discussion around ReferencePolicy, cross namespace forwarding, and route inclusion.

```release-note
NONE
```

/cc @youngnick @stevesloka @hbagdi @jpeach @danehans


robscott:
https://deploy-preview-711--kubernetes-sigs-gateway-api.netlify.app/geps/gep-709/

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,161 @@
...
+* `ReferencePolicy.From` and `ReferencePolicy.To` would require at least 1 entry.
+* `Group`, `Kind`, and `Namespace` would all be required fields in
+  `ReferencePolicy.From`.
+* `Group` and `Kind` would both be required fields in `ReferencePolicy.To`.
+* There would be no way to further refine the `From` fields. By trusting a kind
+  of resource in a namespace, you're effectively trusting anyone in that
+  namespace that can administer that kind of resource. Specifying a name does
+  not provide any real protection.
+* Although we may want to consider adding a `Name` field to `ReferencePolicy.To`
+  in the future, it is not clear that it is necessary at this point.
```
hbagdi, line 88: Can we enumerate which resources are supported in `from.GVK` and those in `to.GVK` in this proposal?

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,161 @@
...
+  to:
+  - group: core
+    kind: Service
+```
+
+### Benefits
+
+* Conceptually similar to NetworkPolicy.
+* A separate resource enables admins to restrict who can allow cross namespace
+  references.
```
hbagdi, line 72: No action required.
Do we need any specific commentary on our Security model that might be useful in future when we look back on this GEP? I don't think so. Food for thought for you and others.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,161 @@
...
+resource that enables app admins to describe where they trust references from.
+
+## Motivation/User Journeys/Background
+
+This GEP enables the following capabilities:
+
+1. Retaining full control of Gateway and Routes in an infra namespace, while
+   targeting apps in different namespaces.
+1. Traffic splitting between Services in different namespaces.
+1. Mesh overrides to target Services in different namespaces.
```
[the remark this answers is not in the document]
robscott: Relates to the other GEP in flight. Essentially the idea that as a consumer in namespace foo, I want to override my requests to service bar in a different namespace and send them to bar-canary. https://deploy-preview-715--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713#policy-attachment-for-mesh

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,161 @@
...
+      backend:
+      - name: bar
+        namespace: bar
+---
+kind: ReferencePolicy
+metadata:
+  name: bar
+  namespace: bar
+spec:
+  from:
```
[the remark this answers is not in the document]
robscott: I think due to the additive nature of this API I'd rather have multiple resources than one large resource, but open to ideas. I'm concerned that a list of lists would become hard to manage.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,161 @@
...
+
+Anytime we allow crossing a namespace boundary, we need to be very cautious. To
+ensure that this feature is safe, we need to enforce a handshake mechanism that
+requires resources in both namespaces to agree to this reference. To accomplish
+that, a new ReferencePolicy resource should be introduced.
+
+![Reference Policy](images/709-referencepolicy.png)
+
+With this model, Routes would be able to directly reference Routes and Services
+in other namespaces. These references would only be considered valid if a
```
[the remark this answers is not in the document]
robscott: Yep, agreed with all of that. Ended up moving route -> route to out of scope section.

site-src/geps/gep-709.md
robscott: I think we need to be clear that this is only for Gateway API implementations, but due to the possibility of custom Route types or any other potential extension points I'm not sure we can have a complete list of allowed resources here.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Namespace string `json:"namespace,omitempty"`
+}
+
+// ReferencePolicyTo describes trusted kinds.
+type ReferencePolicyTo struct {
```
howardjohn, line 138: Having a policy per-service will likely be a bottleneck in some scenarios. For example, if we have an egress proxy, and have an `external-namespace`. I probably just want to say `from: egress-gateway, to: any service in external-namespace`. I am not sure that is re-presentable here, without 1 ReferencePolicy per service?

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+    // +kubebuilder:validation:MaxLength=253
+    Kind string `json:"kind"`
+}
+```
+
+
+### Benefits
+
+* Conceptually similar to NetworkPolicy.
+* A separate resource enables admins to restrict who can allow cross namespace
```
howardjohn, line 161: Can you give an example of when this would actually be useful? I am having a hard time thinking of a scenario where I would want to disallow a namespace from exposing themselves to other namespaces

site-src/geps/gep-709.md
howardjohn: Similarly would be `from: anywhere`

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### Exceptions
+If traffic is originating from the same location as the configured Route (such
+as sidecars in some mesh implementations) implementations may choose to ignore
+ReferencePolicy. This should only be done if:
+* Other mechanisms like NetworkPolicy can be used to effectively limit
+  cross-namespace references.
+* The implementation clearly documents that ReferencePolicy is not honored.
+
+This exception is very unlikely to apply to any ingress implementations of the
```
howardjohn, line 176: Generally agree, but this could be plausible in a single-tenant ingress. For example `ingress` namespace has all Gateways and Routes for the ingress. Some routes reference other namespace. There is no security risk here.

site-src/geps/gep-709.md
robscott: I think this is conceptually similar to organizations who want to restrict the namespaces that can create Services of type LB and use quotas to achieve that. It's possible that you want users to be able to create Services without allowing them to expose them externally.

site-src/geps/gep-709.md
robscott: How is there no security risk in that scenario? Would it be possible the implementation to know that it was always going to be deployed in that way?

robscott: After discussing this a bit more, I think there are some instances where ingress implementations deployed in this way could potentially deserve an exception. Fundamentally the distinction is if the implementation would be subject to some other cross-namespace restrictions such as NetworkPolicy. I'll update this section to be a bit more clear.

site-src/geps/gep-709.md
robscott: Both `to` and `from` structs don't have names. This just represents trust from resources of type a in namespace foo to resources of type b within the same namespace as the ReferencePolicy. So in this case I think the `to: any service in external-namespace` is quite straightforward.

On the other hand `from: anywhere` is intentionally not allowed. I'm not entirely convinced it's a use case we want to encourage, but if it were, I think using a namespace selector instead of a string would be the appropriate way to do it. That does feel like a potential extension point, but I'm pretty hesitant to start with it unless there are compelling use cases.

site-src/geps/gep-709.md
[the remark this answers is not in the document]
youngnick: I think it's better to encourage people to have more, smaller objects rather than less, larger objects as well.

I note that the `from` and `two` stanzas are slices here though, but we don't have any discussion on how they should be combined. I'll add more here on the structs themselves.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+// ReferencePolicySpec identifies a cross namespace relationship that is trusted
+// for Gateway API.
+type ReferencePolicySpec struct {
+    // From describes the trusted namespaces and kinds that can reference the
+    // resources described in "To".
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinItems=1
+    From []ReferencePolicyFrom `json:"from"`
```
youngnick, line 99: Both `from` and `to` are slices, but we don't have anything describing how multiple entries in the slice must be combined.

Seems like it would probably be something like "Each entry in `from` must be considered to be an additional place that references can be valid from, or to put this another way, entries must be combined using OR."

Similarly for the `to` stanza, I think that combining entries with OR is the only thing that makes sense. But regardless, we should have the behavior documented here.

site-src/geps/gep-709.md
youngnick: Yeah, I think one of the key parts about this proposal is that it allows access *from* a Group/Kind and namespace *to* a Group/Kind and namespace (where the `to` namespace is always "the namespace the object is in".) I think that if we want to make it more flexible than "named namespace", it should be a label selector across namespaces only.

As it stands, we can do "Allow from a group of things in a list of namespaces" using these structs, since `from` is a list.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+
+// ReferencePolicyFrom describes trusted namespaces and kinds.
+type ReferencePolicyFrom struct {
+    // Group is the group of the referrent.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Group string `json:"group"`
```
youngnick, line 118: Because all three of these values are required, each `from` reference must fully specify a group, kind, and namespace. This is nice for implementers because it's very straightforward, but will this be flexible enough, or should we consider label selectors here? 

The downside of including label selectors is that you then need two sets of fields that are both optional, so the zero values of the fields and slices of this struct really matter. This way is much simpler.

site-src/geps/gep-709.md
youngnick: The use case for not allowing things to expose themselves to other namespaces is any cluster where a namespace admin for one namespace is not totally trusted. In that case, allowing inbound references is a critical step in privilege escalation (another reason to only allow this interaction in one direction).

For example, I've worked before on CI clusters, where each namespace ran a step in a CI pipeline. Now, CI pipelines are essentaily remote-code-execution-as-a-service, so those workloads are very untrusted. Would someone here be able to create a Service record? Possibly, and if so, it's almost certainly good to ensure that they cannot create cross-namespace references to their namespace.

This is a little contrived, I know, but I think the principle of "relatively untrusted namespace owner" is a pretty sound usecase for this overall.

site-src/geps/gep-709.md
youngnick: If we do update anything, it's probably only to point out something like "the purpose of this is to allow the person who controls the referent object the requirement to accept that reference, explicitly, with the intent of making cross-namespace references more secure by default."

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+* A single ReferencePolicy resource can be used for a namespace in place of
+  separate handshake config on each Service or Route resource.
+
+#### Exceptions
+If traffic is originating from the same location as the configured Route (such
+as sidecars in some mesh implementations) implementations may choose to ignore
+ReferencePolicy. This should only be done if:
+* Other mechanisms like NetworkPolicy can be used to effectively limit
+  cross-namespace references.
+* The implementation clearly documents that ReferencePolicy is not honored.
```
youngnick, line 174: Small change to make this clearer:
```suggestion
ReferencePolicy. This may only be done if:
* Other mechanisms like NetworkPolicy are used to effectively limit
  cross-namespace references.
* The implementation must clearly document that ReferencePolicy is not honored.
```

I think that "may" is okay here because we are following it up with clarifying "must" clauses straight afterwards.

site-src/geps/gep-709.md
youngnick: I suggested a wording update here, I think that if we make that change, it covers this usecase already.

youngnick (commented):
I really like this API, and think it's a great evolution of the earlier work.

site-src/geps/gep-709.md
howardjohn: Hmm, I am not sure I understand. Can you let me know where I am going wrong? This is my understanding:

Lets say we have a namespace `ingress`, which contains a Gateway. We have a namespace `untrusted`, which has a Service.

I think the statement here is we are trying to avoid letting `untrusted` expose their service through `ingress`' LoadBalancer. We are NOT trying to restrict `untrusted` from creating an HTTPRoute though (that is independent of this and driven by Gateway/Route selection agreement); rather, we are trying to restrict `ingress` from creating an HTTPRoute referencing `untrusted`.

So by restricting `untrusted` from creating a ReferencePolicy, we _think_ we have achieved this. In reality, the `ingress` namespace can just point to something in its own namespace that forwards to `untrusted` (there are many ways to do this, for simplicity lets just say they set up a dumb TCP proxy that forwards all requests to `untrusted`). As a result, I would say that the use of restrictive RBAC policy on ReferencePolicy does not achieve the goal of not allowing a Service to expose itself.

youngnick: The point of the ReferencePolicy is to require the owner of the `untrusted` namespace to accept incoming references. If they can't create ReferencePolicy objects, then they can't accept incoming references, and so say, the admin can create a more-open HTTPRoute that may reference a lot of Services, but only allow some to complete the two-way handshake.

I agree that it's a bit convoluted, but the idea here is to put the control over incoming references into the hands of the person who wons the referent. The main advantage of doing this with a separate object is that it allows us to apply this pattern to things whose spec we can't change (like Service), while still keeping that control in the hands of the object's owner. I think that the RBAC advantage of having a separate object is a minor one at best compared to that, but it may prove useful to people who are extremely careful with their RBAC policies.

I probably overemphasised this usecase previously, sorry about that.

howardjohn: > The point of the ReferencePolicy is to require the owner of the untrusted namespace to accept incoming references. If they can't create ReferencePolicy objects, then they can't accept incoming references, and so say, the admin can create a more-open HTTPRoute that may reference a lot of Services, but only allow some to complete the two-way handshake.

I think I understand what you are discussing now, but I don't think its a good idea. The fact its possible to do this with the current design is no problem - obviously we cannot stop users from doing obscure things, but I wouldn't want to recommend this.

Instead of the admin "handshaking" every namespace, but then restricting their ability to handshake back, it seems like they should just not "handshake" in the first place?

howardjohn: Might be getting a bit side tracked here though :slightly_smiling_face:

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
+# GEP-709: Cross Namespace References from Routes
+
+* Issue URL: https://github.com/kubernetes-sigs/gateway-api/issues/709
+* Status: Implementable
+
+## TLDR
+
+This GEP attempts to tackle both cross namespace forwarding and route inclusion.
```
bowei, line 8: It doesn't seem like we are actually defining route inclusion in this GEP.

We should probably say that this is: prepare the API for route inclusion.

bowei: BTW if there are references, it would be good to include to define what route inclusion means.

site-src/geps/gep-709.md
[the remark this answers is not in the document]
bowei: can we add the reference so people can see the link

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+spec:
+  rules:
+  - matches:
+    - path: /bar
+    forwardTo:
+      backend:
+      - name: bar
+        namespace: bar
+---
+kind: ReferencePolicy
```
bowei, line 53: We might want to call this "InboundReferencePolicy" or something make the direction obvious

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+### API
+This proposed API is fairly straightforward, but comes with a couple notable
+decisions:
+
+1. Each ReferencePolicy only supports a single From and To section. Additional
+   trust relationships can be modeled with additional ReferencePolicy resources.
+1. Resource names are intentionally excluded from this policy for simplicity and
+   because they rarely provide any meaningful protection.
```
bowei, line 74: ... as the owner of the namespace can always rename resources/change the content of the resources to match a given policy.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+
+This GEP attempts to tackle both cross namespace forwarding and route inclusion.
+These are closely related concepts that can be solved with a new ReferencePolicy
+resource that enables app admins to describe where they trust references from.
+
+## Motivation/User Journeys/Background
+
+This GEP enables the following capabilities:
+
+1. Retaining full control of Gateway and Routes in an infra namespace, while
```
bowei, line 16:
```
1. Keep same namespace references simple.
```

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+    // Namespace is the namespace of the referrent.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Namespace string `json:"namespace,omitempty"`
+}
+
+// ReferencePolicyTo describes trusted kinds.
```
bowei, line 137: "ReferencePolicyTo describes what Kinds are allowed as targets of the references."

bowei: (Not a trusted kind, it's object that we own anyway)

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+* Conceptually similar to NetworkPolicy.
+* A separate resource enables admins to restrict who can allow cross namespace
+  references.
+* Provides consistent way to control references to any resource from a Route.
+* Can be extended in the future for additional use cases.
+* A single ReferencePolicy resource can be used for a namespace in place of
+  separate handshake config on each Service or Route resource.
+
+#### Exceptions
+If traffic is originating from the same location as the configured Route (such
```
bowei, line 169: locations => "trusted isolation domain"

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
...
+as sidecars in some mesh implementations) implementations may choose to ignore
```
bowei, line 170: use MAY (capital)

site-src/geps/gep-709.md
bowei: MAY, MUST, SHOULD should be capitalized so it's obvious we are using them in LOUD RFC TALK.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,260 @@
+# GEP-709: Cross Namespace References from Routes
+
+* Issue URL: https://github.com/kubernetes-sigs/gateway-api/issues/709
+* Status: Implementable
+
+## TLDR
```
bowei, line 6: Do we say very explicitly what fields and kinds are intended to be covered as part of "core" behavior in this proposal?

Also do we say that implementations may cover more Kinds and fields, but that is Custom behavior and there is no conformance there. (is this a good idea?)

site-src/geps/gep-709.md
robscott: I've been tempted to introduce selectors here but so far have avoided it due to the increased complexity they add. In v1alpha1 I think we have had a few too many selectors and it became difficult to keep track of everything. Direct references can still be confusing, but I think they're generally easier to understand. One of my primary concerns with selectors is that their default is so open. Interpreted literally, an empty selector would accept references from all namespaces. 

So all of that to say that I agree with you. If we really _need_ selectors, we can always consider adding them in the future.

site-src/geps/gep-709.md
youngnick: Yes, that's a great way to do it.

site-src/geps/gep-709.md
robscott: That's a good idea. Any thoughts on this name change @youngnick @hbagdi @jpeach @howardjohn?

site-src/geps/gep-709.md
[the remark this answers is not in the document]
robscott: Good idea, added a link

site-src/geps/gep-709.md
robscott: Added guidance for which resources are considered "core" here.

site-src/geps/gep-709.md
robscott: Completely reworded this section, PTAL.

site-src/geps/gep-709.md
robscott: Good catch, this was not well defined. Added your suggestion verbatim.

site-src/geps/gep-709.md
robscott: Now that the CVE is live, I've reworked this entire section, including "LOUD RFC TALK" :), PTAL.

site-src/geps/gep-709.md
robscott: Unfortunately we only have feature request issues that can be solved with Route inclusion (#634, #695) and a couple design docs. For now I've just clarified this sentence and added a brief description of Route inclusion.

site-src/geps/gep-709.md
robscott: Good call, added this extra context.

site-src/geps/gep-709.md
robscott: Added as part of intro.

site-src/geps/gep-709.md
robscott: Much better, thanks!

site-src/geps/gep-709.md
robscott: Refactored this entire section, PTAL.

site-src/geps/gep-709.md
robscott: Refactored this entire section, PTAL.

site-src/geps/gep-709.md
robscott: I've added this to the proposed spec. I think non-core kinds would fall under extended conformance since the expected behavior is clear, it's just a difference of which types we expect all implementations to support.

robscott:
Thanks for all the great feedback on this! I think I've responded to all the feedback now, let me know if I missed anything. I just pushed a round of updates that should get this closer. It also includes a reference to the relevant CVE now: https://github.com/kubernetes/kubernetes/issues/103675.

site-src/geps/gep-709.md
```diff
@@ -0,0 +1,289 @@
...
+
+1. Retaining full control of Gateway and Routes in an infra namespace, while
+   targeting apps in different namespaces.
+1. Traffic splitting between Services in different namespaces.
+1. Mesh overrides to target Services in different namespaces. (For more info,
+   see GEP [#713](https://github.com/kubernetes-sigs/gateway-api/issues/713))
+
+## ReferencePolicy
+
+Anytime we allow crossing a namespace boundary, we need to be very cautious. To
```
bowei, line 27:
```
[very cautious](cite CVE)
```

site-src/geps/gep-709.md
youngnick: I'm okay with it, although it's a little long. I had discounted it before because I don't see us adding an `OutboundReferencePolicy` at any point, so I didn't think the disambiguation was necessary. But I don't object if others think it makes things clearer, I'm a little deep into this design to be able to have a good idea of how the name looks at first glance.

site-src/geps/gep-709.md
youngnick: Agreed. I think you make a fair point, that it's a pretty unlikely scenario that we already provide betters tools to deal with. But I think that having the ability to RBAC the ReferencePolicy object separately is an advantage, just not that big of one.

youngnick (commented):
/lgtm

I'll leave it up to others to make the call as to if `InboundReferencePolicy` or `ReferencePolicy` is clearer. I don't have a strong enough opinion to have it affect my call. :)

youngnick:
uh, I didn't think my lgtm would be enough to merge this, oops. Sorry.

bowei:
doh - rob you forgot to hold it

robscott:
Yep, that's on me, meant to add a hold on these, will open another one with next round of changes.

youngnick:
I think we were just down to the name of the object anyway, pretty much.

robscott:
Yeah agreed, think this one was pretty close

site-src/geps/gep-709.md
bowei: I think ReferencePolicy is find. (per discussion)

PR #714, opened 2021-07-11
update to our cross-namespace routing guide

mark-church:
This is an update to https://gateway-api.sigs.k8s.io/guides/multiple-ns/

Cross-namespace routing and binding is a very under-documented capability so I wanted to provide a more detailed and realistic example for this guide. This doc was originally written by @danehans.

Here are some of the changes I made:

- Restructured the example/scenario
- Added traffic flow and resource relationship diagram
- Provided more description around why you would want to configure route binding in different ways depending on your circumstances

cc @robscott @youngnick @hbagdi @nicholasjackson

/kind documentation

Preview: https://deploy-preview-714--kubernetes-sigs-gateway-api.netlify.app/guides/multiple-ns/


site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+
+In this example, there are two independent teams, store and site, operating in the same Kubernetes cluster, in the `store` and `site` Namespaces. These are their requirements:
+
+- The site team has two applications, home and login, that are running behind `foo.example.com`. They have decided to deploy independent HTTPRoutes for each application so that the routing can be configured indendently for each app. By having an independent HTTPRoute for each app, it allows them to update the canary weights for the login app without having to touch any routing configuration for home.
```
howardjohn, line 7: Delegating based on domain is also very common (I would say more common, but I have no evidence :slightly_smiling_face: ), which I think deserves a mention. https://github.com/istio/istio/blob/58867c37948598b2a03e23fa451a8211d7dad18b/pilot/pkg/config/kube/gateway/testdata/delegated.yaml#L24 is an example

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+- `httproute/home` which acts as a default routing rule, matching for all traffic to `foo.example.com/*` that is not matched by an existing routing rule and routing it to `service/home`
+- `httproute/login` which routes traffic for `foo.example.com/login` to `service/login-v1` and `service/login-v2` according to the canary traffic weights 
+
+Both of these Routes use the same Gateway binding configuration which specifies `gateway/shared-gateway` in the `infra` Namespace as the only Gateway that these Routes can bind with.
+
+```yaml
+{% include 'cross-namespace-routing/site-route.yaml' %}
+```
+
+After these three Routes are deployed, they will all be bound to `gateway/shared-gateway`. The Gateway merges its bound Routes into a single flat list of routing rules. [Routing precedence](/references/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule) within an HTTPRoute is determined by most specific match and direct conflicts between Routes would result in an error and prevent Routes from binding to that Gateway.
```
howardjohn, line 63: How does the merging work? What if the both claim `prefix: /`? Or even more complex: overlapping regex

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+```yaml
+{% include 'cross-namespace-routing/gateway.yaml' %}
 ```
-{% include 'routes-in-multiple-namespaces.yaml' %}
+
+A couple notes about this Gateway:
+
+- It is matching for the `foo.example.com` domain. This is configured on the Gateway so that each HTTPRoute does not also have to configure hostname matching, since they are all using the same domain.
+- The Gateway is configured for HTTPS and references the `foo-example-com` Secret. This allows the certificate to be managed centrally for all applications which are using this Gateway.
+- `routes.Namespaces.from: All` is the configuration to allow routes from any Namespace to use this Gateway. This impacts how routes are able to bind to this Gateway.
```
howardjohn, line 32: This is fairly insecure... ideally we would have route delegation here but since that is not implemented, should we at least explicitly select `store` and `site` namespaces so `attacker` namespace cannot attach some random route?

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+
+Both of these Routes use the same Gateway binding configuration which specifies `gateway/shared-gateway` in the `infra` Namespace as the only Gateway that these Routes can bind with.
+
+```yaml
+{% include 'cross-namespace-routing/site-route.yaml' %}
+```
+
+After these three Routes are deployed, they will all be bound to `gateway/shared-gateway`. The Gateway merges its bound Routes into a single flat list of routing rules. [Routing precedence](/references/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteRule) within an HTTPRoute is determined by most specific match and direct conflicts between Routes would result in an error and prevent Routes from binding to that Gateway.
+
+Thanks to cross-Namespace routing, the Foobar Corporation can distribute ownership of their infrastructure more evenly, while still retaining centralized control. This gives them the best of both worlds, all delivered through declarative and open source APIs.
```
hbagdi, line 65: nit: I can't comment on the svg so here: Can you remove `*` from the `/` route since that is not canonical of the API definition? I think it would be best to represent the paths form the API as as in the diagram.

site-src/guides/multiple-ns.md
hbagdi: Our [conflict resolution principles](https://gateway-api.sigs.k8s.io/concepts/guidelines/) kick in.

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+The infrastructure team deploys the `shared-gateway` Gateway into the `infra` Namespace. 
+
+```yaml
+{% include 'cross-namespace-routing/gateway.yaml' %}
 ```
-{% include 'routes-in-multiple-namespaces.yaml' %}
+
+A couple notes about this Gateway:
+
+- It is matching for the `foo.example.com` domain. This is configured on the Gateway so that each HTTPRoute does not also have to configure hostname matching, since they are all using the same domain.
```
hbagdi, line 30: some addition: This also allows for using the same HTTPRoute resources across development and production environments where the development environment might be hosted at `foo.dev.corp.example.com`. (dev/prod have separate clusters).

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
+- The Gateway is configured for HTTPS and references the `foo-example-com` Secret. This allows the certificate to be managed centrally for all applications which are using this Gateway.
+- `routes.Namespaces.from: All` is the configuration to allow routes from any Namespace to use this Gateway. This impacts how routes are able to bind to this Gateway.
+
+The store team deploys their route for the `store` Service in the `store` Namespace:
+
+```yaml
+{% include 'cross-namespace-routing/store-route.yaml' %}
+```
+
+This Route has straightforward routng logic as it just matches for `/store` traffic which it sends to the `store` Service. The following snippet of the [`gateways` field](/references/spec/#networking.x-k8s.io/v1alpha1.RouteGateways) controlls which Gateways this Route can bind to:
```
hbagdi, line 40: s/routng/routing

hbagdi: s/controlls/controls

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,65 @@
...
 ```
 
-Please note that this guide illustrates this feature for HTTPRoute resource
-only as an example. The same can be accomplished with other route types as
-well.
+`gateways.allow` can be configured for Gateways in the same Namespace as the Route (the default), all Gateways, or a list of specific Gateways. In this example the store and site teams decide to reference a specific Gateway. This is the least permissive choice which ensures that other Gateways in the cluster (perhaps created in the future at some point) will not bind with these Routes. If cluster administrators have full control over how Gateways are deployed in a cluster then a more permissive binding option could be configured on Routes. The less permissive the Gateway selection is, the less that application owners need to know about which Gateways are deployed. 
+
+The site team now deploys Routes for their applications. They deploy two HTTPRoutes into the `site` Namespace:
+
+- `httproute/home` which acts as a default routing rule, matching for all traffic to `foo.example.com/*` that is not matched by an existing routing rule and routing it to `service/home`
```
hbagdi, line 54: This `type/name` form is new for our docs. I'd recommend avoiding it and using "HTTPRoute `home`" style for consistency purposes.

hbagdi (commented):
Thanks for working on this, Mark!

I want to add some more minor language simplification suggestions but that's after the 80 char length change else the suggested diffs become unreadable.

site-src/guides/multiple-ns.md
howardjohn: Ack, thanks for the link. `More specific matches should be given precedence over less specific ones.` can be ambiguous I think (regex being the biggest issue), but that is outside the scope of this doc. I suppose it may be based on the length of the regex match (which is mostly meaningless, but best we can do).

site-src/guides/multiple-ns.md
bowei: We should say this what the Ingress controllers do as part of the standard install, but point out that it's not very secure.

site-src/guides/multiple-ns.md
bowei: We should clarify regex (I thought we handled that case?)

site-src/guides/multiple-ns.md
mark-church: Separating app components of a public URL I'd say is probably 50/50 by domain/path. I'll make a mention of separating by domain. I think in the future we probably need a separate guide to focus on the actual delegation of parts of the URL in a more secure manner. In this example, I'm assuming all actors have good intent and are somewhat coordinated.

When Route inclusion comes, we will have 2 ways of delegating domain ownership - through Route inclusion and also through Gateway listeners. I think it's a bit early to document good patterns and best practices about these things at this time.

site-src/guides/multiple-ns.md
mark-church: Okay good point, I could use the Namespace label selector on the Gateway. It does make this example more complicated because I have to go through the fact that Namespaces must be labelled. @robscott I think it's probably a good thing that the new Gateway->Route permissions allow it to reference Namespaces directly by name.

site-src/guides/multiple-ns.md
mark-church: Good point, I'll add it in

site-src/guides/multiple-ns.md
hbagdi: Is it statically and correctly possible to validate regex conflicts? I've yet to write a bug-free regex in my career so take that question from a beginners perspective.

History: We have generally stayed away from regexes because of the inherent complexity they bring.

I personally think that using with regex without another distinct metadata (host, another path prefix or header) is not going to work well for any shared infrastructure.

site-src/guides/multiple-ns.md
howardjohn: All namespaces are labeled with `kubernetes.io/metadata.name` in recent Kubernetes versions. I think we should encourage this label, as all other namespace labels are not enforceable - Kubernetes will guarantee `metadata.name` == the name of the namespace.

I think https://github.com/istio/istio/blob/58867c37948598b2a03e23fa451a8211d7dad18b/pilot/pkg/config/kube/gateway/testdata/delegated.yaml#L1 is the best practice, although its for delegating domains not paths (which is more secure)

mark-church: @howardjohn I agree that if dealing with untrusted namespaces, it should be done a little more explicitly. I realized though that for this use-case (single domain, multiple paths in different namespaces), it actually doesn't work because the Namespace label selector is an AND and not a OR (I think). I would need to be able to select multiple Namespaces from a single listener.

I am leaving this as-is until the new route binding proposal is merged in in v1alpha2. Then it will be much easier for a single listener to select multiple Namespaces cc @robscott

site-src/guides/multiple-ns.md
mark-church: any mentions of regex with routing precedence should be mentioned in the conflict resolution guide itself rather than this one so im closing this comment

site-src/guides/multiple-ns.md
howardjohn:
```
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - ns1
      - ns2
```

I think you can do something like that to get OR ^. But fine to do later

site-src/guides/multiple-ns.md
```diff
@@ -1,38 +1,158 @@
...
-      Route.
-    - `http-app-2` HTTPRoute has `spec.gateways.allow` set to `FromList` and
-      contains a reference to the `multi-ns-gateway` in `default` namespace.
-      This means that only the specified Gateway resource can bind to this
-      route.  Additional Gateways may be added to this list to allow them to
-      bind to this route.
+# Cross-Namespace routing
 
+The Gateway API has core support for cross Namespace routing. This is useful
+when more than one user or team is sharing underlying networking infrastructure,
```
jpeach, line 4:
```suggestion
when more than one user or team is sharing the underlying networking infrastructure,
```

youngnick (commented):
This looks great for v1alpha1, but will need a complete rewrite for v1alpha2, now that we have ReferencePolicy. I suggest we merge this for now, and rewrite as part of v1alpha2.

mark-church:
@youngnick yup that's what I was planning on doing. The re-write won't be that complex actually. With https://deploy-preview-725--kubernetes-sigs-gateway-api.netlify.app/geps/gep-724/ I believe that cross-namespace Route-Gateway binding will not even require a ReferencePolicy. Is that correct @robscott?

hbagdi:
@mark-church This is really close. Let me know if you need a hand to get this over the line.

mark-church:
@robscott and @hbagdi I have tried a few different things but I can't figure out why these examples are failing the tests. I have successfully run these YAMLs with the GKE Gateway controller (with the gatewayClassName changed) and they apply to the cluster and run successfully. Does it have something to do with the fact that these resources use non-default Namespaces?

```
Verifying hack/../hack/verify-boilerplate.sh
[0;32mSUCCESS[0m
Verifying hack/../hack/verify-codegen.sh
[0;32mSUCCESS[0m
Verifying hack/../hack/verify-crds.sh
[0;32mSUCCESS[0m
Verifying hack/../hack/verify-examples-kind.sh
[0;31mFAILED[0m
Verifying hack/../hack/verify-golint.sh
[0;32mSUCCESS[0m
make: *** [Makefile:69: verify] Error 1
```

bowei:
@robscott I've had the presubs fail for me recently -- haven't had time to investigate. I'm assuming it is passing for some people, so it looks like the builds are not hermetic.

robscott:
@mark-church this is the error I'm getting when I try to run `kubectl apply --recursive -f examples`. It looks like you already have the namespaces configured in a separate file, that file just isn't first alphabetically. I'd recommend using a `0-namespaces.yaml` name to ensure it runs first. Of course once it runs once, all further installs will work unless you run on a new cluster or delete previous examples. 

Since it's possible that someone will try to uninstall examples with `kubectl delete -f examples`, I think it would also be useful to have more gateway specific namespace names so users are less likely to delete namespaces they're using for real. For example if someone actually had an `infra` namespace, it would be quite painful if uninstalling our examples deleted it.

robscott:
@bowei I have noticed a small number of flakes relating to kind failing to spin up, but 99% of my failures have been related to mixing up some kind of yaml example. I think we do need to find a way to provide better test output here. Unfortunately when we were outputting everything it became very difficult to tell where the actual test failure was. Need to figure out how to get prow to highlight a specific failure.

robscott:
Created https://github.com/kubernetes-sigs/gateway-api/issues/785 to track this.

site-src/v1alpha1/guides/multiple-ns.md
```diff
@@ -1,38 +1,158 @@
...
-    - `http-app-2` HTTPRoute has `spec.gateways.allow` set to `FromList` and
-      contains a reference to the `multi-ns-gateway` in `default` namespace.
-      This means that only the specified Gateway resource can bind to this
-      route.  Additional Gateways may be added to this list to allow them to
-      bind to this route.
+# Cross-Namespace routing
 
+The Gateway API has core support for cross Namespace routing. This is useful
+when more than one user or team is sharing the underlying networking infrastructure,
+yet control and configuration must be segmented to minimize security and fault
```
robscott, line 5: Nit: "minimize security" feels a bit strange here, would "minimize failure domains" or "minimize blast radius" be sufficient here?

site-src/v1alpha1/guides/multiple-ns.md
```diff
@@ -1,38 +1,158 @@
...
+is explored in this guide. The guide shows how two independent teams can safely
+share the same Gateway from different Namespaces.
+
+In this guide there are two independent teams, _store_ and _site_, operating
+in the same Kubernetes cluster in the `store-ns` and `site-ns` Namespaces. These are
+their requirements:
+
+- The site team has two applications, _home_ and _login_, that are running 
+behind `foo.example.com`. They want to isolate access and configuration across
+their apps as much as possible to minimize security and failure domains.
```
robscott, line 21: Same comment as above.

robscott (commented):
This is a great improvement, thanks @mark-church! Just found a couple non-blocking nits.

/approve

robscott:
Thanks!

/lgtm

Issue #713, opened 2021-07-10
GEP: Pattern for policy attachment


robscott:
**What would you like to be added**:
This GEP will add a formal proposal for attaching policies to Gateway API. Although many of these resources will be implementation specific, it will be helpful to have consistent patterns for accomplishing this.

**Why is this needed**:
Although Gateway API already significantly increases what's possible with portable routing APIs for Kubernetes, there are still some implementation specific concepts that will need to be attached to this API. A consistent pattern will improve the overall portability and UX of this API.

GEP: https://gateway-api.sigs.k8s.io/geps/gep-713/

hbagdi:
The working document for this issue is here: https://docs.google.com/document/d/13fyptUtO9NV_ZAgkoJlfukcBf2PVGhsKWG37yLkppJo

robscott:
I think this can be closed now that it has been implemented and documented.

/close

robscott:
Reopening this to track stability of this feature and eventual graduation to standard channel.

youngnick:
This one still needs further work.

/lifecycle frozen

youngnick:
After some discussion on this, we're going to break up this GEP into at least:
- Problem overview and discussion (GEP-713, this GEP)
- Direct Policy Attachment
- Inherited Policy Attachment

Currently the plan is to include status for each of the types into the relevant GEP, and leave the general status discussion and cookbook in GEP-713.

The reason for this is so that we can (hopefully) graduate these things on separate timeframes, and don't need to have complete agreement about Inherited to be able to graduate Direct Policy Attachment.

shaneutt:
We talked about this in the sync today, @guicassolato has some upcoming PRs and really wants to help out with this:

/assign @guicassolato 

Let us know how we can support you, reach out with any questions!

PR #715, opened 2021-07-12
Adding GEP-713: Policy Attachment

robscott:
**What type of PR is this?**
/kind gep

This adds a GEP for policy attachment (#713).

```release-note
NONE
```

Deploy Preview: https://deploy-preview-715--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/


robscott:
Deploy preview: https://deploy-preview-715--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+refer to the Policy resource that has been applied.
+
+### Hierarchy
+Policy precedence is determined in the same direction a request flows. For
+ingress, that means from Gateway to Backend, and for mesh that means from
+Consumer to Backend. 
+
+| Ingress | Sidecar Consumer |
+|-|-|
+| Gateway Namespace | Consumer Namespace |
```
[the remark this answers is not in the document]
howardjohn: The most common usage for the 6 levels in mesh I think would be:
* Consumer NS: set defaults for my namespace
* Consumer: set defaults for my workload
* Route NS: presumably this is either consumer or backend ns, so I don't think this makes sense at all actually
* Route: overrides for `consumer` or `consumer NS` settings
* Backend NS: defaults for backends in my namespace
* Backend: polite hints to clients to follow these settings, never enforceable.

I do not think namespace level enforcement are common, just defaults, as its not really something you can enforce anyways. A consumer can use whatever timeout they want, regardless of what the API decides.

So the problem here is that I want a route policy to override a consumer, so I make the consumer one `default`, but I don't want the backend's rules to override, so I need to make it `enforce`. As a result we cannot really use this pattern.

I may be stuck in the model that our product currently uses, but I cannot really wrap my head around how this is expected to work

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+```yaml
+kind: RetryPolicy
+apiVersion: networking.acme.io/v1alpha1
+metadata:
+  name: foo
+spec:
+  default:
+    maxRetries: 5
+  targetRef:
+    group: networking.acme.io
```
howardjohn, line 211: way outside of scope, but would be interesting to standardize this

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+}
+```
+
+### Kubectl Plugin
+To help improve UX and standardization, a kubectl plugin will be developed that
+will be capable of describing the computed sum of policy that applies to a given
+resource, including policies applied to parent resources.
+
+Each Policy CRD that wants to be supported by this plugin will need to follow
+the API structure defined above and add a `networking.x-k8s.io/policy: true` to
```
howardjohn, line 299:
```suggestion
the API structure defined above and add a `networking.x-k8s.io/policy: true` label to
```
or annotation. label would be easier to query

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+the CRD.
+
+### Interaction with Custom Filters
+The concept of policy attachment is rather similar to attaching custom filters
+to a Route rule. Although it's possible that both concepts could be supported
+concurrently in the future, it is strongly recommended that implementations not
+recycle the same resource types for both concepts. To help simplify
+configuration, a concept should be configurable as either policy or a route
+filter, not both. This guidance may change in the future as we gain a better
```
howardjohn, line 308: We may need guidance on when each one is applicable

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+    
+    // ClassName is the name of the class this policy should apply to. When
+    // unspecified, the policy will apply to all classes that support it.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    // +optional
+    ClassName string `json:"className,omitempty"`
```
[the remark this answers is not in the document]
robscott: I think attaching policy to GatewayClass would require a cluster-scoped policy resource, otherwise anyone in any namespace could attach policy to the class. This field was intended as a means to limit the scope of what a policy applies to.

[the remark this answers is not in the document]
howardjohn: That's fair.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+}
+```
+
+### Status
+A new `Policies` field will be added to status on Gateways and Routes. This will
```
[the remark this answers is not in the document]
robscott: To clarify, this is only intended to include policy that has been directly attached to the resource, not any it might inherit through other means. I think that's where the real scale issues come into play and the proposed kubectl plugin would be helpful. Of course only showing directly attached policies may lead to more confusion than it's worth.

site-src/geps/gep-713.md
robscott: woops, meant to say label here, as you mentioned for the simpler querying.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
howardjohn: Good point, you probably wouldn't actually have thousands then, unless you have extensive usage of sectionName.

Given the missing status on the 4/6 of the attachment points and complexity around multiple controllers, we may consider just having the plugin, at least initially? I cannot really think how it can work with multiple controllers

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+Each policy resource would include 2 structs within the spec. One with enforced
+values and the other with default values. In the following example, the policy
+attached to the Gateway requires cdn to be enabled and provides some default
+configuration for that. The policy attached to the Route changes the value for
+one of those fields (includeQueryString).
+
+```yaml
+kind: GKEServicePolicy # Example of implementation specific policy name
+spec:
+  enforce:
```
davidewatson, line 173: During the 7/12 meeting @youngnick suggested to rename this as `override`, and someone else suggested `overrideChild` or `overrideChildren`.

site-src/geps/gep-713.md
robscott: Yep, agreed. I almost used `gateway.networking.k8s.io` here as the example but wanted to leave that for a later discussion.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+be a list of `PolicyTargetReference` structs with the fields instead used to
+refer to the Policy resource that has been applied.
+
+### Hierarchy
+Policy precedence is determined in the same direction a request flows. For
+ingress, that means from Gateway to Backend, and for mesh that means from
+Consumer to Backend. 
+
+| Ingress | Sidecar Consumer |
+|-|-|
```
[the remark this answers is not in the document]
hbagdi: Or even a cluster-level default without a reference to class (I'm not sure what would be the target ref in that case)?

hbagdi (commented):
Thanks for working on this, Rob.
Huge lift and you have been phenomenally pushing this forward, great work!

I'm still reading through the linked Gdoc and I will have more comments here.

site-src/geps/gep-713.md
mark-church: also +1 to `override`, which I prefer to `overrideChild` and `overrideChildren` because of key verbosity.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+  Backends, and maybe Namespaces
+* Provide a means of attachment that works for both ingress and mesh uses of
+  this API
+* Provide a consistent way that both included and implementation-specific
+  policies can be interpreted
+
+## API
+
+This approach is building on concepts from all of the alternatives discussed
+below. This is very similar to the existing BackendPolicy resource in the API,
```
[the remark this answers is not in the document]
mark-church: This work fully supersedes backendPolicy (in that there's nothing BackendPolicy can do that this policy proposal cannot), so it does feel like we should drop BackendPolicy before anybody wastes their time in trying to implement support for it.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+
+Each Policy CRD that wants to be supported by this plugin will need to follow
+the API structure defined above and add a `networking.x-k8s.io/policy: true` to
+the CRD.
+
+### Interaction with Custom Filters
+The concept of policy attachment is rather similar to attaching custom filters
+to a Route rule. Although it's possible that both concepts could be supported
+concurrently in the future, it is strongly recommended that implementations not
+recycle the same resource types for both concepts. To help simplify
```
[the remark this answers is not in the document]
mark-church: I would suggest making this paragraph more generic:

```
### Guidance for implementors on when and how to use policy attachment
There are multiple methods of custom extension in the Gateway API. Policy attachment and custom Route filters are two of these. Policy attachment is designed to provide arbitrary configuration fields that decorate Gateway API resources. Route filters provide custom request/response filters embedded inside Route resources. Both are extension methods for fields that cannot easily be standardized as core or extended fields of the Gateway API. The following guidance should be considered when introducing a custom field into any Gateway controller implementation:

1) For any given field that a Gateway controller implementation needs, the possibility of using core or extended should always be considered before using custom policy resources. This is encouraged to promote standardization and, over time, to absorb capabilities into the API as 1st class fields, which offer a more streamlined UX than custom policy attachment. 

2) Although it's possible that arbitrary fields could be supported by custom policy, custom route filters, and core/extended fields concurrently, it is strongly recommended that implementations not use multiple mechanisms for representing the same fields. A given field should only be supported through a single extension method. An example of potential conflict is policy precedence and structured hierarchy, which only applies to custom policies. Allowing a field to exist in custom policies and also other areas of the API, which are not part of the structured hierarchy, breaks the precedence model. Note that this guidance may change in the future as we gain a better understanding for extension mechanisms of the Gateway API can interoperate.
```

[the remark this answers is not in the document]
mark-church: Also, it doesn't have to be in this GEP, but in our docs we should probably have a more detailed picture of how core, extended, and custom portions of the API relate to each other. "Custom" is more of a concept at this point than a tangible thing as custom is really two different things - custom policy attachment and route fillters. Both are two very different custom extension methods, but we don't currently make it very clear in our docs that these two things are what comprise our "custom API".

image

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+    ClassName string `json:"className,omitempty"`
+}
+```
+
+### Status
+A new `Policies` field will be added to status on Gateways and Routes. This will
+be a list of `PolicyTargetReference` structs with the fields instead used to
+refer to the Policy resource that has been applied.
+
+### Hierarchy
```
mark-church, line 144: I would recommend inserting a diagram here to demonstrate the policy example in this section. Otherwise it's a bit difficult for a lay person to understand how this all fits together. Also, maybe make this an example of 3 layers of policy precedence. Something similar to this diagram:

image

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+Consumer to Backend. 
+
+| Ingress | Sidecar Consumer |
+|-|-|
+| Gateway Namespace | Consumer Namespace |
+| Gateway | Consumer |
+| Route Namespace | Route Namespace |
+| Route | Route |
+| Backend Namespace | Backend Namespace |
+| Backend | Backend |
```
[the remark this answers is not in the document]
mark-church: Is this the configuration you're thinking of:

```
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 30
  targetRef:
    kind: HTTPRoute
    name: my-route
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 10
  targetRef:
    kind: Service
    name: reports-v1
```

But if `/v1/reports` maps to the `reports-v1` Service that takes longer to respond, then wouldn't the developer just want to increase the timeout for that Service directly then? It seems like in this case there's no need to use the HTTPRoute to customize the timeout for an individual path since the delay is caused by the Service behind that path anyways. Am I thinking about this correctly?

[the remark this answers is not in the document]
howardjohn: For example, if I have /slow and /fast, and want different timeouts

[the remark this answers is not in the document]
mark-church: Oh you want to provide different timeouts to paths which are destined for the same Service? I think you would just need different routes for that since policies can't be applied within a resource. I'm not sure if it feels worth it to allow policies to reference at sections within a resource, at least not initially. The user can always create mirrored routes or services to apply per-routing rule policies.

```
kind: HTTPRoute
metadata:
  name: fast-timeout
spec:
  rules:
  - matches:
    - path:
        type: Prefix
        value: /fast
    forwardTo:
    - serviceName: my-service
      port: 8080
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 10
  targetRef:
    kind: HTTPRoute
    name: fast-timeout
---
kind: HTTPRoute
metadata:
  name: slow-timeout
spec:
  rules:
  - matches:
    - path:
        type: Prefix
        value: /slow
    forwardTo:
    - serviceName: my-service
      port: 8080
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 30
  targetRef:
    kind: HTTPRoute
    name: slow-timeout
```

site-src/geps/gep-713.md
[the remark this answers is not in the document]
howardjohn: I asked the same somewhere, the answer was that it would need to be a cluster scoped policy to reference a cluster scoped resource. Which seems to make sense to me.

For Istio specifically we do have similar cluster-level policies, but we have a special namespace that is highly privileged and considered the "root", which allows us to bypass the cluster scoping. It works well for us, but I am not sure it makes sense as a part of the API?

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Yeah I think this is a good point, I'll incorporate a section around BackendPolicy into this. I think it does make sense to drop it from v1alpha2 and plan on adding a new shared policy resource as we have a better idea of shared concepts that it could contain.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+    // * Gateway: Listener Name
+    // * Route: Rule Name
+    // * Service: Port Name
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    // +optional
+    SectionName string `json:"sectionName,omitempty"`
```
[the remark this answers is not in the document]
robscott: I'd really hate to have multiple levels of hierarchy here, though I admit it is possible. If we really have to, we could consider adding a `SectionType` or `SubsectionName` field.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Since I'm proposing direct + local references from policy to target, the policy resource needs to exist in the same scope as the target. That means that any policy attached to a GatewayClass would need to be a cluster scoped resource as well. So far I've thought that GatewayClass parameters would be the simplest way to express defaults at this level. I'll add some notes to the GEP to summarize this.

site-src/geps/gep-713.md
robscott: This is a great idea, thanks!

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Yep, that's exactly it, I'm going to add a diagram like @mark-church suggested below that should make this a bit easier to understand.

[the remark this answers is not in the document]
robscott: Missed the latest comment from @mark-church. Since we can attach policy to a `sectionName`, we'd be able to attach policy to a specific rule in a route, as long as that rule had a name. So instead we'd end up with this:


```yaml
kind: HTTPRoute
metadata:
  name: timeout-example
spec:
  rules:
  - name: fast 
    matches:
    - path:
        type: Prefix
        value: /fast
    forwardTo:
    - serviceName: my-service
      port: 8080
  - name: slow 
    matches:
    - path:
        type: Prefix
        value: /slow
    forwardTo:
    - serviceName: my-service
      port: 8080
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 10
  targetRef:
    kind: HTTPRoute
    name: timeout-example
    sectionName: fast
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 30
  targetRef:
    kind: HTTPRoute
    name: timeout-example
    sectionName: slow
```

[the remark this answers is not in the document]
hbagdi: The solution above solutions work only if there is no ServicePolicy associated with `my-service`. What if one wishes to override the timeout in one route _and_ have a default at the Service?

Would something like the following be supported?


```yaml
kind: ServicePolicy 
spec:
  enforce: # NOTE THIS
    requestTimeout: 30
  targetRef:
    kind: HTTPRoute
    name: my-slow-route
---
kind: ServicePolicy 
spec:
  default:
    requestTimeout: 10
  targetRef:
    kind: Service
    name: reports-v1
```

Here, the default request-timeout is 10seconds and is configured at the Service-level for all requests, no matter which path the request traverses with one exception: if the request traverses my-slow-route, then the request-timeout will be 30 seconds.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
hbagdi: > That means that any policy attached to a GatewayClass would need to be a cluster scoped resource as well. So far I've thought that GatewayClass parameters would be the simplest way to express defaults at this level. I'll add some notes to the GEP to summarize this.

Parameters ref can point to a namespaced resource?
What if someone decides to have two policy resources: TimeoutPolicy and TimeoutClusterPolicy, where one is namespace scoped and the other is cluster scoped?

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Yep, that's a good example the motivation for the override/enforce field.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+* Route status could get confusing when Routes were referenced as a policy by
+  ServiceBinding
+* Introduces a new mesh specific resource.
+
+### 2. Attaching Policies for Ingress
+An earlier proposal for policy attachment in the Gateway API suggested adding
+policy references to each Resource. This works very naturally for Ingress use
+cases where all requests follow a path through Gateways, Routes, and Backends.
+Adding policy attachment at each level enables different roles to define
+defaults and allow overrides at different levels.
```
hbagdi, line 379: Based on @howardjohn 's comment: https://github.com/kubernetes-sigs/gateway-api/pull/715/files#r668252011

Did we consider having a structure of this sort:
```yaml
spec:
  overridableConfiguration:
  enforcedConfiguration:
```

The `default` word in a precedence order seems confusing. The above proposal is verbose but it reduces the chances of misunderstanding.

hbagdi: This is out of scope for this GEP (as noted already in the doc) so please ignore it.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+  annotations or references from policy to Service(s)
+* No way to attach policy to other resources such as namespace or ServiceImport
+* May be difficult to modify Routes and Services if other components/roles are
+  managing them (eg Knative)
+
+
+## Out of scope
+
+* Define all potential policies that may be attached to resources
+* Design the full structure and configuration of policies
```
hbagdi, line 401: @robscott, have we considered having a GenericPolicy.gateway.networking.k8s.io resource which defines the API and allows for an arbitrary map[string]string?
Pros:
- For some tiny features, implementers can use that
- Implementers have a concrete knowledge of what is it that we are recommending
Cons:
- If this results in a wide-spread adoption, having type-less CRDs will result in more pain elsewhere in the community

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+could be used to refer to that specific section of the resource. It would refer
+to the following concepts on these resources:
+
+* Gateway.Listeners.Name
+* xRoute.Rules.Name
+* Service.Ports.Name
+
+```yaml
+kind: HTTPRoute
+apiVersion: networking.x-k8s.io/v1alpha1
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Policy Attachment for Ingress
+Attaching policy to Gateway resources for ingress use cases is relatively
+straightforward. A policy can reference the resource it wants to apply to.
+Access is granted with RBAC - anyone that has access to create a RetryPolicy in
+a given namespace can attach it to any resource within that namespace.
+
+![Simple Ingress Example](images/713-ingress-simple.png)
+
+Attaching policy to Gateway resources for ingress use cases is relatively
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Added guidance around attaching policy to a GatewayClass resource.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Yeah, I think you're right. I've summarized this and recommended that we may want to consider adding this at a later point but should focus on a kubectl plugin first.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Thanks for the great discussion + feedback here. I've switched around the hierarchy here based on some further discussion and I think it all makes more sense. Most notably, I've differentiated between how hierarchy applies differently to defaults and overrides. Let me know if any parts still don't make sense.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: >Parameters ref can point to a namespaced resource?

That's completely fine because it still requires someone with cluster-scoped write access to GWC to make that reference.

> What if someone decides to have two policy resources: TimeoutPolicy and TimeoutClusterPolicy, where one is namespace scoped and the other is cluster scoped?

I think that's the only way policy attachment to GatewayClass could work. I've added some notes to describe this.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Thanks for the great content! I've added your suggestions to the GEP. Agree that we should include a better description of the distinctions between custom, extended, and core but that it can be added separately.

site-src/geps/gep-713.md
robscott: Added this in the latest version, thanks!

site-src/geps/gep-713.md
robscott: I've tried to clear up the difference between `default` and `override` in terms of precedence in my latest commit. Hopefully that helps here. I'm also open to different name than `default` here but would prefer somewhat shorter names if possible.

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: I think this is much clearer now with the suggestions from @mark-church, feel free to reopen if not.

site-src/geps/gep-713.md
robscott: updated now, thanks!

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,409 @@
...
+    // +kubebuilder:validation:MaxLength=253
+    Kind string `json:"kind"`
+
+    // Name is the name of the target resource.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Name string `json:"name"`
```
howardjohn, line 100: One thing weird about namespace references:
```
kind: ServicePolicy
metadata:
  namespace: foo
spec:
  targetRef:
    kind: Namespace
    name: bar
```

This is obviously not valid. We could make `targetRef` optional and apply to its own Namespace by default if omitted as one way to give users an easy way to avoid this footgun.

One thing that is really nice about this is that you can just have a single YAML file and apply it to N namespaces. Real world example that is common in Istio:
```
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: default
spec:
  mtls:
    mode: STRICT
```

Then can easily enable mtls in a namespace with `kubectl apply -f strict.yaml -n my-namespace`

site-src/geps/gep-713.md
robscott: Yeah this is a really interesting idea. I think it's significantly better than annotations, but I'm concerned that it may pull away some implementations that would otherwise create CRDs (which I think results in the ideal UX). After chatting a bit more about this with you I've added this as an alternative in the GEP. It's really more of a potential variation but I wasn't sure where to put that.

site-src/geps/gep-713.md
robscott: Although that would be a nice shortcut, I'm a bit concerned about how easy it would make it to accidentally apply policy too broadly. This shortcut would be a great way to cut down yaml/templating, but I think making accidental policy applications less likely should be prioritized. I think many would expect a policy with an empty targetRef to apply to nothing, not to the entire namespace it was in. If anyone made this mistake, it could have a pretty large blast radius.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+  this API
+* Provide a consistent way that both included and implementation-specific
+  policies can be interpreted
+
+## Out of scope
+
+* Define all potential policies that may be attached to resources
+* Design the full structure and configuration of policies
+
+## API
```
youngnick, line 32: I think this document needs a section here that describes what a Policy object is. It seems like they have the following properties:
- Can be an object of any group or kind
- Must have a `targetRef` field, as described below.

I actually think it's almost worth formalizing that a Policy object is an object that will decode into a (provided) skeleton struct that only has the `targetRef` field (along with the usual object fields).

site-src/geps/gep-713.md
youngnick: I think that having the default apply to nothing, and having the user need to take definite action to have the Policy apply to larger objects is much, much safer. Obviously the tradeoff is that you'll need more and more specific Policy objects.

site-src/geps/gep-713.md
youngnick: This skeleton should also include the presence of `default` and `override` fields, with an explanation of what they do. I think that providing an expected skeleton will make building new, compliant Policy objects much clearer.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+spec:
+  maxRetries: 5
+  targetRef:
+    name: foo
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    sectionName: bar
+```
+
+This would require adding a `Name` field to Gateway listeners and Route rules:
```
youngnick, line 288: I think this is a good idea regardless.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+  policies that had been directly applied to the resource. This could get
+  confusing as it would not provide any insight into policies attached above or
+  below.
+* Since we only control some of the resources a policy might be attached to,
+  adding policies to status would only be possible on Gateway API resources, not
+  Services or other kinds of backends.
+
+Although these concerns are not unsolvable, they lead to the conclusion that
+a Kubectl plugin should be our primary approach to providing visibility here,
+with a possibility of adding policies to status at a later point.
```
youngnick, line 349: I agree, but I would really like to end up with some sort of status eventually. I think it makes sense to use the API for a bit before deciding on what it looks like, particularly if there's a kubectl plugin to help.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+   Note that this guidance may change in the future as we gain a better
+   understanding for extension mechanisms of the Gateway API can interoperate.
+
+### Advantages
+* Incredibly flexible approach that should work well for both ingress and mesh
+* Conceptually similar to existing ServicePolicy proposal and BackendPolicy
+  pattern
+* Easy to attach policy to resources we don’t control (Service, ServiceImport,
+  etc)
+* Minimal API changes required (just expansion of status)
```
youngnick, line 384: I guess given the status section above, status expansion is not required any more?

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+* Simplifies packaging an application for deployment as policy references do not
+  need to be part of the templating
+
+### Disadvantages
+* May be difficult to understand which policies apply to a request
+
+## Removing BackendPolicy
```
youngnick, line 391: Conceivably, BackendPolicy could end up as the example Policy object, provided under the project's api group?

Probably outside the scope of this design though.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+## Out of scope
+
+* Define all potential policies that may be attached to resources
+* Design the full structure and configuration of policies
+
+## API
+
+This approach is building on concepts from all of the alternatives discussed
+below. This is very similar to the existing BackendPolicy resource in the API,
+but also borrows some concepts from the ServicePolicy proposal.
```
youngnick, line 36: We should have a link to this doc or something so that people can pick up the extra context if they want to.

youngnick (commented):
I think this proposal is excellent, and a great distillation of lots of conversation.

However, I think that we need to make sure that the document is accessible for people without much context, and we need to add a bit more explanation early on to get there. Otherwise, we're going to spend a lot of time giving context to API reviewers that already lack bandwidth.

site-src/geps/gep-713.md
howardjohn: My personal experience with k8s RBAC is this namespace-in-spec requirement makes things really annoying. Almost all Kubernetes types can elide the namespace, which makes re-usable configs, helm charts, and docs much easier.

I see the point about the policy broadness, and would be fine with making empty targetRef apply to nothing (or probably invalid if it does nothing?), but I really want to avoid having the namespace in the spec if possible. If you got through the top helm charts, almost half of them break if you do `helm template CHART | kubectl apply -f my-namespace` because of RBAC namespace references.

One alternative is an alias for "same namespace as policy is in". Like:
```
kind: ServicePolicy
metadata:
  namespace: foo
spec:
  targetRef:
    kind: Namespace
    name: .
```
(or some better syntax)

shaneutt (commented):
As an implementer this seems powerful, extensible, and practical. I like it a lot. I have a couple of small comments, and then I may have some confusion about how policy hierarchy is intended to work which I've put up for discussion. Thank you :bow:

site-src/geps/gep-713.md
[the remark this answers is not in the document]
ccfishk: Does the "SectionName" hierarchy provides scalability like  
"Namespace.SectionName1.SectionName2.svc...."

or just one more level "Namespace.SectionName.svc..."

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+default values while an override value attached will have the highest precedence
+overall.
+
+![Ingress and Sidecar Hierarchy](images/713-hierarchy.png)
+
+To illustrate this, consider 3 resources with the following hierarchy:
+A > B > C. When attaching the concept of defaults and overrides to that, the
+hierarchy would be expanded to this:
+
+A override > B override > C override > C default > B default > A default.
```
ccfishk, line 158: Do I miss anything if A > B > C, why   C default > B default > A default ?

site-src/geps/gep-713.md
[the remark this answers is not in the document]
mark-church: Question - do we really need sectionName for v1alpha2? Are there any core/MVP use-cases that need it right now? It feels like we're a little less sure how this should work and because it would be a pretty simple additive change, we could push this off for later to first get requests for it from users to understand the use-cases before we decide to design it.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+### Policy TargetRef API
+
+Each Policy resource should include a single `targetRef` field. It must not
+target more than one resource at a time, but it can be used target larger
+resources such as Gateways or Namespaces that may apply to multiple child
+resources. This field should have the following structure:
+
+```go
+// PolicyTargetReference identifies an API object to apply policy to.
+type PolicyTargetReference struct {
```
hbagdi, line 82: Question from present me and potentially future us:
Why not label selectors here as well? I'm not advocating for them at all, rather the opposite. Will documenting them here help?
Take this or leave it, not attached to this comment at all.

site-src/geps/gep-713.md
hbagdi: +1 to this one.

site-src/geps/gep-713.md
robscott: Yeah I need to explain this better. The rationale is that overrides are usually top down whereas defaults are usually bottom up. IE as a Gateway admin I want to have final say on requirements in my cluster, but defaults should be overridable by app owners or others down the stack. That means that an override attached to the top of the stack should have the highest precedence and a default attached at the same location should have the lowest. I don't know, does that make any sense?

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+    kind: HTTPRoute
+    name: example
+```
+
+In this final example, we can see how the override attached to the Gateway has
+precedence over the default drainTimeout value attached to the Route. At the
+same time, we can see that the default connectionTimeout attached to the Route
+has precedence over the default attached to the Gateway.
+
+![Hierarchical Policy Example](images/713-policy-hierarchy.png)
```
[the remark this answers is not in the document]
robscott: Yeah you're completely right, I need to add some conflict resolution guidance here. A validating webhook that prevented conflicts entirely does seem like the best UX.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+
+## Goals
+
+* Establish a pattern for policy attachment which will be used for any policies
+  included in the Gateway API spec
+* Establish a pattern for policy attachment which should be used for any
+  implementation specific policies used with Gateway API resources
+* Provide a way to distinguish between required and default values
+* Enable policy attachment at all relevant scopes, including Gateways, Routes,
+  Backends, and maybe Namespaces
```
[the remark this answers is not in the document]
robscott: Great idea, I like this change.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
shaneutt: :vulcan_salute: ok sounds good, for what it's worth my personal preference is the webhook approach: just don't allow colliding priorities.

site-src/geps/gep-713.md
robscott: Yeah I agree with that. I think we could/should try to find enough portable concepts that could fit in a shared policy resource, but don't think that BackendPolicy provides any value as it exists today. I think it was mostly meant to establish a pattern for policy attachment, but the actual config included is so limited that I don't think it would be broadly used.

site-src/geps/gep-713.md
robscott: Great idea, added a sample policy resource to the GEP.

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Good idea, added a link

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+
+* Issue URL: https://github.com/kubernetes-sigs/gateway-api/issues/713
+* Status: Implementable
+
+## TLDR
+
+This GEP aims to standardize policy attachment to resources associated with
+Gateway API. This will be important for providing a consistent experience across
+implementations of the API, even for configuration details that may not be fully
+portable.
```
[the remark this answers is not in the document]
robscott: Much better, thanks!

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+
+## Goals
+
+* Establish a pattern for policy attachment which will be used for any policies
+  included in the Gateway API spec
+* Establish a pattern for policy attachment which should be used for any
+  implementation specific policies used with Gateway API resources
+* Provide a way to distinguish between required and default values
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+* Enable policy attachment at all relevant scopes, including Gateways, Routes,
+  Backends, and maybe Namespaces
+* Provide a means of attachment that works for both ingress and mesh uses of
+  this API
+* Provide a consistent way that both included and implementation-specific
+  policies can be interpreted
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,496 @@
...
+workload to a backend in another namespace. A route can be used to intercept
+these requests and split them between different backends (foo-a and foo-b in
+this case).
+
+![Complex Mesh Example](images/713-mesh-complex.png)
+
+### Policy TargetRef API
+
+Each Policy resource should include a single `targetRef` field. It must not
+target more than one resource at a time, but it can be used target larger
```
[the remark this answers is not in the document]
robscott: Good call, added some rationale here.

site-src/geps/gep-713.md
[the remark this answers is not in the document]
robscott: Added a bunch of conflict resolution guidance here + a recommendation that a validating webhook could provide a better experience. Thanks for catching this!

site-src/geps/gep-713.md
robscott: Added what I hope is a clearer explanation here, let me know if I missed anything or if anything doesn't make sense.

site-src/geps/gep-713.md
robscott: Yeah this is a good point. I added an explanation for the rationale to start with a single targetRef per policy instead of multiple targetRefs or a selector. I don't think it needs to be this way forever, but I'd rather start with a simpler API that can be expanded than introduce too much complexity up front.

site-src/geps/gep-713.md
robscott: Agreed, I think this will end up being useful for more things in the future.

site-src/geps/gep-713.md
robscott: Yep, completely agree on this one.

robscott:
Thanks to everyone for the great feedback on this! I just pushed a pretty large update to this that I think addresses all the feedback so far. Some highlights:

* A new sample Policy resource that can be used as a template
* Improved guidance around hierarchy
* New section covering conflict resolution

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+### Hierarchy
+Each policy may include default or override values. Default values are given
+precedence from the bottom up, while override values are top down. That means
+that a default attached to a Backend will have the highest precedence among
+default values while an override value attached will have the highest precedence
+overall.
```
banks, line 226: I couldn't quite parse this sentence, was it intended to read as below?
```suggestion
default values while an override value attached to a `GatewayClass` will have the highest precedence
overall.
```

banks (commented):
This GEP LGTM. One minor issue inline.

Thanks for your hard work on this!

site-src/geps/gep-713.md
ccfishk: The clarification is fine. Thanks.
Just curious, why not just follow both override and defaults top down to avoid footgun, I mean, easier.

shaneutt (approved):

site-src/geps/gep-713.md
robscott: Yeah I do think there are legitimate arguments for both. It would be simpler if both forms of hierarchy flowed in the same direction. With that said, my thinking here is that most users will want more specific defaults to be given precedence. In most cases, a Gateway will refer to multiple Routes which will refer to multiple Services. In that case, I think it makes sense for defaults applied to the Gateway to be overridden by defaults attached to a specific Route.

Admittedly it is possible for this to work the other way with multiple Gateways referring to the same Route, but I think that's going to be much less common. I do want to get some broader feedback here though.

@hbagdi @jpeach @youngnick @howardjohn What do you think makes the most sense for hierarchy here?

site-src/geps/gep-713.md
robscott: Although I'm a bit hesitant to loosen validation here, I think it might make sense at some point to make name optional when the kind is Namespace. Unfortunately that's impossible to enforce with CRD validation, and since each policy is going to be a different resource, it would be difficult to validate consistently with a webhook.

I'm not sure there's any precedent in k8s for interpreting "." as same namespace, so that may not hold up to Kubernetes API review. It is of course possible for individual implementations to support this as an extended concept though. Since we're trying to avoid breaking changes going post-v1alpha2, I'm biased towards stricter validation up front to retain the option to loosen it in the future if needed. 

Given that this is mostly a UX issue, I don't think we absolutely need to solve this now. In the interim, this is absolutely something that can be solved with templating, whether it's Helm, Kustomize, or something else.

howardjohn: I mostly agree with everything you said. I would add, as a real world example of why this matters: https://github.com/istio/istio/pull/34035.
Istio fails to properly document security best practices due to this exact same thing, ultimately leading a "High" severity finding in NCC audit https://istio.io/latest/blog/2021/ncc-security-assessment/.

site-src/geps/gep-713.md
banks: My 2 cents would be that the proposal makes the most sense to me as it stands. It happens to be exactly the same model I came up with independently for the same problem a few years ago in Consul based on our user requests:

 *  Infrastructure operators want to provide "easy defaults" while empowering service owners to customize where needed without needing central ops intervention (hence defaults gain precedence the more specific their attachment).
 * Infrastructure operators still need to retain ultimate control to enforce regulatory compliance issues, or to mitigate security threats without needing to coordinate with many app owners, hence overrides take precedence the more general their attachment.

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+    Name string `json:"name"`
+
+    // SectionName is the name of a section within the target resource. When
+    // unspecified, this targets the entire resource. In the following
+    // resources, SectionName is interpreted as the following:
+    // * Gateway: Listener Name
+    // * Route: Rule Name
+    // * Service: Port Name
+    //
+    // Support: Core
```
robscott, line 134: Needs to be extended

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+
+    // Name is the name of the target resource.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Name string `json:"name"`
+
+    // SectionName is the name of a section within the target resource. When
```
robscott, line 127: Not necessarily unique, need to define what happens if it isn't

site-src/geps/gep-713.md
robscott: Remove from field level

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+this case).
+
+![Complex Mesh Example](images/713-mesh-complex.png)
+
+### Policy TargetRef API
+
+Each Policy resource should include a single `targetRef` field. It must not
+target more than one resource at a time, but it can be used to target larger
+resources such as Gateways or Namespaces that may apply to multiple child
+resources.
```
youngnick, line 88:
```suggestion
Each Policy resource MUST include a single `targetRef` field. It MUST NOT
target more than one resource at a time, but it MAY be used to target larger
resources such as Gateways or Namespaces that may apply to multiple child
resources.
```

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+
+As with most APIs, there are countless ways we could choose to expand this in
+the future. This includes supporting multiple targetRefs and/or label selectors.
+Although this would enable compelling functionality, it would increase the
+complexity of an already complex API and potentially result in more conflicts
+between policies. Although we may choose to expand the targeting capabilities
+in the future, at this point it is strongly preferred to start with a simpler
+pattern that still leaves room for future expansion.
+
+The `targetRef` field should have the following structure:
```
youngnick, line 98:
```suggestion
The `targetRef` field MUST have the following structure:
```

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+    // +optional
+    // +listType=map
+    // +listMapKey=type
+    // +kubebuilder:validation:MaxItems=8
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+### Hierarchy
+Each policy may include default or override values. Default values are given
```
youngnick, line 222:
```suggestion
Each policy MAY include default or override values. Default values are given
```

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+policy, they can attach it as an override to a Gateway. That would have
+precedence over Routes and Services below it. On the other hand, an app owner
+may want to set a default timeout for their Service. That would have precedence
+over defaults attached at higher levels such as Route or Gateway.
+
+Each policy resource would include 2 structs within the spec. One with override
+values and the other with default values. In the following example, the policy
+attached to the Gateway requires cdn to be enabled and provides some default
+configuration for that. The policy attached to the Route changes the value for
+one of those fields (includeQueryString).
```
youngnick, line 248: Suggestion here, but something like:
```suggestion
If using defaults and overrides, each policy resource MUST include 2 structs
within the spec. One with override values and the other with default values.

In the following example, the policy attached to the Gateway requires cdn to
be enabled and provides some default configuration for that. The policy attached
to the Route changes the value for one of those fields (includeQueryString).
```

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+precedence over the default drainTimeout value attached to the Route. At the
+same time, we can see that the default connectionTimeout attached to the Route
+has precedence over the default attached to the Gateway.
+
+![Hierarchical Policy Example](images/713-policy-hierarchy.png)
+
+#### Supported Resources
+It is important to note that not every implementation will be able to support
+policy attachment to each resource described in the hierarchy above. When that
+is the case, implementations should clearly document which resources a policy
```
youngnick, line 287:
```suggestion
is the case, implementations MUST clearly document which resources a policy
```

site-src/geps/gep-713.md
```diff
@@ -0,0 +1,598 @@
...
+  should be given precedence for the specified section over a targetRef that
+  does not include one.
+* The oldest Policy based on creation timestamp. For example, a Policy with a
+  creation timestamp of "2021-07-15 01:02:03" is given precedence over a Policy
+  with a creation timestamp of "2021-07-15 01:02:04".
+* The Policy appearing first in alphabetical order by "<namespace>/<name>". For
+  example, foo/bar is given precedence over foo/baz.
+
+When a conflict occurs, implementations MUST set a "Conflict" condition in the
+`status.conditions` list for all affected policy resources.
```
youngnick, line 415: We haven't really done a status design for Policy objects yet.

We really should be providing some basic Conditions and Reasons as part of this design that are expected for all policies.

I suggest:
Condition: `type: Attached` - positive polarity (true is good). Indicates that the policy successfully attached to the `targetRef`.  If that fails, we provide Reasons that summarise common errors (of which `Conflict` would be one, as would things like `NotFound` or similar). Exact reasons to be added should be part of the implementation, but this main flow is important to get right in the design.

(I chose "Attached" although we've used "Admitted" in other places for a similar Condition, I think that it's clearer in this case because that's what the Policy is doing.

That means there should be some documentation in the `Status` section, or earlier in the structs that says something like:
```
#### Policy Conditions 

Implementations implementing policy MUST ensure an `Attached` condition is
present in the `status.conditions` stanza on any objects that reference an object
under their control (that is, on Policy objects that reference Gateway API objects
that fall under a GatewayClass they implement.)

The `Attached` Condition MUST be set to `status: true` if and only if the policy object
is successfully attached to the referenced object with no errors. If there are errors,
the `Attached` Condition MUST be set to `status: false`, with a Reason indicating the
reason why the policy could not attach. Sample Reasons are provided in the
Gateway API structs.

When an object falls out of the implementation's scope, it MUST NOT remove the
Condition itself, that will be handled by the next thing to bring the policy into scope.
```

Then, this section can be amended to:
```suggestion
When a conflict occurs, implementations MUST set the "Admitted" condition in the
`status.conditions` list for all affected policy resources to `status: false`, with a Reason of `Conflict`,
and a message that indicates what other policies are conflicting.
```

This design does, however, have the problems mentioned in the `Status` section that, if a Policy object is relevant to more than one implementation, then they will fight over the `Admitted` condition.

After reading this over again, I think we should either do what I have here or leave the status flow completely alone for now.

youngnick (commented):
LGTM aside from one concern about status.

I think we either need to define one Condition, or leave status explicitly undefined.

I lean towards defining an `Attached` Condition for now, and worrying about what happens when multiple implementations use the same Policy later (if that indeed ever becomes a problem).

I also suggested a few make-the-language-shouty-RFC changes.

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Updated, thanks!

site-src/geps/gep-713.md
robscott: Thanks for catching this! You're completely right that we end up with all the same status problems we had with BackendPolicy. One solution might be to have a controller name in the condition struct, but even that's insufficient as a given controller may be implementing multiple Gateways or even classes that target the same backend. I think this is solvable, but not something we need to solve right now. 

I've pulled out any requirements to set conditions in status, and have left the only reference to status as part of the skeleton resource which just includes a generic list of conditions in status. If that makes sense, I'll plan on creating a tracking issue for this discussion if/when this GEP merges.

robscott:
Thanks for all the great feedback on this GEP! Based on the conversation at today's community meeting, it sounds like we're really close to consensus on this one, and no one has any major hesitations with this merging. Given that, I'm going to remove the hold on this so any LGTM will result in this merging in.

/hold cancel

site-src/geps/gep-713.md
robscott: Updated, thanks!

robscott:
I should also note the major changes from the last commit I pushed:

* SectionName is now described as a possible future extension
* There are no longer any firm guidelines or requirements for status, I'll create a follow up tracking issue for that
* Support level annotations have been removed from each field and instead the entire concept has been listed as "EXTENDED" conformance

site-src/geps/gep-713.md
howardjohn: What about making name optional, to allow:
```
kind: ServicePolicy
metadata:
  namespace: foo
spec:
  targetRef:
    kind: Namespace
```

to refer to same namespace?

robscott: That very quickly opens the door to a way to apply policy to all resources of a certain kind within the local namespace. Although that could also be helpful, it adds additional complexity in terms of hierarchy and conflict resolution. I'm currently biased towards explicit config even if it requires a bit more templating. 

With v1alpha2 and the goal to not make any breaking changes after this release, any optional field is optional forever, we have no path to make it required. On the other hand, we can always go from required -> optional in the future if we're completely sure that's what we'd want.

youngnick:
With those latest changes, I think this is all good.

/lgtm

Amila-Rukshan:
Hi @robscott 
I'm in the requirement of having name property in route rule as described in https://gateway-api.sigs.k8s.io/geps/gep-713/?h=multiple+targetrefs#apply-policies-to-sections-of-a-resource-future-extension section of this GEP. Since this GEP is in experimental stage, can I send a PR for what is described in the future extension section?

Thanks for your time!

robscott:
Hey @Amila-Rukshan, thanks for asking! I think we could probably justify adding this to the existing GEP. Because policy attachment is already complex, I think the most important thing will be to show that adding support for targeting sections of resources is actually widely useful. It may be helpful to create an issue describing this first to try to gauge the level of interest in this.

sunjayBhatia:
@Amila-Rukshan started a similar discussion here if you want to chime in as well: https://github.com/kubernetes-sigs/gateway-api/discussions/1489

PR #702, opened 2021-07-01
Add a virtualenv convenience target for mkdocs.

jpeach:
Add a make target to create the virtualenv with mkdocs installed to make
it easier to remember how to get the right tooling. Add an example to
the contribution docs.

/kind documentation

PR #717, opened 2021-07-13
gitignore: add .venv/

hbagdi:
**What type of PR is this?**

/kind cleanup
#702 results in a `.venv/` directory in the working directory, this patch ignores that directory from version control.

NONE

```release-note
NONE
```


robscott:
Good catch, thanks!

/lgtm

Issue #718, opened 2021-07-13
GEP: consolidate serviceName shortcut in RouteForwardTo


hbagdi:
**What would you like to be added**:

Drop the shortcut for services from BackendRef

**Why is this needed**:

To reduce the number of corner-cases like these: #685 #578

PR #719, opened 2021-07-13
GEP-718: consolidation of ServiceName in RouteForwardTo

hbagdi:
**What type of PR is this?**

/kind gep
This adds a GEP for droping ServiceName field from RouteForwardTo(#718)

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-718.md
+++ b/site-src/geps/gep-718.md
@@ -1,0 +1,338 @@
+# GEP-718: Rework forwardTo segment in routes
+
+Issue: [#718](https://github.com/kubernetes-sigs/gateway-api/issues/718)
+
+Status: Implementable
+
+## Motivation
+
+### Complications due to the `serviceName` shortcut
+
+Within `RouteForwardTo` (and `HTTPRouteForwardTo` by extension) there exists
+two mechanisms to specify the upstream network endpoint where the traffic should
+be forwarded to:
+- `ServiceName`: the name of the Kubernetes Service
+- `BackendRef`: a LocalObjectReference to a resource, this is an extension point in the API
+
+While working on [#685](https://github.com/kubernetes-sigs/gateway-api/pull/685),
+it came to light that:
+1. `BackendRef` itself could be used to point to a Kubernetes Service
+2. With [GEP-709](https://github.com/kubernetes-sigs/gateway-api/pull/711),
+there will be a need to introduce a new `namespace` field which will enable use-cases to forward
+traffic to a services and backends located in different namespaces
+
+While (1) can be fixed with more validation and documentation, it only solves for
+the specific case. (2) requires adding two fields: `serviceNamespace`
+and `BackendRef.Namespace` - something we would like to avoid since the `serviceName`
+field was introduced only as a shortcut and adding more service-specific fields
+defeats the original purpose.
+
+These problems are in addition to the original problem that
+[#685](https://github.com/kubernetes-sigs/gateway-api/pull/685) attempts to solve:
+clarifying port requirements when a port is required or not.
+We have been updating documentation to tackle such corner-cases but that
+continues to result in more and more elaborations. Some excerpts from our
+existing documentation:
+```
+	// ServiceName refers to the name of the Service to mirror matched requests
+	// to. When specified, this takes the place of BackendRef. If both
+	// BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence.
+
+	// BackendRef is a local object reference to mirror matched requests to. If
+	// both BackendRef and ServiceName are specified, ServiceName will be given
+	// precedence.
+
+	// Weight specifies the proportion of HTTP requests forwarded to the backend
+	// referenced by the ServiceName or BackendRef field. This is computed as
+
+	// Port specifies the destination port number to use for the
+	// backend referenced by the ServiceName or BackendRef field.
+	// If unspecified, the destination port in the request is used
+	// when forwarding to a backendRef or serviceName.
+```
+
+### Simplifying `RouteForwardTo`
+
+RouteForwardTo is not the best of the names. Using a noun instead of a verb
+would help with in-person communication and documentation.
+Since we are already considering dropping the special case of `service` as a
+backend, we have an opportunity to further simplify `RouteForwardTo` for better
+UX.
+
+## Proposal
+
+- Remove the `ServiceName` field from `RouteForwardTo`
+- Introduce `Service` as a default for `BackendRef.Kind`
+- Pull fields from `RouteForwardTo.BackendRef` into `RouteForwardTo` itself
+- Rename `RouteForwardTo` to `BackendRef`
+- Rename `Route.Spec.Rules[].ForwardTo[]` to `Route.Spec.Rules[].BackendRefs[]` in
+  all routes
+- Apply the same to HTTP types and filters
+
+## API
+
+The updated `BackendRef` (formerly `RouteForwardTo`) struct will be:
+(HTTP version has been omitted for brevity)
+
+```go
+// BackendRef defines how and where a request should be forwarded from the Gateway.
+type BackendRef struct {
+	// Group is the group of the backend resource.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Group string `json:"group"`
+
+	// Kind is kind of the backend resource.
+	//
+	// Support: Core (Kubernetes Service)
+	// Support: Custom (any other resource)
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:default="service"
+	Kind *string `json:"kind"`
+
+	// Name is the name of the backend resource to forward matched requests to.
+	//
+	// If the referent cannot be found, the rule is not included in the route.
+	// The controller should raise the "ResolvedRefs" condition on the Gateway
+	// with the "DegradedRoutes" reason. The gateway status for this route should
+	// be updated with a condition that describes the error more specifically.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Port specifies the destination port number to use for the
+	// backend resource.
+	// This field is required when the backend is a Kubernetes Service.
+	//
+	// Support: Core
+	//
+	// +optional
+	Port *PortNumber `json:"port,omitempty"`
+
+	// Weight specifies the proportion of HTTP requests forwarded to the backend
+	// This is computed as weight/(sum of all weights in this Backends list).
+	// For non-zero values, there may be some epsilon from the exact proportion
+	// defined here depending on the precision an implementation supports. Weight
+	// is not a percentage and the sum of weights does not need to equal 100.
+	//
+	// If only one backend is specified and it has a weight greater than 0, 100%
+	// of the traffic is forwarded to that backend. If weight is set to 0, no
+	// traffic should be forwarded for this entry. If unspecified, weight
+	// defaults to 1.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=1000000
+	Weight *int32 `json:"weight,omitempty"`
+}
+```
+
+A summarized diff for the changes being proposed:
+
+```patch
+diff --git a/apis/v1alpha1/shared_types.go b/apis/v1alpha1/shared_types.go
+index 458145f..de720cd 100644
+--- a/apis/v1alpha1/shared_types.go
++++ b/apis/v1alpha1/shared_types.go
+@@ -94,51 +94,39 @@ type GatewayReference struct {
+ 	Namespace string `json:"namespace"`
+ }
+ 
+-// RouteForwardTo defines how a Route should forward a request.
+-type RouteForwardTo struct {
+-	// ServiceName refers to the name of the Service to forward matched requests
+-	// to. When specified, this takes the place of BackendRef. If both
+-	// BackendRef and ServiceName are specified, ServiceName will be given
+-	// precedence.
++// BackendRef defines how and where a request should be forwarded from the Gateway.
++type BackendRef struct {
++	// Name is the name of the backend resource to forward matched requests to.
+ 	//
+ 	// If the referent cannot be found, the rule is not included in the route.
+ 	// The controller should raise the "ResolvedRefs" condition on the Gateway
+ 	// with the "DegradedRoutes" reason. The gateway status for this route should
+ 	// be updated with a condition that describes the error more specifically.
+ 	//
+-	// The protocol to use is defined using AppProtocol field (introduced in
+-	// Kubernetes 1.18) in the Service resource. In the absence of the
+-	// AppProtocol field a `networking.x-k8s.io/app-protocol` annotation on the
+-	// BackendPolicy resource may be used to define the protocol. If the
+-	// AppProtocol field is available, this annotation should not be used. The
+-	// AppProtocol field, when populated, takes precedence over the annotation
+-	// in the BackendPolicy resource. For custom backends, it is encouraged to
+-	// add a semantically-equivalent field in the Custom Resource Definition.
++	// +kubebuilder:validation:MinLength=1
++	// +kubebuilder:validation:MaxLength=253
++	Name string `json:"name"`
++
++	// Kind is kind of the backend resource.
+ 	//
+-	// Support: Core
++	// Support: Core (Kubernetes Service)
++	// Support: Custom (any other resource)
+ 	//
+ 	// +optional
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-718.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+## Note on future consolidation
+
+It has been noted that `forwardTo` doesn't flow as good as `backendRefs`.
+We could consider further consolidating the YAML if API conventions allow for the
+following:
+
+```yaml
+backendRefs:
```
robscott, line 175: Could we further simplify to this:
```suggestion
forwardTo:
```

robscott: Actually, reading the [conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field) you linked below, it seems like the closest we could potentially get to that would be `forwardToRefs`, but that doesn't feel quite right.

hbagdi: Nouns are generally better than verbs in these cases so I resorted to `backendRefs`.

robscott: If/when we add route inclusion what would this look like? Would a Route be described in `backendRefs` or in something like `routeRefs`? Originally I'd thought we'd have separate `include` and `forwardTo` sections, but that wouldn't make as much sense after this change. Sorry, I know this is not the core part of the GEP, just trying to figure out where we'd eventually end up after this change.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,223 @@
...
+
+## API
+
+
+The updated `RouteForwardTo` will be:
+(`HTTPRouteForwardTo` has been omitted for brevity)
+
+```go
+// RouteForwardTo defines how a Route should forward a request.
+type RouteForwardTo struct {
```
bowei, line 74: Is it possible to put an indicator as to if there are changes to the long comment text

like this?

```
// no change
// blah blah blah    <<
// no change
```

site-src/geps/gep-718.md
bowei: targetRef?

bowei:
/hold (in case of accidental merge)

This looks reasonable to me

LGTM

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,223 @@
...
+    kind: server
+    group: networking.acme.io
+  port: 80
+  weight: 20
+...
+```
+
+## Note on future consolidation
+
+It has been noted that `forwardTo` doesn't flow as good as `backendRefs`.
```
mark-church, line 170: Question - why wait to make this consolidation? Every change will be harder in the future and v1alpha2 is the best milestone to make these changes. The `forwardTo` block has a big impact on the YAML size and new user experience. Even the most basic Route with no matches must use it.

site-src/geps/gep-718.md
hbagdi: @bowei I added another summarized diff in the proposal for a clearer understanding of what is changing. PTAL.

site-src/geps/gep-718.md
hbagdi: > targetRef?

Target is a more generic word that `backend` in this context so my personal preference goes towards `backendRef`.


> If/when we add route inclusion what would this look like? Would a Route be described in backendRefs or in something like routeRefs? Originally I'd thought we'd have separate include and forwardTo sections, but that wouldn't make as much sense after this change. Sorry, I know this is not the core part of the GEP, just trying to figure out where we'd eventually end up after this change.

I don't know the answer but including a route inside a backendRef would open up other can of worms: how do weights come into play? What happens if a route forwards to multiple routes? Now, we could have weights at two separate levels, that would lead to some theoretically interesting math and practically unnecessary confusion.

The other side to this is that route inclusion, at least on the onset, seems like an advanced use-case that would be better served with an API of its own (like `include`) with its own conformance level and tests.

I'll reference this thread in another GEP that I'll write up for this rename and let's continue the discussion there.
I want this GEP to be focused on the other change.

site-src/geps/gep-718.md
hbagdi: I should have used the word `future` more carefully.
I'll be writing another small GEP to dedicate a discussion on this matter and then implement this and the future GEP together for v1a2.

hbagdi:
@robscott @bowei Ping for a review.

robscott:
Deploy Preview: https://deploy-preview-719--kubernetes-sigs-gateway-api.netlify.app/geps/gep-718/

robscott:
@hbagdi Thanks for writing this up! This generally LGTM and seems reasonable, but I think I'd rather explore the future consolidation in the same GEP, especially if this and the future consolidation would be part of the v1alpha2 release. I'm hesitant to commit to adding the extra yaml (backend.name vs serviceName) without exploring the other options currently listed as either future expansion or alternatives.

site-src/geps/gep-718.md
robscott: Yeah I agree that we shouldn't try to combine the concepts of Route inclusion/delegation with forwarding. My actual concern here was purely naming. IE `include` and `forwardTo` seemed to work well together, it seemed slightly harder to find a natural parallel to `backendRefs`, but I don't think that needs to block this. Whether we choose something like `include` or `routeRefs` doesn't matter that much here and should be out of scope for this GEP.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,278 @@
...
+We could consider further consolidating the YAML if API conventions allow for the
+following:
+
+```yaml
+backendRefs:
```
robscott, line 230: The more I've thought about this, I actually think this approach could make the most sense. It's just as concise as the existing API while also being less confusing. I'm a little concerned that there will not be a way to reference/apply policy to a specific item in this list, but that seems like an edge case to me. In most cases that could be resolved by applying policy either to the route rule or to the backend itself.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,278 @@
...
+	// AppProtocol field, when populated, takes precedence over the annotation
+	// in the BackendPolicy resource. For custom backends, it is encouraged to
+	// add a semantically-equivalent field in the Custom Resource Definition.
+	//
+	// Support: Core (Kubernetes Service)
+	// Support: Custom (any other resource)
+	//
+	// +optional
+    // +kubebuilder:default={kind: "service"}
+	BackendRef *LocalObjectReference `json:"backendRef,omitempty"`
```
jpeach, line 96: Maybe `TargetRef`, if we are considering names :)

site-src/geps/gep-718.md
jpeach: My preference is the structure shown here. I read it as "forward to target X, with parameters A,B,C". I don't think that collapsing improves the readability.

hbagdi:
> I think I'd rather explore the future consolidation in the same GEP, especially if this and the future consolidation would be part of the v1alpha2 release. I'm hesitant to commit to adding the extra yaml (backend.name vs serviceName) without exploring the other options currently listed as either future expansion or alternatives.

Got it. I'll update the GEP to include this. Expect an update early next week.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,278 @@
...
+ 	//
+ 	// Support: Core
+ 	//
+```
+For Kubernetes Services, the API will look like the following:
+
+```yaml
+...
+forwardTo:
+- backendRef:
```
stevesloka, line 190: Could we remove the `backendRef` field and just have those fields exist in the `RouteForwardTo` struct? 

You'd end up with this: 
```
forwardTo:
  - name: foo-service-v1
    port: 80
    weight: 80
  - name: foo-service-canary
    port: 80
    weight: 20 
---
forwardTo:
-  name: foo-v1
    kind: server
    group: networking.acme.io
    port: 80
    weight: 80
-  name: foo-v1-canary
    kind: server
    group: networking.acme.io
    port: 80
    weight: 20
```

stevesloka: This is mentioned here (https://github.com/kubernetes-sigs/gateway-api/pull/719/files#diff-200562f21cd2e117d439d1a59d8b3c844217b91f3d6224f34222687e9a8fa5faR267) to move the port/weight fields into the `backendRef` struct. 

I'm suggesting something similar but just the opposite, moving everything into the `forwardTo`.

hbagdi: @robscott @youngnick Do you know if this is a violation of the upstream API convention or not?
I lean towards an xRef containing only the reference (as API evolves, having such a standard makes the UX much better, envoy is a good example here) but happy to tweak this if that's helpful.

hbagdi:
@robscott @stevesloka @jpeach I've updated this GEP to include the larger `RouteForwardTo` rework. Please take another look.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,339 @@
...
+Since we are already considering dropping the special case of `service` as a
+backend, we have an opportunity to further simplify `RouteForwardTo` for better
+UX.
+
+## Proposal
+
+- Remove the `ServiceName` field from `RouteForwardTo`
+- Introduce `Service` as a default for `BackendRef.Kind`
+- Rename `RouteForwardTo.BackendRef` to `RouteForwardTo.TargetRef`
+- Rename `RouteForwardTo` to `Backend`
```
robscott, line 68: I personally think that `forwardTo` is better unless we're able to flatten everything into a `backendRefs` list. In this current proposal we end up with the same levels of nesting (`forwardTo.backendRef` becomes `backends.targetRef`). That feels like a slightly different way to accomplish the same thing and I'm not sure it justifies a breaking change. 

I am very curious if we could have a list of `backendRefs` that also included concepts like weight and port. If not, I think I'd prefer to fallback to `forwardTo.backendRef`.

site-src/geps/gep-718.md
robscott: This is a good question. I can't find anything in the k8s API conventions about this. @lavalamp would it ever be acceptable to include concepts like port or weight in an object reference? We're trying to figure out if we can flatten our API a bit. Essentially translating this:

```
forwardTo:
- backendRef:
    group: ""
    kind: Service
    name: foo
  port: 80
  weight: 10
```

to something like this:

```
backendRefs:
- group: ""
  kind: Service
  name: foo
  port: 80
  weight: 10
```

lavalamp: I think additional fields should be fine. As long as they don't collide with any standard fields, or fields we're likely to add. I doubt object references will ever have "port" or "weight" fields so you should be good. @deads2k will probably agree.

site-src/geps/gep-718.md
hbagdi: > I am very curious if we could have a list of backendRefs that also included concepts like weight and port. If not, I think I'd prefer to fallback to forwardTo.backendRef.

I'm fine with flattening those out if we have answers to the concerned raised below in the doc (alternatives).

hbagdi: Read Daniel's comment after posting the above comment, let me update this GEP in that case.

hbagdi: @robscott updated the GEP.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+spec:
+  hostnames:
+  - "bar.example.com"
+  rules:
+  - matches:
+    - headers:
+        type: Exact
+        values:
+          env: canary
+    backends:
```
hbagdi, line 290: Suggestions so far on this name:
- `backendRefs`: `refs` doesn't add much value here
- `targetRefs`: `target` feels a bit too generic here

lavalamp: Destination and/or source is almost always preferable to "target". Target is ambiguous because it could be talking about the source's target (which is the destination) or the *rule's* target, (which is often the source).

mark-church:
I'll just add my support for this GEP:

- The new structure is very straightforward and less indentation makes it easier to read
- "backend" as a noun will make language very clear when referring to it. 
- It still supports weights, namespaces, and if we add `name` policy references too

hbagdi:
> It still supports weights, namespaces, and if we add `name` policy references too.

Interesting reference to PolicyReference there.
Since we have flattened out the fields, `name` is used to identify the backend resource (as in the API resource name).
So, if the same name is used for referencing policy as well, are there any gotchas?

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+- Apply the same to HTTP types and filters
+
+## API
+
+The updated `Backend` (formerly `RouteForwardTo`) struct will be:
+(HTTP version has been omitted for brevity)
+
+```go
+// Backend defines how and where a request should be forwarded from the Gateway.
+type Backend struct {
```
robscott, line 80: I think this struct should be `BackendRef` and the field name should be `backendRefs`:

>The name of the reference field should be of the format "{field}Ref", with "Ref" always included in the suffix.

and 

>In the case of a list of object references, the field should be of the format "{field}Refs", with the same guidance as the singular case above.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+## Examples
+
+
+For Kubernetes Services, an updated `forwardTo` section will read as follows:
+
+```yaml
+...
+backends:
```
robscott, line 242:
```suggestion
backendRefs:
```

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+...
+```
+
+Here, the `kind` field is omitted as it will be injected as a default.
+
+For custom backends, the API will look like the following:
+
+```yaml
+...
+backends:
```
robscott, line 258:
```suggestion
backendRefs:
```

site-src/geps/gep-718.md
robscott, line 290:
```suggestion
    backendRefs:
```

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+- Introduce `Service` as a default for `BackendRef.Kind`
+- Pull fields from `RouteForwardTo.BackendRef` into `RouteForwardTo` itself
+- Rename `RouteForwardTo` to `Backend`
+- Rename `Route.Spec.Rules[].ForwardTo[]` to `Route.Spec.Rules[].Backends[]` in
+  all routes
+- Apply the same to HTTP types and filters
+
+## API
+
+The updated `Backend` (formerly `RouteForwardTo`) struct will be:
```
robscott, line 75:
```suggestion
The updated `BackendRef` (formerly `RouteForwardTo`) struct will be:
```

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
...
+-	// be updated with a condition that describes the error more specifically.
++	// Group is the group of the backend resource.
+ 	//
+-	// Support: Custom
+-	//
+-	// +optional
+-	BackendRef *LocalObjectReference `json:"backendRef,omitempty"`
++	// +kubebuilder:validation:MinLength=1
++	// +kubebuilder:validation:MaxLength=253
++	Group string `json:"group"`
```
robscott, line 206: Nit: I'm more used to Group>Kind>Name ordering to more closely match the hierarchy.

site-src/geps/gep-718.md
```diff
@@ -0,0 +1,338 @@
+# GEP-718: Rework forwardTo segment in routes
+
+Issue URL: https://github.com/kubernetes-sigs/gateway-api/issues/718
+
+Status: implementable
```
robscott, line 5: I meant to update this in the template, but haven't yet, do you mind making the following change to match the other GEPS?

```suggestion
Issue: [#718](https://github.com/kubernetes-sigs/gateway-api/issues/718)

Status: Implementable
```

robscott (commented):
Thanks for the work on this! Just a few nits and a suggested tweak to the name but otherwise LGTM.

mark-church:
Sorry I was incorrect in suggesting that. `name` could not be used for policy references because the same backend can exist multiple times in the same Route but it may be desirable to have different policies for different matches to the same backend. I like the current usage of `name` though and don't feel that we should change it.

It seems that we have two options to support backend-level policies:
- Use separate Routes for each unique match and have the policy attach directly to the route. This feels very reasonable to me since this is somewhat of a niche policy attachment point and it's not too difficult to separate them into separate routes.
- We add a field that serves as the backend reference id, whose value must be unique within a route. I don't think we can force `name` to be unique because that would prevent backends from getting referenced more than once so this would probably need to be an additional field.

I don't feel strongly that this is something that needs to be solved now, but figure I'd bring it up.

site-src/geps/gep-718.md
hbagdi: I'll take care of this in the actual PR. Please poke again if you feel strongly.

hbagdi:
@robscott Updated.

robscott:
Thanks @hbagdi!

/lgtm

robscott:
I think this has broad enough feedback and consensus to remove the hold.

/hold cancel

Issue #655, opened 2021-05-07
v1alpha2: Clarify HTTPRouteFilter conformance levels


banks:
**What happened**:

Reading the API definition in https://github.com/kubernetes-sigs/gateway-api/blob/3aa586db603203a32005bf1c2589ea45ab4ddbf4/apis/v1alpha2/httproute_types.go#L443 there seems to be some inconsistency in the conformance level annotations for `HTTPRouteFilter`.

It's possible I'm just reading it wrong, but if so perhaps the presentation could be clarified to ensure the spec is clear to other implementors.

Specifically, in `HTTPRouteFilter` I see `RequestHeaderModifier` is `Core` while `RequestMirror` is `Extended` (which seems sensible):

https://github.com/kubernetes-sigs/gateway-api/blob/3aa586db603203a32005bf1c2589ea45ab4ddbf4/apis/v1alpha2/httproute_types.go#L468-L481

But then in the annotations on those nested struct types:
 - all the fields in the `Core` `HTTPRequestHeaderFilter ` are marked as `Extended`. If all config fields only require conformance by "extended" implementations, what does it mean that the filter itself is required by `Core` implementations?
 - `HTTPRequestMirrorFilter` which is marked as `Extended` above, has a field `ServiceName` which is marked as `Core` conformance level. What does it mean for a field to be required by `Core` conforming implementations if it's part of an `Extended` struct?

**What you expected to happen**:

Conformance level annotations for nested configuration to be the same or weaker than the parent configuration level.

Thanks for your great work on this spec, I'm following closely!

*[edit]* I notice afterwards that `v1alpha2` where I saw this is unstable so perhaps this is still a WIP and tracked elsewhere? I didn't see anything obvious though.

hbagdi:
Doing a bit of git spelunking, we didn't have enough discussion around the conformance level for those fields and they got marked as `extended` in a cleanup: https://github.com/kubernetes-sigs/gateway-api/commit/034c3234f1f053621a3f6a6f6bd384781b8508ec#diff-d1c59de21c79ea2bd1a99c2f52d9879981dd785848408064289287cb262fe600R422-R424

I _think_ fields inside `RequestHeaderModifier` shouldn't even have conformance level as the behavior of the filter is well-defined (except #480).

hbagdi:
#721 is a partial fix.
#719 should simplify the filter a bit. I'll take care of updating the language to match the discussion here.
Let me know if I'm missing anything else from here.

[...]

PR #721, opened 2021-07-13
fix support level for HTTPRequestHeaderFilter

hbagdi:
**What type of PR is this?**

/kind bug
As noted in #655, a core filter with extended support fields is
confusing and also seems to be incorrect.
This patch drops the extended field from the filter and is a partial fix
for the referenced issue.

HTTPRequestMirroFilter has not been patch as GEP-718 could potentially
simplify that.

Partial fix for #655

```release-note
bugfix: To avoid confusion, fields within HTTPRequestHeaderFilter are no more marked as 'extended'. All fields must be supported by implementations for this core filter.
```


robscott:
Thanks!

/lgtm

PR #722, opened 2021-07-15
Tweaks to ReferencePolicy GEP

robscott:
**What type of PR is this?**
/kind gep

This is a quick follow up to #711 that addresses a few leftover comments on that PR + does some minor cleanup.

```release-note
NONE
```

Deploy Preview: https://deploy-preview-722--kubernetes-sigs-gateway-api.netlify.app/geps/gep-709/

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -33,6 +33,8 @@ nav:
       - API specification: references/spec.md
       - Releases: references/releases.md
       - Implementations: references/implementations.md
+  - Enhancement Proposals:
+      - "709: Cross Namespace References from Routes": geps/gep-709.md
   - Contributing:
       - Developer guide: contributing/devguide.md
       - Gateway Enhancement Proposal: contributing/gep.md
--- a/site-src/geps/gep-709.md
+++ b/site-src/geps/gep-709.md
@@ -1,6 +1,6 @@
 # GEP-709: Cross Namespace References from Routes
 
-* Issue URL: https://github.com/kubernetes-sigs/gateway-api/issues/709
+* Issue: [#709](https://github.com/kubernetes-sigs/gateway-api/issues/709)
 * Status: Implementable
 
 ## TLDR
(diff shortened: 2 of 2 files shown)


robscott:
I think we're pretty close to consensus on this, but will hold as a precaution.

/hold

site-src/geps/gep-709.md
```diff
@@ -82,8 +88,8 @@ decisions:
    more powerful it may encourage unnecessarily insecure configuration.
 
 ```go
-// ReferencePolicy identifies cross namespace relationships that are trusted for
-// Gateway API.
+// ReferencePolicy identifies cross namespace relationships between types of
```
robscott, line 91: In the previous PR, a suggestion was made to rename this to [InboundReferencePolicy](https://github.com/kubernetes-sigs/gateway-api/pull/711#discussion_r670025962). Wanted to pick up the conversation in this PR. I think I slightly prefer the shorter `ReferencePolicy` name, but admit that `InboundReferencePolicy` would make room for something like `OutboundReferencePolicy`. That might allow admins to restrict where Gateways in their namespace could forward to. Thoughts?

/cc @youngnick @bowei @danehans @hbagdi @jpeach

mkdocs.yml
```diff
@@ -33,6 +33,8 @@ nav:
       - API specification: references/spec.md
       - Releases: references/releases.md
       - Implementations: references/implementations.md
+  - Enhancement Proposals:
+      - "709: Cross Namespace References from Routes": geps/gep-709.md
```
robscott, line 37: Adding full GEP names here could end up being a bit much for our existing nav structure, would anyone prefer a single GEP index page that then linked to each individual GEP? I don't feel strongly either way, but wanted to start with something. (https://deploy-preview-722--kubernetes-sigs-gateway-api.netlify.app/geps/gep-709/)

site-src/geps/gep-709.md
jpeach: Yeh, my preference is also for a single ReferencePolicy with From+To.

robscott: Yeah as I've been thinking about it more, it would not be that difficult to expand ReferencePolicy to cover both inbound and outbound references. I don't think that's something we actually want to do, but the only required change would be adding `namespace` to the `to` struct. The larger hurdle would be the complexity for users in trying to restrict/allow both inbound and outbound references.

robscott: As discussed in community meeting today, we believe it is unlikely that we will ever have a need for an OutboundReferencePolicy. Given that, the preference is to keep the simpler `ReferencePolicy` name. I'll update the comments around ReferencePolicy to make it clearer that it refers to inbound references.

robscott: I've updated the wording around this, PTAL

robscott:
Thanks to everyone for the great feedback around this! Based on the discussion at the community meeting today, it sounds like this was pretty close to consensus with no major objections. Given that, I'm going to remove the hold and move this back to a place where an LGTM can merge this in.

/hold cancel

youngnick:
I did the last one, so I should probably leave this to someone else? :)

howardjohn (approved):

robscott:
ping @hbagdi @jpeach @bowei for final review

site-src/geps/gep-709.md
```diff
@@ -207,19 +217,21 @@ effective safeguards are in place.
...
 
-    // Namespace is the namespace of the referrent.
+    // Namespace is the namespace of the backend. When unspecified, the local
+    // namespace is inferred.
     //
     // Support: Core
     //
     // +kubebuilder:validation:MinLength=1
     // +kubebuilder:validation:MaxLength=253
+    // +optional
```
jpeach, line 234: Should the type also change to `string*`?

robscott: Updated, thanks!

jpeach (approved):
Looks good. Omitting `/lgtm` until other reviewers can take a look.

PR #723, opened 2021-07-15
Clarifying Redirect Support levels

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This clarifies support levels in the Redirect filter and documents how overlapping support levels should be interpreted.

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -657,6 +657,8 @@ type HTTPRequestRedirect struct {
 	// header in the response.
 	// When empty, the protocol of the request is used.
 	//
+	// Support: Extended
+	//
 	// +optional
 	// +kubebuilder:validation:Enum=HTTP;HTTPS
 	Protocol *string `json:"protocol,omitempty"`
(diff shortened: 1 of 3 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -657,23 +657,31 @@ type HTTPRequestRedirect struct {
 	// header in the response.
 	// When empty, the protocol of the request is used.
 	//
+	// Support: Extended
```
robscott, line 660: For GCP we have some cases where we just support same protocol or HTTP->HTTPS, not HTTPS->HTTP. I think that means this field should be extended.

apis/v1alpha2/httproute_types.go
```diff
@@ -657,23 +657,31 @@ type HTTPRequestRedirect struct {
...
 	//
+	// Support: Core
+	//
 	// +optional
 	Hostname *string `json:"hostname,omitempty"`
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
 	// When empty, port (if specified) of the request is used.
 	//
+	// Support: Extended
```
robscott, line 677: Redirecting to arbitrary ports is not supported as broadly as I'd hoped.

Issue #699, opened 2021-06-28
Add support for multiple `kinds` within a single Gateway listener


howardjohn:
**What would you like to be added**:
Replace `kind` with `kinds` in gateway listener

**Why is this needed**:

We may want to select multiple kinds. For example, perhaps I have a listener I want TCP and TLS routes on. Or I make a custom GrpcRoute and want it on the same port as an HTTPRoute. Or I have a legacy resource (say Istio VirtualService) which I want on the same port as HTTPRoute.

The primary alternative would be to have multiple listeners defined and merge them. However, in some cases this is impossible. For example, if two listeners for the same domain are configure with different TLS properties, merging is ~impossible. Having multiple kinds seems more desirable.

Another alternative: make `kind` optional. If left empty, the controller will decide which routes to use based on the protocol. This may be desirable for other reasons as well (simpler config - `protocol: HTTP` will probably always have `kind: HTTPRoute`)

hbagdi:
> Another alternative: make kind optional. If left empty, the controller will decide which routes to use based on the protocol. This may be desirable for other reasons as well (simpler config - protocol: HTTP will probably always have kind: HTTPRoute)

I think it would be best to be explicit about `kind` in configuratoin for two reasons:
- Makes the controller behavior deterministic and makes the controller a little dumber (in a good way)
- [Cluster operators](https://gateway-api.sigs.k8s.io/concepts/security-model/) continue to have explicit control over which routes are being selected to route traffic.

howardjohn:
> Isn't this already possible?

In your example it is, although quite verbose. But that only works well if listeners are merge-able. For example:

> ```yaml
> listeners:
>   - protocol: HTTPS
>     hostname: foo.com
>     port: 80
>     routes:
>       kind: HTTPRoute
>     tls: passthrough
>   - protocol: HTTPS
>     hostname: foo.com
>     port: 443
>     routes:
>       kind: VirtualService
>     tls: terminate
> ```

We cannot merge these together since TLS settings are not compatible. This means it will become very unclear when we can actually merge them and when we cannot, and subtle changes may lead to breakage as they all need to be kept in sync, etc. Additionally, for our use case we will almost always want multiple `kinds`, so that would require all users to duplicate listeners.

Allowing optional also makes the GW more portable. If I just want to declare I am opening an HTTP port 80, why do I need to say what type of routes? I may be fine with any routes that the controller understands that will handle HTTP traffic.

[...]

Issue #724, opened 2021-07-16
GEP: Refresh Route-Gateway Binding


robscott:
**What would you like to be added**:
Refactor cross-namespace Route-Gateway binding to:

* Be more consistent with cross-namespace references from Routes (#709)
* Provide a clear path to enable Route inclusion (Routes including Routes).
* Simplify user experience based on initial feedback.
* Enable other kinds of Route parents in addition to Gateway, this could include:
  * Routes (as part of Route inclusion)
  * Custom Gateway resources
  * Mesh resources

**Why is this needed**:
The existing state is a common source of confusion when interacting with the API and has room for improvement, especially as we're considering adding new capabilities to the API.

PR #725, opened 2021-07-17
Adding GEP 724: Refresh Route-Gateway Binding

robscott:
**What type of PR is this?**
/kind gep

This PR adds GEP #724 as a follow up to various proposals and docs around how we can improve Route-Gateway binding.

```release-note
NONE
```

Deploy Preview: https://deploy-preview-725--kubernetes-sigs-gateway-api.netlify.app/geps/gep-724/


site-src/geps/gep-724.md
```diff
@@ -0,0 +1,426 @@
...
+  // * Route: Rule Name
+  //
+  // Support: Core
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +optional
+  SectionName string `json:"sectionName,omitempty"`
+
+  // Namespace is the namespace of the referent. When unspecified, the local
```
howardjohn, line 225: :+1: the "same namespace" and ability to omit for cluster scoped is great

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,426 @@
...
+
+  // Name is the name of the referent.
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  Name string `json:"name"`
+
+  // SectionName is the name of a section within the target resource. When
+  // unspecified, this targets the entire resource. In the following
+  // resources, SectionName is interpreted as the following:
```
howardjohn, line 214: nit: I assume for custom attachment points they can choose to make this something else? If so it may be best to clarify that in the comment

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,426 @@
...
+* Aligns with initial user feedback.
+
+### Disadvantages
+
+* Attaching a Route to a named listener with SectionName may be a bit confusing.
+* Does not utilize existing ReferencePolicy mechanism.
+* May be more difficult to understand which Routes are attached to a Gateway.
+* Adding/replacing a Gateway requires changes to Routes.
+
+### Potential Expansion
```
howardjohn, line 260: How does this impact possible `route -> route`?

howardjohn (commented):
I really like this. Istio has a pretty similar model (but this is much clearer, IMO) that hasn't shown any obvious issues

site-src/geps/gep-724.md
robscott: Yeah good point, will clarify that.

site-src/geps/gep-724.md
robscott: I think we'd end up with 3 options here, I'm not sure which is best:

1. 1:many Route->Routes: Nearly identical to Route->Gateway pattern described here, parent Route describes trusted namespaces per rule, child Routes use attachTo mechanism to attach to a parent.
2. 1:1 Route->Route: Require direct reference in both directions (include + attachTo would both be direct refs).
3. 1:1 Route->Route: Direct reference from parent to child, ReferencePolicy to denote trust from child ns.

I think 1 and 3 make the most sense, it really just depends on how we want the feature to be used. If we want Routes to be able to somewhat blindly delegate a path to a namespace, 1 makes more sense. If we want Routes to include a very specific Route in another namespace, 3 probably makes more sense.

howardjohn: (3) seems to make most sense to me.

site-src/geps/gep-724.md
bowei: Make the text more direct:

When unspecified (empty string), this will either be:

1. local namespace of the target is a namespace scoped resource
2. no namespace (not applicable) if the target is cluster-scoped.

bowei:
LGTM

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,426 @@
...
+* Follows pattern of direct reference in one direction with a broader trust
+  reference in the other direction.
+* Aligns with initial user feedback.
+
+### Disadvantages
+
+* Attaching a Route to a named listener with SectionName may be a bit confusing.
+* Does not utilize existing ReferencePolicy mechanism.
+* May be more difficult to understand which Routes are attached to a Gateway.
+* Adding/replacing a Gateway requires changes to Routes.
```
hbagdi, line 258: This one is a deviation from our previous conversations around the separation of routes and gateways, which is okay.
Any commentary on how to avoid this?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,426 @@
...
+* Enables attaching Routes to arbitrary parents, such as custom Gateways, other
+  Routes (to be defined), or Meshes.
+* Prevents accidental exposure of Routes.
+* Easy to understand which Gateways/parents a Route is attached to.
+* Further simplifies path to Route inclusion.
+* Follows pattern of direct reference in one direction with a broader trust
+  reference in the other direction.
+* Aligns with initial user feedback.
+
+### Disadvantages
```
hbagdi, line 253: The easy path becomes complex since Routes are now Gateway aware.

site-src/geps/gep-724.md
robscott: Yeah this is one that made me a bit sad. I really liked the ability to seamlessly spin up new Gateways and target the same Routes. The initial feedback we've received from our first UXR sessions suggests that's not actually as valuable as I thought and that app owners would rather have full control on the Gateways their app is attached to.

One note here is that this proposal isn't really adding anything new, it's just taking away other ways to connect Gateways and Routes. I think it makes sense to start with the simplest form possible for v1alpha2 and as requests come in, we can always consider adding back additional concepts such as selectors. I'm just not sure that request is going to be as common as I'd originally thought.

site-src/geps/gep-724.md
robscott: What is the easy path here? The default of a Gateway selecting all Routes in a namespace?

hbagdi: Yes, a route allowing any gateways to bind to itself.

robscott: Although those may be simpler, I think they're poor/insecure defaults. I'd prefer to move away from them in favor of more explicit attachment.

robscott:
Thanks to everyone for the great feedback on this in the community meeting yesterday! As suggested, I've added a rationale section and made the conceptual scope of the change I'm proposing more clear. PTAL

site-src/geps/gep-724.md
robscott: Added something in the new rationale section that covers this.

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+* Route owners have limited control in terms of which Gateways their Route is
+  attached to.
+
+### 2. ReferencePolicy with Routes referencing Gateways
+
+![ReferencePolicy with Routes referencing Gateways](images/724-alt2.png)
+
+The other way we could use ReferencePolicy would be with Routes referencing
+Gateways. Unfortunately the nested structure of Gateways makes this nearly
+impossible to do effectively. A core concept for Gateways is that each listener
```
jpeach, line 411: My main objection to this approach is that `ReferencePolicy` is a access control object, not a binding object. Best to not mix those concerns.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+```
+
+#### Advantages
+
+* Consistent use of ReferencePolicy throughout the API.
+* Provides a single way of binding Gateways to Routes.
+
+#### Disadvantages
+
+* Even the simplest cross-namespace reference from Gateway -> Route would
```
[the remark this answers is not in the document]
robscott: This relies on the existing namespace selector in Gateway listeners to denote where Routes are trusted to bind from. Although it would be nice to reuse ReferencePolicy here, the nested list of listeners makes it significantly more complex + verbose as shown in the alternatives.

site-src/geps/gep-724.md
robscott: Agree that we have to be careful here, but in this case ReferencePolicy would effectively been allowing access/references from Routes to Gateway listeners.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+Route - Gateway relationship. Given that each Gateway is likely to support many
+Routes, it would not be practical to support direct references from Gateways to
+Routes. Instead, it is simpler if Routes directly reference the Gateways they
+want to attach to. Gateways can then specify the namespaces they trust Routes
+to attach from.
+
+![Routes directly reference Gateway](images/724-proposal.png)
+
+In the following example, the lb Gateway indicates that it trusts Routes from
+the foo Namespace, and the HTTPRoute in that namespace attaches directly to the
```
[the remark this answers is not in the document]
robscott: Yeah that's an aspect I'm very open to changing, but didn't really want to cover it in this GEP. If there's enough interest I can expand the scope. As I understand backwards compat, we would still be able to translate a field to a list (and/or optional value) post v1alpha2.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +kubebuilder:default=networking.x-k8s.io
+  Group string `json:"group"`
+
+  // Kind is kind of the referent.
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  Kind string `json:"kind"`
```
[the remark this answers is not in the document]
robscott: Yeah completely agree here, I came very close to adding a Gateway default as well but ultimately decided against it.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+```go
+  // Selector specifies a set of route labels used for selecting
+  // routes to associate with the Gateway. If this Selector is defined,
+  // only routes matching the Selector are associated with the Gateway.
+  // An empty Selector matches all routes.
+  //
+  // Support: Core
+  //
+  // +optional
+  Selector *metav1.LabelSelector `json:"selector,omitempty"`
```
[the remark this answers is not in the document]
robscott: At this point the idea is that they would restrict the kinds of Routes that could be attached to the Gateway, but I'm increasingly thinking they don't make sense anymore, at least not in their current form. With direct references we have clear indication of intent and an easy way to communicate via status that attachment was unsuccessful.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+### 2. ReferencePolicy with Routes referencing Gateways
+
+![ReferencePolicy with Routes referencing Gateways](images/724-alt2.png)
+
+The other way we could use ReferencePolicy would be with Routes referencing
+Gateways. Unfortunately the nested structure of Gateways makes this nearly
+impossible to do effectively. A core concept for Gateways is that each listener
+should be able to attach to an entirely different set of Routes. For example,
+a Gateway may want to delegate foo.com to the foo namespace and bar.com to the
+bar namespace. Unfortunately thay would be very difficult to recreate with
```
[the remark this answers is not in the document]
robscott: Woops, updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+  //
+  // * local namespace of the target is a namespace scoped resource
+  // * no namespace (not applicable) if the target is cluster-scoped.
+  //
+  // Support: Extended
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +optional
+  Namespace string `json:"namespace,omitempty"`
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+  // resources, SectionName is interpreted as the following:
+  // * Gateway: Listener Name
+  // * Route: Rule Name
+  //
+  // Support: Core
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +optional
+  SectionName string `json:"sectionName,omitempty"`
```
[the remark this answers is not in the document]
robscott: Good question, I added some detail to the proposed spec for this.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  Name string `json:"name"`
+
+  // SectionName is the name of a section within the target resource. When
+  // unspecified, this targets the entire resource. In the following
+  // resources, SectionName is interpreted as the following:
+  // * Gateway: Listener Name
+  // * Route: Rule Name
```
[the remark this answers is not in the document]
robscott: Good point - added below. I think when I was initially writing this GEP they were included in GEP #713, added them here as well now.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+The proposed changes here can be summarized as:
+
+* Remove Route selector from Gateways.
+* Replace the 3 options from Route -> Gateway (All, FromList, SameNamespace)
+  with a reference list that supports arbitrary kinds.
+
+### Gateway
+
+On Gateway, the only change involves removing the Route selector field.
+Everything else remains the same.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+    sectionName: foo
+  rules:
+  - name: abc
+    matches:
+    - path: /bar
+```
+
+## Rationale
+
+#### 1. Remove Complexity While Possible
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,494 @@
...
+## Goals
+
+Refactor cross-namespace Route-Gateway binding to:
+
+* Be more consistent with [cross-namespace references from
+  Routes](/geps/gep-709.md)
+* Provide a clear path to enable Route inclusion (Routes including Routes).
+* Simplify user experience based on initial feedback.
+* Enable other kinds of Route parents in addition to Gateway, this could include:
+    * Routes (as part of Route inclusion)
```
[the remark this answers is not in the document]
robscott: Agree with the risks of implementing Route inclusion this way. One distinction I should probably make more clear is that I see this as a potential advantage of the flexibility of this approach, not that this should ultimately be how we implement Route inclusion. Maybe we do need to start on a GEP for Route inclusion so we can get the final piece of this puzzle solved, even if it doesn't make it into the initial v1alpha2 release...

I think it's finding itself in an awkward place between ReferencePolicy and Route->Gateway, both patterns are very similar and could potentially apply to Route->Route, but based on what you're saying here and the comment from @howardjohn above, ReferencePolicy may make more sense for that relationship. If that's the case, we just need to be sure that this `attachTo` concept is more clearly scoped to either just Gateway, or Gateway/Mesh-like parent resources.

jpeach:
> I really like this. Istio has a pretty similar model (but this is much clearer, IMO) that hasn't shown any obvious issues

@howardjohn could you post a link to the istio mechanism?

howardjohn:
> > I really like this. Istio has a pretty similar model (but this is much clearer, IMO) that hasn't shown any obvious issues
> 
> @howardjohn could you post a link to the istio mechanism?

https://istio.io/latest/docs/reference/config/networking/gateway/, grep for "some-config-namespace/my-gateway". Basically VirtualService points to a specific Gateway, and the two way handshake is based on hostname alignment

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Updated this GEP to change them as well since that does seem highly related here.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Ended up adding that to this GEP after all, agree that it is quite relevant.

robscott:
Thanks for the great feedback on this GEP! I've expanded the scope slightly to also restructure how Gateway listeners can restrict supported Route kinds based on feedback from @jpeach. This would also fix #699. Does this latest change make sense?

site-src/geps/gep-724.md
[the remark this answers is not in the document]
howardjohn: is it backwards compatible to add a default later? Seems like it should be?

I don't have a strong preference, but suspect 99% of use cases will be `Gateway` so I could see folks liking the defaulting. Good to be able to add it later if its an issue

howardjohn (approved):

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+	Kind string `json:"kind"`
+```
+
+#### Added
+```go
+type ListenerRoutes struct {
+  // ...
+  // Kinds specifies the groups and kinds of Routes that are allowed to bind to
+  // this Gateway listener. When unspecified or empty, the only limitation on
+  // the kinds of Routes supported is the Listener protocol. Kind MUST
```
hbagdi, line 238: It would be better to have an explicit list over an implied reference. The map of protocol->[]route-types-on-that-protocol will change based on the implementations and even within core as time goes on, which could lead to unnecessary exposure to other protocols.

If YAML verbosity is the reason, IMHO we should pick clarity in this case over conciseness.
Another solution would be to introduce defaulting via the webhook (need a mutating one) based on the listener protocol.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  Kinds []RouteGroupKind `json:"kinds,omitempty"`
+}
+
+type RouteGroupKind struct {
+  // Group is the group of the Route.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:default=networking.gateway.k8s.io
```
hbagdi, line 257:
```suggestion
	// +kubebuilder:default=gateway.networking.k8s.io
```

site-src/geps/gep-724.md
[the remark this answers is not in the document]
hbagdi: > is it backwards compatible to add a default later?

I don't think so but I'm not sure.

I'd recommend to adding a default here, I expect a large majority would be gateway.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+To Route Specs:
+```go
+  // AttachTo defines the parent resources this Route should be attached to. The
+  // only kind of target resource with "Core" support is Gateway. This API may
+  // be extended in the future to support additional kinds of parent resources
+  // such as Routes.
+  //
+  // +optional
+  // +kubebuilder:validation:MaxItems=16
+  AttachTo []AttachRef `json:"gateways,omitempty"`
```
hbagdi, line 323: No action required here but a comment:
If we settle on this, we should move away from "binding" terminology and instead move to "attach/detach" terminology between Routes/Gateways.

"BindTo" would also have been fine, that would avoid rephrasing documentation.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  require a ReferencePolicy in each target namespace.
+* Existing demos and examples would become significantly more verbose.
+* Does not suppport attaching Routes to arbitrary parents.
+* Does not prevent accidental exposure of Routes.
+* Route owners have limited control in terms of which Gateways their Route is
+  attached to.
+
+### 2. ReferencePolicy with Routes referencing Gateways
+
+![ReferencePolicy with Routes referencing Gateways](images/724-alt2.png)
```
hbagdi, line 527: Are the images for the alternatives swapped or am I understanding these alternatives wrong?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  // "ResolvedRefs" condition to false for this listener with the
+  // "InvalidRoutesRef" reason.
+  //
+  // Support: Core
+  //
+  // +optional
+  Kinds []RouteGroupKind `json:"kinds,omitempty"`
+}
+
+type RouteGroupKind struct {
```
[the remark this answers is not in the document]
hbagdi: My preference would be to require both and then make them optional in future (not a breaking change).

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+}
+```
+
+#### Added
+To Route Specs:
+```go
+  // AttachTo defines the parent resources this Route should be attached to. The
+  // only kind of target resource with "Core" support is Gateway. This API may
+  // be extended in the future to support additional kinds of parent resources
+  // such as Routes.
```
hbagdi, line 319: Just want to call out (no action required): This is not the case today because RouteStatus explicitly reference Gateway.

Question: Should we rename fields in the status block as well?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  Kind string `json:"kind"`
+
+  // Namespace is the namespace of the referent. When unspecified (empty
+  // string), this will either be:
+  //
+  // * local namespace of the target is a namespace scoped resource
+  // * no namespace (not applicable) if the target is cluster-scoped.
```
hbagdi, line 349: @robscott  https://github.com/kubernetes/enhancements/pull/2366#discussion_r567114299

We should avoid implicit cluster-scoped fields.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  // interpreted.
+  //
+  // When unspecified (empty string), this will reference the entire resource.
+  // For the purpose of status, an attachment is considered successful if at
+  // least one section in the parent resource accepts it. For example, Gateway
+  // listeners can restrict which Routes can bind to them by Route kind,
+  // namespace, or hostname. If 1 of 2 Gateway listeners accept attachment from
+  // the referencing Route, the Route MUST be considered successfully
+  // attached. If no Gateway listeners accept attachment from this Route, the
+  // Route MUST be considered detached from the Gateway.
```
hbagdi, line 380: Did we consider a []string for this type to support explicitly tartgetting ports? Not a requirement?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+  GatewayRefs []GatewayReference `json:"gatewayRefs,omitempty"`
+}
+```
+
+#### Added
+To Route Specs:
+```go
+  // AttachTo defines the parent resources this Route should be attached to. The
+  // only kind of target resource with "Core" support is Gateway. This API may
+  // be extended in the future to support additional kinds of parent resources
```
hbagdi, line 318: Clarify: It is an error to reference a Gateway resource multiple times list this list.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,614 @@
...
+    // Name is the name of the Route rule. If more than one Route Rule is
+    // present, each Rule MUST specify a name. The names of Rules MUST be unique
+    // within a Route.
+    //
+    // Support: Core
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    // +optional
+    Name string `json:"name,omitempty"`
```
hbagdi, line 421: Is this assuming that we will build route inclusion using this mechanism?
If that is the only reason, please remove this field. We can add it in later once we are more sure (we don't seem to be sure right now: https://github.com/kubernetes-sigs/gateway-api/pull/725/files#r677977878).

Does ReferencePolicy have anything to do here?

site-src/geps/gep-724.md
howardjohn: I disagree. Not sure I can provide any concrete reason why other than personal preference though :slightly_smiling_face:.

It seems to me there is minimal risk of accidentally selecting the wrong kind - and even if it does happen, is it so bad? I am assuming we don't add an InsecureHTTPRoute that violates a bunch of security rules in the future (or a less contrived example) where it would be bad to select extra routes.

site-src/geps/gep-724.md
hbagdi: This is in multiple places.

site-src/geps/gep-724.md
howardjohn: It feels like I am probably wrong if I am disagreeing with an API lead, but having to specify adding `scope` seems like a really bad UX. The GVK already determines if its scoped or not, feels redundant...

site-src/geps/gep-724.md
howardjohn: I think its still used to attach policy to a route, even if not Route<->Route. Ie apply TimeoutPolicy to a route

site-src/geps/gep-724.md
hbagdi: Another note, with `ForwardTo` going away with #718, we will have only one "verbTo" field.
We have already seen how verb fields are a little challenging to document and communicate. Might be worth considering an alternative name here.
`attachments`, `from` (to match RP resource), `binders`, `attachers`, `referrer`?

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: > My preference would be to require both and then make them optional in future

Agree with that. That's the current state as I understand it - even though Group is marked as "optional" that's only because there is a default value. There is no way for it to be empty.

site-src/geps/gep-724.md
robscott: Good point. Any preference between the following field names?

- parentRefs
- attachRefs
- bindingRefs

I think I still slightly prefer the clarity of `attachTo` but I don't think that actually follows the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-of-the-reference-field).

/cc @bowei @howardjohn @jpeach @mark-church @youngnick

site-src/geps/gep-724.md
robscott: I think it's always easier to go from a single value to a list than the other way around (for example Pod IP -> IPs now). My preference is to start on the simpler side and expand if/when it's needed.

site-src/geps/gep-724.md
robscott: Yeah this was copied verbatim from the original GEP for policy attachment where this was proposed. My rationale for adding it here was so that it could also apply to policy attachment, but I'll pull that out here because it's out of scope. It will be fairly easy to add later if necessary.

site-src/geps/gep-724.md
robscott: Good call, added a section renaming this.

site-src/geps/gep-724.md
robscott: Good catch, I really don't love the idea of a `scope` field, but it does seem to be the apimachinery preference + has prior art for Ingress now.

site-src/geps/gep-724.md
robscott: Good call, added

site-src/geps/gep-724.md
robscott: Removed this, we can add it later if needed.

site-src/geps/gep-724.md
robscott: Going with parentRefs for now, open to better names though.

site-src/geps/gep-724.md
robscott: 🤦 updated, thanks!

site-src/geps/gep-724.md
robscott: I'm not tied to either of these approaches, but do prefer the more lenient interpretation I proposed due to ease of use. With the change to direct references proposed here, the risk of routes being accidentally bound to Gateways seems very low.

If we were still doing selection, a change to the Routes supported for a given protocol could immediately result in more Routes being attached to a Gateway when Route types were not specified. On the other hand, with direct references from Routes, a Route of the new type would have to directly reference the Gateway it wanted to attach to. That seems unlikely to cause things to break in significant ways. 

I'd completely agree that we should avoid this if we were still using selection, but with this new approach, I think the only thing most Gateway admins should have to worry about is the hostname, port, and protocol of a listener + the namespaces it supports. If you need to further restrict Route kinds, that's possible, but that seems like an advanced and usually unnecessary use case.

hbagdi: > If we were still doing selection, a change to the Routes supported for a given protocol could immediately result in more Routes being attached to a Gateway when Route types were not specified. On the other hand, with direct references from Routes, a Route of the new type would have to directly reference the Gateway it wanted to attach to. That seems unlikely to cause things to break in significant ways.

Fair. I'm sold.

How about having defaults in here?
If I'm debugging or trying to understand how the API works, and I see '[]RouteGroupKind' as empty, how I'm supposed to know which types routes are binding to this listener since there is no concept of 'protocol' in Routes.
I could do filtering and what not but seeing the route types in there will give a clear pointer to look at those other resources.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  // correspond to kinds of Routes that are compatible with the application
+  // protocol specified in the Listener's Protocol field. If an implementation
+  // does not support or recognize this resource type, it SHOULD set the
+  // "ResolvedRefs" condition to false for this listener with the
+  // "InvalidRoutesRef" reason.
+  //
+  // Support: Core
+  //
+  // +optional
+  Kinds []RouteGroupKind `json:"kinds,omitempty"`
```
hbagdi, line 248: Unbounded list. Can be taken care of in the implementation directly.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  //
+  // +optional
+  GatewayRefs []GatewayReference `json:"gatewayRefs,omitempty"`
+}
+```
+
+#### Added
+To Route Specs:
+```go
+  // ParentRefs references the parent resources this Route should be attached
```
hbagdi, line 315: should is lenient and doesn't get the point across.
We want to communicate "anyone can attach to this route ONLY IF they are in this list".
Something along the lines:
```suggestion
  // ParentRefs references the resources that can process/attach/use this Route.
```

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  Group string `json:"group"`
+
+  // Kind is kind of the referent.
+  //
+  // Support: Core (Gateway)
+  // Support: Extended (Other Resources)
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +kubebuilder:default=Gateway
```
hbagdi, line 349: optional and pointer; can be taken care of in the implementation PR

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  // the referencing Route, the Route MUST be considered successfully
+  // attached. If no Gateway listeners accept attachment from this Route, the
+  // Route MUST be considered detached from the Gateway.
+  //
+  // Support: Core
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +optional
+  SectionName string `json:"sectionName,omitempty"`
```
Miciah, line 407: `MinLength=0` or pointer type. 

I suggested on a call that we could use `ObjectFieldSelector` or similar to specify the section, but I neglected to comment on the PR earlier.  Would it make sense to use `ObjectFieldSelector` so the API is explicit about which section this references?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+* Namespaces: Same, Selector, or All
+
+Routes then have three options as far as which Gateways can bind to them:
+
+* Same namespace (default)
+* From a list of Gateways
+* All
+
+Although this enables a great deal of flexibility, it can lead to confusion. For
+example, 2 separate label selectors from Gateway can be challenging to compute.
```
Miciah, line 46: What does "challenging to compute" mean?  Difficult for humans to reason about? Difficult for an implementation to compute the intersection of routes matching the route selector and routes in namespaces matching the namespace selector?

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+```go
+  // Selector specifies a set of route labels used for selecting
+  // routes to associate with the Gateway. If this Selector is defined,
+  // only routes matching the Selector are associated with the Gateway.
+  // An empty Selector matches all routes.
+  //
+  // Support: Core
+  //
+  // +optional
+  Selector *metav1.LabelSelector `json:"selector,omitempty"`
```
Miciah, line 197: This should be indented (or the lines further down should be unindented).

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Group *string `json:"group,omitempty"`
+	// Kind is the kind of the route resource to select.
+	//
+	// Kind MUST correspond to kinds of routes that are compatible with the
+	// application protocol specified in the Listener's Protocol field.
+	//
+	// If an implementation does not support or recognize this
+	// resource type, it SHOULD set the "ResolvedRefs" condition to false for
```
Miciah, line 225:
```suggestion
	// resource type, it SHOULD set the "ResolvedRefs" condition to "False" for
```

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+}
+```
+
+#### Added
+To Route Specs:
+```go
+  // ParentRefs references the parent resources this Route should be attached
+  // to. The only kind of parent resource with "Core" support is Gateway. This
+  // API may be extended in the future to support additional kinds of parent
+  // resources such as Routes. It is invalid to reference the same parent
```
Miciah, line 318: There's no "Route" kind, so it would be clearer to use a specific example:
```suggestion
  // resources, such as HTTPRoute. It is invalid to reference the same parent
```
or use more general language:
```suggestion
  // resources, such as one of the route kinds. It is invalid to reference the same parent
```

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  //
+  // * local namespace of the target is a namespace scoped resource
+  // * no namespace (not applicable) if the target is cluster-scoped.
+  //
+  // Support: Extended
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  // +optional
+  Namespace string `json:"namespace,omitempty"`
```
Miciah, line 363: I think you need to use `MinLength=0` or a pointer type for this to be optional.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  // Scope represents if this refers to a cluster or namespace scoped resource.
+  // This may be set to "Cluster" or "Namespace".
+  //
+  // Support: Core (Namespace)
+  // Support: Extended (Cluster)
+  //
+	// +kubebuilder:validation:Enum=Cluster;Namespace
+  // +kubebuilder:default=Namespace
+  // +optional
+  Scope string `json:"scope,omitempty"`
```
Miciah, line 374: I think you need to add `"";` to the enum or make this a pointer type for it to be optional.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+```
+
+This would also require adding a `Name` field to Gateway listeners and Route
+rules to support the SectionName concept:
+
+```go
+type Listener struct {
+    // Name is the name of the Listener. If more than one Listener is present
+    // each Listener MUST specify a name. The names of Listeners MUST be unique
+    // within a Gateway.
```
Miciah, line 418: We should also specify `+listType=map` with `+listMapKey=name`.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+    Name string `json:"name,omitempty"`
+    // ...
+}
+```
+
+#### Changed
+
+To accomodate Routes with arbitrary types of parents, `RouteGatewayStatus` would
+be renamed to `RouteParentStatus`. Similarly, the `GatewayRef` inside that
+struct would be
```
Miciah, line 434: The sentence is unterminated and the thought incomplete.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+    port: 80
+---
+kind: ReferencePolicy
+metadata:
+  name: foo-lb
+  namespace: infra
+spec:
+  from:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
```
Miciah, line 571: Shouldn't `from:` specify HTTPRoute, and possibly a namespace?
```suggestion
  from:
  - group: gateway.networking.k8s.io
    kind: HTTPRoute
    namespace: foo
```

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+* Provides a single way of binding Gateways to Routes.
+* Supports attaching Routes to arbitrary parents.
+* Prevents accidental exposure of Routes.
+
+#### Disadvantages
+
+* In most cases, each listener in a Gateway would require a unique
+  ReferencePolicy resource.
+* Even the simplest cross-namespace reference from Route -> Gateway would
+  require a ReferencePolicy in each target namespace.
```
Miciah, line 606: Maybe worth calling out explicitly that this point would rule out self-service use-cases (if I understand correctly).  Most likely, I, as an application developer, do not have permission to create a ReferencePolicy in the infra namespace.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: I think the 1:many scenario described in a different [comment](https://github.com/kubernetes-sigs/gateway-api/pull/725#discussion_r672455899) is really the only way parentRef could make sense to refer/attach to a parent Route. With that kind of approach, we'd mirror the Route-Gateway relationship pretty closely for Route-Route. That has some appeal in that it would allow for broader delegation of a path to a namespace and not just a specific route.

On the other hand, it would add a lot of complexity to an already complex API. The other alternative of direct refs to Routes to include with ReferencePolicy as the trust mechanism feels _much_ simpler. Maybe that means we need some kind of more specific word here to indicate "resource that exists conceptually above Routing". For 99% of use cases, I think that's just Gateway, but I do like the idea of a term that's generic enough to represent a reference to a mesh or other similar non-Gateway concept.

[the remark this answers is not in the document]
robscott: One positive here is that renaming a field is technically backwards compatible, so we do have some flexibility here if we don't get the name right the first time.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+```
+
+#### Added
+To Route Specs:
+```go
+  // ParentRefs references the parent resources this Route should be attached
+  // to. The only kind of parent resource with "Core" support is Gateway. This
+  // API may be extended in the future to support additional kinds of parent
+  // resources such as Routes. It is invalid to reference the same parent
+  // resource more than once.
```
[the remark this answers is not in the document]
robscott: That seems reasonable, I can add something to clarify that.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  //
+  // +optional
+  // +kubebuilder:validation:MaxItems=16
```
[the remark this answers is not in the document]
robscott: Just trying to avoid unbounded lists. Hard to think of any use case where this value would be more than 3 or 4. I guess it's possible that a single Route could be used to redirect to HTTPS, and that could get reused frequently. Did you have a use case for >16?

site-src/geps/gep-724.md
youngnick: > I'd completely agree that we should avoid this if we were still using selection, but with this new approach, I think the only thing most Gateway admins should have to worry about is the hostname, port, and protocol of a listener + the namespaces it supports. If you need to further restrict Route kinds, that's possible, but that seems like an advanced and usually unnecessary use case.

I think this is the argument that clinches this for me. As long as we mandate what controllers must do for the status of an object that attempts to attach to a listener that can't support it, I think it's okay.

However, I do think that it's important to have some status information that makes it clear to a Gateway owner that Routes have attempted to attach to a Listener that doesn't support them - we need to ensure there's a trail in both Gateway -> Route and Route -> Gateway directions for people to follow when things go wrong, because the objects have different owners.

In terms of having defaults, I think that it's important for the described behavior that this be defaulted to an empty list, but we can specify defaults in the RouteGroupKind struct. I feel like @hbagdi's concern here is about discoverability and status though?

site-src/geps/gep-724.md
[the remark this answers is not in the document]
youngnick: This means that the way these fields are used is that if this stanza is not specified, match anything. If it is, match only this. You must specify at least a Kind, in which case the group defaults to `gateway.networking.io`. Otherwise, you may specify any Kind.

I like it.

site-src/geps/gep-724.md
youngnick: yeah, agreed, although I would nit that it should be `attach to`. :)

site-src/geps/gep-724.md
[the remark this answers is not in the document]
youngnick: I lean towards explicitness on this one as well. +1 to no default.

site-src/geps/gep-724.md
youngnick: Could be left for the implementation, but yes, I agree that this should probably be a pointer, and have the omitted behavior defined to be the same as setting to "Namespace".

site-src/geps/gep-724.md
youngnick: We designed Contour's HTTPProxy to be more like (3), but we've *definitely* had requests for not requiring a specific Route name, which I think is more like (1).

I think that (3) requires more out-of-band coordination between parties than (1) does, which could be good or bad. I'm still inclined to err on the side of more coordination and explicit config, because blanked includes scare me a bit. But it's not a strong preference.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
youngnick: The thing that makes Route -> Route inclusion much more complex is the semantics about how you carve up the URL/request metadata space. In HTTPProxy, we explicitly decided that the rules for assigning certain requests to a child HTTPProxy (in the Gateway API, this would be a child HTTPRoute) would become surrounding, uneditable *constraints* on that child, and that the child *did not need to know* about those constraints beforehand.

This has tradeoffs, obviously, but the key part here is that you can't design the Route -> Route reference mechanism without also designing how much the included Routes will know about where they are included.

In short, I think that what we have here is broad enough to *allow* us to experiment in either way, and we shouldn't hold this GEP up on figuring out how Route->Route inclusion works.

youngnick (commented):
For the direction and overall design, this LGTM, but I think there are few more tidy-up changes needed.

I think we need to ensure that we `s/networking.x-k8s.io/gateway.networking.k8s.io` everywhere. :)

site-src/geps/gep-724.md
robscott: Do we think "can attach to" is sufficient here? Not exactly sure how to phrase this, but I think the behavior we want is for these referenced resources to be attached as long as they allow it. Exceptions to that would include conflicts or being referenced from a namespace or kind that isn't allowed. I think we need to be clear that this expresses a desire to attach to the given resource, not just permission for a resource to be attached.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: If I'm interpreting this correctly, the current votes for setting Gateway as default here are:

@youngnick: No
@Miciah: Leaning yes
@hbagdi: Yes
@howardjohn: Fine either way
@jpeach: Leaning no

If I've interpreted that correctly, it seems like we're effectively tied. Fortunately we can change defaults post-v1alpha2 release. Although I initially omitted a default here, I'm leaning towards adding one here since it does represent 99% of use cases. If anyone feels strongly that we should leave out a default, I'm happy to leave it off though.

site-src/geps/gep-724.md
robscott: Will update to a pointer, thanks!

As far as ObjectFieldSelector ([ref for anyone curious](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#objectfieldselector-v1-core)), I'm not sure it's quite what we're looking for here. As I understand it, it allows us to select a single field within a resource. That would allow for the valid `spec.listeners[1]`, but lots of invalid values as well (someone may select some other part of a Gateway that may not make any sense). 

Even the valid example of `spec.listeners[1]` has issues in that it would be tied to an index and not a name. That would require Gateway admins to maintain a very specific ordering of their listeners. If that order ever switched, lots of things could break. A similar argument could be made for changing a listener name, but that seems both more manageable and expected.

If I'm understanding the [relevant API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#selecting-fields) correctly, this field should be interpreted with standard JavaScript syntax, so I don't think we'd even get the flexibility of jsonpath here.

site-src/geps/gep-724.md
robscott: Yeah the more I've thought about this, the more I think 3 is the best approach. That means Route->Route would work nearly identically to Route->Service, which seems reasonable to me.

site-src/geps/gep-724.md
robscott: Thanks @youngnick, I think that's a good distinction. We could expand status for Gateway listeners so controllers could populate the kinds of Routes they were supporting for that listener. I've got some larger thoughts around status that will require a bigger update, so I'll work on that in a subsequent commit. @hbagdi would a status based approach work for you in terms of improving debuggability?

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Thanks for the great summary @youngnick! I think we're all leaning in the same direction of not trying to be too prescriptive about how Route inclusion could work, but instead leave this as one potential approach.  I've updated this line in the GEP to read "as part of one potential approach to Route inclusion".

site-src/geps/gep-724.md
robscott: That's a good question, I think my primary concern was user complexity, added some more clarification here.

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,624 @@
...
+  // routes to associate with the Gateway. If this Selector is defined,
+  // only routes matching the Selector are associated with the Gateway.
+  // An empty Selector matches all routes.
+  //
+  // Support: Core
+  //
+  // +optional
+  Selector *metav1.LabelSelector `json:"selector,omitempty"`
+	// Group is the group of the route resource to select. Omitting the value or specifying
+	// the empty string indicates the networking.x-k8s.io API group.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Good catch, thanks! I'm leaving this as is though because it denotes code that we're removing.

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Added some clarification, let me know what you think.

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: 🤦 woops! Should be updated now.

site-src/geps/gep-724.md
robscott: Updated, thanks!

site-src/geps/gep-724.md
robscott: Added, thanks!

site-src/geps/gep-724.md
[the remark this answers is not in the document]
jpeach: I can't figure out what this thread refers to any more :joy:

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,636 @@
...
+  // resource type, it SHOULD set the "ResolvedRefs" condition to false for
+  // this listener with the "InvalidRoutesRef" reason.
+  //
+  // Support: Core
+  Kind string `json:"kind"`
+```
+
+#### Added
+```go
+type ListenerRoutes struct {
```
jpeach, line 235: Nit: this would be set on the listener `Routes` field, which implies the following addition:

```Go
type Listener struct {
...
Routes ListenerRoutes
}
```

Or is this being added to new fields in `RouteSelector`?

jpeach:
Overall, this is in good shape. Fantastic work by @robscott to integrate the deluge of feedback :)

/lgtm

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Haha this is in the context of `route.spec.parentRefs.kind` and the question is if we should default it to "Gateway". It seems like no one feels overly strongly on this one, but if anyone does I'm happy to go that way. At present, I'm leaning towards leaving it with a "Gateway" default because that represents such a large majority of use cases.

site-src/geps/gep-724.md
robscott: Yeah this was confusing, I added some text for this + a related change to describe where they would fit in.

site-src/geps/gep-724.md
robscott: Added a new status based approach as described above.

site-src/geps/gep-724.md
```diff
@@ -0,0 +1,698 @@
...
+  //
+  // Support: Core
+  //
+  // +kubebuilder:validation:MinLength=1
+  // +kubebuilder:validation:MaxLength=253
+  Kind string `json:"kind"`
+}
+```
+
+### Gateway Status
```
robscott, line 299: @jpeach @youngnick @hbagdi I've added a new status section to describe suggested additions to status. I think this is relevant to different comments from all of you throughout the review of this GEP, so wanted to get your feedback on this addition if you have time.

youngnick: This is great, love it.

youngnick (commented):
Status changes are excellent, as are the others.

/lgtm

site-src/geps/gep-724.md
Miciah: What I had in mind was that `ParentRef` would have a field like `SectionFieldRef corev1.ObjectFieldSelector` in addition to the `SectionName string` field,  and the user would specify `sectionFieldRef: spec.listeners` along with `sectionName`.  The API could require that the field identified by `sectionFieldRef` be a `+listType=map`, and then `sectionName` would specify the value to match against the field identified by `+listMapKey`: 
```go
type GatewaySpec struct {
	// ...
	// +listType=map
	// +listMapKey=name
	Listeners []Listener `json:"listeners"`

	// ...
}
type ParentRef struct {
	// ...

	// SectionFieldRef specifies a field containing sections within the
	// referent.  The specified field must be listType=map.  If
	// SectionFieldRef is specified, then SectionName must be specified as
	// well, and the value of SectionName will be used to identify the
	// specific section by matching it against the field specified by
	// listTypeMap.
	//
	// Support: Core
	//
	// +optional
	SectionFieldRef corev1.ObjectFieldSelector `json:"sectionFieldRef,omitempty"`

	// SectionName is ...
	SectionName string `json:"sectionName,omitempty"`

	// ...
}
```
Example:
```yaml
kind: Gateway
metadata:
  namespace: infra
  name: lb
spec:
  listeners:
  - name: foo
    # ...
---
kind: HTTPRoute
# ...
spec:
  attachTo:
  - kind: Gateway
    namespace: infra
    name: lb
    sectionFieldRef.fieldPath: spec.listeners
    sectionName: foo
```

Miciah: `ObjectFieldSelector` might not be the best type (maybe it would be better to define a new  type that includes the field path and the item name), but the idea is to make explicit what `SectionName` is identifying.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
Miciah: I'm thinking of route-to-route.  Maybe I want to add the same microservice to many different applications.  I don't have a concrete use-case though, so I don't feel too strongly about this.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
Miciah: Sorry, I should have been more clear.  My question was about this statement in particular: "It is invalid to reference the same parent resource more than once."  If the implementation does merge two listeners, and one route is attached to both listeners, should that be treated as the route's referencing the same parent resource more than once and thus be considered invalid?

site-src/geps/gep-724.md
robscott: Thanks for the explanation @Miciah! That does make a lot more sense. I can see that being a natural extension point if/when we want to support attaching to other arbitrary fields in Gateway or elsewhere. If I'm understanding correctly, it looks like it would be easy to add as a new optional field in the future without any kind of breaking changes.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Yeah I think it's unlikely that we'll use this mechanism for attaching to a Route, but since we don't have a GEP for that yet, we can't know for sure. At least with values like this, we can always increase the number of items allowed, nearly impossible to decrease in the future.

site-src/geps/gep-724.md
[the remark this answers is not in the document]
robscott: Oh wow, not sure how I missed that. The spec I had here was very unclear. I've added some clarification that it should be valid to attach a Route to multiple sections/listeners within a Parent/Gateway.

robscott:
Thanks to everyone for the great feedback on this GEP! Now that it's gotten 3 different LGTMs, I'm going to remove the hold and hope for one final LGTM to get it in. My last update clarifies that it is valid to attach a Route to multiple sections/listeners within the same Parent/Gateway. I'm thinking we should be able to cover any additional tweaks to this GEP in the implementation PR if necessary.

/hold cancel

hbagdi:
Impressive heavy lifting here.
Thanks @robscott.
/lgtm

site-src/geps/images/724-alt1.png
jsoref, line 1: @robscott what font did you use for this / is there an original? I ask because `ReferencePolicy` was renamed but these two pictures weren't updated which left some of the documentation not making much sense.

For this picture, that's:
https://github.com/kubernetes-sigs/gateway-api/blob/9f820af97aaf6e8587712fec7b9266f42da81258/geps/gep-724/index.md#1-referencegrant-with-gateways-selecting-routes

site-src/geps/images/724-alt2.png
jsoref, line 1: For this picture, that's: https://github.com/kubernetes-sigs/gateway-api/blob/9f820af97aaf6e8587712fec7b9266f42da81258/geps/gep-724/index.md#2-referencegrant-with-routes-referencing-gateways

PR #729, opened 2021-07-20
Adding suggested reasons for when conditions are healthy

robscott:
**What type of PR is this?**
/kind cleanup

This PR adds reasons for when conditions are healthy. When the conditions are positive, the healthy reason is identical to the condition name. 

```release-note
Guidance has been added for condition reasons when they are healthy.
```

Slack thread with a bit more discussion: https://kubernetes.slack.com/archives/CR0H13KGA/p1616184252018300

/cc @jpeach @howardjohn @lelenanam

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -576,14 +576,18 @@ type GatewayStatus struct {
 // field.
 type GatewayConditionType string
 
-// GatewayConditionReason defines the set of reasons that explain
-// why a particular Gateway condition type has been raised.
+// GatewayConditionReason defines the set of reasons that explain why a
+// particular Gateway condition type has been raised.
 type GatewayConditionReason string
 
 const (
 	// This condition is true when the controller managing the
 	// Gateway has scheduled the Gateway to the underlying network
 	// infrastructure.
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Scheduled"
 	//
 	// Possible reasons for this condition to be false are:
 	//
@@ -595,6 +599,10 @@ const (
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
 	GatewayConditionScheduled GatewayConditionType = "Scheduled"
+
+	// This reason is used with the "Scheduled" condition when the condition is
+	// true.
+	GatewayReasonScheduled GatewayConditionReason = "Scheduled"
 
 	// This reason is used with the "Scheduled" condition when
 	// been recently created and no controller has reconciled it yet.
@@ -624,6 +632,10 @@ const (
 	// reasons are true, the Gateway controller should prefer the
 	// "ListenersNotValid" reason.
 	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Ready"
+	//
 	// Possible reasons for this condition to be false are:
 	//
 	// * "ListenersNotValid"
@@ -634,6 +646,10 @@ const (
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
 	GatewayConditionReady GatewayConditionType = "Ready"
+
+	// This reason is used with the "Ready" condition when the condition is
+	// true.
+	GatewayReasonReady GatewayConditionReason = "Ready"
 
 	// This reason is used with the "Ready" condition when one or
 	// more Listeners have an invalid or unsupported configuration
@@ -698,6 +714,10 @@ const (
 	// * "ProtocolConflict"
 	// * "RouteConflict"
 	//
+	// Possible reasons for this condition to be false are:
+	//
+	// * "NoConflicts"
+	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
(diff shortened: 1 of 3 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -688,9 +691,9 @@ type ListenerConditionReason string
 
 const (
 	// This condition indicates that the controller was unable to resolve
-	// conflicting specification requirements for this Listener. If a
-	// Listener is conflicted, its network port should not be configured
-	// on any network elements.
+	// conflicting specification requirements for this Listener. If a Listener
+	// is conflicted, its network port should not be configured on any network
+	// elements. This condition MUST only be present when it is true.
```
howardjohn, line 696: So its illegal to have `False`? or am I parsing this wrong?

robscott: Yeah, I can clarify this, but the rationale was that `Conflicted=False` is a generally confusing state to have. I think it may be clearer to not have a condition in this case.

jpeach: IIRC, the general guidance for Conditions is that they are always published?

robscott: That's a good point, somehow I'd thought that guidance was limited to positive conditions. Taking another look at the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) suggests that conditions should always be set:

> Condition status values may be True, False, or Unknown. The absence of a condition should be interpreted the same as Unknown. 

I'll update this to instead provide custom reasons for healthy states of `Conflicted` and `Detached`.

robscott:
Thanks for the feedback @howardjohn and @jpeach! Refactored this to just add reasons for all healthy condition states. PTAL.

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -132,36 +132,48 @@ type ParametersReference struct {
...
+	// * "InvalidParameters"
+	// * "Waiting"
 	//
 	// Controllers should prefer to use the values of GatewayClassConditionReason
 	// for the corresponding Reason, where appropriate.
 	GatewayClassConditionStatusAdmitted GatewayClassConditionType = "Admitted"
 
+	// This reason is used with the "Admitted" condition when the condition is
+	// true.
+	GatewayClassReasonStatusAdmitted GatewayClassConditionReason = "Admitted"
```
jpeach, line 165: Should be `GatewayClassReasonAdmitted`?

stevesloka (commented):
I'm a bit confused as to what conditions should always be on an object and when some can come and go. For instanced "Scheduled" should always be there (which still to me I don't follow completely). But you'd also have "Ready" all the time. 

When would other types like "Conflicted" or "Detached" be set? Should they always be there and in the happy path they are in the "valid" state?

I'm mixing the Gateway & Listener status a bit here I understand, but not sure if we should have these set all the time.

apis/v1alpha2/gateway_types.go
```diff
@@ -764,12 +792,20 @@ const (
 	// the Listener could not be attached to the Gateway because the
 	// requested address is not supported.
 	ListenerReasonUnsupportedAddress ListenerConditionReason = "UnsupportedAddress"
+
+	// This reason is used with the "Detached" condition when the condition is
+	// false.
+	ListenerReasonAttached ListenerConditionReason = "Attached"
```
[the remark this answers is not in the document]
robscott: Yeah that was the intent, the reason would be "Attached" any time Detached was false. I agree it's rather confusing. My initial proposal was that these negative conditions simply shouldn't be set when they're false. The [API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) do state that unset == unknown though, which is not really what we'd be trying to communicate if we left these unset.

apis/v1alpha2/gateway_types.go
```diff
@@ -720,6 +740,10 @@ const (
 	// For example, a Listener that specifies "UDP" as the protocol
 	// but a route selector that resolves "TCPRoute" objects.
 	ListenerReasonRouteConflict ListenerConditionReason = "RouteConflict"
+
+	// This reason is used with the "Conflicted" condition when the condition
+	// is false.
+	ListenerReasonNoConflicts ListenerConditionReason = "NoConflicts"
```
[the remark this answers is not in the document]
robscott: Yep, same rationale here.

jpeach:
> I'm a bit confused as to what conditions should always be on an object and when some can come and go. For instanced "Scheduled" should always be there (which still to me I don't follow completely). But you'd also have "Ready" all the time.

Yeh, the API conventions recommend that all conditions be present at times, so that observers can canonically know their values. I think that this makes more sense for programmatic observers than for humans, but at least it's consistent :)

> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.

jpeach (approved):
LGTM, but holding off merging while there's still some discussion.

@robscott Did you want to add any reasons for  `ConditionRouteAdmitted`?

robscott:
> Did you want to add any reasons for ConditionRouteAdmitted?

Thanks for catching that! I completely missed that one, added a couple conditions now.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
stevesloka: Ok that's what I thought that they always need to be there, thanks for clarifying. 

I wonder if we could document this somehow, so that it's clear all the different conditions should be present on each type at all times.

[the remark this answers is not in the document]
stevesloka: Or do they get set by default when the object is created? That would be super cool, then you wouldn't need to deal with the initial state from the controller.

jpeach (approved):
/lgtm

Issue #623, opened 2021-04-22
Conflicting SNI's between HTTPRoute & TLSRoute


stevesloka:
**What happened**:
If I have the same `fqdn` defined in both an `HTTPRoute` as well as a `TLSRoute`, what could the conflict resolution be in that case? 

The current docs on a `TLSRouteMatch` sort of lead me to think that TLSRoutes should match first before HTTPRoutes, but might be nice to raise a Condition for this in the event that one is ignored over another. 

Just curious if others have thought about this yet or if I'm maybe misunderstanding the spec. =)

hbagdi:
At any time, a listener on the Gateway can only select routes of one `kind`, so conflicts shouldn't happen between route of different types.
I'm curious how you are running into this.

stevesloka:
I thought you could duplicate listener ports in a Gateway and then the controller should collapse them:

```
If the implementation does collapse compatible Listeners, the hostname provided in the incoming client request MUST be matched to a Listener to find the correct set of Routes. The incoming hostname MUST be matched using the Hostname field for each Listener in order of most to least specific. That is, exact matches must be processed before wildcard matches.
```

I guess we can apply the same conflict logic within the same type (e.g. HTTPRoutes), but do across types which would make the behavior consistent in a way, but would need to raise a condition on the object to alert the user.

// ref: https://gateway-api.sigs.k8s.io/references/spec/#networking.x-k8s.io/v1alpha1.Gateway

hbagdi:
I see the problem.

```
 87     // 1. Either each Listener within the group specifies the "HTTP"            
 88     //    Protocol or each Listener within the group specifies either                                                                                                                                          
 89     //    the "HTTPS" or "TLS" Protocol.
```

I think the `or` in ` "HTTPS" or "TLS"` is the reason for the confusion.
I could be wrong here but I think multiplexing HTTPS and TLS routing configuration on the same port isn't something we have discussed before. Collapsing should really happen only for listeners which have the same protocol type.
I don't think many implementations would be able to support such a configuration. Is that something you already support (or even plan to support)?

/cc @robscott @jpeach @Miciah

stevesloka:
> Is that something you already support (or even plan to support)?

Today Contour supports a single insecure/secure listener, so all traffic routes through those two listeners essentially. Someone could potentially create a TLS route with the same fqdn as an HTTPRoute and you'd have a conflict. We handle this today by only supporting one or the other if users create them on the same fqdn.

[...]

PR #730, opened 2021-07-22
Fix TLS guide with TLSRoute can be terminated

mike1808:
Previously guide stated that TLSRoutes cannot be terminated on the gateway, however, there is no requirement that the proxy to the backend hop has to be TLS when using TLSRoute. As it's about matching not the traffic protocol.

**What type of PR is this?**

/kind documentation

Fix TLS guide with info that TLSRoutes can be used with TLS terminated at the gateway

N/A

No

### DIFF
--- a/site-src/guides/tls.md
+++ b/site-src/guides/tls.md
@@ -22,6 +22,7 @@ Depending on the Listener Protocol, different TLS modes and Route types are supp
 Listener Protocol | TLS Mode | Route Type Supported
 --- | --- | ---
 TLS | Passthrough | TLSRoute
+TLS | Terminate | TLSRoute
 TLS | Terminate | TCPRoute
 HTTPS | Terminate | HTTPRoute
 


mike1808:
@k8s-ci-robot fixed my email in the commit as I already singed SLA

hbagdi:
Yes, this seems correct.
TLSRoute can support Passthrough or Terminate, the downstream TLS behavior is configured in the listener and not the route.

/lgtm

Will leave approval to another pair of eyes.

howardjohn:
/hold

I have some concerns, typing them up

howardjohn:
I agree this change is implementable and solves a real problem. However, I do have some concerns.

Currently, there are essentially 3 protocols in play:
The downstream connection, the protocol we use for routing internally, and the upstream connection.


Often these are the same for all three - sometimes they are different. For example, we could have TLS, which terminated gives us HTTP for routing, and forward QUIC. Or we could just have HTTP for all three.

Currently, the Route type is determining the middle one (Ill call it "routing protocol"). This PR changes that - the middle protocol is TCP but we are controlling it with a TLSRoute.

This seems even more concerning if we think of HTTP. Lets say I want to Terminate HTTPS traffic, then I want a route to match on the X-My-Header AND the SNI value - so I am matching on TLS properties *and* HTTP properties. Is this something we want to support? If so, does that mean every route needs to contain the entirety of TLSRoute match (which is small today but can grow quite a bit)? If not, then is it ok that routing raw TLS traffic is more capable than HTTPS traffic?

hbagdi:
> Currently, there are essentially 3 protocols in play:
The downstream connection, the protocol we use for routing internally, and the upstream connection.

> Often these are the same for all three - sometimes they are different. For example, we could have TLS, which terminated gives us HTTP for routing, and forward QUIC. Or we could just have HTTP for all three.

In this case, the Gateway terminating protocol and the matching protocol is the same = TLS.

>  This PR changes that - the middle protocol is TCP but we are controlling it with a TLSRoute.

My understanding was that TLSRoute matching in case of "Terminate" will be based on the TLS handshake that is terminating at the Gateway and not the TCP bytestream within it.

howardjohn:
> My understanding was that TLSRoute matching in case of "Terminate" will be based on the TLS handshake that is terminating at the Gateway and not the TCP bytestream within it.

Right, my point is that changes the current semantics and introduces inconsistencies with HTTP(S)

jpeach:
> > My understanding was that TLSRoute matching in case of "Terminate" will be based on the TLS handshake that is terminating at the Gateway and not the TCP bytestream within it.
> 
> Right, my point is that changes the current semantics and introduces inconsistencies with HTTP(S)

TLSRoute has never said anything about the protocol used to reach the backend service, it's all about matching on TLS-level metadata. The BackendPolicy and service AppProtocol is what ttriggers TLS to the backend service.

> Lets say I want to Terminate HTTPS traffic, then I want a route to match on the X-My-Header AND the SNI value - so I am matching on TLS properties and HTTP properties. 

This is supported since we recommend that SNI and HTTP authorities match (so the SNI match is implicit I suppose). I don't know of an in-principle objections to allowing TLS-layer match criteria in HTTP routes, but I haven't seen a proposal either. Maybe implementation experience with a custom match type would be interesting.

robscott:
Relevant to https://github.com/kubernetes-sigs/gateway-api/issues/623

Issue #726, opened 2021-07-19
GEP: Add Path Redirects and Rewrites


robscott:
**What would you like to be added**:
This GEP attempts to cover 2 related but different features:

* Path Redirects
* Rewrites

Once this GEP is complete, HTTPRoute should include filters that enable basic path and hostname redirects and rewrites. Although there is definitely variation among implementations as far as what is possible, there appears to be enough common ground for a portable API.

**Why is this needed**:
This is a common feature request, as noted by https://github.com/kubernetes-sigs/gateway-api/issues/200 and https://github.com/kubernetes-sigs/gateway-api/issues/678.

dacloutier:
How can I go about requesting a change to the gateway-api to support regex/capture group Path rewrites?

Istio won't merge the PR that would support this until it gets added to the gateway api.... and many people ( including me ;D ) need regex path and capture group rewrites. (see: https://github.com/istio/istio/issues/22290)

The feature is discussed in https://github.com/kubernetes-sigs/gateway-api/pull/731 but only mentions it as a "Future extension"

Thanks for helping me get some movement on this.

youngnick:
@dacloutier, thanks for the feedback.

I understand this feature is important to you and a lot of people, and the reason we've left it for a future extension in #731 is that it's actually pretty complicated to write a standard way for it to work, that will work across multiple proxies/Gateway implementations, and still allow you to migrate your config between Gateway implementations (which is a primary goal of this project).

We definitely plan to get to this, but need to make sure that whatever will do will be usable to a broad base of people. Just as a quick example, it's very common for different implementations to use different regex flavors, but for the Gateway API, we need to be able to do something that will work across as many as possible.

Also, what we've found up until this point is that some things that you would think are pretty basic functionality can vary wildly between implementations and make it very difficult for there to be a single, standard way to implement them. Timeouts for HTTP connections are a great example of this - every implementation has there own set of timeouts that mean subtly different things, and we haven't been able to build a good common set to be able to make a HTTPRoute TimeoutPolicy.

Given that regex is already pretty complicated, and there are different flavors, *and* different implementations, we need to be a very careful with the design. 

I know "we're going to get to it" is a _very_ unsatisfying answer, but unfortunately, that's all we have at the moment. I certainly think that your enthusiasm here does a great job of pushing it up my priority list. :)

I should also say that, as always, contributions are welcomed - often the hardest part is getting the discussion started!

rufreakde:
[...]

PR #731, opened 2021-07-22
Adding GEP 726: Path Rewrites and Redirects

robscott:
**What type of PR is this?**
/kind gep

This adds GEP #726, a proposed expansion of redirect capabilities along with new rewrite capabilities. 

```release-note
NONE
```

Deploy Preview: https://deploy-preview-731--kubernetes-sigs-gateway-api.netlify.app/geps/gep-726/

### DIFF
--- a/site-src/geps/gep-726.md
+++ b/site-src/geps/gep-726.md
@@ -42,20 +42,20 @@ type HTTPPathModifierType string
 const (
   // This type of modifier indicates that the complete path will be replaced by
   // the path redirect value.
-  AbsoluteHTTPPathRedirect HTTPPathModifierType = "Absolute"
+  AbsoluteHTTPPathModifier HTTPPathModifierType = "Absolute"
 
   // This type of modifier indicates that any prefix path matches will be
   // replaced by the substitution value. For example, a path with a prefix match
-  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
+  // of "/foo" and a ReplacePrefixMatch substitution of "/bar" will have the
   // "/foo" prefix replaced with "/bar" in matching requests.
-  PathMatchHTTPPathRedirect HTTPPathModifierType = "ReplacePrefixPathMatch"
+  PrefixMatchHTTPPathModifier HTTPPathModifierType = "ReplacePrefixMatch"
 )
 
 // HTTPPathModifier defines configuration for path modifiers.
 type HTTPPathModifier struct {
   // Type defines the type of path modifier.
   //
-  // +kubebuilder:validation:Enum=Absolute;ReplacePrefixPathMatch
+  // +kubebuilder:validation:Enum=Absolute;ReplacePrefixMatch
   Type HTTPPathModifierType `json:"type"`
 
   // Substitution defines the HTTP path value to substitute. An empty value ("")
@@ -90,11 +90,12 @@ spec:
         requestRedirect:
           hostname: foo.com
           path:
-            type: ReplacePrefixPathMatch
+            type: ReplacePrefixMatch
             value: /bar
 ```
 
-This would be represented with the following API additions:
+This would be represented with the following API addition to the existing
+HTTPRequestRedirect filter:
 ```go
 // HTTPRequestRedirect defines a filter that redirects a request. At most one of
 // these filters may be used on a Route rule. This may not be used on the same
@@ -114,8 +115,8 @@ type HTTPRequestRedirect struct {
 
 ### Rewrites
 
-A new `RequestRewrite` filter can be added to support rewrites. In the following
-example, a a request to `example.com/foo/abc` would be rewritten to
+A new `URLRewrite` filter can be added to support rewrites. In the following
+example, a request to `example.com/foo/abc` would be rewritten to
 `example.net/bar/abc`.
 
 ```yaml
@@ -132,22 +133,22 @@ spec:
           type: Prefix
           value: /foo
       filters:
-      - type: RequestRewrite
+      - type: URLRewrite
         requestRewrite:
           hostname: example.net
           path:
-            type: ReplacePrefixPathMatch
+            type: ReplacePrefixMatch
             substitution: /bar
 ```
 
 This would be represent with the following API additions:
 ```go
-// HTTPRequestRewrite defines a filter that modifies a request during
-// forwarding. At most one of these filters may be used on a Route rule. This
-// may not be used on the same Route rule as a HTTPRequestRedirect filter.
+// HTTPURLRewrite defines a filter that modifies a request during forwarding.
+// At most one of these filters may be used on a Route rule. This may not be
+// used on the same Route rule as a HTTPRequestRedirect filter.
 //
 // Support: Extended
-type HTTPRequestRewrite struct {
+type HTTPURLRewrite struct {
   // Hostname is the value to be used to replace the Host header value during
   // forwarding.
   //
@@ -165,6 +166,10 @@ type HTTPRequestRewrite struct {
   Path *HTTPPathModifier `json:"path,omitempty"`
 }
 ```
+
+Note: `RequestRewrite` was originally considered as a name for this filter.
+`URLRewrite` was chosen as it more clearly represented the capabilities of the
+filter and would not be confused with header or query param modification.
 
 ## Portability
 
@@ -265,7 +270,7 @@ from the initial implementation.
 ## Alternatives
 
 ### 1. Generic Path Match Replacement
-Instead of the `ReplacePrefixPathMatch` option proposed above, we could have a
+Instead of the `ReplacePrefixMatch` option proposed above, we could have a
 `ReplacePathMatch` option. This would provide significantly more flexibility and
 room for growth than prefix replacement.
 
--- a/site-src/geps/gep-726.md
+++ b/site-src/geps/gep-726.md
@@ -31,6 +31,43 @@ there is significantly more variation in both if and how they are implemented.
 Given the wide support for this concept, it is important to design the API in a
 way that would make it easy to add this capability in the future.
 
+### Path Modifiers
+
+Both redirects and rewrites would share the same `PathModifier` types:
+
+```go
+// HTTPPathModifierType defines the type of path redirect.
+type HTTPPathModifierType string
+
+const (
+  // This type of modifier indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathRedirect HTTPPathModifierType = "Absolute"
+
+  // This type of modifier indicates that any prefix path matches will be
+  // replaced by the substitution value. For example, a path with a prefix match
+  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
+  // "/foo" prefix replaced with "/bar" in matching requests.
+  PathMatchHTTPPathRedirect HTTPPathModifierType = "ReplacePrefixPathMatch"
+)
+
+// HTTPPathModifier defines configuration for path modifiers.
+type HTTPPathModifier struct {
+  // Type defines the type of path modifier.
+  //
+  // +kubebuilder:validation:Enum=Absolute;ReplacePrefixPathMatch
+  Type HTTPPathModifierType `json:"type"`
+
+  // Substitution defines the HTTP path value to substitute. An empty value ("")
+  // indicates that the portion of the path to be changed should be removed from
+  // the resulting path. For example, a request to "/foo/bar" with a prefix
+  // match of "/foo" would be modified to "/bar".
+  //
+  // +kubebuilder:validation:MaxLength=1024
+  Substitution string `json:"substitution"`
+}
+```
+
 ### Redirects
 
 The existing `RequestRedirect` filter can be expanded to support path redirects.
@@ -70,39 +107,8 @@ type HTTPRequestRedirect struct {
   // Support: Extended
   //
   // +optional
-  Path *HTTPPathRedirect `json:"path,omitempty"`
+  Path *HTTPPathModifier `json:"path,omitempty"`
   // ...
-}
-
-// HTTPPathRedirectType defines the type of path redirect.
-type HTTPPathRedirectType string
-
-const (
-  // This type of redirect indicates that the complete path will be replaced by
-  // the path redirect value.
-  AbsoluteHTTPPathRedirect HTTPPathRedirectType = "Absolute"
-
-  // This type of redirect indicates that any prefix path matches will be
-  // replaced by the substitution value. For example, a path with a prefix match
-  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
-  // "/foo" prefix replaced with "/bar" in matching requests.
-  PathMatchHTTPPathRedirect HTTPPathRedirectType = "ReplacePrefixPathMatch"
-)
-
-// HTTPPathRedirect defines configuration for path redirects.
-type HTTPPathRedirect struct {
-  // Type defines the type of path redirect.
-  //
-  // +kubebuilder:validation:Enum=Absolute;ReplacePrefixPathMatch
-  Type HTTPPathRedirectType `json:"type"`
-
-  // Value defines the HTTP path value to redirect to. An empty value ("")
-  // indicates that the portion of the path to be changed in the redirect should
-  // be removed from the redirect path. For example, a request to "/foo/bar"
-  // with a prefix match of "/foo" would be redirected to "/bar".
-  //
-  // +kubebuilder:validation:MaxLength=1024
-  Substitution string `json:"substitution"`
 }
 ```
 
@@ -156,34 +162,7 @@ type HTTPRequestRewrite struct {
   // Support: Extended
   //
   // +optional
-  Path *HTTPPathRewrite `json:"path,omitempty"`
-}
-
-// HTTPPathRewriteType defines the type of path rewrite.
-type HTTPPathRewriteType string
-
-const (
-  // This type of rewrite indicates that any prefix path matches will be
-  // replaced by the substitution value. For example, a path with a prefix match
-  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
-  // "/foo" prefix replaced with "/bar" in matching requests.
-  PathMatchHTTPPathRedirect HTTPPathRedirectType = "ReplacePrefixPathMatch"
-)
-
-// HTTPPathRewrite defines configuration for path rewrites.
-type HTTPPathRewrite struct {
-  // Type defines the type of path rewrite.
-  //
-  // +kubebuilder:validation:Enum=ReplacePrefixPathMatch
-  Type HTTPPathRewriteType `json:"type"`
-
-  // Substitution defines the HTTP path value to rewrite. An empty value ("")
-  // indicates that the portion of the path to be rewritten should be removed
-  // from the rewritten path. For example, a request to "/foo/bar" with a prefix
-  // match of "/foo" would be rewritten to "/bar".
-  //
-  // +kubebuilder:validation:MaxLength=1024
-  Substitution string `json:"substitution"`
+  Path *HTTPPathModifier `json:"path,omitempty"`
 }
 ```
 
@@ -255,7 +234,7 @@ covered by this proposal:
 
 Both of the following can be represented by adding a new field new types. For
 example, this config would result in a request to `/foo/baz` to be rewritten to
-`/foo/bar`:
+`/bar/baz`:
 
 ```yaml
 filters:
--- a/site-src/geps/gep-726.md
+++ b/site-src/geps/gep-726.md
@@ -1,0 +1,274 @@
+# GEP-726: Add Path Redirects and Rewrites
+
+* Issue: [#726](https://github.com/kubernetes-sigs/gateway-api/issues/726)
+* Status: Implementable
+
+## TLDR
+
+This GEP proposes adding support for path redirects and rewrites in addition to
+host rewrites. This would augment the existing host redirection capabilities.
+
+## Goals
+
+* Implement path redirects.
+* Implement the most portable and simple forms of path rewrites.
+* Describe how more advanced rewrite and redirect and redirect capabilities
+  could be added in the future.
+
+## API
+
+Although many implementations support very advanced rewrite and redirect
+capabilities, the following are the most [portable](#portability) concepts that
+are not already supported by the Gateway API:
+
+* Path redirects
+* Path prefix redirects
+* Path prefix rewrites
+* Host rewrites
+
+Although regular expression based redirects and rewrites are commonly supported,
+there is significantly more variation in both if and how they are implemented.
+Given the wide support for this concept, it is important to design the API in a
+way that would make it easy to add this capability in the future.
+
+### Redirects
+
+The existing `RequestRedirect` filter can be expanded to support path redirects.
+In the following example, a request to `/foo/abc` would be redirected to
+`/bar/abc`.
+
+```yaml
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1alpha2
+metadata:
+  name: http-filter-1
+spec:
+  rules:
+    - matches:
+      - path:
+          type: Prefix
+          value: /foo
+      filters:
+      - type: RequestRedirect
+        requestRedirect:
+          hostname: foo.com
+          path:
+            type: PathMatch
+            value: /bar
+```
+
+This would be represented with the following API additions:
+```go
+// HTTPRequestRedirect defines a filter that redirects a request. At most one of
+// these filters may be used on a Route rule. This may not be used on the same
+// Route rule as a HTTPRequestRewrite filter.
+type HTTPRequestRedirect struct {
+  // Path defines a path redirect.
+  //
+  // Support: Extended
+  //
+  // +optional
+  Path *HTTPPathRedirect `json:"path,omitempty"`
+  // ...
+}
+
+// HTTPPathRedirectType defines the type of path redirect.
+type HTTPPathRedirectType string
+
+const (
+  // This type of redirect indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathRedirect HTTPPathRedirectType = "Absolute"
+
+  // This type of redirect indicates that the path match will be replaced by
+  // the path redirect value.
+  PathMatchHTTPPathRedirect HTTPPathRedirectType = "PathMatch"
+)
+
+// HTTPPathRedirect defines configuration for path redirects.
+type HTTPPathRedirect struct {
+  // Type defines the type of path redirect.
+  //
+  // +kubebuilder:validation:Enum=Absolute;PathMatch
+  Type HTTPPathRedirectType `json:"type"`
+
+  // Value defines the HTTP path value to redirect to. An empty value ("")
+  // indicates that the portion of the path to be changed in the redirect should
+  // be removed from the redirect path. For example, a request to "/foo/bar"
+  // with a prefix match of "/foo" would be redirected to "/bar".
+  //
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+      - path:
+          type: Prefix
+          value: /
+      filters:
+      - type: RequestRedirect
+        requestRedirect:
+          hostname: foo.com
+          path:
+            type: Prefix
+            value: /bar
```
jpeach, line 55: Suggest that we use something like `/app` as the matching prefix. Then we can give a clear example of exactly what redirect location header is generated.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+const (
+  // This type of redirect indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathRedirect HTTPPathRedirectType = "Absolute"
+
+  // This type of redirect indicates that the prefix match will be replaced by
+  // the path redirect value. This type only applies to path matches of type
+  // Prefix. If a Route has other match types, this redirect MUST NOT be applied
+  // to them.
+  PrefixHTTPPathRedirect HTTPPathRedirectType = "Prefix"
```
jpeach, line 85: So, for example,

* the request path is `/api/foo/bar
* and the route is a prefix match on `/api`
* and there is a prefix redirect filter with a path of `/overthere`

The final location path would be `/overthere/foo/bar`?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+  // ...
+}
+
+// HTTPPathRedirectType defines the type of path redirect.
+type HTTPPathRedirectType string
+
+const (
+  // This type of redirect indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathRedirect HTTPPathRedirectType = "Absolute"
```
jpeach, line 79: I'm wondering whether these options are unnecessarily constrained? There's really 2 cases

1. Emit the entire location path
2. Generate the location path by replacing the matched portion

In case (2), the matched portion could be a prefix, or a regex, or implementation-defined.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+// forwarding. At most one of these filters may be used on a Route rule. This
+// may not be used on the same Route rule as a HTTPRequestRedirect filter.
+type HTTPRequestRewrite struct {
+  // Hostname is the value to be used to replace the Host header value during
+  // forwarding.
+  //
+  // Support: Extended
+  //
+  // +optional
+  // +kubebuilder:validation:MaxLength=255
```
jpeach, line 138: MinLength=3 (for "h.c" as the minimum hostname)

howardjohn: Hostname should probably be valid without `.` I think? Like `kubernetes` (when in the default namespace) is valid.

howardjohn: Its probably a bad idea in many cases, but not sure it should be rejected?

robscott:
Missed adding a hold on this earlier for further discussion. Will remove if/when it appears we have some consensus.

/hold

site-src/geps/gep-726.md
robscott: Yep, that's exactly the goal, will work on a better explanation for this.

site-src/geps/gep-726.md
robscott: On 2 I agree that there's room for other kinds of replacement, but I don't think those are quite as portable. I wanted to start with prefix because that seems to be the most universally supportable concept.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+  Hostname *string `json:"hostname,omitempty"`
```
jpeach, line 139: So another approach is to specify that the header filter can set the ':authority` [pseudo header](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3), and the controller can program the underlying proxy to do the right thing?

Perhaps mention this in alternatives and reject it on the basis that this is more direct and explicit?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+
+// HTTPPathRewriteType defines the type of path rewrite.
+type HTTPPathRewriteType string
+
+const (
+  // This type of rewrite indicates that the prefix match will be replaced by
+  // the path rewrite value. This type only applies to path matches of type
+  // Prefix. If a Route has other match types, this rewrite MUST NOT be applied
+  // to them.
+  PrefixHTTPPathRewrite HTTPPathRewriteType = "Prefix"
```
jpeach, line 157: Same comment as for redirect, can we generalize this?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+type HTTPPathRewrite struct {
+  // Type defines the type of path rewrite.
+  //
+  // +kubebuilder:validation:Enum=Prefix
+  Type HTTPPathRewriteType `json:"type"`
+
+  // Value defines the HTTP path value to rewrite.
+  //
+  // +kubebuilder:validation:MaxLength=1024
+  Value HTTPPathRewriteType `json:"value"`
```
jpeach, line 170: Given this structure, how would we add a regex rewrite type? I think that we would need to add a whole new filter type. That's OK, but wanted to confirm your thinking in this area.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,266 @@
...
+  // to them.
+  PrefixHTTPPathRewrite HTTPPathRewriteType = "Prefix"
+)
+
+// HTTPPathRewrite defines configuration for path rewrites.
+type HTTPPathRewrite struct {
+  // Type defines the type of path rewrite.
+  //
+  // +kubebuilder:validation:Enum=Prefix
+  Type HTTPPathRewriteType `json:"type"`
```
jpeach, line 165: Does `Absolute` make sense here too? ie. rewrite `/app/foo/any/thing` to `/new/path`?

robscott: I think it would, it just wasn't quite as broadly supported. I guess any implementation that supports regex would also be able to accomplish this though.

site-src/geps/gep-726.md
robscott: Yeah I should spell that out. I was actually thinking of adding an optional `regex` or `pattern` field here that was only valid with certain rewrite types.

site-src/geps/gep-726.md
robscott: To clarify, are you suggesting using the same type for both rewrite and redirect?

kmai:
I believe this proposal can add a lot of value to enrich routing capabilities in Kubernetes.

Most gateway implementations don't have a normalized API and that makes interchangeability very complicated.

I think it's the best way to encourage modularity in api management tools, edge gateways and service meshes.

site-src/geps/gep-726.md
youngnick: I think it's important to allow single-label rewrites, sadly. People use them as an escape valve all the time.

youngnick (commented):
I hadn't taken a look at this yet, this is looking good on an initial read to me. I would guess it's coming post v1alpha2 though.

robscott:
Hey @dm3ch, thanks for checking in on this! We actually just discussed it at our community meeting today. It missed the cut for the initial v1alpha2 release (v0.4.0), but now that that's out, this is back in focus. I'm hoping to get this in for the next release of the API.

site-src/geps/gep-726.md
robscott: Updated to use `ReplacePathPrefixMatch`. Given @jpeach's comment below, I think `ReplacePathMatch` could be a simpler option long term, but we'd need to be very careful when it came to documenting support levels (ie some implementations can only support replacing prefix path matches). In any case, we'll also have to document how this is handled if multiple matches and/or match types are present.

site-src/geps/gep-726.md
robscott: Updated to use a clearer example for both this and rewrites.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,270 @@
...
+A new `RequestRewrite` filter can be added to support rewrites. In the following
+example, a a request to `example.com/foo/abc` would be rewritten to
+`example.net/bar/abc`.
+
+```yaml
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1alpha2
+metadata:
+  name: http-filter-1
+spec:
```
hbagdi, line 115:
```suggestion
spec:
  hostnames:
  - example.com
```

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,270 @@
...
+  Path *HTTPPathRewrite `json:"path,omitempty"`
+}
+
+// HTTPPathRewriteType defines the type of path rewrite.
+type HTTPPathRewriteType string
+
+const (
+  // This type of rewrite indicates that the prefix match will be replaced by
+  // the path rewrite value. This type only applies to path matches of type
+  // Prefix. If a Route has other match types, this rewrite MUST NOT be applied
```
hbagdi, line 159: What if I want to route based on hostname and not path, but I still want to strip a part of the path?
So example.com/foo/bar/baz goes to internal.example.com/bar/baz? This is a very common use-case where users use the routing infrastructure to "fix" such things.

robscott: I think this would still be possible though? The prefix match would be "foo" and the redirect/rewrite value would be "". I need to document that that's possible, but I don't think there's currently anything that would prevent that..

robscott: Added some clarification here, may have overcomplicated the wording though.

site-src/geps/gep-726.md
robscott: Updated, thanks!

site-src/geps/gep-726.md
robscott: OK, I've been thinking about it more, and I think `ReplacePathMatch` is going to provide significantly more flexibility long term. As noted above, it does mean that we'll need to document the subset of path match types that implementations support, but it seems acceptable to continue to consider 100% of this "Extended" support.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,274 @@
...
+
+This would be represented with the following API additions:
+```go
+// HTTPRequestRedirect defines a filter that redirects a request. At most one of
+// these filters may be used on a Route rule. This may not be used on the same
+// Route rule as a HTTPRequestRewrite filter.
+type HTTPRequestRedirect struct {
+  // Path defines a path redirect.
+  //
+  // Support: Extended
```
hbagdi, line 68: Why not core?

My concern is that field-level support-level within filters is going to result in end-user UX issues.
This patch would not introduce the issue (we already have Scheme that is extended) but it is increasing the scope of the problem a bit.

site-src/geps/gep-726.md
hbagdi: So, we are open to adding an ImplementationSpecific option here in the future?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,274 @@
...
+          path:
+            type: ReplacePathMatch
+            value: /bar
+```
+
+This would be represent with the following API additions:
+```go
+// HTTPRequestRewrite defines a filter that modifies a request during
+// forwarding. At most one of these filters may be used on a Route rule. This
+// may not be used on the same Route rule as a HTTPRequestRedirect filter.
```
hbagdi, line 137: So, we essentially have two ways to munge a request depending on if it is being forwarded or if it is being redirected. Did you explore the alternative of one filter to munge the request and if it is a redirect, then this another filter can be used?
Is this something worth mentioning in the alternatives considered section for posterity?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,274 @@
...
+type HTTPRequestRewrite struct {
```
hbagdi, line 138: What is the conformance level of this filter?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,274 @@
...
+  // Path defines a path rewrite.
+  //
+  // Support: Extended
+  //
+  // +optional
+  Path *HTTPPathRewrite `json:"path,omitempty"`
+}
+
+// HTTPPathRewriteType defines the type of path rewrite.
+type HTTPPathRewriteType string
```
hbagdi, line 157: I think having a single path rewrite type would the API more intuitive. Is it possible to combine this and HTTPPathRedirectType types and use the same type in both filters?
I think the cognitive load on the user is going to be high otherwise.

site-src/geps/gep-726.md
robscott: This is tricky because not all types of path redirects are broadly supported. I think it's likely safe to say that absolute and prefix redirects are broadly supportable, not so clear for regex.

site-src/geps/gep-726.md
robscott: To clarify do you mean that Rewrite + Redirect should not overlap in functionality and instead should be used together when both sets of capabilities are needed?

site-src/geps/gep-726.md
robscott: I think 2 is actually solved now, and 1 always was. This comment came from a time when the options were "Absolute" and "Prefix" which did not scale with other match types. Given our removal of ImplementationSpecific types elsewhere, I don't think it would make sense to add it back without a clear rationale.

site-src/geps/gep-726.md
robscott: Apparently I defined conformance at a field level instead of for the filter as a whole, but since both fields have extended conformance, I should probably move that up a level.

robscott:
@bowei @hbagdi @howardjohn @Jeff-Apple @jpeach @youngnick (and anyone else I may have missed), I've updated this PR based on our discussion yesterday at the community meeting. The end result is something a bit closer to what I started with, most notably moving back from "ReplacePathMatch" to "ReplacePrefixPathMatch". 

I've added "ReplacePathMatch" to the alternatives section of the GEP with a set of reasons for why I don't think it is as helpful as it may have seemed on the surface.

I've also added a "future extension" section that describes how we can add additional path rewrite/redirect types in the future. Importantly this shows how regex substitution could work, and I think it ends up being preferable to the more generic "ReplacePathMatch" approach.

Interested in what you think of these changes, hopefully the updated direction makes sense.

site-src/geps/gep-726.md
hbagdi: No. I mean that HTTPPathRedirectType and HTTPPathRewriteType types should be merged into one.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,336 @@
...
+  // +kubebuilder:validation:MaxLength=255
+  Hostname *string `json:"hostname,omitempty"`
+
+  // Path defines a path rewrite.
+  //
+  // Support: Extended
+  //
+  // +optional
+  Path *HTTPPathRewrite `json:"path,omitempty"`
+}
```
hbagdi, line 160: There will be two filters: `RequestHeaderModifier` and a `RequestRewrite`, the UX seems a bit off here.
Options are:
- Deprecated RequestHeaderModifier and fold it into this new filter
- Introduce more targetted filters (break this new filter into two): RequestHostnameModifier, RequestPathModifier
- Accept this emergent behavior
- Something else?

robscott: Although I agree the filters are similar in purpose, I think "Rewrite" is a common and reasonably well understood concept here. In most of the implementations I can find, rewriting is limited to either the request hostname and path or the full URL. There are a small set of implementations that extend "rewrite" terminology to headers, but that does not seem to be nearly as consistent.

I think the term "rewrite" is important here given its widespread use throughout other implementations. I think this is going to be a common term that users are looking for. Although it may technically more consistent to have a `RequestModifier` filter or separate `RequestHostnameModifier` and `RequestPathModifier` filters, I think it would lead to more confusion.

We could also bundle header modification into this proposed rewrite filter, but I'm not sure that provides sufficient value to introduce a breaking change and/or duplicate capabilities.

site-src/geps/gep-726.md
robscott: This has been addressed in more detail in our latest community meeting + https://github.com/kubernetes-sigs/gateway-api/pull/731#issuecomment-964682943 - I ended up going back to `ReplacePathPrefixMatch`.

site-src/geps/gep-726.md
robscott: I think my latest commit has resolved this, let me know if not.

site-src/geps/gep-726.md
robscott: Added absolute as an option in both places as part of transition to shared types.

site-src/geps/gep-726.md
```diff
@@ -238,9 +246,63 @@ The NGINX rewrite module contains the following relevant capabilities
...
+## Future Extension
+There are two relatively common types of path rewrite/redirect that are not
+covered by this proposal:
+
+1. Replace a path prefix separate from the match
+2. Replace with a Regular Expression substitution
+
+Both of the following can be represented by adding a new field new types. For
+example, this config would result in a request to `/foo/baz` to be rewritten to
+`/foo/bar`:
```
youngnick, line 258:
```suggestion
`/bar/baz`:
```
I think?

youngnick (commented):
LGTM aside from one small thing.

site-src/geps/gep-726.md
robscott: Great catch, thanks!

site-src/geps/gep-726.md
hbagdi: So this is where implementations doing a thing shouldn't be governing our design, right?
The API is to express intent, how the implementations do shouldn't matter - as long as implementations can satisfy the intent.
With that bit, from a purely product and API perspective, would a user not expect a RequestRewrite filter to also be able to modify headers because headers are part of the request?

As a user, wouldn't you be surprised if you had a filter names "RewriteRequest" which won't be able to modify headers? And to modify headers, you had to use a filter called "RequetHeaderModifier"?

Being consistent can help us keep complexity under control in this area.
What are your concerns with duplication of header modification in the RequestRewrite filter as well? We could keep the behavior and intent exactly the same, that should minimize overheads.

robscott: >The API is to express intent, how the implementations do shouldn't matter - as long as implementations can satisfy the intent.

💯 Agree with this.

>What are your concerns with duplication of header modification in the RequestRewrite filter as well? We could keep the behavior and intent exactly the same, that should minimize overheads.

Any time we provide multiple ways to do something, we have to handle conflicts between those. We also have to take extra effort to document and explain the difference between the ways to do things.

>As a user, wouldn't you be surprised if you had a filter names "RewriteRequest" which won't be able to modify headers? And to modify headers, you had to use a filter called "RequetHeaderModifier"?

It's hard for me to say since I'm already so far into the API design. Some thoughts:

1. Header modification may happen at any point, including before a redirect.
1. The current proposal considers rewriting distinct from redirection, merging these together would lead to more confusion as capabilities diverged.
1. We could consider renaming "RequestRewrite" to "URLRewrite" which would more clearly communicate the proposed scope. That would be somewhat inconsistent with the existing "RequestRedirect" filter name, but this is also smaller in scope.
1. We could consider moving everything into smaller filters - PathModifier, HostnameModifier, HeaderModifier, SchemeModifier, etc and then have one final filter that indicated all of the above should result in a redirect and not just a rewrite. I'm concerned that this could result in unnecessary confusion around conformance levels + breaking API changes.

robscott: Chatted with @hbagdi a bit and decided that 3 was the most reasonable approach going forward, I've updated the GEP to reflect that. I've also created https://github.com/kubernetes-sigs/gateway-api/issues/928 as a related follow up from our conversation.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+
+Similarly, this config would result in a request to `/foo/bar/baz` being
+rewritten to `/foo/other/baz`.
+```yaml
+filters:
+- type: RequestRewrite
+  requestRewrite:
+    path:
+      type: RegularExpression
+      pattern: /foo/(.*)/baz
```
ccfishk, line 261: isn't "*" enough here ?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+- type: RequestRewrite
+  requestRewrite:
+    path:
+      type: ReplacePrefix
+      pattern: /foo
+      substitution: /bar
+```
+
+Similarly, this config would result in a request to `/foo/bar/baz` being
+rewritten to `/foo/other/baz`.
```
ccfishk, line 254: Shouldn't be A request to `/foo/baz` being
rewritten to `/bar/baz` ?
Confused.

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+const (
+  // This type of modifier indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathModifier HTTPPathModifierType = "Absolute"
+
+  // This type of modifier indicates that any prefix path matches will be
+  // replaced by the substitution value. For example, a path with a prefix match
+  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
+  // "/foo" prefix replaced with "/bar" in matching requests.
+  PathMatchHTTPPathModifier HTTPPathModifierType = "ReplacePrefixPathMatch"
```
bowei, line 51: I find this naming a bit confusing, "path" appears twice:

*Path*MatchHTTP*Path*Modifier

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+  // match of "/foo" would be modified to "/bar".
+  //
+  // +kubebuilder:validation:MaxLength=1024
+  Substitution string `json:"substitution"`
+}
+```
+
+### Redirects
+
+The existing `RequestRedirect` filter can be expanded to support path redirects.
```
bowei, line 73: To be very clear, redirect means that the server sends a 3xx back to the client?

Which 3xx is sent? Does it matter?

site-src/geps/gep-726.md
bowei: Why not just MatchHTTPPathModifier?

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+        requestRedirect:
+          hostname: foo.com
+          path:
+            type: ReplacePrefixPathMatch
+            value: /bar
+```
+
+This would be represented with the following API additions:
+```go
+// HTTPRequestRedirect defines a filter that redirects a request. At most one of
```
bowei, line 99: We should state very explicitly this sends a 3xx response to the client.

site-src/geps/gep-726.md
bowei: I think you are thinking of UNIX glob expression vs regular expression.

The dot "." means any character.

site-src/geps/gep-726.md
bowei: If this is to be extended to do more than just an absolute substitution, what would that look like?

bowei:
Generally LGTM -- probably hash out the details in the actual API review.

site-src/geps/gep-726.md
jpeach:
```suggestion
  ReplacePrefixHTTPPathModifier HTTPPathModifierType = "ReplacePrefix"
```

WDYT?

site-src/geps/gep-726.md
jpeach: the redirect filter sets 302 as the default status code

site-src/geps/gep-726.md
```diff
@@ -0,0 +1,319 @@
...
+type HTTPPathModifierType string
+
+const (
+  // This type of modifier indicates that the complete path will be replaced by
+  // the path redirect value.
+  AbsoluteHTTPPathModifier HTTPPathModifierType = "Absolute"
+
+  // This type of modifier indicates that any prefix path matches will be
+  // replaced by the substitution value. For example, a path with a prefix match
+  // of "/foo" and a ReplacePrefixPathMatch substitution of "/bar" will have the
```
jpeach, line 49: From an API perspective, I think that I would prefer to decouple prefix matching from prefix rewriting. This is a  bit awkward in Envoy IIRC, but it's doable (at least Contour does it).

If we keep this as it is, then a rewrite would become invalid if the path match changes, and it would also rewrite differently depending on the match prefix. This seems operationally fraught to me, and likely to trick people with unexpected side-effects.

robscott: My understanding of the [Envoy spec](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-redirectaction-prefix-rewrite) is that the only kind of prefix rewriting it supports is based on the corresponding prefix match. That's also true for the `pathPrefixRewrite` in [GCP UrlMaps](https://cloud.google.com/compute/docs/reference/rest/v1/urlMaps#UrlMap.FIELDS.inlinedField_22). Of course with Envoy and other proxies, regex path rewrites could be used to enable more advanced use cases including rewrites based on an arbitrary prefix (covered later in this GEP as possible future extensions), but that doesn't seem to be quite as widely supported as rewriting based on a prefix match.

site-src/geps/gep-726.md
robscott: This is a purely theoretical future implementation of regex path rewrites, but in this case, users would provide a separate pattern (regex) and substitution. 

```
      pattern: /foo/(.*)/baz
      substitution: other
```

So in this case, the `substitution` would be replacing the `(.*)` part of the pattern.

site-src/geps/gep-726.md
robscott: This was renamed a few too many times and I missed the actual variable name. Gave this another update, this time to `PrefixMatchHTTPPathModifier` and `ReplacePrefixMatch` respectively. Hopefully that makes a bit more sense.

site-src/geps/gep-726.md
robscott: Reference: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L703

site-src/geps/gep-726.md
robscott: This part of the GEP is confusing because it was not clear that the redirect filter itself already exists and is more clearly defined than this snippet suggests. I've updated the GEP to clarify that this is an addition to an existing filter. https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L703

site-src/geps/gep-726.md
jpeach: Contour implements this by implicitly expanding it to multiple matches (IIRC I wrote that, but I forget the details). As an API feature, I'd prefer that Envoy limitations didn't leak through, though I'm sympathetic to the complexity of implementing this in the Envoy API.

robscott: >I'd prefer that Envoy limitations didn't leak through

I agree with that, though in this case it's something that also affects GCP APIs, where the proposed `PrefixMatch` is the only path rewrite currently supported. Although I don't want to copy APIs, I also want to build an API that is broadly implementable. I think the current GEP along with the proposed future extensions for Prefix and Regex rewrites based on custom patterns will provide the best possible balance between an implementable and understandable API, but open to other potential approaches.

robscott:
Discussed this at today's community meeting, sounded like there was fairly broad consensus, going to remove the hold. Thanks to everyone for all the help with this one!

/hold cancel

PR #732, opened 2021-07-22
Removing BackendPolicy

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This is part of the GEP #713 implementation.

```release-note
BackendPolicy has been removed from the API for v1alpha2. This is in favor of the new policy attachment guidelines introduced by GEP #713.
```


apis/v1alpha2/shared_types.go
```diff
@@ -32,23 +32,6 @@ const (
...
-	// AppProtocol field is available, this annotation should not be used. The
-	// AppProtocol field, when populated, takes precedence over this annotation.
-	// The value of this annotation must be also be a valid value for the
-	// AppProtocol field.
-	//
-	// Examples:
-	//
-	// - `gateway.networking.k8s.io/app-protocol: https`
-	// - `gateway.networking.k8s.io/app-protocol: tls`
-	AnnotationAppProtocol = "gateway.networking.k8s.io/app-protocol"
```
robscott, line 49: This is a significant removal, but given that Kubernetes 1.18+ is so widely used already, I don't think it's worth including workarounds for older Kubernetes versions anymore. Since this one depended on BackendPolicy, it would be difficult to recreate without some kind of intermediate resource.

site-src/v1alpha2/guides/tls.md
```diff
@@ -130,28 +130,6 @@ there are two HTTPRoute resources which specify certificates for
...
-
-### Example
-
-The following example shows how upstream TLS can be configured. We have
-omitted downstream TLS configuration for simplicity. As noted before, it
-doesn't matter how downstream TLS is configured for the specific listener or
-route.
-
-```yaml
-{% include 'upstream-tls.yaml' %}
```
hbagdi, line 152: Please delete this file.

apis/v1alpha2/shared_types.go
hbagdi: I opened up https://github.com/kubernetes-sigs/gateway-api/issues/737 to track the versions we support.
I was ambivallent about the change but given where we are and where this API is, reworking this in won't have much ROI.
+1

hbagdi:
/lgtm

/hold because:
- remove the one file that no longer should be present
- approvals from other maintainers if you feel the need

site-src/v1alpha2/guides/tls.md
```diff
@@ -130,28 +130,6 @@ there are two HTTPRoute resources which specify certificates for
...
-## Upstream TLS
-
-Upstream TLS configuration applies to the connection between the Gateway
-and Service.
-
-There is only one way to configure upstream TLS: using the `BackendPolicy`
-resource.
-
-Please note that the TLS configuration is related to the Service or backend
-resource and not related to a specific route resource.
```
hbagdi, line 142: Can we keep some of this text and inform the user to use ReferencePolicy for these purposes? it could help with some education and getting our users and implementers be more aware about the new feature.

site-src/v1alpha2/guides/tls.md
robscott: To clarify, this is just the upstream-tls example right? In that case, it's still used by v1alpha1 docs, and probably will be as long as we keep v1alpha1 docs around.

site-src/v1alpha2/guides/tls.md
robscott: I think that the Policy Attachment GEP is likely a better fit here, though it will be easier to reference that when #736 merges.

robscott: Added a hold on #736 to deal with this there.

robscott:
Removing the hold on this one since I think we already have clear consensus on the GEP that proposed this.

/hold cancel

PR #733, opened 2021-07-23
feat(HTTPRouteMatcher): implement http method filter

### LINKED ISSUE #624: Method based routing
I did a quick search and did not see a ticket for this, if I missed it I am sorry.

Method based routing would be a nice addition. This is something that we are currently looking to implement in another part of our stack, but having it in a gateway-api would make things easier for us.

An example of this can be found in Ambassador 

```
---
apiVersion: getambassador.io/v2
kind: Mapping
metadata:
  name: get
spec:
  prefix: /backend/get_only/
  method: GET
  service: quote
```
https://www.getambassador.io/docs/edge-stack/latest/topics/using/method/#method-based-routing

**Why is this needed**:
There are many use cases for this, but the my primary need for this is to split read (GET) and write (PUT) traffic to different endpoints.

secustor:
**What type of PR is this?**
/kind feature
/kind api-change

This PR adds a HTTPMatcher to allow match one or more specific HTTP methods.      
If multiple are defined it is defined as logical OR. 
 
Fixes #624

```release-note
HTTPRoute now supports matching based on HTTP methods.
```


apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,10 @@ type HTTPQueryParamMatch struct {
 	Value string `json:"value"`
 }
 
+// HTTPMethod describes how to select a HTTP route by matching the HTTP
+// method.
+type HTTPMethod string
```
brunoarueira, line 413: Is it an appropriate layer to add some kind of enum to allow specific values? E.g. the http methods defined by the HTTP spec.

See https://datatracker.ietf.org/doc/html/rfc7231#section-4 and here https://datatracker.ietf.org/doc/html/rfc5789#section-2

I thinking more about a way to avoid mistakes when defining the paths and only when the config is applied to not route through to the correct service

secustor: Alternatively we could reuse the HttpMethod from the `genproto` package https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/scheduler/v1#HttpMethod

apis/v1alpha2/httproute_types.go
```diff
@@ -457,6 +461,13 @@ type HTTPRouteMatch struct {
 	//
 	// +optional
 	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
+
+	// Method specifies HTTP method matchers. Multiple match
+	// values are ORed together, meaning, a request must match any of the
+	// specified methods to select the route.
+	//
+	// +optional
+	Method []HTTPMethod `json:"methods,omitempty"`
```
robscott, line 470: I think I'd prefer starting with a single method here. If multiple methods are needed they can be communicated with multiple matches.

apis/v1alpha2/httproute_types.go
robscott: I like the idea of an enum here. I'd also recommend adding enum validation to the field as shown here: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L271. 

I think I'd prefer defining our own custom types instead of reusing the ones defined in genproto. That will allow us more flexibility in terms of documentation and support levels.

Also appreciate the RFC references by @brunoarueira. Where possible, we like to reference specific RFCs when documenting a feature/field in the comments (example: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L342).

robscott (commented):
Thanks for the work on this!

/ok-to-test

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,10 @@ type HTTPQueryParamMatch struct {
 	Value string `json:"value"`
 }
 
+// HTTPMethod describes how to select a HTTP route by matching the HTTP
+// method.
```
robscott, line 412: We should also include a support level here. Given that this is just a shortcut for header matching, I think it probably makes sense to use the same support level we've defined there (Core): https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L328.

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,25 @@ type HTTPQueryParamMatch struct {
 	Value string `json:"value"`
 }
 
+// HTTPMethod describes how to select a HTTP route by matching the HTTP
+// method as defined by [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4) and [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789#section-2).
+// The value is expected in upper case
```
robscott, line 413: Nit: Please wrap at 80 char line lengths.

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,25 @@ type HTTPQueryParamMatch struct {
...
+// +kubebuilder:validation:Enum=GET;HEADER;POST;PUT;DELETE;CONNECT;OPTIONS;TRACE;PATCH
+type HTTPMethod string
+
+const (
+	HTTPMethodUndefined HTTPMethod = "NONE"
```
robscott, line 418: Do we need this for anything? It does not look like it's a valid value for the field.

apis/v1alpha2/httproute_types.go
```diff
@@ -457,6 +476,14 @@ type HTTPRouteMatch struct {
...
 	// +optional
 	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
+
+	// Method specifies HTTP method matchers.
+	// If specified this matcher will only be picked if the request has the corresponding method
+	//
+	// Support: Core
+	//
+	// +optional
+	Method HTTPMethod `json:"method,omitempty"`
```
robscott, line 486: I think this should be a pointer as it is an optional field.

robscott (commented):
Mostly LGTM, just a few nits.

apis/v1alpha2/httproute_types.go
secustor: It is not a valid HTTP method, if that's what you mean. That is the reason I didn't add it to the validator. 

I use such values `not_defined` values in my code and as `genproto` adds one too.  
Should you have reservations against it then I will remove it.

apis/v1alpha2/httproute_types.go
secustor: I don't understand the reasoning behind it, but you seem to be right. Every optional value which is not an array is a pointer.

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,25 @@ type HTTPQueryParamMatch struct {
...
+	HTTPMethodUndefined HTTPMethod = "NONE"
+	HTTPMethodGet       HTTPMethod = "GET"
+	HTTPMethodHeader    HTTPMethod = "HEADER"
+	HTTPMethodPost      HTTPMethod = "POST"
+	HTTPMethodPut       HTTPMethod = "PUT"
+	HTTPMethodDelete    HTTPMethod = "DELETE"
+	HTTPMethodConnect   HTTPMethod = "CONNECT"
+	HTTPMethodOptions   HTTPMethod = "OPTIONS"
+	HTTPMethodTrace     HTTPMethod = "TRACE"
+	HTTPMethodPatch     HTTPMethod = "PATCH"
```
hbagdi, line 429: I'm not sure if this is a good idea or not.
Custom HTTP methods are a thing, although not common in practice.
For the 99% of the cases, these defined methods will suffice.
If there comes a time when we want to support custom HTTP methods, how can we do so without a breaking change?

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,27 @@ type HTTPQueryParamMatch struct {
...
+// [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4) and
+// [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789#section-2).
+// The value is expected in upper case
+// +kubebuilder:validation:Enum=GET;HEADER;POST;PUT;DELETE;CONNECT;OPTIONS;TRACE;PATCH
+type HTTPMethod string
+
+const (
+	HTTPMethodUndefined HTTPMethod = "NONE"
+	HTTPMethodGet       HTTPMethod = "GET"
+	HTTPMethodHeader    HTTPMethod = "HEADER"
```
hbagdi, line 422: Is this same as `HEAD` method? I've not heard about a `HEADER` method, it seems confusing since we already have HTTP headers as a concept.

secustor: Typo on my part.

apis/v1alpha2/httproute_types.go
hbagdi: > If I'm not mistaken it is already possible to match the HTTP method via header value match.

How so?


cc @robscott @jpeach

howardjohn: With `:method` pseudo-header (https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.1)

If we go down this route we need to make it explicit in the spec that implementation must accept pseudo-headers though.

hbagdi: > If we go down this route we need to make it explicit in the spec that implementation must accept pseudo-headers though.

We should clarify this regardless. Else implementations will drift in this aspect.

apis/v1alpha2/httproute_types.go
robscott: Although I think an undefined type like this can make sense in many cases, we've generally avoided them in k8s APIs unless they're going to be explicitly specified. I'd lean towards removing this.

apis/v1alpha2/httproute_types.go
```diff
@@ -457,6 +478,14 @@ type HTTPRouteMatch struct {
 	//
 	// +optional
 	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
+
+	// Method specifies HTTP method matchers.
+	// If specified this matcher will only be picked if the request has the corresponding method
+	//
+	// Support: Core
```
robscott, line 485: I was wrong on the support level here, a bit more discussion at the community meeting yesterday made it clear that it should actually be "Extended" to start with.

hbagdi: +1 to Extended. I feel this could potentially move into Core at some point.

apis/v1alpha2/httproute_types.go
jpeach: FWIW there is an IANA registry for HTTP methods

https://www.iana.org/assignments/http-methods/http-methods.xhtml

apis/v1alpha2/httproute_types.go
jpeach: Yup, +1 on defining pseudo-header matching. Controllers can map the pseudo header name to the underlying proxy control.

hbagdi: I opened up https://github.com/kubernetes-sigs/gateway-api/issues/740 to track this conversation.

My only slight hesitation is that then there are two ways to do the same thing. In this case, that might be okay because custom HTTP methods are not common. Not really sure.

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,26 @@ type HTTPQueryParamMatch struct {
 	Value string `json:"value"`
 }
 
+// HTTPMethod describes how to select a HTTP route by matching the HTTP
+// method as defined by
+// [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4) and
+// [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789#section-2).
+// The value is expected in upper case
```
hbagdi, line 415:
```suggestion
// The value is expected in upper case.
```
The enum below communicates the same thing so you could consider removing it as well.

apis/v1alpha2/httproute_types.go
```diff
@@ -457,6 +477,14 @@ type HTTPRouteMatch struct {
 	//
 	// +optional
 	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
+
+	// Method specifies HTTP method matchers.
+	// If specified this matcher will only be picked if the request has the corresponding method
```
hbagdi, line 482:
```suggestion
	// Method specifies HTTP method matcher.
	// When specified, this route will be matched only if the request has the specified method.
```

apis/v1alpha2/httproute_types.go
secustor: As I understand it this comments will be used to auto generate the docs, correct?  If that's the case I think this valuable information.

apis/v1alpha2/httproute_types.go
```diff
@@ -457,6 +477,14 @@ type HTTPRouteMatch struct {
...
 	// +optional
 	ExtensionRef *LocalObjectReference `json:"extensionRef,omitempty"`
+
+	// Method specifies HTTP method matcher.
+	// When specified, this route will be matched only if the request has the specified method.
+	//
+	// Support: Extended
+	//
+	// +optional
+	Method *HTTPMethod `json:"method,omitempty"`
```
hbagdi, line 487: nit: Please move Method before ExtensionRef.

apis/v1alpha2/httproute_types.go
```diff
@@ -408,6 +408,26 @@ type HTTPQueryParamMatch struct {
 	Value string `json:"value"`
 }
 
+// HTTPMethod describes how to select a HTTP route by matching the HTTP
+// method as defined by
+// [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4) and
+// [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789#section-2).
+// The value is expected in upper case.
+// +kubebuilder:validation:Enum=GET;HEADER;POST;PUT;DELETE;CONNECT;OPTIONS;TRACE;PATCH
```
hbagdi, line 416:
```suggestion
// +kubebuilder:validation:Enum=GET;HEAD;POST;PUT;DELETE;CONNECT;OPTIONS;TRACE;PATCH
```

This is already facing copy-paste challenges. 😞

hbagdi (commented):
I was about to lgtm this but found a couple of small issues.

apis/v1alpha2/httproute_types.go
```diff
@@ -445,6 +465,14 @@ type HTTPRouteMatch struct {
 	// +optional
 	QueryParams []HTTPQueryParamMatch `json:"queryParams,omitempty"`
 
+	// Method specifies HTTP method matcher.
+	// When specified, this route will be matched only if the request has the specified method.
```
robscott, line 469: Nit: Please wrap these lines at 80 chars.

robscott:
Thanks! Will leave LGTM for someone else.

/approve

hbagdi:
@secustor Thank you for your contribution!
/lgtm

Issue #611, opened 2021-04-13
Extensible Service policy and configuration


mark-church:
[This proposal ](https://docs.google.com/document/d/1Wh52e7pupV0H1lEA5l_sqDmvRC7lk2cf1fufCz8vxzM/edit?usp=sharing) draws inspiration from our current TLS design and offers a uniform method of attaching arbitrary and implementation-specific configuration to the Gateway API core resources. 

It aims to achieve three different goals that I think are very important for a standard API spec. This is based on the assumptions that 1) every implementation will have proprietary functionality that's also unique and 2) we can't predict what this functionality is or where in the load balancing hierarchy it should be configured. As a result, to make Gateway a standard that can actually replace proprietary APIs, it must have a robust way of accomplishing the following three things:

1. Provide a standard way of extending arbitrary and proprietary configuration and policy on top of the core resources of the Gateway API
2. Provide the ability to configure (arbitrary and proprietary) policy across different scopes of the load balancing infrastructure, from GatewayClass, to Gateway, to HTTPRoute, to Services, to individual workloads.
3. Provide a way for policy to compose: policy should be able to be configurable simultaneously at multiple scopes (both broad and granular) by non-coordinating users and that policy should compose together to a single effective policy for a given service.

This is just a proposal so I'm happy to take any suggestions on what should change or how we can make it better. cc @bowei @thockin @robscott @hbagdi @youngnick @danehans @howardjohn

youngnick:
We had a lot of similar problems with HTTPProxy, Contour's own CRD, which we tried solving in a similar way, but we found it very quickly grew unwieldy because of the interactions between different policy options.

The best example for HTTPProxy was client request timeout. In HTTPProxy's API, we allow setting this to the special value "infinity", which means that Envoy will disable this timeout. However, when running a central ingress proxy, some organisations don't want to allow people to do this. We experimented with a few ways to do this in the API, and eventually just recommended that, if you need that level of granularity, you use Gatekeeper as an admission controller, and we provide example PolicyTemplates and Policies. That lets an infra admin put strong bounds around settings, on a per-field level, that give early UX feedback (objects that violate the policy are never accepted). It also lets administrators who need even greater flexibility build that for themselves (if you want to allow a single namespace to set client request timeout to infinity, that's very possible with Gatekeeper).

That's not to say that we shouldn't do what you're saying in the doc (I've taken a quick read, will go over again momentarily), just that we should consider the UX and portability of the API as well.

[...]

Issue #99, opened 2020-02-15
Specify circuit breaking/passive health-check behavior for each service


hbagdi:
**What would you like to be added**:

Circuit breaking is a commonly implemented pattern where if a specific instance (pod) of a Service starts throwing 5xx, then the traffic is not forwarded to that specific pod.

**Why is this needed**:

Proxies and load-balancers already share these features and a load-balancing layer is a good place to add these features to deliver a resilient and high-available service.

hbagdi:
In search for a solution, we were unsuccessful in coming up with a portable way to support this feature.
As a compromise, we came up with an easy and portable way to support such implementation-specific features in a consistent manner. The solution is documented in #713.

If there are more portable ways to tackle this, we would love to hear about them.

For now, I'll close this issue.
/close

Issue #98, opened 2020-02-15
Specify load-balancing behavior for each service that is being proxied


hbagdi:
**What would you like to be added**:

For any service that is exposed via service-apis, a service owner should be able to specify the load-balancing behavior that should be used. This can include a common set of algorithms like round-robin, weighted, hash based routing, cookie/session based routing.

**Why is this needed**:

Most services in Kubernetes are designed to scale horizontally. A lot of legacy services that are ported to run on Kubernetes have the requirement of stick sessions. Most proxies and load balancers support load-balancing as well and this is a widely adopted feature in most of current ingress implementations.

hbagdi:
There is certainly an overlap with kube-proxy and the `corev1.Service` resource here.

bowei:
Yes, this API would give us an opportunity to add better load balancing knobs without adding more to the `core.Service` bucket

hbagdi:
@robscott Seems like there is already an issue open for load-balancing.

hbagdi:
In search for a solution, we were unsuccessful in coming up with a portable way to support this feature.
As a compromise, we came up with an easy and portable way to support such implementation-specific features in a consistent manner. The solution is documented in #713.

If there are more portable ways to tackle this, we would love to hear about them.

For now, I'll close this issue.
/close

PR #736, opened 2021-07-26
Implementing policy attachment GEP

robscott:
**What type of PR is this?**
/kind documentation
/kind feature

This adds documentation for the policy attachment pattern described in GEP #713 and adds the associated `PolicyTargetReference` type. When combined with #732, this should be all that's required to implement the GEP.

```release-note
A new policy attachment model has been added.
```

Deploy Preview: https://deploy-preview-736--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/policy-attachment/


robscott:
Adding a hold to follow up with https://github.com/kubernetes-sigs/gateway-api/pull/732#discussion_r676920640.

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,56 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+// PolicyTargetReference identifies an API object to apply policy to.
```
youngnick, line 19: I think we should have some (brief) guidance on how this struct should be used, with a pointer to the GEP, as a start.

I also think that we should actually supply template objects for Policy objects, for Route objects, and any other one that we expect people to do their own versions of. But that's a separate thing I'll log an issue about.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
+# Policy Attachment
+
+Policies attached to Gateway API resources and implementations should use the
```
youngnick, line 3: Nit:
```suggestion
Policies attached to Gateway API resources and implementations must use the
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+* A shared means of attaching policy to resources.
+* Support for configuring both default and override values within policy
+  resources.
+* A shared hierarchy to illustrate how default and override values should
+  interact.
+
+This kind of standardization will not only enable consistent patterns, it will
+enable tooling such as kubectl plugins to be able to visualize all policies that
+have been applied to a given resource.
```
youngnick, line 15: Nit: We're already doing this, so "will enable" is not really correct.
```suggestion
This kind of standardization not only enables consistent patterns, it allows
future tooling such as kubectl plugins to visualize all policies that
have been applied to a given resource.
```

youngnick (commented):
This is looking great, just a few small wording changes from me.

hbagdi:
xref #98  #99 #91 #89 #114 #97 #611

apis/v1alpha2/policy_types.go
hbagdi: > I also think that we should actually supply template objects for Policy objects, for Route objects, and any other one that we expect people to do their own versions of. But that's a separate thing I'll log an issue about.

Heavy +1
As an implementer, I want to simply embed the struct in my struct and call it done.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
+# Policy Attachment
```
hbagdi, line 1: As a reader, I think it would help me if Policy is defined early on in this document.
I can't seem to comprehend what we mean by policy. But then as soon as I read the GKEServicePolicy example, it starts to make sense.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+Policies attached to Gateway API resources and implementations should use the
+following approach to ensure consistency across implementations of the API.
+There are three primary components of this pattern:
+
+* A shared means of attaching policy to resources.
+* Support for configuring both default and override values within policy
+  resources.
+* A shared hierarchy to illustrate how default and override values should
```
hbagdi, line 10:
```suggestion
* A hierarchy to illustrate how default and override values should
```

Not sure if shared really helps here. NBD.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
+# Policy Attachment
+
+Policies attached to Gateway API resources and implementations should use the
+following approach to ensure consistency across implementations of the API.
+There are three primary components of this pattern:
+
+* A shared means of attaching policy to resources.
```
hbagdi, line 7:
```suggestion
* A standardized means of attaching policy to resources.
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+* Support for configuring both default and override values within policy
+  resources.
+* A shared hierarchy to illustrate how default and override values should
+  interact.
+
+This kind of standardization will not only enable consistent patterns, it will
```
hbagdi, line 13:
```suggestion
This kind of standardization not only enables consistent patterns, it also
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+enable tooling such as kubectl plugins to be able to visualize all policies that
+have been applied to a given resource.
+
+## Policy Attachment for Ingress
+Attaching policy to Gateway resources for ingress use cases is relatively
+straightforward. A policy can reference the resource it wants to apply to.
+Access is granted with RBAC - anyone that has access to create a RetryPolicy in
+a given namespace can attach it to any resource within that namespace.
+
+![Simple Ingress Example](/geps/images/713-ingress-simple.png)
```
hbagdi, line 23: 404

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+To build on that example, it’s possible to attach policies to more resources.
+Each policy applies to the referenced resource and everything below it in terms
+of hierarchy. Although this example is likely more complex than many real world
+use cases, it helps demonstrate how policy attachment can work across
+namespaces.
+
+![Complex Ingress Example](/geps/images/713-ingress-complex.png)
```
hbagdi, line 31: 404

hbagdi: 404 for all pngs

hbagdi: I'd also recommending separating GEP content from site content. No Big Deal if later.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+    // +listType=map
+    // +listMapKey=type
+    // +kubebuilder:validation:MaxItems=8
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+### Hierarchy
+Each policy MAY include default or override values. Default values are given
+precedence from the bottom up, while override values are top down. That means
```
hbagdi, line 115: We should first explain the hierarchy and then use top-down bottom-up terminology.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+that a default attached to a Backend will have the highest precedence among
+default values while an override value attached to a GatewayClass will have the
+highest precedence overall.
+
+![Ingress and Sidecar Hierarchy](/geps/images/713-hierarchy.png)
```
hbagdi, line 120:
```suggestion
![Policy Hierarchy](/geps/images/713-hierarchy.png)
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+If using defaults and overrides, each policy resource MUST include 2 structs
+within the spec. One with override values and the other with default values.
+
+In the following example, the policy attached to the Gateway requires cdn to
+be enabled and provides some default configuration for that. The policy attached
+to the Route changes the value for one of those fields (includeQueryString).
+
+```yaml
+kind: GKEServicePolicy # Example of implementation specific policy name
```
hbagdi, line 144: Like in most of our examples, being vendor neutral would be best.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+same time, we can see that the default connectionTimeout attached to the Route
+has precedence over the default attached to the Gateway.
+
+![Hierarchical Policy Example](/geps/images/713-policy-hierarchy.png)
+
+#### Supported Resources
+It is important to note that not every implementation will be able to support
+policy attachment to each resource described in the hierarchy above. When that
+is the case, implementations MUST clearly document which resources a policy may
+be attached to.
```
hbagdi, line 181: We should document the conformance level and then mention this paragraph.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+given scope will be able to modify policy at that level. Since GatewayClass is a
+cluster scoped resource, this means that any policy attached to it must also be
+cluster scoped.
+
+GatewayClass parameters provide an alternative to policy attachment that may be
+easier for some implementations to support. These parameters can similarly be
+used to set defaults and requirements for an entire GatewayClass.
+
+### Targeting External Services
+In some cases (likely limited to mesh) we may want to apply policies to requests
```
hbagdi, line 196: > we may want to

No more a proposal so please reword.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+ties:
+
+* The oldest Policy based on creation timestamp. For example, a Policy with a
+  creation timestamp of "2021-07-15 01:02:03" is given precedence over a Policy
+  with a creation timestamp of "2021-07-15 01:02:04".
+* The Policy appearing first in alphabetical order by "<namespace>/<name>". For
+  example, foo/bar is given precedence over foo/baz.
+
+For a better user experience, a validating webhook can be implemented to prevent
+these kinds of conflicts all together.
```
hbagdi, line 228: What sort of validation are we referring to here?
Isn't the recommendation to avoid validation across resources as much as possible?

site-src/v1alpha2/references/policy-attachment.md
hbagdi: This is the case through-out the doc and we should fix before merge.

site-src/v1alpha2/references/policy-attachment.md
hbagdi: I see that you have done this at the bottom of the doc, combining these makes sense to me.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Kubectl Plugin
+To help improve UX and standardization, a kubectl plugin will be developed that
+will be capable of describing the computed sum of policy that applies to a given
+resource, including policies applied to parent resources.
+
+Each Policy CRD that wants to be supported by this plugin will need to follow
+the API structure defined above and add a `gateway.networking.k8s.io/policy:
+true` label to the CRD.
```
hbagdi, line 237:
```suggestion
Each Policy CRD that wants to be supported by this plugin will need to follow
the API structure defined above and add a `gateway.networking.k8s.io/policy-attachment:
true` label to the CRD.
```

We have ReferencePolicy resource in core so that might cause some confusion, hence the suggestion.
I don't know how to make this clearer: the label is on the CRD and not CRs. Maybe it is clear enough?

site-src/v1alpha2/references/policy-attachment.md
robscott: Not sure how to make this work on github previews + docs site, but deploy preview makes it look like it works there at least: https://deploy-preview-736--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/policy-attachment/#policy-attachment-for-ingress.

site-src/v1alpha2/references/policy-attachment.md
robscott: Replaced with "users may want to", other similar uses seem to be limited to things like "an app owner may want to" which seems reasonable to me.

site-src/v1alpha2/references/policy-attachment.md
robscott: Yeah cross-resource validation is tricky to get right, and something we generally try to avoid. I think it would provide a better UX, but it's not well defined and adds no value here. I've removed it.

site-src/v1alpha2/references/policy-attachment.md
robscott: Yeah this is better, updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Makes sense, updated to Acme.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Good call, moved all images to a separate /images/policy directory.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Good idea, added a paragraph to start this section off, let me know what you think.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

apis/v1alpha2/policy_types.go
robscott: Good call, I've added a more detailed description here + a reference to the docs that already include a sample policy object.

site-src/v1alpha2/references/policy-attachment.md
robscott: Good call. I restructured this a bit to start with a slightly more generic introduction that was focused on the roles, hopefully that helps.

robscott:
Thanks for the great feedback on this @hbagdi and @youngnick! I think I've responded to all feedback now, PTAL.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
+# Policy Attachment
+
+Many implementations of the API support additional functionality that is not
+covered by the API directly. For example, concepts like timeouts, retries, or
```
hbagdi, line 4: nit: "the API" is a bit abstract in the beginning and qualifying it as "Gateway API" might help set the context.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+even CDN configuration are not sufficiently portable to be defined within the
```
hbagdi, line 5: nit: load-balancing, health-checks might be more suited examples than CDN.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+Gateway API. Instead, implementations may choose to represent these concepts
+with custom Policy resources.
```
hbagdi, line 7: It might be helpful for the reader to understand why these are not part of the API, cause that is the first question that comes to my mind.

Something like:
"While features like timeouts, x, y are present in most implementations, their details vary since there are no standards (RFCs) around them. This makes these features less portable. So instead of pulling these into the API, we offer a middle ground: a standard way to plug these features in the API and offer a uniform UX across implementations."

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+
+Policies attached to Gateway API resources and implementations must use the
```
hbagdi, line 9:
```suggestion
Policies are attached to Gateway API resources and implementations must use the
```

Is this statement true? Can't one target a policy to a core.Service resource, which is not part of the gw api?

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+following approach to ensure consistency across implementations of the API.
+There are three primary components of this pattern:
```
hbagdi, line 11: nit: dedup into a single sentence?

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+* A hierarchy to illustrate how default and override values should interact.
+
+This kind of standardization not only enables consistent patterns, it allows
+future tooling such as kubectl plugins to be able to visualize all policies that
+have been applied to a given resource.
+
+## Policy Attachment for Ingress
+Attaching policy to Gateway resources for ingress use cases is relatively
+straightforward. A policy can reference the resource it wants to apply to.
+Access is granted with RBAC - anyone that has access to create a RetryPolicy in
```
hbagdi, line 25:
```suggestion
Access is granted with RBAC - for example, anyone that has access to create a RetryPolicy in
```

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,60 @@
...
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Kind string `json:"kind"`
+
+	// Name is the name of the target resource.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
```
hbagdi, line 41: Question: how do I target a policy to an entire namespace if this is required?

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Target Reference API
+
+Each Policy resource MUST include a single `targetRef` field. It MUST not
+target more than one resource at a time, but it can be used to target larger
+resources such as Gateways or Namespaces that may apply to multiple child
+resources.
+
+The `targetRef` field MUST be an exact replica of the `PolicyTargetReference`
+struct included in the Gateway API. Where possible, we recommend using that
```
hbagdi, line 65: nit
```suggestion
struct included in the Gateway API. Where possible, it is recommend to use that
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,290 @@
...
+struct directly instead of duplicating the type.
+
+### Policy Boilerplate
+The following structure can be used as a starting point for any Policy resource
```
hbagdi, line 69:
```suggestion
The following structure must be used for any Policy resource
```

hbagdi: We should then in the code comments document which field can be omitted from a specific resource.

hbagdi: > in the code comments document which field can be omitted from a specific resource.

You have done this to a certain extent in the document itself below.

site-src/v1alpha2/references/policy-attachment.md
robscott: Yeah it's hard to describe this accurately, I tried to go with "resources and implementations" to be broad here. This is not meant to limit where these policies may be attached but instead require that policies attached to Gateway API resources and implementations use this pattern. Does that make sense?

apis/v1alpha2/policy_types.go
robscott: Need to add an example of this at some point, but my thought is that it would be like any other reference:

```yaml
- group: ""
  kind: "Namespace"
  name: "foo"
```

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated the comment + added some guidance that policy resources could choose to omit one of the "Override" or "Default" fields.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: The sentence before is pretty long, haven't found a great way to dedup yet.

site-src/v1alpha2/references/policy-attachment.md
robscott: Great idea, refactored the first paragraph to look like this.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,60 @@
...
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+
+	// ClassName is the name of the class this policy should apply to. When
+	// unspecified, the policy will apply to all classes that support it.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	ClassName string `json:"className,omitempty"`
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Namespace is the namespace of the referent. When unspecified, the local
+	// namespace is inferred. Even when policy targets a resource in a different
+	// namespace, it may only apply to traffic originating from the same
+	// namespace as the policy.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Name is the name of the target resource.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the referent. When unspecified, the local
+	// namespace is inferred. Even when policy targets a resource in a different
+	// namespace, it may only apply to traffic originating from the same
```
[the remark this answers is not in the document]
robscott: Yep, you're completely right, switched to "must".

site-src/v1alpha2/references/policy-attachment.md
hbagdi: That makes sense. Being broad makes sense here.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+### Hierarchy
+Each policy MAY include default or override values. Overrides enable admins to
+enforce policy from the top down. Defaults enable app owners to provide default
+values from the bottom up for each individual application.
+
+![Policy Hierarchy](/images/policy/hierarchy.png)
```
hbagdi, line 127: Replacing sidecar with mesh in this diagram might be for the best.
Mesh is a concept. Sidecar is a deployment paradigm.

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+
+To illustrate this, consider 3 resources with the following hierarchy:
+A > B > C. When attaching the concept of defaults and overrides to that, the
```
hbagdi, line 130:
```suggestion
A (highest) > B > C. When attaching the concept of defaults and overrides to that, the
```

apis/v1alpha2/policy_types.go
```diff
@@ -0,0 +1,60 @@
...
+	// Name is the name of the target resource.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the referent. When unspecified, the local
+	// namespace is inferred. Even when policy targets a resource in a different
+	// namespace, it MUST only apply to traffic originating from the same
+	// namespace as the policy.
```
hbagdi, line 46: So, let's say we have gateway-a in gateway-ns namespace and route-b in route-ns namespace.

If I've a policy that in gateway-ns that targets a route-b and I also have a policy in route-ns that targets route-b, the hierarchy is policy in gateway-ns > policy in route-ns.
Is that right?

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+hierarchy would be expanded to this:
+
+A override > B override > C override > C default > B default > A default.
```
hbagdi, line 133: I just realized that if someone has RBAC permissions to XPolicy in a namespace, they can essentially target anything in that namepace and traffic originating from that namespace.

hbagdi: No action required but just noting.

hbagdi (commented):
nits and questions, this LGTM.

site-src/v1alpha2/references/policy-attachment.md
robscott: Yep, it's a different security model, and something we should be clear about. For implementations, I think it means policy resources should be designed with a coherent set of permissions in mind, and for users, it means they should consider this when determining where to place Gateways and Routes.

apis/v1alpha2/policy_types.go
robscott: Yes, that's my interpretation, but you've got me thinking that we may want to add more restrictions around cross-namespace policy attachment. I think the primary use case was for a mesh implementation to be able to apply policy to requests originating from the local namespace to a given destination. There's clearly also use cases for Gateway as you described above, but they may become messier than intended and/or be difficult for an implementation to support. I was about to suggest that we should make this an "Extended" concept, but the whole idea of policy attachment is already extended. So that's a long way of saying that I think your interpretation is correct, but we probably need to communicate that implementations can choose to not support cross-namespace policy.

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+  example, foo/bar is given precedence over foo/baz.
+
+### Kubectl Plugin
+To help improve UX and standardization, a kubectl plugin will be developed that
+will be capable of describing the computed sum of policy that applies to a given
+resource, including policies applied to parent resources.
+
+Each Policy CRD that wants to be supported by this plugin will need to follow
+the API structure defined above and add a
+`gateway.networking.k8s.io/policy-attachment: true` label to the CRD.
```
Miciah, line 235: Is it too late to get rid of the Boolean, maybe just use an empty string?  Booleans are annoying in yaml.  
```suggestion
`gateway.networking.k8s.io/policy-attachment: ""` label to the CRD.
```

apis/v1alpha2/policy_types.go
Miciah: > Need to add an example of this at some point, but my thought is that it would be like any other reference:
> 
> ```yaml
> - group: ""
>   kind: "Namespace"
>   name: "foo"
> ```

Interesting.  If you mean "target all resources within a namespace", I woulda thunk I'd need this:

```yaml
- group: ""
  kind: ""
  name: ""
  namespace: "foo"
```

Miciah: Anyway, doesn't at least `Group` need to have `MinLength=0`, or be an optional pointer value?

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+given scope will be able to modify policy at that level. Since GatewayClass is a
+cluster scoped resource, this means that any policy attached to it must also be
+cluster scoped.
+
+GatewayClass parameters provide an alternative to policy attachment that may be
+easier for some implementations to support. These parameters can similarly be
+used to set defaults and requirements for an entire GatewayClass.
+
+### Targeting External Services
+In some cases (likely limited to mesh) users may want to apply policies to
```
howardjohn, line 197: nit: its not really limited to mesh. Egress gateways are a good example. Also sometimes users proxy external traffic back to external sites for various reasons

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -0,0 +1,294 @@
...
+It is possible for multiple policies to target the same resource. When this
+happens, merging is the preferred outcome. If multiple policy resources target
+the same resource _and_ have an identical field specified with different values,
+precedence MUST be determined in order of the following criteria, continuing on
+ties:
+
+* The oldest Policy based on creation timestamp. For example, a Policy with a
+  creation timestamp of "2021-07-15 01:02:03" is given precedence over a Policy
+  with a creation timestamp of "2021-07-15 01:02:04".
+* The Policy appearing first in alphabetical order by "<namespace>/<name>". For
```
howardjohn, line 225: nit: during a conflict namespace should always be the same, right?  due to the hierarchy.

howardjohn (approved):

site-src/v1alpha2/references/policy-attachment.md
robscott: Updated, thanks!

site-src/v1alpha2/references/policy-attachment.md
robscott: I think that's likely true, but I'd prefer to avoid any possibility of conflict here if any form of hierarchy changes in the future. This is the same guidance we've used for other conflict resolution throughout the API, so I think it makes sense to keep it the same.

apis/v1alpha2/policy_types.go
hbagdi: > but we probably need to communicate that implementations can choose to not support cross-namespace policy.

Since we are not sure, I recommend keeping cross-namespace policy for gateway out of scope. There is a need for sure but punting on would be for the best. There are a lot of other features here and we shouldn't feel pressured.

PR #739, opened 2021-07-28
Clarifying TLS Spec on HTTPRoute, Moving to Extended Support

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This PR clarifies the TLS spec on HTTPRoute and moves it to extended support. This has been one of the more confusing  parts of the API, and it seemed like an update to the spec around it could help. I think there are at least some implementations that have reservations around supporting this feature, so I'm also proposing dropping the support level to "Extended" here.

```release-note
Usage of TLS Certificates in HTTPRoutes has been clarified, support level has changed from "Core" to "Extended".
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -113,21 +113,26 @@ type HTTPRouteSpec struct {
 
 	// TLS defines the TLS certificate to use for Hostnames defined in this
 	// Route. This configuration only takes effect if the AllowRouteOverride
-	// field is set to true in the associated Gateway resource.
+	// field is set to true in the associated Gateway Listener.
 	//
-	// Collisions can happen if multiple HTTPRoutes define a TLS certificate
-	// for the same hostname. In such a case, conflict resolution guiding
-	// principles apply, specifically, if hostnames are same and two different
-	// certificates are specified then the certificate in the
-	// oldest resource wins.
+	// Attaching a TLS certificate to HTTPRoutes provides a way for HTTPRoute
+	// owners to effectively populate certificates on Gateway Listeners. Note
+	// that only one certificate may be attached to a Gateway Listener for a
+	// specified hostname. It is not possible to use different certificates for
+	// different HTTPRoutes for the same hostname on the same Gateway Listener.
 	//
-	// Please note that HTTP Route-selection takes place after the
-	// TLS Handshake (ClientHello). Due to this, TLS certificate defined
-	// here will take precedence even if the request has the potential to
-	// match multiple routes (in case multiple HTTPRoutes share the same
-	// hostname).
+	// If multiple HTTPRoutes define a TLS certificate for the same hostname and
+	// are bound to the same Gateway Listener, precedence MUST be determined in
+	// order of the following criteria, continuing on ties:
 	//
-	// Support: Core
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2021-07-28 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2021-07-28 01:02:04".
+	// * The Route appearing first in alphabetical order by
+	//   "<namespace>/<name>". For example, foo/bar is given precedence over
+	//   foo/baz.
+	//
+	// Support: Extended
 	//
 	// +optional
 	TLS *RouteTLSConfig `json:"tls,omitempty"`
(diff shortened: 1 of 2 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -113,21 +113,26 @@ type HTTPRouteSpec struct {
...
+	// field is set to true in the associated Gateway Listener.
 	//
-	// Collisions can happen if multiple HTTPRoutes define a TLS certificate
-	// for the same hostname. In such a case, conflict resolution guiding
-	// principles apply, specifically, if hostnames are same and two different
-	// certificates are specified then the certificate in the
-	// oldest resource wins.
+	// Attaching a TLS certificate to HTTPRoutes provides a way for HTTPRoute
+	// owners to effectively populate certificates on Gateway Listeners. Note
+	// that only one certificate may be attached to a Gateway Listener for a
```
bowei, line 120: do we need to talk about how wildcards are handled?

hbagdi:
> I think there are at least some implementations that have reservations around supporting this feature, so I'm also proposing dropping the support level to "Extended" here.

If I understand correctly, support levels have been mostly been decided based on implementation feasibility and not on reservations on whether something is good or not.
Could you please elaborate on your concerns here?

We can clarify if the spec seems ambiguous or revisit this part of the API entirely if need be.

apis/v1alpha2/httproute_types.go
```diff
@@ -113,21 +113,26 @@ type HTTPRouteSpec struct {
...
-	// match multiple routes (in case multiple HTTPRoutes share the same
-	// hostname).
+	// If multiple HTTPRoutes define a TLS certificate for the same hostname and
+	// are bound to the same Gateway Listener, precedence MUST be determined in
+	// order of the following criteria, continuing on ties:
 	//
-	// Support: Core
+	// * The oldest Route based on creation timestamp. For example, a Route with
+	//   a creation timestamp of "2021-07-28 01:02:03" is given precedence over
+	//   a Route with a creation timestamp of "2021-07-28 01:02:04".
```
ccfishk, line 130: Shouldn't we apply latest timestamp win previous timestamp, which is similar to overwrite. 
Do I miss any restrict/rules somewhere ?

jpeach: https://gateway-api.sigs.k8s.io/concepts/guidelines/#conflicts

Oldest-wins is least likely to break existing configurations.

apis/v1alpha2/httproute_types.go
```diff
@@ -113,21 +113,26 @@ type HTTPRouteSpec struct {
...
 	//
-	// Collisions can happen if multiple HTTPRoutes define a TLS certificate
-	// for the same hostname. In such a case, conflict resolution guiding
-	// principles apply, specifically, if hostnames are same and two different
-	// certificates are specified then the certificate in the
-	// oldest resource wins.
+	// Attaching a TLS certificate to HTTPRoutes provides a way for HTTPRoute
+	// owners to effectively populate certificates on Gateway Listeners. Note
+	// that only one certificate may be attached to a Gateway Listener for a
+	// specified hostname. It is not possible to use different certificates for
```
jpeach, line 121: Only one _**secret**_ can be attached to a listener, but we don't specify what is in that secret. You could have a whole pile of certificates if you want :)

apis/v1alpha2/httproute_types.go
```diff
@@ -113,21 +113,26 @@ type HTTPRouteSpec struct {
...
+	// different HTTPRoutes for the same hostname on the same Gateway Listener.
 	//
-	// Please note that HTTP Route-selection takes place after the
-	// TLS Handshake (ClientHello). Due to this, TLS certificate defined
-	// here will take precedence even if the request has the potential to
-	// match multiple routes (in case multiple HTTPRoutes share the same
-	// hostname).
+	// If multiple HTTPRoutes define a TLS certificate for the same hostname and
+	// are bound to the same Gateway Listener, precedence MUST be determined in
```
jpeach, line 125: Are you are referring to the precedence of the certificate here?

So in this case
* all the routes are valid
* only one of the matching certificates is configured on the proxy
* there's no update to the route's status conditions

robscott:
Adding a hold on this while we discuss #749.

robscott:
With #749 merged in, this PR is irrelevant.

/close

Issue #582, opened 2021-03-11
Allow cross namespace references


howardjohn:
**What would you like to be added**:
Cross namespaces references in Route and BackendPolicy.

Examples:

A gateway may want to expose a route for a service in another namespace. For example, maybe we set up a dedicated gateway namespace a define a few routes to expose Services in other namespaces. This may sound like a security issue (a user can expose my secret Service externally without my permission), but it isn't inherently so - a user could already do this by manually configuring the proxy. If the user actually wants to restrict traffic from other namespaces, like our `gateway` namespace here, they need to do that in NetworkPolicy or similar, not assume someone won't proxy external traffic to their service:
```yaml
apiVersion: networking.x-k8s.io/v1alpha1
kind: BackendPolicy
metadata:
  name: policy
  namespace: gateway
spec:
  backendRefs:
  - name: foo
    kind: service
    namespace: bar # NEW: cross namespace reference
    port: 80
---
apiVersion: networking.x-k8s.io/v1alpha1
kind: HTTPRoute
metadata:
  name: foo-route
  namespace: gateway
spec:
  hostnames:
  - foo.example.com
  rules:
  - forwardTo:
    - serviceName: foo
      namespace: bar
      port: 80
```

A user may migrate to a new namespace and need to do a traffic split
```yaml
apiVersion: networking.x-k8s.io/v1alpha1
kind: HTTPRoute
metadata:
  name: foo-route
  namespace: gateway
spec:
  hostnames:
  - foo.example.com
  rules:
  - forwardTo:
    - serviceName: foo
      namespace: bar
      port: 80
      weight: 2
    - serviceName: foo-new
      namespace: bar-new
      port: 80
      weight: 1
```

robscott:
I think the cross namespace traffic-splitting is a pretty compelling use case. This actually feels somewhat related to https://github.com/kubernetes-sigs/gateway-api/issues/573 (cc @nak3).

My concern is that it risks making the API too complex. We already provide one place in the API for cross-namespace references (Gateway->Route). Adding a second cross-namespace reference (Route->Backend) could potentially become confusing. 

With the Gateway->Route relationship, we added a two way handshake that required both the Route and Gateway to effectively opt-in to references from/to a different namespace. In the Route->Backend relationship, we don't control both resources, so we couldn't achieve the same kind of handshake. Instead it seems like our best way to limit this functionality would be some kind of opt-in on either the Gateway or GatewayClass resource.

[...]

PR #741, opened 2021-08-02
Implements GEP 709 - ReferencePolicy

### LINKED ISSUE #709: GEP: Enable cross namespace references from Routes

youngnick:
**What type of PR is this?**

/kind feature
/kind documentation

This implements the design from GEP 709.

It adds a couple of extra things we missed in the GEP, that I noticed while implementing:
- Condition Reason for Gateway status and an explanation of how the status must work.
- Adds an `ObjectReference` type for cross-namespace references, to keep the existing `LocalObjectReference` type more similar to the core one.
 
Fixes #709 

```release-note
Added a ReferencePolicy object for controlling cross-namespace references.
```


apis/v1alpha2/object_reference_types.go
```diff
@@ -36,3 +36,33 @@ type LocalObjectReference struct {
...
+	// +kubebuilder:validation:MaxLength=253
+	Kind string `json:"kind"`
+
+	// Name is the name of the referent.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the backend. When unspecified, the local
```
jpeach, line 59:
```suggestion
// Namespace is the namespace of the backend. When not specified, the local
```

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+	//
+	// +optional
+	BackendRef *ObjectReference `json:"backendRef,omitempty"`
+
+    // The BackendRef in the more general RouteForwardTo object in shared_types.go
+    // has also been updated.
+```
+### Conformance Level
+
+ReferencePolicy support is a "CORE" coformance level for the following
```
jpeach, line 266:
```suggestion
ReferencePolicy support is a "Core" conformance level for the following
```

apis/v1alpha2/httproute_types.go
```diff
@@ -762,10 +762,15 @@ type HTTPRouteForwardTo struct {
 	// The gateway status for this route should be updated with a
 	// condition that describes the error more specifically.
 	//
+	// If there is a cross-namespace reference to an *existing* object
+	// with no ReferencePolicy, the controller must ensure the "ResolvedRefs"
+	// condition on the Gateway is set to `status: true`, with the "RefNotPermitted"
```
jpeach, line 767: I think `ResolvedRefs` would be false in this case? i.e the reference could not be resolved because there was no policy that permitted access.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
jpeach, line 2: Time warp!

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+
+package v1alpha2
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=gateway-api,shortName=refpol
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```
jpeach, line 25: We can revisit this, but the age doesn't seem particularly useful?

apis/v1alpha2/object_reference_types.go
robscott: We've used "unspecified" in quite a few places around the API,  any reason for the suggested change?

site-src/v1alpha2/references/cross-namespace-references.md
robscott: I'm not sure which is better here, but we should be consistent. Our [docs around conformance levels](https://gateway-api.sigs.k8s.io/concepts/guidelines/#conformance) use all caps, but usually the spec does not. Do we have a strong preference one way or the other?

apis/v1alpha2/referencepolicy_types.go
robscott: I think it's a pretty standard column for k8s resources. I think it could be helpful to know when a resource was created.

apis/v1alpha2/gateway_types.go
```diff
@@ -837,6 +838,12 @@ const (
 	// not resolve any Routes, and the "ResolvedRefs" status condition
 	// should not be raised in that case.
 	ListenerReasonInvalidRoutesRef ListenerConditionReason = "InvalidRoutesRef"
+
+	// This reason is used with the "ResolvedRefs" condition when
+	// one of the Listener's Routes has a BackendRef to an object in
+	// another namespace, where the object in the other namespace does
+	// not have a ReferencePolicy explicitly allowing the reference.
+	ListenerReasonRefNotPermitted ListenerConditionReason = "RefNotPermitted"
```
robscott, line 846: Nit: This does not match the "ReferenceNotPermitted" name above. I think I prefer "Ref" here to match the other names.

apis/v1alpha2/referencepolicy_types.go
howardjohn: +1, should be on all resources.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+### API design decisions
+This proposed API is fairly straightforward, but comes with a few notable decisions:
+
+1. Each ReferencePolicy only supports a single From and To section.
+   Additional trust relationships must be modeled with additional ReferencePolicy resources.
+1. Resource names are intentionally excluded from this policy for simplicity and because they rarely provide any meaningful protection.
+   A user that is able to write to resources of a certain kind within a namespace can always rename
+   resources or change the structure of the resources to match a given policy.
+1. A single Namespace is allowed per "From" struct.
```
howardjohn, line 69: is "all"/"any" not a common use case? I see the concerns of making it too permissive but generally APIs offer some way to be fairly permissive, even if its not easy.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+    // * TLSRoute
+    // * UDPRoute
+    //
+    // +kubebuilder:validation:MinLength=1
+    // +kubebuilder:validation:MaxLength=253
+    Kind string `json:"kind"`
+}
+```
+
+### Benefits
```
robscott, line 170: I think this section is also unnecessary for the docs.

robscott (commented):
Thanks for the work on this @youngnick! Mostly LGTM.

apis/v1alpha2/referencepolicy_types.go
hbagdi, line 25: `namespace` of the policy itself might be helpful here.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ReferencePolicy identifies kinds of resources in other namespaces that are
+// trusted to reference the specified kinds of resources in the local namespace.
```
hbagdi, line 28: `local` is good but `same` might reduce the scope of confusion.
```suggestion
// trusted to reference the specified kinds of resources in the same namespace as the policy.
```

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+// Each ReferencePolicy can be used to represent a unique trust relationship.
+// Additional Reference Policies can be used to add to the set of trusted
+// sources of inbound references for the namespace they are defined within.
```
hbagdi, line 31:
```suggestion
// sources of inbound references for the namespace they are defined within.

// All cross-namespace references in Gateway API (with the exception of cross-namespace
// Gateway-route attachment) require a ReferencePolicy.
```

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+type ReferencePolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the desired state of ReferencePolicy.
+	Spec ReferencePolicySpec `json:"spec,omitempty"`
+
+	// Note that we are explicitly *excluding* ReferencePolicy status at the
+	// moment, as designing it is more difficult than it would seem.
+	// As it is an additive change, we can make changes later.
```
hbagdi, line 41: Write from an end-user view
```suggestion
	// Note that `Status` sub-resource has been excluded at the
	// moment as it was difficult to work out the design.
	// `Status` sub-resource may be added in future.
```

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+type ReferencePolicySpec struct {
+	// From describes the trusted namespaces and kinds that can reference the
+	// resources described in "To". Each entry in this list must be considered
+	// to be an additional place that references can be valid from, or to put
+	// this another way, entries must be combined using OR.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:validation:MinItems=1
+	From []ReferencePolicyFrom `json:"from"`
```
hbagdi, line 63: This should have a max item limit like every other list in this API.
16 seems like a reasonable default to start with. We can bump it up if required.
This limit might seem very low since each item represent a namespace. I personally think if someone is using this list more exhaustively than this, it is unlikely to yield good UX.

(I was going to argue for 8, I couldn't come up with a good reason to defend that)

hbagdi: The same comment applies to 'To' field.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,132 @@
...
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// ReferencePolicyTo describes what Kinds are allowed as targets of the
+// references.
+type ReferencePolicyTo struct {
+	// Group is the group of the referent.
+	//
+	// Support: Core
+	//
```
hbagdi, line 114: 1. Please mention empty group implies core.
2. Any meaningful defaulting we can provide here?

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
+# Cross namespace references and ReferencePolicy
+
+In the Gateway API, it is possible to have references between objects cross namespace boundaries.
+In particular, it's expected that Gateways and Routes may exist in different namespaces,
+or that Services may be referred to by Routes in a another namespace.
```
hbagdi, line 5: > that Gateways and Routes may

ReferencePolicy is not expected to be applied to Gateway-Route references so it would be best to avoid mentioning that here.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+However, this significantly violates the idea of a namespace as the edge of a trust domain.
+In order to bring cross-namespace references under the control of the owner of the referent object's namespace,
+the Gateway API has a ReferencePolicy object that must be created in the referent namespace for the reference to be successful.
```
hbagdi, line 9: nit: 80 char limit throughout the file

hbagdi: > object that must be created in the referent namespace

I know we use 'referent' a lot and assume that it is understood by our users but I think explaining the referent and referee would help here. This paragraph is very hard to parse for me even though I'm familiar with the concept.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+To put this another way, if an object is referred to from outside its namespace,
+the object's owner must create a ReferencePolicy object that describes how that reference is allowed.
+This page explains how this process works.
+
+In the past, we've seen that forwarding traffic across namespace boundaries is a desired feature,
+but without the kinds of safeguards proposed here, [vulnerabilities](https://github.com/kubernetes/kubernetes/issues/103675)
+can emerge.
```
hbagdi, line 17: Maybe start with this paragraph to build up to the problem we are trying to solve and then propose the solution.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+## ReferencePolicy
+
+To ensure that Gateway API is able to safely provide this functionality,
+we need to enforce a handshake mechanism that requires resources in both namespaces to agree to this reference.
+To accomplish that, a ReferencePolicy resource has been be introduced.
+
+![Reference Policy](images/referencepolicy.png)
+
+With this model, Routes are able to directly reference Routes and Services in other namespaces.
```
hbagdi, line 27: We don't have Route->Route yet so it would be best not to mention it here. Can spring confusion.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+  - group: networking.gateway.k8s.io
+    kind: HTTPRoute
+    namespace: foo
+  to:
+  - group: core
+    kind: Service
+```
+
+### API design decisions
+This proposed API is fairly straightforward, but comes with a few notable decisions:
```
hbagdi, line 62: nit:
```suggestion
While the API is simplistic in nature, it comes with a few notable decisions:
```

Avoid blanket statements that the user might not grasp.

site-src/v1alpha2/references/cross-namespace-references.md
hbagdi: +1 to both the comments above

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+1. Each ReferencePolicy only supports a single From and To section.
+   Additional trust relationships must be modeled with additional ReferencePolicy resources.
+1. Resource names are intentionally excluded from this policy for simplicity and because they rarely provide any meaningful protection.
+   A user that is able to write to resources of a certain kind within a namespace can always rename
+   resources or change the structure of the resources to match a given policy.
+1. A single Namespace is allowed per "From" struct.
+   Although a selector would be more powerful it may encourage unnecessarily insecure configuration.
```
hbagdi, line 70: I don't mind being more assertive here and seeing how it plays out in the community.
```suggestion
   Although a selector would be more powerful, it encourages unnecessarily insecure configuration.
```

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+    // has also been updated.
+```
+### Conformance Level
+
+ReferencePolicy support is a "CORE" coformance level for the following
+objects:
+- HTTPRoute
+- TLSRoute
+- TCPRoute
+- UDPRoute
```
hbagdi, line 271: I find this conformance guidance is overly broad and vague.
We should be more explicit here: XRoute -> cross-namespace service/secrets require RP resource.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+Other "ImplementationSpecific" objects and references are *strongly recommended* to also use this flow for cross-namespace references.
```
hbagdi, line 273: We should use "must" here rather than a recommendation.
I do understand that we can't enforce it.
A "recommendation" could be neglected as opposed to "must".
That also helps with security issues that come down in future.

hbagdi (commented):
Great stuff! Thanks Nick.

apis/v1alpha2/httproute_types.go
youngnick: 🤦 Yes, thanks.

apis/v1alpha2/object_reference_types.go
youngnick: I'll leave this one as is for now.

apis/v1alpha2/referencepolicy_types.go
youngnick: 🤦

apis/v1alpha2/referencepolicy_types.go
youngnick: I think that adding extra columns is a good idea, I can see a case for both the `from` namespace and the `from` GroupKind  and `To` GroupKind or something. That will need joining support for print columns though, which I don't think is ready yet.

I think we should leave this for now and come back and add once we've used this for a little bit, and have a better idea what's useful on the command-line.

apis/v1alpha2/referencepolicy_types.go
youngnick: That's much better, thanks.

apis/v1alpha2/referencepolicy_types.go
youngnick: Agreed, done.

apis/v1alpha2/referencepolicy_types.go
youngnick: 1. Agreed, done.
2. I think that the extra explicitness we get from not having a default is better UX. Happy to discuss more though.

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: Added a note on terminology at the start, I think that's a great idea.

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: Thanks, that is a bad habit of mine.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,273 @@
...
+
+1. Each ReferencePolicy only supports a single From and To section.
+   Additional trust relationships must be modeled with additional ReferencePolicy resources.
+1. Resource names are intentionally excluded from this policy for simplicity and because they rarely provide any meaningful protection.
+   A user that is able to write to resources of a certain kind within a namespace can always rename
+   resources or change the structure of the resources to match a given policy.
+1. A single Namespace is allowed per "From" struct.
+   Although a selector would be more powerful it may encourage unnecessarily insecure configuration.
+
+```go
```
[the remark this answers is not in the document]
youngnick: Yeah, good idea. Done.

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: I lifted this phrasing from one of the other GEPs - I don't mind changing, but if so, we should change the others as well.

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: I've clarified a bit, does this read better?

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: Agreed, I made it a recommendation precisely because we can't really enforce it, but putting it as a MUST is better.

youngnick:
Okay, thanks for the reviews everyone, updated. I think I got all the actions I could. Please take another look!

apis/v1alpha2/referencepolicy_types.go
robscott: Agree that this can be a later extension, one potential here would be to just include the following columns:

- from namespace
- from kind
- to kind

Although that would be ideal, we still don't have anyway to do this for a list of structs (kubectl would just output the first matching value for a jsonpath). I've added a note about this in https://github.com/kubernetes-sigs/gateway-api/issues/588.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,141 @@
...
+
+// ReferencePolicyTo describes what Kinds are allowed as targets of the
+// references.
+type ReferencePolicyTo struct {
+	// Group is the group of the referent.
+	// When empty, the "core" API group is inferred.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:validation:MinLength=1
```
robscott, line 124: Nit: I missed this in the GEP, but we can't enforce a min length of 1 for Group. We have to support "" for core.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Kind is the kind of the referent. Although implementations may support
+	// additional resources, the following types are part of the "Core"
+	// support level for this field:
+	//
+	// * Service
+	// * HTTPRoute
+	// * TCPRoute
+	// * TLSRoute
+	// * UDPRoute
```
robscott, line 136: Maybe we should remove these until we actually implement Route inclusion?

apis/v1alpha2/shared_types.go
```diff
@@ -107,10 +107,15 @@ type RouteForwardTo struct {
...
+	// If there is a cross-namespace reference to an *existing* object
+	// with no ReferencePolicy, the controller must ensure the "ResolvedRefs"
+	// condition on the Gateway is set to `status: false`, with the "RefNotPermitted"
+	// reason and not configure this route in the underlying implementation.
+	//
 	// Support: Custom
 	//
 	// +optional
-	BackendRef *LocalObjectReference `json:"backendRef,omitempty"`
+	BackendRef *ObjectReference `json:"backendRef,omitempty"`
```
robscott, line 118: This is on a collision course with GEP #718, but since there's not an implementation PR for that one yet, I think this one will likely make it in first.

robscott:
A few small things but otherwise LGTM, thanks @youngnick!

apis/v1alpha2/object_reference_types.go
```diff
@@ -36,3 +36,34 @@ type LocalObjectReference struct {
 	// +kubebuilder:validation:MaxLength=253
 	Name string `json:"name"`
 }
+
+// ObjectReference identifies an API object including its namespace.
+type ObjectReference struct {
+	// Group is the group of the referent.
+	// When empty, the "core" API group is inferred.
```
hbagdi, line 43: So interestingly, when the field is empty, it doesn't show up in 'kubectl get'. Just leaving a note here. No action required.

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,141 @@
...
+// ReferencePolicy identifies kinds of resources in other namespaces that are
+// trusted to reference the specified kinds of resources in the same namespace
+// as the policy.
+//
+// Each ReferencePolicy can be used to represent a unique trust relationship.
+// Additional Reference Policies can be used to add to the set of trusted
+// sources of inbound references for the namespace they are defined within.
+//
+// All cross-namespace references in Gateway API (with the exception of cross-namespace
+// Gateway-route attachment) require a ReferencePolicy.
```
hbagdi, line 36:
```suggestion
// Gateway-route attachment) require a ReferencePolicy.
//
// Support: Core
//
```

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,141 @@
...
+	Items           []ReferencePolicy `json:"items"`
+}
+
+// ReferencePolicySpec identifies a cross namespace relationship that is trusted
+// for Gateway API.
+type ReferencePolicySpec struct {
+	// From describes the trusted namespaces and kinds that can reference the
+	// resources described in "To". Each entry in this list must be considered
+	// to be an additional place that references can be valid from, or to put
+	// this another way, entries must be combined using OR.
```
hbagdi, line 63: nit: I personally got more confused. Feel free to reject.
```suggestion
	// to be an additional place that references can be valid from.
```

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,141 @@
...
+	// Support: Core
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	From []ReferencePolicyFrom `json:"from"`
+
+	// To describes the resources that may be referenced by the resources
+	// described in "From". Each entry in this list must be considered to be an
+	// additional place that references can be valid to, or to put this another
+	// way, entries must be combined using OR.
```
hbagdi, line 74: Same comment.

apis/v1alpha2/referencepolicy_types.go
hbagdi: +1

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,127 @@
...
+[CVE-2021-25740](https://github.com/kubernetes/kubernetes/issues/103675).
+Implementations of this API need to be very careful to avoid confused deputy
+attacks. ReferencePolicy provides a safeguard for that. Exceptions MUST only
+be made by implementations that are absolutely certain that other equally
+effective safeguards are in place.
+
+### Conformance Level
+
+ReferencePolicy support is a "CORE" conformance level requirement for
+cross-namespace references inside the following objects:
```
hbagdi, line 115:
```suggestion
cross-namespace references that originate from the following objects:
```

hbagdi (commented):
LGTM, some nits.

robscott:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/582

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,127 @@
...
+- TCPRoute
+- UDPRoute
+
+That is, all implemenations MUST use this flow for any cross namespaces
+in any of the core xRoute types, except as noted in the Exceptions section
+above.
+
+Other "ImplementationSpecific" objects and references MUST also use this flow
+for cross-namespace references, except as noted in the Exceptions section
+above.
```
robscott, line 127: More a note for me, but when I implement the Route-Gateway binding GEP, I'll need to update this to discuss the uniqueness of that relationship.

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,127 @@
+# Cross namespace references and ReferencePolicy
+
+### Terminology note
+When discussing the process of creating cross-namespace object references, this
+document and the documentation on the API itself refers to the object being
+referred to as "the referent object", using the meaning of "referent" to be
+"the person, thing, or idea that a word, phrase, or object refers to".[1](https://dictionary.cambridge.org/dictionary/english/referent)
```
robscott, line 7: This footnote looks a bit off in the deploy preview. Maybe either use `<sup>` tag or a full word like `(source)`.
https://deploy-preview-741--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/cross-namespace-references/

site-src/v1alpha2/references/cross-namespace-references.md
```diff
@@ -0,0 +1,127 @@
+# Cross namespace references and ReferencePolicy
+
+### Terminology note
```
robscott, line 3: This feels a bit out of place. Not sure how this would render with mkdocs, but maybe something like a blockquote or callout format with this content added in the Intro section right before/after "Referent" is mentioned for the first time.

robscott (commented):
Thanks for all the work on this @youngnick! Have a bunch of tiny nits, most are non-blocking though. Otherwise LGTM.

site-src/v1alpha2/references/cross-namespace-references.md
youngnick: Yeah, I didn't like it there either, but couldn't think of the right place to put it. I'll see how we can do a blockquote or something more inline instead.

youngnick: I think I found a "Note" syntax that should work, I'll check the deploy preview.

youngnick: I added in an "admonitions" extension, that we can use with:

```
!!! admonition
    blockquote text is indented by four spaces
```

robscott:
Thanks for all the work on this!

/lgtm

site-src/v1alpha2/references/cross-namespace-references.md
robscott: This looks perfect! Thanks for adding it in

PR #742, opened 2021-08-03
Moving method match to v1alpha2 example

robscott:
**What type of PR is this?**
/kind bug
/kind documentation

Missed this when I was reviewing #733 

```release-note
NONE
```

Note: I'm starting this PR to see if I can get `verify-examples-kind` to fail, will update to complete this transition and remove hold at that point.

/hold

### DIFF
--- a/examples/basic-http.yaml
+++ b/examples/basic-http.yaml
@@ -55,7 +55,6 @@ spec:
       path:
         type: Prefix
         value: /some/thing
-      method: GET
     forwardTo:
     - serviceName: my-service2
       port: 8080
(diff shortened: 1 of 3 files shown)


robscott:
For posterity, that added delay caused this test failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_gateway-api/742/pull-gateway-api-verify/1422370697329512448. That means we should be able to catch these regressions in the future with this test. I also created an upstream issue to track this: https://github.com/kubernetes/kubernetes/issues/104090.

hbagdi:
Interesting bug.
Regarding adding to v1a1 examples, I wish there was an automated way to flag if that directory was edited at all in a PR review.
/lgtm

robscott:
+1 Completely agree that it would be nice to flag changes there in PR reviews automatically. Removing hold on this one now that test is passing and temporary fix is in place.

/hold cancel

PR #743, opened 2021-08-03
Add READY and ADDRESS to gateway printer columns

howardjohn:
**What type of PR is this?**

/kind feature

Add new columns. This allows at glance finding out how to access the gateway, and if its ready to be accessed

Example:
```
NAME                  CLASS         ADDRESS        READY   AGE
coolgateway           something                            19m
coolgateway-foobar    foobar        10.149.54.25           19m
coolgatewayl-myclass  myclass                      True    19m
```

Ingress for comparison:
```
NAME      CLASS    HOSTS              ADDRESS          PORTS     AGE
ingress   <none>   foo.org            172.21.255.155   80        4d22h
```

### DIFF
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -25,6 +25,8 @@ import (
 // +kubebuilder:resource:categories=gateway-api,shortName=gtw
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Class",type=string,JSONPath=`.spec.gatewayClassName`
+// +kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.status.addresses[*].value`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // Gateway represents an instantiation of a service-traffic handling
--- a/apis/v1alpha1/gateway_types.go
+++ b/apis/v1alpha1/gateway_types.go
@@ -25,8 +25,6 @@ import (
 // +kubebuilder:resource:categories=gateway-api,shortName=gtw
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Class",type=string,JSONPath=`.spec.gatewayClassName`
-// +kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.status.addresses[*].value`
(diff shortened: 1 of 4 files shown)


robscott:
This looks great, thanks @howardjohn!

/ok-to-test
/approve

robscott:
Actually, missed that there were changes to v1a1 CRDs. Although these are not breaking in any way, I think we should generally consider v1alpha1 frozen.

/hold

apis/v1alpha1/gateway_types.go
```diff
@@ -25,6 +25,8 @@ import (
 // +kubebuilder:resource:categories=gateway-api,shortName=gtw
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Class",type=string,JSONPath=`.spec.gatewayClassName`
+// +kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.status.addresses[*].value`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
```
robscott, line 29: Suggest removing from v1alpha1 since we haven't made any other v1alpha1 changes since v0.3.0

howardjohn: done

apis/v1alpha2/gateway_types.go
```diff
@@ -26,6 +26,8 @@ import (
 // +kubebuilder:subresource:status
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="Class",type=string,JSONPath=`.spec.gatewayClassName`
+// +kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.status.addresses[*].value`
```
robscott, line 29: Chatted a bit about this, I'm a bit concerned that this would lead to more confusion than it's worth since it will only be able to print the first address. I'd lean towards removing this instead.

robscott: Actually maybe this is fine. It will be remarkably better for >99% of use cases where there will only be one address, so optimizing for that probably makes sense.

hbagdi: Is there a way to list this only when `-o wide` is applied?

> Actually maybe this is fine. It will be remarkably better for >99% of use cases where there will only be one address, so optimizing for that probably makes sense.

In that case, why not do `addresses[0].value`?

robscott: I think it's fine to include this in regular output to match ingress, agree that `addresses[0].value` makes more sense given that that's how `*` would effectively be interpreted.

hbagdi:
Let's optimize for the regular case.
We can revisit this during beta if required.
/lgtm

PR #745, opened 2021-08-04
Remove RouteTLSConfig requirement for gateway TLS passthrough.

jpeach:
TLS passthrough never terminates TLS sessions, so it doesn't make
sense for TLS certificates to be required in this configuration.

/kind bug

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -147,9 +147,6 @@ type RouteTLSConfig struct {
 	// handshake for requests that match the hostname of the associated HTTPRoute.
 	// The referenced object MUST reside in the same namespace as HTTPRoute.
 	//
-	// This field is required when the TLS configuration mode of the associated
-	// Gateway listener is set to "Passthrough".
-	//
 	// CertificateRef can reference a standard Kubernetes resource, i.e. Secret,
 	// or an implementation-specific custom resource.
 	//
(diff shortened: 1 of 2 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -147,9 +147,6 @@ type RouteTLSConfig struct {
 	// handshake for requests that match the hostname of the associated HTTPRoute.
 	// The referenced object MUST reside in the same namespace as HTTPRoute.
 	//
-	// This field is required when the TLS configuration mode of the associated
-	// Gateway listener is set to "Passthrough".
```
robscott, line 151: Would this be accurate if "Passthrough" was replaced with "Terminate"?

jpeach: So `certificateRef` is a required field (syntactically), and IIUC you would need to set the `RouteTLSConfig` if the listener protocol is `HTTPS`, the mode is `Terminate`, you don't want the listener to use the default certificate and there is a TLS route override policy of `Allow`.

Phew! :joy:

robscott: Yeah, that's a good point. This comment never really made sense, and a replacement one is not particularly straightforward or valuable, I'm fine with just removing this.

robscott:
Thanks!

/lgtm

PR #750, opened 2021-08-06
implement GEP-718: rework RouteForwardTo

### LINKED ISSUE #718: GEP: consolidate serviceName shortcut in RouteForwardTo

hbagdi:
**What type of PR is this?**

/kind api-change

This PR implement GEP #718. Please refer to the GEP for more details.

Fixes #718

```release-note
API for forwarding requests from routes to backends has been reworked (breaking change).
```


apis/v1alpha2/udproute_types.go
```diff
@@ -92,12 +92,17 @@ type UDPRouteRule struct {
 	// +kubebuilder:validation:MaxItems=8
 	Matches []UDPRouteMatch `json:"matches,omitempty"`
 
-	// ForwardTo defines the backend(s) where matching requests should
-	// be sent.
+	// BackendRefs defines the backend(s) where matching requests should be
+	// sent. If unspecified, this rule MUST not match.
```
robscott, line 96: Same comment as above.

apis/v1alpha2/udproute_types.go
```diff
@@ -92,12 +92,17 @@ type UDPRouteRule struct {
...
 	//
-	// +kubebuilder:validation:MinItems=1
+	// Support: Core for Kubernetes Service
+	// Support: Custom for any other resource
+	//
+	// Support for weight: Extended
```
robscott, line 101: I think we can communicate the "extended" weight support on that field based on the updated guidelines: https://github.com/kubernetes-sigs/gateway-api/pull/723/files#diff-5733e6fd292e68a0c67232b5291bcc3275622ef579fe6dc7893ffb202c7e217d.

apis/v1alpha2/tcproute_types.go
```diff
@@ -92,12 +92,17 @@ type TCPRouteRule struct {
 	// +kubebuilder:validation:MaxItems=8
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
-	// ForwardTo defines the backend(s) where matching requests should
-	// be sent.
+	// BackendRefs defines the backend(s) where matching requests should be
+	// sent. If unspecified, this rule MUST not match.
```
robscott, line 96: This seems confusing to me. Why not just require a BackendRef for TCPRoute since it's currently useless without one?

apis/v1alpha2/tlsroute_types.go
```diff
@@ -127,12 +127,17 @@ type TLSRouteRule struct {
 	// +kubebuilder:validation:MaxItems=8
 	Matches []TLSRouteMatch `json:"matches,omitempty"`
 
-	// ForwardTo defines the backend(s) where matching requests should be
-	// sent.
+	// BackendRefs defines the backend(s) where matching requests should be
+	// sent. If unspecified, this rule MUST not match.
```
robscott, line 131: Same comment as above.

robscott (commented):
Thanks @hbagdi! This mostly LGTM, just note that this will conflict with #741 from @youngnick.

apis/v1alpha2/tcproute_types.go
hbagdi: I updated this to match HTTPRoute.
One use case: As a user, I want a placeholder TLS/UDP/TCP Route but I don't want to define my Service yet.

apis/v1alpha2/udproute_types.go
hbagdi: Can you elaborate a bit here? I'm not sure what is it that you are recommending to change and why.

robscott: Sorry that wasn't very clear. That linked guidance basically states that a support level of a given field is the max of that field and any parent support levels. That would allow you to annotate the weight field directly with `Support: Extended`.

apis/v1alpha2/tcproute_types.go
robscott: That just seems confusing to me since there's so little that's actually in these L4 route types. It seems like a placeholder L4 route would contain nothing at all, except maybe a reference to a Gateway, but that would be strange if you didn't already have backends. I'd lean towards keeping the validation strict here with the idea that we can always expand on it in the future.

apis/v1alpha2/udproute_types.go
hbagdi: > That would allow you to annotate the weight field directly with Support: Extended.

You mean Core? Weight field is set at Core for HTTPRoute.

apis/v1alpha2/tcproute_types.go
hbagdi: Okay. Let's punt this to later. Updated.

hbagdi:
> that this will conflict with #741 from @youngnick.

And so let's get this in quick. ;)

apis/v1alpha2/udproute_types.go
robscott: Oh woops, missed the context here, that is a tricky overlap. This seems fine then.

apis/v1alpha2/shared_types.go
```diff
@@ -77,53 +77,49 @@ type GatewayReference struct {
...
-	// when forwarding to a backendRef or serviceName.
+	// Name is the name of the referent.
 	//
-	// Support: Core
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Port specifies the destination port number to use for this resource.
+	// Port is required when the referent is a Kubernetes Service.
```
robscott, line 107: Created https://github.com/kubernetes-sigs/gateway-api/issues/753 to track webhook validation for this.

robscott:
Thanks @hbagdi!

/lgtm

PR #752, opened 2021-08-06
document support levels for each resource

hbagdi:
**What type of PR is this?**

/kind documentation
Documents conformance level for each resource.
This is something we discussed and assumed but I couldn't find any documentation surrounding it.

None

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -37,6 +37,9 @@ import (
 // finalizer on the associated GatewayClass whenever Gateway(s) is running.
 // This ensures that a GatewayClass associated with a Gateway(s) is not
 // deleted while in use.
+//
+// Support: Core
+//
 type Gateway struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/apis/v1alpha2/gatewayclass_types.go
+++ b/apis/v1alpha2/gatewayclass_types.go
@@ -34,6 +34,9 @@ import (
 // for creating Gateway resources.
 //
 // GatewayClass is a Cluster level resource.
+//
+// Support: Core
+//
 type GatewayClass struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -29,6 +29,9 @@ import (
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // HTTPRoute is the Schema for the HTTPRoute resource.
+//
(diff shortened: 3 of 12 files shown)


robscott:
I think this is largely accurate for the existing implementations of this API, but I think we're going to need a bit more nuance here as far as support levels. It seems at least plausible that there could be some L4 only implementations of the API in the future. Or maybe there are some mesh implementations that really only use the Routing and Policy concepts of this API.

hbagdi:
> It seems at least plausible that there could be some L4 only implementations of the API in the future. Or maybe there are some mesh implementations that really only use the Routing and Policy concepts of this API.

That probably will be a broader discussion.
Do you think we should document at least this verbal agreement right now and then solve the above after that?

robscott:
>That probably will be a broader discussion.
Do you think we should document at least this verbal agreement right now and then solve the above after that?

Yeah I'm not exactly sure how to summarize this yet, but I think we're going to broadly have different classes of implementations of this API that broadly fall into the following categories with the potential for a single implementation to cover all of these:

1. L7 ingress
2. L4 ingress
3. Mesh

For 1 and 2, both GatewayClass and Gateway resources would be considered core, and there would likely be no overlap in Route support. For 3, I'm assuming Mesh implementations would not have a use for Gateway, but it seems at least possible that both L4 and L7 routes may be relevant. I don't think our current conformance structure has a good way to communicate that, but if we are able to come up with something like conformance "profiles", maybe that would help?

jpeach:
This looks good to me, given the existing support level concepts.

/lgtm
/hold

hbagdi:
> Yeah I'm not exactly sure how to summarize this yet, but I think we're going to broadly have different classes of implementations of this API that broadly fall into the following categories with the potential for a single implementation to cover all of these:

I completely share that feeling.

This is documenting what we all understand and agree on today. You are proposing something that requires additional work.
I think mesh is something we are all trying to figure out. There are a lot of details that we don't know yet.
The thing is we can iterate here.

robscott:
Agree that we shouldn't hold off on this purely for mesh. I'd also like to see this API as something that makes sense for L4 implementations even if we don't have any right now. If we considered `HTTPRoute` core, wouldn't that rule out any purely L4 implementations?

hbagdi:
> If we considered HTTPRoute core, wouldn't that rule out any purely L4 implementations?

Indeed. We need a GEP to think about L4 implementation.
They are supported in a first-class fashion but our attention has been largely focused on HTTP.

youngnick:
I agree that this clearly documents the support levels for ingress controller implementations. I have spoken to people who are already planning on using this API for Layer 4 implementations, so I'm a bit worried about this PR effectively enshrining ingress controllers as the only supported implementations for now.

I think I'd rather leave this vague until we do the work to define conformance profiles (which has been on my personal wishlist for ages).

hbagdi:
Okay. Tracking in https://github.com/kubernetes-sigs/gateway-api/issues/764.
/close

PR #756, opened 2021-08-09
Clarify docs on invalid serviceName

### LINKED ISSUE #638: Missing serviceName behavior
I had another question on how we should handle errors (sorry for all the spam, but implementing is causing me to think about all these edge cases). 

If I have an HTTPRoute like the following example, and one of the `forwardTos` are invalid (say the service is missing from the cluster). The docs say I should raise a `ResolvedRefs` condition and the route must be dropped. 

My question is should the entire route be dropped or just that Match? If I had multiple matches for other paths, I'd ideally want to keep those serving traffic, but looking to be consistent since setting `Admitted: false` for a partial route also doesn't make sense. 

The [Implementation Guide](https://gateway-api.sigs.k8s.io/concepts/guidelines/) also says we should strive to serve traffic. I did have a quick conversation on Slack with @hbagdi regarding this (https://kubernetes.slack.com/archives/CR0H13KGA/p1612459180030400), so wanted to see what others thought on how to handle. 

> If the referent cannot be found, the route must be dropped from the Gateway. The controller should raise the “ResolvedRefs” condition on the Gateway with the “DegradedRoutes” reason. The gateway status for this route should be updated with a condition that describes the error more specifically.

Example: 
```
kind: HTTPRoute
metadata:
  name: http-filter-1
  namespace: projectcontour-roots 
  labels:
    type: http
spec:
  gateways:
    allow: All
  hostnames:
    - local.projectcontour.io
  rules:
  - matches:
      - path:
          type: Prefix
          value: /
      forwardTo:
      - serviceName: validservice    # <--- valid serviceName/match
        port: 80
    - matches:
      - path:
          type: Prefix
          value: /blog
      forwardTo:
      - serviceName: boohiss    # <--- invalid serviceName
        port: 80
```

howardjohn:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/638

**What type of PR is this?**

/kind documentation

Fixes #638

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -241,9 +241,11 @@ type HTTPRouteRule struct {
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent. If unspecified or invalid, the rule performs no forwarding; if
-	// no filters are specified that would result in a response being sent,
-	// a HTTP 503 status code is returned.
+	// sent. If unspecified or invalid (refers to a non-existent resource or a Service with no endpoints),
+	// the rule performs no forwarding; if no filters are specified that would result in a
+	// response being sent, a HTTP 503 status code is returned. 503 responses must be sent so that the overall
+	// weight is respected; if an invalid backend is requested to have 80% of requests, then 80% of requests
+	// must get a 503 instead.
 	//
 	// Support: Core for Kubernetes Service
 	// Support: Custom for any other resource
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -241,7 +241,7 @@ type HTTPRouteRule struct {
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent. If unspecified, the rule performs no forwarding. If unspecified and
+	// sent. If unspecified or invalid, the rule performs no forwarding; if
 	// no filters are specified that would result in a response being sent,
 	// a HTTP 503 status code is returned.
 	//
--- a/apis/v1alpha2/tcproute_types.go
+++ b/apis/v1alpha2/tcproute_types.go
@@ -93,7 +93,9 @@ type TCPRouteRule struct {
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent.
+	// sent. If unspecified or invalid, the rule performs no forwarding; if
+	// no filters are specified that would result in a response being sent,
+	// a HTTP 503 status code is returned.
 	//
 	// Support: Core for Kubernetes Service
 	// Support: Custom for any other resource
--- a/apis/v1alpha2/tcproute_types.go
+++ b/apis/v1alpha2/tcproute_types.go
@@ -93,11 +93,13 @@ type TCPRouteRule struct {
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent. If unspecified or invalid (refers to a non-existent resource or a Service with no endpoints),
-	// the rule performs no forwarding; if no filters are specified that would result in a
-	// response being sent, the underlying implementation must actively reject connection attempts to this backend.
-	// Connection rejections must respect weight; if an invalid backend is requested to have 80% of connection,
-	// then 80% of connection must be rejected instead.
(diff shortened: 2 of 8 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -241,7 +241,7 @@ type HTTPRouteRule struct {
 	Filters []HTTPRouteFilter `json:"filters,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent. If unspecified, the rule performs no forwarding. If unspecified and
+	// sent. If unspecified or invalid, the rule performs no forwarding; if
```
hbagdi, line 244: Should we further clarify what invalid means?
Does a regular k8s Service with no endpoints count as invalid?

robscott: Good point, wouldn't hurt to clarify that a Service with no endpoints should also not have traffic forwarded for it either. Maybe something like this:

> A BackendRef is considered invalid when it refers to a non-existent resource or a Service that does not have any endpoints.

stevesloka: Would be nice to clarify too if you have multiple backendrefs what happens. From the discussion in the issue, you'd lad balance across some valid and some that return 503s?

apis/v1alpha2/tcproute_types.go
```diff
@@ -93,7 +93,9 @@ type TCPRouteRule struct {
 	Matches []TCPRouteMatch `json:"matches,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent.
+	// sent. If unspecified or invalid, the rule performs no forwarding; if
+	// no filters are specified that would result in a response being sent,
+	// a HTTP 503 status code is returned.
```
robscott, line 98: Missed this the first time, but the HTTP 503 bit doesn't make sense for TCP or other L4 protocols.

apis/v1alpha2/httproute_types.go
robscott: I personally think it makes more sense to leave the weighting as requested. IE if a user says that 80% of requests should go to an invalid backend, could we just make that 80% of requests 503? Although that's harsh, it may be better than the alternative of sending all traffic to an alpha endpoint, a failure that could potentially go unnoticed for a long time.

stevesloka: I just feel we need to call out whatever behavior is decided. In my head, you'd set status for the invalid but skip the invalid services for serving. But that's ok if others don't agree. 🙃

Given this model, if you only had one invalid service defined for a route, you'd serve 503s, bit would you set a status on the route?

jpeach: > I personally think it makes more sense to leave the weighting as requested. IE if a user says that 80% of requests should go to an invalid backend, could we just make that 80% of requests 503? Although that's harsh, it may be better than the alternative of sending all traffic to an alpha endpoint, a failure that could potentially go unnoticed for a long time.

This was the consensus in the linked issue #638.

You could set an unresolved refs condition somewhere to report the inability to resolve an invalid service name.

apis/v1alpha2/tcproute_types.go
hbagdi: Close the connection for TCP and ignore the packet for UDP?

youngnick: Maybe just "the rule must not be configured in the underlying implementation"?

apis/v1alpha2/tlsroute_types.go
```diff
@@ -128,7 +128,9 @@ type TLSRouteRule struct {
 	Matches []TLSRouteMatch `json:"matches,omitempty"`
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent.
+	// sent. If unspecified or invalid, the rule performs no forwarding; if
+	// no filters are specified that would result in a response being sent,
+	// a HTTP 503 status code is returned.
```
youngnick, line 133: As above with TCPRoute, it's not always given that TLSRoutes are using HTTP, so maybe something like this:
```suggestion
	// the underlying implementation should actively reject connection attempts to this backend,
	// either by sending a HTTP 503 status code, or closing the connection for non-HTTP TLSRoutes.
```

apis/v1alpha2/httproute_types.go
youngnick: So, something like this?
```suggestion
	// sent. If unspecified or invalid (refers to a non-existent resource or a Service with no endpoints),
	// the rule performs no forwarding; if
```
With another bit added afterwards:
```go
	// 503 responses must be sent so that the overall weight is respected; if an invalid backend is requested
	// to have 80% of requests, then 80% of requests must get a 503 instead.
```

robscott:
I think this update lines up well with all the feedback above, thanks @howardjohn!

/lgtm

PR #757, opened 2021-08-09
version selection on docs site

hbagdi:
**What type of PR is this?**

/kind cleanup
Provides a version selector in the sidebar.

None

```release-note
NONE
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -15,6 +15,9 @@ plugins:
 markdown_extensions:
   - admonition
 nav:
+  - VERSIONS:
+      v1alpha1: /
+      v1alpha2: /v1alpha2
   - Introduction: index.md
   - Concepts:
       API overview: concepts/api-overview.md


hbagdi:
I tried to put this in the header but couldn't get that to work.

robscott:
@hbagdi Do you know if there's a way to make this work for the rest of the links in the navigation? The v1alpha2 link gets me to the main v1alpha2 page but all the navigation still takes me back to v1alpha1. It looks like versioning is supported with mkdocs but would require some more effort + an additional tool: https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/.

youngnick:
I had a bit of a look at this, I think I'm going to push ahead with a bit of a redesign here - there are bunch of things that I don't think need to be versioned, and some that definitely do. I'm going to try and push the versioned stuff into separate versioned directories, and that should help with fixing up the nav.

I've got something like this so far:
image

I also spent some time looking into using `mkdocs-material`'s recommended versioning option, [`mike`](https://github.com/jimporter/mike), but it would require very substantial changes to the site build processes, and I'm not sure how they would be wired into the Prow infra or Netlify. It would probably be better in the long run to put further effort into migrating to Hugo or something, and taking more cues from k/k.

I'll post a draft PR once I've got something worth looking at, so we can all check the deployment preview and see what we think.

hbagdi:
> @hbagdi Do you know if there's a way to make this work for the rest of the links in the navigation? The v1alpha2 link gets me to the main v1alpha2 page but all the navigation still takes me back to v1alpha1. It looks like versioning is supported with mkdocs but would require some more effort + an additional tool: https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/.

I looked for that too but couldn't find an easy way out so I thought of putting this up instead.

hbagdi:
No longer needed with the various other improvements being shipped currently.
/close

PR #765, opened 2021-08-11
document multi-value headers as undefined

### LINKED ISSUE #213: bug: multiple header value matching semantics
HTTP headers can have multiple values for the same key.  
Clarify what should be the route matching behavior for headers with multiple values to avoid surprises for the end user when they switch from one implementation to other.

A good example for this is matching based on `X-forwarded-for` header.

Questions to be answered:
- Should a single match for value among a set of values be considered a positive match?
- Should it be required to list out all values in case of multiple headers for a valid match? Is ordering important in this case?

hbagdi:
**What type of PR is this?**

Clarifies that matching behavior for multi-value headers is undefined. This seems like a very rare use-case and specifying something that may cause conformance headaches doesn't seem to be worth it.

Fixes #213

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -343,11 +343,17 @@ type HTTPHeaderMatch struct {
 	// Name is the name of the HTTP Header to be matched. Name matching MUST be
 	// case insensitive. (See https://tools.ietf.org/html/rfc7230#section-3.2).
 	//
-	// If multiple entries specify equivalent header names, only the first entry
-	// with an equivalent name MUST be considered for a match. Subsequent
+	// If multiple entries specify equivalent header names, only the first
+	// entry with an equivalent name MUST be considered for a match. Subsequent
 	// entries with an equivalent header name MUST be ignored. Due to the
 	// case-insensitivity of header names, "foo" and "Foo" are considered
 	// equivalent.
+	//
+	// When a header is repeated in an HTTP request, it is
+	// implementation-specific behavior as to how this is represented.
+	// Generally, proxies should follow the guidance from the RFC:
+	// https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2 regarding
+	// processing a repeated header, with special handling for "Set-Cookie".
 	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=256
--- a/config/crd/bases/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/bases/gateway.networking.k8s.io_httproutes.yaml
@@ -872,7 +872,13 @@ spec:
                                     name MUST be considered for a match. Subsequent
                                     entries with an equivalent header name MUST be
                                     ignored. Due to the case-insensitivity of header
-                                    names, \"foo\" and \"Foo\" are considered equivalent."
+                                    names, \"foo\" and \"Foo\" are considered equivalent.
+                                    \n When a header is repeated in an HTTP request,
+                                    it is implementation-specific behavior as to how
+                                    this is represented. Generally, proxies should
+                                    follow the guidance from the RFC: https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2
+                                    regarding processing a repeated header, with special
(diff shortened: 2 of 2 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -349,6 +349,9 @@ type HTTPHeaderMatch struct {
 	// case-insensitivity of header names, "foo" and "Foo" are considered
 	// equivalent.
 	//
+	// Matching behavior for headers with multiple values (like `Forwarded-for`)
```
bowei, line 352: I was trying to get a PR together, but was hitting validation issues. This is what I wrote:

```
+++ b/apis/v1alpha2/httproute_types.go
@@ -343,12 +343,19 @@ type HTTPHeaderMatch struct {
        // Name is the name of the HTTP Header to be matched. Name matching MUST be
        // case insensitive. (See https://tools.ietf.org/html/rfc7230#section-3.2).
        //
-       // If multiple entries specify equivalent header names, only the first entry
-       // with an equivalent name MUST be considered for a match. Subsequent
+       // If multiple entries specify equivalent header names, only the first
+       // entry with an equivalent name MUST be considered for a match. Subsequent
        // entries with an equivalent header name MUST be ignored. Due to the
        // case-insensitivity of header names, "foo" and "Foo" are considered
        // equivalent.
        //
+       // When a header is repeated in an HTTP request, it is
+       // implementation-specific behavior as to how this is represented.
+       // Generally, proxies should follow the guidance from the RFC:
+       // https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2 regarding
+       // processing a repeated header values as a single comma (",")
+       // separated header, with special handling for "Set-Cookie".
+       //
        // +kubebuilder:validation:MaxLength=256
        Name string `json:"name"`
```

robscott: > If multiple entries specify equivalent header names, only the first entry with an equivalent name MUST be considered for a match. Subsequent entries with an equivalent header name MUST be ignored.

I think that interpretation would run into the same [vulnerability](https://github.com/envoyproxy/envoy/security/advisories/GHSA-2v25-cjjq-5f4w) that Envoy ran into:

>Attackers can set multiple values of a non-inline header, e.g. x-foo: bar and x-foo: baz. Affected Envoy components will only observe the first value, x-foo: bar, in matchers, but both x-foo: bar and x-foo: baz will be forwarded to upstreams. Upstreams may take both values into consideration, resulting in an inconsistency between Envoy’s request matching and the upstream view of the request.

>An example of how this might be exploited is if an Envoy filter validates both x-request-credential and x-request-scope headers, checking whether the given credentials match the request scopes. Only the first scope header would be validated, but the upstream may interpret all provided scopes as being valid.

robscott: For reference, their fix involved changing matching to use a comma separated list of header values, which seems like an appropriate requirement here: https://github.com/envoyproxy/envoy/blob/a12869fa9e9add4301a700978d5489e6a0cc0526/docs/root/version_history/v1.14.5.rst

apis/v1alpha2/httproute_types.go
```diff
@@ -349,6 +349,9 @@ type HTTPHeaderMatch struct {
...
+	// is not defined by this spec.
+	//
```
howardjohn, line 354: > Subsequent entries with an equivalent header name MUST be ignored.

So.. is someone going to change envoy to match like this or are we all just going to be non-compliant :slightly_smiling_face:

apis/v1alpha2/httproute_types.go
bowei: I think we have two things, maybe want to clarify more:

1. multiple HTTPHeaderMatch entries (which is what the first thing is talking about)
2. multiple of the same header in the HTTP request -- which is somewhat malformed except for the comma ones; and the handling of those is up to the implementation. Good secure implementations merge header together using commas and drop extra headers for fields that don't support commas etc and Set-Cookie is an odd duck.

robscott: That makes a lot of sense, I was confusing the two. Agree with your suggested updates given that distinction.

robscott:
Thanks!

/lgtm

PR #768, opened 2021-08-12
Implementing and Documenting GEP 746 (TLS Updates)

### LINKED ISSUE #746: GEP: Replace Certificate Refs on HTTPRoute with Cross Namespace Refs from Gateway
This proposes replacing Certificate Refs on HTTPRoute with the ability for Gateways to reference certificates in other namespaces. This is not a complete replacement, but it does significantly reduce the potential for confusion. 

**Why this is needed**:
This is a confusing part of the API that does not appear to be supported by any implementation yet. We can always choose to add this concept back if it's necessary, but leaving it in for v1alpha2 may leave us stuck with this feature for a long time.

This is a follow up on my [original proposal doc](https://docs.google.com/document/d/1Cv95XFCL6S_9pIyS0drnsDLsfinWc2tHOFl_x3-_SWI/edit).

robscott:
**What type of PR is this?**
/kind feature

This implements and documents GEP #746.

Fixes #746

```release-note
* Certificate refs from Gateways can now cross namespaces (with ReferencePolicy)
* Certificate refs no longer exist on HTTPRoute
```

Note: The CRD verification will fail on this PR until #754 gets in.

Deploy Preview: https://deploy-preview-768--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/guides/tls/

/cc @hbagdi @howardjohn @jpeach @Miciah @youngnick


apis/v1alpha2/gateway_types.go
```diff
@@ -316,8 +287,14 @@ type GatewayTLSConfig struct {
 
 	// CertificateRef is a reference to a Kubernetes object that contains a TLS
 	// certificate and private key. This certificate is used to establish a TLS
-	// handshake for requests that match the hostname of the associated listener.
-	// The referenced object MUST reside in the same namespace as Gateway.
+	// handshake for requests that match the hostname of the associated
+	// listener.
+	//
+	// When this references a resource in a different namespace than the
```
bowei, line 293: We should probably word this in the negative so it's clear:

CertificateRefs to a different namespace from the Gateway are denied UNLESS there
is a ReferencePolicy allowing the certificate to be attached. ...

robscott:
Thanks for the feedback @bowei! I think I've updated everything, PTAL.

examples/v1alpha2/tls-basic.yaml
```diff
@@ -0,0 +1,25 @@
...
+  gatewayClassName: acme-lb
+  listeners:
+  - name: foo-https
+    protocol: HTTPS
+    port: 443
+    hostname: foo.example.com
+    tls:
+      certificateRef:
+        kind: Secret
+        group: ""
```
jpeach, line 15: consider omitting `group` since the default is "core"

jpeach (approved):
/lgtm

examples/v1alpha2/tls-basic.yaml
robscott: I know we once had a default here, but it looks like we've removed it in favor of a generic ObjectRef, so I think this still needs to be specified.

Issue #629, opened 2021-04-23
Summary of out-of-tree problems


hbagdi:
This is an umbrella issue to track several problems that this project is running into because of it being out-of-tree.

- GatewayClass cannot be immutable https://github.com/kubernetes-sigs/gateway-api/issues/189.
- User-friendly `kubectl` output https://github.com/kubernetes-sigs/gateway-api/issues/588 https://github.com/kubernetes-sigs/gateway-api/pull/587 https://github.com/kubernetes-sigs/gateway-api/pull/586
- Lack of built-in CRD validation has required us to build and maintain a validating webhook. This also makes the installation and setup process for Gateway API a bit more cumbersome than the ideal case.
- Validation of kubernetes names requires going through k8s code and looking for regexes. Having API types defined for common types like meta.v1.Group meta.v1.Namespace etc would be helpful. Threads with more context: https://github.com/kubernetes-sigs/gateway-api/pull/772#discussion_r688140333, ObjectName in https://github.com/kubernetes-sigs/gateway-api/pull/871
- Validation of LabelSelector is not possible because it is an upstream type: https://github.com/kubernetes-sigs/gateway-api/issues/833#
- Controllers cannot trust and rely on validation of custom resources via webhook or CR definition as both of them can be bypassed. https://github.com/kubernetes-sigs/gateway-api/pull/923#discussion_r744031322

hbagdi:
This was discussed a while ago in one of the community meetings but I only got to this action item today.
There are several other issues that we have run into with CRDs but they are not logged and I can't recall others from the top of my head.

robscott:
Thanks for creating this! I think another key one is the lack of built-in CRD validation for many of our use cases which has required us to create and maintain a validating webhook.

bowei:
We need to feed a lot of these requirements to api machinery and CRD.

Maybe we should create x-ref issues on their repo(s)?

youngnick:
I think that all of the things we've mentioned here are definitely problems for other CRD authors, so being able to get some solutions to them would be really helpful for everyone. Thanks for making sure this stuff gets tracked @hbagdi!

hbagdi:
> Maybe we should create x-ref issues on their repo(s)?

That's the plan but I was thinking of getting a fair representation of the problems we have in this project before doing that.
Are there others that come to mind?

robscott:
[...]

PR #772, opened 2021-08-12
Tightening validation + adding basic validation tests

### LINKED ISSUE #767: Stricter regex-based validation

Add regex based validation in the API: https://github.com/kubernetes-sigs/gateway-api/pull/754#discussion_r687174710

**Why this is needed**:

Decreasing the surface area of the API further. We can relax (not a breaking change) these validations later on when need be.

robscott:
**What type of PR is this?**
/kind cleanup

This adds regex validation and map keys throughout the v1alpha2 API, consolidates shared types, and adds some very simple validation tests. Much of this logic was pulled from [upstream Kubernetes validation](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go).

Fixes #767

```release-note
Many string fields are now validated with regular expressions.
```

/cc @hbagdi @Miciah @bowei

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -262,6 +262,7 @@ type HTTPPathMatch struct {
 	//
 	// +optional
 	// +kubebuilder:default="/"
+	// +kubebuilder:validation:MaxLength=1024
 	Value *string `json:"value,omitempty"`
 }
 
@@ -281,6 +282,24 @@ const (
 	HeaderMatchRegularExpression      HeaderMatchType = "RegularExpression"
 	HeaderMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
 )
+
+// HTTPHeaderName is the name of an HTTP header.
+//
+// Valid values include:
+//
+// * "Authorization"
+// * "Set-Cookie"
+//
+// Invalid values include:
+//
+// * ":method" - ":" is an invalid character. This means that pseudo headers are
+//   not currently supported by this type.
+// * "/invalid" - "/" is an invalid character
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=256
+// +kubebuilder:validation:Pattern=`^[A-Za-z0-9!#$%&'*+\-.^_\x60|~]+$`
+type HTTPHeaderName string
 
 // HTTPHeaderMatch describes how to select a HTTP route by matching HTTP request
 // headers.
@@ -314,10 +333,7 @@ type HTTPHeaderMatch struct {
 	// Generally, proxies should follow the guidance from the RFC:
 	// https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2 regarding
 	// processing a repeated header, with special handling for "Set-Cookie".
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=256
-	Name string `json:"name"`
+	Name HTTPHeaderName `json:"name"`
 
 	// Value is the value of HTTP Header to be matched.
 	//
@@ -423,14 +439,20 @@ type HTTPRouteMatch struct {
 	// ANDed together, meaning, a request must match all the specified headers
 	// to select the route.
 	//
+	// +listType=map
+	// +listMapKey=name
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
 	Headers []HTTPHeaderMatch `json:"headers,omitempty"`
 
 	// QueryParams specifies HTTP query parameter matchers. Multiple match
 	// values are ANDed together, meaning, a request must match all the
 	// specified query parameters to select the route.
 	//
+	// +listType=map
+	// +listMapKey=name
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
 	QueryParams []HTTPQueryParamMatch `json:"queryParams,omitempty"`
 
 	// Method specifies HTTP method matcher.
@@ -574,10 +596,7 @@ type HTTPHeader struct {
 	// entries with an equivalent header name MUST be ignored. Due to the
 	// case-insensitivity of header names, "foo" and "Foo" are considered
 	// equivalent.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=256
-	Name string `json:"name"`
+	Name HTTPHeaderName `json:"name"`
 
 	// Value is the value of HTTP Header to be matched.
 	//
@@ -604,6 +623,9 @@ type HTTPRequestHeaderFilter struct {
 	//   my-header: bar
 	//
 	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
 	Set []HTTPHeader `json:"set,omitempty"`
 
 	// Add adds the given header(s) (name, value) to the request
@@ -623,6 +645,9 @@ type HTTPRequestHeaderFilter struct {
 	//   my-header: bar
 	//
 	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
 	Add []HTTPHeader `json:"add,omitempty"`
 
 	// Remove the given header(s) from the HTTP request before the
@@ -659,6 +684,7 @@ type HTTPRequestRedirect struct {
 	// +optional
 	// +kubebuilder:validation:Enum=HTTP;HTTPS
 	Protocol *string `json:"protocol,omitempty"`
+
 	// Hostname is the hostname to be used in the value of the `Location`
 	// header in the response.
 	// When empty, the hostname of the request is used.
@@ -666,7 +692,8 @@ type HTTPRequestRedirect struct {
 	// Support: Core
 	//
 	// +optional
-	Hostname *string `json:"hostname,omitempty"`
+	Hostname *Hostname `json:"hostname,omitempty"`
+
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
 	// When empty, port (if specified) of the request is used.
@@ -674,14 +701,15 @@ type HTTPRequestRedirect struct {
 	// Support: Extended
 	//
 	// +optional
-	Port *int `json:"port,omitempty"`
+	Port *PortNumber `json:"port,omitempty"`
+
 	// StatusCode is the HTTP status code to be used in response.
 	//
 	// Support: Core
 	//
 	// +optional
 	// +kubebuilder:default=302
-	// +kubebuilder:validation=301;302
+	// +kubebuilder:validation:Enum=301;302
 	StatusCode *int `json:"statusCode,omitempty"`
 }
 
--- a/apis/v1alpha2/object_reference_types.go
+++ b/apis/v1alpha2/object_reference_types.go
@@ -19,16 +19,12 @@ package v1alpha2
 // LocalObjectReference identifies an API object within the namespace of the
 // referrer.
 type LocalObjectReference struct {
-	// Group is the group of the referent.
-	//
-	// +kubebuilder:validation:MaxLength=253
-	Group string `json:"group"`
+	// Group is the group of the referent. For example, "networking.k8s.io".
+	// When unspecified (empty string), core API group is inferred.
+	Group Group `json:"group"`
 
-	// Kind is kind of the referent.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind string `json:"kind"`
+	// Kind is kind of the referent. For example "HTTPRoute" or "Service".
+	Kind Kind `json:"kind"`
 
 	// Name is the name of the referent.
 	//
@@ -39,21 +35,18 @@ type LocalObjectReference struct {
 
 // ObjectReference identifies an API object including its namespace.
 type ObjectReference struct {
-	// Group is the group of the referent.
+	// Group is the group of the referent. For example, "networking.k8s.io".
 	// When unspecified (empty string), core API group is inferred.
 	//
 	// +optional
 	// +kubebuilder:default=""
-	// +kubebuilder:validation:MaxLength=253
-	Group *string `json:"group"`
+	Group *Group `json:"group"`
 
-	// Kind is kind of the referent.
+	// Kind is kind of the referent. For example "HTTPRoute" or "Service".
 	//
 	// +optional
 	// +kubebuilder:default=Service
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind *string `json:"kind"`
+	Kind *Kind `json:"kind"`
 
 	// Name is the name of the referent.
 	//
@@ -71,10 +64,8 @@ type ObjectReference struct {
 	//
 	// Support: Core
 	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
 	// +optional
-	Namespace *string `json:"namespace,omitempty"`
+	Namespace *Namespace `json:"namespace,omitempty"`
 }
 
 // BackendObjectReference defines how an ObjectReference that is
@@ -86,21 +77,18 @@ type ObjectReference struct {
 // owner to accept the reference. See the ReferencePolicy documentation
 // for details.
 type BackendObjectReference struct {
-	// Group is the group of the referent.
+	// Group is the group of the referent. For example, "networking.k8s.io".
 	// When unspecified (empty string), core API group is inferred.
 	//
 	// +optional
 	// +kubebuilder:default=""
-	// +kubebuilder:validation:MaxLength=253
-	Group *string `json:"group,omitempty"`
+	Group *Group `json:"group,omitempty"`
 
-	// Kind is kind of the referent.
+	// Kind is kind of the referent. For example "HTTPRoute" or "Service".
 	//
 	// +optional
 	// +kubebuilder:default=Service
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind *string `json:"kind,omitempty"`
+	Kind *Kind `json:"kind,omitempty"`
 
 	// Name is the name of the referent.
 	//
@@ -118,10 +106,8 @@ type BackendObjectReference struct {
 	//
 	// Support: Core
 	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
 	// +optional
-	Namespace *string `json:"namespace,omitempty"`
+	Namespace *Namespace `json:"namespace,omitempty"`
 
 	// Port specifies the destination port number to use for this resource.
 	// Port is required when the referent is a Kubernetes Service.
--- a/apis/v1alpha2/policy_types.go
+++ b/apis/v1alpha2/policy_types.go
@@ -23,16 +23,10 @@ package v1alpha2
 // for Gateway API.
 type PolicyTargetReference struct {
 	// Group is the group of the target resource.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Group string `json:"group"`
+	Group Group `json:"group"`
 
 	// Kind is kind of the target resource.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind string `json:"kind"`
+	Kind Kind `json:"kind"`
 
 	// Name is the name of the target resource.
 	//
@@ -45,10 +39,8 @@ type PolicyTargetReference struct {
 	// namespace, it MUST only apply to traffic originating from the same
 	// namespace as the policy.
 	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
 	// +optional
-	Namespace *string `json:"namespace,omitempty"`
+	Namespace *Namespace `json:"namespace,omitempty"`
 
 	// ClassName is the name of the class this policy should apply to. When
 	// unspecified, the policy will apply to all classes that support it.
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -29,22 +29,18 @@ type ParentRef struct {
 	//
 	// Support: Core
 	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:default=gateway.networking.k8s.io
 	// +optional
-	Group *string `json:"group,omitempty"`
+	Group *Group `json:"group,omitempty"`
 
 	// Kind is kind of the referent.
 	//
 	// Support: Core (Gateway)
 	// Support: Custom (Other Resources)
 	//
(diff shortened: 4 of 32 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,41 @@ type RouteStatus struct {
...
+type Kind string
+
+// Namespace refers to a Kubernetes namespace. It must be a RFC 1123 label.
+//
+// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=63
+type Namespace string
+
+// SectionName is the name of a section in a Kubernetes resource. It must be a
```
howardjohn, line 264: I don't think this should be restrictive. Its an extension point, section names in custom types may not have such strict validation

apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,41 @@ type RouteStatus struct {
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:Pattern=`^(\*.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
howardjohn, line 241: This matches IPv4 but not IPv6. Spec declares IPs are not allowd at all though..

hack/invalid-examples/httproute/invalid-path.yaml
```diff
@@ -0,0 +1,10 @@
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1alpha2
+metadata:
+  name: invalid-path
+spec:
+  rules:
+  - matches:
+    - path:
+        type: Prefix
+        value: /<>invalid
```
howardjohn, line 10: Why is this invalid? `<>` is valid in a url

hack/invalid-examples/gateway/duplicate-listeners.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Gateway
+apiVersion: gateway.networking.k8s.io/v1alpha2
+metadata:
+  name: duplicate-listeners
+spec:
+  gatewayClassName: acme-lb
+  listeners:
+  - name: same
```
howardjohn, line 8: how is this validated?

robscott: The wonders of `+listMapKey=name`

apis/v1alpha2/shared_types.go
robscott: Any particular suggestions for loosening this? This was just pulled from the upstream validation for an RFC 1123 subdomain that's used pretty widely throughout k8s. That seemed like a fairly generic starting point which we can always loosen later if necessary.

apis/v1alpha2/shared_types.go
robscott: Is there a regex that could allow all valid hostnames but still invalidate IPv4 addresses? For reference, this is pulled from the upstream kubernetes wildcard subdomain validation (with the exception that the leading `*.` is optional here.

hack/invalid-examples/httproute/invalid-path.yaml
robscott: Wouldn't this be an invalid path segment per the RFC? Regardless, this was a good reminder that my validation was too strict here simply because we allow RegExp and ImplementationSpecific matching here. I've removed this.

apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,41 @@ type RouteStatus struct {
...
 type Hostname string
+
+// Group refers to a Kubernetes Group.
+//
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:Pattern=`^$|^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+type Group string
```
hbagdi, line 248: I wish this was defined in upstream and we could simply pull it from there. Some other day. No action required.

apis/v1alpha2/policy_types.go
```diff
@@ -23,16 +23,10 @@ package v1alpha2
...
-	// +kubebuilder:validation:MaxLength=253
-	Group string `json:"group"`
+	Group Group `json:"group"`
 
 	// Kind is kind of the target resource.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind string `json:"kind"`
+	Kind Kind `json:"kind"`
```
hbagdi, line 29: rant: Why can't github highlight the specific text that has changed in this line.

apis/v1alpha2/shared_types.go
hbagdi: Start small and tight, loosen when needed.

apis/v1alpha2/httproute_types.go
```diff
@@ -696,30 +706,34 @@ type HTTPRequestRedirect struct {
 	//
 	// +optional
 	// +kubebuilder:validation:Enum=HTTP;HTTPS
+
```
hbagdi, line 709: we do not usually have this line jump

apis/v1alpha2/httproute_types.go
```diff
@@ -461,14 +466,20 @@ type HTTPRouteMatch struct {
 	// ANDed together, meaning, a request must match all the specified headers
 	// to select the route.
 	//
+	// +listType=map
+	// +listMapKey=name
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
 	Headers []HTTPHeaderMatch `json:"headers,omitempty"`
```
hbagdi, line 473: That is a lot of parameters to base a match on practically. No need to change, I don't feel strongly enough to ask for a change.

hbagdi (commented):
I'm really bad with regex so I didn't review them. I'm worse with regex than a person who knows nothing about them.
This PR looks good otherwise.
The deal with such PRs is that to review them you have to look at the part of the codebase that _hasn't_ changed. Will improve iteratively if something is missed.

Thanks for pulling this in last minute.

apis/v1alpha2/httproute_types.go
robscott: Yeah I almost started with something lower, but didn't have a great reason. We use 16 for a lot of similar limits so it seemed like a relatively non-controversial starting point.

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go
```diff
@@ -320,6 +321,13 @@ const (
 	HeaderMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
 )
 
+// HTTPHeaderName is the name of an HTTP header.
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=256
+// +kubebuilder:validation:Pattern=`^[-A-Za-z0-9]+$`
```
howardjohn, line 328: this forbids pseudo headers and a bunch of other valid things like periods, slashes, percent signs, etc. rfc7230 3.2.6 describes the full list

robscott: Interesting, are you able to understand from the RFC which characters are valid for a header name? It's hard for me to distinguish, since that specific section seems to be focused on the field value? If the `token` represents the name, it seems like the allowed set of chars would be a-z0-9 + this:

```
"!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" /  "." / "^" / "_" / "`" / "|" / "~" 
```

Then to further complicate things, the spec seems to explicitly rule out `:` as a value, which seems like it would be an issue for pseudo headers...

robscott: For reference, here's where the current regex is coming from: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L388-L400. You're completely right that it would prevent pseudo headers though.

howardjohn: I think all of those are valid.

The trick for pseudo header is that `:` is also the separator for the k/v - so in some cases it ends up being treated as `""` as the key. Not sure if its standard though: 
```
$ curl httpbin.org/get -H ':foo: bar'
{
  "args": {},
  "headers": {
    "": "foo: bar"
```

howardjohn: Also the RFC is hard to follow but if you look at https://datatracker.ietf.org/doc/html/rfc7230#section-3.2, a header key is a `token`, and a token is one of those you listed above. Looking at the golang source was easier than the RFC...

robscott: So to clarify, you're saying that we wouldn't need `:` to be a valid header name character?

howardjohn: If we want to support only _real_ headers - no `:`.

If we want to support pseudo headers though, then we do want it. So that would allow header match on `:path`, `:scheme`, etc.

I think either one is valid, but we should be deliberate. If all pseudoheaders are representable based on other first class fields, then its plausble we do not allow them

apis/v1alpha2/shared_types.go
youngnick: I don't think there is, because IPv4 addresses are still technically valid domain names as well, sadly.

apis/v1alpha2/shared_types.go
youngnick: +1 to starting with RFC1123 subdomain first, and seeing if we need to expand out from there.

apis/v1alpha2/httproute_types.go
robscott: I think there's been at least some discussion around pseudo header matching being useful in some cases, but since I don't think we've reached any consensus on that, it seems safer to leave it out with the possibility of adding it later if necessary.

youngnick: Allowing pseudoheader pros:
- allows people to quickly try out some more advanced routing things, before the API supports it.
Cons:
- less incentive to have the API support it, because you can use pseudoheaders.

I think it would be better long-term to have the functionality explicitly defined rather than do the pseudoheader thing, but it's not a strong preference.

youngnick (commented):
Overall LGTM, I'm inclined to err on the side of not allowing functionality we might want to take away later (in favour of explicit config).

apis/v1alpha2/httproute_types.go
robscott: Extended regex to include characters listed above, avoided pseudo headers for now as support will be much easier to add than remove in the future.

apis/v1alpha2/shared_types.go
bowei: I think trying to encode really complicated things into a regex will be problematic. It becomes quite hard to verify that a given regex does exactly what you want by looking at it. 

I would comment that this validation may not remove all cases and controller will need to check.

apis/v1alpha2/shared_types.go
bowei: Do we have a reference as to where this regex is using as the basis for tis definition?

apis/v1alpha2/shared_types.go
bowei: Which specific section of rfc1123 is this definition from? can you put a href? I'm searching the rfc for the word subdomain and I can't seem to find it easily.

 Isn't this just a RFC1123 domain name (not sub)? 

Also, we should give examples in each of the comments. People are not going to be able to stare at the regex and figure out what the intent is. This applies to all of the really complex validations.

apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,41 @@ type RouteStatus struct {
...
+// SectionName is the name of a section in a Kubernetes resource. It must be a
+// RFC 1123 subdomain.
+//
+// +kubebuilder:validation:Pattern=`^$|^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type SectionName string
+
+// GatewayController is the name of a Gateway API controller. It must be a
+// domain prefixed path.
```
bowei, line 273: Add example

bowei:
LGTM -- but please put examples in each of the docstrings. the regexes are pretty complicated to parse to understand what a generally valid string looks like.

apis/v1alpha2/shared_types.go
hbagdi: I added a note for this in one of my favorite issues: https://github.com/kubernetes-sigs/gateway-api/issues/629

apis/v1alpha2/shared_types.go
robscott: Added examples for most of these + references to where they originated from. Due to how the regexes are combined together in upstream source it can be difficult to find the true source of the logic, but it's fairly easy to get the regex used by just applying a resource with an invalid value. For example, here's how I figured out the validation for Group:

```
The CustomResourceDefinition "gateways.gateway.networking.k8s.io" is invalid:
* spec.group: Invalid value: "gateway.networ<king.k8s.io": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

apis/v1alpha2/shared_types.go
robscott: Good call, added examples throughout.

robscott:
Thanks for the great feedback on this @bowei, @hbagdi, @howardjohn, and @youngnick! I think I've addressed everything, PTAL.

apis/v1alpha2/httproute_types.go
```diff
@@ -262,6 +262,7 @@ type HTTPPathMatch struct {
 	//
 	// +optional
 	// +kubebuilder:default="/"
+	// +kubebuilder:validation:MaxLength=1024
```
bowei, line 265: apparently 2000 is the accepted folk limit around the web

apis/v1alpha2/httproute_types.go
```diff
@@ -282,6 +283,23 @@ const (
 	HeaderMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
 )
 
+// HTTPHeaderName is the name of an HTTP header.
+//
+// Valid values include:
+//
+// * Authorization
```
bowei, line 290: put "quotes" to make it clear these are strings

apis/v1alpha2/httproute_types.go
```diff
@@ -282,6 +283,23 @@ const (
 	HeaderMatchImplementationSpecific HeaderMatchType = "ImplementationSpecific"
 )
 
+// HTTPHeaderName is the name of an HTTP header.
```
bowei, line 286: Do we need to say it's case-insensitive?

apis/v1alpha2/httproute_types.go
```diff
@@ -282,6 +283,23 @@ const (
...
+//
+// Valid values include:
+//
+// * Authorization
+// * Set-Cookie
+//
+// Invalid values include:
+//
+// * :method
```
bowei, line 295: quotes, say why it's invalid (":", "/" are not a valid character for header names)

apis/v1alpha2/httproute_types.go
```diff
@@ -423,14 +438,20 @@ type HTTPRouteMatch struct {
...
 	Headers []HTTPHeaderMatch `json:"headers,omitempty"`
 
 	// QueryParams specifies HTTP query parameter matchers. Multiple match
 	// values are ANDed together, meaning, a request must match all the
 	// specified query parameters to select the route.
 	//
+	// +listType=map
+	// +listMapKey=name
 	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
bowei, line 454: Is there any harm to increase this to 30-50? I can see hitting 16 limit, probably not 30

apis/v1alpha2/httproute_types.go
```diff
@@ -659,29 +683,32 @@ type HTTPRequestRedirect struct {
...
+	Port *PortNumber `json:"port,omitempty"`
+
 	// StatusCode is the HTTP status code to be used in response.
 	//
 	// Support: Core
 	//
 	// +optional
 	// +kubebuilder:default=302
-	// +kubebuilder:validation=301;302
+	// +kubebuilder:validation:Enum=301;302
```
bowei, line 711: I feel like we should just include all the valid redirects from here:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages

apis/v1alpha2/object_reference_types.go
```diff
@@ -20,15 +20,10 @@ package v1alpha2
 // referrer.
 type LocalObjectReference struct {
 	// Group is the group of the referent.
```
bowei, line 22: For example: "core", "networking"

apis/v1alpha2/object_reference_types.go
```diff
@@ -20,15 +20,10 @@ package v1alpha2
...
-	// +kubebuilder:validation:MaxLength=253
-	Group string `json:"group"`
+	Group Group `json:"group"`
 
 	// Kind is kind of the referent.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	Kind string `json:"kind"`
+	Kind Kind `json:"kind"`
```
bowei, line 26: For example: Service, HTTPRoute

apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,90 @@ type RouteStatus struct {
...
+
+// Group refers to a Kubernetes Group. It must either be an empty string or a
+// RFC 1123 subdomain.
+//
+// This validation is based off of the corresponding Kubernetes validation:
+// https://git.k8s.io/apimachinery/pkg/util/validation/validation.go#L208
+//
+// Valid values include:
+//
+// * ""
```
bowei, line 252: (empty string)

bowei:
A few minor comments 

LGTM, ping me after changes, I will apply real LGTM

apis/v1alpha2/httproute_types.go
hbagdi: opinion:
Can you imagine anyone using a 2000 character path to do matching though?
If anything, I think we are already generous here.

apis/v1alpha2/httproute_types.go
robscott: Agree that we should add this, but didn't want to expand the scope of this PR so added https://github.com/kubernetes-sigs/gateway-api/issues/793 to track this.

apis/v1alpha2/httproute_types.go
hbagdi: We say in the place where this type is used. I'm not sure what is the right place for it. The place of use seems more correct because that gets added to the CRD. I can argue either way though.

apis/v1alpha2/httproute_types.go
hbagdi: We should explicitly call out that psuedo headers are not allowed.

apis/v1alpha2/httproute_types.go
hbagdi: There was a conformance concern for not adding them in the first place.

config/crd/v1alpha2/gateway.networking.k8s.io_httproutes.yaml
```diff
@@ -1219,6 +1289,7 @@ spec:
                             detached from the Gateway. \n Support: Core"
                           maxLength: 253
                           minLength: 1
+                          pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
                           type: string
```
hbagdi, line 1293: What is the largest CRD yet? We should register in that competition if one exists.

apis/v1alpha2/shared_types.go
hbagdi: Also mention that it implies core API group. Not a widely known thing AFAICT.

apis/v1alpha2/shared_types.go
```diff
@@ -249,4 +238,90 @@ type RouteStatus struct {
...
+// +kubebuilder:validation:Pattern=`^[a-zA-Z]([-a-zA-Z0-9]*[a-zA-Z0-9])?$`
+type Kind string
+
+// Namespace refers to a Kubernetes namespace. It must be a RFC 1123 label.
+//
+// This validation is based off of the corresponding Kubernetes validation:
+// https://git.k8s.io/apimachinery/pkg/util/validation/validation.go#L187
+//
+// This is used for Namespace name validation here:
+// https://git.k8s.io/apimachinery/pkg/api/validation/generic.go#L63
```
hbagdi, line 277: These line references will get out of date. Opt for only the file or even better, target a specific commit like so:
https://github.com/kubernetes/apimachinery/blob/02cfb53916346d085a6c6c7c66f882e3c6b0eca6/pkg/api/validation/generic.go#L63

Open the above link and press 'y'.

apis/v1alpha2/httproute_types.go
robscott: I hadn't really thought of a use case with that many matches, but I can increase to 32.

robscott: Actually we're using 16 as a limit in a lot of places, including header matching. Unless there's a clear use case I'd be tempted to leave that as is to minimize the changes here.

apis/v1alpha2/httproute_types.go
robscott: I think it's more appropriate to specify that it's case insensitive on the match field itself (current state). There's nothing on this type definition that can require that matching is case insensitive.

apis/v1alpha2/shared_types.go
robscott: Updated, thanks! Thanks for letting me know about that "y" shortcut, super helpful.

apis/v1alpha2/shared_types.go

apis/v1alpha2/object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha2/object_reference_types.go
robscott: Updated, thanks!

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go

robscott:
Thanks for the last round of feedback @bowei and @hbagdi! I think I've responded to everything, PTAL.

apis/v1alpha2/httproute_types.go
bowei: Maybe? There is data that 2000 is effectively the max which is well known. For the others, yes, we have less info so just have to pick a number. I'm not super tied to increase this -- just that we know that more about what the max is.

apis/v1alpha2/httproute_types.go
bowei: I'm ok with anything as long as the generated documentation is clear.

hbagdi:
Great job Rob!
/hold cancel

PR #762, opened 2021-08-10
Documenting Route attachment for v1alpha2

robscott:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
This is a follow up to #754 and provides documentation for GEP #724.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

Deploy Previews:

* https://deploy-preview-762--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/concepts/api-overview/
* https://deploy-preview-762--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/concepts/security-model/

PR #792, opened 2021-08-19
Index and Overview update

youngnick:
**What type of PR is this?**

/kind documentation

This updates the Index and Overview docs for v1alpha2. At this point, we have to start having things point to v1alpha2, or it's going to be easy to miss things as we release.

I've focussed on the Index and Overview pages to help the API reviewers out, I think this explanation is much clearer, and mirrors some discussions we've had on the Slack channel.

```release-note
NONE
```


site-src/concepts/api-overview.md
```diff
@@ -17,9 +17,10 @@ section in the Security model for details.
 
 ## Resource model
 
-> Note: Resources will initially live in the `networking.x-k8s.io` API group as
-> Custom Resource Definitions (CRDs). Unqualified resource names will implicitly
-> be in this API group.
+!!! note
+    As of v1alpha2, resources are in the `gateway.networking.k8s.io` API group as
+    Custom Resource Definitions (CRDs). Unqualified resource names below will implicitly
+    be in this API group. Prior to v1alpha1, the apigroup was `networking.x-k8s.io`.
```
jpeach, line 23:
```suggestion
    be in this API group. Prior to v1alpha1, the API group was `networking.x-k8s.io`.
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,42 @@ Status object.
...
 
-### {HTTP,TCP,Foo}Route
+### Route objects
 
 Route objects define protocol-specific rules for mapping requests from a Gateway
 to Kubernetes Services.
 
-`HTTPRoute` and `TCPRoute` are currently the only defined Route objects.
-Additional protocol-specific Route objects may be added in the future.
+As of v1alpha2, there are four Route types defined in this repo, although it's
```
jpeach, line 84:
```suggestion
As of v1alpha2, the API defines four Route types, although it's
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,42 @@ Status object.
...
+expected that in the future implementations may create their own custom Route
+types.
 
-### BackendPolicy
+#### HTTPRoute
+HTTPRoute is for multiplexing HTTP or terminated HTTPS connections onto a single port or set of ports. It's intended for use in cases where you want to inspect the HTTP stream and use HTTP-level data for either routing or modification. (An example here is using HTTP Headers for routing, or modifying them in-flight).
```
jpeach, line 89:
```suggestion
HTTPRoute is for multiplexing HTTP or terminated HTTPS connections. It's intended for use in cases where you want to inspect the HTTP stream and use HTTP-level data for either routing or modification, for example using HTTP Headers for routing, or modifying them in-flight.
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,42 @@ Status object.
...
 
-BackendPolicy provides a way to configure connections between a Gateway and a
-backend. For the purpose of this API, a backend is any resource that a route can
-forward traffic to. A common example of a backend is a Service. Configuration at
-this level is currently limited to TLS, but will expand in the future to support
-more advanced policies such as health checking.
+#### TLSRoute
+TLSRoute is for multiplexing TLS connections, discriminated via SNI, onto a single port or set or ports. It's intended for where you want to use the SNI as the main routing method, and are not interested in properties of the underlying connection (such as whether or not it is HTTP).
```
jpeach, line 92:
```suggestion
TLSRoute is for multiplexing TLS connections, discriminated via SNI. It's intended for where you want to use the SNI as the main routing method, and are not interested in properties of the underlying connection (such as whether or not it is HTTP)
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,42 @@ Status object.
...
+TCPRoute (and UDPRoute) are intended for use as a mapping between a single port or set of ports and a single backend. In this case, there is no discriminator you can use to choose different backends on the same port, so each TCPRoute really needs a different port on the listener (in general, anyway).
+
+#### Route summary table
+The "Routing Discriminator" column below refers to what information can be used to allow multiple Routes to share ports on the Listener.
+
+|Object|OSI Layer|Routing Discriminator|TLS Support|Purpose|
+|------|---------|---------------------|-----------|-------|
+|HTTPRoute| Layer 7 | Anything in the HTTP Protocol | Terminated only, can be reencrypted| HTTP and HTTPS Routing|
+|TLSRoute| Somewhere between layer 4 and 7| SNI or other TLS properties| Passthrough or terminated, can be reencrypted if terminated. | Routing of TLS protocols including HTTPS where inspection of the HTTP stream is not required.|
+|TCPRoute| Layer 4| None | None (but passthrough will work because the connection is forwarded) | Allows for forwarding of a TCP stream from the Listener to the Backends |
```
jpeach, line 104: TCPRoute can be TLS terminated or passed through, but not multiplexed using SNI. Is that distinction too complicated for this table?

site-src/concepts/api-overview.md
```diff
@@ -133,91 +141,89 @@ different relationships that Gateways and Routes can have:
   a single Route to control application exposure simultaneously across different
   IPs, load balancers, or networks.
 
-*In summary, Gateways select Routes and Routes control their exposure. When a
-Gateway selects a Route that allows itself to be exposed, then the Route will
-bind to the Gateway. When Routes are bound to a Gateway it means their
-collective routing rules are configured on the underlying load balancers or
-proxies that are managed by that Gateway. Thus, a Gateway is a logical
+*In summary, Routes attach to Gateways and Gateways choose what attachments to
```
jpeach, line 144:
```suggestion
*In summary, Routes attach to Gateways and Gateways choose which attachments to
```

Sorry, an extremely marginal nit :)

site-src/concepts/api-overview.md
```diff
@@ -133,91 +141,89 @@ different relationships that Gateways and Routes can have:
...
+  name: foo-gateway
+  namespace: foo-namespace
+spec:
+  listeners:
+  - name: prod-web
+    routes:
+      kinds:
+      - HTTPRoute
+      namespaces:
+      - from: bar-namespace
```
jpeach, line 206: From has to be one of "All;Selector;Same"

site-src/concepts/api-overview.md
youngnick: I don't think so, I'll fix that up. That's why I put the "Routing discriminator" column in.

site-src/concepts/api-overview.md
youngnick: Right, thanks, I'll fix that up.

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
+kind: Gateway
+metadata:
+  name: foo-gateway
+  namespace: foo-namespace
+spec:
+  listeners:
+  - name: prod-web
+    routes:
+      kinds:
+      - HTTPRoute
```
hbagdi, line 213: This line is not valid. It should be `kind: HTTPRoute`.
Can we please instead put this into the examples directory and reference it here so that we can ensure that such code segements are valid?

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
+
+```yaml
+kind: Gateway
+...
+spec:
+  listeners:
+  - name: prod-web
+    routes:
+      kinds:
+      - HTTPRoute
```
hbagdi, line 234: Same comment as above.

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+### Route objects
 
 Route objects define protocol-specific rules for mapping requests from a Gateway
 to Kubernetes Services.
 
-`HTTPRoute` and `TCPRoute` are currently the only defined Route objects.
-Additional protocol-specific Route objects may be added in the future.
+As of v1alpha2, the API defines four Route types, although it's
+expected that in the future implementations may create their own custom Route
+types.
```
hbagdi, line 86:
```suggestion
As of v1alpha2, four Route types are included with the API. Custom Route types
that are implementation-specific are encouraged for other protocols. New route
types may be added to the API in future.
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
 
-### BackendPolicy
+#### HTTPRoute
+HTTPRoute is for multiplexing HTTP or terminated HTTPS connections. It's intended
```
hbagdi, line 89:
```suggestion
HTTPRoute is for multiplexing HTTP or TLS-terminated HTTPS connections. It's intended
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+for use in cases where you want to inspect the HTTP stream and use HTTP-level data
```
hbagdi, line 90:
```suggestion
for use in cases where you want to inspect the HTTP stream and use HTTP request data
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
 
-BackendPolicy provides a way to configure connections between a Gateway and a
-backend. For the purpose of this API, a backend is any resource that a route can
-forward traffic to. A common example of a backend is a Service. Configuration at
-this level is currently limited to TLS, but will expand in the future to support
-more advanced policies such as health checking.
+#### TLSRoute
+TLSRoute is for multiplexing TLS connections, discriminated via SNI. It's intended
+for where you want to use the SNI as the main routing method, and are not interested
+in properties of the underlying connection (such as whether or not it is HTTP).
```
hbagdi, line 97:
```suggestion
in properties of the higher-level protocols like HTTP.  The byte stream of the connection is proxied
without any inspection to the backend.
```

To me, underlying connections seemed like TCP. I think I'm a bit too colored by OSI.

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+
+#### TCPRoute and UDPRoute
+TCPRoute (and UDPRoute) are intended for use as a mapping between a single port or
+set of ports and a single backend. In this case, there is no discriminator you can
```
hbagdi, line 101:
```suggestion
TCPRoute (and UDPRoute) are intended for use for mapping one or more ports
to a single backend. In this case, there is no discriminator you can
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+use to choose different backends on the same port, so each TCPRoute really needs a
+different port on the listener (in general, anyway). You can still terminate TLS,
+and passthrough will work since the stream is passed through unchanged.
```
hbagdi, line 104:
```suggestion
different port on the listener (in general, anyway). You can terminate TLS in
which case the unencrypted byte stream is passed through to the backend.
You can chose to not terminate TLS, in which case the encrypted byte stream
is passed through to the backend. 
```
More explicit here because I expect major confusion here.

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+and passthrough will work since the stream is passed through unchanged. 
+
+#### Route summary table
+The "Routing Discriminator" column below refers to what information can be used to allow multiple Routes to share ports on the Listener.
+
+|Object|OSI Layer|Routing Discriminator|TLS Support|Purpose|
+|------|---------|---------------------|-----------|-------|
+|HTTPRoute| Layer 7 | Anything in the HTTP Protocol | Terminated only, can be reencrypted| HTTP and HTTPS Routing|
+|TLSRoute| Somewhere between layer 4 and 7| SNI or other TLS properties| Passthrough or terminated, can be reencrypted if terminated. | Routing of TLS protocols including HTTPS where inspection of the HTTP stream is not required.|
+|TCPRoute| Layer 4| None | Termination (passthrough also will work because the connection is forwarded) | Allows for forwarding of a TCP stream from the Listener to the Backends |
```
hbagdi, line 113:
```suggestion
|TCPRoute| Layer 4| destination port | Passthrough or terminated | Allows for forwarding of a TCP stream from the Listener to the Backends |
```

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+|UDPRoute| Layer 4| None | None | Allows for forwarding of a UDP stream from the Listener to the Backends. |
```
hbagdi, line 114:
```suggestion
|UDPRoute| Layer 4| destination port | None | Allows for forwarding of a UDP stream from the Listener to the Backends. |
```

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
-  - routes:
-      kind: HTTPRoute
-      selector:
-        matchLabels:
-          expose: prod-web-gw
-      namespaces:
-        from: All
-```
+A Route *must* select what Gateway it wants to attach to, based on the
+`parentRefs` field, which allows the selection of the Group, Kind, name,
```
hbagdi, line 164:
```suggestion
A Route *must* specify the Gateways that can attach to itself using the
`parentRefs` field, which allows the selection of the Group, Kind, name,
```

The current text read as if "specify the gateway and you are done" while in reality it is only an "allow list".

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
+this is intended to allow for later extension. Additionally, the `parentRefs`
+stanza is a list, so a Route may request to attach to more than one Gateway
+(or other parent object).
 
-#### Route Exposure
+Additionally, Gateways *may* specify what kind of Routes they support
+(defaults to Routes that match the Listener protocol if not specified), and
+where those Routes can be (defaults to same namespace). If a Route wants
+to attach to a Gateway in another namespace, that Gateway must *explicitly*
+allow Routes from its namespace for the binding to succeed.
```
hbagdi, line 175:
```suggestion
allow Routes from the Route's namespace for the binding to succeed.
```

So many references...

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
 spec:
-  gateways:
-    allow: FromList
-    gatewayRefs:
-    - name: foo-gateway
-      namespace: foo-namespace
+  parentRefs:
+  - kind: Gateway
+    name: foo-gateway
+    namespace: foo-namespace
```
hbagdi, line 197: Can we move this to examples and reference it instead?

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
 A Gateway MAY contain one or more *Route references which serve to direct
 traffic for a subset of traffic to a specific service.*
 
-### {HTTP,TCP,Foo}Route
+### Route objects
```
robscott, line 79: Maybe "Types" or "Resources" would be better than "objects" here?

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
-bind to the Gateway. When Routes are bound to a Gateway it means their
-collective routing rules are configured on the underlying load balancers or
-proxies that are managed by that Gateway. Thus, a Gateway is a logical
+*In summary, Routes attach to Gateways and Gateways choose which attachments to
+allow. When a Route tries to attach to a Gateway that does not prevent it, then
+the Route will bind to the Gateway. When Routes are bound to a Gateway it means
+their collective routing rules are configured on the underlying load balancers
+or proxies that are managed by that Gateway. Thus, a Gateway is a logical
 representation of a networking data plane that can be configured through
 Routes.*
```
robscott, line 159: I'm not sure what happened here, but I'd actually already updated a good chunk of the Route->Gateway attachment section for v1alpha2 when this page was versioned (https://github.com/kubernetes-sigs/gateway-api/pull/762/files#diff-bb2765a3020ed8962e5e749402a8ace4539c6218bdc621a98625ca19b2700e82R89). I think a lot of what you've done here is better than my initial work, but it may be helpful to merge some of the concepts together here.

I remember in that PR, every way I tried to reword this still sounded confusing to me. I think it may be better to remove or simplify this paragraph in favor of further explanation below.

site-src/concepts/api-overview.md
```diff
@@ -75,35 +76,51 @@ Status object.
...
+|TCPRoute| Layer 4| None | Termination (passthrough also will work because the connection is forwarded) | Allows for forwarding of a TCP stream from the Listener to the Backends |
+|UDPRoute| Layer 4| None | None | Allows for forwarding of a UDP stream from the Listener to the Backends. |
 
-Some backend configuration may vary depending on the Route that is targeting the
-backend. In those cases, configuration fields will be placed on Routes and not
-BackendPolicy. For more information on what may be configured with this resource
-in the future, refer to the related [GitHub
-issue](https://github.com/kubernetes-sigs/gateway-api/issues/196).
 
 ### Route binding
```
robscott, line 117: Maybe just "Attaching Routes to Gateways"

site-src/concepts/api-overview.md
```diff
@@ -133,91 +150,94 @@ different relationships that Gateways and Routes can have:
...
-    - `All` will select Routes from all Namespaces.
-    - `Selector` means that Routes from a subset of Namespaces selected by a
-      Namespace label selector will be selected by this Gateway. When `Selector`
-      is used then the `listeners.routes.namespaces.selector` field can be used
-      to specify label selectors. This field is not supported with `All` or
-      `SameNamespace`.
-
-The below Gateway will select all HTTPRoute resources with the `expose:
-prod-web-gw` across all Namespaces in the cluster.
+#### Route binding handshake
```
robscott, line 161: Maybe this can be merged with this: https://github.com/kubernetes-sigs/gateway-api/pull/762/files#diff-bb2765a3020ed8962e5e749402a8ace4539c6218bdc621a98625ca19b2700e82R126-R150.

robscott (commented):
Thanks for the work on this @youngnick! These are some great improvements. I'm not 100% sure how I feel about all of these pages moving to v1alpha2 concepts while being unversioned. Maybe we need to move version-specific content somewhere else? If not, maybe we should wait until v1alpha2 is released until we merge this in.

site-src/concepts/api-overview.md
hbagdi: What is the k8s standard here? I've seen all three being used in this repo.

youngnick:
>  I'm not 100% sure how I feel about all of these pages moving to v1alpha2 concepts while being unversioned. Maybe we need to move version-specific content somewhere else? If not, maybe we should wait until v1alpha2 is released until we merge this in.

Yeah, I thought the same as I was writing it, but v1alpha2 is such a big change that it needs to be covered in the concepts docs, and we're not going to change the concepts docs after this (hopefully), so I guessed that having them unversioned was better.

With versioned concepts docs though, updating the index page is *very* difficult, since it needs to be relevant to all versions.

PR #780, opened 2021-08-13
SIG-NETWORK v1alpha2 Review

robscott:
**What type of PR is this?**
/kind api-change

**What this PR does / why we need it**:
This is to facilitate the sig-network review of v1alpha2 (#716). It represents a review of the entirety of the `apis/v1alpha2` directory from master at this point in time. Updates to the master branch will require manual updates to this PR as well. I've tried alternatives like .gitignore and .gitattributes that would not require rebasing (https://github.com/robscott/gateway-api/pull/1), but none have worked well for me. 

**References:** 
* [Intro presentation](https://docs.google.com/presentation/d/1Bey3eMFd9dsA8iU_r1wOC5VVXa93oF2CGCl2EakQHWs/edit?usp=sharing&resourcekey=0-7x4CEVRr2ptpaAurz82muw)
* [Docs](gateway-api.sigs.k8s.io) (still WIP for v1alpha2)
* [k/enhancements issue](https://github.com/kubernetes/enhancements/issues/2829)
* [k/enhancements PR](https://github.com/kubernetes/enhancements/pull/2830)

/hold
/cc @andrewsykim @danwinship @khenidak @thockin

PR #796, opened 2021-08-20
Variety of fixes in response to feedback on #780

robscott:
**What type of PR is this?**
/kind cleanup

There's been a lot of great feedback in #780, this makes some of the easier fixes.

```release-note
* listener.routes has been renamed to listener.allowedRoutes
* The "NoSuchGatewayClass" has been removed after it was deprecated in v1alpha1
* "*" is no longer a valid hostname. Instead, leaving hostname unspecified is interpreted as "*".
```


apis/v1alpha2/gateway_types.go
```diff
@@ -30,13 +30,8 @@ import (
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
-// Gateway represents an instantiation of a service-traffic handling
-// infrastructure by binding Listeners to a set of IP addresses.
-//
-// Implementations should add the `gateway-exists-finalizer.gateway.networking.k8s.io`
-// finalizer on the associated GatewayClass whenever Gateway(s) is running.
-// This ensures that a GatewayClass associated with a Gateway(s) is not
-// deleted while in use.
```
robscott, line 39: Move to GatewayClass spec

apis/v1alpha2/gateway_types.go
```diff
@@ -118,18 +113,17 @@ type GatewaySpec struct {
...
+	// depend on the implementation. If a value is set in the spec and the
+	// requested address is invalid or unavailable, the implementation MUST
+	// indicate this in the associated entry in GatewayStatus.Addresses.
 	//
-	// If no Addresses are specified, the GatewayClass may
-	// schedule the Gateway in an implementation-defined manner,
-	// assigning an appropriate set of Addresses.
+	// If no Addresses are specified, the implementation MAY schedule the
+	// Gateway in an implementation-specific manner, assigning an appropriate
+	// set of Addresses.
```
youngnick, line 123: Do we need to put something in here like "All assigned addresses MUST be added in GatewayStatus.Addresses`?

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -47,12 +60,19 @@ type GatewayClass struct {
 	Status GatewayClassStatus `json:"status,omitempty"`
 }
 
+const (
+	// GatewayClassFinalizerGatewaysExist should be added as a finalizer to the
```
youngnick, line 64:
```suggestion
	// GatewayClassFinalizerGatewaysExist MUST be added as a finalizer to the
```

youngnick (commented):
Two small suggestions, but aside from that, LGTM.

apis/v1alpha2/httproute_types.go
```diff
@@ -62,44 +62,29 @@ type HTTPRouteSpec struct {
...
+	//   one hostname that matches the Listener hostname. For example,
+	//   `test.example.com` and `*.example.com` would both match. On the other
+	//   hand, `a.b.example.com`, `example.com`, and `test.example.net` would
+	//   not match.
+	//
+	// If both the Listener and HTTPRoute have specified hostnames, any
+	// HTTPRoute hostnames that do not match the Listener hostname MUST be
+	// ignored. For example, if a Listener specified `*.example.com`, and the
+	// HTTPRoute specified `test.example.com` and `test.example.net`,
+	// `test.example.net` would be ignored.
```
hbagdi, line 83:
```suggestion
	// `test.example.net` must not be considered for a match.
```
Same everywhere else would be best.
The reason I don't like 'ignored' is that implementation may surface the 'ignored' ones in status, which in some ways means that those rules are not ignored.

Ignored also has a negative connotation w.r.t API definitions.

apis/v1alpha2/shared_types.go
```diff
@@ -166,9 +166,13 @@ type BackendRef struct {
 type RouteConditionType string
 
 const (
-	// This condition indicates whether the route has been admitted
-	// or rejected by a Gateway, and why.
+	// This condition indicates whether the route has been admitted or rejected
+	// by a Gateway, and why.
```
hbagdi, line 170: MUST be set only if all rules are satisfied.

apis/v1alpha2/tcproute_types.go
```diff
@@ -111,12 +109,10 @@ type TCPRouteRule struct {
...
-	// "match" behavior.  For example, resource "mytcproutematcher" in group
-	// "networking.acme.io". If the referent cannot be found, the rule is not
-	// included in the route. The controller should raise the "ResolvedRefs"
-	// condition on the Gateway with the "DegradedRoutes" reason. The gateway
-	// status for this route should be updated with a condition that describes
-	// the error more specifically.
+	// "match" behavior. For example, resource "mytcproutematcher" in group
+	// "networking.example.net". If the referent cannot be found, the rule is
+	// not included in the route. The controller must ensure the "ResolvedRefs"
+	// condition on the Route status is set to `status: False`.
```
hbagdi, line 115:
```suggestion
	// "networking.example.net". If the referent cannot be found, the rule MUST
	// not be included in the route. The controller must ensure the "ResolvedRefs"
	// condition on the Route status is set to `status: False`.
```

apis/v1alpha2/gateway_types.go
```diff
@@ -207,32 +196,33 @@ type Listener struct {
...
 	TLS *GatewayTLSConfig `json:"tls,omitempty"`
 
-	// Routes specifies which Routes may be attached to this Listener.
+	// AllowedRoutes specifies which Routes may be attached to this Listener.
 	//
-	// Although a client request may technically match multiple route rules,
-	// only one rule may ultimately receive the request. Matching precedence
-	// MUST be determined in order of the following criteria:
+	// Although a client request may match multiple route rules, only one rule
+	// may ultimately receive the request. Matching precedence MUST be
```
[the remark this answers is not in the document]
robscott: I think one of the things we're struggling with here is what the intended audience is. I think we may need to start differentiating implementation guidance (words like `MUST`) with what users should expect (words like `will`). Maybe implementation guidance belongs somewhere other than spec?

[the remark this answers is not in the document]
hbagdi: I've struggled with that distinction and I brought it before as well.
I think most of the documentation is geared towards implementation guidance and we should stick to that. The larger audience will require much more context and the implementation authors can take care of that.

apis/v1alpha2/gateway_types.go
```diff
@@ -30,13 +30,8 @@ import (
...
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
-// Gateway represents an instantiation of a service-traffic handling
-// infrastructure by binding Listeners to a set of IP addresses.
-//
-// Implementations should add the `gateway-exists-finalizer.gateway.networking.k8s.io`
-// finalizer on the associated GatewayClass whenever Gateway(s) is running.
-// This ensures that a GatewayClass associated with a Gateway(s) is not
-// deleted while in use.
+// Gateway represents an instance of a service-traffic handling infrastructure
```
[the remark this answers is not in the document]
robscott: I think we want to be broader than service-bound traffic. It's entirely possible that Gateways and Routes can be used to route to non-service backends.

apis/v1alpha2/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha2/shared_types.go
robscott: Do you mean if all route rules are valid?

apis/v1alpha2/httproute_types.go

apis/v1alpha2/gateway_types.go
```diff
@@ -770,23 +751,14 @@ const (
...
 
-	// This reason is used with the "ResolvedRefs" condition when
-	// the Listener's Routes selector or kind is invalid or cannot
-	// be resolved. Note that it is not an error for this selector to
-	// not resolve any Routes, and the "ResolvedRefs" status condition
-	// should not be raised in that case.
-	ListenerReasonInvalidRoutesRef ListenerConditionReason = "InvalidRoutesRef"
+	// This reason is used with the "ResolvedRefs" condition when an invalid or
+	// unsupported Route kind is specified by the Listener.
+	ListenerReasonInvalidRoutesRef ListenerConditionReason = "InvalidRouteKinds"
```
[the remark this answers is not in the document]
robscott: Yes, but I'm not sure what that behavior should be.

apis/v1alpha2/gateway_types.go
```diff
@@ -368,18 +357,21 @@ type ListenerRoutes struct {
...
-	// Only Routes in the same namespace as the Gateway may be used by this Gateway.
-	RouteSelectSame RouteSelectType = "Same"
+	// Routes in all namespaces may be attached to this Gateway.
+	NamespacesFromAll NamespacesFrom = "All"
+	// Only Routes in namespaces selected by the selector may be attached to
+	// this Gateway.
+	NamespacesFromSelector NamespacesFrom = "Selector"
+	// Only Routes in the same namespace as the Gateway may be attached to this
+	// Gateway.
+	NamespacesFromSame NamespacesFrom = "Same"
```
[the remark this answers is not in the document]
robscott: Need to think about this one a bit more, don't love both options either.

apis/v1alpha2/gateway_types.go
```diff
@@ -339,11 +329,10 @@ const (
 	TLSModePassthrough TLSModeType = "Passthrough"
 )
 
-// ListenerRoutes defines which Routes may be attached to this Listener.
-type ListenerRoutes struct {
-	// Namespaces indicates which namespaces Routes may be attached to this
-	// Listener from. This is restricted to the namespace of this Gateway by
-	// default.
+// AllowedRoutes defines which Routes may be attached to this Listener.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

apis/v1alpha2/gateway_types.go
```diff
@@ -207,32 +196,33 @@ type Listener struct {
...
 	//   example, foo/bar is given precedence over foo/baz.
 	//
-	// All valid portions of a Route selected by this field should be supported.
-	// Invalid portions of a Route can be ignored (sometimes that will mean the
-	// full Route). If a portion of a Route transitions from valid to invalid,
-	// support for that portion of the Route should be dropped to ensure
-	// consistency. For example, even if a filter specified by a Route is
-	// invalid, the rest of the Route should still be supported.
+	// All valid rules within a Route attached to this Listener should be
+	// supported. Invalid Route rules can be ignored (sometimes that will mean
```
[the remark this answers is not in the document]

apis/v1alpha2/gateway_types.go
```diff
@@ -207,32 +196,33 @@ type Listener struct {
...
-	// MUST be determined in order of the following criteria:
+	// Although a client request may match multiple route rules, only one rule
+	// may ultimately receive the request. Matching precedence MUST be
+	// determined in order of the following criteria:
 	//
-	// * The most specific match. For example, the most specific HTTPRoute match
-	//   is determined by the longest matching combination of hostname and path.
+	// * The most specific match as defined by the Route type. For example, the
+	//   most specific HTTPRoute match is determined by the longest matching
+	//   combination of hostname and path.
```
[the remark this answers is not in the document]
robscott: Ended up removing this from here as it will easily become outdated.

apis/v1alpha2/gatewayclass_types.go

apis/v1alpha2/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha2/gateway_types.go
```diff
@@ -192,9 +181,9 @@ type Listener struct {
 	// Support: Core
 	Protocol ProtocolType `json:"protocol"`
 
-	// TLS is the TLS configuration for the Listener. This field
-	// is required if the Protocol field is "HTTPS" or "TLS" and
-	// ignored otherwise.
+	// TLS is the TLS configuration for the Listener. This field is required if
+	// the Protocol field is "HTTPS" or "TLS". It MUST be ignored when the
+	// Protocol field is "HTTP", "TCP", or "UDP".
```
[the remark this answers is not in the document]
youngnick: Feels like this is a good case for a Condition of some sort, but I'm not sure which one.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
youngnick: Is it allowed to do `FromNamespaces`? That seems much clearer. Then the constants would be:
```
NamespacesFromAll
NamespacesFromSelector
NamespacesFromSame
```

apis/v1alpha2/shared_types.go
youngnick: And I think we may say in other places that a Route should only be not admitted if all routes are *not* valid. Regardless, this should tell a consistent story with the rest of the API, and should say it clearly.

youngnick:
I think I've only really got one thing that needs addressing here, but there are a few followup things, especially around Conditions and status (which is probably due for a recheck and overhaul). I don't think we should block at least the RC on status though, and I feel like a status review is possibly bigger than we should bite off bevore v1alpha2.

apis/v1alpha2/shared_types.go
hbagdi: > Do you mean if all route rules are valid?

Yes. Is that not what we already want?

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
ccfishk: > I think we want to be broader than service-bound traffic. It's entirely possible that Gateways and Routes can be used to route to non-service backends.

Agree with this point : could redirect to deployment backends also, which might not exposed as a service with  label/selector(s).

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -47,12 +60,19 @@ type GatewayClass struct {
 	Status GatewayClassStatus `json:"status,omitempty"`
 }
 
+const (
+	// GatewayClassFinalizerGatewaysExist MUST be added as a finalizer to the
```
ccfishk, line 64: literally "must" is already "commanded or requested",  strong.
dont think the uppercase is still needed, people might feel a little bit scaring. 
 ```suggestion
	// GatewayClassFinalizerGatewaysExist must be added as a finalizer to the
```

File: apis/v1alpha2/gatewayclass_types.go [hunk 16]
```diff
@@ -47,12 +60,19 @@ type GatewayClass struct {
...
+	GatewayClassFinalizerGatewaysExist = "gateway-exists-finalizer.gateway.networking.k8s.io"
+)
+
 // GatewayClassSpec reflects the configuration of a class of Gateways.
 type GatewayClassSpec struct {
 	// Controller is a domain/path string that indicates the
 	// controller that is managing Gateways of this class.
 	//
-	// Example: "acme.io/gateway-controller".
+	// Example: "example.net/gateway-controller".
```
ccfishk, line 75: 👍

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -47,12 +60,19 @@ type GatewayClass struct {
 	Status GatewayClassStatus `json:"status,omitempty"`
 }
 
+const (
+	// GatewayClassFinalizerGatewaysExist MUST be added as a finalizer to the
+	// GatewayClass whenever there are provisioned Gateways using a
+	// GatewayClass.
+	GatewayClassFinalizerGatewaysExist = "gateway-exists-finalizer.gateway.networking.k8s.io"
```
ccfishk, line 67: After reading your comments,  
```suggestion
	ProvisionedGatewayClassFinalizer = "provisioned-gateway-finalizer.gateway.networking.k8s.io"
```

is more aligned to what you mean ?

apis/v1alpha2/httproute_types.go
```diff
@@ -62,44 +62,29 @@ type HTTPRouteSpec struct {
...
-	//   must match exactly. So, `test.foo.com`, `*.foo.com` or `blog.foo.com` match.
-	//   `test.blog.foo.com`, `test.bar.com`, or `bar.com` do not. Hostnames that do
-	//   not match will be ignored.
+	// If a hostname is specified by both the Listener and HTTPRoute, their
+	// must be at least one intersecting hostname for the HTTPRoute to be
+	// attached to the Listener. For example:
+	//
+	// * A Listener with `test.example.com` as the hostname matches HTTPRoutes
+	//   that have either not specified any hostnames, or have specified at
+	//   least one of `test.example.com` or `*.example.com`.
```
ccfishk, line 71: or "*"

ccfishk, line 71:
```suggestion
	//   least one of `test.example.com` or `*.example.com` or `*`.
```

apis/v1alpha2/httproute_types.go
```diff
@@ -62,44 +62,29 @@ type HTTPRouteSpec struct {
...
-	// - A HTTPRoute with unspecified, empty, or `*` values for Hostname matches
-	//   any Listener hostname.
-	// - A Listener with `test.foo.com` as the hostname matches *only*
-	//   `test.foo.com` or `*.foo.com`. Any other hostnames present must be ignored.
-	// - A Listener with `*.foo.com` as hostname, all hostnames in the HTTPRoute
-	//   must have any single label where the star is, and the rest of the hostname
-	//   must match exactly. So, `test.foo.com`, `*.foo.com` or `blog.foo.com` match.
-	//   `test.blog.foo.com`, `test.bar.com`, or `bar.com` do not. Hostnames that do
-	//   not match will be ignored.
+	// If a hostname is specified by both the Listener and HTTPRoute, their
```
ccfishk, line 65:
```suggestion
	// If a hostname is specified by both the Listener and HTTPRoute, there 
```

apis/v1alpha2/httproute_types.go
```diff
@@ -62,44 +62,29 @@ type HTTPRouteSpec struct {
...
+	//   `test.example.com` and `*.example.com` would both match. On the other
+	//   hand, `a.b.example.com`, `example.com`, and `test.example.net` would
+	//   not match.
+	//
+	// If both the Listener and HTTPRoute have specified hostnames, any
+	// HTTPRoute hostnames that do not match the Listener hostname MUST be
+	// ignored. For example, if a Listener specified `*.example.com`, and the
+	// HTTPRoute specified `test.example.com` and `test.example.net`,
+	// `test.example.net` must not be considered for a match.
+	//
```
ccfishk, line 84: Do you mean as following ? 
// If both the Listener and HTTPRoute have specified hostnames, only exact match 
// will be considered. 
// For example, if a Listener specified `test.example.com`, and the
// HTTPRoute specified `test.example.com` and `test.example.net`,
// `test.example.net` is not be considered for a match.

apis/v1alpha2/httproute_types.go
```diff
@@ -199,11 +186,14 @@ type HTTPRouteRule struct {
...
 
 	// BackendRefs defines the backend(s) where matching requests should be
-	// sent. If unspecified or invalid (refers to a non-existent resource or a Service with no endpoints),
-	// the rule performs no forwarding; if no filters are specified that would result in a
-	// response being sent, a HTTP 503 status code is returned. 503 responses must be sent so that the overall
-	// weight is respected; if an invalid backend is requested to have 80% of requests, then 80% of requests
-	// must get a 503 instead.
+	// sent.
+
+	// If unspecified or invalid (refers to a non-existent resource or a Service
```
ccfishk, line 191: is there any k8s service (Service with no endpoints) that does not have endpoints ? 
maybe "resource" is more accurate here ?

apis/v1alpha2/httproute_types.go
```diff
@@ -537,7 +526,7 @@ type HTTPRouteFilter struct {
 
 	// ExtensionRef is an optional, implementation-specific extension to the
 	// "filter" behavior.  For example, resource "myroutefilter" in group
-	// "networking.acme.io"). ExtensionRef MUST NOT be used for core and
+	// "networking.example.net"). ExtensionRef MUST NOT be used for core and
```
ccfishk, line 529: 👍

apis/v1alpha2/tcproute_types.go
```diff
@@ -111,12 +109,10 @@ type TCPRouteRule struct {
...
 type TCPRouteMatch struct {
 	// ExtensionRef is an optional, implementation-specific extension to the
-	// "match" behavior.  For example, resource "mytcproutematcher" in group
-	// "networking.acme.io". If the referent cannot be found, the rule is not
-	// included in the route. The controller should raise the "ResolvedRefs"
-	// condition on the Gateway with the "DegradedRoutes" reason. The gateway
-	// status for this route should be updated with a condition that describes
-	// the error more specifically.
+	// "match" behavior. For example, resource "mytcproutematcher" in group
+	// "networking.example.net". If the referent cannot be found, the rule MUST
```
ccfishk, line 113:
```suggestion
	// "networking.example.net". If the referent cannot be found, the rule must
```

site-src/v1alpha2/references/policy-attachment.md
```diff
@@ -199,15 +199,15 @@ policies to requests to external services. To accomplish this, implementations
 can choose to support a refernce to a virtual resource type:
 
 ```yaml
-apiVersion: networking.acme.io/v1alpha1
+apiVersion: networking.example.net/v1alpha1
```
ccfishk, line 202:
```suggestion
apiVersion: networking.example.net/v1alpha2
```

apis/v1alpha2/httproute_types.go
robscott: This is actually something that came out of API review, I can't remember which of the reviewers suggested it, but the recommendation was to remove "*" as a possible value. An empty or unspecified value here would continue to be interpreted at "*", but there was no point in differentiating between empty and "*" since they were interpreted the same way. Let me add that to the PR description.

apis/v1alpha2/gatewayclass_types.go
robscott: This probably could have a better name, but in this case I'm not actually changing the name, just moving it from gateway_types to gatewayclass_types. The idea is for it to indicate that at least one Gateway is using the GatewayClass and therefore the GatewayClass should not be deleted.

apis/v1alpha2/gatewayclass_types.go
robscott: I'm not really sure how to distinguish between `MUST` and `must` for our godocs, but I think historically we've been using all caps to emphasize/match RFC terminology.

apis/v1alpha2/shared_types.go
robscott: So to clarify, what we want to say here is that admitted means that at least one Route rule has been admitted, not necessarily that the entirety of the Route has been admitted?

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: Thinking about this more alongside the "prefer not to break things" guidance, it seems like we need to both support the requested route kinds we know how to + add this condition to indicate that we don't support these other specific kinds.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: Was talking with @bowei about this recently, maybe there are some guidelines we can be applying when we're writing our godocs here?

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -47,12 +60,19 @@ type GatewayClass struct {
 	Status GatewayClassStatus `json:"status,omitempty"`
 }
 
+const (
+	// GatewayClassFinalizerGatewaysExist MUST be added as a finalizer to the
+	// GatewayClass whenever there are provisioned Gateways using a
```
howardjohn, line 65: is it really a MUST? Does IngressClass have this?

IMO every product that uses `finalizers` has a bad user experience. I am fairly strongly against using one in our product. There is nothing worse than things not deleting when you want them too...

apis/v1alpha2/httproute_types.go
robscott: I think what we're trying to communicate here is a Service that does not have any endpoints, likely because there are no Pods running that have been selected by that service.

apis/v1alpha2/httproute_types.go
robscott: This is trying to describe how wildcard hostname matching should work with hostnames specified on HTTPRoutes. I don't think removing the wildcard from the example would help here, but I may be misunderstanding what you're suggesting.

apis/v1alpha2/gatewayclass_types.go
robscott: I don't think we _need_ to enforce this behavior, there is no parallel for IngressClass. One thing that we need to be clear about though it should not be easy to delete all Gateways of a certain class simply by deleting the GatewayClass.

apis/v1alpha2/tcproute_types.go
robscott: I think there's probably room for a broader discussion around the usability of these all caps RFC terms, but so far the guidance from other reviewers has been to move more into this capitalization style. We can add to the agenda for next community meeting and/or create an issue to discuss if that's helpful. I don't have a very strong opinion here other than trying to be consistent with what we've done elsewhere.

apis/v1alpha2/gatewayclass_types.go
robscott: Moved back to "should" here to match previous state.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: I agree with that interpretation. For now I've just stated that this is invalid, but also created https://github.com/kubernetes-sigs/gateway-api/issues/827 to track this.

apis/v1alpha2/httproute_types.go

apis/v1alpha2/shared_types.go
robscott: Added some guidance around this as well.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: Added guidance for this approach.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: 👍 I like that idea, updated to match that pattern.

robscott:
Thanks to everyone for the great feedback on this one! I think I've responded to everything now, PTAL. Once we get this in, we should be able to move forward with v1alpha2-rc1.

robscott:
Looks like a deploy/netlify flake?

/retest

apis/v1alpha2/tcproute_types.go
youngnick: This is definitely a thing we've copied from RFCs, where the increased emphasis is very purposeful - it's a guide for people implementing the spec that allows the reader to easily find the things they MUST do when reading what is often a long and dry specification. That's why I'm supportive of the all-caps.

youngnick (commented):
/lgtm

I think that @ccfishk raises some fair questions to discuss, particularly about the use of all-caps for MUST, MAY etc, but I also think that's a larger discussion that would be better suited for its own issue/PR.

apis/v1alpha2/gatewayclass_types.go
jpeach: https://www.rfc-editor.org/rfc/rfc2119

PR #819, opened 2021-08-24
Make tls.certificateRef default to Secret (was Service)

howardjohn:
This seems like a mistake cause by reusing a type. Not sure if there is
a more idiomatic way to represent the Go portion, but this makes the
YAML work as expected.


robscott:
Thanks for catching this @howardjohn! Can you also rename `ObjectReference` to `ServiceObjectReference` (or something similar) so we don't make this mistake again?
 
/ok-to-test

howardjohn:
> Thanks for catching this @howardjohn! Can you also rename `ObjectReference` to `ServiceObjectReference` (or something similar) so we don't make this mistake again?
> 
> /ok-to-test

Looks like I can do one better - delete it. This was the only usage.

robscott:
Thanks!

/lgtm
/approve

PR #829, opened 2021-08-25
implement GEP-820: remove extension points from route match types

### LINKED ISSUE #820: GEP: Drop extension points from Route matches

This GEP will add a formal prosposal to drop match extension points from all route types within the API. 

**Why this is needed**:

Although these extension points have been present from early days of the API, they are not understood clearly - we do not have concrete use-cases for them and we do not have any feedback on them. We should drop these and introduce them back into the API when we have more clarity.

/kind gep

hbagdi:
**What type of PR is this?**

/kind api-change
Implements GEP #820 Drop extension points from Route matches

Fixes #820

```release-note
Extension points within match blocks from all Routes have been removed
```

GEP isn't merged (#822).
/hold


youngnick:
/lgtm

I think that this is a good change, and that removing the match functionality makes TCPRoute and UDPRoute easier to explain, at least initially.

robscott:
Thanks!

/lgtm

robscott:
Based on our discussion at the community meeting yesterday, I think we have sufficiently broad consensus to remove the hold on this.

/hold cancel

PR #837, opened 2021-08-27
Make MirrorFilter.BackendRef required

howardjohn:
/kind cleanup

/kind api-change

    Currently this is optional. I cannot see any reason why it would not be
    required, as the whole point of mirroring is the request goes to some
    backend?
Does this PR introduce a user-facing change?:

```notes
* Make `MirrorFilter.BackendRef` a required field when the mirror filter is used
```


apis/v1alpha2/httproute_types.go
```diff
@@ -722,9 +722,7 @@ type HTTPRequestMirrorFilter struct {
 	//
 	// Support: Extended for Kubernetes Service
 	// Support: Custom for any other resource
-	//
-	// +optional
-	BackendRef *BackendObjectReference `json:"backendRef,omitempty"`
+	BackendRef BackendObjectReference `json:"backendRef"`
```
hbagdi, line 725: CI failing because tests are not compiling.

hbagdi:
LGTM from an API perspective.

apis/v1alpha2/httproute_types.go
howardjohn: thanks, should be good now

hbagdi:
@howardjohn Can you update the release notes section of the PR description so that it makes sense to controller authors and end-users?

/lgtm

robscott:
Thanks!

/approve

Issue #790, opened 2021-08-18
v1alpha2 feedback


howardjohn:
I started working on implementing v1alpha2 and came across a few minor things that I think could use tweaks. Opening a single issue to capture them, and may use as a running log if I have any others.. we can split out larger ones if needed

* [x] #821
* [x] "If all hostnames do not match, then the HTTPRoute is not admitted". Should be "any", not "all", I think? In the same doc it says "Hostnames that do not match will be ignored." which implies its ok for only a subset to match.
* [ ] #838
* [x] ParentRefs - if the kind is unsupported, should we silently ignore it or report an error in status? It seems like maybe we shouldn't because it could be handled by another controller. What about invalid Scope/Namespace? That one we also probably can't tell we own it
* [x] Listener.Name and corresponding status - should be +optional and a pointer? to be consistent with the docs. This impacts the YAML - name is required but docs say its optional.
* [x] ListenerStatus.SupportedKinds - maybe nitpick, but "must represent the kinds an implementation supports for the specified protocol". Isn't it a property of protocol+tls settings?
* [x] #840
* [x] multi crds: v1alpha1 "wins" on `kubectl get gateway`. We should have the kustomize only deploy v1alpha2 to the cluster. PR https://github.com/kubernetes-sigs/gateway-api/pull/789/
* [x] listener.Hostname ambiguity for protocols. "This field can be omitted for protocols that don't require hostname based matching." - what does it mean if its there for TCPRoute? Is it valid? Does it do anything? Since TLSRoute now has `hostnames` like HTTPRoute, it should IMO also have the same logic described for HTTPRoute there as well.
* [x] ReferencePolicy.spec.from.group: `When empty, the "core" API group is inferred.`. But its required. so are users supposed to do `group: ""`? Is `group: "core"` legal? EDIT: read examples, looks like `group: ""` is the expectation. A bit odd but ok. I feel like default of core may be a good idea here - the most common usage will be Service and Secret probably (and routes I guess)

robscott:
Thanks @howardjohn! This is a great list, I've added this issue to the v1alpha2 milestone.

robscott:
> Listener.Name and corresponding status - should be +optional and a pointer? to be consistent with the docs. This impacts the YAML - name is required but docs say its optional.

[...]

Issue #72, opened 2020-02-11
Address SNI binding and bypass for TLS listeners


jpeach:
**What would you like to be added**:

In the current API, certificates are a property of the Listener and Routes are a property of the Gateway. Since there is no direct relationship between the certificates and the backend service (i.e. the Route), this structure encourages the separation of SNI routing from HTTP request routing. This can lead to a number of different security issues.

**Why is this needed**:

Operators need the ability to pin the set of certificates to be used for a specific backend service. That is, for a TLS session bound to SNI name `foo.example.com`, always forward HTTP requests to the backend session that is bound to the `foo.example.com` host name.

For a more thorough discussion of SNI bypass, see https://hal.inria.fr/hal-01202712/document 

xref #49

jpeach:
Also [RFC 6066](https://tools.ietf.org/html/rfc6066#section-11.1), which strongly suggests that the HTTP host should be bound to the SNI name.

>    Since it is possible for a client to present a different server_name
   in the application protocol, application server implementations that
   rely upon these names being the same MUST check to make sure the
   client did not present a different name in the application protocol.

bowei:
How do the major proxy implementations handle this? 
Is there possibly a use case to enable SNI/HTTP Host mismatch (I'm guessing no)
Is this dependent on how you set up your config (e.g. if we mandate such behavior, is it possible to achieve)

youngnick:
Contour *intends* to stop this from happening, but in practice it's very easy to do it accidentally, via implementation details.

There are some use cases to do this (basically, allowing [Domain fronting](https://en.wikipedia.org/wiki/Domain_fronting) and similar use cases), but I think that they are specialised enough that we should default to not supporting that and figure it out later.

The key thing here is that, if we include the possibility of client certificates as part of the TLS config, SNI/HTTP Host mismatch is a big security risk. If you allow the TLS negotation to succeed for host foo, including the client certificate exchange, and that exchange is being used for authentication, then allowing the inner HTTP transaction to go to a different host completely bypasses it.

The use case here is that I'm using a Gateway for `ultrasecure.foo.com`, which requires TLS with a client certificate, and the same Gateway for `insecure.foo.com`, which requires a serving TLS only. TLS is terminated at the Gateway, and may be reencrypted afterwards.

In this case, requests with an SNI of `insecure.foo.com` will complete the TLS with the proxy, and, if you have a Host header of `ultrasecure.foo.com`, you will totally bypass the client certificate step.

[...]

PR #839, opened 2021-08-27
Another round of v1alpha2 cleanup

robscott:
**What type of PR is this?**
/kind cleanup

This takes care of another big chunk of feedback from #780 as well as some of the items in #790.

```release-note
* "Controller" has been renamed to "ControllerName"
* "Admitted" condition has been renamed to "Accepted" and now defaults to an "Unknown" state instead of "False"
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -41,7 +41,7 @@ type Gateway struct {
 
 	// Status defines the current state of Gateway.
 	//
-	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "Unknown", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
 	Status GatewayStatus `json:"status,omitempty"`
 }
 
@@ -71,6 +71,9 @@ type GatewaySpec struct {
 	// Listeners associated with this Gateway. Listeners define
 	// logical endpoints that are bound on this Gateway's addresses.
 	// At least one Listener MUST be specified.
+	//
+	// Each listener in a Gateway must have a unique combination of Hostname,
+	// Port, and Protocol.
 	//
 	// An implementation MAY group Listeners by Port and then collapse each
 	// group of Listeners into a single Listener if the implementation
@@ -142,10 +145,8 @@ type GatewaySpec struct {
 	Addresses []GatewayAddress `json:"addresses,omitempty"`
 }
 
-// Listener embodies the concept of a logical endpoint where a Gateway can
-// accept network connections. Each listener in a Gateway must have a unique
-// combination of Hostname, Port, and Protocol. This will be enforced by a
-// validating webhook.
+// Listener embodies the concept of a logical endpoint where a Gateway accepts
+// network connections.
 type Listener struct {
 	// Name is the name of the Listener.
 	//
@@ -157,10 +158,20 @@ type Listener struct {
 	// field is ignored for protocols that don't require hostname based
 	// matching.
 	//
+	// Implementations MUST apply Hostname matching appropriately for each of
+	// the following protocols:
+	//
+	// * TLS: The Listener Hostname MUST match the SNI.
+	// * HTTP: The Listener Hostname MUST match the Host header of the request.
+	// * HTTPS: The Listener Hostname SHOULD match at both the TLS and HTTP
+	//   protocol layers as described above. If an implementation does not
+	//   ensure that both the SNI and Host header match the Listener hostname,
+	//   it MUST clearly document that.
+	//
 	// For HTTPRoute and TLSRoute resources, there is an interaction with the
 	// `spec.hostnames` array. When both listener and route specify hostnames,
-	// there must be an intersection between the values for a Route to be
-	// admitted. For more information, refer to the Route specific Hostnames
+	// there MUST be an intersection between the values for a Route to be
+	// accepted. For more information, refer to the Route specific Hostnames
 	// documentation.
 	//
 	// Support: Core
@@ -175,18 +186,6 @@ type Listener struct {
 	Port PortNumber `json:"port"`
 
 	// Protocol specifies the network protocol this listener expects to receive.
-	// The GatewayClass MUST apply the Hostname match appropriately for each
-	// protocol:
-	//
-	// * For the "TLS" protocol, the Hostname match MUST be
-	//   applied to the [SNI](https://tools.ietf.org/html/rfc6066#section-3)
-	//   server name offered by the client.
-	// * For the "HTTP" protocol, the Hostname match MUST be
-	//   applied to the host portion of the
-	//   [effective request URI](https://tools.ietf.org/html/rfc7230#section-5.5)
-	//   or the [:authority pseudo-header](https://tools.ietf.org/html/rfc7540#section-8.1.2.3)
-	// * For the "HTTPS" protocol, the Hostname match MUST be
-	//   applied at both the TLS and HTTP protocol layers.
 	//
 	// Support: Core
 	Protocol ProtocolType `json:"protocol"`
@@ -429,8 +428,6 @@ type RouteGroupKind struct {
 type GatewayAddress struct {
 	// Type of the address.
 	//
-	// Support: Extended
-	//
 	// +optional
 	// +kubebuilder:default=IPAddress
 	Type *AddressType `json:"type,omitempty"`
@@ -506,7 +503,7 @@ type GatewayStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{type: "Scheduled", status: "False", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	// +kubebuilder:default={{type: "Scheduled", status: "Unknown", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
 	// Listeners provide status for each unique listener port defined in the Spec.
@@ -599,12 +596,11 @@ const (
 	// more Listeners are not ready to serve traffic.
 	GatewayReasonListenersNotReady GatewayConditionReason = "ListenersNotReady"
 
-	// This reason is used with the "Ready" condition when the requested
-	// address has not been assigned to the Gateway. This reason
-	// can be used to express a range of circumstances, including
-	// (but not limited to) IPAM address exhaustion, invalid
-	// or unsupported address requests, or a named address not
-	// being found.
+	// This reason is used with the "Ready" condition when none of the requested
+	// addresses have been assigned to the Gateway. This reason can be used to
+	// express a range of circumstances, including (but not limited to) IPAM
+	// address exhaustion, invalid or unsupported address requests, or a named
+	// address not being found.
 	GatewayReasonAddressNotAssigned GatewayConditionReason = "AddressNotAssigned"
 )
 
@@ -614,17 +610,14 @@ type ListenerStatus struct {
 	Name SectionName `json:"name"`
 
 	// SupportedKinds is the list indicating the Kinds supported by this
-	// listener. When this is not specified on the Listener, this MUST represent
-	// the kinds an implementation supports for the specified protocol. When
-	// there are kinds specified on the Listener, this MUST represent the
-	// intersection of those kinds and the kinds supported by the implementation
-	// for the specified protocol.
+	// listener. This MUST represent the kinds an implementation supports for
+	// that Listener configuration.
 	//
-	// If kinds are specified in Spec that are not supported, an implementation
-	// MUST set the "ResolvedRefs" condition to "False" with the
-	// "InvalidRouteKinds" reason. If both valid and invalid Route kinds are
-	// specified, the implementation should support the valid Route kinds that
-	// have been specified.
+	// If kinds are specified in Spec that are not supported, they MUST NOT
+	// appear in this list and an implementation MUST set the "ResolvedRefs"
+	// condition to "False" with the "InvalidRouteKinds" reason. If both valid
+	// and invalid Route kinds are specified, the implementation MUST
+	// reference the valid Route kinds that have been specified.
 	//
 	// +kubebuilder:validation:MaxItems=8
 	SupportedKinds []RouteGroupKind `json:"supportedKinds"`
@@ -721,8 +714,12 @@ const (
 	// interoperability.
 	ListenerConditionDetached ListenerConditionType = "Detached"
 
-	// This reason is used with the "Detached" condition when the
-	// Listener requests a port that cannot be used on the Gateway.
+	// This reason is used with the "Detached" condition when the Listener
+	// requests a port that cannot be used on the Gateway. This reason could be
+	// used in a number of instances, including:
+	//
+	// * The port is already in use.
+	// * The port is not supported by the implementation.
 	ListenerReasonPortUnavailable ListenerConditionReason = "PortUnavailable"
 
 	// This reason is used with the "Detached" condition when the
@@ -736,9 +733,12 @@ const (
 	// protocol type is not supported.
 	ListenerReasonUnsupportedProtocol ListenerConditionReason = "UnsupportedProtocol"
 
-	// This reason is used with the "Detached" condition when
-	// the Listener could not be attached to the Gateway because the
-	// requested address is not supported.
+	// This reason is used with the "Detached" condition when the Listener could
+	// not be attached to the Gateway because the requested address is not
+	// supported. This reason could be used in a number of instances, including:
+	//
+	// * The address is already in use.
+	// * The type of address is not supported by the implementation.
 	ListenerReasonUnsupportedAddress ListenerConditionReason = "UnsupportedAddress"
 
 	// This reason is used with the "Detached" condition when the condition is
--- a/apis/v1alpha2/gatewayclass_types.go
+++ b/apis/v1alpha2/gatewayclass_types.go
@@ -56,7 +56,7 @@ type GatewayClass struct {
 
 	// Status defines the current state of GatewayClass.
 	//
-	// +kubebuilder:default={conditions: {{type: "Admitted", status: "False", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	// +kubebuilder:default={conditions: {{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Waiting", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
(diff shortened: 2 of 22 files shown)


apis/v1alpha2/tlsroute_types.go
```diff
@@ -49,26 +49,35 @@ type TLSRouteSpec struct {
...
-	// 2. If SNI is a wildcard, then the request matches this Route if the
-	//    SNI is to equal to the suffix (removing the first label) of the
-	//    wildcard.
-	// 3. If SNIs are unspecified, all requests associated with the gateway TLS
-	//    listener will match. This can be used to define a default backend
-	//    for a TLS listener.
+	// SNI attribute of TLS ClientHello message in TLS handshake. This matches
+	// the RFC 1123 definition of a hostname with 2 notable exceptions:
+	//
+	// 1. IPs are not allowed.
```
howardjohn, line 55: IPs are not allowed in SNI in the RFC

apis/v1alpha2/tlsroute_types.go
```diff
@@ -49,26 +49,35 @@ type TLSRouteSpec struct {
...
+	//
+	// If both the Listener and TLSRoute have specified hostnames, any
+	// TLSRoute hostnames that do not match the Listener hostname MUST be
+	// ignored. For example, if a Listener specified `*.example.com`, and the
+	// TLSRoute specified `test.example.com` and `test.example.net`,
+	// `test.example.net` must not be considered for a match.
+	//
+	// If hostnames do not match with the criteria above, then the TLSRoute is
+	// not accepted, and the implementation must raise an 'Accepted' Condition
+	// with a status of `False` for the target Listener(s).
```
howardjohn, line 80: This is a bit confusing when not using `sectionName` in parentRefs. Since the target is not a listener

apis/v1alpha2/tlsroute_types.go
```diff
@@ -49,26 +49,35 @@ type TLSRouteSpec struct {
...
+	//   `test.example.com` and `*.example.com` would both match. On the other
+	//   hand, `example.com` and `test.example.net` would not match.
+	//
+	// If both the Listener and TLSRoute have specified hostnames, any
+	// TLSRoute hostnames that do not match the Listener hostname MUST be
+	// ignored. For example, if a Listener specified `*.example.com`, and the
+	// TLSRoute specified `test.example.com` and `test.example.net`,
+	// `test.example.net` must not be considered for a match.
+	//
+	// If hostnames do not match with the criteria above, then the TLSRoute is
```
howardjohn, line 78: I would clarify: `If <any,all> hostnames ...`. And should be All IMO

apis/v1alpha2/gateway_types.go
```diff
@@ -72,6 +72,9 @@ type GatewaySpec struct {
 	// logical endpoints that are bound on this Gateway's addresses.
 	// At least one Listener MUST be specified.
 	//
+	// Each listener in a Gateway must have a unique combination of Hostname,
+	// Port, and Protocol. This will be enforced by a validating webhook.
```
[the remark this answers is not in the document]
hbagdi: Recommend adding an issue instead.

apis/v1alpha2/gateway_types.go
```diff
@@ -143,9 +146,7 @@ type GatewaySpec struct {
 }
 
 // Listener embodies the concept of a logical endpoint where a Gateway can
-// accept network connections. Each listener in a Gateway must have a unique
-// combination of Hostname, Port, and Protocol. This will be enforced by a
-// validating webhook.
+// accept network connections.
```
hbagdi, line 149: // Listener embodies the concept of a logical endpoint where a Gateway ~can~ accepts network connections.

apis/v1alpha2/gateway_types.go
```diff
@@ -157,10 +158,22 @@ type Listener struct {
...
+	// the following protocols:
+	//
+	// * TLS: The Hostname match MUST be applied to the
+	//   [SNI](https://tools.ietf.org/html/rfc6066#section-3) server name
+	//   offered by the client.
+	// * HTTP: The Hostname match MUST be applied to the host portion of the
+	//   [effective request URI](https://tools.ietf.org/html/rfc7230#section-5.5)
+	//   or the [:authority pseudo-header](https://tools.ietf.org/html/rfc7540#section-8.1.2.3)
+	// * HTTPS: The Hostname match MUST be applied at both the TLS and HTTP
+	//   protocol layers.
```
hbagdi, line 171: @youngnick Is this sufficient to close the domain fronting issue (#72) as well?

apis/v1alpha2/tcproute_types.go
```diff
@@ -27,7 +27,9 @@ import (
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
-// TCPRoute is the Schema for the TCPRoute resource.
+// TCPRoute provides a way to route TCP requests. When combined with a Gateway
+// listener, it can be used to forward requests to the port specified by the
```
hbagdi, line 31:
```suggestion
// listener, it can be used to forward connections on the port specified by the
```

apis/v1alpha2/tlsroute_types.go
hbagdi: Any relevant use cases for supporting IPs?
Can we start narrow and open up later?

apis/v1alpha2/gateway_types.go
robscott: Although this isn't entirely new content, now that I'm seeing it here, I think this wording may be a bit too strong. I'm not sure that every implementation _can_ apply a match at both layers.

apis/v1alpha2/tlsroute_types.go
robscott: I think you're both agreeing that IPs should not be allowed, but John is also saying we probably don't even need to specify that here.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
youngnick: I'd include the TLS details in that list as well, tbh.

apis/v1alpha2/gateway_types.go
youngnick: I think the direction is good, but maybe we need to specify this in terms of API objects, like:
```
- TLS: the Listener hostname must match the SNI.
- HTTP: The HTTPRoute hostname must match the Host header of the request.
- HTTPS: The Listener Hostname and the HTTPRoute hostname must agree with each other, and also match request properties as above.
```

youngnick: Because the important part is that we expect that Listeners and HTTPRoutes that do not match these rules should not end up attached.

apis/v1alpha2/gateway_types.go
```diff
@@ -610,17 +608,14 @@ type ListenerStatus struct {
...
-	// "InvalidRouteKinds" reason. If both valid and invalid Route kinds are
-	// specified, the implementation should support the valid Route kinds that
-	// have been specified.
+	// listener. This MUST represent the kinds an implementation supports for
+	// that Listener configuration.
+	//
+	// If kinds are specified in Spec that are not supported, they MUST NOT
+	// appear in this list and an implementation MUST set the "ResolvedRefs"
+	// condition to "False" with the "InvalidRouteKinds" reason. If both valid
+	// and invalid Route kinds are specified, the implementation should support
```
youngnick, line 617:
```suggestion
	// and invalid Route kinds are specified, the implementation should reference
```

apis/v1alpha2/shared_types.go
```diff
@@ -230,19 +230,24 @@ type RouteStatus struct {
...
 	// means the route has not been attached to any Gateway.
 	//
 	// +kubebuilder:validation:MaxItems=32
 	Parents []RouteParentStatus `json:"parents"`
 }
 
-// Hostname is the fully qualified domain name of a network host, as defined
-// by RFC 3986. Note the following deviations from the "host" part of the
-// URI as defined in the RFC:
+// Hostname is the fully qualified domain name of a network hostThis matches
```
youngnick, line 245:
```suggestion
// Hostname is the fully qualified domain name of a network host. This matches
```

apis/v1alpha2/tlsroute_types.go
youngnick: This is copied from other places which abide exactly by RFC1123, which *does* say that it SHOULD be possible to enter an IP in dotted-decimal format as a hostname, so it's important in those cases.

I agree that since SNI explicitly disallows literal addresses in the hostname field, it's not actually needed here, but I think it doesn't hurt and makes the rules across the different types of resources in our API more similar.

In short, I think we should probably keep this.

apis/v1alpha2/tlsroute_types.go
youngnick: +1 that it the Route should only not be accepted if *all* hostnames do not match, and if at least one does, a ResolvedRefs condition should be raised as well.

apis/v1alpha2/tlsroute_types.go
youngnick: Yeah, this comes down to the behavior with partial Listener acceptance, I think if we mandate "all", and have a ResolvedRefs in the event not all match, then this covers the sectionName case.

youngnick (commented):
Overall LGTM, some small comments.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: Created https://github.com/kubernetes-sigs/gateway-api/issues/847 to track webhook validation.

> I'd include the TLS details in that list as well, tbh.

Do you mean that hostname, listener, and protocol can be identical across listener if TLS is different?

apis/v1alpha2/tlsroute_types.go
robscott: I agree that we need a better way to handle status when `sectionName` is not specified. 2 options I've been thinking of:

1. A standard/reserved listener name such as "all" that is used in status to cover failed attachment to the Gateway as a whole.
2. A new set of condition types for Gateway status specifically for this purpose. (We could potentially reuse the existing listener condition types).

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: Removed reference to webhook in godocs here.

apis/v1alpha2/gateway_types.go
robscott: I don't think we should mix up the route type with the listener protocol here. While there's likely a ~99% overlap here, if we tie route types too closely here, we prevent or at least complicate extended route types. I've copied most of your suggestions from above, but removed any references to Routes - the following paragraph already covers that intersection.

apis/v1alpha2/gateway_types.go

apis/v1alpha2/gatewayclass_types.go
```diff
@@ -130,8 +126,8 @@ type ParametersReference struct {
 	Scope *string `json:"scope,omitempty"`
 
 	// Namespace is the namespace of the referent.
-	// This field is required when scope is set to "Namespace" and ignored when
-	// scope is set to "Cluster".
+	// This field is required when scope is set to "Namespace" must be unset
```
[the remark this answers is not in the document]

apis/v1alpha2/shared_types.go

apis/v1alpha2/tcproute_types.go
robscott: Updated, thanks!

apis/v1alpha2/tlsroute_types.go
robscott: Kept in place since the starting point is RFC 1123, but added a reference to RFC 6066 for this bullet point to explain that IP addresses were not allowed in SNI names.

apis/v1alpha2/tlsroute_types.go
robscott: >+1 that it the Route should only not be accepted if all hostnames do not match

Updated, thanks!

> and if at least one does, a ResolvedRefs condition should be raised as well.

Wouldn't this result in a successful attachment? Why would we raise a ResolvedRefs condition?

robscott:
@youngnick @hbagdi @howardjohn Thanks for the great feedback on this! I think I've responded to everything, PTAL.

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
youngnick: > Do you mean that hostname, listener, and protocol can be identical across listener if TLS is different?

Oh, you're quite right, that makes no sense. I was more thinking that TLS details are part of the key for this, but you're right that if the hostname, port, and protocol are identical, then the TLS details must not be different.

apis/v1alpha2/tlsroute_types.go
youngnick: I think option 2 seems better - if no listeners are valid, that seems very relevant for top-level status, but I would have thought we had that covered already somehow.

apis/v1alpha2/tlsroute_types.go
youngnick: I guess you're right, the thing I wanted to make sure we have is to indicate what hostnames *are* in use somehow. If you've got a bunch of hostnames and only one is in use, that seems like it would be good to know.

apis/v1alpha2/httproute_types.go
```diff
@@ -487,6 +484,10 @@ type HTTPRouteFilter struct {
 	// Implementers are encouraged to define custom implementation types to
 	// extend the core API with implementation-specific behavior.
 	//
+	// If a reference to a custom filter type cannot be resolved, the filter
+	// MUST NOT be skipped. Instead, requests that would have been processed by
+	// that filter SHOULD receive a HTTP error response.
```
youngnick, line 489: Why would they not do this? If it's only because the implementation doesn't support it, then we should say "MUST, except if".

youngnick (commented):
LGTM aside from me being nitpicky about SHOULD.

I think that we need to ensure that every time we use SHOULD, we clearly explain the reasons that is not MUST, and consider using "MUST, except if" instead.

I think that SHOULD leaves too much room for interpretation without clarification.

apis/v1alpha2/gateway_types.go
```diff
@@ -157,10 +158,18 @@ type Listener struct {
 	// field is ignored for protocols that don't require hostname based
 	// matching.
 	//
+	// Implementations MUST apply Hostname matching appropriately for each of
+	// the following protocols:
+	//
+	// * TLS: The Listener Hostname MUST match the SNI.
+	// * HTTP: The Listener Hostname MUST match the Host header of the request.
+	// * HTTPS: The Listener Hostname SHOULD match at both the TLS and HTTP
+	//   protocol layers as described above.
```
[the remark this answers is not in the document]
howardjohn: At the very least, the SNI MUST match the listener hostname I would think. And the HTTP Host MUST match the HTTPRoute hostname. And the listener/route hostnames MUST form an intersection. Therefor, there MUST be some intersection between the two if both are specified. For example, if I have listener hostname `*.example.com`, then both MUST have `*.example.com`.

Not that we should write that as is.. just brain dumping.

apis/v1alpha2/tlsroute_types.go
robscott: https://github.com/kubernetes-sigs/gateway-api/issues/848

apis/v1alpha2/tlsroute_types.go
robscott: https://github.com/kubernetes-sigs/gateway-api/issues/849

apis/v1alpha2/httproute_types.go
robscott: Yeah I think you're right, can't think of a reason not to use MUST here.

apis/v1alpha2/gateway_types.go
```diff
@@ -611,17 +605,14 @@ type ListenerStatus struct {
...
-	// specified, the implementation should support the valid Route kinds that
-	// have been specified.
+	// listener. This MUST represent the kinds an implementation supports for
+	// that Listener configuration.
+	//
+	// If kinds are specified in Spec that are not supported, they MUST NOT
+	// appear in this list and an implementation MUST set the "ResolvedRefs"
+	// condition to "False" with the "InvalidRouteKinds" reason. If both valid
+	// and invalid Route kinds are specified, the implementation should
+	// reference the valid Route kinds that have been specified.
```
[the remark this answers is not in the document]

apis/v1alpha2/gateway_types.go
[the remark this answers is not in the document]
robscott: I took a shot at updating this. What I'm trying to communicate here is that if implementations can implement the match at both levels they should, but not every implementation can.

youngnick:
I think you got it @robscott, nice work.

/lgtm

/hold
for another lgtm though.

apis/v1alpha2/gateway_types.go
```diff
@@ -718,8 +711,11 @@ const (
 	// interoperability.
 	ListenerConditionDetached ListenerConditionType = "Detached"
 
-	// This reason is used with the "Detached" condition when the
-	// Listener requests a port that cannot be used on the Gateway.
+	// This reason is used with the "Detached" condition when the Listener
+	// requests a port that cannot be used on the Gateway. This reason could be
+	// used in a number of instances, including:
```
jpeach, line 716: Need a newline here to render the markdown bulleted list.

apis/v1alpha2/gateway_types.go
```diff
@@ -733,9 +729,11 @@ const (
 	// protocol type is not supported.
 	ListenerReasonUnsupportedProtocol ListenerConditionReason = "UnsupportedProtocol"
 
-	// This reason is used with the "Detached" condition when
-	// the Listener could not be attached to the Gateway because the
-	// requested address is not supported.
+	// This reason is used with the "Detached" condition when the Listener could
+	// not be attached to the Gateway because the requested address is not
+	// supported. This reason could be used in a number of instances, including:
```
jpeach, line 734: Need a newline here to render the markdown bulleted list.

apis/v1alpha2/httproute_types.go
```diff
@@ -455,16 +453,15 @@ type HTTPRouteMatch struct {
...
-// +union
+// HTTPRouteFilter defines processing steps that must be completed during the
+// request or response lifecycle. HTTPRouteFilters are meant as an extension
+// point to express processing that may be done in Gateway implementations. Some
+// examples include request or response modification, implementing
+// authentication strategies, rate-limiting, and traffic shaping. API
+// guarantee/conformance is defined based on the type of the filter.
+// TODO(hbagdi): re-render CRDs once controller-tools supports union tags: -
+// https://github.com/kubernetes-sigs/controller-tools/pull/298 -
+// https://github.com/kubernetes-sigs/controller-tools/issues/461 +union
```
jpeach, line 464: The `+union` should be on a line by itself. Separate TODO into a separate paragraph, or drop it into an issue.

jpeach:
Just a couple of formatting nits.

/lgtm

apis/v1alpha2/httproute_types.go
robscott: Good call, removed entirely: https://github.com/kubernetes-sigs/gateway-api/issues/854.

apis/v1alpha2/gateway_types.go
robscott: Updated, thanks!

apis/v1alpha2/gateway_types.go
robscott: Updated, thanks!

PR #842, opened 2021-08-31
Add protocol specific validation for Gateway api

### LINKED ISSUE #827: Validation should prevent protocol specific config errors on Gateways
We should validate the following:

TLSConfig is empty when the protocol is HTTP, TCP, or UDP.
Hostname is empty when the protocol is TCP or UDP.

**Why this is needed**:
This would help provide a better UX.

bishtsaurabh5:
This commit adds the following validations

- TLSConfig is empty when the protocol is HTTP, TCP, or UDP.
- Hostname is empty when the protocol is TCP or UDP.

**What type of PR is this?**

kind/feature

Introduces protocol specific validations to
provide better UX

Fixes #827 

```release-note
This change introduces protocol specific validations
*  TLSConfig is empty when the protocol is HTTP, TCP, or UDP
*  Hostname is empty when the protocol is TCP or UDP.

```

### DIFF
--- a/apis/v1alpha2/validation/gateway.go
+++ b/apis/v1alpha2/validation/gateway.go
@@ -17,9 +17,25 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+var (
+	// set of protocols for which we need to validate that hostname is empty
+	protocolsHostnameInvalid = map[gatewayv1a2.ProtocolType]struct{}{
+		gatewayv1a2.TCPProtocolType: {},
+		gatewayv1a2.UDPProtocolType: {},
+	}
+	// set of protocols for which TLSConfig shall not be present
+	protocolsTLSInvalid = map[gatewayv1a2.ProtocolType]struct{}{
+		gatewayv1a2.HTTPProtocolType: {},
+		gatewayv1a2.UDPProtocolType:  {},
+		gatewayv1a2.TCPProtocolType:  {},
+	}
 )
 
 // ValidateGateway validates gw according to the Gateway API specification.
@@ -29,5 +45,47 @@ import (
 // Validation that is not possible with CRD annotations may be added here in the future.
 // See https://github.com/kubernetes-sigs/gateway-api/issues/868 for more information.
 func ValidateGateway(gw *gatewayv1a2.Gateway) field.ErrorList {
-	return nil
+	return validateGatewaySpec(&gw.Spec, field.NewPath("spec"))
+}
+
+// validateGatewaySpec validates whether required fields of spec are set according to the
+// Gateway API specification.
+func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.ErrorList {
+	return validateGatewayListeners(spec.Listeners, path.Child("listeners"))
+}
+
+// validateGatewayListeners validates whether required fields of listeners are set according
+// to the Gateway API specification.
+func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errs = append(errs, validateListenerTLSConfig(listeners, path)...)
+	errs = append(errs, validateListenerHostname(listeners, path)...)
+	return errs
+}
+
+func validateListenerTLSConfig(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, l := range listeners {
+		if isProtocolInSubset(l.Protocol, protocolsTLSInvalid) && l.TLS != nil {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("tls"), fmt.Sprintf("should be empty for protocol %v", l.Protocol)))
+		}
+	}
+	return errs
+}
+
+func isProtocolInSubset(protocol gatewayv1a2.ProtocolType, set map[gatewayv1a2.ProtocolType]struct{}) bool {
+	_, ok := set[protocol]
+	return ok
+}
+
+// validateListenerHostname validates each listener hostname
+// should be empty in case protocol is TCP or UDP
+func validateListenerHostname(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, h := range listeners {
+		if isProtocolInSubset(h.Protocol, protocolsHostnameInvalid) && h.Hostname != nil {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("hostname"), fmt.Sprintf("should be empty for protocol %v", h.Protocol)))
+		}
+	}
+	return errs
 }
--- a/apis/v1alpha2/validation/gateway_test.go
+++ b/apis/v1alpha2/validation/gateway_test.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+func TestValidateGateway(t *testing.T) {
+	listeners := []gatewayv1a2.Listener{
+		{
+			Hostname: nil,
+		},
+	}
+	baseGateway := gatewayv1a2.Gateway{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Namespace: metav1.NamespaceDefault,
+		},
+		Spec: gatewayv1a2.GatewaySpec{
+			GatewayClassName: "foo",
+			Listeners:        listeners,
+		},
+	}
+	tlsConfig := gatewayv1a2.GatewayTLSConfig{}
+
+	testCases := map[string]struct {
+		mutate             func(gw *gatewayv1a2.Gateway)
+		expectErrsOnFields []string
+	}{
+		"tls config present with http protocol": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				gw.Spec.Listeners[0].Protocol = gatewayv1a2.HTTPProtocolType
+				gw.Spec.Listeners[0].TLS = &tlsConfig
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].tls"},
+		},
+		"tls config present with tcp protocol": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				gw.Spec.Listeners[0].Protocol = gatewayv1a2.TCPProtocolType
+				gw.Spec.Listeners[0].TLS = &tlsConfig
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].tls"},
+		},
+		"hostname present with tcp protocol": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				hostname := gatewayv1a2.Hostname("foo.bar.com")
+				gw.Spec.Listeners[0].Hostname = &hostname
+				gw.Spec.Listeners[0].Protocol = gatewayv1a2.TCPProtocolType
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].hostname"},
+		},
+		"hostname present with udp protocol": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				hostname := gatewayv1a2.Hostname("foo.bar.com")
+				gw.Spec.Listeners[0].Hostname = &hostname
+				gw.Spec.Listeners[0].Protocol = gatewayv1a2.UDPProtocolType
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].hostname"},
+		},
+	}
+
+	for name, tc := range testCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			gw := baseGateway.DeepCopy()
+			tc.mutate(gw)
+			errs := ValidateGateway(gw)
+			if len(tc.expectErrsOnFields) != len(errs) {
+				t.Fatalf("Expected %d errors, got %d errors: %v", len(tc.expectErrsOnFields), len(errs), errs)
+			}
+			for i, err := range errs {
+				if err.Field != tc.expectErrsOnFields[i] {
+					t.Errorf("Expected error on field: %s, got: %s", tc.expectErrsOnFields[i], err.Error())
+				}
+			}
+		})
+	}
+}


bishtsaurabh5:
@robscott can you please review this PR .

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +46,32 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
 // to the Gateway API specification.
 func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
 	// TODO [danehans]: Add additional validation of listener fields.
-	return validateListenerHostname(listeners, path)
+	var errs field.ErrorList
+	e1:=validateListenerTLSConfig(listeners, path)
+	errs = append(errs,e1...)
+	e2:=validateListenerHostname(listeners, path)
+	errs = append(errs,e2...)
```
robscott, line 53: Tiny Nit:
```suggestion
	e1 := validateListenerTLSConfig(listeners, path)
	errs = append(errs, e1...)
	e2 := validateListenerHostname(listeners, path)
	errs = append(errs, e2...)
```

robscott: This same general comment applies through most of this code, essentially to add spaces:

- after commas
- before and after `:=`

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +46,32 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
...
+	return errs
+}
+
+func validateListenerTLSConfig(listeners []gatewayv1a2.Listener,path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	tlsHttpsProtocols := []gatewayv1a2.ProtocolType{gatewayv1a2.HTTPSProtocolType,gatewayv1a2.TLSProtocolType}
+	for i,l := range listeners{
+		if !contains(tlsHttpsProtocols,l.Protocol) && l.TLS != nil {
```
robscott, line 61: Our protocol validation allows for custom protocols. That means we actually have to invert this logic to only prevent these fields from being set if the protocol is "TCP", "UDP", or "HTTP".

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +46,32 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
...
+			errs = append(errs,field.Forbidden(path.Index(i).Child("tls"),fmt.Sprintf("should be empty for protocol %v",l.Protocol)))
+		}
+	}
+	return errs
+}
+
+func contains(protocolTypes []gatewayv1a2.ProtocolType, protocolType gatewayv1a2.ProtocolType)  bool {
+	for _, v := range protocolTypes {
+		if reflect.DeepEqual(v,protocolType) {
```
robscott, line 70: Why do we need DeepEqual() here?

apis/v1alpha2/validation/gateway.go
```diff
@@ -55,8 +82,12 @@ func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path
 func validateListenerHostname(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
 	for i, h := range listeners {
+		tcpUdpProtocols := []gatewayv1a2.ProtocolType{gatewayv1a2.TCPProtocolType,gatewayv1a2.UDPProtocolType}
 		// When unspecified, “”, or *, all hostnames are matched.
 		if h.Hostname == nil || (*h.Hostname == "" || *h.Hostname == "*") {
```
robscott, line 87: Not introduced by your PR, but would be good to fix while in the vicinity.
```suggestion
		if h.Hostname == nil {
```

apis/v1alpha2/validation/gateway.go
```diff
@@ -55,8 +82,12 @@ func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path
...
+			if contains(tcpUdpProtocols,h.Protocol) {
+				errs = append(errs, field.Required(path.Index(i).Child("hostname"), fmt.Sprintf("hostname is required for protocol type %v",h.Protocol)))
+			}
```
robscott, line 90: This is in a block where hostname is effectively empty. Instead we'd want to move this below.

robscott: The validation we actually want here is to say that Hostname must be empty/nil if the protocol is TCP or UDP.

apis/v1alpha2/validation/gateway_test.go
```diff
@@ -77,6 +77,24 @@ func TestValidateGateway(t *testing.T) {
 			},
 			expectErrsOnFields: []string{},
 		},
+		"tls config present with tls protocol": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				protocol := gatewayv1a2.TLSProtocolType
+				tlsConfig := gatewayv1a2.GatewayTLSConfig{}
+				gw.Spec.Listeners[0].Protocol = protocol
+				gw.Spec.Listeners[0].TLS = &tlsConfig
```
robscott, line 85: Would this work? If so, I'd recommend updating all the similar tests below as well.
```suggestion
				gw.Spec.Listeners[0].Protocol = gatewayv1a2.TLSProtocolType
				gw.Spec.Listeners[0].TLS = &gatewayv1a2.GatewayTLSConfig{}
```

robscott (commented):
Thanks for the work on this @bishtsaurabh5!

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: can hostname not have a wildcard character ? Will the empty string be translated to nil ?

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: Fixed this . Do you have any linter for this ? I did go fmt before submitting the patch . I am new to go here :)

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: removed the DeepEqual apparently the ProtocolType is a simple wrapper over string

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: Have not done this change in the latest patch as I am unsure whether this should be removed as comments have already been added by someone for empty and * hostnames , tests are also present for same

apis/v1alpha2/validation/gateway_test.go
bishtsaurabh5: Declared tlsConfig as a global variable for all tests so that we can use that directly over there as we are not modifying that

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +45,32 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
...
+	e1 := validateListenerTLSConfig(listeners, path)
+	errs = append(errs, e1...)
+	e2 := validateListenerHostname(listeners, path)
+	errs = append(errs, e2...)
+	return errs
+}
+
+func validateListenerTLSConfig(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	nonTlsProtocols := []gatewayv1a2.ProtocolType{gatewayv1a2.HTTPProtocolType, gatewayv1a2.UDPProtocolType, gatewayv1a2.TCPProtocolType}
```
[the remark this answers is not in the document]
bishtsaurabh5: Not sure if this would be used globally by  other functions too

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +45,32 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
 // to the Gateway API specification.
 func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
 	// TODO [danehans]: Add additional validation of listener fields.
-	return validateListenerHostname(listeners, path)
+	var errs field.ErrorList
```
howardjohn, line 48: nit: 
```go
var errs field.ErrorList
errs = append(errs, validateListenerTLSConfig(listeners, path)...)
errs = append(errs, validateListenerHostname(listeners, path)...)
return errs
```

otherwise we will end up with e3, e4, ..., e99

bishtsaurabh5: seems fair to me will fix this :) . Thanks for your review

apis/v1alpha2/validation/gateway.go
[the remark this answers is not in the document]
bishtsaurabh5: Thanks for your review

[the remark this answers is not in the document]
bishtsaurabh5: @tokers can we discuss as why do you think it should be placed outside ? I am fairly new to this project so I am  not aware of specific usecases where this would be helpful

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: Fixed in the latest patch

apis/v1alpha2/validation/gateway.go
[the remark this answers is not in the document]
tokers: The `nonTlsProtocols` is composed of constants, so it can be created before the function call, and we just read it, we don't have to create multiple copies when we enter this function, this is also good for the performance (but maybe it's trivial).

apis/v1alpha2/validation/gateway_test.go
```diff
@@ -141,6 +156,44 @@ func TestValidateGateway(t *testing.T) {
 			},
 			expectErrsOnFields: []string{"spec.listeners[0].hostname"},
 		},
```
hbagdi, line 158: add tests and validation code for:
- Protocol is TLS but TLSConfig is not specified - failure
- Protocol is HTTPS but TLSConfig is not specified - failure

robscott: I don't think we can safely validate that without conflicting with cert generation or other non-secret based certificates. In at least some cases users will want to leave this blank initially and have it filled in later with automation (or connected to some other kind of certificate).

hbagdi: Don't you need at least the `Mode` bit set?

robscott: There's a default value of `Terminate` for that field which seems sufficient to me

hbagdi: You are right. Resolving the thread.

apis/v1alpha2/validation/gateway.go
robscott: This was changed in response to sig-network review since "*" was identical to empty/unspecified, it did not make sense to allow the value. Here's the updated regex validation: https://github.com/kubernetes-sigs/gateway-api/blob/5ac109cf2d1c98006e2672e4484e352ec320a412/apis/v1alpha2/shared_types.go#L263.

bishtsaurabh5: Fixed

apis/v1alpha2/validation/gateway.go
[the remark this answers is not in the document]
hbagdi: I think moving this and the tcpUdpProtocols outside would be for the best because I expect other validations to use these as well.

Also, once you move them outside, replace the `contains()` with:
```go
func isProtocolInSubset(protocol gatewayv1a2.ProtocolType, set []gatewayv1a2.ProtocolType) bool {}
```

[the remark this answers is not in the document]
bishtsaurabh5: Thanks @tokers @hbagdi  will fix this

apis/v1alpha2/validation/gateway.go
```diff
@@ -59,7 +85,11 @@ func validateListenerHostname(listeners []gatewayv1a2.Listener, path *field.Path
 		if h.Hostname == nil || (*h.Hostname == "" || *h.Hostname == "*") {
 			continue
 		}
+		tcpUdpProtocols := []gatewayv1a2.ProtocolType{gatewayv1a2.TCPProtocolType, gatewayv1a2.UDPProtocolType}
```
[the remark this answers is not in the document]
bishtsaurabh5: Fixed

apis/v1alpha2/validation/gateway.go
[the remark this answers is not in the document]
bishtsaurabh5: Fixed

apis/v1alpha2/validation/gateway.go
```diff
@@ -55,11 +83,13 @@ func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path
...
 	var errs field.ErrorList
 	for i, h := range listeners {
-		// When unspecified, “”, or *, all hostnames are matched.
-		if h.Hostname == nil || (*h.Hostname == "" || *h.Hostname == "*") {
+		if h.Hostname == nil {
 			continue
 		}
 		hostname := string(*h.Hostname)
+		if isProtocolInSubset(h.Protocol, tcpUdpProtocols) {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("hostname"), fmt.Sprintf("hostname should not be present for protocol type %v", h.Protocol)))
```
jpeach, line 91: Nit: make this message consistent with the other one

```suggestion
errs = append(errs, field.Forbidden(path.Index(i).Child("hostname"), fmt.Sprintf(" should  be empty for protocol  %v", h.Protocol)))
```

apis/v1alpha2/validation/gateway.go
```diff
@@ -26,6 +27,11 @@ import (
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
+var (
+	tcpUdpProtocols = []gatewayv1a2.ProtocolType{gatewayv1a2.TCPProtocolType, gatewayv1a2.UDPProtocolType}
+	nonTlsProtocols = []gatewayv1a2.ProtocolType{gatewayv1a2.HTTPProtocolType, gatewayv1a2.UDPProtocolType, gatewayv1a2.TCPProtocolType}
+)
```
jpeach, line 33: Since these are only used for `isProtocolInSubset`, I'd make them a set, eg:
```Go
var map[ProtocolType]struct{} = map[ProtocolType]struct{}{
        gatewayv1a2.HTTPProtocolType: struct{}{},
        gatewayv1a2.UDPProtocolType: struct{}{},
        gatewayv1a2.TCPProtocolType: struct{}{},
}

```

You could take it a bit further to use a convenience type, but not sure if that's worth the effort:

```Go
type ProtocolSubset map[gatewayv1a2.ProtocolType]struct{}

func (p ProtocolSubset) Contains(proto gatewayv1a2.ProtocolType) bool {
        _, ok := p[proto]
        return ok
}
```

apis/v1alpha2/validation/gateway.go
jpeach: > Fixed this . Do you have any linter for this ? I did go fmt before submitting the patch . I am new to go here :)

`make verify` will run more linters

jpeach (commented):
Just need to make the error messages consistent.

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: I think making a map is a nice idea here . Making a separate convenicence type I am not sure about that . Uploading the patch for the map

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: Thanks Nice catch !! The new message looks more elegant

apis/v1alpha2/validation/gateway.go
robscott: As a note, this is very familiar to the widely used k8s set utils: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go.

apis/v1alpha2/validation/gateway_test.go
```diff
@@ -141,6 +143,57 @@ func TestValidateGateway(t *testing.T) {
...
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].hostname"},
+		},
+		"empty string hostname": {
+			mutate: func(gw *gatewayv1a2.Gateway) {
+				hostname := gatewayv1a2.Hostname("")
+				gw.Spec.Listeners[0].Hostname = &hostname
+			},
+			expectErrsOnFields: []string{"spec.listeners[0].hostname"},
+		},
```
robscott, line 196: Now that we're using regex validation on the hostname, we likely don't need the validation in the webhook. With that said, it is much nicer to be able to use upstream validation functions included here. Since this validation wasn't actually added in this PR, I don't think this needs to block anything, I've added https://github.com/kubernetes-sigs/gateway-api/issues/868 to track this.

robscott:
Thanks for the work on this! Will defer to someone who's been more involved in the review for the final LGTM (maybe @jpeach or @hbagdi can take a final look).

/approve

apis/v1alpha2/validation/gateway.go
```diff
@@ -26,6 +27,19 @@ import (
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
+var (
+	exists          = struct{}{}
```
jpeach, line 31: `exists` seems a bit non-idiomatic for Go. I'd just inline the struct literals where needed.

apis/v1alpha2/validation/gateway.go
```diff
@@ -26,6 +27,19 @@ import (
...
+	tcpUdpProtocols = map[gatewayv1a2.ProtocolType]struct{}{
+		gatewayv1a2.TCPProtocolType: exists,
+		gatewayv1a2.UDPProtocolType: exists,
+	}
+	nonTlsProtocols = map[gatewayv1a2.ProtocolType]struct{}{
```
jpeach, line 36: Nit: idiomatic Go capitalizes initialisms, so Udp and Tls should be UDP and TLS.

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +58,25 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
...
+	return errs
+}
+
+func validateListenerTLSConfig(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, l := range listeners {
+		if isProtocolInSubset(l.Protocol, nonTlsProtocols) && l.TLS != nil {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("tls"), fmt.Sprintf("should be empty for protocol %v", l.Protocol)))
+		}
+	}
```
jpeach, line 73: Sorry, I just realized that the TLS field should be populated for TLS-layered protocols and not for non-TLS protocols, so maybe something like this:

```suggestion
        for i, l := range listeners {        
                switch {    
                case isProtocolInSubset(l.Protocol, nonTlsProtocols):    
                        if l.TLS != nil {                                
                                errs = append(errs, field.Forbidden(path.Index(i).Child("tls"),    
                                        fmt.Sprintf("should be empty for protocol %v", l.Protocol)))    
                        }                                                                                                                               
                case isProtocolInSubset(l.Protocol, tlsProtocols):    
                        if l.TLS == nil {    
                                errs = append(errs, field.Forbidden(path.Index(i).Child("tls"),    
                                        fmt.Sprintf("should not be empty for protocol %v", l.Protocol)))    
                        }    
    
                }      
        }  
```

jpeach (changes requested):
Just a couple of small issues that I messed on the first pass.

apis/v1alpha2/validation/gateway.go
bowei: +1 this is not something we have elsewhere in the codebase, let's leave this out.

apis/v1alpha2/validation/gateway.go
```diff
@@ -26,6 +27,19 @@ import (
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
+var (
+	exists          = struct{}{}
+	tcpUdpProtocols = map[gatewayv1a2.ProtocolType]struct{}{
```
bowei, line 32: Is it not possible to use the set package from the k/k codebase

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +58,25 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
 // to the Gateway API specification.
 func validateGatewayListeners(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
 	// TODO [danehans]: Add additional validation of listener fields.
-	return validateListenerHostname(listeners, path)
+	var errs field.ErrorList
+	errs = append(errs, validateListenerTLSConfig(listeners, path)...)
```
bowei, line 62: I think we should delete the TODO from danehans since we have implementation now.

apis/v1alpha2/validation/gateway.go
```diff
@@ -44,7 +58,25 @@ func validateGatewaySpec(spec *gatewayv1a2.GatewaySpec, path *field.Path) field.
...
+	var errs field.ErrorList
+	for i, l := range listeners {
+		if isProtocolInSubset(l.Protocol, nonTlsProtocols) && l.TLS != nil {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("tls"), fmt.Sprintf("should be empty for protocol %v", l.Protocol)))
+		}
+	}
+	return errs
+}
+
+func isProtocolInSubset(protocol gatewayv1a2.ProtocolType, set map[gatewayv1a2.ProtocolType]struct{}) bool {
```
bowei, line 77: can we use the standard `set` package? the code will be more understandable.

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: Hey can you tell me the name of the package ?

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: @jpeach I understand but make caps for tcpUdpProtocols to TCPUDPProtocols would export it and I think camel case would be better since we append both the protocols in the name. another var nonTlsProtocols could be made to nonTLSProtocols but then one name would be exported and other would not be so it would be a bit inconsistent .

apis/v1alpha2/validation/gateway.go
bishtsaurabh5: @bowei I think the package in k/k does have it for generic type and rather it is for primitive types

apis/v1alpha2/validation/gateway.go
robscott: Unfortunately the k8s set package only works with a few primitive types (including string). In some other cases like EndpointSlice controller [we've created sets based on that interface that support specific custom types](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/util/endpointslice/endpointset.go). We could do that here, but it may be too complex for this specific use case.

robscott:
Hey @bishtsaurabh5, thanks for your work on this! Do you have time in the next few days to follow up with this PR? If not, that's completely fine, we can fork a PR from this to finish up the last few bits of this. We're hoping to publish a release next week, and this is one of the items we'd like to include.

bishtsaurabh5:
Sure @robscott thanks for your review. I will update the PR today been a little bit busy for the last few days :)

apis/v1alpha2/validation/gateway.go
robscott: I think the appropriate capitalization is `tcpUDPProtocols` and `nonTLSProtcols`, does that line up with what you were thinking @jpeach?

apis/v1alpha2/validation/gateway.go
robscott: Yep, I think there are valid reasons for TLS config to be empty even when the protocol is HTTPS or TLS. Between certs being attached through different means and controllers that may attach cert refs later, I don't think we can/should tighten the validation to require TLS config in those cases.

robscott:
@bishtsaurabh5 Thanks for the work on this! I just had the same nit as @jpeach on naming, but otherwise this LGTM.

apis/v1alpha2/validation/gateway.go
tokers: nonTLSProtcols is good but `tcpUDPProtocols` is a bit strange, is there any words we can use to represent TCP and UDP protocols?

robscott: Maybe we should focus on why we care about these sets of protocols:

- `nonTLSProtcols` -> `protocolsTLSInvalid`
- `tcpUDPProtocols` -> `protocolsHostnameInvalid`

Along with that change, it would probably be helpful to have a brief comment for each explaining the purpose of each set.

apis/v1alpha2/validation/gateway.go
```diff
@@ -17,17 +17,75 @@ limitations under the License.
...
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
+var (
+	// set of protocols for which we need to validate if hostname is empty
```
robscott, line 28: Tiny nit:
```suggestion
	// set of protocols for which we need to validate that hostname is empty
```

robscott:
Thanks!

/lgtm
/approve

bishtsaurabh5:
I want to thank everyone for actively taking part in the code review taking time out of their busy schedule I am new to open source and go , So It was really helpful folks :)

PR #850, opened 2021-09-03
httproute: drop ImplementationSpecific match types

### LINKED ISSUE #832: HTTPRoute matching should support custom match types

See the discussion here for more detail (https://github.com/kubernetes-sigs/gateway-api/pull/822#discussion_r696163543), but in short:

In GEP-820 (#820 and #822) we are removing the extensionRef in the HTTPRoute matches section. This issue covers work to add something back in that allows implementations to try custom matching of some sort.

@bowei suggested:
Modify `HTTPPathMatch.Type`:

* Allow qualified names (e.g. `acme.io/my-custom-match`). These represent custom match types
* The standard match types (`Exact`, `Prefix`, `RegularExpression`) stay
* Remove `ImplementationSpecific`

**Why this is needed**:
We agreed to go ahead with #822  on the proviso we come back, this is to make sure this doesn't get lost.

hbagdi:
**What type of PR is this?**

/kind api-change

ImplementationSpecific is trappy match type: it provides an escape hatch
to do more but it doesn't take into account that multiple
implementation-specific match types could be possible. Use-cases exist
in this area but we have not thought through them deeply.
Being conservative and dropping this feature in v1alpha2 to avoid any potential
breaking change in the future. We may revisit this in the future as we gain a
better understanding.

Fixes #832

```release-note
ImplementationSpecific match types in HTTPRoute's path, query, and header matches have been removed.
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -211,7 +211,6 @@ type HTTPRouteRule struct {
 // * "Exact"
 // * "Prefix"
 // * "RegularExpression"
-// * "ImplementationSpecific"
 //
 // Prefix and Exact paths must be syntactically valid:
 //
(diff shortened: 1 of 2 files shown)


robscott:
Adding a hold for one more person to sign off on this, but LGTM.

/lgtm
/hold

PR #855, opened 2021-09-08
docs: note the pseudo headers should be allowed explicitly

### LINKED ISSUE #740: Clarify if psuedo-headers are supported within HTTP header matching
**What happened**:

Pseudo-headers can be used to specify matches against non-header properties of HTTP requests: https://github.com/kubernetes-sigs/gateway-api/pull/733#discussion_r676935043

The documentation of the API clarify if pseudo-headers are supported or not.

N/A

tokers:
**What type of PR is this?**

/kind documentation

Document the pseudo headers should be allowed by implementations.

Fixes #740 

```release-note

```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -319,6 +319,10 @@ type HTTPHeaderMatch struct {
 	// Generally, proxies should follow the guidance from the RFC:
 	// https://www.rfc-editor.org/rfc/rfc7230.html#section-3.2.2 regarding
 	// processing a repeated header, with special handling for "Set-Cookie".
+	//
+	// Note implementations must allow the use of HTTP/2 pseudo headers
+	// (See https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.1)
+	// like ":method", ":path".
 	Name HTTPHeaderName `json:"name"`
 
 	// Value is the value of HTTP Header to be matched.


youngnick:
/ok-to-test

LGTM, but I'll let some others look before I issue the command. :)

jpeach:
@tokers @youngnick I'm not that confident about which implementations support matching on pseudo-headers. Maybe a "SHOULD" would be better?

tokers:
> @tokers @youngnick I'm not that confident about which implementations support matching on pseudo-headers. Maybe a "SHOULD" would be better?

If we use "SHOULD", then we also need to note the result if a solution doesn't implement it.

hbagdi:
The current (in the mainline dev branch) guidance here is opposite: https://github.com/kubernetes-sigs/gateway-api/blob/5ac109cf2d1c98006e2672e4484e352ec320a412/apis/v1alpha2/httproute_types.go#L270-L281

The linked issue (#740) can probably be closed since we have the above text now.

I think we should refrain from adding this right now and come back post v1alpha2 if there is a demand.

youngnick:
Sorry to lead you astray here @tokers.

tokers:
> Sorry to lead you astray here @tokers.

Never mind. I also didn't notice the guidance pasted by @hbagdi . So it looks like now we can close this PR?

hbagdi:
@tokers Seems like it.
I'll also close the linked issue.
/close

Issue #851, opened 2021-09-03
GEP: Allow Multiple Certificate Refs per Gateway Listener


robscott:
**What would you like to be added**:
The `CertificateRef` field in Listener TLSConfig should be replaced with a `CertificateRefs` field.

**Why this is needed**:
This would help with a number of use cases, including:

- When hostname is empty on the Listener, and Routes attach multiple unique Hostnames
- Providing RSA and EC certs
- Temporarily including new and old certs during an upgrade

Relevant Slack thread: https://kubernetes.slack.com/archives/CR0H13KGA/p1629912749150800

tokers:
Hi, there, what's the status of this GEP? I'd like to implement it :)

robscott:
Reopening because we need conformance test coverage for this.

PR #856, opened 2021-09-08
Adding GEP 851: Allow Multiple Certificate Refs per Gateway Listener

robscott:
**What type of PR is this?**
/kind gep

This adds GEP #851: Allow Multiple Certificate Refs per Gateway Listener.

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-851.md
+++ b/site-src/geps/gep-851.md
@@ -1,0 +1,102 @@
+# GEP-851: Allow Multiple Certificate Refs per Gateway Listener
+
+* Issue: [#851](https://github.com/kubernetes-sigs/gateway-api/issues/851)
+* Status: Implementable
+
+## TLDR
+
+Replace `CertificateRef` field with a `CertificateRefs` field in Gateway
+Listeners.
+
+## Goals
+
+* Allow certs to be attached for multiple unique Hostnames for the same
+  Listener. This is most helpful in the following situations:
+  * The hostname on the Listener is empty or unspecified allowing Routes to
+    attach arbitrary hostnames to the Listener with Routes.
+  * The hostname on the Listener is a wildcard and Route owners want to attach
+    multi-level subdomains. For example, `*.example.com` on the Listener and
+    `app.dev.example.com` on a Route.
+* Enable specifying RSA and ECDSA certs on the same Listener.
+* Improve cert upgrades by allowing new and old certs to be attached to the same
+  Listener.
+
+## Non-Goals
+
+* Be too prescriptive about how multiple certificates are handled for a Gateway
+  Listener. Each implementation may handle this slightly differently.
+
+## Introduction
+
+When we removed certificate references from HTTPRoute ([GEP
+746](/geps/gep-746/)) we lost the capability to attach multiple certificates to
+a Listener. This GEP proposes adding that capability back, directly to the
+Gateway Listener this time.
+
+In addition to helping with the self service approach, where app owners may
+attach multiple hostnames (and certs) to a single Listener, this can help with
+other use cases:
+
+* Attaching RSA and ECDSA certs.
+* Attaching both new and old certs during an upgrade.
+
+## API
+
+The `CertificateRef` field in `GatewayTLSConfig` would be replaced with the
+following `CertificateRefs` field:
+
+```go
+    // CertificateRefs contains a series of references to Kubernetes objects that
+    // contains TLS certificates and private keys. These certificates are used to
+    // establish a TLS handshake for requests that match the hostname of the
+    // associated listener.
+    //
+    // When multiple TLS certificates are specified, SNI matching MUST be used
+    // to determine the appropriate certificate to use.
+    //
+    // Implementations SHOULD verify that there is an intersection between the
+    // hostname(s) attached to this Listener and the CN and SAN of each attached
+    // certificate. If there is no intersection, implementations SHOULD NOT
+    // attach the certificate and SHOULD set the ResolvedRefs condition to False
+    // with the "InvalidCertificateRef" reason. Implementations MAY choose to
+    // skip this validation to support certain use cases like self-signed
+    // certificates with missing or invalid CN or SANs. If implementations do
+    // not implement this validation, they MUST clearly document that.
+    //
+    // References to a resource in different namespace are invalid UNLESS there
+    // is a ReferencePolicy in the target namespace that allows the certificate
+    // to be attached. If a ReferencePolicy does not allow this reference, the
+    // "ResolvedRefs" condition MUST be set to False for this listener with the
+    // "InvalidCertificateRef" reason.
+    //
+    // This field is required to have at least one element when the mode is set
+    // to "Terminate" (default) and is optional otherwise.
+    //
+    // CertificateRefs can reference to standard Kubernetes resources, i.e.
+    // Secret, or implementation-specific custom resources.
+    //
+    // Support: Core (Kubernetes Secrets)
+    //
+    // Support: Implementation-specific (Other resource types)
+    //
+    // +optional
+	// +kubebuilder:validation:MaxItems=64
+    CertificateRefs []*SecretObjectReference `json:"certificateRefs,omitempty"`
+```
+
+## Self Service Model
+
+This change helps enable a self service model for TLS certificates. As an
+example, a Gateway may have a listener that does not specify a hostname. Two
+HTTP Routes may be attached, both with different hostnames. This change enables
+attaching different certificates for each of those hostnames to a Gateway
+Listener.
+
+Taking this a step further, it would be possible to build a controller that
+attached these certificates to the Gateway listener. This would enable a self
+service model where application owners do not need direct access to a Gateway to
+attach a certificate.
+
+## Alternatives
+
+N/A
--- a/site-src/geps/gep-851.md
+++ b/site-src/geps/gep-851.md
@@ -10,35 +10,33 @@ Listeners.
 
 ## Goals
 
-* Allow certs to be attached for multiple unique Hostnames for the same
-  Listener. This is most helpful in the following situations:
-  * The hostname on the Listener is empty or unspecified allowing Routes to
-    attach arbitrary hostnames to the Listener with Routes.
-  * The hostname on the Listener is a wildcard and Route owners want to attach
-    multi-level subdomains. For example, `*.example.com` on the Listener and
-    `app.dev.example.com` on a Route.
-* Enable specifying RSA and ECDSA certs on the same Listener.
-* Improve cert upgrades by allowing new and old certs to be attached to the same
-  Listener.
+Provide a path for implementations to support:
+
+* RSA and ECDSA certs on the same Listener.
+* Referencing certificates for different hostnames from the same Listener (maybe
+  as part of a self-service TLS approach).
+* Including new and old certificates as part of renewal process.
+* Certificate pinning: enable implementations that require certain certificates
+  for legacy clients while exposing a "normal" certificate to non-legacy
+  clients.
 
 ## Non-Goals
 
-* Be too prescriptive about how multiple certificates are handled for a Gateway
-  Listener. Each implementation may handle this slightly differently.
+* Define how implementations should support these features. Many implementations
+  have limited control as far as how certs are handled. Some implementations
+  just pass certs directly to the dataplane and rely on that implementation
+  specific behavior to determine which certs are used for a given request. That
+  makes it difficult to define any truly portable handling of multiple
+  certificates.
 
 ## Introduction
 
-When we removed certificate references from HTTPRoute ([GEP
-746](/geps/gep-746/)) we lost the capability to attach multiple certificates to
-a Listener. This GEP proposes adding that capability back, directly to the
-Gateway Listener this time.
-
-In addition to helping with the self service approach, where app owners may
-attach multiple hostnames (and certs) to a single Listener, this can help with
-other use cases:
-
-* Attaching RSA and ECDSA certs.
-* Attaching both new and old certs during an upgrade.
+As described above, there are a number of potential use cases for attaching
+multiple certificates to a Gateway Listener. The most straightforward reason for
+that involves attaching RSA and ECDSA certs. Although this is not a very common
+use case, it is a clearly understood and broadly supported example of why this
+change would be helpful. This change will enable implementations to support
+these more advanced use cases while still providing a portable core.
 
 ## API
 
@@ -51,17 +49,9 @@ following `CertificateRefs` field:
     // establish a TLS handshake for requests that match the hostname of the
     // associated listener.
     //
-    // When multiple TLS certificates are specified, SNI matching MUST be used
-    // to determine the appropriate certificate to use.
-    //
-    // Implementations SHOULD verify that there is an intersection between the
-    // hostname(s) attached to this Listener and the CN and SAN of each attached
-    // certificate. If there is no intersection, implementations SHOULD NOT
-    // attach the certificate and SHOULD set the ResolvedRefs condition to False
-    // with the "InvalidCertificateRef" reason. Implementations MAY choose to
-    // skip this validation to support certain use cases like self-signed
-    // certificates with missing or invalid CN or SANs. If implementations do
-    // not implement this validation, they MUST clearly document that.
+    // A single CertificateRef to a Kubernetes Secret has "Core" support.
+    // Implementations MAY choose to support attaching multiple certificates to
+    // a Listener, but this behavior is implementation-specific.
     //
     // References to a resource in different namespace are invalid UNLESS there
     // is a ReferencePolicy in the target namespace that allows the certificate
@@ -75,27 +65,14 @@ following `CertificateRefs` field:
     // CertificateRefs can reference to standard Kubernetes resources, i.e.
     // Secret, or implementation-specific custom resources.
     //
-    // Support: Core (Kubernetes Secrets)
+    // Support: Core - A single reference to a Kubernetes Secret
     //
-    // Support: Implementation-specific (Other resource types)
+    // Support: Implementation-specific (More than one reference or other resource types)
     //
     // +optional
-	// +kubebuilder:validation:MaxItems=64
+    // +kubebuilder:validation:MaxItems=64
     CertificateRefs []*SecretObjectReference `json:"certificateRefs,omitempty"`
 ```
-
-## Self Service Model
-
-This change helps enable a self service model for TLS certificates. As an
-example, a Gateway may have a listener that does not specify a hostname. Two
-HTTP Routes may be attached, both with different hostnames. This change enables
-attaching different certificates for each of those hostnames to a Gateway
-Listener.
-
-Taking this a step further, it would be possible to build a controller that
-attached these certificates to the Gateway listener. This would enable a self
-service model where application owners do not need direct access to a Gateway to
-attach a certificate.
 
 ## Alternatives
 


robscott:
Adding a hold until there's sufficient consensus on this GEP:

/hold

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,88 @@
...
+The `CertificateRef` field in `GatewayTLSConfig` would be replaced with the
+following `CertificateRefs` field:
+
+```go
+	// CertificateRefs contains a series of references to Kubernetes objects that
+	// contains TLS certificates and private keys. These certificates are used to
+	// establish a TLS handshake for requests that match the hostname of the
+	// associated listener.
+  //
+  // When multiple TLS certificates are specified, SNI matching MUST be used to
```
youngnick, line 50: So, the behavior here is that you can specify as many certs as you like, and they'll only be used if one of the cert's CN or SAN fields contains a match for a hostname in either the Listener or the Route?

Are implementations expected to check the CN and SAN of all certs and tell users if something doesn't match up? If they do, should the Gateway API present a standard way to indicate that in status/Conditions? I kind of feel like the best user experience will be if implementations do have to check, and tell you if there's zero valid matches between TLS, Listener Hostname, and Route hostname.

youngnick (commented):
I'm +1 on the idea, but I really want to make sure we consider what behavior is required and what behavior is left up to the implementation. I think we should mandate quite a bit, maybe all of the hostname matching behavior.

site-src/geps/gep-851.md
tokers: > Are implementations expected to check the CN and SAN of all certs and tell users if something doesn't match up? If they do, should the Gateway API present a standard way to indicate that in status/Conditions

I think it makes sense, we should check the CN and SAN with the hostname in either the Listener or the Route.

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,88 @@
...
+## TLDR
+
+Replace `CertificateRef` field with an `CertificateRefs` field in Gateway
+Listeners.
+
+## Goals
+
+* Allow certs to be attached for multiple unique Hostnames for the same
+  Listener. This is most helpful when the hostname is empty on the Listener and
+  attached Routes specify multiple hostnames.
```
hbagdi, line 15: TLS cert per hostname for the same port (listener collapsing) is another alternative.

The feature is helpful only in the case when you don't have a hostname present at all.
We should note that approach here in this GEP.

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,88 @@
...
+* Enable specifying RSA and ECDSA certs on the same Listener.
+* Improve cert upgrades by allowing new and old certs to be attached to the same
+  Listener.
```
hbagdi, line 18: These are the two compelling reasons for me.
How horrible would it be to limit to 2 certs per listener (per hostname when hostname specified in listener)?
We can expand the limit in future.

site-src/geps/gep-851.md
hbagdi: > Are implementations expected to check the CN and SAN of all certs and tell users if something doesn't match up? If they do, should the Gateway API present a standard way to indicate that in status/Conditions? I kind of feel like the best user experience will be if implementations do have to check, and tell you if there's zero valid matches between TLS, Listener Hostname, and Route hostname.

+1 with one exception:

Sometimes, users configure a self-signed certificate (with no or some random CN/SANs) for all of their listeners, really common with dev environments and some other times when TLS trust is not important. Another case is when you initially start with cert-manager as well I think (at least that used to be a thing). How do we handle for those cases?

To handle the above, we currently use listener hostname -> TLS cert to establish the relationship between the SNI and a TLS cert (and don't use TLS SNI in the handshake to look up the cert).

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,88 @@
...
+	CertificateRefs []*SecretObjectReference `json:"certificateRefs,omitempty"`
+```
+
+## Self Service Model
+
+This change helps enable a self service model for TLS certificates. As an
+example, a Gateway may have a listener that does not specify a hostname. Two
+HTTP Routes may be attached, both with different hostnames. This change enables
+attaching different certificates for each of those hostnames to a Gateway
+Listener.
```
hbagdi, line 79: Isn't it possible to enable the self-service model without this feature with listener collapsing and explicit per hostname listeners (with each of those specifying the TLS certificate)?

I'm a bit hesitant with this position because if the cluster-admin doesn't specify a hostname in the listener, they are leaving a wide hole in their infrastructure and trusting route owners with more than necessary for most organizations.
Listeners with an empty hostname are an escape hatch and are not meant for widespread usage. At least, that is my understanding.

robscott: Yeah, this is a good point, maybe my example should be broader. Let's say the hostname on the Listener is `*.example.com` and we want to allow self service attachment of `store.staging.example.com` and `app.staging.example.com`. I think that's a reasonable use case for attaching certs to multiple subdomains.

Even that could be accomplished with multiple listeners, but that may not be something an infra admin wants to care about, they simply want to delegate everything under example.com to this set of namespaces/routes.

site-src/geps/gep-851.md
robscott: I think it's also helpful when you are trying to support multiple layers of subdomains `*.example.net` on listener, `app.staging.example.net` on route, I can clarify that.

site-src/geps/gep-851.md
robscott: It seems like we all agree that we should have a way to communicate/warn when the CN and SAN do not overlap with any values that are allowed by the listener and attached routes. Based on @hbagdi's scenario above, it seems like we need to leave some flexibility as far as what implementations do when that happens. I think it's reasonable to require that implementations add a warning when these conflict, but I can't think of a good way to require implementations to actually prevent this while still allowing for Harry's use case.

robscott: I updated this to match what I said above, let me know what you think @hbagdi @tokers @youngnick.

site-src/geps/gep-851.md
robscott: Added a better example here. Agree that this is still possible by creating lots of listeners, but I'm not sure that's practical at a certain scale, especially with the limit on the number of listeners we have.

site-src/geps/gep-851.md
robscott: I personally think the other reason is also compelling, so would rather start with a higher limit if possible.

site-src/geps/gep-851.md
hbagdi: > Let's say the hostname on the Listener is *.example.com and we want to allow self service attachment of store.staging.example.com and app.staging.example.com. I think that's a reasonable use case for attaching certs to multiple subdomains.

In my understanding, a single `*` implies one DNS label.
I _think_ our documentation notes the same: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/shared_types.go#L252-L254

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
+# GEP-851: Allow Multiple Certificate Refs per Gateway Listener
+
+* Issue: [#851](https://github.com/kubernetes-sigs/gateway-api/issues/851)
+* Status: Implementable
+
+## TLDR
+
+Replace `CertificateRef` field with an `CertificateRefs` field in Gateway
```
hbagdi, line 8:
```suggestion
Replace `CertificateRef` field with a `CertificateRefs` field in Gateway
```

site-src/geps/gep-851.md
hbagdi: As per the current spec, the admin has two options:
- don't specify a hostname at all (I personally do not recommend this)
- specify hostnames: *.example.com *.staging.example.com - this is fairly common

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Goals
+
+* Allow certs to be attached for multiple unique Hostnames for the same
+  Listener. This is most helpful in the following situations:
+  * The hostname on the Listener is empty or unspecified allowing Routes to
+    attach arbitrary hostnames to the Listener with Routes.
+  * The hostname on the Listener is a wildcard and Route owners want to attach
+    multi-level subdomains. For example, `*.example.com` on the Listener and
+    `app.dev.example.com` on a Route.
```
hbagdi, line 19: Please see this thread: https://github.com/kubernetes-sigs/gateway-api/pull/856/files#r705706904.
I don't think multi-level subdomains are possible with a wildcard hostname in Listener.

site-src/geps/gep-851.md
robscott: As I understand it, [Envoy can't restrict wildcard matching to a single DNS label](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost). I know our docs were once more prescriptive in terms of a wildcard only matching a single DNS label, but I don't think they say that now. We should clarify this, but my interpretation of "prefixed with a single wildcard label" is that `*.example.com` is valid, but `*.*.example.com` is not. It does not actually limit what `*.example.com` should match.

/cc @howardjohn @jpeach @youngnick

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
...
+following `CertificateRefs` field:
+
+```go
+    // CertificateRefs contains a series of references to Kubernetes objects that
+    // contains TLS certificates and private keys. These certificates are used to
+    // establish a TLS handshake for requests that match the hostname of the
+    // associated listener.
+    //
+    // When multiple TLS certificates are specified, SNI matching MUST be used
+    // to determine the appropriate certificate to use.
```
hbagdi, line 55: Does the API provide any guidance on how to pick a certificate when multiple TLS certs can be used for a given TLS handshake context (context = SNI+cipher+othe TLS params)?

hbagdi (commented):
I'm on board with this.
We are trying to fit in the "self-service TLS model" in here but I find it only tangentially related and it adds more confusion than clarity. I believe that we have other (enough) compelling reasons to add this in and we should omit that as part of the motivation or goal and tackle the self-service problem separately (maybe no API change and only documentation).
This comment is in the spirit to add clarity to the GEP for future-selves.

site-src/geps/gep-851.md
robscott: I'm not sure if there's any guidance we can provide that could be universally implemented here. I think at least in some cases this behavior is going to be pretty locked into the implementation without a lot of room for customization.

hbagdi: In that case, we have two options:
1. no mention of this
2. mention "Such situations are possible and behaviors will be implementation-specific"

I don't have a preference. You pick one and consider this thread resolved.

robscott: I tried 2 first, but every combination of wording I could come up with felt more confusing to me than not saying anything. Open to any suggestions though.

site-src/geps/gep-851.md
robscott: Followed up in that thread.

site-src/geps/gep-851.md
robscott: Updated, thanks!

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
...
+
+* Allow certs to be attached for multiple unique Hostnames for the same
+  Listener. This is most helpful in the following situations:
+  * The hostname on the Listener is empty or unspecified allowing Routes to
+    attach arbitrary hostnames to the Listener with Routes.
+  * The hostname on the Listener is a wildcard and Route owners want to attach
+    multi-level subdomains. For example, `*.example.com` on the Listener and
+    `app.dev.example.com` on a Route.
+* Enable specifying RSA and ECDSA certs on the same Listener.
+* Improve cert upgrades by allowing new and old certs to be attached to the same
```
howardjohn, line 21: I don't think Envoy can support this (happy to be proven wrong though - not 100% confident): https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto.html?highlight=downstreamtlscontext#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-commontlscontext

' In server contexts, the first RSA certificate is used for clients that only support RSA and the first ECDSA certificate is used for clients that support ECDSA.'

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
...
+```go
+    // CertificateRefs contains a series of references to Kubernetes objects that
+    // contains TLS certificates and private keys. These certificates are used to
+    // establish a TLS handshake for requests that match the hostname of the
+    // associated listener.
+    //
+    // When multiple TLS certificates are specified, SNI matching MUST be used
+    // to determine the appropriate certificate to use.
+    //
+    // Implementations SHOULD verify that there is an intersection between the
```
howardjohn, line 57: This assumes the control plane can always introspect the certificate. Is that always the case? I can imagine a lot of cases where this would not be the case.

Istio used to be this way actually, although its not anymore.

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,102 @@
...
+The `CertificateRef` field in `GatewayTLSConfig` would be replaced with the
+following `CertificateRefs` field:
+
+```go
+    // CertificateRefs contains a series of references to Kubernetes objects that
+    // contains TLS certificates and private keys. These certificates are used to
+    // establish a TLS handshake for requests that match the hostname of the
+    // associated listener.
+    //
+    // When multiple TLS certificates are specified, SNI matching MUST be used
```
howardjohn, line 54: So for an Envoy implementation, is the expectation that we would take a listener like:
```
hostname: "*"
certs:
 - foo.com
 - bar.com
 - baz.com
```

is a syntactical sugar for
```
- hostname: foo.com
   certs:
   - foo.com
- hostname: bar.com
   certs:
   - bar.com
```

(although the route binding is slightly different with  `sectionName` match)

site-src/geps/gep-851.md
youngnick: We discussed this in the meeting today, but to recap my understanding:
- it's important to distinguish here between the Gateway API rules for Gateway-Route attachment, and the way that those rules will be implemented.
- The Gateway API rules currently state that a `*` character only ever matches a single DNS label. So if the Listener has `*.example.com`, only `foo.example.com`, not `subdomain.foo.example.com` would match _if it was specified in the Route_.
- In terms of Envoy's _implementation_ of this, if the Route specifies `*.example.com` as the hostname, then when Envoy implements that, Envoy will match more than one label under the `*`. (I actually think that we should consider not allowing wildcards on Routes because of the combinatorial number of use-cases it allows, but not strongly enough to make it anything other than an aside).

Adding this feature means that there are effectively three places that a hostname needs to be matched (in the HTTPS case):
- SNI hostname via whatever is in all the certs. This means that most implementations will probably need separate routing chains per cert (Envoy certainly does).
- Listener Hostname, as we do now
- Route Hostname, as we do now.

For each hostname match, we allow either a precise or a wildcard match, or the Listener and Route hostnames can be unspecified. So, our number of combinations is 2 x 3 x 3 = 18 combinations that we need to define rules on, for when things line up. If hostnames are precise but don't all match, then that's a separate degree of freedom. What I'm saying by this paragraph is that, in adding this feature, we have to carefully define the rules for what supersedes what, and what is valid.

To choose some examples out of the space:
- For a cert that handles `*.example.com`, a Listener hostname of `foo.example.com`, and a Route of `*.example.com`, everything will attach, but the actual routing config must necessarily be limited to `foo.example.com`.
- For a cert that handles `foo.example.com`, a Listener hostname of `*.example.com`, and a Route of `bar.example.com`, obviously, no match.
- For a cert that handles `foo.example.com`, a LIstener hostname of `*.example.com`, and a Route of `foo.example.com`, we have a match, and the routing chain must also be `foo.example.com`.
- The case we talked about today is when there is a cert that handles `*.example.com`, and the Listener or Route hostnames are either unset or also `*.example.com`, then the eventual routing chain that's created will do SNI for *.example.com, and then Envoy will route host headers for `*.example.com`, which could conceivably include multiple levels of routing. However, without having the SNI and Host header being different, this won't be a problem - this is an example of why we should ensure that the SNI and Host header must match.

Maybe the shortest way to sy what I think the rules should be are:
- wildcard matches are only ever one label in the Gateway API.
- If wildcard matches and precise matches are being combined, the precise one is the result (as long as they match).
- In general you should try to keep wildcard matches at the Gateway level and leave Route hostname matches for precise matches.

site-src/geps/gep-851.md
youngnick: I think we need to define the hostname matching rules like we do with the Listener/Route ones, something like:

"If there is an overlap between certificate SNI hostnames, then any precise matches take precedence. When two precise matches are relevant, the longest hostname wins. For multiple matches outside of the cases, the implementation decides".

site-src/geps/gep-851.md
youngnick: I think that's correct in that case, but I also think that we'll need to be pretty specific with rules to cover all the possible cases.

site-src/geps/gep-851.md
robscott: To clarify does this mean that Istio can no longer derive details from certs?

site-src/geps/gep-851.md
robscott: Do you know what would happen if one of the attached certs was expired? Do you know if it has any way to prioritize certs that haven't expired?

site-src/geps/gep-851.md
youngnick: This basically demands that every control plane unwrap certificates referenced here and check the CN and SANs. It *should* be possible for everything to pull the certs from Kuberrnetes themselves (since you need to have read access to them to be able to watch them anyway), unwrap them enough to check the CN and SAN, and then throw the other info away.

For Contour, this is extra functionality we'll need to add, but it is doable.

site-src/geps/gep-851.md
robscott: Thanks for taking the time to write this up @youngnick! This is a really great summary of the levels of complexity this change would add. As I've been reading through this and other comments here, I'm wondering if it makes sense to limit the scope of this change to support for RSA and EC certs. Although much smaller than the original intent, this accomplishes several important things:

- Supports attaching both RSA and EC certs to the same listener
- Transitions CertRefs to a list, allowing future expansion of this field simply by loosening validation

The vast majority of the complexity in this discussion is based on unspecified future work to enable self service certificate attachment. It would likely make more sense to wait until we were ready to fully tackle that before we determine how this particular use of multiple certs should be handled.

site-src/geps/gep-851.md
robscott: What if we start with a simpler version of this that doesn't require unwrapping? Up to 2 certificates may be attached to a listener, 1 RSA and 1 EC. I don't want to give up on the more complex self-service use cases here, but we may not need to solve them in this first v1a2 release.

site-src/geps/gep-851.md
jpeach: I read this as meaning that the ability to provide more than one certificate would allow you to provision different certificates for the two names. We can just make that explicit in the text though.

site-src/geps/gep-851.md
jpeach: Even if Envoy doesn't support that directly, a control plane can still make a choice about which certificate to send to Envoy (e.g. the one with the longest validity time remaining).

site-src/geps/gep-851.md
howardjohn: Istio can implement it today, for kind=secret. Previously though the data plane read the Secret directly.

For other kinds (Especially something like a file reference mounting an HSM or similar) the control plane wouldn't have any clue. Or CSI driver, etc etc.

Overall I think it will come back to bite us if we require control plane can see the cert.

For dual rsa+ecdsa maybe it's fine, although it's not clear to me how common that use case is to introduce this complexity

jpeach: If a control plane can't or won't examine the certificates, this is a SHOULD, so that implementation would sill be conforming.

This SHOULD aligns with [RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2.2).

Personally, I think that the less we say in the standard about how this works, the better. There's a lot of divergence across proxy implementations and TLS changes can move very rapidly. It's good to keep the implementation flexibility as long as users have this general guidance about what to put in a secret.

howardjohn: If I don't inspect the contents then how would we know which one to present to clients? It seems like we would need to just pick a random one (or make Envoy smarter to do the selection dynamically?)

jpeach: Sorry, what I meant was that some theoretical control plane (not you specifically) might not examine it. Maybe it would make sense for it do dump them all down to the data plane and let that figure it out.

robscott:
Thanks to everyone for the great feedback on this PR! This has really highlighted how complex it would be to support certs for different hostnames on a single listener. As I've thought about this more, I think the most important thing is to change `certificateRef` to `certificateRefs`. Although it would be great to figure out the details of how multiple certs of differing hostnames are handled, I don't think that's required to make this change. 

Although not a common use case, providing RSA and ECDSA certs is a valid one. It is also clearly understood and widely supported. I updated the GEP to start with only that goal. It does seem likely that we'll want to expand this in the future to support more broad use cases, but we don't need to figure out all the details for those to get this in.

I think my latest changes have made this GEP significantly simpler while still allowing all the same room for future expansion.

/cc @bowei @hbagdi @howardjohn @jpeach @youngnick

site-src/geps/gep-851.md
tokers: Would it be too strict to ask the control plane to unwrap the certificates and check the CN and SAN, what if people just want to use a certificate with different SAN (as a placeholder) on the staging env? Maybe it should be a preference but not a striction.

robscott: The current wording of "Should" denotes a recommendation and not a requirement (https://datatracker.ietf.org/doc/html/rfc2119#section-3), which I think is what you're suggesting.

youngnick:
I like this simpler approach.

I'll leave the hold until some US folk have had a chance to review.

/lgtm

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,80 @@
...
+Listeners.
+
+## Goals
+
+* Enable specifying RSA and ECDSA certs on the same Listener.
+* Provide a path to future expansion that could include:
+  * Support for referencing certificates for different hostnames (maybe as part
+    of a self-service TLS approach).
+  * Support for including new and old certificates as part of renewal process.
+
```
bowei, line 18: * Certificate pining: enable implementations that require certain certificates for legacy clients while exposing a "normal" certificate to non-legacy clients.

site-src/geps/gep-851.md
```diff
@@ -0,0 +1,80 @@
...
+The `CertificateRef` field in `GatewayTLSConfig` would be replaced with the
+following `CertificateRefs` field:
+
+```go
+    // CertificateRefs contains a series of references to Kubernetes objects that
+    // contains TLS certificates and private keys. These certificates are used to
+    // establish a TLS handshake for requests that match the hostname of the
+    // associated listener.
+    //
+    // A maximum of 2 certificates may be referenced from this field - one RSA
```
bowei, line 46: I don't think we should be this prescriptive. This is going to cause problems when moving to having more than one certificate of differing types etc.

I think we should just leave it not specified that this requires a certain kind of cert etc.

site-src/geps/gep-851.md
bowei: We may want to turn SHOULD to MAY

bowei:
Rob -- I think we should not be quite so prescriptive. It will make future expansion more confusing for implementations.

hbagdi:
I prefer this iteration of the GEP over the previous one - it is more concise and has more concrete reasoning behind it.
A tangential question (out of scope from this PR but still related): how are SNI->Certificates tied? Based on Listener.Hostname->CertificatesRefs? I know we discussed this in the community meeting, I'm confused so asking again.

site-src/geps/gep-851.md
robscott: Added, thanks!

site-src/geps/gep-851.md
robscott: 👍 removed.

robscott:
@bowei @hbagdi @jpeach @youngnick I've updated this PR based on our [discussion in Slack](https://kubernetes.slack.com/archives/CR0H13KGA/p1631733176077900). PTAL.

howardjohn (approved):

youngnick:
/lgtm

/hold cancel
because @hbagdi +1ed in Slack.

PR #859, opened 2021-09-10
Clarifying how HTTP and TLS Route status should be populated

### LINKED ISSUE #849: Add top level Gateway conditions when Routes fail to attach
Top level Gateway conditions that can be used to indicate when a Route has not been attached to any listeners within the Gateway.

**Why this is needed**:
The alternative of adding this condition to each listener is rather unwieldy. 

This idea came from a [comment thread](https://github.com/kubernetes-sigs/gateway-api/pull/839#discussion_r700722524) with @howardjohn and @youngnick.

robscott:
**What type of PR is this?**
/kind bug

This fixes the confusing guidance (originally added by me) in HTTP and TLS Routes that suggested that conditions should be populated on Gateway Listeners. I had originally built out conditions that could be populated on Gateway and Listeners, but I'm not sure that's the right approach. I think the simplest and most straightforward answer here is to populate Route status. If Route owners can't attach to a Gateway they want to, they can simply contact the Gateway owner.

Fixes #849

```release-note
Updated guidance on how HTTP and TLS Route status should be populated when hostnames do not match.
```

/cc @youngnick @howardjohn

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -83,9 +83,10 @@ type HTTPRouteSpec struct {
 	// HTTPRoute specified `test.example.com` and `test.example.net`,
 	// `test.example.net` must not be considered for a match.
 	//
-	// If hostnames do not match with the criteria above, then the HTTPRoute is
-	// not accepted, and the implementation must raise an 'Accepted' Condition
-	// with a status of `False` for the target Listener(s).
+	// If both the Listener and HTTPRoute have specified hostnames, and none
+	// match with the criteria above, then the HTTPRoute is not accepted. The
+	// implementation must raise an 'Accepted' Condition with a status of
+	// `False` in the corresponding RouteParentStatus.
 	//
 	// Support: Core
 	//
--- a/apis/v1alpha2/tlsroute_types.go
+++ b/apis/v1alpha2/tlsroute_types.go
@@ -75,9 +75,10 @@ type TLSRouteSpec struct {
 	// TLSRoute specified `test.example.com` and `test.example.net`,
 	// `test.example.net` must not be considered for a match.
 	//
-	// If all hostnames do not match with the criteria above, then the TLSRoute
-	// is not accepted, and the implementation must raise an 'Accepted'
-	// Condition with a status of `False` for the target Listener(s).
+	// If both the Listener and TLSRoute have specified hostnames, and none
+	// match with the criteria above, then the TLSRoute is not accepted. The
+	// implementation must raise an 'Accepted' Condition with a status of
+	// `False` in the corresponding RouteParentStatus.
 	//
 	// Support: Core
 	//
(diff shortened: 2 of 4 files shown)


howardjohn (approved):

apis/v1alpha2/httproute_types.go
```diff
@@ -83,9 +83,10 @@ type HTTPRouteSpec struct {
 	// HTTPRoute specified `test.example.com` and `test.example.net`,
 	// `test.example.net` must not be considered for a match.
 	//
-	// If hostnames do not match with the criteria above, then the HTTPRoute is
-	// not accepted, and the implementation must raise an 'Accepted' Condition
-	// with a status of `False` for the target Listener(s).
+	// If hostnames are specified, and none match with the criteria above, then
+	// the HTTPRoute is not accepted, and the implementation must raise an
+	// 'Accepted' Condition with a status of `False` in the corresponding
+	// RouteParentStatus.
```
hbagdi, line 89: What if the route specifies two hostnames - one satisfied the above criteria but the other one doesn't?

hbagdi: > must be at least one intersecting hostname for the HTTPRoute to be

Maybe this line clarifies that?

apis/v1alpha2/httproute_types.go
```diff
@@ -83,9 +83,10 @@ type HTTPRouteSpec struct {
 	// HTTPRoute specified `test.example.com` and `test.example.net`,
 	// `test.example.net` must not be considered for a match.
 	//
-	// If hostnames do not match with the criteria above, then the HTTPRoute is
-	// not accepted, and the implementation must raise an 'Accepted' Condition
-	// with a status of `False` for the target Listener(s).
+	// If hostnames are specified, and none match with the criteria above, then
```
hbagdi, line 86:
```suggestion
	// When hostnames specified in Route and none match with the criteria above, then
```

robscott: There's a bit of nuance here, it's really only if hostnames are specified on both the Gateway and Route. If they are absent in either place, everything should match. I'll work on clarifying that.

hbagdi: Yes, that's what I was trying to capture as well but probably didn't do a good job either.
I was trying to imply gateway's listener hostname with the "criteria above" part of the sentence but that feels indirect as well.

apis/v1alpha2/httproute_types.go
robscott: Yep, that's what I was going for. I also hope that "and none match the criteria" helps convey that only one would need to match.

apis/v1alpha2/httproute_types.go
robscott: Thanks for catching this @hbagdi! I've updated the PR to make this more clear, PTAL.

PR #869, opened 2021-09-15
Align path prefix matching with Ingress.

### LINKED ISSUE #866: Define whether path prefix matches are element-wise or byte-wise
Gateway defines a [prefix match](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.PathMatchType)  like this:

> Prefix and Exact paths must be syntactically valid:
> * Must begin with the ‘/’ character
> * Must not contain consecutive ‘/’ characters (e.g. /foo///, //).
> * For prefix paths, a trailing ‘/’ character in the Path is ignored, e.g. /abc and /abc/ specify the same match.

But Ingress V1 [says](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types):

> Prefix: Matches based on a URL path prefix split by /. Matching is case sensitive and done on a path element by element basis. A path element refers to the list of labels in the path split by the / separator. A request is a match for path p if every p is an element-wise prefix of p of the request path.

I don't recall a discussion about this, but we should discuss whether to align the Gateway prefix semantics with Ingress.

jpeach:
Explicitly align path prefix matching with the Ingress API by specifying
it as an prefix of path elements, not a prefix of bytes.

**What type of PR is this?**
/kind bug
/kind api-change

This fixes #866.

```release-note
Aligned path prefix matching with Ingress by clarifying that it is a prefix of path elements.
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -214,18 +214,33 @@ type HTTPRouteRule struct {
 //
 // Prefix and Exact paths must be syntactically valid:
 //
-// - Must begin with the '/' character
-// - Must not contain consecutive '/' characters (e.g. /foo///, //).
-// - For prefix paths, a trailing '/' character in the Path is ignored,
-// e.g. /abc and /abc/ specify the same match.
+// - Must begin with the `/` character
+// - Must not contain consecutive `/` characters (e.g. `/foo///`, `//`).
 //
 // +kubebuilder:validation:Enum=Exact;Prefix;RegularExpression
 type PathMatchType string
 
-// PathMatchType constants.
 const (
-	PathMatchExact             PathMatchType = "Exact"
-	PathMatchPrefix            PathMatchType = "Prefix"
+	// Matches based on a URL path prefix split by `/`. Matching is
+	// case sensitive and done on a path element by element basis. A
+	// path element refers to the list of labels in the path split by
+	// the `/` separator. A request is a match for path _p_ if every
+	// _p_ is an element-wise prefix of the request path.
+	//
+	// For example, `/abc`, `/abc/` and `/abc/def` match the prefix
+	// `/abc`, but `/abcd` does not.
+	PathMatchExact PathMatchType = "Exact"
+
+	// Matches the URL path exactly and with case sensitivity.
+	PathMatchPrefix PathMatchType = "Prefix"
+
+	// Matches if the URL path matches the given regular expression with
+	// case sensitivity.
+	//
+	// Since `"RegularExpression"` has custom conformance, implementations
+	// can support POSIX, PCRE, RE2 or any other regular expression dialect.
+	// Please read the implementation's documentation to determine the supported
+	// dialect.
 	PathMatchRegularExpression PathMatchType = "RegularExpression"
 )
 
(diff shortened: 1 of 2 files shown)


howardjohn:
Bonus points if someone adds envoy-native support so we don't have to convert it to a regex 🙂

jpeach:
> Bonus points if someone adds envoy-native support so we don't have to convert it to a regex (performance+ simplicity - not a huge concern)

FWIW, I was planning to expand this into two routes with an exact and prefix match. Maybe that won't work out :)

howardjohn:
> > Bonus points if someone adds envoy-native support so we don't have to convert it to a regex (performance+ simplicity - not a huge concern)
> 
> FWIW, I was planning to expand this into two routes with an exact and prefix match. Maybe that won't work out :)

I think the options are to use regex (runtime cost) or 2 routes (control plane cost, its not just a list of 2 strings but the whole route duplicated I believe). So both aren't ideal... but certainly shouldn't block the API here.

hbagdi:
/lgtm
Created https://github.com/kubernetes-sigs/gateway-api/issues/873 to track further validation.

apis/v1alpha2/httproute_types.go
```diff
@@ -214,18 +214,33 @@ type HTTPRouteRule struct {
...
 
-// PathMatchType constants.
 const (
-	PathMatchExact             PathMatchType = "Exact"
-	PathMatchPrefix            PathMatchType = "Prefix"
+	// Matches based on a URL path prefix split by `/`. Matching is
+	// case sensitive and done on a path element by element basis. A
+	// path element refers to the list of labels in the path split by
+	// the `/` separator. A request is a match for path _p_ if every
+	// _p_ is an element-wise prefix of _p_ of the request path.
```
Miciah, line 228: The last sentence is difficult to parse.  Did you mean to write something like the following?
```suggestion
	// the `/` separator. A request is a match for path _p_ if every
	// element of _p_ matches the element at the corresponding place
	// in the request path.
```
or simply the following?
```suggestion
	// the `/` separator. A request is a match for path _p_ if _p_
	// is an element-wise prefix of the request path.
```

apis/v1alpha2/httproute_types.go
```diff
@@ -214,18 +214,33 @@ type HTTPRouteRule struct {
...
+	//
+	// For example, `/abc` and `/abc/` match the prefic `/abc`, but `/abcd`
```
Miciah, line 230:
```suggestion
	// For example, `/abc` and `/abc/` match the prefix `/abc`, but `/abcd`
```

apis/v1alpha2/httproute_types.go
jpeach: Nice catch. The latter was the intention :)

youngnick:
@jpeach, tide/prow won't let us merge things that have "fixes #something" in the commit message - it has to only be in the PR text.

Aside from that,
/lgtm

Once you've changed the commit message, feel free to cancel the hold.

apis/v1alpha2/httproute_types.go
```diff
@@ -214,18 +214,33 @@ type HTTPRouteRule struct {
...
+	// path element refers to the list of labels in the path split by
+	// the `/` separator. A request is a match for path _p_ if every
+	// _p_ is an element-wise prefix of the request path.
+	//
+	// For example, `/abc`, `/abc/` and `/abc/def` match the prefix
+	// `/abc`, but `/abcd` does not.
+	PathMatchExact PathMatchType = "Exact"
+
+	// Matches the URL path exactly and with case sensitivity.
+	PathMatchPrefix PathMatchType = "Prefix"
```
skriss, line 235: @jpeach :wave: was just looking at this and the godoc for these two (`Exact` and `Prefix`) are switched.

PR #871, opened 2021-09-17
Add validation for Kubernetes object name fields and ProtocolType

### LINKED ISSUE #867: All required values should have a noted behavior for the zero value

As discussed in the community meeting on the 13th of September, 2021, we should make it a rule for this API that all required values must have a definition of what the zero value does. (It could be "cause the config to be rejected").

This means that, should any required field need to change to optional in the future, all controllers that implement the spec will be able to cope with the zero value introduced by making the field optional. Plus, it seems like a good idea to have the zero value have a definition, to prevent it being overloaded in different ways by different implementations.

If we agree on this, I'll get started on the PR.

youngnick:
**What type of PR is this?**

/kind cleanup

This PR ensures that there are no required fields that allow empty values, and that all required fields have consistent validation.

It does this by introducing a new string type alias, ObjectName, for Kubernetes object names, and by ensuring that the Listener ProtocolType has validation.

On the plus side, that was all the required types that could be zero I could find. The Port number field already has validation to ensure it's between 1 and 65535.

Fixes #867 

```release-note
Validation for Kubernetes object names has been updated
Listener ProtocolType now has validation
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -245,9 +245,19 @@ type Listener struct {
 // protocols. Any protocol defined by implementations will fall under custom
 // conformance.
 //
-// +kubebuilder:validation:MinLength:=1
-// +kubebuilder:validation:MinLength:=255
-// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9]([A-Z-a-z0-9\/]*[A-Za-z0-9])?(\.[A-Za-z0-9]([A-Z-a-z0-9\/]*[A-Za-z0-9])?)*$`
+// Valid values include:
+//
+// * "HTTP" - Core support
+// * "example.com/bar" - Implementation-specific support
+//
+// Invalid values include:
+//
+// * "example.com" - must include path if domain is used
+// * "foo.example.com" - must include path if domain is used
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=255
+// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([-a-zSA-Z0-9]*[a-zA-Z0-9])?$|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9]+$`
 type ProtocolType string
 
 const (
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -63,10 +63,7 @@ type GatewayList struct {
 type GatewaySpec struct {
 	// GatewayClassName used for this Gateway. This is the name of a
 	// GatewayClass resource.
-	//
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	GatewayClassName string `json:"gatewayClassName"`
+	GatewayClassName ObjectName `json:"gatewayClassName"`
 
 	// Listeners associated with this Gateway. Listeners define
 	// logical endpoints that are bound on this Gateway's addresses.
@@ -247,6 +244,10 @@ type Listener struct {
 // `mycompany.com/my-custom-protocol`. Un-prefixed names are reserved for core
 // protocols. Any protocol defined by implementations will fall under custom
 // conformance.
+//
+// +kubebuilder:validation:MinLength:=1
+// +kubebuilder:validation:MinLength:=255
+// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9]([A-Z-a-z0-9\/]*[A-Za-z0-9])?(\.[A-Za-z0-9]([A-Z-a-z0-9\/]*[A-Za-z0-9])?)*$`
 type ProtocolType string
 
 const (
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -53,10 +53,7 @@ type ParentRef struct {
 	// Name is the name of the referent.
 	//
 	// Support: Core
-	//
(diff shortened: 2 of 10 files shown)


howardjohn (commented):
Sorry if my regex knowledge is wrong here but I think this breaks the examples for ExternalService types where we have things like google.com?

youngnick:
I have assumed that the ObjectReferences here must be to Kubernetes objects. I can see that it would be handy to be able to fudge a direct reference to a non-Kubernetes thing, but I guess I had assumed that in that case there would be some Kubernetes object to represent the external object first.

If we are not going to require that references are to Kubernetes objects, then we should record that fact alongside the all the object name references.

apis/v1alpha2/gateway_types.go
```diff
@@ -247,6 +244,10 @@ type Listener struct {
 // `mycompany.com/my-custom-protocol`. Un-prefixed names are reserved for core
 // protocols. Any protocol defined by implementations will fall under custom
 // conformance.
+//
+// +kubebuilder:validation:MinLength:=1
+// +kubebuilder:validation:MinLength:=255
```
hbagdi, line 249:
```suggestion
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MinLength=255
```

hbagdi:
+1 to both points by Nick.

youngnick:
Dang, I should have talked about this in the meeting today. I'll start a Slack thread so we can go async.

jpeach:
> I have assumed that the ObjectReferences here must be to Kubernetes objects.

Yeh, I had the same assumption.

youngnick:
Okay, I'll leave this until we can come to some agreement about the project's stance on if the references must be to Kubernetes objects only, https://kubernetes.slack.com/archives/CR0H13KGA/p1632187311091400 is the Slack thread to discuss.

robscott:
The Slack thread referenced by @youngnick above has gotten fairly long. Here's a summary of what I think the current state is for each of the two relevant questions.

**1. Should we validate Name in ObjectRefs?**
This is nuanced, and I don't think there's consensus here yet. My personal opinion is that we should not. Each Kubernetes resource can define its own name validation (example: [EndpointSlice](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/discovery/validation/validation.go#L54-L58), [others](https://cs.k8s.io/?q=ValidateObjectMeta%5C(&i=nope&files=validation.go&excludeFiles=&repos=kubernetes/kubernetes)). Although the validation proposed here currently represents the least restrictive validation I'm aware of among current k8s objects, there is no guarantee that will be true in the future. The corresponding [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) highlight the validation described here as the most common name validations, but do not provide any guarantees that resources will always fall into these buckets. The lack of upstream validation for ObjectRef names makes me think we should avoid this.

**2. Should we be open to using ObjectRefs to reference non-k8s objects?**
There are significant concerns raised in the Slack thread about using ObjectRefs to refer to non-k8s objects. @hbagdi summarizes it well with the following:

>This API is complex enough making extension points more loosely defined and arbitrary could have effects that are hard for us to foresee today.

Although I'm personally open to the idea of referencing virtual or non-k8s resources, I don't feel as strongly about it. It seems mostly reasonable to start without this concept with the idea that we can add it later if necessary with a simple addition to docs and no change in validation.

youngnick:
Okay, finishing up for the day but tomorrow after the meeting I will update this and remove the object name validation, and clarify that objects must be Kubernetes resources.

youngnick:
In the latest commit, I:
- removed the validation on the ObjectName field
- Updated docs on all the reference fields with "must reference a valid Kubernetes object" language
- Added examples for ProtocolType
- Updated the regex for ProtocolType to match the given examples

youngnick:
Well, the system works, turned out there were a couple of regex problems, which the tests picked up! Fixed now.

apis/v1alpha2/shared_types.go
```diff
@@ -298,6 +302,14 @@ type Group string
 // +kubebuilder:validation:Pattern=`^[a-zA-Z]([-a-zA-Z0-9]*[a-zA-Z0-9])?$`
 type Kind string
 
+// ObjectName refers to the name of a Kubernetes object.
+// Object names can have a variety of forms, including RFC1123 subdomains,
+// RFC 1123 labels, or RFC 1035 labels.
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type ObjectName string
```
robscott, line 311: I [followed up with sig-api-machinery](https://groups.google.com/g/kubernetes-sig-api-machinery/c/V-qnyODuMt4) and it sounds like there is a [minimum set of validation we can provide here](https://github.com/kubernetes/kubernetes/blob/b290671b1b28ab4af8164e2b69b334fab584558e/staging/src/k8s.io/apimachinery/pkg/api/validation/path/name.go#L30-L46). It may be a bit difficult to translate into regex validation though, and others recommended against adding validation here. Although that minimum set of validation does not include a max length, 253 seems like a reasonable value to me based on the most common validation + ensure we don't end up with an unbounded list.

robscott (commented):
Thanks for the work on this @youngnick! I personally don't think we _need_ regex validation for ObjectName, but also would be fine if it was added. I'm leaving a hold on here in case you want to add something based on the [sig-api-machinery thread](https://groups.google.com/g/kubernetes-sig-api-machinery/c/V-qnyODuMt4).

/lgtm
/hold

youngnick:
I don't think that doing that level of checking is worthwhile when the apiserver will stop you creating objects with that name anyway; I think it's more important to get the "must refer to an existing type" clauses in.

/hold cancel

robscott:
Thanks!

/lgtm

PR #875, opened 2021-09-22
Fix HTTP path match documentation.

jpeach:
The documentation for the prefix and exact path match types was attached
to the wrong constants.

Thanks @skriss!

**What type of PR is this?**
/kind bug
/kind documentation

Documentation mixup from #869.


youngnick:
/lgtm

Nice catch @skriss, can't believe we all missed it!

PR #879, opened 2021-09-24
Clarifying that HTTP listeners may be used for Cleartext HTTP/2

### LINKED ISSUE #817: Cleartext HTTP/2 (h2c) should be allowed on HTTP listeners
>         // Accepts cleartext HTTP/1.1 sessions over TCP.    
>        HTTPProtocolType ProtocolType = "HTTP" 

Though this isn't couched in spec language (MUST, SHOULD, etc), this argues against allowing implementations to support HTTP/2 over cleartext. Implementations should be allowed to offer h2c when the "HTTP" protocol is specified.

robscott:
**What type of PR is this?**
/kind feature

Clarifies that HTTP listeners may be used for Cleartext HTTP/2.

Fixes #817

```release-note
HTTP listeners may now be used for Cleartext HTTP/2
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -250,7 +250,10 @@ type Listener struct {
 type ProtocolType string
 
 const (
-	// Accepts cleartext HTTP/1.1 sessions over TCP.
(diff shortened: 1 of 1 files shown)


youngnick:
/lgtm

I don't think this one's contentious enough to need a hold.

PR #882, opened 2021-09-27
Removing scope fields from object references

### LINKED ISSUE #838: Scope in object references may be unnecessary
As a follow up from @howardjohn's v1alpha2 feedback in #790: 

>ParentRef.Scope is redundant. I know sig-apimachinery suggested it, but I disagree. Its inconsistent with Kubernetes - I don't run `kubectl get clusterrole --scope Cluster`, I run `kubectl get clusterrole` and the api-server knows it is cluster scoped. Why make users specify this boilerplate? It also gives users another way to create invalid config (ie Scope=Cluster Kind=Gateway); a great API makes it hard to create invalid configs.

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

Our object references use kind instead of resource which requires implementations to have a predefined kind->resource mapping. That means there can never be ambiguity within an implementation as far as whether a reference is to a cluster-scoped or namespace-scoped resource.

This also removes the ability for ParentRef to refer to cluster-scoped resources. Although we can add that in the future with a docs update. There do not appear to be any use cases for that today.

Fixes #838 

```release-note
* The `scope` field has been removed from all object references.
* ParentRefs can no longer refer to cluster-scoped resources.
```


howardjohn (approved):
:tada:

robscott:
Discussed at community meeting yesterday as well, it seems like there's pretty broad consensus on this change.

/hold cancel

PR #886, opened 2021-09-29
Adding TLS options validation as final sig-network review follow up

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

I found a comment from @khenidak that I'd missed from the original sig-network review suggesting more thorough validation here. This adds generic "AnnotationKey" and "AnnotationValue" types in hopes that they'll be useful for more than just TLS options in the future.

```release-note
* TLS options keys are now subject to the same validation as Kubernetes annotations.
* TLS options values now have a max length of 1024 characters.
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -352,3 +352,34 @@ type SectionName string
 // +kubebuilder:validation:MaxLength=253
 // +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type GatewayController string
+
+// AnnotationKey is the key of an annotation in Gateway API. This is used for
+// validation of maps such as TLS options. This matches the Kubernetes
+// "qualified name" validation that is used for annotations and other common
+// values.
+//
+// Valid values include:
+//
+// * example
+// * example.com
+// * example.com/path
+// * example.com/path.html
+//
+// Invalid values include:
+//
+// * example~ - "~" is an invalid character
+// * example.com. - can not start or end with "."
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:Pattern=`^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]/?)*$`
+type AnnotationKey string
+
+// AnnotationValue is the value of an annotation in Gateway API. This is used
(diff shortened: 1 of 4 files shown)


howardjohn (approved):
lgtm. Note your release note has "TCP" instead of "TLS"

apis/v1alpha2/shared_types.go
```diff
@@ -352,3 +352,34 @@ type SectionName string
...
+// +kubebuilder:validation:Pattern=`^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]/?)*$`
+type AnnotationKey string
+
+// AnnotationKey is the key of an annotation in Gateway API. This is used for
+// validation of maps such as TLS options. This roughly matches Kubernetes
+// annotation validation, although the length validation in that case is based
+// on the entire size of the annotations struct.
+//
+// +kubebuilder:validation:MinLength=0
+// +kubebuilder:validation:MaxLength=1024
```
jpeach, line 384: That probably prevents keys that are JSON or PEM objects, which seems quire useful for TLS config.

robscott: My first thought is that those may be better served by a reference to a separate secret or config map depending on the use case. With that said, we could increase the limit here to something like 16k if that helps unblock some specific use cases.

For reference, [upstream validation](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validation/objectmeta.go#L36) ensures that the entire size of the annotations struct is less than 256kB. It's difficult to translate that perfectly, but since we currently allow for a max of 100

apis/v1alpha2/shared_types.go
```diff
@@ -352,3 +352,34 @@ type SectionName string
...
+//
+// * example~ - "~" is an invalid character
+// * example.com. - can not start or end with "."
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:Pattern=`^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]/?)*$`
+type AnnotationKey string
+
+// AnnotationKey is the key of an annotation in Gateway API. This is used for
```
youngnick, line 378:
```suggestion
// AnnotationValue is the value of an annotation in Gateway API. This is used for
```

youngnick:
Great change, just needs the docstring for `AnnotationValue` updated, then LGTM.

apis/v1alpha2/shared_types.go
robscott: Ended up bumping this up to 4096, hoping that's sufficient for now. That matches our longest length validation and still leaves us room to expand if necessary.

apis/v1alpha2/shared_types.go

youngnick:
/lgtm

/hold
for @jpeach

PR #898, opened 2021-10-08
Final changes for RC2 based on sig-net review feedback

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

There are three changes here, all based on feedback from @thockin in #861:

1. The "Prefix" path match type has been renamed "PathPrefix" to leave room for a future "StringPrefix" match.
2. The "ClassName" field in PolicyTargetReference has been removed as it is not clearly necessary at this point.
3. A new optional "Name" field has been added to ReferencePolicyTo, enabling more fine grained control over which resource(s) may be referenced.

More thoughts on all of the above:

1. I think it could be useful to add StringPrefix matching in the future. Currently Envoy based implementations of this API need to use RegEx matching to support our PathPrefix matching. Although this is fine as a starting point, I can imagine at least some users of the API would prefer to have access to the StringPrefix matching they're familiar with for the implementations that support it. This doesn't commit us to providing StringPrefix as a match type, but does make that simpler if we want to.
2. I was the one who originally added "ClassName", but after looking through the docs, GEP, and asking around, it does not seem like this field was clearly needed. It _may_ be helpful in certain use cases, but I don't think it's necessary at this point. 
3. The idea of adding `to.name` in ReferencePolicy has come up _many_ times already, and even reappeared when I presented this concept to sig-auth. Although it's true that we could add this at a later point, it would be messier. Users would need to know whether the implementation they were using was new enough to support that particular field of ReferencePolicy, and if it wasn't you may be unknowingly allowing more references than intended. Although I've generally pushed to keep ReferencePolicy as simple as possible to start, this feels like an inevitable addition and it is significantly simpler to add this up front.

```release-note
* The "Prefix" path match type has been renamed "PathPrefix".
* The "ClassName" field in PolicyTargetReference has been removed.
* A new optional "Name" field has been added to ReferencePolicyTo.
```

/hold for consensus
/cc @bowei @hbagdi @jpeach @youngnick


youngnick:
I agree with all three of these changes, I think that the call on ReferencePolicy `name` is particularly good.

/lgtm

/hold
in case anyone else wants to go.

I'll update the changelog again once this merges.

PR #899, opened 2021-10-12
Clarifying docs around TLS options

robscott:
**What type of PR is this?**
/kind cleanup

This is a follow up from https://github.com/kubernetes-sigs/gateway-api/discussions/896 and our related discussion at the community meeting. 

```release-note
Added clarification that implementation-specific TLS options MUST be domain-prefixed.
```

/cc @jpeach @youngnick


robscott:
To clarify, I don't think this _needs_ to get in before we cut v1alpha2 or RC2, just wanted to file the PR while the topic was still fresh.

PR #918, opened 2021-10-26
Add GEP-917 to document conformance principles

### LINKED ISSUE #917: GEP: Clarify what we mean by Gateway API conformance

We need a GEP to cover:
- what is conformance for the Gateway API?
- what will the Gateway API project define and supply,
- what will implementors of the Gateway API be required to do to have their implementation be certified as conformant?

We've discussed this in a community meeting - the October 25th, 2021 meeting, to be precise. We also discussed the following documents:
Purpose summary:
https://docs.google.com/document/d/18iECeKMp1OewSGISskv6Chfmjo9u2U0_iUH0jhPdKOk/edit#
Requirements doc:
https://docs.google.com/document/d/1QL-MpIVzqxe32Y2BZ_dYOB8zNsF9c4pnKEIB9ZLt118/edit

This GEP should provide a high-level overview of *why* we will do conformance, *what* the success criteria or end result will be, and *how* implementations will gain compliance.

This is not a full implementation plan, it's a plan for the plan.

**Why this is needed**:
Right now, we are stuck doing conformance design because we don't have agreement on the terms of reference - this document is about making sure we are all coming from the same place.

youngnick:
**What type of PR is this?**
/kind gep

Adds GEP-917 to document conformance principles and outline what a conformance process for the Gateway API should and must provide.

Fixes #917

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-917.md
+++ b/site-src/geps/gep-917.md
@@ -1,0 +1,194 @@
+# GEP-917: Gateway API conformance principles
+
+* Issue: [#917](https://github.com/kubernetes-sigs/gateway-api/issues/917)
+* Status: Provisional
+
+(See definitions in [Kubernetes KEP][kep-status].
+
+[kep-status]: https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml#L9
+
+## TLDR
+
+This GEP outlines the reasons for and principles by which the Gateway API will
+design its conformance profiles and testing regime.
+
+## Goals
+
+- Record why we are doing conformance and what we hope to achieve with it
+- Record the success criteria for the conformance process and associated artifacts
+
+## Non-Goals
+
+- Designing the conformance profiles at anything more than a very basic level
+- Designing the conformance testing framework or implementation
+- Desgning the process for implementations to prove they are conformant
+
+## Introduction
+
+### What is Conformance
+
+Conformance is the creation of a process that allows everyone, implementors and
+users alike, to check that an implementation conforms to the defined spec.
+
+For core Kubernetes, this also allows for the use of specific badges and branding.
+
+It usually includes some form of test harness that can produce a standard output,
+which can be submitted somewhere for validation. The place where the validations
+are held is then the canonical source of information about what implementations
+are conformant.
+
+### Why do it for Gateway API?
+The Gateway API is a large, complex API with many use cases and implementations.
+Not all implementations support the same features, and some features have
+different required support levels.
+
+One of the primary goals of the Gateway API project is to make it safer and
+easier for end users to move their traffic configuration between implementations.
+Without some form of conformance to guarantee the same behavior between
+implementations, this is simply not achievable.
+
+Possibly a better way to say this is that we are looking to have Route objects
+portable between Gateways with a minimum of spec change.
+
+By creating a standard set of conformance tests and information, we can:
+- Make it easier for API consumers to understand what a particular API does
+- Make it possible for tooling to be constructed to check for portability
+between implementations
+
+Additionally, as the first project to develop an "official" set of CRDs, we have
+a responsibility to the community to build out a set of best practices for
+similar efforts in the future. Ensuring that whatever we build is reusable for
+other projects will help to lift everyone who works with CRDs.
+
+### What do we need out of conformance for Gateway API?
+Must have:
+- Support for different types of conformance, at a minimum Layer 4 and Layer 7
+Gateway API support. We wil call this “Conformance Profiles”, which is
+inline with upstream Kubernetes.
+- Support for fields or features that have Core, Extended, and
+ImplementationSpecific support. In particular, it should be possible for
+implementations to only support some subset of Extended fields, and to be able
+to use the framework for their own ImplementationSpecific features if required.
+- A testing suite that can validate that an implementation meets the conformance
+profiles it claims.
+- A way to retrieve conformance information in a machine-parseable way.
+
+## Proposal
+
+### Conformance profiles principles
+
+#### Basics
+We have two conformance profiles (to begin with), covering what we will loosely
+call "Layer 7" (meaning focussed around HTTP and HTTPS, like a traditional
+Ingress controller), and "Layer 4", focussed around forwarding TCP and UDP
+traffic.
+
+We may add more conformance profiles later.
+
+These conformance profiles are mainly defined in terms of the objects
+(specifically, the Route objects) they support, but can have additional features.
+
+However we end up naming them, at the end of the day, they are syntactic sugar
+for a set of tests to be run against an implementation, that guarantees that it
+behaves in the required way. We should aim to have the usual range of tests for
+both the happy and unhappy paths for various types of operations.
+
+When describing the idea of a conformance profile, it’s useful to think of it
+like a Git or Docker tag - it tracks the required features for a specific
+version of the API, and muyst be included in and updated by a version bump in the
+bundle version of the Gateway API. (The _bundle version_ is the thing that we
+mark as a "release", that looks like `v0.4.0`, not `v1alpha2`).
+
+This will enable an implementation to say that it supports version whatever of
+the Gateway API. This is again similar to upstream in that implementations need
+to submit conformance test results for each version of Kubernetes they support.
+
+It’s intended that conformance profiles are composable - a single implementation
+MAY support both conformance profiles, and the same will apply if we add more
+conformance profiles in the future.
+
+#### Interaction with existing support levels
+
+Conformance profiles will ensure that an implementation can provide all the
+features currently marked as "Core" support in the API documentation.
+
+Fields marked "Extended" support will eventually have conformance tests that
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-917.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Proposal
+
+### Conformance profiles principles
+
+#### Basics
+We have two conformance profiles (to begin with), covering what we will loosely
+call "Layer 7" (meaning focussed around HTTP and HTTPS, like a traditional
+Ingress controller), and "Layer 4", focussed around forwarding TCP and UDP
+traffic.
```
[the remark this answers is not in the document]
youngnick: The idea of profiles is to make it easier to understand what Routes can move between Gateways. If the Gateways have the same profile, if you are using features covered by conformance (ie Core or Extended), then you can safely move those Routes between those Gateways. 

I think that doing it the way you're talking about will mean that users will need to take a "guess I'll try it out and see if it works" approach, which I don't think is desirable.

Obviously, we can't mandate that everything works the same exactly between different controllers, but the support levels and conformance profiles should mean that application developer persona users can be confident that a Route can safely move between Gateways, regardless of implementation.

To put this another way, in my mind, one of the primary goals of the Gateway API project is to work towards Route portability between Gateways. The API is complex enough that we can't guarantee this across the board, but by providing subsets of common functionality, I think we can come pretty close. That's what the profiles are intended to achieve.

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,194 @@
...
+welcome.
+
+#### Layer 7 conformance profile
+
+Must support the following objects:
+- GatewayClass
+- Gateway
+- ReferencePolicy
+- HTTPRoute
+- TLSRoute
```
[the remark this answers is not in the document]
youngnick: Once we have conformance tests, it should be, but in practice right now probably isn't.

site-src/geps/gep-917.md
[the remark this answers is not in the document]
youngnick: We had a great discussion about this at the community meeting today, and I agreed that I'll rework this document to put the focus on the per-object conformance, and let the conformance profiles happen organically later. I'd be pretty shocked if we don't end up with some way to talk about the common sets of objects supported later, but this was probably a premature optimization on my part.

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+profiles it claims.
+- A way to retrieve conformance information in a machine-parseable way.
+
+## Proposal
+
+### Conformance profiles principles
+
+#### Basics
+
+The Gatway API project defines conformance purely in terms of what objects an
```
robscott, line 83:
```suggestion
The Gateway API project defines conformance purely in terms of what objects an
```

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+implementation supports.
+By "supports" in this context, we mean the functionality covered by Core support.
+Extended support fields are handled by a separate mechanism.
```
robscott, line 86:
```suggestion
To support a resource, an implementation must support all "Core" functionality defined by the resource. Support for "Extended" functionality will be indicated separately.
```

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+
+All implementations must support all the Core functions on the following objects:
+- GatewayClass
```
robscott, line 89: Tiny nit: Need new line before this for markdown list parsing

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+- Gateway
+- ReferencePolicy
+
+The following objects are optional to support, but have defined behavior if you
+do:
+- HTTPRoute
```
robscott, line 95: Tiny nit: Need new line before this for markdown list parsing

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+
+For all of these objects, we should aim to have the usual range of tests for
+both the happy and unhappy paths for various types of operations.
+
+The conformance is versioned - it tracks the required features for a specific
+version of the API, and must be included in and updated by a version bump in the
+bundle version of the Gateway API. (The _bundle version_ is the thing that we
+mark as a "release", that looks like `v0.4.0`, not `v1alpha2`). 
+
+This will enable an implementation to say that it supports version whatever of
```
robscott, line 108:
```suggestion
This will enable an implementation to say that it supports a specific version of
```

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+the Gateway API. This is again similar to upstream in that implementations need
+to submit conformance test results for each version of Kubernetes they support.
+
+Because the support is defined in terms of the objects that an implementation
```
robscott, line 112:
```suggestion
Because the support is defined in terms of the resources that an implementation
```

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+
+Fields marked "Extended" support will eventually have conformance tests that
+lock in the behavior of that feature, and there will be a mechanism for implementations
+to tell the testing framework what extended fields they support.
+
+#### Testing framework
+
+The Gateway API project will provide a set of tests and harness to run them, such
+that an implementation may point the test harness at a Gateway or set of Gateways
+managed by that implenentation and have the testing framework deliver a report
```
robscott, line 130:
```suggestion
managed by that implementation and have the testing framework deliver a report
```

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+features currently marked as "Core" support in the API documentation.
+
+Fields marked "Extended" support will eventually have conformance tests that
+lock in the behavior of that feature, and there will be a mechanism for implementations
+to tell the testing framework what extended fields they support.
+
+#### Testing framework
+
+The Gateway API project will provide a set of tests and harness to run them, such
+that an implementation may point the test harness at a Gateway or set of Gateways
```
robscott, line 129: I've actually thought of conformance targeting a GatewayClass and then creating/testing Gateways, would that be workable?

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+mark as a "release", that looks like `v0.4.0`, not `v1alpha2`). 
+
+This will enable an implementation to say that it supports version whatever of
+the Gateway API. This is again similar to upstream in that implementations need
+to submit conformance test results for each version of Kubernetes they support.
+
+Because the support is defined in terms of the objects that an implementation
+supports, the conformance is composable, and orthogonal for each object type.
+It's valid to only support TLSRoute and not HTTPRoute, for example, although
+it would be odd.
```
robscott, line 115: Tiny nit: Maybe focus on a more common example such as supporting HTTPRoute but not TCPRoute.

robscott (commented):
This is great, thanks for the work on this @youngnick! Just a few nits but generally LGTM and will be great to have this merged in. Maybe this should serve as our primary conformance GEP? If/when we want to add more detail around how conformance should work, it could be added to this GEP instead of in a new GEP.

jpeach:
This seems like a very sound approach.

/lgtm
/hold

site-src/geps/gep-917.md
```diff
@@ -0,0 +1,162 @@
...
+- ReferencePolicy
+
+The following objects are optional to support, but have defined behavior if you
+do:
+- HTTPRoute
+- TLSRoute
+- TCPRoute
+- UDPRoute
+
+For all of these objects, we should aim to have the usual range of tests for
```
hbagdi, line 100:
```suggestion
For all of these resources, we should aim to have the usual range of tests for
```
Use a replace-all for this.

site-src/geps/gep-917.md
hbagdi: Also, remove the "it would be odd", we may be potentially shooting down some cases. Others in the community will be (are?) more creative and use this API in interesting ways.

site-src/geps/gep-917.md
hbagdi: Maybe?
In-cluster proxies will need something akin to an operator for that to work.
Kong is not considering Gateway lifecycle management in the first iteration.

robscott: If Kong is deployed behind a Service, will it be sufficiently aware of that to set an IP in Gateway status? I believe that's how Istio's currently setting this status, but not sure, @howardjohn would know best.

youngnick: I think that just putting in here that the testing may be pointed at a GatewayClass or Gateway will cover us no matter what.

But I think that full dynamic Gateway creation may be a ways away for a few implementations at least, Kong and Contour among them. So we should have a way to set whether a new Gateway can be created or not as well.

youngnick:
I've fixed up all the comments, so I'm just going to remove the hold; next LGTM will merge.

/hold cancel

robscott:
Thanks!

/lgtm

Issue #922, opened 2021-10-29
GEP: Gateway API Versioning


robscott:
**What would you like to be added**: 
Clarification around compatibility guarantees, experimental fields, and versioning plans.

**Why this is needed**: 
We have a large number of implementations being built around this API. As we approach a beta release, we need a well defined plan for how that and future releases will work.

youngnick:
Looks like we missed closing this when #923 was merged.

/close

PR #923, opened 2021-10-29
Adding GEP 922: Gateway API Versioning Plan

robscott:
**What type of PR is this?**
/kind gep

This proposes a set of versioning guidelines for future Gateway API releases. This is a follow up to the [related doc](https://docs.google.com/document/d/16tvd2j-zUdL1tqzFJFHGOI2WFCWs6dosj3QPkVDAdOY/edit#) that has been discussed in gateway-api and sig-apimachinery meetings.

Related to #922, will not be fixed until this GEP moves to "Implemented"

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-922.md
+++ b/site-src/geps/gep-922.md
@@ -1,0 +1,178 @@
+# GEP-922: Gateway API Versioning
+
+* Issue: [#922](https://github.com/kubernetes-sigs/gateway-api/issues/922)
+* Status: Implementable
+
+## TLDR
+Each Gateway API release will be represented by a bundle version that represents
+that specific combination of CRDs, API versions, and validating webhook. To
+enable experimental fields, future releases of the API will include stable and
+experimental CRD tracks. Users will be able to access experimental features by
+installing the experimental CRDs. A cluster can contain either an experimental
+or stable CRD for any resource at a given time. 
+
+
+## Goals
+Provide a path for implementations to support:
+
+* Users should have a predictable experience when interacting with Gateway API. 
+* The API should be able to continue to safely evolve, including:
+    * Loosening validation
+    * Transitioning required fields to optional
+    * Adding new fields at an experimental stability level
+
+## Introduction
+This GEP details how we should approach future releases of Gateway API. This is
+intended as another step towards building consensus. Once merged, it will be
+considered implemented once this content is fully documented.
+
+## Version Terminology
+Gateway API has two related but different versioning schemes:
+
+### 1. API Versions (ex: v1alpha2)
+Each API version provides a unique way to interact with the API. Significant
+changes such as removing or renaming fields will be represented here.
+
+### 2. Bundle Versions (ex: v0.4.0)
+Bundle versions are broader in scope. They use semantic versioning and include
+the following:
+
+* API Types/CRDs
+* Validating Webhook
+
+Each bundle may include multiple API versions, potentially introducing new ones
+and/or removing old ones as part of a new bundle.
+
+## Limitations of Webhook and CRD Validation
+CRD and webhook validation is not the final validation i.e. webhook is “nice UX”
+but not schema enforcement. This validation is intended to provide immediate
+feedback to users when they provide an invalid configuration, but can not
+completely be relied on because it:
+
+* Is not guaranteed to be present or up to date in all clusters.
+* Will likely never be sufficient to cover all edge cases.
+* May be loosened in future API releases.
+
+## Persona Requirements
+When implementing or using Gateway API, each persona has a unique set of
+responsibilities to ensure we're providing a consistent experience.
+
+### API Authors:
+* MUST provide conversion between API versions (excluding experimental fields),
+  starting with v1alpha2.
+* MAY include the following changes to an existing API version with a new bundle
+ **patch** version:
+    * Clarifications to godocs
+    * Loosening validation to fix a bug (not to enable a new feature)
+    * Fixes to typos
+* MAY include the following changes to an existing API version with a new bundle
+  **minor** version:
+    * Everything that is valid in a patch release
+    * New experimental API fields or resources
+    * Loosened validation
+    * Making required fields optional
+    * Removal of experimental fields
+    * Removal of experimental resources
+    * Graduation of fields or resources from experimental to stable track
+* MAY introduce a new **API version** with a new bundle minor version, which may
+  include:
+    * Everything that is valid in a minor release
+    * Renamed fields
+    * Anything else that is valid in a new Kubernetes API version
+    * Removal/tombstoning of beta fields
+* MAY release a new major bundle version (v1.0) as part of graduating the API to
+  GA and releasing a new API version.
+
+Note that each new bundle version, no matter how small, may include updated
+CRDs, webhook, or both. Implementations may read annotations on Gateway API CRDs
+(defined below) to determine the version and channel of CRDs that have been
+installed in the cluster.
+
+### Implementers:
+* MUST handle fields with loosened validation without crashing
+* MUST handle fields that have transitioned from required to optional without
+  crashing
+* MUST NOT rely on webhook or CRD validation as a security mechanism. If field
+  values need to be escaped to secure an implementation, both webhook and CRD
+  validation can be bypassed and cannot be relied on. Instead, implementations
+  should implement their own escaping or validation as necessary. To avoid
+  duplicating work, Gateway API maintainers are considering adding a shared
+  validation package that implementations can use for this purpose. This is
+  tracked by [#926](https://github.com/kubernetes-sigs/gateway-api/issues/926).
+
+### Installers:
+* MUST install a full Gateway API bundle, with matching CRD and webhook
+  versions.
+
+## Adding Experimental Fields
+Over time, it will be useful to add experimental fields to the API. In upstream
+Kubernetes, those would generally be guarded with feature gates. With Gateway
+API we will accomplish by releasing experimental versions of our CRDs.
+
+With this approach, we achieve a similar result. Instead of using feature gates
+and validation to prevent fields from being set, we just release separate CRDs.
+Once the API reaches beta, each bundle release can include 2 sets of CRDs,
+stable and experimental.
+
+New fields will be added to the experimental set of CRDs first, and may graduate
+to stable APIs later. Experimental fields will be marked with the
+`+experimental` annotation in Go type definitions. Gateway API CRD generation
+will exclude these fields from stable CRDs. Experimental fields may be removed
+from the API. Due to the experimental nature of these CRDs, they are not
+recommended for production use.
+
+If experimental fields are removed or renamed, the original field name should be
+removed from the go struct, with a tombstone comment ensuring the field name
+will not be reused. 
+
+Each CRD will be published with annotations that indicate their bundle version
+and channel:
+
+```
+gateway.networking.k8s.io/bundle-version: v0.4.0
+gateway.networking.k8s.io/channel: stable|experimental
+```
+
+## Alternatives Considered
+### 1. Use Webhook to Implement Feature Gates
+This option would involve using the webhook to prevent experimental fields from
+being set unless the corresponding feature gate was enabled. This would make use
+of the existing Kubernetes feature gate tooling.
+
+This approach would be familiar to Kubernetes users, but would put a lot of
+emphasis on the webhook. It would require webhook versioning to be precisely in
+line with the CRDs, and edge cases during the upgrade process could get rather
+messy.
+
+### 2: Reuse Alpha API versions
+With this approach, we would only allow the use of these fields when using the
+alpha API. 
+
+To simplify conversion, new fields will be added to all versions of the API,
+with some key exceptions:
+
+* New fields will be added with an `alpha` prefix to beta and GA API versions.
+* Webhook validation would ensure that new fields will only be writable when
+  using alpha API versions.
+* New fields may be removed without replacement while they are still considered
+  alpha.
+
+An alpha API field could graduate to beta in the subsequent minor release of the
+API. That process would involve:
+
+* Loosening webhook validation to allow writes to the field with beta API
+  versions.
+* Removing the alpha prefix from the field name in the beta API Changing the
+  alpha prefix to a beta prefix in the GA API.
+
+A beta API field could graduate to GA in the subsequent minor release of the
+API. That process would involve:
+
+* Loosening webhook validation to allow writes to the field with GA API
+  versions.
+* Removing the beta prefix from the field name in the GA API.
+
+This is potentially the most complex of all the options presented. It would
+effectively require us to consistently maintain an alpha API version. Of all the
+options presented, this is probably the most likely to confuse users and
+implementers of the API.


site-src/geps/gep-922.md
```diff
@@ -0,0 +1,150 @@
...
+  include:
+    * Everything that is valid in a minor release
+    * Renamed fields
+    * Removal/tombstoning of beta fields
+
+### Implementers:
+* MUST handle fields with loosened validation without crashing
+* MUST handle fields that have transitioned from required to optional without
+  crashing
+* MUST NOT rely on webhook or CRD validation as a security mechanism. If field
```
youngnick, line 84: Maybe some examples here? I think that it's pretty safe to rely on webhook and CRD validation to ensure maximum field lengths, for example, but not to ensure referential integrity across objects.

youngnick (commented):
This is looking pretty good overall to me, but there's a couple of things that could be clearer.

site-src/geps/gep-922.md
robscott: Although I think it's reasonable to implement something like max length validation in the webhook or CRD, I think it's still important for implementations to ensure they don't crash if they receive a value outside of those bounds. Unfortunately it is entirely possible for users to bypass webhook or CRD validation by modifying/removing them.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,150 @@
...
+
+Each CRD will be published with annotations that indicate their bundle version
+and channel:
+
+```
+gateway.networking.k8s.io/bundle-version: v0.4.0
+gateway.networking.k8s.io/channel: stable|experimental
+```
+
+## Alternatives Considered
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-922.md
robscott: I've been trying to find a sentence or two that we can safely add here, but so far everything I've come up with seems like something we may regret saying. For example, if we say that an implementation can avoid length validation, there's a chance that someone overloads a value so far that it causes their controller to crash or maybe even exploit some underlying vulnerability. Essentially every possible thing I can think of suggesting implementers could safely skip seems to have at least one edge case where we'd regret giving that guidance.

hbagdi: I think there are limits to the degree of uncertainty that controllers will be able to tolerate.
Let's take length validation for example. A field like the name of the Gateway has length validation within the resource definition. Now, the user can bypass that by editing the CRD or passing --validate=false to kubectl but I think there are practical limits to how much of these controllers can tolerate. Asking controller authors to all of the validation again is a bit unreasonable I think - we would be only simplifying our lives by passing on the problem.

Also, if it is expected from the controllers that they do not rely on any external validation _at all_, then we will need to invest in a validation library that controllers can use as they resolve resources.

hbagdi: Also, I'm not sure how but we need to also involve sig-architecture and api-machinery folks in here and see what (if anything) is that they recommend. This should not be a problem that is being discussed for the first time.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+feedback to users when they provide an invalid configuration, but can not
+completely be relied on because it:
+
+* Is not guaranteed to be present or up to date in all clusters.
+* Will likely never be sufficient to cover all edge cases.
+* May be loosened in future API releases.
+
+## Role Requirements
+When implementing or using Gateway API, each role has a unique set of
+requirements to ensure we're providing a consistent experience.
```
hbagdi, line 55:
```suggestion
## Persona Requirements
When implementing or using Gateway API, each persona has a unique set of
responsibilities to ensure we're providing a consistent experience.
```

I got confused with roles in our security model and roles as in RBACRole in kube.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+
+### API Authors:
+* MUST support API versions for at least 1 year, starting with v1alpha2.
```
hbagdi, line 58: What does support mean?
Like the user can configure v1alpha2 for a year? Or does that mean that we will be pushing security fixes (if they come up) for a year? Is there any value in doing this for a year, especially for an alpha API?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+* MUST provide conversion between API versions, starting with v1alpha2.
```
hbagdi, line 59: To be clear, this doesn't imply conversion webhook, correct?
Does this imply round-trip to lower versions is required?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+* MAY include the following changes to an existing API version with a new bundle
+ **patch** version:
+    * Clarifications to godocs
+    * Loosening validation to fix a bug (not to enable a new feature)
+    * Fixes to typos
```
hbagdi, line 64: This _implies_ updates to CRDs. Can we make that explicit since not all readers will understand the implication?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+* MAY include the following changes to an existing API version with a new bundle
+  **minor** version:
+    * Everything that is valid in a patch release
+    * New alpha API fields or resources
```
hbagdi, line 68:
```suggestion
    * New experimental API fields or resources
```

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+    * Loosened validation
+    * Making required fields optional
+    * Removal of experimental fields
+    * Removal of experimental resources
+    * Graduation of fields or resources (alpha->beta, beta->ga)
```
hbagdi, line 73: Will fields have alpha->beta->ga or will it be more like experimental->ga? The former at the field-level sounds like a recipe for a lot of pain and confusion.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+* MAY introduce a new **API version** with a new bundle minor version, which may
+  include:
+    * Everything that is valid in a minor release
+    * Renamed fields
+    * Removal/tombstoning of beta fields
```
hbagdi, line 78: Is any commentary necessary on the major version or is that we plan to always be on 0.x.y?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Implementers:
+* MUST handle fields with loosened validation without crashing
+* MUST handle fields that have transitioned from required to optional without
+  crashing
+* MUST NOT rely on webhook or CRD validation as a security mechanism. If field
+  values need to be escaped to secure an implementation, both webhook and CRD
+  validation can be bypassed and cannot be relied on. Instead, implementations
+  should implement their own escaping or validation as necessary.
```
hbagdi, line 87: See https://github.com/kubernetes-sigs/gateway-api/pull/923#discussion_r741311631
I think this seems a bit draconian on our part.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+removed from the go struct, with a tombstone comment ensuring the field name
+will not be reused. 
+
+Each CRD will be published with annotations that indicate their bundle version
+and channel:
+
+```
+gateway.networking.k8s.io/bundle-version: v0.4.0
+gateway.networking.k8s.io/channel: stable|experimental
+```
```
hbagdi, line 119: So, at the end of the day, a cluster can either have a stable or an experimental CRD. It is exclusive because you can only have one CRD at a time for a GK.

Why have the distinction in that case?
What if we just ship the experimental fields as part of the stable channel?

youngnick: If we do that, then the only way for YAML consumers (and let's be real, most people will use YAML, not read the full spec) to know what fields are experimental is for us to prefix them somehow - that is, we'd have to call all the experimental fields `experimentalFieldName` or something.

Other things I can think of that only supplying on CRD spec would imply:
- Implementations would need a switch of some sort to turn on experimental field processing - effectively, we're moving the feature gate to the implementing controller, and trusting them to figure it out.
- We'd also need to supply a method for implementations to determine what the experimental fields are, so that they can tell people they're using an experimental field but haven't enabled it.

Having two separate CRDs solves all of those problems - the implementation can switch experimental field processing on or off based on the annotation on the CRD object itself (this also means you can't enable experimental field processing without having modify CRDs access, which I think is fine).

site-src/geps/gep-922.md
youngnick: I agree that it's very difficult to draw a middle ground here. Either we rely on the CRD and webhook validation, or we don't. I agree that implementations should ideally not crash if things are not present, but if we don't trust the validations to some extent, there's no choice other than to replicate all of them inside the implementations, in which case I agree with @hbagdi that we should supply validation code along with the CRD types themselves, and expect that implementation authors run that code before touching the object.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,163 @@
...
+
+* Issue: [#922](https://github.com/kubernetes-sigs/gateway-api/issues/922)
+* Status: Implementable
+
+## TLDR
+Each Gateway API release will be represented by a bundle version that represents
+that specific combination of CRDs, API versions, and validating webhook. To
+enable experimental fields, a new experimental set of CRDs will be included with
+future releases.
+
```
bowei, line 11: We should state what we are calling this, to avoid any confusion.

"Stable" vs "Experimental" CRD version.

Should also probably say the API version (e.g. v1alphaX) does not represent the right model for CRD versioning.

site-src/geps/gep-922.md
robscott: Updated, thanks!

site-src/geps/gep-922.md
robscott: I think you're right, this doesn't really make sense quite yet. I think LTS is something we should really be considering as part of a beta release, but not something we need to worry about for alpha.

site-src/geps/gep-922.md
robscott: Added a note below this to clarify that.

site-src/geps/gep-922.md
robscott: Updated, thanks!

site-src/geps/gep-922.md
robscott: Good catch, I've updated this to clarify that fields will graduate from experimental -> stable, not alpha->beta->ga as I originally had here.

site-src/geps/gep-922.md
robscott: Good catch, I've added a note about a major release to accompany a v1 API.

site-src/geps/gep-922.md
robscott: Agree with @youngnick's points above. I think it's important to have users explicitly opt in to experimental functionality, and I think separate CRDs are likely the best mechanism we have for this. Although I'd really like something better like kubectl support/gating for alpha fields, I don't think that's going to be practical anytime soon.

site-src/geps/gep-922.md
robscott: Updated to make this a bit more clear

site-src/geps/gep-922.md
robscott: As far as I can tell, a conversion webhook is necessary for the following:

1. Renaming fields
2. Restructuring fields

I'd like to avoid a conversion webhook as long as possible, but I do think it's likely preferable to having incompatible API versions, especially if we're expecting our next release to be v1beta1.

site-src/geps/gep-922.md
robscott: I think it's important to note that what we're saying here is that _any_ validation we provide either in webhook or CRD form can be bypassed. This is a fundamental limitation of CRDs. I don't think we can really change the advice given here because the validation we're providing is simply not something that can be securely relied on.

 I think that likely means that it would be useful to export some shared validation code with each bundle release that _can_ be imported + relied on. Maybe this means that this shared validation code is the first thing that runs through a resource, and if it fails, an implementation triggers a warning event and otherwise ignores the resource?

site-src/geps/gep-922.md
robscott: Agree this is tricky, added a comment in the thread you referenced.

site-src/geps/gep-922.md
hbagdi: > I think it's important to note that what we're saying here is that any validation we provide either in webhook or CRD form can be bypassed. This is a fundamental limitation of CRDs. I don't think we can really change the advice given here because the validation we're providing is simply not something that can be securely relied on.

Added this to https://github.com/kubernetes-sigs/gateway-api/issues/629.

>  Maybe this means that this shared validation code is the first thing that runs through a resource, and if it fails, an implementation triggers a warning event and otherwise ignores the resource?

Yes. Although, we have a lot on our plate and I'm afraid that this could fall off, resulting in fragmentation, which will have a direct impact on the UX of the API.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+    * Everything that is valid in a minor release
+    * Renamed fields
+    * Removal/tombstoning of beta fields
+* MAY release a new major bundle version (v1.0) as part of graduating the API to
+  GA and releasing a v1 API version.
+
+Note that each new bundle version, no matter how small, may include updated
+CRDs, webhook, or both.
+
+### Implementers:
```
hbagdi, line 85: What if an implementation doesn't want to work in a cluster unless the installed version matches a support matrix that the it is programmed against? Is that invalid behavior?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+With this approach, we achieve a similar result. Instead of using feature gates
+and validation to prevent fields from being set, we just release separate CRDs.
+Once the API reaches beta, each bundle release can include 2 sets of CRDs,
+stable and experimental.
+
+New fields will be added to the experimental set of CRDs first, and may graduate
+to stable APIs later. Experimental fields will be marked with the
+`+experimental` annotation in Go type definitions. Gateway API CRD generation
+will exclude these fields from stable CRDs. Experimental fields may be removed
+from the API.
```
hbagdi, line 112: As an end-user, should I use experimental CRDs in production? Case-by-case basis?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+    * Fixes to typos
+* MAY include the following changes to an existing API version with a new bundle
+  **minor** version:
+    * Everything that is valid in a patch release
+    * New experimental API fields or resources
+    * Loosened validation
+    * Making required fields optional
+    * Removal of experimental fields
+    * Removal of experimental resources
+    * Graduation of fields or resources from experimental to stable track
```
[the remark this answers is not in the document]
robscott: Good point - we do have annotations proposed further down that would enable this. Annotations should be able to be read programmatically, as long as the controller requests read access on CRDs which seems reasonable and likely necessary.

site-src/geps/gep-922.md
robscott: I think that's valid, but likely unnecessarily constrained. If all the guidelines here are followed, newer additions to the API would simply not be supported by an older implementation of the API. A warning or set of warnings is likely preferable to stopping altogether.

site-src/geps/gep-922.md
robscott: Added https://github.com/kubernetes-sigs/gateway-api/issues/926 to track this.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+* MAY introduce a new **API version** with a new bundle minor version, which may
+  include:
+    * Everything that is valid in a minor release
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+    * Renamed fields
+    * Removal/tombstoning of beta fields
+* MAY release a new major bundle version (v1.0) as part of graduating the API to
+  GA and releasing a v1 API version.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-922.md
robscott: Good point, clarified that this is "not recommended"

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+installing the experimental CRDs.
+
+## Goals
+Provide a path for implementations to support:
+
+* Users should have a predictable experience when interacting with Gateway API. 
+* The API should be able to continue to safely evolve, including:
+    * Loosening validation
+    * Transitioning required fields to optional
+    * Adding new fields at an alpha stability level
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,168 @@
...
+
+* Issue: [#922](https://github.com/kubernetes-sigs/gateway-api/issues/922)
+* Status: Implementable
+
+## TLDR
+Each Gateway API release will be represented by a bundle version that represents
+that specific combination of CRDs, API versions, and validating webhook. To
+enable experimental fields, future releases of the API will include stable and
+experimental CRD tracks. Users will be able to access experimental features by
+installing the experimental CRDs.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/geps/gep-922.md
hbagdi: Can we add a little bit of commentary to the GEP itself on this? For posterity.

site-src/geps/gep-922.md
hbagdi: Okay. Can we explicitly mention that controllers can rely on the annotation to decide their behavior?

site-src/geps/gep-922.md
robscott: Good call - I added some context here that we're considering a shared validation package + a link to #926.

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,177 @@
...
+### 1. API Versions (ex: v1alpha2)
+Each API version provides a unique way to interact with the API. Significant
+changes such as removing or renaming fields will be represented here.
+
+### 2. Bundle Versions (ex: v0.4.0)
+Bundle versions are broader in scope. They use semantic versioning and include
+the following:
+
+* API Types/CRDs
+* Validating Webhook
```
youngnick, line 41: Should we mention other validation code here?

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,177 @@
...
+* Is not guaranteed to be present or up to date in all clusters.
+* Will likely never be sufficient to cover all edge cases.
+* May be loosened in future API releases.
+
+## Persona Requirements
+When implementing or using Gateway API, each persona has a unique set of
+responsibilities to ensure we're providing a consistent experience.
+
+### API Authors:
+* MUST provide conversion between stable API versions, starting with v1alpha2.
```
youngnick, line 61:
```suggestion
* MUST provide conversion between API versions, starting with v1alpha2.
```
Feels weird to call an alpha API "stable".

site-src/geps/gep-922.md
```diff
@@ -0,0 +1,177 @@
...
+(defined below) to determine the version and channel of CRDs that have been
+installed in the cluster.
+
+### Implementers:
+* MUST handle fields with loosened validation without crashing
+* MUST handle fields that have transitioned from required to optional without
+  crashing
+* MUST NOT rely on webhook or CRD validation as a security mechanism. If field
+  values need to be escaped to secure an implementation, both webhook and CRD
+  validation can be bypassed and cannot be relied on. Instead, implementations
```
youngnick, line 96: This wording is fine for now, but if we do implement the validation module, feels like it should be a MUST to me.

youngnick (commented):
This lgtm with a few non-blocking nits.

site-src/geps/gep-922.md
robscott: Good catch, I added a note to https://github.com/kubernetes-sigs/gateway-api/issues/926 so we remember to come back to this GEP and update it.

site-src/geps/gep-922.md
robscott: Yeah, the terminology here is confusing. I've updated this to "excluding experimental fields", which I think is a clearer representation of what I was aiming for here.

site-src/geps/gep-922.md
robscott: I think we should definitely add that to our bundle versions if/when we have it. I have a note in #926 to come back to this KEP and update it if/when we add that logic.

hbagdi:
Here is to hoping we have thought this through.
/lgtm

robscott:
I think there's broad enough consensus on this to merge it in, we can revisit through smaller tweaks if needed.

/hold cancel

PR #925, opened 2021-11-04
docs: unmanaged gateway development guide

### LINKED ISSUE #892: docs: Gateway implementation guide for managed vs. unmanaged gateway lifecycles and intra vs extra cluster

This is a request for documentation for implementations which would provide additional guidance and prescription on how to implement `Gateway` controllers for different kinds of underlying Gateway infrastructure options. Particularly the guide should cover:

- `Gateway`s which are "unmanaged" and refer to an existing deployment

Additional Goals:

- define clear terminology about the underlying lifecycle management and network location of `Gateways` 

**Why this is needed**:

This is a direct follow up from a community meeting where the overall topic was discussed, and was a follow up to:

- @howardjohn's [Document](https://docs.google.com/document/d/1BjmBUaNnDsk8LcDF0YbItV1kRRtc3eiWBdgvRg-1hfo) on the subject
- A related discussion which suggested some implementation details for how to solve some of the problems in this domain with spec changes: https://github.com/kubernetes-sigs/gateway-api/discussions/877

shaneutt:
**What type of PR is this?**

/kind documentation

In a previous [discussion](https://github.com/kubernetes-sigs/gateway-api/discussions/877) as well as a [document shared with contributors](https://docs.google.com/document/d/1BjmBUaNnDsk8LcDF0YbItV1kRRtc3eiWBdgvRg-1hfo) we pondered and talked about how we might handle "unmanaged gateways", which are `Gateway` resources that would reference an existing server that would otherwise not be provisioned by or have its lifecycle managed by the `Gateway` controller implementation itself. Several implementors noted that supporting an unmanaged gateway option was something they would
need ultimately, but we decided it wasn't something that would be easy (or even ultimately correct) to express via the `Gateway` specification. Given that context and during a community sync it was suggested to start a guide which would provide **high level**, and **example style** ideas about how to implement this, if nothing else just so we had some guidance for future implementors, this patch adds such a guide.

Fixes #892

```release-note
NONE
```

**Additional Notes**:

I only ended up covering the topics of:

- nomenclature: naming and describing "unmanaged" and "managed" gateways accordingly
- configuration: how one might actually differentiate between unmanaged and managed gateways
- controller strategy: a high level strategy for implementing unmanaged gateway reconcilation logic using reference `Services`

I know this is very limited at this point, but I wanted to start small to see how this was received. I would invite others who've solved similar problems to continue growing the guide according to their experience.

### DIFF
--- a/site-src/geps/gep-917.md
+++ b/site-src/geps/gep-917.md
@@ -1,0 +1,166 @@
+# GEP-917: Gateway API conformance principles
+
+* Issue: [#917](https://github.com/kubernetes-sigs/gateway-api/issues/917)
+* Status: Provisional
+
+## TLDR
+
+This GEP outlines the reasons for and principles by which the Gateway API will
+design its conformance and testing regime.
+
+## Goals
+
+- Record why we are doing conformance and what we hope to achieve with it
+- Record the success criteria for the conformance process and associated artifacts
+
+## Non-Goals
+
+- Designing the conformance process at anything more than a very basic level
+- Designing the conformance testing framework or implementation
+- Designing the process for implementations to prove they are conformant
+
+## Introduction
+
+### What is Conformance
+
+Conformance is the creation of a process that allows everyone, implementors and
+users alike, to check that an implementation conforms to the defined spec.
+
+For core Kubernetes, this also allows for the use of specific badges and branding.
+
+It usually includes some form of test harness that can produce a standard output,
+which can be submitted somewhere for validation. The place where the validations
+are held is then the canonical source of information about what implementations
+are conformant.
+
+### Why do it for Gateway API?
+The Gateway API is a large, complex API with many use cases and implementations.
+Not all implementations support the same features, and some features have
+different required support levels.
+
+One of the primary goals of the Gateway API project is to make it safer and
+easier for end users to move their traffic configuration between implementations.
+Without some form of conformance to guarantee the same behavior between
+implementations, this is simply not achievable.
+
+Possibly a better way to say this is that we are looking to have Route resources
+portable between Gateways with a minimum of spec change.
+
+By creating a standard set of conformance tests and information, we can:
+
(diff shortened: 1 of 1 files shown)


site-src/v1alpha2/guides/unmanaged-gateway-development.md
```diff
@@ -0,0 +1,107 @@
...
+for every implementation, if your implementation does in fact include an
+existing `Service` resource which is distinct and well defined and it might be
+feasible to consider it the "source of truth" for the `Listeners` and
+`Addresses` that would otherwise be represented by the `Gateway`, then perhaps
+consider coupling the two together:
+
+- require a reference to a `Service` to be present for unmanaged `Gateway`
+  functionality to be available, perhaps via annotations or via controller
+  manager command line flags and environment variables.
+- add reconcilation logic which can convert a `Service` spec into `Listeners`
```
howardjohn, line 91: Isn't this then pretty much 'managed'?

If you do this then you cannot use a *lot* of functionality. TLS, scoping the route selection, etc.

I don't think creating Gateway from Service is a good idea, personally. The reverse is.

howardjohn (commented):
Here is our approach: https://github.com/istio/istio/pull/35876/files

Which I feel like can be generally applicable. I am not a huge fan of requiring an annotation when we already have and `addresses` field that can serve the same purpose (and effectively does for cloud LBs, if we consider 'managed' == provision a new IP/LB and 'unmanaged' == use a preprovisioned IP/LB).

site-src/v1alpha2/guides/unmanaged-gateway-development.md
shaneutt: I'll have to think on it some, there are definitely limitations here.

shaneutt:
Given initial feedback I've decided to hold off on this for the moment. I'll take some time to consider other options.

PR #932, opened 2021-11-15
GEP 735 L4 Traffic Matching

### LINKED ISSUE #735: Add Address Matching to TCPRoute and UDPRoute
Add address matching to `TCPRouteMatch` and `UDPRouteMatch`. 

**Why this is needed**:
TCP and UDP Routes currently only support custom matching extensions. It would be helpful to include some basic matching capabilities within the Routes. Address matching seems like a natural starting point. Although for most ingress use cases this could be covered by Gateway listeners, this concept could be quite helpful for mesh routing.

Also, see transparent proxy discussion in #727

shaneutt:
**What type of PR is this?**
/kind feature
/kind design
/kind gep
/kind api-change
-->

This adds traffic matching to enable more expressive routing for `TCPRoute` and `UDPRoute`.

Fixes #735

```release-note
- TCPRouteRule now includes an optional "matches" field for traffic matching
- UDPRouteRule now includes an optional "matches" field for traffic matching
```

### DIFF
--- a/site-src/geps/gep-735.md
+++ b/site-src/geps/gep-735.md
@@ -52,6 +52,9 @@ type AddressMatch struct {
 	//
 	// For IPAddress the implementor may expect either IPv4 or IPv6.
 	//
+	// Support: Core (IPAddress)
+	// Support: Custom (NamedAddress)
+	//
 	// +optional
 	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
 	// +kubebuilder:default=IPAddress
@@ -61,6 +64,8 @@ type AddressMatch struct {
 	// on the type and support by the controller.
 	//
 	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
+	//
+	// Support: Core
 	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
@@ -77,10 +82,14 @@ Using the new `AddressMatch` type matches can be expressed in topical lists on
 type TrafficMatches struct {
 	// SourceAddresses indicates the originating (source) network
 	// addresses which are valid for routing traffic.
+	//
+	// Support: Core
 	SourceAddresses []AddressMatch `json:"sourceAddresses"`
 
 	// DestinationAddresses indicates the destination network addresses
 	// which are valid for routing traffic.
+	//
+	// Support: Core
 	DestinationAddresses []AddressMatch `json:"destinationAddresses"`
 ```
 
--- a/site-src/geps/gep-735.md
+++ b/site-src/geps/gep-735.md
@@ -31,59 +31,57 @@ level of tuning options for L4 traffic routing at a level below the `Gateway`.
 
 The API changes include the following new types:
 
-- `NetworkMatch` to indicate the IP for address matching
+- `AddressMatch` to indicate the IP for address matching
 - `TrafficMatches` to configure matching according to network address
 
 These types enable the address matching required, with some active
 considerations about how to leave these open ended for later expansion.
 
-### NetworkMatch Type
+### AddressMatch Type
 
-A new `NetworkMatch` type provides the targeting mechanism for match inclusion
+A new `AddressMatch` type provides the targeting mechanism for match inclusion
 of a given network address:
 
 ```go
-type NetworkMatch struct {
-	// Address is an IP address which refers to either an IPv4 or IPv6 address.
+type AddressMatch struct {
+	// Type of the address, either IPAddress or NamedAddress.
 	//
-	// Examples: `127.0.0.1`, `::1`
+	// If NamedAddress is used this is a custom and specific value for each
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
+	// +kubebuilder:default=IPAddress
+	Type *AddressType `json:"type,omitempty"`
+
+	// Value of the address. The validity of the values will depend
+	// on the type and support by the controller.
+	//
+	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
 	//
 	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=45
-	Address string `json:"address"`
+	// +kubebuilder:validation:MaxLength=253
+	Value string `json:"value"`
 }
 ```
-
-The purpose of calling this `NetworkMatch` and wrapping the single `IP` attribute in
-a structure is to enable forward expansion in case later we're interested in
-adding more specificity, such as perhaps the network mask or `/<int>` CIDR to
-allow matching within a network range, instead of a specific IP address:
-
-```go
-type NetworkMatch struct {
-	Address string `json:"address"`
-	Mask    *string `json:"mask"`
-}
-```
-
-However for this iteration the goal is to keep the initial spec simple and to
-the immediate goals so only `Address` is intended for inclusion until further
-requirements develop.
 
 ### TrafficMatches Type
 
-Using the new `NetworkMatch` type matches can be expressed in topical lists on
+Using the new `AddressMatch` type matches can be expressed in topical lists on
 `TCPRoute` and `UDPRoute` using the new `TrafficMatches` type:
 
 ```go
 type TrafficMatches struct {
 	// SourceAddresses indicates the originating (source) network
 	// addresses which are valid for routing traffic.
-	SourceAddresses []NetworkMatch `json:"sourceAddresses"`
+	SourceAddresses []AddressMatch `json:"sourceAddresses"`
 
 	// DestinationAddresses indicates the destination network addresses
 	// which are valid for routing traffic.
-	DestinationAddresses []NetworkMatch `json:"destinationAddresses"`
+	DestinationAddresses []AddressMatch `json:"destinationAddresses"`
 ```
 
 This type becomes an optional field and shared by both `TCPRouteRule` and
@@ -109,10 +107,12 @@ spec:
   rules:
   - matches:
       sourceAddresses:
-      - address: "192.168.1.1"
-      - address: "FE80::0202:B3FF:FE1E:8329"
+      - value: "192.168.1.1"
+      - value: "FE80::0202:B3FF:FE1E:8329"
+      - type: NamedAddress
+        value: "my-custom-name"
       destinationAddresses:
-      - address: "10.96.0.1"
+      - value: "10.96.0.1"
     backendRefs:
     - name: my-service
       port: 8080
@@ -150,6 +150,17 @@ listeners for the Gateway make it much harder to understand the value at this
 stage. We're deferring port matching to focus on address matching for this
 iteration so that we can come back around to it separately once we've gathered
 more use case information.
+
+### CIDR AddressType
+
+When using `AddressType` as a component to `AddressMatch` it was considered to
+add a new type `CIDRAddress` which would allow matching against an entire
+subnet. This sounds good, but given a lack of concrete feedback on an
+immediate need for this in the original issue [#727][issue-727] it was decided
+that this could wait for now and just as easily be added later in a backwards
+compatible manner.
+
+[issue-727]:https://github.com/kubernetes-sigs/gateway-api/issues/727
 
 ## References
 
--- a/site-src/geps/gep-735.md
+++ b/site-src/geps/gep-735.md
@@ -63,6 +63,11 @@ type AddressMatch struct {
 	// Value of the address. The validity of the values will depend
 	// on the type and support by the controller.
 	//
+	// If implementations support proxy-protocol (see:
+	// https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) they
+	// are expected to respect the connection meta-data from proxy-protocol
+	// in the match logic implemented for these address values.
+	//
 	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
 	//
 	// Support: Core
--- a/site-src/geps/gep-735.md
+++ b/site-src/geps/gep-735.md
@@ -65,7 +65,7 @@ type AddressMatch struct {
 	//
 	// If implementations support proxy-protocol (see:
 	// https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) they
-	// are expected to respect the connection meta-data from proxy-protocol
+	// must respect the connection metadata from proxy-protocol
 	// in the match logic implemented for these address values.
 	//
 	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
--- a/site-src/geps/gep-735.md
+++ b/site-src/geps/gep-735.md
@@ -1,0 +1,159 @@
+# GEP-735: TCP and UDP addresses matching
+
+* Issue: [#735](https://github.com/kubernetes-sigs/gateway-api/issues/735)
+* Status: Provisional
+
+## TLDR
+
+Spec for matching source and destination addresses on L4 APIs.
+
+## Goals
+
+- add matching rules for address to `TCPRoute`
+- add matching rules for address to `UDPRoute`
+- intentionally avoid type definitions that would make it hard to expand later
+
+## Non-Goals
+
+- define rules for port matching
+
+## Introduction
+
+While `TCPRoute` and `UDPRoute` currently support custom matching extensions,
+there is desire among the community to include some "fundamental" matching
+options in the spec that cover the most common requirements. In this GEP we
+request address matching for these APIs in order to support a standard
+for some of the commonplace setups of gateway implementations. Matching is
+intended to be covered for both _source_ and _destination_ to enable a finer
+level of tuning options for L4 traffic routing at a level below the `Gateway`.
+
+## API
+
+The API changes include the following new types:
+
+- `NetworkMatch` to indicate the IP for address matching
+- `TrafficMatches` to configure matching according to network address
+
+These types enable the address matching required, with some active
+considerations about how to leave these open ended for later expansion.
+
+### NetworkMatch Type
+
+A new `NetworkMatch` type provides the targeting mechanism for match inclusion
+of a given network address:
+
+```go
+type NetworkMatch struct {
+	// Address is an IP address which refers to either an IPv4 or IPv6 address.
+	//
+	// Examples: `127.0.0.1`, `::1`
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=45
+	Address string `json:"address"`
+}
+```
+
+The purpose of calling this `NetworkMatch` and wrapping the single `IP` attribute in
+a structure is to enable forward expansion in case later we're interested in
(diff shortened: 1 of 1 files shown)


apis/v1alpha2/tcproute_types.go
```diff
@@ -59,6 +59,18 @@ type TCPRouteStatus struct {
...
+	// and ports. Any traffic that matches the rule should be routed to the
+	// configured BackendRefs accordingly. Implementations should consider
+	// matches an exclusive configuration: if any traffic does NOT match
+	// according to the given TrafficMatches configuration then the traffic
+	// should not be handled by this route.
+	//
+	// Support: Extended
+	//
+	// +optional
+	Matches *TrafficMatches `json:"matches"`
```
howardjohn, line 72: One concern I have - what if I want to do this for HTTP/TLS/etc? It seems like today we have HTTPRoute as a superset of TCPRoute; this would be a change from that direction.

In particular, we have a requirement for port matching on HTTP.

shaneutt: I did consider this but I was wondering if it might be a bit out of scope considering the conversation thus far in #735, but it seems to me we can just as easily add `TrafficMatches` to `HTTPRouteMatch` as an optional field if we like, no?

howardjohn: I think its very reasonable to defer this though

shaneutt: :heavy_check_mark:

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,202 @@
...
+
+```go
+type NetworkMatch struct {
+	// Address is an IP address which refers to either an IPv4 or IPv6 address.
+	//
+	// Examples: `127.0.0.1`, `::1`
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=45
+	Address string `json:"address"`
```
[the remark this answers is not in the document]
shaneutt: Just want to make sure I'm clear: are you suggesting:

a) type for IP addresses (v4, v6)
b) type for address (e.g. can include hostnames)
c) both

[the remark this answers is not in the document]
robscott: I think the idea of a "NamedAddress" here makes sense, similar to the address type we have on Gateway. I don't think we'd need to distinguish between v4 and v6 addresses though, but don't feel too strongly about that.

robscott (commented):
Thanks for your work on this @shaneutt!

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,202 @@
...
+
+```go
+type PortMatch struct {
+	// Port is a network port.
+	//
+	// Examples: `80`, `443`
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=5
+	Port PortNumber `json:"port"`
```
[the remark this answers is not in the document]
shaneutt: Whoops yes it was, but I've decided to remove Port from this GEP anyhow due to lack of real use cases to support it in light of listener ports on gateways.

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,202 @@
...
+}
+```
+
+However for this iteration the goal is to keep the initial spec simple and to
+the immediate goals so only `Address` is intended for inclusion until further
+requirements develop.
+
+### PortMatch Type
+
+a new `PortMatch` type provides the targeting mechanism for match inclusion of
```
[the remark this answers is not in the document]
shaneutt: port was removed from the draft in favor of addresses only, so this is gone now.

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,202 @@
...
+	// connections, then 80% of connections must be rejected instead.
+	//
+	// Support: Core for Kubernetes Service
+	// Support: Custom for any other resource
+	//
+	// Support for weight: Extended
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	BackendRefs []BackendRef `json:"backendRefs,omitempty"`
```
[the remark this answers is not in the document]
shaneutt: Sounds good, I've just gone ahead and removed this entirely.

site-src/geps/gep-735.md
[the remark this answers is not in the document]
shaneutt: let me know your thoughts f82c90bf0b696ec7ca333a2625e34d17f8a5d472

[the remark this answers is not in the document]
shaneutt: 1462088fdeada16de4ed7f624f465b5aa0d103e0 changes this to support both `IPAddress` and `NamedAddress` options, let me know your thoughts :bow:

tokers:
Another useful route match type is the TLS matching (according to the SNI), is this also in the plan?

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,170 @@
...
+	// Value of the address. The validity of the values will depend
+	// on the type and support by the controller.
+	//
+	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Value string `json:"value"`
+}
+```
```
Miciah, line 69: Have you considered making `AddressMatch` a [union type](https://github.com/kubernetes/enhancements/blob/0e4d5df19d396511fe41ed0860b0ab9b96f46a2d/keps/sig-api-machinery/1027-api-unions/README.md), with `Type` being the discriminator?  Having a generic `Value` field looks odd in a Kubernetes API, and it forfeits the opportunity to enforce stricter validation.

shaneutt: I had not as I was very intentionally trying to make this similar to `GatewayAddress` so that we didn't end up with two different ways of describing an address, do you have any thoughts on that?

shaneutt:
> Another useful route match type is the TLS matching (according to the SNI), is this also in the plan?

For the purposes of this KEP only L4 matching was originally intended (re #735) so while it may be reasonable to add something like this in a later iteration, in an effort to keep this GEP simple and succinct and avoid a longer review process I would like to avoid adding it at this stage.

hbagdi:
> Another useful route match type is the TLS matching (according to the SNI), is this also in the plan?

Gateway API already supports that. Please take a look at TLSRoute.

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,170 @@
...
+	// If NamedAddress is used this is a custom and specific value for each
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
+	// +kubebuilder:default=IPAddress
+	Type *AddressType `json:"type,omitempty"`
```
hbagdi, line 58: Please add conformance levels for these features.

site-src/geps/gep-735.md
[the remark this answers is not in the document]
hbagdi: NamedAddress feels a bit like ImplementationSpecific in path match (which we got rid off).
@robscott @howardjohn Does NamedAddress by itself (without the additional context of the provider) enough here?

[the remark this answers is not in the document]
robscott: That's a good point, NamedAddress is going to have a different meaning to most implementations. That's easier to solve on Gateways because each will only be implemented by a single controller. This feels similar in nature to RegEx path matching - each implementation will have a different take on it, but it means something conceptually similar. 

In most cases you would not want to specify a RegEx match on a Route implemented by multiple controllers unless you knew they supported the same variation of RegEx. I think this would come with the same constraints. Although that's not ideal, I do think it's a sufficiently valuable concept to include. We likely need to document the limitations of these implementation-specific types though.

site-src/geps/gep-735.md
Miciah: Ah, I can see how making `AddressMatch` similar to `GatewayAddress` could reduce confusion if the user wanted to specify the same IP address or named address on both the listener and an address match.  On the other hand, `AddressMatch` is missing `Hostname`, the semantics of `NamedAddress` across the two types gets a little funny, and I'm wary of making things appear alike superficially when they are fundamentally different.  

Elaborating on what I mean by funny semantics, do the names that the user can specify for a `NamedAddress` in the listener refer to the same set of things as the names that the user can specify for a `NamedAddress` in a source match?  What about for destination matches  Is this relationship (or lack thereof) entirely up to the implementation to define?  

(I now kind of wish we had used a union type for `GatewayAddress`, but it's probably too late to change that now without an extremely compelling need.)

robscott: For better or worse, we've used the `Type` and `Value` pair heavily throughout the API, including path, header, and query param matching. I believe HTTPRoute filters are the only place we've used a proper union type. Although union types are common in upstream Kubernetes, I think the pattern proposed here is also quite common. 

Since we're dealing with CRDs and there is not proper union support yet, we'd be stuck with adding additional webhook validation if we transitioned to a union type. My personal preference would be to continue with the proposed approach to more closely match the pattern we've used elsewhere in Gateway API.

Miciah: Using union types probably would have made sense in those other contexts (and in hindsight, it would have simplified defining support levels).  It makes even more sense for `AddressMatch` because we can specify OpenAPI validation to enforce that any value that the user puts in, say, the "address" field must be either an IPv4 or an IPv6 address (the following is incomplete and not tested, but this example should illustrate what I mean):
```yaml
sourceAddresses:
  items:
    oneOf:
    - properties:
        address:
          format: ipv4
    - properties:
        address:
          format: ipv6
    properties:
      address:
        type: string
    type: object
  type: array
```
Granted, CRD validation may have gaps (if I recall correctly, it admits an object definition that specifies a field that doesn't correspond to the discriminator value).  However, I figure that some validation is better than none.  

I think we'd need webhook validation anyway to validate that the user provided a valid regexp for the path, header, and query param matching, but regexp support is "Custom" and underspecified anyway, so it's fair to leave that up to implementors.  

Using a union type is just a suggestion for enabling slightly improved validation and clarity, so I won't press it if the goal is to make `AddressMatch` match `GatewayAddress`.

youngnick: I think the problem with both using a union type and having more interesting OpenAPI validation is that neither fits well through the CRD-validation-shaped hole we have available for now.

I also think that having some more complex validation in either the webhook or an exported library (as we've discussed briefly a few times) would be beneficial.

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,170 @@
...
+	// Type of the address, either IPAddress or NamedAddress.
+	//
+	// If NamedAddress is used this is a custom and specific value for each
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
```
youngnick, line 56: Another option that would allow more experimentation by implementations is to have this Type field be a named string type, that we provide standard constants for, but that can also accept a domain-prefixed string. The constants we export would have a defined support policy, anything domain-prefixed is ImplementationSpecific.

So, we would make `IPAddress` available now for sure (with Extended or even Core support), but we can leave `NamedAddress` out if we want - implementations are free to build out their own `implementation.io/NamedAdddress` behavior, and then bring it back to us for Extended support consideration.

This avoids the version-incompatibilty problem of adding values to an enum being a breaking change, at the cost of a bit more complexity, and less standardization.

(This is cribbed from the metav1.Condition `Type` field, which has similar concerns).

youngnick (commented):
I really like the shape of this, but have a couple of small things about the exact field semantics. In general, I think we should add enum fields very carefully, as they are tricky to manage, because adding values to an enum is a breaking API change (by the API guidelines).

site-src/geps/gep-735.md
shaneutt: Now added in c25e4f68885700d948c10c5534c7e52bd2dbf297 :+1:

site-src/geps/gep-735.md
shaneutt: I feel generally good about this idea, once I have gathered some more feedback above about `NamedAddress` I'll address this further and ideally implement it.

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,178 @@
...
+type AddressMatch struct {
+	// Type of the address, either IPAddress or NamedAddress.
+	//
+	// If NamedAddress is used this is a custom and specific value for each
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
+	//
+	// Support: Core
```
hbagdi, line 55:
```suggestion
	// Support: Core (IPAddress)
	// Support: Custom (NamedAddress)
```

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,178 @@
...
+      sourceAddresses:
+      - value: "192.168.1.1"
+      - value: "FE80::0202:B3FF:FE1E:8329"
+      - type: NamedAddress
+        value: "my-custom-name"
+      destinationAddresses:
+      - value: "10.96.0.1"
+    backendRefs:
+    - name: my-service
+      port: 8080
```
hbagdi, line 126: Any commentary on constitutes as the source/destination address when there are multiple hopes in the middle? For instance, is proxy protocol respected or not?

hbagdi:
LGTM once there is clarity on the proxy protocol.

site-src/geps/gep-735.md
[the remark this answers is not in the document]
robscott: >NamedAddress feels a bit like ImplementationSpecific in path match

@hbagdi I think there's a significant difference here. One of the key problems with "ImplementationSpecific" path matching was that any given implementation may have multiple extended kinds of path matching they'd like to support, and "ImplementationSpecific" could only really support one per implementation. With NamedAddress, I think there will be at most one interpretation per implementation. We already have a "NamedAddress" concept in the API that's fairly well defined, so it's not so much a question of if we should add the concept of a "NamedAddress", but if it makes sense here. 

> given the contention I'm leaning towards removing the NamedAddress for now 

@shaneutt my personal preference is that we leave this in place. I think it's a natural extension point and helps show the purpose of even having an address type here. Although it is something that will have a different meaning per implementation, that matches how NamedAddress already works elsewhere in the API, and seems conceptually similar to RegEx matching.

site-src/geps/gep-735.md
robscott: I do like the idea of vendor prefixed names here like `example.com/NamedAddress`, but I'm struggling to find additional use cases beyond NamedAddress. Since we already have NamedAddress elsewhere in the API (Gateway Spec+Status), it could get confusing to have different ways to represent that concept. If there were other types of addresses we might want to match on, I think domain prefixed types could still make sense though.

site-src/geps/gep-735.md
youngnick: Proxy protocol is an important point! I feel like I should send you a can opener for all the cans of worms you keep opening up. (It's a good thing!)

Personally, I feel like proxy protocol should really mean that the Gateway should be transparent and do minimal routing. So in the case that a Gateway has proxy protocol configured, that should mean that address-match routing like in this GEP should be disabled or ignored.

shaneutt: It's funny you mention proxy protocol because I had a feeling this was going to come up. I'm somewhat inclined to add documentation that states we aren't considering it at this time and allow future contributors to expand into that and change that as needed. Does that sound like a reasonable course of action, or what would be your preference?

youngnick: I think that saying that proxy protocol support is undefined for now, PRs welcomed, here's a tracking issue, would make sense to get this functionality in.

shaneutt: Sounds good, @hbagdi does this sound OK to you as a resolution?

hbagdi: Proxy protocol being undefined _could_ break implementations in the future when we do define it.
For example, consider an implementation has decided to read proxy headers from the previous hop and takes that into account for routing. Now, if the spec clarifies later on that this must not be the case - that's a breaking change.
Is this understanding correct?


A safer approach would be to explicitly state if proxy protocol should be respected for routing decisions or not.

shaneutt: I am persuaded, I have made an explicit point in the GEP about proxy protocol connection meta-data needing to be respected for match inclusions: 3ac0601fbbb1cb4b10f4b93e4b4f463f046c6130

site-src/geps/gep-735.md
shaneutt: @Miciah just wanted to double check this?

site-src/geps/gep-735.md
```diff
@@ -0,0 +1,184 @@
...
+	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
+	// +kubebuilder:default=IPAddress
+	Type *AddressType `json:"type,omitempty"`
+
+	// Value of the address. The validity of the values will depend
+	// on the type and support by the controller.
+	//
+	// If implementations support proxy-protocol (see:
+	// https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) they
+	// are expected to respect the connection meta-data from proxy-protocol
```
hbagdi, line 68:
```suggestion
	// must respect the connection metadata from proxy-protocol
```

site-src/geps/gep-735.md
shaneutt: Given Harry's new commit https://github.com/kubernetes-sigs/gateway-api/pull/932/commits/57f90a8b3a22bbf69155c2f42ec13eb16258a0f4 it seems like we can resolve this comment now.

site-src/geps/gep-735.md
Miciah: I agree, please disregard the suggestion to use a union type as far as this PR is concerned.

site-src/geps/gep-735.md
shaneutt: As per our discussion in the sync today I created https://github.com/kubernetes-sigs/gateway-api/issues/958 as a follow up to do this holistically (e.g. for `Gateway` as well) so that we can use what exists for now to move forward, but make sure to reconsider how this works before `v1beta1`.

robscott:
Thanks! Will leave final LGTM for someone else (cc @bowei @hbagdi @jpeach @youngnick).  

/approve
/hold cancel

candita:
@shaneutt what happens once the GEP is accepted?  Who completes the work?

shaneutt:
> @shaneutt what happens once the GEP is accepted? Who completes the work?
> 
> Update: I found the [PR](https://github.com/kubernetes-sigs/gateway-api/pull/976/files) in v1alpha2, and the revert of the GEP to provisional. Bummer.

Indeed, given the current state this GEP's actual status is closest to being considered `declined`. If you are feeling strongly about it, you could move it forward or even create a new one that follows up on it with a different implemention. Extra context: I created this GEP entirely because there was an open ask to do this, I did not have and still do not have a strong need for this functionality in my own implementations currently so I'm inclined to call it `declined` and wait for someone else who has a need to come along later.

PR #937, opened 2021-11-16
Adding details to Conformance GEP

robscott:
**What type of PR is this?**
/kind documentation
/kind gep

This is an addition to GEP #917 that provides a general structure for how we can define conformance tests. Although this could have potentially been a separate GEP, it felt sufficiently small to just bundle with the existing conformance GEP. Happy to split this out if it's easier though.

This is largely a follow up to the [earlier doc](https://docs.google.com/document/d/1QL-MpIVzqxe32Y2BZ_dYOB8zNsF9c4pnKEIB9ZLt118/edit), but it also adds a section to discuss how this could work with prepopulated Gateways. Open to other approaches here, this just seemed like it may be the simplest starting point.

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-917.md
+++ b/site-src/geps/gep-917.md
@@ -1,21 +1,21 @@
-# GEP-917: Gateway API conformance principles
+# GEP-917: Gateway API Conformance Testing
 
 * Issue: [#917](https://github.com/kubernetes-sigs/gateway-api/issues/917)
 * Status: Provisional
 
 ## TLDR
 
-This GEP outlines the reasons for and principles by which the Gateway API will
-design its conformance and testing regime.
+This GEP outlines the principles and overarching structure that Gateway API
+conformance tests will be built with.
 
 ## Goals
 
 - Record why we are doing conformance and what we hope to achieve with it
 - Record the success criteria for the conformance process and associated artifacts
+- Provide a set of guidelines for conformance test structure
 
 ## Non-Goals
 
-- Designing the conformance process at anything more than a very basic level
 - Designing the conformance testing framework or implementation
 - Designing the process for implementations to prove they are conformant
 
@@ -76,9 +76,7 @@ profiles it claims.
 
 ## Proposal
 
-### Conformance profiles principles
-
-#### Basics
+### Conformance Profiles
 
 The Gateway API project defines conformance purely in terms of what resources an
 implementation supports.
@@ -117,7 +115,7 @@ supports, the conformance is composable, and orthogonal for each object type.
 For example, it's valid to only support HTTPRoute and not TCPRoute, or TLSRoute
 and not HTTPRoute.
 
-#### Interaction with existing support levels
+### Interaction with existing support levels
 
 Conformance definitions will ensure that an implementation can provide all the
 features currently marked as "Core" support in the API documentation.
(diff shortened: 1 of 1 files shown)


robscott:
This is adding to Nick's GEP, so adding him to reviewers list as well:

/cc @youngnick

youngnick:
Yes, I agree with the approach here, I feel like it will put some pressure on those of us with currently-non-dynamic Gateway provisioning to get it built (😄 ), but the support for preloaded Gateways will cover us in the meantime.

/lgtm

site-src/geps/gep-917.md
```diff
@@ -126,23 +124,46 @@ Fields marked "Extended" support will eventually have conformance tests that
...
 to tell the testing framework what extended fields they support.
 
-#### Testing framework
+## Testing framework
+
+Conformance tests will initially be structured as a CLI that runs on existing
+Kubernetes clusters. This may be expanded to include a testing library that can
+be imported by implementations. This tool will output test results in a
+structured format that can be used to generate reports. Where possible it will
+share code with the existing ingress-controller-conformance test suite.
```
shaneutt, line 133: For your considerations: while I think a CLI is good in any case, I think structuring this as a Kubernetes `Job` as the sort of defacto way to run this may have some merit.

site-src/geps/gep-917.md
```diff
@@ -126,23 +124,46 @@ Fields marked "Extended" support will eventually have conformance tests that
...
+
+### Test Definitions
+
+Each set of tests will be built around a set of YAML manifests representing
+Gateway API resources. The tests will focus on how they expect a given
+controller to process those resources, configured via GatewayClass name. For
+example, tests may check to ensure that a controller correctly populates status,
+or that a request is appropriately routed to the desired backend with any
+relevant modifications.
```
shaneutt, line 142: This vaguely reminds me of https://github.com/kudobuilder/kuttl, which I worked on with some friends at a previous job. While I don't necessarily suggest we use `kuttl` itself it might be worth reviewing it for ideas and influences if nothing else.

site-src/geps/gep-917.md
```diff
@@ -126,23 +124,46 @@ Fields marked "Extended" support will eventually have conformance tests that
...
-of tests is a requirment for the API to graduate to `v1beta1` API stability level.
-(As per our upstream KEP). It's acceptable to begin with a small set and expand
-outward while the project is in beta, but having a full set of conformance tests
-that cover most of the API scope should be a requirement for declaring the API
-stable.
+* Docs URL
+* Function to determine feature enablement
 
-#### Certification process
+### Prepopulated Gateways
```
shaneutt, line 155: FWIW I'm personally in favor of calling these "unmanaged" gateways as I feel that language is more clear at a glance.

site-src/geps/gep-917.md
```diff
@@ -126,23 +124,46 @@ Fields marked "Extended" support will eventually have conformance tests that
...
+
+Some implementations of Gateway API are not able to provision arbitrary Gateways
+and instead are deployed with a 1:1 relationship to a Gateway. To ensure that
```
shaneutt, line 158: While it's basically a nitpick, I think the language here about ability might not be necessary because there's implementations like Istio which have the ability but also support "unmanaged" gateways.

My thoughts:
```suggestion
Some implementations of Gateway API support unmanaged `Gateway` resources 
that refer to an underlying server that was created by other means (e.g. `Helm`)
rather than provisioning and managing the server according to the `Gateway`
resource. To ensure that
```

site-src/geps/gep-917.md
```diff
@@ -1,21 +1,21 @@
...
+# GEP-917: Gateway API Conformance Testing
 
 * Issue: [#917](https://github.com/kubernetes-sigs/gateway-api/issues/917)
 * Status: Provisional
 
 ## TLDR
 
-This GEP outlines the reasons for and principles by which the Gateway API will
-design its conformance and testing regime.
+This GEP outlines the principles and overarching structure that Gateway API will
```
hbagdi, line 8:
```suggestion
This GEP outlines the principles and overarching structure that Gateway API
```

hbagdi:
This is reasonable and we should move quickly here. We will learn so much more once we have 2 or 3 conformance tests and a few implementations running those in their respective CIs.
/lgtm

site-src/geps/gep-917.md
robscott: Agree that that's a natural extension point, but I think even a k8s job would be based around a CLI.

site-src/geps/gep-917.md
robscott: Thanks! I took a look at kuttl and it has some pretty great patterns. I think our conformance tests will need to be somewhat more complex, specifically making potentially advanced requests and ensuring their responses match what we'd expect. This does seem like a great reference point for any of our tests that will start with a set of Kubernetes resources and expect status to be populated in a specific way.

site-src/geps/gep-917.md
robscott: Although "unmanaged" makes sense for the most common use case here, I think "prepopulated" is a bit more broad here. For example, some implementations may have very slow provisioning, or some manual steps during initial provisioning followed by automated updates afterwards. What this is really referring to is the ability for implementations to prepopulate Gateways, for any reason, when running conformance tests.

site-src/geps/gep-917.md
robscott: Good clarification - updated.

site-src/geps/gep-917.md
robscott: Updated, thanks!

robscott:
I think there's enough consensus on this one. Removing the hold.

/hold cancel

PR #939, opened 2021-11-17
Removing v1alpha1 generated code and validation

robscott:
**What type of PR is this?**
/kind cleanup

This removes the generated code and validation for v1alpha1. Since v1alpha1 was essentially frozen with the v0.3.0 release, it does not make sense to maintain the code generation and validation beyond that point. For reference, the [Kubernetes deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api) states that we need to maintain deprecated alpha API versions for _0_ releases.

Note that this intentionally leaves all of the following in place as they are required for our documentation:

- API types
- Examples
- Docs

```release-note
The v1alpha1 API version has been removed. 
```

/hold for consensus


robscott:
For more context on this PR, I'm getting back into code generation to support experimental fields as part of our updated versioning strategy. As I was updating code generation, it didn't make sense to workaround v1alpha1. Although it's still helpful to have the docs in place (maintained by this PR), the codegen is effectively wasted effort at this point in my opinion.

youngnick:
I agree that getting rid of this is probably for the best.

/lgtm
/hold for further review.

PR #940, opened 2021-11-18
Fixing make docs sed to work on BSD and Linux

crmejia:
**What type of PR is this?**
/kind bug

Is it currently not possible to run `make docs` on macos because it's running BSD's sed which has a different behavior than GNU's sed on the `-i` option.

```release-note
NONE
```


crmejia:
Here it is @jpeach

hack/make-docs.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Copyright 2014 The Kubernetes Authors.
```
jpeach, line 3: 2021 :joy:

hack/make-docs.sh
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
```
jpeach, line 20:
```bash
 readonly SCRIPT_ROOT="$(cd "$(dirname "${BASH_SOURCE}")"/.. && pwd)"
```

This sets `$SCRIPT_ROOT` to the parent of the "hack" directory. You can then use this to locate the paths you use below, so that running this script will work correctly from different working directories.

hack/make-docs.sh
```diff
@@ -0,0 +1,41 @@
...
+run::sed() {
+    local -r vers="$(sed --version < /dev/null 2>&1 | grep -q GNU && echo gnu || echo bsd)"
+    case "$vers" in
+        gnu) sed -i "$@" ;;
+        *) sed -i '' "$@" ;;
+esac
+}
+
+# Ensure site dir exists
+mkdir -p site
```
jpeach, line 31:
```suggestion
mkdir -p ${SCRIPT_ROOT}/site
```

and so on ...

hack/make-docs.sh
crmejia: It was a good year for me 😅

hack/make-docs.sh
crmejia: That is clever. Fixed.

hack/make-docs.sh
```diff
@@ -0,0 +1,43 @@
...
+    case "$vers" in
+        gnu) sed -i "$@" ;;
+        *) sed -i '' "$@" ;;
+esac
+}
+
+# Ensure site dir exists
+mkdir -p ${SCRIPT_ROOT}/site
+# Generate docs with mkdocs
+mkdocs build
```
jpeach, line 35: Oops, I missed this one ... you probably need to cd into SCRIPT_ROOT to find mkdocs.yml, so maybe it best to just do that once at the start 

sorry I didn't flag this at the start

crmejia: No, problem. Should I still keep the absolute paths on the commands after doing `cd`?

jpeach: Nah it's not necessary after that IMHO

robscott:
Thanks!

/ok-to-test
/lgtm
/approve

PR #944, opened 2021-11-19
Validate that Service backend refs have a port.

### LINKED ISSUE #941: Validate the port field in the BackendObjectReference
https://github.com/kubernetes-sigs/gateway-api/blob/f7b77a78fee608226bae51dfd93d7fb4af060fd7/apis/v1alpha2/object_reference_types.go#L124-L130

When a route specified a BackendObjectReference to a Service, the admission controller should validate that the `Port` field is present.

Add support for each route type:
- [x] HTTPRoute - https://github.com/kubernetes-sigs/gateway-api/pull/944
- [ ] TCPRotue
- [ ] UDPRoute
- [ ] TLSRoute

jpeach:
**What type of PR is this?**
/kind bug

Validate tha backend refs of type Service specify a port number.

Updates #941

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/validation/httproute.go
+++ b/apis/v1alpha2/validation/httproute.go
@@ -46,7 +46,38 @@ func ValidateHTTPRoute(route *gatewayv1a2.HTTPRoute) field.ErrorList {
 // validateHTTPRouteSpec validates that required fields of spec are set according to the
 // HTTPRoute specification.
 func validateHTTPRouteSpec(spec *gatewayv1a2.HTTPRouteSpec, path *field.Path) field.ErrorList {
-	return validateHTTPRouteUniqueFilters(spec.Rules, path.Child("rules"))
+	var errs field.ErrorList
+
+	errs = append(errs, validateHTTPRouteUniqueFilters(spec.Rules, path.Child("rules"))...)
+	errs = append(errs, validateHTTPRouteBackendServicePorts(spec.Rules, path.Child("rules"))...)
+
+	return errs
+}
+
+// validateHTTPRouteBackendServicePorts validates that v1.Service backends always have a port.
+func validateHTTPRouteBackendServicePorts(rules []gatewayv1a2.HTTPRouteRule, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+
+	for i, rule := range rules {
+		path = path.Index(i).Child("backendRefs")
+		for i, ref := range rule.BackendRefs {
+			if ref.BackendObjectReference.Group != nil &&
+				*ref.BackendObjectReference.Group != "" {
+				continue
+			}
+
+			if ref.BackendObjectReference.Kind != nil &&
+				*ref.BackendObjectReference.Kind != "Service" {
+				continue
+			}
+
+			if ref.BackendObjectReference.Port == nil {
+				errs = append(errs, field.Required(path.Index(i).Child("port"), "missing port for Service reference"))
+			}
+		}
+	}
+
+	return errs
 }
 
 // validateHTTPRouteUniqueFilters validates whether each core and extended filter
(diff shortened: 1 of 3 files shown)


apis/v1alpha2/validation/httproute.go
```diff
@@ -46,7 +46,38 @@ func ValidateHTTPRoute(route *gatewayv1a2.HTTPRoute) field.ErrorList {
...
+
+	for i, rule := range rules {
+		path = path.Index(i).Child("backendRefs")
+		for i, ref := range rule.BackendRefs {
+			if ref.BackendObjectReference.Group != nil &&
+				*ref.BackendObjectReference.Group != "" {
+				continue
+			}
+
+			if ref.BackendObjectReference.Kind != nil &&
```
tokers, line 69: It seems that we don't have some documents to mark the default value of `Kind` is "Service".

Quote from the website:

> (Optional)
Kind is kind of the referent. For example “HTTPRoute” or “Service”.

tokers (commented):
Just a little doubt.

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -431,3 +431,119 @@ func TestValidateHTTPPathMatch(t *testing.T) {
...
+		{
+			name:     "default groupkind with port",
+			errCount: 0,
+			route: &gatewayv1a2.HTTPRoute{
+				Spec: gatewayv1a2.HTTPRouteSpec{
+					Rules: []gatewayv1a2.HTTPRouteRule{{
+						BackendRefs: []gatewayv1a2.HTTPBackendRef{{
+							BackendRef: gatewayv1a2.BackendRef{
+								BackendObjectReference: gatewayv1a2.BackendObjectReference{
+									Port: portPtr(99),
```
hbagdi, line 465: Isn't this an invalid ObjectReference anyway because it is missing the required Name and Kind (Kind will defaulted _usually_)? I guess it is okay to unit test it.

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -431,3 +431,119 @@ func TestValidateHTTPPathMatch(t *testing.T) {
...
+			},
+		}, {
+			name:     "default groupkind with no port",
+			errCount: 1,
+			route: &gatewayv1a2.HTTPRoute{
+				Spec: gatewayv1a2.HTTPRouteSpec{
+					Rules: []gatewayv1a2.HTTPRouteRule{{
+						BackendRefs: []gatewayv1a2.HTTPBackendRef{{
+							BackendRef: gatewayv1a2.BackendRef{
+								BackendObjectReference: gatewayv1a2.BackendObjectReference{},
```
hbagdi, line 480: Can we at least add a Name to the BackendObjectReference object everywhere? Minor but helps with readability.

apis/v1alpha2/validation/httproute.go
jpeach: The default is set in the CRD:

https://github.com/kubernetes-sigs/gateway-api/blob/25607266669cc3616ed90404db4ff6e6d854a3b7/apis/v1alpha2/object_reference_types.go#L104-L106

apis/v1alpha2/validation/httproute_test.go
jpeach: Yeh, there's a case for that. I felt bad about assuming the fields were non-nil in the validator, since there was an argument that we can't guarantee that the API server has validated things. And then I wanted to test that case.

hbagdi:
The upstream issue is not specific to HTTP and yet this patch implements a solution specific to HTTP.
We could merge this in and develop a solution for other route types using this or we could go in all at once. I don't have a preference as long as the end result is achieved.

apis/v1alpha2/validation/httproute.go
jpeach: Yup, done.

jpeach:
> The upstream issue is not specific to HTTP and yet this patch implements a solution specific to HTTP. We could merge this in and develop a solution for other route types using this or we could go in all at once. I don't have a preference as long as the end result is achieved.

Yeh that's a good point. I switched this PR to "update" the issue rather than fix. We can come back and address other types in a general way.

PR #945, opened 2021-11-19
Implementing GEPs 726 and 922: Rewrites + Experimental CRDs

robscott:
**What type of PR is this?**
/kind feature

This is a huge PR that implements GEP #726 and #922. These are bundled because they're difficult to do independently. To add new features to Gateway API, they need to be experimental. To understand how to generate experimental CRDs, we needed to have some kind of experimental feature to test with.

A big thanks to @jpeach for the [inspiration](https://github.com/jpeach/gateway-api-install) here for CRD generation and showing this kind of custom approach was _relatively_ straightforward.

I spent a non-trivial amount of time trying to get these field level annotations to work as "markers" like the other kubebuilder tags. Unfortunately I could not make that work, so chose a different syntax `<>` so these values would still be parseable after crdgen was complete. If anyone wants to experiment with better approaches here, I'd be very open. 

Alternatives I'd considered here:
- Separate types files for experimental CRDs
- A relatively complex set of shell scripts

Although this approach is not perfect, I think it's preferable to either of the above.

This implements both #726 and #922, but both still require documentation updates. I'm leaving that out of this PR because it's already so big.

```release-note
* Gateway API now includes "Experimental" release channel
* Gateway API CRDs now include `gateway.networking.k8s.io/bundle-version` and `gateway.networking.k8s.io/channel` annotations
* URL Rewrites and Path redirects have been added as new "Experimental" features
```


robscott:
This is a big PR, adding a hold until we have consensus on the approach.

/hold

robscott:
Although I'd like to improve the reference docs, the rendering with this change actually isn't as bad as I'd originally expected. Search for "gateway:experimental" in the [deploy preview](https://deploy-preview-945--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/spec/) to take a look.

apis/v1alpha2/httproute_types.go
```diff
@@ -663,7 +686,43 @@ type HTTPRequestHeaderFilter struct {
...
+	// the "/foo" prefix replaced with "/bar" in matching requests.
+	PrefixMatchHTTPPathModifier HTTPPathModifierType = "ReplacePrefixMatch"
+)
+
+// HTTPPathModifier defines configuration for path modifiers.
+// <gateway:experimental>
+type HTTPPathModifier struct {
+	// Type defines the type of path modifier.
+	//
+	// <gateway:experimental>
```
hbagdi, line 709: Can we change the generator to automatically use the presence of this tag at a parent level to be propagated down?
Repeating this on each member seems error-prone to me.

I can't think of a situation where the struct is itself marked as experimental and any of the fields within would not be experimental, hence the suggestion.

pkg/generator/main.go
```diff
@@ -0,0 +1,158 @@
...
+
+	apiext "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/controller-tools/pkg/crd"
+	"sigs.k8s.io/controller-tools/pkg/loader"
+	"sigs.k8s.io/controller-tools/pkg/markers"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	ExperimentalMarker      = "<gateway:experimental>"
```
hbagdi, line 34: If you use `+gateway:experimental` here, does the parser remove it from the description of the field in the JSON schema?

What I'm thinking is if we can use the same string pattern as tags, while using the grep logic you have implemented below. It looks and _feels_ more readable.

apis/v1alpha2/httproute_types.go
youngnick: I agree, this would be really nice. But I don't know how we would ensure that the generated API docs had the tags then.

hack/verify-examples-kind.sh
```diff
@@ -53,7 +53,7 @@ kind create cluster --name "${CLUSTER_NAME}" --kubeconfig "${KUBECONFIG}" || res
 
 ##### Test v1alpha2 CRD apply and that invalid examples are invalid.
 # Install CRDs
-kubectl apply --kubeconfig "${KUBECONFIG}" -f config/crd/v1alpha2 || res=$?
+kubectl apply --kubeconfig "${KUBECONFIG}" -f config/crd/stable || res=$?
```
youngnick, line 56: Feels like we probably should still be testing that the experimental CRDs apply successfully? That does kind of imply that we should have a test corpus of YAMLs that tests the experimental CRD fields in addition to the stable ones as well though.

May be outside the scope of this PR, I guess.

youngnick (commented):
This LGTM overall, with just a couple of small queries.

apis/v1alpha2/httproute_types.go
robscott: Yeah it's actually unnecessary to have these tags on the child fields - the propagation @hbagdi describes already works when it comes to CRD generation. The problem is that it's not quite as clear when it comes to documentation, so I added this for _everything_ that was experimental at every level.

pkg/generator/main.go
robscott: >If you use +gateway:experimental here, does the parser remove it from the description of the field in the JSON schema?

Yep, unfortunately those get pulled out as "markers". Unfortunately I have not found a good way to access those. I tried to add support for a custom marker like this (https://github.com/kubernetes-sigs/controller-tools/pull/635), but it ended up being complex to get it working at every level I wanted (and represented significantly more code and complexity). If we ever want to upstream the concept of experimental fields, that's likely what we'd do, but I think that's something we can save for later.

apis/v1alpha2/httproute_types.go
```diff
@@ -663,7 +686,43 @@ type HTTPRequestHeaderFilter struct {
...
+	// +kubebuilder:validation:Enum=Absolute;ReplacePrefixMatch
+	Type HTTPPathModifierType `json:"type"`
+
+	// Substitution defines the HTTP path value to substitute. An empty value
+	// ("") indicates that the portion of the path to be changed should be
+	// removed from the resulting path. For example, a request to "/foo/bar"
+	// with a prefix match of "/foo" would be modified to "/bar".
```
hbagdi, line 716: I'm a bit hesitant about empty fields, especially string semantics. They are not usually the most intuitive, even more so in kube land.
Have we considered another "Type": "RemovePrefixMatch"?

apis/v1alpha2/httproute_types.go
```diff
@@ -663,7 +686,43 @@ type HTTPRequestHeaderFilter struct {
...
+	//
+	// <gateway:experimental>
+	// +kubebuilder:validation:MaxLength=1024
+	Substitution string `json:"substitution"`
+}
+
+// HTTPRequestRedirect defines a filter that redirects a request. At most one of
+// these filters may be used on a Route rule. This may not be used on the same
+// Route rule as a HTTPURLRewrite filter.
```
hbagdi, line 725:
```suggestion
// HTTPRequestRedirect defines a filter that redirects a request. This filter may not be used on the same
// Route rule as a HTTPURLRewrite filter.
```

And a question: Can we change the may to must to have a more defined and a narrower API that can be later relaxed?

pkg/generator/main.go
```diff
@@ -0,0 +1,158 @@
...
+	BundleVersionAnnotation = "gateway.networking.k8s.io/bundle-version"
+	ChannelAnnotation       = "gateway.networking.k8s.io/channel"
+)
+
+// This generation code is largely copied from sigs.k8s.io/controller-tools/pkg/crd/gen.go
```
hbagdi, line 39: Can you reference the commit so that we can navigate up and down the git history as that file changes upstream?

pkg/generator/main.go
```diff
@@ -0,0 +1,158 @@
...
+			log.Printf("generating %s CRD for %v\n", channel, groupKind)
+
+			parser.NeedCRDFor(groupKind, nil)
+			crdRaw := parser.CustomResourceDefinitions[groupKind]
+
+			// Inline version of "addAttribution(&crdRaw)" ...
+			if crdRaw.ObjectMeta.Annotations == nil {
+				crdRaw.ObjectMeta.Annotations = map[string]string{}
+			}
+			crdRaw.ObjectMeta.Annotations[BundleVersionAnnotation] = "v0.5.0-dev"
```
hbagdi, line 91: Not a biggie but if you can pull this out of code into the script invoking this code, that may help as we do releases.

hbagdi: I guess we will need to also change the PR URL for every release as well. Your call.

apis/v1alpha2/httproute_types.go
```diff
@@ -521,6 +523,15 @@ type HTTPRouteFilter struct {
 	// +optional
 	RequestRedirect *HTTPRequestRedirectFilter `json:"requestRedirect,omitempty"`
 
+	// URLRewrite defines a schema for a filter that responds to the
+	// request with an HTTP redirection.
+	//
+	// Support: Core
```
hbagdi, line 529: The type `HTTPURLRewriteFilter` says it is Extended. Should this be extended?

hbagdi:
The generated YAML has nothing that indicates that those fields are experimental: https://github.com/kubernetes-sigs/gateway-api/blob/7821e4e2db1c577f22af0a74cbba7ed1f801842e/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml#L977-L1014

Can we not strip out the experimental tags from the YAML?

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go
```diff
@@ -684,6 +743,14 @@ type HTTPRequestRedirectFilter struct {
 	// +optional
 	Hostname *Hostname `json:"hostname,omitempty"`
 
+	// Path defines a path redirect.
```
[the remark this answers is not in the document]

apis/v1alpha2/httproute_types.go
```diff
@@ -703,6 +770,28 @@ type HTTPRequestRedirectFilter struct {
 	StatusCode *int `json:"statusCode,omitempty"`
 }
 
+// HTTPURLRewriteFilter defines a filter that modifies a request during
+// forwarding. At most one of these filters may be used on a Route rule. This
+// may not be used on the same Route rule as a HTTPRequestRedirect filter.
+//
+// Support: Extended
```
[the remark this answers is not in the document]
robscott: Although it might help for documentation, it is unfortunately useless when it comes to CRD generation. Added just to be safe.

apis/v1alpha2/httproute_types.go
```diff
@@ -703,6 +770,28 @@ type HTTPRequestRedirectFilter struct {
...
+	//
+	// +optional
+	Hostname *Hostname `json:"hostname,omitempty"`
+
+	// Path defines a path rewrite.
+	//
+	// Support: Extended
+	//
+	// +optional
+	Path *HTTPPathModifier `json:"path,omitempty"`
```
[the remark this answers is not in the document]
robscott: Isn't needed, but probably helps with docs, added.

hack/verify-examples-kind.sh
robscott: Updated, thanks!

apis/v1alpha2/httproute_types.go
brianehlert: What if explicitly a NULL is the substitution value for the remove case?  Something explicit in the value as opposed to an additional type (too many types, while explicit to implement can cause usability issues and API friction)
Also, many JSON parsers and automation tools drop an empty value - thus preventing passing an empty value in a document through a tool chain, and YAML empty value = NULL

robscott: @hbagdi We discussed this at the community meeting yesterday. I suggested something similar to what @brianehlert said above, essentially trying to ensure that users had to specify at least something here, even if that was just "". I tested the current CRDs out with the following YAML in a redirect filter:

```yaml
path:
  type: ReplacePrefixMatch
```

That resulted in the following error:
```
error validating data: ValidationError(HTTPRoute.spec.rules[0].filters[0].requestRedirect.path): missing required field "substitution" in io.k8s.networking.gateway.v1alpha2.HTTPRoute.spec.rules.filters.requestRedirect.path
```

That error could be avoided by specifying an empty string:

```yaml
path:
  type: ReplacePrefixMatch
  substitution: ""
```

I think that ends up being pretty clear. 

The other key point that came up in the community meeting was that "replace with nothing" is a fairly common way to represent "remove" in a wide variety of tooling. Although I agree that we need to be careful about inferring meaning from unspecified values, I think we're safely avoiding that by requiring user input. 

With all that said, my personal preference is to avoid an additional unique filter type here and instead move forward with the ability to replace a prefix match with nothing.

pkg/generator/main.go
robscott: I wasn't sure where to place these. I thought about another package, but wasn't sure we wanted to commit to exporting them. For now I've just moved these up to the top in a `const` block. My current thought process is that these would be changed as part of every changelog PR, and then immediately changed back in a follow up PR post release. There's likely a better way to handle this, but I think that approach would technically work. Trying not to make this PR much bigger, but will work on a follow up for docs.

pkg/generator/main.go
robscott: Updated, thanks!

robscott:
>Can we not strip out the experimental tags from the YAML?

Good call, updated to leave them in place.

robscott:
Thanks for the great feedback on this @hbagdi and @youngnick! I think I've addressed everything now, PTAL.

youngnick:
/lgtm

I think the docs followup is a great idea. Thanks @robscott!

apis/v1alpha2/httproute_types.go
```diff
@@ -663,7 +686,42 @@ type HTTPRequestHeaderFilter struct {
...
+	// This type of modifier indicates that any prefix path matches will be
+	// replaced by the substitution value. For example, a path with a prefix
+	// match of "/foo" and a ReplacePrefixMatch substitution of "/bar" will have
+	// the "/foo" prefix replaced with "/bar" in matching requests.
+	PrefixMatchHTTPPathModifier HTTPPathModifierType = "ReplacePrefixMatch"
+)
+
+// HTTPPathModifier defines configuration for path modifiers.
+// <gateway:experimental>
+type HTTPPathModifier struct {
```
hbagdi, line 706: Do we need to clarify that path doesn't include query string?
I think this may be a potential source of confusion. Not a blocker.

hbagdi:
Great work Rob!
/lgtm

apis/v1alpha2/httproute_types.go
robscott: Agree, that's a good point. I'll add that in a follow up.

robscott:
Thanks for the reviews!

/hold cancel

PR #946, opened 2021-11-23
Validating HTTPRouteFilter type consistency

### LINKED ISSUE #942: Validate HTTP filter discriminated union consistency.
`HTTPRouteFilter` is a discriminated union where the `Type` field is the discriminator. The admission controller should validate that the only the matching field is present.

crmejia:
**What type of PR is this?**
/kind feature

This PR adds validates that a HTTPRouterFilter Type matches it's value.

Fixes #942

```release-note
NONE
```


crmejia:
Hey @jpeach, this PR looks to address  #942

Makefile
```diff
@@ -42,7 +42,7 @@ vet:
 
 # Run go test against code
 test:
-	go test -race -cover ./...
+	go test ./...
```
jpeach, line 45: Bad merge?

crmejia: whoops..

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +166,29 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
 	return allErrs
 }
+
+// validateHTTPRouteFilterTypeMatchesValues validates that the filter type matches the filter value.
+func validateHTTPRouteFilterTypeMatchesValues(filter *gatewayv1a2.HTTPRouteFilter, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+
+	switch filter.Type {
+	case gatewayv1a2.HTTPRouteFilterExtensionRef:
+		if filter.ExtensionRef == nil || filter.RequestHeaderModifier != nil || filter.RequestMirror != nil || filter.RequestRedirect != nil {
```
tokers, line 176: These if logics might not so feasible, if a new filter type is introduced, we have to change a few places, and it's easy to forget.

crmejia: I see what you mean. It's hard to mantain. I've researched alternatives:
1. I started by looking at ways to implement discriminated unions in Go and they usually point to using interfaces(see [Alternatives to sum types in Go](https://making.pusher.com/alternatives-to-sum-types-in-go/), [StackOverflow](https://stackoverflow.com/questions/21553398/best-practice-for-unions-in-go)). Correct me If I'm wrong but I don't think it's right to add interfaces to the API just to solve this problem. 
2. Something that occurred to me is to compare the json tags with the type but I think is  "hacky".

Do you have any ideas or tips for the correct implementation? I have a hunch that there is a more elegant way to solve this and I just didn't look in the right place.

tokers: If we can relax the constraints, say unused filter config can still be reserved there, than the switch statement can be simple.

jpeach: The best I could come up with was this:

```Go
func validateUnionField(discriminator string, name string, value interface{}) {
        var isNil bool

        switch value := value.(type) {
        case *gatewayv1a2.HTTPRequestHeaderFilter:   
                isNil = (value == nil)
        case *gatewayv1a2.HTTPRequestMirrorFilter:
                isNil = (value == nil)
        case *gatewayv1a2.HTTPRequestRedirectFilter:
                isNil = (value == nil)
        case *gatewayv1a2.LocalObjectReference:
                isNil = (value == nil)
        default:
                // Puke an error so someone updates this code ...
        
        }

        if discriminator == name {
                if isNil {               
                        // Filter spec is not set, but it should be.
                }
        } else {
                if !isNil {
                        // Filter spec is set, but it should not be.
                }
        }
}

validateUnionField(filter.Type, gatewayv1a2.HTTPRouteFilterExtensionRef, filter.ExtensionRef)
validateUnionField(filter.Type, gatewayv1a2.HTTPRouteFilterRequestHeaderModifier, filter.RequestHeaderModifier)
         
```

crmejia: Made the changes. Does this look good to you?

crmejia: I've relaxed the constraints to make the switch statement simpler. I think that `validateUnionField`  is more complex and still has the issue of referencing the types so new types need to be added.

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +166,31 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
 	}
 	return allErrs
 }
+
+// validateHTTPRouteFilterTypeMatchesValues validates that the filter type matches the filter value.
+func validateHTTPRouteFilterTypeMatchesValues(filter *gatewayv1a2.HTTPRouteFilter, path *field.Path) field.ErrorList {
```
robscott, line 171: It doesn't look like this is actually being called anywhere other than the test. I think the best place to fit this in would be in `validateHTTPRouteUniqueFilters`, with that func renamed to `validateHTTPRouteFilters`

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +166,31 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
+	var errs field.ErrorList
+
+	switch filter.Type {
+	case gatewayv1a2.HTTPRouteFilterExtensionRef:
+		if filter.ExtensionRef == nil {
+			return append(errs, field.Invalid(path.Child("filters"), path, "type of filter must match type of value"))
```
robscott, line 177: I think it would be helpful to have more specific error messages for each of these.
```suggestion
			return append(errs, field.Invalid(path.Child("filters"), path, "extensionRef must be specified for ExtensionRef filters"))
```

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -553,3 +553,89 @@ func TestValidateServicePort(t *testing.T) {
...
+		},
+		{
+			name:        "invalid nil type filter",
+			routeFilter: &gatewayv1a2.HTTPRouteFilter{},
+			errCount:    1,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			errs := validateHTTPRouteFilterTypeMatchesValues(tc.routeFilter, field.NewPath("spec").Child("rules"))
```
robscott, line 635: I think the actual path you'd want here would be something like `spec.rules[0].filters[0]`, but since this is just a test, an empty value is probably sufficient:

```suggestion
			errs := validateHTTPRouteFilterTypeMatchesValues(tc.routeFilter, field.NewPath(""))
```

robscott (commented):
Thanks for the work on this @crmejia!

apis/v1alpha2/validation/httproute.go
crmejia: Would `validateHTTPRouteSpec` be a better fit to add the test? I noticed that `validateHTTPRouteUniqueFilters` is being called there along with `validateHTTPRouteBackendServicePorts`.

robscott: Yeah not sure where the best place is, something needs to get called in `validateHTTPRouteSpec`. My suggestion would be to make that a generic function that validated a list of filters, including both uniqueness and that required fields were non-empty. That would also lead to another area for improvement, we would no longer need separate `validateHTTPRouteUniqueFilters` and `validateHTTPBackendUniqueFilters` functions.

jpeach, line 171: you need to call this from somewhere, I suggest from `validateHTTPRouteSpec

apis/v1alpha2/validation/httproute.go
jpeach, line 177:
```suggestion
			errs = append(errs, field.Invalid(path.Child("filters"), path, "type of filter must match type of value"))
```

Follow this pattern in the function, so that returning errs makes more sense.

crmejia: Oh yes, good catch. Thanks.

apis/v1alpha2/validation/httproute.go
crmejia: I did some refactoring. Please let me know what you think.

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +146,32 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
 	}
 	return allErrs
 }
+
+// validateHTTPRouteFilterTypeMatchesValue validates that each filter type matches
+//the filter value.
+func validateHTTPRouteFilterTypeMatchesValue(filters []gatewayv1a2.HTTPRouteFilter, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for _, filter := range filters {
+		switch filter.Type {
```
crmejia, line 155: That current error is coming from this switch. 
```
 apis/v1alpha2/validation/httproute.go:155:3: missing cases in switch of type v1alpha2.HTTPRouteFilterType: HTTPRouteFilterURLRewrite (exhaustive)
		switch filter.Type { 
```
But I'm not getting that error when I run this verification locally. Also, that switch statement has a `default` clause so it doesn't make sense to get this error. Any ideas what could be the reason?

apis/v1alpha2/validation/httproute.go
```diff
@@ -47,10 +47,17 @@ func ValidateHTTPRoute(route *gatewayv1a2.HTTPRoute) field.ErrorList {
 // HTTPRoute specification.
 func validateHTTPRouteSpec(spec *gatewayv1a2.HTTPRouteSpec, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
+	for i, rule := range spec.Rules {
+		errs = append(errs, validateHTTPRouteUniqueFilters(rule.Filters, path.Child("rules").Index(i))...)
+		errs = append(errs, validateHTTPRouteFilterTypeMatchesValue(rule.Filters, path.Child("rules").Index(i))...)
```
robscott, line 52: This is already a great simplification, but can you take it one step further and just have a generic `validateHTTPRouteFilters` func instead that covers both of these? Would save us from looping through the same list of filters 2x.

apis/v1alpha2/validation/httproute.go
robscott: It looks like this is complaining about a missing case for the `URLRewrite` filter type. That may have been added while you were working on this PR, so would've been easy to miss.

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -553,3 +558,89 @@ func TestValidateServicePort(t *testing.T) {
...
+	tests := []struct {
+		name        string
+		routeFilter []gatewayv1a2.HTTPRouteFilter
+		errCount    int
+	}{
+		{
+			name: "valid HTTPRouteFilterRequestHeaderModifier type filter with matching field",
+			routeFilter: []gatewayv1a2.HTTPRouteFilter{{
+				Type:                  gatewayv1a2.HTTPRouteFilterRequestHeaderModifier,
+				RequestHeaderModifier: &gatewayv1a2.HTTPRequestHeaderFilter{},
```
robscott, line 572: I wonder if we need to also make sure that the value of these fields is non empty? Even a simple check like `filter.RequestRedirect == nil || (gatewayv1a2.HTTPRequestHeaderFilter{}) == *filter.RequestRedirect)` might be helpful to ensure completely empty structs were also invalid.

robscott (commented):
Thanks for your work on this @crmejia!

apis/v1alpha2/validation/httproute.go
crmejia: Great catch by the linter. I guess the compiler evaluates the `default:` as correct but the linter is a bit more clever. 👍🏾

apis/v1alpha2/validation/httproute.go
crmejia: I kept them separate to have a clear boundary for each validation but I can see the inefficiency.  I'll simplify 👍🏾

apis/v1alpha2/validation/httproute_test.go
crmejia: Done. I had to use `reflect.DeepEqual` in some cases.

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +144,34 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
 	return allErrs
 }
+
+// validateHTTPRouteFilterTypeMatchesValue validates that each filter type matches
+//the filter value and that values are not empty.
+func validateHTTPRouteFilterTypeMatchesValue(filter gatewayv1a2.HTTPRouteFilter, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	switch filter.Type {
+	case gatewayv1a2.HTTPRouteFilterExtensionRef:
+		if filter.ExtensionRef == nil || (gatewayv1a2.LocalObjectReference{}) == *filter.ExtensionRef {
```
jpeach, line 154:
```suggestion
		if filter.ExtensionRef == nil || gatewayv1a2.LocalObjectReference{} == *filter.ExtensionRef {
```

I think that this is the correct pattern to apply here. We don't need deep equal, because if the filter is empty, the top-level fields will all equal the top level fields of the empty struct.

jpeach:
Once the deep equal issue is done, this LGTM, pending Rob's concurrence.

/lgtm
/hold

apis/v1alpha2/validation/httproute.go
crmejia: I've changed these and others. However, in the case of `HTTPRequestHeaderFilter` it cannot be compared as it contains slices. Here is the error : `struct containing []ref cannot be compared`. So what I did instead was to check that those slices are not nil, instead of using deep equal.

jpeach: OK, that sounds ok :+1:

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +143,35 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
+			errs = append(errs, field.Invalid(path.Child("filters"), path, "extensionRef must be specified for ExtensionRef filter"))
+		}
+	case gatewayv1a2.HTTPRouteFilterRequestHeaderModifier:
+		//Request header modifier contains slice whi
```
jpeach, line 157: Need to complete this comment :)

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +143,35 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
+		if filter.RequestHeaderModifier == nil || filter.RequestHeaderModifier.Add == nil || filter.RequestHeaderModifier.Set == nil || filter.RequestHeaderModifier.Remove == nil {
+			errs = append(errs, field.Invalid(path.Child("filters"), path, "requestHeaderModifier must be specified for RequestHeaderModifier filter"))
+		}
+	case gatewayv1a2.HTTPRouteFilterRequestMirror:
+		if filter.RequestMirror == nil || (gatewayv1a2.HTTPRequestMirrorFilter{}) == *filter.RequestMirror {
```
jpeach, line 162:
```suggestion
		if filter.RequestMirror == nil || gatewayv1a2.HTTPRequestMirrorFilter{} == *filter.RequestMirror {
```

and the other places in this function

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +143,35 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
+func validateHTTPRouteFilterTypeMatchesValue(filter gatewayv1a2.HTTPRouteFilter, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	switch filter.Type {
+	case gatewayv1a2.HTTPRouteFilterExtensionRef:
+		if filter.ExtensionRef == nil || (gatewayv1a2.LocalObjectReference{}) == *filter.ExtensionRef {
+			errs = append(errs, field.Invalid(path.Child("filters"), path, "extensionRef must be specified for ExtensionRef filter"))
+		}
+	case gatewayv1a2.HTTPRouteFilterRequestHeaderModifier:
+		//Request header modifier contains slice whi
+		if filter.RequestHeaderModifier == nil || filter.RequestHeaderModifier.Add == nil || filter.RequestHeaderModifier.Set == nil || filter.RequestHeaderModifier.Remove == nil {
```
jpeach, line 158:
```suggestion
		if filter.RequestHeaderModifier == nil || (filter.RequestHeaderModifier.Add == nil && filter.RequestHeaderModifier.Set == nil && filter.RequestHeaderModifier.Remove == nil) {
```
IIUC the filter is missing if it is nil, or if all of the slices are empty.

crmejia: removed the switch statement.

crmejia:
After some discussion with Rob regarding validation we decided to check for `nil` instead of empty and to check these two cases for each filter:
1. `if filter != nil  && filter.Type != filter return some error`
2. `if filter == nil  && filter.Type == filter return some other error`
Please let me know what y'all think.

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +143,43 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
 	}
 	return allErrs
 }
+
+// validateHTTPRouteFilterTypeMatchesValue validates that each filter type matches
+//the filter value and that values are not empty.
```
robscott, line 148: Tiny nit:
```suggestion
// validateHTTPRouteFilterTypeMatchesValue validates that only the expected fields are
// set for the specified filter type.
```

apis/v1alpha2/validation/httproute.go
```diff
@@ -166,3 +143,43 @@ func validateHTTPPathMatch(path *gatewayv1a2.HTTPPathMatch, fldPath *field.Path)
...
+	}
+	if filter.URLRewrite != nil && filter.Type != gatewayv1a2.HTTPRouteFilterURLRewrite {
+		errs = append(errs, field.Invalid(path, filter.URLRewrite, "must be nil if the HTTPRouteFilter.Type is not URLRewrite"))
+	}
+	if filter.URLRewrite == nil && filter.Type == gatewayv1a2.HTTPRouteFilterURLRewrite {
+		errs = append(errs, field.Required(path, "filter.URLRewrite must be specified for URLRewrite HTTPRouteFilter.Type"))
+	}
+	if filter.Type == "" {
+		errs = append(errs, field.Required(path, "filter.Type cannot be empty"))
+	}
```
robscott, line 183: This is actually already covered by CRD validation, no need to also have it here: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L497-L498

robscott (commented):
Thanks @crmejia! Just a couple tiny nits but otherwise LGTM.

robscott:
Thanks @crmejia! Will leave final LGTM for someone else, but this looks great.

/approve

jpeach (approved):
/lgtm

\o/

PR #951, opened 2021-12-01
Fixing BackendRef docs for ResolvedRefs condition

robscott:
**What type of PR is this?**
/kind bug
/kind documentation

Fixes a mistake in BackendRef docs that suggested ResolvedRefs should be set to "True" when it really should have been "False".

```release-note
Clarifies that ResolvedRefs should be set to "False" when a ReferencePolicy does not exist to allow a cross-namespace reference.
```


youngnick:
oof, nice catch.

/lgtm

PR #953, opened 2021-12-03
Fix ReferencePolicy docs about empty value in `spec.to.name`

nak3:
Current docs explains about `spec.to.name` as:

```
Name is the name of the referent. When unspecified or empty, this policy
refers to all resources of the specified Group and Kind in the local
namespace.
```

However, the ReferencePolicy with empty value (like below):

```shell
kubectl apply -f - <<EOF
kind: ReferencePolicy
metadata:
  name: allow-gateways-to-ref-secrets
  namespace: default
spec:
  from:
    kind: Gateway
    namespace: istio-system
  to:
  - group: ""
    kind: Secret
    name: ""
EOF
```

It gets the following error `Invalid value: "": spec.to.name in body should be at least 1 chars long` as:

```
The ReferencePolicy "allow-gateways-to-ref-secrets" is invalid: spec.to.name: Invalid value: "": spec.to.name in body should be at least 1 chars long
```
It should be fine to support only "unspecified" case. So, this patch removes "empty" from docs.

/kind documentation

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/referencepolicy_types.go
+++ b/apis/v1alpha2/referencepolicy_types.go
@@ -123,7 +123,7 @@ type ReferencePolicyTo struct {
 	// * Service
 	Kind Kind `json:"kind"`
 
-	// Name is the name of the referent. When unspecified or empty, this policy
+	// Name is the name of the referent. When unspecified, this policy
 	// refers to all resources of the specified Group and Kind in the local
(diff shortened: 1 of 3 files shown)


robscott:
Thanks @nak3!

/lgtm
/approve

PR #956, opened 2021-12-06
Introduce wildcardhostname field

### LINKED ISSUE #949: HTTPRequestRedirectFilter allows for wildcard hostnames
**What happened**:

Looking through the `HTTPRequestRedirectFilter` spec, the `Hostname` used has a default kubebuilder validation regex which would allow a user to specify a redirect of `*.foo.com` which might not make sense as a redirect hostname. 

Controllers could catch this and set a condition, we could update the field to use a new `hostname` type which removes the ability to have a wildcard in the value, or maybe there's a use-case I'm not thinking of that folks would want this? 

Hostnames with wildcards would be rejected. 

Here's a sample HTTPRoute:

```yaml
kind: HTTPRoute
metadata:
  name: kuard
  namespace: projectcontour
  labels:
    app: kuard
spec:
  parentRefs:
    kind: Gateway
    name: contour
  hostnames:
  - "local.projectcontour.io"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    filters:
      - requestRedirect:
          hostname: "*.stevesloka.com"
          scheme: https
          statusCode: 302
        type: RequestRedirect
    backendRefs:
    - kind: Service
      name: kuard
      port: 80
```

Right now Contour configures Envoy: 
```
$ curl -i local.projectcontour.io                                                                                                                                 
HTTP/1.1 302 Found
location: https://*.stevesloka.com/
vary: Accept-Encoding
date: Tue, 30 Nov 2021 17:37:35 GMT
server: envoy
content-length: 0
```

But doesn't route properly:
image

stevesloka:
**What type of PR is this?**
/kind bug

/kind api-change

Update the HTTPRequestRedirectFilter.Hostname to use a Hostname type which validates the value cannot
contain wildcard characters. Updates all other references which used Hostname to use WildcardHostname
which does allow for wildcard characters.

Signed-off-by: Steve Sloka <slokas@vmware.com>

Fixes #949

```release-note
Introduced WildcardHostname which allows for wildcard characters in relevant Hostname values. 
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -252,6 +252,24 @@ type RouteStatus struct {
 // +kubebuilder:validation:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 type Hostname string
 
+// PreciseHostname is the fully qualified domain name of a network host. This matches
+// the RFC 1123 definition of a hostname with 2 notable exceptions:
+//
+// 1. IPs are not allowed.
+// 2. A hostname may not be prefixed with a wildcard label (`*.`).
+//
+// Hostname can be "precise" which is a domain name without the terminating
+// dot of a network host (e.g. "foo.example.com").
+//
+// Note that as per RFC1035 and RFC1123, a *label* must consist of lower case
+// alphanumeric characters or '-', and must start and end with an alphanumeric
+// character. No other punctuation is allowed.
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+type PreciseHostname string
+
 // Group refers to a Kubernetes Group. It must either be an empty string or a
 // RFC 1123 subdomain.
 //
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -253,12 +253,10 @@ type RouteStatus struct {
 type Hostname string
 
 // PreciseHostname is the fully qualified domain name of a network host. This matches
-// the RFC 1123 definition of a hostname with 2 notable exceptions:
+// the RFC 1123 definition of a hostname with 1 notable exception that
+// numeric IP addresses are not allowed.
 //
-// 1. IPs are not allowed.
-// 2. A hostname may not be prefixed with a wildcard label (`*.`).
-//
-// Hostname can be "precise" which is a domain name without the terminating
+// PreciseHostname can be "precise" which is a domain name without the terminating
 // dot of a network host (e.g. "foo.example.com").
 //
 // Note that as per RFC1035 and RFC1123, a *label* must consist of lower case
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -740,7 +740,7 @@ type HTTPRequestRedirectFilter struct {
 	// Support: Core
 	//
 	// +optional
-	Hostname *Hostname `json:"hostname,omitempty"`
+	Hostname *PreciseHostname `json:"hostname,omitempty"`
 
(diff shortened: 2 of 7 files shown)


hbagdi:
Can we add a test case  'wildcard hostname on the redirect filter fails'? You can do so by adding it to hack/invalid-examples.

apis/v1alpha2/gateway_types.go
```diff
@@ -174,7 +174,7 @@ type Listener struct {
 	// Support: Core
 	//
 	// +optional
-	Hostname *Hostname `json:"hostname,omitempty"`
+	Hostname *WildcardHostname `json:"hostname,omitempty"`
```
jpeach, line 177: To me, it seems like it would be cleaner to change the type in the filter to `string` and add the custom  validation there. That's a smaller change and lets us keep the (IMHO nicer) `Hostname` typedef for the primary use case.

stevesloka: I agree about the smaller change @jpeach might be easier in the end and less user-breaking. 

@robscott does that work for you? I was originally going off of your suggestion of changing to a `WildcardHostname` as in the current commit.

robscott: We discussed this at the community meeting yesterday, meant to follow up sooner. I think at least @youngnick and I prefer using custom types like `Hostname` and `WildcardHostname` over a string so they're simpler to reuse in the future.

stevesloka:
@hbagdi I added an invalid test case for the redirect hostname.

apis/v1alpha2/gateway_types.go
jpeach: In this case, however, it's one place that has custom behavior, so introducing a new type for that seems like a lot. If there are other hostname fields that can't be wildcards, I'd prefer to add the type at that time.

Continuing to rathole on this, what to people think of the type name `PreciseHostname` for a hostname that cannot contain a wildcard label? That's the terminology we use in description, so maybe it makes sense as a type too?

jpeach: I'm sorry to keep bikeshedding this ... `WildcardHostname` implies the name is a wildcard, but it's not. It might be wildcard or precise, so I don't think it's a good choice to replace `Hostname` (which I think is fine).

stevesloka: I agree with @jpeach that `Wilcardhostname` does imply it's "only" a wildcard. The `PreciseHostname` might be a good route to use for the redirect field and leave `Hostname` as-is.

robscott: 👍 that works for me.

jpeach: Me too, thanks @stevesloka :)

youngnick: I like the use of `PreciseHostname` as well, that helps to cement that usage as a way of distinguishing them in the code as well as the docs.

youngnick:
LGTM once the field name changes are done as discussed.

robscott:
@stevesloka I was about to release 0.4.1, but realized this change also probably belongs in that release. Agree with @youngnick that this PR LGTM once that final name change gets in.

stevesloka:
@robscott sorry for loosing track of this one, I've updated the PR with the PreciseHostname. Thanks!

apis/v1alpha2/shared_types.go
```diff
@@ -252,6 +252,24 @@ type RouteStatus struct {
...
 type Hostname string
 
+// PreciseHostname is the fully qualified domain name of a network host. This matches
+// the RFC 1123 definition of a hostname with 2 notable exceptions:
+//
+// 1. IPs are not allowed.
+// 2. A hostname may not be prefixed with a wildcard label (`*.`).
+//
+// Hostname can be "precise" which is a domain name without the terminating
+// dot of a network host (e.g. "foo.example.com").
```
jpeach, line 262:
```suggestion
// the RFC 1123 definition of a hostname except that numeric IP addresses are not allowed.
```

stevesloka: 👍

robscott (commented):
Thanks!

/lgtm
/approve

jpeach (approved):
/lgtm

robscott:
Woops, we missed the hold on this.

/hold cancel

Issue #962, opened 2021-12-09
ReferencePolicy.Spec.From.Namespace is Optional


robscott:
**What happened**:
ReferencePolicy.Spec.From.Namespace is Optional.

**What you expected to happen**:
It should be required. 

**Anything else we need to know?**:
It looks like the accidental inclusion of `omitempty` on that field resulted in an optional field. Thanks to @howardjohn for catching this.

I'm not actually sure what the appropriate mitigation is here. Our [current versioning guidelines](https://gateway-api.sigs.k8s.io/geps/gep-922/#api-authors) do not allow us to tighten validation. To me this seems to very clearly be a bug, and making this field required could be considered a bug fix, but I think that would also break API conversion. We could consider a new  v1alpha2 release just to fix this, but that would also require an update to our versioning guidelines. If we want to fix this, we should probably do it before we hit beta. Alternatively, we could leave this indefinitely and update documentation to state that the lack of a namespace is invalid.

youngnick:
I don't feel like we have any good options here. Issuing a v1alpaha3 just for this feels bad, updating our guidelines feels bad, leaving this indefinitely feels even worse.

I agree that this is a bug, and we should fix it before v1beta1.

The least worst option seems to me to be updating the versioning guidelines to say "in alpha only, we can tighten validation if there's a bug in the validation, only". That is, we allow fixes to this kind of bug only, and then shout loudly about it in all our comms channels.

jpeach:
https://github.com/kubernetes-sigs/gateway-api/blob/4c34b35210b57fd223d551241330930494dd86b6/apis/v1alpha2/shared_types.go#L315-L318

Does this validation (esp. the minimum length) make the field required in practice? Or is the omitempty handled before the validation?

PR #963, opened 2021-12-13
Updating versioning guidelines to allow for bug fixes in patch releases

robscott:
**What type of PR is this?**
/kind documentation

This updates versioning guidelines to allow for bug fixes in patch releases and creates a new docs page to mirror the GEP content. (This is part of the mitigation for https://github.com/kubernetes-sigs/gateway-api/issues/962).

```release-note
Clarification that bug fixes may be broader than just loosened validation.
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -40,6 +40,7 @@ nav:
         API overview: concepts/api-overview.md
         Security Model: concepts/security-model.md
         Implementation Guidelines: concepts/guidelines.md
+        Versioning: concepts/versioning.md
     - Implementations: implementations.md
     - FAQ: faq.md
   - Guides:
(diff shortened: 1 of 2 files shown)


site-src/geps/gep-922.md
```diff
@@ -63,7 +63,8 @@ responsibilities to ensure we're providing a consistent experience.
 * MAY include the following changes to an existing API version with a new bundle
  **patch** version:
     * Clarifications to godocs
-    * Loosening validation to fix a bug (not to enable a new feature)
+    * Updating validation and/or CRD annotations to fix a bug (not to enable a
+      new feature)
```
hbagdi, line 67: Updates to CRDs and/or code to fix a bug

Why be specific to validation or annotations? (I could be missing something)

robscott: Good call, updated.

PR #964, opened 2021-12-13
Fixing bug where namespace could be unspecified in ReferencePolicy

### LINKED ISSUE #962: ReferencePolicy.Spec.From.Namespace is Optional

robscott:
**What type of PR is this?**
/kind bug
/kind api-change

This fixes a bug where namespace could be unspecified in ReferencePolicy

Fixes #962

```release-note
Namespace can no longer be left unspecified in ReferencePolicy. 
```

#963 needs to merge before this one
/hold for consensus

### DIFF
--- a/apis/v1alpha2/referencepolicy_types.go
+++ b/apis/v1alpha2/referencepolicy_types.go
@@ -104,7 +104,7 @@ type ReferencePolicyFrom struct {
 	// Namespace is the namespace of the referent.
 	//
 	// Support: Core
-	Namespace Namespace `json:"namespace,omitempty"`
+	Namespace Namespace `json:"namespace"`
 }
 
 // ReferencePolicyTo describes what Kinds are allowed as targets of the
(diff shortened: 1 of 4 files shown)


robscott:
I think we've got consensus on this one + #963 has now merged.

/hold cancel

PR #966, opened 2021-12-13
Adding GEP 957 for port matching support.

### LINKED ISSUE #957: GEP: Route Port Matching
**What would you like to be added**: support for port matching in routes

**Why this is needed**:

_Mesh_ (primary use case): for mesh, we are using a parentRef of `Mesh`, instead of Gateway. As such, there is no port matching available, since we do not get ports in the Gateway.listener like we do with Gateway cases. Matching on ports is quite common in mesh; for example, if I am defining some policy for google.com, I know I only need this for port 80 since google isn't serve on 8080.
_Ingress_ (secondary use case): This is similar to hostname attachment in many ways. Rather than requiring a bunch of `sectionNames`, I want to attach to multiple sections at once by filtering by port instead of explicit listener reference

Prior art: https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest has a port field

Design ideas:
- Add it to `TrafficMatches` from https://github.com/kubernetes-sigs/gateway-api/pull/932. This only applies to TCP/UDP route though; we would need this in HTTP and TLS as well
- Use `sectionName` (ie `sectionName: "80"`). This would work but is a bit odd since port is a string here. Would only apply to mesh
- Add a new `port` field to parentRef. For Gateway, this would act as an additional filter on listeners bound. For mesh, this would scope to that port.

cxhiano:
**What type of PR is this?**
/kind gep
/kind api-change

Proposal to support destination port matching.

Fixes #957 

```release-note
Add a new field `Port` to `ParentRef` for destination port matching
```

### DIFF
--- a/site-src/geps/gep-957.md
+++ b/site-src/geps/gep-957.md
@@ -1,0 +1,138 @@
+# GEP-957: Destination Port Matching
+
+* Issue: [#957](https://github.com/kubernetes-sigs/gateway-api/issues/957)
+* Status: Implementable
+
+## TLDR
+
+Add a new `port` field to ParentRef to support port matching in Routes.
+
+## Goals
+
+* Support port matching in routes based on the destination port number of the
+  request.
+
+## Non-Goals
+
+* Support port matching based on port name.
+
+## Introduction
+
+Port matching is a common service mesh use case where traffic policies/rules
+need to be applied to traffic to certain destination ports. For ingress, while
+the API today already supports attaching a route to a specific listener, it may
+be useful to support attaching routes to listener(s) on a specified port. This
+allows route authors to apply networking behaviors on a fixed port.
+
+## API
+
+The proposal is to add a new field `Port` to `ParentRef`:
+
+```go
+type ParentRef struct {
+  ...
+  // Port is the network port this Route targets. It can be interpreted
+  // differently based on the type of parent resource:
+  //
+  // Gateway: All listeners listening on the specified port that also support
+  // this kind of Route(and select this Route). It's not recommended to set
+  // `Port` unless the networking behaviors specified in a Route must
+  // apply to a specific port as opposed to a listener(s) whose port(s) may
+  // be changed.
+  // When both Port and SectionName are specified, the name and port of the
+  // selected listener must match both specified values.
+  //
+  // Implementations MAY choose to support other parent resources.
+  // Implementations supporting other types of parent resources MUST clearly
+  // document how/if Port is interpreted.
+  //
+  // For the purpose of status, an attachment is considered successful as
+  // long as the parent resource accepts it partially. For example, Gateway
+  // listeners can restrict which Routes can attach to them by Route kind,
+  // namespace, or hostname. If 1 of 2 Gateway listeners accept attachment from
+  // the referencing Route, the Route MUST be considered successfully
+  // attached. If no Gateway listeners accept attachment from this Route, the
+  // Route MUST be considered detached from the Gateway.
+  //
+  // Support: Core
+  //
+  // +optional
+  Port *PortNumber `json:"port,omitempty"`
+  ...
+}
+```
+
+The following example shows how an HTTPRoute could be applied to port 8000. In
+this example, the HTTPRoute will be attached to listeners foo and bar on port
+8000 but not listener baz on port 8080.
+```yaml
+kind: HTTPRoute
+metadata:
+  name: example
+  namespace: example
+spec:
+  parentRef:
+  - name: my-gateway
+    port: 8000
+  ...
+---
+kind: Gateway
+metadata:
+  name: my-gateway
+  namespace: example
+spec:
+  listners:
+  - name: foo
+    port: 8000
+    protocol: HTTP
+    ...
+  - name: bar
+    port: 8000
+    protocol: HTTP
+    ...
+  - name: baz
+    port: 8080
+    ...
+```
+
+The following example shows how a TCPRoute could be attached to an Mesh CRD to
+route all traffic in a service mesh whose original destination port is 8000 to
+port 8080 of service foo.
+```yaml
+kind: TCPRoute
+metadata:
+  name: example
+  namespace: example
+spec:
+  parentRef:
+  - name: my-mesh
+    group: example.io
+    kind: Mesh
+    port: 8000
+  rules:
+  - backendRefs
+    - name: foo
+      port: 8080
+```
+
+## Alternatives
+### 1. Use SectionName in ParentRef for port matching
+Port matching can be supported if SectionName accepts port numbers in addition
+to listener names. This approach results in a less explicit API when a ParentRef
+points to a resource that is not `Gateway`. For example, an implementation may
+attach a route to an `Mesh` CRD. In this case, it's less inituitive to set
+`ParentRef.SectionName` to `443` to express `route all traffic whose destination
+port is 443 to ...`. It also complicates the validation on SectionName in order
+to differentiate between a listener name and a port number.
+
+### 2. Update TrafficMatches to support port matching
+TrafficMatches was proposed in
+[gep-735](https://gateway-api.sigs.k8s.io/geps/gep-735/) to support TCP and UDP
+address matching. TrafficMatches can be extended to support port matching.
+TrafficMatches will need to be added to HTTPRoute/TLSRoute if the feature is
+desired there.
+
+While this proposal works for mesh, it may be confusing for ingress because a
+user can specify port matching behavior in a route that is incompatible with
+the listeners the route attaches to. For example, a user can specify a match
+on port 443 in a route while the route only attaches to a listener on port 80.


site-src/geps/gep-957.md
```diff
@@ -0,0 +1,77 @@
...
+* Issue: [#957](https://github.com/kubernetes-sigs/gateway-api/issues/957)
+* Status: Provisional
+
+## TLDR
+
+Support destination port matching.
+
+## Goals
+
+* Support destination port matching based on port number.
```
robscott, line 12: In the community meeting yesterday I think there was a bit of confusion around what destination port meant, this might be a bit clearer, but not sure:
```suggestion
* Support port matching in routes based on the destination port number of the request.
```

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,77 @@
+# GEP-957: Destination Port Matching
+
+* Issue: [#957](https://github.com/kubernetes-sigs/gateway-api/issues/957)
+* Status: Provisional
+
+## TLDR
+
+Support destination port matching.
```
robscott, line 8:
```suggestion
Add a new `port` field to ParentRef to support port matching in Routes.
```

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,77 @@
...
+  Port *PortNumber `json:"port,omitempty"`
+  ...
+}
+```
+
+## Alternatives
+### 1. Use SectionName in ParentRef for port matching
+Port matching can be supported if SectionName accepts port numbers in addition
+to listener names. This approach results in a less explicit API. It can also
+complicate the validation on SectionName.
```
robscott, line 65: Maybe add a bit more detail here such as an example of how it would complicate SectionName validation and/or how this would provide a poor user experience.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,77 @@
...
+## Introduction
+
+Port matching is a common service mesh use case where traffic policies/rules
+need to be applied to traffic to certain destination ports. For ingress, while
+the API today already supports attaching a route to a specific listener, it may
+be useful to support attaching routes to listener(s) on a specified port. For
+example, some users may only want to expose their application on ports 80 or 443
+on the Gateway.
+
+## API
```
robscott, line 27: Probably need a section in the GEP and/or godoc below describing the interaction between SectionName and Port. I'm not actually sure what the best approach is though.

We could say they were mutually exclusive, but that is likely unnecessarily restrictive. For ingress, one could say that they only wanted to attach to Gateway listener named foo if it was also listening on port 80. I'm assuming that wouldn't be super common, but it does seem fairly easy to interpret. For mesh, section name could represent some other implementation specific attribute such as a network name. In that case, it seems like specifying both network/sectionName and port could be quite helpful.

Given the potential usefulness, maybe we should suggest that these fields are complementary?

robscott (commented):
Thanks for your work on this @cxhiano!

site-src/geps/gep-957.md
cxhiano: Yeah, I don't have a strong opinion on this. Agreed that it may be better to make these fields complementary given that there is a reasonable interpretation on the meaning when both fields are present.

cxhiano:
Thanks for the comments! PR updated. PTAL!

robscott:
@hbagdi had a good comment in Slack asking if port would really be sufficient in `parentRef` or if we'd also want to add more fields in the future. Although it's hard to predict the future, I think it's at least possible that more fields could be added here. When I originally proposed `sectionName`, I remember at least 2 other alternatives being discussed:

1. A label selector that could enable selecting multiple parent resources. Although it's possible we may add this in the future, I think all the selectors in v1alpha1 helped show just how complex selectors can become. I personally don't think this is likely.
2. A jsonpath field that would enable matching arbitrary values in any resource. This would also provide a great deal of flexibility (and complexity), but I'm not currently aware of any use cases for it. I'd personally be hesitant to add in this kind of complexity.

It is possible that there are other potential extension points I'm missing here, but I can't think of any obvious ones. Even if we did want to add a field in the future, I don't think there's anything in this proposal that would prevent that from happening. With that said, very interested in what others think.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+## Alternatives
+### 1. Use SectionName in ParentRef for port matching
+Port matching can be supported if SectionName accepts port numbers in addition
+to listener names. This approach results in a less explicit API for mesh use
+cases. For example, it's less inituitive to create a route with
+`ParentRef.SectionName` set to `443` to express `route all traffic whose
+destination port is 443 to ...`. It also complicates the validation on
+SectionName in order to differentiate between a listener name and a port number.
```
jpeach, line 70: IIUC the only reason that section name isn't sufficient is where the parent is some resource that is not a Gateway. It would be good to call that out here because it's pretty important IMHO.

cxhiano: Rephrased this section. Please let me know if it is more clear now.

robscott:
@cxhiano we discussed this PR in more detail at today's community meeting. There was a lot of good feedback, I'll try to summarize it here, but of course anyone should feel free to add anything I miss.

1. It would be useful to have some more concrete examples of how this would be used for both Gateway and Mesh. This would help explain the rationale for the GEP and likely provide a better basis for the rest of the discussion. 
2. There's some concern that users using port number will more tightly couple the relationship between Gateways and Routes. When app devs attach Routes with a port number, it means that Gateway admins no longer have the flexibility to switch ports on the Gateway without also updating the Routes. I think this will be a reasonable tradeoff to make for some, but we should add some guidance to the GEP, and eventually documentation around when users would want to use one approach vs the other.
3. There was also some concern about this providing too many ways to do the same thing. I think that could largely be resolved by addressing the first 2 items (clarifying the rationale for adding this capability + clearly defining when each field should be used).

cxhiano:
Thanks for the feedback! Updated the API section to provide a recommendation on when `port` should be used. Also added two examples on how it can be used. (One for Gateway, the other for Mesh)

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+Add a new `port` field to ParentRef to support port matching in Routes.
+
+## Goals
+
+* Support port matching in routes based on the destination port number of the
+  request.
+
+## Non-Goals
+
+* Support port matching based on named port.
```
hbagdi, line 17:
```suggestion
* Support port matching based on port number.
```
Named port is ambiguous in this context (port names in k8s, section names, etc)

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Introduction
+
+Port matching is a common service mesh use case where traffic policies/rules
+need to be applied to traffic to certain destination ports. For ingress, while
+the API today already supports attaching a route to a specific listener, it may
+be useful to support attaching routes to listener(s) on a specified port. For
+example, some users may only want to expose their application on ports 80 or 443
+on the Gateway.
```
hbagdi, line 26: > be useful to support attaching routes to listener(s) on a specified port

Why might that be useful?
Please elaborate on that for the purpose of this GEP.

80/443 is not the best example here probably. Readers are not imaginative so picking a better example will bring more clarity.

hbagdi:
My concern is that this has the potential to complicate Ingress use-cases.
Route authors, upon seeing this, could become more defensive than necessary and put in port numbers.
Can we be more prescriptive about specific cases where this must be used?

site-src/geps/gep-957.md
cxhiano: Updated this section and also the field comment in API section to be more prescriptive. 

Admittedly while there are clear, concrete mesh use cases, it's only potentially useful for ingress(@robscott please chime in if I am missing a concrete ingress use case here). I am hoping that this GEP is a reasonable middle ground which enhances this API to be more inclusive towards use cases outside of ingress(such as mesh) while not making it more confusing/complicated for the existing ingress support.

pleshakov:
If the proposed `port` field can potentially complicate the API for the Gateway kind, perhaps it makes sense to consider adding extension options to the `parentRef`, so that third-party kinds can leverage them?

For example:
```yaml
kind: HTTPRoute
metadata:
  name: myroute
spec:
  parentRefs:
  - name: my-mesh
    group: example.io
    kind: Mesh
    options:
      port: "8080"
  hostnames:
  - example.com
. . .
```
where `options` is similar to https://github.com/kubernetes-sigs/gateway-api/blob/60e1838a179b4960ed5d0f825bf8ebfa189b8580/apis/v1alpha1/gateway_types.go#L360

robscott:
> My concern is that this has the potential to complicate Ingress use-cases. Route authors, upon seeing this, could become more defensive than necessary and put in port numbers. Can we be more prescriptive about specific cases where this must be used?

I think for Ingress this is adding one more optional component to a handshake. While it could lead to more defensive configuration from Route owners, I don't think that would always be a bad thing. I think it would be highly disruptive to app/route owners if a Gateway admin changed the ports that are being listened on. It seems unlikely that this would be a frequently changed attribute of a Gateway.

I think it's more likely that a Gateway admin would want to change a listener name than the listener port. For example, maybe a listener was initially focused on just app A, but they want to expand that listener to cover app A and B. Or maybe they want to split a shared listener into different ones? 

If a Route owner can attach to example.com on port 80, that may be more resilient (and easier to understand) than attaching to a relatively abstract section name. To be clear, I think section name will still be a useful mechanism for many orgs, I'm just starting to think that may become the advanced case, not the default.

robscott:
>If the proposed port field can potentially complicate the API for the Gateway kind, perhaps it makes sense to consider adding extension options to the parentRef, so that third-party kinds can leverage them?

Although we have that for TLS options, we've generally tried to avoid open ended configs like that. I think wherever we have a widely understand concept, we should try to surface that to a top level field if possible. On the other hand, if there are some use cases that rely on implementation specific concepts and terminology, it may be worth exploring a more open ended `options` field. @pleshakov do you have any use cases in mind beyond port?

pleshakov:
@robscott 
> , we've generally tried to avoid open ended configs like that. I think wherever we have a widely understand concept, we should try to surface that to a top level field if possible. 

That makes sense.

> On the other hand, if there are some use cases that rely on implementation specific concepts and terminology, it may be worth exploring a more open ended options field. @pleshakov do you have any use cases in mind beyond port?

Unfortunately, I don't have specific examples beyond the port. But I'm not too familiar with the service mesh use cases.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
+# GEP-957: Destination Port Matching
+
+* Issue: [#957](https://github.com/kubernetes-sigs/gateway-api/issues/957)
+* Status: Provisional
```
hbagdi, line 4: Implementable?

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+allows route authors to apply networking behaviors on a fixed port.
+
+## API
+
+The proposal is to add a new field `Port` to `ParentRef`:
+
+```go
+type ParentRef struct {
+  ...
+  // Port is the network port this resource targets. It can be interpreted
```
hbagdi, line 34: nit:
> Port is the network port this resource targets

What is `this`? Very ambiguous.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+  // differently based on the type of parent resource:
+  //
+  // Gateway: All listeners listening on the specified port that also support
+  // this kind of Route. It's only recommended to set `Port` when the networking
```
hbagdi, line 38:
```suggestion
  // this kind of Route (and select this Route). It's only recommended to set `Port` when the networking
```

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+  // behaviors specified in a Route should only apply to a specific port as
+  // opposed to listener(s) whose port(s) may get switched by Gateway admins.
+  // When both Port and SectionName are specified, the name and port of the
+  // selected listener must match both specified values.
+  //
+  // Implementations MAY choose to support other parent resources.
+  // If that is the case, they MUST clearly document how Port is
+  // interpreted.
```
hbagdi, line 46:
```suggestion
  // Implementations supporting other types of parent resources
  // MUST clearly document how/if Port is interpreted.
```

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+  //
+  // When both Port and SectionName are unspecified, this will reference the
+  // entire resource.
```
hbagdi, line 49: Recommend leaving this out. Port field is a more specific matcher and should refrain from general behavior guidance.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,140 @@
...
+  //
+  // For the purpose of status, an attachment is considered successful as
+  // long as the parent resource accepts it partially. For example, Gateway
```
hbagdi, line 52:
```suggestion
  // For the purpose of status, an attachment is considered successful if
  // at least one listener accepts the Route. For example, Gateway
```

hbagdi:
> I think for Ingress this is adding one more optional component to a handshake. While it could lead to more defensive configuration from Route owners, I don't think that would always be a bad thing. I think it would be highly disruptive to app/route owners if a Gateway admin changed the ports that are being listened on. It seems unlikely that this would be a frequently changed attribute of a Gateway.

It really comes down to abstractions that should be exposed to a human. I think abstracting away port at this layer is really hard and this GEP furthers this fact.
Like you say it is not a bad thing, it just increases the number of things a Gateway and a Route owner must cross-check.

I still think we should strongly advise users to not use this field for Ingress-like use-cases to unnecessarily avoid complicating the simpler use-cases. If we can add that bit, I'm good with this GEP.

site-src/geps/gep-957.md
cxhiano: I wonder if we can phrase this in a way that is agnostic to the type of parent resource here while the example following this general statement is specifically about Gateway/Listener?

site-src/geps/gep-957.md
cxhiano: Updated to "this Route"

cxhiano:
Thanks for the review!

> I still think we should strongly advise users to not use this field for Ingress-like use-cases to unnecessarily avoid complicating the simpler use-cases. If we can add that bit, I'm good with this GEP.

I updated the API documentation to say "It's not recommended to use `Port` unless...". Please let me know if that is sufficient.

site-src/geps/gep-957.md
```diff
@@ -0,0 +1,138 @@
...
+type ParentRef struct {
+  ...
+  // Port is the network port this Route targets. It can be interpreted
+  // differently based on the type of parent resource:
+  //
+  // Gateway: All listeners listening on the specified port that also support
+  // this kind of Route(and select this Route). It's not recommended to set
+  // `Port` unless the networking behaviors specified in a Route should only
+  // apply to a specific port as opposed to listener(s) whose port(s) may
+  // get switched by Gateway admins.
```
hbagdi, line 41:
```suggestion
  // this kind of Route(and select this Route). It's not recommended to set
  // `Port` unless the networking behaviors specified in a Route must
  // apply to a specific port as opposed to a listener(s) whose port may
  // be changed.
```

hbagdi (approved):

robscott:
Thanks @cxhiano!

/lgtm

PR #969, opened 2021-12-20
Initial Conformance Tests for Gateway API

robscott:
**What type of PR is this?**
/kind feature

This PR adds a very basic initial framework for conformance tests. Although there is still plenty of room for both improvement and expansion, I want to keep this initial starting point relatively small and straightforward. This PR is ready for full review, I'm especially interested in feedback on the general structure and concepts in this PR though.

```release-note
NONE
```


robscott:
At long last, I think these tests are ready for full review. They line up with the earlier conformance related GEPs and cover the following:

- GatewayClass
  - Status:
    - Accepted
- Gateway:
  - Provisioning (provisioning in advance is also supported)
  - Status:
    - Ready
- HTTPRoute:
  - Matching:
    - Multiple routes
    - Multiple rules
    - Path
    - Hostname
    - Header
  - Status:
    - Parents

robscott:
Adding a hold because this a big PR and want to have at least a couple LGTMs before merging:

/hold

conformance/manifests/base.yaml
```diff
@@ -0,0 +1,300 @@
...
+kind: Namespace
+metadata:
+  name: gateway-conformance-infra
+  labels:
+    gateway-conformance: infra
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: Gateway
+metadata:
+  name: gateway-same-namespace
```
hbagdi, line 11: Having more specific names with some semantic meaning like "shared-infra-gateway" would be better, especially as more and more tests use this.

conformance/tests/httproute.go
```diff
@@ -0,0 +1,254 @@
...
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace",
+	Manifests:   []string{"httproute-simple-same-namespace.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := v1alpha2.Namespace("gateway-conformance-infra")
+		routeNN := types.NamespacedName{Name: "gateway-conformance-infra-test", Namespace: string(ns)}
+		gwNN := types.NamespacedName{Name: "gateway-same-namespace", Namespace: string(ns)}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, gwNN, routeNN)
+
+		t.Run("Simple HTTP request should reach infra-backend", func(t *testing.T) {
+			t.Logf("Making request to http://%s", gwAddr)
```
hbagdi, line 46: Good starting point. We may need to have a CLI arg for gwAddr. Just leaving a note, no action is necessary yet.

conformance/utils/http/http.go
```diff
@@ -0,0 +1,111 @@
...
+type ExpectedRequest struct {
+	Host    string
+	Method  string
+	Path    string
+	Headers map[string]string
+}
+
+// MakeRequestAndExpectResponse makes a request with the given parameters and
+// verifies the response matches the provided ExpectedResponse.
+func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwAddr string, expected ExpectedResponse) {
```
hbagdi, line 50: Have you considered having retry here in case the resources are being stabilized while this runs?
Or should these be run only once the status of resources have been verified (and thus ensures a deterministic state).

robscott: These tests wait for the appropriate ready status to be set on the Route and Gateway. I think that should be interpreted as the Gateway being ready to handle these requests. Although it's possible that some implementations are setting status prior to proxies being completely updated, I think we should try to keep conformance tests as tight as possible initially so that users can trust a ready status across all conformant implementations.

conformance/tests/httproute.go
robscott: I'd been working with the assumption that Gateway status addresses would always be set on a ready Gateway. Would that not be possible for some of the implementations relying on provisioning Gateways in advance of running these tests?

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	defer resp.Body.Close()
+
+	if d.Debug {
+		dump, dumpErr := httputil.DumpResponse(resp, true)
+		if err != nil {
+			return nil, nil, dumpErr
+		}
+
+		fmt.Printf("Received response:\n%s\n\n", formatDump(dump, "< "))
```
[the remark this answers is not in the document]
robscott: I think that's already the case? One is "Sending Request: ...", and this is "Received Response: ...", any changes you'd recommend?

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,160 @@
...
+	"time"
+
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// RoundTripper is an interface used to make requests within conformance tests.
+// This can be overridden with custom implementations whenever necessary.
+type RoundTripper interface {
+	CaptureRoundTrip(Request) (*CapturedRequest, *CapturedResponse, error)
+}
```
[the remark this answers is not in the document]
robscott: These are great questions, I struggled with the appropriate interface here, but ended up with something that was pretty similar to what [ingress-controller-conformance implemented](https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/master/test/http/http.go#L39-L66). I think there are 3 main reasons I avoided the net/http interfaces:

1. I want to be able to support more protocols than HTTP. I actually don't know how to test UDP, but TCP could theoretically recycle the same interface.
2. I want tests to be able to specify a request to example.com/foo and know that it will get routed to the appropriate Gateway automatically. That could technically be done by modifying/copying a http request that was passed in, but feels messier. We could also just make each test case specify this as they structure a request.
3. I wanted to use the same echoserver image as ingress-controller-conformance so we didn't have to maintain a separate one for this. That requires a struct that matches the output from that container.

Open to different approaches here, but that's how I landed up with the current approach.

conformance/manifests/httproute-cross-namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,160 @@
...
+
+	client := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+
+	method := "GET"
+	if request.Method != nil {
+		method = string(*request.Method)
+	}
+	req, err := http.NewRequestWithContext(context.TODO(), method, request.URL.String(), nil)
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

conformance/utils/suite/suite.go
```diff
@@ -0,0 +1,96 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,160 @@
...
+			return nil, nil, dumpErr
+		}
+
+		fmt.Printf("Received response:\n%s\n\n", formatDump(dump, "< "))
+	}
+
+	body, _ := ioutil.ReadAll(resp.Body)
+
+	// we cannot assume the response is JSON
+	if isJSON(body) {
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

conformance/manifests/base.yaml
robscott: I've added some comments at the top of this file describing the key resources, but unfortunately don't have great ideas for how to make the names of the Gateways more descriptive right now. At this point, the only real variation between them is where they accept routes to bind from:

- gateway-same-namespace (only supports route in same ns)
- gateway-all-namespaces (supports routes in all ns)
- gateway-backend-namespaces (supports routes in ns with backend label)

In the future when we add resources with different protocols, multiple listeners, etc, naming may become a bit more clear. Open to ideas though.

conformance/tests/httproute.go
```diff
@@ -0,0 +1,254 @@
...
+		testCases := []http.ExpectedResponse{{
+			Request:   http.ExpectedRequest{Path: "/"},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request:   http.ExpectedRequest{Path: "/example"},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request:   http.ExpectedRequest{Path: "/", Headers: map[string]string{"version": "one"}},
```
howardjohn, line 85: Header usage doesn't work I think. By the time we get here go will have canonicalized the header name. IE should be checking `Version` (or some other fix)

conformance/utils/kubernetes/helpers.go
```diff
@@ -0,0 +1,251 @@
...
+
+// HTTPRouteMustHaveParents waits for the specified HTTPRoute to have parents
+// in status that match the expected parents. This will cause the test to halt
+// if the specified timeout is exceeded.
+func HTTPRouteMustHaveParents(t *testing.T, client client.Client, routeName types.NamespacedName, parents []v1alpha2.RouteParentStatus, seconds int) {
+	t.Helper()
+
+	var actual []v1alpha2.RouteParentStatus
+	waitFor := time.Duration(seconds) * time.Second
+	waitErr := wait.PollImmediate(5*time.Second, waitFor, func() (bool, error) {
```
howardjohn, line 167: poll every 5s is needlessly slow for a lot of implementations, IMO. I see about 500ms on my env

conformance/utils/kubernetes/helpers.go
```diff
@@ -0,0 +1,251 @@
...
+		if reflect.DeepEqual(aParent.ParentRef.Kind, eParent.ParentRef.Kind) {
+			t.Logf("ParentRef.Kind doesn't match")
+			return false
+		}
+		if aParent.ParentRef.Name != eParent.ParentRef.Name {
+			t.Logf("ParentRef.Name doesn't match")
+			return false
+		}
+		if aParent.ParentRef.Namespace != eParent.ParentRef.Namespace {
+			t.Logf("ParentRef.Namespace doesn't match")
```
howardjohn, line 207: nit: helpful to log what values were actually received vs expected

conformance/utils/kubernetes/helpers.go
```diff
@@ -0,0 +1,251 @@
...
+
+		actual = route.Status.Parents
+
+		return parentsMatch(t, parents, actual), nil
+	})
+	require.NoErrorf(t, waitErr, "error waiting for HTTPRoute to have parents matching expectations")
+}
+
+func parentsMatch(t *testing.T, expected, actual []v1alpha2.RouteParentStatus) bool {
+	if len(expected) != len(actual) {
```
howardjohn, line 182: puttting `t.Helper()` may help logs

conformance/utils/kubernetes/helpers.go
```diff
@@ -0,0 +1,251 @@
...
+	})
+	require.NoErrorf(t, waitErr, "error waiting for Gateway to have at least one IP address in status")
+	return ipAddr, waitErr
+}
+
+// HTTPRouteMustHaveParents waits for the specified HTTPRoute to have parents
+// in status that match the expected parents. This will cause the test to halt
+// if the specified timeout is exceeded.
+func HTTPRouteMustHaveParents(t *testing.T, client client.Client, routeName types.NamespacedName, parents []v1alpha2.RouteParentStatus, seconds int) {
+	t.Helper()
```
howardjohn, line 163: Need this within `PollImmediate` to capture the logs from there, if that was the intent

howardjohn: Actually that makes it log in wait.go code ... not sure how to get the call of `HTTPRouteMustHaveParents` in the output

conformance/utils/kubernetes/helpers.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		if reflect.DeepEqual(aParent.ParentRef.Kind, eParent.ParentRef.Kind) {
+			t.Logf("ParentRef.Kind doesn't match")
+			return false
+		}
+		if aParent.ParentRef.Name != eParent.ParentRef.Name {
+			t.Logf("ParentRef.Name doesn't match")
+			return false
+		}
+		if aParent.ParentRef.Namespace != eParent.ParentRef.Namespace {
```
howardjohn, line 206: `!=` is not effective on pointer types

conformance/utils/kubernetes/helpers.go
robscott: Good call, updated to 1 second for now, can lower further if needed.

conformance/tests/httproute.go
robscott: Good catch, updated to check "Version" for now, but will work on a better fix that makes the header check case insensitive.

conformance/manifests/base.yaml
bowei: I know it's more verbose, but does it make sense to repeat the YAML across test cases (i.e. make each case hermetic) rather than trying to deduplicate? I know it's kind of a pain, but sharing the setup can result in coupling between test cases and the creation of a complex "base" case that makes it harder to understand the relationship later on.

conformance/tests/httproute.go
bowei: My recommendation from an organization perspective is to try to avoid nesting by resource type in the test cases, as we will end up having conformance things that are split between gwc, gw, and *route behaviors.

instead put that in the name of the test go file. 
Also, if we could have the tests be encapsulated into a file each makes it easier to understand organizationally:

HTTPRouteSimpleSameNamespace_test.go:

```
func init() {
  ConformanceTests = append(ConformanceTests, &ConformanceTest{ ...
  })
}
```

robscott: @bowei thanks for the suggestion! I've reworked this to use that structure and I think it will make it much easier to add future test cases + understand existing ones.

conformance/utils/kubernetes/helpers.go
robscott: Updated, thanks!

conformance/utils/kubernetes/helpers.go
robscott: Woops, missed this, thanks!

conformance/manifests/base.yaml
robscott: Yeah it's going to be tough to find the right balance here. For any route to work, it needs the following:

- Namespace
- Gateway
- Service
- Pods

Spinning all of those up and down for every test would quickly become quite time consuming. Alternatively if we were to leave them running indefinitely, we'd need each test case to be sure that it wasn't overlapping with resources defined by other test cases. This would also result in rather verbose yaml definitions for each test case. A final argument in favor of this shared set of base resources is that it enables conformance tests to be meaningful for implementations that are not currently able to support dynamically provisioning Gateways (this generally means there's a 1:1 relationship between Gateway and controller).

With all that said, more isolated tests would enable us to run tests in parallel which could significantly speed up a test run when there were a large number of tests in play. It would also be easier to understand each individual test case than our current approach that requires an understanding of the base resources.

I think we'll inevitably want to move towards a test suite that has more isolated test definitions, but fortunately that should be a fairly straightforward transition from this current state. I think that the current proposal is an acceptable starting point given the constraints above, but that we should try to avoid expanding the base set of resources, and in the future add test cases that are not entirely dependent on them.

conformance/utils/http/http.go
hbagdi: Okay. That's fine. We can get feedback and tweak this if needed.

conformance/tests/httproute-cross-namespace.go
```diff
@@ -0,0 +1,67 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
hbagdi, line 2: Can you keep a single test case and add the remaining ones in a separate PR? That would minimize the likelihood of shipping buggy tests. This PR is big as is.

conformance/tests/httproute-cross-namespace.go
```diff
@@ -0,0 +1,67 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteCrossNamespace)
+}
+
+var HTTPRouteCrossNamespace = suite.ConformanceTest{
+	ShortName:   "HTTPRouteCrossNamespace",
+	Description: "A single HTTPRoute in the gateway-conformance-web-backend namespace attaching to Gateway in other namespace",
```
hbagdi, line 39: If the description tells me that it is a negative or positive test, it is better - code and words must match up. Else, the reader has to read the code to figure out the purpose of the test.
```suggestion
	Description: "A single HTTPRoute in the gateway-conformance-web-backend namespace attaches to Gateway in other namespace",
```

conformance/tests/httproute-cross-namespace.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: HTTPRoute
+metadata:
+  name: httproute-cross-namespace
```
hbagdi, line 4:
```suggestion
  name: web-backend
```

We should avoid putting namespace/name/types in the Name or Namespace keys. That makes the whole thing much harder to parse and often unnecessarily.

shaneutt (approved):
Once other people's comments are resolved this looks like a great place to start for me. We'll be adding these to our test suite soon, the only thing I was really concerned about going into this was making sure there wasn't any friction for testing "unmanaged mode" (existing gateways) but I don't think what's here will cause any problem with that.

conformance/tests/httproute-cross-namespace.go
robscott: Good call, I've updated most test descriptions, hopefully they're clearer now.

conformance/tests/httproute-cross-namespace.yaml
robscott: I definitely agree with the idea of removing type prefixes like `httproute-`, I've done that. I'm less sold on giving the routes more generic names. I think names can be used to provide context on what a Gateway/HTTPRoute is for. For example, a Gateway with a name `same-namespace-only` could make it easier to read and understand a test than a Gateway with a more generic name like `app-lb`.

conformance/tests/httproute-cross-namespace.go
robscott: Good call, I left the simplest test in and have pulled the others out for a follow up PR.

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// Request is the primary input for making a request.
+type Request struct {
+	URL      url.URL
+	Host     string
+	Protocol *v1alpha2.ProtocolType
+	Method   *v1alpha2.HTTPMethod
+	Headers  map[string][]string
+	Gateway  *v1alpha2.Gateway
```
hbagdi, line 46: Where is this being used?

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// CapturedResponse contains request metadata captured from an echoserver
+// response.
```
hbagdi, line 50: copy pasta error

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,154 @@
...
+type RoundTripper interface {
+	CaptureRoundTrip(Request) (*CapturedRequest, *CapturedResponse, error)
+}
+
+// Request is the primary input for making a request.
+type Request struct {
+	URL      url.URL
+	Host     string
+	Protocol *v1alpha2.ProtocolType
+	Method   *v1alpha2.HTTPMethod
```
hbagdi, line 44: Use string for Protocol/Method - simpler API. This doesn't seem to have an any advantage.

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -0,0 +1,154 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
hbagdi, line 2: I'd recommend starting with a single `conformance/utils/http` package which includes this file.
Any reason to have so manage packages upfront?

conformance/utils/http/http.go
```diff
@@ -0,0 +1,114 @@
...
+// ExpectedRequest can be used as both the request to make and a means to verify
+// that echoserver received the expected request.
+type ExpectedRequest struct {
+	Host    string
+	Method  string
+	Path    string
+	Headers map[string]string
+}
+
+// MakeRequestAndExpectResponse makes a request with the given parameters and
```
hbagdi, line 48: Is this function actually used anywhere?

conformance/utils/suite/suite.go
```diff
@@ -0,0 +1,98 @@
...
+limitations under the License.
+*/
+
+package suite
+
+import (
+	"testing"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
```
hbagdi, line 23: Unwanted line jump.

conformance/utils/kubernetes/apply.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		namespacedName := types.NamespacedName{Namespace: uObj.GetNamespace(), Name: uObj.GetName()}
+		fetchedObj := uObj.DeepCopy()
+		err := c.Get(context.TODO(), namespacedName, fetchedObj)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				require.NoErrorf(t, err, "error getting resource")
+			}
+			t.Logf("Creating %s %s", uObj.GetName(), uObj.GetKind())
+			err = c.Create(context.TODO(), &uObj)
```
hbagdi, line 68: Please replace all contexts with a 10 or 15-second timeout to avoid long-hanging tests. We can fine-tune the timeout later on.
A better (more tedious) approach would be to have a context being propagated up from the top. Either is fine, please avoid requests without timeout.

conformance/utils/kubernetes/apply.go
```diff
@@ -0,0 +1,92 @@
...
+			t.Logf("manifest: %s", string(b))
+			require.NoErrorf(t, decodeErr, "error parsing manifest")
+		}
+		if len(uObj.Object) == 0 {
+			continue
+		}
+
+		if uObj.GetKind() == "Gateway" {
+			err = unstructured.SetNestedField(uObj.Object, gcName, "spec", "gatewayClassName")
+			require.NoErrorf(t, err, "error setting GatewayClass")
```
hbagdi, line 58:
```suggestion
			require.NoErrorf(t, err, "error setting field `spec.GatewayClass` on a Gateway resource")
```

conformance/utils/kubernetes/apply.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		namespacedName := types.NamespacedName{Namespace: uObj.GetNamespace(), Name: uObj.GetName()}
+		fetchedObj := uObj.DeepCopy()
+		err := c.Get(context.TODO(), namespacedName, fetchedObj)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				require.NoErrorf(t, err, "error getting resource")
+			}
+			t.Logf("Creating %s %s", uObj.GetName(), uObj.GetKind())
```
hbagdi, line 67: Add a similar log line before c.Get() on line 62.

hbagdi (commented):
I didn't look at all the code but I think we can iterate here.
I'm comfortable merging.

conformance/utils/suite/suite.go
robscott: This looks strange but was intentional. I'm trying to group imports into three separate categories: 

- stdlib
- other
- local

A bit more explanation of the structure here https://github.com/golang/tools/commit/ed69e84b1518b5857a9f4e01d1f9cefdcc45246e. And an example of us doing similar grouping elsewhere in the project: https://github.com/kubernetes-sigs/gateway-api/blob/master/cmd/admission/main.go.

conformance/utils/roundtripper/roundtripper.go
robscott: Since this is an interface that implementations can override, I've wanted to keep it as generic and possible and leave the door open for at least TCP. That's a ways out though, and I'm not really sure how we'd structure that. The idea was to separate the interface + default implementation from HTTP specific helper funcs.

conformance/utils/kubernetes/apply.go
robscott: I'm a little hesitant to do that because our test output is already fairly verbose here. Here's the output I'm currently getting:

```
    suite.go:56: Test Setup: Ensuring GatewayClass has been accepted
    suite.go:59: Test Setup: Applying base manifests
    apply.go:81: Updating gateway-conformance-infra Namespace
    apply.go:81: Updating same-namespace Gateway
    apply.go:81: Updating all-namespaces Gateway
    apply.go:81: Updating backend-namespaces Gateway
    apply.go:81: Updating infra-backend-v1 Service
    apply.go:81: Updating infra-backend-v1 Deployment
    apply.go:81: Updating infra-backend-v2 Service
    apply.go:81: Updating infra-backend-v2 Deployment
    apply.go:81: Updating gateway-conformance-app-backend Namespace
    apply.go:81: Updating app-backend-v1 Service
    apply.go:81: Updating app-backend-v1 Deployment
    apply.go:81: Updating app-backend-v2 Service
    apply.go:81: Updating app-backend-v2 Deployment
    apply.go:81: Updating gateway-conformance-web-backend Namespace
    apply.go:81: Updating web-backend Service
    apply.go:81: Updating web-backend Deployment
    suite.go:62: Test Setup: Ensuring Gateways and Pods from base manifests are ready
```

Sometimes there's a mix of `Updating` and `Creating` depending on if I've enabled cleanup or added new resources to the base manifests. In either case, adding a `Getting` before each `Updating` or `Creating` line may add more confusion or noise than it's worth. I think it could be helpful to log before the first API call we're doing in case of a timeout or other connection issue, but that's actually happening earlier in the test flow. Do you think there's a better way to log this output as a whole?

conformance/utils/kubernetes/apply.go
robscott: Good call, I've updated to use context more intentionally throughout, but will work on a larger follow up in the next PR to see where it makes sense to propagate context further.

conformance/utils/kubernetes/apply.go
robscott: Updated, thanks!

conformance/utils/http/http.go
robscott: It is used quite heavily by the tests I pulled out to #994, have pulled out this function to add back with that PR.

conformance/utils/roundtripper/roundtripper.go
robscott: Updated, thanks!

conformance/utils/roundtripper/roundtripper.go
robscott: Updated, thanks!

conformance/utils/roundtripper/roundtripper.go
robscott: Woops, removed, thanks!

conformance/utils/suite/suite.go
```diff
@@ -0,0 +1,98 @@
...
+		GatewayClassName: gcName,
+		Debug:            debug,
+		Cleanup:          cleanup,
+	}
+}
+
+// Setup ensures the base resources required for conformance tests are installed
+// in the cluster. It also ensures that all relevant resources are ready.
+func (suite *ConformanceTestSuite) Setup(t *testing.T) {
+	suite.ControllerName = kubernetes.GWCMustBeAccepted(t, suite.Client, suite.GatewayClassName, 180)
```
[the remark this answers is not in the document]
robscott: I tried `t.Run` first, but that didn't feel right because these items are not really tests as much as they are setup steps. (The GatewayClass status == accepted is relatively close to a test, but the rest are not). Instead I've added logging before each step, which does seem to make the output much clearer here.

robscott:
Thanks for all the great feedback on this @hbagdi, @howardjohn, @jpeach, and anyone else who's commented. I think I've resolved everything and this should be good to go now, PTAL.

conformance/utils/suite/suite.go
hbagdi: Ah, TIL

conformance/utils/suite/suite.go
```diff
@@ -0,0 +1,102 @@
...
+	Client           client.Client
+	RoundTripper     roundtripper.RoundTripper
+	GatewayClassName string
+	ControllerName   string
+	Debug            bool
+	Cleanup          bool
+}
+
+// New returns a new ConformanceTestSuite.
+func New(c client.Client, gcName string, debug, cleanup bool, roundTripper roundtripper.RoundTripper) *ConformanceTestSuite {
```
hbagdi, line 40: Since new arguments may be added to this function, replacing the arguments of this function with a "SuiteOpts" struct could be better.

robscott: Updated, thanks!

hbagdi:
@robscott Linter unhappy.

robscott:
Woops, fixed the failing test, thanks for catching that @hbagdi! Removing the hold so next LGTM will result in a merge.

/hold cancel

hbagdi:
/lgtm
Thank you for all your hard work on this, Rob! This one took a lot of effort over the past year.

PR #970, opened 2021-12-21
Fix HTTPRoute validation tests.

jpeach:
**What type of PR is this?**

/kind bug
/kind cleanup

The HTTPRoute validation tests were directly calling internal API
functions to try to pinpoint the set of tests that they expected to
be run. This is a risky pattern, because there's nothing that ensures
that the internal functions are actually wired up to the validation
logic correctly.

This change establishes a new pattern, where each test can set up fixtures
in theor own way (for the sake of brevity), but the test is always driven
by the top-level validation call. This exposed a number of bugs where
the validation functions were not wired up correctly.

In the future, this can be improved by updating the tests to supply
specific expecations on the returned errors rather than just the error
count.

N/A

```release-note
NONE
```


robscott:
Thanks @jpeach! These are some great improvements.

/lgtm

PR #974, opened 2021-12-23
Document that ParentRefs require ReferencePolicy.

### LINKED ISSUE #973: Route ParentRef field does not mention ReferencePolicy
**What happened**:

The `Namespace` field of the `ParentRef` structure doesn't mention whether a `ReferencePolicy` is required for the controller to follow a cross-namespace reference.

I would expect all cross-namespace references to check a `ReferencePolicy`.

jpeach:
**What type of PR is this?**

/kind bug
/kind api-change

Consistent use of ReferencePolicy across the API makes it more
predictable and easier to reason about. Document that ParentRef
references that cross namespaces are required to obey ReferencePolicy
restrictions. I believe that this was always the intention, though
the documentation text was overlooked.

Fixes #973.

```release-note
Controller implementations must enforce ReferencePolicy polocies when
attaching routes to gateways in different namespaces.
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -52,6 +52,11 @@ type ParentRef struct {
 	// Namespace is the namespace of the referent. When unspecified (or empty
 	// string), this refers to the local namespace of the Route.
 	//
+	// Note that when a namespace is specified, a ReferencePolicy object
+	// is required in the referent namespace to allow that namespace's
+	// owner to accept the reference. See the ReferencePolicy documentation
+	// for details.
+	//
 	// Support: Core
 	//
 	// +optional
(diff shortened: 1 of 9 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -52,6 +52,11 @@ type ParentRef struct {
 	// Namespace is the namespace of the referent. When unspecified (or empty
 	// string), this refers to the local namespace of the Route.
 	//
+	// Note that when a namespace is specified, a ReferencePolicy object
+	// is required in the referent namespace to allow that namespace's
+	// owner to accept the reference. See the ReferencePolicy documentation
+	// for details.
```
robscott, line 58: I think this is a bit more nuanced, maybe something like this would work?
```suggestion
	// Note that when a different namespace is specified, something in the
	// target namespace needs to allow that reference. When referring to a
	// Gateway in a different namespace, that Gateway needs to explicitly
	// allow for Routes to be attached from this namespace. For other kinds
	// of parent resources that don't include this configuration, a
	// ReferencePolicy must be present to accept the reference. See the 
	// ReferencePolicy documentation for details.
```

jpeach:
Closing this out. Talking about ReferencePolicy would just add to the confusion that I'm suffering from. I'll think about some different language.

PR #976, opened 2022-01-03
feat: implement L4 traffic matching

shaneutt:
**What type of PR is this?**

/kind feature
/kind api-change

This implements [GEP 735](https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/geps/gep-735.md).

```release-note
- optional L4 traffic matching has been added for `TCPRoute` and `UDPRoute`
```


apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,54 @@ type AnnotationKey string
 // +kubebuilder:validation:MinLength=0
 // +kubebuilder:validation:MaxLength=4096
 type AnnotationValue string
+
+// TrafficMatches defines AddressMatch rules for inbound traffic according to
+// source and/or destination of that traffic.
+type TrafficMatches struct {
```
hbagdi, line 389: nit: I find `TrafficMatches` a bit too vague as a type name here but I can't seem to come up with a better one. This is not a very big deal since it doesn't show up in the YAML API.

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,54 @@ type AnnotationKey string
...
+
+// AddressMatch defines matching rules for network addresses by type.
+type AddressMatch struct {
+	// Type of the address, either IPAddress or NamedAddress.
+	//
+	// If NamedAddress is used this is a custom and specific value for each
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
```
hbagdi, line 411: Upstream differentiates between v4 and v6:
- https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services
- https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/

Should we?

Furthermore, do we need to think of cases where an implementation doesn't support IPv6 for any reason? Too hypothetical?

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,54 @@ type AnnotationKey string
...
+	// Value of the address. The validity of the values will depend
+	// on the type and support by the controller.
+	//
+	// If implementations support proxy-protocol (see:
+	// https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) they
+	// must respect the connection metadata from proxy-protocol
+	// in the match logic implemented for these address values.
+	//
+	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
+	//
```
hbagdi, line 430: Can we be explicit if CIDRs are allowed or not?

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,54 @@ type AnnotationKey string
 // +kubebuilder:validation:MinLength=0
 // +kubebuilder:validation:MaxLength=4096
 type AnnotationValue string
+
+// TrafficMatches defines AddressMatch rules for inbound traffic according to
+// source and/or destination of that traffic.
```
hbagdi, line 388:
```suggestion
// source and/or destination address of a packet or connection.
```

config/crd/experimental/gateway.networking.k8s.io_udproutes.yaml
```diff
@@ -218,6 +218,98 @@ spec:
...
+                            properties:
+                              type:
+                                allOf:
+                                - enum:
+                                  - IPAddress
+                                  - Hostname
+                                  - NamedAddress
+                                - enum:
+                                  - IPAddress
+                                  - NamedAddress
```
hbagdi, line 283: Two enum sets. Something is wrong here.

apis/v1alpha2/tcproute_types.go
```diff
@@ -59,6 +59,11 @@ type TCPRouteStatus struct {
 
 // TCPRouteRule is the configuration for a given rule.
 type TCPRouteRule struct {
+	// Matches add rules for filtering traffic to backends based on addresses.
```
hbagdi, line 62:
```suggestion
	// Matches are rules for routing traffic to backends based on addresses.
```

apis/v1alpha2/shared_types.go
shaneutt: > Upstream differentiates between v4 and v6 should we?

My initial thoughts are: one reason we might want to differentiate would be if one or the other is not available on the `Gateway` network (e.g. the `Gateway` only has IPv4 OR IPv6). Given that we're simply using the existing `IPAddress` type which already supported both IPv4 and IPv6 I feel like we would need a strong reason and a separate body of work to add this differentiation as it seems possible it would end up being backwards incompatible.

> Furthermore, do we need to think of cases where an implementation doesn't support IPv6 for any reason? Too hypothetical?

To expand on that I think there could be cases where IPv4 isn't supported too, but as you've already suggested I personally wouldn't want to spend cycles trying to predict these kinds of situations. We have until February (at least) before we're going to start talking beta release, there's time in the interim for such use cases to present themselves.

apis/v1alpha2/shared_types.go
shaneutt: It's inherent to the existing `IPAddressType` type, but yes I added documentation to that type explicitly pointing out that CIDR ranges are not allowed:

a6a982a7901da297494387d4ef62ebd880e48d7a

config/crd/experimental/gateway.networking.k8s.io_udproutes.yaml
shaneutt: Yes, unfortunately re-using the `AddressType` from `Gateway` ended up with validation from the base type combining with the top-level validation. I didn't find a way to indicate to the generators to only use the highest level validation, so I ended up making a new type purpose-built to replace this: 05616c8daf9fbb4d48e7c2084a35ed609e3c5400

apis/v1alpha2/gateway_types.go
```diff
@@ -477,6 +477,10 @@ const (
 	// must be in a standard IPv6 text representation
 	// (see [RFC 5952](https://tools.ietf.org/html/rfc5952)).
 	//
+	// Note that this type is intended for specific addresses and address ranges
+	// are not supported (e.g. you can not use a CIDR range like 127.0.0.0/24 as
+	// an IPAddress).
+	//
```
hbagdi, line 483:
```suggestion
	// This type is intended for specific addresses. Address ranges
	// are not supported (e.g. you can not use a CIDR range like 127.0.0.0/24 as
	// an IPAddress).
	//
```

apis/v1alpha2/shared_types.go
hbagdi: IPAddressType is not used it seems. I think that is because you introduced L4AddressType to fix the problem with CRD generation.

apis/v1alpha2/shared_types.go
hbagdi: > Given that we're simply using the existing IPAddress type which already supported both IPv4 and IPv6 I feel like we would need a strong reason and a separate body of work to add this differentiation as it seems possible it would end up being backwards incompatible.

Let's discuss this in the office hour today. We will be asked about dual-stack as soon as this API hits beta and we need to have a good answer/plan for it. Best to follow upstream closely.

hbagdi: My bad. No meeting today. 
cc @robscott who may be more familiar with upstream dual-stack work

apis/v1alpha2/tcproute_types.go
```diff
@@ -59,6 +59,11 @@ type TCPRouteStatus struct {
 
 // TCPRouteRule is the configuration for a given rule.
 type TCPRouteRule struct {
+	// Matches are rules for routing traffic to backends based on addresses.
+	//
+	// +optional
+	Matches *TrafficMatches `json:"matches,omitempty"`
```
jpeach, line 65: HTTPRoute has a slice here (e.g. `Matches []TrafficMatch`) which expresses an OR condition. I don't see this discussed in the GEP, but is there a good reason to diverge from the existing API pattern?

apis/v1alpha2/shared_types.go
jpeach: This struct defines a route match, so we can follow the pattern in HTTPRoute and call it `AddressRouteMatch`, maybe with aliases `TCPRouteMatch` and `UDPRouteMatch`.

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,58 @@ type AnnotationKey string
...
+	// addresses which are valid for routing traffic.
+	//
+	// Support: Core
+	SourceAddresses []AddressMatch `json:"sourceAddresses"`
+
+	// DestinationAddresses indicates the destination network addresses
+	// which are valid for routing traffic.
+	//
+	// Support: Core
+	DestinationAddresses []AddressMatch `json:"destinationAddresses"`
```
jpeach, line 400: If we followed the API pattern from HTTPRoute then this struct could have the following fields:

```go
SourceAddress *Address
SourceNetwork *NetworkAddress
DestinationAddress *Address
DestinationNetwork *NetworkAddress
```

The conditions in this struct would be AND'ed. The matches slice would be OR'ed.

apis/v1alpha2/shared_types.go
shaneutt: I think this is still accurate as is, the reference to `IPAddress` as an enum value here references the actual string `"IPAddress"` which is in turn type referenced by the `IPAddressType`? :thinking:

apis/v1alpha2/shared_types.go
shaneutt: I'm not opposed to calling this `AddressRouteMatches` if that suits you as well @hbagdi

apis/v1alpha2/gateway_types.go
shaneutt: resolved via e8bf1e64ca41b2ec24ae8d60f8ef51f4406fff58

apis/v1alpha2/tcproute_types.go
shaneutt: I'm OK with switching to the OR style list of traffic matches, cc @robscott any thoughts?

apis/v1alpha2/shared_types.go
shaneutt: It seems like your suggestion here proposes two things:

1. adding the `NetworkAddress` type
2. removing the `AddressMatch` type

Removing the `AddressMatch` type in favor of flat `Address` removes the forward flexibility that was intended by the `type+value` definition of an `Address` which followed the previous pattern established by `GatewayAddress` for how a network address can be defined. In your proposed model would you expect those new types to be added horizontally alongside `SourceAddress`, `DesitinationAddress`, e.t.c. as they come up? Just trying to make sure I'm understanding correctly.

If I'm understanding your suggestion correctly It seems that `NetworkAddress` is an alternate from the previously suggested (but noted as a consideration for later, see the alternatives section of the GEP) permutation of the `Address` type (such as a `CIDR` type) for match inclusion and solves the problem by adding horizontal options. The existing way solves the problem by nesting the permutations via the `type+value` option which intentionally followed the design pattern already present in the `GatewayAddress` type.

I'm not opposed to the change but It is not immediately clear to me that there is a benefit, could you help me to better understand your reasoning? Also given that this seems like a significant divergence from the original GEP I'm inclined to say it should be a follow up so that we can discuss that reasoning in isolation and forum at length and not overload this PR with that conversation, but I'm flexible and if most people prefer to sort this out now I will go along with it.

apis/v1alpha2/shared_types.go
hbagdi: That's an improvement so +1.

apis/v1alpha2/tcproute_types.go
hbagdi: +1. It would be for the best to not deviate from the existing API patterns.

apis/v1alpha2/shared_types.go
shaneutt: Alright sounds good, 19f1646b93f7db2e095771848c30fa1c7088c8ea changes this and updates the GEP accordingly.

apis/v1alpha2/tcproute_types.go
shaneutt: Sounds good, this was changed to a list in 19f1646b93f7db2e095771848c30fa1c7088c8ea

shaneutt: 19f1646b93f7db2e095771848c30fa1c7088c8ea includes documentation updates for this, and 9876eede43b01fa0472336eafc60a6422e0d3219 includes the changes to make this a list :+1:

apis/v1alpha2/shared_types.go
robscott: >Upstream differentiates between v4 and v6

Although that's true, at least part of that was motivated by simpler implementation. For example, EndpointSlices initially started with a single IPAddress type in v1alpha1 that was replaced with separate v4 and v6 types in large part to simplify proxy implementations where v4 and v6 addresses were handled entirely separately. 

It does seem like it would be useful to be able to request the IP families you want to have assigned to a Gateway. I'm less sure that it's helpful to have separate matching types for v4 and v6 addresses. In that case, the families can already be derived by the value provided and a separate type does not seem as useful to me.

@khenidak has been most involved in upstream dual stack work, what do you think about representing IP matching? Should we have separate address types per IP family or rely on deriving that information from the address/cidr itself?

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,58 @@ type AnnotationKey string
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Value string `json:"value"`
+}
+
+// AddressType defines how a network address is represented as a text string.
+//
+// +kubebuilder:validation:Enum=IPAddress;NamedAddress
+// +kubebuilder:default=IPAddress
+type AddressType string
```
robscott, line 440: This seems like it needs corresponding type definitions for IPAddress and NamedAddress, roughly matching what we do for GatewayAddressType. `AddressType` does seem like it may be too vague here, maybe `L4AddressType` would be better?

apis/v1alpha2/shared_types.go
robscott: Added a comment below, I think we also need type definitions for the new `AddressType`. Maybe both this and `GatewayAddressType` should be moved into a new `address_types.go` file?

apis/v1alpha2/shared_types.go
robscott: I agree with @shaneutt here, I'm not opposed to a change, but don't see a clear benefit. Some additional explanation could be helpful. If we were to go ahead with a change like that, maybe the best approach would be a separate PR to propose changes to the GEP itself.

robscott (commented):
Thanks for your work on this @shaneutt! Mostly LGTM, just a few small nits.

apis/v1alpha2/shared_types.go
jpeach: Sorry, the main gist of this suggestion is that we no longer need slices here. Since the outer struct has a slice (with OR semantics), this struct must have AND semantics, so the slice becomes moot (you can only use 1 element in almost all cases).

The proposal of a `NetworkAddress` type is to clarify the distinction between individual addresses and CIDR addresses, which (based on the discussion is not crystal clear). I think that it makes sense in this API to be able to configure network ranges, and that is probably also makes sense to have an explicit type for that with it's own nomenclature and validation.

Maybe it makes sense to mirror the existing `AddressType` to `NetworkType` so you can have a CIDR network, or a named network if it makes sense. I think that's pretty consistent with HTTPRoute and you wouldn't need to duplicate the existing gateway address.

Still thinking out loud here, but if you can match traffic on a destination CIDR network, that would imply that you have make a gateway listen on a network address, and I don't think there's a way to express that.

Happy to discuss further on slack or zoom :)

apis/v1alpha2/shared_types.go
jpeach: > It's inherent to the existing `IPAddressType` type,

I'm not sure about this. There's no explicit validation, but do we document that you are allowed to put a CIDR range in `GatewayAddress`?

apis/v1alpha2/shared_types.go
```diff
@@ -383,3 +383,58 @@ type AnnotationKey string
...
+
+	// DestinationAddresses indicates the destination network addresses
+	// which are valid for routing traffic.
+	//
+	// Support: Core
+	DestinationAddresses []AddressMatch `json:"destinationAddresses"`
+}
+
+// AddressMatch defines matching rules for network addresses by type.
+type AddressMatch struct {
```
jpeach, line 404: What's the motivation for a new type that is practically the same as `GatewayAddress`? Is this still needed after the refactoring?

apis/v1alpha2/shared_types.go
shaneutt: > Sorry, the main gist of this suggestion is that we no longer need slices here. Since the outer struct has a slice (with OR semantics), this struct must have AND semantics, so the slice becomes moot (you can only use 1 element in almost all cases).

In the GEP we discussed (and originally had) horizontal options at this level for destination vs source, and address vs port. To me it still seems useful to have this distinction for forward growth, e.g. direction of traffic and type of identifier (address, port, e.t.c.) would be horizontal here and then for each of those identifiers there would be a nested type to further configure those individual matches.

> The proposal of a NetworkAddress type is to clarify the distinction between individual addresses and CIDR addresses, which (based on the discussion is not crystal clear).
> I think that it makes sense in this API to be able to configure network ranges, and that is probably also makes sense to have an explicit type for that with it's own nomenclature and validation.
> Maybe it makes sense to mirror the existing AddressType to NetworkType so you can have a CIDR network, or a named network if it makes sense. I think that's pretty consistent with HTTPRoute and you wouldn't need to duplicate the existing gateway address.
> Still thinking out loud here, but if you can match traffic on a destination CIDR network, that would imply that you have make a gateway listen on a network address, and I don't think there's a way to express that.

CIDR was noted but specifically left out of the first iteration of the GEP (https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/geps/gep-735.md#cidr-addresstype). I think we should be doing follow up GEP work to add CIDR matching rather than trying to solve for that problem in a PR that is only meant to implement the state of the GEP so far.

apis/v1alpha2/shared_types.go
shaneutt: I had `L4AddressType` in a previous iteration, I'm fine with that. I've changed it back to that in b8a6fd9146123bef6fc3995bb19fde1a1ef5807e

apis/v1alpha2/shared_types.go
shaneutt: We don't accept `Hostname` for L4 matching.

apis/v1alpha2/shared_types.go
jpeach: I see .. that makes sense. Thanks!

apis/v1alpha2/shared_types.go
jpeach: Deferring CIDR network address types makes sense to me.

> In the GEP we discussed (and originally had) horizontal options at this level for destination vs source, and address vs port. 

Sorry, I'm not familiar with what you mean by "horizontal options", so not sure whether we are in agreement or not :)

shaneutt: Attributes on the same struct, vs sub-structs.

apis/v1alpha2/shared_types.go
shaneutt: > Added a comment below, I think we also need type definitions for the new AddressType. Maybe both this and GatewayAddressType should be moved into a new address_types.go file?

@robscott I might be misunderstanding you, but at least currently (now named `IPAddressType`) I don't think we need additional types, as `IPAddress` and `NamedAddress` are in use by way of the enum? :thinking:

apis/v1alpha2/shared_types.go
jpeach: Hmm, is that tradeoff worth it? We could reject `Hostname` in the webhook and not bifurcate the types ..

apis/v1alpha2/shared_types.go
jpeach: OK. All I think that we need here now it to switch to 
```Go
	SourceAddress *AddressMatch `json:"sourceAddress"`
	DestinationAddress *AddressMatch `json:"destinationAddress"`
```

and document that these fields are AND match conditions. We should crib some of the language fro HTTPRouteMatch to document the semantics.

apis/v1alpha2/shared_types.go
robscott: Sorry for the confusion here, I'm really just suggesting one of two things:

1. We just use AddressType and move the enum validation from the type definition to each field definition.
2. We split AddressType as you've currently proposed, and if so, we define valid values for the new type. For example, with the current AddressType, we have:

```go
type AddressType string

const (
    HostnameAddressType AddressType = "Hostname"
    // ...
)
```

But with the new type you've added, we only have the type definition, not the corresponding consts. The more I think about it though, the more I'm leaning towards just moving the enums to the fields instead of having a separate type everywhere we need an address in the API.

robscott: I think @jpeach is also suggesting something similar to 1 in his [comment below](https://github.com/kubernetes-sigs/gateway-api/pull/976#discussion_r779965079). In both cases we're suggesting a shared type, the only distinction would be how we handle the difference in enum validation depending on where it's used.

apis/v1alpha2/shared_types.go
shaneutt: I cherry picked 563873c25289a4abe0bb9f1fb09060bd16c6b4e5 into this branch from your fork (as we discussed separately in slack). I'm OK with the suggested change, so I will consider this resolved but if anyone feels there's more to be discussed here please do feel free to unresolve this comment :+1:

apis/v1alpha2/udproute_types.go
```diff
@@ -59,6 +59,14 @@ type UDPRouteStatus struct {
 
 // UDPRouteRule is the configuration for a given rule.
 type UDPRouteRule struct {
+	// Matches add rules for filtering traffic to backends based on addresses.
+	// Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 67:
```suggestion
	// +kubebuilder:validation:MaxItems=16
	//  <gateway:experimental>
```

apis/v1alpha2/tcproute_types.go
```diff
@@ -59,6 +59,14 @@ type TCPRouteStatus struct {
 
 // TCPRouteRule is the configuration for a given rule.
 type TCPRouteRule struct {
+	// Matches are rules for routing traffic to backends based on addresses.
+	// Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
robscott, line 67:
```suggestion
	// +kubebuilder:validation:MaxItems=16
	//  <gateway:experimental>
```

config/crd/stable/gateway.networking.k8s.io_tcproutes.yaml
```diff
@@ -218,6 +218,84 @@ spec:
                       maxItems: 16
                       minItems: 1
                       type: array
+                    matches:
```
robscott, line 221: Stable CRDs should not change with this PR.

config/crd/stable/gateway.networking.k8s.io_udproutes.yaml
```diff
@@ -218,6 +218,84 @@ spec:
                       maxItems: 16
                       minItems: 1
                       type: array
+                    matches:
```
robscott, line 221: Stable CRDs should not change with this PR.

examples/v1alpha2/traffic-matching-tcp.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: I think this should actually be in `examples/v1alpha2/experimental` and we'd also need a corresponding update in https://github.com/kubernetes-sigs/gateway-api/blob/master/hack/verify-examples-kind.sh#L63 that only tried to install experimental examples when `CHANNEL == "experimental"`.

apis/v1alpha2/gateway_types.go
```diff
@@ -462,38 +462,44 @@ type GatewayAddress struct {
...
+// string, specifically in the context of listener addresses for Gateways and
+// includes options over the base AddressType type such as Hostname.
 //
 // If the requested address is unsupported, the controller
 // should raise the "Detached" listener status condition on
 // the Gateway with the "UnsupportedAddress" reason.
 //
 // +kubebuilder:validation:Enum=IPAddress;Hostname;NamedAddress
-type AddressType string
+type GatewayAddressType string
```
robscott, line 474: The more I think about this, the more I think that we should share the single AddressType for both use cases and move the enum to the field. Unfortunately field based validation is currently the only way to add experimental additions to enum validation (https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L498). That won't actually be needed here yet, but if we add a new type in the future, it would be helpful to have it already split out that way.

robscott (commented):
Thanks for all the work on this @shaneutt! I think you and @jpeach may be working on bigger changes here, but just remembered that we need to guard all these new fields as `experimental` to start, so added some comments related to that.

apis/v1alpha2/shared_types.go
```diff
@@ -399,3 +399,65 @@ type AnnotationKey string
...
+	// +optional
+	SourceAddress *NetworkAddress `json:"sourceAddresses"`
+
+	// DestinationAddresses indicates the destination network addresses
+	// which are valid for routing traffic.
+	//
+	// Support: Core
+	//
+	// +optional
+	DestinationAddress *NetworkAddress `json:"destinationAddresses"`
```
robscott, line 423: @shaneutt @jpeach I wanted to follow up a bit on why I was hesitant about this change. As I understand it, we've gone from separate lists of source and dest addresses, to lists of source/dest address pairs. That is fine as long as users are only trying to match on source address or dest address, but gets frustrating if they're trying to match on both.

For example, previously the following would have been possible:
```yaml
- source:
  - 10.1.2.3
  - 10.1.2.4
- dest:
  - 10.2.3.4
  - 10.2.3.5
 ```

To represent that now, we end up with an exponential increase in config:
```yaml
- source: 10.1.2.3
  dest:   10.2.3.4
- source: 10.1.2.4
  dest:   10.2.3.4
- source: 10.1.2.3
  dest:   10.2.3.5
- source: 10.1.2.4
  dest:   10.2.3.5
```

Both examples are actually simplified versions of the config that would actually be needed, but hopefully that helps illustrate the concern in terms of exponential growth when both source and dest matching need to be combined.

A potential solution would be to make both SourceAddress and DestinationAddress lists here or say that we can _mostly_ solve this with CIDRs later, but neither of those options feel great.

jpeach: Yeh, the first example is definitely cleaner. I am not sure how realistic this example is (do people really need to deal with routing lots of individual addresses?), but we should probably still prefer the cleaner syntax.

apis/v1alpha2/shared_types.go
shaneutt: As suggested I've re-converged the types. I particularly made `AddressType` a shared type as this seems like the type we want to center on.

Everything looks right _except_ we're back to the problem where the generators are adding `Hostname` to the L4 matching address type. I've tried to override or cancel this but I'm not sure if that can be done with the generator tags. If someone has a hot tip on how this can be done I need some help.

apis/v1alpha2/shared_types.go
shaneutt: This is now re-converged and there's no longer a new type.

config/crd/stable/gateway.networking.k8s.io_tcproutes.yaml
shaneutt: things have been moved to `experimental` as you suggested

examples/v1alpha2/traffic-matching-tcp.yaml
shaneutt: things have been moved to `experimental` as you suggested

apis/v1alpha2/gateway_types.go
shaneutt: I couldn't find another way to do this, so I implemented this as you suggested. This did not result in a change to anything else during generation so I think we're looking good :+1:

shaneutt: 1245aab692077a7c130730bf0e30121af9e850d8

apis/v1alpha2/shared_types.go
shaneutt: This has been changed as suggested, there is now only one `AddressType` :+1:

apis/v1alpha2/shared_types.go
brianehlert: I am looking at this now for CRDs (outside of Gateway) and I am leaning to follow Service which uses ipfamilies to follow existing semantics.
To easily express: match ipv6 traffic, match ipv4 traffic, match both
@shaneutt is that what you are thinking AddressType would do, equivalent-ish to Service ipfamilies?
This could allow matching on family alone without requiring the person to define a match pattern.
To your point, always requiring a match pattern to signify anything other than 'both' is only one key.

apis/v1alpha2/shared_types.go
```diff
@@ -399,3 +399,86 @@ type AnnotationKey string
...
 // +kubebuilder:validation:MaxLength=4096
 type AnnotationValue string
+
+// AddressRouteMatches defines AddressMatch rules for inbound traffic according to
+// source and/or destination address of a packet or connection.
+type AddressRouteMatches struct {
+	// SourceAddresses indicates the originating (source) network
+	// addresses which are valid for routing traffic.
+	//
+	// Support: Core
```
robscott, line 409: Nit: I think this should probably be "Extended" to start, and we can potentially graduate to "Core" in the future if we see wide enough support.

apis/v1alpha2/shared_types.go
```diff
@@ -399,3 +399,86 @@ type AnnotationKey string
...
+	SourceAddresses []AddressMatch `json:"sourceAddresses"`
+
+	// DestinationAddresses indicates the destination network addresses
+	// which are valid for routing traffic.
+	//
+	// Support: Core
```
robscott, line 415: Same comment as above re support level.

site-src/v1alpha2/guides/tcp.md
```diff
@@ -43,6 +43,24 @@ In this way each `TCPRoute` "attaches" itself to a different port on the
 `Gateway` so that the service `my-foo-service` is taking traffic for port `8080`
 from outside the cluster and `my-bar-service` takes the port `8090` traffic.
 
+## Alternatives: Routing Using Traffic Matching
+
+**EXPERIMENTAL**
```
robscott, line 48: Maybe use an [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#inline-blocks) for this? I'm not sure if this should be "info" or "warning". Something like this might work:

```
??? info "Experimental Channel"

    This functionality is currently only included in the "Experimental" channel
    of Gateway API. For more information on release channels, refer to the
    [related documentation](https://gateway-api.sigs.k8s.io/concepts/versioning/#adding-experimental-fields).
```

robscott:
Thanks for the work on this @shaneutt! This LGTM other than a couple tiny nits.

apis/v1alpha2/shared_types.go
shaneutt: If I'm understanding correctly, yes that's basically what I had in mind.

apis/v1alpha2/shared_types.go
shaneutt: updated in bb3672a1517853dbbfec47b7c02a5b784d376bc1 :+1:

apis/v1alpha2/shared_types.go
shaneutt: updated in bb3672a1517853dbbfec47b7c02a5b784d376bc1

site-src/v1alpha2/guides/tcp.md
shaneutt: added in bb3672a1517853dbbfec47b7c02a5b784d376bc1 :+1:

robscott: @shaneutt apparently my snippet above must have relied on a newer version of the plugin or something. Replacing `???` with `!!!` seems to work though.

shaneutt: Ok, now changed in 5f448f55bb08f3bb857fed145c336e8716969b45 :+1:

robscott:
Thanks @shaneutt!

/lgtm
/approve

apis/v1alpha2/shared_types.go
bowei: As the value is a specific Value, having an additional IP family field seems to be redundant.

Minor style point, we should write all of the comment from the perspective of the users (the implemetors should...) style comments need to live somewhere else...

PR #978, opened 2022-01-05
Add kind matching helpers to object reference types.

jpeach:
**What type of PR is this?**

/kind feature
/kind api-change

Every controller needs to test whether an object referene matches some
known group and kind, so it's probably convenient if we add a helper
directly to the API package.

N/A

```release-note
Added the `HasKind()` member function to object reference types.
```

### DIFF
--- a/apis/v1alpha2/object_reference_types.go
+++ b/apis/v1alpha2/object_reference_types.go
@@ -16,6 +16,22 @@ limitations under the License.
 
 package v1alpha2
 
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+func makeGroupKind(g *Group, k *Kind) schema.GroupKind {
+	var gk schema.GroupKind
+
+	if g != nil {
+		gk.Group = string(*g)
+	}
+
+	if k != nil {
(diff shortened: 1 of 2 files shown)


jpeach:
> @jpeach Thanks for the work on this! I think an expanded set of helpers would likely be helpful to include in this project. If I'm understanding this correctly, it's mostly useful because Group and Kind are both pointers, and comparing the values of pointers can be tedious?

In practice, it is safe to dereference them because we always provide defaults, but comparing to a GroupKind directly makes controller code cleaner IMHO.

My agenda with this PR is to gauge the community appetite for small API extensions that can improve the quality of life for controller authors :)

> As far as where this and other helpers should live, I think it likely belongs in a different package, maybe `v1alpha2/utils`?

Yeh I thought about a different package, and the possible approaches I could think of are

1. Use a different function for each API type
1. Use an `interface{}` parameter and type switch internally
1. Use Go generics

I think that (2) would be the lease worst if we want to go that way.

jpeach:
Closing this, and will post a new PR that doesn't change the CRD types.

PR #981, opened 2022-01-08
Cherry Pick #956 into release-0.4

robscott:
**What type of PR is this?**
/kind bug

This cherry-picks #956 into the release-0.4 branch in advance of a v0.4.1 release.

```release-note
Introduced PreciseHostname which prevents wildcard characters in relevant Hostname values. 
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -683,7 +683,7 @@ type HTTPRequestRedirectFilter struct {
 	// Support: Core
 	//
 	// +optional
-	Hostname *Hostname `json:"hostname,omitempty"`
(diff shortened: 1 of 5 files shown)


brianehlert:
I am not following the use case why this is required.
I understand there is some link to validation enforcement, but that seems more like an implementation convenience than a feature that makes the api easier to use.

robscott:
@brianehlert That's a good question. This change actually tightens validation, which is technically backwards incompatible. We're trying to get as many of these changes out of the way while the API is still in alpha. The rationale for even allowing this change in is that the previously allowed value was entirely invalid and not implementable. Redirecting to *.example.com does not make sense.

PR #982, opened 2022-01-10
Rename ParentRef to be consistent with other reference types.

jpeach:
**What type of PR is this?**
/kind cleanup

Across the API, the object reference types are all called
`FooBarReference`. To be consistent with this pattern, `ParentRef`
should be called `ParentReference`.

A few issues with this PR

1. Not convinced that this is a compatible change for code that consumes this module
2. deepcopy-gen generates code for both the type and its alias, which is deeply wrong

N/A

```release-note
ParentRef type renamed to ParentReference.
```


apis/v1alpha2/shared_types.go
```diff
@@ -86,6 +86,9 @@ type ParentRef struct {
 	SectionName *SectionName `json:"sectionName,omitempty"`
 }
 
+// ParentRef is a backwards compatibility alias for ParentReference.
+type ParentRef = ParentReference
```
hbagdi, line 90: I think this is upsetting the code generation?

jpeach: Yup

> deepcopy-gen generates code for both the type and its alias, which is deeply wrong

hbagdi: I think not having this type of backward compatibility with Go APIs like this is reasonable. Larger changes may pose problems but this is okay.
Sensitive area so would like to hear others.

howardjohn: +1, I am not too concerned with go compatibility. Especially when its a trivial change

robscott: Agreed, this seems fine. Upstream kubernetes has also not been concerned about go compat historically.

jpeach: Sounds great! I removed the compatibility alias, so let's see what CI says about that.

hbagdi:
LGTM.
Grepped for places that could need updates but didn't find any.
Leaving the final approval for another maintainer.

robscott:
Thanks!

/lgtm

PR #986, opened 2022-01-12
Added name to HTTPRoute

hzxuzhonghu:
**What type of PR is this?**

/kind design

Istio convert HTTPRoute to its own VirtualService, and then use it to build xds Router, however in xds route, there is a name telling envoy the router and it can be shown in access log. So users can know how their request goes through gateway.

FYI: https://github.com/istio/istio/issues/36790

Fixes #

```release-note
Added name to HTTPRouteRule and HTTPRouteMatch to tell the users which route a request passes through.
```


hzxuzhonghu:
cc @howardjohn

howardjohn (commented):
This is very useful if/when we add `sectionName` to `PolicyTargetReference` as well

brianehlert:
I am consciously not trying not to come across in a negative way with my comments and struggling.  I am going to admit that up front.  I realize this is also set as optional.
I seriously want to better understand: 
What are the use cases beyond Istio?  Can this value be inferred from other values?
I am trying to get a better understanding of the reason / problem behind the request and why it can't be solved in a way that does not require another value or can't be accomplished with an implementation specific extension.
Anyone willing to help me grok this?

howardjohn:
I think there are two main use cases:
* Ability to add debugging information, such as adding the route matched to some access log
* Ability to have more fine grained matches for policy references IF we add `sectionName` to the policy object. Which IMO we should

robscott:
Although it's only an addition of an optional field, I think this would still benefit from a small GEP to explain the rationale and potential use cases. GEPs can also provide a better way to debate the merits of a feature and a reference for the future as far as why something was added. 

With that said, this is a familiar concept. I actually initially proposed this in the policy attachment GEP, but at that point it got tabled for later as the use cases weren't sufficiently clear. Now that some time has passed, I see the following potential use cases:

- Policy Attachment
- Route Attachment/Delegation
- Improved Debugging

I think a GEP for this would need to explore all 3 potential use cases and ensure that anything we added made sense for all. We'd also need to determine if/how we could validate the uniqueness of name here to roughly correspond with the uniqueness validation we have on listener name.

/hold

hzxuzhonghu:
Sorry, what is GEP?

> uniqueness validation we have on listener name.

Where is this documented?

hbagdi:
> Sorry, what is GEP?

GEP is a lightweight process to introduce enhancements. 
Please see:
- https://gateway-api.sigs.k8s.io/contributing/enhancement-requests/
- https://gateway-api.sigs.k8s.io/contributing/gep/

It is similar to KEP in the upstream Kubernetes project.

> >  uniqueness validation we have on listener name.

> Where is this documented?

We haven't documented this in various places because it was something that was taken for granted (a well-defined pattern in Kubernetes).
This is enforced via these markers: https://github.com/kubernetes-sigs/gateway-api/blob/85dab2ac7ed6f64d5f9c225df92cd73f4ab85893/apis/v1alpha2/gateway_types.go#L110-L111

rahulanand16nov:
Any news on the status of this PR?

youngnick:
We're still discussing what implications this has for the rest of the API.

I think that everyone is broadly in favor of adding an optional name field, but we need to define the semantics around what to do when it's not present.

@robscott, I know I'm missing some detail here, but I can't find the references. Help?

robscott:
@Jeff-Apple's GEP for route delegation (#1058) will also rely on an optional name field for Route rules. I think once that gets finalized, we'll have a clear rationale for adding this field.

youngnick:
Yes, I think that what we have here implements the mechanics of actually having the field present, but we will need a GEP (that is, a document that describes the change in #995 ) that runs through the caveats and answers some questions around the lifecycle:
- If a name is specified, what does it mean? What can it be used for? If it's not present, what happens?
- Does the name need to be unique? (Seems like it would be) If so, how do we enforce that? Does the webhook check? (Seems like that's the obvious place).
- Can you refer to a HTTPRoute rule that doesn't have a name?
- What happens if you specify a name that doesn't exist?

PR #987, opened 2022-01-13
Clarify that listener names are unique.

### LINKED ISSUE #984: Clearly document that listener SectionName field values must be unique within a Gateway

The spec should say whether the listener SectionName field values need to be unique within a Gateway. If so, the validation package should enforce it.

**Why this is needed**:

The listener status includes the SectionName field, so it's a reasonable implication that you should be able to match this up to the SectionName field in the Gateway spec. If SectionName values are not unique, then the matching is ambiguous.

jpeach:
**What type of PR is this?**
/kind documentation

Listener names must be unique within a Gateway, but the documentation never
mentions that.

Fixes #984

```release-note
NONE
```


robscott:
Thanks @jpeach!

/lgtm

PR #993, opened 2022-01-13
GEP-992 HTTPRoute LB Strategies

### LINKED ISSUE #992: Explicit LoadBalancing Configurations for HTTPBackendRefs

Clear, explicit and top-level configuration for the load-balancing strategy when using multiple `HTTPBackendRefs` in an `HTTPRoute` ruleset.

**Why this is needed**:

Currently only a "weighted" load-balancing strategy is supported, and it's effectively _implicit_ (rather than ideally _explicit_) in configuration currently. There is no way to define other strategies, or for implementations to have their own specific strategies available.

shaneutt:
**What type of PR is this?**

/kind gep
/kind api-change

This GEP aims to make it possible to define load-balancing strategies for `HTTPRoutes`.

Resolves #992


site-src/geps/gep-992.md
```diff
@@ -0,0 +1,184 @@
...
+	// effect, otherwise it is required.
+	//
+	// Support: Core for RoundRobinStrategy
+	// Support: Core for WeightedStrategy
+	// Support: Extended for LeastConnectionsStrategy
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=RoundRobin,WeightedStrategy,LeastConnectionsStrategy
+	// +kubebuilder:default=RoundRobin
+	LoadBalancerStrategy *LoadBalancerStrategy `json:"loadBalancerStrategy,omitempty"`
```
howardjohn, line 106: Copying my comment from the issue so we have 1 place for discussion


I think Route is the wrong place for this - or at least the wrong place to be the exclusive location. if its a Policy, then it can attach at different points. I think the most important points are Service and global. For example, global is useful if you want to say "All services should use the 'failover to another region' load balance strategy". Probably not all your services in the cluster need to define that. Another example would be a cluster admin could decide they like Round Robin by default, but a specific service needs sticky load balancer or something.

Briefly looking at some proxies, it seems most are putting things at the backend level?:
Google: https://cloud.google.com/compute/docs/reference/rest/v1/backendServices
Envoy: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto.html?highlight=load
Possibly nginx (not as familiar): https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/
haproxy: https://www.digitalocean.com/community/tutorials/an-introduction-to-haproxy-and-load-balancing-concepts

One implementation note for those proxies is that you end up needing to create extra backends on the order of `backends * number of routes configuring backend properties`. This can explode to very large numbers

site-src/geps/gep-992.md
```diff
@@ -0,0 +1,184 @@
...
+	// RoundRobin is a simple LoadBalancerStrategy which distributes inbound
+	// matching requests across all HTTPBackendRefs evenly according to the
+	// overall number of requests.
+	RoundRobinStrategy LoadBalancerStrategy = "RoundRobin"
+
+	// WeightedStrategy distributes inbound matching requests across
+	// HTTPBackendRefs according to the "weight" value set on each individual
+	// ref, and uses those weights to determine what percentage of traffic each
+	// backend should take on.
+	WeightedStrategy LoadBalancerStrategy = "Weighted"
```
howardjohn, line 128: This is concerning to me because it means that `weight` doesn't apply to other load balancing algorithms.

Its very reasonable to have weight and {RoundRobin,LeastConnections}. As a two tier selection:
* First, pick the 'backend' by weight.
* Next, pick the exact upstream IP to use by LB algorithm

shaneutt: I considered this when I originally wrote it, but so far hadn't included it because it seemed simple to add later and I wanted to keep the first draft tiny. If you feel that it really does need to be added up front I can simply add it.

howardjohn: This would break us I think, in its current state. We have other resources (eg Policy) that define load balancing state. By not having the strategy optional here, that enforces that we would have to use the Route policy (unless we just blatantly ignore what they configured...), which is very restrictive compared to our current policy options.

Which goes back to my other comment about it being in route, really.

site-src/geps/gep-992.md
shaneutt: What you're saying here seems to align with some conversation I had with @robscott on the same subject, which was ultimately the subject of the current contents of `# Alternatives`. I would like to get his input here as well.

site-src/geps/gep-992.md
```diff
@@ -0,0 +1,184 @@
...
+	//
+	// If only one backend is provided in BackendRefs this property has no
+	// effect, otherwise it is required.
+	//
+	// Support: Core for RoundRobinStrategy
+	// Support: Core for WeightedStrategy
+	// Support: Extended for LeastConnectionsStrategy
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=RoundRobin,WeightedStrategy,LeastConnectionsStrategy
```
robscott, line 104: Reading through this GEP got me to spend some time looking at what was possible across different implementations. Similar to the rewrite GEP, I think it would be useful to provide a brief overview of the capabilities of various proxies/load balancers to clarify why we're choosing theses strategies as a base. Here's a rough starting point for what I've found so far. 

**GCP supports [3 modes:](https://cloud.google.com/load-balancing/docs/backend-service#balancing-mode)**
- Connection
- Rate 
- Utilization

**Envoy supports [5 balancing modes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#arch-overview-load-balancing-types-least-request):**
* Weighted round robin
* Weighted least request
* Ring hash
* Maglev
* Random

**NGINX supports [6 balancing modes:](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/):**
* Round Robin
* Least Connections
* IP Hash
* Generic Hash
* Least Time
* Random

**HAPROXY supports [10 balancing modes:](http://cbonte.github.io/haproxy-dconv/2.0/configuration.html#4-balance)**
* roundrobin
* static-rr
* leastconn
* first
* source
* uri
* url_param
* hdr
* random
* rdp-cookie


Some top level concerns:
- Different proxies support attaching this config at different levels
- They almost universally accept additional configuration with varying structure

I need to think about this more, but I'm wondering if this could be better served with policy attachment.

site-src/geps/gep-992.md
```diff
@@ -0,0 +1,184 @@
...
+      port: 80
+    - name: service2
+      port: 80
+    - name: service3
+      port: 80
+    loadBalancerStrategy: "LeastConnections"
+```
+
+TODO: determine the best way to add "NamedStrategy" which would allow
+      individual implementations to define their own strategies.
```
robscott, line 166: Agree with this. If we're going to take the route of built in configuration, we'll need to leave room for custom types + likely custom configuration to go along with that.

robscott (commented):
Thanks for writing this up @shaneutt!

shaneutt:
@howardjohn and @robscott I appreciate all your feedback thus far, and I'm continuing to think and draft on this and will eventually have some updates to this draft, but I have run into something I would like to get some specific feedback on.

One thing that seems common between both of your comments (and my own needs as an implementor) is that we're _going_ to need sooner or later the lower-level definitions for `LoadBalancerStrategy` as was suggested by @howardjohn in the beginning.

Consider the following topology items:

httproute-lb-layered
---
httproute-lb-flat
---
httproute-lb-mixed

There's been an open suggestion that the top level `LoadBalancerStrategy` could just be a default that cascades down to lower levels, but I see two dimensions to this: one where the implementation really could have a real balancer at both levels with a potentially different strategy, and one where _only_ the top level strategy makes any sense.

In my own use case I have real options where the backend might just be a `Service` IP (modeled by diagram 2) and other real options where the `HTTPBackendRefs` resolve to individual addresses (modeled by 1), so ultimately I see a need to support defining a strategy for both groups.

Does that resonate with you @howardjohn and @robscott? I think this may have been what both of you were already getting at, but I want to make sure we're on the same page.

site-src/geps/gep-992.md
```diff
@@ -0,0 +1,184 @@
...
+      port: 80
+```
+
+This implies a load-balancing strategy for traffic which matches but is limited
+in how it can be configured today.
+
+### Problems
+
+Currently only one load-balancing strategy is available, which we'll call the
+`weighted` strategy (and by proxy, you can implement a `round-robin` strategy
```
bowei, line 49: We should be careful as round robin is distinct from weighted. For example, flipping a coin for each request is ok for the weighted implementation, but round robin is more deterministic in behavior.

site-src/geps/gep-992.md
bowei: I agree with this -- I think this is something that is going to be hard to bridge in a good way, especially between distributed vs single instance implementations.

site-src/geps/gep-992.md
shaneutt: Correct, thus why in the implementation hereafter they are referred to as distinct items and have their own distinct types.

site-src/geps/gep-992.md
pleshakov: > One implementation note for those proxies is that you end up needing to create extra backends on the order of backends * number of routes configuring backend properties. This can explode to very large numbers

I think this is a very serious implication for the implementations. Right now nothing dictates that each backendRef needs to have its own dedicated load balancing pool: pools can be shared among routes. More pools = more memory is taken on a proxy (I assume this is true for most of the proxies).

site-src/geps/gep-992.md
pleshakov: As mentioned in the other comments, it is common to have two levels of load balancing:
- first, choose a backend (load balancing pool, upstream, cluster, etc.)
- second, choose a backend instance according to the load balancing method.

I don't think it is common for proxies to support least connections or round robin methods at the first level expect splitting traffic by weight . As a result, to implement the proposed configuration, it will be necessary to create a new pool to include members from different pools. 

My concern is that we will remove the first level this way. And as mentioned here https://github.com/kubernetes-sigs/gateway-api/pull/993/files#r784229605, it will be necessary to create a pool per each route.

shaneutt:
Closing this for the time being as my priority for it has changed and the feedback received thus far has informed other changes in implementation. I will be revisiting this in time, perhaps after further conversation in the community sync and once priority for it comes back around.

PR #1001, opened 2022-01-18
Fix v1alpha2 "HTTP routing" example

acnodal-tc:
**What type of PR is this?**
/kind documentation

The example "bar-route" HTTPRoute from the "HTTP routing" web page had a bug where both backendRefs belonged to
the same rule so traffic would either match the header in which case it
would be split 50/50 between bar-svc-canary and bar-svc, or it wouldn't
match the header in which case I'm not sure what would happen.

This PR adds a second HTTPRouteRule to the bar-route HTTPRoute so the
yaml manifest matches the descriptive text.

https://kubernetes.slack.com/archives/CR0H13KGA/p1642527563035400

```release-note
This PR fixes the bar-route example on https://gateway-api.sigs.k8s.io/v1alpha2/guides/http-routing/
```

### DIFF
--- a/examples/v1alpha2/http-routing/bar-httproute.yaml
+++ b/examples/v1alpha2/http-routing/bar-httproute.yaml
@@ -16,5 +16,6 @@ spec:
     backendRefs:
     - name: bar-svc-canary
       port: 8080
+  - backendRefs:
     - name: bar-svc
       port: 8080


robscott:
Thanks @acnodal-tc!

/ok-to-test

robscott:
Thanks!

/lgtm
/approve

examples/v1alpha2/http-routing/bar-httproute.yaml
```diff
@@ -16,5 +16,6 @@ spec:
     backendRefs:
     - name: bar-svc-canary
       port: 8080
+  - backendRefs:
```
hbagdi, line 19: This is really hard on the eyes. An improvement would be to move this array element to position 0.
A further improvement would be to also specify the default match criteria - the explicit match will make the object much more self-documenting.

robscott: Yeah, that would make this more readable.

acnodal-tc: Got it, thank you for taking a look. I'll open a new PR with these changes.

acnodal-tc: Sorry folks, I need some clarification. Rearranging the rules is no problem.

If, though, I add the default match (which is PathPrefix of "/" I think) as Harry suggests I get
```yaml
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - name: bar-svc
      port: 8080
  - matches:
    - headers:
      - type: Exact
        name: env
        value: canary
    backendRefs:
    - name: bar-svc-canary
      port: 8080
```
The problem is that path matches are more "specific" than header matches so now the match that we want to be the catch-all would match everything.

If I add a PathPrefix "/" match to the second match then it would have two things that match but I can't find anything that says that the count of things that match has an impact on how specific the match is.

robscott: You're right, this is quite confusing. Matching should give precedence to the largest number of header matches in a tie, which should result in the behavior we want. 

https://github.com/kubernetes-sigs/gateway-api/blob/398f7e14fcefd85c0dc411928e49121d958be781/apis/v1alpha2/httproute_types.go#L141-L148

The way defaulting works, any implementation will not be able to differentiate between specified and default values when they are equivalent. So in this case, both rules would end up with the same indistinguishable path match, even though only one specified it. With all that said, that's pretty complicated to communicate in what should be a relatively simple example. 

Maybe the simplest thing is to just add yaml comments above each rule describing their purpose. Something like:

```yaml
  rules:
  # canary only
  - matches:
    - headers:
      - type: Exact
        name: env
        value: canary
    backendRefs:
    - name: bar-svc-canary
      port: 8080
  # catch all
  - backendRefs:
  # ...
```

Interested in what @hbagdi thinks here too.

hbagdi: My interpretation is:

```yaml
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - name: bar-svc
      port: 8080
  - matches:
    - path:
        type: PathPrefix
        value: /
      headers:
      - type: Exact
        name: env
        value: canary
    backendRefs:
    - name: bar-svc-canary
      port: 8080
```

This translates essentially to gateway sending all requests to bar-svc except requests containing the header 'env: canary', which are forwarded to bar-svc-canary.


> Matching should give precedence to the largest number of header matches in a tie

Yes, but parameters for matching withing a single match are ANDed together.

acnodal-tc: Thank you both! I think that I understand now, please check my work:

> https://github.com/kubernetes-sigs/gateway-api/blob/398f7e14fcefd85c0dc411928e49121d958be781/apis/v1alpha2/httproute_types.go#L144-L148

Each criterion in those 5 is checked in turn for each HTTPRouteRule and if a winner can be found based on one criterion then no further criteria are checked. In this example both HTTPRouteRules would tie on "Characters in a matching non-wildcard hostname" (none), "Characters in a matching hostname" (none), and "Characters in a matching path" (one) so the deciding criterion will be "Header matches" because bar-svc-canary has one match and bar-svc has none. "Query param matches" does not need to be checked because a decision was made earlier.

Is this correct?

robscott: @acnodal-tc yep, that's a great way to explain it.

PR #1002, opened 2022-01-26
Implement GEP-957: Destination Port Matching.

cxhiano:
**What type of PR is this?**
/kind feature
/kind api-change

Implement destination port matching proposed in [gep-957](https://gateway-api.sigs.k8s.io/geps/gep-957/)

```release-note
- Allow routes to select listeners by port number.
```


howardjohn (approved):

hbagdi:
/ok-to-test
We should add an example (examples/ directory) to make sure we have some CI around it.
We should also add a guide for this. I'm not sure if that has to be part of this PR but I think that should be in the scope of the GEP.

cxhiano:
Example added. PTAL 

Agreed that we could have a guide(probably in a separate PR). Should the guide be centered around Route <> Listener attachment or Route <> custom CRD(e.g. Mesh) attachment? I am not sure though.

apis/v1alpha2/shared_types.go
```diff
@@ -65,7 +65,9 @@ type ParentReference struct {
 	// SectionName is the name of a section within the target resource. In the
 	// following resources, SectionName is interpreted as the following:
 	//
-	// * Gateway: Listener Name
+	// * Gateway: Listener Name. When both Port and SectionName are specified,
```
robscott, line 68: Noticed that this shows up in stable API definitions. Until we have a better way to differentiate, something like this might work:
```suggestion
	// * Gateway: Listener Name. When both Port (experimental) and SectionName are specified,
```

robscott (commented):
Thanks @cxhiano! This mostly LGTM, do you mind creating a follow up issue to track adding docs for this feature?

apis/v1alpha2/shared_types.go
cxhiano: Updated.

cxhiano:
I created https://github.com/kubernetes-sigs/gateway-api/issues/1021 for tracking documentation improvement.

apis/v1alpha2/shared_types.go
youngnick: Uh, there's no mention of the experimental status of Port here? Did I miss something?

youngnick:
I think this is looking great, aside from the small note about the experimental status of Port in the stable docs.

apis/v1alpha2/shared_types.go
cxhiano: Oops, for some reason my git push did not go through. It's updated now.

robscott:
@cxhiano looks like you need to run hack/update-codegen.sh, otherwise LGTM.

/approve

robscott:
Thanks!

/lgtm

PR #1004, opened 2022-01-31
Propose GRPCRoute Resource

gnossen:
**What type of PR is this?**

/kind gep

Propose a `GRPCRoute` resource.

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1016.md
+++ b/site-src/geps/gep-1016.md
@@ -97,122 +97,33 @@ in a proxyless service mesh.
 
 ### Cross Serving
 
-A new route type (and especially one encapsulated in an already supported
-protocol) raises questions about precedence of routing rules. Today, there is
-already a complicated algorithm dictating which routing rule to apply to a
-particular HTTP request when two `HTTPRoutes` both apply to the request's
-hostname. What semantics should apply when a `GRPCRoute` and an `HTTPRoute` both apply to
-a request?
+Occasionally, gRPC users will place gRPC services on the same hostname/port
+combination as HTTP services. For example, `foo.com:443/v1` might serve
+REST+JSON while `foo.com:443/com.foo.WidgetService/` serves gRPC. Such an
+arrangement in the Gateway APIs poses complex technical challenges. How are
+GRPCRoutes to be reconciled with HTTPRoutes? And how should invididual
+implementations accomplisht this?
 
-We propose that the GRPCRoute's matchers be translated into the equivalent
-HTTPRoute matchers and that the normal HTTPRoute precedence rules apply.
-This translation is mostly straightforward, but one significant deviation between
-GRPCRoute and HTTPRoute poses difficulties: the gRPC method matcher.
+After a long look at the implementations with which the author is familiar, it
+was deemed technically infeasible. Furthermore, after surveying the gRPC
+community, this was found to be a niche use case to begin with.
 
-gRPC methods are encoded in the URI of an HTTP/2 message in the format
-`/SERVICE/METHOD`. The matcher allows the user to supply these independently, so
-that the user may leave either blank, to be interpreted as an unconditional
-match. Further, this field may either be an exact match or a regex match.
+In any case, users wishing to accomplish this always have the option of using
+`HTTPRoute` resources to achieve this use case, at the cost of a degraded user
+experience.
 
-To translate this field to an HTTPRoute path matcher, we propose the following
-scheme:
+If at some point in the future, demand for this use case increases and we have
+reason to believe that the feasibility of implementation has improved, this
+would be a backward compatible change.
 
-- If neither service nor method is supplied and the type is `Exact`, translate to a
-  path match of `/` with type `PathPrefix`
-- If service and method are supplied and the type is `Exact`, translate to a
-  path match of `/SERVICE/METHOD` with type `Exact`
-- If service is supplied but not method and the type is `Exact`, translate to a
-  path match of `/SERVICE/` with type `PathPrefix`
-- Otherwise, translate the service and the method individually into regular
-  expressions (with an empty field interpreted as `.*`), and translate to an
-  HTTPRoute path matcher of type `RegularExpression` with value `/SERVICE_REGEX/MATCHER_REGEX`
+As such, implementations that support GRPCRoute must enforce uniqueness of
+hostnames between `GRPCRoute`s and `HTTPRoute`s. If a route (A) of type `HTTPRoute` or
+`GRPCRoute` is attached to a Listener and that listener already has another Route (B) of
+the other type attached and the intersection of the hostnames of A and B is
+non-empty, then the implementation must reject Route A. That is, the
+implementation must raise an 'Accepted' condition with a status of 'False' in
+the corresponding RouteParentStatus.
 
-
-#### An Example Implementation
-
-In the interest of proving the feasibility of implementation of this proposal,
-we include a sketch of an implementation for an xDS-based control plane.
-
-Suppose the following resources are applied to the cluster:
-
-```yaml
-apiVersion: gateway.networking.k8s.io/v1alpha2
-kind: Gateway
-metadata:
-  name: my-gateway
-spec:
-  gatewayClassName: acme-lb
-  listeners:
-  - name: http
-    protocol: HTTP
-    port: 80
----
-kind: GRPCRoute
-apiVersion: gateway.networking.k8s.io/v1alpha2
-metadata:
-  name: foo-grpcroute
-spec:
-  parentRefs:
-  - name: my-gateway
-  hostnames:
-  - foo.com
-  rules:
-    matches:
-    - method:
-        service: foo.bar
-    backendRefs:
-    - name: grpc-backend
-      port: 50051
----
-kind: HTTPRoute
-apiVersion: gateway.networking.k8s.io/v1alpha2
-metadata:
-  name: foo-httproute
-spec:
-  parentRefs:
-  - name: my-gateway
-  hostnames:
-  - foo.com
-  rules:
-    matches:
-    - path:
-        type: PathPrefix
-        value: /foo
-    backendRefs:
-    - name: http-backend
-      port: 80
-```
-
-According to the above scheme, we translate the gRPC method matcher to an HTTP
-path matcher of type `PathPrefix` and value `/foo.bar/`. Following the HTTPRoute
-precedence rules, `/foo.bar/` has more characters than `/foo`, so it has higher
-precedence. Hence, the GRPCRoute's rule must be evaluated before the HTTPRoute's
-rule.
-
-Accordingly, an xDS server should create something like the following
-[`RouteConfiguration`](https://github.com/envoyproxy/envoy/blob/379a7567b0d15cc9d0806052b82670c8cad76fc5/api/envoy/config/route/v3/route.proto#L26):
-
-```json
-{
-  "virtual_hosts": [
-    "domains": ["foo.com"],
-    "routes": [
-      {
-        "prefix": "/foo.bar/",
-        "route": {
-          "cluster": "grpc-backend-cluster"
-        }
-      },
-      {
-        "prefix": "/foo",
-        "route": {
-          "cluster": "http-backend-cluster"
-        }
-      }
-    ]
-  ]
-}
-```
 
 ## API
 
@@ -272,9 +183,11 @@ spec:
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // GRPCRoute provides a way to route gRPC requests. This includes the capability
-// to match requests by hostname, service, method, or header. Filters can be
+// to match requests by hostname, gRPC service, gRPC method, or HTTP/2 header. Filters can be
 // used to specify additional processing steps. Backends specify where matching
 // requests should be routed.
+//
+// Support: Extended
 type GRPCRoute struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
@@ -327,6 +240,13 @@ type GRPCRouteSpec struct {
 	// implementation must raise an 'Accepted' Condition with a status of
 	// `False` in the corresponding RouteParentStatus.
 	//
+	// If a Route (A) of type HTTPRoute or GRPCRoute is attached to a
+	// Listener and that listener already has another Route (B) of the other
+	// type attached and the intersection of the hostnames of A and B is
+	// non-empty, then the implementation must reject Route A. That is, the
+	// implementation must raise an 'Accepted' condition with a status of
+	// 'False' in the corresponding RouteParentStatus.
+	//
 	// Support: Core
 	//
 	// +optional
@@ -371,37 +291,18 @@ type GRPCRouteRule struct {
 	// See the documentation for GRPCRouteMatch on how to specify multiple
 	// match conditions that should be ANDed together.
 	//
-	// If no matches are specified, the matches every gRPC request.
+	// If no matches are specified, the implementation must match every gRPC request.
 	//
-	// Proxy or Load Balancer routing configuration generated from GRPCRoutes or HTTPRoutes
+	// Proxy or Load Balancer routing configuration generated from GRPCRoutes
 	// MUST prioritize rules based on the following criteria, continuing on
-	// ties. When comparing GRPCRoutes and HTTPRoutes, GRPCRoute matches are first converted
-	// to HTTPRoute matches according to the following scheme:
-	//
-	// * HeaderMatches are translated exactly, since the field are
-	//    one-to-one
-	// * If `MethodMatch.type` is `Exact` and neither the
-	//    `MethodMatch.service` nor `MethodMatch.method` fields are present,
-	//    translate to a `PathMatch` of `{"type": "PathPrefix", value: "/"}`
-	// * If `MethodMatch.type` is `Exact` and both the `MethodMatch.service`
-	//    and `MethodMatch.method` fields are present, translate to a `PathMatch`
-	//    of `{"type": "Exact", value: "/" + service + "/" + method}`
-	// * If `MethodMatch.type` is `Exact` and the `MethodMatch.service`
-	//    field is present but not the `MethodMatch.method` field, translate to a
-	//    `PathMatch` of `{"type": "PathPrefix", value: "/" + service}`
-	// - Otherwise, translate the service and the method individually into
-	//    regular expressions (with an absent field interpreted as `.*`) and
-	//    translate to a `PathMatch` of
-	//    `{"type": "RegularExpression", value: "/": serviceRegex + "/" + methodRegex}`
-	//
-	// Once the translation is applied, precedence must be given to the Rule with
-	// the largest number of:
+	// ties. Merging must not be done between GRPCRoutes and HTTPRoutes.
+	// Precedence must be given to the rule with the largest number of:
 	//
 	// * Characters in a matching non-wildcard hostname.
 	// * Characters in a matching hostname.
-        // * Characters in a matching path.
+        // * Characters in a matching service.
+        // * Characters in a matching method.
 	// * Header matches.
-        // * Query param matches.
 	//
 	// If ties still exist across multiple Routes, matching precedence MUST be
 	// determined in order of the following criteria, continuing on ties:
@@ -413,7 +314,6 @@ type GRPCRouteRule struct {
 	// If ties still exist within the Route that has been given precedence,
 	// matching precedence MUST be granted to the first matching rule meeting
 	// the above criteria.
-        //
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
@@ -499,7 +399,9 @@ type GRPCRouteMatch struct {
 // request service and/or method..
 type GRPCMethodMatch struct {
 	// Type specifies how to match against the service and/or method.
-	// Support: Core (Exact)
+	// Support: Core (Exact with method specified)
+	//
+	// Support Custom (Exact with no method specified)
 	//
 	// Support: Custom (RegularExpression)
 	//
@@ -794,10 +696,44 @@ type GRPCRequestMirrorFilter struct {
 }
 ```
 
-No additional options will be added to the `ProtocolType` string. The `"HTTP"`
-and `"HTTPS"` options correspond exactly to the required functionality for the
-two main modes of operation in gRPC. This also enables a `GRPCRoute` to share
-the same listener as an `HTTPRoute`.
+In addition, two new `ProtocolType` options will be added:
+
+```diff
+--- a/apis/v1alpha2/gateway_types.go
++++ b/apis/v1alpha2/gateway_types.go
+@@ -165,6 +165,11 @@ type Listener struct {
+        //   protocol layers as described above. If an implementation does not
+        //   ensure that both the SNI and Host header match the Listener hostname,
+        //   it MUST clearly document that.
++       // * H2C: The Listener Hostname MUST match the :authority pseudo-header of the request.
++       // * H2: The Listener Hostname SHOULD match at both the TLS and HTTP
++       //   protocol layers as described above. If an implementation does not
++       //   ensure that both the SNI and :authority pseudo-header match the Listener hostname,
++       //   it MUST clearly document that.
+        //
+        // For HTTPRoute and TLSRoute resources, there is an interaction with the
+        // `spec.hostnames` array. When both listener and route specify hostnames,
+@@ -279,6 +284,12 @@ const (
+ 
+        // Accepts UDP packets.
+        UDPProtocolType ProtocolType = "UDP"
++
++       // Accepts HTTP/2 sessions over TLS.
++       H2ProtocolType ProtocolType = "H2"
++
++       // Accepts cleartext HTTP/2 sessions.
++       H2CProtocolType ProtocolType = "H2C"
+ )
+ 
+ // GatewayTLSConfig describes a TLS configuration.
+```
+
+TLS-encryption is the default in HTTP/2, so H2C is the
+[marked](https://en.wikipedia.org/wiki/Markedness) option here.
+
+If the listener supports neither `H2` nor `H2C`, then the implementation MUST
+raise an `Accepted` Condition with a status of `False` in the corresponding
+RouteParentStatus.
 
 ## Future Enhancements
 
--- a/site-src/geps/gep-1016.md
+++ b/site-src/geps/gep-1016.md
@@ -667,6 +667,9 @@ If the listener supports neither `H2` nor `H2C`, then the implementation MUST
 raise an `Accepted` Condition with a status of `False` in the corresponding
 RouteParentStatus.
 
+Implementations supporting `H2` or `H2C` _must_ support HTTP/2 without first
+upgrading from HTTP/1.
+
 ## Beta Graduation Criteria
 
 - `GRPCRoute` has been implemented by at least 2 controllers.
--- a/site-src/geps/gep-1016.md
+++ b/site-src/geps/gep-1016.md
@@ -117,6 +117,8 @@ match. Further, this field may either be an exact match or a regex match.
 To translate this field to an HTTPRoute path matcher, we propose the following
 scheme:
 
+- If neither service nor method is supplied and the type is `Exact`, translate to a
+  path match of `/` with type `PathPrefix`
 - If service and method are supplied and the type is `Exact`, translate to a
   path match of `/SERVICE/METHOD` with type `Exact`
 - If service is supplied but not method and the type is `Exact`, translate to a
@@ -335,7 +337,7 @@ type GRPCRouteSpec struct {
         // 
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
-	// +kubebuilder:default={{matches: {{method: {type: "RegularExpression", service: ".*", method: ".*"}}}}}
+	// +kubebuilder:default={{matches: {{method: {type: "Exact"}}}}}
 	Rules []GRPCRouteRule `json:"rules,omitempty"`
 }
 
@@ -371,17 +373,35 @@ type GRPCRouteRule struct {
 	//
 	// If no matches are specified, the matches every gRPC request.
 	//
-	// Proxy or Load Balancer routing configuration generated from HTTPRoutes
+	// Proxy or Load Balancer routing configuration generated from GRPCRoutes or HTTPRoutes
 	// MUST prioritize rules based on the following criteria, continuing on
-	// ties. Precedence must be given to the the Rule with the largest number
-	// of:
+	// ties. When comparing GRPCRoutes and HTTPRoutes, GRPCRoute matches are first converted
+	// to HTTPRoute matches according to the following scheme:
+	//
+	// * HeaderMatches are translated exactly, since the field are
+	//    one-to-one
+	// * If `MethodMatch.type` is `Exact` and neither the
+	//    `MethodMatch.service` nor `MethodMatch.method` fields are present,
+	//    translate to a `PathMatch` of `{"type": "PathPrefix", value: "/"}`
+	// * If `MethodMatch.type` is `Exact` and both the `MethodMatch.service`
+	//    and `MethodMatch.method` fields are present, translate to a `PathMatch`
+	//    of `{"type": "Exact", value: "/" + service + "/" + method}`
+	// * If `MethodMatch.type` is `Exact` and the `MethodMatch.service`
+	//    field is present but not the `MethodMatch.method` field, translate to a
+	//    `PathMatch` of `{"type": "PathPrefix", value: "/" + service}`
+	// - Otherwise, translate the service and the method individually into
+	//    regular expressions (with an absent field interpreted as `.*`) and
+	//    translate to a `PathMatch` of
+	//    `{"type": "RegularExpression", value: "/": serviceRegex + "/" + methodRegex}`
+	//
+	// Once the translation is applied, precedence must be given to the Rule with
+	// the largest number of:
 	//
 	// * Characters in a matching non-wildcard hostname.
 	// * Characters in a matching hostname.
-	// * GRPCRoute over HTTPRoute
-	// * Characters in a matching service.
-	// * Characters in a matching method.
+        // * Characters in a matching path.
 	// * Header matches.
+        // * Query param matches.
 	//
 	// If ties still exist across multiple Routes, matching precedence MUST be
 	// determined in order of the following criteria, continuing on ties:
@@ -393,10 +413,11 @@ type GRPCRouteRule struct {
 	// If ties still exist within the Route that has been given precedence,
 	// matching precedence MUST be granted to the first matching rule meeting
 	// the above criteria.
+        //
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{method: {type: "RegularExpression", service: ".*", method: ".*"}}}
+	// +kubebuilder:default={{method: {type: "Exact"}}}
 	Matches []GRPCRouteMatch `json:"matches,omitempty"`
 
 	// Filters define the filters that are applied to requests that match
@@ -439,11 +460,34 @@ type GRPCRouteRule struct {
 	BackendRefs []GRPCBackendRef `json:"backendRefs,omitempty"`
 }
 
+// GRPCRouteMatch defines the predicate used to match requests to a given
+// action. Multiple match types are ANDed together, i.e. the match will
+// evaluate to true only if all conditions are satisfied.
+//
+// For example, the match below will match a gRPC request only if its service
+// is `foo` AND it contains the `version: v1` header:
+//
+// ```
+// match:
+//   method:
+//     type: Exact
+//     service: "foo"
+//   headers:
+//   - name: "version"
+//     value "v1"
+// ```
 type GRPCRouteMatch struct {
+	// Path specifies a gRPC request service/method matcher. If this field is not
+	// specified, all services and methods will match.
+	//
 	// +optional
-	// +kubebuilder:default={type: "RegularExpression", service: ".*", method: ".*"}
+	// +kubebuilder:default={type: "Exact"}
 	Method *GRPCMethodMatch `json:"path,omitempty"`
 
+	// Headers specifies gRPC request header matchers. Multiple match values are
+	// ANDed together, meaning, a request must match all the specified headers
+	// to select the route.
+	//
 	// +listType=map
 	// +listMapKey=name
 	// +optional
@@ -451,7 +495,10 @@ type GRPCRouteMatch struct {
 	Headers []GRPCHeaderMatch `json:"headers,omitempty"`
 }
 
+// GRPCPathMatch describes how to select a gRPC route by matching the gRPC
+// request service and/or method..
 type GRPCMethodMatch struct {
+	// Type specifies how to match against the service and/or method.
 	// Support: Core (Exact)
 	//
 	// Support: Custom (RegularExpression)
@@ -460,31 +507,70 @@ type GRPCMethodMatch struct {
 	// +kubebuilder:default=Exact
 	Type *GRPCMethodMatchType `json:"type,omitempty"`
 
+
+	// Value of the service to match against. If left empty or omitted, will
+	// match all services.
 	// +optional
 	// +kubebuilder:default=""
 	// +kubebuilder:validation:MaxLength=1024
 	Service *string `json:"value,omitempty"`
 
+	// Value of the method to match against. If left empty or omitted, will
+	// match all services.
 	// +optional
 	// +kubebuilder:default=""
 	// +kubebuilder:validation:MaxLength=1024
 	Method *string `json:"value,omitempty"`
-
-	// +optional
-	// +kubebuilder:default=true
-	CaseSensitive *bool `json:"value,omitempty"`
 }
 
+// MethodMatchType specifies the semantics of how gRPC methods and services should be compared.
+// Valid MethodMatchType values are:
+//
+// * "Exact"
+// * "RegularExpression"
+//
+// Exact paths must be syntactically valid:
+//
+// - Must not contain `/` character
+//
+// +kubebuilder:validation:Enum=Exact;PathPrefix;RegularExpression
 // +kubebuilder:validation:Enum=Exact;RegularExpression
 type GRPCMethodMatchType string
 
+const (
+	// Matches the service and/or method exactly and with case sensitivity.
+	PathMatchExact PathMatchType = "Exact"
+
+	// Matches if the service and/or method matches the given regular expression with
+	// case sensitivity.
+	//
+	// Since `"RegularExpression"` has custom conformance, implementations
+	// can support POSIX, PCRE, RE2 or any other regular expression dialect.
+	// Please read the implementation's documentation to determine the supported
+	// dialect.
+	PathMatchRegularExpression PathMatchType = "RegularExpression"
+)
+
+// GRPCHeaderMatch describes how to select a gRPC route by matching gRPC request
+// headers.
 type GRPCHeaderMatch struct {
+	// Type specifies how to match against the value of the header.
+	//
 	// +optional
 	// +kubebuilder:default=Exact
 	Type *HeaderMatchType `json:"type,omitempty"`
 
+	// Name is the name of the gRPC Header to be matched.
+	//
+	// If multiple entries specify equivalent header names, only the first
+	// entry with an equivalent name MUST be considered for a match. Subsequent
+	// entries with an equivalent header name MUST be ignored. Due to the
+	// case-insensitivity of header names, "foo" and "Foo" are considered
+	// equivalent.
 	Name GRPCHeaderName `json:"name"`
 
+	// Value is the value of the gRPC Header to be matched.
+	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=4096
 	Value string `json:"value"`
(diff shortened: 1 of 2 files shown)


site-src/geps/gep-XXXX.md
```diff
@@ -0,0 +1,248 @@
...
+spec:
+  parentRefs:
+  - name: my-gateway
+  hostnames:
+  - foo.com
+  - bar.com
+  rules:
+  - matches:
+      method:
+        grpcService: helloworld.Greeter
```
howardjohn, line 62: nit: is `grpc` needed? We know its grpc - we are in a GRPCroute

site-src/geps/gep-XXXX.md
```diff
@@ -0,0 +1,248 @@
...
+        grpcMethod:  SayHello
+      headers:
+      - type: Exact
+        name: magic
+        value: foo
+
+    filters:
+    - type: GRPCRequestHeaderModifierFilter
```
howardjohn, line 70: same GRPC prefix comment here and below

site-src/geps/gep-XXXX.md
gnossen: No. Good catch. I had already made this change in the structs, but forgot to reflect it in the example. Fixed.

site-src/geps/gep-XXXX.md
gnossen: Fixed.

site-src/geps/gep-XXXX.md
```diff
@@ -0,0 +1,248 @@
+# GEP-XXXX: GRPCRoute
```
youngnick, line 1: When you next take a pass on this @gnossen, it's definitely worth logging an "Consider adding a GRPCRoute" issue with the `kind/gep` label, to get a number to put here. Even if we end up not doing this, having this GEP preserved as a record of the discussion is valuable.

brianehlert:
Is the intention that this would also incorporate grpc-web in some way? Or the implementation would be strict to grpc.
I am asking since grpc-web is a project technically distinct from grpc with its own nuances.

gnossen:
I believe this proposal is now ready for a first round of reviews. @howardjohn @youngnick @robscott

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+      port: 80
+```
+
+According to the above scheme, we translate the gRPC method matcher to an HTTP
+path matcher of type `PathPrefix` and value `/foo.bar/`. Following the HTTPRoute
+precedence rules, `/foo.bar/` has more characters than `/foo`, so it has higher
+precedence. Hence, the GRPCRoute's rule must be evaluated before the HTTPRoute's
+rule.
+
+Accordingly, an xDS server should create something like the following
```
hbagdi, line 192: Best to remove specifics of implementation, I don't _think_ we do this anywhere else.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+  parentRefs:
+  - name: my-gateway
+  hostnames:
+  - foo.com
+  - bar.com
+  rules:
+  - matches:
+      method:
+        service: helloworld.Greeter
+        method:  SayHello
```
hbagdi, line 239: method.method feels off, how about?
```suggestion
      service: helloworld.Greeter
      method:  SayHello
```

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+HTTPRoute matchers and that the normal HTTPRoute precedence rules apply.
+This translation is mostly straightforward, but one significant deviation between
+GRPCRoute and HTTPRoute poses difficulties: the gRPC method matcher.
+
+gRPC methods are encoded in the URI of an HTTP/2 message in the format
+`/SERVICE/METHOD`. The matcher allows the user to supply these independently, so
+that the user may leave either blank, to be interpreted as an unconditional
+match. Further, this field may either be an exact match or a regex match.
+
+To translate this field to an HTTPRoute path matcher, we propose the following
```
hbagdi, line 117: Question: Is all of this translation visible to the end-user or is it implementation guidance?
Is that we expect a controller in the cluster that reconciles GRPCRoute into HTTPRoute?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+
+- If neither service nor method is supplied and the type is `Exact`, translate to a
+  path match of `/` with type `PathPrefix`
+- If service and method are supplied and the type is `Exact`, translate to a
+  path match of `/SERVICE/METHOD` with type `Exact`
+- If service is supplied but not method and the type is `Exact`, translate to a
+  path match of `/SERVICE/` with type `PathPrefix`
+- Otherwise, translate the service and the method individually into regular
+  expressions (with an empty field interpreted as `.*`), and translate to an
+  HTTPRoute path matcher of type `RegularExpression` with value `/SERVICE_REGEX/MATCHER_REGEX`
```
hbagdi, line 128: There is one missing piece - the protocol. We don't have an explicit HTTP/2 protocol and our current guidance is loose:

https://github.com/kubernetes-sigs/gateway-api/blob/3c8ea05823d3a4215f86b9b3d56316bf2e74a066/apis/v1alpha2/gateway_types.go#L265-L272

One could translate a gRPCRoute to HTTPRoute but if the underlying gateway doesn't support HTTP/2 or has it disabled, none of this would work.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=gateway-api
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Hostnames",type=string,JSONPath=`.spec.hostnames`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// GRPCRoute provides a way to route gRPC requests. This includes the capability
+// to match requests by hostname, service, method, or header. Filters can be
```
hbagdi, line 275:
```suggestion
// to match requests by hostname, gRPC-service, gRPC-method, or HTTP header. Filters can be
```

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+// used to specify additional processing steps. Backends specify where matching
+// requests should be routed.
+type GRPCRoute struct {
```
hbagdi, line 278: What is the conformance level for this route?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+  - CORS
+
+
+#### Proxyless Service Mesh
+
+The gRPC library supports proxyless service mesh, a system by which routing
+configuration is received not by an in-line proxy or sidecar proxy but by the client
+itself. Eventually, `GRPCRoute` in the Gateway APIs should support this feature.
+However, to date, there are no HTTP client libraries capable of participating
+in a proxyless service mesh.
```
hbagdi, line 94: I'm not very familiar with proxy-less SM. What are the responsibilities of implementation to support this? What changes (if any) are required in this API to support this?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	//    `{"type": "RegularExpression", value: "/": serviceRegex + "/" + methodRegex}`
+	//
+	// Once the translation is applied, precedence must be given to the Rule with
+	// the largest number of:
+	//
+	// * Characters in a matching non-wildcard hostname.
+	// * Characters in a matching hostname.
+        // * Characters in a matching path.
+	// * Header matches.
+        // * Query param matches.
```
hbagdi, line 404: No query-parameter matching is supported yet.
```suggestion
```

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	// determined in order of the following criteria, continuing on ties:
+	//
+	// * The oldest Route based on creation timestamp.
+	// * The Route appearing first in alphabetical order by
+	//   "{namespace}/{name}".
+	//
+	// If ties still exist within the Route that has been given precedence,
+	// matching precedence MUST be granted to the first matching rule meeting
+	// the above criteria.
+        //
```
hbagdi, line 416: What must a gateway do when a request doesn't match any route?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=1024
+	Method *string `json:"value,omitempty"`
+}
+
+// MethodMatchType specifies the semantics of how gRPC methods and services should be compared.
+// Valid MethodMatchType values are:
+//
+// * "Exact"
+// * "RegularExpression"
```
hbagdi, line 530: Conformance levels?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	Type *HeaderMatchType `json:"type,omitempty"`
+
+	// Name is the name of the gRPC Header to be matched.
+	//
+	// If multiple entries specify equivalent header names, only the first
+	// entry with an equivalent name MUST be considered for a match. Subsequent
+	// entries with an equivalent header name MUST be ignored. Due to the
+	// case-insensitivity of header names, "foo" and "Foo" are considered
+	// equivalent.
+	Name GRPCHeaderName `json:"name"`
```
hbagdi, line 570: Isn't a gRPC header the same as a HTTP header? My knowledge in this area is weak so please bear with me.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+
+	// Filters defined at this level should be executed if and only if the
+	// request is being forwarded to the backend defined here.
+	//
+	// Support: Custom (For broader support of filters, use the Filters field
+	// in GRPCRouteRule.)
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	Filters []GRPCRouteFilter `json:"filters,omitempty"`
```
hbagdi, line 618: Since the proposal is specific to gRPC using HTTP/2, what are the downsides of not introducing a new filter type and all the corresponding filters and instead re-using what exists in HTTP?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	PathMatchExact PathMatchType = "Exact"
+
+	// Matches if the service and/or method matches the given regular expression with
+	// case sensitivity.
+	//
+	// Since `"RegularExpression"` has custom conformance, implementations
+	// can support POSIX, PCRE, RE2 or any other regular expression dialect.
+	// Please read the implementation's documentation to determine the supported
+	// dialect.
+	PathMatchRegularExpression PathMatchType = "RegularExpression"
```
pleshakov, line 551: Would it make sense to also consider matching based on the prefix ? For example, to match based on the package:
```yaml
  rules:
    matches:
    - service: foo. # route traffic that belongs to package foo
      type: Prefix
    backendRefs:
    - name: grpc-backend
      port: 50051
 ```
 
It looks like it is possible to configure that using the RegularExpression match type. However, prefix matching in proxies are more efficient than regex-based matching.

Is it a viable use case?

File: site-src/geps/gep-1016.md [hunk 17]
```diff
@@ -0,0 +1,813 @@
...
+- [the CRI](https://github.com/kubernetes/cri-api/blob/49fe8b135f4556ea603b1b49470f8365b62f808e/README.md),
+- [the device plugin framework](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/)
+
+Given gRPC's importance in the application-layer networking space and to
+the Kubernetes project in particular, we must ensure that the gRPC control plane
+configuration landscape does not Balkanize.
+
+### Encapsulated Network Protocols
+
+At the time of writing, the only kind of route officially defined by the Gateway
```
youngnick, line 43: This is a little incorrect, we curently have HTTPRoute, TLSRoute, TCPRoute, and UDPRoute. Could we say that HTTPRoute is the only layer 7 specific route type? Probably.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+the level of HTTP. This is why we propose a new resource.
+
+In setting this precendent, we must also introduce a coherent policy for _when_
+to introduce a custom `Route` resource for an encapsulated protocol for which a
+lower layer protocol already exists. We propose the following criteria for such
+an addition.
+
+- Users of the encapsulated protocol would miss out on significant conventional features from their ecosystem if forced to route at a lower layer.
+- Users of the enapsulated protocol would experience a degraded user experience if forced to route at a lower layer.
+- The encapsulated protocol has a significant user base, particularly in the Kubernetes community.
```
youngnick, line 59: No change here, just wanted to point out that this is a great set of rules that we probably should pull out into a separate doc somewhere.

site-src/geps/gep-1016.md
youngnick: This feels like implementation guidance to me - that the GRPCRoute object should be translated into data-path config equivalent to the object. I am a -1 on actually automatically generating an actual HTTPRoute object.

site-src/geps/gep-1016.md
youngnick: I agree we haven't done this very much, but I think the example benefits from showing the eventual LB config in this case - the fact that's an xDS config isn't that important.

site-src/geps/gep-1016.md
youngnick: Since we only have one conformance profile at the moment, essentially, I'd say its Extended.

youngnick (commented):
Not much extra to add on what others have commented.

site-src/geps/gep-1016.md
hbagdi: +1 on your -1

site-src/geps/gep-1016.md
gnossen: This section has been removed due to a change in merge semantics.

site-src/geps/gep-1016.md
gnossen: Do note that there's another field here specifiying the _type_ of the match. Should the `method.type` field become a `methodType` field?

site-src/geps/gep-1016.md
gnossen: This section has been removed due to a change in merge semantics.

site-src/geps/gep-1016.md
gnossen: I see. I was under the impression that the `HTTP` protocol implied `HTTP2`. I will firm this up.

site-src/geps/gep-1016.md
gnossen: Added.

site-src/geps/gep-1016.md
gnossen: Let me start by caveating that proxyless service mesh support is not actually part of *this* proposal.

With that said, I don't believe any changes will be necessary to the `GRPCRoute` resource to enable this. Instead, attachment to a `Mesh` object, referenced through the`parentRefs` field will activate this support. Another proposal for a `Mesh` resource is planned for sometime in the next couple of quarters.

site-src/geps/gep-1016.md
gnossen: Hm. This was pulled from the HTTPRoute docstrings. Removed.

site-src/geps/gep-1016.md
gnossen: My instinct is that we should do the same thing as HTTPRoute here, but I can't find documentation on what HTTPRoute implementations must do in this case.

site-src/geps/gep-1016.md
gnossen: This is specified a bit higher up, in the `GRPCMethodMatch` struct.

site-src/geps/gep-1016.md
gnossen: This is true, roughly speaking. More details [here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).

site-src/geps/gep-1016.md
gnossen: The downside is that then we would not be able to add extra gRPC-specific filters. I imagine most of the items in the "Future Enhancements" section will need to be implemented as new gRPC-specific filters.

site-src/geps/gep-1016.md
gnossen: I've considered this. The problem is that allowing prefix matching on the _service_ with a method specified would actually be an HTTP/2 _infix_ matcher. For example, consider the following:

```
  rules:
    matches:
    - service: foo.
      method: bar
      type: Prefix
```

This would match HTTP/2 URIs conforming to this regex:

```
/foo\..*/bar.*
```

It also becomes much more difficult to justify the fact that both the service and the method share a matcher type if you allow prefix match. Even if you had both a `serviceType` and a `methodType`, you still have this infix matcher problem though.

My preference is to keep it simple for the moment and go with what's currently included in the proposal. If folks feel strongly about this though, I'm happy to hear the feedback.

site-src/geps/gep-1016.md
gnossen: Wasn't actually aware of this. I was going off of the this page, which seems to list only [`HTTPRoute`](https://gateway-api.sigs.k8s.io/v1alpha2/api-types/gatewayclass/). I'll correct this.

site-src/geps/gep-1016.md
gnossen: Thanks. Any guidance on placement?

gnossen:
Sorry for the long turnaround time on addressing review comments. I was resolving a discussion on the implementability of this proposal. As a result, I've changed it slightly. We no longer allow hostnames to be shared between `GRPCRoute`s and `HTTPRoute`s. Users will always have the ability to use an `HTTPRoute` if they really must share the hostname and it's a backward-compatible change to allow these in the future if we find strong demand for it and also find a way to make it feasible from a technical perspective. Details are included in the updated proposal.

I've also added two new protocol types: "H2" and "H2C".

gnossen:
@youngnick @bowei @pleshakov @hbagdi Friendly ping. Ready for the next round of review.

gnossen:
@youngnick @bowei @pleshakov @hbagdi Friendly ping. Waiting on the next round of review.

site-src/geps/gep-1016.md
robscott: I'd assumed this only supported an exact match, how would a prefix or regex match work when there are 2 values specified (service and method)?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	// Proxy or Load Balancer routing configuration generated from GRPCRoutes or HTTPRoutes
+	// MUST prioritize rules based on the following criteria, continuing on
+	// ties. When comparing GRPCRoutes and HTTPRoutes, GRPCRoute matches are first converted
+	// to HTTPRoute matches according to the following scheme:
+	//
+	// * HeaderMatches are translated exactly, since the field are
+	//    one-to-one
+	// * If `MethodMatch.type` is `Exact` and neither the
+	//    `MethodMatch.service` nor `MethodMatch.method` fields are present,
+	//    translate to a `PathMatch` of `{"type": "PathPrefix", value: "/"}`
```
robscott, line 385: Would exact match on "/" be a better translation here?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+	// * If `MethodMatch.type` is `Exact` and both the `MethodMatch.service`
+	//    and `MethodMatch.method` fields are present, translate to a `PathMatch`
+	//    of `{"type": "Exact", value: "/" + service + "/" + method}`
+	// * If `MethodMatch.type` is `Exact` and the `MethodMatch.service`
+	//    field is present but not the `MethodMatch.method` field, translate to a
+	//    `PathMatch` of `{"type": "PathPrefix", value: "/" + service}`
+	// - Otherwise, translate the service and the method individually into
+	//    regular expressions (with an absent field interpreted as `.*`) and
+	//    translate to a `PathMatch` of
+	//    `{"type": "RegularExpression", value: "/": serviceRegex + "/" + methodRegex}`
```
robscott, line 395: To clarify, this would essentially translate an empty Method match with a `foo` Service to `type: PathPrefix, value: /foo`, and a regex would only be required if the match specified a path but not a Service? Not very familiar with gRPC, but is that last option a very common one? Would it be valid to just not support that?

site-src/geps/gep-1016.md
robscott: Do we need match types here? If so, do they apply to both Service and Method? Should we split those into separate matches?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+
+	// RequestHeaderModifier defines a schema for a filter that modifies request
+	// headers.
+	//
+	// Support: Core
+	//
+	// Support: Core
+	//
+	// +optional
+	RequestHeaderModifier *GRPCRequestHeaderFilter `json:"requestHeaderModifier,omitempty"`
```
robscott, line 666: Can we reuse the same type we have for HTTP header modifier here?

robscott (commented):
Thanks for your work on this @gnossen!

site-src/geps/gep-1016.md
gnossen: The only two supported types here are "exact" and "regular expression". Here's a full matrix of combinations and their equivalent URI matchers:

## Type Exact

|Service|Method|URI Matcher|
|----------|----------|-----------|
|Specified|Specified|Exact `/${SERVICE}/${METHOD}`|
|Specified|Unspecified|Prefix `/${SERVICE}/`|
|Unspecified|Specified|Suffix `/${METHOD}/` or Regex `/.+/${METHOD}`|
|Unspecified|Unspecified|Prefix `/`|

## Type RegularExpression

|Service|Method|URI Matcher|
|----------|----------|-----------|
|Specified|Specified|Regex `/${SERVICE}/${METHOD}`|
|Specified|Unspecified|Regex `/${SERVICE}/.+`|
|Unspecified|Specified|Regex `/.+/${METHOD}`|
|Unspecified|Unspecified|Prefix `/`|

I'll add this into the proposal to clear this up.

site-src/geps/gep-1016.md
gnossen: Removed this section since we're no longer proposing merge between `GRPCRoute` and `HTTPRoute` (but my last `git push` failed -- whoopsie)

Regardless, this semantic still applies. I think an unspecified method and unspecified match should match _all_ gRPC traffic. Do you think there's an issue with this?

An exact match on `/` is not even a valid gRPC+protobuf URI, so I don't think this makes much sense.

site-src/geps/gep-1016.md
gnossen: > To clarify, this would essentially translate an empty Method match with a foo Service to type: PathPrefix, value: /foo

Correct.

> a regex would only be required if the match specified a path but not a Service

You can imagine many different use cases for regexes. Many gRPC APIs are resource oriented. You may want to route only methods that start with `Get`. In this case, you'd want to do something like:

```
method:
 type: RegularExpression
 service: foo.org
 method: Get.+
```

Or you might do something similar with services. Say you want to route both v1 and v2 of a service. What's nice here is that exact matches are a subset of regular expressions, so the fact that there's only one type for two values here doesn't actually get in the way of expressing the routing logic you want.

> Would it be valid to just not support that?

It would be valid, but it would also be limiting. This is why I have listed "Custom" support for the RegularExpression matcher type.

site-src/geps/gep-1016.md
gnossen: > Do we need match types here?

I think so. See my use cases for regex matches [here](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r826430565).

> If so, do they apply to both Service and Method?

Yes.

> Should we split those into separate matches?

I don't see the value add in doing this. Again, see [this comment](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r826430565).

site-src/geps/gep-1016.md
gnossen: Sure. That seems reasonable. I wasn't sure if there was some rule against sharing filters between different route types.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,768 @@
...
+ // GatewayTLSConfig describes a TLS configuration.
+```
+
+TLS-encryption is the default in HTTP/2, so H2C is the
+[marked](https://en.wikipedia.org/wiki/Markedness) option here.
+
+If the listener supports neither `H2` nor `H2C`, then the implementation MUST
+raise an `Accepted` Condition with a status of `False` in the corresponding
+RouteParentStatus.
+
```
youngnick, line 756: I know we don't always ask for this, but because this is the first time we're adding a new Route resource, I'd like to see something about the graduation plan.

Based on what we have in #1039, I'd suggest something like:
- gRPC resource implemented by a couple of implementations
- Some conformance tests in place
- Users are using the API surface
- API review done by upstream reviewers.

Graduation to GA should be similar but more stringent (exhaustive conformance tests, implemented by lots o implementations, etc).

youngnick (commented):
This is looking great, just a couple of outstanding questions from me. Sorry about the review delay.

gnossen:
@youngnick @bowei @pleshakov @hbagdi Friendly ping. Ready for the next round of review.

site-src/geps/gep-1016.md
youngnick: Okay, that makes sense to me. Consider my objection resolved, but perhaps something addressing the stutter in the doc to preserve this for later?

site-src/geps/gep-1016.md
youngnick: I think the expectation is that unsupported entries in this list should result in admission failure for the resource, just like it does for HTTPRoute, so reusing the same type spec should be fine.

youngnick (commented):
This LGTM with one small request to record the discussion about stuttering so we don't have to rehash it at some point.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,781 @@
...
+While certain gRPC implementations support multiple transports and multiple
+interface definition languages (IDLs), this proposal limits itself to
+[HTTP/2](https://developers.google.com/web/fundamentals/performance/http2) as
+the transport and [Protocol Buffers](https://developers.google.com/protocol-buffers)
+as the IDL, which makes up the vast majority of gRPC traffic in the wild.
+
+## Introduction
+
+While it would be possible to support gRPC via CRDs, in the long run, this would
```
bowei, line 24: To be clear, the whole API is done with CRDs. Do you mean custom, out-of-tree resources?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,781 @@
...
+#### User Experience
+
+The user experience would also degrade significantly if forced to route at the level of HTTP.
+
+- Encoding services and methods as URIs (an implementation detail of gRPC)
+- The [Transfer Encoding header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding) for trailers
+- Many features supported by HTTP/2 but not by gRPC, such as
+  - Query parameters
+  - Methods besides `POST`
+  - CORS
```
bowei, line 78: CORS headers (might want to add a link https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS),

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,781 @@
...
+
+
+#### Proxyless Service Mesh
+
+The gRPC library supports proxyless service mesh, a system by which routing
+configuration is received not by an in-line proxy or sidecar proxy but by the client
```
bowei, line 84: nit: configuration and routing is implemented by the client stub itself (receiving is just one part of the story)

bowei (commented):
LGTM 

We can take the specific API review comments to the actual API PR

youngnick:
/hold for maintainer LGTMs

@hbagdi @howardjohn, do you have any further comments? We'll give one more week starting from the time this is posted.

I'll LGTM once my comments are addressed, @bowei can do the same, and then once we have one more maintainer LGTM, we are good to go.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+- Users of the encapsulated protocol would miss out on significant conventional features from their ecosystem if forced to route at a lower layer.
+- Users of the enapsulated protocol would experience a degraded user experience if forced to route at a lower layer.
+- The encapsulated protocol has a significant user base, particularly in the Kubernetes community.
+
+gRPC meets _all_ of these criteria and is therefore, we contend, a strong
+candidate for inclusion in the Gateway APIs.
+
+#### HTTP/2 Cleartext
+
+gRPC allows HTTP/2 cleartext communication (H2C). This is conventionally deployed for
```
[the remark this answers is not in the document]
gnossen: This _is_ in scope. Note the two new added `ProtocolType` options.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+testing. Many control plane implementations do not support this by default and
+would require special configuration to work properly.
+
+#### Content-Based Routing
+
+While not included in the scope of this initial GEP, a common use case cited for
```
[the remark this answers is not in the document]
gnossen: I'm not sure I follow? This sentence is calling out a commonly requested use case for `GRPCRoute` as a justification for its existence while simultaneously calling that use case out of scope for this GEP.

site-src/geps/gep-1016.md
gnossen: Yes. Clarified.

site-src/geps/gep-1016.md
gnossen: Clarified.

site-src/geps/gep-1016.md
gnossen: Done.

site-src/geps/gep-1016.md
gnossen: Switched `GRPCRequestHeaderFilter` to `HTTPRequestHeaderFilter` and `GRPCRequestMirrorFilter` to `HTTPRequestMirrorFilter`.

gnossen:
Sorry for the delay on responding to these last few review comments.

@bowei @youngnick I believe I've addressed everything you've brought up.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,694 @@
...
+        //
+        // For HTTPRoute and TLSRoute resources, there is an interaction with the
+        // `spec.hostnames` array. When both listener and route specify hostnames,
+@@ -279,6 +284,12 @@ const (
+ 
+        // Accepts UDP packets.
+        UDPProtocolType ProtocolType = "UDP"
++
++       // Accepts HTTP/2 sessions over TLS.
++       H2ProtocolType ProtocolType = "H2"
```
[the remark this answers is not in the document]
gnossen: Because `HTTPS` doesn't necessarily imply support for HTTP/2. You could have an `HTTPS` implementation that only supports HTTP/1.1. An implementation supporting `H2` would guarantee that HTTP/2 over TLS with no upgrade from HTTP/1 works.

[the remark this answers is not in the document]
howardjohn: It seems a bit inconsistent... so we now have:

* `HTTP`: h1, optionally h2c
* `HTTPS`: h1 or h2 (over TLS)
* `H2C`: only h2c
* `HTTP2`: only h2 (over TLS)

I can't imagine any user could see those 4 options and understand what they mean without a close inspection of the docs

Also...
> You could have an HTTPS implementation that only supports HTTP/1.1

Doesn't the spec not allow that?

```
	// Accepts HTTP/1.1 or HTTP/2 sessions over TLS.
	HTTPSProtocolType ProtocolType = "HTTPS"
```

> An implementation supporting H2 would guarantee that HTTP/2 over TLS with no upgrade from HTTP/1 works.

It guarantees HTTP/2 works, but does it guarantee HTTP/1 does *not* work?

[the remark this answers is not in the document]
howardjohn: How does it work when I have `HTTPS` and `H2` (or `HTTP` and `H2C`) on the same port?

If we say its a conflict, then that means that if I want to do h1 and h2 on the same port, I need to name the port `HTTP`, not `H2C`. So we will need to tell users "If you want to use HTTP2 over clear text, do NOT set protocol: h2c".

If its not a conflict and they can merge, then that would imply we can accept HTTP/1.1 over the H2C/HTTP2 protocols which seems unexpected?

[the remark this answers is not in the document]
gnossen: > Doesn't the spec not allow that?

The word "or" is doing a lot of work here. If we should be interpreting this to mean "Accepts HTTP/1.1 or (HTTP/1.1. and HTTP/2)", then I suppose so.

> How does it work when I have HTTPS and H2 (or HTTP and H2C) on the same port?

I would expect `HTTPS` and `H2` on the same port to mean that the port would accept any of the following over TLS:

- HTTP/1.1
- HTTP/2 via upgrade from HTTP/1.1
- HTTP/2 directly (i.e. without upgrade from HTTP/1.1)

Whereas `H2` on its own would guarantee only:

- HTTP/2 directly (i.e. without upgrade from HTTP/1.1)

While `HTTPS` would guarantee:

- HTTP/1.1
- HTTP/2 via upgrade from HTTP/1.1

> If its not a conflict and they can merge, then that would imply we can accept HTTP/1.1 over the H2C/HTTP2 protocols which seems unexpected?

I don't think this is true. A listener with both `HTTPS` and `H2` that accepts HTTP/1.1 would be doing so under the capabilities of `HTTPS`, not of `H2`. This seems similar to what you'd expect for a port that accepts both TCP and HTTP (or is this not allowed?)


More generally, I need to solve two problems here:

- Providing an indication that an implementation definitely supports HTTP/2 (this may be solved already, as you pointed out, depending on how you read the current spec for HTTPS)
- Providing an indication that an implementation definitely accepts HTTP/2 without first upgrading from HTTP/1. Not all servers do this.

If you can propose an alternative that accomplishes these two goals, I'd be more than happy to adopt it.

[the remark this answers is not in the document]
howardjohn: When you say upgrade are you referring to https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header#Use_with_HTTP/2 or something else? Since "The HTTP Upgrade mechanism is used only for cleartext HTTP2 (h2c)"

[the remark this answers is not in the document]
howardjohn: Given both the HTTP and HTTPS protocol, in the current spec, allow HTTP2, which don't we just say that GRPC can only bind to HTTP/HTTPS, and the GRPCRoutes only accept HTTP/2 requests.

I don't see why we need 'Providing an indication that an implementation definitely supports HTTP/2'? if the implementation accepts GRPCRoutes then they must accept HTTP/2 (we can declare that in the spec if needed)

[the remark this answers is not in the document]
gnossen: > When you say upgrade are you referring to https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header#Use_with_HTTP/2

Yes.

> if the implementation accepts GRPCRoutes then they must accept HTTP/2 (we can declare that in the spec if needed)

I was under the impression that this sort of requirement was expected to be expressed through the `ProtocolType`. I have no problem with this just being part of the spec for GRPCRoute.

I have switched to making this requirement in the spec for GRPCRoute and have removed the `H2` and `H2C` ProtocolTypes entirely.

site-src/geps/gep-1016.md
hbagdi: Is there a 404 equivalent?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,685 @@
...
+	// Matches define conditions used for matching the rule against incoming
+	// gRPC requests. Each match is independent, i.e. this rule will be matched
+	// if **any** one of the matches is satisfied.
+	//
+	// For example, take the following matches configuration:
+	//
+	// ```
+	// matches:
+	// - method:
+	//     method: foo.bar
```
hbagdi, line 343: nit: `foo.bar` a valid gRPC method name?

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,685 @@
...
+	// Conformance-levels at this level are defined based on the type of filter:
+	//
+	// - ALL core filters MUST be supported by all implementations.
+	// - Implementers are encouraged to support extended filters.
+	// - Implementation-specific custom filters have no API guarantees across
+	//   implementations.
+	//
+	// Specifying a core filter multiple times has unspecified or custom
+	// conformance.
+	// Support: Core
```
hbagdi, line 404: This has come before so I want to clarify.
Since gRPC filters are part of a bigger feature, gRPCRoute, they shouldn't have a conformance level that is higher than the parent.

Should this be `extended`?

site-src/geps/gep-1016.md
hbagdi: It may make sense to call this "HeaderName" instead to avoid unnecessary confusion. The "GRPC" prefix doesn't bring any additional value but can cause confusion here.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,685 @@
...
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=1024
+	Service *string `json:"value,omitempty"`
+
+	// Value of the method to match against. If left empty or omitted, will
+	// match all services.
+	// +optional
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=1024
+	Method *string `json:"value,omitempty"`
```
hbagdi, line 491: I find this pretty confusing and we should do a bit of optimization for the easy path.

Based on the defaults, the values are:
```
type: Exact
method: ""
service: ""
```

No matter what we document, this - 'exact match on empty strings' makes little sense to me and I _feel_ it will make little sense to anyone.
If the type is Exact, then the values should contain something.
My experience is that using empty values to mean everything doesn't work well, especially in k8s.

Some options we can explore:
1. Can the code require at least one of method or service?
2. Can we ask the user to provide an explicit `*` as the value of the `method` or `service` to mean that they would like to match any method or service respectively? IMHO, that makes the configuration much more self-documenting.
3. If type=Exact, the code should be exacting. Splitting `type` into two types makes the configuration verbose but it also makes it a bit more machine-readable. Let's say the type=regex and then one of them is a regex while the other one is not, how is the code supposed to figure it out?
4. Exact with the method specified is core and no method specified is custom seems arbitrary and confusing. Can we classify both at the same level?

I should note that the above is done with the right technical reasons, I don't doubt. It doesn't feel it is intuitive enough.

Thoughts?

site-src/geps/gep-1016.md
gnossen: I think I need some guidance here.

> Since gRPC filters are part of a bigger feature, gRPCRoute, they shouldn't have a conformance level that is higher than the parent.

I absolutely agree with this. In the places where I've written "Core" in this proposal, I've meant "Core to GRPCRoute."  Likewise, "Custom" here means "Custom to GRPCRoute."

If I write the support levels in absolute terms, then I think that would flatten things out and everything would become "Support: extended." This loses much of the detail that I've written in about support levels between different GRPCRoute implementations.

Thoughts on how to resolve this?

site-src/geps/gep-1016.md
gnossen: Ah, good catch. Should be `service` here, not `method`. Fixed.

site-src/geps/gep-1016.md
gnossen: > Based on the defaults, the values are:

This is true. Those are the default values. But people generally don't see the default values, they see the default _behavior_. And the default _behavior_ is to match all traffic.

> My experience is that using empty values to mean everything doesn't work well, especially in k8s.

Can you provide some concrete examples?

> Can the code require at least one of method or service?

Considering that supplying no match at all already matches all traffic, I'm totally open to this option. Going ahead and making this change. Anyone objecting, please jump in.

> Can we ask the user to provide an explicit * as the value of the method or service to mean that they would like to match any method or service respectively? IMHO, that makes the configuration much more self-documenting.

I'm not the biggest fan of this option. As far as I know, no equivalent syntax exists for HTTPRoute. This could lead to confusion.

> If type=Exact, the code should be exacting. Splitting type into two types makes the configuration verbose but it also makes it a bit more machine-readable.

I don't quite see how this solves the problem you introduced at the beginning of this comment? And as you point out, this would not be as nice a UX.

> Let's say the type=regex and then one of them is a regex while the other one is not, how is the code supposed to figure it out?

Discussed already [here](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r826430565).

Specifically, this:

> What's nice here is that exact matches are a subset of regular expressions, so the fact that there's only one type for two values here doesn't actually get in the way of expressing the routing logic you want. 

In practice, this shouldn't be a limitation.

> Exact with the method specified is core and no method specified is custom seems arbitrary and confusing. Can we classify both at the same level?

This actually was mistaken. The combination that presents trouble for implementations is `type: EXACT` with service unspecified and method specified. I've corrected this in the proposal.

I expect that many implementations will add support by translating to HTTPRoutes. In [this chart](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r826425912) from earlier, you see that this combination is equivalent to either a suffix matcher or a regex matcher of the form `/.+/${METHOD}`. Suffix matchers don't exist for HTTPRoute and Regex matchers have "Custom" support.  This case is called out to have the same support level as an HTTPRoute regex matcher to allow implementations that support HTTPRoute but not regex to also support GRPCRoute at some level.

If you have an alternative to address this problem, I'm open to suggestions.

site-src/geps/gep-1016.md
hbagdi: > I absolutely agree with this. In the places where I've written "Core" in this proposal, I've meant "Core to GRPCRoute." Likewise, "Custom" here means "Custom to GRPCRoute."

That is how I read this as well but we run into issues such as #655.

> If I write the support levels in absolute terms, then I think that would flatten things out and everything would become "Support: extended." This loses much of the detail that I've written in about support levels between different GRPCRoute implementations.

Extended features must be portable _if_ they are supported. Is that not enough for defining semantics here?

cc @robscott @youngnick

youngnick: I don't think we've explicitly discussed this before, but I've always conceived of the field-level support working like @gnossen described - this is because we clearly _have_ to allow implementations not to support some objects, because Layer 7 ingress controllers may not be able to do Layer 4 resources (like TCPRoute and UDPRoute), and Layer 4 load balancers may not be able to do higher-layer resources (like HTTPRoute or TLSRoute).

So, the "Support: Core" markings on fields have to be on a per-object basis.

But like I said, this is a subtle point that I don't think we've explicitly made before.

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,813 @@
...
+
+gRPC allows HTTP/2 cleartext communication (H2C). This is conventionally deployed for
+testing. Many control plane implementations do not support this by default and
+would require special configuration to work properly.
+
+#### Content-Based Routing
+
+While not included in the scope of this initial GEP, a common use case cited for
+routing gRPC is payload-aware routing. That is, routing rules which determine a
+backend based on the contents of the protocol buffer payload.
```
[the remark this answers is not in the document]
gnossen: Yes. Note added about this.

youngnick:
/lgtm

but will leave this for @hbagdi to come back to for approval and his lgtm

site-src/geps/gep-1016.md
```diff
@@ -0,0 +1,704 @@
...
+	Headers []GRPCHeaderMatch `json:"headers,omitempty"`
+}
+
+// GRPCPathMatch describes how to select a gRPC route by matching the gRPC
+// request service and/or method..
+//
+// At least one of Service and Method must be a non-empty string.
+type GRPCMethodMatch struct {
+	// Type specifies how to match against the service and/or method.
+	// Support: Core (Exact)
```
hbagdi, line 481:
```suggestion
	// Support: Core (Exact with service and method specified)
```

site-src/geps/gep-1016.md
hbagdi: > This is true. Those are the default values. But people generally don't see the default values, they see the default behavior. And the default behavior is to match all traffic.

In that case, the default values _should_ communicate the default behavior well.

> Can you provide some concrete examples?

Empty label selector behavior comes to mind.


> Considering that supplying no match at all already matches all traffic, I'm totally open to this option. Going ahead and making this change. Anyone objecting, please jump in.

Great. Could you create a tracking issue for this? This will require updates to validating webhook; the last time I looked, controller-runtime didn't allow for "at least one" JSON-schema validations. 


> > Can we ask the user to provide an explicit * as the value of the method or service to mean that they would like to match any method or service respectively? IMHO, that makes the configuration much more self-documenting.

> I'm not the biggest fan of this option. As far as I know, no equivalent syntax exists for HTTPRoute.

In HTTPRoute, the default value for path is `/` with the type `PathPrefix`.

gnossen: > Great. Could you create a tracking issue for this? This will require updates to validating webhook; the last time I looked, controller-runtime didn't allow for "at least one" JSON-schema validations.

[Created.](https://github.com/kubernetes-sigs/gateway-api/issues/1098)

site-src/geps/gep-1016.md
robscott: I agree with the interpretation here. We do cover this in the docs, but it may not be prominent enough (https://gateway-api.sigs.k8s.io/concepts/guidelines/#overlapping-support-levels).

site-src/geps/gep-1016.md
robscott: I agree with Nick and Harry here that the stutter does not feel quite right. The explanation here makes me mostly OK with it, but since I think this will be a common critique, it would be good to document the rationale in the GEP.

robscott:
Thanks for the work on this @gnossen! This is a well written and massive GEP. I'm deferring to @hbagdi for final LGTM.

/approve

robscott:
Looks like we're good to merge this in, thanks for all the work on this @gnossen!

/hold cancel

PR #1019, opened 2022-02-14
API overview fixes

markmc:
A few small fixes:

- **don't specify full namespaces.selector path:** the section describing allowedRoutes.namespaces field confusingly uses the full path for the namespace selector, suggesting it's talking about different sections of the spec.
- **qualify allowedRoutes namespaces and kinds:** without qualification, we're currently suggesting that namespaces and kinds are listener fields, like hostname.
- **fix namespace typos:** the route attachment examples moved from foo-namespace and bar-namespace in commit #792

/kind cleanup
/kind documentation


robscott:
Great fixes, thanks @markmc!

/lgtm
/approve

PR #1023, opened 2022-02-18
feat: allow URL overrides for conformance test manifests

shaneutt:
**What type of PR is this?**

/kind feature

The conformance tests include some hard-coded file paths for testing manifests. This patch makes it possible to override them with your own file paths and additionally enables URL support so that Golang projects can more easily run the test suite by using the [Golang Package](https://pkg.go.dev/sigs.k8s.io/gateway-api).


youngnick:
/lgtm

Nice change, I can see it will make running downstream conformance much easier.

robscott:
Thanks @shaneutt!

/approve

PR #808, opened 2021-08-23
docs: update tcp guide for v1alpha2

shaneutt:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

This updates the `TCP` guide for `v1alpha2`.

**Which issue(s) this PR fixes**:

This resolves the 5th task of https://github.com/kubernetes-sigs/gateway-api/issues/783

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1025, opened 2022-02-28
tcp guide: remove section on metadata based tcp matches

markmc:
In #808, we documented how the tcp route match extension point can be used for metadata based tcp matching.

Soon after, in #829, this extension point was removed, but the docs remained.

Remove this section to reflect the current state of the API.

/kind cleanup
/kind documentation


robscott:
Thanks @markmc!

/ok-to-test
/lgtm
/approve

PR #1027, opened 2022-03-02
fix(conformance): don't use canceled context to Delete resources

michaelbeaumont:
**What type of PR is this?**
/kind bug

Fixes the conformance tests cleanup. Functions registered with `Cleanup` will run after `ctx` from the outer `for` loop scope has already been canceled.

```release-note
NONE
```


robscott:
Thanks @michaelbeaumont! 

/lgtm
/approve

PR #1029, opened 2022-03-04
Fixing error handling and typo in roundtripper

robscott:
**What type of PR is this?**
/kind bug

A couple small fixes for roundtripper.

```release-note
Conformance tests: Improved error handling for HTTP requests
```

### DIFF
--- a/conformance/utils/roundtripper/roundtripper.go
+++ b/conformance/utils/roundtripper/roundtripper.go
@@ -43,7 +43,7 @@ type Request struct {
 	Headers  map[string][]string
 }
 
-// CapturedResponse contains request metadata captured from an echoserver
+// CapturedRequest contains request metadata captured from an echoserver
 // response.
 type CapturedRequest struct {
 	Path     string              `json:"path"`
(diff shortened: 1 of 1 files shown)


conformance/utils/roundtripper/roundtripper.go
```diff
@@ -116,7 +116,7 @@ func (d *DefaultRoundTripper) CaptureRoundTrip(request Request) (*CapturedReques
 
 	if d.Debug {
 		dump, dumpErr := httputil.DumpResponse(resp, true)
```
bowei, line 118: Why not reuse err so we don't have to keep track of stuff? Does the value in err live past this block? (would be a maintenance issue)

robscott: It takes a bit more code to avoid error shadowing (have to declare dump separately). I don't feel strongly either way though, so adjusted the code to use separate declarations + reuse err var.

youngnick:
/lgtm

I didn't notice it before, and it should be in a separate PR, but if for some reason the response is not JSON, then the caller will get an empty object with nil error? That doesn't seem ideal.

Issue #1017, opened 2022-02-10
Admission webhook: Provide install YAML as part of release bundle YAML


youngnick:
As part of installing a gateway-api release bundle, we need to provide a YAML or set of YAMLs that contains both the CRDs (whether they be stable or experimental), and the install for the admission-server webhook.

youngnick:
No, the plan here is to have something a bit more complete - a single file (and command) to install both the CRDs and webhook, since I believe it should be mandatory. For those who want to customize the install a bit, I'd like to provide a directory of versioned individual YAMLs as well as the single file.

youngnick:
Okay, here's my thoughts on the install process:

I think that we want to have a single YAML file that requires no further rendering, that includes both the CRDs and the webhook install.

We add the following files:
`install/stable.yaml` - contains the stable CRD channel and the current webhook version.
`install/experimental.yaml` - contains the experimental CRD channel and the current webhook version.

These will just be the source files concatenated together, with comments on each included file saying where that file was sourced from.

I think that we should keep the source files in their current locations, so as not to break people who are using them from there already.

We should also add the new single-file sources to `make generate` and `make verify`, so that new PRs that update anything ensure that everything is up-to-date correctly.

Release docs need an update with a checklist of things to do (like updating `pkg/generator/main.go` etc. ) - they're already out of date because we need to update once we fork the `release-x` branch.

PR #1047, opened 2022-03-17
Add master label to the most recent image build

youngnick:

**What type of PR is this?**
/kind cleanup

As I started working on #1017, I realized that we'll need to have a way to refer to "whatever image is the latest master" for our living-dangerously users, so I propose we also add a `master` label, which will go on anything that doesn't get a semver label.

Updates #1017

```release-note
NONE
```

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -45,8 +45,8 @@ then
 fi
 
 
-# We tag the image with :latest for the most recent PR merge.
-LATEST=true
+# We tag the image with :latest for the most recent PR merge to master.
+LATEST=false
 
 VERSION_TAG=$GIT_TAG
 
@@ -62,10 +62,12 @@ then
 
     # We want the binary version to show up correctly too.
     BINARY_VERSION="${BASE_REF}"
-
-    # If we're on a semver baseref, then we don't want to tag the image with :latest
-    LATEST=false
+else
+    # If our base ref == "master" then we will tag :latest.
+    LATEST=true
 fi
+
+
 
 # First, build the image, with the version info passed in.
 # Note that an image will *always* be built tagged with the GIT_TAG, so we know when it was built.
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -45,7 +45,8 @@ then
 fi
 
 
-LATEST=false
+# We tag the image with :latest for the most recent PR merge.
+LATEST=true
 
 VERSION_TAG=$GIT_TAG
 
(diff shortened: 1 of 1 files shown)


youngnick:
This only needs to be merged to master, not backported to release-0.4.

hack/build-and-push.sh
```diff
@@ -77,10 +77,16 @@ docker build --build-arg COMMIT=${BINARY_VERSION} --build-arg TAG=${VERSION_TAG}
...
 # Then, we add extra tags if required.
+# If the version tag and the git tag aren't the same, we're on a release branch, so
+# we need to push the release tag.
 if [[ $VERSION_TAG != $GIT_TAG ]]
 then
     docker tag ${REGISTRY}/admission-server:${GIT_TAG} ${REGISTRY}/admission-server:${VERSION_TAG}
     docker push ${REGISTRY}/admission-server:${VERSION_TAG}
+else
+# Otherwise, we're on master and we should update the master image here too.
+    docker tag ${REGISTRY}/admission-server:${GIT_TAG} ${REGISTRY}/admission-server:master
```
robscott, line 88: Since we're planning to move to "main", should we get ahead of it here? Alternatively should we just use "latest"?

youngnick: Yeah, making this PR is what reminded me about moving to `main`. You're right that `latest` is probably better used for a nightly-style thing anyway. I'll update to that effect.

hack/build-and-push.sh
```diff
@@ -58,15 +59,12 @@ if [[ "${BASE_REF}" != "master" ]]
...
-    # Use some bash magic to check if the semver does not end with -sometext, that
-    # would indicate a prerelease version. If this is not a prerelease, then we want to set
-    # the `latest` tag too.
-    if [[ ! "${BASE_REF}" =~ -(.+)$ ]];
-    then
-    LATEST=true
-    fi
+
+    # If we're on a semver baseref, then we don't want to tag the image with :latest
+    LATEST=false
```
robscott, line 67: Does this mean we won't tag latest as part of semver release like v0.5.0?

youngnick: Yes, this means that `:latest` is the latest dev build rather than the latest stable. Maybe this is too far outside the usual contract? I think that using `:latest` for things you want to be stable is an antipattern anyway.

robscott: I usually think of `:latest` representing the latest build on the master/main branch. If I'm interpreting this change correctly, we'd actually end up pushing a new image anytime a commit merged to a different branch like release-0.4. I agree that relying on a `:latest` image is a bad idea in general, but I'd generally expect updates to that tag to be relatively sequential in nature, so a latest referencing a master commit getting replaced by one that references a release-0.4 commit would be surprising.

I know it's also fairly common for `:latest` to represent the latest release, not the latest commit, but since these images are already limited to the staging repo, it's probably fine to just track the latest commit.

youngnick: Ah, you're right, I have a logic error, I need to check if we're on master, and set LATEST to `true` in that case only, not assume it's true and set it to false if semver. I'll update.

youngnick:
Thanks, closing in favor of #1068.

PR #1050, opened 2022-03-17
Updating Security Model and Reference Policy documentation

### LINKED ISSUE #1033: docs: improve explanation of secure usage of AllowedRoutes selectors
Add additional content (perhaps in the [Security Model](https://gateway-api.sigs.k8s.io/concepts/security-model/) section?) to better describe how [`AllowedRoutes`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.AllowedRoutes) should be configured for secure operation, likely linking out to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors for additional context on how to use selectors.

**Why this is needed**:
It's quite easy to configure a permissive policy with `FromNamespaces: All`, but there's not much information on how `FromNamespaces: Selector` can be used to narrow the allowed scope from where Routes may bind, or potential risks of specifying a selector like `Label` without restrictive RBAC instead of explicitly selecting a namespace by `metadata.Name`.

As an example provided by @robscott 
> I think many users may end up with something like this:
```
namespaces:
  from: Selector
  selector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - foo
      - bar
```

robscott:
**What type of PR is this?**
/kind documentation

As I set out to fix #1033, it seemed like everything related to that documentation needed to be refreshed. I've updated our Security Model documentation and then renamed our related "Cross Namespace References" doc to a "ReferencePolicy" doc since that's all it was really covering.

Fixes #1033

```release-note
NONE
```


youngnick (commented):
Nice updates!

/lgtm

PR #1051, opened 2022-03-17
Adding webhook to e2e tests with kind, moving deploy to gateway-system namespace

### LINKED ISSUE #1040: Add e2e testing for validating webhook
This does not need to be complex, but should ensure that the webhook can be deployed successfully, valid resources can be created, and invalid ones are prevented. This could likely be added to https://github.com/kubernetes-sigs/gateway-api/blob/master/hack/verify-examples-kind.sh.

robscott:
**What type of PR is this?**
/kind cleanup

This PR includes 3 related changes:
1. Moves webhook deployment from gateway-api to gateway-system
2. Adds webhook deployment as part of kind e2e verification tests
3. Adds invalid examples that are only covered by webhook validation

Fixes #1040

```release-note
The Gateway API webhook is now deployed in a `gateway-system` namespace instead of `gateway-api`.
```

### DIFF
--- a/hack/verify-examples-kind.sh
+++ b/hack/verify-examples-kind.sh
@@ -22,8 +22,8 @@ readonly GO111MODULE="on"
 readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
 readonly CLUSTER_NAME="verify-gateway-api"
-readonly KUBECONFIG="${GOPATH}/.kubeconfig"
 
+export KUBECONFIG="${GOPATH}/.kubeconfig"
 export GOFLAGS GO111MODULE GOPATH
 export PATH="${PATH}:${GOPATH}/bin"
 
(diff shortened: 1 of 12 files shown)


hack/verify-examples-kind.sh
```diff
@@ -49,34 +49,53 @@ res=0
...
-      kubectl apply --kubeconfig "${KUBECONFIG}" --recursive -f examples/experimental/v1alpha2 || res=$?
+      kubectl apply --recursive -f examples/experimental/v1alpha2 || res=$?
   fi
 
   # Install invalid gateway-api resources.
   # None of these examples should be successfully configured
   # This is very hacky, sorry.
   # Firstly, apply the examples, remembering that errors are on stdout
-  kubectl apply --kubeconfig "${KUBECONFIG}" --recursive -f hack/invalid-examples 2>&1 | \
+  kubectl apply --recursive -f hack/invalid-examples 2>&1 | \
```
youngnick, line 93: Every time I look at this I shudder a bit. I'm worried it's a bit fragile, but I can't think of a way to improve it without writing our own kubectl apply thing. :)

youngnick (commented):
This looks great to me, although I think the day is coming that we'll need to move the mechanisms to actually apply-and-check the things to Go code.

PR #1055, opened 2022-03-21
Adding hostnames to Route Status

### LINKED ISSUE #848: Add Hostnames to Status for HTTP and TLS Routes
A new `hostnames` field in `RouteParentStatus` that could be used to indicate which hostnames a Route was attached with.

**Why this is needed**:
This could be useful if a Route was attached to different Gateways for different hostnames. For example, it could be attached to Gateway `foo` for `foo.com` and Gateway `bar` for `bar.com`. This would happen if the Route specified both hostnames but each Gateway only specified one of the hostnames. 

For reference: This idea was came from a [comment thread](https://github.com/kubernetes-sigs/gateway-api/pull/839#discussion_r700722952) with @youngnick

robscott:
**What type of PR is this?**
/kind feature

In the last round of sig-network review, this specific issue/idea was raised. It reappeared over the weekend and I thought I'd try to implement it in time for this round of API review. It ended up being a larger change than I'd initially expected, so I'm not tied to getting this into v0.5.0. 

Fixes #848 

```release-note
Experimental: RouteParentStatus for HTTP and TLS Routes now includes hostnames.
```


robscott:
Adding a hold because this is an API change and should not merge without consensus.

/hold

apis/v1alpha2/shared_types.go
```diff
@@ -263,6 +263,58 @@ type RouteStatus struct {
...
+	// manages the parent must add an entry to this list when the controller
+	// first sees the route and should update the entry as appropriate when the
+	// route or gateway is modified.
+	//
+	// Note that parent references that cannot be resolved by an implementation
+	// of this API will not be added to this list. Implementations of this API
+	// can only populate Route status for the Gateways/parent resources they are
+	// responsible for.
+	//
+	// A maximum of 32 Gateways will be represented in this list. An empty list
```
youngnick, line 311:
```suggestion
	// A maximum of 32 parent resources will be represented in this list. An empty list
```

apis/v1alpha2/shared_types.go
```diff
@@ -263,6 +263,58 @@ type RouteStatus struct {
 	Parents []RouteParentStatus `json:"parents"`
 }
 
+// RouteParentStatusWithHostnames describes the status of a route with respect
+// to an associated Parent. In the experimental channel, this includes hostnames
+// used as part of the Parent binding.
+type RouteParentStatusWithHostnames struct {
```
youngnick, line 269: The `WithHostnames` suffix feels a little verbose, but I can't think of anything better that's not worse. 😞

youngnick (commented):
This makes sense to me, couple of small nits.

/lgtm

apis/v1alpha2/httproute_types.go
```diff
@@ -853,5 +853,5 @@ type HTTPBackendRef struct {
 
 // HTTPRouteStatus defines the observed state of HTTPRoute.
 type HTTPRouteStatus struct {
-	RouteStatus `json:",inline"`
+	RouteStatusWithHostnames `json:",inline"`
```
bowei, line 856: Why not keep the name of the struct the same? will we also have RouteStatusWithoutHostnames?

robscott: Good question - this is because TCPRoute and UDPRoute don't support hostnames so we're leaving that out of status. So RouteStatusWithHostnames is really just that, it is a type that inlines RouteStatus + adds a hostnames field.

apis/v1alpha2/shared_types.go
robscott: Haha same.

apis/v1alpha2/httproute_types.go
bowei: Does it make sense to pull up the Hostnames field instead of creating a new type? It is all inline in any event.

robscott: So just add the hostnames field to both HTTPRouteStatus and TLSRouteStatus? Although that might be simpler now, I think we'd regret it as more routes like GRPCRoute are added.

apis/v1alpha2/shared_types.go

apis/v1alpha2/shared_types.go
```diff
@@ -256,13 +256,65 @@ type RouteStatus struct {
 	// can only populate Route status for the Gateways/parent resources they are
 	// responsible for.
 	//
-	// A maximum of 32 Gateways will be represented in this list. An empty list
-	// means the route has not been attached to any Gateway.
+	// A maximum of 32 parent resources will be represented in this list. An
```
bowei, line 259: We should leave the max number to the kubebuilder validation limit if possible to avoid potential drift between comments and actual value.

apis/v1alpha2/httproute_types.go
bowei: Not clear why writing it out is better/worse to be honest -- this makes the usage of the API for the go consumer somewhat cumbersome (there are multiple levels of inline being collapsed into the YAML representation).  

I don't think there is a requirement to keep the status struct the same Go types between the route types?

robscott: I dug into this further, and it's actually substantially more complicated to split this out than I'd thought.

Essentially RouteStatus includes a list of RouteParentStatus structs, and we want to add a Hostnames field to the RouteParentStatus structs. The RouteParentStatus struct is sufficiently complex that I don't think we should duplicate it. So we could potentially replace RouteStatusWithHostnames with something like this for each route type that includes hostnames (currently HTTP and TLS, will likely include GRPC soon):

```go
HTTPRouteStatus struct {
	// Parents is a list of parent resources (usually Gateways) that are
	// associated with the route, and the status of the route with respect to
	// each parent. When this route attaches to a parent, the controller that
	// manages the parent must add an entry to this list when the controller
	// first sees the route and should update the entry as appropriate when the
	// route or gateway is modified.
	//
	// Note that parent references that cannot be resolved by an implementation
	// of this API will not be added to this list. Implementations of this API
	// can only populate Route status for the Gateways/parent resources they are
	// responsible for.
	//
	// A maximum of 32 parent resources will be represented in this list. An
	// empty list means the route has not been attached to any parent resource.
	//
	Parents []RouteParentStatusWithHostnames `json:"parents"`
}
```

That would make for a larger diff and would increase the chance that one of these got out of sync. I'm not sure it's worth the change, but I may be missing a better way to do it.

markmc (commented):
Might be good to mention in the "If both the Listener and TLSRoute have specified hostnames" `HTTPRoute.spec.hostnames` docs how the result would be reflected in status. Same for TLS

apis/v1alpha2/httproute_types.go
markmc: In [route delegation status draft](https://hackmd.io/@youngnick/gateway-api-statusl) I asked:

> If we go with RouteParentStatusWithHostnames, would we add this per-rule attachment status list there so that it only applies to HTTP and TLS?
>
> Or if we went with HTTPRouteParentStatus, would we add it there so it applies only to HTTP?
>
> Or would we add it to both RouteParentStatus and RouteParentStatusWithHostnames/HTTPRouteParentStatus?

mikemorris:
Is there a specific reason why `hostnames` in status should be considered differently than `port` regarding https://github.com/kubernetes-sigs/gateway-api/issues/1060#issuecomment-1079412191?

PR #1065, opened 2022-03-23
Clarifying TLS Reencryption Docs

### LINKED ISSUE #968: Docs mentions Reencrypt for HTTPRoute and TLSRoute is available
**What happened**:

[Route summary table](https://gateway-api.sigs.k8s.io/concepts/api-overview/?h=reencrypt#route-summary-table) explains as HTTPRoute and TLSRoute as:

HTTPRoute
> Terminated only, can be reencrypted

TLSRoute
> Passthrough or terminated, can be reencrypted if terminated.	

https://github.com/kubernetes-sigs/gateway-api/blob/1deb7bc86193f670c149acf35426dd1ceaf588a5/site-src/concepts/api-overview.md?plain=1#L115-L116

image

But `reencrypt` is not available yet from my understanding. The docs should not mention about reencrypted? Or I am wrong and reencrypted is actually possible?

`reencrypted` is not available yet from my understanding. The docs should not mention about reencrypted. (Again, I'm sorry if I misunderstand it.)

If Reencrypt is possible, how to configure it :sweat_smile: 

TLSRoute with terminated is also still under discussion as https://github.com/kubernetes-sigs/gateway-api/pull/730

robscott:
**What type of PR is this?**
/kind documentation

This fixes a confusing part of our documentation that describes reencryption as being possible without clarifying that it is not currently configurable with the API. If/when this merges, I'll create a follow up issue to track the feature request/GEP for configuring reencryption.

Fixes #968

```release-note
NONE
```

### DIFF
--- a/site-src/concepts/api-overview.md
+++ b/site-src/concepts/api-overview.md
@@ -112,11 +112,16 @@ to allow multiple Routes to share ports on the Listener.
 
 |Object|OSI Layer|Routing Discriminator|TLS Support|Purpose|
 |------|---------|---------------------|-----------|-------|
-|HTTPRoute| Layer 7 | Anything in the HTTP Protocol | Terminated only, can be reencrypted| HTTP and HTTPS Routing|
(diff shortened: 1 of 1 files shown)


youngnick (commented):
/lgtm

I like this wording.

Issue #1006, opened 2022-02-01
Admission Webhook: Document that installation is required


youngnick:
Running the webhook is an expected part of installing the APIs, but currently our documentation does not make this clear.

It should say that installing the version of the webhook that comes with a release bundle is required to use the APIs themselves, and conformance tests will test this. (#1005 covers the larger work to actually make that happen).

mehabhalodiya:
@youngnick May I know, in which page you would like to document this?

youngnick:
Definitely the Getting Started guide (https://gateway-api.sigs.k8s.io/v1alpha2/guides/getting-started/), but we'll also need to update the instructions for implementors. I think this one is probably blocked by me finishing up the rest of the changes in #1005, since we'll need to update the install instructions with whatever I figure out there.

mehabhalodiya:
So you are working on this, nice!

PR #1066, opened 2022-03-23
Updating docs related to versioning and webhook

### LINKED ISSUE #1037: Clarify how Gateway API versioning and release channels work
We should add/rework docs to clarify how Gateway API versioning works. This can include content from the following working docs:

- [Gateway API Stability](https://docs.google.com/document/d/1jz1DGci9YJkFx3R5AIBhRziPPJpGEFtEDTgOPy0gfcE/edit)
- [Gateway API Beta Release Plan](https://docs.google.com/document/d/1F3dcs_J8Y6GVlfMGjRtom5ljovnKa_MMIC34GJxUdBs/edit)

robscott:
**What type of PR is this?**
/kind documentation

I started by trying to update versioning guidelines, but that turned into a bit of yak shaving where a lot of the content that was in the existing doc felt like it really belonged in other docs. That resulted in me moving some of the webhook related discussion to the getting started page, and potentially fixing #1006 as part of this.

Fixes #1037 and #1006

```release-note
NONE
```


site-src/v1alpha2/guides/getting-started.md
```diff
@@ -21,33 +21,43 @@ Gateway API. By installing a Gateway controller in your Kubernetes cluster,
...
-| kubectl delete -f -
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.4.2/deploy/admission_webhook.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.4.2/deploy/certificate_config.yaml
 ```
 
+### Cleanup
+After you're done, you can clean up after yourself by uninstalling the Gateway
+API CRDs and webhook by replacing "apply" with "delete" in the commands above.
+If these resources are in-use or if they were installed by a Gateway controller,
+then do not uninstall them. This will uninstall the Gateway API respirces for
```
markmc, line 61:
```suggestion
then do not uninstall them. This will uninstall the Gateway API resources for
```

site-src/concepts/guidelines.md
```diff
@@ -38,6 +38,28 @@ prescribed resolution, the following guiding principles should be applied:
...
+implemented by a variety of different products within the same cluster. That
+means that the API version your implementation was developed with may be
+different than the API version it is used with. At a minimum, the following
+requirements must be met to ensure future versions of the API do not break your
+implementation:
+
+* Handle fields with loosened validation without crashing
+* Handle fields that have transitioned from required to optional without
+  crashing
+* DO NOT rely on webhook or CRD validation as a security mechanism. If field
```
markmc, line 54: I'm probably missing some important nuance, but [the discussion on this (around 48m)](https://youtu.be/EUUtwv6mFGg?t=2874) in this week's meeting sounded to me like the webhook is an important security mechanism, and that's why its presence (or equivalent) should be checked by conformance tests.

Maybe these two statements aren't contradictory, but I would have expected to see documentation added about how it's required for implementors to deploy the webhook to be conformant and an explanation of why it's a potential security issue if it's not deployed.

site-src/concepts/guidelines.md
```diff
@@ -118,6 +140,16 @@ The "implementation-specific" designation allows a CORE or EXTENDED feature to
 be well-defined taking into account the realities of some features that are
 mostly but not entirely portable.
 
+## Limitations of Webhook and CRD Validation
+CRD and webhook validation is not the final validation i.e. webhook is “nice UX”
+but not schema enforcement. This validation is intended to provide immediate
```
markmc, line 145: Same comment here

site-src/concepts/guidelines.md
```diff
@@ -118,6 +140,16 @@ The "implementation-specific" designation allows a CORE or EXTENDED feature to
...
+feedback to users when they provide an invalid configuration, but can not
+completely be relied on because it:
```
markmc, line 147: "Relied on to (what?) because it:"

It's not obvious to me from this what it shouldn't be relied on for?

Maybe the "DO NOT rely on webhook or CRD validation" text from above would fit better in this section?

site-src/concepts/guidelines.md
markmc: Ah, I see you're just moving existing docs around. Maybe best addressed separately then?

markmc:
By the way, how does the `concepts/versioning/` relate to `references/releases/` - they seem overlapping? Also `RELEASE.md` only references the latter?

site-src/concepts/guidelines.md
robscott: @markmc this is an important distinction, and we should be more clear about exactly what a webhook provides. I think it's worth strongly installing the installation of a webhook and even requiring it as part of a conformant deployment, but we need to recognize that a webhook can fail or be bypassed. 

/cc @youngnick for more thoughts on how we should approach this.

site-src/concepts/guidelines.md
robscott: Good points. Updated this section, hopefully it's a bit more clear now.

site-src/v1alpha2/guides/getting-started.md
robscott: Updated, thanks!

site-src/v1alpha2/guides/getting-started.md
```diff
@@ -1,7 +1,7 @@
 # Getting started with Gateway APIs
 
 **1.**  **[Install a Gateway controller](#installing-a-gateway-controller)**
- _OR_  **[install the Gateway API CRDs manually](#installing-gateway-api-crds-manually)**
+ _OR_  **[install the Gateway API CRDs manually](#installing-gateway-api-bundle)**
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

robscott:
> By the way, how does the concepts/versioning/ relate to references/releases/ - they seem overlapping? Also RELEASE.md only references the latter?

Good catch @markmc! I'll try to update those files as part of this PR since they both seem very relevant. Adding a hold until I can get to them.

/hold

site-src/concepts/guidelines.md
youngnick: I think that we're aiming for a defence-in-depth posture here. Kubebuilder validations and webhook validations are part of it, but should not be the whole story.

However, I do think it's reasonable to expect that implementations provide equivalent functionality to the webhook, and ensure that it's part of conformance testing. Of course it's possible to bypass it, and it's not the whole story, but enabling it does increase security for API users.

Using the words "you can't rely on this" kind of implies that there's no point to having them. I think that something like "implentors should ensure that, even if the protections are bypassed, their implementations are still as secure as possible" or something, particularly for things like filtering fields.

The validation protects against mistakes, the webhook against attackers without a way to bypass them, but nothing protects against a dedicated, motivated attacker with high cluster access.

site-src/concepts/guidelines.md
```diff
@@ -118,6 +140,18 @@ The "implementation-specific" designation allows a CORE or EXTENDED feature to
...
 
+## Limitations of Webhook and CRD Validation
+CRD and webhook validation is not the final validation i.e. webhook is "nice UX"
+but not schema enforcement. This validation is intended to provide immediate
+feedback to users when they provide an invalid configuration. Write code
+defensively with the assumption that at least some invalid input (Gateway API
+resources) will reach your controller. Both Webhook and CRD validation is not
+fully reliable because it:
+
+* Is not guaranteed to be present or up to date in all clusters.
```
youngnick, line 151: I think that if we mandate that a Gateway API install _must_ include either the webhook or something that implements the same function, this can be removed.

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
-Each API version provides a unique way to interact with the API. Significant
-changes such as removing or renaming fields will be represented here.
-
-### 2. Bundle Versions (ex: v0.4.0)
-Bundle versions are broader in scope. They use semantic versioning and include
-the following:
-
-* API Types/CRDs
+* API Types
+* CRDs
```
youngnick, line 8:
```suggestion
* API Types (Go bindings for the resources)
* CRDs (Kubernetes definitions of the resources)
```

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
 * Validating Webhook
```
youngnick, line 9:
```suggestion
* Validating Webhook install
```

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
+* Graduation of fields or resources from experimental to stable track
+* Introduction of a new **API version**, which may also include:
+  * Renamed fields
+  * Anything else that is valid in a new Kubernetes API version
+  * Removal/tombstoning of beta fields
+* Removal of an API resource following [Kubernetes deprecation
+  policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)
+
+### Major version (v0.x to v1.0)
+* Everything that is valid in a patch release
```
youngnick, line 69:
```suggestion
* Everything that is valid in a minor release
```

youngnick (commented):
This is looking pretty good, but I think we need to be really careful with the wording around why implementations need to be defensive on top of the validations and webhook.

site-src/concepts/guidelines.md
bowei, line 54: What is "escaping" referring to in the webhook? We don't transform any field values?

site-src/concepts/guidelines.md
```diff
@@ -38,6 +38,28 @@ prescribed resolution, the following guiding principles should be applied:
...
+
+An important consideration when implementing this API is how it might change in
+the future. Similar to the Ingress API before it, this API is designed to be
+implemented by a variety of different products within the same cluster. That
+means that the API version your implementation was developed with may be
+different than the API version it is used with. At a minimum, the following
+requirements must be met to ensure future versions of the API do not break your
+implementation:
+
+* Handle fields with loosened validation without crashing
```
bowei, line 51: We should be more specific here:

"Fields values may contain values with less restrictive validation in newer versions of the API. In this case, the controller is expected to handle this gracefully. The most common response would be to reject the configuration as malformed and signal the user via a condition in the `status` block.

site-src/concepts/guidelines.md
```diff
@@ -118,6 +140,18 @@ The "implementation-specific" designation allows a CORE or EXTENDED feature to
 be well-defined taking into account the realities of some features that are
 mostly but not entirely portable.
 
+## Limitations of Webhook and CRD Validation
+CRD and webhook validation is not the final validation i.e. webhook is "nice UX"
```
bowei, line 144: We should clarify why we are making this statement, otherwise the reader will be somewhat left with a directive without the associated reasoning.

- Are we saying that webhooks are inherently not reliable because people can uninstall them?
- Or is it because of API revisions,

Or both...

site-src/concepts/guidelines.md
```diff
@@ -118,6 +140,18 @@ The "implementation-specific" designation allows a CORE or EXTENDED feature to
...
+but not schema enforcement. This validation is intended to provide immediate
+feedback to users when they provide an invalid configuration. Write code
+defensively with the assumption that at least some invalid input (Gateway API
+resources) will reach your controller. Both Webhook and CRD validation is not
+fully reliable because it:
+
+* Is not guaranteed to be present or up to date in all clusters.
+* Will likely never be sufficient to cover all edge cases.
```
bowei, line 152: Probably delete this one? It just sounds like "webhook might not be implemented all the way?" which is somewhat confusing statement.

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
-## Summary
-Each Gateway API release is represented by a bundle version that represents
-that specific combination of CRDs, API versions, and validating webhook. To
-enable experimental fields, future releases of the API will include stable and
-experimental CRD tracks. Users will be able to access experimental features by
-installing the experimental CRDs. A cluster can contain either an experimental
-or stable CRD for any resource at a given time. 
+## Overview
+Each new release of Gateway API is defined with a "bundle version" such as
+v0.4.0. This contains the following:
```
bowei, line 5: if bundle version is the git tag, we should say so.

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
+Release channels are used to indicate the stability of fields within a resource.
+For example, even if a resource has already graduated to beta, we need a way to
+test out new fields and features before committing to their long term stability.
+With upstream Kubernetes APIs, this is accomplished with [feature
+gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).
+Unfortunately, CRDs do not have a similar concept yet.
+
+Instead of trying to recreate feature gates in this project, we've introduced
+release channels. Starting in v0.5.0, all new fields and features will start in
+the experimental set of CRDs. From that point, these may graduate to the stable
```
bowei, line 29: experimental channel (keep wording consistent)

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
 ```
 gateway.networking.k8s.io/bundle-version: v0.4.0
 gateway.networking.k8s.io/channel: stable|experimental
 ```
+
+## What can Change
+When using or implementing this API, it is important to understand what can
+change across API versions.
+
+### Patch version (v0.4.0 -> v0.4.1)
```
bowei, line 48: ### Patch version Major.Minor.Patch +1 (e.g. v0.4.0 to v0.4.1)

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
+* Clarifications to godocs
+* Updates to CRDs and/or code to fix a bug
+* Fixes to typos
+
+### Minor version (v0.4.0 -> v0.5.0)
```
bowei, line 53: see above

site-src/concepts/versioning.md
bowei, line 69: Better to say: There are no API compatibility guarantees when the major version changes.

site-src/v1alpha2/contributing/devguide.md
```diff
@@ -59,6 +59,17 @@ You can kick off an overall build from the top-level makefile:
...
 
+## Adding Experimental Fields
+
+Starting with v0.5.0, all additions to the API must start in the experimental
+release channel. Experimental fields must be marked with the
+`<gateway:experimental>` annotation in Go type definitions. Gateway API CRD
+generation will exclude these fields from stable CRDs.
+
+If experimental fields are removed or renamed, the original field name should be
+removed from the go struct, with a tombstone comment ensuring the field name
```
bowei, line 70: give example of tombstone comment format

site-src/v1alpha2/guides/getting-started.md
```diff
@@ -21,33 +21,43 @@ Gateway API. By installing a Gateway controller in your Kubernetes cluster,
...
-After you're done, you can clean up after yourself by uninstalling the
-Gateway API CRDs. The following command will remove all GatewayClass, Gateway,
-and associated resources in your cluster. If these resources are in-use or
-if they were installed by a Gateway controller, then do not uninstall them.
-This will uninstall the Gateway API CRDs for the entire cluster. Do not do
-this if they might be in-use by someone else as this will break anything using
-these resources.
+### Install the Webhook
 
+The validating webhook included with Gateway API is still in active development
```
bowei, line 46: Shouldn't this be more for release notes? It's going to be hard to find this paragraph and remove it later.

site-src/v1alpha2/guides/getting-started.md
```diff
@@ -21,33 +21,43 @@ Gateway API. By installing a Gateway controller in your Kubernetes cluster,
...
 ```
-kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.4.0" \
-| kubectl delete -f -
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.4.2/deploy/admission_webhook.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/gateway-api/v0.4.2/deploy/certificate_config.yaml
 ```
 
+### Cleanup
+After you're done, you can clean up after yourself by uninstalling the Gateway
+API CRDs and webhook by replacing "apply" with "delete" in the commands above.
```
bowei, line 59: we should just give the commands directly

```
kubectl delete -f ...

```

site-src/concepts/versioning.md
robscott: I'm not sure I understand this change. I think the bundle version includes a webhook and CRDs, which you can separately install in a cluster.

site-src/v1alpha2/guides/getting-started.md
robscott: We already have an issue to track improving the installation process (https://github.com/kubernetes-sigs/gateway-api/issues/1017) that @youngnick is working on. I'll update that issue to ensure it includes updating this text. I think our next release will include those fixes and can also be used to update this text. The latest released version of the webhook (v0.4.2) does have noticeable bugs and omissions, so I think it's important to leave this in until we've resolved them as part of the v1b1 milestone.

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Not sure I understand this suggestion, I think this could result in more confusing headings?

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/guidelines.md
robscott: Updated, thanks!

site-src/concepts/guidelines.md
robscott: Reworked a lot of this section, hopefully it's clearer now.

site-src/concepts/guidelines.md
robscott: Updated, thanks!

site-src/concepts/guidelines.md
robscott: I've removed this comment, but for example our CRD validation uses regexes to ensure that only "safe" values will exist for some fields. That does not mean that implementations should blindly trust this input and add it to a DB query for example.

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/guidelines.md
robscott: Good call, replaced with "May not be deployed correctly"

site-src/concepts/guidelines.md
robscott: Updated, thanks!

site-src/concepts/guidelines.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
youngnick: Yeah, agreed, this is more confusing. Resolving.

site-src/concepts/versioning.md
youngnick: Maybe this?

```suggestion
### Patch version (e.g. v0.4.0 -> v0.4.1)
```

site-src/concepts/versioning.md
youngnick:
```suggestion
### Minor version (e.g. v0.4.0 -> v0.5.0)
```

youngnick (commented):
Couple of small changes that I think will help with @bowei's feedback. LGTM aside from that though.

site-src/v1alpha2/contributing/devguide.md
robscott: Added, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

hbagdi:
LGTM, @bowei has unresolved comments above.

site-src/concepts/guidelines.md
```diff
@@ -38,6 +38,41 @@ prescribed resolution, the following guiding principles should be applied:
...
+  crashing
+
+## Limitations of CRD and Webhook Validation
+
+CRD and webhook validation is not the final validation i.e. webhook is "nice UX"
+but not schema enforcement. This validation is intended to provide immediate
+feedback to users when they provide an invalid configuration. Write code
+defensively with the assumption that at least some invalid input (Gateway API
+resources) will reach your controller. Both Webhook and CRD validation is not
+fully reliable because it:
```
bowei, line 62: (note: this is a generic CRD/webhook problem, not just for this specific API)

bowei: (would be good to make sure this is clear)

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
-will exclude these fields from stable CRDs. Experimental fields may be removed
-from the API. Due to the experimental nature of these CRDs, they are not
-recommended for production use.
-
-If experimental fields are removed or renamed, the original field name should be
-removed from the go struct, with a tombstone comment ensuring the field name
-will not be reused. 
-
+### API Versions (v1alpha2, v1beta1)
+Within Gateway API, API versions are primarily used to indicate the stability of
```
bowei, line 13: API versions (e.g. "xxxxxx")

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
+gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).
+Unfortunately, CRDs do not have a similar concept yet.
+
+Instead of trying to recreate feature gates in this project, we've introduced
+release channels. Starting in v0.5.0, all new fields and features will start in
+the experimental release channel. From that point, these may graduate to the
+stable release channel or be dropped from the API entirely.
+
+Due to their unstable nature, experimental CRDs should not be used for
+production workloads.
```
bowei, line 34: I think for clarity, we should have the definitions explicit in a list

```
* Stable: Stable channel features have long-term stability...
* Experimental: Experimental channel features ...
```

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
+
```
bowei, line 35: Rob -- I think  we need an example to illustrate the differences between version, channel?

What do you think? I think it will make it very clear for the reader what is intended for this.

site-src/concepts/guidelines.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,70 @@
...
-
+### API Versions (v1alpha2, v1beta1)
+Within Gateway API, API versions are primarily used to indicate the stability of
+a resource. For example, if a resource has not yet graduated to beta, it is
+still possible that it could either be removed from the API or changed in
+backwards incompatible ways. For more information on API versions, refer to the
+[full Kubernetes API versioning
+documentation](https://kubernetes.io/docs/reference/using-api/#api-versioning).
+
+### Release Channels (experimental, stable)
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: I think I went overboard here, but hopefully the diagrams here help: https://deploy-preview-1066--kubernetes-sigs-gateway-api.netlify.app/concepts/versioning/

youngnick:
This is heaps more clear to me, LGTM.

I'll leave it to @bowei to push the final button though.

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,86 @@
...
+example, even if a resource has already graduated to beta, we need a way to test
+out new fields and features before committing to their long term stability. With
+upstream Kubernetes APIs, this is accomplished with [feature
+gates](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).
+Unfortunately, CRDs do not have a similar concept yet.
+
+Instead of trying to recreate feature gates in this project, we've introduced
+release channels. Starting in v0.5.0, all new fields and features will start in
+the Experimental release channel. From that point, these may graduate to the
+Stable release channel or be dropped from the API entirely.
```
markmc, line 48:
```suggestion
Stable release channel or be dropped from the API entirely. See [GEP-922](/geps/gep-922) for a more detailed discussion of this approach to new fields and features.
```

site-src/v1alpha2/guides/getting-started.md
```diff
@@ -21,33 +21,43 @@ Gateway API. By installing a Gateway controller in your Kubernetes cluster,
...
 ```
 
-After you're done, you can clean up after yourself by uninstalling the
-Gateway API CRDs. The following command will remove all GatewayClass, Gateway,
-and associated resources in your cluster. If these resources are in-use or
-if they were installed by a Gateway controller, then do not uninstall them.
-This will uninstall the Gateway API CRDs for the entire cluster. Do not do
-this if they might be in-use by someone else as this will break anything using
-these resources.
+### Install the Webhook
```
markmc, line 44: Is it just me, or is there no obvious page to link to which would explain what "the webhook" is?

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,86 @@
...
-to stable APIs later. Experimental fields will be marked with the
-`+experimental` annotation in Go type definitions. Gateway API CRD generation
-will exclude these fields from stable CRDs. Experimental fields may be removed
-from the API. Due to the experimental nature of these CRDs, they are not
-recommended for production use.
-
-If experimental fields are removed or renamed, the original field name should be
-removed from the go struct, with a tombstone comment ensuring the field name
-will not be reused. 
+### API Versions (e.g. v1alpha2, v1beta1)
```
markmc, line 12: One thing I found odd from trying to put together some Gateway API bits for the sig-network annual report (kubernetes/community#6614) is that we have nice release notes for ... uh ... releases, but not for API versions, which is probably what folks should be more interested in?

Just a side observation, not something to address in this PR. Maybe something to think about for when v1beta1 is cut?

mehabhalodiya:
We need to add some information regarding:
- How do we update the code for new versions? and 
- How do we update the documentation when a new version is released?

I am an LFX mentee and I have done a docs assessment of this project. Already project mentors (@youngnick and Nate) are agreed on this. You can see the comment in [Google Docs](https://docs.google.com/document/d/1qeSwZ4WVevybTQoL5hG20Dcsh2qMzyPKv-z7qM3ho1c/edit?disco=AAAAWqiXVKA).

So I would suggest that we need to write up the process for versioning the documentation clearly. 
Reference: [cluster-api docs](https://cluster-api.sigs.k8s.io/contributing#versioning)

site-src/concepts/versioning.md
```diff
@@ -1,116 +1,86 @@
...
+The Stable release channel includes:
+
+* Resources that have graduated to beta
+* The fields in those resources that have graduated to the stable channel
+
+The Experimental release channel includes everything in the Stable release
+channel, plus:
 
+* Alpha API resources
+* Experimental fields that have not graduated to the stable channel
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

site-src/concepts/versioning.md
robscott: Updated, thanks!

site-src/v1alpha2/guides/getting-started.md
robscott: Good point, created #1133 to track that.

site-src/concepts/versioning.md
robscott: I think that's actually by design. An API version like v1beta1 really just indicates the stability of a resource, it's fairly rare for major changes to occur at the same time an API version is graduating to a new stability level. Of course when we transitioned from v1alpha1 to v1alpha2 there were pretty significant release notes included with v0.4.0, but I think most API version bumps from here on out are going to be bumping stability, not making significant changes. Instead, each new release of Gateway API will likely include new features, and I think that's what the release notes should focus on. Does that seem reasonable?

robscott:
@mehabhalodiya I agree with you. Unfortunately we don't have a way to version our docs yet. Doing that seems to require a transition off of mkdocs and to some other system. I'm very supportive of that kind of transition, but don't personally have time to work on it. Until we have proper docs versioning in place, I don't think there's much we can document about releasing version specific docs.

PR #1069, opened 2022-03-24
conformance: embed manifests in Go binary

howardjohn:
This allows importing the suite as a go package and running in
controller repos. This was the same stated purpose as
https://github.com/kubernetes-sigs/gateway-api/pull/1023; we would
prefer not to pull manifests from URL though. Its possible the URL
support is obsolete from this, or we can keep both depending on
feedback.

**What type of PR is this?**
/kind feature

```release-note
NONE
```

### DIFF
--- a/conformance/conformance_test.go
+++ b/conformance/conformance_test.go
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package conformance
+// conformance_test contains code to run the conformance tests. This is in its own package to avoid circular imports.
+package conformance_test
 
 import (
 	"flag"
--- a/conformance/embed.go
+++ b/conformance/embed.go
@@ -1,0 +1,22 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


conformance/conformance_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package conformance
+package conformance_test
```
bowei, line 17: can you put a comment why this package needs to be different name

howardjohn:
@shaneutt is there still interest in the URL support with this? I don't mind keeping it, wasn't sure if there were other reasons it was useful (always pulling from HEAD for example).

conformance/conformance_test.go
howardjohn: I will put a comment but the reason is to avoid a circular import

shaneutt:
> @shaneutt is there still interest in the URL support with this? I don't mind keeping it, wasn't sure if there were other reasons it was useful (always pulling from HEAD for example).

Always pulling from `main` is potentially useful as a smoke signal for upstream changes (or even breakage) which I think will be more helpful in the short term while things are changing rapidly, which is what I had in mind when I just made it possible to use the URL to the manifests. That said we're growing close enough to beta and times when I don't think things will be moving very much that I wouldn't ask to stop this change based on that.

youngnick:
I really like this change.

I'll leave someone else to hit this with the approved bat though.

/lgtm

howardjohn:
@bowei approval didn't go through for some reason

bowei:
hmmm -- is prow hooked up?

robscott:
Maybe it'll work this time?

/approve

PR #1073, opened 2022-03-26
feat: introduce v1beta1 types

### LINKED ISSUE #1041: Introduce v1beta1 types for resources graduating to beta
As one of the final steps to graduate GatewayClass, Gateway, and HTTPRoute to beta, we need to create v1beta1 type definitions for these resources. These MUST be identical to the corresponding v1alpha2 type definitions.

mehabhalodiya:
Signed-off-by: mehabhalodiya <mehabhalodiya@gmail.com>

**What type of PR is this?**
/kind feature

As suggested by @robscott, this is a follow-up PR on one of our last steps before the v1beta1 (v0.5.0) release.
The key is that we’d need ` GatewayClass`, `Gateway`, and `HTTPRoute` type definitions in `apis/v1beta1` in addition to the alpha directories they already exist in.

Fixes #1041 

```release-note
Introduce v1beta1 types for resources graduating to beta
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -1,0 +1,806 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=gateway-api,shortName=gtw
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Class",type=string,JSONPath=`.spec.gatewayClassName`
+// +kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.status.addresses[*].value`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// Gateway represents an instance of a service-traffic handling infrastructure
+// by binding Listeners to a set of IP addresses.
+type Gateway struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the desired state of Gateway.
+	Spec GatewaySpec `json:"spec"`
+
+	// Status defines the current state of Gateway.
+	//
+	// +kubebuilder:default={conditions: {{type: "Scheduled", status: "Unknown", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
+	Status GatewayStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// GatewayList contains a list of Gateways.
+type GatewayList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Gateway `json:"items"`
+}
+
+// GatewaySpec defines the desired state of Gateway.
+//
+// Not all possible combinations of options specified in the Spec are
+// valid. Some invalid configurations can be caught synchronously via a
+// webhook, but there are many cases that will require asynchronous
+// signaling via the GatewayStatus block.
+type GatewaySpec struct {
+	// GatewayClassName used for this Gateway. This is the name of a
+	// GatewayClass resource.
+	GatewayClassName ObjectName `json:"gatewayClassName"`
+
+	// Listeners associated with this Gateway. Listeners define
+	// logical endpoints that are bound on this Gateway's addresses.
+	// At least one Listener MUST be specified.
+	//
+	// Each listener in a Gateway must have a unique combination of Hostname,
+	// Port, and Protocol.
+	//
+	// An implementation MAY group Listeners by Port and then collapse each
+	// group of Listeners into a single Listener if the implementation
+	// determines that the Listeners in the group are "compatible". An
+	// implementation MAY also group together and collapse compatible
+	// Listeners belonging to different Gateways.
+	//
+	// For example, an implementation might consider Listeners to be
+	// compatible with each other if all of the following conditions are
+	// met:
+	//
+	// 1. Either each Listener within the group specifies the "HTTP"
+	//    Protocol or each Listener within the group specifies either
+	//    the "HTTPS" or "TLS" Protocol.
+	//
+	// 2. Each Listener within the group specifies a Hostname that is unique
+	//    within the group.
+	//
+	// 3. As a special case, one Listener within a group may omit Hostname,
+	//    in which case this Listener matches when no other Listener
+	//    matches.
+	//
+	// If the implementation does collapse compatible Listeners, the
+	// hostname provided in the incoming client request MUST be
+	// matched to a Listener to find the correct set of Routes.
+	// The incoming hostname MUST be matched using the Hostname
+	// field for each Listener in order of most to least specific.
+	// That is, exact matches must be processed before wildcard
+	// matches.
+	//
+	// If this field specifies multiple Listeners that have the same
+	// Port value but are not compatible, the implementation must raise
+	// a "Conflicted" condition in the Listener status.
+	//
+	// Support: Core
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=64
+	Listeners []Listener `json:"listeners"`
+
+	// Addresses requested for this Gateway. This is optional and behavior can
+	// depend on the implementation. If a value is set in the spec and the
+	// requested address is invalid or unavailable, the implementation MUST
+	// indicate this in the associated entry in GatewayStatus.Addresses.
+	//
+	// The Addresses field represents a request for the address(es) on the
+	// "outside of the Gateway", that traffic bound for this Gateway will use.
+	// This could be the IP address or hostname of an external load balancer or
+	// other networking infrastructure, or some other address that traffic will
+	// be sent to.
+	//
+	// The .listener.hostname field is used to route traffic that has already
+	// arrived at the Gateway to the correct in-cluster destination.
+	//
+	// If no Addresses are specified, the implementation MAY schedule the
+	// Gateway in an implementation-specific manner, assigning an appropriate
+	// set of Addresses.
+	//
+	// The implementation MUST bind all Listeners to every GatewayAddress that
+	// it assigns to the Gateway and add a corresponding entry in
+	// GatewayStatus.Addresses.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	Addresses []GatewayAddress `json:"addresses,omitempty"`
+}
+
+// Listener embodies the concept of a logical endpoint where a Gateway accepts
+// network connections.
+type Listener struct {
+	// Name is the name of the Listener. This name MUST be unique within a
+	// Gateway.
+	//
+	// Support: Core
+	Name SectionName `json:"name"`
+
+	// Hostname specifies the virtual hostname to match for protocol types that
+	// define this concept. When unspecified, all hostnames are matched. This
+	// field is ignored for protocols that don't require hostname based
+	// matching.
+	//
+	// Implementations MUST apply Hostname matching appropriately for each of
+	// the following protocols:
+	//
+	// * TLS: The Listener Hostname MUST match the SNI.
+	// * HTTP: The Listener Hostname MUST match the Host header of the request.
+	// * HTTPS: The Listener Hostname SHOULD match at both the TLS and HTTP
+	//   protocol layers as described above. If an implementation does not
+	//   ensure that both the SNI and Host header match the Listener hostname,
+	//   it MUST clearly document that.
+	//
+	// For HTTPRoute and TLSRoute resources, there is an interaction with the
+	// `spec.hostnames` array. When both listener and route specify hostnames,
+	// there MUST be an intersection between the values for a Route to be
+	// accepted. For more information, refer to the Route specific Hostnames
+	// documentation.
+	//
+	// Support: Core
+	//
+	// +optional
+	Hostname *Hostname `json:"hostname,omitempty"`
+
+	// Port is the network port. Multiple listeners may use the
+	// same port, subject to the Listener compatibility rules.
+	//
+	// Support: Core
+	Port PortNumber `json:"port"`
+
+	// Protocol specifies the network protocol this listener expects to receive.
+	//
+	// Support: Core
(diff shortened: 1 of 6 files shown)


apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+// GatewaySpec defines the desired state of Gateway.
+//
+// Not all possible combinations of options specified in the Spec are
+// valid. Some invalid configurations can be caught synchronously via a
+// webhook, but there are many cases that will require asynchronous
+// signaling via the GatewayStatus block.
+type GatewaySpec struct {
+	// GatewayClassName used for this Gateway. This is the name of a
+	// GatewayClass resource.
+	GatewayClassName ObjectName `json:"gatewayClassName"`
```
akutz, line 66: Is there a reason to not just name the field `ClassName`? Or even just `GatewayClass`? The latter would be consistent with [`Service.Spec.LoadBalancerClass`](https://github.com/kubernetes/api/blob/b754a94214be15ffc8d648f9fe6481857f1fc2fe/core/v1/types.go#L4509-L4521). That said, the `Name` suffix _is_ used by [`Ingress.Spec.IngressClassName`](https://github.com/kubernetes/api/blob/cd271e7523ef7b6b485aa42c0218c294ca00a001/networking/v1/types.go#L258-L269).  In fact, the `Name` suffix is _also_ used by [`PersistentVolumeClaim.Spec.StorageClassName`](https://github.com/kubernetes/api/blob/b754a94214be15ffc8d648f9fe6481857f1fc2fe/core/v1/types.go#L348-L351), so it appears as if the outlier is `Service.Spec.LoadBalancerClass` and the convention _is_ to use the `Name` suffix. Carry on 😃 

Also, the comment could be simplified:

```go
// GatewayClassName is the name of the GatewayClass resource used by
// this Gateway.
```

Finally, the use of `ObjectName` is interesting. I see it has a validation on it for min and max length. However, [the comments](https://github.com/mehabhalodiya/gateway-api/blob/205398acbffdebcaacd89ebbd159b5356ee49576/apis/v1beta1/shared_types.go#L338-L344) on the type _also_ indicate additional, possible validations:

```go
// ObjectName refers to the name of a Kubernetes object.
// Object names can have a variety of forms, including RFC1123 subdomains,
// RFC 1123 labels, or RFC 1035 labels.
//
```

I wonder if there is a reason to not just use `string` and let the webhook that will likely do the additional validation also be responsible for the min/max length validation?

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// That is, exact matches must be processed before wildcard
+	// matches.
+	//
+	// If this field specifies multiple Listeners that have the same
+	// Port value but are not compatible, the implementation must raise
+	// a "Conflicted" condition in the Listener status.
+	//
+	// Support: Core
+	//
+	// +listType=map
```
akutz, line 110: Ah, cool. I have not seen this used yet, so it's neat to see it!

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=64
```
akutz, line 113: So this is odd to me. Why a max of 64 listeners? That seems arbitrary?

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	//
+	// If this field specifies multiple Listeners that have the same
+	// Port value but are not compatible, the implementation must raise
+	// a "Conflicted" condition in the Listener status.
+	//
+	// Support: Core
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MinItems=1
```
akutz, line 112: I do not think a non-zero list should be required to set up a gateway. I can easily see a gateway being created to reserve or grab an IP address before any listeners are defined. I would like there to be an allowed min items of zero and mark this field as `+optional` and `omitempty`. Maybe there was some discussion about this at one point that I missed? Thanks!

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// For HTTPRoute and TLSRoute resources, there is an interaction with the
+	// `spec.hostnames` array. When both listener and route specify hostnames,
+	// there MUST be an intersection between the values for a Route to be
+	// accepted. For more information, refer to the Route specific Hostnames
+	// documentation.
+	//
+	// Support: Core
+	//
+	// +optional
+	Hostname *Hostname `json:"hostname,omitempty"`
```
akutz, line 178: Why is this not `HostName`? The Kubernetes [API conventions for naming](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions) state:

> Go field names must be PascalCase. JSON field names must be camelCase. Other than capitalization of the initial letter, the two should almost always match. No underscores or dashes in either.

So shouldn't this field be:

```go
Hostname *HostName `json:"hostName,omitempty"
```

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// implemented. Invalid Route rules can be ignored (sometimes that will mean
+	// the full Route). If a Route rule transitions from valid to invalid,
+	// support for that Route rule should be dropped to ensure consistency. For
+	// example, even if a filter specified by a Route rule is invalid, the rest
+	// of the rules within that Route should still be supported.
+	//
+	// Support: Core
+	// +kubebuilder:default={namespaces:{from: Same}}
+	// +optional
+	AllowedRoutes *AllowedRoutes `json:"allowedRoutes,omitempty"`
```
akutz, line 232: I'm not certain I think `AllowedRoutes` should be named as such. It is really just `AllowedRouteTypes` now, as the label selector from `RouteBindingSelector` in the initial v1alpha1 API is no longer present. The naming here gives the illusion of something that does not exist.

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+// * "example.com/bar" - Implementation-specific support
+//
+// Invalid values include:
+//
+// * "example.com" - must include path if domain is used
+// * "foo.example.com" - must include path if domain is used
+//
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=255
+// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([-a-zSA-Z0-9]*[a-zA-Z0-9])?$|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9]+$`
```
akutz, line 261: From where does this pattern originate? It may be a good idea to document the reference for this.

akutz (changes requested):
Thank you for submitting this PR! I am not finished with my review, but I have to go for now. I will come back and finish looking through the PR as soon as I am able. Certainly by the end of tomorrow.

hbagdi:
It may be best to not review such humungous PRs and find other alternatives to review the patch-diff between v1alpha2 and v1beta1. Thoughts?

robscott:
Yep, sorry for the confusion on this PR! These are great comments @akutz and I'll try to respond to them soon, but this particular PR was actually intended to be more of a checklist item on the milestone than the actual API review. I've been super busy, but my goal is to get a PR out that is just a diff of aps/v1alpha2 between v0.4.0 release and now, like @hbagdi suggested above. Of course if anyone else wants to take that on, I'm happy to let someone else own it. Either way, hopefully we can have a smaller diff ready for review on Monday.

akutz:
> It may be best to not review such humungous PRs

@hbagdi, I absolutely know what you mean, but at the same time I could not help but smile at just this part 😄

mehabhalodiya:
@hbagdi @robscott I have not done it correctly. Right?
I will make changes according to your comments. Sorry for this!
I am a newbie and trying to understand, it would be great if you people help me 🙂

robscott:
Hey @mehabhalodiya, sorry for the confusion! Your PR is headed in the right direction, I'd just missed creating the corresponding PR that compared v0.4.0 and master API types for the sake of a simpler API review. I'm hoping to have that in place on Monday, but will also work to respond to the great feedback here as well.

akutz:
If y'all want to ignore my feedback on this PR, I will post it on the newer PR when Robb posts it. I will still wait to resolve this PR's feedback until the other is posted, just so it is a little easier for me to find the comments.

mehabhalodiya:
@akutz @robscott @hbagdi So please let me everyone's final review after looking onto [Gateway API v0.5.0 API Review](https://github.com/kubernetes-sigs/gateway-api/pull/1086).

apis/v1beta1/gateway_types.go
youngnick: Hey @akutz! We're coming to the end of the v1beta1 release window, and both @robscott and I realized that your feedback had slipped through the cracks. So, I'll respond now where I think we probably won't make changes, and make issues for ones where we will.

For the min-max length validations and others, the general guideline we're using is to only add custom validation when we can't get it from the kubebuilder/OpenAPI validation.

apis/v1beta1/gateway_types.go
youngnick: We haven't written this down anywhere, I don't think, but we prefer not to have unbounded lists, since people will inevitably try to stuff 1000 things in there.

For times when we don't have a good idea, we do just pick an arbitrary limit that seems larger than what we think most uses will need, and accept that we may need to increase later. This is the case for all the limits you've called out here. Yes, they are arbitrary, but we needed a value >1 and < infinity, and we have no data about actual usage, so any number we pick is arbitrary.

apis/v1beta1/gateway_types.go
youngnick: In general, Kubernetes resources treat `Hostname` as a single word, rather than `Host Name`, and we've kept that convention going.

apis/v1beta1/gateway_types.go
youngnick: I think we discussed renaming this when we changed over the binding structure, and came down on the side that the amount of implementation churn this rename would generate was not worth the increase in readability. @robscott, am I mistaken?

apis/v1beta1/gateway_types.go
youngnick: I don't know if it originated anywhere other than someone building it for here.

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+type ProtocolType string
+
+const (
+	// Accepts cleartext HTTP/1.1 sessions over TCP. Implementations MAY also
+	// support HTTP/2 over cleartext. If implementations support HTTP/2 over
+	// cleartext on "HTTP" listeners, that MUST be clearly documented by the
+	// implementation.
+	HTTPProtocolType ProtocolType = "HTTP"
```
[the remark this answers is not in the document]
youngnick: Agreed, we should fix this.

mehabhalodiya:
Sorry for the delay!
I am re-copying the files from `v1alpha2`, just give me a moment! 
Further, Nick and Rob were going to split up responding to the comments from Andrew.
Thank you for your patience 🙂

apis/v1beta1/gateway_types.go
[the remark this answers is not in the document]
youngnick: I started making this change, but it kind of doesn't work because the same docstrings are used for both the Godoc documentation and the YAML API reference documentation, which does not care about the constant names. So, this convention of using "This constant does" means that the docstring can work with both.

apis/v1beta1/gateway_types.go
youngnick: I just realized that I didn't make clear that the other reason for the limits is the max record size in etcd, which defaults to about 1MB of text in deserialized form.

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// +optional
+	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+}
+
+// RouteGroupKind indicates the group and kind of a Route resource.
+type RouteGroupKind struct {
+	// Group is the group of the Route.
+	//
+	// +optional
+	// +kubebuilder:default=gateway.networking.k8s.io
```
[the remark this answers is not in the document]
youngnick: The group is the domain part of the `apiVersion` field, `v1` is the version. We've chosen to leave out version here to make it so that users don't have to manually update versions as we move through them - on the assumption we'll move through versions much faster than upstream Kubernetes does.

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// Kind is the kind of the Route.
+	Kind Kind `json:"kind"`
+}
+
+// GatewayAddress describes an address that can be bound to a Gateway.
+type GatewayAddress struct {
+	// Type of the address.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=IPAddress;Hostname;NamedAddress
```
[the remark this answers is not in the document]
youngnick: I've addressed this and other concerns in #1178.

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// set of Addresses.
+	//
+	// The implementation MUST bind all Listeners to every GatewayAddress that
+	// it assigns to the Gateway and add a corresponding entry in
+	// GatewayStatus.Addresses.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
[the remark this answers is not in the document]
youngnick: See ["No unbounded lists" comment.](https://github.com/kubernetes-sigs/gateway-api/pull/1073#discussion_r890829214)

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	//
+	// A set of common keys MAY be defined by the API in the future. To avoid
+	// any ambiguity, implementation-specific definitions MUST use
+	// domain-prefixed names, such as `example.com/my-custom-option`.
+	// Un-prefixed names are reserved for key names defined by Gateway API.
+	//
+	// Support: Implementation-specific
+	//
+	// +optional
+	// +kubebuilder:validation:MaxProperties=16
```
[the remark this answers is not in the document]
youngnick: See ["No unbounded lists" comment.](https://github.com/kubernetes-sigs/gateway-api/pull/1073#discussion_r890829214)

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+
+// GatewayStatus defines the observed state of Gateway.
+type GatewayStatus struct {
+	// Addresses lists the IP addresses that have actually been
+	// bound to the Gateway. These addresses may differ from the
+	// addresses in the Spec, e.g. if the Gateway automatically
+	// assigns an address from a reserved pool.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
```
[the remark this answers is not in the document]
youngnick: See ["No unbounded lists" comment.](https://github.com/kubernetes-sigs/gateway-api/pull/1073#discussion_r890829214)

apis/v1beta1/gateway_types.go
```diff
@@ -0,0 +1,806 @@
...
+	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:default={{type: "Scheduled", status: "Unknown", reason:"NotReconciled", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Listeners provide status for each unique listener port defined in the Spec.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=64
```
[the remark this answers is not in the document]
youngnick: See ["No unbounded lists" comment.](https://github.com/kubernetes-sigs/gateway-api/pull/1073#discussion_r890829214)

youngnick:
I think I've hit most of Andrew's concerns.

I think that we should actually get this merged into master asap, and then if we need to make any further v1alpha2 changes, we copy them over.

PR #1075, opened 2022-03-28
Fix golint on systems with selinux enforcing

markmc:
We run golangci-lint in a container with the local directory bind-mounted as a volume. However, without relabelling the directory
contents, selinux will disallow access. We can use the `:Z` volume option to ensure relabelling occurs.

See docs like https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label and https://docs.podman.io/en/latest/markdown/podman-run.1.html

On systems without selinux enforcing, this won't have any affect.

Without this change:

```
$ ./hack/verify-golint.sh
2022/03/28 11:59:12 Failed to write README file to cache dir /cache: stat /cache/README: no such file or directory
```

/kind bug


markmc:
/hold

Let's see what approach kubernetes/test-infra#25798 settles on

markmc:
> Let's see what approach [kubernetes/test-infra#25798](https://github.com/kubernetes/test-infra/pull/25798) settles on

Ok, that eventually merged with the `--security-opt="label=disable"` approach which turns off label separation, rather than the relabelling approach I originally proposed

Updated now, and should be good to merge!

/hold cancel

youngnick:
I'll approve this one, and leave someone else to check and LGTM.

/approve

shaneutt (approved):
/lgtm

PR #1078, opened 2022-03-29
respect --cleanup=0 in ConformanceTest::Run

mikemorris:
**What type of PR is this?**
/kind cleanup

Feel free to close if the intent is that ConformanceTest::Run should always delete resources after running each test, but I was a little confused when working on adding a new test and running it individually why my resources were always gone afterwards even when passing `--cleanup=0`

```release-note
NONE
```

### DIFF
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -116,7 +116,7 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 	}
 	for _, manifestLocation := range test.Manifests {
 		t.Logf("Applying %s", manifestLocation)
-		kubernetes.MustApplyWithCleanup(t, suite.Client, manifestLocation, suite.GatewayClassName, true)
+		kubernetes.MustApplyWithCleanup(t, suite.Client, manifestLocation, suite.GatewayClassName, suite.Cleanup)
 	}
 
 	test.Test(t, suite)


robscott:
My intention here was actually for --cleanup=0 to allow the base manifests to be persisted across test runs, not the resources from all individual test cases. That is admittedly rather confusing though. If we allowed test case resources to be persisted across test runs I think we'd run into test failures/unexpected outputs due to tests interacting with each other. (IE HTTPRoutes from different test cases both attaching to a Gateway that's defined by one of the base resources).

hbagdi:
Maybe an explicit `--disable-resource-cleanup` could help (which disables only cleaning up resources of test, not crds)? Could one not use a breakpoint?

youngnick:
Presumably we almost certainly only want to keep the test details for failed tests? Maybe there's something we can do there.

I agree that for something like this to work, there needs to be pretty careful structuring so that test resources left hanging around don't corrupt the run of later tests?

acnodal-tc:
LGTM

I fixed this myself because it's very useful to have the configuration hang around for post-mortem debugging after a failure. It sounds like there's a more substantial change up for discussion but in the mean time I'm +1 for merging this while we decide what to do.

mikemorris:
Per discussion on the weekly call, this currently conflates two distinct but valid use cases, which should likely be implemented as separate flags:
- Persisting the base resources between between runs
  - For efficiency when they make take several minutes to create - this is the current functionality of `--cleanup=0`, but which should maybe be renamed
- Not destroying test-specific resources on teardown for debugging a single failing test
  - This should likely be a separate flag, with guardrails to avoid polluting state on other tests
  - Maybe this is appropriate to link with a flag to exit early on failure like https://jestjs.io/docs/cli#--bail?

Closing this for now, but would love to see anyone interested pick up these changes.

acnodal-tc:
Would the first use case be handled by renaming the "cleanup" flag to something like "cleanup-base-resources"? I'm using --cleanup so I'd volunteer to make the name change.

acnodal-tc:
FYI: https://github.com/kubernetes-sigs/gateway-api/pull/1119

PR #1080, opened 2022-03-29
Support eventual consistency in conformance tests

nathancoleman:
**What type of PR is this?**

/kind bug
(best effort ^, is there a better option?)

We discussed in the sig meeting recently that "accepted" (or "ready" in the future) can only really indicate that the controller has synced all necessary configuration and that routes will begin working "soon".

In order for the conformance tests to support this, they must accept that HTTP requests to new routes may fail for some period of time but should become eventually consistent.

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -20,6 +20,7 @@ import (
 	"net/url"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -72,8 +73,29 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
 			req.Headers[name] = []string{value}
 		}
 	}
-	cReq, cRes, err := r.CaptureRoundTrip(req)
-	require.NoErrorf(t, err, "error making request")
+
+	var cReq *roundtripper.CapturedRequest
+	var cRes *roundtripper.CapturedResponse
+
+	// We expect eventual consistency once the resources are ready, so wait until
+	// we've made a successful request and gotten a response with desired status
+	// before making assertions on the response body.
+	require.Eventually(t, func() bool {
+		var err error
+
+		cReq, cRes, err = r.CaptureRoundTrip(req)
+		if err != nil {
+			t.Log("request failed, not ready yet")
+			return false
+		}
+
+		if cRes.StatusCode != expected.StatusCode {
+			t.Log("response missing expected status, not ready yet")
+			return false
+		}
+
+		return true
+	}, 60*time.Second, 1*time.Second, "error making request, never got expected status")
 
 	ExpectResponse(t, cReq, cRes, expected)
 }
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -17,6 +17,7 @@ limitations under the License.
 package http
 
 import (
+	"fmt"
 	"net/url"
 	"strings"
 	"testing"
@@ -45,9 +46,15 @@ type ExpectedRequest struct {
 	Headers map[string]string
 }
 
-// MakeRequestAndExpectResponse makes a request with the given parameters and
-// verifies the response matches the provided ExpectedResponse.
-func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwAddr string, expected ExpectedResponse) {
+const maxConsistencyPeriodPerRequest = 60 * time.Second
+const numConsistencyChecksPerRequest = 3
+
+// MakeRequestAndExpectEventuallyConsistentResponse makes a request with the given parameters,
+// understanding that the request may fail for some amount of time.
+//
+// Once the request succeeds and the response has the expected status code, assert that
+// a response matching the provided ExpectedResponse is returned consistently.
+func MakeRequestAndExpectEventuallyConsistentResponse(t *testing.T, r roundtripper.RoundTripper, gwAddr string, expected ExpectedResponse) {
 	t.Helper()
 
 	if expected.Request.Method == "" {
@@ -74,16 +81,13 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
 		}
 	}
 
-	var cReq *roundtripper.CapturedRequest
-	var cRes *roundtripper.CapturedResponse
-
 	// We expect eventual consistency once the resources are ready, so wait until
 	// we've made a successful request and gotten a response with desired status
 	// before making assertions on the response body.
 	require.Eventually(t, func() bool {
 		var err error
 
-		cReq, cRes, err = r.CaptureRoundTrip(req)
+		_, cRes, err := r.CaptureRoundTrip(req)
 		if err != nil {
 			t.Logf("Request failed, not ready yet: %v", err.Error())
 			return false
@@ -95,9 +99,18 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
 		}
 
 		return true
-	}, 60*time.Second, 1*time.Second, "error making request, never got expected status")
+	}, maxConsistencyPeriodPerRequest, 1*time.Second, "error making request, never got expected status")
 
-	ExpectResponse(t, cReq, cRes, expected)
+	// Once we've made a successful request and gotten a response with the
+	// desired status, assert that we get the expected response consistently.
+	for i := 1; i <= numConsistencyChecksPerRequest; i++ {
+		t.Run(fmt.Sprintf("consistency_check_%d", i), func(t *testing.T) {
+			cReq, cRes, err := r.CaptureRoundTrip(req)
+			require.NoError(t, err, "error making request after previous success")
+
+			ExpectResponse(t, cReq, cRes, expected)
+		})
+	}
 }
 
 // ExpectResponse verifies that a captured request and response match the
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -17,7 +17,6 @@ limitations under the License.
 package http
 
 import (
-	"fmt"
 	"net/url"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


conformance/utils/http/http.go
```diff
@@ -72,8 +73,29 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
...
-	cReq, cRes, err := r.CaptureRoundTrip(req)
-	require.NoErrorf(t, err, "error making request")
+
+	var cReq *roundtripper.CapturedRequest
+	var cRes *roundtripper.CapturedResponse
+
+	// We expect eventual consistency once the resources are ready, so wait until
+	// we've made a successful request and gotten a response with desired status
+	// before making assertions on the response body.
+	require.Eventually(t, func() bool {
```
howardjohn, line 83: Do we want to test eventually **consistent**? Technically a proxy doing a 50/50 200 or 404 passes currently. We could do a check that does backoff on failures, and on success retry immediately and ensure we get the "success" N times.

In Istio tests, we do all of ours this way fwiw and it seems to work pretty well. We do exponential backoff and ensure we get 3 successes consecutively

conformance/utils/http/http.go
```diff
@@ -72,8 +73,29 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
...
+		var err error
+
+		cReq, cRes, err = r.CaptureRoundTrip(req)
+		if err != nil {
+			t.Log("request failed, not ready yet")
+			return false
+		}
+
+		if cRes.StatusCode != expected.StatusCode {
+			t.Log("response missing expected status, not ready yet")
```
howardjohn, line 93: Can we log what we expect vs what we got?

conformance/utils/http/http.go
```diff
@@ -72,8 +73,29 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
...
+			return false
+		}
+
+		return true
+	}, 60*time.Second, 1*time.Second, "error making request, never got expected status")
```
nathancoleman, line 98: I think it makes sense to have this 60s cap be configurable, but I'm not sure at what granularity. Same cap for every test in the suite? Add as a func arg so every caller can specify their own?

Looking for some input from others here

conformance/utils/http/http.go
nathancoleman: @howardjohn if you would, please take a look at latest commits and let me know what ya think. Each test that needs to make assertions using this path now looks like this:

image

evankanderson:
Any chance that we could instead require a conservative `Ready` status for the `HttpRoute`, rather than expecting each client to build in eventual consistency handling?

I ask because making the Gateway API eventually consistent (rather than conservatively `Ready`) means that every development team writing an end-to-end test through the Gateway API will need to add this logic to their tests and potentially have to debug the "was it a Gateway-induced eventual consistency failure or an application-level failure" when their tests fail or time out. As an end-user developer writing tests, being able to distinguish between "the Gateway controller didn't finish" and "my application fails for 30s at startup" is pretty useful.

nathancoleman:
Hi @evankanderson ! There was some discussion about adding a `Ready` status to the `HTTPRoute`, and it sounds like that will be coming in the near term; however, I believe even then, things are expected to be eventually consistent due to the way these systems work.

The context from the discussion at yesterday's meeting might be helpful. It's up on YouTube [here](https://youtu.be/IwWuwGneXno?t=941) (linked to appropriate timestamp).

skriss (approved):
FYI I've been able to run the conformance tests using this PR against Contour's implementation successfully, whereas the tests in `master` where flaky for us. Thanks!

conformance/utils/http/http.go
```diff
@@ -72,10 +80,37 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
...
+
+		if cRes.StatusCode != expected.StatusCode {
+			t.Logf("Expected response to have status %d but got %d, not ready yet", expected.StatusCode, cRes.StatusCode)
+			return false
+		}
+
+		return true
+	}, maxConsistencyPeriodPerRequest, 1*time.Second, "error making request, never got expected status")
+
+	// Once we've made a successful request and gotten a response with the
```
howardjohn, line 104: Sorry to nitpick this a lot... a bit worried about the logic here. Typically what I have seen is that config changes will go from "all old config" to "mix of old and new" before arriving at "new config". So if we assert that once we see the 'new' state its immediately always 'new', we may see issues.

For example, if we have 2 pods implementing the gateway and 1 gets the new config first.

Here is how the Istio logic for this works: https://github.com/istio/istio/blob/30a866fdb5ad65eccde3dafe21d19f455addbd72/pkg/test/util/retry/retry.go#L182.

Basically we test that we get a success N times in a row -- but if we fail part way through the check we continue on retrying.

Also a bit of nitpick, but I am not sure we need the `consistency_check_%d` to show up in the test result? Doing the above would make that pretty hard to retain anyways

robscott: I agree with the logic suggested here by @howardjohn. Maybe a fairly low timeout like 30 seconds to get 3 consecutive successful requests would work?

robscott (commented):
Thanks for the work on this @nathancoleman! Agree with John's suggestion but otherwise this LGTM.

robscott:
>Any chance that we could instead require a conservative Ready status for the HttpRoute, rather than expecting each client to build in eventual consistency handling?

@nathancoleman I really wish that was possible. That was my original intent here, but unfortunately I think it was a bit too optimistic. Many (maybe most) of the implementations of this API simply don't have a way to know when the config they've written has actually been picked up and implemented. If there are good ways around this, I'm very open to alternatives, because I agree that a reliable "ready" condition on Routes would be more helpful than a condition that indicated that config had been updated and should be ready to use soon.

evankanderson:
Two thoughts:

1. Does it make sense to call this condition `Accepted` or `Programmed` rather than `Ready` to make the distinction clearer to users?
2. Conservative here could include adding a 30s delay after `Accepted`, assuming that's something like 2.5 stddev beyond the mean.

conformance/utils/http/http.go
nathancoleman: 👍 Not nitpicking IMO @howardjohn ! This makes a lot of sense and will help us avoid similar issues when we make the number of pods implementing the gateway configurable beyond `1` in the near future.

nathancoleman: @howardjohn I've pushed some updates to handle intermittent success in the style of what you linked. Let me know what ya think!

You can see an example run w/ log output under **Run tests** [here](https://github.com/hashicorp/consul-api-gateway/runs/5793540213?check_suite_focus=true).

nathancoleman:
@evankanderson @robscott I believe that future status additions for `HTTPRoute` -- "ready" vs. "accepted" vs. "programmed", etc., none of which exists today -- would be a good topic for the next kube-sig meeting on Monday, April 4.

Those evolutions in the Gateway API are only tangentially related to this PR. My goal with this change is just to unblock several implementations that are unable to run the conformance tests successfully today.

Would it make sense to you guys to move this discussion over to that venue? I think that would allow for more participation as well.

conformance/utils/http/http.go
```diff
@@ -44,9 +45,15 @@ type ExpectedRequest struct {
 	Headers map[string]string
 }
 
-// MakeRequestAndExpectResponse makes a request with the given parameters and
-// verifies the response matches the provided ExpectedResponse.
-func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwAddr string, expected ExpectedResponse) {
+const maxConsistencyPeriodPerRequest = 60 * time.Second
```
robscott, line 48: As I understand it, these tests require 3 consecutive successful requests to have occurred within this time period, and this time starts once a Route transitions to a status of accepted. If that's all correct, I think I'd prefer starting with a lower value of 30 seconds + renaming this to something like `maxTimeToConsistency`. Not sure on naming, but the `PerRequest` part throws me off in the current name.

conformance/utils/http/http.go
```diff
@@ -44,9 +45,15 @@ type ExpectedRequest struct {
...
+const numConsistencyChecksPerRequest = 3
```
robscott, line 49: As I understand this, this variable controls the number of consecutive successful requests we require for a test to be considered successful. If that's a correct interpretation, the current name doesn't clearly communicate that to me. Maybe something like `requiredConsecutiveSuccesses` or similar would be clearer. It's hard to come up with a perfectly descriptive name for either of these vars, so it may also make sense to add comments describing the purpose of each of them.

conformance/utils/http/http.go
```diff
@@ -72,12 +79,49 @@ func MakeRequestAndExpectResponse(t *testing.T, r roundtripper.RoundTripper, gwA
...
+		cRes         *roundtripper.CapturedResponse
+		err          error
+		numSuccesses int
+	)
+
+	require.Eventually(t, func() bool {
+		cReq, cRes, err = r.CaptureRoundTrip(req)
+		if err != nil {
+			numSuccesses = 0
+			t.Logf("Request failed, not ready yet: %v", err.Error())
```
robscott, line 102: Looking at the ways CaptureRoundTrip can return an error, I'm wondering if we should exit early with a test failure here like we were before (`require.NoErrorf(t, err, "error making request")`). For example, if the error we're getting is from `http.NewRequestWithContext` (https://github.com/kubernetes-sigs/gateway-api/blob/master/conformance/utils/roundtripper/roundtripper.go#L87), we're not going to have any more success retrying with the same params.

robscott (commented):
Thanks @nathancoleman! A few more tiny nits but otherwise this LGTM.

conformance/utils/http/http.go
nathancoleman: `CaptureRoundTrip` will return an error in the event that Envoy has not stood up listeners on the gateway yet even if it's been "programmed", so I don't think we should fail the test at this point. The general progression we see in our conformance test runs is:

```log
# Envoy isn't ready yet, so this returns an error from CaptureRoundTrip
http.go:102: Request failed, not ready yet: Get "http://172.18.255.201/": dial tcp 172.18.255.201:80: connect: connection refused

http.go:114: Request has passed 1 times in a row of the desired 3, not ready yet
http.go:114: Request has passed 2 times in a row of the desired 3, not ready yet
http.go:118: Request has passed 3 times in a row of the desired 3, ready!
```
(taken from the **Run tests** step in our Actions workflow [here](https://github.com/hashicorp/consul-api-gateway/runs/5821189907?check_suite_focus=true))

conformance/utils/http/http.go
nathancoleman: The route transition happens outside the scope of the functions I've modified/added. The timer starts when the function is called essentially. The caller has already verified the status transition before calling `MakeRequestAndExpectEventuallyConsistentResponse`

conformance/utils/http/http.go
robscott: Good catch, that makes sense to me.

conformance/utils/http/http.go
nathancoleman: Renamed + added comments

robscott:
Thanks @nathancoleman!

/lgtm
/approve

Issue #482, opened 2020-11-25
Define standard conformance format for a GatewayClass to communicate what it supports to operators


mark-church:
**What would you like to be added**:
GatewayClasses encapsulate the capabilities that an implementation of Service APIs supports. However, for the platform operators or application operators who might deploy Gateways in a cluster may not know exactly what within the API a given GatewayClass supports. This issue proposes a standard way for all GatewayClasses to communicate to a user what it supports.

Let's say I have an `acme-lb` GatewayClass. The GatewayClass should be able to tell me the following:

```
kind: Gateway
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: supported
spec:
  gatewayClassName: supported
  listeners:  
  - protocol: 
      http: supported
      tcp: unsupported
      udp: unsupported
    port: supported
    routes:
      kind: 
        HTTPRoute: supported
        TCPRoute: unsupported
        UDPRoute: unsupported
      selector:
        matchLabels: supported
...
---
HTTPRoute:
  spec:
    hostnames: supported
    rules:
    - matches:
      - path:
          type: 
            prefix: supported
            exact: supported
            regularexpression: unsupported
            implementationspecific: unsupported
        headers:
          type:
            exact: supported
            implementationspecific: 
              contains: supported # Will match traffic with any single header that contains this string
              doesnotcontain: supported # Will match traffic if no headers contain this string 
...
```

This is a terrible example of the format but you get the idea :) Perhaps OpenAPI has some kind of capability for this.

The goals this aims to achieve:

- Provide a standardized way for all GatewayClasses to communicate which aspects of the API they support.
- Core fields must be supported to achieve conformance, but since extended fields can be optional it will be very useful to understand which extended fields are implemented by a given GatewayClass.
- Custom fields (like annotations, ImplementationSpecific fields, or CRDs that can be referenced) should also be able to be communicated in some way so operators can understand how a GatewayClass has been extended with additional functionality.
- Provide a way for the provider to give additional context to describe implementation-specific aspects of a core, extended, or custom field. The example above with `contains` illustrates this.

How a GatewayClass provider communicates to the operator what its capabilities are can be done in many different ways:
- Perhaps conformance tests spits this out as a standardized output and providers can simply put this in their docs
- Perhaps comments or fields within a GatewayClass implementation can fill out these fields. This would give the opportunity for implementors to add special comments and context regarding support for specific fields. This might be the case where a field is supported but how it functions requires more context or explanation. 

[...]

PR #1081, opened 2022-03-29
Add conformance tests for ReferencePolicy

mikemorris:
**What type of PR is this?**

/kind feature

Adds conformance tests for ReferencePolicy. Given that [ReferencePolicy is _not_ among the resources graduating to v1beta1](https://github.com/kubernetes-sigs/gateway-api/issues/1041), we may need to add some ability fro implementations to opt-in to conformance tests for specific experimental or alpha features, similar to the functionality proposed in #482 

~~The `HTTPRouteMissingReferencePolicy` test could likely be consolidated into the primary valid `HTTPRouteReferencePolicy` test~~

I plan to handle lifecycle behavior such as validating a mutation like deleting a ReferencePolicy causing an HTTPRoute (and associated Listener?) to switch their `ResolvedRefs` status to `RefNotPermitted` and _stop_ sending traffic in a follow-up PR once we get alignment on expected behavior and if we want to mutate Kubernetes objects like that inside these tests before implementing.

Refs #20

```release-note
Add conformance tests for ReferencePolicy
```


conformance/tests/httproute-reference-policy.go
```diff
@@ -0,0 +1,65 @@
...
+			http.ExpectResponse(t, cReq, cRes, http.ExpectedResponse{
+				Request: http.ExpectedRequest{
+					Method: "GET",
+					Path:   "/",
+				},
+				StatusCode: 200,
+				Backend:    "web-backend",
+				Namespace:  "gateway-conformance-web-backend",
+			})
+		})
```
nathancoleman, line 63: This will need to use [this function](https://github.com/kubernetes-sigs/gateway-api/blob/df0cc90c639239d9126fac11f178fd1c0c6d7a09/conformance/utils/http/http.go#L63) since https://github.com/kubernetes-sigs/gateway-api/pull/1080 was merged

conformance/tests/httproute-missing-reference-policy.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteMissingReferencePolicy)
```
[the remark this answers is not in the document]
mikemorris: That definitely makes sense for most cases and would be a reason to keep this as a separate test from the successful cross-namespace ref using `ReferencePolicy`, I just wasn't sure about the [exception](https://gateway-api.sigs.k8s.io/v1alpha2/api-types/referencepolicy/#exceptions) case - if an implementation falling in that category should show this test as failing with an annotation/explanation or be able to skip it?

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,67 @@
...
+	ShortName:   "HTTPRouteInvalidReferencePolicy",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should fail to attach to a Gateway in the same namespace if the route has a backendRef Service in the gateway-conformance-app-backend namespace and a ReferencePolicy exists but does not grant permission to route to that specific Service",
+	Manifests:   []string{"tests/httproute-invalid-reference-policy.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		routeName := types.NamespacedName{Name: "invalid-reference-policy", Namespace: "gateway-conformance-infra"}
+		gwName := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
+
+		// TODO: Determine if this is actually what we want. It is likely
+		// preferable to have status set with some kind of warning/error message
+		// but that is also unlikely to be universally achievable.
```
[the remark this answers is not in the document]
mikemorris: Added a draft implementation of checking for an `Accepted: false` status condition, which seems like it should be sufficiently standard for these cases.

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteInvalidReferencePolicy)
```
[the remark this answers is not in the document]
mikemorris: Added a proposed implementation of an `ExtendedSupport` field to the `ConformanceTest` struct, taking a slice of `ConformanceTestSupport` objects but introducing just `SupportReferencePolicy` to start. Figured having some amount of fine-grained control is going to quickly be more valuable than opting into conformance for all Extended conformance features or none. Could consider adding a CLI flag for these, but that would probably be best addressed in a separate PR.

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,86 @@
...
+		ns := v1alpha2.Namespace(gwNN.Namespace)
+		kind := v1alpha2.Kind("Gateway")
+
+		t.Run("Route status should have a route parent status with an Accepted condition set to False", func(t *testing.T) {
+			parents := []v1alpha2.RouteParentStatus{{
+				ParentRef: v1alpha2.ParentReference{
+					Group:     (*v1alpha2.Group)(&v1alpha2.GroupVersion.Group),
+					Kind:      &kind,
+					Name:      v1alpha2.ObjectName(gwNN.Name),
+					Namespace: &ns,
```
nathancoleman, line 58: The namespace assertion won't work here for our implementation unless the `HTTPRoute` includes the namespace on its `parentRefs` over in `httproute-invalid-reference-policy.yaml`. I'm not sure if this is how it's supposed to work - in which case simply adding the namespace to `parentRefs[0]` has us passing these tests - or if we should be implicitly adding the `HTTPRoute`'s own namespace to the parent ref if not specified.

Can we look into this together later today or next week?

conformance/tests/httproute-missing-reference-policy.go
```diff
@@ -0,0 +1,81 @@
...
+		ns := v1alpha2.Namespace(gwNN.Namespace)
+		kind := v1alpha2.Kind("Gateway")
+
+		t.Run("Route status should have a route parent status with an Accepted condition set to False", func(t *testing.T) {
+			parents := []v1alpha2.RouteParentStatus{{
+				ParentRef: v1alpha2.ParentReference{
+					Group:     (*v1alpha2.Group)(&v1alpha2.GroupVersion.Group),
+					Kind:      &kind,
+					Name:      v1alpha2.ObjectName(gwNN.Name),
+					Namespace: &ns,
```
nathancoleman, line 53: Same thing from the above comment applies to the namespace assertion here.

nathancoleman, line 53: Same thing from the above comment applies to the namespace assertion here.

nathancoleman (commented):
Our `ReferencePolicy` implementation is passing these tests except for the namespace issue. Not sure which path is the "correct" one for resolving yet.

conformance/tests/httproute-invalid-reference-policy.go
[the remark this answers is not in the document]
robscott: Good catch, thanks @mikemorris! I agree we should probably just change that linked test to the ideal case and only change it back to its current more lenient state if we determine it's necessary.

conformance/tests/httproute-missing-reference-policy.go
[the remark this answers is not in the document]
robscott: Good call on that exception. Maybe we should be introducing two separate but related concepts here. One would be `SupportedFeatures`, the other would be `Exemptions`, both would behave similarly. Each test could define a list of Features and/or Exemptions. If the test runner did not have a Feature from the `supportedFeatures` list specified or had an Exemption from the `exemptions` list specified, the test would be skipped.

Related comment: We should probably rename that test now that we'll have 2 tests related to HTTPRoute references crossing namespace boundaries in invalid ways. Maybe `httproute-invaliid-cross-namespace-backend-ref` and `httproute-invaliid-cross-namespace-parent-ref`?

conformance/conformance_test.go
```diff
@@ -52,6 +52,9 @@ func TestConformance(t *testing.T) {
 		GatewayClassName: *gatewayClassName,
 		Debug:            *showDebug,
 		Cleanup:          *shouldCleanup,
+		ExtendedSupport: []suite.ConformanceTestSupport{
+			suite.SupportReferencePolicy,
+		},
```
robscott, line 57: Nit: I think we should just call this "Features" or something more generic, and not opt-in to any extended feature testing by default.

```suggestion
		Features: []suite.ConformanceTestFeatures{},
```

conformance/tests/httproute-reference-policy.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+var HTTPRouteReferencePolicy = suite.ConformanceTest{
+	ShortName:   "HTTPRouteReferencePolicy",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace, with a backendRef in the gateway-conformance-web-backend namespace, should attach to Gateway in the gateway-conformance-infra namespace",
+	Manifests:   []string{"tests/httproute-reference-policy.yaml"},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+		if !slices.Contains(s.ExtendedSupport, suite.SupportReferencePolicy) {
+			t.Skip("Skipping ReferencePolicy conformance test")
+		}
```
robscott, line 41: Instead of adding this inside the `Test` func, maybe add a `Features` field to `suite.ConformanceTest` and update our suite.Run func to consider this when running tests.

conformance/utils/suite/suite.go
```diff
@@ -25,6 +25,15 @@ import (
 	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
 )
 
+// ConformanceTestSupport is a specific subset of extended support for opting in to
```
robscott, line 28: Maybe call this something more generic like `SupportedFeature`?

conformance/tests/httproute-invalid-reference-policy.go
mikemorris: Pushed up a commit which I hope should fix this - I believe passing `false` as the `namespaceRequired` argument to `HTTPRouteMustHaveParents` should allow this field to be omitted from the route config (with the prior assumption that the route and expected parent being checked are in the same namespace).

I think there's a bit of an edge case that this boolean flag lets slip through in that it wouldn't be able to properly validate a route with multiple parents where one parent is in the same namespace (so namespace on the status could be optional), but a second parent in a different namespace, where the namespace would be required - I don't think we have any tests exercising a scenario like that yet though, so maybe something to improve as we look into adding route delegation conformance tests.

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,84 @@
...
+			// There are two valid ways to represent this:
+			// 1. No listeners in status
+			// 2. One listener in status with 0 attached routes
+			if len(gw.Status.Listeners) == 0 {
+				// No listeners in status.
+			} else if len(gw.Status.Listeners) == 1 {
+				require.Equal(t, int32(0), gw.Status.Listeners[0].AttachedRoutes)
+			} else {
+				t.Errorf("Expected no more than 1 listener in status, got %d", len(gw.Status.Listeners))
+			}
```
skriss, line 81: Is there a canonical definition for "Attached"? In this case, if I'm not mistaken, the route should still result in `app-backend-v1` being programmed with 90% weight, with a 10% weight being applied to a 503 response for the `app-backend-v2` backend that's not covered by a ReferencePolicy, based on https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/httproute_types.go#L191-L196. We (contour) had been considering the route to be "Attached" in this case since it's being at least partially programmed, but I'm not seeing a clear definition one way or another.

conformance/utils/suite/suite.go
```diff
@@ -114,6 +128,15 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 	if test.Parallel {
 		t.Parallel()
 	}
+
+	// Check that all features excerised by the test have been opted into by
+	// the suite.
+	for feature := range test.Features {
```
skriss, line 134:
```suggestion
	for _, feature := range test.Features {
```

skriss, line 134:
```suggestion
	for _, feature := range test.Features {
```

conformance/tests/httproute-invalid-reference-policy.go
nathancoleman: I'm highly in favor of clarifying optional fields as much we possibly can in the spec. IMO they represent the biggest risk of divergence between implementations.

conformance/tests/httproute-invalid-reference-policy.go
mikemorris: Agreed that the spec could be more clear on how these situations are intended to be handled - it's potentially more complicated when either resource changes after having been iniitally accepted. I added a discussion topic for the weekly meeting later today to hopefully hear perspectives from various implementations.

mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1112 for further discussion.

conformance/tests/httproute-missing-reference-policy.go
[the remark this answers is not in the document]
nathancoleman: Though I'm not sure how frequently we expect this to be used, [ReferencePolicy exceptions](https://gateway-api.sigs.k8s.io/v1alpha2/api-types/referencepolicy/#exceptions) includes this note which adds to the need for exceptions here:
> There are some situations where it MAY be acceptable to ignore ReferencePolicy in favor of some other security mechanism. This MAY only be done if other mechanisms like NetworkPolicy can effectively limit cross-namespace references by the implementation.

conformance/tests/httproute-invalid-reference-policy.go
mikemorris: @skriss I updated this test to instead check for `ResolvedRefs { status: False, reason: RefNotPermitted }` conditions on both the Route and ListenerStatus, and added a check that HTTP requests eventually start returning 503 status codes - does this match your expected behavior?

conformance/tests/httproute-invalid-reference-policy.go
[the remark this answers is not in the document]
mikemorris: I think the expected handling of the cross-namespace status is still slightly fuzzy so I added a `not accepted` section to https://github.com/kubernetes-sigs/gateway-api/issues/1112 but will hold off on any changes to that test for now.

mikemorris:
Pushed up some relatively significant logical changes in response to reviews and discussion, also extracted https://github.com/kubernetes-sigs/gateway-api/pull/1114 for adding `RouteConditionReason` if that would be easier to review separately. This should be ready for another review pass now, and hopefully reaching some consensus around minimum expectations for Routes with invalid BackendRefs.

conformance/tests/httproute-invalid-reference-policy.go
skriss: Definitely makes sense for the HTTPRoute - our implementation is currently using a different reason, but it's wrong & should be switched to `RefNotPermitted`.

We're not currently setting that condition on the Listener status, though - I do see in the API docs that the `RefNotPermitted` listener condition reason says "This reason is used with the “ResolvedRefs” condition when one of the Listener’s Routes has a BackendRef to an object in another namespace, where the object in the other namespace does not have a ReferencePolicy explicitly allowing the reference.", so I guess we should be, though I also wonder why this condition is needed on the Listener (where it could be relevant to any one of the listener's routes) rather than just on the HTTPRoute itself. Looks like this has already been raised in #1112 (specifically in https://github.com/kubernetes-sigs/gateway-api/issues/1112#issuecomment-1099655762).

mikemorris: Thanks for the detail @skriss! I skipped the listener check for now to hopefully get this merge-able and can follow up on that later once we can reach consensus across implementations on expected behavior.

conformance/utils/kubernetes/helpers.go
```diff
@@ -231,21 +231,88 @@ func parentsMatch(t *testing.T, expected, actual []v1alpha2.RouteParentStatus, n
 				return false
 			}
 		}
-		if len(aParent.Conditions) < len(eParent.Conditions) {
-			t.Logf("Expected more conditions to be present")
+		if !conditionsMatch(t, aParent.Conditions, eParent.Conditions) {
```
skriss, line 234: expected should come before actual:
```suggestion
		if !conditionsMatch(t, eParent.Conditions, aParent.Conditions) {
```

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteInvalidCrossNamespaceBackendRef)
+}
+
+var HTTPRouteInvalidCrossNamespaceBackendRef = suite.ConformanceTest{
+	ShortName:   "HTTPRouteInvalidCrossNamespaceBackendRef",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should set a ResolvedRefs status False with reason RefNotPermitted when attempting to bind to a Gateway in the same namespace if the route has a BackendRef Service in the gateway-conformance-web-backend namespace and a ReferencePolicy granting permission to route to that Service does not exist",
+	Exemptions: []suite.Exemptions{
```
nathancoleman, line 37:
```suggestion
	Exemptions: []suite.ExemptFeature{
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -231,21 +231,88 @@ func parentsMatch(t *testing.T, expected, actual []v1alpha2.RouteParentStatus, n
...
+		}
+		if !reflect.DeepEqual(aListener.SupportedKinds, eListener.SupportedKinds) {
+			t.Logf("Expected SupportedKinds to be %v, got %v", eListener.SupportedKinds, aListener.SupportedKinds)
+			return false
+		}
+		if aListener.AttachedRoutes != eListener.AttachedRoutes {
+			t.Logf("Expected AttachedRoutes to be %v, got %v", eListener.AttachedRoutes, aListener.AttachedRoutes)
+			return false
+		}
+		if !conditionsMatch(t, aListener.Conditions, eListener.Conditions) {
```
mikemorris, line 292:
```suggestion
		if !conditionsMatch(t, eListener.Conditions, aListener.Conditions) {
```

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,113 @@
...
+				}},
+			}}
+
+			kubernetes.HTTPRouteMustHaveParents(t, s.Client, routeNN, parents, false, 60)
+		})
+
+		// TODO(mikemorris): Un-skip check for Listener ResolvedRefs
+		// RefNotPermitted condition and/or add check for attached
+		// routes and any expected Listener conditions once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
```
robscott, line 74: It seems like #1114 may be nearly good enough to unblock this?

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,113 @@
...
+
+			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, gwNN, listeners, 60)
+		})
+
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, s.Client, s.ControllerName, gwNN, routeNN)
+
+		// TODO(mikemorris): Add check for HTTP requests successfully reaching
+		// app-backend-v1 at path "/" if it is determined that a Route with at
+		// at least one allowed BackendRef should be accepted by a Gateway
+		// and partially configured.
```
robscott, line 98: Can you add a tracking issue for this?

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
```diff
@@ -0,0 +1,75 @@
...
+				}},
+			}}
+
+			kubernetes.HTTPRouteMustHaveParents(t, suite.Client, routeNN, parents, false, 60)
+		})
+
+		// TODO(mikemorris): Add check for Listener attached routes or
+		// ResolvedRefs RefNotPermittedAccepted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
```
robscott, line 73: Can you attach these TODOs to #1112?

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
```diff
@@ -0,0 +1,75 @@
...
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-backend-ref", Namespace: "gateway-conformance-infra"}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
+
+		ns := v1alpha2.Namespace(gwNN.Namespace)
+		kind := v1alpha2.Kind("Gateway")
+
+		// TODO(mikemorris): Add check for Accepted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
```
robscott, line 50: Can you add this to a list on #1112?

robscott (commented):
Thanks @mikemorris! Just a few small things but otherwise LGTM

conformance/utils/suite/suite.go
```diff
@@ -19,32 +19,55 @@ package suite
...
 	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
 	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
 )
 
+// ExemptFeature allows opting out of core conformance tests at an
+// individual feature granularity.
+type ExemptFeature string
+
+const (
+	// This option indicates support for the ReferencePolicy object.
```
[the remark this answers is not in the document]
mikemorris: 🤦

conformance/utils/kubernetes/helpers.go
```diff
@@ -231,21 +231,88 @@ func parentsMatch(t *testing.T, expected, actual []v1alpha2.RouteParentStatus, n
...
+		}
+
+		actual = gw.Status.Listeners
+
+		return listenersMatch(t, listeners, actual), nil
+	})
+	require.NoErrorf(t, waitErr, "error waiting for Gateway status to have listeners matching expectations")
+}
+
+// TODO(mikemorris): this and parentsMatch could possibly be rewritten as a generic function?
```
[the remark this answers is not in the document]
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1142

conformance/tests/httproute-invalid-reference-policy.go
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1143 which also addresses the TODO above about checking for an `Accepted` condition

conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Run("Simple HTTP request should not reach app-backend-v2", func(t *testing.T) {
+			http.MakeRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, gwAddr, http.ExpectedResponse{
+				Request: http.ExpectedRequest{
+					Method: "GET",
+					Path:   "/v2",
+				},
+				StatusCode: 503,
+				// TODO: should these fields be populated when the BackendRef is invalid?
+				Backend:   "app-backend-v2",
+				Namespace: "gateway-conformance-app-backend",
```
robscott, line 109: I missed this earlier, but I think these lines should be removed since they won't be checked in any way (https://github.com/kubernetes-sigs/gateway-api/blob/master/conformance/utils/http/http.go#L137).

robscott (commented):
Thanks @mikemorris! One more tiny nit otherwise LGTM.

/approve

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
mikemorris: Captured as TODO items in the "unpermitted" section of #1112

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
```diff
@@ -0,0 +1,75 @@
...
+					Status: metav1.ConditionFalse,
+					Reason: string(v1alpha2.RouteReasonRefNotPermitted),
+				}},
+			}}
+
+			kubernetes.HTTPRouteMustHaveParents(t, suite.Client, routeNN, parents, false, 60)
+		})
+
+		// TODO(mikemorris): Add check for Listener attached routes or
+		// ResolvedRefs RefNotPermittedAccepted condition once
```
mikemorris, line 71:
```suggestion
		// Listener ResolvedRefs RefNotPermitted condition once
```

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
mikemorris: Added an unresolved item to the TODO list in the "unpermitted" section of #1112

conformance/tests/httproute-invalid-reference-policy.go
mikemorris:
```suggestion
```

skriss (commented):
I re-ran this against Contour with the latest revisions and it's still passing so :+1: from me.

robscott:
Thanks @mikemorris!

/lgtm

PR #1083, opened 2022-03-31
backlinks from example manifests to guides

### LINKED ISSUE #446: Add backlinks from example manifests to guides

Add YAML comments in `examples/*.yaml` to list the guides in which an example manifests is being used:
```yaml
# used in:
# - docs-src/foo-guide.md
# - docs-src/bar-guide.md
```

**Why is this needed**:

Changes to example manifests and docs can get out of sync and skew.

kundan2707:
/kind documentation

Fixes #446 

```release-note
None
```

### DIFF
--- a/examples/experimental/v1alpha2/traffic-matching-tcp.yaml
+++ b/examples/experimental/v1alpha2/traffic-matching-tcp.yaml
@@ -1,3 +1,6 @@
+# Used in
+# - site-src/v1alpha2/guides/tcp.md
+
 apiVersion: gateway.networking.k8s.io/v1alpha2
 kind: TCPRoute
 metadata:
--- a/examples/v1alpha1/basic-http.yaml
+++ b/examples/v1alpha1/basic-http.yaml
@@ -1,3 +1,6 @@
+# Used in
+# - site-src/v1alpha1/api-types/httproute.md
+
 apiVersion: networking.x-k8s.io/v1alpha1
 kind: GatewayClass
 metadata:
--- a/examples/v1alpha1/basic-tcp.yaml
+++ b/examples/v1alpha1/basic-tcp.yaml
@@ -1,3 +1,6 @@
+# Used in
+# - site-src/v1alpha1/guides/tcp.md
+
 apiVersion: networking.x-k8s.io/v1alpha1
 kind: GatewayClass
 metadata:
--- a/examples/v1alpha1/cross-namespace-routing/gateway.yaml
+++ b/examples/v1alpha1/cross-namespace-routing/gateway.yaml
@@ -1,3 +1,6 @@
+# Used in
+# - site-src/v1alpha1/guides/multiple-ns.md
+
 apiVersion: networking.x-k8s.io/v1alpha1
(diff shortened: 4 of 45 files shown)


youngnick:
Nice work!

/lgtm but will leave for someone else to hit the approve button.

robscott:
Thanks for the work on this @kundan2707! This looks great, but unfortunately it had some unintended consequences :(. All of these comments end up being included in the docs as well: https://deploy-preview-1083--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/guides/traffic-splitting/#guide. I spent a bit of time trying to figure out if there was a workaround here that would allow us to exclude comments from the rendered docs, but couldn't find a great solution for this.

Looking back at the original issue, it seems like the goal was to prevent code snippets from being updated without first considering all the files that would be affected by the change. Maybe a better solution here would be to have some kind of GitHub action/presubmit that did the following:

1. Check to see if any examples had been modified
2. Search to see if those examples have been included in any markdown
3. If 1 and 2 are true, add a comment on the PR with a list of examples that have been changed + where they are referenced in docs

I realize that's significantly more complicated than this approach, so open to alternatives. Unfortunately I think having these comments visible in our docs would be too distracting, so I want to put a hold on this until we can find a way around that or build some kind of automation like I suggested above.

/hold

hbagdi:
Great work.
Rob's concern is fair.
Another path would be to play around with the plugin to see if we can exclude some (not all maybe) comments from the YAML.
For example: https://pypi.org/project/mkdocs-include-markdown-plugin/
This would require a bit more work on the tooling side.

kundan2707:
> Thanks for the work on this @kundan2707! This looks great, but unfortunately it had some unintended consequences :(. All of these comments end up being included in the docs as well: https://deploy-preview-1083--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/guides/traffic-splitting/#guide. I spent a bit of time trying to figure out if there was a workaround here that would allow us to exclude comments from the rendered docs, but couldn't find a great solution for this.
> 
> Looking back at the original issue, it seems like the goal was to prevent code snippets from being updated without first considering all the files that would be affected by the change. Maybe a better solution here would be to have some kind of GitHub action/presubmit that did the following:
> 
> 1. Check to see if any examples had been modified
> 2. Search to see if those examples have been included in any markdown
> 3. If 1 and 2 are true, add a comment on the PR with a list of examples that have been changed + where they are referenced in docs
> 
> I realize that's significantly more complicated than this approach, so open to alternatives. Unfortunately I think having these comments visible in our docs would be too distracting, so I want to put a hold on this until we can find a way around that or build some kind of automation like I suggested above.
> 
> /hold

@robscott Thanks for you suggestion. I will look for an alternative and discuss once found.

youngnick:
I think that something like this is a good idea, but we need a way to not have the links show up in the rendered site version. I'm going to close this out for now, and encourage @kundan2707 to open a new issue to talk about the approach first.

Thanks for the work @kundan2707, hopefully we can work together to figure out the right way to proceed.

/close

PR #1085, opened 2022-04-04
GEP-1058 Add support for Route Inclusion and Delegation

Jeff-Apple:
**What type of PR is this?**
/kind gep

This is the formal GEP-10058 document

```release-note
NONE
```


robscott:
Thanks @Jeff-Apple!

/ok-to-test

robscott:
Looks like this test failure was an issue caused by my webhook integration, hopefully it's just a flake.

/retest

youngnick:
I haven't got to it today, but I will open a PR back to your branch @Jeff-Apple, with my changes to the status section tomorrow.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+
+During the reconciliation process, any conflict of configuration settings
+between the parent and child routes is considered fatal to the transaction and
+the entire Binding fails.
+
+A configuration conflict exists when the reconciliation process would result in
+an Ephemeral route configuration that would not be valid as a single, standalone
+`Route`.  This means, in concept, a user could save the generated Ephemeral
+route config to `my-foo-route.yaml`, remove the parent and child routes from the
+cluster, then load `my-foo-route.yaml` into the cluster, and the system system
```
robscott, line 309:
```suggestion
cluster, then load `my-foo-route.yaml` into the cluster, and the system
```

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+would accept and deploy it as a valid config.
+
+A config conflict may not be between config elements in the child and the
+parent. Instead, the conflict may be between config elements from two (or more)
+child routes that attach to the same `RouteRule` in teh parent route. For this
```
robscott, line 314:
```suggestion
child routes that attach to the same `RouteRule` in the parent route. For this
```

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+
+### Hostname Settings
+
+In the current version of the standard,  `hostname` can be specified at both the
+listener and route level. When both are specified, each is treated as a partial
+FQDN and the two strings are combined at the point of overlap. In this proposal,
+that model is extended to allow `hostname` to be specified at the listener,
+parent route **and** child route levels. Specifying `hostname` at any
+combination of these levels is allowed but the `hostnames` are always combined
+in the sequence: listener, parent route, child route.
```
robscott, line 337: Is this an important use case? If not, would it simplify implementation if we said that child routes could not specify hostnames?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+   `RouteRule` in the child route. This creates a set of `RouteRule`\(s\) that
+   are added to the Ephemeral route.
+   1. If the child route attaches to multiple `RouteRule`\(s\) in the parent,
+      each `RouteRule` from the parent is independently reconciled with the
+      child's `RouteRules`\(s\). For each `RouteRul` in the parent, an
+      additional set of `RouteRule`\(s\) is created and added to the Ephemeral
+      route.
+6. Path Matching: If `HTTPPathMatch` is configured in both the parent and child
+   `Routes`, allow **only** substring prefix matching on path, without any
+   wildcard or RegEx matching,
```
robscott, line 425: I think path matching will always be configured because our default path match is prefix + "/". A lot of these limitations seem like things we could enforce with webhook validation. Maybe we should have a section covering what additions we'll need to webhook validation.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+    1. As suggested in [this comment in GitHub Issue #634][gh-issue-634-comment]
+       to avoid cycles.
+    2. The two path strings are concatenated to form the complete path to be
+       matched.
+7. Hostname: Extend, to child routes, the current model of configuring the
+   hostname at both the Listener and Route.
+    1. Allow setting hostname matching at any combination of lister, parent
```
robscott, line 432:
```suggestion
    1. Allow setting hostname matching at any combination of listener, parent
```

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+Gateway API community.
+
+There are multiple other documents, GitHub Issues and discussions that are
+directly related to this topic and influenced the design of this GEP. This
+include the following:
+
+- [Request Filtering Between Gateways and Namespaced Routes · Issue #634 · kubernetes-sigs/gateway-api (github.com)](https://github.com/kubernetes-sigs/gateway-api/issues/634)
+- [Request Filtering Between Gateways and Namespaced Routes - Google Docs](https://docs.google.com/document/d/1-0mgRRAY784OgGQ1_LCOshpLLbeAtIr4eXd0YVYK4RY/edit#heading=h.8cfxzle5tmqb)
+- [Clarify how RouteGateways would work if we supported Route->Route delegation · Issue #411 · kubernetes-sigs/gateway-api (github.com)](https://github.com/kubernetes-sigs/gateway-api/issues/411)
+- [Route inclusion/delegation security and policy implications · Issue #1042 · kubernetes-sigs/gateway-api (github.com)](https://github.com/kubernetes-sigs/gateway-api/issues/1042)
```
robscott, line 614: I think we likely need some more coverage in this GEP about how policy attachment interacts with route inclusion. Unfortunately I'm not sure what the answer is yet, but it does add another dimension of complexity here.

robscott (commented):
Thanks @Jeff-Apple! This is a well written and massive GEP. I took a first pass, but still need to think about some of the details more here. At a high level, I think we should be trying to start with a lot of guardrails around route inclusion to simplify the implementation wherever possible. For example, maybe Child Routes can only have Parent Routes as Parents, not Gateways. Maybe Child Routes can't specify Hostnames? This is a _very_ complex topic, so wherever we have an opportunity to simplify the scope, I think we should consider it.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+
+1. Support for Parent/child Route relationships is an "Exteneded" feature in the
+   Gateway API specification
+2. Add `AllowedRoutes` to [HTTPRouteRule][httpRouteRule] as optional
+    1. Default value is nil
+    2. A single `RouteRule` can contain `BackendRefs` OR `AllowedRoutes`, not both.
+    3. A parent `Route` may include `RouteRules` that use `BackendRefs` **AND**
+       `RouteRules` that use `AllowedRoutes`.
+    4. `AllowedRoutes` MUST be non-nil in a `RouteRule` to permit child `Routes`
+       to attach to that `RouteRule`.
```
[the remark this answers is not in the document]
youngnick: Yeah, I think that the complexity of allowing an explicit deny doesn't add enough - I really prefer that parent Route owners just have to explicitly _opt-in_ by requiring `AllowedRoutes` to be non-nil.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+6. A `Route` MUST NOT be a parent to a `Route` AND a child of another `Route`.
+    1. Explanation: This limitation avoids the need for specific
+       internal/terminal Route types, and the design of using a ParentRef from
+       the child to an AllowedRoutes slot on a parent allows a Route to know if
+       it is either a parent or child (or misconfigured as both) without
+       requiring the graph traversal that a design like ForwardTo may need to
+       avoid cycles.
+    2. If an implementation chooses to not enforce this limitation, that must be
+       documented along with how many tiers of parent/child route relationships
+       are allowed
```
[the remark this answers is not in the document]
youngnick: Thinking about it more, I tend to agree.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+7. A Route’s `ParentRefs` can list more than one type of parent.
```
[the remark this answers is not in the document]
youngnick: I'd put this as "all parents of a Route must be of the same Group and Kind" if we want to do this. It certainly will make some implementation things easier.

site-src/geps/gep-1058.md
youngnick: It would simplify things, yes.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+that mechanism.
+
+One advantage of allowing Routes to include groups of settings by reference is
+that, when one of those settings needs to be changed, it only has to be changed
+in one place instead of editing all of the relevant Routes. Another advantage is
+that it prevents those settings from diverging across the Routes over time.
+
+The remainder of this document covers the proposed enhancements to the Gateway API.
+
+## Patent/Child Route Relationships
```
caboteria, line 122: Parent/Child

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+## Binding Parent and Child Routes
+
+Binding a child route to a parent route refers to the process of reconciling
+(i.e. combining) the two routes. Binding is treated as a single transaction for
+each Parent/Child route pair. Even when a child route attaches to multiple
+`RouteRules` in the parent route, the binding between the two routes is still a
+single transaction.
+
+The binding transaction has two stages; reconciliation and activation. The
```
howardjohn, line 241: I am a bit confused by this.. it sounds like its describing an implementation. Can we rephrase it as a spec?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+NOTE: An Ephemeral route is not a formal Kubernetes object. It is a logical
+concept used, in this document, to help describe behavior. In some cases, the
+terms "parent" and "ephemeral" route are interchangeable depending on whether or
+not an ephemeral route already exists due to a previous, successful binding of a
+child to that parent.
+
+### Failed Binding
+
+The binding transaction MUST fail if there is A non-recoverable error during the
+binding process. A failed binding should result in no changes to the
```
howardjohn, line 266: So say I have parent route v1 and child route v1. This is all valid config ("active Ephemeral").

Now, I change the child route to 'v2'. Its now invalid. Per the spec, I am supposed to use the old "active Ephemeral". 

Do I understand right? if so, this seems un-implementable without a stateful implementation storing prior state outside of k8s?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+When a parent route's binding to a child route fails, that should not block the
+parent route's bindings with other child routes. In the same way, a child route
+failing to bind with a parent route should not block that child's bindings to
+other parent routes.
+
+## Reconciling Parent and Child Routes
+
+**A key principle guiding the design of how parent and child routes are
+combined, is that the data plane can apply the resulting route settings in a
+single processing cycle (per direction) of the network traffic.**
```
howardjohn, line 283: can we add a requirement that the parent and child routes can statically (in control plane) be reduced into a single route?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+During the reconciliation process, any conflict of configuration settings
+between the parent and child routes is considered fatal to the transaction and
+the entire Binding fails.
+
+A configuration conflict exists when the reconciliation process would result in
+an Ephemeral route configuration that would not be valid as a single, standalone
+`Route`.  This means, in concept, a user could save the generated Ephemeral
+route config to `my-foo-route.yaml`, remove the parent and child routes from the
+cluster, then load `my-foo-route.yaml` into the cluster, and the system
+would accept and deploy it as a valid config.
```
howardjohn, line 310: Not literally though, right? There is a key different in that the child route can refer to things in their own namespace but not in the parent's namespace without a ReferenceGrant?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+A config conflict may not be between config elements in the child and the
+parent. Instead, the conflict may be between config elements from two (or more)
+child routes that attach to the same `RouteRule` in the parent route. For this
+reason, when a config conflict is detected, implementations should identify the
+config elements that conflict, including the routes that are the original source
+of each element, and provide that info to users via appropriate methods of
+notification  (e.g., logs, message in GUI, etc.).
```
howardjohn, line 318: Should we prescribe a status message that should occur as a result? Talking about UI, etc seem to be outside of scope -- all metadata like this in the API today is via status. And extra layers on top should be outside of the spec and depend on status

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+NOTE: A notification of the type just described may have security implications
```
howardjohn, line 320: Can we be prescriptive on how status messages are reported to avoid this?

site-src/geps/gep-1058.md
howardjohn: +1, 3 way wildcard match is too much

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+#### Path Matching
+
+The model described above is simple to understand and apply with all types of
+matching criteria except for path matching. There are two reasons path matching
+is different.
+
+First, the user may want to match routes in the child using paths that are
+relative to the path in the parent route’s match criteria.
+
+The second reason is that there are multiple methods supported for doing path
```
howardjohn, line 359: This applies to all matches, doesn't it? Except Method

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+path to be matched. When being concatenated. the path from the child is added to
+the end of the parent’s path.
+
+### RouteFilters
+
+The filters specified in the associated sectionName of the parent route’s
+`RouteRules`, are added (logical AND) to each `RouteFilter` in the child route’s
+`RouteRules`.
+
+Configuration conflicts are determined as described in the [Configuration
```
howardjohn, line 375: What is a conflict? Is it illegal to try to add a header that is already added? To add a header in parent and remove in child? To set the same header in both? To define two different mirror policies in parent/child?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+```
+We add a new `grandparents` stanza inside the the top-level `parents` status. This stanza will be `+optional` and `omitempty`, with the following rules:
+- `grandparents` will be nil in all non-parent HTTPRoutes.
+- Once a HTTPRoute is a parent, then this stanza must also be non-nil.
+
+We need to add this inside the parents stanza because it's possible that a parent HTTPRoute may be attached to more than one parent (probably Gateways) of its own, and there may be different status between different Gateways. For example, if a parent HTTPRoute is attached to two Gateways, one of which has a wildcard hostname, whether or not the child HTTPRoute attempting to attach to the parent HTTPRoute will be accepted is dependent on the hostname setting on the Gateway Listener. An example demonstrating this is included below.
+
+The struct itself has the same `parentRef` and `conditions` fields as the `parents` struct.
+
+The `parentRef` in the `grandparents` struct must be the same as the one on the *parent* Route's status. See the example for more clarity here.
```
howardjohn, line 482: This is really really hard for me to wrap my head around.

In hierarchy we have:

- "Grandparent" (Gateway)
- Parent (HTTPRoute 1)
- Child (HTTPRoute 2)

But it is HTTPRoute 1 that has "grandparent" status added? This feels odd to me.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+For Conditions:
+- when the controller first sees the parent Route, it must add the following Condition Types to the *parent* Route:
+  - `Attached`: one or more Routes have successfully attached to this parent. False if zero Routes have attached successfully. Note that a new Reason has also been added, `PartiallyAttached` which indicates that *some* routes have tried to attach and failed.
```
howardjohn, line 486: This is the only mention of PartiallyAttached, is it meant to be included in more detail somewhere?

howardjohn, line 486: In general, it seems like we should have the same conditions as Listener (given its a Parent) - which would lean towards `Detached`. Or just having nothing there, since it doesn't really make sense to be `Detached`, and have a `attachedRoutes` like listener?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+  - `Conflicted`: one or more settings in the parent conflict with settings in one of the child routes. The message should contain more details. Routes that cause this to be true do not attach.
+  - `ResolvedRefs`: one or more Routes have failed to attach because there is a problem with their reference. (for example, they could not be found). Those routes do not attach.
+  - `NotSupported`: the controller that is reconciling this Route does not support attaching at least one of the routes that has attempted to attach, or the parent route is already a child route itself. Those routes do not attach.
```
howardjohn, line 489: Do we need a whole new condition for this? How does it differ from Conflicted?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+It must also add the same types to the *child* Route.
```
howardjohn, line 491: Also under `grandparent`? That means in two different places we are referring to the same thing as "grandparent" which doesn't seem consistent with how family relations are described and may cause confusion.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+      sectionName: store
+    controllerName: prod-gw
+    conditions:
+    - type: Attached
+      status: true
+      reason: Attached
+      message: Successfully attached to `store` section.      
+    grandparents:
+    - parentRef:
+          kind: Gateway
```
howardjohn, line 653: I am not convinced this is something the child route needs to know about. 

For example, when I bind a HTTPRoute to a Gateway, that Gateway binds to a GatewayClass and that gateway class to a controller, and that controller probably sets up Foo and Bar resources in the process. But HTTPRoute doesn't care about any of that -- it just needs to know if it attached to the Gateway or not.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+          name: prod-web-gw
+        conditions:
+        - type: Accepted
+          status: true
+          reason: Accepted
+          Message: Route successfully attached to parent and gateway.
+```
+
+This is pretty verbose, but the extra information will be very important once we
```
howardjohn, line 682: I would worry about scalability here. If we have many child routes, and change to the parent may require updating 1000s of status resources?

howardjohn: Not to mention we are placing all of those dis-aggregated routes into an aggregated status which defeats some of a scalability benefits of decoupling. For the same reasons, Gateway doesn't list all of the attached routes. At large scales it is unsustainable and may not even fit in etcd.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+spec:
+  parentRefs:
+  - name: gateway-a
+    sectionName: http
+  - name: gateway-b
+    sectionName: http
+  rules:
+    - allowedRoutes:
+        kinds: 
+          - kind: HTTPRoute
```
howardjohn, line 732: Why is this allowed?  I thought we could only route to the same type

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+    conditions:
+    - type: Accepted
+      status: true
+      reason: Accepted
+      message: Successfully accepted
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: HTTPRoute
+metadata:
+  name: http-child
```
howardjohn, line 760: why is this a child? it's parentRef is a gateway

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+          status: true
+          reason: Accepted
+          message: Successfully accepted
+      - controllerName: somevalue
+        parentRef:
+          name: gateway-b
+        conditions:
+        - type: Accepted
+          status: false
+          reason: HostnameMismatch
```
howardjohn, line 796: Listener hostname mismatch doesn't trigger an error so it seems like this shouldn't either?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+          status: true
+          reason: Accepted
+          message: Successfully accepted
+  - controllerName: nadir
+    parentRef:
+      name: http-parent
+    conditions:
+    - type: Accepted
+      status: true
+      reason: GrandparentConfigError
```
howardjohn, line 910: It seems to confusing to say "This is accepted because of GrandparentConfigError". We are essentially saying its valid because ("reason") of an error?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+
+
+## Alternatives
+
+### Extending backendRef to allow targeting a Route
+
+This idea was to allow `backendRef` to, optionally, specify a route instead of a
+service. Some of the reasons for not taking this approach include:
+
+- It would require changes to `backendRefs`. No changes to `allowedRoutes` are
```
howardjohn, line 931: I disagree. backendRef can refer to any type, including route. Quite the opposite - this proposal requires changes to all route types to add `allowedRoutes`

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+  needed to use it for this. 
+- To minimize the changes to `backendRef`, it would require the owner of the
```
howardjohn, line 933: Not necessarily - we could allow abstractions such as just selecting the namespace and leaving name blank, which would be functionally equivalent to allowNamespaces.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+  parent route to know the exact name of the child route. If the two routes are
+  in different namespaces, that might require cross-organization coordination
```
howardjohn, line 935: This is already needed though, just in one direction.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+  between administrators.
+- It would result in there being one way to attach routes to listeners and a
```
howardjohn, line 937: Maybe from an implementation perspective. From a user perspective a route can be seen as a backend -- its logically going through both of the routes, they just happen to be merged together as part of the implementation details (of some implementations! some may not actually do this at all).

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+  namespace: prod-gw
+spec:
+  parentRefs:
+    name:  prod-web-gw
+  hostname: www.example.com
+  rules:
+  ...
+    - name: store
+...
+      allowedRoutes:
```
howardjohn, line 161: This seems extremely complex for what we are doing. First, I would assume the majority of use cases are to a single namespace -- maybe even go as far as to say we should only allow a single namespace, as allowing multiple is problematic from an ownership perspective. We also don't need `namespaces` nesting, since `kinds` is not relevant here since we only allow same-kind.

What that leaves us with is the need to refer to a single namespace. This means everything else is boilerplate, really. We could consider simplifying this entire to something like `routeNamespace: store-ns` even.

I do prefer using `backendRef` anyways ,but if we are going with this approach I think it could be dramatically simplified

howardjohn (commented):
Largely copied over comments from the original doc: https://docs.google.com/document/d/1iumYQQMaB9y-PCtMsRwVfzKlZXZDejr_fO0krXn3F90/edit

costinm:
Not sure policy attachment is well defined - if a policy is attached to a parent, does it apply to child ? Or the other way ? What happens if policies conflict as result of merging ?

And more important - what are the security implications, child seems to have a lot of power influencing parent routing.

site-src/geps/gep-1058.md
robscott: Yeah as I've been thinking about this more it seems simpler and safer to limit this capability to a single namespace and replace this with a single string instead of the struct. I think I'd originally argued for consistency with Gateway -> Route, but that's very much overkill here.

site-src/geps/gep-1058.md
robscott: I think what we've done elsewhere in the spec is probably relevant here - we should probably suggest that in the event of a conflict we detach/break immediately. We can still leave rooms for stateful implementations to provide a more graceful fallback if possible, but I don't think that's broadly implementable.

site-src/geps/gep-1058.md
robscott: This is a good question, maybe we should start with one of the following limitations:

- Filters can not be specified on child routes
- Filters can not be specified on parent routes
- Filters of the same type can only be specified in child or parent route but never both

site-src/geps/gep-1058.md
robscott: I think as a Route owner it would be very useful to know where my Route is exposed without having to traverse a tree. With that said, I recognize that this could become complex to implement and can be added separately from the rest of the delegation API. Maybe this particular change (grandparents in status) could be pulled off into another dependent GEP so it could be reviewed independently if necessary?

site-src/geps/gep-1058.md
robscott: Another way to look at this would be that route delegation may only make sense for a subset of routes, maybe just HTTPRoute to start. I personally don't like using backendRef for this concept because another route is not really a backend, it's just additional routing configuration. 

Maybe more importantly, the separation of concerns here seems like a critical part of the design. I want to be able to say that I trust Route admins in the foo namespace to handle routing for anything under `/foo`. I shouldn't have to care about what the specific Route(s) are called in that namespace.

site-src/geps/gep-1058.md
robscott: I think it's much more reasonable to require a child to know details of the parent resource than the other way around. As an owner of a parent route, I don't want to have to worry about what Route(s) are named in namespace foo, just that I trust Route owners in that namespace to handle requests under `/foo`.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+        1. A child route “attaches" to the parent route’s `RouteRule` that matches
+           that `SectionName`. The child route must also match the criteria
+           specified in the `RouteRule’s AllowedRoutes`.
+    4. If the `ParentRefs` of a child route doesn’t include a `SectionName`, the
+       child route attaches to all of the Parent Route’s `RouteRules` having
+       `AllowedRoutes` criteria that the child route meets.
+5. A parent `Route` and any related child `Routes` must be the same type of
+   `Routes`.
+    1. Example: A child httpRoute can not point to a parent tcpRoute.
+6. A `Route` MUST NOT be a parent to a `Route` AND a child of another `Route`.
```
bowei, line 219: Can we add:

Specifically 

1. A ParentRef refers to a Route Kind, no rules can have a non-nil AllowedRoutes
2. Similarly, a Route with non-nil AllowRoutes cannot have ParentRef to another Route.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+    1. Explanation: This limitation avoids the need for specific
+       internal/terminal Route types, and the design of using a ParentRef from
+       the child to an AllowedRoutes slot on a parent allows a Route to know if
+       it is either a parent or child (or misconfigured as both) without
+       requiring the graph traversal that a design like ForwardTo may need to
+       avoid cycles.
+    2. If an implementation chooses to not enforce this limitation, that must be
```
bowei, line 226: Do we need to have this carve for the time being?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+       documented along with how many tiers of parent/child route relationships
+       are allowed
+7. A Route’s `ParentRefs` can list more than one type of parent.
+    1. Example: A `HTTPRoute`’s ParentRefs can list one (or more) `Gateways` and
+       one (or more) `HTTPRoutes` at the same time.
+
+## Binding Parent and Child Routes
+
+Binding a child route to a parent route refers to the process of reconciling
+(i.e. combining) the two routes. Binding is treated as a single transaction for
```
bowei, line 236: I think we should refer to the Gateway/ Route binding.

We should also (can be separate subsection) comparing where these are same/different.

site-src/geps/gep-1058.md
bowei, line 241: As a declarative API, we need to make sure that states we are talking about is represented explicitly as transitions in the stored state/status of the resource. I think this section will be the most challenging to get right.

When we talk about a sequential ordering, we need to note:

It doesn't sound like we are saying there is a specific ordering in which children are presented to be bound, but there is some serialized order which the control performs the evaluation (e.g. like lock acquisition between competing threads).

To make sure I understand this, is this something like:

* `P`: parent route
* `BC`: set of bound children to parent (i.e. children that have been successfully bound)
* `UC`: set of yet-unbound children (i.e. children that the controller is looking at that are not bound).

Binding is the event that takes a unbound child and successfully binds it to the parent. This is an atomic decision -- looking at the current configuration generated (ephemeral route) by looking at `P + BC`, adding the incremental route `r` from `UC` -- if it the config is valid, then `r` is bound (i.e. a member of `BC`).

As part of the transition, the status of `r` and parent will be updated to reflect the successful binding.

Note: we may want to make the notion of ephemeral route concrete -- especially if we require that all route inclusion is a rewrite to something that is representable as a single route. In this case, it's actually an entirely valid resource.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+- **Delegation:** when an included route crosses a persona ownership boundary
+(e.g. a parent route is managed by an infrastructure team and a child route is
+managed by an app dev team)
+- **Inclusion:** When a child Route’s configuration is combined with a parent
+Route’s config during runtime. The two routes may or may not be owned by the same
+persona or organization. All delegations are inclusions, but not all inclusions
+are delegations.  
+
+## Introduction
+
```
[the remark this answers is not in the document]
bowei: I would recommend that we can work on this as a library that can be included in every implementation (if they desire). It will resolve a lot of ability to implement issues wrt to the spec.

site-src/geps/gep-1058.md
youngnick: It's a Reason, not a Condition, so I think we only ended up mentioning it here. I agree that this section could use more fleshing out though.

youngnick: I should note that we've been talking about moving the `Detached` Condition to a positive-polarity `Attached` condition instead, to make this easier to understand. I think @mikemorris logged the issue?

I may have been jumping ahead in this bit.

site-src/geps/gep-1058.md
youngnick: I thought that having a separate condition to indicate that "this isn't supported" was better than overloading "Conflicted" (which I read to mean "this is a supported action, but some detail of it is incorrect").

site-src/geps/gep-1058.md
youngnick: Agreed, this needs more work.

site-src/geps/gep-1058.md
youngnick: Yes, I think the concept needs work, and it's probably worth implementing the rest and seeing what would be useful to have in each status.

The overall design goal here is for the owner of the child Route not to need to also fetch the parent Route and the Gateway to determine if their thing is being exposed correctly.

site-src/geps/gep-1058.md
youngnick: Again, the idea here is to avoid the owner of the child Route needing to do a tree traversal to know the state of their child route. I can see the fanout concerns, but was hoping that by having the status refer back _up_ the tree, rather than down, we should avoid some of those problems.

site-src/geps/gep-1058.md
youngnick: This is from an earlier version that allowed other Kinds, and hasn't been updated.

site-src/geps/gep-1058.md
youngnick: Yeah, I agree this is confusing, and needs to be updated.

site-src/geps/gep-1058.md
youngnick: The other thing I don't think we mentioned here is that I think that the backendRef implies an extra routing hop, which I don't think will be common. This is a way to break up the config, not a way to represent extra routing hops.

site-src/geps/gep-1058.md
youngnick: Yes, I think that there's absolutely an action here to do some definition about how policy attachment interacts with route inclusion.

youngnick (commented):
Big review from me, sorry.

I'd summarize this as:
- we need another pass over the status design, especially keeping in mind some of the changes we've made to what counts as "Attached".
- we need to define the interaction between this and Policy Attachment.

Before we do all of that though, I think that there seems to be some reasonably strong objections to this whole concept - it seems like the value proposition here is not clear, and I think that we may need to do more to explain:
- the circumstances under which an implementation would support this
- that it's completely valid for an implementation that already has some of these capabilities to not support it (in particular, I think that Service Mesh implementations have many similar functions as part of their mesh).

costinm:
We were discussing the mesh use cases in GAMMA - in particular how
HttpRoutes may be attached to a
Service ( or something similar ).

What happens if a user has both a mesh and gateway, and they have Gateway
forward to a backend Service,
which in turns has HttpRoutes attached ? One way or another - we need to
address this, since it'll be a common
case for users of both APIs. Is it treated as a case of 'inclusion or
delegation' ? What if the impl of Gateway and
mesh is different ? If not - wouldn't it be pretty confusing for users ?

My proposal to do delegation with the equivalent semantics of having 2
chained gateways ( even if it
is implemented in one physical gateway ) holds for this mode - if Gateway
is also a Mesh implementation
it may optimize, else users may have 2 proxies. But either way - the model
for gateway delegation
will be the same with the model for using gateway and mesh routes.

Keeping the flexibility of having 2 layers of proxies as needed is also
useful when policies like
WASM or other extensions are attached - the implementation may choose to
not allow the 'child'
owner to inject risky code or policies into the trusted top level gateway,
and instantiate a dedicated
proxy for childrens that have such policies.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,623 @@
...
+There are two primary use cases that this proposal is intended to address. There
+are also a number of secondary use cases that are a combination of the two
+primary ones. The twp primary use cases are described below.
+
+The diagram below is an example scenario that is referred to
+throughput this document and the example configurations are based on it.
+
+**Example Scenario 1:**  
+![Example Scenario 1](images/1058-example-scenario.png "Example Scenario 1")
```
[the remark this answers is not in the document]
ramaraochavali: /s/throughput/throughout

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+ways this technique can be used. In the first case, traffic needs to be routed
+to "Core Services" (`core-svces-ns`) from two different starting points. Using
+route inclusion, the routing configuration for the Core Services are put in a
+child route and then that child route is included in two sections of a parent
+route where they are needed.
+
+Another way route inclusion is used in Scenario 1 is by configuring a
+HTTPRouteFilter, that inserts a header, in the parent route. This results in the
+filter being added to each RouteRule in the child route `store-route`. The child
+route is both shorter and easier to read than it would have been without this
```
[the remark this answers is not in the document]
ramaraochavali: > The important part of having the resources be in separate namespaces is not that they're shorter, it's that they can be owned by different people, and those people can coordinate without having to coordinate edits of a single resource.

I agree. There are different personas that delegation is used for.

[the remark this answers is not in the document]
howardjohn: @youngnick I 100% agree with that use case and I think is a good motivation for this GEP. I think that is "Partial Administrative Delegation of Routing" though - you are delegating to different roles. This section ("Compose routes from reusable config fragments") seems to be suggesting it should be used as just a way to DRY config within the same ownership which I don't agree is a good idea.

The design may make this usage _possible_ (I think we would have to go out of our way to stop it, if we wanted, by not allowing same-namespace delegation, which seems odd), but I don't think we should list it as a use case or encourage it.

Totally possible I am missing more context on why this is useful though.

site-src/geps/gep-1058.md
howardjohn: I think so. Might be nice to list it explicitly. It was a bit unclear to me since 'ephemeral route' is describing an implementation. This may be overly pedantic though

site-src/geps/gep-1058.md
howardjohn: That definitely seems like a valid use case. It actually brings up a bit of a gap in Gateway<->Route. Can a route succesfully bind to a Gateway, while the Gateway itself is not Ready? It seems like possibly that should be indicated (strawman: `Attached=false, Reason=parentNotReady` or something). A similar concept could be used here.

The key difference I am trying to make here is its a ~boolean - we just say the parent is ready or not, not all the details. This ties in to the fanout concerns mentioned below as well

site-src/geps/gep-1058.md
bowei: If that is the case, do we need to update the proposal? The current language (as I understand it), talks about a conceptually sequential algorithm for resolving conflicts. Which is not necessarily a bad approach, just pointing out we need to make sure it is well-defined.

site-src/geps/gep-1058.md
howardjohn: For migration, could you do this:

Pre:  `routeNamespace: before`, route defined in `before`
Transition:  `routeNamespace: before`, route defined in `before` and the same route defined in `after`
Switch over:  `routeNamespace: after`, route defined in `before` and the same route defined in `after`
Complete:  `routeNamespace: after`, route defined in `after`

?

It seems mostly the same flow, except with multiple namespaces you have both child routes 'active' at one time and are relying on conflict resolution to ensure only one is picked at time? 

I am not confident in ^, LMK if that makes sense...

site-src/geps/gep-1058.md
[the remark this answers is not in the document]
costinm: This is very important - 'different personas' also means different roles and privileges. The Gateway admin has control over the host DNS and certificates, and can decide which namespaces to trust and for what hostname and path prefix, and can set policies for the host.

Maintaining security and separation of roles should be a clear requirement for any delegation proposal. My concern with the complexity of this proposal is in large part because it makes the security properties hard to understand - which will lead to CVEs.

site-src/geps/gep-1058.md
costinm: An implementation may choose to use a controller to combine the routes and generated 'flat' routes - instead of implementing this at 'runtime'. Or may use tools in CI/CD to combine the routes and deploy flat routes. 

I agree a spec should not require or describe a particular implementation.

site-src/geps/gep-1058.md
costinm: There is another thread on 'roles' and permissions: I don't think a bad 'child' route in a delegated namespace should ever break or invalidate a 'parent' route, no matter when the parent is added. 

The first level of routes should be validated independently and be applied first, even if it conflicts with any child route. 

I would also require that child route validation and application is isolated at namespace boundary - I still don't know how delegation a /prefix to multiple namespaces works, but an error in one namespace should not impact a different one or prevent it from adding routes.

site-src/geps/gep-1058.md
costinm: I would extend this to: the requirement should not conflict with an implementation that reduces routes statically in an external CI/CD system, allowing the feature to work with Gateway implementations that opt out from implementing this in the control plane. 

That means the entire 'ephemeral' stuff should not be part of the spec, and the rules for resolving conflict should 
be deterministic and stateless.

site-src/geps/gep-1058.md
costinm: Agree with the last point - and we should match whatever GAMMA does. 

For example, if we go with 'attach routes to Services' - then the backend remains a Service ( no change ), and we can delegate to the HttpRoute attached to the service in the child namespace. 

If we add a new CRD to represent a virtual service - this can be the backendRef. If mesh would treat this virtual service with attached routes as a backend - so should the gateway spec. 

Service and whatever resource GAMMA defines will be named entities - you shouldn't care how the HttpRoute is named, but the name of the Service or virtual Service is reflected in DNS and exposed to users.

A nice side-effect is that the 'child' routes could be directly used from the mesh - don't know if we want to allow this, see my comment on interactions between gateways and mesh.

site-src/geps/gep-1058.md
costinm: +1, or we can have the backendRef match whatever we define in GAMMA. 

Leaving it blank is a bit problematic because in a namespace you may have children of multiple gateways. For example /admin may be delegated by foo.com and bar.com.  I don't think requiring hostname in child routes is ideal - and also not sufficient, I think it is possible for 2 different gateways in 2 namespaces to define 'foo.com' with different IPs - and still delegate.

site-src/geps/gep-1058.md
costinm: I think an important use case is to allow helm charts to be written and independent of cluster config.

For example in servlets web.xml routes are independent of hostname ( and even context prefix ). This is a very useful property and allows a .war to be deployed in any server without changes. 

Not sure if 'parent' knowing the naming details in the child namespace is true - the BackendRef has a name and namespace, and the name is quite significant and required and must be the exact Service name. Why would delegating
to a route be different ?

site-src/geps/gep-1058.md
costinm: Why would backendRef 'imply' an extra routing hop ? It allows implementations to use an extra hop, if strong isolation between parent/child is required - or to flatten the route. 

Technically Service is not a real backend and does require CNI to route requests to a Pod, and the spec allows any CRD
to be used as backend - there is no mention of 'hops' in the spec.

site-src/geps/gep-1058.md
costinm: That's why we have a list of BackendRefs and weights. 

To migrate you add the second namespace with weight 0, and gradually shift. The parent has full control, and you can't accidentally have a random namespace using same labels mess up, or complicated conflict resolution between namespaces.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+- **Ephemeral Route:** The route configuration that results from combining (i.e.
+ reconciling) a parent route and the child route(s) that attaches to it. The
+ ephemeral route represents the configuration that is actually applied to the
+ data plane.
+- **Delegation:** when an included route crosses a persona ownership boundary
+(e.g. a parent route is managed by an infrastructure team and a child route is
+managed by an app dev team)
+- **Inclusion:** When a child Route’s configuration is combined with a parent
+Route’s config during runtime. The two routes may or may not be owned by the same
+persona or organization. All delegations are inclusions, but not all inclusions
```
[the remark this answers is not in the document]
costinm: Also defining requirements for both inclusion and delegation, and maybe having separate docs since I don't think one solution can satisfy both.

I don't think it is safe for inclusion to cross namespace boundaries - while delegation main purpose is for crossing. 
The biggest problem in this spec is mixing the 2 and failing to address either problem in a simple way.

costinm:
> Before we do all of that though, I think that there seems to be some reasonably strong objections to this whole concept - it seems like the value proposition here is not clear, and I think that we may need to do more to explain:

I don't think I've seen any comment objecting to the concept or value of delegation - quite the opposite. 

We may disagree about the requirements of how delegation should work ( in particular around security,policies, isolation), and on the spec imposing a specific implementation, or the complexity.

caboteria (changes requested):
Hi! I have a few specific comments and I guess that I need another one to make them active.

caboteria:
Looked at this over the weekend and the biggest problem that I had was from overloading the HTTPRoute object to be two different things depending on the context. This will likely be an ongoing problem, e.g. "is this HTTPRoute a parent route or a child route?", "should this HTTPRoute have done something or is it waiting for a child?" etc. The ephemeral route idea is an attempt to solve this problem but for me it just adds complexity.

I'd like to suggest an experiment: create a new object called "HTTPRouteTemplate" which would be what we're now calling a "parent route." It would attach to Gateways and would be attached to by HTTPRoutes. It would be a template for attaching HTTPRoutes and it would constrain its attached HTTPRoutes, just like parent routes do today. In other words, this is mostly the same functionality, just with clearer nomenclature.

This would eliminate a lot of confusion about what HTTPRoute is supposed to do, when, and in which context. Users would no longer need to try to figure out whether a given HTTPRoute was a parent or a child based on context. It no longer needs the ephemeral route idea: HTTPRouteTemplate does what it does, and HTTPRoute does what it does, and they're two different things. To me this is basically the same functionality as the current proposal but I find it much easier to understand if we give up the "grandparent-parent-child" idea.

robscott:
I think I like @caboteria's idea of a new resource focused entirely on HTTP delegation. If we kept it very limited in scope, it could end up simplifying this approach and resolving a number of the comments above. Each new resource we add comes with a significant cost, so we need to be careful here, but if it looks like delegation would only support a tiny subset of the overall HTTPRoute functionality it could be quite compelling.

I've been thinking about how we could move forward with this GEP as a whole, and I think one of the most important steps will be to outline how each capability of HTTPRoute today would or would not be supported by delegation. This seems like it could also help determine if a new delegation focused resource makes sense or if we really need most of the capabilities of an HTTPRoute for both parent and child resources.

Beyond that, here are some ideas that could potentially help get this GEP closer to merging. Some of these may not be relevant if we end up going further with @caboteria's suggestion above for a new resource to model this.

1. Move expansion of status to later GEP, keep initial GEP focused on new reasons and conditions within existing structure
2. Clearly state in the GEP that route delegation must be statically computable
3. Simplify API by replacing "AllowedRoutes" struct with "DelegateToNamespace" string (name is not important here)
4. Merge GEP as "Provisional" with the requirement that an OSS library that can support the requirements above can be developed within Gateway API before GEP graduates to "Implementable".

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+## TLDR
+
+Add support for a 2-tier hierarchy of route configuration to allow partial
+delegation of route settings and, provide more flexibility in how route
+configurations are organized and managed.
+
+## Goals
+
+- Provide a 2-tier hierarchy of route configuration to allow partial delegation
+of route settings.
```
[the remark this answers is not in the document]
Jeff-Apple: I agree and will add a statement to make it clear.

site-src/geps/gep-1058.md
hzxuzhonghu, line 161: What will happen if a delegation Route rule created later than the root rule?

hzxuzhonghu: Even a single namespace allowed, there will be cases, multiple delegate rules referenced, how can we process it?

In istio, we allow specifying the delegate by namespace/name,  there can be only one delegate for a root rule.

costinm:
I also like the idea of having a separate CRD for the parent route. In the spirit of bike shedding, I would call it HttpRouteDelegate or RouteDelegation - 'template' has a slightly different use.

That will make the use extremely clear and simple - and fits the 'policy attachment' model, it is a policy set on the gateway defining what routes can be attached by which namespace. We can start with what is likely the most common use cases - gateway admin delegating by URL prefix, etc. Would be great to have a list of the use cases and evaluate which are common enough.

Since you suggested Template - I can see the benefit of allowing an actual template - like /${namespace}/ - in the delegation policy.

As mentioned, keeping the HttpRoute ( which will be the most frequently used CR ) clean and simple and having the delegation
as a separate policy is better for users.

Route 'inclusion' or some actual templating may also be a useful - but it would be ideal to separate delegation (which has a clear security and role separation purpose) from templating and inclusion.

site-src/geps/gep-1058.md
mikemorris: Refs https://github.com/kubernetes-sigs/gateway-api/issues/1110 for flipping `Detached` -> `Attached`

For Listener though, as I understand it the `Detached` condition currently represents its attachment to Gateway it's defined on (e.g. no port/protocol conflict, etc), not the status of any routes binding to the Listener.

I think `attachedRoutes` would be a better fit here, and deferring to a status condition (`Accepted` or possibly `Detached`/`Attached`) on each child route to show whether it has successfully bound to a parent route would fit better with the child -> parent reference pattern more common throughout the rest of Gateway API currently.

site-src/geps/gep-1058.md
mikemorris: This will be relevant for GAMMA too, e.g. thinking about how a service owner defining timeout policy on a mesh HTTPRoute for a service interacts with timeout defaults or override set by a gateway administrator. I think it might actually be relatively straightforward, just an additional level deep but with the same default/override pattern as already established.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,962 @@
...
+4. Add `Name` (type is [SectionName][sectionName]) to
+   [HTTPRouteRule][httpRouteRule] as optional
+    1. The `Name` field is expected to be added in the implementation of
+       [GEP-995 - Add Name to HTTPRouteRule and HTTPRouteMatch][gep-995]
+    2. Default value is nil
+    3. A child route may include a `SectionName` when its `ParentRef` is a route.
+        1. A child route “attaches" to the parent route’s `RouteRule` that matches
+           that `SectionName`. The child route must also match the criteria
+           specified in the `RouteRule’s AllowedRoutes`.
+    4. If the `ParentRefs` of a child route doesn’t include a `SectionName`, the
```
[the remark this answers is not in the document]
mikemorris: I _think_ this is the same logic as HTTPRoutes binding to Gateway Listeners without a `SectionName` - keeping this behavior consistent feels reasonable to me to simple cases? Not opposed to adding a restriction though if it helps encourage clarity.

site-src/geps/gep-1058.md
mikemorris: My original intent in proposing [`AllowedRoutes`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.AllowedRoutes) was to offer config parity between how Gateways allow routes to bind and how HTTPRoute could similarly allow child routes (as @robscott mentioned above).

**With the added same-kind constraint, we can definitely skip at least the first level.** (Even if we wanted to consider extensibility for e.g. TCPRoute -> HTTPRoute delegation in a future iteration, I don't think _additionally_ adding `AllowedRoutes` as a peer at that time would be a breaking change or backwards-incompatible, and this does reduce boilerplate for the simpler case.)

Using the [`RouteNamespaces`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.RouteNamespaces) struct directly (which we may want to rename to `AllowedNamespaces` for clarity in the standalone use case?) would support both the single-namespace

**For the simplest case, could we add an additional enum variant `"Namespace"` to [`FromNamespaces`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.FromNamespaces)?** Following that, we could add an additional optional field `name` to `RouteNamespaces`, with similar constraints as `selector`:
> `name` string _(Optional)_
Name must be specified when From is set to “Namespace”. In that case, only Routes in Namespaces matching this Name will be selected by this Gateway (or Route). This field is ignored for other values of “From”.

This would have the benefit of keeping our struct types limited _and_ additionally improving the config UX for simple cases of regular Gateway -> Route ownership delegation.

> Even a single namespace allowed, there will be cases, multiple delegate rules referenced, how can we process it?
>
> In istio, we allow specifying the delegate by namespace/name, there can be only one delegate for a root rule.

I think @hzxuzhonghu's point is valid, even for the single-namespace-only case, and we should offer prescriptive guidance/conformance tests for how this should be handled. I think specifying `name` in a delegation slot leans too much toward requiring coordination across the delegation boundary though, and doesn't offer any meaningful additional constraint (team in delegated namespace could rename any route to match, it's just annoying).

site-src/geps/gep-1058.md
mikemorris: Wouldn't this be parallel/identical to the `NoMatchingListenerHostname` [RouteConditionReason](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteConditionReason)?

site-src/geps/gep-1058.md
mikemorris: I think adding an `Attached` RouteConditionType is actually a _great_ fit for this - the similar `ListenerConditionType` already describes almost this exact scenario, including differentiation from `Accepted` (rejected from _immediate_ parentRef target due to invalid syntax/semantics, such as trying to delegate to a TCPRoute):
> This condition indicates that, even though the listener is syntactically and semantically valid, the controller is not able to configure it on the underlying Gateway infrastructure.

site-src/geps/gep-1058.md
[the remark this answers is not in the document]
mikemorris: This could have some bearing on GAMMA, but I think it's manageable as long as we're careful/explicit in how routes will be composited together.

The case I'm thinking of is defining a canary HTTPRoute on the mesh for a service (which would have ParentRef mesh or service). I _think_ a viable way for this to composite safely, with expected behavior, would be for a supporting mesh-integrated Gateway that specifies an HTTPRoute with the mesh service as a backendRef to bascially parse the mesh HTTPRoute as a child route rather than expecting that route to explicitly bind to the Gateway too.

shaneutt:
Talked about this in the sync: not considering a blocker for v0.6.0

Issue #1077, opened 2022-03-28
Update HTTPRoute and other status definitions and documentation


youngnick:
**What would you like to be added**:

In implementing HTTPRoute and Gateway, implementors have found things that we've missed or that are not clear in the definitions of how status should be updated.

There are some condition types that need to be added, some edge cases that need to be addressed, and so on.

This issue covers doing a pass over this area before we finalize v1beta1.

This top-level comment will be updated with a checklist of tasks as we find them.

**Why this is needed**:
Having a clear explanation of the status flow is vital to making sure the user experience for Gateway API is preserved.

youngnick:
Candidates for tasks so far:
- Add a `Ready` condition type for HTTPRoute, with explanation of what Ready means. This has conformance test implications.
- Add a `NotImplemented` or `NotSupported` condition type for all resources so that there's another place to hold more information about why something was not accepted, with explanations of what that means.

youngnick:
Need to fold feedback from #935 in here as well.

evankanderson:
There's also feedback in https://github.com/kubernetes-sigs/gateway-api/pull/1080#issuecomment-1082412501 (sorry @nathancoleman for diverting your PR).

I'd also like to specify some update semantics of the Gateway API -- in particular, it would be nice if Gateway API consumers could update from one "working" (not serving 500s) configuration to another without serving 500s in between. IIRC, at least some Envoy-based implementations may update the `RouteConfiguration` before updating the `Cluster` configuration, leading to routes which point to not-yet defined clusters. It would be nice if the Gateway API and conformance tests could make it clear that this is not a correct implementation. (We work around this in Knative by creating a dummy route with the next set of Clusters, and probe that route until it has live backends, but it would be better if this could be handled by the Gateway implementation).

cc @dprotaso @carlisia

youngnick:
We discussed this in our community meeting today and agreed that:
- Many implementations (not just Envoy-based ones) can't guarantee that the configuration requested has made it to the entire data plane, so a strict `Ready` condition (as @evankanderson reasonably asked for on #1080) is not possible for them.
- Because of this, we can't mandate a `Ready` condition, because people will bring the expectation of strict behavior because of the name.
- We're going to work on another condition type (proposed names: `Configured`, `Programmed` or similar, there was the usual naming discussion, we'll need to ~bikeshed~ discuss it some more. :) ) that indicates that the the controller has finished working and that the data plane should be ready "soon".
[...]

Issue #1062, opened 2022-03-22
L4 Gateway Sharing


markmc:
The Gateway vs Route separation allows a cluster operator gateway owner to share a gateway and its resources with many application developers.

However, in the L4 routing model (i.e. TCPRoute and UDPRoute), routes attaching to a shared gateway will generally be associated with a dedicated gateway listener, thereby preventing an application owner from creating a route without some cooperation from the gateway owner.

However, the gateway may represent a resource that is desirable to share - for example, a cloud load-balancer, or even a public IP address - so it seems important that we allow sharing.

Sharing a L4 forwarding gateway would require that a cluster operator could define a gateway from which independent application owners could be assigned listener TCP/UDP ports, without any risk of other application owners interfering with or conflicting with their ports.

Once a route has attached to such a listener, the application owner would retrieve the assigned port from status, and share this port and the gateway address with clients that wish to connect.

One way of thinking of this is in the context of the "the goal of using Gateway to fill the role that is currently filled by Services of type LoadBalancer" - this would allow independent application developers to have `type=LoadBalancer` behavior without requiring expensive resources (load-balancers, public IP addresses, etc.) to be allocated to each.

For a longer discussion, see [Gateway API - Transport Layer (L4) Routing vs Gateway Sharing](https://docs.google.com/document/d/1KGqdpXUPxTxvyOUih4ggmKBI5d8ePMF3_YLIvh28LfI)

#818 is also relevant because the proposed modelling in the doc involves a listener having a port range to allocate from:

> If a Listener could be associated with a range of ports, the Gateway controller could be responsible for assigning ports from the range to routes which match that Listener.
>
> In Gateway.status, the attachment status of each port would need to be represented somehow. One way is to add a ListenerStatus for each port, or perhaps only for each allocated port. In this way, listeners with port ranges behave like a Listener template from which per-port Listeners are generated.
>
> In TCPRoute.status, sufficient information should be provided to determine which generated Listener the route attached to, so the application owner can determine which port was allocated. We could make use of the recently added ParentReference.Port field to expose this.

See also [this work-in-progress branch](https://github.com/kubernetes-sigs/gateway-api/compare/master...markmc:tcp-gateway-sharing) showing the API changes and [this work-in-progress Istio branch](https://github.com/istio/istio/compare/master...markmc:tcp-gateway-sharing) showing an Istio-based proof-of-concept.

youngnick:
[...]

Issue #1061, opened 2022-03-22
Make listener port optional


markmc:
There are multiple cases where it may make sense to not require a gateway owner to specify a listener port:

1. Where a port number is irrelevant or nonsensical - #1052 is an example AIUI
2. Where the port is well-known and can be derived from the listener protocol - e.g. where `protocol==HTTP` default to port 80 if unspecified
3. Where a port should be auto-allocated for this listener - e.g. similar to auto-allocating an ephemeral port if you call `bind()` with the port specified as 0

One proposal is to specify that an unspecified port should be defined to have per-protocol semantics - so e.g. we'd define defaults for HTTP and HTTPS, but TLS, TCP, and UDP would be auto-allocated a random port. Implementation-specific protocols would define their own default behaviors.

Another alternative would be to use `port==0` for auto-allocating a random port, and only have defaults for protocols where one makes sense.

See also #1060, #818, and [Gateway API: Multi Port Approach](https://docs.google.com/document/d/1NN-IDoN-mIxtv6f7JR_G9kK1kKbZHMqQPcUSy6tPDbA)

markmc:
#1066 adds docs for implementors about the possibility of required fields become optional in future:

> Handle fields that have transitioned from required to optional without crashing

hbagdi:
Gateway resource consists of two parts:
1. Network configuration that is closely tied to the underlying infrastructure. Ports, IPs and hostnames are good example of this category.
2. Routing configuration such as TLS, route selectors, etc.

Of these, most/all use-cases consider (2) to be the source of truth and a Gateway should _lead_ and not _follow_.

(1) is more complicated.
In most of our designs up until this point, we have focused on Gateway resource "representing" a network endpoint.
Up until now, we have assumed that such a network endpoint exists or is managed out of band, and the Gateway resource reflects the state of that resource and the API has no semantics around that.

There are two class large classes of use-cases
- Treat networking configuration such as ports in the Gateway as the source of truth. Reconcile the rest of the infrastructure to this. This is similar to throw Service resource works and has been the focus so far.
- Gateway represents the existing state, which is updated based on what exists in the underlying infrastructure. We have not focused much on the use-cases of these classes or waived them off with "another controller could be creating the gateway resource". I've seen some use-cases from proxy vendors, mesh vendors and L4-level use-cases that run into this problem.

I don't propose a solution here. I want to express my understanding and validate it with how others are thinking about it.
[...]

Issue #818, opened 2021-08-23
Address port range and "all ports" use cases


jpeach:
In the API review the topic of "all ports" services was raised.

https://github.com/kubernetes-sigs/gateway-api/pull/780#discussion_r693210917

And there is a KEP around this for Services:

https://github.com/kubernetes/enhancements/pull/2611

This issue is for capturing use cases for gateways that need large numbers of ports, possibly all of them.

jpeach:
@beriberikix In all ports KEP, SIP/RTP/gaming servers are mentioned. Do you know of any use cases that would need lots of Gateway ports?

beriberikix:
In my experience, they tend to have just a handful or simply one port per protocol. But paging the expert on all things WebRTC @Sean-Der!

prameshj:
A survey was sent out as part of the services KEP to better understand the usecases. The results are in tiny.cc/allportsslides which details some usecases.

youngnick:
Feels like I could summarise most of those use cases as "protocols designed before everyone wanted to fit everything through NAT/PAT".

I think the other use case to keep in mind for "all ports" is also "no port", for Routes that don't fit into the L4 bucket. I think that with a minimum of effort, the Gateway construct could be applied to things like IP routes, having the Gateway be a more general router. A Gateway object with a set of IPRoutes that indicated CIDR and next-hop could be a reasonable standard for a basic BGP config, for example.

More usefully, I think that a Gateway could also be considered as a VPN endpoint, allowing an IPRoute to specify what subnets are available behind the Listener.

In both of those cases, the "port" is not meaningful, so we could either:
- leave it unspecified
- ignore it
- have a way to say "all ports"

which are all roughly equivalent in terms of the actual use case.

beriberikix:
> Feels like I could summarise most of those use cases as "protocols designed before everyone wanted to fit everything through NAT/PAT".

I think that's accurate for the use cases mentioned (which is plenty!) but there are modern protocols that don't or can't rely on NAT, ex. telco and IoT. I won't get started on L3/L2 or even non-IP protocols that are part of 5G :) That said, besides what's covered in the survey, covering "ports are not meaningful" would suffice.

ccfishk:
> Feels like I could summarise most of those use cases as "protocols designed before everyone wanted to fit everything through NAT/PAT".
Pretty all-inclusive all use cases ! 👍 
[...]

PR #1086, opened 2022-04-04
Gateway API v0.5.0 API Review

robscott:
**What type of PR is this?**
/kind api-change

This PR represents the diff between v0.4.x and a proposed v0.5.0. This was accomplished by copying the apis/v1alpha2 from master to a branch based on the release-0.4 branch. 

### Current status
We believe GatewayClass, Gateway, and HTTPRoute have met the criteria we defined earlier for graduation to beta ([details](https://docs.google.com/document/d/1F3dcs_J8Y6GVlfMGjRtom5ljovnKa_MMIC34GJxUdBs/edit)). The following work is still in progress: 

* Adding [new conditions to HTTPRoute status](https://github.com/kubernetes-sigs/gateway-api/issues/1077#issuecomment-1081237680)
* A variety of [documentation updates and webhook deployment cleanup](https://github.com/kubernetes-sigs/gateway-api/milestone/5)

### Changes from v0.4
The proposed API is largely identical to what we released in v1alpha2, with the following exceptions:

1. We want to graduate GatewayClass, Gateway, and HTTPRoute to v1beta1 ([More Details](https://docs.google.com/document/d/1F3dcs_J8Y6GVlfMGjRtom5ljovnKa_MMIC34GJxUdBs/edit))
2. We have introduced a new "release channels" concept. The new "experimental" release channel allows us to add new features to Gateway API in a similar way to how we use alpha feature gates in upstream to guard new fields initially. ([More Details](https://deploy-preview-1066--kubernetes-sigs-gateway-api.netlify.app/concepts/versioning/#release-channels-experimental-stable))

### New experimental features
There are 3 new experimental features included in this change:

- Path redirects and rewrites [GEP 726](https://gateway-api.sigs.k8s.io/geps/gep-726/)
- Address matching on TCPRoute and UDPRoute [GEP 735](https://gateway-api.sigs.k8s.io/geps/gep-735/)
- Port matching in ParentRef [GEP 957](https://gateway-api.sigs.k8s.io/geps/gep-957/)

### Future port changes
As a follow up to the previous review, there has been a lot of discussion around Ports, and potential for port ranges, all ports, and auto assignment:

- [All Ports and Port Ranges](https://github.com/kubernetes-sigs/gateway-api/issues/818)
- [Optional Listener Port](https://github.com/kubernetes-sigs/gateway-api/issues/1061)
- [L4 Gateway Sharing](https://github.com/kubernetes-sigs/gateway-api/issues/1062)

At this point, we believe that all of the above can be added in the future with [backwards compatible changes/additions](https://docs.google.com/document/d/1NN-IDoN-mIxtv6f7JR_G9kK1kKbZHMqQPcUSy6tPDbA/edit). Note: this is not saying we should or will make these additions, just that they will still be possible after graduating to beta.

### Goals of this review

Since we're an official Kubernetes API, we need formal approval for this change from SIG-Network API Reviewers, assigned below. Of course feedback from others is also welcome here. Our goal is to complete this process by mid to late April.

Note: The aim is not to merge this PR, just to provide a way to review what has changed between v0.4 and v0.5.

/assign @danwinship @khenidak @thockin


apis/v1alpha2/httproute_types.go
```diff
@@ -704,6 +769,31 @@ type HTTPRequestRedirectFilter struct {
...
+// Support: Extended
+type HTTPURLRewriteFilter struct {
+	// Hostname is the value to be used to replace the Host header value during
+	// forwarding.
+	//
+	// Support: Extended
+	//
+	// <gateway:experimental>
+	// +optional
+	Hostname *Hostname `json:"hostname,omitempty"`
```
robscott, line 786:
```suggestion
	Hostname *PreciseHostname `json:"hostname,omitempty"`
```

apis/v1alpha2/shared_types.go
```diff
@@ -401,3 +430,86 @@ type AnnotationKey string
...
 type AnnotationValue string
+
+// AddressRouteMatches defines AddressMatch rules for inbound traffic according to
+// source and/or destination address of a packet or connection.
+type AddressRouteMatches struct {
+	// SourceAddresses indicates the originating (source) network
+	// addresses which are valid for routing traffic.
+	//
+	// Support: Extended
+	SourceAddresses []AddressMatch `json:"sourceAddresses"`
```
thockin, line 441: I'm suspicious of the need for this.  It feels like firewall behavior, but it doesn't purport to drop non-matching traffic, just to ignore it, IIUC?  That seems potentially problematic for some impls (e.g. iptables would not match and allow the packet to continue for further processing).

Can you help me understand what it is for?

apis/v1alpha2/shared_types.go
```diff
@@ -401,3 +430,86 @@ type AnnotationKey string
...
+
+	// DestinationAddresses indicates the destination network addresses
+	// which are valid for routing traffic.
+	//
+	// Support: Extended
+	DestinationAddresses []AddressMatch `json:"destinationAddresses"`
```
thockin, line 447: How is this different than listener?  Is this related to SNI?  No?

I'd like to understand how this is used - hostname isn't part of a TCP or UDP packet.  Is someone planning to do DNS lookups and filter on that?

apis/v1alpha2/shared_types.go
```diff
@@ -401,3 +430,86 @@ type AnnotationKey string
...
+	// must respect the connection metadata from proxy-protocol
+	// in the match logic implemented for these address values.
+	//
+	// Examples: `1.2.3.4`, `128::1`, `my-named-address`.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Value string `json:"value"`
```
thockin, line 482: IPv4, IPv6, hostname are unambiguously parseable (with some effort) - is this justified?

If we want to keep this, should it follow the same 1-field-per enum value pattern?

apis/v1alpha2/shared_types.go
robscott: I think there are at least two key use cases here:

1. For routing requests going through Gateway(s) that are listening on a wide set of addresses
2. For routing requests not going through Gateway(s) such as mesh implementations

I think @shaneutt and/or @howardjohn might be able to add some more detail for potential use cases here.

apis/v1alpha2/shared_types.go
robscott: Good question, it's been a while since we reviewed this, and I can't remember the specific use cases for source IP matching, maybe @shaneutt does?

apis/v1alpha2/shared_types.go
robscott: I think it's at least possible that we'd add additional types in the future that may not be as easy to differentiate between. Even with named address, it's possible that in some systems, a name is just a number, or a name is a valid hostname, or something else. I could get on board with the 1-field-per enum pattern here though.

apis/v1alpha2/shared_types.go
howardjohn: The strongest use case, IMO, is for east-west traffic that can convey a destination address. This could be from CONNECT, PROXY protocol, service mesh redirection, etc.

For example, if we want to say "All requests to httpbin should use canary logic":

```yaml
kind: HTTPRoute
metadata:
  name: httpbin-canary
spec:
  parentRefs:
  - name: my-gateway
  matches:
  - destinationAddresses:
      type: NamedAddress
      value: httpbin.default.svc.cluster.local
  rules:
  - backendRefs:
    - name: httpbin-v2
      port: 80
      weight: 1
    - name: httpbin-v1
      port: 80
      weight: 1
```

(note: HTTP doesn't actually have this field currently, but the same applies to all protocols).

Matching only on Host header is not actually sufficient, IMO. For example, consider if we are trying to configure sending traffic to an nginx proxy, which then in turn does Host header routing (that is, the application does `curl nginx.default.svc.cluster.local -H "Host: host-for-nginx"`. Matching `hostnames: [nginx.default.svc.cluster.local]` would not help here.

danwinship:
> The proposed API is largely identical to what we released in v1alpha2, with the following exceptions:
> 
> 1. We want to graduate GatewayClass, Gateway, and HTTPRoute to v1beta1 ([More Details](https://docs.google.com/document/d/1F3dcs_J8Y6GVlfMGjRtom5ljovnKa_MMIC34GJxUdBs/edit))

I see no references to "v1beta1" here... I had assumed when I read this that you'd be duplicating the "graduating" parts of `apis/v1alpha2/` into a new `apis/v1beta1/`, so that people using only beta APIs could do

    kind: Gateway
    ...

youngnick:
@danwinship I think that's the plan for actually cutting the release, but I think that @robscott felt this review would be easier if you didn't need to review files that were just all added lines (since a `v1beta1` would be net-new and not produce a useful diff.)

The plan is absolutely to take the three resources mentioned and copy them into a `v1beta1` version however.

robscott:
@danwinship Good point, I can make a separate PR with `/apis/v1beta1`, I was just trying to make it as easy as possible to see what was actually changing. The new v1beta1 types will be identical to the v1alpha2 types shown here, so I'd hoped just seeing a diff covering the current state with what you'd currently reviewed would be more manageable. Let me know if adding `/apis/v1beta1` at this point would be helpful.

apis/v1alpha2/httproute_types.go
```diff
@@ -664,7 +684,42 @@ type HTTPRequestHeaderFilter struct {
...
+	Type HTTPPathModifierType `json:"type"`
+
+	// Substitution defines the HTTP path value to substitute. An empty value
+	// ("") indicates that the portion of the path to be changed should be
+	// removed from the resulting path. For example, a request to "/foo/bar"
+	// with a prefix match of "/foo" would be modified to "/bar".
+	//
+	// <gateway:experimental>
+	// +kubebuilder:validation:MaxLength=1024
+	Substitution string `json:"substitution"`
```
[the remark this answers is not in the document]
robscott: Will be fixed in #1124

apis/v1alpha2/shared_types.go
robscott: Discussed in community meeting and it seems like we may need to rethink this a bit, more discussion in #1127.

apis/v1alpha2/shared_types.go
robscott: Discussed in community meeting and it seems like we may need to rethink this a bit, more discussion in #1127.

youngnick:
Looks like this got closed in error?

apis/v1alpha2/shared_types.go
```diff
@@ -253,12 +284,10 @@ type RouteStatus struct {
...
 // PreciseHostname is the fully qualified domain name of a network host. This matches
-// the RFC 1123 definition of a hostname with 2 notable exceptions:
-//
-// 1. IPs are not allowed.
-// 2. A hostname may not be prefixed with a wildcard label (`*.`).
+// the RFC 1123 definition of a hostname with 1 notable exception that
+// numeric IP addresses are not allowed.
 //
-// Hostname can be "precise" which is a domain name without the terminating
+// PreciseHostname can be "precise" which is a domain name without the terminating
```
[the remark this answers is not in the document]
robscott: Will be fixed in https://github.com/kubernetes-sigs/gateway-api/pull/1124

apis/v1alpha2/httproute_types.go
```diff
@@ -227,11 +227,10 @@ const (
 	// Matches based on a URL path prefix split by `/`. Matching is
 	// case sensitive and done on a path element by element basis. A
 	// path element refers to the list of labels in the path split by
-	// the `/` separator. A request is a match for path _p_ if every
-	// _p_ is an element-wise prefix of the request path.
+	// the `/` separator. When specified, a trailing `/` is ignored.
 	//
-	// For example, `/abc`, `/abc/` and `/abc/def` match the prefix
-	// `/abc`, but `/abcd` does not.
+	// For example. the paths `/abc`, `/abc/`, and `/abc/def` would all match
```
candita, line 232: nit
```suggestion
	// For example, the paths `/abc`, `/abc/`, and `/abc/def` would all match
```

danwinship:
/unassign
(I skimmed over it, but I haven't used this API at all, or thought about it since the last review, so I don't know the context of most of these changes, etc. None of it seems obviously wrong...)

apis/v1alpha2/httproute_types.go
```diff
@@ -664,7 +684,42 @@ type HTTPRequestHeaderFilter struct {
...
+	// removed from the resulting path. For example, a request to "/foo/bar"
+	// with a prefix match of "/foo" would be modified to "/bar".
+	//
+	// <gateway:experimental>
+	// +kubebuilder:validation:MaxLength=1024
+	Substitution string `json:"substitution"`
+}
+
+// HTTPRequestRedirect defines a filter that redirects a request. This filter
+// MUST not be used on the same Route rule as a HTTPURLRewrite filter.
```
candita, line 722: nit
```suggestion
// MUST NOT be used on the same Route rule as a HTTPURLRewrite filter.
```

apis/v1alpha2/httproute_types.go
```diff
@@ -704,6 +769,31 @@ type HTTPRequestRedirectFilter struct {
 	StatusCode *int `json:"statusCode,omitempty"`
 }
 
+// HTTPURLRewriteFilter defines a filter that modifies a request during
+// forwarding. At most one of these filters may be used on a Route rule. This
+// may not be used on the same Route rule as a HTTPRequestRedirect filter.
```
candita, line 774: nit - for consistency
```suggestion
// MUST NOT be used on the same Route rule as a HTTPRequestRedirect filter.
```

apis/v1alpha2/shared_types.go
```diff
@@ -84,6 +86,35 @@ type ParentRef struct {
 	//
 	// +optional
 	SectionName *SectionName `json:"sectionName,omitempty"`
+
+	// Port is the network port this Route targets. It can be interpreted
+	// differently based on the type of parent resource:
+	//
+	// * Gateway: All listeners listening on the specified port that also
```
candita, line 93:
```suggestion
	// If the parent resource is a Gateway, all listeners listening on the specified port that also
```

apis/v1alpha2/shared_types.go
```diff
@@ -84,6 +86,35 @@ type ParentRef struct {
...
+	// support this kind of Route(and select this Route). It's not
```
candita, line 94: This sentence is a little unclear.  Is it a definition of the Gateway interpretation of Port?  Or is Gateway a type of parent resource?

apis/v1alpha2/shared_types.go
```diff
@@ -401,3 +430,86 @@ type AnnotationKey string
...
+	// implementation to handle (and add validation for) according to their
+	// own needs.
+	//
+	// For IPAddress the implementor may expect either IPv4 or IPv6.
+	//
+	// Support: Core (IPAddress)
+	// Support: Custom (NamedAddress)
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=IPAddress;NamedAddress
```
candita, line 464: The Enum lacks Hostname
```suggestion
	// +kubebuilder:validation:Enum=IPAddress;NamedAddress;Hostname
```

candita:
@robscott looks good to me, as someone new to the review.  I just had a few nits and minor comments.

apis/v1alpha2/shared_types.go
robscott: This has been removed from the API while we reconsider how it could work.

apis/v1alpha2/shared_types.go
robscott: This has been removed from the API while we reconsider how it could work.

apis/v1alpha2/shared_types.go
robscott: This ended up getting pulled out entirely with the latest update.

apis/v1alpha2/shared_types.go
robscott: Good catch - will be fixed by #1160

apis/v1alpha2/httproute_types.go
robscott: Good catch - will be fixed by #1160

apis/v1alpha2/httproute_types.go
robscott: Good catch - will be fixed by #1160

apis/v1alpha2/httproute_types.go
```diff
@@ -664,7 +699,42 @@ type HTTPRequestHeaderFilter struct {
...
+
+const (
+	// This type of modifier indicates that the complete path will be replaced
+	// by the path redirect value.
+	AbsoluteHTTPPathModifier HTTPPathModifierType = "Absolute"
+
+	// This type of modifier indicates that any prefix path matches will be
+	// replaced by the substitution value. For example, a path with a prefix
+	// match of "/foo" and a ReplacePrefixMatch substitution of "/bar" will have
+	// the "/foo" prefix replaced with "/bar" in matching requests.
```
thockin, line 713: Is this a path-token match or a string match?  If the input path is `/foo/qux` it will become `/bar/qux`, yes?  What if the input was `/fooqux` ?

aojea: the ingress one is `path` token https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types , should this be the same?

apis/v1alpha2/shared_types.go
```diff
@@ -155,22 +182,59 @@ type BackendRef struct {
...
-	ConditionRouteAccepted RouteConditionType = "Accepted"
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Accepted"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	RouteConditionAccepted RouteConditionType = "Accepted"
```
aojea, line 199: naive question, "Accepted" is a Type?

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go

apis/v1alpha2/httproute_types.go

apis/v1alpha2/shared_types.go
robscott: Not sure I understand the question. We're defining it as a type along with all the corresponding reasons we recommend setting for that condition. This matches the pattern we've used elsewhere for status conditions and reasons.

apis/v1alpha2/shared_types.go

apis/v1alpha2/shared_types.go
aojea: ah, I saw below reason "accepted" and type "ResolvedRefs" so I get confused, it was curiosity , disregard it

apis/v1alpha2/shared_types.go
robscott: Good catch, this was not clearly documented in the context of rewrites and redirects, I've filed https://github.com/kubernetes-sigs/gateway-api/pull/1189 to fix that.

apis/v1alpha2/httproute_types.go
robscott: Good catch, this was not clearly documented in the context of rewrites and redirects, I've filed https://github.com/kubernetes-sigs/gateway-api/pull/1189 to fix that.

thockin (commented):
Discussed in real-time - couple small followups Rob noted.  LGTM.  Thanks.

youngnick:
So, I think that once we resolve #1200, #1201, and #1202, we can close this out as complete?

thockin:
SGTM

youngnick:
Yes, I think we may just close this out - this is really a placeholder to give the API reviewers an easier time. Unless there's something else @robscott wants to do here?

robscott:
Yep, I think this was good to go. All the changes reviewed here were released as part of v0.5.0-rc1.

PR #1088, opened 2022-04-05
Duplicate the kustomization.yaml for the stable and experimental folders

jaison-tiu:
**What type of PR is this?**
kind/feature

This PR moves the kustomization.yaml in the config/crd folder to the experimental/ and stable/ subfolders. This will allow the user the install either the stable or experimental CRDs based on their need instead of locking the user to only the stable CRDs.

N/A

```release-note
Allow the user to install either the stable or experimental set of CRDs.
```


youngnick:
This is a great idea, but may be made obsolete when we finish up #1017. I haven't finished that work yet, so that's on me.

I'll update now and let everyone else decide what we want to do here.

PR #1090, opened 2022-04-06
Supply single install YAMLs for all required config

### LINKED ISSUE #1017: Admission webhook: Provide install YAML as part of release bundle YAML

youngnick:
*What type of PR is this?**
/kind feature

Adds `install/experimental.yaml` and `install/stable.yaml` to the repo, along with tooling to update them.

Moves webhook deployment YAMLs from `deploy` to `config/webhook` since they are intended to be used to generate install YAMLs.

Adds tooling to update versions as required in webhook YAMLs.

Updates `make generate` to make sure everything is up to date. This also means that `make verify` will catch any changes in the new files as well.

Updates RELEASE.md to cover what I believe to be the current process for releasing a new version. This can definitely be optimized, but this should document what we are doing right now.

I've tried to keep changes in distinct commits to make reviewing a bit easier.

Fixes #1017

```release-note
Installing Gateway API now only requires using a single YAML file. The install instructions have been updated.
Webhook base YAMLs have been moved from `deploy/` to `config/webhook`.
```


youngnick:
This will also need to be backported to `release-0.4` if we don't want to wait for `v0.5.0` to release the changes.

RELEASE.md
```diff
@@ -19,9 +19,17 @@ documentation].
...
-- Once the above PR is merged, the author must publish a new Git tag. This can
-  be done using the `git` CLI or Github's [release][release]
-  page. This step can be performed only by [Gateway API maintainers][gateway-api-team].
+
+The following steps must be done by one of the [Gateway API maintainers][gateway-api-team]:
+- If you are making a new major or minor release, cut a `release-major.minor` branch that we
+  can tag things in as needed.
+- Check out the `release-major.minor` branch locally
+- Update `pkg/generator/main.go` with the new semver tag and any updates to the API review URL.
+- Run the following commnand `BASE_REF=vmajor.minor.patch make generate` which will update generated docs
```
markmc, line 28: Do you need to do this for every patch release too?

hack/build-install-yaml.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
```
markmc, line 3:
```suggestion
# Copyright 2022 The Kubernetes Authors.
```

hack/update-webhook-yaml.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Kubernetes Authors.
```
markmc, line 3:
```suggestion
# Copyright 2022 The Kubernetes Authors.
```

hack/update-webhook-yaml.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [[ -z "${BASE_REF-}" ]];
+then
+    echo "BASE_REF env var must be set and nonempty."
+    exit 1
+fi
+
+semver='^v[0-9]+\.[0-9]+\.[0-9]+$'
+
+if [[ "${BASE_REF}" =~ $semver ]]
```
markmc, line 29: build-and-push does the reverse - if it's not master, assumes it's semver. Would it not be easier to mirror that logic?

markmc (commented):
`hack/verify-examples-kind.sh` needs updating to use `config/webhook/` instead of `deploy/`

And more broadly, I wonder do we really want to keep this monster autogenerated convenience file in git? Could build-and-push not publish this ... somewhere?

RELEASE.md
youngnick: Yes, good point. I'll try and make this clearer.

hack/build-install-yaml.sh
youngnick: 🤦‍♂️

hack/update-webhook-yaml.sh
youngnick: build-and-push is only run by Prow, not by people, whereas this is supposed to be run by people, so I wanted to err on the side of not doing anything unless the semver was explicitly specified (to prevent accidentally updating things).

youngnick: I'll comment in the script to that effect.

youngnick:
I think I've hit most of the feedback in that round, but let me know if I missed anything.

I did have a think about other places to put the single-file YAMLs, but everything I could think of wasn't that much better than just dropping it into Git. As a plus, our existing `make verify` tooling will now ensure that the files are kept up to date for `master` as well.

RELEASE.md
robscott: Although the individual bullet points look good now, I think this still needs to be split up by subheading. So instead of everything being under "For a major or minor release", you could have ~3 sections:

- Major or minor release
- Patch release
- All release

There's not much unique for a patch release, but should at least mention that every patch release needs to originate from a release-major.minor branch.

config/webhook/admission_webhook.yaml
```diff
@@ -30,7 +30,7 @@ kind: Service
 metadata:
   labels:
     name: gateway-api-webhook-server
-    version: 0.0.1
+    version: dev
```
robscott, line 33: I'd vote to remove this since it seems likely that this will be difficult to maintain or at least easy to mess up.

config/webhook/admission_webhook.yaml
```diff
@@ -49,6 +49,7 @@ metadata:
   namespace: gateway-system
   labels:
     name: gateway-api-admission-server
+    version: dev
```
robscott, line 52: Same comment here about removing this

install/experimental.yaml
```diff
@@ -0,0 +1,4560 @@
+# Copyright 2022 The Kubernetes Authors.
```
robscott, line 1: These files should probably be included in https://github.com/kubernetes-sigs/gateway-api/blob/master/.gitattributes

robscott: Actually, I'm starting to think these files shouldn't be checked in at all. I was looking at what other projects do, and really like cert-manager's solution. They attach manifests to the GitHub release to end up with an installation command like `kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml`. They have a lot of tooling to automate this process, but even if we don't get all the automation in place, it seems like it could be a good approach for us.

robscott (commented):
Thanks for all the work on this @youngnick! Some small comments/questions, but like the direction here.

install/experimental.yaml
markmc: +1 nice find. There sure is a lot of tooling behind it though!

In the end though, the key part is attaching the yaml to the draft github release. See https://github.com/cert-manager/release/blob/cccdd578bf20a63e366563ee3bb18e9c71fdf7a1/cmd/cmrel/cmd/gcb_publish.go#L455-L459

The github CLI does support this in https://cli.github.com/manual/gh_release_create or https://cli.github.com/manual/gh_release_upload

markmc:
> I think I've hit most of the feedback in that round, but let me know if I missed anything.

:+1:

> I did have a think about other places to put the single-file YAMLs, but everything I could think of wasn't that much better than just dropping it into Git. As a plus, our existing `make verify` tooling will now ensure that the files are kept up to date for `master` as well.

`make verify` could still generate it temporarily for testing. We're only version-controlling it as an easy way of publishing it though?

youngnick:
> `make verify` could still generate it temporarily for testing. We're only version-controlling it as an easy way of publishing it though?

Yeah, that's correct. I like what Rob is suggesting about adding the files to the release, but that seems like a significant amount of extra work.

I'm out next week so probably won't have bandwidth to do that on top, what do we want to do here? I'll fix the other comments, but feels like we can either:
- merge this and use the `install` YAMLs for now
- I pull the install YAML changes out
- We don't merge this at all until I come back and we can solve the publishing problem.
-

robscott:
I think this is still a net improvement, maybe we go with the following approach to get this in:

1. Leave reorganized YAML in place
2. Remove `/install` generation from `make generate`
3. Move `/install` to `/release` and add it to `.gitignore`
4. Document that the artifacts generated in `/release` are intended to be uploaded as part of a release

I think the rest of this PR can largely stay as is, with the exception of some relatively minor doc updates.

markmc:
> I think this is still a net improvement, maybe we go with the following approach to get this in:
> 
> 1. Leave reorganized YAML in place
> 2. Remove `/install` generation from `make generate`
> 3. Move `/install` to `/release` and add it to `.gitignore`
> 4. Document that the artifacts generated in `/release` are intended to be uploaded as part of a release
> 
> I think the rest of this PR can largely stay as is, with the exception of some relatively minor doc updates.

+1 good plan

I've taken Nick's branch, cleaned up the commits a bit, made the changes above, and pushed to here: https://github.com/kubernetes-sigs/gateway-api/compare/master...markmc:single-install-yaml

I'm away next week too. Maybe Rob could pull all that together and merge it?

Note this doesn't include Nick's idea to use the single install YAMLs in `verify-examples-kind.sh'

youngnick:
I'm really sorry I didn't get this updated in time, I think that folding in Mark's changes would be great.

youngnick:
Okay, I merged in @markmc's diff, and removed the `install` dir. I *think* this should be mergable now.

youngnick:
I hadn't removed the labels as @robscott requested, that's done now.

robscott:
Thanks @youngnick and @markmc! This is a great improvement

PR #1091, opened 2022-04-06
verify-examples-kind: reverse path order for kind binary

markmc:
We're installing kind in a temporary `GOPATH`, let's make sure it gets chosen over any older version installed.

I was hitting an issue with 0.7.0 previously installed earlier in my path.

/kind bug


robscott:
Thanks @markmc!

/ok-to-test
/lgtm
/approve

PR #1093, opened 2022-04-07
conformance: add option to read manifests from disk

michaelbeaumont:

Adds the ability to run the conformance tests using manifests on disk. Useful for example if the manifests are generated by taking the embedded files and tweaking them.

**What type of PR is this?**
/kind feature

```release-note
NONE
```

### DIFF
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


robscott:
@michaelbeaumont which manifests do you need to modify? I'm hesitant to introduce skew between what it means to run conformance tests.

michaelbeaumont:
@robscott We need to annotate the namespaces and change the `Listener` ports.

robscott:
@michaelbeaumont that use case generally makes sense to me. I know this is a slippery slope, but I'd be interested in adding flags for these specific modifications - something like `--add-namespace-labels` and `--non-privileged-ports`? I'm trying to avoid a set of conformance tests that are so open that anyone can pass them just by modifying the manifests. (I know that's not what you're trying to do here, it just would open the door for making more significant modifications to the manifests).

An alternative approach would be to clearly document the types of modifications that may be made to manifests while still remaining conformant. Open to what others think here as well, just want to make sure we're not opening up the possible variations of conformance too wide.

michaelbeaumont:
@robscott Yeah, adding explicit options makes sense to me. For non-privileged ports, passing in a range that the tests would allocate listener ports from (in particular, ports need to be unique per listener) would be ideal I think. I'll open some PRs in any case.

At least for us, those are the only modifications needed to run the tests and the only ones I can imagine needing for now. I suppose that means the URL option will go away at some point?

michaelbeaumont:
Closing for https://github.com/kubernetes-sigs/gateway-api/pull/1100 and https://github.com/kubernetes-sigs/gateway-api/pull/1101

PR #1094, opened 2022-04-07
Fix conformance tests for IPV6 gateway addresses

acnodal-tc:
**What type of PR is this?**
/kind feature

The conformance tests don't work with Gateways that have IPV6 addresses. When using an IPV6 IP address in a URL it has to be wrapped in [].

```release-note
Conformance tests now work with Gateways that have IPV6 addresses.
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -17,6 +17,7 @@ limitations under the License.
 package http
 
 import (
+	"net"
 	"net/url"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


conformance/utils/http/http.go
```diff
@@ -71,7 +72,17 @@ func MakeRequestAndExpectEventuallyConsistentResponse(t *testing.T, r roundtripp
 		expected.StatusCode = 200
 	}
 
-	t.Logf("Making %s request to http://%s%s", expected.Request.Method, gwAddr, expected.Request.Path)
+	// Parse gwAddr so we can tell which family it belongs to.
```
michaelbeaumont, line 75: Maybe we could handle this centrally in `WaitForGatewayAddress`

acnodal-tc: We could, but "[2001:470:154:5::1]" isn't a valid IP address so I'd be hesitant to pass that around to code that might want to use it as an address. The square-brackets are part of the URL, not part of the IP address.

The good news is that this is the only place in the conformance tests that formats URLs.

robscott: I think @michaelbeaumont's updated comment that #1095 should fix this makes sense, can you confirm @acnodal-tc?

acnodal-tc: I'll give it a try.

PR #1096, opened 2022-04-07
Allow the user to tell the conformance tests leave config in place

acnodal-tc:
**What type of PR is this?**
/kind feature

With this small change you can run the conformance tests and leave the configuration in place at the end of the run:

` $ go test ./conformance/conformance_test.go --gateway-class={your-gw-class} --cleanup=false`

This is helpful when a test fails because you can then explore what went wrong.

```release-note
To run the conformance tests and leave the configuration in place at the end of the run:
 $ go test ./conformance/conformance_test.go --gateway-class={your-gw-class} --cleanup=false
```


nathancoleman:
@acnodal-tc this duplicates https://github.com/kubernetes-sigs/gateway-api/pull/1078 which already has some conversation on it

acnodal-tc:
Thanks for the tip! I added my $0.02 to that thread so I'll close this one.

PR #1100, opened 2022-04-08
conformance: add option to annotate Namespaces

michaelbeaumont:
**What this PR does / why we need it**:
Adds the option to annotate Namespaces created by the conformance tests.

**What type of PR is this?**
/kind feature

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1101, opened 2022-04-08
conformance: add option to pass valid ports for listeners and label Namespaces

michaelbeaumont:
Adds the option to label `Namespaces` and specify which ports to use for `Gateway` listeners in the conformance tests.

**What type of PR is this?**
/kind feature

```release-note
NONE
```

### DIFF
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -34,11 +34,63 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/conformance"
 )
 
+// Applier prepares manifests depending on the available options and applies
+// them to the Kubernetes cluster.
 type Applier struct {
-	NamespaceLabels map[string]string
+	NamespaceLabels    map[string]string
+	ValidListenerPorts []v1alpha2.PortNumber
+}
+
+// prepareGateway adjusts both listener ports and the gatewayClassName. It
+// returns an index pointing to the next valid listener port.
+func prepareGateway(t *testing.T, uObj *unstructured.Unstructured, gatewayClassName string, validListenerPorts []v1alpha2.PortNumber, portIndex int) int {
+	err := unstructured.SetNestedField(uObj.Object, gatewayClassName, "spec", "gatewayClassName")
+	require.NoErrorf(t, err, "error setting `spec.gatewayClassName` on %s Gateway resource", uObj.GetName())
+
+	if len(validListenerPorts) > 0 {
+		listeners, _, err := unstructured.NestedSlice(uObj.Object, "spec", "listeners")
+		require.NoErrorf(t, err, "error getting `spec.listeners` on %s Gateway resource", uObj.GetName())
+
+		for i, uListener := range listeners {
+			require.Less(t, portIndex, len(validListenerPorts), "not enough unassigned valid ports for `spec.listeners[%d]` on %s Gateway resource", i, uObj.GetName())
+
+			listener, ok := uListener.(map[string]interface{})
+			require.Truef(t, ok, "unexpected type at `spec.listeners[%d]` on %s Gateway resource", i, uObj.GetName())
+
+			nextPort := validListenerPorts[portIndex]
+			err = unstructured.SetNestedField(listener, int64(nextPort), "port")
+			require.NoErrorf(t, err, "error setting `spec.listeners[%d].port` on %s Gateway resource", i, uObj.GetName())
+
+			portIndex++
+			listeners[i] = listener
+		}
+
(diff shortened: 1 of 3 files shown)


conformance/utils/kubernetes/apply.go
```diff
@@ -61,6 +69,44 @@ func MustApplyWithCleanup(t *testing.T, c client.Client, location string, gcName
 		if uObj.GetKind() == "Gateway" {
```
robscott, line 69: I think we have enough modifications here now that we should probably extract this logic out to a separate function that we can write tests for. IE maybe a generic `prepareUnstructured` func that extracts all of the logic from lines 69-109 in this file.

conformance/utils/suite/suite.go
```diff
@@ -35,16 +36,21 @@ type ConformanceTestSuite struct {
...
+	Client               client.Client
+	GatewayClassName     string
+	Debug                bool
+	Cleanup              bool
+	RoundTripper         roundtripper.RoundTripper
+	BaseManifests        string
+	NamespaceAnnotations map[string]string
+	// ValidListenerPorts holds a list of ports assignable to Gateway listeners.
+	// If empty, every listener port will be left as is.
+	ValidListenerPorts []v1alpha2.PortNumber
```
robscott, line 53: Was there any intention to allow these to be configured with flags?

conformance/utils/kubernetes/apply.go
```diff
@@ -34,17 +34,25 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/conformance"
 )
 
+type Applier struct {
```
robscott, line 41: Could probably use a comment here describing the purpose of this struct.

robscott (commented):
Thanks for the work on this @michaelbeaumont! Just reviewing this PR since it's a superset of #1100, but still happy to merge them separately if that's easier for you.

conformance/utils/suite/suite.go
michaelbeaumont: We're using the conformance tests as a library so I didn't bother.

michaelbeaumont:
> Thanks for the work on this @michaelbeaumont! Just reviewing this PR since it's a superset of #1100, but still happy to merge them separately if that's easier for you.

I figured they're distinct features so I opened 2 PRs, and figured #1100 is a bit simpler. But either way for me.

michaelbeaumont:
@robscott I switched to namespace labels, I suspect that won't change much in terms of whether the options still make sense.

robscott:
Thanks @michaelbeaumont! This change LGTM, do you mind adding some simple unit tests for the new `prepareGateway` and `prepareNamespace` functions?

michaelbeaumont:
@robscott added some tests

conformance/utils/kubernetes/apply.go
```diff
@@ -34,17 +34,75 @@ import (
...
 // provided YAML file and registers a cleanup function for resources it created.
 // Note that this does not remove resources that already existed in the cluster.
-func MustApplyWithCleanup(t *testing.T, c client.Client, location string, gcName string, cleanup bool) {
+func (a Applier) MustApplyWithCleanup(t *testing.T, c client.Client, location string, gcName string, cleanup bool) {
 	data, err := getContentsFromPathOrURL(location)
 	require.NoError(t, err)
 
 	decoder := yaml.NewYAMLOrJSONDecoder(data, 4096)
+
+	portIndex := 0
```
robscott, line 105: It would be helpful to add a comment here describing the purpose of this index + why it's here any not per-Gateway. I think some may expect the unique ports to apply per Gateway not per manifest file. So if someone has 2 Gateways defined in a manifest with 2 ports each, they may be surprised that they need to define 4 ports and not just 2. (I could also be misinterpreting the code here).

robscott (commented):
Thanks for the work on this @michaelbeaumont!

conformance/utils/kubernetes/apply.go
michaelbeaumont: Yeah that's the behavior. I'll add a comment that the idea is to keep listener ports unique.

conformance/utils/kubernetes/apply_test.go
```diff
@@ -0,0 +1,205 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+func Test_prepareNamespace_empty(t *testing.T) {
```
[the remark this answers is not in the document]
michaelbeaumont: Done, also refactored a bit more to make testing easier.

robscott:
Thanks @michaelbeaumont!

/lgtm
/approve

PR #1107, opened 2022-04-11
Gateway.Spec.Addresses `Extended` Support Level

### LINKED ISSUE #1106: Gateway.Spec.Addresses Support Level Likely Inaccurate
`Gateway.Spec.Addresses` is [currently listed as having "Core" support](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha2/gateway_types.go#L138). This does not seem like something that some implementations will be able to support. It would be helpful to get some comments here from other implementers to confirm my theory that this is not universally supportable.

mehabhalodiya:
**What type of PR is this?**
/kind bug

This PR changes the support level i.e. from `Core` to `Extended` of `Gateway.Spec.Addresses`.
You may have  a look at [this comment](https://github.com/kubernetes-sigs/gateway-api/issues/1106#issuecomment-1094434674)

Fixes #1106 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -135,7 +135,7 @@ type GatewaySpec struct {
 	// it assigns to the Gateway and add a corresponding entry in
 	// GatewayStatus.Addresses.
 	//
-	// Support: Core
+	// Support: Extended
 	//
 	// +optional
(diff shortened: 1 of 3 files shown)


robscott:
Thanks for working on this @mehabhalodiya! This change looks good, but I think you also need to run `make generate` and commit the results to get the tests to pass here.

PR #1108, opened 2022-04-11
Log when cleaning up test resources

nathancoleman:
**What type of PR is this?**

/kind cleanup

I'm digging into an issue that I believe may be related to eventually consistent cleanup, and it's helpful to know when resources are being deleted in the same way we log when they're created.

Fixes #

```release-note
NONE
```


robscott:
Thanks!

/lgtm
/approve

Issue #1112, opened 2022-04-13
Route handling of invalid BackendRefs needs clarification


mikemorris:
**What would you like to be added**:

Add documentation, possibly a new page in the Reference section of the website, and additional `RouteConditionReason` and `ListenerConditionReason` types as needed to clarify intended handling of invalid BackendRefs configurations. This is a bit more complicated than just #1109, as the implications of initially applying or resolving changes to a Route's BackendRefs or a ReferencePolicy may require triggering actions or settings statuses across several objects.

**Why this is needed**:

There are a few distinct cases by which a `BackendRef` could be or become invalid, and I think it may be helpful to define and use more specific terms than "invalid" in the spec, as well as centralize and specify the intended behavior for each case more explicitly.

One specific difficulty is the lack of specificity around when a Route `Accepted` condition should be set, and whether an `Accepted { status: false, reason: Invalid }` condition should be set in any of these cases (and/or reflected in a [`ListenerConditionType`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.ListenerConditionType) status) given the lack of granularity in indicating status per BackendRef versus the status of the Route or Listener as a whole.

It's also slightly unclear if `attachedRoutes` on [`ListenerStatus`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.ListenerStatus) is intended to communicate the same thing as a route having an `Accepted { status: true, reason: Accepted }` route parent status condition referencing that listener because of the difference between the words used ("attached" vs "accepted", refs #1111).

## "not accepted"

This case is when the Route has not been accepted by a Gateway Listener. The `HTTPRouteInvalidCrossNamespace` conformance test illustrates one example where this is an intended outcome, but it's not clear how this status should be communicated. The test suggests that _no_ RouteParentStatus should be set on the Route because, having been rejected, it does not have an actual parent, only an intended parent - setting an `Accepted { status: False }` condition (what RouteConditionReason?) could be an alternative way to communicate this status though.

#### RouteParentStatus conditions
- None?
- **[Alternatively?]** `Accepted { status: False }`

#### Listener handling
- A Route that is in a namespace not allowed by a Gateway Listener's `allowedRoutes` should be rejected by the GatewayController.
[...]

PR #1114, opened 2022-04-14
api: add RouteConditionReason to v1alpha2

### LINKED ISSUE #1109: docs: add RouteConditionReason to v1alpha2 docs, enumerate possible reasons for each RouteConditionType

A section to the API Specification for `RouteConditionReason`, mirroring similar content from [`ListenerConditionReason`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.ListenerConditionReason)

**Why this is needed**:

Only [`RouteConditionType`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.RouteConditionType) is currently documented, and it does not offer sufficient clarity on what reasons are allowed for each condition type, and when those reasons should be used.

mikemorris:
**What type of PR is this?**
/kind documentation
/kind feature
/kind api-change
/kind deprecation

Adds a `RouteConditionReason` type with known required reasons and renames two `RouteConditionType` constants. Not sure if the release note formatting would make more sense to split this into two separate PRs.

Fixes #1109, refs #1112

```release-note
adds RouteConditionReason type with RouteReasonAccepted, RouteReasonResolvedRefs and RouteReasonRefNotPermitted constants
```
```release-note
deprecates ConditionRouteAccepted and ConditionRouteResolvedRefs constants, adds RouteConditionAccepted and RouteConditionResolvedRefs constants
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -189,11 +189,11 @@ type RouteConditionType string
 const (
 	// This condition indicates whether the route has been accepted or rejected
 	// by a Gateway, and why.
-	ConditionRouteAccepted RouteConditionType = "Accepted"
+	RouteConditionAccepted RouteConditionType = "Accepted"
 
 	// This condition indicates whether the controller was able to resolve all
 	// the object references for the Route.
-	ConditionRouteResolvedRefs RouteConditionType = "ResolvedRefs"
+	RouteConditionResolvedRefs RouteConditionType = "ResolvedRefs"
 )
 
 // RouteParentStatus describes the status of a route with respect to an
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -186,9 +186,20 @@ type BackendRef struct {
 // RouteConditionType is a type of condition for a route.
 type RouteConditionType string
 
+// RouteConditionReason is a reason for a route condition.
+type RouteConditionReason string
+
 const (
 	// This condition indicates whether the route has been accepted or rejected
 	// by a Gateway, and why.
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Accepted"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
 	RouteConditionAccepted RouteConditionType = "Accepted"
 
 	// FIXME: alias for backwards compatibility in v1alpha2, remove in next release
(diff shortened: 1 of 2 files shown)


rainest:
In the course of writing our Gateway.Listeners[].allowedRoutes implementation @shaneutt noted that there was no way we could indicate rejections on the resource. He originally suggested a `ConditionRouteRejected` in line with the existing constants, but assuming this is accepted it looks like we should instead add a false RouteConditionReason for RouteConditionAccepted.

Not sure if we could piggyback off this PR to do so; I submitted https://github.com/mikemorris/gateway-api/pull/1 if so, or can retarget it to this repo after this is merged.

mikemorris:
Agreed that's a valid case that's difficult to communicate currently @rainest, and I do think communicating this through a `RouteConditionReason` feels safer than crossing a disallowed namespace boundary to set a `ListenerStatus` condition.

I think I'd lean towards retargeting after merging this though given that it's somewhat new/unspecified scope - might be nice to add a conformance test (or augment the existing `HTTPInvalidCrossNamespace` test) illustrating expected usage too!

apis/v1alpha2/shared_types.go
```diff
@@ -186,14 +186,57 @@ type BackendRef struct {
...
+	//
+	// * "Accepted"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	RouteConditionAccepted RouteConditionType = "Accepted"
+
+	// FIXME: alias for backwards compatibility in v1alpha2, remove in next release
+	ConditionRouteAccepted RouteConditionType = RouteConditionAccepted
```
robscott, line 206: We've generally had a policy that breaking changes/renames to our Go types/APIs are acceptable in any release as long as they don't result in breaking changes to the CRDs. That matches the upstream behavior, so we could just remove these lines. I'm also open to changing this to be stricter.

apis/v1alpha2/shared_types.go
```diff
@@ -186,14 +186,57 @@ type BackendRef struct {
...
+	//
+	// * "RefNotPermitted"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	RouteConditionResolvedRefs RouteConditionType = "ResolvedRefs"
+
+	// FIXME: alias for backwards compatibility in v1alpha2, remove in next release
+	ConditionRouteResolvedRefs RouteConditionType = RouteConditionResolvedRefs
```
robscott, line 225: Same comment as above.

apis/v1alpha2/shared_types.go
```diff
@@ -186,14 +186,57 @@ type BackendRef struct {
...
+
+	// This reason is used with the "Accepted" condition when the Route has been
+	// accepted by the Gateway.
+	RouteReasonAccepted RouteConditionReason = "Accepted"
```
robscott, line 229: Nit: In other place we've grouped conditions and acceptable reasons together in type definitions. So essentially

Condition Accepted
- Reason Accepted

Condition ResolvedRefs
- Reason ResolvedRefs
- ReasonRefNotPermitted

robscott:
@mikemorris A few tiny comments but otherwise LGTM. @rainest I like the change you proposed in https://github.com/mikemorris/gateway-api/pull/1, do you mind filing a follow up PR here once this one merges?

apis/v1alpha2/shared_types.go
```diff
@@ -186,14 +186,51 @@ type BackendRef struct {
...
 
 	// This condition indicates whether the controller was able to resolve all
 	// the object references for the Route.
-	ConditionRouteResolvedRefs RouteConditionType = "ResolvedRefs"
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "ResolvedRefs"
+	//
+	// Possible reasons for this condition to be False are:
```
robscott, line 216: Tiny nit, capitalization is inconsistent with "true" above. No strong preference either way, just as long as they're consistent.
```suggestion
	// Possible reasons for this condition to be false are:
```

robscott (commented):
Found one more very tiny nit, but does not need to block.

/approve

PR #1115, opened 2022-04-14
Implement GRPCRoute (GEP 1016)

### LINKED ISSUE #1016: GEP: GRPCRoute

A `GRPCRoute` resource for idiomatically routing gRPC traffic.

**Why this is needed**:

Full rationale detailed in [the GEP](https://github.com/kubernetes-sigs/gateway-api/pull/1004).

gnossen:
**What type of PR is this?**
/kind feature
/kind api-change

Fixes #1016

```release-note
Introduce GRPCRoute resource.
```


apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+
+	// BackendRefs defines the backend(s) where matching requests should be
+	// sent.
+
+	// If unspecified or invalid (refers to a non-existent resource or a Service
+	// with no endpoints), the rule performs no forwarding. If there are also no
+	// filters specified that would result in a response being sent, a gRPC `UNAVAILABLE`
+	// status is returned. `UNAVAILABLE` responses must be sent so that the overall
+	// weight is respected; if an invalid backend is requested to have 80% of
+	// requests, then 80% of requests must get a `UNAVAILABLE` instead.
```
markmc, line 217: Is any of #1151 relevant to carry over here?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+// GRPCRouteMatch defines the predicate used to match requests to a given
+// action. Multiple match types are ANDed together, i.e. the match will
+// evaluate to true only if all conditions are satisfied.
+//
+// For example, the match below will match a gRPC request only if its service
+// is `foo` AND it contains the `version: v1` header:
+//
+// ```
+// match:
+//   method:
```
markmc, line 237: Should this be?

```
// matches:
//  - method:
```

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+	// +listMapKey=name
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	Headers []GRPCHeaderMatch `json:"headers,omitempty"`
+}
+
+// GRPCMethodMatch describes how to select a gRPC route by matching the gRPC
+// request service and/or method.
+//
+// At least one of Service and Method must be a non-empty string.
```
markmc, line 266: Aren't we saying if both are omitted then it matches all? That's why the default match is just Type=Exact?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+	// `False` in the corresponding RouteParentStatus.
+	//
+	// If a Route (A) of type HTTPRoute or GRPCRoute is attached to a
+	// Listener and that listener already has another Route (B) of the other
+	// type attached and the intersection of the hostnames of A and B is
+	// non-empty, then the implementation must reject Route A. That is, the
+	// implementation must raise an 'Accepted' condition with a status of
+	// 'False' in the corresponding RouteParentStatus.
+	//
+	// Support: Core
```
markmc, line 115: Just thinking about what it means for GRPC specific stuff to be marked as Core if GRPCRoute itself is Extended

Referring to: https://gateway-api.sigs.k8s.io/concepts/guidelines/#overlapping-support-levels

Since these fields are never embedded in a struct with Core support level, then these fields can never be Core?

So ... it's just a nod to "when GRPCRoute becomes Core, some fields will be Core and some will be Extended" ?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+	// +kubebuilder:validation:Enum=RequestHeaderModifier;RequestMirror;ExtensionRef
+	// <gateway:experimental:validation:Enum=RequestHeaderModifier;RequestMirror;ExtensionRef>
+	Type GRPCRouteFilterType `json:"type"`
+
+	// RequestHeaderModifier defines a schema for a filter that modifies request
+	// headers.
+	//
+	// Support: Core
+	//
+	// Support: Core
```
markmc, line 387: Duplicated

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+	// +optional
+	RequestMirror *HTTPRequestMirrorFilter `json:"requestMirror,omitempty"`
+
+	// ExtensionRef is an optional, implementation-specific extension to the
+	// "filter" behavior.  For example, resource "myroutefilter" in group
+	// "networking.example.net"). ExtensionRef MUST NOT be used for core and
+	// extended filters.
+	//
+	// Support: Implementation-specific
+	// Support: Implementation-specific
```
markmc, line 407: Duplicated

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+	// with the "RefNotPermitted" reason and not configure this backend in the
+	// underlying implementation.
+	//
+	// In either error case, the Message of the `ResolvedRefs` Condition
+	// should be used to provide more detail about the problem.
+	//
+	// Support: Custom
+	//
+	// +optional
+	BackendRefs []BackendRef `json:",inline"`
```
markmc, line 434: Hmm, why is this a slice? That's not what's in HTTP, nor in the GEP?

hack/update-codegen.sh
```diff
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -x
```
markmc, line 17: Spurious change

markmc (changes requested):
One other comment - it would be good to squash the commits (it's all one logical change, unless you wanted to have the autogen stuff in a separate commit) and at least add to the commit message the GEP number you're implementing

Just trying to smooth the way to merging. HTH

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,445 @@
...
+//
+// Implementations supporting `GRPCRoute` with the `HTTPS` `ProtocolType` must
+// accept HTTP/2 connections without an initial upgrade from HTTP/1.1. If the
+// implementation does not support this, then it should raise a "Detached"
+// condition for the affected listener with a reason of "UnsupportedProtocol"
+//
+// Implementations supporting `GRPCRoute` with the `HTTP` `ProtocolType` must
+// support cleartext HTTP/2 without an initial upgrade from HTTP/1.1. If the
+// implementation does not support this, then it should raise a "Detached"
+// condition for the affected listener with a reason of "UnsupportedProtocol"
```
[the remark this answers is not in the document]
gnossen: > Are we just talking about what an implementation must support in order to claim GRPCRoute support?

That's exactly the intent.

> And so the user-facing docs would just say "your route may be rejected if the GatewayClass does not support GRPCRoute"?

Right.

apis/v1alpha2/grpcroute_types.go
gnossen: Thanks for the reference. Yes. I have translated this change to the appropriate correlate for gRPC.

apis/v1alpha2/grpcroute_types.go
gnossen: Indeed! Good catch.

apis/v1alpha2/grpcroute_types.go
gnossen: No. This was the original idea, but [it was shot down in review](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r840953154).

apis/v1alpha2/grpcroute_types.go
gnossen: No, the idea here is that this feature is Core to the `GRPCRoute` feature. Independent of that, `GRPCRoute` is extended relative to the Gateway API. There was previously a discussion on that topic [here](https://github.com/kubernetes-sigs/gateway-api/pull/1004#discussion_r840945168).

apis/v1alpha2/grpcroute_types.go
gnossen: That's embarrassing. Fixed.

apis/v1alpha2/grpcroute_types.go
gnossen: Fixed.

apis/v1alpha2/grpcroute_types.go
gnossen: Whoops. Fixed.

hack/update-codegen.sh
gnossen: Ah, indeed. Though it's pretty hard to debug docs generation without it.

gnossen:
@markmc Thanks for the review. Sorry for the turnaround time on the response.

As for squashing, that's going to be a little more difficult now that I've merged master in. Is it not possible to enable auto-squashing on the repo if that is the preferred commit history style?

config/crd/kustomization.yaml
```diff
@@ -1,4 +1,3 @@
-resources:
```
robscott, line 1: Was this intentional?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+	// requests that would have otherwise been routed to an invalid backend. If
+	// multiple backends are specified, and some are invalid, the proportion of
+	// requests that would otherwise have been routed to an invalid backend
+	// MUST receive an `UNIMPLEMENTED` status.
+	//
+	// When a BackendRef refers to a Service that has no ready endpoints, it is
+	// recommended to return an `UNAVAILABLE` status.
+	//
+	// Support: Core for Kubernetes Service
+	// Support: Custom for any other resource
```
youngnick, line 236: Could you see about updating this to be similar to the new(ish) HTTPRoute BackendRefs field reference (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/httproute_types.go#L201-L231) here?

In particular, I think that we need to be clear about the status behavior; When a BackendRef is invalid, the Attached Condition and ResolvedRefs Conditions should be updated in a similar way to the HTTPRoute ones.

In particular, I think we need to be specific about what happens if zero, one out of one, or one out of more than one BackendRefs are invalid.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+	// "ResolvedRefs"  condition on the Route is set to `status: False`,
+	// with the "RefNotPermitted" reason and not configure this backend in the
+	// underlying implementation.
+	//
+	// In either error case, the Message of the `ResolvedRefs` Condition
+	// should be used to provide more detail about the problem.
+	//
+	// Support: Custom
+	//
+	// +optional
```
youngnick, line 449: Could you please do another pass here with the updated HTTPRoute mechanisms in mind (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/httproute_types.go#L934-L962)?

youngnick (commented):
Mostly looks LGTM, but I think we need to backport some of the changes we made to HTTPRoute for v0.5.0 about BackendRefs. I'm not one hundred percent sure the semantics are the same for GRPCRoute, but we need to review at least.

apis/v1alpha2/httproute_types.go
```diff
@@ -93,6 +93,13 @@ type HTTPRouteSpec struct {
 	// implementation must raise an 'Accepted' Condition with a status of
 	// `False` in the corresponding RouteParentStatus.
 	//
+	// If a Route (A) of type HTTPRoute or GRPCRoute is attached to a
+	// Listener and that listener already has another Route (B) of the other
+	// type attached and the intersection of the hostnames of A and B is
+	// non-empty, then the implementation must reject Route A. That is, the
+	// implementation must raise an 'Accepted' condition with a status of
+	// 'False' in the corresponding RouteParentStatus.
```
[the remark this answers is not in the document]
robscott: Agree with Nick here, this is slightly inconsistent with our other conflict resolution guidance which generally relies on the creationTimestamp of the route and not when it was attached since creationTimestamp is more reliable for stateless controllers.

https://github.com/kubernetes-sigs/gateway-api/blob/ca079e6d90f2cb7521443dd5722a6c9fe345b8be/apis/v1beta1/httproute_types.go#L143-L159

apis/v1alpha2/httproute_types.go
```diff
@@ -478,11 +486,13 @@ const (
...
-//   headers:
-//   - name: "version"
-//     value "v1"
+//
+//	path:
+//	  value: "/foo"
+//	headers:
+//	- name: "version"
+//	  value "v1"
+//
```
robscott, line 495: IS this change intentional?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Hostnames",type=string,JSONPath=`.spec.hostnames`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// GRPCRoute provides a way to route gRPC requests. This includes the capability
+// to match requests by hostname, gRPC service, gRPC method, or HTTP/2 header. Filters can be
+// used to specify additional processing steps. Backends specify where matching
+// requests should be routed.
+//
+// Implementations supporting `GRPCRoute` with the `HTTPS` `ProtocolType` must
```
robscott, line 36: Nit: We've generally used all caps for keywords like SHOULD and MUST.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+//
+// “`
+// matches:
+// - method:
+//
+//	  type: Exact
+//	  service: "foo"
+//	headers:
+//	- name: "version"
+//	  value "v1"
```
robscott, line 260: Indentation looks off here

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+	Type *GRPCMethodMatchType `json:"type,omitempty"`
+
+	// Value of the service to match against. If left empty or omitted, will
+	// match all services.
+	//
+	// At least one of Service and Method must be a non-empty string.
+	// +optional
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=1024
+	Service *string `json:"service,omitempty"`
```
robscott, line 305: Can we add some regex validation here? What characters would be valid?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+
+	// Value of the method to match against. If left empty or omitted, will
+	// match all services.
+	//
+	// At least one of Service and Method must be a non-empty string.
+	// +optional
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=1024
+	Method *string `json:"method,omitempty"`
```
robscott, line 314: Can we add some regex validation here?

robscott (commented):
Thanks @gnossen! This mostly LGTM.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+// +kubebuilder:printcolumn:name="Hostnames",type=string,JSONPath=`.spec.hostnames`
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// GRPCRoute provides a way to route gRPC requests. This includes the capability
+// to match requests by hostname, gRPC service, gRPC method, or HTTP/2 header. Filters can be
+// used to specify additional processing steps. Backends specify where matching
+// requests should be routed.
+//
+// Implementations supporting `GRPCRoute` with the `HTTPS` `ProtocolType` must
+// accept HTTP/2 connections without an initial upgrade from HTTP/1.1. If the
```
hbagdi, line 37: I.e. using ALPN? If so, should we mention it?

apis/v1alpha2/grpcroute_types.go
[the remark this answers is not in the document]
hbagdi: Is there no leeway for implementations that don't support upgrades over plain text but do support HTTP/2?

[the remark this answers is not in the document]
shaneutt: Could we please also add the reason why this is a requirement to the documentation?

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
gnossen: It seems like this could cause issues if someone tried to attach a route to a gateway with a route already attached. Is it possible for the controller to check for an `Accepted: true` condition at startup?

apis/v1alpha2/shared_types.go
```diff
@@ -333,7 +333,8 @@ type RouteStatus struct {
 //
 // 1. IPs are not allowed.
 // 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
-//    label must appear by itself as the first label.
+//
```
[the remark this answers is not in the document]
gnossen: Done.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+// implementation does not support this, then it should raise a "Detached"
```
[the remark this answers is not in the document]
gnossen: Done.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -0,0 +1,461 @@
...
+// condition for the affected listener with a reason of "UnsupportedProtocol"
+//
+// Implementations supporting `GRPCRoute` with the `HTTP` `ProtocolType` must
+// support cleartext HTTP/2 without an initial upgrade from HTTP/1.1. If the
+// implementation does not support this, then it should raise a "Detached"
```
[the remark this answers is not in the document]
gnossen: Done.

apis/v1alpha2/grpcroute_types.go
gnossen: Done.

apis/v1alpha2/httproute_types.go
gnossen: Looks like this was an issue with my Go toolchain. Fixed.

apis/v1alpha2/grpcroute_types.go
gnossen: Changed here and throughout.

apis/v1alpha2/grpcroute_types.go
gnossen: Formatter strikes again. Fixed.

apis/v1alpha2/grpcroute_types.go
gnossen: Trying to get consensus from the gRPC community on what's reasonable here since this isn't codified. For the moment, I'm just disallowing the slash character.

apis/v1alpha2/grpcroute_types.go
gnossen: Same answer as above.

gnossen:
> Is there no leeway for implementations that don't support upgrades over plain text but do support HTTP/2?

@hbagdi I'm not sure why I'm not able to respond to this comment directly, so I'm responding here.

I'm thinking this must be a typo? The spec requires that implementations supporting GRPCRoute with the `HTTP` protocol type support HTTP/2 _without_ an upgrade. An implementation as in your comment is absolutely conformant to the current spec as written.

I'm guessing you meant

> Is there no leeway for implementations that only support HTTP/2 over plain text with an upgrade from HTTP/1 but do support HTTP/2?

If so, the answer is "not really." AFAIK, this would not be useful to any of the existing gRPC clients and servers in the wild. It would more or less be a broken implementation.

apis/v1alpha2/grpcroute_types.go
gnossen: Yes, using ALPN. I've made this explicit.

apis/v1alpha2/grpcroute_types.go
gnossen: Done.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
gnossen: Discussed in the community meeting. Going ahead and sticking in line with HTTPRoute/HTTPRoute conflict resolution for now but @youngnick is looking into solving this issue for the Gateway API in a more holistic manner.

apis/v1alpha2/grpcroute_types.go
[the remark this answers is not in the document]
gnossen: Done.

apis/v1alpha2/httproute_types.go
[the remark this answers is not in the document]
mikemorris: Refs https://github.com/kubernetes-sigs/gateway-api/issues/1211#issuecomment-1156710105 for prior discussion of whether checking status for conflict resolution is a viable path forward.

robscott:
Hey @gnossen I think this is good to merge, do you mind running `make generate` again with go 1.19rc2? Kubernetes updated it's test infra a couple days ago which means we needed to update our codegen: https://github.com/kubernetes-sigs/gateway-api/pull/1288

gnossen:
> Hey @gnossen I think this is good to merge, do you mind running make generate again with go 1.19rc2? Kubernetes updated it's test infra a couple days ago which means we needed to update our codegen: https://github.com/kubernetes-sigs/gateway-api/pull/1288

Done

robscott:
@gnossen Actually one more step - it looks like this needs a rebase + `make generate` run to get presubmits to pass. (This is because of a conflict with https://github.com/kubernetes-sigs/gateway-api/pull/1276).

robscott:
Thanks for all the work on this @gnossen! Discussed at community meeting today and there were no concerns about merging this in. Will plan to merge this at some point tomorrow unless else someone LGTMs before then (that's also welcome).

/approve

PR #1118, opened 2022-04-18
Removing confusing part of ParentRef godoc

### LINKED ISSUE #988: Route references to unknown types need to be followed before checking GatewayClass ownership
**What happened**:

If a controller observes a change in a route, it needs to follow the parent refs to see whether it attaches to a gateway class that the controller is implementing. If a user specifies a (valid) parent group and kind that the controller doesn't implement, it can't follow the reference to know whether it was supposed to configure the route or not.

ParentRef says:
```
// References to objects with invalid Group and Kind are not valid, and must
// be rejected by the implementation, with appropriate Conditions set
// on the containing object.
```

If the route parent is a reference to a type the controller doesn't understand (e.g. "Mesh"), the controller should be able to know that it should not update the route status.

robscott:
**What type of PR is this?**
/kind documentation

This removes a confusing part of the ParentRef doc. The problem was that an invalid ParentRef would mean that no implementation could possibly understand that they were the intended implementor, since the only possible link to an implementation (parentRef) was invalid. I considered leaving in "References to objects with invalid Group and Kind are not valid", but that was really just repeating what was stated in the previous paragraph.

Fixes #988

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -27,10 +27,6 @@ import (
 //
 // The API object must be valid in the cluster; the Group and Kind must
 // be registered in the cluster for this reference to be valid.
-//
-// References to objects with invalid Group and Kind are not valid, and must
-// be rejected by the implementation, with appropriate Conditions set
-// on the containing object.
 type ParentReference struct {
 	// Group is the group of the referent.
 	//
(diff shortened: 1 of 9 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -27,10 +27,6 @@ import (
 //
 // The API object must be valid in the cluster; the Group and Kind must
 // be registered in the cluster for this reference to be valid.
-//
-// References to objects with invalid Group and Kind are not valid, and must
```
youngnick, line 31: Feels like this is trying to say something like "if the reference is to an object you don't control, don't accept it", but that seems pretty redundant, agreed.

PR #1119, opened 2022-04-20
Add a "cleanup-base-resources" conformance flag, and a way to use it

acnodal-tc:
**What type of PR is this?**
/kind cleanup

There was a flag called "cleanup" but that wasn't specific enough[1],
so this "cleanup-base-resources" flag does the same thing but leaves
room for other flags that will clean up other things.

This commit adds a Make variable called CONFORMANCE_FLAGS which
provides a way to pass that flag (or any other) into the conformance
tests. For example, if I wanted to override the test GatewayClass
and *not* clean up the base resources, this would be the invocation:

 $ make conformance CONFORMANCE_FLAGS="-cleanup-base-resources=false -gateway-class=acndev-http4"

[1] https://github.com/kubernetes-sigs/gateway-api/pull/1078

```release-note
This commit adds a Make variable called CONFORMANCE_FLAGS which provides a way to pass that flag (or any other) into the conformance tests. For example, if I wanted to override the test GatewayClass and *not* clean up the base resources, this would be the invocation:

 $ make conformance CONFORMANCE_FLAGS="-cleanup-base-resources=false -gateway-class=acndev-http4"
```

### DIFF
--- a/conformance/utils/kubernetes/apply_test.go
+++ b/conformance/utils/kubernetes/apply_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kubernetes
 
 import (
+	"flag"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


conformance/utils/kubernetes/apply_test.go
```diff
@@ -27,6 +28,15 @@ import (
...
 
+var (
+	// These aren't used by this test but since this test is run using
+	// the same invocation as the conformance test it needs to parse the
+	// same set of flags.
+	_ = flag.String("gateway-class", "", "unused")
+	_ = flag.Bool("debug", false, "unused")
+	_ = flag.Bool("cleanup-base-resources", true, "unused")
+)
+
```
robscott, line 39: Maybe we need a config or flags package so we don't have to repeat these in both places?

robscott (commented):
Thanks for the work on this!

PR #1121, opened 2022-04-21
Mention BackendObjectReference.Kind default in docstring

rainest:
**What type of PR is this?**
/kind documentation

Mention that the BackendObjectReference.Kind default is `Service` in its godoc string. Doc generation pulls in that this is optional, but not the kubebuilder tag that indicates the default, and a BackendObjectReference with an _empty_ Kind doesn't make much sense.

```release-note
NONE
```


robscott:
@rainest Thanks for adding this! The change LGTM, but it looks like you need to run `make generate` + commit the results of that.

robscott:
Thanks @rainest!

/lgtm
/approve

PR #1122, opened 2022-04-21
Update RouteParentStatus.ControllerName Description

spencerhance:
/kind documentation

Add additional comment explaining that this field cannot be empty or updates to status will be rejected.  It is not immediately clear otherwise from looking at the RouteParentStatus struct that this is required.

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -249,6 +249,10 @@ type RouteParentStatus struct {
 	// The format of this field is DOMAIN "/" PATH, where DOMAIN and PATH are
 	// valid Kubernetes names
 	// (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names).
+	//
+	// Controllers MUST populate this field when writing status. Controllers should ensure that
+	// entries to status populated with their ControllerName are cleaned up when they are no
+	// longer necessary.
 	ControllerName GatewayController `json:"controllerName"`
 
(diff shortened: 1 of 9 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -249,6 +249,8 @@ type RouteParentStatus struct {
 	// The format of this field is DOMAIN "/" PATH, where DOMAIN and PATH are
 	// valid Kubernetes names
 	// (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names).
+	//
+	// This field cannot be empty.
```
robscott, line 253: In general we've only added explicit comments about a field being optional, but otherwise it is assumed to be required. With that said, I think we do need some more clarification here, so maybe something like this that also helps explain the purpose of the field?

```suggestion
	// Controllers MUST populate this field when writing status. Controllers should ensure that
	// entries to status populated with their ControllerName are cleaned up when they are no
	// longer necessary.
```

robscott (commented):
Thanks for the work on this @spencerhance!

apis/v1alpha2/shared_types.go
spencerhance: Good call, added!

spencerhance:
> I think you may need to commit the results of make generate before tests will pass.

Ah makes sense - added to PR

PR #1124, opened 2022-04-21
Refactoring path modifier to use union discriminator

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This is a follow up to @thockin's feedback in https://github.com/kubernetes-sigs/gateway-api/pull/1086#discussion_r846356699. It makes the following changes:

- HTTPPathModifier API
  - Replaces "Substitution" with new type-specific "ReplaceFullPath" and "ReplacePrefixMatch" fields
  - Renames "Absolute" to "ReplaceFullPath" for more consistency
- Validation
  - Adds webhook validation to ensure that only type-appropriate fields are set in HTTPPathModifier (this could be done with oneOf validation, but that is not supported with Kubebuilder annotations: https://github.com/kubernetes-sigs/controller-tools/issues/461)
  - Adds webhook validation to ensure that rewrite and redirect validation 
  - Adds examples of path rewrites and redirects that will be used in verification script
- Use "PreciseHostname" instead of "Hostname" URLRewrite filter

Not sure this needs a release note because none of the updated fields were ever released, so up to author of changelog as far as if this should be included:

```release-note
- HTTPPathModifier API
  - Replaces "Substitution" with new type-specific "ReplaceFullPath" and "ReplacePrefixMatch" fields
  - Renames "Absolute" to "ReplaceFullPath" for more consistency
- Validation
  - Adds webhook validation to ensure that only type-appropriate fields are set in HTTPPathModifier (this could be done with oneOf validation, but that is not supported with Kubebuilder annotations: https://github.com/kubernetes-sigs/controller-tools/issues/461)
  - Adds webhook validation to ensure that rewrite and redirect validation 
  - Adds examples of path rewrites and redirects that will be used in verification script
- Use "PreciseHostname" instead of "Hostname" URLRewrite filter
```


robscott:
This is a relatively large change, adding a hold until we can get some broader consensus.

/hold

robscott:
There's a [kubernetes-api-reviewers thread](https://groups.google.com/g/kubernetes-api-reviewers/c/Q7UC6hptRlw) that's relevant to this PR.

apis/v1alpha2/httproute_types.go
```diff
@@ -705,17 +705,25 @@ type HTTPPathModifier struct {
 	// Type defines the type of path modifier.
 	//
 	// <gateway:experimental>
-	// +kubebuilder:validation:Enum=Absolute;ReplacePrefixMatch
+	// +kubebuilder:validation:Enum=ReplaceFullPath;ReplacePrefixMatch
```
apelisse-test, line 708: Since you really don't have any other fields than union members in that struct, I don't think the discriminator is useful at all.

apelisse: ^ this was me, I posted with a temporary account ;-)

robscott: I think this adds room for growth. The goal here is to add additional types in the future. Since we already have 14 implementations and integrations, and have more on the way, it's very likely that at least some of these will trail the version of the API deployed in the cluster. A discriminator field allows those implementations to communicate that they don't support a specific type, instead of just seeing an empty struct and not knowing what to do with it.

apelisse: Yeah, they don't get a lot more value from seeing the struct empty vs seeing the discrimator set to a value that doesn't make sense (but they get a little value). In both cases, they can manually clear everything/set the discriminator to the right value.

robscott: I chatted with @thockin about this, and while I think he still has significant reservations about using a union discriminator, I think he's on board with starting with one. The idea is that it's impossible to retroactively add a discriminator, but we could always make a discriminator optional and/or provide a smart default in the future. (Please correct me if I misunderstood anything).

apis/v1alpha2/httproute_types.go
```diff
@@ -705,17 +705,25 @@ type HTTPPathModifier struct {
 	// Type defines the type of path modifier.
 	//
```
youngnick, line 706: I think we need a callout here that we expect new values to be added and that implementations should ensure they handle values not in this list without crashing etc.

youngnick (commented):
LGTM with one caveat about the kubebuilder enum.

apis/v1alpha2/httproute_types.go
robscott: Updated, thanks!

robscott:
Discussed at community meeting today. Seems like there's consensus with this approach. I've pushed an update to address feedback and going to go ahead and remove the hold so we can get this in with the next LGTM.

/hold cancel

Issue #1104, opened 2022-04-08
Conformance Tests: L7 Gateways


robscott:
This issue is intended to track conformance test development for L7 Gateways. Comment below if you're interested in working on covering any of these areas.

**Core Capabilities:**
- [x] Gateway - Route Binding
  - [x] Successful cross namespace configuration
  - [x] Invalid cross namespace configuration
  - [x] Invalid binding from Route kind not specified in AllowedRoutes.Kinds (@skriss) 
- [x] Hostname matching (@skriss) 
  - [x] Multiple listeners specifying different hostnames and same port, forwarding to different Routes
  - [x] Single listener specifying hostname that intersects with one of several hostnames specified on a Route 
  - [x] Single listener specifying hostname that intersects with none of several hostnames specified on a Route 
- [x] TLS Certificates (Prior art in https://github.com/kubernetes-sigs/ingress-controller-conformance that can be a starting point) - tracked in #1330

**Extended Capabilities:**
- [ ] #2035
  - [ ] Specify 1 address
  - [ ] Specify multiple addresses

skriss:
I'm interested in picking up some of these - can start with the remaining route binding one and the hostname matching ones. @robscott were you thinking that the Gateway manifests for these would go into the individual tests, rather than the "base" manifests, since these aren't likely to be reused across many tests?

robscott:
@skriss that's a great question. I think it will be most straightforward to start with Gateways being defined by the tests directly. We'll need to discuss more broadly though, maybe at the next community meeting. I think some implementations don't support automatically provisioning Gateways yet, so depending on how widespread that is, we may need to move Gateway definitions into base manifests. I'm thinking it should be easy to change where Gateways are defined after the tests are written though.

shaneutt:
At this point the remaining items are broken off as separate issues, so we're good to close this.

PR #1130, opened 2022-04-28
conformance: add test for disallowed route kind binding

skriss:
**What type of PR is this?**
conformance test (looks like we need a new kind)

Adds a test for a route that has a parentRef to a
Gateway whose listeners don't allow the route type.

This test largely follows the pattern from "invalid cross namespace", except it creates a new Gateway that disallows HTTPRoutes.

Updates #1104.

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-disallowed-kind.go
+++ b/conformance/tests/httproute-disallowed-kind.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteDisallowedKind)
+}
+
+var HTTPRouteDisallowedKind = suite.ConformanceTest{
+	ShortName:   "HTTPRouteDisallowedKind",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should fail to attach to a Listener that does not allow the HTTPRoute kind",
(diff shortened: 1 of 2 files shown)


conformance/tests/httproute-disallowed-kind.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - name: tls
+    port: 443
+    protocol: TLS
+    tls:
+      mode: Passthrough
+    allowedRoutes:
+      namespaces:
+        from: Same
+      kinds:
+      - kind: TLSRoute
```
skriss, line 18: Per https://github.com/kubernetes-sigs/gateway-api/issues/1104#issuecomment-1112515816 it may be problematic for some implementations not to have the Gateway definitions in the base manifests, need to sort out the pattern here.

robscott: cc @hbagdi and @shaneutt

conformance/tests/httproute-disallowed-kind.yaml
```diff
@@ -0,0 +1,32 @@
...
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: HTTPRoute
+metadata:
+  name: httproute-disallowed-kind
```
robscott, line 23: Nit: no need for httproute prefix in name.
```suggestion
  name: disallowed-kind
```

conformance/tests/httproute-disallowed-kind.go
```diff
@@ -0,0 +1,71 @@
...
+		routeName := types.NamespacedName{Name: "httproute-disallowed-kind", Namespace: "gateway-conformance-infra"}
+		gwName := types.NamespacedName{Name: "tlsroutes-only", Namespace: "gateway-conformance-infra"}
+
+		// TODO: Determine if this is actually what we want. It is likely
+		// preferable to have status set with some kind of warning/error message
+		// but that is also unlikely to be universally achievable.
+		t.Run("Route should not have Parents set in status", func(t *testing.T) {
+			parents := []v1alpha2.RouteParentStatus{}
+			kubernetes.HTTPRouteMustHaveParents(t, suite.Client, routeName, parents, true, 60)
+		})
```
robscott, line 53: I don't think we need to block on this since other tests are already doing the same thing, but created https://github.com/kubernetes-sigs/gateway-api/issues/1135 to track this.

conformance/tests/httproute-disallowed-kind.yaml
robscott: For reference, the alternative here would be to add more Gateways to our base set of resources instead. That would work, but could become difficult to maintain.

robscott (commented):
Thanks @skriss!

This LGTM, but want to get feedback from other implementers on this one.

conformance/tests/httproute-disallowed-kind.yaml
skriss: :+1: removed.

conformance/tests/httproute-disallowed-kind.yaml
robscott: Discussed in community meeting on Monday in reference to this and https://github.com/kubernetes-sigs/gateway-api/pull/1137#discussion_r863243598. Apparently this ended up being a non issue for implementations now.

robscott:
Thanks @skriss!

/lgtm
/approve

PR #1137, opened 2022-05-02
conformance: add test for listener hostname matching

skriss:
**What type of PR is this?**
/kind conformance-test

Adds a test that verifies that different hostnames
for different HTTP listeners that all use the same
port are matched correctly.

Updates #1104.

```release-note
NONE
```

Note, this test also creates a new Gateway, so same question as for https://github.com/kubernetes-sigs/gateway-api/pull/1130#discussion_r861332705. There will be at least one more hostname-related test coming that requires a new Gateway as well.

### DIFF
--- a/conformance/tests/httproute-listener-hostname-matching.yaml
+++ b/conformance/tests/httproute-listener-hostname-matching.yaml
@@ -1,0 +1,74 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: Gateway
+metadata:
+  name: httproute-listener-hostname-matching
+  namespace: gateway-conformance-infra
+spec:
+  gatewayClassName: {GATEWAY_CLASS_NAME}
+  listeners:
+  - name: listener-1
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: Same
+    hostname: bar.com
+  - name: listener-2
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: Same
+    hostname: foo.bar.com
+  - name: listener-3
+    port: 80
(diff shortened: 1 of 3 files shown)


conformance/tests/httproute-listener-hostname-matching.yaml
```diff
@@ -0,0 +1,74 @@
...
+  parentRefs:
+  - name: httproute-listener-hostname-matching
+    namespace: gateway-conformance-infra
+    sectionName: listener-2
+  rules:
+  - backendRefs:
+    - name: infra-backend-v2
+      port: 8080
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
howardjohn, line 61: Should we add some HTTPRoutes with hostname as well to test intersection?

Especially something like listener: `*.com`, route: `foo.example.com`

skriss: I have another test coming that specifically tests hostname intersections -- I thought it was simpler to split them into separate tests, but can combine them if folks prefer.

robscott: 👍 Agree that it would be good to have. I'd also tend to lean towards a separate PR for the additional tests, but don't have a strong preference.

conformance/tests/httproute-listener-hostname-matching.yaml
```diff
@@ -0,0 +1,74 @@
...
+kind: Gateway
```
nathancoleman, line 2: Noting from the SIG meeting that the Consul team is working on other conformance tests for this repo that spin up Gateways dynamically.

LGTM

robscott:
Thanks @skriss!

/lgtm 
/approve

PR #1139, opened 2022-05-02
Reverting GEP 735 from implemented to provisional

### LINKED ISSUE #1127: Consider holding GEP 735 for a future release (after v0.5.0)
As discussed in today's community meeting, [API Review on v0.5.0](https://github.com/kubernetes-sigs/gateway-api/pull/1086) has raised some good questions about address matching. I've tried to answer some of them, but have realized that others will require more significant thought/changes. Here are some of the questions discussed in today's meeting:

1. Does source address matching act purely as a matching mechanism or is it more of a firewall? 
2. How does merging and conflict resolution work?
3. Would Gateway be a more natural place for firewall config?
4. What are primary use cases for dest address matching? I think it's likely mesh + egress implementations, but if so, we should more clearly define how that would work. Does it make sense for traffic flowing through an ingress Gateway?
5. More broadly, if these concepts don't make sense in TCPRoute, what does?

So at this point I think we may have more questions than answers, and it might make the most sense to leave this for a later release once we've had some time to spend more time answering these kinds of questions. Open to what others think here.

robscott:
**What type of PR is this?**
/kind cleanup
/kind gep
/kind api-change

API review for v0.5.0 raised some important questions that we didn't have great answers for. This temporarily removes this change until we can more clearly answer the questions and/or adjust the approach.

Fixes #1127 

```release-note
Address matching has been removed from TCP and UDP routes. This feature was never released, but represents a broader concept we're still interested in including in the API, it just likely requires more work to ensure we've thought through every possible use case and interpretation.
```


robscott:
This is a big change, adding a hold for consensus. I did chat with @shaneutt this morning to ensure this made sense to him, and there has been support on the corresponding issue (#1127) and community meeting.

/hold

robscott:
Discussed at community meeting, seems like we've got consensus on this, removing the hold.

/hold cancel

PR #1145, opened 2022-05-04
api: remove ListenerReasonRouteConflict

mikemorris:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind api-change
/kind deprecation

This appears to be an artifact of the v1alpha1 configuration style where a listener selects routes directly. In v1alpha2 an unsupported route attempting to attach to a listener should instead be rejected by the gateway, and optionally set an `Accepted` RouteCondition with status `False`.

A somewhat-related mismatch between the Listener's [`AllowedRoutes`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.AllowedRoutes) `kinds` field and the supported [`protocol`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.ProtocolType) is explicitly expected to be expressed by the `ResolvedRefs` ListenerCondition setting status `False` with the `InvalidRouteKinds` reason (but could possibly be covered instead by webhook validation for the Core support level ProtocolType values and xRoute kinds?)

Refs #1077, https://github.com/kubernetes-sigs/gateway-api/discussions/935#discussioncomment-1695833

```release-note
removes the RouteConflict ListenerConditionReason intended to describe an error state that is no longer possible in the v1alpha2 API
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -641,7 +641,6 @@ const (
 	//
 	// * "HostnameConflict"
 	// * "ProtocolConflict"
-	// * "RouteConflict"
 	//
 	// Possible reasons for this condition to be False are:
 	//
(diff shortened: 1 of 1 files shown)


robscott:
Thanks!

/lgtm
/approve

PR #1146, opened 2022-05-04
api: remove ListenerReasonUnsupportedExtension

mikemorris:
**What type of PR is this?**

/kind cleanup
/kind documentation
/kind api-change
/kind deprecation

This appears to be an artifact of the v1alpha1 configuration style where a listener selects routes directly. If I understand correctly, this is a reference to the `ExtensionRef` field on [`HTTPRouteRule`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRouteRule) (and other xRouteRule objects), which are the only other uses of the term `Extension` currently in the spec codebase.

`UnsupportedExtension` should likely become a RouteConditionReason instead, but I'm not quite sure with which RouteConditionType it should be used.

The intended behavior now is a bit unclear (it's an edge case not currently described in #1112). While rejecting a route entirely because of a single unsupported filter on one rule might have a large blast radius, I'm thinking through a few possible scenarios and potential impact/mitigation and wondering if that might actually be okay (and might even be desirable if using `ExtensionRef` to implement something like JWT auth):
- Moving from one Gateway API implementation to another - some degree of breaking change is likely expected here and would hopefully be discovered in a test environment and mitigated manually.
- Upgrading a Gateway API implementation to a new version which drops support for an old filter - this seems like something that should be clearly identified by the implementation as a breaking change in a changelog or mitigated safely through a deprecation process.
- Attempting to add a new filter to an existing route rule and making a typo - this feels like the most likely common operator risk, could we perhaps suggest implementations add custom webhook validation for this field to prevent accepting invalid `ExtensionRef` configurations that cause an attached route to become detached?

Refs #1077, https://github.com/kubernetes-sigs/gateway-api/discussions/935#discussioncomment-1695833

```release-note
removes the UnsupportedExtension ListenerConditionReason intended for a route selection model that has changed in the v1alpha2 API
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -689,7 +689,6 @@ const (
 	// Possible reasons for this condition to be true are:
 	//
 	// * "PortUnavailable"
-	// * "UnsupportedExtension"
 	// * "UnsupportedProtocol"
 	// * "UnsupportedAddress"
 	//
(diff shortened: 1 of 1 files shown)


robscott:
I agree that this removal makes sense. We should not detach a listener if a Route fails to attach to it. I think we should create a follow up issue to discuss what we should do instead of this. Will leave this for @hbagdi, @youngnick, or someone else to sign off. Or we can get final sign off at community meeting on Monday.  

/approve

PR #1151, opened 2022-05-06
Updating HTTP status code documentation to use 404 instead of 503

robscott:
**What type of PR is this?**
/kind documentation
/kind api-change
/area conformance

As suggested by @michaelbeaumont, this updates our HTTP status code guidance to require 404 status codes and removes any recommendations to return 503 status codes.

```release-note
HTTP 404 responses should be returned whenever a matching route rule does not exist or a backend reference can not be resolved.
```

Adding a hold for consensus.
/hold

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -158,6 +158,9 @@ type HTTPRouteRule struct {
 	// matching precedence MUST be granted to the first matching rule meeting
 	// the above criteria.
 	//
+	// When no rules matching a request have been successfully attached to the
+	// parent a request is coming from, a HTTP 404 status code MUST be returned.
+	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
 	// +kubebuilder:default={{path:{ type: "PathPrefix", value: "/"}}}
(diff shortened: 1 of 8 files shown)


robscott:
cc @mikemorris

youngnick:
I thought the reason we specified a 503 originally is that it allows users to tell that the error is originating from something in the path rather than the end app? Personally, I would expect a 404 to mean that the request got to the end app and then it wasn't found, not that the proxy is misconfigured.

robscott:
Meant to link the [context from Slack](https://kubernetes.slack.com/archives/CR0H13KGA/p1651785041134049).

>If we're sending 404s for misconfigurations, how will the app dev know that those 404s are not supposed to happen? 503s are used because they are a signal that something is wrong, and someone needs to take some action.

I think it's useful to think about what an implementation should do when a matching route rule does not exist. I think that a 404 is the most appropriate response in that case. The question then becomes if we should return different HTTP status code if someone has tried and failed to attach a Route. My thought is that failed attempts to attach a Route should be treated the same as if the Route was not attached to the Gateway. I think the same logic applies when a BackendRef is not valid/resolvable. 

For reference here are the [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231) definitions of the HTTP status codes:

> The 503 (Service Unavailable) status code indicates that the server
>    is currently unable to handle the request due to a temporary overload
>    or scheduled maintenance, which will likely be alleviated after some
>    delay.  The server MAY send a Retry-After header field
>    ([Section 7.1.3](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3)) to suggest an appropriate amount of time for the
>    client to wait before retrying the request.


>   The 404 (Not Found) status code indicates that the origin server did
>    not find a current representation for the target resource or is not
>    willing to disclose that one exists.  A 404 status code does not
>    indicate whether this lack of representation is temporary or
>    permanent; the 410 (Gone) status code is preferred over 404 if the
>    origin server knows, presumably through some configurable means, that
>    the condition is likely to be permanent.

Although neither seem like a perfect fit here, 404 seems more accurate to me.

robscott:
Another use case that we should document - Service exists and is valid to reference but no healthy endpoints.

robscott:
Also cover unsupported types.

robscott:
Updated to account for feedback from today's community meeting.

youngnick:
After the discussion in our meeting, I'm behind this change.

/lgtm

robscott:
Seems like we had consensus at community meeting.

/hold cancel

mikemorris:
Sorry for not looking at this closely earlier - the part that feels a bit strange to me with this change is
> // If multiple backends are specified, and some are invalid, the proportion of
// requests that would otherwise have been routed to an invalid backend
// MUST receive a 404 status code.

Because (I think?) some sort of route would have been matched for the logic to reach the backend weighting, and if _some_ backends are valid, retrying the request could potentialy reach a live backend (closer to 503 logic).

Issue #1135, opened 2022-04-29
Conformance tests should allow an error state in Route Parent status


robscott:
Currently we have several tests with a TODO to allow implementations to actually provide an error in Route parent status if they are able to ([example](https://github.com/kubernetes-sigs/gateway-api/blob/fcb4b5ce5caa87087e745f5db52214cfba4fce06/conformance/tests/httproute-invalid-cross-namespace.go#L43-L49)). 

I think there are 2 valid ways to represent this:

1. len(Route.Status.Parents) == 0
2. len(Route.Status.Parents) == 1 and the first entry has an Accepted condition set to false

Our tests currently only allow the first one, but they should allow both.

PR #1152, opened 2022-05-09
conformance: allow error status in Route

howardjohn:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1135

This changes tests to assert their is either no parent at all, or a
non-accepted parent (indicating an error message, presumably).

/kind bug

Fixes #

```release-note
NONE
```


conformance/utils/kubernetes/helpers.go
```diff
@@ -177,6 +177,41 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, gwName types.Name
 	return net.JoinHostPort(ipAddr, port), waitErr
 }
 
+// HTTPRouteMustHaveParents waits for the specified HTTPRoute to have either no parents
```
skriss, line 180:
```suggestion
// HTTPRouteMustHaveNoAcceptedParents waits for the specified HTTPRoute to have either no parents
```

skriss (commented):
LGTM other than one comment nit

conformance/utils/kubernetes/helpers.go
```diff
@@ -177,6 +177,41 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, gwName types.Name
...
+// or a single parent that is not acceptted. This is used to validate HTTPRoute errors.
```
robscott, line 181:
```suggestion
// or a single parent that is not accepted. This is used to validate HTTPRoute errors.
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -177,6 +177,41 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, gwName types.Name
...
+func HTTPRouteMustHaveNoAcceptedParents(t *testing.T, client client.Client, routeName types.NamespacedName, seconds int) {
+	t.Helper()
+
+	var actual []v1alpha2.RouteParentStatus
+	waitFor := time.Duration(seconds) * time.Second
+	waitErr := wait.PollImmediate(1*time.Second, waitFor, func() (bool, error) {
```
robscott, line 187: I'm not really sure what the solution is here, but this seems like something that could accidentally pass if a controller just hadn't gotten around to populating status yet. I hate to add delay for the sake of delay, but maybe we need to wait a few seconds before accepting `len(actual) == 0` as a successful result. If status is populated that seems sufficient to pass immediately, but not quite sure about the lack of conditions resulting in an immediate pass.

robscott (commented):
Thanks @howardjohn! Mostly LGTM, just a couple nits.

conformance/utils/kubernetes/helpers.go
howardjohn: added a delay only in this case

PR #1153, opened 2022-05-09
conformance: do not delay successes

howardjohn:
Currently WaitForConsistency requires 3 requests to succeed in a row,
and waits 1s between requests. This means each tests takes a minimum of
3s even in success. This is ok for now, but as we scale up may be quite
slow.

Instead, change the convergence logic so that we only delay on failures

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -147,9 +147,9 @@ func WaitForConsistency(t *testing.T, r roundtripper.RoundTripper, req roundtrip
 			return false
 		}
 
-		t.Logf("Request passed, ready!")
 		return true
 	})
+	t.Logf("Request passed")
 
 	return cReq, cRes
 }
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -23,8 +23,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


nathancoleman (approved):

conformance/utils/http/http.go
```diff
@@ -91,40 +89,67 @@ func MakeRequestAndExpectEventuallyConsistentResponse(t *testing.T, r roundtripp
...
 		}
 
-		numSuccesses++
-		if numSuccesses < threshold {
-			t.Logf("Request has passed %d times in a row of the desired %d, not ready yet", numSuccesses, threshold)
-			return false
-		}
-
-		t.Logf("Request has passed %d times in a row of the desired %d, ready!", numSuccesses, threshold)
+		t.Logf("Request passed, ready!")
```
skriss, line 150: nit: I probably wouldn't log "ready!" until after all the required requests have succeeded, so could either move this log statement to after `awaitConvergence` returns, or drop the "ready!" bit, or replace with logging within `awaitConvergence` that indicates how many requests have succeeded so far.

skriss (approved):
LGTM with one nit on logging

robscott:
Thanks!

/lgtm
/approve

PR #1154, opened 2022-05-09
conformance: allow converence time for gateway status check

howardjohn:
Without this, other tests may be creating valid Routes that are
attached. This test can run and then immediately fail due to those
routes from other tests.

Instead, we should assert eventual consistency.

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
+++ b/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
@@ -19,6 +19,7 @@ package tests
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


robscott:
> Without this, other tests may be creating valid Routes that are attached. This test can run and then immediately fail due to those routes from other tests.

@howardjohn are you referring to a state where the same Gateway is used in consecutive tests, and the attachment state of a Route from a previous test case has not been cleaned up before the next test starts?

howardjohn:
> > Without this, other tests may be creating valid Routes that are attached. This test can run and then immediately fail due to those routes from other tests.
> 
> @howardjohn are you referring to a state where the same Gateway is used in consecutive tests, and the attachment state of a Route from a previous test case has not been cleaned up before the next test starts?

Yes

skriss (approved):
LGTM, I believe we've run into this a couple times with Contour but I hadn't had a chance to dig into it yet.

robscott:
Thanks!

/lgtm
/approve

PR #1155, opened 2022-05-09
api: add RouteReasonNotAllowedByListeners

rainest:
**What type of PR is this?**
/kind feature

/kind api-change

Adds two `RouteConditionReason` constants for `RouteConditionAccepted==false`:
- `NotAllowedByListeners` indicates that none of the Gateway's Listeners have an `allowedRoutes` that permits the route.
- `NoMatchingListenerHostname` indicates that none of the Gateway's compatible Listeners have a Hostname that matches one of the route's hostnames (and implicitly that there is no compatible Listener without a Hostname, which would match any route).

Related to #1077 

```release-note
adds RouteReasonNotAllowedByListeners and RouteReasonNoMatchingListenerHostname RouteConditionReason constants
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -193,6 +193,10 @@ const (
 	//
 	// * "Accepted"
 	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NotAllowedByListeners"
+	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
@@ -201,6 +205,11 @@ const (
 	// This reason is used with the "Accepted" condition when the Route has been
 	// accepted by the Gateway.
 	RouteReasonAccepted RouteConditionReason = "Accepted"
+
+	// This reason is used with the "Accepted" condition when the route has not
+	// been accepted by a Gateway because the Gateway had no Listener whose
+	// allowedRoutes criteria permit the route
+	RouteReasonNotAllowedByListeners RouteConditionReason = "NotAllowedByListeners"
 
 	// This condition indicates whether the controller was able to resolve all
 	// the object references for the Route.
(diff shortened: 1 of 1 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -202,6 +207,15 @@ const (
 	// accepted by the Gateway.
 	RouteReasonAccepted RouteConditionReason = "Accepted"
 
+	// This reason is used with the "Accepted" condition when the route has not
+	// been accepted by a Gateway because the Gateway had no Listener whose
```
robscott, line 211: Tiny nit for consistent tense
```suggestion
	// been accepted by a Gateway because the Gateway has no Listener whose
```

robscott (commented):
Thanks @rainest!

/ok-to-test
/approve

rainest:
I had a brief discussion with @mikemorris re separating NotAllowedByListeners into separate constants for AllowedRoutes Kind and Namespace sections to be more specific and allow more targeted conformance tests, but I don't think we can do this since we're checking a route against all available Listeners, and can encounter scenarios like:

```
namespaceFoo/TCPRoute

ListenerA.allowedRoutes: (namespace selector for some label present on namespaceFoo), (kind UDPRoute)
ListenerB.allowedRoutes: (namespace selector that _does not_ match namespaceFoo), (kind TCPRoute)
```
where the answer to "was the route denied because of AllowedRoutes.Kinds or because of AllowedRoutes.Namespaces?" is "Both?".

Therefore, conformance tests should probably test that each of "only rejected because of Kind criteria", "only rejected because of (each type of) Namespace criteria", and mixed cases like the above result in the same NotAllowedByListeners.

Intuitively, I'd expect that Hostname checks, and receiving status NoMatchingListenerHostname, are performed after AllowedRoutes checks, and that routes should not expect to see NoMatchingListenerHostname if they also qualify for NotAllowedByListeners. However, I don't think this actually matters in practice, so conformance tests should _not_ test behavior when a route is neither allowed by AllowedRoutes nor matches any Hostnames. Conformance tests for NotAllowedByListeners should probably omit Hostnames and tests for NoMatchingListenerHostname should probably omit AllowedRoutes (or rather, provide Listeners with compatible protocols and the implied AllowedRoutes.Kinds) from test Listeners as such.

PR #1158, opened 2022-05-10
conformance: add test cases for Gateway->Secret ReferenceGrant

### LINKED ISSUE #1105: Conformance Tests: ReferencePolicy
This issue is intended to track conformance test development for ReferencePolicy. Comment below if you're interested in working on covering any of these areas.

**Extended Capabilities:**
- [ ] Gateway -> Secret (@nathancoleman)
  - [ ] Cross namespace reference not allowed by a ReferencePolicy
  - [ ] Cross namespace reference allowed by a ReferencePolicy
  - [ ] Multiple cross namespace references with only 1 allowed by a ReferencePolicy
- [x] Route -> Service (#1081)
  - [x] Cross namespace reference not allowed by a ReferencePolicy
  - [x] Cross namespace reference allowed by a ReferencePolicy
  - [x] Multiple cross namespace references with only 1 allowed by a ReferencePolicy

nathancoleman:
**What type of PR is this?**

/kind feature

Builds on #1081 to add test coverage for ReferenceGrant as it applied to cross-namespace `Gateway` -> `Secret` references.

Fixes #1105

```release-note
Add conformance tests for ReferenceGrant as it applies to cross-namespace Gateway->Secret references
```


conformance/base/manifests.yaml
```diff
@@ -350,3 +350,20 @@ spec:
...
...WFlBOG1aUGMwRlUKRVNWSXhNeFNoOEdQZWV1dHNlTVVNb2tXTmFqNEZSU2JRR1V1bWpSZHNoV0MrV3ZIQU9zV2pRSTJhbFA5Y3cxYgpDS05CbXpiRURKLytZalJ1bXZGYUo3TmtzUDZVRklLVXdJS1VmOWdOOWs5UXFpQlU1TWIxQ1lmZWJteEJtN3oyCjQ0b2JBZ01CQUFFd0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dJQkFJZS9iRzZZcVBIYmdHbG5WMUpTckpTcERVSVcKL0REdFhuclFudzFZYkhrV3J0VWwyczdicTZobSthNndobGNqOXJWTUhwZ0JDUEE2bHY5OWprbXEyL2hsNk5EeQp0ZFNmVGVmNWhtcjJOTG1xVzhRQUQrQm5rTjhpQ3JnV1krSHpHMW5PQWhaVGFINE93azgvVmZUYmFPMWhzVStQCjFnb3NyU3Q1Zk5SSGErQ3RGWWVPZHJSUFRQMUh5TDZDdFppV2R3THA3Y0dDWXZxY28ySjZOMzQ3ank3Nmx5dzUKNFc1Zm02RW5RaXFzZGo1VitYWmt6L3k4Nnp2ZUpPLzZMN1VmeE9BR09KclFETFNXLzUzSEFYcTN4RWhYbnd2aQpweWhISkpWVmxPclZWcEdxSlRlUDVya0VXOWJYNGtqdkZoQklvZXQyKy9leTQvdU5aanREdGtVWVlrZHAwaWdKClVabjlEeGV0ZzRxZEN1aDdCcFphdmJvTWJRNGdlTXNQckFRVXBrcVJmeVh3OGFNTXV0bk0zalV1dWpSN0FiWngKc2puYWJQOTdWY0MxRTdhenkraEJTM2pjTHlyTmdzdWd6REhXb1QyeHhSNXo3NEdSMFozSGdSQlRJWnRmYTJ4RwpZeDdpeWI4ZnFTMGZrQ3YyL1hxTVh0NisvUGdabGdhaDBBSXRoSlVkOGtoOE1OZmRwZURyQVNIQWF2eEFFTzNqCjMrWTZwZzVpVWg2OEE0SVlMOHNsSlNTZlE0NzZuaGpyekp6NU5DSHUzVC9yMVEwaVZqTGNjdEFrc1NWdTNlSHcKdzB1YTVKeG5pTDUzYmwxakVPTEV0QzFVS1RNeXR5Uy94M1I2UTNoSUlNR0dyNE5nV1VGdU9aRnJURWxkODNiVwpIa3JJS0JZTlY1Zy9nWmJnCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K
```
nathancoleman, line 368: This is a self-signed throwaway cert; however, it might trigger scanners somewhere. Any conventions for this that I don't know about?

conformance/utils/kubernetes/helpers.go
```diff
@@ -320,6 +320,7 @@ func conditionsMatch(t *testing.T, expected, actual []metav1.Condition) bool {
 func findConditionInList(t *testing.T, conditions []metav1.Condition, condName, condValue string) bool {
 	for _, cond := range conditions {
 		if cond.Type == condName {
+			// TODO(nathancoleman): Compare Reason
```
nathancoleman, line 323: Still need to address this before marking as ready for review. I suspect we have some tests today that _think_ the `Reason` is being asserted when it actually isn't. Not sure if this will cause some other unrelated tests to start failing.

conformance/base/manifests.yaml
```diff
@@ -19,7 +19,7 @@ metadata:
   name: same-namespace
   namespace: gateway-conformance-infra
 spec:
-  gatewayClassName: {GATEWAY_CLASS_NAME}
+  gatewayClassName: "{GATEWAY_CLASS_NAME}"
```
nathancoleman, line 22: We're still in a position where we need to patch an annotation onto the `Deployments` using `kubectl --kustomize`; however, this was previously always getting mangled as an object instead of a string. It's just a placeholder that gets replaced by the test runner [here](https://github.com/kubernetes-sigs/gateway-api/blob/5a69157e5ca3e7831834af4ca27e3f6ac26db612/conformance/utils/kubernetes/apply.go#L57).

nathancoleman (commented):
Personal review

skriss (commented):
Thanks @nathancoleman, had a couple comments from running this against Contour's implementation. Modulo the comments plus one other fix I need to make to Contour, was able to get these to pass.

conformance/base/manifests.yaml
robscott: I think we may want to create this in the Go test definition instead of a yaml manifest. Maybe we could have a function similar to what was used in ingress-controller-conformance? (https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/cec7cf25b0291933426e256065ef4afc981ea1fa/test/kubernetes/kubernetes.go#L207-L244)

conformance/tests/gateway-secret-reference-policy.go
```diff
@@ -0,0 +1,63 @@
...
+				}},
+				Conditions: []metav1.Condition{
+					{
+						Type:   string(v1alpha2.ListenerConditionReady),
+						Status: metav1.ConditionTrue,
+					},
+					{
+						Type:   string(v1alpha2.ListenerConditionResolvedRefs),
+						Status: metav1.ConditionTrue,
+					},
```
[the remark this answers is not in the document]
robscott: We haven't been clear about that, my initial thought was to only _require_ conditions like this when they represent an error state. With that said, that would make conformance tests more complex, and it doesn't hurt anything to have a ResolvedRefs condition present... Interested in what others think.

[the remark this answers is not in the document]
nathancoleman: If they're not currently required in a healthy state, I'm tempted to leave that contract as-is unless it becomes problematic one day. Can adjusts these tests accordingly

conformance/utils/kubernetes/helpers.go
nathancoleman: This change winds up being non-trivial as all of our test setup assertions break if you start asserting the `Condition.Reason`.

If no one feels strongly, I'd prefer addressing in a followup PR, understanding that it's a pre-existing hole in our test coverage.

Thoughts @robscott ?

robscott: This func only accepts a condition name and value as string right? So it would currently be impossible for any test to expect that a reason was also being checked. I think it's fine to leave checks on reason for a later point.

nathancoleman: Right. I went down the path of accepting a reason arg as well, but plumbing that all the way down without just creating a new set of `...WithReason` functions alongside what we have today is more scope expansion than I'd like in this PR.

conformance/base/manifests.yaml
nathancoleman: This turned out to be a good approach, thanks @robscott ! I did need to add a hook to install in-memory objects during `TestSuite.Setup` and have them clean up alongside the yaml manifests that we use for base resources today

conformance/utils/suite/suite.go
```diff
@@ -124,6 +124,10 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	t.Logf("Test Setup: Applying base manifests")
 	suite.Applier.MustApplyWithCleanup(t, suite.Client, suite.BaseManifests, suite.GatewayClassName, suite.Cleanup)
 
+	t.Logf("Test Setup: Applying programmatic resources")
+	secret := kubernetes.MustCreateSelfSignedCertSecret(t, "gateway-conformance-web-backend", "certificate", []string{"*"})
+	suite.Applier.MustApplyObjectsWithCleanup(t, suite.Client, []client.Object{secret}, suite.Cleanup)
```
nathancoleman, line 129: This is the first resource we're creating for testing that doesn't reside in a yaml file somewhere in the repo, so I needed to add this into the existing setup hook

conformance/utils/suite/suite.go
```diff
@@ -124,6 +124,10 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	t.Logf("Test Setup: Applying base manifests")
 	suite.Applier.MustApplyWithCleanup(t, suite.Client, suite.BaseManifests, suite.GatewayClassName, suite.Cleanup)
 
+	t.Logf("Test Setup: Applying programmatic resources")
+	secret := kubernetes.MustCreateSelfSignedCertSecret(t, "gateway-conformance-web-backend", "certificate", []string{"*"})
```
nathancoleman, line 128: The `*` hostname might stick out, but the changes made in this PR only cover ReferencePolicy. We may want tests in the future that cover the request-response when using a certificate; however, that is outside the scope of this PR.

nathancoleman:
@skriss @robscott please review latest commits when ya get a chance

conformance/utils/suite/suite.go
skriss, line 129: It might be nice to be able to attach these kinds of resource creates to the tests for which they're needed rather than having to put them all in the suite-level setup; could add `Setup`/`Teardown`-style functions to the `ConformanceTest` struct. I don't think this needs to block now, but could be a refactor to keep in mind if/when we add more of these.

nathancoleman:
@skriss I've added many different permutations for inapplicable ReferencePolicies to the test that already existed, asserting that none of them apply.

I've also added a new case w/ a ReferencePolicy granting access to all Secrets in the target namespace vs. a specific one. All are passing for Consul. Please confirm for Contour and refresh your review when you get a chance.

skriss:
> @skriss I've added many different permutations for inapplicable ReferencePolicies to the test that already existed, asserting that none of them apply.
> 
> I've also added a new case w/ a ReferencePolicy granting access to all Secrets in the target namespace vs. a specific one. All are passing for Consul. Please confirm for Contour and refresh your review when you get a chance.

Nice! All are passing for Contour. I'll take a look through the code shortly.

youngnick:
/lgtm

Nice work @nathancoleman!

nathancoleman:
@robscott who's responsible for providing the approval to merge?

robscott:
🤦 Somehow I'd thought Nick had approved and I was trying to do the other half.

/approve

PR #1162, opened 2022-05-14
Add HTTP Rewrite and Redirect Filter guide

### LINKED ISSUE #928: Document HTTPRoute Filter Compatibility
With GEP #726, we're introducing a filter that is incompatible with another existing one (Rewrite vs Redirect). This will likely not be the only time this happens. It would be helpful to add clear documentation about what filters we expect to be compatible with each other. At this point, it may be fair to say something like the following:

> All filters are expected to be compatible with each other except for the URLRewrite and RequestRedirect filters which may not be combined. If an implementation can not support other combinations of filters, they must clearly document that limitation. In all cases where incompatible or unsupported filters are specified, implementations MUST add a warning condition to status.

I believe @jpeach is already working on a standard status condition for that last point.

### LINKED ISSUE #1038: Document Path Redirects and Rewrites (GEP 726)
This GEP was implemented but has not been documented yet.

rainest:
**What type of PR is this?**
/kind documentation

Adds a guide for GEP-726 features, including the changes currently under review in #1124 

It is a WIP for two reasons:

- For HTTP->HTTPS redirects, are implementations expected to not redirect requests that are already HTTPS, or does the spec allow misconfigurations that result in redirect loops? If so, how should users properly set up their HTTPRoutes and Gateway Listener configuration to only match these redirect rules on HTTP requests? I expect this generalizes to path/hostname/etc. also, but this was the most prominent case I could think of.
- Do we care about path types for path modifiers? It seems that you can use prefix modifiers for exact paths, though it's probably pointless (if harmless), and I'm not sure whether RegularExpression paths would be considered prefixes here (i.e. ReplacePrefixMatch changes only the path segment matching the regex).
- I added the text suggested in #928 mostly as-is, but it raises the question "what status must we set?". It seems like we need a new `RouteConditionType` for filter-related statuses if we want this, or a new `RouteConditionReason` (it doesn't seem like it'd fit under any of the existing types unless we reject routes with invalid filters entirely, and I wouldn't expect to since they're per-rule and since the text indicates this should just be a warning, not an error).

As a side note, I have bad timing and will be out from 5/18-5/28, so it may make sense to use this as a starting point for someone else if review takes time or it needs significant changes. Edit: then I forgot this is the week that a bunch of people are at Kubecon :slightly_smiling_face: I should actually be able to check things on the 26th even though I'm still technically out.

Fixes #1038 
Fixes #928 

```release-note
NONE
```


site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+  name: http-filter-redirect
+spec:
+  hostnames:
+    - redirect.example
+  rules:
+    - filters:
+      - type: RequestRedirect
+	    requestRedirect:
+          scheme: https
+		  statusCode: 301
```
youngnick, line 36: Formatting fix needed here.

youngnick: I *think* maybe this is some tabs instead of spaces?

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+	    requestRedirect:
+          path:
+		    type: ReplaceFullPath
+			replaceFullPath: /paprika
+		  statusCode: 302
+      backendRefs:
+      - name: example-svc
+        weight: 1
+        port: 80
+```
```
robscott, line 87: Same comment as above - move to separate file.

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+portion matching `matches.path.value`. Changing the filter in the above to:
+
+```
+      - type: RequestRedirect
+	    requestRedirect:
+          path:
+		    type: ReplacePrefixMatch
+			replacePrefixMatch: /paprika
+		  statusCode: 302
+```
```
robscott, line 105: Indentation looks off here, maybe a combination of tabs and spaces going on here?

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+https://redirect.example/paprika/pinch` and `location:
+https://redirect.example/paprika/teaspoon` response headers.
+
+<!---
+Do these behave at all differently based on matches.path.type? I expect no, and
+that implementations should just always replace only the matching path prefix
+or the entire path. Using ReplacePrefixMatch on Exact is the same as
+ReplaceFullPath. RegularExpression looks a bit ambiguous, since the
+PathMatchType looks unclear on whether they're handled as prefixes or not.
+-->
```
robscott, line 117: Good point, I think https://github.com/kubernetes-sigs/gateway-api/issues/1166 will resolve this.

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+  rules:
+    - filters:
+      - type: URLRewrite
+	    urlRewrite:
+		  hostname: elsewhere.example 
+      backendRefs:
+      - name: example-svc
+        weight: 1
+        port: 80
+```
```
robscott, line 143: Same comment as above about needing a separate file.

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+ReplaceFullPath. RegularExpression looks a bit ambiguous, since the
+PathMatchType looks unclear on whether they're handled as prefixes or not.
+-->
+
+## Rewrites
+
+Rewrites modify components of a client request before proxying it upstream. A
+[`URLRewrite`
+filter](/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPURLRewriteFilter)
+can change the upstream request hostname and/or path. For example,
```
robscott, line 124: I think this might read better if you merged the content from before and after the example, something like this:
```suggestion
can change the upstream request hostname and/or path. For example, the following HTTPRoute will accept...
```

Same comment applies elsewhere.

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+	    urlRewrite:
+		  hostname: elsewhere.example 
+          path:
+		    type: ReplaceFullPath
+			replaceFullPath: /fennel
+      backendRefs:
+      - name: example-svc
+        weight: 1
+        port: 80
+```
```
robscott, line 175: Same comment as above.

robscott (commented):
Thanks @rainest, this is really great!

site-src/v1alpha2/guides/http-redirect-rewrite.md
rainest: Yeah, my editor was doing some :thinking: stuff there. The now separated examples should be using 4 space indents throughout.

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,180 @@
...
+hostname (`redirect.example`), path (`/cinnamon`), and port (implicit) remain
+unchanged.
+
+<!---
+The above raises some questions. Do redirects apply unconditionally? If my
+request is already HTTPS, will the above create a redirect loop? If redirects
+are unconditional (i.e. they _do not_ nothing and proxy upstream if the request
+already uses HTTPS), how should you properly write configuration elsewhere to
+ensure this only applies to HTTP requests?
+-->
```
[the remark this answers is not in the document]
rainest: Have added a comment indicating that redirects are noops and just proxy if their redirect URL would match the original input URL. I think that being inherent to implementations is good (there shouldn't be any case where you _want_ a loop, so config interpretation should just assume you don't). Is there somewhere else we should indicate that in the spec?

site-src/v1alpha2/guides/http-redirect-rewrite.md
rainest: Original in case the diff does something silly now that it's gone:

> Do these behave at all differently based on matches.path.type? I expect no, and that implementations should just always replace only the matching path prefix or the entire path. Using ReplacePrefixMatch on Exact is the same as ReplaceFullPath. RegularExpression looks a bit ambiguous, since the PathMatchType looks unclear on whether they're handled as prefixes or not.

I simply removed this without adding additional explanation since it looks like 1166's answer is just "you can't configure them on those other types". Do we think we'd need explanation of that here or is that too far off into the weeds?

rainest:
`examples/experimental/http-redirect-rewrite/httproute-redirect-https.yaml` is technically already valid under v1alpha2, but I expect it's easier to not split hairs and just call the whole thing experimental, assuming it'll imminently become not experimental and we'll move the whole directory back.

youngnick:
/lgtm

Nice work @rainest, this is very clear.

shaneutt (approved):
Looks like we're at a point with this one where we can remove `[WIP]` from the name. Thanks Travis for the effort on this one, lookg great.

/lgtm

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,88 @@
...
+301 response with a `location: https://redirect.example/cinnamon` header. The
+hostname (`redirect.example`), path (`/cinnamon`), and port (implicit) remain
+unchanged.
+
+Redirect filters do not return redirects if the redirected URL and original URL
+are the same. Although the above configuration applies to both HTTP and HTTPS
+requests, HTTPS requests will proxy upstream, not return a redirect response
+that would result in a loop.
+
+### Path redirects
```
robscott, line 39: I think this section needs an "Experimental Channel" disclaimer like what I have below:
```
!!! info "Experimental Channel"

    The `Port` field described below is currently only included in the
    "Experimental" channel of Gateway API. For more information on release
    channels, refer to the [related documentation](https://gateway-api.sigs.k8s.io/concepts/versioning).
```

site-src/v1alpha2/guides/http-redirect-rewrite.md
```diff
@@ -0,0 +1,88 @@
...
+
+```yaml
+{% include 'experimental/http-redirect-rewrite/httproute-redirect-prefix.yaml' %}
+```
+
+will result in redirects with `location:
+https://redirect.example/paprika/pinch` and `location:
+https://redirect.example/paprika/teaspoon` response headers.
+
+## Rewrites
```
robscott, line 65: This section also needs an "Experimental Channel" disclaimer.

examples/experimental/http-redirect-rewrite/httproute-rewrite.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: Mind adding backrefs like this here? https://github.com/kubernetes-sigs/gateway-api/blob/20683afe4018a42280d933971d44c258c328c964/examples/v1alpha2/basic-http.yaml#L1-L2

examples/experimental/http-redirect-rewrite/httproute-redirect-prefix.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: Mind adding backrefs like this here? https://github.com/kubernetes-sigs/gateway-api/blob/20683afe4018a42280d933971d44c258c328c964/examples/v1alpha2/basic-http.yaml#L1-L2

examples/experimental/http-redirect-rewrite/httproute-redirect-https.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: Mind adding backrefs like this here? https://github.com/kubernetes-sigs/gateway-api/blob/20683afe4018a42280d933971d44c258c328c964/examples/v1alpha2/basic-http.yaml#L1-L2

examples/experimental/http-redirect-rewrite/httproute-redirect-full.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: Mind adding backrefs like this here? https://github.com/kubernetes-sigs/gateway-api/blob/20683afe4018a42280d933971d44c258c328c964/examples/v1alpha2/basic-http.yaml#L1-L2

robscott (commented):
Thanks @rainest! Mostly LGTM, just a few small nits. On an unrelated note, if anyone wants to remove https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/css/extra.css, I would welcome that. The alignment that results in is particularly bad with this new page.

rainest:
@robscott do we indeed want the backref comments in examples? Those do actually render in the guide:

backref

The other examples don't include the same; their usage is just loosely indicated by the directory name, though I'm not sure others have the same quality where the guide is under v1alpha2 and the examples are under experimental. I removed the original commit from the PR, [backrefs.diff.txt](https://github.com/kubernetes-sigs/gateway-api/files/8855780/backrefs.diff.txt) if we want to re-add it.

robscott:
@rainest any chance you checked that after https://github.com/kubernetes-sigs/gateway-api/pull/1186 merged? I had to change the prefix characters to `#?` to avoid hiding all subheadings. If that doesn't work, I think it's fine to just leave that part out of this PR. This is a great improvement and would love to get it in.

rainest:
Alright, preview properly stripping the new header (albeit only after the second try for some reason), so this should be all good to go.

youngnick:
LGTM, I'll just approve though.

/approve

shaneutt (approved):
/lgtm

Issue #1103, opened 2022-04-08
Conformance Tests: HTTPRoute Filters


robscott:
This issue is intended to track conformance test development for HTTPRoute filters. Comment below if you're interested in working on covering any of these areas.

**Core Capabilities:**
- [x] Header modifier (@skriss)
  - [x] Set
  - [x] Add
  - [x] Remove
- [x] Request Redirect: (@LiorLieberman)
  - [x] Scheme
  - [x] Port Number

**Experimental Capabilities:**
- [x] Request Redirect:
  - [x] Path
- [x] Request Rewrite: (@LiorLieberman )
  - [x] Host
  - [x] Path

**Extended Capabilities:**
- [x] Request mirroring
- [ ] Combinations of the above with core capabilities

skriss:
I've submitted https://github.com/kubernetes-sigs/gateway-api/pull/1163 for the header modifier set/add/remove cases.

sunjayBhatia:
I think this issue should be augmented (or another created) to add testing for HTTPRoute Filters per-BackendRef as well, since the existing tests only test per-RouteRule

Part of that will include adding a test helper that can conditionally check a response has the correct content, given it was routed to a particular backend

robscott:
@LiorLieberman is working on conformance tests for HTTP Redirects.

LiorLieberman:
@robscott is request conformance for request mirroring still wanted? 

also proposing changes to the description

...

Extended Capabilities:
 Request Redirect: (@LiorLieberman)
   [x] Scheme
   [x] Port Number
   
Experimental  Capabilities:
 Request Redirect:
    [x] Path
 Request Rewrite: (@LiorLieberman )
   [x] host
   [x] Path
   
...

robscott:
Thanks @LiorLieberman! Updated the issue to match. We are still looking for tests covering request mirroring.

shaneutt:
At this point we only have one item left, talked with @robscott he's gonna close this and spin off the last one.

/assign @robscott

LiorLieberman:
What is the one item left? @shaneutt

robscott:
@LiorLieberman added https://github.com/kubernetes-sigs/gateway-api/issues/2037 to track the last part of this (combining HTTPRoute filters together in tests).

Going to close this one out in favor of that more specific one.

PR #1163, opened 2022-05-16
conformance: add request header modifier test

skriss:
**What type of PR is this?**
/area conformance

Adds a request header modifier test.

Updates #1103.

```release-note
NONE
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -29,14 +29,14 @@ import (
 
 // ExpectedResponse defines the response expected for a given request.
 type ExpectedResponse struct {
-	// ExpectedRequest defines the request that
-	// is expected to arrive at the backend.
-	ExpectedRequest ExpectedRequest
-
 	// Request defines the request to make.
-	// If not specified, ExpectedRequest.Request
-	// will be used as the request to make.
-	Request *Request
+	Request Request
+
+	// ExpectedRequest defines the request that
+	// is expected to arrive at the backend. If
+	// not specified, the backend request will be
+	// expected to match Request.
+	ExpectedRequest *ExpectedRequest
 
 	StatusCode int
 	Backend    string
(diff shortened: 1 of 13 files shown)


conformance/utils/http/http.go
```diff
@@ -28,19 +28,33 @@ import (
...
 type ExpectedResponse struct {
-	Request    ExpectedRequest
+	// Request defines the request to make.
+	Request Request
+
+	// BackendRequest defines the request that
+	// is expected to arrive at the backend. If
+	// not specified, it is assumed to be the
+	// same as the original request.
+	BackendRequest *Request
```
markmc, line 38: This could have been reversed so that the "make request" code isn't dealing with a struct that includes AbsentHeaders

To do that, ExpectedRequest  would embed Request, and adding AbsentHeaders. You'd always supply an ExpectedRequest, and it would be used for making the request if no Request was specified

But ... meh

markmc (commented):
Looks pretty comprehensive to me. A case insensitivity check could be a good addition?

/lgtm

skriss:
> Looks pretty comprehensive to me. A case insensitivity check could be a good addition?

:+1: added the case-insensitivity test case.

robscott:
It looks like there's a conflict here, so tests will fail initially, but hopefully this helps post rebase.

/ok-to-test

skriss:
Thanks @robscott, conflict fixed

skriss:
If possible, it'd be nice to get this merged before the v0.5.0 release.

conformance/utils/http/http.go
robscott: Agree that this structure would be a bit better if it's not too much trouble.

robscott:
@skriss this LGTM but agree with @markmc's suggestions above, specifically the adjustments to types and adding a check that covers case insensitive matching.

conformance/utils/http/http.go
skriss: OK, I can make this change, just noting that this makes all of the struct literals more verbose in the test cases due to the additional nesting (which I think is why I decided against this approach originally).

skriss: Will push a commit shortly that I think addresses the essence of this comment while keeping the test case struct literals simple for the normal cases.

skriss: Alright, the latest version splits out two structs, `Request` and `ExpectedRequest`, with the latter embedding the former and adding `AbsentHeaders`. `Request` is the required one, though, with `ExpectedRequest` defaulting to matching `Request` if not specified. I think this is the best of both worlds, PTAL.

skriss:
> and adding a check that covers case insensitive matching.

I added this in https://github.com/kubernetes-sigs/gateway-api/pull/1163/commits/d2990dd01d5634f54f9d42bac623b497acc30e94

skriss:
@robscott just a nudge here, the feedback has been addressed and would be great to have this included in v0.5.0 if possible, thanks!

robscott:
Sorry it took me so long to get back to this one! Took the time to verify that these tests pass locally for me.

/lgtm
/approve

robscott:
Looks like this conflicts with https://github.com/kubernetes-sigs/gateway-api/pull/1228. Will LGTM again once this is rebased.

skriss:
Rebased and fixed conflict, thanks @robscott!

robscott:
Seems like the first one didn't work...

/lgtm

PR #1170, opened 2022-05-20
conformance: fixes for invalid reference policy test

skriss:
**What type of PR is this?**
/kind bug
/area conformance

#1167 commented out a call to `t.Skip(...)`, however that `t.Skip(...)` call had been written in a misleading way; it was skipping _the rest_ of the test, not just the function passed to t.Skip(...) as an arg. So after commenting it out, the rest of the test now executes and uncovers a couple small issues with the test:

- fixes path prefix match YAML indentation
- removes check for HTTPRoute to be Accepted pending #1112

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-invalid-reference-policy.go
+++ b/conformance/tests/httproute-invalid-reference-policy.go
@@ -46,9 +46,6 @@ var HTTPRouteInvalidReferencePolicy = suite.ConformanceTest{
 		ns := v1alpha2.Namespace(gwNN.Namespace)
 		gwKind := v1alpha2.Kind("Gateway")
 
-		// TODO(mikemorris): Add check for Accepted condition once
-		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
-		// has been resolved
 		t.Run("Route status should have a route parent status with a ResolvedRefs condition with status False and reason RefNotPermitted", func(t *testing.T) {
 			parents := []v1alpha2.RouteParentStatus{{
 				ParentRef: v1alpha2.ParentReference{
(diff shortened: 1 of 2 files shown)


conformance/tests/httproute-invalid-reference-policy.go
```diff
@@ -90,7 +90,7 @@ var HTTPRouteInvalidReferencePolicy = suite.ConformanceTest{
 		// 	kubernetes.GatewayStatusMustHaveListeners(t, s.Client, gwNN, listeners, 60)
 		// })
 
-		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, s.Client, s.ControllerName, gwNN, routeNN)
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, s.Client, s.ControllerName, gwNN)
```
skriss, line 93: Dropped the check for the route to be Accepted here pending the TODO on L49-51 to be resolved.

mikemorris: Could you move the TODO from L49 down to here? I actually hadn't looked at the impl of `GatewayAndHTTPRoutesMustBeReady`, didn't realize it was actually checking the `Accepted` condition on routes.

mikemorris (approved):
Thanks for fixing this up, LGTM!

skriss:
> 🤦 re the t.Skip logic, was mixing up conventions from testing frameworks in other languages.

No worries, it looked right and we all missed it! I had been wondering why this test didn't start failing for us when the 503 -> 404 change was made, but this explains it now 😄

conformance/tests/httproute-invalid-reference-policy.go
skriss: :+1: moved and reworded slightly

robscott:
Thanks!

/lgtm
/approve

PR #1171, opened 2022-05-23
Adding validation to ensure prefix path match exists for path modifier

### LINKED ISSUE #1166: Add webhook validation to ensure that ReplacePathPrefix in rewrites has corresponding prefix match
When a URLRewrite filter is specified on a Route rule and it includes ReplacePathPrefix configuration, we should verify that exactly one path prefix match is specified on the same Route rule. We should also verify that no other path matches are specified as part of this.

robscott:
**What type of PR is this?**
/kind cleanup

This adds webhook validation to ensure that a path match exists when required by path modifier in filter.

Fixes #1166 

```release-note
Webhook validation now ensures that a path match exists when required by path modifier in filter.
```

### DIFF
--- a/apis/v1alpha2/validation/httproute.go
+++ b/apis/v1alpha2/validation/httproute.go
@@ -48,9 +48,9 @@ func ValidateHTTPRoute(route *gatewayv1a2.HTTPRoute) field.ErrorList {
 func validateHTTPRouteSpec(spec *gatewayv1a2.HTTPRouteSpec, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
 	for i, rule := range spec.Rules {
-		errs = append(errs, validateHTTPRouteFilters(rule.Filters, path.Child("rules").Index(i))...)
+		errs = append(errs, validateHTTPRouteFilters(rule.Filters, rule.Matches, path.Child("rules").Index(i))...)
 		for j, backendRef := range rule.BackendRefs {
-			errs = append(errs, validateHTTPRouteFilters(backendRef.Filters, path.Child("rules").Index(i).Child("backendsrefs").Index(j))...)
+			errs = append(errs, validateHTTPRouteFilters(backendRef.Filters, rule.Matches, path.Child("rules").Index(i).Child("backendsrefs").Index(j))...)
 		}
 		for j, m := range rule.Matches {
 			if m.Path != nil {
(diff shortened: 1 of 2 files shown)


apis/v1alpha2/validation/httproute.go
```diff
@@ -212,5 +212,23 @@ func validateHTTPPathModifier(modifier gatewayv1a2.HTTPPathModifier, path *field
 	if modifier.ReplacePrefixMatch == nil && modifier.Type == gatewayv1a2.PrefixMatchHTTPPathModifier {
 		errs = append(errs, field.Invalid(path, modifier.ReplacePrefixMatch, "must not be nil if the HTTPRouteFilter.Type is ReplacePrefixMatch"))
 	}
+
+	if len(errs) == 0 && modifier.ReplacePrefixMatch != nil {
+		if !hasExactlyOnePrefixMatch(matches) {
```
candita, line 217: Why not add this error to the list of errors?

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -328,6 +330,26 @@ func TestValidateHTTPRoute(t *testing.T) {
...
+				URLRewrite: &gatewayv1a2.HTTPURLRewriteFilter{
+					Path: &gatewayv1a2.HTTPPathModifier{
+						Type:               gatewayv1a2.PrefixMatchHTTPPathModifier,
+						ReplacePrefixMatch: utilpointer.String("foo"),
+					},
+				},
+			}},
+		}},
+	}, {
+		name:     "rewrite path modifier missing path match",
```
candita, line 351: Maybe add one more test for > 1 path match?

apis/v1alpha2/validation/httproute.go
youngnick: It reads to me like this adds the error to the list of errors in the case that there's not exactly one prefix match that's not replaced, which seems right. @robscott, maybe some comments here would help make it clearer?

apis/v1alpha2/validation/httproute_test.go
youngnick: +1 from me, that seems like a good one to add.

apis/v1alpha2/validation/httproute.go
robscott: Good call, this was confusing. I was actually using `len(errs) == 0` as a shortcut for a check on `modifier.Type == gatewayv1a2.PrefixMatchHTTPPathModifier` but that ended up being more confusing than it was worth. Updated to just use the latter check since it's much easier to understand.

apis/v1alpha2/validation/httproute_test.go
robscott: Added, thanks!

PR #1173, opened 2022-05-26
Clarifying that wildcard hostname matching is suffix matching

### LINKED ISSUE #1159: Wildcard hostname matching behavior is not clearly specified
As @howardjohn [noted in Slack](https://kubernetes.slack.com/archives/CR0H13KGA/p1652128690078659), our hostname matching behavior is not clearly specified. Other than saying "exact matches must be processed before wildcard matches", we don't really cover how wildcard matches should be implemented. Specifically, which of the following should `*.example.com` match?

1. `a.example.com` 
2. `a.b.example.com` 

We [recently added a conformance test](https://github.com/kubernetes-sigs/gateway-api/pull/1137) that stated that only 1 should match. On further thought, I'm not sure that's what we want. 

Here's my understanding of the default behavior for a variety of different products:

1. nginx defaults to match 1 and 2 [ref](https://nginx.org/en/docs/http/server_names.html#wildcard_names)
2. envoy defaults to match 1 and 2 [ref](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualhost)
3. haproxy can definitely support suffix matching (hdr_end), I'm not sure about the other [ref](https://cbonte.github.io/haproxy-dconv/2.6/configuration.html#7.3.6-req.hdr)
4. GCP can only support suffix matching (not well documented yet)

Given the above, it seems that most underlying implementations treat wildcard host matching as suffix matching as the default, and some may only be able to support that approach. My suggestion would be to adjust the spec and conformance tests to reflect that.

**Why this is needed**:
This is an important gap in our spec today.

robscott:
**What type of PR is this?**
/kind documentation
/area conformance

This is a follow up from #1159 that clarifies that wildcard hostname matching should be considered equivalent to suffix matching. This is a change from the Ingress spec but is more widely implementable and matches the most widespread default behavior, including nginx, envoy, GCP, and Azure.

Fixes #1159

```release-note
Wildcard hostname matching behavior has been documented as equivalent to suffix matching.
```

If merged, this will cause conformance tests to start failing for some implementations, adding a temporary hold so everyone has a chance to weigh in.

/hold

/cc @bowei @shaneutt @youngnick @mikemorris @howardjohn

### DIFF
--- a/conformance/tests/httproute-listener-hostname-matching.go
+++ b/conformance/tests/httproute-listener-hostname-matching.go
@@ -66,7 +66,15 @@ var HTTPRouteListenerHostnameMatching = suite.ConformanceTest{
 			Backend:   "infra-backend-v3",
 			Namespace: ns,
 		}, {
-			Request:    http.ExpectedRequest{Host: "too.many.prefixes.bar.com", Path: "/"},
+			Request:   http.ExpectedRequest{Host: "multiple.prefixes.bar.com", Path: "/"},
+			Backend:   "infra-backend-v3",
+			Namespace: ns,
+		}, {
+			Request:   http.ExpectedRequest{Host: "multiple.prefixes.foo.com", Path: "/"},
+			Backend:   "infra-backend-v3",
+			Namespace: ns,
+		}, {
+			Request:    http.ExpectedRequest{Host: "foo.com", Path: "/"},
 			StatusCode: 404,
 		}, {
 			Request:    http.ExpectedRequest{Host: "no.matching.host", Path: "/"},
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -172,6 +172,10 @@ type Listener struct {
 	// accepted. For more information, refer to the Route specific Hostnames
 	// documentation.
 	//
+	// Hostnames that are prefixed with a wildcard label (`*.`) are interpreted
+	// as a suffix match. That means that a match for `*.example.com` would match
+	// both `test.example.com`, and `foo.test.example.com`, but not `example.com`.
+	//
 	// Support: Core
 	//
 	// +optional
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -74,8 +74,13 @@ type HTTPRouteSpec struct {
 	// * A Listener with `*.example.com` as the hostname matches HTTPRoutes
 	//   that have either not specified any hostnames or have specified at least
 	//   one hostname that matches the Listener hostname. For example,
-	//   `test.example.com` and `*.example.com` would both match. On the other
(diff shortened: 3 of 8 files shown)


youngnick:
As I said in #1159, although I would slightly prefer to keep this similar to a single label, it's not worth excluding already-working implementations from conformance (which would be the cost of not making this change).

So, this LGTM to me, but interested to hear from others.

howardjohn:
I would slightly prefer to keep it a single label [and have someone make Envoy support single labels efficiently].

robscott:
> I would slightly prefer to keep it a single label

Although I agree that it would be nice to match the Ingress spec and more closely align with TLS certs, I think there's likely a compelling reason that so many APIs/implementations (envoy, nginx, GCP, Azure, etc) chose to interpret this as suffix matching. The fact that some of those can _only_ implement suffix matching makes me think it's our only real option here.

conformance/tests/httproute-listener-hostname-matching.go
```diff
@@ -66,8 +66,9 @@ var HTTPRouteListenerHostnameMatching = suite.ConformanceTest{
 			Backend:   "infra-backend-v3",
 			Namespace: ns,
 		}, {
-			Request:    http.ExpectedRequest{Host: "too.many.prefixes.bar.com", Path: "/"},
-			StatusCode: 404,
+			Request:   http.ExpectedRequest{Host: "multiple.prefixes.bar.com", Path: "/"},
+			Backend:   "infra-backend-v3",
+			Namespace: ns,
```
shaneutt, line 71: I might be reading this wrong: this test proves the change by making sure the `*` match works right? Are we missing the test for the assertion that the suffix without a prefix doesn't match?

e.g. we need these tests:

- `*.ocean.example.com` matches `whales.ocean.example.com` and `sharks.ocean.example.com`
- `*.ocean.example.com` does **not** match `ocean.example.com`

shaneutt (commented):
Overall I don't have a problem with the approach, it may not be ideal for our implementation but it's implementable and if it's the thing that _everyone_ can easily implement then that's fine.

I do have a small comment/question about the testing for this, but it's possible I'm just missing something.

conformance/tests/httproute-listener-hostname-matching.go
robscott: Good call, added test case for that. On a related note - I'm a big fan of those example domains. Trying to keep the surface area of this test small, but if anyone wants to move our tests to ocean themed example.com/net/org domains, that would also be great.

shaneutt: fits right in with our already nautical software themes :joy:

robscott:
Also ran this by @mikemorris to confirm this made sense for their implementation. I think there's enough consensus here to remove the hold - still need someone to add an LGTM though.

/hold cancel

PR #1174, opened 2022-05-27
docs: add note on RFC2119 keyword semantics

markmc:
We should probably explain that these keywords have precise semantics, and we're not just shouting :smiley:

/kind documentation


youngnick:
Agreed.

/approve

I'll let someone else lgtm.

site-src/concepts/api-overview.md
```diff
@@ -2,6 +2,12 @@
 
 This document provides an overview of Gateway API.
 
+!!! note
+    The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
+    NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
+    "OPTIONAL" in this documentation are to be interpreted as
+    described in RFC 2119.
```
robscott, line 9: I think this is a great addition, but I'm not sure it's quite the right place. I think API overview and anything else in "concepts" should ideally be focused on users of the API, and the specifics of this RFC are probably not that important to that audience. I do think this is highly relevant to implementors of the API though, so maybe move this to "Implementation Guidelines"? Now that I'm looking at that page, it looks like it needs some help, but in general that feels like the right place for this kind of content.

As far as how it's worded, we may want to clarify that this applies throughout our docs and spec, I'd consider that optional though:

>Throughout the Gateway API documentation and specification, keywords such as "MUST", "MAY", and "SHOULD" are used broadly. These should be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).

robscott (commented):
Thanks @markmc!

site-src/concepts/api-overview.md
markmc: > I think this is a great addition, but I'm not sure it's quite the right place. I think API overview and anything else in "concepts" should ideally be focused on users of the API, and the specifics of this RFC are probably not that important to that audience.

I agree, and personally I'm not a fan of using these keywords except where we're trying to give super specific guidance to implementors.

However ... we do use the keywords a bunch of times on this page?

> I do think this is highly relevant to implementors of the API though, so maybe move this to "Implementation Guidelines"? Now that I'm looking at that page, it looks like it needs some help, but in general that feels like the right place for this kind of content.
> 
> As far as how it's worded, we may want to clarify that this applies throughout our docs and spec, I'd consider that optional though:
> 
> > Throughout the Gateway API documentation and specification, keywords such as "MUST", "MAY", and "SHOULD" are used broadly. These should be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).

I like that :thumbsup:

So, remove the keywords from API overview page and add this note to implementation guidelines?

youngnick: > So, remove the keywords from API overview page and add this note to implementation guidelines?

I approved earlier, but this sounds like it will be much clearer.

markmc: Ok, done now. I've just downcased those keywords for now. The doc could probably do with a bit more work to make it more user-facing than implementor-facing

PR #1176, opened 2022-05-27
Fix certificateRefs optionality.

briantkennedy:
The GatewayTLSConfig CertificateRefs member is a slice of pointer to struct which implies that the struct is optional.  Change this to a slice of struct to better fit with API type conventions.

/kind cleanup


youngnick:
/ok-to-test

The slice-of-pointers does seem redundant, agreed.

/lgtm

PR #1177, opened 2022-05-31
docs: backlinks from example manifests to guides

### LINKED ISSUE #446: Add backlinks from example manifests to guides

markmc:
Fixes #446

Based on #1083

We would like our example YAML files to include comments that do not show up in the guides where the files are used as jinja templates, while still ensuring they are valid YAML files.

Configuring the jinja Environment with "line_comment_prefix" allows you to control the syntax of the jinja comment directive. If we set that to "##", then such lines are ommitted during jinja templating, and are treated as YAML comments.

See also https://jinja.palletsprojects.com/en/2.11.x/templates/#line-statements

Generated the comments using this awfulness:

```
$ sed -i '1s/^/## Used in:\n/' $( \
    for iii in $(git grep -l '{% include' site-src/); do \
      awk -F "'" '/{% include/ { print "examples/"$2 }' $iii ; \
    done | sort | uniq)

$ for iii in $(git grep -l '{% include' site-src/); do \
    for jjj in $(awk -F "'" '/{% include/ { print "examples/"$2 }' $iii); do \
      sed -i "2s|^|## - $iii\n|" $jjj; \
    done; \
  done
```

/kind documentation

### DIFF
--- a/examples/v1alpha1/basic-http.yaml
+++ b/examples/v1alpha1/basic-http.yaml
@@ -1,3 +1,5 @@
+## Used in:
+## - site-src/v1alpha1/api-types/httproute.md
 apiVersion: networking.x-k8s.io/v1alpha1
 kind: GatewayClass
 metadata:
--- a/examples/v1alpha1/basic-tcp.yaml
+++ b/examples/v1alpha1/basic-tcp.yaml
@@ -1,3 +1,5 @@
+## Used in:
+## - site-src/v1alpha1/guides/tcp.md
 apiVersion: networking.x-k8s.io/v1alpha1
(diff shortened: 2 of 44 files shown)


robscott:
This is brilliant, thanks @markmc! I'd love to see some automation around adding these comments in the future, but this is already a great upgrade as is.

/lgtm
/approve

PR #1179, opened 2022-06-01
Renaming ReferencePolicy to ReferenceGrant

robscott:
**What type of PR is this?**
/kind api-change

Renames ReferencePolicy to ReferenceGrant. Renaming this resource was originally suggested in the [community meeting on Jan 24](https://docs.google.com/document/d/1eg-YjOHaQ7UD28htdNxBR3zufebozXKyI28cl2E11tU/edit#bookmark=id.2mgmr9ue1p10), and then revisited in several additional meetings. I ran a [poll in March](https://kubernetes.slack.com/archives/CR0H13KGA/p1646244022335919) at which point "ReferenceGrant" was the most popular name. I think this has largely been accepted as something that will happen, but no one has had the time to actually make the change. With v0.5.0 around the corner, it seems like we should try to fit this change in if possible.

Getting this change in would also help with a [KEP that is relying on this resource](https://github.com/kubernetes/enhancements/pull/3295/files#r886082664) and is concerned that a rename at a later point could result in confusion/instability. 

```release-note
ReferencePolicy has been renamed to ReferenceGrant
```


robscott:
Adding a temporary hold so everyone has a chance to add feedback, will remove this sometime tomorrow.

/cc @shaneutt @youngnick @howardjohn @mikemorris 
/hold

youngnick:
Looks like this was a straightforward search-and-replace, I agree with the ReferenceGrant name, and this LGTM.

/lgtm

shaneutt (approved):
/lgtm

site-src/geps/gep-709.md
```diff
@@ -32,17 +32,17 @@ emerge.
 
 To ensure that Gateway API is able to safely provide this functionality, we need
 to enforce a handshake mechanism that requires resources in both namespaces to
-agree to this reference. To accomplish that, a new ReferencePolicy resource
+agree to this reference. To accomplish that, a new ReferenceGrant resource
 should be introduced.
 
-![Reference Policy](images/709-referencepolicy.png)
+![Reference Policy](images/709-referencegrant.png)
```
mikemorris, line 38: I think this image source needs to be renamed too.

site-src/v1alpha2/api-types/referencegrant.md
```diff
@@ -1,35 +1,39 @@
-# ReferencePolicy
+# ReferenceGrant
 
-A ReferencePolicy can be used to enable cross namespace references within
+!!! note
+    This resource was originally named "ReferencePolicy". It was renamed
+    to "ReferenceGrant" to avoid any confusion with policy attachment.
```
mikemorris, line 5: Should a note like this be added to GEP-709 too?

mikemorris (approved):
/lgtm

site-src/geps/gep-709.md
robscott: Updated, thanks!

site-src/v1alpha2/api-types/referencegrant.md
robscott: Updated, thanks!

robscott:
Thanks for the reviews @mikemorris, @shaneutt, and @youngnick! Updated to address feedback, PTAL.

robscott:
Think we've got enough consensus around this, will let next LGTM merge this in.

/hold cancel

PR #1181, opened 2022-06-03
Update ReferenceGrant docs to include Gateway -> Secret use case

nathancoleman:
**What type of PR is this?**

/kind documentation

The docs for ReferenceGrant in some cases only cover the route -> backend use case where they should also include the gateway -> secret use case for certificate references.

Fixes #

```release-note
Update ReferenceGrant docs to include Gateways that reference a Secret in a different namespace
```


youngnick:
Nice catch @nathancoleman!

/approved
I'll leave the LGTM for a second review.

shaneutt (approved):
/lgtm

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -95,6 +95,7 @@ type ReferencePolicyFrom struct {
 	// additional resources, the following Route types are part of the "Core"
 	// support level for this field:
 	//
+	// * Gateway
```
robscott, line 98: I think we need some more nuance here. I'm not really sure how this will scale since we really need to document each valid pair, ie HTTPRoute -> Secret doesn't make sense, and there's not currently any way to do Gateway -> Service. I think the best option here is to add a note here that ReferencePolicy is only supported as explicitly documented on object reference fields. Then we can add a comment on the cert ref namespace field in Gateways that references to Secrets in other namespaces can be allowed by ReferenceGrant. Similarly, we can update the BackendRef namespace field to state that references to Services in other namespaces can be allowed by ReferenceGrant.

robscott (commented):
Thanks for the work on this @nathancoleman! I agree with the goal here, but I think it may require a bit more restructuring.

apis/v1alpha2/referencepolicy_types.go
mikemorris: The status Condition -> Reason mappings (see example in https://github.com/kubernetes-sigs/gateway-api/pull/1114) may provide a pattern for clarifying these relationships a bit.

robscott:
Thanks @nathancoleman!

/lgtm
/approve

PR #1189, opened 2022-06-07
Clarifying how ReplacePrefixMatch works

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This is a follow up from the [comments](https://github.com/kubernetes-sigs/gateway-api/pull/1086#discussion_r883131526) by @thockin and @aojea. Although I think we've clearly defined how the PathPrefix match works, we haven't been as clear with the corresponding path rewrite. This copies some of the text we have for prefix path matching so it's also clearly documented here.

```release-note
NONE
```


aojea:
lgtm

youngnick:
This LGTM, not sure if I should hit the go button though.

shaneutt (approved):
Makes sense to me.

/lgtm

thockin (commented):
LGTM

Issue #1183, opened 2022-06-06
CHANGELOG for v0.5.0


shaneutt:
**What would you like to be added**:

We need a CHANGELOG for `v0.5.0`.

**Why this is needed**:

We can not release `v0.5.0` otherwise.

PR #1193, opened 2022-06-08
docs: update PR template for CHANGELOG reminder

shaneutt:
**What type of PR is this?**

/kind documentation

While working on https://github.com/kubernetes-sigs/gateway-api/issues/1183 I found it was quite laborious to search all PR history and fill in the `CHANGELOG.md` retroactively.

This PR is a proposal instead to indicate that user-facing changes must be accompanied by a `CHANGELOG.md` update in the PR they are introduced, adding a PR checklist to the PR template. Consequently, this also proposes that we keep a heading of `Unreleased` in the CHANGELOG as a place for contributors to place their entries prior to release to reduce confusion about where it should go.


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
 - [v0.1.0-rc2](#v010-rc2)
 - [v0.1.0-rc1](#v010-rc1)
 
+## Unreleased
+
+Changes that are in `main` but have not yet made it to a release.
```
robscott, line 19: Maybe add some kind of note here to indicate that unreleased changes may be rolled back or changed prior to a release.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -51,3 +51,7 @@ If yes, please enter a release note below:
 ```release-note
 
 ```
+
+**PR Checklist**
+
+- [ ] the `CHANGELOG.md` was updated with any user-facing changes
```
robscott, line 57: If we're doing this, we probably want to remove the release-note section above

robscott (commented):
Thanks @shaneutt! I like this idea. I'm a bit concerned that we may end up with a lot of merge conflicts on CHANGELOG.md if we take this approach, but it could be worth trying to see if that really is an issue.

CHANGELOG.md
shaneutt:
```suggestion
Changes that are in `main` but have not yet made it to a release.

**NOTE**: unreleased changes are subject to rollback or change prior to landing in a release
```

shaneutt: Sounds good. Change in 1e06e53e677c12d76f8fe9ec6570ffea61908737

.github/PULL_REQUEST_TEMPLATE.md
shaneutt: :+1: Changed in 1e06e53e677c12d76f8fe9ec6570ffea61908737

shaneutt:
@youngnick any feelings on this one?

youngnick:
In Contour, we use a `changelogs` directory, with an `unreleased` directory below that, and each PR has to include a separate .md file with the changelog in it. This avoids conflicts in a singular changelog file, but makes the release process pretty easy. Ideally, we can make it so your PR won't go green without the changelog present (in Contour we use Actions, but we'd need to figure out a way to add this to Prow's build, I think).

shaneutt:
@youngnick, @robscott just wanted to ping on this one because it's been sitting a while: think we're OK to give this one a try?

robscott:
Actually before we go too far down this path, maybe we should try the upstream scripting for release notes generation: https://github.com/kubernetes/release/blob/master/cmd/release-notes/README.md.

shaneutt:
Alright, that goes a bit beyond the scope of what I was trying to accomplish here, so I'll consider this one closed for the time being and then we should probably create a follow up issue for that :+1:

PR #1194, opened 2022-06-08
Change ReferenceGrant shortname from "refpol" to "refgrant"

caboteria:
I couldn't think of an abbreviation for "grant" so I went with the
whole word.

**What type of PR is this?**
/kind cleanup
/kind api-change

In the rename from ReferencePolicy->ReferenceGrant the shortname wasn't changed so it's no longer an abbreviation of the full name. This is confusing to users because what's a "pol"? It's annoying to developers since the shortname conflict prevents both objects from co-existing.

```release-note
The ReferenceGrant shortname is "refgrant".
```


robscott:
🤦 thanks @caboteria!

/ok-to-test
/approve

PR #1192, opened 2022-06-08
recopy the v1beta1 types

mehabhalodiya:
Signed-off-by: mehabhalodiya [mehabhalodiya@gmail.com](mailto:mehabhalodiya@gmail.com)

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
As suggested by @robscott, this is a follow-up PR on one of our last steps before the v1beta1 (v0.5.0) release.
The key is that we’d need `GatewayClass`,` Gateway`, and `HTTPRoute` type definitions in `apis/v1beta1` in addition to the alpha directories they already exist in.

**Which issue(s) this PR fixes**:

Fixes #1041 

**Does this PR introduce a user-facing change?**:

```release-note
Introduce v1beta1 types for resources graduating to beta
```

PR #1197, opened 2022-06-09
Generating CRDs and client code for v1beta1, removing v1alpha1

robscott:
**What type of PR is this?**
/kind cleanup

This PR makes a number of changes, primarily related to generating code for v1beta1 and removing v1alpha1 types:

- Removes v1alpha1 docs and types
- Updates CRDs to include v1beta1 where applicable
- Removes CRDs that haven't graduated to beta from standard channel
- Updates v1beta1 types to strictly match v1alpha2 types (there were a couple more recent changes that were missed in #1192)
- Updates webhook to validate v1beta1 (in addition to v1alpha2)
- Copies v1alpha2 examples to v1beta1 for verification

```release-note
NONE
```


shaneutt (changes requested):
I think we're pretty close to good, just a couple things:

The README is still linking directly to `v1alpha1`:

```
[Go docs for the package](https://pkg.go.dev/sigs.k8s.io/gateway-api/apis/v1alpha1)
```

Looks like we should update that to `v1beta1`.

Also looks like HAProxy refers to specifically having `v1alpha1` support in `implementations.md`:

```
HAProxy Ingress v0.13 partially supports the Gateway API's v1alpha1 specification. See the [controller's Gateway API documentation][h2] to get informed about conformance and roadmap.
```

Seems like we might want to give them a heads up, perhaps this is just out of date?

Other than that, I see a couple other references but those look like they can stay, historical stuff like blog posts and geps, e.t.c.

robscott:
Good catches @shaneutt, thanks!

>The README is still linking directly to v1alpha1:

🤦 updated to v1alpha2 now (should eventually be v1b1).

>HAProxy Ingress v0.13 partially supports the Gateway API's v1alpha1 specification. See the [controller's Gateway API documentation][h2] to get informed about conformance and roadmap.

@jcmoraisjr may have an update here

shaneutt (approved):
So there are a lot of changes in a single commit, but I looked through this and minus the things I pointed out above, I think we're good. :+1: 

/lgtm

apis/v1beta1/gateway_types.go
```diff
@@ -24,7 +24,6 @@ import (
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:categories=gateway-api,shortName=gtw
 // +kubebuilder:subresource:status
-// +kubebuilder:storageversion
```
youngnick, line 27: Just to confirm, we're leaving the storage version at v1alpha2 for now, and will migrate it later? This makes sense to me, but is probably worth calling out.

youngnick (commented):
This LGTM, with one small comment.

I did realize when reviewing this though, that I think that, if you only install the `standard` channel CRDs, I don't think conformance will pass because ReferenceGrant is not present?

PR #1198, opened 2022-06-09
[WIP] conformance: add local filesystem path support to getContentsFromPathOrURL helper

### LINKED ISSUE #1180: conformance: allow reading BaseManifests from non-embedded local filesystem

Update [`getContentsFromPathOrURL`](https://github.com/kubernetes-sigs/gateway-api/blob/d1b86648e2771763d5f782bfd762c33936034fd2/conformance/utils/kubernetes/apply.go#L201-L237) to support reading files from the local filesystem rather than only the [embedded filesystem](https://github.com/kubernetes-sigs/gateway-api/blob/master/conformance/embed.go).

**Why this is needed**:
We're currently working on switching our implementation of Gateway API conformance tests from a local git checkout to re-implementing [conformance_test.go](https://github.com/kubernetes-sigs/gateway-api/blob/master/conformance/conformance_test.go) directly (in order to specify `SupportedFeatures`, which IIUC is a recommended pattern), but we're currently blocked due to our need to make some slight tweaks to the base manifests (lowering the number of replicas to work around resource constraints in CI and adding an annotation to all Deployments), which we have previously been doing with a `kustomize` patch.

Specifying a string path for [`BaseManifests`](https://github.com/kubernetes-sigs/gateway-api/blob/d1b86648e2771763d5f782bfd762c33936034fd2/conformance/utils/suite/suite.go#L71) _is_ currently supported as an option when initializing a `ConformanceTestSuite` struct, but unless the string is a remote path beginning with `https://`, it is hardcoded to [only read from the read-only embedded filesystem](https://github.com/kubernetes-sigs/gateway-api/blob/d1b86648e2771763d5f782bfd762c33936034fd2/conformance/utils/kubernetes/apply.go#L232).

I've managed to read the base manifests from the embedded filesystem, patch them with [`krusty`](https://pkg.go.dev/sigs.k8s.io/kustomize/api/krusty) and write them to a local temp file with [`kyaml/filesys`](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml/filesys), but would need to add support for reading from the local filesystem to be able to use the patched manifests in this way - is this something we could consider supporting, or is there a different way I should be approaching this problem?

mikemorris:
**What type of PR is this?**
/kind feature

Consul API Gateway currently needs to slightly tweak the conformance tests base manifests with kustomize (to add an annotation to Deployments and reduce replicas to 1 due to resource limitations in GitHub Actions). Specifying a string path for [`BaseManifests`](https://github.com/kubernetes-sigs/gateway-api/blob/d1b86648e2771763d5f782bfd762c33936034fd2/conformance/utils/suite/suite.go#L71) _is_ currently supported as an option when initializing a `ConformanceTestSuite` struct, but unless the string is a remote path beginning with `https://`, it is hardcoded to [only read from the read-only embedded filesystem](https://github.com/kubernetes-sigs/gateway-api/blob/d1b86648e2771763d5f782bfd762c33936034fd2/conformance/utils/kubernetes/apply.go#L232).

This is currently marked as WIP as it may need some additional input validation checks.

Fixes #1180 

```release-note
conformance: allow configuring BaseManifests from local filesystem with "local://" path prefix
```

### DIFF
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 	"testing"
 	"time"
@@ -228,6 +229,15 @@ func getContentsFromPathOrURL(location string) (*bytes.Buffer, error) {
 			return nil, fmt.Errorf("received %d bytes from %s, expected %d", count, location, resp.ContentLength)
 		}
 		return manifests, nil
+	} else if strings.HasPrefix(location, "local://") {
+		_, path, _ := strings.Cut(location, "local://")
+
+		b, err := os.ReadFile(path)
+		if err != nil {
+			return nil, err
+		}
+
+		return bytes.NewBuffer(b), nil
 	}
 	b, err := conformance.Manifests.ReadFile(location)
 	if err != nil {


youngnick:
I think the URL-style approach is fine, and we can just document that unprefixed filenames are assumed to be in the embedded filesystem. I added some commends on #1180 about what validation I think we should do for the filenames.

robscott:
@mikemorris is this still something you want to get in? It looks like it's still in WIP state, is that intentional? Does John's [related PR](https://github.com/kubernetes-sigs/gateway-api/pull/1069) help here?

mikemorris:
@robscott I just wasn't quite sure how to best implement the safeguards mentioned in https://github.com/kubernetes-sigs/gateway-api/issues/1180#issuecomment-1160983779
> This is currently marked as WIP as it may need some additional input validation checks.

#1069 is the limitation I'm trying to get around - only the files in `tests` and `base` are included in the embedded filesystem when using gateway-api as a library to run the conformance suite.

PR #1199, opened 2022-06-09
deps: update k8s.io modules to v0.24

mikemorris:
**What type of PR is this?**
/kind cleanup
/kind api-change

Updates the k8s.io Go modules from v0.22 to v0.24. Not strictly an API change, but wanted to somehow flag that this is a breaking change due to Go module conflicts.

This would be worth doing this now, before the v0.5.0 release, due to the dependency change from `github.com/googleapis/gnostic` to `github.com/google/gnostic` - this was handled in https://github.com/kubernetes/kubernetes/pull/108994 to resolve the dependency conflict, but requires updating all dependencies in lockstep, and will likely cause a similar cascading impact for any Gateway API implementations.

When running the conformance tests with both `google/gnostic` and `googleapis/gnostic` in the `go.mod` dependency tree, the following panic happens:
```
panic: proto: file "extensions/extension.proto" is already registered
        previously from: "github.com/google/gnostic/extensions"
        currently from:  "github.com/googleapis/gnostic/extensions"
See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```

The `sigs.k8s.io/controller-runtime` update is necessary to adapt to the breaking change in https://github.com/go-logr/logr/pull/42 hit due to the dependency bump to `github.com/go-logr/logr` from updating the Kubernetes libraries to v0.24

```release-note
BREAKING: update to k8s.io v0.24 modules, requires a lockstep update for most uses of k8s.io modules in Gateway API implementations due to google/gnostic module fork
```


robscott:
Thanks @mikemorris!

/lgtm

PR #1204, opened 2022-06-13
Replacing webhook validation with CRD validation for AddressType

### LINKED ISSUE #1202: Use CRD based RegEx validation instead of webhook where possible
I believe this is limited to AddressType validation right now, but one of the final bits of feedback from @thockin on #1086 was that if it was possible to validate something with CRD validation, we should do that. Here's the rationale:

* If something can be done with CRD validation, it should, webhook should be last resort
* It seems unlikely that we'd frequently be adding additional address types
* We should be able to reuse one of the regexes we're already using for domain prefixed path ([GatewayController](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1beta1/shared_types.go#L448)) and add a `IPAddress|Hostname|` prefix

robscott:
**What type of PR is this?**
/kind cleanup

In response to feedback on #1086, this moves validation from the webhook into the CRD.

Fixes #1202

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -487,12 +487,13 @@ type AnnotationValue string
 //
 // Values `IPAddress` and `Hostname` have Extended support.
 //
-// All other values, including domain-prefixed values have Custom support,
-// which are used in implementation-specific behaviors.
+// All other values, including domain-prefixed values have Custom support, which
+// are used in implementation-specific behaviors. Support for additional
+// predefined CamelCase identifiers may be added in future releases.
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
-// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9])+$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\/[a-zA-Z0-9]+$`
+// +kubebuilder:validation:Pattern=`^Hostname|IPAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type AddressType string
 
 const (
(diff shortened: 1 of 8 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -487,12 +487,13 @@ type AnnotationValue string
...
-// All other values, including domain-prefixed values have Custom support,
-// which are used in implementation-specific behaviors.
+// All other values, including domain-prefixed values have Custom support, which
+// are used in implementation-specific behaviors. Support for additional
+// predefined CamelCase identifiers may be added in future releases.
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
-// +kubebuilder:validation:Pattern=`^([a-zA-Z0-9])+$|^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])\/[a-zA-Z0-9]+$`
+// +kubebuilder:validation:Pattern=`^Hostname|IPAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
```
robscott, line 496: Note: This is copied from the ControllerName regex with `Hostname|IPAddress|` added as a prefix. https://regex101.com/r/k892is/1 for reference.

apis/v1beta1/shared_types.go
```diff
@@ -480,6 +480,20 @@ type AnnotationKey string
 type AnnotationValue string
 
 // AddressType defines how a network address is represented as a text string.
+// This may take two possible forms:
+//
+// * A predefined CamelCase string identifier (currently limited to `IPAddress` or `Hostname`)
```
shaneutt, line 485: Just because these types of references with point-in-time statements of facts about other symbols in the code run the risk of bit-rotting and being missed when other things change, would it be worth making this change?

```suggestion
// * A predefined CamelCase string identifier (e.g. `IPAddress`, `Hostname`)
```

shaneutt (approved):
No blockers, but one potential comment change for your considerations.

/lgtm

apis/v1beta1/shared_types.go
robscott: I definitely agree that it's a risk that they get out of sync, but I prefer this to the alternative of requiring users to parse a difficult to understand regex to understand which values are currently valid. We do need to be very careful with all reviews around enums like this that when we change the set of allowed values, we update that _everywhere_.

PR #1205, opened 2022-06-13
Adding back ReferencePolicy with a note that it will be removed in v0.6.0

### LINKED ISSUE #1201: Return ReferencePolicy CRD, deprecated, for v0.5.0
To make migration easier, we should deprecate the v1alpha2 ReferencePolicy resource in v0.5.0, rather than removing it.

This means we'll need to add the types and CRD back, and figure out how to get the deprecation field and warning set on the generated CRD object.  (This will need to use https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation).

ReferencePolicy will then be removed in v0.6.0.

robscott:
**What type of PR is this?**
/kind cleanup

This addresses #1201 since at least some implementations will want to support both ReferencePolicy and ReferenceGrant concurrently.

For reference, this is what the deprecation warning looks like:
Screen Shot 2022-06-13 at 2 17 01 PM

Fixes #1201

```release-note
NONE
```


apis/v1alpha2/referencegrant_types.go
```diff
@@ -29,7 +29,7 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 // as the policy.
 //
 // Each ReferenceGrant can be used to represent a unique trust relationship.
-// Additional Reference Policies can be used to add to the set of trusted
+// Additional Reference Grants can be used to add to the set of trusted
```
shaneutt, line 32: Good catch :+1:

apis/v1alpha2/referencepolicy_types.go
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+package v1alpha2
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:categories=gateway-api,shortName=refpol
+// +kubebuilder:deprecatedversion:warning="ReferencePolicy has been renamed to ReferenceGrant. ReferencePolicy will be removed in v0.6.0 in favor of the identical ReferenceGrant resource."
```
shaneutt, line 24: Looks good. Should we add an issue that references this line to [v0.6.0](https://github.com/kubernetes-sigs/gateway-api/milestone/6) so that we get a upfront reminder?

shaneutt (approved):
Looks straightforward, just one comment about follow-up.

apis/v1alpha2/referencepolicy_types.go
robscott: Good call: https://github.com/kubernetes-sigs/gateway-api/issues/1206

nathancoleman (approved):

PR #1207, opened 2022-06-13
docs: update description of GatewayAddresses

shaneutt:
**What type of PR is this?**

/kind documentation

Related to a conversation that I brought up in a community meeting, the purpose of this PR is to tweak the wording for `GatewayAddresses` so that it's more clear that the current intention is for these addresses to be addresses like `LoadBalancer` service addresses, or otherwise addresses used for the Gateway to service ingress traffic from outside the cluster, and not simply every kind of address any particular `Gateway` implementation might have. The previous language was just "all addresses bound to the Gateway", but we're not currently expecting (for instance) that implementations would add cluster or pod IPs here.

In that same talk we discussed the possibility of a GEP where could try and add networking contexts to `GatewayAddresses`, but at least for now I felt this distinction was a small helpful thing that could be done for the moment, and we could think about grander improvements later.

```release-note
The documentation around the GatewayAddress type now makes the distinction that these are meant to be "public" IPs (not necessarily public internet, but at least addresses that route traffic from outside of the cluster network in).
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -469,10 +469,10 @@ type GatewayAddress struct {
 
 // GatewayStatus defines the observed state of Gateway.
 type GatewayStatus struct {
-	// Addresses lists the IP addresses that have actually been
-	// bound to the Gateway. These addresses may differ from the
-	// addresses in the Spec, e.g. if the Gateway automatically
-	// assigns an address from a reserved pool.
+	// Addresses lists the IP addresses that are used to serve ingress traffic from
+	// outside the cluster to the Gateway. These addresses may differ from the
+	// addresses in the Spec, e.g. if the Gateway automatically assigns an address
+	// from a reserved pool.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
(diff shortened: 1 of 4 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -469,10 +469,10 @@ type GatewayAddress struct {
 
 // GatewayStatus defines the observed state of Gateway.
 type GatewayStatus struct {
-	// Addresses lists the IP addresses that have actually been
-	// bound to the Gateway. These addresses may differ from the
-	// addresses in the Spec, e.g. if the Gateway automatically
-	// assigns an address from a reserved pool.
+	// Addresses lists the IP addresses that are used to serve ingress traffic from
+	// outside the cluster to the Gateway. These addresses may differ from the
```
robscott, line 473: I think this may be a bit too restrictive, even if it is true 99% of the time today. For example, some Gateway implementations may be limited to exposing traffic within a cluster. I don't think that's going to be common, but I don't want to prevent that use case. I've also seen some interesting discussions around using this API to represent egress, which I think this clarification would prevent.

shaneutt: Alright, then for now i'll just leave it as is.

PR #1208, opened 2022-06-14
Move error status code from 404 to 500

### LINKED ISSUE #1200: Move error status code to 500

**Why this is needed**:

dpasiukevich:
* If there are no matching routes, a 404 should be generated
* If there are matching routes but no/missing backends/endpoints to
  route to, a 500 (specifically) should be generated.

**What type of PR is this?**

/kind api-change

Update status code requirements to return 500 if there are no backends for the route instead of 404

Fixes #1200

```release-note
The Gateway API now requires to return http 500 status code if there are no backends for existing route
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -202,7 +202,7 @@ type HTTPRouteRule struct {
 	// BackendRefs defines the backend(s) where matching requests should be
 	// sent.
 	//
-	// A 404 status code MUST be returned if there are no BackendRefs or filters
+	// A 500 status code MUST be returned if there are no BackendRefs or filters
 	// specified that would result in a response being sent.
 	//
 	// A BackendRef is considered invalid when it refers to:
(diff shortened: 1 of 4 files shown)


robscott:
Thanks @dpasiukevich!

/lgtm
/approve

apis/v1alpha2/httproute_types.go
```diff
@@ -212,11 +212,11 @@ type HTTPRouteRule struct {
...
-	// When a BackendRef is invalid, 404 status codes MUST be returned for
+	// When a BackendRef is invalid, 500 status codes MUST be returned for
 	// requests that would have otherwise been routed to an invalid backend. If
 	// multiple backends are specified, and some are invalid, the proportion of
 	// requests that would otherwise have been routed to an invalid backend
-	// MUST receive a 404 status code.
+	// MUST receive a 500 status code.
 	//
 	// When a BackendRef refers to a Service that has no ready endpoints, it is
 	// recommended to return a 503 status code.
```
mikemorris, line 222: Should this have been changed to 500 too?

dpasiukevich: I agree, seems we should update here as well.
@robscott should I make the PR to change this status code 503 -> 500 as well?

dpasiukevich:
I could create the test. Is it needed to be done before the v0.5.0 release?

apis/v1alpha2/httproute_types.go
dpasiukevich: Added the PR for that: https://github.com/kubernetes-sigs/gateway-api/pull/1210

robscott:
@dpasiukevich I don't think this is required for v0.5.0 release, but we should add a conformance test for it sometime soon. As I was writing up an issue for that test I ran into https://github.com/kubernetes-sigs/gateway-api/issues/1211, so may be a bit until we figure out exactly what that test should look like.

PR #1209, opened 2022-06-14
Move error status code from 404 to 500

### LINKED ISSUE #1200: Move error status code to 500

dpasiukevich:
Update API for TLSRouteRule to return 500 status code if there are
unspecified or invalid BackendRefs

**What type of PR is this?**

/kind api-change

Update status code requirements to return 500 if there are no backends for the route instead of 404 for TLSRouteRule

Fixes #1200

```release-note
The Gateway API now requires to return http 500 status code if there are no backends for existing TLSRouteRule
```


robscott:
Thanks @dpasiukevich!

/lgtm
/approve

PR #1210, opened 2022-06-14
Move error status code from 503 to 500

### LINKED ISSUE #1200: Move error status code to 500

dpasiukevich:
Update HTTPRouteRule API spec to return 500 http status code if it
refers to a Service that has no ready endpoints for both v1alpha2 & v1beta1

**What type of PR is this?**

/kind api-change

Update status code requirements to return 500 if there are no backends for the route instead of 503

Fixes #1200

```release-note
The Gateway API now requires to return http 500 status code if there are no backends for existing route
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -219,7 +219,7 @@ type HTTPRouteRule struct {
 	// MUST receive a 500 status code.
 	//
 	// When a BackendRef refers to a Service that has no ready endpoints, it is
-	// recommended to return a 503 status code.
+	// recommended to return a 500 status code.
 	//
 	// Support: Core for Kubernetes Service
 	// Support: Custom for any other resource
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -218,7 +218,7 @@ type HTTPRouteRule struct {
 	// MUST receive a 500 status code.
 	//
 	// When a BackendRef refers to a Service that has no ready endpoints, it is
-	// recommended to return a 503 status code.
+	// recommended to return a 500 status code.
 	//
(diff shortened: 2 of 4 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -219,7 +219,7 @@ type HTTPRouteRule struct {
 	// MUST receive a 500 status code.
 	//
 	// When a BackendRef refers to a Service that has no ready endpoints, it is
-	// recommended to return a 503 status code.
+	// recommended to return a 500 status code.
```
robscott, line 222: This is a tricky one. I think 503 could actually be appropriate here. In this specific case the Service reference is correct, there just aren't any endpoints available at that point in time. That seems more likely to be the kind of temporary outage suggested by a 503 than any of the other cases we've considered (ie a reference to a resource that simply doesn't exist).

With that said, a 500 is more generic and consistent with what we've used elsewhere. I think I'm good with this change, but want a few more people to chime in to confirm.

robscott (commented):
Thanks @dpasiukevich! Would like a few more people to look at this one, but I _think_ it makes sense.

/cc @mikemorris @youngnick @shaneutt

apis/v1alpha2/httproute_types.go
mikemorris: Agree 503 could actually make sense here, I had assumed it was inconsistent, but it's more just splitting out "valid backendRef with zero instances" as a separate case from "invalid backendRef pointing to nonexistant thing"

dpasiukevich: Do we keep 503 then?

youngnick: I think we should keep 503 for this specific case, only.

PR #1217, opened 2022-06-15
conformance: fix gateway listener ports for multiple calls to MustApplyWithCleanup

michaelbeaumont:
This PR fixes the logic for managing `Gateway` listener ports added in https://github.com/kubernetes-sigs/gateway-api/pull/1101 for new tests where the `Applier` needs to keep track of which ports are assigned to which `Gateway`s for correctness.
Tests are added.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

### DIFF
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -42,7 +42,6 @@ import (
 // Applier prepares manifests depending on the available options and applies
 // them to the Kubernetes cluster.
 type Applier struct {
-	NamespaceLabels map[string]string
 	// ValidUniqueListenerPorts maps each listener port of each Gateway in the
 	// manifests to a valid, unique port. There must be as many
 	// ValidUniqueListenerPorts as there are listeners in the set of manifests.
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -84,9 +84,10 @@ func NewApplier(namespaceLabels map[string]string, validPorts []v1beta1.PortNumb
 	}
 }
 
-// freePortsFor is used to mark ports used by a Gateway as available. It does
-// not lock the availablePorts map.
-func (a *Applier) freePortsFor(name types.NamespacedName) {
+// markPortsAvailable is used to mark ports used by a Gateway as available. It does
+// not lock the availablePorts map because it's intended to be called within a
+// critical section.
+func (a *Applier) markPortsAvailable(name types.NamespacedName) {
 	ports, ok := a.assignedPorts[name]
 	if !ok {
 		return
@@ -100,11 +101,15 @@ func (a *Applier) freePortsFor(name types.NamespacedName) {
 // prepareGateway adjusts both listener ports and the gatewayClassName. It
 // returns the ports used by the listeners if they came from validPorts.
 func (a *Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured) {
+	a.portsLock.Lock()
+	defer a.portsLock.Unlock()
+
 	name := types.NamespacedName{
 		Namespace: uObj.GetNamespace(),
 		Name:      uObj.GetName(),
 	}
-	a.freePortsFor(name)
+	// Mark the ports allocated to this Gateway as available
+	a.markPortsAvailable(name)
 
 	err := unstructured.SetNestedField(uObj.Object, a.GatewayClass, "spec", "gatewayClassName")
 	require.NoErrorf(t, err, "error setting `spec.gatewayClassName` on %s Gateway resource", uObj.GetName())
@@ -129,7 +134,7 @@ func (a *Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured) 
 			newPort, ok := newPorts[port]
 			if !ok {
 				var portIsValid bool
-				newPort, portIsValid = a.availablePorts.AssignAvailablePort()
+				newPort, portIsValid = a.availablePorts.PopPort()
 				require.True(t, portIsValid, "not enough unassigned valid ports for Gateway resource")
 				newPorts[port] = newPort
 			}
@@ -181,9 +186,6 @@ func prepareNamespace(t *testing.T, uObj *unstructured.Unstructured, namespaceLa
 // a set of manifests.
 func (a *Applier) prepareResources(t *testing.T, decoder *yaml.YAMLOrJSONDecoder) ([]unstructured.Unstructured, error) {
 	var resources []unstructured.Unstructured
-	a.portsLock.Lock()
-	defer a.portsLock.Unlock()
-
 	for {
 		uObj := unstructured.Unstructured{}
 		if err := decoder.Decode(&uObj); err != nil {
@@ -284,7 +286,7 @@ func (a *Applier) MustApplyWithCleanup(t *testing.T, c client.Client, timeoutCon
 						require.NoErrorf(t, err, "error deleting resource")
 					}
 					a.portsLock.Lock()
-					a.freePortsFor(namespacedName)
+					a.markPortsAvailable(namespacedName)
 					a.portsLock.Unlock()
 				})
 			}
@@ -305,7 +307,7 @@ func (a *Applier) MustApplyWithCleanup(t *testing.T, c client.Client, timeoutCon
 					require.NoErrorf(t, err, "error deleting resource")
 				}
 				a.portsLock.Lock()
-				a.freePortsFor(namespacedName)
+				a.markPortsAvailable(namespacedName)
 				a.portsLock.Unlock()
 			})
 		}
--- a/conformance/utils/kubernetes/apply.go
+++ b/conformance/utils/kubernetes/apply.go
@@ -42,13 +42,18 @@ import (
 // Applier prepares manifests depending on the available options and applies
 // them to the Kubernetes cluster.
 type Applier struct {
+	// Labels to apply to namespaces created by the Applier
+	NamespaceLabels map[string]string
+
 	// ValidUniqueListenerPorts maps each listener port of each Gateway in the
-	// manifests to a valid, unique port. There must be as many
-	// ValidUniqueListenerPorts as there are listeners in the set of manifests.
-	// For example, given two Gateways, each with 2 listeners, there should be
-	// four ValidUniqueListenerPorts.
+	// manifests to a valid, unique port.
+	// There must be as many validPorts as the maximum number of ports
+	// used by listeners simultaneously.
+	// For example, given one Gateway with 2 listeners on the same port and one
+	// Gateway with 2 listeners on different ports, there should be at least three
+	// validPorts.
 	// If empty or nil, ports are not modified.
-	ValidUniqueListenerPorts []v1beta1.PortNumber
+	ValidUniqueListenerPorts PortSet
 
 	// GatewayClass will be used as the spec.gatewayClassName when applying Gateway resources
 	GatewayClass string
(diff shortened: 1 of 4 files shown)


conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
+func NewApplier(namespaceLabels map[string]string, validPorts []v1alpha2.PortNumber) *Applier {
+	return &Applier{
+		namespaceLabels: namespaceLabels,
+		availablePorts:  validPorts,
+		assignedPorts:   map[types.NamespacedName]PortSet{},
+	}
+}
+
+// freePortsFor is used to mark ports used by a Gateway as available. It does
+// not lock the availablePorts map.
```
robscott, line 70: Why doesn't this lock the map?

conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
-	// For example, given two Gateways, each with 2 listeners, there should be
-	// four ValidUniqueListenerPorts.
-	// If empty or nil, ports are not modified.
-	ValidUniqueListenerPorts []v1alpha2.PortNumber
+	namespaceLabels map[string]string
+	// availablePorts holds ports that can still be assigned
+	availablePorts PortSet
+	// assignedPorts tracks ports used for Gateway objects
+	assignedPorts map[types.NamespacedName]PortSet
+	portsLock     sync.Mutex
```
robscott, line 50: Is this lock intended to guard assignedPorts, availablePorts or both?

conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
+// If empty or nil, Gateway listener ports are not modified.
+func NewApplier(namespaceLabels map[string]string, validPorts []v1alpha2.PortNumber) *Applier {
+	return &Applier{
+		namespaceLabels: namespaceLabels,
+		availablePorts:  validPorts,
+		assignedPorts:   map[types.NamespacedName]PortSet{},
+	}
+}
+
+// freePortsFor is used to mark ports used by a Gateway as available. It does
```
robscott, line 69: Is this intended to be used when a Gateway is being deleted? The name made me think this was just returning a list of available ports for the given input.

conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
 // prepareGateway adjusts both listener ports and the gatewayClassName. It
-// returns an index pointing to the next valid listener port.
-func prepareGateway(t *testing.T, uObj *unstructured.Unstructured, gatewayClassName string, validListenerPorts []v1alpha2.PortNumber, portIndex int) int {
+// returns the ports used by the listeners if they came from validPorts.
+func (a *Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured, gatewayClassName string) {
+	name := types.NamespacedName{
+		Namespace: uObj.GetNamespace(),
+		Name:      uObj.GetName(),
+	}
+	a.freePortsFor(name)
```
robscott, line 89: What is this actually doing? A comment may help explain the logic here.

conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
+		a.availablePorts = append(a.availablePorts, port)
+	}
+	delete(a.assignedPorts, name)
 }
 
 // prepareGateway adjusts both listener ports and the gatewayClassName. It
-// returns an index pointing to the next valid listener port.
-func prepareGateway(t *testing.T, uObj *unstructured.Unstructured, gatewayClassName string, validListenerPorts []v1alpha2.PortNumber, portIndex int) int {
+// returns the ports used by the listeners if they came from validPorts.
+func (a *Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured, gatewayClassName string) {
```
robscott, line 84: Is the idea that this function can only be called if portsLock has already been locked?

conformance/utils/kubernetes/apply.go
```diff
@@ -41,45 +42,95 @@ import (
...
+		for _, port := range newPorts {
+			allocatedPorts = append(allocatedPorts, port)
 		}
 
 		err = unstructured.SetNestedSlice(uObj.Object, listeners, "spec", "listeners")
 		require.NoErrorf(t, err, "error setting `spec.listeners` on %s Gateway resource", uObj.GetName())
 	}
 
-	return portIndex
+	a.assignedPorts[name] = allocatedPorts
```
robscott, line 133: I'm not sure how easy this is, but it seems like it would be much safer to pull all the port logic out to a separate function and handle locking inside there. All these calls like this feel rather dangerous given the size of this function and that it's unclear if/when locking has occurred.

conformance/utils/kubernetes/portset.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (s *PortSet) AssignAvailablePort() (v1alpha2.PortNumber, bool) {
+	if len(*s) == 0 {
+		return 0, false
+	}
+
+	newPort := v1alpha2.PortNumber((*s)[0])
+	*s = (*s)[1:]
+	return newPort, true
+}
```
robscott, line 34: Although more heavyweight, it may be good to just use the [apimachinery sets util](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/sets/int.go) or recreate some subset of that here for the PortNumber type. A `PopAny()` function name like that uses would probably be more familiar than `AssignAvailablePort()`.

robscott (commented):
Sorry it took me so long to review this! Thanks for all the work @michaelbeaumont!

conformance/utils/kubernetes/apply.go
michaelbeaumont: The map is locked in `prepareResources` but this is called multiple times in the critical section, I'll add to the comment.

conformance/utils/kubernetes/apply.go
michaelbeaumont: Yeah that's free as a verb not an adjective. I'll rename

conformance/utils/kubernetes/apply.go
michaelbeaumont: Yes.

conformance/utils/kubernetes/apply.go
michaelbeaumont: The lock actually belongs in `prepareGateway` so I've put it there, which at least keeps it in this function. I'm not sure how to refactor to remove the port logic any further or whether it would help. The only thing `prepareGateway` does besides set the gateway class is iterating through the listeners to do this port logic.

conformance/utils/kubernetes/apply.go
michaelbeaumont: It not only guards the map from concurrent writes but creates a critical section for freeing and allocating ports every time we have a Gateway.

robscott:
@michaelbeaumont can you rebase this PR?

shaneutt:
@michaelbeaumont just checking in, is there anything you need in order to help move this forward? let us know :vulcan_salute:

conformance/utils/kubernetes/apply.go
```diff
@@ -55,41 +56,112 @@ type Applier struct {
 
 	// ControllerName will be used as the spec.controllerName when applying GatewayClass resources
 	ControllerName string
+
+	namespaceLabels map[string]string
```
shaneutt, line 60: We already have a nearly identical `NamespaceLabels` attribute on this struct, do we actually need to make a distinction here? If so, can we add comments explaining why?

michaelbeaumont: I meant to make all fields private, I think `NamespaceLabels` crept back up in a rebase. I introduced the `New` pattern to avoid the writing to a `nil` map issue, then more public fields were introduced. I'll just get rid of the `New` pattern and deal with the `nil` map instead.

shaneutt:
We lost track of this one! :facepalm: 

@michaelbeaumont do you wanna try and rebase here and we'll see how this applies? Or what would you like to do?

michaelbeaumont:
> I'm gonna consider it a draft for the time being as there's some work that would need to be done first, let me know how we can help!

I think it needs to be decided whether we want to add this option for running the conformance test/the review needs to be finished. The PR is ready, we use it for our conformance testing.

conformance/utils/kubernetes/apply.go
```diff
@@ -41,55 +42,114 @@ import (
...
+	}
+	delete(a.assignedPorts, name)
 }
 
 // prepareGateway adjusts both listener ports and the gatewayClassName. It
-// returns an index pointing to the next valid listener port.
-func (a Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured, portIndex int) int {
+// returns the ports used by the listeners if they came from validPorts.
+func (a *Applier) prepareGateway(t *testing.T, uObj *unstructured.Unstructured) {
+	// This locks serves as a critical section for the logic around manipulating
```
shaneutt, line 88:
```suggestion
	// This lock serves as a critical section for the logic around manipulating
```

conformance/utils/kubernetes/portset.go
```diff
@@ -0,0 +1,34 @@
...
+*/
+
+package kubernetes
+
+import (
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// PortSet is a wrapper for handling sets of ports.
+type PortSet []v1beta1.PortNumber
```
shaneutt, line 24: We have `Set` types available in Kubernetes utilities `k8s.io/apimachinery/pkg/utils/sets`, which you can see we already use in [conformance/utils/suite/suite.go:77](https://github.com/kubernetes-sigs/gateway-api/blob/2ae5227364ccf25652fb233554861ffd4c0601c8/conformance/utils/suite/suite.go#L77). Could we perhaps just use these?

michaelbeaumont: Actually, this changes the semantics to an unordered set, which makes the output non deterministic. Not necessarily a problem in practice, but it does make the tests significantly harder to write.

shaneutt: Right, I consider this resolved at your discretion/preference.

shaneutt:
This one is currently our oldest PR, and we've just recently started going into multiple pages in the PR pagination (:fearful:) we should probably try and get this one sorted out.

For some of the people who are often involved in conformance, check this one out and let us know what you think:

/cc @mlavacca @arkodg @sunjayBhatia

michaelbeaumont:
Note that as is, without this PR, the "unique listener ports" functionality is broken.

conformance/utils/suite/suite.go
```diff
@@ -62,12 +62,19 @@ type Options struct {
...
-	// four ValidUniqueListenerPorts.
-	// If empty or nil, ports are not modified.
+	// ValidUniqueListenerPorts is the set of ports that listeners may be
+	// assigned. This ensures that listeners are assigned only ports from
+	// this set, as opposed to the defaults that are set in the manifests.
+	// This is useful when the test infrastructure is constrained to exposing
+	// only a limited set of ports outside of the cluster.
+	// The Applier maps each listener port of each Gateway in the manifests
+	// to a port from ValidUniqueListenerPorts.
+	// There must be as many validPorts as the maximum number of ports
```
sunjayBhatia, line 72: nit (can add elsewhere this same documentation is included in this change too): 
```suggestion
	// There must be as many validPorts as the maximum number of distinct ports
```

conformance/utils/kubernetes/portset.go
```diff
@@ -0,0 +1,34 @@
...
+package kubernetes
+
+import (
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// PortStack is a wrapper for handling sets of ports.
+type PortStack []v1beta1.PortNumber
+
+func (s *PortStack) PopPort() (v1beta1.PortNumber, bool) {
```
sunjayBhatia, line 26: could use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set ?

michaelbeaumont: See https://github.com/kubernetes-sigs/gateway-api/pull/1217#discussion_r1136225015 :)

sunjayBhatia: ha! nice feel free to ignore 👍🏽

michaelbeaumont:
Change of gears on this one. Due to a recent change from servicelb to metallb for our k3s e2e test environment, this functionality is no longer needed. servicelb allocates a port on the same IP for every LoadBalancer Service, so ports have to be non-overlapping, where metallb allocates a new IP for each LoadBalancer Service.

Given that it's broken as is, I propose we remove the `ValidUniqueListenerPorts` functionality entirely: https://github.com/kubernetes-sigs/gateway-api/pull/1992

PR #1221, opened 2022-06-21
conformance: Use valid YAML string for gatewayClassName placeholder

nathancoleman:
**What type of PR is this?**

/kind cleanup

YAML parsers bork this placeholder because it appears as a map instead of the string it should be interpreted as

Fixes #

```release-note
NONE
```


howardjohn (approved):

shaneutt (approved):
/lgtm

PR #1224, opened 2022-06-22
docs: add support details for kong's implementation

shaneutt:
**What type of PR is this?**

/kind documentation

It occurred to me that from time to time we might get end-users of specific implementations reporting issues here, rather than on their respective support pages. It's not much, but as a tiny way to help with that this patch adds support and slack links to the Kong section of the implementations page in hopes of helping to direct people to the right place to get help.

```release-note
NONE
```


robscott:
Thanks @shaneutt!

/lgtm
/approve

PR #1228, opened 2022-06-24
Adds new tiny type for HTTPRoute Listeners in conformance tests

### LINKED ISSUE #1219: HTTPRoute Listener Hostname Matching conformance tests have invalid ParentRef expectations
**What happened**:
Our conformance test helpers expect [exactly one ParentRef to be present in HTTPRoute status](https://github.com/kubernetes-sigs/gateway-api/blob/9e271efb321fb67708f880d3a83a00c7d4229c8f/conformance/utils/kubernetes/helpers.go#L113-L144). That was true until we added a test case [with 2 parent refs](https://github.com/kubernetes-sigs/gateway-api/blob/9e271efb321fb67708f880d3a83a00c7d4229c8f/conformance/tests/httproute-listener-hostname-matching.yaml#L68-L80) (same Gateway, different ListenerName).

The test case itself is valid, we just need to rethink how we're verifying parent status here. For now I think all we  instead of passing in NamespacedName for Routes, we need a new construct that includes Namespace, Name, and an optional list of ListenerNames. A more thorough solution would also support a single Route referring to multiple unique Gateways, but that would likely be more complex, open to ideas.

spencerhance:
**What type of PR is this?**

/kind bug
/kind failing-test

Adds a tiny type for specifying HTTPRoute ParentRefs to Gateways
in the conformance tests.  This allows us to test cases with 2 ParentRefs (to the same Gateway).

Fixes #1219

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-cross-namespace.go
+++ b/conformance/tests/httproute-cross-namespace.go
@@ -37,7 +37,7 @@ var HTTPRouteCrossNamespace = suite.ConformanceTest{
 	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
 		routeNN := types.NamespacedName{Name: "cross-namespace", Namespace: "gateway-conformance-web-backend"}
 		gwNN := types.NamespacedName{Name: "backend-namespaces", Namespace: "gateway-conformance-infra"}
-		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, gwNN, routeNN)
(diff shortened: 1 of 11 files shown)


youngnick:
Looks like you need a `gofmt -s` pass, @spencerhance. Aside from that, seems like a reasonable change to me.

spencerhance:
Thanks @youngnick !

shaneutt (approved):
Yep, seems like a fair way to handle the multiple parents :+1: 

/lgtm

PR #1230, opened 2022-06-27
apis: validate header and query param matches

### LINKED ISSUE #1225: clarify behavior of multiple query param matches for the same key
The spec does not indicate what should be done if a route ~rule~ match has multiple query param matches for the same key.

For header matches, guidance is given to use the first match for a given header name, and to ignore subsequent ones for the same name (see https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPHeaderMatch).

I'd suggest adding the same guidance for query params, with the exception that query param names be case-sensitive. If this makes sense, I'll open a PR to add this to the spec. (I should also be able to add a conformance test for query params that includes this soon).

skriss:
**What type of PR is this?**
/kind feature

Adds webhook validation to ensure that no
HTTP header or query param is matched more
than once in a given route rule. Also updates
the godoc for HTTPQueryParamMatch to be
consistent with HTTPHeaderMatch.

Fixes #1225 

```release-note
Adds webhook validation to ensure that no HTTP header or query param is matched more than once in a given route rule.
```

### DIFF
--- a/apis/v1alpha2/validation/httproute_test.go
+++ b/apis/v1alpha2/validation/httproute_test.go
@@ -19,6 +19,8 @@ package validation
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	utilpointer "k8s.io/utils/pointer"
 
@@ -576,11 +578,11 @@ func TestValidateHTTPHeaderMatches(t *testing.T) {
 	tests := []struct {
 		name          string
 		headerMatches []gatewayv1a2.HTTPHeaderMatch
-		errCount      int
+		expectErr     string
 	}{{
 		name:          "no header matches",
 		headerMatches: nil,
-		errCount:      0,
+		expectErr:     "",
 	}, {
 		name: "no header matched more than once",
 		headerMatches: []gatewayv1a2.HTTPHeaderMatch{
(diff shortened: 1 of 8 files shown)


apis/v1alpha2/validation/httproute_test.go
```diff
@@ -572,6 +572,128 @@ func TestValidateHTTPPathMatch(t *testing.T) {
...
+				}},
+			}}
+
+			errs := ValidateHTTPRoute(&route)
+			if len(errs) != tc.errCount {
+				t.Errorf("got %d errors, want %d errors: %s", len(errs), tc.errCount, errs)
+			}
+		})
+	}
+}
```
shaneutt, line 634: Perhaps a bit of a nitpick: How do you feel about verifying the error contents as well?

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -572,6 +572,128 @@ func TestValidateHTTPPathMatch(t *testing.T) {
...
```
shaneutt, line 695: Similar to the above, any thoughts on verifying the error contents?

shaneutt (commented):
No blockers, thanks for taking the time and thanks for adding the validation!

Just a couple smaller comments but I think this is just about ready to go.

apis/v1alpha2/validation/httproute_test.go
skriss: Yeah I can do that; the rest of the tests in this file are only verifying error counts so I went with that for consistency but agree it'd be a bit better to verify specific errors.

shaneutt (approved):
/lgtm

Issue #1229, opened 2022-06-27
Documentation needs to be updated as part of v0.5.0 release


robscott:
I think at a minimum this requires the following changes, these can be combined or split up as needed, comment on whichever ones you can take on:

- [x] _completed by @EmilyShepherd:_ Add v1beta1 examples
  * Copy https://github.com/kubernetes-sigs/gateway-api/tree/master/examples/v1alpha2 into a new v1beta1 directory, ensure all new files use v1beta1 API group
  * Copy https://github.com/kubernetes-sigs/gateway-api/tree/master/hack/invalid-examples/v1alpha2 into a new v1beta1 directory, ensure new files use v1beta1 API group
  * Update any docs (`site-src` dir) that include v1alpha2 examples (example `{% include 'v1alpha2/http-filter.yaml' %}`) to use v1beta1
- [x] _completed by @EmilyShepherd:_ Any GatewayClass, Gateway, or HTTPRoute yaml snippets in `site-src` markdown files should use `v1beta1` instead of `v1alpha2`
- [x] _completed by @EmilyShepherd:_ The `v1alpha2/` prefix in documentation links (and structure) should be removed - bonus points for finding a way to redirect from new to old path
- [x] _completed by @keithmattix:_ Any concepts that are only included in experimental channel need to be clearly documented as that (ReferenceGrant, TCPRoute, TLSRoute, and UDPRoute are all included in this) - [example](https://github.com/kubernetes-sigs/gateway-api/blame/master/site-src/v1alpha2/guides/http-redirect-rewrite.md#L36-L40)
- [ ] As part of release process: Update documentation to explain how to install standard and experimental CRDs.

I may have missed some items here, feel free to add any additional suggestions below.

keithmattix:
I'm willing to help out as well!

robscott:
Awesome, thanks @EmilyShepherd and @keithmattix! Just comment whichever item(s) you want to start with.

keithmattix:
I'll start off with a PR for calling out items in the experimental channel

keithmattix:
FYI @robscott the instructions on the [getting started](https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/v1alpha2/guides/getting-started.md#install-the-crds) for installing the CRDs also needs to be updated to reflect the new channel structure. I couldn't find another issue tracking that, so I'm commenting here

robscott:
[...]

PR #1234, opened 2022-06-27
Add experimental warning for resources that are experimental

keithmattix:
Signed-off-by: Keith Mattix II <keithmattix2@gmail.com>

**What type of PR is this?**

/kind documentation

This PR adds an warning to prominent references of resources that exist only in the experimental channel so that readers are made aware of the implications of the versioning scheme as early in their Gateway API journey as possible.

Addresses #1229

```release-note
Add warning headers for experimental resources/concepts
```


EmilyShepherd:
Note to self that this will cause merge conflicts with my change (removing the `v1alpha2` prefix from docs paths).

Assuming this PR is merged first I'll rebase and resolve.

PR #1235, opened 2022-06-28
conformance: fix GatewayAndHTTPRoutesMustBeReady for multiple parentRefs

michaelbeaumont:
**What type of PR is this?**
/kind bug

Makes the conformance test `HTTPRouteListenerHostnameMatching` possible to pass. It's the first test with an `HTTPRoute` that has multiple `parentRefs`. Previously this code assumed every `HTTPRoute` had one `parentRef`. It now checks the `HTTPRoute` to see how many `parentRefs` it has.

```release-note
NONE
```


michaelbeaumont:
@robscott Yes it does overlap, though with that PR the test is still broken and this API forces us to duplicate the listener names in the YAML and the test code.

Is there a reason not to just derive the correct status from the `parentRefs` in the `HTTPRoute` resource?

Opened https://github.com/kubernetes-sigs/gateway-api/pull/1237 to fix the test

PR #1238, opened 2022-06-28
Update docs to use v1beta1

EmilyShepherd:
[EDIT: In light of @robscott's suggestion, the change of the documentation reference to `v1beta1` has been removed from this PR - all examples will continue to use `v1alpha2` for the moment.]

**What type of PR is this?**
/kind documentation

This PR is part of #1229 which requires updating examples and documentation to use the newer `v1beta1` versions of resources, where appropriate. It does the following:

- Adds v1beta1 examples to `hack/invalid-examples`
- Adds v1beta1 examples to `examples/experimental`
- Updates the api-docs auto generator scripts to generate both v1beta1 and v1alpha2
- ~Updates the documentation to refer to v1beta1 examples were possible.~
- Updates the documentation's file structure to remove the v1alpha2 prefix (v1alpha2 symlinks have been left in place to ensure that links to the old URLs still work).
- ~Updates the "Used In" metadata headers in all examples (either by removing them for examples that are no longer used, or by ensuring that their path does not contain the old-style v1alpha2 prefix).~

Notable points:
- The `hack/api-docs/generate.sh` script had to be refactored slightly to have to generate both versions of the api
- The `apis/docs.go` stub had to be added to get gen-crd-api-reference-docs to recognise `apis` as a go package and therefore scan its subdirectories.
- ~`Gateway`, `GatewayClass`, and `HTTPRoute` have been explictly excluded from the `v1alpha2` api-docs generation, as they already appear in `v1beta1`. It is debatable if this is "correct" behaviour as they do, of course, still exist in the `v1alpha2` version, but it felt redundant to include them twice in the docs.~

Addresses #1229

```release-note
Add examples and documentation for v1beta1
```


robscott:
Thanks @EmilyShepherd, this is great! Now that I'm realizing that our actual release date is 2 weeks away (July 12), I'm  realizing that we could result in a confusing state of our docs if we're not careful. 

Specifically if our examples all show the `v1beta1` API group but our getting started section still recommends installing `v0.4.3` (without v1beta1). I think that leaves us with two options:

1. Update docs to recommend installing v0.5.0 release candidate
2. Hold off on this specific part of the change until we release v0.5.0

I think 2 is the much safer option. If that makes sense to you, do you mind keeping the last 2 commits of this PR for a later PR that we can merge right after we release v0.5.0?

EmilyShepherd:
> Thanks @EmilyShepherd, this is great! Now that I'm realizing that our actual release date is 2 weeks away (July 12), I'm realizing that we could result in a confusing state of our docs if we're not careful.
> 
> Specifically if our examples all show the `v1beta1` API group but our getting started section still recommends installing `v0.4.3` (without v1beta1). I think that leaves us with two options:
> 
> 1. Update docs to recommend installing v0.5.0 release candidate
> 2. Hold off on this specific part of the change until we release v0.5.0
> 
> I think 2 is the much safer option. If that makes sense to you, do you mind keeping the last 2 commits of this PR for a later PR that we can merge right after we release v0.5.0?

If it were one of my own projects, I'd normally just go straight to recommending the latest version, but that is most definately not a call for me to make.

So I am therefore more than happy to take the last two commits out and split them into a seperate PR. I'll have to update the api-docs gen code to no longer exclude `HTTPRoute`, `Gateway` and `GatewayClass` from the v1alpha2 version (as without the change in the docs, many pages will still be referencing the v1alpha2 versions, so it is important that they exist in the spec).

I'll pull those commits out, make that change (and also do an intermediate repair of "Used By" headers, which will be wrong if we merge just the first few commits in) now.

If you would like, I could also amend this PR to still include the dropping of the `v1alpha2` url prefix (but not change any of the examples / spec references) as that feels like it would be safe to do pre-release?

hack/make-docs.sh
```diff
@@ -35,6 +35,6 @@ mkdir -p site
 # Generate docs with mkdocs
 mkdocs build
 # Generate v1alpha2 API docs
-./hack/api-docs/generate.sh v1alpha2 site/v1alpha2-spec.html
+./hack/api-docs/generate.sh site/spec.html
 # Add them to spec page originally generated by mkdocs
-run::sed -e '/REPLACE_WITH_GENERATED_CONTENT/{r site/v1alpha2-spec.html' -e 'd;}' site/v1alpha2/references/spec/index.html
```
robscott, line 40: For the sake of the old link, it would be nice to leave this line in place:
https://deploy-preview-1238--kubernetes-sigs-gateway-api.netlify.app/v1alpha2/references/spec/

hack/invalid-examples/v1beta1/httproute/invalid-filter-wrong-field.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: invalid-filter-wrong-field
+spec:
+  rules:
+  - filters:
+    - type: RequestHeaderModifier
+      requestRedirect: 
+        port: 443
```
robscott, line 10: I think this probably already existed, but mind adding a trailing newline?

robscott (commented):
This is really great, thanks @EmilyShepherd! Just a couple nits but otherwise LGTM.

/approve

robscott:
>If it were one of my own projects, I'd normally just go straight to recommending the latest version, but that is most definately not a call for me to make.

I'd usually agree, but since we still only have a release candidate, we historically have recommended installing the latest stable release until we get another full release out.

hack/make-docs.sh
EmilyShepherd: Good catch - will fix

robscott:
Thanks @EmilyShepherd!

/lgtm

PR #1243, opened 2022-06-30
Update HTTPRoute BackendRef docs to clarify error behavior

### LINKED ISSUE #1211: Conflicting guidance when a HTTPRoute refers to an invalid backend
I was trying to write an issue describing what conformance tests should look like for this, and I think.

On the HTTPBackendRef type definition we say that if the referent can't be found or is not allowed by a ReferenceGrant the BackendRef must be dropped from the Gateway. https://github.com/kubernetes-sigs/gateway-api/blob/3f4b981dd6669f67398d753a6f278b241d669953/apis/v1beta1/httproute_types.go#L870-L879

On the BackendRefs field we say "When a BackendRef is invalid, 500 status codes MUST be returned for requests that would have otherwise been routed to an invalid backend." "Invalid" is defined to include both of the scenarios described above. https://github.com/kubernetes-sigs/gateway-api/blob/3f4b981dd6669f67398d753a6f278b241d669953/apis/v1beta1/httproute_types.go#L207-L218

I think the rationale behind both of these conflicting guidelines makes sense in isolation. Maybe the following guidelines make sense:

1. Don't attach a Route to a Gateway if a Route has no valid BackendRefs or include a filter with an invalid reference - this will result in a 404 unless another Route matches.
2. Do attach a Route to a Gateway if only a portion of BackendRefs are invalid - this will result in a 500 for traffic that would have otherwise been routed to the invalid BackendRef.

The motivation for 2 is to avoid dropping a Route entirely if someone adds an invalid reference as part of a traffic split, ie send 10% to canary service, but that new reference to canary service is broken. I'm not sure that 2 is worthwhile, so an alternative would be to just say that Routes should not be attached if they contain any invalid references.

Note: I think this is a blocker for v0.5.0, but not v0.5.0-rc1.

I know a lot of people have been involved in this broader discussion around status codes, apologies if I miss anyone here. As always, appreciate anyone's thoughts on this.

/cc @mikemorris @nathancoleman @youngnick @howardjohn @shaneutt

youngnick:
**What type of PR is this?**
/kind cleanup
/kind documentation

- Moves reasons for an individual BackendRef to be invalid so they are included in the HTTPBackendRef struct docs ( the BackendRef field).
- Clarifies rules for handling invalid BackendRefs in the BackendRefs field of the HTTPRouteRule struct.

Fixes #1211
Updates #1112

```release-note
The rules for BackendRefs have been updated to make clear that invalid BackendRefs must produce HTTP response codes of 500 in most cases.
```

This PR is still WIP, since I will need to change the conformance tests to check for the behavior. But I wanted to check that we really are in agreement on this behavior while I work on those.

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -207,7 +207,7 @@ type HTTPRouteRule struct {
 	//
 	// If *all* entries in BackendRefs are invalid, and there are also no filters
 	// specified in this route rule, *all* traffic which matches this rule MUST
-	// recieve a 500 status code (exactly).
+	// receive a 500 status code (exactly).
 	//
 	// See the HTTPBackendRef definition for the rules about what makes a single
 	// HTTPBackendRef invalid.
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -207,7 +207,7 @@ type HTTPRouteRule struct {
 	//
 	// If *all* entries in BackendRefs are invalid, and there are also no filters
 	// specified in this route rule, *all* traffic which matches this rule MUST
-	// receive a 500 status code (exactly).
+	// receive a 500 status code.
 	//
 	// See the HTTPBackendRef definition for the rules about what makes a single
 	// HTTPBackendRef invalid.
@@ -221,9 +221,6 @@ type HTTPRouteRule struct {
 	// For example, if two backends are specified with equal weights, and one is
 	// invalid, 50 percent of traffic must receive a 500. Implementations may
 	// choose how that 50 percent is determined.
-	//
-	// When a HTTPBackendRef refers to a Service that has no ready endpoints,
-	// implementations MAY return a 503 for requests to that backend instead.
 	//
 	// Support: Core for Kubernetes Service
 	//
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -202,24 +202,27 @@ type HTTPRouteRule struct {
 	// BackendRefs defines the backend(s) where matching requests should be
 	// sent.
 	//
-	// A 500 status code MUST be returned if there are no BackendRefs or filters
-	// specified that would result in a response being sent.
+	// Failure behavior here depends on how many BackendRefs are specified and
+	// how many are invalid.
 	//
-	// A BackendRef is considered invalid when it refers to:
+	// If *all* entries in BackendRefs are invalid, and there are also no filters
+	// specified in this route rule, *all* traffic which matches this rule MUST
+	// recieve a 500 status code (exactly).
 	//
-	// * an unknown or unsupported kind of resource
-	// * a resource that does not exist
-	// * a resource in another namespace when the reference has not been
-	//   explicitly allowed by a ReferenceGrant (or equivalent concept).
+	// See the HTTPBackendRef definition for the rules about what makes a single
+	// HTTPBackendRef invalid.
 	//
-	// When a BackendRef is invalid, 500 status codes MUST be returned for
+	// When a HTTPBackendRef is invalid, 500 status codes MUST be returned for
 	// requests that would have otherwise been routed to an invalid backend. If
 	// multiple backends are specified, and some are invalid, the proportion of
 	// requests that would otherwise have been routed to an invalid backend
 	// MUST receive a 500 status code.
 	//
-	// When a BackendRef refers to a Service that has no ready endpoints, it is
-	// recommended to return a 503 status code.
+	// For example, if two backends are specified with equal weights, and one is
+	// invalid, 50 percent of traffic must randomly receive a 500.
+	//
+	// When a HTTPBackendRef refers to a Service that has no ready endpoints,
+	// implementations MAY return a 503 for requests to that backend instead.
 	//
 	// Support: Core for Kubernetes Service
 	//
@@ -933,21 +936,30 @@ type HTTPRequestMirrorFilter struct {
 type HTTPBackendRef struct {
 	// BackendRef is a reference to a backend to forward matched requests to.
 	//
-	// If the referent cannot be found, this HTTPBackendRef is invalid and must
-	// be dropped from the Gateway. The controller must ensure the
-	// "ResolvedRefs" condition on the Route is set to `status: False` and not
-	// configure this backend in the underlying implementation.
+	// A BackendRef can be invalid for the following reasons. In all cases, the
+	// implementation MUST ensure the `ResolvedRefs` Condition on the Route
+	// is set to `status: False`, with a Reason and Message that indicate
+	// what is the cause of the error.
 	//
-	// If there is a cross-namespace reference to an *existing* object
-	// that is not covered by a ReferenceGrant, the controller must ensure the
-	// "ResolvedRefs"  condition on the Route is set to `status: False`,
-	// with the "RefNotPermitted" reason and not configure this backend in the
-	// underlying implementation.
+	// A BackendRef is invalid if:
 	//
-	// In either error case, the Message of the `ResolvedRefs` Condition
-	// should be used to provide more detail about the problem.
+	// * It refers to an unknown or unsupported kind of resource. In this
+	//   case, the Reason and Message of the Condition must explain which kind
+	//   of resource is unknown or unsupported.
 	//
-	// Support: Custom
+	// * It refers to a resource that does not exist. In this case, the Reason
+	//   and Message of the Condition must explain which resource does not exist.
+	//
+	// * It refers a resource in another namespace when the reference has not been
+	//   explicitly allowed by a ReferenceGrant (or equivalent concept). In this
+	//   case, the Reason and Message of the Condition must explain which cross-
+	//   namespace reference is not allowed.
+	//
+	// Support: Core for Kubernetes Service
+	//
+	// Support: Custom for any other resource
+	//
+	// Support for weight: Core
 	//
 	// +optional
 	BackendRef `json:",inline"`
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -944,16 +944,17 @@ type HTTPBackendRef struct {
 	// A BackendRef is invalid if:
 	//
 	// * It refers to an unknown or unsupported kind of resource. In this
-	//   case, the Reason and Message of the Condition must explain which kind
-	//   of resource is unknown or unsupported.
+	//   case, the Reason must be set to `InvalidKind` and Message of the
+	//   Condition must explain which kind of resource is unknown or unsupported.
 	//
-	// * It refers to a resource that does not exist. In this case, the Reason
-	//   and Message of the Condition must explain which resource does not exist.
+	// * It refers to a resource that does not exist. In this case, the Reason must
+	//   be set to `BackendNotFound` and the Message of the Condition must explain
+	//   which resource does not exist.
 	//
 	// * It refers a resource in another namespace when the reference has not been
 	//   explicitly allowed by a ReferenceGrant (or equivalent concept). In this
-	//   case, the Reason and Message of the Condition must explain which cross-
-	//   namespace reference is not allowed.
+	//   case, the Reason must be set to `RefNotPermitted` and the Message of the
+	//   Condition must explain which cross-namespace reference is not allowed.
 	//
 	// Support: Core for Kubernetes Service
 	//
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -219,7 +219,8 @@ type HTTPRouteRule struct {
 	// MUST receive a 500 status code.
 	//
 	// For example, if two backends are specified with equal weights, and one is
-	// invalid, 50 percent of traffic must randomly receive a 500.
+	// invalid, 50 percent of traffic must receive a 500. Implementations may
+	// choose how that 50 percent is determined.
 	//
 	// When a HTTPBackendRef refers to a Service that has no ready endpoints,
 	// implementations MAY return a 503 for requests to that backend instead.
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -246,6 +246,15 @@ const (
 	// another namespace, where the object in the other namespace does
 	// not have a ReferenceGrant explicitly allowing the reference.
 	RouteReasonRefNotPermitted RouteConditionReason = "RefNotPermitted"
+
+	// This reason is used with the "ResolvedRefs" condition when
+	// one of the Route's rules has a reference to an unknown or unsupported
+	// Group and/or Kind.
+	RouteReasonInvalidKind RouteConditionReason = "InvalidKind"
+
+	// This reason is used with the "ResolvedRefs" condition when
+	// one of the Route's rules has a reference to a resource that does not exist.
+	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
 )
 
 // RouteParentStatus describes the status of a route with respect to an
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -252,8 +252,8 @@ const (
 	// Group and/or Kind.
 	RouteReasonInvalidKind RouteConditionReason = "InvalidKind"
 
-	// This reason is used with the "ResolvedRefs" condition when
-	// one of the Route's rules has a reference to a resource that does not exist.
+	// This reason is used with the "ResolvedRefs" condition when one of the
+	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
 )
 
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -206,7 +206,7 @@ type HTTPRouteRule struct {
 	//
 	// If *all* entries in BackendRefs are invalid, and there are also no filters
 	// specified in this route rule, *all* traffic which matches this rule MUST
-	// receive a 500 status code (exactly).
+	// receive a 500 status code.
 	//
 	// See the HTTPBackendRef definition for the rules about what makes a single
 	// HTTPBackendRef invalid.
@@ -220,9 +220,6 @@ type HTTPRouteRule struct {
 	// For example, if two backends are specified with equal weights, and one is
 	// invalid, 50 percent of traffic must receive a 500. Implementations may
 	// choose how that 50 percent is determined.
-	//
-	// When a HTTPBackendRef refers to a Service that has no ready endpoints,
-	// implementations MAY return a 503 for requests to that backend instead.
 	//
 	// Support: Core for Kubernetes Service
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -201,24 +201,28 @@ type HTTPRouteRule struct {
 	// BackendRefs defines the backend(s) where matching requests should be
 	// sent.
 	//
-	// A 500 status code MUST be returned if there are no BackendRefs or filters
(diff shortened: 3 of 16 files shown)


apis/v1alpha2/httproute_types.go
```diff
@@ -202,24 +202,27 @@ type HTTPRouteRule struct {
 	// BackendRefs defines the backend(s) where matching requests should be
 	// sent.
 	//
-	// A 500 status code MUST be returned if there are no BackendRefs or filters
-	// specified that would result in a response being sent.
+	// Failure behavior here depends on how many BackendRefs are specified and
+	// how many are invalid.
 	//
-	// A BackendRef is considered invalid when it refers to:
+	// If *all* entries in BackendRefs are invalid, and there are also no filters
```
howardjohn, line 208: Is this saying something different than "When a HTTPBackendRef is invalid, 500 status codes MUST be returned for"? They seem the same to me, and I wasn't sure if there was some subtle difference I was missing or if we were just re-iterating? Especially the "...also no filters..." part?

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
```diff
@@ -71,5 +72,17 @@ var HTTPRouteInvalidCrossNamespaceBackendRef = suite.ConformanceTest{
 		// Listener ResolvedRefs RefNotPermitted condition once
 		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
 		// has been resolved
+
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN))
```
mikemorris, line 76:
```suggestion
		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
```
I believe this PR resolves the TODO on line 49 - if we're expecting a 500 response, the route should be `Accepted` by the gateway because _something_ is being programmed in the underlying proxy for this route.

conformance/tests/httproute-invalid-reference-grant.go
```diff
@@ -98,13 +98,13 @@ var HTTPRouteInvalidReferenceGrant = suite.ConformanceTest{
 		// at least one allowed BackendRef should be accepted by a Gateway
 		// and partially configured.
 
-		t.Run("Simple HTTP request should not reach app-backend-v2", func(t *testing.T) {
+		t.Run("HTTP Request to invalid backend with missing referenceGrant should receive a 500", func(t *testing.T) {
```
mikemorris, line 101: Returning a 500 here as distinct from the not found/unmatched 404 behavior resolves the TODO on line 90 - `routeNN` should be added as an additional argument to `kubernetes.GatewayAndHTTPRoutesMustBeReady` because the route should be `Accepted` for this difference to happen.

conformance/tests/httproute-invalid-backendref-unknown-kind.go
```diff
@@ -0,0 +1,88 @@
...
+
+			kubernetes.HTTPRouteMustHaveParents(t, suite.Client, routeNN, parents, false, 60)
+		})
+
+		// TODO(mikemorris): Add check for Listener attached routes or
+		// Listener ResolvedRefs RefNotPermitted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
+
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN))
```
mikemorris, line 76:
```suggestion
		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
```
Resolves the TODO on line 49

conformance/tests/httproute-invalid-backendref-unknown-kind.go
```diff
@@ -0,0 +1,88 @@
...
+		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-backend-ref", Namespace: "gateway-conformance-infra"}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
+
+		ns := v1alpha2.Namespace(gwNN.Namespace)
+		kind := v1alpha2.Kind("Gateway")
+
+		// TODO(mikemorris): Add check for Accepted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
+		t.Run("Route status should have a route parent status with a ResolvedRefs condition with status False and reason RefNotPermitted", func(t *testing.T) {
```
mikemorris, line 52:
```suggestion
		t.Run("Route status should have a route parent status with a ResolvedRefs condition with status False and reason InvalidKind", func(t *testing.T) {
```

conformance/tests/httproute-invalid-backendref-nonexistent.go
```diff
@@ -0,0 +1,88 @@
...
+
+			kubernetes.HTTPRouteMustHaveParents(t, suite.Client, routeNN, parents, false, 60)
+		})
+
+		// TODO(mikemorris): Add check for Listener attached routes or
+		// Listener ResolvedRefs RefNotPermitted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
+
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN))
```
mikemorris, line 76:
```suggestion
		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeReady(t, suite.Client, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
```

conformance/tests/httproute-invalid-backendref-nonexistent.go
```diff
@@ -0,0 +1,88 @@
...
+		routeNN := types.NamespacedName{Name: "invalid-nonexistent-backend-ref", Namespace: "gateway-conformance-infra"}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
+
+		ns := v1alpha2.Namespace(gwNN.Namespace)
+		kind := v1alpha2.Kind("Gateway")
+
+		// TODO(mikemorris): Add check for Accepted condition once
+		// https://github.com/kubernetes-sigs/gateway-api/issues/1112
+		// has been resolved
+		t.Run("Route status should have a route parent status with a ResolvedRefs condition with status False and reason RefNotPermitted", func(t *testing.T) {
```
mikemorris, line 52:
```suggestion
		t.Run("Route status should have a route parent status with a ResolvedRefs condition with status False and reason BackendNotFound", func(t *testing.T) {
```

conformance/tests/httproute-invalid-backendref-nonexistent.go
```diff
@@ -0,0 +1,88 @@
...
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteInvalidCrossNamespaceBackendRef)
+}
+
+var HTTPRouteInvalidNonExistentBackendRef = suite.ConformanceTest{
+	ShortName:   "HTTPRouteInvalidNonExistentBackendRef",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should set a ResolvedRefs status False with reason RefNotPermitted and return 500 when binding to a Gateway in the same namespace if the route has a BackendRef Service that does not exist",
+	Exemptions: []suite.ExemptFeature{
+		suite.ExemptReferenceGrant,
+	},
```
mikemorris, line 40:
```suggestion
	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should set a ResolvedRefs status False with reason BackendNotFound and return 500 when binding to a Gateway in the same namespace if the route has a BackendRef Service that does not exist",
```

conformance/tests/httproute-invalid-backendref-unknown-kind.go
```diff
@@ -0,0 +1,88 @@
...
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteInvalidCrossNamespaceBackendRef)
+}
+
+var HTTPRouteInvalidBackendRefUnknownKind = suite.ConformanceTest{
+	ShortName:   "HTTPRouteInvalidBackendRefUnknownKind",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should set a ResolvedRefs status False with reason RefNotPermitted when attempting to bind to a Gateway in the same namespace if the route has a BackendRef that points to an unknown Kind.",
+	Exemptions: []suite.ExemptFeature{
+		suite.ExemptReferenceGrant,
+	},
```
mikemorris, line 40:
```suggestion
	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should set a ResolvedRefs status False with reason InvalidKind when attempting to bind to a Gateway in the same namespace if the route has a BackendRef that points to an unknown Kind.",
```

mikemorris (commented):
Will review the proposed docs changes more closely once this is out of draft, but suggested a few tweaks to the conformance tests for now.

Tagging @skriss for review as I'm suggesting restoring the route acceptance check that was removed in https://github.com/kubernetes-sigs/gateway-api/pull/1170#pullrequestreview-980407532

conformance/tests/httproute-invalid-cross-namespace-backend-ref.go
youngnick: Yeah, I agree.

apis/v1alpha2/httproute_types.go
youngnick: Not really in this update, no, but there are a few places where "no valid backends" and "some valid backends" were treated differently, and I wanted to make doubly sure that there's no confusion.

youngnick:
Okay, I think I've covered everything I can for now in here, so marking this ready for review.

The one case I haven't written a conformance test for is the "two backendRefs where one is invalid means that half of the traffic gets a 500", since I think we'll need some big helper changes for that, and I think we should talk about how we do that. I'll open an issue.

youngnick:
Nope, I missed a bunch of things, fixing them first. Pardon my dust.

robscott (commented):
Thanks @youngnick! This mostly LGTM, but I need to spend a bit more time to run through conformance tests.

apis/v1alpha2/httproute_types.go
```diff
@@ -202,24 +202,28 @@ type HTTPRouteRule struct {
...
 	// MUST receive a 500 status code.
 	//
-	// When a BackendRef refers to a Service that has no ready endpoints, it is
-	// recommended to return a 503 status code.
+	// For example, if two backends are specified with equal weights, and one is
+	// invalid, 50 percent of traffic must receive a 500. Implementations may
+	// choose how that 50 percent is determined.
+	//
+	// When a HTTPBackendRef refers to a Service that has no ready endpoints,
+	// implementations MAY return a 503 for requests to that backend instead.
```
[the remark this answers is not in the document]
youngnick: Agreed, it's not very clear. I guess our two options are either make it mandatory, or just mandate 500 for everything. I'm leaning towards the latter, so I'll just remove this.

apis/v1alpha2/shared_types.go
```diff
@@ -242,6 +242,15 @@ const (
...
 	// not have a ReferenceGrant explicitly allowing the reference.
 	RouteReasonRefNotPermitted RouteConditionReason = "RefNotPermitted"
+
+	// This reason is used with the "ResolvedRefs" condition when
+	// one of the Route's rules has a reference to an unknown or unsupported
+	// Group and/or Kind.
+	RouteReasonInvalidKind RouteConditionReason = "InvalidKind"
+
+	// This reason is used with the "ResolvedRefs" condition when
+	// one of the Route's rules has a reference to a resource that does not exist.
```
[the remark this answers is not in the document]
youngnick: Manual justification of text, sigh. Fixed.

apis/v1beta1/httproute_types.go
```diff
@@ -201,26 +201,31 @@ type HTTPRouteRule struct {
 	// BackendRefs defines the backend(s) where matching requests should be
 	// sent.
 	//
-	// A 500 status code MUST be returned if there are no BackendRefs or filters
-	// specified that would result in a response being sent.
+	// Failure behavior here depends on how many BackendRefs are specified and
+	// how many are invalid.
```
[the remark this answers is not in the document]
youngnick: Yeah, I almost forgot to update the v1beta1 definitions, that seems like it would be very easy to do.

conformance/tests/httproute-invalid-backendref-unknown-kind.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: HTTPRoute
+metadata:
+  name: invalid-cross-namespace-backend-ref
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - backendRefs:
+    - group: unknownkind.io
```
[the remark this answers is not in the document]
youngnick: of course, yeah.

conformance/utils/kubernetes/helpers.go
```diff
@@ -342,6 +343,38 @@ func GatewayStatusMustHaveListeners(t *testing.T, client client.Client, gwNN typ
 	require.NoErrorf(t, waitErr, "error waiting for Gateway status to have listeners matching expectations")
 }
 
+// HTTPRouteMustHaveConditions checks that the supplied HTTPRoute has the supplied Condition,
+// halting after the specified timeout is exceeded.
+func HTTPRouteMustHaveCondition(t *testing.T, client client.Client, routeNN types.NamespacedName, condition metav1.Condition, seconds int) {
```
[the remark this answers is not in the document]
youngnick: Good point, I changed this to take the Gateway name it should be checking for.

conformance/utils/kubernetes/helpers.go
```diff
@@ -390,13 +424,16 @@ func conditionsMatch(t *testing.T, expected, actual []metav1.Condition) bool {
...
 }
 
-func findConditionInList(t *testing.T, conditions []metav1.Condition, condName, condValue string) bool {
+func findConditionInList(t *testing.T, conditions []metav1.Condition, condName, condValue, condReason string) bool {
 	for _, cond := range conditions {
 		if cond.Type == condName {
-			// TODO(nathancoleman): Compare Reason
 			if cond.Status == metav1.ConditionStatus(condValue) {
-				return true
+				if cond.Reason == condReason {
```
skriss, line 431: I'm seeing a couple of tests start failing with this PR that I think should still pass:

- httproute-disallowed-kind
- httproute-invalid-cross-namespace-parent-ref

These both use the [HTTPRouteMustHaveNoAcceptedParents function](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/kubernetes/helpers.go#L214), which asserts that a Route has either no Accepted condition, or an "Accepted: false" condition. 

At least in Contour's implementation, we're setting an "Accepted: false" condition in both of these scenarios. The problem arises because `HTTPRouteMustHaveNoAcceptedParents` [does not pass a reason](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/kubernetes/helpers.go#L245-L248) when checking for an "Accepted: false" condition, so `findConditionsInList` expects the actual condition to have an empty reason.

Not sure what the best fix is here, but one option would be to consider an empty reason string to mean "match any reason" instead of "match an empty string"? Otherwise, we might have to plumb a reason through from the place where `HTTPRouteMustHaveNoAcceptedParents` is called.

youngnick: Yeah, I agree that having an empty string passed for Reason can mean "match any Reason", since Reason strings should not normally be empty.

conformance/tests/httproute-invalid-reference-grant.go
```diff
@@ -43,69 +43,39 @@ var HTTPRouteInvalidReferenceGrant = suite.ConformanceTest{
...
+			})
+		})
+
+		t.Run("HTTP Request to valid sibling backend should succeed", func(t *testing.T) {
+			http.MakeRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, gwAddr, http.ExpectedResponse{
+				Request: http.Request{
+					Method: "GET",
+					Path:   "/",
+				},
+				StatusCode: 200,
```
skriss, line 76:
```suggestion
				StatusCode: 200,
				Backend: "app-backend-v1",
				Namespace: "gateway-conformance-app-backend",
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -342,6 +343,39 @@ func GatewayStatusMustHaveListeners(t *testing.T, client client.Client, gwNN typ
...
+		err := client.Get(ctx, routeNN, route)
+		if err != nil {
+			return false, fmt.Errorf("error fetching HTTPRoute: %w", err)
+		}
+
+		parents := route.Status.Parents
+
+		var conditionFound bool
+		for _, parent := range parents {
+			if parent.ParentRef.Name == v1alpha2.ObjectName(gwNN.Name) && string(*parent.ParentRef.Namespace) == gwNN.Namespace {
```
skriss, line 366:
```suggestion
			if parent.ParentRef.Name == v1alpha2.ObjectName(gwNN.Name) && (parent.ParentRef.Namespace == nil || string(*parent.ParentRef.Namespace) == gwNN.Namespace) {
```

skriss (changes requested):

conformance/utils/kubernetes/helpers.go
```diff
@@ -81,7 +81,7 @@ func GWCMustBeAccepted(t *testing.T, c client.Client, gwcName string, seconds in
 		}
 
 		controllerName = string(gwc.Spec.ControllerName)
-		return findConditionInList(t, gwc.Status.Conditions, "Accepted", "True"), nil
+		return findConditionInList(t, gwc.Status.Conditions, "Accepted", "True", "Accepted"), nil
```
robscott, line 84: I don't think we _need_ to require reason to exactly match the condition name when it is in a success state. For example, when Istio sets the Ready condition, it sets the reason to "ListenersValid". Adding this kind of change will cause any implementations like that to immediately fail all conformance tests since this is part of the fundamental check if the base resources are ready. I think we could consider adding this check in the future if interested, but would prefer to leave it open to any reason now to avoid significant disruption before a release.

conformance/utils/kubernetes/helpers.go
```diff
@@ -106,7 +106,7 @@ func NamespacesMustBeReady(t *testing.T, c client.Client, namespaces []string, s
 				t.Errorf("Error listing Gateways: %v", err)
 			}
 			for _, gw := range gwList.Items {
-				if !findConditionInList(t, gw.Status.Conditions, "Ready", "True") {
+				if !findConditionInList(t, gw.Status.Conditions, "Ready", "True", "Ready") {
```
robscott, line 109: Same comment as above, would prefer "" instead of "Ready" as the last arg. (If I understand that code, that will allow any reason).

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: https://github.com/kubernetes-sigs/gateway-api/issues/1256

robscott (commented):
Thanks @youngnick! This mostly LGTM, just a couple comments.

skriss (approved):
Thanks for the updates @youngnick, this LGTM

conformance/utils/kubernetes/helpers.go
youngnick: I changed this to the empty string, with a comment.

conformance/utils/kubernetes/helpers.go
youngnick: I changed this to the empty string, with a comment.

robscott:
Thanks @youngnick!

/lgtm

PR #1245, opened 2022-06-30
Update examples to v1beta1

EmilyShepherd:
**What type of PR is this?**
/kind documentation
/hold
(hold until the recommended version is >= 0.5.0)

This change was originall part of #1238, but it was agreed that updating the documentation's examples themselves should wait until 0.5.0 is the recommended version (currently it is still 0.4.x). This PR is those remaining changes which can be merged as and when they are needed.

This updates all examples to use v1beta1 variants, where possible, and links to the API Specification to use the v1beta1 versions.

It does **not** do any of the following:

- GEPs are not touched
- The "introducing v1alpha2" blog post still uses v1alpha2 for all of its examples
- Examples that require the use of v1alpha2 resources, will continue to use v1alpha2 variants for their whole example, even for beta-graduated resources like the `Gateway`. This is in line with the [advice][api-advice] to continue to use v1alpha2 internally, if you want to use alpha features. (The main examples of this are `guides/tcp` which uses exclusively v1alpha2 examples, and `guides/tls` which will use v1beta1 for examples that can be impleneted using _just_ v1beta1 resources, or v1alpha2 for examples that require the alpha Reference Grant).

[api-advice]: https://kubernetes.slack.com/archives/CR0H13KGA/p1656533075102939

```release-note
Update documentation examples to v1beta1
```

### DIFF
--- a/examples/experimental/v1alpha2/http-redirect-rewrite/httproute-redirect-full.yaml
+++ b/examples/experimental/v1alpha2/http-redirect-rewrite/httproute-redirect-full.yaml
@@ -1,5 +1,3 @@
-#$ Used in:
-#$ - site-src/v1alpha2/api-types/httproute.md
 apiVersion: gateway.networking.k8s.io/v1alpha2
 kind: HTTPRoute
 metadata:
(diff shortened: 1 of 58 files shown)


robscott:
Thanks @EmilyShepherd!

/lgtm
/approve
/hold until release

PR #1252, opened 2022-07-08
Returning NamedAddress type and marking it deprecated

robscott:
**What type of PR is this?**
/kind bug
/kind cleanup

This returns our NamedAddress type and marks it deprecated. Unfortunately I think we went a step too far when we removed it altogether. Although we've left room to add new values to enums in new minor versions, we have not provided any precedent for removing supported enum values. Unfortunately I think removing "NamedAddress" from the list of supported values would make it so HTTPRoutes that were valid in v0.4.x would not be valid in v0.5.x, which is something we should strive to avoid. 

I think the best we can do here is continue to allow the value to be set but document that it has been deprecated in favor of domain prefixed strings. Hopefully this approach makes sense.

```release-note
NamedAddress type is back to support backwards compatibility but is formally deprecated.
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -488,13 +488,16 @@ type AnnotationValue string
 //
 // Values `IPAddress` and `Hostname` have Extended support.
 //
+// The `NamedAddress` value has been deprecated in favor of implementation
+// specific domain-prefixed strings.
+//
 // All other values, including domain-prefixed values have Custom support, which
 // are used in implementation-specific behaviors. Support for additional
 // predefined CamelCase identifiers may be added in future releases.
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=253
-// +kubebuilder:validation:Pattern=`^Hostname|IPAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
+// +kubebuilder:validation:Pattern=`^Hostname|IPAddress|NamedAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type AddressType string
 
 const (
@@ -517,4 +520,14 @@ const (
 	//
 	// Support: Extended
 	HostnameAddressType AddressType = "Hostname"
+
+	// A NamedAddress provides a way to reference a specific IP address by name.
+	// For example, this may be a name or other unique identifier that refers
+	// to a resource on a cloud provider such as a static IP.
+	//
+	// The `NamedAddress` type has been deprecated in favor of implementation
+	// specific domain-prefixed strings.
+	//
+	// Support: Implementation-Specific
+	NamedAddressType AddressType = "NamedAddress"
 )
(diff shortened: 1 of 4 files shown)


robscott:
Although relatively tiny in scope, this undoes part of a previous change, so want to make sure there's some consensus here.

/hold

apis/v1alpha2/shared_types.go
```diff
@@ -483,18 +483,21 @@ type AnnotationValue string
 // AddressType defines how a network address is represented as a text string.
 // This may take two possible forms:
 //
-// * A predefined CamelCase string identifier (currently limited to `IPAddress` or `Hostname`)
+// * A predefined CamelCase string identifier (currently limited to `IPAddress` or  `Hostname`)
```
shaneutt, line 486: This seems to just be adding extra blank space?

```suggestion
// * A predefined CamelCase string identifier (currently limited to `IPAddress` or `Hostname`)
```

shaneutt:
I'm not necessarily strongly for or against adding this back. For posterity how many implementations were wanting this to be added back and what was the reasoning if other than "need more time to update"?

apis/v1alpha2/shared_types.go
robscott: Woops, fixed!

apis/v1beta1/shared_types.go
```diff
@@ -483,18 +483,21 @@ type AnnotationValue string
 // AddressType defines how a network address is represented as a text string.
 // This may take two possible forms:
 //
-// * A predefined CamelCase string identifier (currently limited to `IPAddress` or `Hostname`)
+// * A predefined CamelCase string identifier (currently limited to `IPAddress` or  `Hostname`)
```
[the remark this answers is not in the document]
robscott: Woops, fixed!

robscott:
>For posterity how many implementations were wanting this to be added back and what was the reasoning if other than "need more time to update"?

My primary argument here is that removing it would break our backwards compatibility guarantees. In general we've said we can loosen validation or add new valid values, but if something was at some point considered valid, we will not prevent it from being set in the future unless it was very clearly a bug in our validation. In this case it represented a defined and valid value, so I don't think we can get away with removing it.

We have removed suggested reasons for status from the API, but I think the important distinction there is that we still allow implementations to populate those reasons, we just don't define them as suggested reasons anymore. 

In upstream Kubernetes we've sometimes tweaked defaults between API versions, but I don't think we've ever tightened validation.

shaneutt:
> My primary argument here is that removing it would break our backwards compatibility guarantees. In general we've said we can loosen validation or add new valid values, but if something was at some point considered valid, we will not prevent it from being set in the future unless it was very clearly a bug in our validation. In this case it represented a defined and valid value, so I don't think we can get away with removing it.

It's common to _not_ have backwards compatibility guarantees when you're at an alpha level maturity. Some might consider that part of the reason to have these kinds of maturity levels: to promote quick iterations and to not get stuck with decisions in the early days of a project. If nobody is specifically complaining, isn't removing it just a reduction in burden for the project that frees up (albeit probably only a little) time for other things?

I'm going to approve as I understand and can appreciate your reasoning, and I don't have a reason to block. I do however have a preference (maybe it would be better to say, it's my default approach) to make alpha a place that moves quickly and isn't afraid to rip stuff out as that approach can have a have an influence on velocity and just wanted to share my sentiment though I don't request any action taken for the purposes of this PR.

nathancoleman (approved):

robscott:
> I do sometimes wonder if it would be better to take a harder and faster approach at these earlier maturity states of a project, but this still ends up being a pretty minor thing in either case.

Yeah I generally agree with that approach, but unfortunately my understanding of k8s API guidelines is that although we could make a breaking change in alpha, that would need to be followed by another alpha API version, not a beta API version. I don't think this is significant enough of a change to merit another alpha API release.

robscott:
Discussed this at today's community meeting and the change had broad consensus. As mentioned above, this change is to preserve backwards compatibility, the type should still be considered deprecated.  Removing the hold on this.

/hold cancel

shaneutt (approved):
/lgtm

PR #1258, opened 2022-07-12
Enum details update

### LINKED ISSUE #1014: Update documentation for use of kubebuilder Enum tags
The [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) strongly recommend against using OpenAPI Enum validation, because there is a strong implication that the set of enumerated values is immutable; implementations can reasonably expect that the entire set of an Enum is known. Because of this, for the core APIs, adding a value to or removing a value from an Enum is a breaking API change and requires an API version bump.

Before we go to beta, I'd like us to discuss our use of kubebuilder Enum tags, and whether it will make version management easier for these types to:
- provide an aliased string type (we already do this in most places I can see)
- provide constants for the valid/Enum values
- document that the constants are supported.

I can see an argument that because we have the experimental channel, the rules apply less, but I'm pretty reluctant to disregard lessons learned in the core APIs without a very good reason.

/kind design
I think?

youngnick:
**What type of PR is this?**
/kind cleanup
/kind documentation

This clarifies the rules around enum values to ensure that we document that additional values may be added in the future, and how implementations must handle unknown values.

I've left conformance tests for this behavior out for now, because it's a bit complex to test with (we'll need a way to push invalid things into Kube or some other workaround).

Fixes #1014 

```release-note
All Enum API fields have had updates to clarify that we may add values at any time, and that implementations must handle unknown Enum values.
```

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -219,7 +219,7 @@ const (
 
 	// This reason is used with the "Accepted" condition when a value for an Enum
 	// is not recognized.
-	RouteReasonInvalidEnumValue RouteConditionReason = "InvalidEnumValue"
+	RouteReasonUnsupportedValue RouteConditionReason = "UnsupportedValue"
 
 	// This condition indicates whether the controller was able to resolve all
 	// the object references for the Route.
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -217,6 +217,10 @@ const (
 	// compatible Listeners whose Hostname matches the route
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
+	// This reason is used with the "Accepted" condition when a value for an Enum
+	// is not recognized.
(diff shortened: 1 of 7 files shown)


apis/v1beta1/shared_types.go
```diff
@@ -217,6 +217,10 @@ const (
 	// compatible Listeners whose Hostname matches the route
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
+	// This reason is used with the "Accepted" condition when a value for an Enum
+	// is not recognized.
+	RouteReasonInvalidEnumValue RouteConditionReason = "InvalidEnumValue"
```
robscott, line 222: Should we make this more generic? Maybe just "UnknownValue" or "UnsupportedValue"?

robscott (commented):
Thanks @youngnick! This mostly LGTM, just one question.

apis/v1beta1/shared_types.go
youngnick: Yeah, that makes sense, I'll update.

robscott:
Thanks @youngnick!

/lgtm

PR #1276, opened 2022-07-20
Bumping versions to v0.6.0-dev

robscott:
**What type of PR is this?**
/kind cleanup

Now that we've released v0.5.0, we need to bump our version refs throughout to avoid future API changes being included in CRDs labeled "v0.5.0".

```release-note
NONE
```

### DIFF
--- a/config/crd/experimental/gateway.networking.k8s.io_gatewayclasses.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_gatewayclasses.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/gateway-api/pull/1086
-    gateway.networking.k8s.io/bundle-version: v0.5.0
+    gateway.networking.k8s.io/bundle-version: v0.6.0-dev
     gateway.networking.k8s.io/channel: experimental
   creationTimestamp: null
   name: gatewayclasses.gateway.networking.k8s.io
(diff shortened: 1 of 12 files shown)


shaneutt (commented):
So with this change people deploying from `main` wont get a `v0.5.0` version tag, when what they're really getting is `main`. Some features could make it into `main` but then back out, meaning that at some point you might deploy `v0.6.0-dev` which would have a feature in it, but then `v0.6.0` would not. Not super likely to happen or be a problem, but just for extra special clarification, would it be reasonable to tag everything as `main` instead? Would that not be more clear about what they're getting?

robscott:
> So with this change people deploying from main wont get a v0.5.0 version tag

Yep, that's the main goal here, don't want to have things labeled as v0.5.0 that are not that release.

> Some features could make it into main but then back out, meaning that at some point you might deploy v0.6.0-dev which would have a feature in it, but then v0.6.0 would not

I personally think that's fine and should be assumed when using something tagged "dev", but I can understand how that could be a poor experience for someone. Maybe we just need clearer documentation around this?

> would it be reasonable to tag everything as main instead?

I don't feel that strongly about name, but I slightly prefer the `-dev` suffix as it represents the release we're working toward and provides some level of context as far as how old the CRD may be. For reference, this matches what we did before the `v0.5.0` release where we had a `v0.5.0-dev` value here. I think one downside of using `main` as the name is that it could be interpreted as communicating a sense of stability - ie someone might thing that this is that "main" set of CRDs for the project.

I'm fine to add more docs or choose a different name, but I do want to get something in ASAP that will remove the v0.5.0 references from our CRDs so we don't have multiple variations of v0.5.0 CRDs floating around.

shaneutt (approved):
I have a slight preference for a different name like `main` where down the road we could start branching for releases rather than having to do manual changes like this, but given that this sentiment isn't shared I also don't feel it's worth blocking over.

/approve

youngnick (commented):
/lgtm

PR #1278, opened 2022-07-20
Clarifying which kinds of bugs should be filed in this repo

robscott:
**What type of PR is this?**
/kind cleanup

We've already gotten some implementation specific bugs filed in this repo, that will likely become more common as this project grows. Hopefully adding this kind of clarification will help users find the correct place to file each kind of bug.

```release-note
NONE
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -4,7 +4,20 @@ about: Report a bug encountered with the gateway-api project
 labels: kind/bug
 
 ---
-<!-- Please use this template while reporting a bug and provide as much info as
+<!--
+Thank you for your interest in Gateway API! Please note that bug reports 
+here should only be used for bugs with the API itself, such as:
+
+- Incomplete or inaccurate validation
+- Problems with the API specification
+- Unexpected errors with the validating webhook
+- Flaws in conformance tests
+
+Bugs that don't fit into one of the above categories should likely be filed
+with the [implementation](https://gateway-api.sigs.k8s.io/implementations/)
+of Gateway API that you're using.
+
+Please use this template while reporting a bug and provide as much info as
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -4,7 +4,19 @@ about: Report a bug encountered with the gateway-api project
...
+Note: Bug reports here should only be used for bugs with the API itself, such
+as:
+
+- Incomplete or inaccurate validation
+- Problems with the API specification
+- Unexpected errors with the validating webhook
+- Flaws in conformance tests
+
+Bugs that don't fit into one of the above categories should likely be filed
+with the implementation of Gateway API you're using.
```
markmc, line 17:
```suggestion
with the [implementation](https://gateway-api.sigs.k8s.io/implementations/) of Gateway API you're using.
```

The implementations page might help the reporter find where to file their issue

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -4,7 +4,19 @@ about: Report a bug encountered with the gateway-api project
 labels: kind/bug
 
 ---
-<!-- Please use this template while reporting a bug and provide as much info as
+<!--
```
markmc, line 7: Since we're about so say "go away!" maybe we could lead with "Thank you for your interest in Gateway API!" ?

.github/ISSUE_TEMPLATE/bug.md
robscott: Updated, thanks!

.github/ISSUE_TEMPLATE/bug.md
robscott: Good call, updated, thanks!

PR #1283, opened 2022-07-26
Rectified Conformance Test Name Generation

### LINKED ISSUE #1272: Conformance Test Name Generation is Flawed
**What happened**:
The test names for HTTP Header Matching in particular are not very helpful. For example, here's the output from one test run:

```
        --- PASS: TestConformance/HTTPRouteHeaderMatching/9_request_to_/_with_headers_should_go_to_ (0.07s)
        --- PASS: TestConformance/HTTPRouteHeaderMatching/4_request_to_/_with_headers_should_go_to_ (0.07s)
        --- PASS: TestConformance/HTTPRouteHeaderMatching/3_request_to_/_with_headers_should_go_to_ (0.08s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/5_request_to_/_with_headers_should_go_to_infra-backend-v1 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/8_request_to_/_with_headers_should_go_to_infra-backend-v2 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/0_request_to_/_with_headers_should_go_to_infra-backend-v1 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/6_request_to_/_with_headers_should_go_to_infra-backend-v1 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/7_request_to_/_with_headers_should_go_to_infra-backend-v2 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/2_request_to_/_with_headers_should_go_to_infra-backend-v1 (30.00s)
        --- FAIL: TestConformance/HTTPRouteHeaderMatching/1_request_to_/_with_headers_should_go_to_infra-backend-v2 (30.00s)
```

Test names to make sense.

Run conformance tests.

I think there are at least two fundamental problems here:

1. When Backend is nil we should look to see if a status code has been specified and use that instead: https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-header-matching.go#L57 
2. We probably want to find a better way to represent "/" in our test case names. As it is, it looks like "2_request_to_" has a subtest named "\_with_headers_should_go_to_infra-backend-v1". I'm not sure what the best solution is here, but the following steps could provide some improvement here:
  A. Remove the first "/" character from paths, replace the rest with another character such as "_"
  B. Update test cases to be able to optionally specify their own test name. When specified, it would be used instead of the generated one.

For reference, our current testName function is here:
https://github.com/kubernetes-sigs/gateway-api/blob/9537337d838cfebe3775c51be83c486a127c0d88/conformance/tests/httproute-matching-across-routes.go#L118-L125

In addition to the updates suggested above, it should likely be moved to some kind of more clearly shared utils/helpers file as part of these updates.

abhijit-dev82:
**What type of PR is this?**
/kind cleanup

Conformance Test Name Generation is Flawed

Fixes #1272 

No

### DIFF
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -88,18 +88,24 @@ effort, check out the #development channel or join our [weekly developer meeting
 
 ### Contour
 
-[Contour][contour] is an open source ingress controller for Kubernetes.
+[Contour][contour] is a CNCF open source Envoy-based ingress controller for Kubernetes.
 
-Contour currently implements much of Gateway API `v0.3.0`.
+Contour implements Gateway API v0.5.0, supporting the v1alpha2 and v1beta1 API versions.
+All [Standard channel][contour-standard] resources (GatewayClass, Gateway, HTTPRoute), plus ReferenceGrant and TLSRoute, are supported.
+Contour's implementation passes all Gateway API conformance tests included in the v0.5.0 release.
 
-See the [Contour Gateway API Guide][contour-1] for information on how to deploy and use Contour's Gateway implementation.
+See the [Contour Gateway API Guide][contour-guide] for information on how to deploy and use Contour's Gateway API implementation.
 
-_Note that not all of the `v0.3.0` specification is completed yet, [contributions welcome!][contour-2]._
+For help and support with Contour's implementation, [create an issue][contour-issue-new] or ask for help in the [#contour channel on Kubernetes slack][contour-slack].
+
+_Some "extended" functionality is not implemented yet, [contributions welcome!][contour-contrib]._
 
 [contour]:https://projectcontour.io
-[latest]:https://doc.traefik.io/traefik/routing/providers/kubernetes-gateway/
-[contour-1]:https://projectcontour.io/guides/gateway-api/
-[contour-2]:https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md
+[contour-standard]:https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels-eg-experimental-standard
+[contour-guide]:https://projectcontour.io/guides/gateway-api/
+[contour-issue-new]:https://github.com/projectcontour/contour/issues/new/choose
+[contour-slack]:https://kubernetes.slack.com/archives/C8XRH2R4J
+[contour-contrib]:https://github.com/projectcontour/contour/blob/main/CONTRIBUTING.md
 
 ### Emissary-Ingress (Ambassador API Gateway)
 


conformance/tests/httproute-matching-across-routes.go
```diff
@@ -116,10 +116,32 @@ var HTTPRouteMatchingAcrossRoutes = suite.ConformanceTest{
...
+	} else {
+		/* Request to non root path */
+		tc.Request.Path = "'" + tc.Request.Path + "'"
+		if tc.Request.Headers != nil {
+			headerStr = " with headers"
+		}
+		reqStr = fmt.Sprintf("%d request to %s%s%s", i, tc.Request.Host, tc.Request.Path, headerStr)
 	}
 
-	return fmt.Sprintf("%d request to %s%s%s should go to %s", i, tc.Request.Host, tc.Request.Path, headerStr, tc.Backend)
```
robscott, line 124: Unfortunately this isn't quite working as I'd hoped:

```
TestConformance/HTTPRouteHostnameIntersection/HTTPRoutes_that_do_intersect_with_listener_hostnames/20_request_to_anotherwildcard.io'/s4'_should_go_to_404
```

I think it makes more sense just to put the path and hostname in quotes. It may be simpler if we replace all of this with:

``` 
fmt.Sprintf("%d request to '%s%s'%s should go to", i, tc.Request.Host, tc.Request.Path, headerStr)
```

We'd still need the fix for StatusCode support below, up to you how to integrate that.

robscott (commented):
Thanks @abhijit-dev82!

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: It would look like the below for root path.

 --- PASS: TestHTTPReqString/9_request_to_'/'_with_headers_should_go_to_404 (0.00s)
    --- PASS: TestHTTPReqString/5_request_to_'/'_with_headers_should_go_to_infra-backend-v1 (0.00s)

Non root path
    --- PASS: TestHTTPReqString/8_request_to_'/foo/try'_with_headers_should_go_to_infra-backend-v2 (0.00s)

Is that ok?

conformance/tests/httproute-matching-across-routes.go
```diff
@@ -116,10 +116,23 @@ var HTTPRouteMatchingAcrossRoutes = suite.ConformanceTest{
...
+
 	if tc.Request.Headers != nil {
 		headerStr = " with headers"
 	}
 
-	return fmt.Sprintf("%d request to %s%s%s should go to %s", i, tc.Request.Host, tc.Request.Path, headerStr, tc.Backend)
+	reqStr = fmt.Sprintf("%d request to '%s%s'%s should go to", i, tc.Request.Host, tc.Request.Path, headerStr)
+
+	if tc.Backend != "" {
+		return fmt.Sprintf("%s %s", reqStr, tc.Backend)
```
robscott, line 135:
```suggestion
		return fmt.Sprintf("%s should go to %s", reqStr, tc.Backend)
```

conformance/tests/httproute-matching-across-routes.go
```diff
@@ -116,10 +116,23 @@ var HTTPRouteMatchingAcrossRoutes = suite.ConformanceTest{
...
+	}
+	return fmt.Sprintf("%s %d", reqStr, tc.StatusCode)
```
robscott, line 137:
```suggestion
	return fmt.Sprintf("%s should receive a %d", reqStr, tc.StatusCode)
```

conformance/tests/httproute-matching-across-routes.go
```diff
@@ -116,10 +116,23 @@ var HTTPRouteMatchingAcrossRoutes = suite.ConformanceTest{
...
+
 	headerStr := ""
+	reqStr := ""
+
 	if tc.Request.Headers != nil {
 		headerStr = " with headers"
 	}
 
-	return fmt.Sprintf("%d request to %s%s%s should go to %s", i, tc.Request.Host, tc.Request.Path, headerStr, tc.Backend)
+	reqStr = fmt.Sprintf("%d request to '%s%s'%s should go to", i, tc.Request.Host, tc.Request.Path, headerStr)
```
robscott, line 132:
```suggestion
	reqStr = fmt.Sprintf("%d request to '%s%s'%s", i, tc.Request.Host, tc.Request.Path, headerStr)
```

conformance/tests/httproute-matching-across-routes.go
```diff
@@ -116,10 +116,23 @@ var HTTPRouteMatchingAcrossRoutes = suite.ConformanceTest{
 }
 
 func testName(tc http.ExpectedResponse, i int) string {
```
robscott, line 118: I think this function should probably be moved. Since it's dependent on `http.ExpectedResponse`, we could move it to the file that defines that and make a function like this:


```
func (er *ExpectedResponse) TestCaseName (num int) {
   ...
}
```

robscott (commented):
Thanks for the work on this @abhijit-dev82! This mostly LGTM. Do you mind rebasing/squashing to remove the merge commits in this PR?

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: Modified as per the comment.

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: Modified as per the comment.

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: Modified as per the comment.

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: Moved the function to utils/http/http.go

conformance/tests/httproute-matching-across-routes.go
abhijit-dev82: Thank you @robscott for guiding me through my first commit to a k8s-sig project. Looking forward to contribute more to the community.

PR #1290, opened 2022-08-01
Rectified Conformance Test Name Generation

abhijit-dev82:
What type of PR is this?
/kind cleanup

What this PR does / why we need it:
Conformance Test Name Generation is Flawed

Which issue(s) this PR fixes:

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1272

Does this PR introduce a user-facing change?:
No

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -40,6 +40,9 @@ type ExpectedResponse struct {
 	StatusCode int
 	Backend    string
 	Namespace  string
+
+	// User Given TestCase name
+	TestCaseName string
 }
 
 // Request can be used as both the request to make and a means to verify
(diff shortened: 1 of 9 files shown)


abhijit-dev82:
While resolving merge conflicts the original https://github.com/kubernetes-sigs/gateway-api/pull/1283 got closed, hence had to open a new PR.

conformance/utils/http/http.go
```diff
@@ -229,3 +232,26 @@ func CompareRequest(cReq *roundtripper.CapturedRequest, cRes *roundtripper.Captu
 	}
 	return nil
 }
+
+// Get User-defined test case name or generate from expected response to a given request.
+func (er *ExpectedResponse) GetTestCaseName(i int) string {
```
robscott, line 237: Tiny nit: Would expect this to be called "TestCaseName":
```suggestion
func (er *ExpectedResponse) TestCaseName(i int) string {
```

(https://go.dev/doc/effective_go#Getters)

conformance/utils/http/http.go
```diff
@@ -229,3 +232,26 @@ func CompareRequest(cReq *roundtripper.CapturedRequest, cRes *roundtripper.Captu
...
+
+	/* If TestCase name is provided then use that or else generate one */
```
robscott, line 239: Tiny nit: Would expect this to be a line comment (`//` instead of `/*`) per https://go.dev/doc/effective_go#commentary.

robscott (commented):
Thanks @abhijit-dev82! This LGTM, works well for me, just a couple tiny nits.

/ok-to-test

conformance/utils/http/http.go
abhijit-dev82: Had kept it "GetTestCaseName" as had added a new field in struct ExpectedResponse "TestCaseName" to carry the user given test case name. Will change the field name to "UsrTestCaseName" and the method name to "TestCaseName"

conformance/utils/http/http.go
abhijit-dev82: Corrected.

conformance/utils/http/http.go
robscott: Good point, I like your original approach in that case with the field named "TestCaseName" and the func named "GetTestCaseName"

robscott:
Thanks @abhijit-dev82!

/lgtm
/approve

PR #1292, opened 2022-08-02
make timeout configurable

### LINKED ISSUE #1287: Conformance Tests Should Support Configurable Timeouts
When running conformance tests, it can be helpful to adjust timeouts to understand what is causing a test failure. As part of making conformance tests configurable, we should also document any maximum timeouts that are acceptable for a conformant implementation.

wstcliyu:
**What type of PR is this?**
/kind feature

make timeout configurable

Fixes #1287

Yes. It enables supplying timeout configuration.

### DIFF
--- a/conformance/utils/config/timeout.go
+++ b/conformance/utils/config/timeout.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import "time"
+
+type TimeoutConfig struct {
+	// CreateTimeout represents the maximum time for a Kubernetes object to be created.
+	// Max value for conformant implementation: None
+	CreateTimeout time.Duration
+
+	// DeleteTimeout represents the maximum time for a Kubernetes object to be deleted.
+	// Max value for conformant implementation: None
+	DeleteTimeout time.Duration
+
+	// GatewayMustHaveAddress represents the maximum time for at least one IP Address has been set in the status of a Gateway.
+	// Max value for conformant implementation: None
+	GatewayMustHaveAddress time.Duration
+
+	// GatewayStatusMustHaveListeners represents the maximum time for a Gateway to have listeners in status that match the expected listeners.
+	// Max value for conformant implementation: None
+	GatewayStatusMustHaveListeners time.Duration
+
+	// GWCMustBeAccepted represents the maximum time for a GatewayClass to have an Accepted condition set to true.
+	// Max value for conformant implementation: None
+	GWCMustBeAccepted time.Duration
+
+	// HTTPRouteMustNotHaveParents represents the maximum time for an HTTPRoute to have either no parents or a single parent that is not accepted.
+	// Max value for conformant implementation: None
+	HTTPRouteMustNotHaveParents time.Duration
+
+	// HTTPRouteMustHaveCondition represents the maximum time for an HTTPRoute to have the supplied Condition.
+	// Max value for conformant implementation: None
+	HTTPRouteMustHaveCondition time.Duration
+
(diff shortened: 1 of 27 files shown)


robscott:
Thanks @wstcliyu!

/ok-to-test

conformance/utils/config/timeout.go
```diff
@@ -0,0 +1,86 @@
...
+package config
+
+import "time"
+
+type TimeoutConfig struct {
+	CreateTimeout                  time.Duration // used by MustApplyObjectsWithCleanup, MustApplyWithCleanup
+	DeleteTimeout                  time.Duration // used by MustApplyObjectsWithCleanup, MustApplyWithCleanup
+	GatewayMustHaveAddress         time.Duration // used by WaitForGatewayAddress
+	GatewayStatusMustHaveListeners time.Duration // used by GatewayStatusMustHaveListeners
+	GWCMustBeReady                 time.Duration // used by GWCMustBeAccepted
```
robscott, line 26: I think "GWCMustBeAccepted" would be more accurate here.

robscott (commented):
Thanks for the work on this @wstcliyu!

conformance/utils/config/timeout.go
```diff
@@ -0,0 +1,129 @@
...
+	// Max value for conformant implementation: None
+	// Used by: getContentsFromPathOrURL
+	ManifestFetchTimeout time.Duration
+
+	// NamespacesMustBeReady represents the maximum time for all Pods and Gateways in a namespaces to be marked as ready.
+	// Max value for conformant implementation: None
+	// Used by: NamespacesMustBeReady
+	NamespacesMustBeReady time.Duration
+
+	// RequestTimeout represents the maximum time for making an HTTP Request or fetching a Kubernetes object.
```
robscott, line 72: I don't think we should include "fetching a Kubernetes object" here. I think "RequestTimeout" should be reserved for HTTP requests made with our roundtripper. The other timeouts could either have different names or just leave them unchanged/not configurable. The places this PR uses request timeout for fetching k8s objects are probably a bit broad, for example one of those timeouts actually covers the following flow:

- request a list of Gateways
- iterate through to confirm each Gateway in list has ready condition set to true
- request a list of Pod
- iterate through to confirm each Pod in list has ready condition set to true

There's a lot of variety in the other places where RequestTimeout is currently be used and I think it would be difficult to create a generic timeout that covered all of them reasonably well. I think I'd lean towards reverting these specific uses of RequestTimeout to not be configurable and leaving this config for HTTP request timeouts.

conformance/utils/config/timeout.go
```diff
@@ -0,0 +1,129 @@
...
+	// Max value for conformant implementation: None
```
robscott, line 73: There is a related value here that I expected to be configurable. I think this may be the only one that truly needs a "Max value for conformant implementation", and we should probably start with the current default of 30 seconds: https://github.com/kubernetes-sigs/gateway-api/blob/361a156b0705bfbf10accc01b8c3e7424acfbc7e/conformance/utils/http/http.go#L69

conformance/utils/config/timeout.go
```diff
@@ -0,0 +1,129 @@
...
+	// Used by: CaptureRoundTrip, GWCMustBeAccepted, NamespacesMustBeReady, WaitForGatewayAddress, HTTPRouteMustHaveNoAcceptedParents, HTTPRouteMustHaveParents, GatewayStatusMustHaveListeners, HTTPRouteMustHaveCondition
```
robscott, line 74: I know I originally suggested this, but given how many functions use these timeouts, it will likely be difficult/impossible to keep these in sync as the codebase evolves. It's probably better just to leave the "Used by" lines out for all of these comments.

robscott (commented):
Thanks for the work on this @wstcliyu!

conformance/utils/config/timeout.go
wstcliyu: What would you suggest as the max value for this?

robscott: I think 30 seconds is fine as both default and what we say in the "Max value for conformant implementation" comment.

conformance/utils/config/timeout.go
wstcliyu: Thanks. Done.

robscott:
Thanks @wstcliyu! Will leave LGTM for someone else since this slightly modifies expectations for a conformant implementation. 

/cc @skriss @shaneutt @youngnick 
/approve

conformance/utils/config/timeout.go
wstcliyu: Done.

conformance/utils/config/timeout.go
wstcliyu: Done.

shaneutt (commented):
/lgtm

PR #1298, opened 2022-08-08
conformance/utils/suite: Move timeout setup before building roundtripper

sunjayBhatia:
/kind bug

Previously timeout config was passed to roundtripper before defaults
were set. When suite was run with default timeouts, for example HTTP requests were
set up with a context deadline of 0, meaning they failed immediately.

N/A

NONE


robscott:
Thanks @sunjayBhatia!

/lgtm
/approve

PR #1300, opened 2022-08-08
Adding Route Delegation Details for each HTTPRoute Capability

robscott:
**What type of PR is this?**
/kind gep

This is meant to augment @Jeff-Apple's  primary PR for this GEP (#1085). This PR will likely not merge independently and is still a WIP.

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1058.md
+++ b/site-src/geps/gep-1058.md
@@ -1,0 +1,89 @@
+# GEP-1058: Route Inclusion and Delegation
+
+- Issue: [#1058](https://github.com/kubernetes-sigs/gateway-api/issues/1058)
+- Status: Provisional
+
+## TLDR
+
+Add support for a 2-tier hierarchy of route configuration to allow partial
+delegation of route settings and, provide more flexibility in how route
+configurations are organized and managed.
+
+## HTTPRoute Delegation Capabilities
+
+### Matching
+| Capability | Importance for Delegation | Complexity | Implementation Details |
+|---|---|---|---|
+| Hostnames | ? | Moderate | A three way intersection between Gateway, Parent Route, and Child Route hostnames |
+| Prefix Path on Parent, Prefix or Exact on Child | High | Low | Prefix path match of parent is prepended to child path match, child can support exact or prefix path matching |
+| Exact Path on Parent | Low | Moderate | Would require only "/" matches on child route |
+| Exact Path on Child | High | Low | Prefix path match of parent is prepended to child path match, child can support exact or prefix path matching |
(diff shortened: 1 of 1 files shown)


robscott:
Deploy Preview: https://deploy-preview-1300--kubernetes-sigs-gateway-api.netlify.app/geps/gep-1058/

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,89 @@
...
+## HTTPRoute Delegation Capabilities
+
+### Matching
+| Capability | Importance for Delegation | Complexity | Implementation Details |
+|---|---|---|---|
+| Hostnames | ? | Moderate | A three way intersection between Gateway, Parent Route, and Child Route hostnames |
+| Prefix Path on Parent, Prefix or Exact on Child | High | Low | Prefix path match of parent is prepended to child path match, child can support exact or prefix path matching |
+| Exact Path on Parent | Low | Moderate | Would require only "/" matches on child route |
+| Exact Path on Child | High | Low | Prefix path match of parent is prepended to child path match, child can support exact or prefix path matching |
+| Regex Path on Parent or Child | Low | Impossible | Maybe some complex merging is possible here, but I can't think of a good solution |
```
mikemorris, line 21: What about delegating something like a `/accounts/[id]` pattern? This feels like it may be not a not-uncommon need - could something like excluding a parent PathPrefix make this possible? RegEx parent path feels like it could be harder and lower-value though, maybe split these cases apart?

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,89 @@
...
+
+```yaml
+matches:
+- path:
+  - prefix: /baz
+```
+
+* Does the child route match `/foo/bar` and `/bar/baz`?
+* Would it be preferable to require the child route to specify the full path?
+* How does this interact with potential matching conflicts for headers or query params?
```
mikemorris, line 65: I think matching both makes sense, and it's preferable to _not_ specify the full path, as it would make config more brittle if the upstream path were to ever change.

site-src/geps/gep-1058.md
```diff
@@ -0,0 +1,89 @@
...
+Child Route:
+
+```yaml
+matches:
+- method: GET
+  path:
+  - prefix: /baz
+```
+
+* Does this only match `GET /foo/baz`?
```
mikemorris, line 89: I think yes, it should only match `GET /foo/baz` (effectively a logical `AND` evaluation). I _think_ it would make sense to apply the same `AND` logic to headers and query params, for consistency with the existing intra-match spec https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.HTTPRouteMatch

PR #1305, opened 2022-08-09
Conformance: fix Listener ConditionReason

### LINKED ISSUE #1297: conformance: gateway tests expecting wrong ListenerConditionReason
**What happened**:

In the conformance tests for the gateway there are two different tests [`GatewaySecretInvalidReferenceGrant`, `GatewaySecretMissingReferenceGrant`] that expect a listener to have the ReasonCondition as `ListenerReasonInvalidCertificateRef`. Both tests create a Gateway with a listener that refers to a secret in another namespace and there is no `ReferenceGrant` granting that reference.

 For what concerns the`ListenerReasonInvalidCertificateRef`, the godoc says that:
```
// This reason is used with the "ResolvedRefs" condition when the
// Listener has a TLS configuration with at least one TLS CertificateRef
// that is invalid or cannot be resolved.
```

while for what concerns the`ListenerReasonRefNotPermitted `, the godoc says that:
```
// This reason is used with the "ResolvedRefs" condition when
// one of the Listener's Routes has a BackendRef to an object in
// another namespace, where the object in the other namespace does
// not have a ReferenceGrant explicitly allowing the reference.
```

Since the listener cannot reference the secret as there is no `ReferenceGrant` that allows that reference, the right reason to use for the `ListenerCondition` is `ListenerReasonRefNotPermitted` instead of `ListenerReasonInvalidCertificateRef`.

mlavacca:
**What type of PR is this?**

/kind bug

In the "GatewaySecretInvalidReferenceGrant" and "GatewaySecretMissingReferenceGrant" conformance tests for gateways, the expected condition has been changed from `InvalidCertificateRef` to `RefNotPermitted`.

Fixes #1297 

```release-note
To be conformant with the API, if there is no ReferenceGrant that grants a listener to reference a secret in another namespace, the ListenerConditionReason for the condition ResolvedRefs must be set to RefNotPermitted instead of InvalidCertificateRef.
```

### DIFF
--- a/conformance/tests/gateway-secret-invalid-reference-grant.go
+++ b/conformance/tests/gateway-secret-invalid-reference-grant.go
@@ -39,7 +39,7 @@ var GatewaySecretInvalidReferenceGrant = suite.ConformanceTest{
 	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
 		gwNN := types.NamespacedName{Name: "gateway-secret-invalid-reference-grant", Namespace: "gateway-conformance-infra"}
 
-		t.Run("Gateway listener should have a false ResolvedRefs condition with reason InvalidCertificateRef", func(t *testing.T) {
+		t.Run("Gateway listener should have a false ResolvedRefs condition with reason RefNotPermitted", func(t *testing.T) {
 			listeners := []v1alpha2.ListenerStatus{{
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
Looks good, thanks for identifying this problem and resolving it :vulcan_salute: 

/lgtm

PR #1306, opened 2022-08-10
Export DefaultTimeoutConfig.

briantkennedy:
Some exported methods like
MakeRequestAndExpectEventuallyConsistentResponse require the
TimeoutConfig resource.  Callers may want to use this with a sensible
default, so we should expose a way to get said value without requiring
the caller to understand which timeout values need to be configured.

**What type of PR is this?**

/kind cleanup

Make it easier to use MakeRequestAndExpectEventuallyConsistentResponse.

None

```release-note
NONE
```

### DIFF
--- a/conformance/utils/config/timeout.go
+++ b/conformance/utils/config/timeout.go
@@ -68,7 +68,8 @@ type TimeoutConfig struct {
 	RequestTimeout time.Duration
 }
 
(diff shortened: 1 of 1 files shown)


robscott:
Thanks!

/ok-to-test
/lgtm
/approve

PR #1310, opened 2022-08-11
Add utils to translate custom types

### LINKED ISSUE #826: Create utils to make it easier to work with custom types
Helper funcs that help with type conversion, especially from and to pointers, for our custom types. Examples of these types include `Hostname`, `Namespace`, and `Group`.

**Why this is needed**:
To simplify CRD validation, we've created a lot of custom types. Unfortunately this makes it difficult to work with our APIs with go. It would be very helpful if we had some utils that made this easier to work with. Otherwise each implementation will likely end up implementing their own set of these.

abhijit-dev82:
**What type of PR is this?**
/kind cleanup

Proposed change set are helper functions to translate to and from custom types.

Fixes #826

No

### DIFF
--- a/apis/v1alpha2/validation/util/utils.go
+++ b/apis/v1alpha2/validation/util/utils.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


apis/v1beta1/helper.go
```diff
@@ -0,0 +1,177 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
robscott, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

apis/v1alpha2/helper.go
```diff
@@ -0,0 +1,177 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
robscott, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

apis/v1beta1/helper.go
```diff
@@ -0,0 +1,177 @@
...
+	pn := PortNumber(port)
+	return &pn
+}
+
+func PortNumFromPtr(name *PortNumber) int32 {
+	portNum := int32(*name)
+	return portNum
+}
+
+func ListenerHostnameToPtr(host string) *Hostname {
```
robscott, line 39: Maybe remove `Listener` prefix here since Hostname can be used elsewhere.

apis/v1beta1/helper.go
```diff
@@ -0,0 +1,177 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+func SectionNameToPtr(sectionName string) *SectionName {
```
robscott, line 19: I think most go linting will usually fail without comments describing each exported function

robscott (commented):
Thanks @abhijit-dev82! This looks like a great start. Will discuss this at the community meeting on Monday to ensure that at least some implementations will actually make use of this. This seems like it could be very helpful, but don't want to introduce extra code to maintain until we're sure it will be used.

carlisia:
@abhijit-dev82 thank you for doing this, I will certainly use it so I can delete some code on my code base!

Everything lgtm except I'm not loving the name and the location of the file, especially when I consider existing helper functionality.

I opened an issue to explain: [Consider a minor restructure of directory/file for util things · Issue #1336 · kubernetes-sigs/gateway-api](https://github.com/kubernetes-sigs/gateway-api/issues/1336). I'd hate to block the merging of this PR for something that is sorta trivial but brakes the interface so it will require some thinking. At the same time, I would love adding such helpful code in a place that makes great sense, given the context around it.

robscott:
I really like @carlisia's idea in #1336. @abhijit-dev82 do you think you could incorporate that restructuring into this PR? 

/ok-to-test

carlisia:
There is one more input I'd like to offer. If we look at the existing conversion helpers (`PathMatchTypePtr` and `PortNumberPtr` here:

https://github.com/kubernetes-sigs/gateway-api/blob/854e2bfc52764bda4d81dd24851ab124bf564e8f/apis/v1alpha2/validation/util/utils.go#L24  

...we see that there is no `To` or `From`. That style is idiomatic of Go because it doesn't duplicate information that we can see in a different way. In this case, that `PathMatchTypePtr` takes a string is indicated in the signature, and it is idiomatic to name the function based on what it returns, in this case the pointer. Sooooo.... long paragraph to suggest changing the names to fit this format:

`s/SectionNameToPtr/SectionNamePtr`
`s/SectionNameFromPtr/SectionNameStr`

apis/v1alpha2/helper.go
```diff
@@ -0,0 +1,177 @@
...
+	pn := PortNumber(port)
+	return &pn
+}
+
+func PortNumFromPtr(name *PortNumber) int32 {
+	portNum := int32(*name)
+	return portNum
+}
+
+func ListenerHostnameToPtr(host string) *Hostname {
```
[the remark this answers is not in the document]
abhijit-dev82: corrected

apis/v1beta1/helper.go
abhijit-dev82: corrected.

apis/v1alpha2/helper.go
abhijit-dev82: Corrected.

carlisia:
This PR adds the directory structure mentioned above: https://github.com/kubernetes-sigs/gateway-api/pull/1337?w=1.

abhijit-dev82:
@carlisia & @robscott Thank you for all the comments. I was away due to health reasons, hence was not able to make the necessary changes. I see that the directory changes are already done, will add the new file in the util dir. Do we need to see we can auto generate these translation routines based on the custom types defined .

abhijit-dev82:
HI @carlisia & @robscott , Had to open a new PR as was facing merge issues with the older branch. Below is the link
https://github.com/kubernetes-sigs/gateway-api/pull/1352

PR #1316, opened 2022-08-11
All route ResolvedRefs reasons added to docs

### LINKED ISSUE #1313: Missing reasons in the `ResolvedRefs` `RouteConditionType` documentation
**What happened**:
The `ResolvedRefs` Route condition (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go#L227-L234) documentation does not include the reasons `InvalidKind` and `BackendNotFound`  among the possible ones that make it False.

What you expected to happen:

Since this godoc section's purpose is to enumerate the possible reasons that may lead to having one `ConditionStatus` instead of another, it is needed to ensure its exhaustivity. Hence, `InvalidKind` and `BackendNotFound` should be added to the possible reasons that can make the Accepted route condition False in the godoc section.

How to reproduce it (as minimally and precisely as possible):

Anything else we need to know?:

mlavacca:
**What type of PR is this?**

/kind documentation

`InvalidKind` and `BackendNotFound` are added to the possible failure reasons of the `ResolvedRefsCondition` in the route godoc.

Fixes #1313 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -231,6 +231,8 @@ const (
 	// Possible reasons for this condition to be false are:
 	//
 	// * "RefNotPermitted"
+	// * "InvalidKind"
+	// * "BackendNotFound"
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
@@ -333,8 +335,7 @@ type RouteStatus struct {
 //
 // 1. IPs are not allowed.
 // 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
-//
-//	label must appear by itself as the first label.
+//    label must appear by itself as the first label.
 //
 // Hostname can be "precise" which is a domain name without the terminating
 // dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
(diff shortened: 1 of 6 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -331,10 +333,9 @@ type RouteStatus struct {
 // Hostname is the fully qualified domain name of a network host. This matches
 // the RFC 1123 definition of a hostname with 2 notable exceptions:
 //
-// 1. IPs are not allowed.
-// 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
-//
-//	label must appear by itself as the first label.
+//  1. IPs are not allowed.
+//  2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
+//     label must appear by itself as the first label.
```
robscott, line 338: Why the extra spaces here? If I'm understanding this correctly, we're moving from one space after `//` to two, is that intentional?

robscott (commented):
Thanks @mlavacca!

apis/v1alpha2/shared_types.go
mlavacca: It was unintentional, `go fmt` somehow produced that. I noticed that the "verify" job failed. I have updated the `yaml` files, should be good now.

robscott:
Thanks @mlavacca!

/lgtm
/approve

PR #1321, opened 2022-08-12
Route ResolvedRefs reason godoc updated

### LINKED ISSUE #1312: Missing reason in the `Accepted` `RouteConditionType` documentation
**What happened**:

The `Accepted` Route condition (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go#L193-L201) documentation does not include the reason  `UnsupportedValue` among the possible ones that make it `False`.

Since this godoc section's purpose is to enumerate the possible reasons that may lead to having one ConditionStatus instead of another, it is needed to ensure its exhaustivity. Hence, `UnsupportedValue` should be added to the possible reasons that can make the `Accepted` route condition `False` in the godoc section.

mlavacca:
**What type of PR is this?**

/kind documentation

Documentation: `UnsupportedValue` has been added to the set of possible reasons when the `ResolvedRefs` status condition is failing. 

Fixes #1312 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -198,6 +198,7 @@ const (
 	//
 	// * "NotAllowedByListeners"
 	// * "NoMatchingListenerHostname"
+	// * "UnsupportedValue"
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
Thanks @mlavacca :vulcan_salute: 

/lgtm

PR #1326, opened 2022-08-15
add GEP-1323: Response Header Filter

### LINKED ISSUE #1323: Add support for manipulating response headers in the `HTTPRouteFilter` API

It'd be great if `HTTPRouteFilter` had a field to manipulate response headers being sent from  a backend (or a group of backends) similar to `RequestHeaderModifier`.

This can be done by adding a new field `ResponseHeaderModifier` of type `HTTPResponseHeaderFilter` which would have the same API as the existing `HTTPRequestHeaderFilter`. Alternatively, `HTTPRequestHeaderFilter` could be renamed to `HTTPHeaderFilter` and both `RequestHeaderModifier` and `ResponseHeaderModifier` would be of type `HTTPHeaderFilter`.

Similar specifications can be seen in service meshes like Istio's `VirtualService` (ref: https://istio.io/latest/docs/reference/config/networking/virtual-service/#Headers) and ingresses like Contour's `HTTProxy` (ref: https://projectcontour.io/docs/v1.22.0/config/api/#projectcontour.io/v1.Route)

**Why this is needed**:

Just like editing request headers can be useful, the same goes for response headers. For example, it allows teams to add/remove cookies for only a certain backend, which can help in identifying certain users that were redirected to that backend previously.

```yaml
kind: HTTPRoute
metadata:
  name: http-response-header
spec:
  hostnames:
    - response.header.example
  rules:
    - filters:
      - type: ResponseHeaderModifier
        responseHeaderModifier:
          set:
            name: Set-Cookie
            value: user=insider
      backendRefs:
      - name: example-svc
        weight: 1
        port: 80
```

Another potential use case could be when you have a frontend that needs to know whether it's talking to a stable or a beta version of the backend server, in order to render different UI or adapt its response parsing accordingly.

```yaml
kind: HTTPRoute
metadata:
  name: http-response-header
spec:
  hostnames:
    - response.header.example
  rules:
    - backendRefs:
      - name: example-svc-beta
        weight: 50
        port: 80
        filters:
           - type: ResponseHeaderModifier
             responseHeaderModifier:
              set:
                name: build
                value: beta
      - name: example-svc-stable
        weight: 50
        port: 80
```

If this requires a GEP, I'd be happy to start working on one.

aryan9600:
Signed-off-by: Sanskar Jaiswal <jaiswalsanskar078@gmail.com>

**What type of PR is this?**
/kind gep

Adds a GEP proposing a way of modifying response headers via a new field `responseHeaderModifier`.

Fixes #1323 

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1323.md
+++ b/site-src/geps/gep-1323.md
@@ -1,0 +1,208 @@
+# GEP 1323: Response Header Filter
+* Issue: [#1323](https://github.com/kubernetes-sigs/gateway-api/issues/1323)
+* Status: Implementable
+
+## TLDR
+Similar to how we have `RequestHeaderModifier` in `HTTPRouteFilter`, which lets users modify request headers before the request is forwarded to a backend (or a group of backends), it’d be helpful to have a `ResponseHeaderModifier` field which would let users modify response headers before they are returned to the client.
+
+## Goals
+* Provide a way to modify HTTP response headers in a `HTTPRoute`.
+* Reuse existing types as much as possible to reduce boilerplate code.
+
+## Non Goals
+* Provide a way to modify other parts of a HTTP response like status code.
+* Add fields specifically for standard headers such as `Cookie`.
+
+## Introduction
+Currently, the `HTTPRouteFilter` API provides a way for request headers to be modified through the `RequestHeaderModifier` field of type `HTTPRequestHeaderModifier`. But, a similar API to modify response headers does not exist. This proposal intends to introduce a new field in `HTTPRouteFilter` named `ResponseHeaderModifier`.
+
+## API
+We could introduce a new API named `HTTPResponseHeaderModifier` which would look exactly like the existing `HTTPRequestHeaderModifier` API. But since HTTP headers have the same semantics for both requests and responses, it makes more sense to rename `HTTPRequestHeaderModifier` to `HTTPHeaderModifier` and use this for both `RequestHeaderModifier` and `ResponseHeaderModifier`.
+
+```golang
+// HTTPHeaderModifier defines a filter that modifies the headers of a HTTP
+// request or response.
+type HTTPHeaderModifier struct {
+    // Set overwrites the request with the given header (name, value)
+    // before the action.
+    // +optional
+    // +listType=map
+    // +listMapKey=name
+    // +kubebuilder:validation:MaxItems=16
+    Set []HTTPHeader `json:"set,omitempty"`
+
+    // Add adds the given header(s) (name, value) to the request
+    // before the action. It appends to any existing values associated
+    // with the header name.
+
+    // +optional
+    // +listType=map
+    // +listMapKey=name
+    // +kubebuilder:validation:MaxItems=16
+    Add []HTTPHeader `json:"add,omitempty"`
+
+    // Remove the given header(s) from the HTTP request before the action. The
+    // value of Remove is a list of HTTP header names. Note that the header
+    // names are case-insensitive (see
+    // https://datatracker.ietf.org/doc/html/rfc2616#section-4.2).
+    // +optional
+    // +kubebuilder:validation:MaxItems=16
+    Remove []string `json:"remove,omitempty"`
+}
+```
+
+Given the fact that this functionality is offered by only a few projects that currently implement Gateway API when using their own traffic routing CRDs, it’s better to support `ResponseHeaderModifier` as an _Extended_ feature, unlike `RequestHeaderModifier` which is a _Core_ feature. This will also not increase the difficulty of implementing Gateway API for any future ingress or service mesh.
+
+This feature can be further extended via [Policy Attachment](../references/policy-attachment.md). The mechanism and use cases of this may be explored in a future GEP.
+
+## Usage
+Adding support for this unlocks a lot of real world use cases. Let’s review a couple of them:
+
+* A team has a frontend web app, along with two different versions of their backends exposed as Kubernetes services. If, the frontend needs to know which backend it’s talking to, this can be easily achieved without any modifications to the application code.
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: http-response-header
+spec:
+  hostnames:
+    - response.header.example
+  rules:
+    - backendRefs:
+      - name: example-svc-beta
+        weight: 50
+        port: 80
+        # set a custom header for all responses being sent from the beta build of the backend server.
+        filters:
+           - type: ResponseHeaderModifier
+             responseHeaderModifier:
+              add:
+                name: build
+                value: beta
+      - name: example-svc-stable
+        weight: 50
+        port: 80
+```
+
+* Cookies can be automatically injected into the response of services. This can enable services to identify users that were redirected to a certain backend.
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: http-response-header
+spec:
+  hostnames:
+    - response.header.example
+  rules:
+    # match against any requests that has the cookie set due to the below rule
+    - matches:
+      - headers:
+        type: Exact
+        name: Cookie
+        value: user=insider
+      backendRefs:
+        - name: foo-svc
+        port: 8080
+
+    - filters:
+      - type: ResponseHeaderModifier
+        # set cookies for all requests being forwarded to this service
+        responseHeaderModifier:
+          set:
+            name: Set-Cookie
+            value: user=insider
+      backendRefs:
+      - name: example-svc
+        weight: 1
+        port: 80
+```
+
+> Note: Some projects like Envoy support interpolating a few predefined [variables into header values](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#custom-request-response-headers). Similar functionality might be supported by other implementations but its unlikely to be portable and thus has been excluded from the API for the time being.
+
+## Prior Art
+A few projects that implement Gateway API already have support for similar functionality (in their custom CRDs), like:
+* Istio’s `VirtualService`:
+
+```yaml
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: reviews-route
+spec:
+  hosts:
+  - reviews.prod.svc.cluster.local
+  http:
+  - headers:
+      request:
+        set:
+          test: "true"
+    route:
+    - destination:
+        host: reviews.prod.svc.cluster.local
+        subset: v2
+      weight: 25
+    - destination:
+        host: reviews.prod.svc.cluster.local
+        subset: v1
+      headers:
+        response:
+          remove:
+          - foo
+      weight: 75
+```
+
+* Contour’s `HTTPProxy`:
+
+```yaml
+apiVersion: projectcontour.io/v1
+kind: HTTPProxy
+metadata:
+  name: basic
+spec:
+  virtualhost:
+    fqdn: foo-basic.bar.com
+  routes:
+    - conditions:
+      - prefix: /
+      services:
+        - name: s1
+          port: 80
+      responseHeadersPolicy:
+        set:
+          name: test
+          value: true
+```
+
+* Ingress NGINX:
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: nginx-headers
+  annotations:
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      add_header ingress nginx;
+spec:
+  ingressClassName: nginx
+  rules:
+  - host: custom.configuration.com
+    http:
+      paths:
+      - path: /
+        pathType: Prefix
+        backend:
+          service:
+            name: http-svc
+            port:
+              number: 80
+```
+
+## External Links
+* [Contour `HeaderValue`](https://projectcontour.io/docs/v1.22.0/config/api/#projectcontour.io/v1.HeaderValue)
+* [Istio `Headers`](https://istio.io/latest/docs/reference/config/networking/virtual-service/#Headers)
+* [Ingress NGINX Configuration Snippets](https://kubernetes.github.io/ingress-nginx/examples/customization/configuration-snippets/)
+* [NGINX `add_header` directive](https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header)
+


site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+## Goals
+* Provide a way to modify HTTP response headers.
+* Reuse existing types as much as possible to reduce boilerplate code.
+
+## Non Goals
+* Provide a way to modify other parts of a HTTP response like status code.
+* Add fields specifically for standard headers such as `Cookie`.
+
+## Introduction
+Currently, the `HTTPRouteFilter` API provides a way for request headers to be modified through the `RequestHeaderModifier` field of type `HTTPRequestHeaderModifier`. But, a similar API to modify response headers does not exist. This proposal intends to introduce a new field in `HTTPRouteFilter` named `ResponseHedaerModifier`.
```
shaneutt, line 17:
```suggestion
Currently, the `HTTPRouteFilter` API provides a way for request headers to be modified through the `RequestHeaderModifier` field of type `HTTPRequestHeaderModifier`. But, a similar API to modify response headers does not exist. This proposal intends to introduce a new field in `HTTPRouteFilter` named `ResponseHeaderModifier`.
```

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+
+## API
+We could introduce a new API named `HTTPResponseHeaderModifier` which would look exactly like the existing `HTTPRequestHeaderModifier` API. But since HTTP headers have the same semantics for both requests and responses, it makes more sense to rename `HTTPRequestHeaderModifier` to `HTTPHeaderModifier` and use this for both `RequestHeaderModifier` and `ResponseHeaderModifier`.
```
shaneutt, line 20: :+1:

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+    // value of Remove is a list of HTTP header names. Note that the header
+    // names are case-insensitive (see
+    // https://datatracker.ietf.org/doc/html/rfc2616#section-4.2).
+    // +optional
+    // +kubebuilder:validation:MaxItems=16
+    Remove []string `json:"remove,omitempty"`
+}
+```
+
+Given the fact that this functionality is offered by only a few projects that currently implement Gateway API when using their own traffic routing CRDs, it’s better to support `ResponseHeaderModifier` as an _Extended_ feature, unlike `RequestHeaderModifier` which is a _Core_ feature. This will also not increase the difficulty of implementing Gateway API for any future ingress or service mesh.
```
shaneutt, line 54: I agree :+1:

shaneutt (commented):
I think it's a good proposal. I particularly like the usage examples, thanks for taking the time to write all this out.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+          set:
+            name: Set-Cookie
+            value: user=insider
+      backendRefs:
+      - name: example-svc
+        weight: 1
+        port: 80
+```
+
+## Prior Art
```
robscott, line 120: These examples are great, but they are both Envoy based. Do you mind adding examples of how this could be configured with something like nginx or haproxy?

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+## Introduction
+Currently, the `HTTPRouteFilter` API provides a way for request headers to be modified through the `RequestHeaderModifier` field of type `HTTPRequestHeaderModifier`. But, a similar API to modify response headers does not exist. This proposal intends to introduce a new field in `HTTPRouteFilter` named `ResponseHedaerModifier`.
+
+## API
+We could introduce a new API named `HTTPResponseHeaderModifier` which would look exactly like the existing `HTTPRequestHeaderModifier` API. But since HTTP headers have the same semantics for both requests and responses, it makes more sense to rename `HTTPRequestHeaderModifier` to `HTTPHeaderModifier` and use this for both `RequestHeaderModifier` and `ResponseHeaderModifier`.
+
+```golang
+// HTTPHeaderFilter defines a filter that modifies the headers of a HTTP
+// request or response.
+type HTTPHeaderFilter struct {
```
robscott, line 25: Nit: I think you meant to say `HTTPHeaderModifier` here and elsewhere in the code snippet to match the paragraph above.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+          port: 80
+      responseHeadersPolicy:
+    set:
+      name: test
+      value: true
+```
+
+## External Links
+* https://projectcontour.io/docs/v1.22.0/config/api/#projectcontour.io/v1.HeaderValue
+* https://istio.io/latest/docs/reference/config/networking/virtual-service/#Headers
```
robscott, line 176: Can you make these markdown links?

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
+# GEP 1323: Response Header Filter
+* Issue: [#1323](https://github.com/kubernetes-sigs/gateway-api/issues/1323)
+* Status: Provisional
```
robscott, line 3:
```suggestion
* Status: Implementable
```

Unless this winds up being unexpectedly controversial, I think we can move directly to Implementable.

robscott (commented):
Thanks @aryan9600!

site-src/geps/gep-1323.md
shaneutt: The only thing that _might_ be controversial that I can see is that I think it's technically up for debate whether the functionality be core or extended? _maybe_? but ultimately I think you're right.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
...
+}
+```
+
+Given the fact that this functionality is offered by only a few projects that currently implement Gateway API when using their own traffic routing CRDs, it’s better to support `ResponseHeaderModifier` as an _Extended_ feature, unlike `RequestHeaderModifier` which is a _Core_ feature. This will also not increase the difficulty of implementing Gateway API for any future ingress or service mesh.
+
+## Usage
+Adding support for this unlocks a lot of real world use cases. Let’s review a couple of them:
+
+* For example, if a team has a frontend web app that along with two different versions of their backends exposed as Kubernetes services, and the frontend needs to know which backend it’s talking to, this can be easily achieved without any modifications to the application code.
+
```
candita, line 60: Could you describe how this might work if the backend application needs to deliver a custom value in the response?  E.g. the backend destination determines a cookie value?

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,177 @@
+# GEP 1323: Response Header Filter
+* Issue: [#1323](https://github.com/kubernetes-sigs/gateway-api/issues/1323)
+* Status: Provisional
+
+## TLDR
+Similar to how we have `RequestHeaderModifier` in `HTTPRouteFilter`, which lets users modify request headers before the request is forwarded to a backend (or a group of backends), it’d be helpful to have a `ResponseHeadederModifier` field which would let users modify response headers before they are forwarded to its destination.
+
+## Goals
+* Provide a way to modify HTTP response headers.
+* Reuse existing types as much as possible to reduce boilerplate code.
```
candita, line 10: What if the admin wanted to add the same response header to all HTTPRoutes (e.g. the same HSTS response for all of them) - can you find a way for the admin to override response headers at a higher level?

aryan9600: I don't see how that's possible with the current structure of the API. Adding a response header to all HTTPRoutes is possible if the Gateway object had a way of modifying the responses that are handled by HTTPRoutes attached to it. From what I understand going through https://gateway-api.sigs.k8s.io/concepts/api-overview/#gateway, a Gateway is not responsible for actually modifying or forwarding requests/responses.
If we want to introduce this behavior, I'd guess that it's out of scope here and needs a RFC of it's own. cc: @robscott @shaneutt

site-src/geps/gep-1323.md
aryan9600: I'm sorry, I don't understand fully. If the backend application needs to deliver a custom cookie that it determined using some business logic, it'd need to set it in while returning the response itself. If I'm not answering your question clearly here, could you kindly elaborate a bit more? thanks

site-src/geps/gep-1323.md
youngnick: That use case (admin wanting to force the addition of something) is one of the things we had in mind with Policy Attachment - I could see a Policy Resource that replicated the same structure used here in an `override` block, that could be attached to a Gateway resource. That would imply that these settings should Override any settings on the HTTPRoutes.

site-src/geps/gep-1323.md
youngnick: It sounds like maybe you're describing something like the backend app setting a cookie, and then the Gateway implementation needing to do something with that value, @candita?

If that is correct, I know some proxies support setting a header value using the value from another, but I don't feel like that's common enough to be included in the API.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,204 @@
...
+    // value of Remove is a list of HTTP header names. Note that the header
+    // names are case-insensitive (see
+    // https://datatracker.ietf.org/doc/html/rfc2616#section-4.2).
+    // +optional
+    // +kubebuilder:validation:MaxItems=16
+    Remove []string `json:"remove,omitempty"`
+}
+```
+
+Given the fact that this functionality is offered by only a few projects that currently implement Gateway API when using their own traffic routing CRDs, it’s better to support `ResponseHeaderModifier` as an _Extended_ feature, unlike `RequestHeaderModifier` which is a _Core_ feature. This will also not increase the difficulty of implementing Gateway API for any future ingress or service mesh.
```
youngnick, line 54: I like this, we can use the same `HTTPHeaderModifier` struct to represent the actual config, but using the field that the struct appears in inside the Route can have different support, and that support can be documented on those fields.

Not sure if we need that in this GEP or not though.

youngnick (commented):
This is looking pretty great to me, just a few small things.

site-src/geps/gep-1323.md
aryan9600: Do you mean we could add a line in the `HTTPRouteFilter` API above the `ResponseHeaderModifier` field mentioning the support similar to this: https://github.com/kubernetes-sigs/gateway-api/blob/bb549d06152f11759358298970a807e8ffb08089/apis/v1beta1/httproute_types.go#L571?
If yes, I didn't mention it in the proposal, because I thought it's implied that the support will be documented in the API docs. But if you think, we need to be explicit about it in the proposal, I'll add something accordingly.

site-src/geps/gep-1323.md
aryan9600: Yes, I agree, the Policy Attachment API is a good place for something like this. But I'm not sure if it falls under the scope of this proposal. If it does, then please let me know and I'll add a section about it. cc: @shaneutt @robscott

shaneutt: @aryan9600 if we're going to go straight to `implementable` with this, then I think adding a section for this and pointing to policy attachment sounds correct.

aryan9600: @youngnick added a section on Policy Attachment, please take a look, thanks.

site-src/geps/gep-1323.md
candita: @aryan9600 for example, this HTTPRoute has two hostnames, and the header response depends on the value of the hostname that delivered this traffic.  If response1.header.example was used, return header value 1, but if response2.header.example was used, return header value 2.  The backend gets to decide what value is sent back in the response header.  Can this be set up to hold a variable name in the `value` of the `add` stanza?

```suggestion
filters:
           - type: ResponseHeaderModifier
             responseHeaderModifier:
              add:
                name:number
                value: $
```
I guess I'm aware of more response headers that are set after arriving at the backend, and based on what the application determines.  This proposal is based more on a hard-coded response header value, which is less challenging, but perhaps also less useful.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
...
+    TargetRef gatewayv1a2.PolicyTargetReference `json:"targetRef"`
+
+    // Override defines policy configuration that should override policy
+    // configuration attached below the targeted resource in the hierarchy.
+    // +optional
+    Override *HTTPHeaderModifier `json:"override,omitempty"`
+
+    // Default defines default policy configuration for the targeted resource.
+    // +optional
+    Default *HTTPHeaderModifier `json:"default,omitempty"`
```
candita, line 227: I don't think `Default` is needed here.  If so, how is the `Default` different than the `Override`?  What happens if both are specified?

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
...
+          service:
+            name: http-svc
+            port:
+              number: 80
+```
+
+## Policy Attachment
+
+This feature can be further extended via [Policy Attachment](../references/policy-attachment.md). Implementations can introduce a new Policy, `ResponseHeaderPolicy`, which can be used to ensure that certain headers are present/absent in a response, regardless of the `responseHeaderModifier` field in the `HTTPRoute` associated with that response.
+
```
candita, line 203: Maybe you don't need to hash all of this out in this GEP, and just a reference to future work could be included.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
...
+}
+
+type ResponseHeaderPolicyStatus struct {
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+`.spec.default` provides a way to modify headers of responses being returned from the resource specified in `.spec.targetRef`. This resource can be a part of the Gateway API like `Gateway` or a core Kubernetes resource like `Namespace`. Using a `Namespace` as the `.spec.targetRef` would result in any response being returned from any resource in that namespace having its headers modified according to the policy.
+`.spec.override` can be used to modify the headers of responses being returned from a resource that is already covered by `.spec.default` of a `ResponseHeaderPolicy`. Let's walk through an example:
+
```
candita, line 237: This is a good example. If the community decides to keep these details within this GEP, then I'll offer a suggestion for a different way to implement a global policy with the possibility of override.  The policy is set at the Gateway to apply to all *Routes globally as you described, but then each *Route could override the global default if needed, using the `ResponseHeaderModifier` mechanism you described earlier.  The policy would then need to include some way to describe whether it is optional or not, i.e. whether it can be overridden or not.  To me, this approach is clearer, but I still think it could be discussed in a different GEP.

site-src/geps/gep-1323.md
pleshakov: Having a policy that overrides a filter creates an implicit policy for a filter in an HTTPRoute :) . It also means we now have two different ways to configure the same thing - through a policy and through a filter. I wonder if it is a good thing.

Perhaps another way is to have a policy that can override all filters rather than a specific one? So it is more general-purpose.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
...
+      backendRefs:
+        - name: foo-svc
+        port: 8080
+
+    - filters:
+      - type: ResponseHeaderModifier
+        # set cookies for all requests being forwarded to this service
+        responseHeaderModifier:
+          set:
+            name: Set-Cookie
```
pleshakov, line 113: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie, "To send multiple cookies, multiple Set-Cookie headers should be sent in the same response."   Will it be possible to set multiple headers with the same name?

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
...
+
+    // Add adds the given header(s) (name, value) to the request
+    // before the action. It appends to any existing values associated
+    // with the header name.
+
+    // +optional
+    // +listType=map
+    // +listMapKey=name
+    // +kubebuilder:validation:MaxItems=16
+    Add []HTTPHeader `json:"add,omitempty"`
```
pleshakov, line 43: does it make sense to always add headers? For example, what if the backend returns 500 response or the backend doesn't exist so that the proxy generates 500.   

For example, in NGINX, it is possible to add headers for a subset of 2xx and 3xx codes only.

site-src/geps/gep-1323.md
```diff
@@ -0,0 +1,274 @@
+# GEP 1323: Response Header Filter
+* Issue: [#1323](https://github.com/kubernetes-sigs/gateway-api/issues/1323)
+* Status: Implementable
+
+## TLDR
+Similar to how we have `RequestHeaderModifier` in `HTTPRouteFilter`, which lets users modify request headers before the request is forwarded to a backend (or a group of backends), it’d be helpful to have a `ResponseHeaderModifier` field which would let users modify response headers before they are forwarded to its destination.
```
tokers, line 6:
```suggestion
Similar to how we have `RequestHeaderModifier` in `HTTPRouteFilter`, which lets users modify request headers before the request is forwarded to a backend (or a group of backends), it’d be helpful to have a `ResponseHeaderModifier` field which would let users modify response headers before they are returned to the client.
```

site-src/geps/gep-1323.md
tokers: @pleshakov I think what you described is more like the "conditions" for running the filter. I think it can be implemented as a common option for all filters, i.e. check out the condition before running the filter.

site-src/geps/gep-1323.md
aryan9600: @pleshakov yes we have two different ways to configure the same thing, but they are meant to be used in a different context. The example in the `Policy Attachment` section describes a real world use case that highlights this.
I'm definitely okay with having a multi-purpose policy instead of this but please keep in mind that this is just a suggestion. As far as I understand, implementations are free to design their policies as they chose and like. In https://gateway-api.sigs.k8s.io/references/policy-attachment/, I can see policies like `TimeoutPolicy` and `RetryPolicy` being used throughout, although they are technically each a filter.

site-src/geps/gep-1323.md
aryan9600: okay, so from what I understand, you're suggesting parsing with logical conditioning? this is a very cool feature, but i doubt if it's generic enough to be included into the API. could you please point me to a couple of service meshes/ingresses that allow for something similar today?

site-src/geps/gep-1323.md
aryan9600: I think the example below describes the usage of `Default` and `Override` sufficiently. About both being specified, I don't think that makes any sense, so maybe I should add something like _only_ one of `Default` and `Override` should be present?

site-src/geps/gep-1323.md
aryan9600: I don't understand what you mean by "a reference to future work"? Do you mean state something like "This API can be extended with Policy Attachment, which will be discussed in a future GEP"?

site-src/geps/gep-1323.md
aryan9600: You can use `add` instead of `set`: https://github.com/kubernetes-sigs/gateway-api/blob/30f2983e9f01368adff82772c4d9178fcdb26344/apis/v1beta1/httproute_types.go#L730

site-src/geps/gep-1323.md
aryan9600: @pleshakov you seem to be suggesting similar to https://github.com/kubernetes-sigs/gateway-api/pull/1326#discussion_r954544963
i agree with @tokers that this could be made more generic. furthermore, i believe it exists outside the scope of this GEP, as deciding how to conditional logic on filters is a GEP of it's own.

site-src/geps/gep-1323.md
aryan9600: If we allow a Route to override the `override` set by a Gateway using its `ResponseHeaderModifier`, how would admins set headers alongside the headers specified  in the Route's `ResponseHeaderModifier`?
About, moving the Policy Attachment section to its own GEP, I'm neither for or against it, so some consensus from the maintainers on this would be helpful.
wdyt? @shaneutt @robscott @youngnick

site-src/geps/gep-1323.md
mikemorris: Would setting different filters on two separate [HTTPBackendRefs](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPBackendRef) be sufficient to enable some of the desired functionality here? The configuration would be static still and a bit more verbose than specifying a filter with a dynamic value at the [HTTPRouteRule](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule) level, but maybe enough for some needs, such as determining if a response was served by a canary backend?

For the hostname matching specifically, I suppose two separate HTTPRouteRules (identical except for the `hostnames` field) could be configured as one way to implement that instead of relying on a response from the backend - that feels like it could be a use case that could be simplified with config templating, which may be best achieved by third-party tools at a layer above Gateway API.

Finally, for header manipulation (add or modify a header based on an existing header value in the response, which feels like the most direct way to implement this without touching the backend app itself), it feels like we might have some of the scaffolding to build from by combining `type: RegularExpression` from [HTTPHeaderMatch](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.HTTPHeaderMatch) and the `add/set` fields in [HTTPRequestHeaderFilter](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.HTTPRequestHeaderFilter) if we were to extend the RegEx implementation from just matching to allow capture groups for setting a value, but then you're already in Custom conformance due to the variance in RegEx implementations, so I'm not sure how much value there is to spec that, although [HTTPPathModifier](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier) could possibly benefit from a similar implementation.

site-src/geps/gep-1323.md
mikemorris: I like this proposed approach and _think_ it's sufficiently clear. [Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) already has the structure for separating defaults and overrides - if a header was added or modified by an `Default` config a filter on the Route object should be able to further modify or remove it, but if added or modified by an `Override` config, a filter on the Route should _not_ be able to modify or remove it.

The main reason this may warrant consideration as a separate GEP is that it contradicts the current warning in https://gateway-api.sigs.k8s.io/references/policy-attachment/#interaction-with-custom-route-filters, although that does explicitly state:
> Note that this guidance may change in the future as we gain a better understanding of how extension mechanisms of the Gateway API can interoperate.

site-src/geps/gep-1323.md
mikemorris: https://gateway-api.sigs.k8s.io/references/policy-attachment/#policy-boilerplate does describe allowing both `Default` and `Override` to be set, although I think that often makes more sense when they modify different values, such as the `AcmeServicePolicy` example in https://gateway-api.sigs.k8s.io/references/policy-attachment/#hierarchy

site-src/geps/gep-1323.md
pleshakov: I think the validation will prevent it though - it will reject any attempt to set or add two headers with the same name
```
    // +listType=map
    // +listMapKey=name
  ```
at the same time, it is probably will be very uncommon requirement

aryan9600: i'm not 100% sure, but i think that'll happen when you have the same header name (the key) present in both `set` _and_ `add`. This conclusion of mine stems from https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules
> map: X + Y performs a merge where the array positions of all keys in X are preserved but the values are overwritten by values in Y when the key sets of X and Y intersect. Elements in Y with non-intersecting keys are appended, retaining their partial order.

site-src/geps/gep-1323.md
aryan9600: @mikemorris having gone through the text, I understand that we disallow overlaps between Policy and _custom_ Route filters. Since this proposed filter is not custom I feel like the current proposal is more or less okay.
I'll update it with your suggestion about Routes being able to modify `Default` config. @candita I now realize that I misread your initial comment 🤦 so yes, a Route should be able to modify the gloabal default

site-src/geps/gep-1323.md
robscott: After discussing this at today's community meeting, it seems like we're all OK to leave policy attachment out of the scope of this GEP. As long as the GEP mentions that this may be further expanded by policy attachment in the future, I don't think we need to get into the details of what that would look like here.

robscott:
Thanks for all the work on this @aryan9600! We talked about this at today's community meeting. I left some follow up comments above, but I think the major themes are:

1. This GEP feels very close to ready to merge.
2. When it comes to variables in header names, we should mention that some underlying implementations like Envoy support interpolating variables into header values, but that is unlikely to be portable across implementations.
3. When it comes to policy attachment, we should mention it as a possible future extension point, but we don't need to cover it in detail in this GEP.

/ok-to-test

robscott:
Thanks @aryan9600! This LGTM but will defer to someone else for the final approval.

/approve

site-src/geps/gep-1323.md
youngnick: Yes, I think that when this is implemented, any new fields should have their support explicitly documented.

youngnick:
Yup, I think this is good to go.

/lgtm

Issue #1282, opened 2022-07-26
GEP: Describe Backend Properties


youngnick:
**What would you like to be added**:

As people have started implementing Gateway API, some functionality has come up that:
- is tightly bound to the backend (which is usually a Service)
- Can't easily be added to the Service resource

But that we need a way to represent in the API.

The first examples that came up are:
- TLS details for (re)encryption between the Gateway and the backend. That is, the backend must have a way to store a serving TLS keypair that represents its service identity. (This one in particular has a lot of crossover with Mesh/GAMMA) use cases.
- HTTP Websocket protocol needs to be able to be turned on somewhere in the API, but requires the backend implementing it to support it.

These are certainly not the only things that we'll need to add to this sort of construct though, so it must be extensible and designed with the future in mind.

This GEP covers the work to design and implement solutions to these problems.

As @shaneutt suggested, the first step will be implementing a provisional GEP that sets out the terms of what we're doing - the "What" and "Why" of the solution, and then once we're on the same page, we'll talk more about the "How".

Part of this GEP process should also be clarifying if we're using Policy Attachment for this, and why we've chosen to use it or not.

**Why this is needed**:

Layer 7 implementations already allow both of the first functionality, solving these in a more general way will give us a path towards adding more things.

youngnick:
Okay, I've started something rough in https://docs.google.com/document/d/1M8EPrZKuhYsQjHnVrdsBxU8mVPKUWUYLXW9qE_Et9-Y/edit?usp=sharing .

Remember that we're working first on a "what" and a "why", not on a "how" yet. Let's do a Provisional GEP agreeing on those things, then we can move to the "how".

mikemorris:
> That is, the backend must have a way to store a serving TLS keypair that represents its service identity. (This one in particular has a lot of crossover with Mesh/GAMMA) use cases.

One interesting twist here is that for mesh cases it may be preferable for the TLS identity to be assigned by a mesh controller rather than specified manually (but maybe possible to override?) which could suggest a status field for exposing this to consumers.

evankanderson:
Would it be reasonable for the mesh to fill in the TLS details in the ServiceMetadata? Perhaps a spec/status split where the mesh could fill in the status if the cert was not specified in spec.

howardjohn:
I mentioned this elsewhere but IMO application TLS and mesh encryption (which is not always tls...) should be kept entirely separated

youngnick:
[...]

PR #1333, opened 2022-08-16
Add Provisional GEP-1282 document

youngnick:
Signed-off-by: Nick Young <nick@isovalent.com>

/kind gep

Adds the provisional GEP-1282 (#1282) document.

Updates #1282 

```release-note
NONE
```


site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,79 @@
+# GEP-696: GEP template
```
shaneutt, line 1: We would appear to want _some kind of title for this_?

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,79 @@
...
+* Issue: [#1282](https://github.com/kubernetes-sigs/gateway-api/issues/1282)
+* Status: Provisional
+
+(See definitions in [Kubernetes KEP][kep-status].
+
+[kep-status]: https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml#L9
+
+## TLDR
+
+End-user feedback of Gateway API has presented a trend: some users are interested in defining functionality that describes capabilities of the service more than the route you take to get to the service. In order to support these more advanced route functions, implementations need programmatic access to deeper insights about a Service’s (that is: a Kubernetes Service that would be used in an *Route backendRef) capabilities, such as TLS or websocket details. The purpose of this GEP is to add APIs for capturing those Service capabilities to enable more complex routing composition.
```
shaneutt, line 12: > some users are interested in defining functionality that describes capabilities of the service more than the route you take to get to the service

To me this statement seems to demand more information about "why": it's not clear to me after reading the entire GEP so far _why_ users are interested in this functionality.

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,79 @@
...
+* The method must be structured (that is, stored in a typed schema), and extensible (the design must include ways in which we can add more properties and capabilities in the future without redesigning it).
+* The method should have a roadmap where we can take what we learn from this design and bring it to the core Kubernetes resources later.
+* Clear separation of concerns with PolicyAttachment for properties which require multiple levels of defaults/overrides (timeouts, retries, etc).
+
+
+## Non-Goals
+
+* We don’t want to require changes to v1.Service to start with, as the turnaround time for those changes can be years.
+* Our solution should not use unstructured data like labels and annotations - these are very sticky and hard to get rid of once you start using them.
+* v1.Service’s appProtocol field is not fit for purpose, because it is defined as accepting values either from the [IANA Service Name registry](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtmly), or domain-prefixed values and we need more flexibility than that.
```
shaneutt, line 26: good non-goals.

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,79 @@
...
+* To be populated as we discuss more.
+
+
+## Introduction
+
+As implementations have started to build out HTTPRoute support, requests for some common capabilities have started coming up, particularly TLS re-encryption (encryption between Gateway and backend), Websockets tracking, and HTTP/2 support.
+
+Evan Anderson opened a [discussion that generated a lot of interest](https://github.com/kubernetes-sigs/gateway-api/discussions/1244), and during that discussion, we brought up a few more points:
+* Whatever we do to solve this in the HTTPRoute case has applicability to describing more general service properties, like identity, which could be useful for other Route use cases.
```
shaneutt, line 35:
```suggestion
* Whatever we do to solve this in the HTTPRoute case may be applicable to more general service properties, like identity, which could be useful for other Route use cases.
```

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,79 @@
...
+The initial list includes, but is not limited to:
+* TLS information for connection from the Gateway to the backend service. Note that this doesn’t include any information used for service mesh encryption, just what a Gateway’s proxy would need to be able to connect to the backend service.
+* Websocket protocol information for the backend service.
+* Protocol disambiguation for “upgradeable” protocols like HTTP/2 and HTTP/1.1 which operate on the same port. (Websockets may be another case of this.)
+
+### Why build something?
+
+The existing resources we have (Service and HTTPRoute) don’t have space for this information currently, and we have feedback from users via various implementations that it’s necessary - you could argue that some of this is table stakes for a layer 7 ingress controller.
+
+In previous discussions, we discussed a variety of ways to achieve this goal, but there’s been a lot of discussion already, so we’re moving to a GEP discussion format to try to capture feedback more efficiently.
```
shaneutt, line 53: Right now, I think this "why" section needs significantly more content, it doesn't seem to be fully supporting the "why should we do this" question:

> The existing resources we have (Service and HTTPRoute) don’t have space for this information currently

This just extends the "what": we're reiterating _what_ we want to see done by stating that it doesn't exist yet, but that isn't part of an answer to "why".

> and we have feedback from users via various implementations that it’s necessary

This comment needs some quantification, can we add the user feedback we've received here (even if paraphrased some) into some specific needs that justify the extra service metadata we're proposing?

> you could argue that some of this is table stakes for a layer 7 ingress controller.

This comment just seems to reiterate the desire of the author, but we need to quantify this: what capabilities of an L7 ingress controller demand the service metadata we're proposing?

>  In previous discussions, we discussed a variety of ways to achieve this goal, but there’s been a lot of discussion already, so we’re moving to a GEP discussion format to try to capture feedback more efficiently.

This doesn't seem to belong in "why": it seems to be more of a note to indicate there have been previous discussions (without links)?

site-src/geps/gep-1282.md
youngnick: 🤦‍♂️

site-src/geps/gep-1282.md
youngnick: It's a bit difficult to explain _why_ people want this things without talking in more detail about what they want, so I guess I'll need to fix this later in the GEP?

Bascially, the thing here is: Users have asked for stuff like backend TLS termination, and we don't have anywhere to put that in the current API. There are a few related features, so this GEP is about trying to make a broad solution to that class of problem - adding the description of extra capabilities to the backend.

I don't know how to say that more concisely than what I have here though.

site-src/geps/gep-1282.md
youngnick: It seems like what we need here is:
- a list of the actual feature requests we have in this repo
- examples of how other ingress controllers have requests for similar things
- More links to evidence.

Does that sound right?

site-src/geps/gep-1282.md
youngnick: I've had a crack after updating the rest of the GEP, PTAL.

site-src/geps/gep-1282.md
shaneutt: Sure, also potentially user stories. Ideally some stuff more written from the perspective of the cluster operator who has to set the ingress up for other teams.

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,100 @@
...
+  * Linkerd uses a [Server resource](https://linkerd.io/2.11/reference/authorization-policy/#server) (which is functionally pretty similar to Service in that it associates a name with a Pod selector, but also has other details like if the service supports Proxy protocol), along with a ServerAuthorization resource that specifies some constructs that sit more at the service mesh level, including identity and access control.
+
+In terms of other implementations existing use cases for features like this:
+- For Contour, there are annotations to allow the configuration of the following on a backend Service:
+  - Max connections
+  - Max pending requests
+  - Max requests
+  - Max retries
+  - Upstream protocol: This is what allows Contour to handle the
```
shaneutt, line 60: It looks like we didn't finish this list item?

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,100 @@
...
+  - Load balancer algorithms
+  - Connection pool settings
+  - Outlier Detection
+  - Tunnel settings for non-HTTP workloads
+
+The properties we're talking about all share two things:
+- They are tightly bound to the backend, rather than being important at the route level
+- And the service owner (application developer in the Gateway API personas) should control these settings, not the owner of the Gateway implementation (the Cluster Operator)
+
+A good example is the load balancing algorithms for backends, as distinct from load balancing for routes. Load balancing for routes is about choosing which route option the implementation should choose, while load balancing for backends is about which endpoint from the backend the implementation should send traffic to if that backend's route is chosen. It's likely that these will need to be set differently (for example, random load balancing at the route level for canarying between services, with Weighted Least Request at the backend level for the actual endpoints).
```
shaneutt, line 71: This distinction is pretty important, and plays in with the mesh use case quite a bit. I tried to take a stab at this previously and we decided on policy attachment, it will be very important where we go from here. No action required, just highlighting this.

shaneutt (approved):
One small comment left about a sentence that wasn't finished, and then I approve as I think the what and the why are at a point of clarity that this is a good `provisional` foundation for the work to come.

I'll leave the lgtm for another reviewer so as to allow more reading and not merge yet.

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,100 @@
...
+  - Max pending requests
+  - Max requests
+  - Max retries
+  - Upstream protocol: This is what allows Contour to handle the 
+- In Istio, the `DestinationRule` resource allows the configuration of many settings like this:
+  - Load balancer algorithms
+  - Connection pool settings
+  - Outlier Detection
+  - Tunnel settings for non-HTTP workloads
+
```
mikemorris, line 66:
```suggestion
- In Consul, [`ServiceDefaults`](https://www.consul.io/docs/connect/config-entries/service-defaults) allows specifying similar configuration:
  - Max connections
  - Max pending requests
  - Max concurrent requests
  - Protocol
  - Passive health check interval and max failures (outlier detection)
  - TLS SNI
  
```

mikemorris (approved):
Read through this (and the comment threads in the original doc too) and feels like it's in good shape for provisional status - only holding off on the LGTM slash command to avoid merging this preemptively.

site-src/geps/gep-1282.md
mikemorris: Worth cross-linking to prior discussion in https://github.com/kubernetes-sigs/gateway-api/pull/993 and https://github.com/kubernetes-sigs/gateway-api/issues/992?

shaneutt: sure

youngnick: I've added in references to this discussion.

youngnick:
Okay, with those changes, I think this has everything addressed. PTAL.

markmc:
Could the "decomposed view of Service" line of thinking from [GEP-1324](https://gateway-api.sigs.k8s.io/geps/gep-1324/) could be useful here?

Let's say:

- Part of the Service resource defines the "backend" - most obviously the selector, target ports/protocols, and ipFamilies
- The rest of the Service resource defines different types of "frontends" - internal load-balancers (aka proxy), external load-balancers, node ports, even "external name"
- We have a case where a Service frontend is used with a manually defined backend - that's a Service without a selector
- We have a case where a Service backend is used without any frontend - that's a "headless" Service - or maybe you'd call this a DNS-only frontend, since a DNS entry is defined with those backend IPs

And in GEP-1324, we're describing use cases where we are using Service to define backends (`a-v1` and `a-v2`) and separately also using Service to define a frontend (`a`). And AIUI, there are use cases for both service producers and consumers to want to define frontends for existing backends.

Given all of that, do we have a clear view on which category each property in this GEP falls into?

For example, for reencryption - the key pair for the proxy is clearly a frontend property. Is it obvious that the destination CA cert is a backend property, or is there an argument that this is how the frontend owner chooses what to trust and so should be a frontend property?

Stuff like max requests, load balancing algorithms, connection pool settings, etc. all seem clearly frontend properties.

Why does it matter? Well, maybe for true backend properties, we'd take the view that long-term it belongs in Service and we consider adding an interim resource. But for frontend properties, we'd define those in Gateway/xRoute?

youngnick:
Yes, I think that the frontend-backend split brought up by GEP-1324 is something that anything that deals with this will need to consider, and we need to make sure that whatever we do is harmonized between the ingress use cases and the mesh use cases for this API.

That said, I feel like spending too much time figuring that out falls into the "how" bucket - where this GEP is meant to represent our agreement that we need to have *somewhere* to put that information. Figuring out where that is - that's the next step.

markmc:
> That said, I feel like spending too much time figuring that out falls into the "how" bucket - where this GEP is meant to represent our agreement that we need to have _somewhere_ to put that information. Figuring out where that is - that's the next step.

Agree, except the description treats all of these properties as one category of information that should all be handled in the same way. The frontend/backend line of thinking might lead to a different conclusion.

youngnick:
To be honest, for the ingress use case (which is what we're targeting here, most of the capabilities I'm thinking of are about the backend properties - since up until now, the Gateway has been mainly concerned with a backend as "a bucket of endpoints", and the capabilities I'm describing here feel (to me at least) more like backend-service things in service mesh terms. Routes actually describe config that's more like service _frontends_ in the east-west use case.

markmc:
Focusing on backend properties sounds good. But the way I read the the "why" section, a lot of the stuff after "In terms of other implementations existing use cases for features like this" - including the "A good example is the load balancing algorithms for backends" bit - sounds like a lot of frontend stuff? Maybe I'm thinking about it wrong, and we should think of that sort of thing stuff as backend properties?

site-src/geps/gep-1282.md
```diff
@@ -0,0 +1,107 @@
...
+- In Consul, [`ServiceDefaults`](https://www.consul.io/docs/connect/config-entries/service-defaults) allows specifying similar configuration:
+  - Max connections
+  - Max pending requests
+  - Max concurrent requests
+  - Protocol
+  - Passive health check interval and max failures (outlier detection)
+  - TLS SNI (we don't support configuring a specific CA at this granularity, but publishing a public key in a status field could be useful)
+  
+The properties we're talking about all share two things:
+- They are tightly bound to the backend, rather than being important at the route level
```
hzxuzhonghu, line 75: Maybe related to the caller as well

shaneutt:
Given that this is `provisional` and doesn't appear to have any comments strongly opposing it: I would move that we get this merged to avoid it getting "stuck" for a long period of time for smaller comments, and then any contributors are free to add language tweaks or new sections as separate PRs thereafter to continue building on it.

youngnick:
> Focusing on backend properties sounds good. But the way I read the the "why" section, a lot of the stuff after "In terms of other implementations existing use cases for features like this" - including the "A good example is the load balancing algorithms for backends" bit - sounds like a lot of frontend stuff? Maybe I'm thinking about it wrong, and we should think of that sort of thing stuff as backend properties?

The load balancing thing is a particularly good example, because it can be both. Route-level load balancing is what I would consider a frontend feature, but endpoint-level load balancing is a backend feature. I'll update to try to make this a bit clearer.

site-src/geps/gep-1282.md
youngnick: They're relevant to the caller in how they take effect, but they are (or should be) _owned_ by the application owner, and apply to the service's pods and their associated endpoints, in Gateway API resource terms anyway.

youngnick:
Okay, I've pushed a small update to clarify the frontend/backend thing a bit. I think that this is probably ready to merge, and we can do any further updates later.

robscott:
Thanks @youngnick!

/lgtm

site-src/geps/gep-1282.md
hzxuzhonghu: Not exactly, in some case, A caller may use LB algorithm and connection pool settings different from caller B. How should gateway API allow this settings

PR #1334, opened 2022-08-16
conformance: listener references invalid secret

### LINKED ISSUE #1318: conformance: test gateway referencing an unexisting tls secret

A new conformance test case testing the presence of the status ResolvedRefs condition failed with the reason `InvalidCertificateRef` should be added to the conformance test suite.

To do so, the test case should create a gateway that references a non-existing TLS secret and check the failed `ResolvedRefs` condition is added to the corresponding listener status with the reason `InvalidCertificateRef`.

**Why this is needed**:

This is needed to ensure that when the referenced secret does not exist, the listener status is set accordingly.

mlavacca:
**What type of PR is this?**

/kind feature

A new conformance test case has been added to ensure that whenever a listener references an unexisting secret as CertificateRef, the Condition ResolvedRefs is set as failed with the reason InvalidCertificateRef.

Fixes #1318 

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-secret-missing-referenced-secret.go
+++ b/conformance/tests/gateway-secret-missing-referenced-secret.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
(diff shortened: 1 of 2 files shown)


conformance/tests/gateway-secret-missing-referenced-secret.go
```diff
@@ -0,0 +1,58 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewaySecretMissingReferencedSecret)
+}
+
+var GatewaySecretMissingReferencedSecret = suite.ConformanceTest{
+	ShortName:   "GatewaySecretMissingReferencedSecret",
+	Description: "A Gateway should fail to become ready if the Gateway has a certificateRef for a not existing Secret",
```
shaneutt, line 36:
```suggestion
	Description: "A Gateway should fail to become ready if the Gateway has a certificateRef for a non-existing Secret",
```

shaneutt (approved):
LGTM, but I'll let someone else put the official lgtm on it just so there's an extra pair of :eyes: on it.

PR #1337, opened 2022-08-17
Add converter pkg to house type conversion helpers

### LINKED ISSUE #1336: Consider a minor restructure of directory/file for util things
Currently, there is a directory named "validation" that houses a folder named "util" and a package and file both named "utils". Nothing in this "utils" package does validation. There are only a couple things and they do some translation from Go types to pointer of custom types, which is entirely the same a currently open PR, which adds a file named "helper.go" to the root of the API directory (for both API versions) and adds the same type of translations.

Current layout:
```
├── apis
│   ├── doc.go
│   ├── v1alpha2
│   │   ├── doc.go
│   │   ├── gateway_types.go
|   |   |...more files....
│   │   ├── validation
│   │   │   ├── doc.go
│   │   │   |...more files....
│   │   │   └── util
│   │   │       ├── utils.go
│   │   │       └── utils_test.go
```

New PR: [Add utils to translate custom types by abhijit-dev82 · Pull Request #1310 · kubernetes-sigs/gateway-api](https://github.com/kubernetes-sigs/gateway-api/pull/1310?w=1). An exception with this PR is that it also adds translation of the other way around (from custom type pointers to Go types). 

That made me wonder if these translations should be split into separate packages. I don't think so. If not, then where, under the existing "validation" folder? One could argue that these are utils exclusively there to support the code that does validation. However, as the PR above signals, and in my own usage, that conversion functionality is very useful for interacting with the entire gateway api code base.

The problem I am thinking to solve:
- keep like things together
- have things be easily discoverable if we can help it

I propose moving the "util" folder out from under "validation" so they are parallel to each other. And rename things so both the package and folder match. I personally prefer util/util (singular).

The content of the "helper.go" file in the above PR should go in this package. The file itself could have a more precise name. Why name it something vague (helper, util) when everything in it is doing something very specific? The package could remain with the name util/utils so it servers as a holding place for future type of utility/helper things, but the file or files could be organized in a way as to reflect their contents. Could be that all of the existing + content of the PR mentioned would go in a file named "translation.go" or "conversion.go", or they could be split into separate files if it would make it helpful to make some further distinction for that content.

Proposal:

```
├── apis
│   ├── doc.go
│   ├── v1alpha2
│   │   ├── doc.go
│   │   ├── gateway_types.go
|   |   |...more files....
│   │   ├── validation
│   │   │   ├── doc.go
│   │   │   |...more files....
│   │   └── util
│   │       ├── conversion.go
│   │       └── conversion_test.go
```

carlisia:
**What type of PR is this?**
/kind cleanup

I created the directory structure to address the issue below, and moved existing helpers to there. To note, I ended up naming the package `converter` because there are quite a number of types that could potentially end up with conversion helpers. 

Fixes #1336 

```release-note
Moved type conversion helpers from the `utils` package to a new package named `converter`.

```


apis/v1alpha2/converter/httproute.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2: This feels a little weird in that this seems somewhere between a file being moved (stick with original year) and a net new file (use current year). I think I'd lean towards sticking with the original year here, but don't feel strongly.

apis/v1alpha2/converter/httproute.go
```diff
@@ -14,20 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package utils
+package converter
```
robscott, line 17: Tiny nit: I'd prefer a name that wasn't so closely linked to conversion. I'm used to that referring to conversion between API versions, and we may want to reserve that terminology in case we ever need a conversion webhook for this project.

robscott (commented):
Thanks @carlisia! Have a couple nits but like the direction of this.

apis/v1alpha2/converter/httproute.go
carlisia: I had the same exact thinking and was split between wanting to keep the year the original content was added or going with the year the file was added 🤔

I think going with the year the content was added best reflects that this is a move and not an original. Will change.

apis/v1alpha2/converter/httproute.go
carlisia: Gotcha. The original terminology used was "translation" (see https://github.com/kubernetes-sigs/gateway-api/blob/96414f1ce72b2277cde5cb8e8abcd97ea039fe4a/apis/v1alpha2/validation/util/utils.go#L23). 

So, `translator`? `typeconverter`? Other idea?

robscott: Maybe we should have a `util` subpackage, and then optionally have a `util/translator` subpackage. That would make it easier to add future utils without adding more noise to the API types dir. Don't feel that strongly here, just prefer avoiding "conversion" terminology.

carlisia: Yes, that sounds helpful. To double check:

`apis/vi1alpha2/utils/translator/httproute.go`?

robscott: LGTM

carlisia: * s/utils/util

robscott:
Thanks @carlisia! This change LGTM, but will wait for another reviewer to formally sign off.

/approve

shaneutt (approved):
/lgtm

PR #1344, opened 2022-08-19
docs: couple of ReferenceGrant cleanups

markmc:
Remove deprecated ReferencePolicy using the hideTypePatterns config for the API docs generator.

ReferencePolicy has been renamed to ReferenceGrant, but we have retained ReferencePolicy as a deprecated type in order to make migration easier.

Keeping ReferencePolicy around in the docs is confusing for readers and probably doesn't provide any value.

Also fix path of ReferenceGrant image.

/kind cleanup

### DIFF
--- a/hack/api-docs/config.json
+++ b/hack/api-docs/config.json
@@ -4,7 +4,8 @@
     ],
     "hideTypePatterns": [
         "ParseError$",
-        "List$"
+        "List$",
+        "ReferencePolicy$"
     ],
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
This makes sense to me :+1: 

Since it has to do with the visibility of an existing API, I'll wait for another maintainer to give the lgtm.

robscott:
Thanks!

/lgtm

PR #1345, opened 2022-08-19
docs: display beta before alpha in API spec

markmc:
We display the alpha types in the API spec before the beta types. So readers see all of our types - including experimental and deprecated - before seeing the package with the more limited set of "standard" types.

Reverse this by manually iterating the packages list so that the beta package is displayed before the alpha package.

I could buy that this is too much of a hack to be worth it though!

/kind documentation


shaneutt (approved):
This seems to make perfect sense to me :+1: 

I looked at the netlify preview and the ordering does seem to be an improvement.

/lgtm

Issue #1347, opened 2022-08-20
Document versioning plan and upgrade recommendations


robscott:
As we're thinking about the long term use of CRDs for this API, we'll want to ensure that both users and implementations have a smooth experience across CRD upgrades. 

Here's a high level starting point based on discussions with folks from sig-apimachinery around the proper way to handle stored and served versions. 

Minor release 1 (v0.5.0 for these resources):

* Release new API version (v1beta1)
* Leave old version as storage version (v1alpha2)
* Serve both API versions to simplify migration

Minor release 2 (v0.6.0 for these resources):

* Set new version as storage version (v1beta1)
* Stop serving old API version (v1alpha2)
* Recommend users run a tool such as [kube-storage-version-migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator) in their clusters if they have resources that were created with the alpha API

Minor release 3 or later (v0.7.0 for these resources):

* Remove old API version from the API

/assign

PR #1348, opened 2022-08-20
Beta resources: Setting storage version to v1beta1, not serving v1alpha2

robscott:
**What type of PR is this?**
/kind cleanup
/kind deprecation

This affects resources that graduated to v1beta1 in the v0.5.0 release.
Although that specific change is relatively small, it required a lot of
related work:

* Restructuring examples to remove v1alpha2 examples where they were no
  longer useful and their installation would fail
* Updating conformance tests to use v1beta1 where possible
* Some related fixes to v1alpha2 blog post to account for removal of
  v1alpha2 example

This is closely related to #1347

```release-note
GatewayClass, Gateway, and HTTPRoute are now only supported with the v1beta1 version of the API. The v1alpha2 API versions of these resources will be fully removed in a future release.
```


robscott:
This is a big change. Adding a hold for consensus.

/hold

hack/verify-examples-kind.sh
```diff
@@ -106,6 +105,9 @@ for CHANNEL in experimental standard; do
         echo Examples failed as expected
 done
 
+# Undo workaround from earlier
```
youngnick, line 108: Ooh, this is a good catch.

youngnick (commented):
Wow, this was a surprisingly big one. Most of the changes are pretty straightforward renames, but this probably needs someone to do another visual check through the docs and make sure they make sense (I am going a bit crosseyed).

/lgtm

hack/verify-examples-kind.sh
markmc: "Temporary workaround for release" sounds like something that should just be removed? If it's needed permanently, maybe we need to update the comment?

markmc:
This needs updating in `admission_webhook.yaml` doesn't it?

```
webhooks:
  - name: validate.gateway.networking.k8s.io
    matchPolicy: Equivalent
    rules:
      - operations: [ "CREATE" , "UPDATE" ]
        apiGroups: [ "gateway.networking.k8s.io" ]
        apiVersions: [ "v1alpha2" ]
        resources: [ "gateways", "gatewayclasses", "httproutes" ]
```

hack/verify-examples-kind.sh
robscott: I think you're right, this probably shouldn't exist at all. There was a time where we wanted the manifests installed from the main branch to be at least somewhat useful and refer to the latest released image instead of the latest image build from the main branch. All of our installation instructions clearly refer to a versioned release so this is likely entirely unnecessary, will revert this change altogether.

robscott:
Good catch @markmc! Should be fixed by https://github.com/kubernetes-sigs/gateway-api/pull/1365

hack/verify-examples-kind.sh
robscott: Actually as I was digging into this more, it looks more complicated than initially expected and I don't want to expand this PR any further. I've created https://github.com/kubernetes-sigs/gateway-api/issues/1366 to track this and added it to the v0.6.0 milestone.

shaneutt (approved):
/lgtm

robscott:
Fixes #1372

/hold cancel

robscott:
Missed updating one conformance test to use v1beta1, PTAL.

youngnick (commented):
/lgtm

PR #1349, opened 2022-08-20
Documenting supported versions

robscott:
**What type of PR is this?**
/kind documentation

As this project matures, it's important to document our support guarantees going forward. This proposes a set of basic commitments that I think are both safe for us as maintainers and helpful for anyone building a product with this API. 

```release-note
NONE
```

### DIFF
--- a/site-src/concepts/versioning.md
+++ b/site-src/concepts/versioning.md
@@ -144,3 +144,15 @@ this project.
 We do not provide stability guarantees for source code imports. The Interfaces
 and behavior may change in an unexpected and backwards-incompatible way in any
 future release.
+
+## Supported Versions
+
+This project aims to provide support for a wide range of Kubernetes versions with
+consistent upgrade experiences across versions. To accomplish that, we commit to:
+
+1. Support a minimum of the most recent 5 Kubernetes minor versions.
+2. Ensure that all standard channel changes between v1beta1 and v1 are fully
+   compatible and convertible.
+3. Take every possible effort to avoid introduction of a conversion webhook. If
+   a conversion webhook needs to be introduced, it will be supported for the
+   lifetime of the API, or at least until an alternative is available.


robscott:
Adding a hold for consensus, want to get consensus from majority of maintainers before merging.

/hold

site-src/concepts/versioning.md
```diff
@@ -144,3 +144,15 @@ this project.
 We do not provide stability guarantees for source code imports. The Interfaces
 and behavior may change in an unexpected and backwards-incompatible way in any
 future release.
+
+## Supported Versions
+
+This project aims to provide support for a wide range of Kubernetes versions with
+consistent upgrade experiences across versions. To accomplish that, we commit to:
+
+1. Support a minimum of the most recent 5 Kubernetes minor versions.
```
shaneutt, line 153: :+1:

site-src/concepts/versioning.md
```diff
@@ -144,3 +144,15 @@ this project.
...
+2. Ensure that all standard channel changes between v1beta1 and v1 are fully
+   compatible and convertible.
```
shaneutt, line 155: :+1:

site-src/concepts/versioning.md
```diff
@@ -144,3 +144,15 @@ this project.
...
+3. Take every possible effort to avoid introduction of a conversion webhook. If
+   a conversion webhook needs to be introduced, it will be supported for the
+   lifetime of the API, or at least until an alternative is available.
```
shaneutt, line 158: :+1:

shaneutt (approved):

site-src/concepts/versioning.md
youngnick: I would recommend no longer than 3 to enable us to move faster, so that we can use features like https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2876-crd-validation-expression-language/README.md once they get to GA.

I acknowledge that this messes with people, and am okay to discuss more, but that's my preference.

robscott: My theory is that a significant majority of users and implementations of this API will use and support versions of Kubernetes that trail Kubernetes upstream by more than 3. I think there are substantial sets of users on Kubernetes 1.21 and 1.22 today that would be unsupported by a policy that limited this to 3. For example, when Kubernetes 1.25 is released tomorrow, [ingress-nginx will support the last 6 minor versions of Kubernetes](https://github.com/kubernetes/ingress-nginx#support-versions-table). Similarly, [Istio has supported up to 5 Kubernetes minor versions](https://istio.io/latest/docs/releases/supported-releases/), and runs tests as far back as 9 versions. Although the ideal would be that all users are using the latest version(s) of Kubernetes, that is not the reality today.

If we can always confidently tell users to install the latest version of CRDs in their cluster, it will significantly simplify our support and release process. This would also be a big win for implementers of the API as they could be confident that they will be able to support the latest release of this API and cover the majority of their customers at the same time. 

I know the Ingress v1 transition was painful for many implementations since it only provided 3 Kubernetes minor versions to transition. That forced many to provide two separate sets of releases depending on the version(s) of the API that were available in the cluster. If we're not careful, we could introduce that kind of painful transition with _each_ minor release of the Gateway API. 

Here's a rough comparison of what the experience might look like if we supported 3 or 5 trailing Kubernetes versions:

| Topic | 3 Trailing Kubernetes Versions | 5 Trailing Kubernetes Versions |
|-|-|-|
| Implementation Releases | Each new release may require different versions targeting different versions of Kubernetes | Single release implementing latest CRDs |
| Gateway Bug Fixes and Security Patches | Should really be rolled back to previous Gateway API releases | No need for backports, everyone can install latest CRDs |
| Documentation | Need to carefully maintain documentation for each previous release of the API | Maintaining docs for the latest release is likely sufficient |
| New Kubernetes Features | Can access new Kubernetes features 8 months earlier | Need to wait 8 additional months to use latest features in CRD API |

As a high level summary, I think by supporting more Kubernetes versions, we'd be able to only support the latest version of Gateway API. Not only does this simplicity help maintainers of this project, it also helps anyone building products on top of it.

Note: I should really clarify that I'm talking about ensuring that anything we expect users to deploy such as CRDs or webhook validation is supported for 5 minor versions. I do not think that our Go codebase or dependencies need to provide the same level of version support.

brianehlert: I would expect following the pattern of upstream.
But I would also not expect so mush version change that it is a significant concern.

From my point of view - if an API is churning so significantly that it constantly needs versioning (implying behavior change) then it should remain alpha / experimental and not move into beta or stable where the compatibility guarantees begin.
I get the impression from the comment that a decent amount of rapid versioning is expected.
IMHO - extending a schema should not drive a version change.  Modifying behavior should.

No API consumer likes dealing with version change.  Especially operations folks (more so than developers).
Too much churn impacts adoption.

robscott: I think the real concern here is about CEL as @youngnick mentioned above. Adopting that for our CRDs would likely mean we didn't need to maintain a validating webhook. However that's only available in Kubernetes 1.25+, so whenever we adopt CEL we'll be effectively dropping support for older versions of Kubernetes. Is 3 minor versions sufficient? Or do we need longer? As we saw with Ingress, 3 minor versions is not a lot of time, and many/most Ingress implementations supported a broader range of Kubernetes versions than that.

In that case there were two options if you wanted to support more that 3 k8s versions:

1. Update your controller to conditionally read/write to v1 or v1beta1 depending on the version of Kubernetes in use.
2. Release separate versions of your controller to support different versions of Kubernetes.

I'm hoping to avoid that here.

youngnick: Yes, CEL in particular is a tricky one, because using it requires access to at least Kubernetes v1.24, and will hopefully let us massively simplify or remove the validating webhook.

I've thought some more on this, and I think 5 versions is about as good a compromise between velocity and pragmatism as we can hope for - with a bit of luck there won't be too many more changes at the Kubernetes level that will require particular versions like CEL does, so 5 versions won't be too bad.

@brianehlert we've specifically designed the graduation requirements for resources around the upstream API conventions so as to make _schema additions_ (like adding fields etc) not require version bumps to the API (in terms of the resources' `apiVersion`, for example `v1beta1`).

However, because we are specific about conformance, we _do_ include the conformance testing (and validating webhook) into the release bundles (like `v0.5.0`). In general, we're trying to keep minor version bumps here as compatible as possible, but breakage between minor versions will probably mostly be of concern to implementation authors rather than users of the API. (For example, in #1364, I'm trying to update and harmonize the way status works, which will change how authors need to pass conformance, and change how you can tell your resource has been accepted, but not change any of the resource definitions themselves).

shaneutt: I'm open to compromising on the 5 versions, but wondering if the time to do that isn't as needed (that is to say, as we see a need jump up to 5 from 3) as opposed to committing to it now where reducing is harder :thinking:

robscott: >I'm open to compromising on the 5 versions, but wondering if the time to do that isn't as needed (that is to say, as we see a need jump up to 5 from 3) as opposed to committing to it now where reducing is harder

This kind of commitment is important for implementations building products on top of the API. My own bias is that implementations like GKE need to make long term support decisions now, and that's much simpler to do when the OSS support windows are clearly documented.

shaneutt: Well, if everyone else is convinced then I can role with it. I consider this comment resolved.

shaneutt: I am persuaded, I consider this comment resolved.

PR #1350, opened 2022-08-22
gep-1324: add a definition of "gateway" to glossary

markmc:
The glossary section goes to great length to create a clear understanding of some terms that are especially important in the
context of a discussion of supporting mesh use-cases in the API.

However, we don't currently define the "gateway" term. Rectify this by adding a definition that makes reference to many other terms that might crop up in this conversation - e.g. traffic translation, network elements, network endpoints, senders/receivers, consumers/producers, reachability information, routing and routing discriminators, load-balancing, filtering, encryption, NAT, proxying, encapsulation, logical gateways, distributed gateways, segmentation, North/South vs East/West, micro-segmentation and zero-trust, and ingress vs egress gateways. Phew!

Hopefully this is neither hilariously incorrect or wildly out of kilter with ongoing GAMMA discussions. The idea is to generate some discussion about whether we all understand the abstract idea of a "gateway" in the same way. HTH

/kind design

### DIFF
--- a/site-src/geps/gep-1324.md
+++ b/site-src/geps/gep-1324.md
@@ -127,6 +127,34 @@ A service **consumer** is someone that utilizes a service by communicating with 
 
 Consumer and producer are personas, not concrete workloads or resources. One important aspect of these personas is that they may live in different trust boundaries (typically namespace).
 
+### **Gateway**
+
+We use the term “gateway” in reference to an abstract concept capturing the idea of a component which translates network traffic.
+
+We are using the term in a deliberately abstract way, therefore it is important that readers do not confuse this abstract interpretation of the term - and associated terms - with more concrete interpretations in more specific contexts. For example, we are specifically not referring to the meaning of “gateway” in IP packet routing, nor even the meaning of “translation” in that context.
+
+We can think of a gateway as being a component (aka “network element”) which implements a layer of indirection between a sender of traffic (e.g. a consumer) and its receivers (e.g. a producer). Senders and receivers are all known as “network endpoints”. Gateways consume
+“reachability information” describing endpoints in order to gain knowledge of available receivers.
+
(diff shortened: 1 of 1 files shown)


markmc:
A thought that only occurs to me now that I've posted - we could say that DNS can play a role in implementing the "layer of indirection" that a gateway could encapsulate? An obvious example is DNS based load-balancing with headless services.

But a DNS name is itself a powerful layer of indirection, and we talk about e.g. StatefulSet replicas having a "stable network identity" so that consumers can refer to a specific replica evens as pods are deleted and recreated.

site-src/geps/gep-1324.md
```diff
@@ -127,6 +127,34 @@ A service **consumer** is someone that utilizes a service by communicating with
 
 Consumer and producer are personas, not concrete workloads or resources. One important aspect of these personas is that they may live in different trust boundaries (typically namespace).
 
+### **Gateway**
+
+We use the term “gateway” in reference to an abstract concept capturing the idea of a component which translates network traffic.
+
+We are using the term in a deliberately abstract way, therefore it is important that readers do not confuse this abstract interpretation of the term - and associated terms - with more concrete interpretations in more specific contexts. For example, we are specifically not referring to the meaning of “gateway” in IP packet routing, nor even the meaning of “translation” in that context.
+
+We can think of a gateway as being a component (aka “network element”) which implements a layer of indirection between a sender of traffic (e.g. a consumer) and its receivers (e.g. a producer). Senders and receivers are all known as “network endpoints”. Gateways consume
```
youngnick, line 136: I've been using "translates between somewhere with less context to somewhere with more context", but I think this is probably a better one for this specific use case.

youngnick (commented):
This LGTM, I'll approve, and leave the second opinion to someone else.

/approve

shaneutt (approved):
/lgtm

PR #1356, opened 2022-08-23
conformance: listener `invalidRouteKind`

### LINKED ISSUE #1320: conformance: test gateway when an unsupported Route kind is specified by the Listener.

A new conformance test case testing the presence of the status ResolvedRefs condition failed with the reason `ListenerReasonInvalidRouteKinds` should be added to the conformance test suite.

To do so, the test case should create a gateway with a listener specifying an unsupported Route Kind and check the `ResolvedRefs` condition is added to the corresponding listener status with the reason `ListenerReasonInvalidRouteKinds`.

**Why this is needed**:

This is needed to ensure that when a listener references an unsupported Route kind, the Listener status is updated accordingly.

mlavacca:
**What type of PR is this?**

/kind feature

A new conformance test case has been added to ensure that whenever a listener supports a Route type incompatible with the ProtocolType, the `ListenerStatus.SupportedKinds` are updated accordingly, and the `ResolvedRefs` listener condition fails with reason `InvalidRouteKinds `.

Fixes #1320 

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-unsupported-route-kind.go
+++ b/conformance/tests/gateway-unsupported-route-kind.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


conformance/tests/gateway-unsupported-route-kind.go
```diff
@@ -0,0 +1,72 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayUnsupportedRouteKind)
+}
+
+var GatewayUnsupportedRouteKind = suite.ConformanceTest{
+	ShortName:   "GatewayUnsupportedRouteKind",
+	Description: "A Gateway in the gateway-conformance-infra namespace should fail to become ready if explicitly supports a route incompatible with the protocol type",
```
shaneutt, line 36: Technically as demonstrated by your tests below, I think this language may be more accurate?

```suggestion
	Description: "A Gateway in the gateway-conformance-infra namespace should fail to become ready if explicitly supports any route type incompatible with the protocol type, even if there are types which are valid as well.",
```

lmkwyt

PR #1319, opened 2022-08-12
broken link under site_src

kundan2707:
**What type of PR is this?**
/kind bug
/kind documentation

**What this PR does / why we need it**:
Multiple links are broken under site-src folder

**Which issue(s) this PR fixes**:

Fixes #1311 

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1357, opened 2022-08-23
Fixing docs links that incorrectly had site-src/ prefix

robscott:
**What type of PR is this?**
/kind documentation

As caught by @plgingembre, some of our docs links incorrectly had the "/site-src" prefix, this fixes that.

```release-note
NONE
```

### DIFF
--- a/site-src/api-types/httproute.md
+++ b/site-src/api-types/httproute.md
@@ -17,7 +17,7 @@ The specification of an HTTPRoute consists of:
   fields.
 
 The following illustrates an HTTPRoute that sends all traffic to one Service:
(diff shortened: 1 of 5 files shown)


markmc:
I guess this is more-or-less a revert of #1319 ? Anything else in there that should also be reverted

/lgtm

markmc:
> I guess this is more-or-less a revert of #1319 ? Anything else in there that should also be reverted

For example:

```
-[httproute]: /references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute
+[httproute]:https://gateway-api.sigs.k8s.io/api-types/httproute/
```

That's making the ```api-types/httproute/``` page link back to itself rather than to the API spec

PR #1358, opened 2022-08-23
Revert "broken_site_src"

robscott:
**What type of PR is this?**
/kind documentation

As pointed out by @markmc when reviewing #1357, we probably want to roll back all of #1319. I think this was caused by a common point of confusion. When browsing links in site-src within the repo, they will be broken, that is intentional. The goal of site-src is to build https://gateway-api.sigs.k8s.io/, not to be useful independently of that. I'm not sure how/where to document that because I know it's confusing, suggestions welcome.

```release-note
NONE
```


shaneutt (approved):
:facepalm: my bad.

/lgtm

PR #1361, opened 2022-08-24
Document the edge case with repeated query params

### LINKED ISSUE #1271: Document the edge case with repeated query params

Extend the documentation of the Name field of the HTTPQueryParamMatch to describe the edge case of having repeated query params with the same name in the query string of a request. For example, `?a=1&a=2`.

**Why this is needed**:

This is needed to
* Ensure compatibility across Gateway implementations.
* To prevent users from relying on implementation-specific behavior which could cause issues when switching implementations.

For the query string `?a=1&a=2`, currently, it is not specified whether the following matches match or not:
(1)
```yaml
queryParams:
- name: a
  value: "1"
```
(2)

```yaml
queryParams:
- name: a
  value: "2"
```

Note that there is no standard that describes how a data plane must handle that case. Envoy, nginx use the first value of a param.
See also https://github.com/kubernetes-sigs/gateway-api/discussions/1269

As discussed in the community meeting on July 18, 2022, in the documentation it makes sense to target the two sides of the Gateway API:
* The implementers: if the data plane allows, they should match against the first value of a param.
* The users: users should not route traffic based on repeated query params to guard themselves against potential differences in the Gateway API implementations.

See also the documentation of the repeated headers case https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/httproute_types.go#L370-L371

pleshakov:
**What type of PR is this?**
/kind documentation

Ensure compatibility across Gateway implementations.
To prevent users from relying on implementation-specific behavior which could cause issues when switching implementations.

Fixes #1271

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -430,6 +430,15 @@ type HTTPQueryParamMatch struct {
 	// entry with an equivalent name MUST be considered for a match. Subsequent
 	// entries with an equivalent query param name MUST be ignored.
 	//
+	// If a query param is repeated in an HTTP request, the behavior
+	// is purposely left undefined, since different data planes have different
+	// capabilities. However, it's *recommended* that implementations should
+	// match against the first value of the param if the data plane supports it,
+	// as this behavior is expected in other load balancing contexts outside of
+	// the Gateway API.
+	// Users should not route traffic based on repeated query params to guard
+	// themselves against potential differences in the implementations.
+	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=256
 	Name string `json:"name"`
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -2961,7 +2961,17 @@ spec:
                                     param names, only the first entry with an equivalent
                                     name MUST be considered for a match. Subsequent
                                     entries with an equivalent query param name MUST
-                                    be ignored."
+                                    be ignored. \n If a query param is repeated in
+                                    an HTTP request, the behavior is purposely left
+                                    undefined, since different data planes have different
+                                    capabilities. However, it's *recommended* that
+                                    implementations should match against the first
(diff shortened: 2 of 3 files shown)


apis/v1beta1/httproute_types.go
```diff
@@ -429,6 +429,13 @@ type HTTPQueryParamMatch struct {
 	// entry with an equivalent name MUST be considered for a match. Subsequent
 	// entries with an equivalent query param name MUST be ignored.
 	//
+	// If a query param is repeated in an HTTP request, it is
+	// implementation-specific behavior as to how this is represented. However,
+	// implementations should match against the first value of the param if
+	// the data plane supports it.
```
youngnick, line 435: This seems a little unclear to me, how about this?
```suggestion
	// If a query param is repeated in an HTTP request, the behavior
	// is purposely left undefined, since different data planes have different
	// capabilities. However, it's *recommended* that implementations should
	// match against the first value of the param if the data plane supports it,
	// as this behavior is expected in other contexts.
```
(Sorry, I have had no bandwidth starting the new job).

pleshakov: @youngnick 

Good suggestion 👍 

> as this behavior is expected in other contexts.

This looks a little unclear to me.  Do you imply any similar Gateway API guidelines?

youngnick: Ah, sorry, I missed this. By "in other contexts", I meant "in other, non-Gateway API load balancing contexts".

pleshakov: @youngnick thanks! I applied your suggestion with added clarification about load balancing contexts.

youngnick:
This LGTM, so I'll approve and leave the final lgtm command for another maintainer.

/approve

youngnick:
Sorry @pleshakov, you'll need to rebase this one, and then run `make generate` to get it to pass CI. (There's a fix for the Netlify deploy preview failure in `main` now).

pleshakov:
I rebased against the latest main and also squashed the commits.

pleshakov:
forgot to `make generate`. fixed.

shaneutt (approved):
/lgtm

Issue #1102, opened 2022-04-08
Conformance Tests: HTTPRoute Matching


robscott:
This issue is intended to track conformance test development for HTTPRoute matching. Comment below if you're interested in working on covering any of these areas.

**Core Capabilities:**
- [x] Path prefix matching
- [x] Exact path matching (#1147)
- [x] Header matching (#1134)
- [ ] Combinations of the above

**Extended Capabilities:**
- [x] Method matching (@skriss)
- [x] Query param matching (@skriss)
- [ ] Combinations of the above with core capabilities

candita:
@robscott Interested

robscott:
@candita great, thanks! Any particular tests described in this issue you want to take on? Let me know if you have any questions as you're getting started.

skriss:
I've submitted https://github.com/kubernetes-sigs/gateway-api/pull/1134 for header matching.

skriss:
I can take on tests for query param and method matching (@robscott note that query param should move under "Extended" conformance per https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPQueryParamMatch).

Has a pattern been discussed for separating extended-conformance tests from core conformance ones? Is it intended to use the `SupportedFeatures` field (https://github.com/kubernetes-sigs/gateway-api/blob/master/conformance/conformance_test.go#L51-L53)? If not, I can put some thought into how to structure things.

robscott:
Thanks @skriss! I've moved query param matching into extended. I think right now our best approach is to have any extended feature gated by the "SupportedFeatures" option, but open to other ideas.

skriss:
Circling back here, @robscott you can mark query param matching as done, and I'll try to get to method matching soon.

Would be great if someone else wanted to pick up the combination tests!

robscott:
Updated, thanks @skriss!

sunjayBhatia:
I can work on some tests for combinations of matching capabilities

Also see https://github.com/kubernetes-sigs/gateway-api/pull/1370 for a relevant PR

youngnick:
/remove-lifecycle stale

@skriss, @sunjayBhatia, is there further work outstanding here?

skriss:
> /remove-lifecycle stale
> 
> @skriss, @sunjayBhatia, is there further work outstanding here?

[...]

PR #1370, opened 2022-08-31
conformance: Add test for ensuring route match order is respected

sunjayBhatia:
**What type of PR is this?**

/kind cleanup
/kind documentation

Adds a conformance test for a currently ambiguous header matching case.

The request with headers `version: two` and `color: blue` matches both

```
  # Matches "version: two"
  - matches:
    - headers:
      - name: version
        value: two
    backendRefs:
    - name: infra-backend-v2
      port: 8080
```
and
```
  # Matches "color: blue" OR "color: green"
  - matches:
    - headers:
      - name: color
        value: blue
    - headers:
      - name: color
        value: green
    backendRefs:
    - name: infra-backend-v1
      port: 8080
```

with Contour, since we generate matches for OR matches as separate routes which are then sorted (by number of header matches and then alphanumerically on header), the second route rule above would have precedence

However, according to this language: https://github.com/kubernetes-sigs/gateway-api/blob/a3f769e5c690546140b6b9f0db5aea4d766b8f68/apis/v1beta1/httproute_types.go#L161-L163

the *first* route should be given precedence.

~~As far as I can tell there isn't anything in the `HTTPRouteMatch` documentation that covers this.~~

~~There is a [note](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteRule) about preference going to the route match with `the largest number of ... Header matches`, however does that does not seem to apply in the OR case here and maybe needs to be clarified. If instead of including the header `color` in our request/match we used something that sorted alphanumerically after `version`, Contour at least would route to `infra-backend-v2` (the first match listed above)~~

~~Seems like we could add to the documentation for route match conflict resolution section something about what to do in this case?~~

Updates: #1102

```release-note
NONE
```


shaneutt (commented):
When looking just at the diff, it's not clear how this differentiates from the test case right above it. Am I missing something?

sunjayBhatia:
> When looking just at the diff, it's not clear how this differentiates from the test case right above it. Am I missing something?

the previous case is distinct as it should exactly match this rule:

```
  # Matches "version: two" AND "color: orange"
  - matches:
    - headers:
      - name: version
        value: two
      - name: color
        value: orange
    backendRefs:
    - name: infra-backend-v1
      port: 8080
```

shaneutt (approved):
Ah! It was too early in the morning :coffee:++

I was looking at the wrong thing and was getting confused. Looks good.

/lgtm

sunjayBhatia:
> Ah! It was too early in the morning ☕++
> 
> I was looking at the wrong thing and was getting confused. Looks good.
> 
> /lgtm

Also generally have a question that goes with this test

```
There is a note about preference going to the route match with the largest number of ... Header matches, however does that does not seem to apply in the OR case here and maybe needs to be clarified. If instead of including the header color in our request/match we used something that sorted alphanumerically after version, Contour at least would route to infra-backend-v2 (the first match listed above)

Seems like we could add to the documentation for route match conflict resolution section something about what to do in this case?
```

The test case here isn't explicitly covered by any of the documentation in the spec as far as I can tell

sunjayBhatia:
There are a few ambiguities like this that conformance tests can show easily but may be missed in the godoc/field descriptions in the API. In this case I think the test change should be accompanied by an addition to the language in the route matching API description. Is this the right process to do so in the same change?

shaneutt:
> There are a few ambiguities like this that conformance tests can show easily but may be missed in the godoc/field descriptions in the API. In this case I think the test change should be accompanied by an addition to the language in the route matching API description. Is this the right process to do so in the same change?

For that we would probably want a couple of reviewers whereas for this it's just adding a functional test case, so I feel it's reasonable to merge this and then add the language change as a separate PR. It's your choice however.

sunjayBhatia:
Actually re-reading the spec

```
	// If ties still exist within the Route that has been given precedence,
	// matching precedence MUST be granted to the first matching rule meeting
	// the above criteria.
```

so actually, this test should route to the v1 backend since the match for `version: two` comes before the other possible match in the list of routes

sunjayBhatia:
hm, didnt mean to remove reviewers, but re-requesting a review did so

shaneutt (approved):
I think it looks good, but I'll just ask @youngnick to do a double tap. :+1:

youngnick (commented):
Thanks for picking this up, @sunjayBhatia!

/lgtm

skriss:
@sunjayBhatia reading the current spec literally, I would think this request should go to `infra-backend-v1`. Per https://github.com/kubernetes-sigs/gateway-api/blob/a3f769e5c690546140b6b9f0db5aea4d766b8f68/apis/v1beta1/httproute_types.go#L143-L152, we should prioritize the **Rule** with the greatest number of header matches, which in this case is:

```
 # Matches "color: blue" OR "color: green"
  - matches:
    - headers:
      - name: color
        value: blue
    - headers:
      - name: color
        value: green
    backendRefs:
    - name: infra-backend-v1
      port: 8080
```

Since this is a single Rule that has 2 header matches (even though each individual Match within the role only has one each).

shaneutt (approved):
/lgtm

sunjayBhatia:
@shaneutt re: https://github.com/kubernetes-sigs/gateway-api/pull/1370#issuecomment-1238472815

looks like we may not have agreement on how this should work and the exact language involved

sunjayBhatia:
See https://github.com/kubernetes-sigs/gateway-api/issues/1381

shaneutt:
> looks like we may not have agreement on how this should work and the exact language involved

Yep my apologies. Let's follow up in #1381 as you've proposed.

PR #1373, opened 2022-09-02
Add `ResponseHeaderModifier` to `HTTPRouteFilter` 

aryan9600:
**What type of PR is this?**
/kind feature
/kind api-change

This PR implements [GEP-1323](https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/geps/gep-1323.md). It renames `HTTPRequestHeaderFilter` to `HTTPHeaderFilter` and introduces a new field `ResponseHeaderModifier` in `HTTPRouteFilter`. It also adds conformance tests for the same.

Relates to #1323

```release-note
A new field `responseHeaderModifier` is added to `.spec.rules.filters`, which allows for modification of HTTP response headers
```


conformance/utils/http/http.go
```diff
@@ -38,9 +38,13 @@ type ExpectedResponse struct {
 	// expected to match Request.
 	ExpectedRequest *ExpectedRequest
 
-	StatusCode int
-	Backend    string
-	Namespace  string
+	// TODO: maybe move these into a dedicated type named Response?
```
shaneutt, line 41: We should either:

a) resolve/remove this TODO as a part of this PR
b) if there's sufficient cause to follow up on this after this PR, create an issue for it and link it here so the TODO doesn't get lost.

aryan9600: I think there is be sufficient cause, but I don't think it'd be wise to mix this with the current PR changes. I'll create an issue for this and open a follow up PR after this gets merged.

youngnick (commented):
LGTM as long as well are all fine with moving this straight to beta (I vote yes to that).

apis/v1beta1/httproute_types.go
```diff
@@ -631,6 +639,14 @@ const (
 	// Support in HTTPBackendRef: Extended
 	HTTPRouteFilterRequestHeaderModifier HTTPRouteFilterType = "RequestHeaderModifier"
 
+	// HTTPRouteFilterResponseHeaderModifier can be used to add or remove an HTTP
+	// header from an HTTP response before it is sent to the client.
+	//
+	// Support in HTTPRouteRule: Extended
+	//
+	// Support in HTTPBackendRef: Extended
+	HTTPRouteFilterResponseHeaderModifier HTTPRouteFilterType = "ResponseHeaderModifier"
```
robscott, line 648: We should mark this value as experimental like this:

https://github.com/kubernetes-sigs/gateway-api/blob/2e82b50d2b91cc2f9c4dc96181fdcc9131c27482/apis/v1beta1/shared_types.go#L113

robscott (commented):
Thanks @aryan9600! Mostly LGTM, just want to be sure that this is added as an experimental feature.

apis/v1beta1/httproute_types.go
```diff
@@ -571,8 +571,8 @@ type HTTPRouteFilter struct {
 	// Reason of `UnsupportedValue`.
 	//
 	// +unionDiscriminator
-	// +kubebuilder:validation:Enum=RequestHeaderModifier;RequestMirror;RequestRedirect;ExtensionRef
-	// <gateway:experimental:validation:Enum=RequestHeaderModifier;RequestMirror;RequestRedirect;URLRewrite;ExtensionRef>
+	// +kubebuilder:validation:Enum=RequestHeaderModifier;ResponseHeaderModifier;RequestMirror;RequestRedirect;ExtensionRef
+	// <gateway:experimental:validation:Enum=RequestHeaderModifier;ResponseHeaderModifier;RequestMirror;RequestRedirect;URLRewrite;ExtensionRef>
```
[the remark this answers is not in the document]
robscott: Woops, yep, you're right.

config/crd/standard/gateway.networking.k8s.io_httproutes.yaml
```diff
@@ -1848,6 +1848,115 @@ spec:
                                       - 302
                                       type: integer
                                   type: object
+                                responseHeaderModifier:
```
robscott, line 1851: This PR should not result in a change to the standard set of CRDs. Once you get `<gateway:experimental>` tags in all the right places, this diff should disappear.

conformance/tests/httproute-response-header-modifier.go
```diff
@@ -0,0 +1,161 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteResponseHeaderModifier)
+}
+
+var HTTPRouteResponseHeaderModifier = suite.ConformanceTest{
```
robscott, line 33: It's really great to add a conformance test up front like this, but I think we need to hide it behind a "SupportedFeature" so it's opt-in while it's still in the experimental channel: https://github.com/kubernetes-sigs/gateway-api/blob/48cf5d9e4756dd65be21fa26e17e3627903c070a/conformance/utils/suite/suite.go#L42-L55

apis/v1beta1/httproute_types.go
```diff
@@ -581,7 +581,15 @@ type HTTPRouteFilter struct {
...
 	// +optional
-	RequestHeaderModifier *HTTPRequestHeaderFilter `json:"requestHeaderModifier,omitempty"`
+	RequestHeaderModifier *HTTPHeaderFilter `json:"requestHeaderModifier,omitempty"`
+
+	// ResponseHeaderModifier defines a schema for a filter that modifies response
+	// headers.
+	//
+	// Support: Extended
+	//
+	// +optional
```
robscott, line 591: Need `<gateway:experimental>` here.

robscott (commented):
Thanks @aryan9600! This looks really close, just a few more comments.

conformance/tests/httproute-response-header-modifier.go
aryan9600: Thanks for bringing this up, I was going to ask about the same thing. I was thinking about what you suggested, but I could not find a way to configure which `SupportedFeature`s should be used while running the conformance tests, as there's no such flag in `conformance/utils/flags/flags.go`.
Atm, it seems to be hardcoded here:
https://github.com/kubernetes-sigs/gateway-api/blob/48cf5d9e4756dd65be21fa26e17e3627903c070a/conformance/conformance_test.go#L53-L55

robscott: Yeah we really should have a flag for this. I think that got missed because so many implementations are running these tests by importing them and running them with the rest of their test suite.

aryan9600: Okay, let me create an issue for this then.

shaneutt (approved):
I think this looks solid, and I think it's an interesting addition: it will be interesting to see the usage reporting for this.

I'm approving, but I'll leave the LGTM to @robscott who was previously reviewing.

robscott:
Thanks @aryan9600!

/lgtm
/approve

aryan9600:
I'm trying to fix the issues causing CI to fail. It looks like https://github.com/kubernetes-sigs/gateway-api/pull/1390 modifies v1alpha2 to use v1beta1 definitions as aliases. Thus, running `make update-codegen` with the changes in this PR causes the new field to be added to the v1alpha2 CRD as well, which leads me to believe that `// +k8s:deepcopy-gen=false` on this line isn't actually working as intended. 
https://github.com/kubernetes-sigs/gateway-api/blob/544d916d2be23f9e7038ae90afcce0063863ebcb/apis/v1alpha2/httproute_types.go#L239

I could manually remove it but then this would need to be done every time someone runs `make update-codegen`, so I'm kinda unsure on how to proceed. cc: @howardjohn @robscott

robscott:
Hey @aryan9600, this was a tricky one. From what I can tell, it requires a rebase on the main branch + the following changes:

```diff
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -427,7 +427,7 @@ type GRPCRouteFilter struct {
        // Support: Core
        //
        // +optional
-       RequestHeaderModifier *HTTPRequestHeaderFilter `json:"requestHeaderModifier,omitempty"`
+       RequestHeaderModifier *HTTPHeaderFilter `json:"requestHeaderModifier,omitempty"`

        // RequestMirror defines a schema for a filter that mirrors requests.
        // Requests are sent to the specified destination, but responses from
diff --git a/apis/v1alpha2/httproute_types.go b/apis/v1alpha2/httproute_types.go
index dfbabce..62ff375 100644
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -292,10 +292,10 @@ const (
 // +k8s:deepcopy-gen=false
 type HTTPHeader = v1beta1.HTTPHeader

-// HTTPRequestHeaderFilter defines configuration for the RequestHeaderModifier
-// filter.
+// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP request
+// or response.
 // +k8s:deepcopy-gen=false
-type HTTPRequestHeaderFilter = v1beta1.HTTPRequestHeaderFilter
+type HTTPHeaderFilter = v1beta1.HTTPHeaderFilter
``` 

After these changes, code generation appeared to work well for me locally.

aryan9600:
@robscott i don't think that'll work, since the problem arises due to
https://github.com/kubernetes-sigs/gateway-api/blob/544d916d2be23f9e7038ae90afcce0063863ebcb/apis/v1alpha2/httproute_types.go#L240

since `v1beta1.HTTPRouteFilter` contains the new field `responseHeaderModifier`,  `v1alpha2.HTTPRouteFilter` now also contains this new field, leading to the codegen making it a part of v1alpha2 CRD.
I think we need a way to disable codegen for the v1alpha2 CRD, so that any changes to the codebase from that commit onwards doesn't affect the CRD. wdyt?

robscott:
@aryan9600 I get the confusion now. It's not well documented, but standard practice with k8s APIs is to add all new fields to all supported API versions. In upstream Kubernetes, new fields are hidden with feature gates and are opt in. Since Gateway API is using CRDs, we came up with the parallel concept of an "experimental" release channel that gets new fields first. With that said, whether you're using CRDs from the experimental or standard release channel, both of the API versions contained within them should have an identical set of fields.

aryan9600:
@robscott got it, i was under the impression that we want the new field in the latest API version only. thanks for clarifying :)

robscott:
Thanks @aryan9600!

/lgtm
/approve

Issue #640, opened 2021-04-28
Audit the code base and make sure that our usage of custom vs implementation-specific is accurate AND/OR fix the description in the guidelines.


bowei:
Audit the code base and make sure that our usage of custom vs implementation-specific is accurate AND/OR fix the description in the guidelines.

References:

* https://github.com/kubernetes-sigs/gateway-api/pull/639
* https://github.com/kubernetes-sigs/gateway-api/issues/530

shaneutt:
This requests a quick check of all places we use custom or implementation-specific and make sure that we've documented those correctly. We ideally should have done this before `v0.5.0` (our first beta release) but since that ship has sailed, I've placed this in `v0.6.0` as something we should take care of before the next release.

LCaparelli:
I'd like to work on this. First contrib ever, so I might need some pointers, but I'll ping on #sig-network-gateway-api if I need any help o/

@shaneutt Is there a deadline for v0.6.0 I should be mindful of?

/assign

shaneutt:
> I'd like to work on this. First contrib ever, so I might need some pointers, but I'll ping on #sig-network-gateway-api if I need any help o/
> 
> @shaneutt Is there a deadline for v0.6.0 I should be mindful of?
> 
> /assign

Thanks @LCaparelli! We are currently targeting the week before Kubecon for the release. We appreciate you helping out with this, please let us know here in the issue or on Kubernetes slack if you need any help while working on it :vulcan_salute:

PR #1375, opened 2022-09-04
Add a conformance Test for HTTPS listener with multiple TLS certs

### LINKED ISSUE #1330: Support and test multiple TLS certificates per Gateway listener

We should clarify the API Spec to clarify what happens when multiple TLS Certificates are specified on a Listener. For example, what happens when those certificates overlap or conflict? Once we clarify the spec, we'll also need a conformance test to verify the behavior.

**Why this is needed**:

Attaching multiple TLS certificates to a Gateway listener should have extended support.

ksankeerth:
**What type of PR is this?**
/kind feature

Add a conformance Test for  HTTPS listener with multiple TLS certs

Fixes #1330

NONE

```release-note

```

### DIFF
--- a/conformance/tests/httplistener-multiple-tls-certs-longest-sni-matching.go
+++ b/conformance/tests/httplistener-multiple-tls-certs-longest-sni-matching.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"fmt"
+	"testing"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+	"sigs.k8s.io/gateway-api/conformance/utils/tls"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPSLitenerMultiTLSCertsLongestMatchingSNI)
+}
+
+var HTTPSLitenerMultiTLSCertsLongestMatchingSNI = suite.ConformanceTest{
+	ShortName:   "HTTPSLitenerMultiTLSCertsLongestMatchingSNI",
+	Description: "HTTPS listener references multiple TLS certificates should use longest matching SNI out of all available certificates",
+	Manifests:   []string{"tests/httplistener-multiple-tls-certs-longest-sni-matching.yaml"},
+	Features:    []suite.SupportedFeature{suite.SupportHTTPListenerMultipleTLSCerts},
+	Test: func(t *testing.T, cts *suite.ConformanceTestSuite) {
+
+		testCases := []tls.TestCase{
+			{
+				DialInfo: tls.DialInfo{
+					Host:       "example.com",
+					Port:       "443",
+					ServerName: "abc.xyz.test.example.com",
+				},
+				ExpectedOutput: tls.ExpectedOutput{
+					SANOrCommonName: "*.xyz.test.example.com",
+				},
+			},
+			{
+				DialInfo: tls.DialInfo{
+					Host:       "example.com",
+					Port:       "443",
+					ServerName: "test1.example.com",
+				},
+				ExpectedOutput: tls.ExpectedOutput{
+					SANOrCommonName: "*.example.com",
+				},
+			},
+			{
+				DialInfo: tls.DialInfo{
+					Host:       "example.com",
+					Port:       "443",
+					ServerName: "internal.test.example.com",
+				},
+				ExpectedOutput: tls.ExpectedOutput{
+					SANOrCommonName: "*.test.example.com",
+				},
+			},
+		}
+
+		for i, tc := range testCases {
+			t.Run(fmt.Sprintf("Test case %d", i+1), func(t *testing.T) {
+				tls.InitiateTLSHandShakeAndValidateSNIMatch(t, tc.DialInfo, tc.ExpectedOutput)
+			})
+		}
+	},
+}
--- a/conformance/utils/tls/tls.go
+++ b/conformance/utils/tls/tls.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tls
+
+import (
+	"crypto/tls"
+	"fmt"
+	"testing"
+)
+
+type DialInfo struct {
+	Host       string
+	Port       string
+	ServerName string
+}
+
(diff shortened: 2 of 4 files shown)


conformance/utils/tls/tls.go
```diff
@@ -0,0 +1,65 @@
...
+	defer tlsConn.Close()
+
+	err = tlsConn.Handshake()
+	if err != nil {
+		t.Logf("TLS handshake failed : %v", err)
+	}
+
+	for _, san := range tlsConn.ConnectionState().PeerCertificates[0].DNSNames {
+		if san == expectedOutput.SANOrCommonName {
+			t.Logf("SNI matching passed")
```
tokers, line 56: We can use `t.Log` if we don't have arguments for the logging.

conformance/utils/tls/tls.go
```diff
@@ -0,0 +1,65 @@
...
+			return
+		}
+	}
+	if tlsConn.ConnectionState().PeerCertificates[0].Subject.CommonName == expectedOutput.SANOrCommonName {
+		t.Logf("SNI matching passed")
```
tokers, line 61: Ditto.

conformance/utils/tls/tls.go
```diff
@@ -0,0 +1,65 @@
...
+type ExpectedOutput struct {
+	SANOrCommonName string
+}
+
+type TestCase struct {
+	DialInfo
+	ExpectedOutput
+}
+
+func InitiateTLSHandShakeAndValidateSNIMatch(t *testing.T, tlsDialInfo DialInfo, expectedOutput ExpectedOutput) {
```
tokers, line 40: How will this function detect the failure? Methods like `t.Fail`, `T.Fatal` are not used.

conformance/utils/tls/tls.go
ksankeerth: Thanks for the suggestion. I'll change the code.

conformance/utils/tls/tls.go
ksankeerth: Will change the code here too.

conformance/utils/tls/tls.go
ksankeerth: Sorry, It's my mistake. Thanks for pointing it out. I'll make changes and push the code.

ksankeerth:
@tokers Thanks for your comments. I have fixed them. Will you be able to review the PR?

conformance/utils/tls/tls.go
```diff
@@ -0,0 +1,65 @@
...
+		if san == expectedOutput.SANOrCommonName {
+			t.Log("SNI matching passed")
+			return
+		}
+	}
+	if tlsConn.ConnectionState().PeerCertificates[0].Subject.CommonName == expectedOutput.SANOrCommonName {
+		t.Log("SNI matching passed")
+		return
+	}
+	t.Fail()
```
tokers, line 64: Should we just fail immediately once the failure occurs?

ksankeerth: @tokers  Thanks again. I'm not sure, Do you think we have to try multiple attempts and log the number of failed attempts?

tokers: I think so. Throwing the error just occurred is better than letting the program do more things, which will cause more confusion.

ksankeerth: @tokers  thanks for your time. I'm assuming, You suggest something like this,  

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/http/http.go#L113-L146
Screenshot 2022-09-20 at 09 11 23


There is a timeout, and 1 second delay if a failure occurred. Shall I try to implement similar logic for TLS Handshake?

Just asking for my curiosity, Are there any better ways to test  longest SNI matching 🤔?

howardjohn (commented):
I am a bit confused. Multi certs are `Implementation-specific` per the spec. How can we have a conformance tests for something Implementation-specific?

ksankeerth:
> I am a bit confused. Multi certs are `Implementation-specific` per the spec. How can we have a conformance tests for something Implementation-specific?

@howardjohn Not sure my understanding is correct. But in the document, It says when multiple certs are configured, GW should use the longest matching SNI. (It's under support: core). I think the longest SNI matching falls under support core but more than one secret refs fall under Implementation specific

> 
image

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig

But supporting multiple certs  are under support: Implementation specific

image

conformance/tests/httplistener-multiple-tls-certs-longest-sni-matching.go
```diff
@@ -0,0 +1,76 @@
...
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPSLitenerMultiTLSCertsLongestMatchingSNI)
+}
+
+var HTTPSLitenerMultiTLSCertsLongestMatchingSNI = suite.ConformanceTest{
+	ShortName:   "HTTPSLitenerMultiTLSCertsLongestMatchingSNI",
+	Description: "HTTPS listener references multiple TLS certificates should use longest matching SNI out of all available certificates",
+	Manifests:   []string{"tests/httplistener-multiple-tls-certs-longest-sni-matching.yaml"},
+	Test: func(t *testing.T, cts *suite.ConformanceTestSuite) {
```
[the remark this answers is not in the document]
ksankeerth: Thanks for reviewing and examples. Got it. If any implementations don't support multiple certs,  can ignore the longest SNI matching test.

ksankeerth:
Rebased

conformance/tests/httplistener-multiple-tls-certs-longest-sni-matching.go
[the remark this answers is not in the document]
ksankeerth: Updated

mikemorris:
This feels like an example of https://github.com/kubernetes-sigs/gateway-api/issues/640 - we may need to revisit how support levels are specified here, and if some subset like "more than one Kubernetes Secret" should actually be Extended support with an opt-in conformance test like this.

mikemorris:
@robscott This seems intended to address the conformance tests requested in https://github.com/kubernetes-sigs/gateway-api/issues/851#issuecomment-1201753415, thoughts on whether that GEP already covers the expected conformance sufficiently, or if we need further discussion on what level this should be?

shaneutt:
@ksankeerth it's been a bit since we've heard back from you, are you still able to continue with this PR?

shaneutt:
@ksankeerth we haven't heard back from you in some time, we're going to mark this one as a draft for the time being since there's a few conflicts that need to be resolved. Let us know how we can help you if you're still interested in this one! And if you're not interested that's totally fine, just let us know and we'll unassign!

ksankeerth:
@shaneutt Apologies for the very late reply. I think there is some confusion in multi-tls support as Implementation specific. 

https://github.com/kubernetes-sigs/gateway-api/pull/1526#discussion_r1019755797

>I think this is going to need some more discussion. If we say that supporting multiple TLS certificates is "Extended" support, we'll need to define how multiple TLS certs should be handled. I'm not entirely sure what that looks like today, but it could be worth getting some input from anyone currently supporting that.

>One concrete change here would be to limit the scope of "Extended" support to Kubernetes TLS secrets. That would result in the following:


As per the doc,https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Listener

>TLS is the TLS configuration for the Listener. This field is required if the Protocol field is “HTTPS” or “TLS”. It is invalid to set this field if the Protocol field is “HTTP”, “TCP”, or “UDP”.

>The association of SNIs to Certificate defined in GatewayTLSConfig is defined based on the Hostname field for this listener.

> The GatewayClass MUST use the longest matching SNI out of all available certificates for any TLS handshake.

> Support: Core


If GW has to support the longest SNI matching as core support, that indirectly means GWs have multiple certificates.  However, support for multi-certificates is currently mentioned as  Implementation specific. Later, I raised this PR to mark it as extended support. I believe there is no conclusion on that https://github.com/kubernetes-sigs/gateway-api/pull/1526.  (Sorry If I missed any discussions on this)

howardjohn:
"multiple certificates" feature is for multiple under a single Listener. I think the text you quoted is about having multiple by having N listeners with 1 certificate each

PR #1376, opened 2022-09-05
changed docstring in shared_types.go to account for empty strings being an error

### LINKED ISSUE #1220: ParentReference.Namespace has unnecessary pointer to string.
https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1beta1/shared_types.go#L48

> "Namespace is the namespace of the referent. When unspecified (or empty string), this refers to the local namespace of the Route."

Given that the zero value in Go for a string type is empty string, there's no need for ParentReference.Namespace to be of type `*string`, rather, `string` would have the same interface with a simpler implementation.

asim-reza:
**Issue**
ParentReference.Namespace has unnecessary pointer to string. [#1220](https://github.com/kubernetes-sigs/gateway-api/issues/1220)

**What type of PR is this?**
/kind documentation

In reference to the Issue mentioned above ([#1220](https://github.com/kubernetes-sigs/gateway-api/issues/1220)) it was decided that setting the value as an empty string for the field `namespace` is an error, and this is now reflected in its docstring

Fixes #1220 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -47,8 +47,7 @@ type ParentReference struct {
 	Kind *Kind `json:"kind,omitempty"`
 
 	// Namespace is the namespace of the referent. When unspecified, this refers
-	// to the local namespace of the Route. Refer to the Namespace type for valid
-	// values.
+	// to the local namespace of the Route.
 	//
 	// Support: Core
 	//
--- a/apis/v1alpha2/shared_types.go
+++ b/apis/v1alpha2/shared_types.go
@@ -46,8 +46,9 @@ type ParentReference struct {
 	// +optional
 	Kind *Kind `json:"kind,omitempty"`
 
-	// Namespace is the namespace of the referent. When unspecified (or empty
-	// string), this refers to the local namespace of the Route.
+	// Namespace is the namespace of the referent. When unspecified, this refers
+	// to the local namespace of the Route. Refer to the Namespace type for valid
+	// values.
 	//
 	// Support: Core
 	//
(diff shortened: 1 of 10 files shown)


apis/v1alpha2/shared_types.go
```diff
@@ -46,8 +46,9 @@ type ParentReference struct {
 	// +optional
 	Kind *Kind `json:"kind,omitempty"`
 
-	// Namespace is the namespace of the referent. When unspecified (or empty
-	// string), this refers to the local namespace of the Route.
+	// Namespace is the namespace of the referent. When unspecified, this refers
+	// to the local namespace of the Route. Refer to the Namespace type for valid
+	// values.
```
markmc, line 51:
```suggestion
	// Namespace is the namespace of the referent. When unspecified, this refers
	// to the local namespace of the Route.
```

I think the "Refer to the Namespace type" bit isn't useful because the docs will include a link to that type. If it was useful, we'd have to say it everywhere for every type.

markmc (changes requested):
Thank you for your patch!

Could you also make this change to the v1beta1 docs?

/label ok-to-test

apis/v1alpha2/shared_types.go
asim-reza: Hey! I think I made the change for both `v1alpha2` and `v1beta1`... Unless I pushed it wrong

markmc:
Thanks!

/lgtm

asim-reza:
@keithmattix 
@youngnick

youngnick:
This LGTM, nice work.

Running `make generate` and pushing the result should clear up the CI failures.

Once that's done, I'll approve. Thanks for making your first change!

PR #1379, opened 2022-09-07
Fix PYTHON_VERSION for build image

youngnick:
Looks like netlify previews are failing, I think this should fix it. If this one has good previews, that should do it.

**What type of PR is this?**
/kind bug
/kind cleanup
/kind regression

Fixes #

```release-note
NONE
```


shaneutt (approved):
Preview is working :+1: 

/lgtm

PR #1382, opened 2022-09-07
replace Attached condition with Accepted in HTTPRoute docs

skriss:
The HTTPRoute docs have several references to
an "Attached" condition, however this condition
does not exist for routes. Replaces these
references with the appropriate "Accepted"
condition.

**What type of PR is this?**
/kind documentation

The HTTPRoute docs have several references to
an "Attached" condition, however this condition
does not exist for routes. Replaces these
references with the appropriate "Accepted"
condition.

See https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteConditionType for the list of valid Route condition types and reasons.

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/httproute_types.go
+++ b/apis/v1alpha2/httproute_types.go
@@ -65,7 +65,7 @@ type HTTPRouteRule = v1beta1.HTTPRouteRule
 // must ensure that unknown values will not cause a crash.
 //
 // Unknown values here must result in the implementation setting the
-// Attached Condition for the Route to `status: False`, with a
+// Accepted Condition for the Route to `status: False`, with a
 // Reason of `UnsupportedValue`.
 //
 // +kubebuilder:validation:Enum=Exact;PathPrefix;RegularExpression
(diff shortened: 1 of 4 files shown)


shaneutt (approved):
So from what I can tell, this appears to be right: we've had this wrong for some time as the `Accepted` condition exists for `HTTPRoute` but not `Attached` (which is only a `Gateway` listener condition elsewhere). Approved, but I would like another maintainer to just double check me and give the LGTM because I can't shake the feeling like I might be missing something :thinking:

youngnick:
This change is one hundred percent correct as per the current spec, but I'm currently working on a status rewrite which proposes we move to `Attached` everywhere. So maybe this should wait until I finish that up and we discuss it?

mikemorris:
This content likely needs to be revisited now that #1383 has merged.

skriss:
So this was just trying to fix a bug in the v0.5.0 spec/docs, but I guess at this point it's probably missed the boat for inclusion in v0.5.1? If so, and the broader v0.6.0 conditions changes will cover this, then we can just close this one out and live with the v0.5.x docs bug.

skriss:
Rebased in case we want to include it in v0.5.1, if not, and will be covered by larger v0.6 changes, feel free to close.

robscott:
Thanks @skriss! If you follow up with a cherry-pick PR into release-0.5, I'll add it to the v0.5.1 changelog.

/lgtm
/approve

Issue #1364, opened 2022-08-26
GEP: Conditions and Status update


youngnick:
We currently have a lot of outstanding issues related to status, mostly around Conditions, whether it be issues with specific Conditions being hard to understand, or general issues around Conditions. It's just because a lot of the status and Conditions design has grown very organically, what we effectively need is a design pass to make sure everything is consistent. There have been a few looks at this previously, but nothing coordinated across the whole API.

This issue covers a review of all the outstanding status issues, figuring out which ones relate, generating a TODO list, and then writing a GEP to cover all the changes.

Based on my initial quick pass, I think it's likely that this will include changes that will break conformance, but I think the longer we leave this, the worse the changes will be.

I propose making this a blocker for v0.6.0 release, so that we can get the pain out of the way now, and move forward with a consistent view of how our status flows work.

Here's my initial notes from doing a quick issue roundup.

### Rough initial notes to link in relevant issues

In general, we need to make sure we're reusing the same Condition types across different resources, and that they have the same or similar definitions across resources as well.

The biggest review to date was #1111, and there are still some outstanding items from that.

GatewayClass has `Accepted`. This seems well-defined.

Gateway has `Ready`, which has significant issues, see #1156. This probably needs a follow-up to describe the change.

I think we definitely need to have a better description of what `Ready` means, and ensure that the definitions are consistent. @evankanderson has raised that it would be really useful for knative's use case for `Ready` to mean "Ready to pass traffic as soon as it's set", but I don't know the extent to which we can mandate that. We may be stuck with setting a "`Ready` means will be ready in <5sec" or something, but that will require implementation feedback.

We need to move the Listener condition to `Attached` instead of `Detached`, as described in #1110. Having a small set of standard positive-polarity (that is, they are `status: true` if everything is good) Conditions is useful, and similar things should have similar names.

#1112 also needs to be updated post the closing of #1211.

We also need to note if conditions should be added if they are in a happy state. The latest comments on #1315 is talking about this.
Personally, I prefer that conditions are always added, even if they are `Unknown`, but this one needs discussion.

#1362 should probably also be done as part of this.

robscott:
I was initially going to create a new issue for this, but will start here for now since I know this is meant to be a bit of an umbrella issue. 

Closely related to your comments about a "Ready" condition on HTTPRoute, I think we really need something to fill the gap in our current conformance tests here. Currently we're limited to an "Accepted" condition with a rather vague definition:

[...]

PR #1383, opened 2022-09-09
Add GEP-1364 proposal

youngnick:
**What type of PR is this?**
/kind gep

This PR adds an implementable proposal for GEP1364 (#1364).

It's a big GEP, for a _very_ big change, that will definitely break conformance tests. But I think that it will set us up with clear rules for Conditions going forward.

It's quite opinionated, so I expect to need to defend my decisions here somewhat, don't be shy about objections.

Updates #1364 

```release-note
Status definitions have been updated across the API.
```
The release note will need something like the tl;dr in the GEP, once we all agree. 😄

### DIFF
--- a/site-src/geps/gep-1364.md
+++ b/site-src/geps/gep-1364.md
@@ -5,7 +5,7 @@
 
 (See definitions in [Kubernetes KEP][kep-status].
 
-[kep-status]: https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml#L9
+
 
 ## TLDR
 
@@ -47,14 +47,6 @@ This level of change is not optimal, and the intent is to make this a one-off ch
 that can be built upon for future resources - since there are definitely more resources
 on the way.
 
-TODO: Further outline some of the changes that have been required already due to
-unclear definitions in the spec, particularly around conformance. (needs some
-issue archaeology.)
-
-[1111]: https://github.com/kubernetes-sigs/gateway-api/issues/1111
-[1110]: https://github.com/kubernetes-sigs/gateway-api/issues/1110
-[1362]: https://github.com/kubernetes-sigs/gateway-api/issues/1362
-
 ## Background: Kubernetes API conventions and prior art on Conditions
 
 Because this GEP is mainly concerned with updating the Conditions we are setting in
@@ -64,33 +56,36 @@ section of the API conventions document.)
 
 * Conditions are a standard type used to represent arbitrary higher-level status from
 a controller.
-* They are a listMapType, a list that is enforced by the apiserver to have only one entry
-of each item, using the `type` field as a key. (So, this is effectively a map that looks like
-a list in YAML form).
-* Each has a number of fields, the most important of which for this discussion are `type`, `status`,
-`season`, and `observedGeneration`.
-  * `type` is a string value indicating the Condition type. `Accepted`, `Scheduled`, and `Ready` are
-  current examples.
-  * `status` indicates the state of the condition, and can be one of three values, `true`, `false`,
-  or `unknown`. Unkown in particular is important, because it means that the controller is unable
-  to determine the status for some reason.
-  * `reason` is a CamelCase string that is a brief description of the reason why the `status` is set
-  the way it is.
-  * `observedGeneration` is an optional field that sets what the `metadata.generation` field was
-  when the controller last saw a resource.
-* Conditions shoud describe the _current state_ of the resource at observation time, which means
-that they should be an adjective (like `Ready`), or a past-tense verb (like `Attached`).
-* Conditions should be applied to a resource the first time the controller sees the resource.
-This seems to imply that _all conditions should be present on every resource owned
-by a controller_, but the rest of the conventions don't make this clear, and it
-is often not complied with.
-* It's helpful to have a top-level condition which summarizes more detailed conditions. The guidelines
-suggest using either `Ready` for long-running processes, or `Succeeded` for bounded exection.
+* They are a listMapType, a list that is enforced by the apiserver to have only
+one entry of each item, using the `type` field as a key. (So, this is effectively
+a map that looks like a list in YAML form).
+* Each has a number of fields, the most important of which for this discussion
+are `type`, `status`, `season`, and `observedGeneration`.
+  * `type` is a string value indicating the Condition type. `Accepted`, `Scheduled`,
+  and `Ready` are current examples.
+  * `status` indicates the state of the condition, and can be one of three values,
+  `true`, `false`, or `unknown`. Unkown in particular is important, because it
+  means that the controller is unable to determine the status for some reason.
+  * `reason` is a CamelCase string that is a brief description of the reason why
+  the `status` is set the way it is.
+  * `observedGeneration` is an optional field that sets what the `metadata.generation`
+  field was when the controller last saw a resource.
+* Conditions shoud describe the _current state_ of the resource at observation
+time, which means that they should be an adjective (like `Ready`), or a past-tense
+verb (like `Attached`).
+* Conditions should be applied to a resource the first time the controller sees
+the resource. This seems to imply that _all conditions should be present on every
+resource owned by a controller_, but the rest of the conventions don't make this
+clear, and it is often not complied with.
+* It's helpful to have a top-level condition which summarizes more detailed conditions.
+The guidelines suggest using either `Ready` for long-running processes, or `Succeeded`
+for bounded exection.
 
-From these guidelines, we can see that Conditions can be either _positive polarity_ (healthy resources
-have them as `status: true`) or _negative polarity_ (healthy resources have them as `status: false`).
-`Ready` is an example of a positive polarity condition, and conditions like `Conflicted` from Listener
-or `NetworkUnavailable`, `MemoryPressure`, or `DiskPressure` from the Node resource are examples of
+From these guidelines, we can see that Conditions can be either _positive polarity_
+(healthy resources have them as `status: true`) or _negative polarity_ (healthy
+resources have them as `status: false`). `Ready` is an example of a positive polarity
+condition, and conditions like `Conflicted` from Listener or `NetworkUnavailable`,
+`MemoryPressure`, or `DiskPressure` from the Node resource are examples of
 negative-polarity conditions.
 
 There is also some extra context that's not in the API conventions doc:
@@ -108,8 +103,6 @@ relevant conditions should always be added, even if their status has to be set t
 `unknown`.
 Gateway API resources do not currently require this, and the practice seems to be
 uncommon.
-
-[typstatus]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
 
 ## Proposed changes
 
@@ -272,8 +265,7 @@ is otherwise okay, then the data plane must generate 500s for traffic that match
 that HTTPRoute.
 * HTTPRoutes with *all* backends in other namespaces, but not permitted by ReferenceGrants.
 In this case, the "nonexistent service backends" rules apply, and 500s must be
-generated. TODO(youngnick) check that this is what the conformance tests actually
-test atm.
+generated.
 
 For ReferenceGrant or not-designed-yet Policy resources, `Attached` means that:
 * the resource has a correctly-defined set of resources that it applies to
@@ -292,17 +284,45 @@ actively closes connections, or if no port should be opened.)
 
 #### Ready
 
-TODO: The Ready condition has unclear semantics.
-It's not clear if it should mean "ready when you see this" or "will be ready soon".
+Currently, the `Ready` condition text for Gateway says:
+```go
+	// This condition is true when the Gateway is expected to be able
+	// to serve traffic. Note that this does not indicate that the
+	// Gateway configuration is current or even complete (e.g. the
+	// controller may still not have reconciled the latest version,
+	// or some parts of the configuration could be missing).
+```
+
+This is pretty unclear - how can the Gateway serve traffic if config is missing?
+In the past, we've been asked to have a Condition that only flips to `true` when
+*all* required configuration is present.
+
+For many implementations (certainly for Envoy-based ones), getting this information
+correctly and avoiding races on applying it is surprisingly difficult. For this
+reason, this GEP proposes we give the `Ready` condition a short window
+
+This condition needs a few changes:
+
+* Each resource should have a `Ready` condition
+* The `Ready` condition should be clear as to *when* the traffic is able to pass
+  the implementation. Since it's _extremely_ difficult to _guarantee_ this, we're
+  stuck with some sort of "will be ready in no longer than x seconds" standard
+  (which we'll also need to write conformance tests for, probably including a 
+  stress test of adding a *lot* of records as well.) As an initial ballpark,
+  I propose 5 seconds.
 
 
 
 ## Alternatives
 
-(List other design alternatives and why we did not go in that
-direction)
+(Most alternatives have been discussed inline. Please comment here if this section
+needs updating.)
 
 ## References
+[kep-status]: https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml#L9
 
-(Add any additional document links. Again, we should try to avoid
-too much content not in version control to avoid broken links)
+[1111]: https://github.com/kubernetes-sigs/gateway-api/issues/1111
+[1110]: https://github.com/kubernetes-sigs/gateway-api/issues/1110
+[1362]: https://github.com/kubernetes-sigs/gateway-api/issues/1362
+
+[typstatus]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
--- a/site-src/geps/gep-1364.md
+++ b/site-src/geps/gep-1364.md
@@ -44,6 +44,13 @@ in what constitutes a healthy or unhealthy resource, and there will be changes
 required for all implementations to be conformant with the release that includes
 these changes.
 
+The constants that mark the deprecated types will be also marked as deprecated,
+and will no longer be tested as part of conformance. They'll still be present,
+and will work, but they won't be part of the spec any more. This should give
+implementations and users a release to transition to the new design (in UX terms).
+This grace period should be one release (so, the constants will be removed in
+v0.7.0.)
+
 This level of change is not optimal, and the intent is to make this a one-off change
 that can be built upon for future resources - since there are definitely more resources
 on the way.
@@ -75,7 +82,8 @@ are `type`, `status`, `reason`, and `observedGeneration`.
 
 4. Conditions shoud describe the _current state_ of the resource at observation
 time, which means that they should be an adjective (like `Ready`), or a past-tense
-verb (like `Attached`).
+verb (like `Attached`). This one in particular is documented pretty closely in the
+[Typical status properties][typstatus] section of the guidelines.
 5. Conditions should be applied to a resource the first time the controller sees
 the resource. This seems to imply that _all conditions should be present on every
 resource owned by a controller_, but the rest of the conventions don't make this
@@ -123,9 +131,19 @@ when it's observed.
 For example, HTTPRoutes must always have `Accepted` and `ResolvedRefs`, regardless
 of their state.
 * Negative polarity error conditions must only be added when the error is True.
-
-(For discussion): Should the summary conditions be added into the CRD definitions
-as default values for `status.Conditions`, with a status of `Unknown`?
+* The `Ready` condition will be moved to Extended conformance, and we'll reevaluate
+if it's used by any implementations after some time has passed. If not, it may be
+removed.
+* To capture the behavior that `Ready` currently captures, `Programmed` will be
+introduced. This means that the implementation has seen the config, has everything
+it needs, parsed it, and sent configuration off to the data plane. The configuration
+should be available "soon". We'll leave "soon" undefined for now.
+* The summary conditions will be added into the CRD definitions
+as default values for `status.Conditions`, with a status of `Unknown`.
+* Resolving a comment that came up, documentation will be added to clarify that
+it's okay to add your own Conditions, and that implementations should namespace
+their custom Conditions with a domain prefix (so `implementation.io/CustomType`
+rather than just `CustomType`), or run the risk of using a word that's reserved later.
 
 The exact list of changes is detailed below. The next few sections detail
 the reasons for these large-scale changes.
@@ -221,7 +239,9 @@ need to assume that any others are negative.
 #### Decision
 
 Gateway API conditions will be positive for conditions that describe the happy
-state of the object, which is currently `Attached` and `ResolvedRefs`.
+state of the object, which is currently `Attached` and `ResolvedRefs`, and will 
+also include the new `Programmed` condition, and the newly-Extended condition
+`Ready`.
 
 
 ### Should conditions always be added?
@@ -229,9 +249,8 @@ state of the object, which is currently `Attached` and `ResolvedRefs`.
 Not all of them.
 
 Positive polarity Conditions that describe the desirable state of the object must
-always be set. These are currently `Attached`, `ResolvedRefs`, and whatever we
-end up choosing to represent "the data path will be done soon", `Ready`, `Programmed`,
-or other.
+always be set. These are currently `Attached`, `ResolvedRefs`, and `Programmed`.
+Implementations that use `Ready` must also add it immediately, set to `false`.
 
 ### New and Updated Conditions
 
@@ -245,7 +264,7 @@ attach to anything).
 That is, the proposal is to replace:
 * `Scheduled` on Gateway
 * `Detached` on Listener
-* `Accpeted` on Route
+* `Accepted` on Route
 
 with `Attached` in all these locations.
 
@@ -360,6 +379,7 @@ tests to verify the behvaior.
 * When it's set, the condition indicates that traffic is ready to flow through
 the data plane _immediately_, not at some eventual point in the future.
 
+We'll need to add conformance testing for this.
 
 #### Programmed
 
--- a/site-src/geps/gep-1364.md
+++ b/site-src/geps/gep-1364.md
@@ -114,12 +114,18 @@ uncommon.
 * All the current Conditions that indicate that the resource is okay and ready
 for processing witll be replaced with `Attached`, except for GatewayClass (since
 it is the root of the resource tree, it will stay with `Accepted`).
-* In general, resources should be considered `Attached` if their config will
-generate some config in the underlying data plane.
-* All Conditions will be positive polarity (`GoodState: true`, not `BadState: false`)
-* All relevant Conditions for a resource must be added when it's observed.
-For example, HTTPRoutes must always have `Accepted`, `Ready`, and `ResolvedRefs`
-conditions, regardless of their state.
+* In general, resources should be considered `Attached` if their config is valid
+enough to generate some config in the underlying data plane.
+* There will be a limited set of positive polarity summary conditions, and a number
+of other specific negative-polarity error conditions.
+* All relevant positive-polarity summary Conditions for a resource must be added
+when it's observed.
+For example, HTTPRoutes must always have `Accepted` and `ResolvedRefs`, regardless
+of their state.
+* Negative polarity error conditions must only be added when the error is True.
+
+(For discussion): Should the summary conditions be added into the CRD definitions
+as default values for `status.Conditions`, with a status of `Unknown`?
 
 The exact list of changes is detailed below. The next few sections detail
 the reasons for these large-scale changes.
@@ -146,8 +152,13 @@ Of course, because we're using all of this configuration to describe some sort o
 path from "outside"/lacking cluster context to "inside"/enriched with cluster context,
 we also need a way to describe when that data path is configured and working.
 
-Again, we already have a word we use for this - "Ready", so this GEP proposes keeping
-that and tightening up the definition.
+We already have a word in the Kubernetes API, but it comes with some expectations
+that implementations are not currently able to meet. That word is `Ready`, but it
+implies that the data path is Ready _when you read the status_, rather than that
+it _will be ready soon_ (which is what most implementations can guarantee currently.)
+
+So we have an unresolved question as to what to do with the `Ready` condition.
+This is addressed further below.
 
 ### Condition polarity
 
@@ -209,19 +220,18 @@ need to assume that any others are negative.
 
 #### Decision
 
-Gateway API resource conditions will all be positive polarity.
+Gateway API conditions will be positive for conditions that describe the happy
+state of the object, which is currently `Attached` and `ResolvedRefs`.
+
 
 ### Should conditions always be added?
 
-Since we are mandating that conditions should always be positive, we should also
-specify that controllers always place all conditions on each resource when they
-reconcile it, even if that update sets the status to `unknown` because of
-information not being available.
+Not all of them.
 
-This is becasue, whatever the API guidelines say, the absence of a positive
-condition means that a user can't be _sure_ that the condition is okay. So, we
-should ensure that all the positive conditions are always present, even if their
-status is `unknown`.
+Positive polarity Conditions that describe the desirable state of the object must
+always be set. These are currently `Attached`, `ResolvedRefs`, and whatever we
+end up choosing to represent "the data path will be done soon", `Ready`, `Programmed`,
+or other.
 
 ### New and Updated Conditions
 
@@ -254,7 +264,8 @@ is no missing information, including but not limited to:
   * Features being used by the resource are supported
 
 All of these rules can be summarized into:
-* The resource will produce some configuration in the underlying data plane.
+* The resource is valid enough to produce some configuration in the underlying
+data plane.
 
 For Gateway, `Attached` also subsumes the functions of `Scheduled`: `Attached`
 set to `true` means that sufficient capacity exists on underlying infrastructure
@@ -266,10 +277,14 @@ Note that some classes of inter-resource reference failure do _not_ cause a reso
 to become unattached (that is, to have the `Attached` condition set to `status: false`).
 * Nonexistent Service backends - if the backend does not exist on a HTTPRoute that
 is otherwise okay, then the data plane must generate 500s for traffic that matches
-that HTTPRoute.
+that HTTPRoute. In this case, the `Attached` Condition must be true, and the
+`ResolvedRefs` Condition must be false, with reasons and messages indicating that
+the backend services do not exist.
 * HTTPRoutes with *all* backends in other namespaces, but not permitted by ReferenceGrants.
 In this case, the "nonexistent service backends" rules apply, and 500s must be
-generated.
+generated. In this case, again, the `Attached` condition is true, and the
+`ResolvedRefs` Condition is false, with reasons and messages indicating that the
+backend services are not reachable.
 
 For ReferenceGrant or not-designed-yet Policy resources, `Attached` means that:
 * the resource has a correctly-defined set of resources that it applies to
@@ -284,6 +299,37 @@ backends may not produce any configuration in the data plane, and so may cause
 the resource to fail to attach. (An example here could be a TCP Route with
 no backends, we need to decide if that means that a port should be opened that
 actively closes connections, or if no port should be opened.)
+
+Examples of Conditions:
+* HTTPRoute with one match with one backend that is valid. `Attached` is true,
+`ResolvedRefs` is true.
+* HTTPRoute with one match with one backend that is a nonexistent Service backend.
+The `Attached` Condition is true, the `ResolvedRefs` condition is false. `Attached`
+is true in this case because the data path must respond to requests that would be
+sent to that backend with a 500 response.
+* HTTPRoute with one match with two backends, one of which is a nonexistent Service
+backend. The `Attached` Condition is true, the `ResolvedRefs` condition is false.
+`Attached` is true in this case because the data path must respond to a percentage
+of the requests matching the rule corresponding to the weighting of the nonexistent
+backend (which would be fifty percent unless weights are applied).
+* HTTPRoute with one match with one backend that is in a different namespace, and
+does _not_ have a ReferenceGrant permitting that access. The `Attached` condition
+is true, and the `ResolvedRefs` Condition is false. As before, `Attached` is true
+because in this case, the data path must be programmed with 500s for the match.
+* TCPRoute with one match with a backend that is a nonexistent Service. `Attached`
+is false, and `ResolvedRefs` is false. `Attached` is false in this case because
+there is not enough information to program any rules to handle the traffic in the
+underlying data plane - TCP doesn't have a way to say "this is a valid destination
+that has something wrong with it".
+* HTTPRoute with one Custom supported filter added that is not supported by the
+implementation. Our spec is currently unclear on what happens in this case, but
+custom HTTP Filters require the use of the `ExtensionRef` filter type, and the
+setting of the ExtensionRef field to the name, group, version, and kind of a
+custom resource that describes the filter. If that custom resource is not supported,
+it seems reasonable to say that this should be a reference failure, and be treated
+like other reference failures (`Attached` will be set to true, `ResolvedRefs` to
+false, and traffic that would have matched the filter should receive a 500 error.)
+
 
 
 #### Ready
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+GatewayClass will maintain the `Accepted` condition.
+
+All of these conditions share the following meanings:
+* The resource has been accepted for processing by the controller
+* The resource is syntatically and semantically valid, and internally consistent
+* The resource fits into a larger system of Gateway API resources, and there is
+is no missing information, including but not limited to:
+  * Any mandatory references resolve to existing resources (examples here are the
+  Gateway's gatewayClass field, or the `parentRefs` field in Route resources)
+  * Any specified TLS secrets exist
```
howardjohn, line 247: Currently the API supports partially-valid resources. I believe this is proposing a partially-valid resource would be `Attached=false`, but the config is actually 'active' (or, I would describe it, "attached").

To me this feels a bit odd. The old API sort of had this issue as well.

Thinking purely from a pedantic-correctness POV (which may not be ideal - sometimes simpler is better), I would expect there to be two conditions (I use bad names intentionally to avoid confusion): "ThisResourceWasApplied" - any part of the rule had some effect on the system, and "ThisRuleIsEntirelyValid" - no errors found in the rule at all.

Its possible my concern would be addressed with a different word than "Attached", rather than more conditions, as well

WDYT?

howardjohn: I guess this is somewhat addressed below.. but I am not sure its accurate? It seems we are defining a very small set of "failures that do not become unattached", but there are a lot more partially-accepted resources I think?

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+* Status: Implementable
+
+(See definitions in [Kubernetes KEP][kep-status].
+
+
+
+## TLDR
+
+The status, particularly the Conditions, across the whole Gateway API has very much grown
+grown organically, and so has many inconsistencies and odd behaviors.
```
shaneutt, line 13:
```suggestion
The status, particularly the Conditions, across the whole Gateway API has very much grown
organically, and so has many inconsistencies and odd behaviors.
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+This condition needs a few changes:
+
+* Each resource should have a `Ready` condition
+* The `Ready` condition should be clear as to *when* the traffic is able to pass
+  the implementation. Since it's _extremely_ difficult to _guarantee_ this, we're
+  stuck with some sort of "will be ready in no longer than x seconds" standard
+  (which we'll also need to write conformance tests for, probably including a 
+  stress test of adding a *lot* of records as well.) As an initial ballpark,
+  I propose 5 seconds.
```
howardjohn, line 312: This probably won't work for cloud providers (at least GKE is slower, not sure about others).

I am also not sure it provides value. What do we lose if we don't define anything here?

If we consider something like Knative, they certainly aren't going to be able to just wait X seconds; I'd expect the same with other users

howardjohn: And if we can't provide any guarantees about Readiness, then it seems like Ready becomes no different than Attached...

howardjohn (commented):
Definitely a good direction

site-src/geps/gep-1364.md
youngnick: The guideline I've been basing the "does something become unattached?" on is "Does this produce at least some configuration in the underlying data plane?". So I guess it sounds like I need to be more explicit with that guideline, but that aside from that, we're largely in agreement?

site-src/geps/gep-1364.md
youngnick: We certainly can mandate `Ready` to be an _extended_ condition, and say "If you supply a Ready condition, it has to work like this". Maybe doing that is a better place to start with `Ready`?

howardjohn: Do we have a feel for which implementations can and cannot support `Ready` properly?

site-src/geps/gep-1364.md
howardjohn: I think it makes sense to have "Is this attached" and "is this 100% valid" as two separate things. 

Part of my confusion, I think, is I am not actually sure which fields result in "not attached". It seems _most_ issues are partially accepted. Which is fine - `Attached` just becomes basically a pretty high level acknowledgement that the controller is handling the resource

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+This level of change is not optimal, and the intent is to make this a one-off change
+that can be built upon for future resources - since there are definitely more resources
+on the way.
+
+## Background: Kubernetes API conventions and prior art on Conditions
+
+Because this GEP is mainly concerned with updating the Conditions we are setting in
+Gateway API resources' `status`, it's worth reviewing some important points about
+Conditions. (Thi information is mainly taken from the [Typical status properties][typstatus]
```
keithmattix, line 54: Missing href for the link here?

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+section of the API conventions document.)
+
+* Conditions are a standard type used to represent arbitrary higher-level status from
+a controller.
+* They are a listMapType, a list that is enforced by the apiserver to have only
+one entry of each item, using the `type` field as a key. (So, this is effectively
+a map that looks like a list in YAML form).
+* Each has a number of fields, the most important of which for this discussion
+are `type`, `status`, `season`, and `observedGeneration`.
```
keithmattix, line 63:
```suggestion
are `type`, `status`, `reason`, and `observedGeneration`.
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Proposed changes
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
+it is the root of the resource tree, it will stay with `Accepted`).
+* In general, resources should be considered `Attached` if their config will
+generate some config in the underlying data plane.
```
keithmattix, line 115: It might be helpful to be explicit here:
```suggestion
generate some config in the underlying data plane (whether that config is valid or not).
```

site-src/geps/gep-1364.md
keithmattix: Yeah I think I agree with @howardjohn's point here: "does this resource produce config" is distinct from "does this resource produce _valid_ config". I'm sure some implementations don't allow invalid (e.g. duplicate route names or something) to be sent to the data plane, but it doesn't feel like that's a requirement for the Gateway API at this juncture.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+For Gateway, `Attached` also subsumes the functions of `Scheduled`: `Attached`
+set to `true` means that sufficient capacity exists on underlying infrastructure
+for the Gateway to be provisioned. If that capacity does not exist, then the
+Gateway cannot be reconciled successfully, and so fails to attach to the
+owning GatewayClass.
+
+Note that some classes of inter-resource reference failure do _not_ cause a resource
+to become unattached (that is, to have the `Attached` condition set to `status: false`).
+* Nonexistent Service backends - if the backend does not exist on a HTTPRoute that
```
keithmattix, line 263: This starts painting an observability problem IMO.  Even if conditions are only meant to be human-readable, how does a human look at an HTTPRoute resource and know "oops I have a typo in my service backend"? Yes, 500s indicate something's wrong, but I would expect to find a condition pointing to the specific issue

site-src/geps/gep-1364.md
robscott: Tiny nit: has -> have

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Goals
+
+* Update Conditions design to be consistent across Gateway API resources
+* Provide a model and guidelines for Conditions for future new resources
+* Specify changes to conformance required for Condition updates
+
+## Non-Goals
+
+
```
robscott, line 25:
```suggestion
* Define the full set of Conditions that will ever be used with Gateway API
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Introduction
+
+Gateway API currently has a lot of issues related to status, especially that
+status is inconsistent ([#1111][1111]), that names are hard to understand ([#1110][1110]),
+Reasons aren't explained properly ([#1362][1362]), and so on.
```
robscott, line 31: Tiny nit (feel free to ignore):
```suggestion
and that Reasons aren't explained properly ([#1362][1362]).
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+So, for example,  we have GatewayClass with `Accepted`, Gateway with `Scheduled`,
+the Gateway Listeners with `Detached` (which you want to be `false`, unlike the previous
+two), and then Gateways and Gateway Listeners have `Ready`, but Route doesn't (and which
+also you want to be `true`).
+
+This document lays out large-scale changes to the way that we talk about resources,
+and the Conditions to match them. This means that there will be an unavoidable break
+in what constitutes a healthy or unhealthy resource, and there will be changes
+required for all conformant implementations to stay conformant.
```
robscott, line 44: Maybe specifically mention the concept of being conformant with a specific release of the API?
```suggestion
required for all implementations to be conformant with the release that includes these changes.
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+  * `status` indicates the state of the condition, and can be one of three values,
+  `true`, `false`, or `unknown`. Unkown in particular is important, because it
+  means that the controller is unable to determine the status for some reason.
+  * `reason` is a CamelCase string that is a brief description of the reason why
+  the `status` is set the way it is.
+  * `observedGeneration` is an optional field that sets what the `metadata.generation`
+  field was when the controller last saw a resource.
+* Conditions shoud describe the _current state_ of the resource at observation
+time, which means that they should be an adjective (like `Ready`), or a past-tense
+verb (like `Attached`).
```
robscott, line 75: Would `Attaching` ever be valid as a condition? Maybe just verb or adjective is sufficient here?

site-src/geps/gep-1364.md
robscott, line 54:
```suggestion
Conditions. (This information is mainly taken from the [Typical status properties][typstatus]
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+Again, we already have a word we use for this - "Ready", so this GEP proposes keeping
+that and tightening up the definition.
+
+### Condition polarity
+
+In terms of the polarity of conditions, we have three options, of which only two are
+really viable:
+* All conditions must be negative polarity
+* All conditions must be positive polarity
+* Some conditions can be positive polarity, but most should be negative.
```
robscott, line 155: My own personal preference would be for conditions with a positive polarity to represent the normal healthy state and always be present, and for conditions with a negative polarity to represent very specific error states and to only be present when true. I think we should do everything we can to avoid the presence of double negatives like "Detached == False", but recognize that there are some error states that are best represented with a negative condition. I just don't think those negative conditions should be present when false.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+In this case, only a limited set of summary conditions are positive, and the rest
+are negative.
+
+Pros:
+* Error states can be described with `Error: true` instead of `NoError: false`.
+* Negative polarity error conditions are more friendly to not being present (since
+absence of `Error: true` implies everything's okay).
+
+Cons:
+* Double negatives in the happy case: `NoError: false`
```
robscott, line 204: This seems like a con of "All Conditions Positive" but not sure it applies when we allow both positive and negative conditions?

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+#### Decision
+
+Gateway API resource conditions will all be positive polarity.
+
+### Should conditions always be added?
+
+Since we are mandating that conditions should always be positive, we should also
+specify that controllers always place all conditions on each resource when they
+reconcile it, even if that update sets the status to `unknown` because of
+information not being available.
```
robscott, line 215: I think this approach would lead to a _very_ noisy conditions list. I'd personally prefer leaving room for negative/error conditions to be unset unless they were true.

site-src/geps/gep-1364.md
robscott: Agree that cloud providers will likely have trouble implementing a `Ready` condition. Could we introduce a `Reconciled` condition that would be more broadly implementable?

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+	// or some parts of the configuration could be missing).
+```
+
+This is pretty unclear - how can the Gateway serve traffic if config is missing?
+In the past, we've been asked to have a Condition that only flips to `true` when
+*all* required configuration is present.
+
+For many implementations (certainly for Envoy-based ones), getting this information
+correctly and avoiding races on applying it is surprisingly difficult. For this
+reason, this GEP proposes we give the `Ready` condition a short window
```
robscott, line 302: I think I'd rather reserve `Ready` for "Ready at this moment" and instead use a different term to represent what we're actually describing - this config has been written to the data plane but we're not sure how long that will take to propagate. Previously `Reconciled` has been proposed to communicate this.

robscott (commented):
Thanks for all the work on this @youngnick! This is a great start

site-src/geps/gep-1364.md
bowei: We should allow room for error conditions in addition to attached in this case so the intformation can be conveyed.

site-src/geps/gep-1364.md
bowei: +1 any distributed (e.g. geo distributed) implementation has a hard time with this (think: config was pushed to 95% of the sites around the world, is it ready?)

youngnick:
Okay, to summarize the comments around condition naming and polarity:

We're going to end up with four possible positive polarity summary conditions:
- `ValidSpec` or similar; there aren't any issues in the config of _this resource on its own_
- `Attached`; there aren't any issues with the way this resources _attaches to others_
- `Programmed` or similar; the resource has been processed and the data plane is in the process of being configured. It should be ready "soon", where "soon" is implementation-specific.
- `Ready`; an optional, *Extended* condition that is only added when the implementation can guarantee that all required configuration has been accepted by the data plane and is running. Because it's Extended, it will have conformance tests to guarantee this behavior.

Other error conditions may be added that are negative polarity - the only thing relevant about these Conditions is if they are `status: true`. If they are not `status: true`, they should not be present.

Three or four summary, positive-polarity conditions seems on the high side to me, but if everyone feels that the additional clarity is worth the overhead of managing more Conditions.

I'm kind of inclined to say that anything we would use a `ValidSpec` summary condition for should probably be in the validation rules - whether they are in the OpenAPI definitions or the validating webhook doesn't really matter. I don't think the cases in which `ValidSpec` is True and `Attached` is false are going to be very common, so I don't really see the utility of that condition as much, personally. But if we really want it, we can do it.

site-src/geps/gep-1364.md
youngnick: I guess I wrote this pooly - by "Does this resource produce config" I meant to read "Is this resource valid enough to produce some config when combined with the rest of the resource set".

The thing I'm thinking of particularly here is that for HTTPRoute, invalid references still produce data path config because the match will respond with 500s. Whereas for TCPRoute etc, if you don't have a service to route to, there's no way you can build a forwarding path - there's nothing to forward to! So that won't produce any data path config at all, and so, in the current design, would cause the resource to fail to be attached.

I think that partial validity is an enormous minefield of weirdness, just like not applying config if it would break things.

site-src/geps/gep-1364.md
youngnick: It's all the way down at the bottom of the doc, with the other references. I thought it was better to have everything in the one place.

site-src/geps/gep-1364.md
youngnick: That sort of usage is specifically called out as not desirable in the API conventions, and I agree. Conditions are supposed to be level-triggered, not edge-triggered, so having a condition that says that something is happening is a bit weird. (Should `Attaching` transition to `false` once the attaching process is complete?)

site-src/geps/gep-1364.md
youngnick: I've been assuming that only valid config will be passed to the data plane. If a resource would generate invalid config, then to my mind it shouldn't be considered attached.

Again, here I'm thinking of HTTPRoute adding 500s for things that would otherwise cause a resource to become unattached (that is, it doesn't  match anything, or it doesn't have a valid `backendRef`, or there are zero endpoints in a valid backendRef). For Route resources that don't have an error code response like HTTP does, presence of that class of error should probably result in it becoming unattached (since you can't generate any meaningful data plane config at all).

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
+it is the root of the resource tree, it will stay with `Accepted`).
+* In general, resources should be considered `Attached` if their config will
+generate some config in the underlying data plane.
+* All Conditions will be positive polarity (`GoodState: true`, not `BadState: false`)
+* All relevant Conditions for a resource must be added when it's observed.
```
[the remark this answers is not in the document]
youngnick: So, does this mean that all conditions that have Core conformance should always be present? Or does this mean that a set of positive-polarity summary endpoints we name should always be present? The former could include any negative-polarity conditions we define as having Core conformance (which would clash with the "don't have double negatives" guidance we're looking for in other places.)

site-src/geps/gep-1364.md
youngnick: If we're going to have a mixed-polarity set of conditions, then I think we need to define:
- which Conditions are positive polarity (Probably `Attached`, `Programmed`, `Ready`, and `Valid` if we do it)
- which Conditions are positive polarity and should always be present (Probably `Attached`, `Programmed`, and `Valid` if we do it).
- which Conditions are core (currently, everything other than `Ready`)
- which Conditions are extended (only `Ready`).
- which Conditions are negative (all the error conditions).

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,328 @@
...
+really viable:
+* All conditions must be negative polarity
+* All conditions must be positive polarity
+* Some conditions can be positive polarity, but most should be negative.
+
+The fact that the user experience of `Ready` or conditions like `Accepted` being `true`
+in the healthy case is much better rules out the first option, so we are left to
+decide between enforcing that all conditions are positive, or that we have a mix.
+
+Having an arbitrary mix will make doing machine-based extraction of information
```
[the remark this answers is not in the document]
youngnick: It means that anything consuming the Conditions will need to have a hard-coded list of the positive polarity conditions. Not a deal breaker, but it's another obstacle in the way of building a generic condition-parsing library.

site-src/geps/gep-1364.md
youngnick: Ah yes, you're right.

site-src/geps/gep-1364.md
youngnick: There's only one struct allowed for Conditions, so we can't change nil-values. What we'll need to say is "For all conditions other than <this list>, they must be negative polarity, and must only be present if they are status: true". Conversely, this also means that the absent value for these Condition types is "False", not "I don't know yet". Not the end of the world, but a little different to the API conventions.

site-src/geps/gep-1364.md
youngnick: Yes, I guess I should make this clearer, that in this case, it's absolutely expected that there be some additional, negative-polarity error conditions that tell you that the backendRef doesn't exist. (This is a distinct case from the backend existing, but having zero endpoints, which would _not_ generate a "backendRef doesn't exist" error, but would end up with the same behavior).

site-src/geps/gep-1364.md
youngnick: Okay, I'll update this section to:
- add `Ready` as an Extended support condition, that, if present, indicates that the data plane is fully configured. But because it's not always present, we'll need `Reconciled` to indicate what we have here - that everything is okay and the data plane will be ready soon.

site-src/geps/gep-1364.md
shaneutt: For posterity, can we cite some sources for that assertion? It would probably actually be good as a part of this status effort to document somewhere _in the code_ that we're avoiding this due to conventions, and then point to those conventions, as this will help future contributors to more quickly avoid pursuing such things (or give them a path to who they need to convince in the greater kubernetes community).

howardjohn:
ValidSpec can have things not feasible to check in a webhook like temporary dependency issues (referenced service doesn't exist) or implementation specific issues (parent has FooGateway, which doesn't support the Bar field), so I think it holds weight.

If we do want to collapse we could have a single `Valid` field with reasons like `NotValid`, `NotAttached`, etc (naming needs work) possibly...?

Also re-iterating my comment on https://github.com/kubernetes-sigs/gateway-api/pull/1383#discussion_r967280691 since I think I wasn't quite clear - I think making a `Ready` optional condition is actually harmful. It seems like the type of field only useful if its required, and I am not sure how feasible it is to require it.

youngnick:
I guess in my mind, `ValidSpec` is always a requirement for `Attached`, so there's less added value in having the two bits of information. For me, in the cases you give:
* Service doesn't exist  would cause a resource either to program 500s into the data plane and add a more specific condition for HTTPRoute (the already existing `ResolvedRefs`), or cause another Route type that doesn't have an error code like HTTP to become unattached
* The FooGateway not supporting the Bar field would cause the HTTPRoute to fail to attach, and so `Attached` would be false anyway, with Reason and Message to indicate the reason.

To put this another way, for me, `ValidSpec` is _always_ a requirement for `Attached`, so the marginal utility of having both seems low. Invalid objects can't attach, and the reason they can't is that they're invalid.

Is it just the naming that's the problem, or is it the pivoting the status around the idea of attachment?

howardjohn:
> I guess in my mind, `ValidSpec` is always a requirement for `Attached`, so there's less added value in having the two bits of information. For me, in the cases you give:
> 
> * Service doesn't exist  would cause a resource either to program 500s into the data plane and add a more specific condition for HTTPRoute (the already existing `ResolvedRefs`), or cause another Route type - that doesn't have an error code like HTTP - to become unattached.

I think I am confused. I thought ValidSpec meant it was 100% valid. So we would make ValidSpec=false if there is a unknown service. I thought the proposal was to have *only* the conditions listed above, so how would `ResolvedRefs` be used?

> * The FooGateway not supporting the Bar field would cause the HTTPRoute to fail to attach, and so `Attached` would be false anyway, with Reason and Message to indicate the reason.

Maybe.. unless the field is "fail open" and we just ignore it...? I am not sure we have a clear spec here on unsupported fields. It would be a bummer to add a filter and then suddenly have an outage because it wasn't supported.

site-src/geps/gep-1364.md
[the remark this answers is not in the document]
youngnick: I've guessed "summary conditions" here - meaning just the positive polarity ones.

site-src/geps/gep-1364.md
youngnick: This is almost a direct quote from the API conventions as it is, should I put a link in here to the exact line?

youngnick: The exact quote is as follows, I can just inline if it's better, but I was trying to avoid just pasting in the entire section:

---

Condition type names should describe the current observed state of the resource, rather than describing the current state transitions. This typically means that the name should be an adjective ("Ready", "OutOfDisk") or a past-tense verb ("Succeeded", "Failed") rather than a present-tense verb ("Deploying"). Intermediate states may be indicated by setting the status of the condition to Unknown.

For state transitions which take a long period of time (e.g. more than 1 minute), it is reasonable to treat the transition itself as an observed state. In these cases, the Condition (such as "Resizing") itself should not be transient, and should instead be signalled using the True/False/Unknown pattern. This allows other observers to determine the last update from the controller, whether successful or failed. In cases where the state transition is unable to complete and continued reconciliation is not feasible, the Reason and Message should be used to indicate that the transition failed.

---

I don't believe that we generally have any process that we want to handle in the way called out in the exception bullet point. Happy to be guided here on what you'd like to see.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,389 @@
...
+status design hasn't been updated since resources were created.
+
+So, for example, we have GatewayClass with `Accepted`, Gateway with `Scheduled`,
+the Gateway Listeners with `Detached` (which you want to be `false`, unlike the previous
+two), and then Gateways and Gateway Listeners have `Ready`, but Route doesn't (and which
+also you want to be `true`).
+
+This document lays out large-scale changes to the way that we talk about resources,
+and the Conditions to match them. This means that there will be an unavoidable break
+in what constitutes a healthy or unhealthy resource, and there will be changes
```
costinm, line 43: Given the beta status of the API - I would keep the old behavior in place and not break anything, maybe until v1.
Implementations generating both old and new is a pain - but far less than impacting users.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,389 @@
...
+SIG-API Machinery has been reluctant to add fields that would aid in machine-parsing
+of Conditions, especially fields that would indicate the polarity, because they
+are intended more for human consumption than machine consumption.
+
+This means that there's no guidance from upstream about condition polarity. We'll
+discuss this more when we talk about new conditions.
+
+The guidance about Conditions being added as soon as a controller sees a resource
+is a bit unclear - as written in the conventions, it seems to imply that _all_ 
+relevant conditions should always be added, even if their status has to be set to
```
costinm, line 105: I think it is reasonable for the controller to do this - as an indication the resource is in use by the controller.

It becomes problematic if a resource may be processed by multiple controllers - for example the discussion
of HttpRoute applying to both mesh and Gateway can only work if the type has a prefix (MeshReady). 

In general it would be good to indicate how to handle resources where multiple controllers operate on. While 
it's good to know which controller mess with the resource - maybe the controller should skip this step if the
resource is not ignored ( a mesh controller will ignore any route that don't target both mesh and gateway as example)

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,389 @@
...
+`unknown`.
+Gateway API resources do not currently require this, and the practice seems to be
+uncommon.
+
+## Proposed changes
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
```
costinm, line 115: Sound good - but I think Gateway may need additional info, like Ready - since the process may involve 2 steps.
For example Attached will be set when the controller generates the configs - but Ready when certificates are provisioned. CertManager would watch for Attached and start the ACME process, create the Secret - and the gateway controller would switch to Ready.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,389 @@
...
+
+This GEP proposes replacing all conditions that indicate syntactic and semantic
+validity with one, `Attached` condition type, with the exception of the
+GatewayClass resource (because it's the root of the resource tree and doesn't
+attach to anything).
+
+That is, the proposal is to replace:
+* `Scheduled` on Gateway
+* `Detached` on Listener
+* `Accpeted` on Route
```
candita, line 248:
```suggestion
* `Accepted` on Route
```

youngnick: Thanks!

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,389 @@
...
+resources have them as `status: false`). `Ready` is an example of a positive polarity
+condition, and conditions like `Conflicted` from Listener or `NetworkUnavailable`,
+`MemoryPressure`, or `DiskPressure` from the Node resource are examples of
+negative-polarity conditions.
+
+There is also some extra context that's not in the API conventions doc:
+
+SIG-API Machinery has been reluctant to add fields that would aid in machine-parsing
+of Conditions, especially fields that would indicate the polarity, because they
+are intended more for human consumption than machine consumption.
```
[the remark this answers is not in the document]
youngnick: I think it's more that API Machinery has said specifically that they won't add a field for indicating the polarity of a Condition to the Condition struct. This was in a discussion raised by @evankanderson, but I can't seem to find the reference now.

site-src/geps/gep-1364.md
youngnick: In most cases, the Conditions slice is in a separate struct that's part of a slice that's namespaced by controller name - so that controllers should be able to update only their own Conditions. I'd expect that mesh controllers should have a distinct controller name string, so this method should continue working.

site-src/geps/gep-1364.md
youngnick: I'm proposing a `Programmed` condition to summarize this without the "this is ready right now" implications of Ready. I'll put those changes into this summary.

site-src/geps/gep-1364.md
youngnick: I added a section about this just below.

youngnick:
Okay, I think I've captured all the outstanding comments in that update. Please take a look, our timebox for this last pass of reviews is in just under 48 hours, so that I can start working on the implementation.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+`unknown`.
+Gateway API resources do not currently require this, and the practice seems to be
+uncommon.
+
+## Proposed changes
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
```
robscott, line 123:
```suggestion
for processing will be replaced with `Attached`, except for GatewayClass (since
```

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+if it's used by any implementations after some time has passed. If not, it may be
+removed.
+* To capture the behavior that `Ready` currently captures, `Programmed` will be
+introduced. This means that the implementation has seen the config, has everything
+it needs, parsed it, and sent configuration off to the data plane. The configuration
+should be available "soon". We'll leave "soon" undefined for now.
+* The summary conditions will be added into the CRD definitions
+as default values for `status.Conditions`, with a status of `Unknown`.
+* Resolving a comment that came up, documentation will be added to clarify that
+it's okay to add your own Conditions, and that implementations should namespace
```
robscott, line 144: We should probably also recommend that implementations publish both new and old conditions for several months to provide a smoother transition.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+
+    * `type` is a string value indicating the Condition type. `Accepted`, `Scheduled`,
+    and `Ready` are current examples.
+    * `status` indicates the state of the condition, and can be one of three values,
+    `true`, `false`, or `unknown`. Unkown in particular is important, because it
+    means that the controller is unable to determine the status for some reason.
+    * `reason` is a CamelCase string that is a brief description of the reason why
+    the `status` is set the way it is.
+    * `observedGeneration` is an optional field that sets what the `metadata.generation`
+    field was when the controller last saw a resource.
```
robscott, line 81: I think we'd had some conversations about making this required for conformant Gateway implementations. Probably out of scope for this GEP, just wanted to make sure it was intentionally excluded.

robscott (commented):
Thanks @youngnick! These are great updates. A few nits and a question about naming but really like this direction.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+Gateway API resources do not currently require this, and the practice seems to be
+uncommon.
+
+## Proposed changes
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
+it is the root of the resource tree, it will stay with `Accepted`).
```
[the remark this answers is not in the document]
skriss: I had the same thought and am +1 to this, `Accepted` seems a little bit easier to understand for users IMO and can be consistently used across all resources.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+  * Any specified TLS secrets exist
+* The resource is supported by the controller by ensuring things like:
+  * Any Kinds being referred to by the resource are supported
+  * Features being used by the resource are supported
+
+All of these rules can be summarized into:
+* The resource is valid enough to produce some configuration in the underlying
+data plane.
+
+For Gateway, `Attached` also subsumes the functions of `Scheduled`: `Attached`
```
howardjohn, line 289: On some impls, provisioning takes about a minute. That may mean that some errors that are immediately known are masked. Or I suppose they aren't masked, but we need to wait the entire minute to know if its valid. The message field could say "all fields valid, waiting for deployment" perhaps, but that isn't in the machine parse-able section.

That may not warrant change, just a thought

howardjohn (commented):
I am not expert here things seem reasonable to me

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Proposed changes summary
+
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
+it is the root of the resource tree, it will stay with `Accepted`).
+* In general, resources should be considered `Attached` if their config is valid
+enough to generate some config in the underlying data plane.
+* There will be a limited set of positive polarity summary conditions, and a number
+of other specific negative-polarity error conditions.
```
evankanderson, line 128: This will make UI display somewhat harder (green dots / checkmarks / exclamation points), but I realize it may align with human reading a bit more easily.

I don't have a strong opinion here, just want to highlight the cost elsewhere in the stack.

site-src/geps/gep-1364.md
evankanderson: Is it possible for something like an HTTPRoute to define two paths, one of which is "valid enough" config, while the other is garbage?

If so, are there expectations or requirements from a conformance PoV ("always apply all or none") that we need to specify?

site-src/geps/gep-1364.md
evankanderson: You might also get a `BackendRefEmpty` condition...

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+    * `observedGeneration` is an optional field that sets what the `metadata.generation`
+    field was when the controller last saw a resource.
+
+4. Conditions shoud describe the _current state_ of the resource at observation
+time, which means that they should be an adjective (like `Ready`), or a past-tense
+verb (like `Attached`). This one in particular is documented pretty closely in the
+[Typical status properties][typstatus] section of the guidelines.
+5. Conditions should be applied to a resource the first time the controller sees
+the resource. This seems to imply that _all conditions should be present on every
+resource owned by a controller_, but the rest of the conventions don't make this
```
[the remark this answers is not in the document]
robscott: I agree that we should have a set of permanent and reliable conditions that are always present, but I personally don't think it makes sense for error conditions to always be present, especially since that tends to lead to confusing double negative states.

site-src/geps/gep-1364.md
robscott: I think our practice of including `controllerName` for each distinct set of conditions should largely allow multiple controllers to cooperate on status for the same resource, but open to other potential improvements.

site-src/geps/gep-1364.md
sunjayBhatia: > Is it possible for something like an HTTPRoute to define two paths, one of which is "valid enough" config, while the other is garbage?

Most cases of an invalid configuration in one route rule (with others being completely valid within the same HTTPRoute) seem to be misconfigurable only in the context of a reference to another resource and fall into the `ResolvedRefs` bucket (which seems stated below and possibly agreed upon that `Attached: true` be set with `ResolvedRefs: false` and return HTTP code 500 for the invalid rule).

One such example not covered is if you were to use a `HTTPRequestRedirectFilter` and `HTTPURLRewrite` in the list of filters for an individual `HTTPRouteRule` ([ref](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter)). At the moment it doesn't appear we have prescribed any status code or other error handling for this error. This seems to fall into the bucket of things that cannot produce data plane configuration.

Going by the statements below, if this rule were the only one present `Attached: false` should be set, since it cannot result in any configuration (with another more specific error condition). If present with other valid rules, `Attached: true` should be set (with another more specific error condition again). I can definitely see the argument for setting `Attached: false` however in the latter case, since stated above, the `Attached` condition means `The resource is syntatically and semantically valid, and internally consistent`.

It seems that the `Valid: true` condition mentioned above ends up being somewhat equivalent to having  for example for HTTPRoute `ResolvedRefs: true` and no error conditions (and vice versa, `Valid: false` equivalent to `ResolvedRefs: false` *or* some more specific error condition). Maybe this framing helps a little with determining possible merits of having another condition?

> If so, are there expectations or requirements from a conformance PoV ("always apply all or none") that we need to specify?

Just doing a quick pass, I don't *think* I saw any existing conformance tests that cover this kind of thing so I agree that would be good to add.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+* `Detached` on Listener
+* `Accepted` on Route
+
+with `Attached` in all these locations.
+
+GatewayClass will maintain the `Accepted` condition.
+
+All of these conditions share the following meanings:
+* The resource has been accepted for processing by the controller
+* The resource is syntatically and semantically valid, and internally consistent
```
sunjayBhatia, line 275: nit

```suggestion
* The resource is syntactically and semantically valid, and internally consistent
```

site-src/geps/gep-1364.md
evankanderson: How do `controllerName` conditions roll up to `Accepted`?

Also, controllers probably need to be sure to use PATCH on the `status` sub-resource or do conditional PUTs and avoid extra writes if the conditions haven't changed.

site-src/geps/gep-1364.md
youngnick: Good point, I updated.

site-src/geps/gep-1364.md
youngnick: For some resources, all conditions are namespaced by `controllerName`. I agree we need to ensure that the rollup behaivor is specified. I'm going to add a section about partial validity that should help.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+* All the current Conditions that indicate that the resource is okay and ready
+for processing witll be replaced with `Attached`, except for GatewayClass (since
+it is the root of the resource tree, it will stay with `Accepted`).
+* In general, resources should be considered `Attached` if their config is valid
+enough to generate some config in the underlying data plane.
+* There will be a limited set of positive polarity summary conditions, and a number
+of other specific negative-polarity error conditions.
+* All relevant positive-polarity summary Conditions for a resource must be added
+when it's observed.
+For example, HTTPRoutes must always have `Accepted` and `ResolvedRefs`, regardless
```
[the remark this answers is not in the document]
youngnick: I hadn't updated, but with the change to `Accepted` everywhere, this is now correct, thanks.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+* Negative polarity error conditions must only be added when the error is True.
+* The `Ready` condition will be moved to Extended conformance, and we'll reevaluate
+if it's used by any implementations after some time has passed. If not, it may be
+removed.
+* To capture the behavior that `Ready` currently captures, `Programmed` will be
+introduced. This means that the implementation has seen the config, has everything
+it needs, parsed it, and sent configuration off to the data plane. The configuration
+should be available "soon". We'll leave "soon" undefined for now.
+* The summary conditions will be added into the CRD definitions
+as default values for `status.Conditions`, with a status of `Unknown`.
```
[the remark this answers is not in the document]
youngnick: I was intending to make it that the non-`controllerName` Conditions will be present, but I'll leave this out for now.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+attach to anything).
+
+That is, the proposal is to replace:
+* `Scheduled` on Gateway
+* `Detached` on Listener
+* `Accepted` on Route
+
+with `Attached` in all these locations.
+
+GatewayClass will maintain the `Accepted` condition.
```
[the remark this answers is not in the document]
youngnick: Yeah, agreed. I've migrated the proposal to this.

site-src/geps/gep-1364.md
youngnick: Avoiding "always apply all or none" is the reason this is tricky. There are mixed expectations about what happens if _some_ config is present but bad, and I'm attempting to thread the needle here and make this as user-friendly as possible while not becoming really difficult to implement or understand.

site-src/geps/gep-1364.md
youngnick: I worded this carefully, so as not to imply that the provisioning needs to have _completed_, just that _it can be started_. The completion of infra provisioning should be handled by `Programmed` or `Ready`, or both.

site-src/geps/gep-1364.md
```diff
@@ -0,0 +1,409 @@
...
+`Ready`.
+
+
+### Should conditions always be added?
+
+Not all of them.
+
+Positive polarity Conditions that describe the desirable state of the object must
+always be set. These are currently `Attached`, `ResolvedRefs`, and `Programmed`.
+Implementations that use `Ready` must also add it immediately, set to `false`.
```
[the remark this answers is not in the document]
youngnick: Yeah, that seems like a good idea.

site-src/geps/gep-1364.md
youngnick: I added a section on partial validity to hopefully make this clearer.

youngnick:
Okay, updated again, this feels like it's getting closer.

Changelog this time:
- Changed references to `Attached` to `Accepted`. I think @robscott and @skriss are right here, this is probably clearer.
- Added a section on partial validity to explain this better.
- Added the example @sunjayBhatia gave to the examples.

robscott:
Thanks @youngnick! This is a really well written GEP. Since previous generations of this PR have already been approved a couple times, I think we've got pretty clear consensus around this approach.

/lgtm

mikemorris (commented):
Quite late in doing a thorough review of this, but really happy with where it ended up - feels like a good step forwards in resolving some of the current ambiguity in status conditions. Thanks @youngnick!

robscott:
@mikemorris these are great bits of feedback, do you have time to create a follow up PR or issue to address any that are straightforward to fix? Concerned that otherwise we'll lose track of these.

PR #1394, opened 2022-09-20
Add flags for SupportedFeatures and ExemptFeatures in conformance tests.

### LINKED ISSUE #1391: Add flags for `SupportedFeature` and `ExemptFeature` in conformance tests
A flag for configuring `SupportedFeature` and another for `ExemptFeature` in the conformance test suite.

**Why this is needed**:
This would let implementations choose which features they want to include in their conformance test runs, depending on which features are supported/unsupported by them.

gyohuangxin:
**What type of PR is this?**

/kind feature

This PR add flags for SupportedFeatures and ExemptFeatures in conformance tests.

Examples:

```
// conformance tests with supported features 
go test ./conformance --gateway-class nginx -v -supported-features ReferenceGrant
go test ./conformance --gateway-class nginx -v -supported-features ReferenceGrant,HTTPRouteQueryParamMatching,HTTPRouteMethodMatchig

// conformance tests with exempt features
go test ./conformance --gateway-class nginx -v -exempt-features ReferenceGrant
```

Fixes #1391

```release-note
1. Add flags for SupportedFeatures and ExemptFeatures in conformance tests.
2. Remove suite.SupportReferenceGrant as the default value of SupportedFeatures.
3. Replace `Features:    []suite.SupportedFeature{suite.SupportReferenceGrant}` from Features list to `Exemptions:  []suite.ExemptFeature{suite.ExemptReferenceGrant}` on all tests.
```

### DIFF
--- a/conformance/conformance_test.go
+++ b/conformance/conformance_test.go
@@ -18,6 +18,7 @@ limitations under the License.
 package conformance_test
 
 import (
+	"strings"
 	"testing"
 
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
@@ -45,15 +46,37 @@ func TestConformance(t *testing.T) {
 
 	t.Logf("Running conformance tests with %s GatewayClass", *flags.GatewayClassName)
 
+	supportedFeatures := parseSupportedFeatures(*flags.SupportedFeatures)
+	exemptFeatures := parseExemptFeatures(*flags.ExemptFeatures)
+
 	cSuite := suite.New(suite.Options{
 		Client:               client,
 		GatewayClassName:     *flags.GatewayClassName,
 		Debug:                *flags.ShowDebug,
 		CleanupBaseResources: *flags.CleanupBaseResources,
-		SupportedFeatures: []suite.SupportedFeature{
-			suite.SupportReferenceGrant,
-		},
+		SupportedFeatures:    supportedFeatures,
+		ExemptFeatures:       exemptFeatures,
 	})
 	cSuite.Setup(t)
 	cSuite.Run(t, tests.ConformanceTests)
+}
+
+// parseSupportedFeatures parses the arguments for supported-features flag,
+// then converts the string to []suite.SupportedFeature
+func parseSupportedFeatures(f string) []suite.SupportedFeature {
+	var res []suite.SupportedFeature
+	for _, value := range strings.Split(f, ",") {
+		res = append(res, suite.SupportedFeature(value))
+	}
+	return res
+}
+
+// parseExemptFeatures parses the arguments for exempt-features flag,
+// then converts the string to []suite.ExemptFeature
+func parseExemptFeatures(f string) []suite.ExemptFeature {
+	var res []suite.ExemptFeature
+	for _, value := range strings.Split(f, ",") {
+		res = append(res, suite.ExemptFeature(value))
+	}
+	return res
 }
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -27,4 +27,6 @@ var (
 	GatewayClassName     = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
 	ShowDebug            = flag.Bool("debug", false, "Whether to print debug logs")
 	CleanupBaseResources = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
+	SupportedFeatures    = flag.String("supported-features", "SupportReferenceGrant", "Supported features included in conformance tests suites")
+	ExemptFeatures       = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
 )
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -27,6 +27,6 @@ var (
 	GatewayClassName     = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
 	ShowDebug            = flag.Bool("debug", false, "Whether to print debug logs")
 	CleanupBaseResources = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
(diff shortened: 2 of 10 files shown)


conformance/utils/flags/flags.go
```diff
@@ -27,4 +27,6 @@ var (
 	GatewayClassName     = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
 	ShowDebug            = flag.Bool("debug", false, "Whether to print debug logs")
 	CleanupBaseResources = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
+	SupportedFeatures    = flag.String("supported-features", "SupportReferenceGrant", "Supported features included in conformance tests suites")
+	ExemptFeatures       = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
```
aryan9600, line 31: Instead of using `flag.String` and then having helper parsing functions we can use `flag.StringSlice`.

aryan9600: To not break the current behavior, let's keep the default value of `SupportedFeatures` as `[]suite.SupportedFeature{suite.SupportReferenceGrant}`

gyohuangxin: > Instead of using `flag.String` and then having helper parsing functions we can use `flag.StringSlice`.

Sorry, I didn't find `flag.StringSlice` in flag packge https://pkg.go.dev/flag.

> To not break the current behavior, let's keep the default value of SupportedFeatures as []suite.SupportedFeature{suite.SupportReferenceGrant}

The helper parsing function will convert the value `SupportedFeatures` to `[]suite.SupportedFeature{suite.SupportReferenceGrant}`, which doesn't break current behavior and makes the arguments looks more concise.

aryan9600: My bad, I though we were using [pflag](https://pkg.go.dev/github.com/spf13/pflag) 🤦

gyohuangxin: It's fine, haha. I hope it could have a `StringSlice` function, but it doesn't, I have to use another helper parsing function.

conformance/conformance_test.go
```diff
@@ -45,15 +46,37 @@ func TestConformance(t *testing.T) {
...
+	supportedFeatures := parseSupportedFeatures(*flags.SupportedFeatures)
+	exemptFeatures := parseExemptFeatures(*flags.ExemptFeatures)
+
 	cSuite := suite.New(suite.Options{
 		Client:               client,
 		GatewayClassName:     *flags.GatewayClassName,
 		Debug:                *flags.ShowDebug,
 		CleanupBaseResources: *flags.CleanupBaseResources,
-		SupportedFeatures: []suite.SupportedFeature{
-			suite.SupportReferenceGrant,
```
mikemorris, line 54: I believe the original intent of having this specific feature hardcoded was to include ReferenceGrant in the default set of conformance tests - I'm not sure if we should try to keep a "default supported features" list somehow (feels awkward), or instead just change conformance tests currently enabled with the flag to instead switch to an ExemptFeature flag to opt-out.

mikemorris:
/ok-to-test

@gyohuangxin Could you add a changelog entry in the "user-facing change" section?

gyohuangxin:
> /ok-to-test
> 
> @gyohuangxin Could you add a changelog entry in the "user-facing change" section?

Updated the "user-facing change" section.

conformance/conformance_test.go
gyohuangxin: Maybe we can make a "default supported features" which runs all the conformance tests, and we can change enabled tests with `SupportedFeatures` flag or just opt-out tests with `ExemptFeatures` flag. What do you think?

gyohuangxin:
@aryan9600 @mikemorris Any more comments?

conformance/conformance_test.go
mikemorris: I think we could just remove the [`suite.SupportReferenceGrant`](https://github.com/kubernetes-sigs/gateway-api/search?q=SupportReferenceGrant) constant, and remove it from the `Features` list on all conformance tests that currently expect it to be set.

This keeps those tests included by default, still allows using the `ExemptReferenceGrant` flag to opt-out, avoids any complexity of an extra "default" list, and feels okay given that we're intending to promote ReferenceGrant to v1beta1 in the upcoming v0.6.0 Gateway API release.

gyohuangxin: @mikemorris Thanks for your reply, I can understand it basically. What to do are:
1. Remove [suite.SupportReferenceGrant](https://github.com/kubernetes-sigs/gateway-api/search?q=SupportReferenceGrant) as the default value of `SupportedFeatures`
2. Remove [suite.SupportReferenceGrant](https://github.com/kubernetes-sigs/gateway-api/search?q=SupportReferenceGrant) from `Features` list on all tests, such as https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/gateway-secret-invalid-reference-grant.go#L37. Then replace it to ```Exemptions: []suite.ExemptFeature{
		suite.ExemptReferenceGrant,
	},```, so that `--exempt-features ExemptReferenceGrant` flag can be set to opt-out.
	
Can I confirm it is correct?

mikemorris: Yes, that should be correct @gyohuangxin!

I think we can then do
3. Remove the constant at https://github.com/kubernetes-sigs/gateway-api/blob/583e4e2541b757d5e72d23c18f2d7201b49bccf2/conformance/utils/suite/suite.go#L47-L48. In upcoming Gateway API v0.6.0, all implementations should assume ReferenceGrant is supported unless explicitly setting `ExemptReferenceGrant`.

gyohuangxin: @mikemorris Thanks! I've updated the code and the "user-facing change" section per your comments, please review again.

gyohuangxin:
@shaneutt Can you help review it? Thanks.

gyohuangxin:
@shaneutt Cloud you please have a look on that?

robscott:
Thanks @gyohuangxin!

/approve

PR #1401, opened 2022-09-21
Clarify HTTPRoute matching precedence

### LINKED ISSUE #1381: Clarify Route matching precedence language
The language describing Route match precedence order still seems a little unclear: https://github.com/kubernetes-sigs/gateway-api/blob/c1adb20f2d476dee1b95c255aedfa9b49e8b59f0/apis/v1beta1/httproute_types.go#L144-L167

In https://github.com/kubernetes-sigs/gateway-api/issues/1370 I interpreted the section about `largest number of...Header matches` to mean the Rule with the largest number of matching headers in a `HTTPRouteMatch` whereas another valid interpretation is below:

@sunjayBhatia reading the current spec literally, I would think this request should go to `infra-backend-v1`. Per https://github.com/kubernetes-sigs/gateway-api/blob/a3f769e5c690546140b6b9f0db5aea4d766b8f68/apis/v1beta1/httproute_types.go#L143-L152, we should prioritize the **Rule** with the greatest number of header matches, which in this case is:

```
 # Matches "color: blue" OR "color: green"
  - matches:
    - headers:
      - name: color
        value: blue
    - headers:
      - name: color
        value: green
    backendRefs:
    - name: infra-backend-v1
      port: 8080
```

Since this is a single Rule that has 2 header matches (even though each individual Match within the Rule only has one each).

_Originally posted by @skriss in https://github.com/kubernetes-sigs/gateway-api/issues/1370#issuecomment-1238472815_

We should clarify this language a bit and make a decision on what is the conformant behavior, to change or keep https://github.com/kubernetes-sigs/gateway-api/issues/1370

sunjayBhatia:
**What type of PR is this?**
/kind cleanup
/kind documentation

HTTPRoute matching precedence documentation seems a little unclear, as shown by https://github.com/kubernetes-sigs/gateway-api/pull/1370 and further discussion. This PR attempts to clarify it by:
- moving the note about hostname matching closer to the hostname field
- explicitly referencing how rules and matches interact, rather than rules alone

Fixes #1381

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -93,6 +93,16 @@ type HTTPRouteSpec struct {
 	// implementation must raise an 'Accepted' Condition with a status of
 	// `False` in the corresponding RouteParentStatus.
 	//
+	// In the event that multiple HTTPRoutes specify intersecting hostnames (e.g.
+	// overlapping wildcard matching and exact matching hostnames), precedence must
+	// be given to rules from the HTTPRoute with the largest number of:
+	//
+	// * Characters in a matching non-wildcard hostname.
+	// * Characters in a matching hostname.
+	//
+	// If ties exist across multiple Routes, the matching precedence rules for
+	// HTTPRouteMatches takes over.
+	//
 	// Support: Core
 	//
 	// +optional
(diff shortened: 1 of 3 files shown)


shaneutt (approved):
/lgtm

Issue #627, opened 2021-04-23
Provide admission server builds for CPU architectures other than amd64


hbagdi:
**What would you like to be added**:

Admission server builds are currently only available for amd64 architecture.
We should provide builds for other architectures as well.

**Why is this needed**:
Our users could be running k8s clusters on other CPU architectures.

hbagdi:
For some context, #621 was the first attempt to get an image built and published. We decided to take an incremental approach to building out our images. There is no blocker for building docker images for other CPU architectures.

youngnick:
/good-first-issue

Outstanding work here is to update our webhook admission build to some other architectures, `linux/arm64` for certain, and to make sure we can add more if needed.

akankshakumari393:
Quick question - 
1. what all arch do we want to add?
2. All arch images need to use to repo that has arch information in its name, earlier we were using `${REGISTRY}/admission-server:${VERSION_TAG}`. Now it should be `${REGISTRY}/admission-server-${ARCH}:${VERSION_TAG}`. So do we need to create new repo or is it handled by infra team.

youngnick:
I think that `amd64` and `arm64` are probably okay for now, we can add more later if required.

I didn't know that was a requirement, maybe we can check with some other things like ingress-nginx as to how they handle this (I don't think that other Kubernetes projects have the arch in the repo name?)

shaneutt:
We now have a multi-arch build with new `arm64` support:

https://console.cloud.google.com/gcr/images/k8s-staging-gateway-api/global/admission-server@sha256:5dd8c42658daeb0a26ba0e4ac5e33c3bba2f1f4cd0f41685f837dd4a3684d2f0/details?tab=pull

PR #1402, opened 2022-09-25
Provide admission server builds for CPU architectures other than amd64

akankshakumari393:
Fixes - #627 

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ WORKDIR /go/src/sig.k8s.io/gateway-api
 COPY  . .
 ARG TAG
 ARG COMMIT
-RUN CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -a -o gateway-api-webhook \
+ARG ARCH
+RUN CGO_ENABLED=0 GOARCH=${ARCH} GOOS=linux go build -a -o gateway-api-webhook \
       -ldflags "-s -w -X main.VERSION=$TAG -X main.COMMIT=$COMMIT" ./cmd/admission
 
 FROM gcr.io/distroless/static:nonroot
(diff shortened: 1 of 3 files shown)


shaneutt (commented):
/ok-to-test

shaneutt (changes requested):
Most of this looks good, but one important thing it does is add separate images for different architectures. It would seem to me that ideally we would want a multi-arch build as this will have the benefit of allowing container runtimes to pull their version automatically without the image name having to be managed at a higher level. @robscott, @youngnick is there any reason we _wouldn't_ want or be able to do this as a multi-arch build? :thinking:

Dockerfile
```diff
@@ -18,7 +18,8 @@ WORKDIR /go/src/sig.k8s.io/gateway-api
 COPY  . .
 ARG TAG
 ARG COMMIT
-RUN CGO_ENABLED=0 GOARCH=amd64 GOOS=linux go build -a -o gateway-api-webhook \
+ARG ARCH
```
howardjohn, line 21: can replace with https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

hack/build-and-push.sh
```diff
@@ -64,10 +70,10 @@ fi
 
 # First, build the image, with the version info passed in.
 # Note that an image will *always* be built tagged with the GIT_TAG, so we know when it was built.
-docker build --build-arg COMMIT=${COMMIT} --build-arg TAG=${BINARY_TAG} \
-  			-t ${REGISTRY}/admission-server:${GIT_TAG} .
-docker push ${REGISTRY}/admission-server:${GIT_TAG}
+docker build --build-arg COMMIT=${COMMIT} --build-arg TAG=${BINARY_TAG} ARCH=${ARCH} \
```
howardjohn, line 73: I think we need to switch to `docker buildx build --platforms={...}`.

As-is, we won't get a single multi-arch image

akankshakumari393: Ok i will check👍

robscott: @akankshakumari393 are you still able to work on this? Anything we can help with?

robscott:
It looks like this PR has gone stale and will be replaced by https://github.com/kubernetes-sigs/gateway-api/pull/1462, closing this one out.

PR #1405, opened 2022-09-26
Marking v1alpha2 as deprecated for beta resources

robscott:
**What type of PR is this?**
/kind cleanup
/kind api-change

This marks v1alpha2 as deprecated for beta resources. This will provide a gentler transition than immediately removing v1alpha2 from the list of served versions. Implementations should still switch to read and write from v1beta1 as soon as possible.

```release-note
v1alpha2 has been deprecated for Gateway, GatewayClass, and HTTPRoute. This version of the API will be removed from CRDs in a future release. Implementations should update to use v1beta1 for these resources as soon as possible.
```


howardjohn (approved):

shaneutt (approved):
/lgtm

PR #1412, opened 2022-09-27
`ControllerName` validation helper

### LINKED ISSUE #1410: `ControllerName` validation helper

The GatewayClass `ControllerName` validation is enforced at the CRD level through the following regex:
`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`

Since the controllerName is a field that may be used by the implementations to check that the `GatewayClass` has to be reconciled by the controller itself, it would be beneficial to provide a helper for checking its validity. This helper could be added under `apis/v1beta1/validation` or `apis/v1beta1/util/validation`

**Why this is needed**:
This is needed to provide a helper to check the `ControllerName` validity by the various implementations.

mlavacca:
**What type of PR is this?**

/kind feature

A utility to check the Controller name validity has been added to apis/<version>/util/validation.

Fixes #1410 

```release-note
NONE
```


apis/v1alpha2/util/validation/gatewayclass.go
```diff
@@ -0,0 +1,34 @@
...
+
+package validation
+
+import (
+	"regexp"
+
+	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// IsControllerNameValid checks that the provided controllerName complaints with the expected
```
sunjayBhatia, line 25: spelling nit, same for the other copy of this

```suggestion
// IsControllerNameValid checks that the provided controllerName complies with the expected
```

apis/v1beta1/util/validation/gatewayclass.go
```diff
@@ -0,0 +1,33 @@
...
+
+package validation
+
+import (
+	"regexp"
+
+	gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// IsControllerNameValid checks that the provided controllerName complaints with the expected
```
akankshakumari393, line 25:
```suggestion
// IsControllerNameValid checks that the provided controllerName complies with the expected
```

apis/v1alpha2/util/validation/gatewayclass_test.go
```diff
@@ -0,0 +1,58 @@
...
+package validation_test
+
+import (
+	"testing"
+
+	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	validationtutils "sigs.k8s.io/gateway-api/apis/v1alpha2/util/validation"
+)
+
+func TestIsControllerNameValid(t *testing.T) {
```
sunjayBhatia, line 26: maybe worth adding another test or two that exercise the regex more, w/ invalid characters etc.

howardjohn (commented):
TBH I don't get the use case. Seems like each impl will make pick `controllerName` as a constant, and if its invalid they will pick a new one? Not sure why we need a Go helper...

I also don't mind though

apis/v1alpha2/util/validation/gatewayclass_test.go
mlavacca: I added a couple more test cases.

apis/v1alpha2/util/validation/gatewayclass.go
```diff
@@ -0,0 +1,34 @@
...
+// format. It must be a non-empty domain prefixed path.
+func IsControllerNameValid(controllerName gatewayv1a2.GatewayController) bool {
+	if controllerName == "" {
+		return false
+	}
+
+	re := regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`)
```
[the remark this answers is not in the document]
mlavacca: I've set it as a global variable so that it is compiled only once.

mlavacca:
> TBH I don't get the use case. Seems like each impl will make pick `controllerName` as a constant, and if its invalid they will pick a new one? Not sure why we need a Go helper...

Recently in the Kong implementation, we have had to make it customizable via argument. This helper is meant to validate the controllerName in such a use case.

akankshakumari393:
Yeah it happens, If you request one, the others would be removed from review request.

youngnick:
For the case that people take the controllerName as an argument, having common code to validate seems like a great idea to me.

/approve

I'll leave the LGTM for someone else though.

shaneutt (changes requested):
Code and idea LGTM, however one thing I would like to talk about possibly changing:

We're putting this in two places with the same regex. Having copied code is usually not ideal. We recently made changes to share types between v1alpha2 and v1beta1 to reduce some of these copies, I think we should do the same here.

mlavacca:
> We're putting this in two places with the same regex. Having copied code is usually not ideal. We recently made changes to share types between v1alpha2 and v1beta1 to reduce some of these copies, I think we should do the same here.

I factorized the regexp and kept only the `v1beta1` instance.

shaneutt (commented):
We should be able to share _all_ the v1beta1 code backwards to v1alpha2, rather than just the regexp no?

PR #1415, opened 2022-09-28
chore: remove workaround logics in verify-examples-kind.sh

### LINKED ISSUE #1366: Update hack/verify-examples-kind.sh to work better with releases
Something that allows us to:

1. Run tests using the latest webhook image that has been built (latest commit, not latest release)
2. Still pass `hack/verify-codegen.sh`
3. Have a simple release process that allows us to at least temporarily set a fixed image version in our webhook deployment for our generated YAML attached to the release + the config/webhook YAML found at the release tag.

**Why this is needed**:
At one point we added the following to the script to ensure that our deployed YAML would use the latest image while our manifests on the main branch would not:

```
# Temporary workaround for release
sed -i 's/v0.5.0/latest/g' config/webhook/admission_webhook.yaml
```

This had the unfortunate affect of making a change that would cause our hack/verify-codegen.sh script to fail. https://github.com/kubernetes-sigs/gateway-api/pull/1348#discussion_r956890842 resolved that issue, but it all feels like a bit of a hack. We need a more resilient solution here.

tokers:
Signed-off-by: Chao Zhang <tokers@apache.org>

**What type of PR is this?**

/kind feature
/kind cleanup

Previously, we have some workaround logics in the `hack/verify-examples-kind.sh`, it modifies the image tag in `config/webhook/admission_webhook.yaml`, so that the latest commit of admission webhook can be used.

This PR removed the hack logic by using the customization feature in Kustomize, see https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#customizing for the details. We create the `config/webhook/kustomization.yaml` file temporarily (and remove it when cleanup), which replaces the webhook image tag from `v0.5.0` (or others in the future) to `latest`.

Fixes #1366

NONE

### DIFF
--- a/hack/verify-examples-kind.sh
+++ b/hack/verify-examples-kind.sh
@@ -22,6 +22,7 @@ readonly GO111MODULE="on"
 readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
 readonly CLUSTER_NAME="verify-gateway-api"
+readonly ADMISSION_WEBHOOK_VERSION="v0.5.1"
 
 export KUBECONFIG="${GOPATH}/.kubeconfig"
 export GOFLAGS GO111MODULE GOPATH
@@ -60,7 +61,7 @@ resources:
   - certificate_config.yaml
   - admission_webhook.yaml
 images:
-  - name: gcr.io/k8s-staging-gateway-api/admission-server:v0.5.0
+  - name: gcr.io/k8s-staging-gateway-api/admission-server:${ADMISSION_WEBHOOK_VERSION}
     newTag: latest
 EOF
 
--- a/hack/verify-examples-kind.sh
+++ b/hack/verify-examples-kind.sh
@@ -33,6 +33,9 @@ cleanup() {
   if [ "$CLEANED_UP" = "true" ]; then
     return
   fi
+
+  rm config/webhook/kustomization.yaml
+
   if [ "${KIND_CREATE_ATTEMPTED:-}" = true ]; then
     kind delete cluster --name "${CLUSTER_NAME}" || true
   fi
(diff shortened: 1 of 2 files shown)


hack/verify-examples-kind.sh
```diff
@@ -51,11 +54,19 @@ res=0
 KIND_CREATE_ATTEMPTED=true
 kind create cluster --name "${CLUSTER_NAME}" || res=$?
 
+cat <<EOF >config/webhook/kustomization.yaml
+resources:
+  - 0-namespace.yaml
+  - certificate_config.yaml
+  - admission_webhook.yaml
+images:
+  - name: gcr.io/k8s-staging-gateway-api/admission-server:v0.5.0
```
robscott, line 63:
```suggestion
  - name: gcr.io/k8s-staging-gateway-api/admission-server:v0.5.1
```

robscott: Would be even better if this was a var that could be set at the top of the file.

hack/verify-examples-kind.sh
```diff
@@ -105,9 +116,6 @@ for CHANNEL in experimental standard; do
         echo Examples failed as expected
 done
 
-# Undo workaround from earlier
-sed -i 's/latest/v0.5.1/g' config/webhook/admission_webhook.yaml
```
robscott, line 109: Removing this line means that running this command locally will leave `config/webhook/kustomization.yaml` in your git diff. Do you mind adding `rm config/webhook/kustomization.yaml` here?

hack/verify-examples-kind.sh
```diff
@@ -33,6 +33,9 @@ cleanup() {
   if [ "$CLEANED_UP" = "true" ]; then
     return
   fi
+
+  rm config/webhook/kustomization.yaml
```
robscott, line 37: Why does this need to be here?

robscott (commented):
Thanks @tokers! This is a creative solution.

hack/verify-examples-kind.sh
tokers: Or if the `config/webhook/kustomization.yaml` is appropriate to be there, we can just leave it.

hack/verify-examples-kind.sh
tokers: Yes, it makes sense. I'll think about it.

hack/verify-examples-kind.sh
tokers: OK.

tokers: OK

tokers: I think leaving the remove logic in `cleanup` would be better.

robscott:
Thanks @tokers!

/lgtm
/approve

PR #1416, opened 2022-09-28
Update golangci-lint (1.46.2->1.50.0)

oscr:
**What type of PR is this?**

/kind bug
/kind cleanup

Updates the golangci-lint version from 1.46.2 -> 1.49.0 and fixes all the findings. This is needed due to linters being disabled currently. Also adds `make lint` target which just makes it easier for developers. 

Here is the current main branch linter output:
```
hack/verify-golint.sh
level=warning msg="[linters context] bodyclose is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] contextcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] nilerr is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] noctx is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] rowserrcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] sqlclosecheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] structcheck is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
level=warning msg="[linters context] unparam is disabled because of go1.18. You can track the evolution of the go1.18 support by following the https://github.com/golangci/golangci-lint/issues/2649."
```

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -272,15 +272,15 @@ type GRPCRouteRule struct {
 // For example, the match below will match a gRPC request only if its service
 // is `foo` AND it contains the `version: v1` header:
 //
-// ```
-// matches:
-// - method:
-//    type: Exact
-//    service: "foo"
-//   headers:
-//   - name: "version"
-//     value "v1"
-// ```
+//	```
+//	matches:
+//	- method:
+//	   type: Exact
+//	   service: "foo"
+//	  headers:
+//	  - name: "version"
+//	    value "v1"
+//	```
 type GRPCRouteMatch struct {
 	// Path specifies a gRPC request service/method matcher. If this field is not
 	// specified, all services and methods will match.
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -340,9 +340,9 @@ const (
 //
 // Invalid values include:
 //
-// * ":method" - ":" is an invalid character. This means that HTTP/2 pseudo
-//   headers are not currently supported by this type.
-// * "/invalid" - "/" is an invalid character
+//   - ":method" - ":" is an invalid character. This means that HTTP/2 pseudo
+//     headers are not currently supported by this type.
+//   - "/invalid" - "/" is an invalid character
 //
 // +kubebuilder:validation:MinLength=1
 // +kubebuilder:validation:MaxLength=256
@@ -495,11 +495,13 @@ const (
 //
 // ```
 // match:
-//   path:
-//     value: "/foo"
-//   headers:
-//   - name: "version"
-//     value "v1"
+//
+//	path:
+//	  value: "/foo"
+//	headers:
+//	- name: "version"
+//	  value "v1"
+//
 // ```
 type HTTPRouteMatch struct {
 	// Path specifies a HTTP request path matcher. If this field is not
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -40,7 +40,7 @@ var (
 
 // ValidateGateway validates gw according to the Gateway API specification.
 // For additional details of the Gateway spec, refer to:
-//  https://gateway-api.sigs.k8s.io/v1beta1/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway
+// https://gateway-api.sigs.k8s.io/v1beta1/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway
 //
 // Validation that is not possible with CRD annotations may be added here in the future.
 // See https://github.com/kubernetes-sigs/gateway-api/issues/868 for more information.
--- a/cmd/admission/main.go
+++ b/cmd/admission/main.go
@@ -27,6 +27,7 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 
 	"k8s.io/klog/v2"
 
@@ -72,7 +73,8 @@ func main() {
 	server := &http.Server{
 		Addr: ":8443",
 		// Require at least TLS12 to satisfy golint G402.
-		TLSConfig: &tls.Config{MinVersion: tls.VersionTLS12, Certificates: []tls.Certificate{certs}},
+		TLSConfig:         &tls.Config{MinVersion: tls.VersionTLS12, Certificates: []tls.Certificate{certs}},
+		ReadHeaderTimeout: 16 * time.Second,
 	}
 	mux := http.NewServeMux()
 	mux.HandleFunc("/validate", admission.ServeHTTP)
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,17 @@ uninstall:
 verify:
 	hack/verify-all.sh -v
 
+# Run Go linter
+.PHONY: lint
+lint:
+	hack/verify-golint.sh
+
+# Run Go linter auto-fixers if supported
+.PHONY: lint-fix
+lint-fix:
+	hack/verify-golint.sh --fix
+
+
 # Build the documentation.
 .PHONY: docs
 docs:
--- a/apis/v1alpha2/referencegrant_types.go
+++ b/apis/v1alpha2/referencegrant_types.go
@@ -40,7 +40,6 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 // grant allowed.
 //
 // Support: Core
-//
 type ReferenceGrant struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/apis/v1alpha2/validation/gateway.go
+++ b/apis/v1alpha2/validation/gateway.go
@@ -40,7 +40,7 @@ var (
 
 // ValidateGateway validates gw according to the Gateway API specification.
 // For additional details of the Gateway spec, refer to:
-//  https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.Gateway
+// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.Gateway
 //
 // Validation that is not possible with CRD annotations may be added here in the future.
 // See https://github.com/kubernetes-sigs/gateway-api/issues/868 for more information.
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -334,9 +334,9 @@ type RouteStatus struct {
 // Hostname is the fully qualified domain name of a network host. This matches
 // the RFC 1123 definition of a hostname with 2 notable exceptions:
 //
-// 1. IPs are not allowed.
-// 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
-//    label must appear by itself as the first label.
+//  1. IPs are not allowed.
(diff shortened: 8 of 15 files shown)


akankshakumari393:
Could we rebase the PR? I see that there are other changes as well which I think are not related to this PR

robscott:
@oscr Thanks for filing this PR! I think the presubmit is failing because you need to run `make generate` to update the  generated CRD yamls.

```
 M config/crd/experimental/gateway.networking.k8s.io_grpcroutes.yaml
 M config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
 M config/crd/experimental/gateway.networking.k8s.io_tlsroutes.yaml
 M config/crd/standard/gateway.networking.k8s.io_httproutes.yaml
```

oscr:
@robscott Thank you! That seems to have solved it.

apis/v1beta1/httproute_types.go
```diff
@@ -495,11 +495,13 @@ const (
 //
 // ```
 // match:
-//   path:
-//     value: "/foo"
-//   headers:
-//   - name: "version"
-//     value "v1"
+//
+//	path:
```
robscott, line 499: Is this additional indentation required to pass the updated linter?

apis/v1beta1/validation/gateway.go
```diff
@@ -40,7 +40,8 @@ var (
 
 // ValidateGateway validates gw according to the Gateway API specification.
 // For additional details of the Gateway spec, refer to:
-//  https://gateway-api.sigs.k8s.io/v1beta1/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway
+//
+//	https://gateway-api.sigs.k8s.io/v1beta1/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway
```
robscott, line 44: This indentation and extra line break seem like they shouldn't be necessary, is this required for the linter update?

cmd/admission/main.go
```diff
@@ -72,7 +73,8 @@ func main() {
 	server := &http.Server{
 		Addr: ":8443",
 		// Require at least TLS12 to satisfy golint G402.
-		TLSConfig: &tls.Config{MinVersion: tls.VersionTLS12, Certificates: []tls.Certificate{certs}},
+		TLSConfig:         &tls.Config{MinVersion: tls.VersionTLS12, Certificates: []tls.Certificate{certs}},
+		ReadHeaderTimeout: 2 * time.Second,
```
robscott, line 77: What is the purpose of this change?

robscott (commented):
Thanks @oscr!

apis/v1beta1/httproute_types.go
oscr: I invoked golangci-lint with fix in in order to resolve the findings. I didn't add a Makefile target for it but it's quite useful. For example cluster-api has it: https://github.com/kubernetes-sigs/cluster-api/blob/1917d52af16a7df70c5e93c9bb0820d5c04f972f/Makefile#L478

apis/v1beta1/validation/gateway.go
oscr: Good catch. This looks strange and when I deleted the extra line and tab indent the linter didn't mind.

cmd/admission/main.go
oscr: There is the finding that the change fixes

```
cmd/admission/main.go:72:13: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
	server := &http.Server{
		Addr: ":8443",
		// Require at least TLS12 to satisfy golint G402.
		TLSConfig: &tls.Config{MinVersion: tls.VersionTLS12, Certificates: []tls.Certificate{certs}},
	}
make: *** [Makefile:118: lint] Error 1

```

robscott: Good catch, that makes sense. I'm not sure what a good value is here, but 2 seconds seems like it might be on the low side. Would 5 seconds make sense?

robscott (commented):
Thanks for the work on this @oscr!

apis/v1beta1/httproute_types.go
oscr: @robscott I can add the lint-fix target. No problem at all. 

Given how gofmt insists on the current format is incorrect and want's the code reformatted, maybe we should just disable that specific linter and undo the changes? Whats your opinion on that?

cmd/admission/main.go
oscr: There seems to be an interesting spread here: cluster-api uses 2 seconds, k/k uses 32. I have no objection to 5 seconds.

robscott: Wow, that is quite the difference. Maybe we should just meet in the middle with something like 16? It does seem unlikely for this to be an issue with any of the suggested values, just don't want this to end up unexpectedly breaking someone.

apis/v1beta1/httproute_types.go
robscott: I'm not sure, disabling all of gofmt for this relatively niche use case does seem unfortunate. We could possibly make an exception like this for gofmt that only applied to our API type definitions: https://github.com/kubernetes-sigs/gateway-api/blob/main/.golangci.yml#L42. Interested in what others think here as well.

/cc @shaneutt @youngnick

cmd/admission/main.go
oscr: Sounds great. That should be plenty of time.

apis/v1beta1/httproute_types.go
oscr: To ensure the changes to api where from the golangci-linter I checked out main and ran it again.

```shell
$ git checkout main apis
Updated 6 paths from 6e0a1739
$ make lint
hack/verify-golint.sh
(warnings omitted...)
apis/v1beta1/httproute_types.go:343: File is not `gofmt`-ed with `-s` (gofmt)
// * ":method" - ":" is an invalid character. This means that HTTP/2 pseudo
//   headers are not currently supported by this type.
// * "/invalid" - "/" is an invalid character
apis/v1beta1/shared_types.go:337: File is not `gofmt`-ed with `-s` (gofmt)
// 1. IPs are not allowed.
// 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
//    label must appear by itself as the first label.
apis/v1alpha2/grpcroute_types.go:277: File is not `gofmt`-ed with `-s` (gofmt)
// - method:
//    type: Exact
//    service: "foo"
//   headers:
make: *** [Makefile:118: lint] Error 1
$ make lint-fix
(warnings omitted...)
$ make lint
hack/verify-golint.sh
(warnings omitted...)
```

oscr: So the gofmt linter fails as is in main.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -274,12 +274,13 @@ type GRPCRouteRule struct {
...
-// - method:
-//    type: Exact
-//    service: "foo"
-//   headers:
+//   - method:
+//     type: Exact
+//     service: "foo"
+//     headers:
 //   - name: "version"
 //     value "v1"
```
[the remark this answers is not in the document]
oscr: Discussed below

apis/v1beta1/httproute_types.go
shaneutt: I wonder if there's some prior art somewhere nearby in the sigs for avoiding this yaml problem :thinking:

oscr: There doesn't seem to be any specific configuration to manage this. For example: 
https://github.com/kubernetes-sigs/kubebuilder/blob/master/.golangci.yml
https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci.yml

I assume they follow the linters suggestions

oscr: https://github.com/kubernetes-sigs/kubebuilder/blob/f4dbb5c82e436c6943d81abb2bf0b5b2b3478921/docs/book/src/cronjob-tutorial/running-webhook.md?plain=1#L37-L39

This is maybe going a different way. But I like this approach by Kubebuilder.

oscr:
Newer version released. Updating pr with latest and greatest.

apis/v1beta1/httproute_types.go
robscott: I think we're stuck between two less than ideal options:

**1. Don't run gofmt and goimports on our types files**
This can be accomplished pretty easily with the following addition to `exclude-rules`:
```
    - path: _types\.go
      linters:
        - gofmt
        - goimports
```

It allows us to keep inline YAML examples in our godocs and therefore generated reference docs. 

**2. Remove YAML examples from inline docs**
Although this may make our generated reference docs slightly harder to follow, it would ensure that our types files were well formatted.

**3. Don't upgrade golangci-lint yet**
This appears to pass with our current generation of golangci-lint which is only slightly older than this one.


At this point, I don't love any of the options, but am slightly leaning towards 3 here unless we can find a better alternative than 1 or 2.

oscr: I think one thing to remember here is that many editors will format on save for Go. I can't remember explicitly enabling it (I might be wrong here) in IDEA but I nevertheless I have it. If the linter expects a certain indentation, but your tooling automatically reformats the code upon save, it's going to be a pain for contributors. I think there is a risk of many prs failing the lint step in the future. 

Ultimately it's your call. I can fix 1, 2 or close it for 3.

seankhliao: it's possible to embed yaml, just the entire block needs to be indented to conform to gofmt's notion of a code block

seankhliao: From https://go.dev/doc/comment
indenting all the yaml by a tab should make it a preformatted code block

oscr: Awesome find @seankhliao! Thank you so for your help here. I've updated the indents with tabs and at least locally it doesn't reformat or linting fail.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -272,15 +272,15 @@ type GRPCRouteRule struct {
...
-// ```
+//	```
+//	matches:
+//	- method:
+//	   type: Exact
+//	   service: "foo"
+//	  headers:
+//	  - name: "version"
+//	    value "v1"
+//	```
```
robscott, line 283: Unfortunately this didn't translate well to the generated reference docs 😢. https://deploy-preview-1416--kubernetes-sigs-gateway-api.netlify.app/references/spec/#gateway.networking.k8s.io%2fv1alpha2.GRPCRouteMatch. Not sure what the solution is, there likely is some combination of indentation that works well for most. I have noticed that inline yaml examples only seem to fail gofmt on top level structs/comments were the actual `//` is not indented.

oscr: Unfortunate, but a good find with the top level comments. Is that a viable way forward? Otherwise I suggest closing the pr because it doesn't seem to be possible to move forward as is.

PR #1419, opened 2022-09-30
Improve the Gateway `ResolvedRefs` condition documentation

### LINKED ISSUE #1362: Clarify RefNotPermitted and InvalidCertificateRef reasons in API Spec
As a follow up to #1297, @skriss suggested that we should clarify this behavior in the API docs.

- currently the RefNotPermitted reason only talks about routes and backends - this should be expanded to include listeners and secrets, or maybe remove references to specific resource types altogether
- we should clarify the scope of InvalidCertificateRef, exactly what an "invalid" certificate ref is, and that it excludes this case of a disallowed cross-namespace reference

mlavacca:
**What type of PR is this?**

/kind documentation

The `ListenerReasonRefNotPermitted` documentation has been improved to specify that this reason can be triggered by either an unpermitted `BackendRef` or an unpermitted `CertificateRef`. Besides, the `ListenerReasonInvalidCertificateRef ` documentation has been improved to specify better what makes a listener `CertificateRef` invalid

Fixes #1362 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -397,17 +397,25 @@ const (
 
 	// This reason is used with the "ResolvedRefs" condition when the
 	// Listener has a TLS configuration with at least one TLS CertificateRef
-	// that is invalid or cannot be resolved.
+	// that is invalid or does not exist.
+	// A CertificateRef is considered invalid when it refers to a nonexistent
+	// or unsupported resource or kind, or when the data within that resource
+	// is malformed.
+	// This reason must be used only when the reference is allowed, either by
+	// referencing an object in the same namespace as the Gateway, or when
+	// a cross-namespace reference has been explicitly allowed by a ReferenceGrant.
+	// If the reference is not allowed, the reason RefNotPermitted must be used
+	// instead.
 	ListenerReasonInvalidCertificateRef ListenerConditionReason = "InvalidCertificateRef"
 
 	// This reason is used with the "ResolvedRefs" condition when an invalid or
 	// unsupported Route kind is specified by the Listener.
 	ListenerReasonInvalidRouteKinds ListenerConditionReason = "InvalidRouteKinds"
 
-	// This reason is used with the "ResolvedRefs" condition when
-	// one of the Listener's Routes has a BackendRef to an object in
-	// another namespace, where the object in the other namespace does
-	// not have a ReferenceGrant explicitly allowing the reference.
+	// This reason is used with the "ResolvedRefs" condition when the
+	// Listener has a TLS configuration that references an object in another
+	// namespace, where the object in the other namespace does not have a
+	// ReferenceGrant explicitly allowing the reference.
 	ListenerReasonRefNotPermitted ListenerConditionReason = "RefNotPermitted"
 )
 
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -747,17 +747,25 @@ const (
 
 	// This reason is used with the "ResolvedRefs" condition when the
 	// Listener has a TLS configuration with at least one TLS CertificateRef
-	// that is invalid or cannot be resolved.
+	// that is invalid or does not exist.
+	// A CertificateRef is considered invalid when it refers to a nonexistent
+	// or unsupported resource or kind, or when the data within that resource
+	// is malformed.
+	// This reason must be used only when the reference is allowed, either by
+	// referencing an object in the same namespace as the Gateway, or when
+	// a cross-namespace reference has been explicitly allowed by a ReferenceGrant.
+	// If the reference is not allowed, the reason RefNotPermitted must be used
+	// instead.
 	ListenerReasonInvalidCertificateRef ListenerConditionReason = "InvalidCertificateRef"
 
 	// This reason is used with the "ResolvedRefs" condition when an invalid or
 	// unsupported Route kind is specified by the Listener.
 	ListenerReasonInvalidRouteKinds ListenerConditionReason = "InvalidRouteKinds"
 
-	// This reason is used with the "ResolvedRefs" condition when
-	// one of the Listener's Routes has a BackendRef to an object in
-	// another namespace, where the object in the other namespace does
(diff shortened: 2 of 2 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -397,17 +397,18 @@ const (
...
 	ListenerReasonInvalidRouteKinds ListenerConditionReason = "InvalidRouteKinds"
 
 	// This reason is used with the "ResolvedRefs" condition when
-	// one of the Listener's Routes has a BackendRef to an object in
-	// another namespace, where the object in the other namespace does
-	// not have a ReferenceGrant explicitly allowing the reference.
+	// one of the Listener's Routes references an object in another namespace,
+	// where the object in the other namespace does not have a ReferenceGrant
+	// explicitly allowing the reference. The unpermitted reference can be
+	// either a BackendRef, or a CertificateRef.
```
robscott, line 411: A couple thoughts that are 

1. It seems strange for this to extend to invalid references in attached routes. Maybe @youngnick has been thinking about this as part of the broader conditions update he's working on?
2. It seems like cross-namespace references that are not allowed by a ReferenceGrant should always receive the RefNotPermitted reason, regardless of if the Secret actually exists.

apis/v1alpha2/gateway_types.go
```diff
@@ -397,17 +397,18 @@ const (
 
 	// This reason is used with the "ResolvedRefs" condition when the
 	// Listener has a TLS configuration with at least one TLS CertificateRef
-	// that is invalid or cannot be resolved.
+	// that is malformed or does not exist.
```
robscott, line 400: I'm not sure that "malformed" is clearer than "invalid" here. Trying to think through what could possibly cause this and it seems like it would either be:

- Not Found
- Nonexistent Kind
- Unsupported Kind

Anything else?

robscott (commented):
Thanks @mlavacca! Would like to get @youngnick's review on this as well.

/cc @youngnick

robscott:
@mlavacca Can you change the PR title to be a bit more descriptive? Easy to lose track of this PR when looking through the list.

apis/v1alpha2/gateway_types.go
mlavacca: It could also be used when the referenced secret data is invalid (it doesn't contain the data expected by the controller, or those data are malformed somehow).

mlavacca: would you highlight these reasons specifically in the docs, instead of putting a generic `invalid` or `malformed` that isn't too meaningful?

apis/v1alpha2/gateway_types.go
mlavacca: > 2. It seems like cross-namespace references that are not allowed by a ReferenceGrant should always receive the RefNotPermitted reason, regardless of if the Secret actually exists.

This is what I had in mind when I wrote this part. In this case, we should decide which reason has the highest severity:  `RefNotPermitted` or `InvalidCertificateRef`. In my opinion, if no reference is allowed, the reason should be`RefNotPermitted`, no matter if the secret exists or is valid.

robscott: I agree with what you're suggesting in this comment, I think this PR still needs some work to make that clear in the spec.

apis/v1alpha2/gateway_types.go
robscott: I think the ideal would be to add a sentence at the end of this paragraph that goes something like this: "For example, a CertificateRef is considered invalid when it refers to a nonexistent or unsupported resource or kind, or when the data within that resource is malformed."

apis/v1alpha2/gateway_types.go
mlavacca: > 1. It seems strange for this to extend to invalid references in attached routes. Maybe @youngnick has been thinking about this as part of the broader conditions update he's working on?

I just figured out that the previous comment here was copy-paste from the `RouteReasonRefNotPermitted`: https://github.com/kubernetes-sigs/gateway-api/blob/1e037bb54ca1051023c0e9b90f56cdf1ebf0adce/apis/v1beta1/shared_types.go#L247-L251

And it doesn't make any sense here; it misled me. The only reason to have `RefNotPermitted` should be the reference to an ungranted secret. So I'm going to fix it.

mlavacca: I addressed both your points, it should be clearer now.

apis/v1alpha2/gateway_types.go
```diff
@@ -397,17 +397,24 @@ const (
...
 	// Listener has a TLS configuration with at least one TLS CertificateRef
-	// that is invalid or cannot be resolved.
+	// that is invalid or does not exist.
+	// A CertificateRef is considered invalid when it refers to a nonexistent
+	// or unsupported resource or kind, or when the data within that resource
+	// is malformed.
+	// This reason must be used only when the reference is granted, which means
+	// that if the referenced secret is invalid or does not exist, but the
+	// reference is not allowed because of the lack of a ReferenceGrant, the
+	// reason RefNotPermitted must be used.
```
mikemorris, line 407: If this suggestion is accepted, it should be applied to the v1beta1 APIs too.

```suggestion
	// This reason must be used only when the reference is allowed, either by
	// referencing an object in the same namespace as the Gateway, or when
	// a cross-namespace reference has been explicitly allowed by a ReferenceGrant.
	// If the reference is not allowed, the reason RefNotPermitted must be used
	// instead.
```

mikemorris (commented):
The `ListenerReasonRefNotPermitted` changes LGTM, and I proposed a phrasing tweak to hopefully clarify `ListenerReasonInvalidCertificateRef` a bit more.

Thanks for your work on this @mlavacca, this feels pretty close now and a good clarification!

apis/v1alpha2/gateway_types.go
mlavacca: done

mikemorris (approved):
/lgtm

robscott (approved):
Thanks @mlavacca, this is a great improvement!

/approve

PR #1426, opened 2022-10-03
GEP: add HTTPRoute Mesh Binding proposal

### LINKED ISSUE #1294: GEP: Mesh Service Binding
As the GAMMA initiative seeks to create a model for service [mesh](https://docs.google.com/document/d/1T_DtMQoq2tccLAtJTpo3c0ohjm25vRS35MsestSL9QU/edit#heading=h.fv9zxi8pjsiq) implementations within Gateway API, it is necessary to have a mechanism for [binding](https://docs.google.com/document/d/1T_DtMQoq2tccLAtJTpo3c0ohjm25vRS35MsestSL9QU/edit#heading=h.qerd5nlyuzyt) routes to services

This GEP covers the work to design and implement this mechanism.. 

**Why this is needed**:

As @shaneutt originally suggested, the first step will be implementing a provisional GEP that sets out the terms of what we're doing - the "What" and "Why" of the solution, and then once we're on the same page, we'll talk more about the "How".

Why this is needed:
Without a way to bind routes to services, no further work can continue

mikemorris:
**What type of PR is this?**
/kind feature
/kind gep
/area mesh

Proposes an approach for service owners to configure service mesh traffic routing rules using an `HTTPRoute` with a Kubernetes `Service` resource as a `parentRef`.

Fixes #1294 

```release-note
NONE
```


site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,383 @@
...
+
+`Service` resource with [`type: NodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) or [`type: LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) MUST NOT be allowed as `parentRefs` or `backendRefs`. It should isntead be preferred to expose services publicly through the North/South Gateway API interfaces.
+
+`Services` with [`type: ExternalName`](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) SHOULD NOT be allowed as `parentRefs` or `backendRefs` due to [security concerns](https://github.com/kubernetes/kubernetes/issues/103675), although might eventually play some role in [configuring egress functionality](https://github.com/kubernetes-sigs/gateway-api/issues/1070).
+
+`Service` resources supported as `backendRefs` SHOULD be consistent with expectations for North/South Gateway API implementations, MUST have associated endpoints, and include:
```
mikemorris, line 100:
```suggestion
`Service` resources supported as `backendRefs` SHOULD be consistent with expectations for North/South Gateway API implementations, MUST have associated endpoints, and include:

```

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Omitting `backendRefs`
+
+Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` `Service`. If the `parentRef` does not have any endpoints (e.g. if it is a `Service` without selectors), implementations MUST return an HTTP 503 response code (see discussion in https://github.com/kubernetes-sigs/gateway-api/pull/1210).
```
mikemorris, line 72:
```suggestion
Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` `Service`. If the `parentRef` does not have any endpoints (e.g. if it is a `Service` without selectors), implementations MUST return an HTTP 503 response code (see discussion in [#1210](https://github.com/kubernetes-sigs/gateway-api/pull/1210)).
```

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+Introduce a new resource to represent the "frontend" role of a service as defined in [GEP-1291: Mesh Representation](https://docs.google.com/document/d/1oyA9uUH7pNNxxwy3WZGSWx-edHDBLrujcezr8q3el70/edit#).
+
+#### Controller manages new DNS hostname
+
+A controller could create a matching selector-less `Service` (i.e. no endpoints), to create a `.cluster.local` name, or could interact with [external-dns](https://github.com/kubernetes-sigs/external-dns) to create a DNS name in an owned domain.
+
+Ownership/trust would remain based on naming pattern: `serviceName.namespace.svc.[USER_DOMAIN]`
+
+Separate `HttpService`, `TlsService` and `TcpService` resources could have the benefit of allowing us to define protocol specific elements to the spec along with an embedded `CommonServiceSpec`, similar to `CommonRouteSpec`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.CommonRouteSpec), and keep similar patterns as `Service`.
```
mikemorris, line 159:
```suggestion
Separate `HttpService`, `TlsService` and `TcpService` resources could have the benefit of allowing us to define protocol specific elements to the spec along with an embedded `CommonServiceSpec`, similar to [`CommonRouteSpec`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.CommonRouteSpec), and keep similar patterns as `Service`.
```

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+        value: "/bar"
+```
+
+The example above would drop all incoming traffic for HTTP paths other than `/bar` to the `foo` `Service`, following the existing spec to return HTTP 404 response codes for unmatched requests, and HTTP 500 response codes for requests excluded due to an [`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).
+
+### Allowed `Service` types
+
+Valid `Service` resources to be selected as a `parentRef` SHOULD provision one or more virtual IP(s) and DNS hostname(s) within the Kubernetes cluster, and include the default [`ClusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) `Service` type, with or without selectors.
+
+`Service` resource with [`type: NodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) or [`type: LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) MUST NOT be allowed as `parentRefs` or `backendRefs`. It should isntead be preferred to expose services publicly through the North/South Gateway API interfaces.
```
mikemorris, line 96:
```suggestion
`Service` resource with [`type: NodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) or [`type: LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) MAY be allowed as `parentRefs` or `backendRefs`, as these do provision virtual IPs and are effectively `ClusterIP` services with additional functionality, but it should typically be preferred to expose services publicly through the North/South Gateway API interfaces instead.
```

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+`Services` with [`type: ExternalName`](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) SHOULD NOT be allowed as `parentRefs` or `backendRefs` due to [security concerns](https://github.com/kubernetes/kubernetes/issues/103675), although might eventually play some role in [configuring egress functionality](https://github.com/kubernetes-sigs/gateway-api/issues/1070).
+
+`Service` resources supported as `backendRefs` SHOULD be consistent with expectations for North/South Gateway API implementations, MUST have associated endpoints, and include:
+
+* `ClusterIP` `Services` with selectors
+* [Headless Services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) with selectors. Headless services without selectors SHOULD NOT be supported due to the possibility of matching `Services` with `type: ExternalName`.
+* `StatefulSet` resources
```
mikemorris, line 104:
```suggestion
* `StatefulSet` resources, by targeting the associated headless service
```
@costinm you mentioned this may be difficult for some meshes to support, should we instead remove `StatefulSet` from this section and just keep it as a MAY further down?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
+# GEP-1426: HTTPRoute Mesh Binding
+
+* Issue: [#1294](https://github.com/kubernetes-sigs/gateway-api/issues/1294)
+* Status: Implementable
+
+## Overview
+
+Similar to how `HTTPRoutes` bind to `Gateways` and mange North/South traffic flows in Gateway API’s ingress use-case, it would be natural to adopt a similar model for traffic routing concerns in service mesh deployments. The purpose of this GEP is to add a mechanism to the Gateway API spec for the purpose of associating `HTTPRoutes` to a service mesh and offering a model for service owners to manage traffic splitting configurations.
```
mikemorris, line 8:
```suggestion
Similar to how `HTTPRoutes` bind to `Gateways` and mange North/South traffic flows in Gateway API’s ingress use-case, it would be natural to adopt a similar model for traffic routing concerns in service mesh deployments. The purpose of this GEP is to add a mechanism to the Gateway API spec for the purpose of associating `HTTPRoutes` to a service mesh and offering a model for service owners to manage traffic splitting configurations.

This GEP is intended to establish an implementable baseline for supporting basic service mesh traffic routing functionality through the Gateway API spec.
```
@robscott thinking about adding this to help clarify your question about when GAMMA might be considered "implementable" - my hope is that advancing this GEP would be enough to get started, even as we continue to work on increasing the scope of supported functionality.

Thoughts @howardjohn @keithmattix? Is it too early to add this, or would adding a caveat about enforcing AuthZ through implementation-specific configuration until supported in Gateway API be sufficient?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* NOT already specified as a `backendRef` of any _other_ `HTTPRoute` with a `Service` `parentRef`
+    * A `Service` MAY still be listed as a `backendRef` on the _same_ `HTTPRoute` where it is specified as a `parentRef`, e.g. to only split off some subset of traffic from the default backend endpoints.
+    * This could be computationally expensive to validate.
+
+This should be enforced by controller validation, which should set an `Accepted: { status: False }` condition with a corresponding new `ParentRefConflict` `RouteConditionReason`.
+
+The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
```
mikemorris, line 137: Would definitely appreciate 👀 on this section to make sure the scope of these rules feels reasonable and sufficient to cover the goal of allowing a Service to both have E/W mesh routing rules applied and be accessible through a Gateway concurrently.

site-src/geps/gep-1424.md
kflynn: I would vote for being restrictive at first and opening things up as we find use cases for them -- my gut feel is that I'd much rather handle ingress with an ingress controller.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+This GEP uses the [roles and personas](https://gateway-api.sigs.k8s.io/concepts/security-model/#roles-and-personas) defined in the Gateway API security model.
+
+## Goals
+
+* SHOULD allow `HTTPRoute` traffic rules to be configurable for a mesh service by the application owner/producer.
+* SHOULD allow control by the cluster operator (mesh administrator) to grant permission for whether `HTTPRoute` resources in a given namespace are allowed to configure mesh traffic routing.
+* MUST only be allowed to configure "producer" traffic rules for a `Service` in the same namespace as the `HTTPRoute`.
+    * Traffic routing configuration defined in this way SHOULD be respected by ALL consumer services in all namespaces in the mesh.
+* SHOULD only be allowed to direct traffic to `backendRefs` within the same namespace when used for configuration of mesh traffic.
```
howardjohn, line 20: Fine with this short term but should be opened up in a fast followup IMO

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* Handling East/West traffic outside the cluster (VMs, etc).
+
+## Introduction
+
+It is proposed that an application owner should configure traffic rules for a mesh service by configuring an `HTTPRoute` with a Kubernetes `Service` resource as a `parentRef`.
+
+This approach is dependent on both the "frontend" role of the Kubernetes `Service` resource as defined in [GEP-1324: Service Mesh in Gateway API](https://gateway-api.sigs.k8s.io/geps/gep-1324/#service) when used as a `parentRef` and the "backend" role of `Service` when used as a `backendRef`. It would use the Kubernetes DNS name to match traffic for meshes implementing "transparent proxy" functionality, but the `backendRef` endpoints would ultimately be used for the canonical IP address(es) to which traffic should be redirected by rules defined in this HTTPRoute. This approach leverages the existing points of extensibility within the Gateway API spec, and would not require introducing any API changes or new resources, only defining expected behavior.
```
howardjohn, line 46:
```suggestion
This approach is dependent on both the "frontend" role of the Kubernetes `Service` resource as defined in [GEP-1324: Service Mesh in Gateway API](https://gateway-api.sigs.k8s.io/geps/gep-1324/#service) when used as a `parentRef` and the "backend" role of `Service` when used as a `backendRef`. It would use the Kubernetes service name to match traffic for meshes implementing "transparent proxy" functionality, but the `backendRef` endpoints would ultimately be used for the canonical IP address(es) to which traffic should be redirected by rules defined in this HTTPRoute. This approach leverages the existing points of extensibility within the Gateway API spec, and would not require introducing any API changes or new resources, only defining expected behavior.
```

Without further explanation, "DNS" is confusing here.

What we are describing is "Calls to service foo". For `Service`, that is identified by the ClusterIP of `foo`. For 99.99% of users, they access this by doing a DNS lookup for `foo` (or variants); but the API itself has no direct relation to DNS, IMO.

Might be worth calling this out (unless we do later on) to clarify, since it is subtle

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+    backendRefs:
+    - kind: Service
+      name: foo
+      weight: 90
+    - kind: Service
+      name: foo_v2
+      weight: 10
+```
+
+In the example above, routing rules have been configured to direct 90% of traffic for the `foo` `Service` to the default "backend" endpoints specified by the `foo` `Service` [`selector`](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) field, and 10% to the `foo_v2` `Service`. This is determined based on the `ClusterIP` (for `Service`) and `ClusterSetIP` (for `ServiceImport`) matching, and for "transparent proxy" mesh implementations would match all requests to `foo.svc.cluster.local` (or arbitrary custom suffix, as the hostname is not specified manually) from within the same namespace, all requests to `foo.store.svc.cluster.local` from other namespaces, and all requests to `foo.store.svc.clusterset.local` for multicluster services, within the scope of the service mesh.
```
howardjohn, line 68: nit: I think its worth saying "This means we are matching the service" then "services can be arbitrary things... if the are Service <behavior>, if ServiceImport <behavior>, else implementation specific".

Just to make it clear that it is one of those things that has arbitrary inputs but for the well-known types we have explicit behavior

howardjohn: I guess we basically say that here, so not sure what my request is :-)

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Omitting `backendRefs`
+
+Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` `Service`. If the `parentRef` does not have any endpoints (e.g. if it is a `Service` without selectors), implementations MUST return an HTTP 503 response code (see discussion in [#1210](https://github.com/kubernetes-sigs/gateway-api/pull/1210)).
```
howardjohn, line 72: "If the `parentRef` does not have any endpoints (e.g. if it is a `Service` without selectors)..."

This part seems correct but confusing. The behavior should be exactly the same as if `backendRefs` was *not* elided. Which I think you have described, but it sounds different.

And "without selectors" seems wrong, you can have endpoints without selectors.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+  parentRefs:
+  - kind: Service
+    name: foo
+  rules:
+    matches:
+    - path:
+        value: "/bar"
+```
+
+The example above would drop all incoming traffic for HTTP paths other than `/bar` to the `foo` `Service`, following the existing spec to return HTTP 404 response codes for unmatched requests, and HTTP 500 response codes for requests excluded due to an [`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).
```
howardjohn, line 90: This is super subtle - I think its correct but should be more explicit.

I think this is saying:

* All traffic implicitly works without routes (this is just how K8s works)
* When you create a route you are *replacing* that implicit logic - not adding to it
* Therefor you may be restricting traffic via a route

Note: restriction != authz

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Allowed `Service` types
+
+Valid `Service` resources to be selected as a `parentRef` SHOULD provision one or more virtual IP(s) and DNS hostname(s) within the Kubernetes cluster, and include the default [`ClusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) `Service` type, with or without selectors.
```
howardjohn, line 94:
```suggestion
Valid `service` resources to be selected as a `parentRef` SHOULD provision one or more virtual IP(s) and DNS hostname(s) within the Kubernetes cluster, and include the default [`ClusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) `Service` type, with or without selectors.
```

anywhere we are refering to the bag of valid "services" we should probably not use `Service`?

howardjohn: But besides terminology - do we want to be that prescriptive, or just that that "services should have a way to identify traffic to them (typically by a virtual IP or name)".

for example, we may have a `S3BucketRoute` (weird, but proves a point), that is identified by the name of the bucket - not an IP or DNS name. The important thing is a proxy can identify the traffic, not _how_

site-src/geps/gep-1424.md
howardjohn: I would allow them. These are supersets of ClusterIP types.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* [Headless Services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) with selectors. Headless services without selectors SHOULD NOT be supported due to the possibility of matching `Services` with `type: ExternalName`.
+* `StatefulSet` resources
+
+#### `Service` without selectors
+
+An alternate pattern additionally supported by this approach would be to target a `Service` without selectors as the `parentRef`. This could be a clean way to create a pure routing construct and abstract a logical frontend, as traffic would resolve to a `backendRef` `Service` with selectors defined on the `HTTPRoute`, or receive a 4xx/5xx error response if no matching path or valid backend was found.
+
+#### StatefulSet
+
+[`StatefulSet`](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) resources MAY be supported as a `backendRef`, but specifics of this integration have not yet been clarified in the full Gateway API spec.
```
robscott, line 112: I don't really understand the purpose of this, I think it would be safest to remove it unless there are well defined use cases we need to support.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+`Service` resources supported as `backendRefs` SHOULD be consistent with expectations for North/South Gateway API implementations, MUST have associated endpoints, and include:
+
+* `ClusterIP` `Services` with selectors
+* [Headless Services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) with selectors. Headless services without selectors SHOULD NOT be supported due to the possibility of matching `Services` with `type: ExternalName`.
+* `StatefulSet` resources
+
+#### `Service` without selectors
+
+An alternate pattern additionally supported by this approach would be to target a `Service` without selectors as the `parentRef`. This could be a clean way to create a pure routing construct and abstract a logical frontend, as traffic would resolve to a `backendRef` `Service` with selectors defined on the `HTTPRoute`, or receive a 4xx/5xx error response if no matching path or valid backend was found.
```
robscott, line 108: How would this work? Services with Selectors make it easier for users to attach their own custom endpoints to the Service, is that a goal here?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Limiting graph depth, preventing conflicts and cycles
+
+A `Service` may only be specified as a `parentRef` of an `HTTPRoute` if it is:
+
+* NOT already specified as a `parentRef` of another `HTTPRoute` [in the same namespace]
+    * This requirement may be relaxed later to follow the existing [precedence rules](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule), but is intended to simplify initial implementation.
```
robscott, line 129: It seems like we may need some kind of conflict resolution logic here similar to what we already have in place for Gateway API.

site-src/geps/gep-1424.md
robscott: For a bit more context, I raised some concerns here because it feels somewhat arbitrary to not support NodePort or LoadBalancer Services. LoadBalancer Services are a superset of NodePort Services which are a superset of ClusterIP Services. This means that we're effectively saying that we don't support a set of Services that happen to have some extra functionality on top. I agree with Mike's suggested change above. Essentially I think our stance should be that:

1. ClusterIP Services are recommended
2. The ClusterIP portion of NodePort and LoadBalancer Services are supported
3. We recommend using ClusterIP Services over alternative types because the additional functionality added by NodePort and LoadBalancer Services will be ignored by mesh implementations.

robscott (commented):
Thanks @mikemorris! Took a first pass through this, it's a great start.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+    * The namespace colocation requirement effectively forecloses using this strategy to delegate route management, but there is service mesh [user interest](https://github.com/istio/istio/issues/22997) in merging separately-defined routes together to manage large configurations, multiple versions or canary deployments for a single mesh service.
+* NOT already specified as a `backendRef` of any _other_ `HTTPRoute` with a `Service` `parentRef`
```
howardjohn, line 131: Intuitively, I don't think we should add this restriction.

My rationale for that depends on the exact semantics of how routes work though. I thought we had defined it but maybe that was in my doc... we will definitely need to say what happens if there is a gateway with a route pointing to the service with route

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
+# GEP-1426: HTTPRoute Mesh Binding
+
+* Issue: [#1294](https://github.com/kubernetes-sigs/gateway-api/issues/1294)
+* Status: Implementable
```
shaneutt, line 4: I am not saying _no_ to marking this _implementable_ right away, but I did want to take a moment to point it out and just ask a couple questions to those who are thinking about building out the implementation. Marking this `Implementable` is meant to indicate that anyone thereafter can make a PR to add the implementation to the API. Is this actually the state we're in, and do we want that? Alternatively, do we want it `Provisional` to give others a bit more time to add more iterative PRs on top of this, perhaps honing in and further expanding and refining some of the key points?

Again, not a blocker because end of the day I don't want to slow down our progress towards an alpha, and even if we had to make significant changes in the alpha that's OK. More just a double tap because once you really get the momentum of an implementation going the refinement process can become a bit slower.

site-src/geps/gep-1424.md
keithmattix: I agree with this change. Targeting a statefulset specifically sounds like the wrong abstraction

site-src/geps/gep-1424.md
keithmattix: This sounds right to me, but would love feedback from others

site-src/geps/gep-1424.md
keithmattix: But this restriction only applies to another HTTPRoute with a Service parentRef (i.e. a mesh route). Therefore, it shouldn't restrict Gateway based HTTPRoutes

site-src/geps/gep-1424.md
mikemorris: I've been thinking that adding a few diagrams of allowed/disallowed topologies to illustrate these rules might be helpful.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* [GEP-1282: Describing Backend Properties](https://gateway-api.sigs.k8s.io/geps/gep-1282/) may be one path to associating a `Service` with a mesh, but likely wouldn't be able to handle the application of multiple `HTTPRoutes` for the same `Service`, but each intended for different mesh implementations
+    * It's currently unclear how relevant this constraint may be, but associating an `HTTPRoute` with a mesh by this method would additionally require an extra graph traversal step.
+* Expecting a `Mesh` `parentRef` or similar reference as proposed in [GEP-1291: Mesh Representation](https://docs.google.com/document/d/1oyA9uUH7pNNxxwy3WZGSWx-edHDBLrujcezr8q3el70/edit#) may be a preferred eventual path forward, but wouldn't be required initially, with the assumption that only one mesh should typically be present in a cluster.
+* No mechanism for egress redirection of traffic from arbitrary hostnames to a mesh service within this approach (but could still be implemented seperately).
+
+## Alternatives
```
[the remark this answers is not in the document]
mikemorris: Agreed, that should be added - it's similar to the nested `parentRefs` approach in functionality, but I like that it inverts it to potentially always have meaningful context (e.g. it could be applicable to "in-mesh" Gateways too)

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
mikemorris: Further discussion in https://github.com/kubernetes-sigs/gateway-api/pull/1426#pullrequestreview-1130314653, feels like there's consensus to only direct traffic to `StatefulSet` indirectly through a headless service.

mikemorris:
```suggestion
```
There's at least three separate threads suggesting to remove this, so I'll get rid of it.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+`Services` with [`type: ExternalName`](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) SHOULD NOT be allowed as `parentRefs` or `backendRefs` due to [security concerns](https://github.com/kubernetes/kubernetes/issues/103675), although might eventually play some role in [configuring egress functionality](https://github.com/kubernetes-sigs/gateway-api/issues/1070).
+
+`Service` resources supported as `backendRefs` SHOULD be consistent with expectations for North/South Gateway API implementations, MUST have associated endpoints, and include:
+
+* `ClusterIP` `Services` with selectors
+* [Headless Services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) with selectors. Headless services without selectors SHOULD NOT be supported due to the possibility of matching `Services` with `type: ExternalName`.
```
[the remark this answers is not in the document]
mikemorris: Ah, I was expecting headless service could be valid as a `backendRef` but NOT as a `parentRef` frontend. I wasn't entirely clear on the headless service without selectors behavior either.

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
mikemorris: Removed.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+#### StatefulSet
+
+[`StatefulSet`](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) resources MAY be supported as a `backendRef`, but specifics of this integration have not yet been clarified in the full Gateway API spec.
+
+#### Multicluster support with `ServiceImport`
+
+`ServiceImport` resources allocate a virtual IP in the cluster, so MAY be allowed as a `parentRef`.
+
+`ServiceImport` would remain a valid backend option for `HTTPRoute` resources (but _not_ currently a requirement for core conformance), and could be specified alongside a `Service` `backendRef` to split traffic across clusters within a `ClusterSet` (as defined in the [Multi-cluster Service (MCS) APIs project](https://github.com/kubernetes-sigs/mcs-api)). This could be a way to solve the need described in [A use case for using Gateway APIs in Multi-Cluster](https://docs.google.com/document/d/1bjr0uAVMmEtTX4mpU_aGXXapQFsEz_Qt0OnDoVswEEI/edit).
```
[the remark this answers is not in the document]
mikemorris: My only caution with speccing `ServiceImport` explicitly as extended conformance is the caveat that all this is still experimental, and I'm not quite sure if `ServiceImport` specifically has had broad enough usage yet to understand the precise expected behavior?

site-src/geps/gep-1424.md
howardjohn: Targetting headless service is not un-ambiguous either btw. Its *possible* to implement, but you can have 2 headless services for 1 pod, and its impossible to distinguish them based on IP ( since there is no VIP ). If we want to allow this, lets clearly document the behavior.

howardjohn: In Istio I think we just pick the oldest or something. its pretty rare to have this scenario

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
howardjohn: I don't have an issue deferring this

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,381 @@
...
+## Goals
+
+* SHOULD allow `HTTPRoute` traffic rules to be configurable for a mesh service by the application owner/producer.
+* SHOULD allow control by the cluster operator (mesh administrator) to grant permission for whether `HTTPRoute` resources in a given namespace are allowed to configure mesh traffic routing.
+* MUST only be allowed to configure "producer" traffic rules for a `Service` in the same namespace as the `HTTPRoute`.
+    * Traffic routing configuration defined in this way SHOULD be respected by ALL consumer services in all namespaces in the mesh.
+* SHOULD only be allowed to direct traffic to `backendRefs` within the same namespace when used for configuration of mesh traffic.
+    * This constraint may be revisited in the future, but is intended to help simplify an initial implementation.
+    * Attempting to specify an `HTTPBackendRef` in a different namespace should raise a status condition on the `HTTPRoute`.
+    * When an `HTTPBackendRef` is invalid, 500 status codes MUST be returned for requests that would have otherwise been routed to an invalid backend.
```
youngnick, line 23: I'd phrase this like so:
```suggestion
    * When a route rule contains a `HTTPBackendRef` that is invalid, 500 status codes MUST be returned for requests that would match that route rule.
```

This is the same behavior as with Gateway attachment, but also makes it clear that this applies to partially-valid HTTPRoutes.

site-src/geps/gep-1424.md
youngnick: I'm in favor of the idea that the services just need a way to identify traffic - Cilium does this slightly differently, and that will cover the way Cilium security identities work.

I'm okay with saying that only Cluster IP is valid _for now_ though.

youngnick (commented):
I'm still not one hundred percent convinced that service binding is the way to go, but I think that this is looking good as a way to prove me wrong. :)

Left some comments, but I don't think there are any blockers from me.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,381 @@
...
+
+### `hostnames` field
+
+GAMMA implementations SHOULD NOT infer any functionality from the `hostnames` field on `HTTPRoute` due to current under-specification and reserved potential for future usage or API changes. For the use case of filtering incoming traffic from selected hostnames, it is recommended to guide users toward configuring [`HTTPHeaderMatch`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderMatch) rules for the [`Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) header. Functionality to be explored in future GEPs may include supporting concurrent usage of an `HTTPRoute` traffic configuration for multiple North/South `Gateways` and East/West mesh use cases or redirection of egress traffic to an in-cluster `Service`.
+
+### Limiting graph depth, preventing conflicts and cycles
+
+A `Service` may only be specified as a `parentRef` of an `HTTPRoute` if it is:
+
+* NOT already specified as a `parentRef` of another `HTTPRoute` [in the same namespace]
```
mateiidavid, line 125: Sorry, bit of a silly question but just wanted to make sure I understand this correctly. There can only be one route resource per `Service`?

steeling:
Adding the document here for posterity. I've collected a set of concerns and suggestions for the current HTTPRoute mesh binding proposal. Tl;dr is I believe that remodeling the definition of a Kubernetes service to split into a frontend and backend is inherently the wrong approach, and we should follow suit with what the multicluster services API is doing by finding another mechanism to register DNS entries and a ClusterIP. More specific reasoning in the document below.

https://docs.google.com/document/d/1yPvLQLTuokU73x_59BKShozBe1fA13xbwaZAWNgjBIg/edit

site-src/geps/gep-1424.md
shaneutt: I don't consider this a blocker, but I would ask that unless there's a strong reason to not mark this as provisional, we mark it provisional before we merge.

shaneutt (approved):
I am approving this with the following conditions:

- please mark the KEP as `provisional` unless there's some really strong reason not to that I'm missing.
- there are a lot of comments here. I would encourage us to think deeply about which things are hard blockers, and which things could be follow ups. Default to creating a follow up issue so we can get this PR through without it becoming even larger _unless absolutely necessary_.

site-src/geps/gep-1424.md
mikemorris:
```suggestion
* NOT already specified as a `backendRef` of any _other_ `HTTPRoute` with a `Service` `parentRef`
   * This restriction is intended to prevent route "inclusion/delegation" patterns or circular references from multiple tiers of routing rules _within mesh configuration_.
   * A `Service` MAY still be listed as a `backendRef` on a different `HTTPRoute` which only targets a `Gateway` as a `parentRef`. The specific behavior of how Gateways might interact with mesh routing rules will be defined in a future GEP.
```

site-src/geps/gep-1424.md
mikemorris: @robscott How does this look?
```suggestion
    * Resolving conflicts from multiple `HTTPRoutes` attempting to target the same `Service` as a `parentRef` should follow the existing [conflict guidelines](https://gateway-api.sigs.k8s.io/concepts/guidelines/#conflicts).
      * When an existing `HTTPRoute` targeting the `Service` as a `parentRef` has already been accepted, any `HTTPRoute` with a newer creation timestamp should be rejected, because the match target considered for the specificity rules (the `Service`) is identical.
      * If creation timestamps are identical (from applying both routes simultaneously), only the `HTTPRoute` with a `name` appearing first in alphabetical order should be accepted.
      * Any rejected `HTTPRoute` should set an `Accepted` condition with `status: "False"`.
    * This requirement may be relaxed later to follow the existing [precedence rules for merging configuration](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule), but is intended to simplify initial implementation.
```

site-src/geps/gep-1424.md
mikemorris: For initial implementation, yes - this will likely be relaxed in the future as described in the notes below because we know there is user interest in splitting large configurations over multiple files.

site-src/geps/gep-1424.md
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1479 to track this.

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1479 to capture this.

site-src/geps/gep-1424.md
mikemorris:
```suggestion
* SHOULD only be allowed to direct traffic to `backendRefs` within the same namespace when used for configuration of mesh traffic, setting a `ResolvedRefs` condition with `status: "False"` and reason `RefNotPermitted` on the `HTTPRoute` if this is attempted.
```

Opened https://github.com/kubernetes-sigs/gateway-api/issues/1480 to capture this longer-term - I'm going to move this out of the Goals section and into a new section for constraints/restrictions if the suggested status condition here seems reasonable.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+## Non-Goals
+
+* Supporting "egress" use cases, which is currently a deferred goal, including:
+    * Defining how "consumer" traffic rules which could override routing for service upstreams only within the local scope of a namespace or service might be configured.
+    * Redirecting calls from arbitrary custom domains to an in-cluster service.
+* Defining how multiple `Services` or `EndpointSlices` representing instances of a single "logical" service should present an identity for AuthN/AuthZ or be associated with each other beyond routing rules.
+* Defining how AuthZ should be implemented to secure East/West traffic between services.
+* Defining how [Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) would bind to `HTTPRoute`, services or a mesh.
+* Defining how `HTTPRoutes` configured for East/West service mesh traffic management might integrate with North/South `Gateways`.
```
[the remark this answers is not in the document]
mikemorris:
```suggestion
* Defining how `HTTPRoutes` configured for East/West service mesh traffic management might integrate with North/South `Gateways`. This will be defined in a future GEP.
```
Opened https://github.com/kubernetes-sigs/gateway-api/issues/1478 to track this, should it be linked here explicitly @robscott?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+Similar to how `HTTPRoutes` bind to `Gateways` and mange North/South traffic flows in Gateway API’s ingress use-case, it would be natural to adopt a similar model for traffic routing concerns in service mesh deployments. The purpose of this GEP is to add a mechanism to the Gateway API spec for the purpose of associating `HTTPRoutes` to a service mesh and offering a model for service owners to manage traffic splitting configurations.
+
+## Personas
+
+This GEP uses the [roles and personas](https://gateway-api.sigs.k8s.io/concepts/security-model/#roles-and-personas) defined in the Gateway API security model.
+
+## Goals
+
+* SHOULD allow `HTTPRoute` traffic rules to be configurable for a mesh service by the application owner/producer.
```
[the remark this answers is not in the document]
mikemorris: Updated.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* SHOULD only be allowed to direct traffic to `backendRefs` within the same namespace when used for configuration of mesh traffic.
+    * This constraint may be revisited in the future, but is intended to help simplify an initial implementation.
+    * Attempting to specify an `HTTPBackendRef` in a different namespace should raise a status condition on the `HTTPRoute`.
+    * When an `HTTPBackendRef` is invalid, 500 status codes MUST be returned for requests that would have otherwise been routed to an invalid backend.
+        * This matches existing behavior as defined in [`HTTPRouteRule`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule).
+* SHOULD set a status field on `HTTPRoute` to show if the routing configuration has been applied to the mesh.
+* MAY assume that a mesh implements "transparent proxy" functionality to redirect calls to the Kubernetes DNS address for a `Service` through mesh routing rules.
+    * For all namespaces configured by a cluster operator in the scope of a mesh deployment.
```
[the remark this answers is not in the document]
mikemorris: Moving a handful of these into an `Implementation Details` section.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+* SHOULD NOT require reconfiguring existing `HTTPRoutes` for North/South Gateway configuration.
+
+## Non-Goals
+
+* Supporting "egress" use cases, which is currently a deferred goal, including:
+    * Defining how "consumer" traffic rules which could override routing for service upstreams only within the local scope of a namespace or service might be configured.
+    * Redirecting calls from arbitrary custom domains to an in-cluster service.
+* Defining how multiple `Services` or `EndpointSlices` representing instances of a single "logical" service should present an identity for AuthN/AuthZ or be associated with each other beyond routing rules.
+* Defining how AuthZ should be implemented to secure East/West traffic between services.
+* Defining how [Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/) would bind to `HTTPRoute`, services or a mesh.
```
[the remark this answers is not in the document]
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1482 to track this.

site-src/geps/gep-1424.md
mikemorris: Does this rephrasing clarify this sufficiently? Agreed on "without selectors" being wrong, removed that.
```suggestion
Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` service. The behavior should be exactly the same as if the `parentRef` service was explicitly defined as the only `backendRef`, so if the `parentRef` does not have any endpoints, implementations MUST return an HTTP 503 response code (see discussion in [#1210](https://github.com/kubernetes-sigs/gateway-api/pull/1210)).
```

site-src/geps/gep-1424.md
mikemorris: @howardjohn Would this specify the behavior more clearly?
```suggestion
The example above would drop all incoming traffic for HTTP paths other than `/bar` to the `foo` `Service`, following the existing spec to return HTTP 404 response codes for unmatched requests, and HTTP 500 response codes for requests excluded due to an [`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).

When no `HTTPRoute` resources or AuthZ configuration are defined, all traffic should implicitly work - this is just how Kubernetes functions. When you create an `HTTPRoute` targeting a service as a `parentRef` you are replacing that implicit logic - not adding to it. Therefore, you may be reshaping or restricting traffic via an `HTTPRoute` configuration (which should be noted is distinct from disallowing traffic by AuthZ).
```

site-src/geps/gep-1424.md
mikemorris: I'm going to merge this with the proposed language (`MAY` and the recommendation to prefer `ClusterIP` services). For a provisional GEP that feels like sufficient leeway to allow implementations to support them, but clear that it's not required. We should definitely start with conformance tests only for ClusterIP services, and use the existing "opt-in" functionality for `NodePort` or `LoadBalancer` `Service` types if/when those tests are added.

site-src/geps/gep-1424.md
mikemorris: Hmm, I was only considering the "frontend" role of a service without selectors as a `parentRef`, not sure what/if should be added to describe any possible role as a `backendRef`? Would be open to adding suggested language here.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,384 @@
...
+
+### `hostnames` field
+
+GAMMA implementations SHOULD NOT infer any functionality from the `hostnames` field on `HTTPRoute` due to current under-specification and reserved potential for future usage or API changes. For the use case of filtering incoming traffic from selected hostnames, it is recommended to guide users toward configuring [`HTTPHeaderMatch`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderMatch) rules for the [`Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) header. Functionality to be explored in future GEPs may include supporting concurrent usage of an `HTTPRoute` traffic configuration for multiple North/South `Gateways` and East/West mesh use cases or redirection of egress traffic to an in-cluster `Service`.
```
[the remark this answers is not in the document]
mikemorris: I'd prefer to avoid the header match behavior so we don't break implementations later - do you have a preference for rejecting versus ignoring @howardjohn?

mikemorris:
/hold

The actual file needs to be renamed from `site-src/geps/gep-1424.md` to `site-src/geps/gep-1426.md` before merging, but I don't want to break any comments accidentally.

howardjohn (commented):
/lgtm
/approve

Thanks!

As discussed in other places, the plan is to send a few followup PRs shortly after address minor gaps, then mark it `implementable`. At that point, we will encourage implementors to implement it *with extreme caution to not become overly dependent on it so the spec doesn't ossify too quickly* and ensure it works as well as we hoped.

site-src/geps/gep-1424.md
howardjohn: :+1:

site-src/geps/gep-1424.md
howardjohn: :+1:

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
howardjohn: I guess we can reject by emitting a status message saying the spec is invalid, but then ignoring it at runtime? Rather than dropping the whole route.

site-src/geps/gep-1424.md
howardjohn: LGTM but I plan to submit a close followup to allow the HTTP merging rules

site-src/geps/gep-1424.md
howardjohn: This is more clear, but I'll note I intend to submit a followup to try to drop this restriction.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,386 @@
...
+        value: "/bar"
+```
+
+The example above would drop all incoming traffic for HTTP paths other than `/bar` to the `foo` `Service`, following the existing spec to return HTTP 404 response codes for unmatched requests, and HTTP 500 response codes for requests excluded due to an [`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).
+
+### Allowed service types
+
+Services valid to be selected as a `parentRef` SHOULD have a way to identify traffic to them - typically by one or more virtual IP(s), DNS hostname(s) or name(s).
+
+Implementations SHOULD support the default [`ClusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) `Service` type as a `parentRef`, with or without selectors.
```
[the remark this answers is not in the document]
mikemorris: Yea, the only reason I didn't yet was unfamiliarity with how implementations like Cillium might work.

/cc @youngnick

mikemorris:
@steeling I did add a specific goal in response to your feedback, I understand we may disagree on this expected behavior, but it should at least be explicitly stated as a goal now, rather than an assumption.

> * SHOULD NOT require downstream "consumer" services to update configuration or DNS addresses for traffic to follow "producer" mesh routing rules configured by upstream services.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+    backendRefs:
+    - kind: Service
+      name: foo
+      weight: 90
+    - kind: Service
+      name: foo_v2
+      weight: 10
+```
+
+In the example above, routing rules have been configured to direct 90% of traffic for the `foo` `Service` to the default "backend" endpoints specified by the `foo` `Service` [`selector`](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) field, and 10% to the `foo_v2` `Service`. This is determined based on the `ClusterIP` (for `Service`) and `ClusterSetIP` (for `ServiceImport`) matching, and for "transparent proxy" mesh implementations would match all requests to `foo.svc.cluster.local` (or arbitrary custom suffix, as the hostname is not specified manually) from within the same namespace, all requests to `foo.store.svc.cluster.local` from other namespaces, and all requests to `foo.store.svc.clusterset.local` for multicluster services, within the scope of the service mesh.
```
costinm: s/foo.svc.cluster.local/foo.store.svc.cluster.local/
 - and not sure what 'arbitrary custom suffix ...' comment means.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+
+### Omitting `backendRefs`
+
+Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` service. The behavior should be exactly the same as if the `parentRef` service was explicitly defined as the only `backendRef`, so if the `parentRef` does not have any endpoints, implementations MUST return an HTTP 503 response code (see discussion in [#1210](https://github.com/kubernetes-sigs/gateway-api/pull/1210)).
+
+This syntax has limited utility currently, but could become more relevant if additional functionality beyond traffic splitting is added to the `HTTPRoute` resource. `HTTPRoutes` configured in this way MAY be used to manually enroll services into a mesh (which could trigger behavior like injecting a sidecar proxy) if they have not already been enrolled by some other mechanism (as proposed in [GEP-1291: Mesh Representation](https://docs.google.com/document/d/1oyA9uUH7pNNxxwy3WZGSWx-edHDBLrujcezr8q3el70/edit) or similar).
```
costinm: Do we support declaring some ports in the Service as http ( using port name or appprotocol ) ? Should a Service with appProtocol=http be considered equivalent to a HttpRoute with parentRef and backendRef pointing to the service ? 

Related: what happens if a Service has multiple ports with appProtocol=http ? Or none ?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+
+When no `HTTPRoute` resources or AuthZ configuration are defined, all traffic should implicitly work - this is just how Kubernetes functions. When you create an `HTTPRoute` targeting a service as a `parentRef` you are replacing that implicit logic - not adding to it. Therefore, you may be reshaping or restricting traffic via an `HTTPRoute` configuration (which should be noted is distinct from disallowing traffic by AuthZ).
+
+### Allowed service types
+
+Services valid to be selected as a `parentRef` SHOULD have a way to identify traffic to them - typically by one or more virtual IP(s), DNS hostname(s) or name(s).
+
+Implementations SHOULD support the default [`ClusterIP`](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) `Service` type as a `parentRef`, with or without selectors.
+
+["Headless" `Services`](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) SHOULD NOT be supported as a `parentRef`, because they do not implement the "frontend" functionality of a service.
```
costinm: Not entirely accurate - they do create DNS entries and a hostname, it just lacks a VIP and kube-driven LB.

They are a pain to support due to the lack of VIP - but useful in some cases, I would relax 'SHOULD NOT'  -maybe optional ?

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+
+`Service` resource with [`type: NodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport) or [`type: LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) MAY be allowed as `parentRefs` or `backendRefs`, as these do provision virtual IPs and are effectively `ClusterIP` services with additional functionality, but it should typically be preferred to expose services publicly through the North/South Gateway API interfaces instead.
```
costinm: How about StatefulSets ( not technical services, but create DNS entries ) ?

site-src/geps/gep-1424.md
costinm: For backendRef - they work as intended. 

As frontendRef - may be worth clarifying 'withtout selector and without user-attached Endpoint/EndpointSlices', i.e. pure virtual services.

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
costinm: Did we decide to not allow a HttpRoute with 2 parents ( Gateway and Service ) ? I would say it is the same as a header match.

site-src/geps/gep-1424.md
mikemorris: This was intended to refer to how a cluster can set a hostname other than `.cluster.local`

site-src/geps/gep-1424.md
mikemorris: We should probably work on specifying that as part of https://github.com/kubernetes-sigs/gateway-api/issues/1487

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+    * This requirement may be relaxed later to follow the existing [precedence rules for merging configuration](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule), but is intended to simplify initial implementation.
+    * The namespace colocation requirement effectively forecloses using this strategy to delegate route management, but there is service mesh [user interest](https://github.com/istio/istio/issues/22997) in merging separately-defined routes together to manage large configurations, multiple versions or canary deployments for a single mesh service.
+* NOT already specified as a `backendRef` of any _other_ `HTTPRoute` with a `Service` `parentRef`
+   * This restriction is intended to prevent route "inclusion/delegation" patterns or circular references from multiple tiers of routing rules _within mesh configuration_.
+   * A `Service` MAY still be listed as a `backendRef` on a different `HTTPRoute` which only targets a `Gateway` as a `parentRef`. The specific behavior of how Gateways might interact with mesh routing rules will be defined in a future GEP.
```
costinm: Would be good to resolve if this is the new 'delegation' for gateways - and this GEP can be promoted as an optional feature for Gateways implementing route delegation.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+    * A `Service` MAY still be listed as a `backendRef` on the _same_ `HTTPRoute` where it is specified as a `parentRef`, e.g. to only split off some subset of traffic from the default backend endpoints.
+    * This could be computationally expensive to validate.
+
+This should be enforced by controller validation, which should set an `Accepted: { status: False }` condition with a corresponding new `ParentRefConflict` `RouteConditionReason`.
+
+The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
```
costinm: I think it is a bit of overkill - a user may deploy an envoy/nginx/etc as a service an have it route to whatever they want, creating cycles anyways. 

I would add a section about 'mesh implementation should implement the Forwarded header, and support a max number of forwards to prevent cycles'. This also provides a standard way for apps to detect the path of the request.

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
costinm: I'm not sure I see the benefit. If a HttpRoute has a Service as parentRef - it is clearly mesh route. What would the extra mesh field in parentRef convey ? 

The Service parentRef provides a clear binding of the route to the service name.

site-src/geps/gep-1424.md
```diff
@@ -0,0 +1,394 @@
...
+
+#### Phantom `parentRef`
+```
+spec:
+  parentRefs:
+  - kind: Mesh
+    name: istio
+```
+
+This is done by configuring the `parentRef`, to point to the `istio` `Mesh`. This resource does not actually exist in the cluster and is only used to signal that the Istio mesh should be used. In Istio's [experimental implementation](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#mesh-traffic), the `hostnames` field on `HTTPRoute` is used to match mesh service traffic to the routing rules.
```
costinm: Worth noting that this also handles egress traffic, not only service.

site-src/geps/gep-1424.md
mikemorris: Would definitely appreciate some help figuring out specific implementation details and what we might be able to support in https://github.com/kubernetes-sigs/gateway-api/issues/1484

site-src/geps/gep-1424.md
mikemorris: Removed StatefulSet from this GEP as it seemed to be a substantial source of confusion or disagreement, hope to address it more explicitly in https://github.com/kubernetes-sigs/gateway-api/issues/1484

site-src/geps/gep-1424.md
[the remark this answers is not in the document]
mikemorris: The main benefit I can recall would be the ability to disambiguate routes if multiple meshes are ever installed on a cluster.

site-src/geps/gep-1424.md
mikemorris: Hmmm, that is an interesting option - I was definitely leaning towards a UX that could validate this when config is created to set a status, but it's definitely more of a pain to implement.

site-src/geps/gep-1424.md
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/issues/1484 to continue this discussion.

Issue #622, opened 2021-04-22
Allow application developer to specify TLS client certificate


tsaarni:
**What would you like to be added**:

Application developer should be able to define a client certificate that the proxy uses when it connects to developer's backend service with TLS.

**Why is this needed**:

Background:

This feature has been previously discussed and it was included as `BackendPolicy.spec.tls.clientCertificateRef`. Later, Gateway level configuration was preferred #395 and feature was removed.  As per discussion in the comments of #400, I would like to propose this capability to be re-introduced as an optional feature,  as an alternative to global (per gateway) configuration.

Use case details:

Assume we have application developer working within Kubernetes namespace. Application developer wants to secure the traffic between proxy and their backend service.

1. Application developer generates their own certificates and configures server certificate to their backend service
2. To get the proxy to validate the server certificate, the application developer configures reference to the CA certificate secret to `BackendPolicy.spec.tls.certificateAuthorityRef`

Next the developer would like to protect the backend service even further, by allowing incoming TLS connections only from the proxy. Symmetrically, application developer could enable client authentication by themselves, like they did for the server authentication in step 2 above but this would require client certificate to be configurable per backend service, not per Gateway which would be outside of application developers domain.

Implementing this enhancement has following advantages:

* Allows "self-service" workflow for the application developer, since they do not need to depend on cluster admin to configure client certificate on gateway level
* Cluster admin does not need to manage / coordinate / distribute single CA certificate for the development teams working in the cluster. Each team can configure their own client certificates for the proxy and CA certificates to their backend for validating the client cert

youngnick:
This seems reasonable to me, with the proviso that the `BackendPolicy` client certificate details are documented as specifying the details for a TLS interaction between the Gateway and the serving Pods.

youngnick:
This API has changed substantially with v1alpha2, we will need to revisit how we can achieve this.

robscott:
It seems like it would be possible to achieve this with some form of policy attachment initially, not sure if there's a better approach or built in policy resource we can offer here.

shaneutt:
While grooming we saw that this one was open for a long period of time without anyone with a strong use case to champion it. We're going to close this as we don't expect anyone's ready to drive it forward, but if you still want this feature and have a strong use case we will be happy to reconsider this and re-open.

tsaarni:
[...]

PR #1430, opened 2022-10-03
GEP-1282 Backend Properties - Update implementation

candita:
**What type of PR is this?**
/kind gep

Add proposed API implementation for Backend Properties.

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1282.md
+++ b/site-src/geps/gep-1282.md
@@ -184,55 +184,45 @@ type BackendCapabilitiesSpec struct {
 	
     // TLS is the section of details on encryption processing in the traffic 
     // hop from Gateway to backend, sometimes known as upstream TLS or re-encrypt.
-    // A xRoute configured as passthrough cannot specify a Forward encrypt type.
+    // A xRoute configured as passthrough cannot specify a TLS encrypt type.
     // TLS is useful for HTTPRoute and should not be used on UDP/TCP/TLSRoute.
     //
     // +optional
     TLS *EncryptType `json:"tls,omitempty"`
 }
 
-// EncryptType describes the details needed to implement encryption from the
-// gateway to the backend.
+// EncryptType describes the details needed by the client (Gateway) to implement 
+// encryption from the client to the destination endpoint.
 type EncryptType struct {
-    // The minimal TLSVersion that is accepted by the backend.
+    // The minimal TLSVersion that is accepted by the endpoint.
     // If empty, there is no minimal backend TLSVersion.
     // +optional	
     MinVersion   *TLSVersion `json:"minVersion,omitempty"`
 
-    // The maximum TLSVersion that is accepted by the backend.
+    // The maximum TLSVersion that is accepted by the endpoint.
     // If empty, there is no maximum backend TLSVersion.
     // +optional    
     MaxVersion   *TLSVersion  `json:"maxVersion,omitempty"`
     
-    // CipherSuite is the minimum cipher suite that the backend accepts.
-    // If empty, anything that is appropriate for the TLS version is accepted.
+    // CipherSuites is the minimum cipher suite that the backend accepts.
+    // If empty, anything that is appropriate for the MinVersion is accepted.
+    // If the MinVersion is not specified, the result is implementation
+    // dependent, e.g. based on the minimimal TLS version supported by the
+    // implementation.
     // +optional
-    CipherSuite   string `json:"cipherSuite,omitempty"`
+    CipherSuites []string `json:"cipherSuites,omitempty"`
     
-    // GatewayCert is the certificate used by the Gateway to prove its identity to
-    // the backend, if the backend so requires.
-    // +optional   
-    GatewayCert  *SecretObjectReference `json:"gatewayCert,omitempty"`
-    
-    // BackendCerts are the CA certificates used by the gateway to authenticate
-    // with the backend. This certificate allows the Gateway to verify and trust the 
-    // backend certificate.
+    // CACerts is the PEM bundle containing Certificate Authority, trusted by both 
+    // client and endpoint, and intermediate certificates needed to complete the 
+    // certificate chain of trust.  If empty, the system root certs should be used.
+    //
     // +optional	
-    BackendCerts []SecretObjectReference `json:"backendCerts,omitempty"`
-
-    // CACerts are the Certificate Authority, trusted by both Gateway and backend,
-    // which complete the certificate chain of trust.
-    // +optional	
-    CACerts      []SecretObjectReference `json:"caCerts,omitempty"`
+    CACerts      *SecretObjectReference `json:"caCerts,omitempty"`
     
-    // SubjectAltNames allow multiple named domains to use a single certificate.
+    // ServiceNames provide multiple named domains which will share a certificate
+    // for the endpoint.
     // +optional
-    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
-    
-    // ServerNameIndication specifies the host name used by this transaction, and
-    // may be required by the backend for TLS negotiation.
-    // +optional
-    ServerNameIndication string `json:"serverNameIndication,omitempty"`
+    ServiceNames []string `json:"serviceNames,omitempty"`
 }
 
 // BackendCapabilitiesStatus describes the status of the backend and its
@@ -249,7 +239,7 @@ type backendCapabilitiesStatus struct {
 ```
 
 ###Example Capabilities
-For a Service private-service defined as:
+For a Service `private-service` defined as:
 
 ```yaml
 apiVersion: v1
@@ -265,7 +255,7 @@ ports:
     targetPort: 9099
 ```
 
-We could create a backendCapabilities for TLS re-encrypt on private-service defined as:
+We could create a backendCapabilities for TLS re-encrypt on `private-service` defined as:
 
 ```yaml
 apiversion: v1beta1
@@ -275,12 +265,17 @@ metadata:
 spec:
   backend:
     kind: Service
-    name:  private-service
+    name: private-service
   TLS:
     minVersion: 1.3
-    gatewayCert: my-gateway-secret
-    caCerts: my-ca-certs
-    backendCerts: my-backend-secret
+    cipherSuites:
+      - TLS_AES_256_GCM_SHA384
+      - TLS_CHACHA20_POLY1305_SHA256
+      - TLS_AES_128_GCM_SHA256
+    caCerts: my-cert-bundle
+    serviceNames:
+      - private-service.example.com
+      - dev.example.com
 ```
 
 Implementers of TLS re-encrypt would have to check for backendCapabilities that shared the same spec.backend.name as
--- a/site-src/geps/gep-1282.md
+++ b/site-src/geps/gep-1282.md
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
 
 ## API
 
-To be written later, once we have agreement on the “what” and the “why”. (This is the “how”).
+A new object, BackendCapabilities, is added to the API that serves as a wrapper to a Kubernetes Service, and provisions
+backend capabilities for a Service that requires them.  As discussed by the Gateway API maintainers and community, this
+architectural choice is made because we would ideally like to locate this capabilities set within the Service but until
+that is possible in Kubernetes, we can start with a wrapper to the Service.
 
+BackendCapabilities contains a BackendCapabilitiesSpec object as mentioned in Option 2, defining each of the different
+capabilities.  In this discussion, the capabilities are TLSReencyrpt, Websocket, and Protocol, but additional
+capabilities may be added as needed in subsequent versions of Gateway API.  BackendCapabilities also contains a
+reference to a backend service name and type, and a BackendCapabilitiesStatus.
+
+In https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go add a new type:
+```go
+// BackendCapabilities references a Kubernetes Service and defines any backend 
+// capabilities that Service requires.  
+//
+// +optional
+type BackendCapabilities struct {
+    // BackendName is the name of the backend service to which these backend 
+    // capabilities apply.
+    // +required
+    BackendName string `json:”backendName”`
+
+    // BackendType is the type of backend to which these backend 
+    // capabilities apply.
+    //
+    // +required
+    BackendType BackendTargetRef `json:”backendType”`
+
+	// Spec specifies the capabilities supported by this
+    // Service wrapper, if any.  The capabilities are predefined, and have two
+    // characteristics in common: they are tightly bound to the backend, 
+    // rather than being important at the route level, and the service owner
+    // (application developer in the Gateway API personas) should control 
+    // these settings, not the owner of the Gateway implementation (the 
+    // Cluster Operator in the Gateway API personas).
+    // 
+    // Support: Extended
+	//
+    // +required
+	Spec BackendCapabilitiesSpec `json:”spec”`
+
+	// Status represents the current state of the backend and its capabilities.
+    // +optional
+	Status *BackendCapabilitiesStatus `json:”status,omitempty”`
+}
+
+// BackendTargetReference identifies an API object to apply BackendCapabilities 
+// to. 
+// 
+type BackendTargetReference struct {
+	// Group is the group of the target resource.
+	Group Group `json:"group"`
+
+	// Kind is kind of the target resource.
+    // Valid values include:
+    // * "Service" (the default if no Kind is specified)
+    // * "ServiceImport"
+    //
+    // Invalid values include:
+    //
+    // * "invalid/kind" - "/" is an invalid character
+    // +required
+	Kind string `json:"kind"`
+
+	// Version is the version of the target resource.
+	Version ObjectName `json:"version"`
+
+	// Name is the name of the target resource.
+	Name ObjectName `json:"name"`
+}
+
+// BackendCapabilitiesSpec defines a group of specific properties held by its
+// affiliated BackendCapabilitiesService, and specified by the Service owner 
+// (typically the application developer).  Examples of properties that 
+// belong in the BackendCapabilitiesSpec include: TLS re-encryption, session
+// state, timeouts, retries, Websockets, HTTP/2, load balancer algorithm, 
+// healthcheck info, external traffic policy, connection limits, error page,
+// protocol switch, alerts, and others. 
+//
+// Support: Extended
+//
+type BackendCapabilitiesSpec struct {
+
+	// +optional
+	Reencrypt *TLSReencryptType `json:”reencrypt,omitempty”`
+
+	// +optional
+    Websocket *WebsocketType `json:”websocket,omitempty”`
+
+	// +optional
+    Protocol *ProtocolType `json:”protocol,omitempty”`
+}
+
+type ReencryptType struct {
+    Version *TLSVersion
+    ClientCertFile string
+    PrivateKeyFile string
+    BackendCertFile string
+    CaCertsFile	  []string
+}
+
+type WebsocketType struct {
+	URL	string
+	BinaryType BinaryTypeType
+}
+
+type ProtocolType struct {
+	Version string
+}
+
+// BackendCapabilitiesStatus describes the status of the backend and its
+// capabilities.
+type backendCapabilitiesStatus struct {
+	// List of conditions that represent the current status of the
+	// the backend capabilities object.  
+    // TODO: Define constants to represent typical status conditions
+    //
+    // +kubebuilder:validation:MinItems=1
+    // +kubebuilder:validation:MaxItems=8
+    Conditions []metav1.Condition `json:”conditions,omitempty"`
+}
+```
+
+###Example Capabilities
+For a Service private-service defined as:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: private-service
+spec:
+  selector:
+    app.kubernetes.io/name: privateApp
+ports:
+  - protocol: TCP
+    port: 80
+    targetPort: 9099
+```
+
+We could create a backendCapabilities for TLS re-encrypt on private-service defined as:
+
+```yaml
+apiversion: v1beta1
+kind: backendCapabilities
+metadata:
+  name: private-service
+backendName: private-service
+backendType:
+  kind: Service
+  version: v1namenum
+  name:  privateApp
+spec:
+  reencrypt:
+    version: 1.3
+    clientCert: /path/to/cert.pem
+    privateKey: /path/to/pkey.pem
+    caCerts: /path/to/cacerts.pem
+    backendCert: /path/to/backendCert.pem
+```
+
+Implementers of TLS re-encrypt would have to check for backendCapabilities that shared the same backendName and
+backendType as the targeted Service, to discover the capabilities specification they would need to perform TLS
+re-encrypt at the Service.
 
 ## Alternatives
 
-Again, we need to wait until we’re writing the “how”.
+### Alternative 1 - map of structs
+
+The API will be composed of several new objects: BackendObjectCapabilities, an instance of BackendObjectCapabilitiesMap,
+defined as a mapping of capability name to an array of BackendObjectCapability objects.  BackendObjectCapability objects
+are simple but expressive Group-Kind-Name objects with an additional Value field implemented as an interface so that it
+can store any type of data.
+
+In https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go:
+```go
+// BackendRef defines how a Route should forward a request to a Kubernetes
+// resource.
+//
+// ...
+type BackendRef struct {
+	// BackendObjectReference references a Kubernetes object.
+	BackendObjectReference `json:",inline"`
+
+   // BackendObjectCapabilities specifies the capabilities supported by this
+   // BackendRef, if any.  The capabilities are not predefined, but have two
+   // characteristics in common: they are tightly bound to the backend, 
+   // rather than being important at the route level, and the service owner // (application developer in the Gateway API personas) should control 
+   // these settings, not the owner of the Gateway implementation (the 
+   // Cluster Operator in the Gateway API personas).
+   // 
+   // There are a maximum of 32 BackendObjectCapabilities per BackendRef.
+   //
+   // Support: Extended
+   //
+   // +optional
+   // +kubebuilder:validation:MaxItems=32
+   BackendObjectCapabilities *BackendObjectCapabilitiesMap `json:”backendObjectCapabilitiesMap,omitempty”`
+	
+	// Weight specifies the proportion of requests forwarded to the referenced
+	//...
+	Weight *int32 `json:"weight,omitempty"`
+}
+
+// BackendObjectCapabilitiesMap defines a mapping of BackendObjectCapability name
+// to an array of BackendObjectCapability
+BackendObjectCapabilitiesMap map[string][]BackendObjectCapability
+
+// BackendObjectCapability defines a capability or property held by its
+// affiliated BackendObject, and specified by the BackendObject owner 
+// (typically the application developer).  Examples of BackendObjectCapability
+// use cases include: TLS re-encryption, session state, timeouts, retries, 
+// Websockets, HTTP/2, load balancer algorithm, healthcheck info, external
+// traffic policy, connection limits, error page, protocol switch, alerts. 
+//
+// The BackendObject must be valid in the cluster; the Group and Kind must
+// be registered in the cluster for this reference to be valid.
+//
+// Support: Extended
+//
+// References to objects with invalid Group and Kind are not valid.
+type BackendObjectCapability struct {
+	// Group is the group of the referent.
+	//
+	// Support: Core
+	//
+	// +kubebuilder:default=gateway.networking.k8s.io
+	// +optional
+	Group *Group `json:"group,omitempty"`
+
+	// Kind is kind of the referent.
+	//
+	// Support: Custom
+	//
+	// +kubebuilder:default=Gateway
+	// +optional
+	Kind *Kind `json:"kind,omitempty"`
+
+	// Name is the name of the referent.
+	//
+	// Support: Core
+	Name ObjectName `json:"name"`
+
+	// Value is the value of the referent.  The type is not specifically
+	// defined and is left up to the implementation to infer.
+	//
+	// Support: Extended
+	// 
+	// +optional
+	Value interface{} `json:”value”`
+}
+
+```
+### Example Capabilities for Alternative 1
+
+TLS information for connection from the Gateway to the backend service. What a Gateway’s proxy would need to be able to
+connect to the backend service.
+
+The BackendCapabilities could be defined by an implementation and look something like:
+
+```yaml
+
+backendCapabilities:
+  tls-reencrypt:
+    - name: version
+    value: 1.3
+    - name: clientCert
+    value: /path/to/cert.pem
+    - name: privateKey
+    value: /path/to/pkey.pem
+    - name: caCerts
+    value: /path/to/cacerts.pem
+    - name: backendCert
+    value: /path/to/backendCert.pem
+```
+Websocket protocol information for the backend service. (I took a best guess at what Websocket might look like)
+
+The BackendCapabilities could be defined by an implementation and look something like:
+```yaml
+backendCapabilities:
+  websocket:
+    - name: url
+    value: scheme://path/to/app
+    - name: binaryType
+    value: blob
+```
+Protocol disambiguation for “upgradeable” protocols like HTTP/2 and HTTP/1.1 which operate on the same port.
+
+The BackendCapabilities could be defined by an implementation and look something like:
+
+```yaml
+backendCapabilities:
+  protocol:
+    - name: version
+    value: HTTP/2
+```
 
 ## References
 
--- a/site-src/geps/gep-1282.md
+++ b/site-src/geps/gep-1282.md
@@ -36,6 +36,7 @@ The purpose of this GEP is to add APIs for capturing those backend capabilities 
 As implementations have started to build out HTTPRoute support, requests for some common capabilities have started coming up, particularly TLS re-encryption (encryption between Gateway and backend), Websockets tracking, and HTTP/2 support.
 
 Evan Anderson opened a [discussion that generated a lot of interest](https://github.com/kubernetes-sigs/gateway-api/discussions/1244), and during that discussion, we brought up a few more points:
+
 * Whatever we do to solve this in the HTTPRoute case may be applicable describing more general service properties, like identity, which could be useful for other Route use cases.
 * This may be very useful for mesh/GAMMA use cases as well as more generally for adding arbitrary future capabilities to the backend service. An example is a CA cert for connecting to the backend - that’s tightly bound to the service, but there’s nowhere to put it at the moment.
 
@@ -46,25 +47,30 @@ This initial form of this GEP is for the Gateway API community to agree on what 
 ### Why build something?
 
 We've got the following feature requests and discussions in the Gateway API repo:
-- [#1244](https://github.com/kubernetes-sigs/gateway-api/discussions/1244) : Unclear how to specify upstream (webserver) HTTP protocol. This issue describes the problems that Evan had in trying to be able to define if a backend behind a HTTPRoute supports HTTP2 over cleartext or websockets. The question of how to tell the Gateway implementation that the backend needs TLS details for a proxy-based implementation to be able to connect it also came up in the discussion.
-- [#1285](https://github.com/kubernetes-sigs/gateway-api/discussions/1285) has a more specific discussion about how different ingress implementations allow this to be configured today, whether that's with the Ingress resource or their own custom one. The great roundup that Candace did is reproduced in the next few bullet points.
+* [#1244](https://github.com/kubernetes-sigs/gateway-api/discussions/1244): Unclear how to specify upstream (webserver) HTTP protocol. This issue describes the problems that Evan had in trying to be able to define if a backend behind a HTTPRoute supports HTTP2 over cleartext or websockets. The question of how to tell the Gateway implementation that the backend needs TLS details for a proxy-based implementation to be able to connect it also came up in the discussion.
+  
+* [#1285](https://github.com/kubernetes-sigs/gateway-api/discussions/1285) has a more specific discussion about how different ingress implementations allow this to be configured today, whether that's with the Ingress resource or their own custom one. The great roundup that Candace did is reproduced in the next few bullet points.
+  
   * Istio uses a [DestinationRule resource with ClientTLSSettings](https://istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings) to capture TLS details, and the DestinationRule resource also holds traffic policy information like load balancing algorithm, connection pool size, and so on.
   * Openshift’s Route resource allows the [configuration of reencryption](https://docs.openshift.com/container-platform/4.10/networking/routes/secured-routes.html#nw-ingress-creating-a-reencrypt-route-with-a-custom-certificate_secured-routes) specifically, along with custom certificate details.
   * Contour’s HTTPProxy captures TLS details using an Envoy client certificate, destination CA certificate, and optional SubjectName which sets what Envoy should expect to see from the backend service, all inside the HTTPProxy resource. It also requires either a Protocol field inside the HTTProxy, or an annotation on the Service that tells Contour that the service expects TLS. This is [all documented](https://projectcontour.io/docs/v1.21.1/config/upstream-tls/), but I should note that Contour’s docs use the Envoy convention where a backend in Gateway parlance is called an Upstream (which may be confusing if you’re not used to it).
   * Linkerd uses a [Server resource](https://linkerd.io/2.11/reference/authorization-policy/#server) (which is functionally pretty similar to Service in that it associates a name with a Pod selector, but also has other details like if the service supports Proxy protocol), along with a ServerAuthorization resource that specifies some constructs that sit more at the service mesh level, including identity and access control.
 
 In terms of other implementations existing use cases for features like this:
+
 - For Contour, there are annotations to allow the configuration of the following on a backend Service:
   - Max connections
   - Max pending requests
   - Max requests
   - Max retries
   - Upstream protocol: This is what allows Contour to handle switching protocols for the backend service.
+
 - In Istio, the `DestinationRule` resource allows the configuration of many settings like this:
   - Load balancer algorithms
   - Connection pool settings
   - Outlier Detection
   - Tunnel settings for non-HTTP workloads
+
 - In Consul, [`ServiceDefaults`](https://www.consul.io/docs/connect/config-entries/service-defaults) allows specifying similar configuration:
   - Max connections
   - Max pending requests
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+
+	// +optional
+    Websocket *WebsocketType `json:”websocket,omitempty”`
+
+	// +optional
+    Protocol *ProtocolType `json:”protocol,omitempty”`
+}
+
+type ReencryptType struct {
+    Version *TLSVersion
```
robscott, line 195: I think many backends support a range of TLS versions, it may be worth allowing room for that here. As a broader question, is this absolutely required? Could we live without it in a v1?

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    ClientCertFile string
```
robscott, line 196: Based on the discussion in the community meeting yesterday, this doesn't seem like the right place for this config. Maybe the Gateway is a better fit?

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    PrivateKeyFile string
+    BackendCertFile string
```
robscott, line 198: What are these for? It may be helpful to see an e2e use case where all of these fields were used with an example of who populated each of these. As a broader note, these feel like something we should derive from Secret/Object refs, not file paths.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    CaCertsFile	  []string
```
robscott, line 199: Is this intended to be a list of CaCerts Files? Would it be sufficient to have a single CA Bundle referenced with an object ref?

robscott (commented):
Thanks for the work on this @candita!

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+// state, timeouts, retries, Websockets, HTTP/2, load balancer algorithm, 
+// healthcheck info, external traffic policy, connection limits, error page,
+// protocol switch, alerts, and others. 
+//
+// Support: Extended
+//
+type BackendCapabilitiesSpec struct {
+
+	// +optional
+	Reencrypt *TLSReencryptType `json:”reencrypt,omitempty”`
```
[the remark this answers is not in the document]
tsaarni: I see this was also discussed in https://github.com/kubernetes-sigs/gateway-api/discussions/1285#discussioncomment-3261405. To me `reencrypt` is quite misleading since, as said, it is just about TLS for backend/upstream hop. Personally, I would rather vote something neutral like `tls`, with a description that it can be used also for re-encrypt use case?

site-src/geps/gep-1282.md
tsaarni: I also think having only `TLSVersion` is bit ambiguous. Usually I see something like `MinTLSVersion`, `MaxTLSVersion` and also a list of `AllowedCipherSuites` (and so far, ciphers have no effect for TLSv1.3).  

Maybe for TLS clients, like we have here, these settings are bit "less visible" than for servers: servers immediately get caught when running TLS scanners towards them and users will request for configurability. In case of clients, the lack of configurability might go under the radar for a while. Regardless, having configurability at least for min/max would be very nice.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    - name: version
+    value: 1.3
+    - name: clientCert
+    value: /path/to/cert.pem
+    - name: privateKey
+    value: /path/to/pkey.pem
+    - name: caCerts
+    value: /path/to/cacerts.pem
+    - name: backendCert
+    value: /path/to/backendCert.pem
```
tsaarni, line 374: Likely a strange question: what is the syntax used in the examples in this chapter? :-) I was rather expecting to see (partial) Kubernetes manifests matching the API / CRD proposal earlier in this document instead of general name/value records.

site-src/geps/gep-1282.md
tsaarni: I was just commenting in old (closed) issue https://github.com/kubernetes-sigs/gateway-api/issues/622#issuecomment-1267937777 that there might have been some confusion, but since I was not present at the meeting yesterday, I miss the context and maybe I might be just confused myself:

Since this GEP proposes encryption for the cluster-internal network hop from the gateway proxy to the backend, shouldn't the fields here be something like:

* `certificateRef` that is a reference to `Secret` of `type: kubernetes.io/tls`, where the secret includes keys `tls.crt` and `tls.key` which are the client credentials that gateway should use when establishing TLS connection to said backend service
* `caCertificateRef` that is a reference to `Secret` of `type: Opaque`, where the secret includes key `ca.crt` which is a CA bundle for validating the backend server certificate.

Sidenote: there is no common API / agreement about the CA certificate. For example, certificate-manager can store it in `type: kubernetes.io/tls` with name `ca.crt` even though `type: kubernetes.io/tls` does not recognize key with that name.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+	Reencrypt *TLSReencryptType `json:”reencrypt,omitempty”`
+
+	// +optional
+    Websocket *WebsocketType `json:”websocket,omitempty”`
+
+	// +optional
+    Protocol *ProtocolType `json:”protocol,omitempty”`
+}
+
+type ReencryptType struct {
```
tsaarni, line 194: Is there a need to add optional `subjectName` for server certificate validation? By default, HTTPS clients are [expected to validate](https://www.rfc-editor.org/rfc/rfc9110#name-https-certificate-verificat) that the identity in server certificate matches with the server name in URI. Sometimes there might be a need to overwrite this behavior and compare the identity to a name that was provisioned manually, instead of using the backends DNS name.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
 ## API
 
-To be written later, once we have agreement on the “what” and the “why”. (This is the “how”).
-
+A new object, BackendCapabilities, is added to the API that serves as a wrapper to a Kubernetes Service, and provisions
+backend capabilities for a Service that requires them.  As discussed by the Gateway API maintainers and community, this
+architectural choice is made because we would ideally like to locate this capabilities set within the Service but until
+that is possible in Kubernetes, we can start with a wrapper to the Service.
+
+BackendCapabilities contains a BackendCapabilitiesSpec object as mentioned in Option 2, defining each of the different
```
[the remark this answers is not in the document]
candita: Agreed

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+capabilities.  In this discussion, the capabilities are TLSReencyrpt, Websocket, and Protocol, but additional
```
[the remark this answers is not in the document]
candita: 👍🏼

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    // +required
+    BackendName string `json:”backendName”`
+
+    // BackendType is the type of backend to which these backend 
+    // capabilities apply.
+    //
+    // +required
+    BackendType BackendTargetRef `json:”backendType”`
+
+	// Spec specifies the capabilities supported by this
```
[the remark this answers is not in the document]
candita: 👍🏼

site-src/geps/gep-1282.md
candita: This is an Alternative option that was considered, a map of structs, offering the most minimal set of predefined fields.  We will not likely choose this, but if anyone has a preference for `Alternative 1`, now would be the time to discuss that.

site-src/geps/gep-1282.md
candita: Good points.  Will replace with min/max version. Optional, so if not specified then there is no check.  If anyone feels strongly that we don't need it, we can remove it.

site-src/geps/gep-1282.md
candita: I'll make some changes on this, using `SecretObjectReference` from https://github.com/kubernetes-sigs/gateway-api/blob/96414f1ce72b2277cde5cb8e8abcd97ea039fe4a/apis/v1beta1/object_reference_types.go#L48  
```suggestion
 type ReencryptType struct {
    // +optional	
    MinVersion   *TLSVersion
    // +optional    
    MaxVersion   *TLSVersion
    // +optional	
    ClientCerts  []SecretObjectReference
    // +optional	
    BackendCerts []SecretObjectReference
    // +optional	
    CaCerts	     []SecretObjectReference
}
```

site-src/geps/gep-1282.md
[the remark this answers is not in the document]
candita: in https://gateway-api.sigs.k8s.io/guides/tls/?h=tls#clientserver-and-tls, since we call the first hop `Terminate`, how about calling the second hop `Extend` or `Forward`?  

I hesitate to call it `tls` because we already use that in `Gateway.spec.Listeners` https://github.com/kubernetes-sigs/gateway-api/blob/96414f1ce72b2277cde5cb8e8abcd97ea039fe4a/apis/v1beta1/gateway_types.go#L207 and documented in https://gateway-api.sigs.k8s.io/guides/tls/?h=tls#clientserver-and-tls

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    // * "ServiceImport"
+    //
+    // Invalid values include:
+    //
+    // * "invalid/kind" - "/" is an invalid character
+    // +required
+	Kind string `json:"kind"`
+
+	// Version is the version of the target resource.
+	Version ObjectName `json:"version"`
```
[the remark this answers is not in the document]
candita: 👍🏼

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    // BackendName is the name of the backend service to which these backend 
+    // capabilities apply.
+    // +required
+    BackendName string `json:”backendName”`
+
+    // BackendType is the type of backend to which these backend 
+    // capabilities apply.
+    //
+    // +required
+    BackendType BackendTargetRef `json:”backendType”`
```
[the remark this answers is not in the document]
candita: I made some changes.  Lmk what you think.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +98,302 @@ Of course, from the section above, you can see that there are many other feature
...
+    // Cluster Operator in the Gateway API personas).
+    // 
+    // Support: Extended
+	//
+    // +required
+	Spec BackendCapabilitiesSpec `json:”spec”`
+
+	// Status represents the current state of the backend and its capabilities.
+    // +optional
+	Status *BackendCapabilitiesStatus `json:”status,omitempty”`
```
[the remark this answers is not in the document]
candita: I like how Gateway API has a status for nearly every object, and was hoping to use that from the beginning, even if there was just one condition listed.

site-src/geps/gep-1282.md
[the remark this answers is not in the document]
candita: I used TLS, and updated some more of the godoc.  Please take a look.

site-src/geps/gep-1282.md
candita: I made some changes. Let me know what you think.

site-src/geps/gep-1282.md
candita: I changed it again.

site-src/geps/gep-1282.md
candita: I added `subjectAlternativeName` and `serverNameIndication`.  One of these (or both) might provide what is needed.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
...
+    // If empty, there is no minimal backend TLSVersion.
+    // +optional	
+    MinVersion   *TLSVersion `json:"minVersion,omitempty"`
+
+    // The maximum TLSVersion that is accepted by the backend.
+    // If empty, there is no maximum backend TLSVersion.
+    // +optional    
+    MaxVersion   *TLSVersion  `json:"maxVersion,omitempty"`
+    
+    // CipherSuite is the minimum cipher suite that the backend accepts.
```
howardjohn, line 207: There is no "minimum" cipher suite as far as I can tell; this should be a list?

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
...
+    // If empty, anything that is appropriate for the TLS version is accepted.
+    // +optional
+    CipherSuite   string `json:"cipherSuite,omitempty"`
+    
+    // GatewayCert is the certificate used by the Gateway to prove its identity to
+    // the backend, if the backend so requires.
+    // +optional   
+    GatewayCert  *SecretObjectReference `json:"gatewayCert,omitempty"`
```
howardjohn, line 215: This shouldn't be specific to gateway

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
...
+    BackendCerts []SecretObjectReference `json:"backendCerts,omitempty"`
+
+    // CACerts are the Certificate Authority, trusted by both Gateway and backend,
+    // which complete the certificate chain of trust.
+    // +optional	
+    CACerts      []SecretObjectReference `json:"caCerts,omitempty"`
+    
+    // SubjectAltNames allow multiple named domains to use a single certificate.
+    // +optional
+    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```
howardjohn, line 230: How does this match? Are prefix matches allowed?

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
...
+    // If empty, anything that is appropriate for the TLS version is accepted.
+    // +optional
+    CipherSuite   string `json:"cipherSuite,omitempty"`
+    
+    // GatewayCert is the certificate used by the Gateway to prove its identity to
+    // the backend, if the backend so requires.
+    // +optional   
+    GatewayCert  *SecretObjectReference `json:"gatewayCert,omitempty"`
+    
+    // BackendCerts are the CA certificates used by the gateway to authenticate
```
howardjohn, line 217: I am confused how this differs from CACerts

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
...
+    // with the backend. This certificate allows the Gateway to verify and trust the 
+    // backend certificate.
+    // +optional
```
howardjohn, line 220: what happens if not set? We don't validate, or use system root certs?

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,297 @@ Of course, from the section above, you can see that there are many other feature
 
 ## API
```
howardjohn, line 107: I mentioned this on the call but want to re-iterate so its in writing: I think BackendCapabilities is the wrong abstraction for this API.

What `BackendCapabilities` says is "Here are the properties of the backend". This may be something like "Listens on TLS1.2 or TLS1.3 over HTTP/2" or "Supports HTTP/1.1 and Http/2", etc.

What we want to do, as implementors, is _program a client_ to decide how to connect. With the information in this API, we don't have enough data to know how to program a client.

If a backend says they support HTTP/1.1 and HTTP/2, how do I know which one to send? 

If a backend says it supports TLS, I have the same problem. As a gateway, I can either forward a request as is or add a layer of TLS. But in many cases, we won't actually know if the original request already *is* TLS encrypted, so we don't know whether we should add a layer of TLS or not.

Ultimately, this configuration is a *consumer configuration*; by treating it as a producer configuration we are not getting the functionality required to implement this effectively.

Note: you may look at DestinationRule and make the same argument, but actually it is a consumer configuration that allows the producer to set defaults for consumers; for reasons mentioned above, setting a default policy of "TLS encrypt" would almost certainly break users.

candita: Thanks @howardjohn for writing this out, I understand the concern better.  

My viewpoint, and I think @youngnick 's, is that by creating `BackendProperties` that reference a backend, the developer or admin is telling the Gateway how to interact with the backend.  The intention is still to use `BackendCapabilities` as a wrapper or decorator for the backend.  

I may miss some nuances here, but I thought what might happen is that the `Gateway` (acting in its client role) would lookup and use `BackendCapabilites` as its reference to the backend. For a `HTTPRoute` that is not a passthrough (assuming the `Gateway` knows which traffic is passthrough), if the `Gateway` finds a `BackendCapabilitesSpec` for the backend, then it uses that information to decide how to  connect.  If it doesn't find a spec, then it connects as it would have without knowing the backend capabilities.

Note I started with `BackendTargetRef` as a part of `BackendCapabilities`, not `BackendCapabilitesSpec`.  Later I moved it into the spec by request from @robscott (https://github.com/kubernetes-sigs/gateway-api/pull/1430#discussion_r987472884).  Does hiding the `BackendTargetRef` within `BackendCapabilitesSpec` make programming a client more difficult?  

Is there any way to salvage the `BackendCapabilies` as an abstraction?  We can definitely add more to make it work properly if it is still missing too much.  Maybe we need to store more of the handshake information for TLS, there are a few pieces missing, like details on cipher suite, compression, curve, etc.  And there are few pieces that still need tweaking.

youngnick: @howardjohn, there are a few reasons why we're doing this with a separate resource:
- To avoid adding fields to the `backendRef` that may not always make sense there
- To avoid having to set fields individually for each `backendRef` - this cost increases for each time a backend is referenced in a Route config.
- To make it so that this config is available to other things than HTTPRoute without reimplementing in every Route type

The basic idea is that there's an object, owned by the _owner of the Service_ that tells consumers what they can do when they connect to the service. You're one hundred percent right that these are _consumer_ rules, but the only person who knows what these rules should be is the person who sets up the Service - the Producer.. So this is a way for the _service owner/producer_ to tell _Service consumers_ about how to connect to the service, in a structured way.

It seems to me like having the Route/Gateway need to guess at these rules is not ideal, unless we assume that the Route owner _always_ is the Service producer. 

If that is the case, then we can easily do this by adding fields to `backendRef`. The tradeoff there is that now, _everywhere_ that uses `backendRef` needs to have rules about what happens when you specify TLS settings, or websocket settings, or whatever. So if you use your backend in more than one backendRef, you now need to duplicate all these settings.

tsaarni: Is the problem just that calling this "backend properties/capabilities" is guiding towards wrong abstraction? For example, the description currently in the TLS fields says:

```go
    // The minimal TLSVersion that is accepted by the backend.
    ...
    // CACerts are the Certificate Authority, trusted by both Gateway and backend,
```

This wording makes sense if one would really describe _capabilities of the backend_. But since we are programming the client/gateway/proxy, speculating about the backend capabilities like this seems bit unexpected. 

What I was expecting to see was simply properties applied to Gateway's _backend connection(s)_ when communicating towards specific backend(s). Something like "the minimum TLSVersion used by the Gateway when connecting to the backend" and "Certificate Authorities used by Gateway to verify backend server certificate".  I'm not sure if the same holds to everything, but at least for TLS it seems so.

howardjohn: The alternative is not to make _every use in `backendRef`_. The core relationship of consumer + producer rules, IMO, is "Producer sets defaults, can be overriden by consumer" (which is essentially what policy attachment is all about, as well). Its not necessarily either-or.

---

> For a HTTPRoute that is not a passthrough (assuming the Gateway knows which traffic is passthrough), if the Gateway finds a BackendCapabilitesSpec for the backend, then it uses that information to decide how to connect.

Focusing overly on HTTPRoute (here and in other GEPs, we tend to) is not always the best path. What if the route is a TCPRoute? The contents are opaque - we cannot tell if its TLS or not.

Even if its a TLSRoute - you know it has _a layer of TLS_, but how many? How many does the backend expect? May sound crazy, but its not uncommon to have multiple layers of TLS at this point with transparent mTLS, things like [onion routing](https://en.wikipedia.org/wiki/Onion_routing), etc.

But most importantly - backendProperties, as proposed, are a property of the backend. That means there is no relationship to Gateway or Routes _at all_. These properties should be respected by any Kubernetes cluster that can support them, not just N/S traffic (mesh).

youngnick: > What I was expecting to see was simply properties applied to Gateway's backend connection(s) when communicating towards specific backend(s). Something like "the minimum TLSVersion used by the Gateway when connecting to the backend" and "Certificate Authorities used by Gateway to verify backend server certificate". I'm not sure if the same holds to everything, but at least for backend TLS hop it seems so.

Is that not what this proposal is doing? Is it just the _name_ `BackendCapabilities` that is the sticking point? The idea here is to allow the _owner of the service_ to say "When connecting to the _thing that's running inside the pod_. do these things" - have a minimum TLS version, use this CA cert and so on. If it's just that the word `Capabilities` is too overloaded, and we need to change the name, I don't think that's a big deal.

@tsaarni, in the quote you've given, I think it's equally correct to have that say:
```
    // The minimal TLSVersion required to connect to the backend.
    ...
    // CACerts are the Certificate Authority, that the serving certificate for the backend service must be validated against.
```

Both of those properties are things that are _set_ by the owner of the Service, but are _needed_ by clients that connect to the Service. Similarly, the point of having websocket details in a resource like this is so that the owner of the Service can say "My service will accept websockets at the following paths", so that clients (like a Gateway proxy) know what to do for those paths.

Again, the point here is to take a whole bunch of configuration that, in the Ingress world, was often done via annotations on the Service, and make it structured. Ideally, we would put this inside the Service in structured fields, but we can't easily change the Service spec, so we've opted for a "decorator" style CRD.

If it's just that the _name_ `BackendCapabilities` is creating the wrong abstraction, that's fine, but we need some guidance around what else to call it.

@howardjohn, I'm not quite sure I understand what you're asking for. Are you saying that this resource should be called BackendHTTPProperties or something, and only be relevant when it's used as a `backendRef` in a HTTPRoute? Or are you saying that we should define semantics for the other Route types?

It's clear that you disagree with the current approach, but I am not sure what actions we need to take to move towards something we can agree on.

howardjohn: Consumers need to be able to control the settings, not just the service producers.

I am definitely not suggesting making it tied to HTTP - my concern was the opposite. That this won't work for non-http clients because we don't know if we should encrypt or not.

youngnick: Okay, I agree that consumers also need a place for some settings like this. But don't we need to have the service producer's settings available at least, and then we can see what things make sense to allow in other places? I'll be honest, in my head I kind of thought that Policy Attachment could be used to set things from the Gateway side - so the producer can say "I accept TLS 1.2 or greater", and a Policy could say "All Routes attached to this Gateway will default to TLS 1.3" and then things would work. But without the bit where the producer says what they expect, then you can't even do simple cases, where there's no extra TLS wrapping etc.

howardjohn: I agree with you but I don't think the GEP behaves that way? Or I read it wrong

youngnick: I don't think we have anything in the GEP right now about Policy Attachment and consumer settings, but that's probably worth adding.

I'll review the GEP again and see if I can see any other changes, and then come back here.

site-src/geps/gep-1282.md
youngnick: Agreed, this should be `[]string` so that it can store a list of ciphers in the usual format.

site-src/geps/gep-1282.md
youngnick: I think this one is more intended to hold the serving certificate being used; it's probably not necessary.

site-src/geps/gep-1282.md
youngnick: Yes, this is more describing a client certificate, but I think that this one can probably be elided for now - the behavior we want is that the _Gateway_ holds a client certificate that can be used with the CACert specified in this object.

site-src/geps/gep-1282.md
youngnick: For the CACerts key, I think that if that field is set, it should be the entire trust chain used for certificate authentication in both directions, and if not set, system root certs should be used. This will let users use Let's Encrypt certs from cert-manager or similar for their service if that's what they want.

youngnick (changes requested):
I spent some time thinking about this proposal some more after some back-and-forth in the comments.

I think it's important that we focus the fields in on "the things a client will need to know to connect to the wrapped service". Since a Gateway is a client, that will enable the Gateway to do reencrypt, and also allow other uses of this object.

This pattern - that the BackendCapabilities describes the information you need to _connect_ to the backend is what I had in mind when I wrote up GEP-1282 originally. I think in the discussions, the original intent has got a bit muddled up - this is on me for not being clearer to everyone, sorry.

It could be that BackendCapabilites should really be called BackendConnectionInfo or the more simple BackendProperties, since it seems that `Capabilities` is a slightly overloaded term. I'd like to discuss this some more in the next community meeting.

For the actual spec though, the good news is that it's only a limited set of things that we need to put in the struct to describe the information you need to know to connect to a TLS backend.
* MinVersion and MaxVersion are great and should stay.

Then, we just need:
* CipherSuites as a list of strings, so that we can store it as a list of cipher names.
* CACerts - you have this as a list, which I think is fine, but it could also be a single secret containing a PEM bundle. (That's how other Secret references in Gateway API work, if you want to put multiple certs in them, just use a PEM bundle).
* SubjectAltNames - I think we should maybe leave this out for the first cut. If we do include it, we should call it ServiceNames and specify that it contains a list of all the names the service will respond to (that is, that it has certificates for.) That way, it can be used by a Gateway implementation to check if the Listener and/or Route hostnames match the backend. That does mean it will be a three-way hostname match, which could get pretty complicated.
* I think we should leave out the other fields - there's no need to supply the serving cert (in the BackendCert field currently), that gets sent as part of the handshake, and the SNI is covered by the ServiceNames field if we put it in. The GatewayCert is configuration that should live somewhere on the Gateway - it's up to the Gateway owner and the implementation to wire a client certificate that matches the CACert to the Gateway's data plane, so that requests can be made with a correct client certificate. I also don't think that we should add in client SNI checking yet, this can be added later if we think that this model will work.

youngnick:
Oh, lastly, I think that we probably need a bit of discussion about how this will interact with Policy Attachment.

In short, I think it should be something like:
- BackendCapabilities sets what the _service producer_ allows/wants for connections.
- Policy Attachment may further default or limit the set of connection properties at a Route, Gateway, or GatewayClass level, using the defaults and overrides functionality.

I can explain this further in a meeting after I've thought it through some more.

tsaarni:
May I ask one clarification for the GEP update? In the GEP is said that the initial list of properties to add includes "_TLS information for connection from the Gateway to the backend service_" which I thought would include client credentials. However, I see from comments from @howardjohn and @youngnick that this is not the case. @youngnick writes:

> The GatewayCert is configuration that should live somewhere on the Gateway - it's up to the Gateway owner and the implementation to wire a client certificate that matches the CACert to the Gateway's data plane, so that requests can be made with a correct client certificate.

What I think is meant here is that Cluster Operator who creates Gateways would provision the client certificate. Cluster Administrator then gives (via some out-of-band communication mechanism) the CA certificate to the Application Developer (service provider) who needs to configure that to their backend to allow it to validate Gateway's client certificate.  Is that the expected workflow?

My question is coming from the "self-service" perspective:  There is a likelihood that the Application Developer (service provider) would like to provision their own client certificate for Gateway, without needing to coordinate this with the Cluster Administrator.  Of course, I'm not sure what @candita's expectation was for the workflow? 

Also, if it is Cluster Administrator who provisions the client certificate for Gateway, I'm wondering why Application Developer is in control of TLS versions that Gateway uses, or cipher suites. What drives the decision which TLS properties are managed by which persona? All these parameters go into client's SSL context.

youngnick:
I think that a part that is confusing here is that it seems likely that we'll need to have two places where backend config *can* live.

Firstly, the config for the service itself. This is the service saying "I will accept TLS connections that look like this". This includes TLS version, the CA cert used for the communication, and so on. I think that these are a good fit for a BackendCapabilities/BackendProperties object.

Secondly, the config used to _connect_ to the service. Depending on how the Gateway implementation is interacting with the services, it could be that the implementation owner wants the Gateway to have _one_ client certificate, or _many_. It could also be that, as @tsaarni says, that the _application owner_ wants every client to have their own, individual client certificate.

For the case that the Gateway implementation has _one_ client certificate across *all* client connections, that's a implementation-level config.

For the case that the Gateway implementation needs individual client certificates for each backend, I think the right place is actually on the HTTPRoute, since that specifies _how the Gateway impl connects to the backend_.

In this model, the BackendCapabilities is the _Service resource owner_ saying what the _serving_ properties are (what TLS versions may _possibly_ be used, what the serving cert CA is, and for cases like websockets, what the paths that support websockets are), and the HTTPRoute would hold the _client_ properties (what version should actually be used by the Gateway impl to connect, what client certificate should be used to connect, and so on).

The difference between the two is that the BackendCapabilities information is (at least theoretically) useful to _other things than a Gateway_, and the HTTPRoute information is needed by _things that want to connect_.

I'm kind of making this up as I go along - this is a classic case that @candita had the very reasonable requirement that she wanted to be able to describe how to do a backend TLS connection in the API at all, and we've ended up stuck on definitional and semantic issues about where config belongs.

site-src/geps/gep-1282.md
candita: Changed it to bundle.

site-src/geps/gep-1282.md
candita: I removed both and replaced with `ServiceNames`.

site-src/geps/gep-1282.md
candita: TLS 1.3 only accepts a subset of what 1.1/1.2 accept.  I changed this to a list so the implementation can validate whatever cipher suites are included.  However, if they were left empty, and there is no min or maxVersion, the implementation would need to determine which cipherSuites to use.

site-src/geps/gep-1282.md
candita: Removed.

site-src/geps/gep-1282.md
candita: If the certificate is for a wildcard domain like *.example.com then prefix matches like abc.example.com and def.example.com are allowed.  Should I be specific in the godoc for this?

site-src/geps/gep-1282.md
candita: Removed.

site-src/geps/gep-1282.md
Miciah: TLSv1.2 cipher suites and TLSv1.3 cipher suites don't map one-to-one.  One option would be to have two separate fields for TLSv1.2 (and earlier) and for TLSv1.3.  That's what HAProxy does: 
- http://cbonte.github.io/haproxy-dconv/2.6/configuration.html#3.1-ssl-default-bind-ciphers
- http://cbonte.github.io/haproxy-dconv/2.6/configuration.html#3.1-ssl-default-bind-ciphersuites

site-src/geps/gep-1282.md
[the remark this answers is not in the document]
Miciah: What would be an example of a useful status condition that would apply to a `BackendCapabilities` and not to the `HTTPRoute` that targets the `Service` that the `BackendCapabilities` modifies?

site-src/geps/gep-1282.md
Miciah: > (and so far, ciphers have no effect for TLSv1.3).

@tsaarni, what do you mean by this?

candita:
> I think that a part that is confusing here is that it seems likely that we'll need to have two places where backend config _can_ live.
>
To me, that is a clear indication that the original design is wrong.  We wanted to make sure that whatever we designed would be a wrapper to a Service, because we would really like to keep all of this information aligned with the Service itself.

> Firstly, the config for the service itself. This is the service saying "I will accept TLS connections that look like this". This includes TLS version, the CA cert used for the communication, and so on. I think that these are a good fit for a BackendCapabilities/BackendProperties object.
> 
> Secondly, the config used to _connect_ to the service. Depending on how the Gateway implementation is interacting with the services, it could be that the implementation owner wants the Gateway to have _one_ client certificate, or _many_. It could also be that, as @tsaarni says, that the _application owner_ wants every client to have their own, individual client certificate.
> 
> For the case that the Gateway implementation has _one_ client certificate across _all_ client connections, that's a implementation-level config.
> 
Yes, we could call that the Gateway's default certificate, used by all clients when nothing else was specified.
We could also have a global default certificate, used by all clients that don't even have a Gateway default certificate.

> For the case that the Gateway implementation needs individual client certificates for each backend, I think the right place is actually on the HTTPRoute, since that specifies _how the Gateway impl connects to the backend_.
> 
No, we want this to apply to more than just HTTPRoute.  Unless you want to put backend capabilities on each route type.  Again, going back to the model of "wrapper on Service", this just doesn't fit in.

I see this is on the agenda tonight for discussion, so perhaps we can hash it out there.

costinm:
- having a large object with all the possible properties of a backend seems very dangerous, will be hard to evolve it
- for TLS1.3 - I don't think it's a common use case for a user to go to each backend and specify the settings. Usually TLS1.3 is required for compliance or because the gateway admin requires 1.3 - and that would be set in the gateway settings.
- I'm concerned with client cert specification - it is pretty tricky for the producer to tell the consumer what to use. I would expect any mTLS from gateway to backend to default to the workload identity or be configured on the client side.

site-src/geps/gep-1282.md
```diff
@@ -90,6 +97,7 @@ We’re looking to add specific, structured extension points somewhere in the re
 Those specific, structured extension points need to be in a place where they can be owned by the person who owns the backend, since that could be different to the person who owns the Route. That is, whatever we choose should be something that extends a thing that is referred to by a `backendRef` in a Route, not inside the Route.
 
 The initial list includes, but is not limited to:
+
 * TLS information for connection from the Gateway to the backend service. Note that this doesn’t include any information used for service mesh encryption, just what a Gateway’s proxy would need to be able to connect to the backend service.
 * Websocket protocol information for the backend service.
```
costinm, line 102: I don't think mixing TLS info with protocol info is ideal. TLS is pretty advanced and scary for many users.
Also in many cases this is negotiated at protocol level, not config ( ALPN, etc)

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,292 @@ Of course, from the section above, you can see that there are many other feature
...
+
+// BackendTargetReference identifies an API object to apply BackendCapabilities to. 
+// 
+type BackendTargetReference struct {
+    // Group is the group of the target resource.
+    Group Group `json:"group"`
+
+    // Kind is kind of the target resource.
+    // Valid values include:
+    // * "Service" (the default if no Kind is specified)
```
costinm, line 156: How can it be a Service if it's a backend/workload capability ? Deployment/ReplicaSet/etc - or pod label selector maybe ? 

If you want this to be a Service extension - call it ServiceProperties or ServiceTLS, etc. 

At least in Istio we make a very big distinction between workloads - pods selected by label - and Services.  I think  many users treat Service as a frontend - and the selected Pods as backend.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,292 @@ Of course, from the section above, you can see that there are many other feature
...
+    // +optional
+    TLS *EncryptType `json:"tls,omitempty"`
+}
+
+// EncryptType describes the details needed by the client (Gateway) to implement 
+// encryption from the client to the destination endpoint.
+type EncryptType struct {
+    // The minimal TLSVersion that is accepted by the endpoint.
+    // If empty, there is no minimal backend TLSVersion.
+    // +optional
```
costinm, line 199: I don't know if this is something each individual service owner should mess with and configure individually.

TLS compliance is an admin choice, may be configured on the Gateway or on cluster or implementation-specific config.  In most cases we want to prevent service owners to downgrade security or break the compliance requirements by missconfiguration.

This is even a bigger problem for CipherSuites - do we expect each individual backend owner to understand what cipher suites are safe and configure them ? This is something very few people know or should mess with.

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,292 @@ Of course, from the section above, you can see that there are many other feature
...
+    // implementation.
+    // +optional
+    CipherSuites []string `json:"cipherSuites,omitempty"`
+    
+    // CACerts is the PEM bundle containing Certificate Authority, trusted by both 
+    // client and endpoint, and intermediate certificates needed to complete the 
+    // certificate chain of trust.  If empty, the system root certs should be used.
+    //
+    // +optional	
+    CACerts      *SecretObjectReference `json:"caCerts,omitempty"`
```
costinm, line 220: That's the biggest problem - we can't have a SecretObjectReference crossing a namespace boundary. We had CVEs related to this, it's not good. 

Also not sure what it means 'trusted by both' - I suspect what you want to define here is the root CA in case the backend is using a private root. SPKI or the public key or cert - inline, like it is used in kubeconfig - may work.
However I don't know if this API is the proper place to define root key managed and distribution - there is a working 
group I think working on certs and security. 

I don't disagree that supporting serlf-signed or private root is important - just that PKI config is probably better as a separate concern, not part of backend properties - and can't be based on cross namespace Secret references ( or expectation that each namespace has a secret with that name).

site-src/geps/gep-1282.md
```diff
@@ -98,12 +106,292 @@ Of course, from the section above, you can see that there are many other feature
...
+    
+    // ServiceNames provide multiple named domains which will share a certificate
+    // for the endpoint.
+    // +optional
+    ServiceNames []string `json:"serviceNames,omitempty"`
```
costinm, line 225: Why ? How does it help the caller to know ahead of time the list that will be returned in the cert anyways ?

You mean 'SANNames' - for supporting spiffee or equivalent ? That may be useful. 
Do you mean this as an override of the normal TLS server verification - where gateway talk with foo.ns.svc.cluster.local but get a certificate for example.com - I think it's important to be very clear with this, security is tricky.

site-src/geps/gep-1282.md
youngnick: I think maybe there's some confusion - I don't think I've ever intended this to be the _only_ way that TLS will be enabled to backends. This construct is to cover when _the service owner wants to manage their own TLS_, otherwise the Gateway would be able to contact the service directly. (I also include in that case the mesh-style magic infrastructure that will transparently encrypt connections for you).

Right now, there is no way to represent the (pretty common) case of someone saying "my service is running its own TLS, I need it exposed via a Gateway". That's the primary use case we're trying to solve here - the extra applicatibility is useful but not essential.

I guess another way to say this is - regardless of what service owners _should_ be doing, they absolutely _do_ want to do what this object is providing, already.

site-src/geps/gep-1282.md
youngnick: This is absolutely the Service being used as a backend, as a shortcut for a selector. That's why we called it _Backend_Capabilities, not just _Service_Capabilities. Remember, a Service is both the _frontend_ (what you spend most of your time thinking of it as, because that's what Istio needs), and the _backend_, a bucket of workload endpoints (which is how people who aren't doing service mesh tend to see it more).

Having this reference a Service resource is a way avoid having to store the workload selector in two places, as well as a way to meet people who are using Ingress controllers for north-south traffic more than they are Services for east-west traffic.

site-src/geps/gep-1282.md
youngnick: This Secret reference is not for some other service to connect to this one, it's for the Gateway controller (which already needs cluster-wide access to secrets to be able to do its job). The main, relevant client here is the Gateway controller. 

And even if users _do_ need to do a cross-namespace Secret reference, that's exactly what ReferenceGrant is for, to enable the _controlled_ use of Secrets or other resources across namespaces. It seems reasonable to assume that, for self-signed certs, a CA cert might live in its own namespace, with a ReferenceGrant then being required to allow access to it. Or, of course, a controller could just copy that into relevant namespaces too.

Lastly, this is a CA _certificate chain_, not a _key_, so it's intended to be public. We could use another object (maybe a configmap?) to store it if necessare.

site-src/geps/gep-1282.md
youngnick: In this case, it's to help the _Gateway implementation_ so that it can determine configuration validity before allowing any connections at all. I'm so-so on this one, because using it makes the hostname match a three-part one (Listener, HTTPRoute, BackendCapabilities), rather than a two-part one, which is already hard to do. The upside is that it means that implementations can easily check and make clear _at a configuration level_ when something is not going to work, rather than waiting to have client requests fail and needing to troubleshoot that.

youngnick:
> * having a large object with all the possible properties of a backend seems very dangerous, will be hard to evolve it

This is a fair point, but the other way to do it - having a separate resource for each type of config - also has problems.

Let's review:

Have one large object:
Pros:
- The "add a nil-able struct for each config type" pattern lets each type of information be reasonably orthogonal to the others, and allows for straightforward expansion.
- Having only one resource to watch means there's less code churn for Gateway implementations.
- Having only one extra wrapper object to look for makes it easier for users to know where to look.
Cons:
- Object could become unusably large.

Have more, smaller objects:
Pros:
- Each object is tightly contained, and has a smaller failure domain.
Cons:
- Each extra resource to add is another round of things for each implementation to add.
- Each extra resource is another kubectl command for users looking for info as to what went wrong.

Personally, I'd prefer to stick with the single large object for a while, because I think the implementation code churn and user experience costs will rapidly increase over time.

> * for TLS1.3 - I don't think it's a common use case for a user to go to each backend and specify the settings. Usually TLS1.3 is required for compliance or because the gateway admin requires 1.3 - and that would be set in the gateway settings.

Again, the idea here is to make it so that for a cluster that requires TLS 1.3, but doesn't provide infrastructure to do it automatically, it's _possible_ to use a Gateway implementation to both connect and to enforce those restrictions.

> * I'm concerned with client cert specification - it is pretty tricky for the producer to tell the consumer what to use. I would expect any mTLS from gateway to backend to default to the workload identity or be configured on the client side.

The idea here is to enable clients to give a Gateway some details about what to use to connect. Maybe we don't have it supply a certificate with a keypair, maybe we do. Having a magic keypair that allows you to connect to a service available in the same namespace as the service seems like a security problem, but the other problem is that, if you _do_ want each client of the backend service to use an individual certificate, how do you tell the Gateway which one to use for _this_ backend? It could be that this is the point at which we throw up our hands and say "if you want that, you want a service mesh, SPIFFE and SPIRE, or some other more advanced thing" though.

youngnick:
Okay, after discussion in the community meeting today, we're working on collecting TLS use cases, along with owners etc, in this new document:
https://docs.google.com/document/d/17sctu2uMJtHmJTGtBi_awGB0YzoCLodtR6rUNmKMCs8/edit?usp=sharing

/hold
this until we figure out what the use cases mean.

candita:
In yesterday's meeting we decided to officially drop GEP-1282.

Issue #1428, opened 2022-10-03
website: render Approved GEPs in numerical order instead of alphabetical


mikemorris:
**What would you like to be added**:

Approved GEPs should be rendered in reverse chronological order in the website navigation menu (roughly, sorting by PR number instead of date is likely preferable). This may require adding additional metadata to existing GEP files.

**Why this is needed**:

Currently, Approved GEPs are rendered in alphabetical order, which can be particularly confusing now that we've reached four-digit GEPs.

<img width="248" alt="Screen Shot 2022-10-03 at 5 35 02 PM" src="https://user-images.githubusercontent.com/1149913/193689419-0b045775-48f1-4ef6-a927-0846113581e3.png">

/kind documentation

robscott:
For reference, this is the line that generates this: https://github.com/kubernetes-sigs/gateway-api/blob/main/.golangci.yml#L42.

PR #1438, opened 2022-10-06
website: render approved GEPs in chronological order

Amila-Rukshan:
**What type of PR is this?**

/kind documentation

This will order the navigation menu for approved GEPs.

Fix is using the support from the already installed `mkdocs-awesome-pages-plugin`
configured as it is given here:  https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin#customize-navigation

Now it order the nav items in chronological order:

Screenshot 2022-10-06 at 19 48 56

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1428

```release-note
NONE
```

### DIFF
--- a/site-src/geps/.pages
+++ b/site-src/geps/.pages
@@ -1,1 +1,20 @@
 title: Approved GEPs
+nav:
+    - gep-696.md
+    - gep-709.md
+    - gep-713.md
+    - gep-718.md
+    - gep-724.md
+    - gep-726.md
+    - gep-735.md
+    - gep-746.md
+    - gep-820.md
+    - gep-851.md
+    - gep-917.md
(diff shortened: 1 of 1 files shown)


mikemorris:
Thanks for finding a way to do this @Amila-Rukshan!

I tried experimenting with using the `order` and `sort_type: natural` options from the docs you linked (instead of listing them manually), but it seemed to not work as expected (four-digit GEPs with a leading 1 sorted as less than three-digit GEPs) unless I added a leading zero to the GEP numbers in filenames 😕

Amila-Rukshan:
@mikemorris I also tried `sort_type: natural` and didn't notice what you have noticed though.

PR #1439, opened 2022-10-06
Fix List type aliases

howardjohn:
If we have the outer List type v1alpha2 but the inner Items type v1beta1, informers get confused and end up panicing. This fixes it

**What type of PR is this?**

Fixes #

```release-note
NONE
```


robscott:
Thanks @howardjohn!

/lgtm
/approve

PR #1436, opened 2022-10-05
docs: merge 'custom' and 'implementation-specific' conformance

LCaparelli:
Signed-off-by: Lucas Caparelli <lucas.caparelli@gympass.com>

**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:

This PR merges the concepts of `custom` and `implementation-specific` conformance into `implementation-specific` only in order to simplify the spec. All `custom` conformance mentions were replaced by `implementation-specific`.

**Which issue(s) this PR fixes**:

Fixes #640 

**Does this PR introduce a user-facing change?**:

```release-note
Changes "custom" conformance level to "implementation-specific"
```

PR #1443, opened 2022-10-07
feat: add guard check for go version

### LINKED ISSUE #1442: Add go version check for code generation
**What would you like to be added**: a version check for Go when running code generation

**Why this is needed**: the project's `go.mod` currently uses Go 1.18, but the code generation script must be executed using go 1.19. However, `make verify` runs fine with 1.18 locally, but will fail in CI/CD that is using 1.19.

LCaparelli:
Signed-off-by: Lucas Caparelli <lucas.caparelli@gympass.com>

**What type of PR is this?**
/kind feature

Adds a guard check to ensure correct Go version when running code generation.

Fixes #1442

```release-note
NONE
```

This is a follow-up from https://github.com/kubernetes-sigs/gateway-api/pull/1436#issuecomment-1270749444

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -26,6 +26,12 @@ readonly GOMODCACHE="$(go env GOMODCACHE)"
 readonly GO111MODULE="on"
 readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
+readonly REQUIRED_GO_VER="1.19"
+
+if [[ $(go version | grep ${REQUIRED_GO_VER} -c) -eq 0 ]]; then
+  echo "Go v${REQUIRED_GO_VER} is required to run code generation"
+  exit 1
+fi
 
 export GOMODCACHE GO111MODULE GOFLAGS GOPATH
 
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -28,7 +28,12 @@ readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
 readonly MIN_REQUIRED_GO_VER="1.19"
 
-if go version | perl -ne "exit 0 unless m{go version go([0-9]+.[0-9]+)}; exit 1 if (\$1 >= ${MIN_REQUIRED_GO_VER})"; then
+function go_version_matches {
+  go version | perl -ne "exit 1 unless m{go version go([0-9]+.[0-9]+)}; exit 1 if (\$1 < ${MIN_REQUIRED_GO_VER})"
+  return $?
+}
+
+if ! go_version_matches; then
   echo "Go v${MIN_REQUIRED_GO_VER} or later is required to run code generation"
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


hack/update-codegen.sh
```diff
@@ -26,6 +26,12 @@ readonly GOMODCACHE="$(go env GOMODCACHE)"
 readonly GO111MODULE="on"
 readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
+readonly REQUIRED_GO_VER="1.19"
+
+if [[ $(go version | grep ${REQUIRED_GO_VER} -c) -eq 0 ]]; then
```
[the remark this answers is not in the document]
LCaparelli: This is nice, thanks for the suggestion! Using `< 1.19` also conveniently continues to work if moving forward there's not a strict requirement for 1.20 or later versions. Nice :-)

hack/update-codegen.sh
```diff
@@ -26,6 +26,12 @@ readonly GOMODCACHE="$(go env GOMODCACHE)"
 readonly GO111MODULE="on"
 readonly GOFLAGS="-mod=readonly"
 readonly GOPATH="$(mktemp -d)"
+readonly MIN_REQUIRED_GO_VER="1.19"
+
+if go version | perl -ne "exit 0 unless m{go version go([0-9]+.[0-9]+)}; exit 1 if (\$1 >= ${MIN_REQUIRED_GO_VER})"; then
```
LCaparelli, line 31: @shaneutt had to change it a little bit, the perl script should exit 0 in order to trigger the if statement, played a bit with the boolean logic to adjust for clarity after that. Also adjusted the regexp to work with first minor releases, which don't have the patch component:
```
❯ gvm use go1.19
Now using version go1.19
❯ go version
go version go1.19 linux/amd64
❯ gvm use go1.19.2
Now using version go1.19.2
❯ go version
go version go1.19.2 linux/amd64
```

Wdyt?

shaneutt: My only thinking is that it might be a _little_ tricky for readers to understand at a glance because we're mixing an explicit success with an explicit failure _and_ "success" (`exit 0`) ultimately means "failure" in this if conditional. It's all kinda minor because at the end of the day, any time you're working with bash scripts running perl/sed/awk it always gets a little messy, but I do have a suggestion for your considerations which I think may help improve the "at a glance" readability here for future code readers:

```bash
function go_version_matches() {
    go version | perl -ne "exit 1 unless m{go version go([0-9]+.[0-9]+)}; exit 1 if (\$1 < 1.19)"
    return $?
}

if ! go_version_matches; then
    echo "Go v1.19 or later is required to run code generation"
    exit 1
fi
```

This way `exit 1` is just failure, and using a function it makes the conditional read a bit more clear. LMKWYT?

LCaparelli: Right, makes sense!

shaneutt (approved):

shaneutt (approved):
/lgtm

PR #1446, opened 2022-10-10
add ListenerConditionAccepted, update docs from "Detached"

### LINKED ISSUE #1110: "good" state for ListenerStatus feels logically inverted, difficult to understand

- Add [`ListenerConditionType`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1alpha2.ListenerConditionType) `Attached` (or possibly `Accepted` for parity with other objects, ref #1111 ), deprecate `Detached`

**Why this is needed**:

From a quick glance, seeing a `ListenerStatus` with `Detached` and `Conflicted` conditions on a Gateway can be a red herring leading an operator to initially think a `Listener` is not working properly - this may be particularly confusing for new users unfamiliar with checking for the "negative" case of `Detached: {status: false, reason: Attached}`, `Conflicted: {status: false, reason: NoConflicts}` meaning that the Listener is actually in a "good" state.

This potential confusion about whether a Listener is working as expected is slightly exacerbated by the `ListenerStatus` field being a child of [`GatewayStatus`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.GatewayStatus) rather than being a `status` field directly on [`Listener`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.Listener).

I initially considered if `Conflicted` should be changed too, but I think just this change alone could make the status conditions for a Listener feel a bit more intuitve.

mikemorris:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind api-change
/kind deprecation

Implements the switch from `ListenerConditionDetached` to a new `ListenerConditionAttached`, as recommended in [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/).

Fixes #1110, refs #1364 

```release-note
Adds `ListenerConditionAccepted` and `ListenerReasonAccepted`
Deprecates `ListenerConditionDetached` and `ListenerReasonAttached`
```

Now that we've settled on the expected behavior, these expectations should likely be added to relevant conformance tests.

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -352,9 +352,7 @@ const (
 	// interoperability.
 	ListenerConditionAccepted ListenerConditionType = "Accepted"
 
-	// The "Detached" ListenerConditionType is deprecated and "Accepted" should
-	// be used instead.
-	// TODO: is there a need/way to add a kubebuilder annotation here?
+	// Deprecated: use "Accepted" instead.
 	ListenerConditionDetached ListenerConditionType = "Detached"
 
 	// This reason is used with the "Accepted" condition when the condition is
@@ -362,9 +360,9 @@ const (
 	ListenerReasonAccepted ListenerConditionReason = "Accepted"
 
 	// This reason is used with the "Detached" condition when the condition is
-	// False. This reason is deprecated and ListenerReasonAccepted should be used
-	// with ListenerConditionAccepted instead.
-	// TODO: is there a need/way to add a kubebuilder annotation here?
+	// False.
+	//
+	// Deprecated: use the "Accepted" condition with reason "Accepted" instead.
 	ListenerReasonAttached ListenerConditionReason = "Attached"
 
 	// This reason is used with the "Accepted" condition when the Listener
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -62,7 +62,7 @@ type Listener = v1beta1.Listener
 // ProtocolType defines the application protocol accepted by a Listener.
 // Implementations are not required to accept all the defined protocols.
 // If an implementation does not support a specified protocol, it
-// should raise a "Detached" condition for the affected Listener with
+// should set the "Accepted" condition to "False" for the affected Listener with
 // a reason of "UnsupportedProtocol".
 //
 // Core ProtocolType values are listed in the table below.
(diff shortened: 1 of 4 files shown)


shaneutt (commented):
Looking good. What are you looking for RE: kubebuilder annotations?

mikemorris:
@shaneutt I'm curious if something like the following can/should be used more granularly, like for these constants?
https://github.com/kubernetes-sigs/gateway-api/blob/006a257d4dfd6bf3bea938def6e20e7ddf256df2/apis/v1alpha2/httproute_types.go#L29
https://github.com/kubernetes-sigs/gateway-api/blob/006a257d4dfd6bf3bea938def6e20e7ddf256df2/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml#L28-L30

Or alternatively if there's some preferred convention like https://github.com/kubernetes-sigs/gateway-api/blob/751ccfc277d8f85a4da342d8864c1b3431e86813/apis/v1beta1/zz_generated.register.go#L34

shaneutt:
> @shaneutt I'm curious if something like the following can/should be used more granularly, like for these constants?
> 
> https://github.com/kubernetes-sigs/gateway-api/blob/006a257d4dfd6bf3bea938def6e20e7ddf256df2/apis/v1alpha2/httproute_types.go#L29
> 
> https://github.com/kubernetes-sigs/gateway-api/blob/006a257d4dfd6bf3bea938def6e20e7ddf256df2/config/crd/standard/gateway.networking.k8s.io_httproutes.yaml#L28-L30
> 
> Or alternatively if there's some preferred convention like
> 
> https://github.com/kubernetes-sigs/gateway-api/blob/751ccfc277d8f85a4da342d8864c1b3431e86813/apis/v1beta1/zz_generated.register.go#L34

Ah, now I see what you're getting at. I am only aware of the `kubebuilder:deprecatedversion` tag as per the [kubebuilder CRD generation documentation](https://book.kubebuilder.io/reference/markers/crd.html?highlight=deprecat#crd-generation), which does not apply here because it marks the entire version as deprecated. I am not aware of any way to call a specific field deprecated. I think we can just use the `godoc` you've already provided explaining that it's deprecated for now, maybe we can open/find an upstream issue to add something that automates this for us in kubebuilder later.

mikemorris:
Removed TODOs and switched deprecations warnings from passive to active voice, think this should be ready to go now.

robscott:
Thanks @mikemorris!

/lgtm
/approve

apis/v1alpha2/gateway_types.go
```diff
@@ -337,45 +337,54 @@ const (
...
 	//
-	// * "Attached"
+	// * "PortUnavailable"
+	// * "UnsupportedProtocol"
+	// * "UnsupportedAddress"
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
+	ListenerConditionAccepted ListenerConditionType = "Accepted"
```
howardjohn, line 353: The docs here (L330) are wrong now, the polarity should be inverted

Issue #1111, opened 2022-04-13
discrepancies between Accepted, Ready, and Detached condition types and reasons


mikemorris:
Several different [`Kubernetes meta/v1.Condition`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#condition-v1-meta) types are defined by the spec, and some use differing terminology to communicate apparently similar status.

**What would you like to be added**:

- [ ] Add `Pending` GatewayClassConditionReason, deprecate `Waiting` (for parity with ListenerConditionType)
  - Tracking in https://github.com/kubernetes-sigs/gateway-api/issues/1449
- [x] Possibly add an `Accepted` GatewayConditionType, with `Accepted`, `Invalid` and `Pending` GatewayConditionReason to disambiguate the status of the object being accepted by the Gateway Controller versus being ready to route traffic.
  - [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/) proposed replacing the current `GatewayConditionScheduled` with a new `GatewayConditionAccepted`
  - https://github.com/kubernetes-sigs/gateway-api/pull/1447
- [x] Add `Attached` or `Accepted` ListenerConditionType, deprecate `Detached`
  - `Attached` could still be a reasonable choice to clarify that this object isn't submitted to the controller individually, but rather evaluated as part of a Gateway object.
  - If choosing `Accepted`, add `Accepted` ListenerConditionReason, deprecate `Attached`
  - See https://github.com/kubernetes-sigs/gateway-api/issues/1110 for more detail
  -  [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/) proposed switching this to `Accepted`
  - https://github.com/kubernetes-sigs/gateway-api/pull/1446
- [ ] ~~Add `InvalidParameters` ListenerConditionReason, prefer over more generic `Invalid` when applicable?~~
- [x] Add `Accepted` RouteConditionReason, deprecate `Admitted` (for parity with GatewayClassConditionType)
  - `Admitted` was only in the v1alpha1 spec, and had already been removed in https://github.com/kubernetes-sigs/gateway-api/pull/839
  -  Documented `Accepted` properly in https://github.com/kubernetes-sigs/gateway-api/pull/1114
- [ ] ~~Possibly add `Ready` RouteConditionType, although this is known to be difficult to implement accurately.~~
  - Tracked in https://github.com/kubernetes-sigs/gateway-api/issues/1156
[...]

PR #1447, opened 2022-10-10
add GatewayConditionAccepted and GatewayReasonAccepted

mikemorris:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind api-change
/kind deprecation

Implements the recommendation in [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/) to replace `GatewayConditionScheduled` with `GatewayConditionAccepted`

Refs #1111, #1364

```release-note
Adds GatewayConditionAccepted and GatewayReasonAccepted
Deprecates GatewayConditionScheduled and GatewayReasonScheduled
```

This status condition does not appear to be currently checked in conformance tests, which mostly rely on the`Ready` condition instead. We could either add checks for this condition now, or in a subsequent pull request.

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -214,9 +214,7 @@ const (
 	// interoperability.
 	GatewayConditionAccepted GatewayConditionType = "Accepted"
 
-	// The "Scheduled" GatewayConditionType is deprecated and "Accepted" should
-	// be used instead.
-	// TODO: is there a need/way to add a kubebuilder annotation here?
+	// Deprecated: use "Accepted" instead.
 	GatewayConditionScheduled GatewayConditionType = "Scheduled"
 
 	// This reason is used with the "Accepted" condition when the condition is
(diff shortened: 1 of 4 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -212,17 +212,28 @@ const (
...
 
+	// This reason is used with the "Accepted" condition when the condition is
+	// True.
+	GatewayReasonAccepted GatewayConditionReason = "Accepted"
+
 	// This reason is used with the "Scheduled" condition when the condition is
-	// true.
+	// True. This reason is deprecated and GatewayReasonAccepted should be used
+	// with GatewayConditionAccepted instead.
+	// TODO: is there a need/way to add a kubebuilder annotation here?
```
shaneutt, line 229: I am only aware of the `kubebuilder:deprecatedversion` tag as per the [kubebuilder CRD generation documentation](https://book.kubebuilder.io/reference/markers/crd.html?highlight=deprecat#crd-generation), which does not apply here because it marks the entire version as deprecated. I am not aware of any way to call a specific field deprecated. I think we can just use the `godoc` you've already provided explaining that it's deprecated for now, maybe we can open/find an upstream issue to add something that automates this for us in kubebuilder later.

mikemorris:
Removed TODOs and switched deprecation warnings from passive to active voice, think this should be ready to go now.

apis/v1alpha2/gateway_types.go
robscott: Something similar was proposed but ended up getting turned down: https://github.com/kubernetes/enhancements/pull/3340.

robscott:
Thanks!

/lgtm
/approve

howardjohn:
Give v0.6.0 is not a new API version, what is the expected behavior of implementors here? Must set both?

PR #1457, opened 2022-10-14
Webhook: validate the combination of port, protocol, and hostname are unique for each listener.

### LINKED ISSUE #847: Webhook: Validate unique port, protocol, and hostname per listener
The validating webhook should ensure that port, protocol, and hostname are unique among Gateway listeners. For example, 2 listeners for foo.com would be valid if they were on different ports or protocols, but not if port and protocol were also identical.

**Why this is needed**:
To provide a better user experience.

gyohuangxin:
**What type of PR is this?**

/kind feature

Add validation to validate the combination of port, protocol, and hostname are unique for each listener.

Fixes #847

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -140,7 +140,7 @@ func ValidateListenerNames(listeners []gatewayv1b1.Listener, path *field.Path) f
 // unique for each listener.
 func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
-	hostnameProtocolPortSets:= sets.Set[string]{}
+	hostnameProtocolPortSets := sets.Set[string]{}
 	for i, listener := range listeners {
 		hostname := new(gatewayv1b1.Hostname)
 		if listener.Hostname != nil {
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -148,12 +148,11 @@ func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.
 		}
 		protocol := listener.Protocol
 		port := listener.Port
-		
 		hostnameProtocolPort := fmt.Sprintf("%s:%s:%d", *hostname, protocol, port)
 		if hostnameProtocolPortSets.Has(hostnameProtocolPort) {
 			errs = append(errs, field.Forbidden(path.Index(i), fmt.Sprintln("combination of port, protocol, and name must be unique for each listener")))
 		}
 		hostnameProtocolPortSets.Insert(hostnameProtocolPort)
-	} 
+	}
 	return errs
 }
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -136,7 +136,7 @@ func ValidateListenerNames(listeners []gatewayv1b1.Listener, path *field.Path) f
 	return errs
 }
 
-// validateHostnameProtocolPort validates that the combination of port, protocol, and name are
+// validateHostnameProtocolPort validates that the combination of port, protocol, and hostname are
 // unique for each listener.
 func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
@@ -150,7 +150,7 @@ func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.
 		port := listener.Port
 		hostnameProtocolPort := fmt.Sprintf("%s:%s:%d", *hostname, protocol, port)
 		if hostnameProtocolPortSets.Has(hostnameProtocolPort) {
-			errs = append(errs, field.Forbidden(path.Index(i), fmt.Sprintln("combination of port, protocol, and name must be unique for each listener")))
+			errs = append(errs, field.Forbidden(path.Index(i), fmt.Sprintln("combination of port, protocol, and hostname must be unique for each listener")))
 		} else {
 			hostnameProtocolPortSets.Insert(hostnameProtocolPort)
 		}
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package validation
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
@@ -70,6 +71,7 @@ func validateGatewayListeners(listeners []gatewayv1b1.Listener, path *field.Path
 	errs = append(errs, validateListenerHostname(listeners, path)...)
 	errs = append(errs, ValidateTLSCertificateRefs(listeners, path)...)
 	errs = append(errs, ValidateListenerNames(listeners, path)...)
+	errs = append(errs, validateHostnameProtocolPort(listeners, path)...)
 	return errs
 }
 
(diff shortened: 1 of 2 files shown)


gyohuangxin:
@robscott I make it a draft PR because I have some question:

1. As you mentioned in #847: 

> The validating webhook should ensure that port, protocol, and hostname are unique among Gateway listeners

but then you give an example: 

> "2 listeners for foo.com would be valid if they were on different ports or protocols, but not if port and protocol were also identical."

They are a bit inconsistent, but my understanding is the webhook should separately validate unique port, protocol per hostname for gateway listeners. Is it correct? Is it needed to verify unique port, protocol across hostnames? Is it needed to verify unique hostname per listener?

2. I add validations in v1beta1 apis, should I add them in v1alpha1 apis too?

shaneutt:
@gyohuangxin where are we at with this one? Is there anything we can do to help the progression here? Looks like maybe we're still waiting on some feedback requested from @robscott?

gyohuangxin:
Yes, it's on hold because I'm still confused with the first question,  I'm not sure my thought is close to the reality.


> @robscott I make it a draft PR because I have some question:
> 
> 1. As you mentioned in [Webhook: Validate unique port, protocol, and hostname per listener #847](https://github.com/kubernetes-sigs/gateway-api/issues/847):
> 
> > The validating webhook should ensure that port, protocol, and hostname are unique among Gateway listeners
> 
> but then you give an example:
> 
> > "2 listeners for foo.com would be valid if they were on different ports or protocols, but not if port and protocol were also identical."
> 
> They are a bit inconsistent, but my understanding is the webhook should separately validate unique port, protocol per hostname for gateway listeners. Is it correct? Is it needed to verify unique port, protocol across hostnames? Is it needed to verify unique hostname per listener?
>

shaneutt:
> Yes, it's on hold because I'm still confused with the first question, I'm not sure my thought is close to the reality.

Ok, I put this on next weeks community sync to discuss as part of the "old PR roundup" so we can talk about it there.

apis/v1beta1/validation/gateway.go
```diff
@@ -117,3 +119,41 @@ func ValidateTLSCertificateRefs(listeners []gatewayv1b1.Listener, path *field.Pa
...
 	return errs
 }
+
+
+// validateHostnameProtocolPort validates listener protocols or ports
+// must be different for the same hostname
+func ValidateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	hostnameProtocolMap := make(map[gatewayv1b1.Hostname][]gatewayv1b1.ProtocolType)
+	hostnamePortMap := make(map[gatewayv1b1.Hostname][]gatewayv1b1.PortNumber)
```
robscott, line 129: Hey @gyohuangxin, I think there's a simpler way to do this. All that we really want to do is ensure that the combination of port, protocol, and name are unique. So for example, the following would be valid:

```yaml
- port: 80
  hostname: foo.com
  protocol: HTTP
- port: 443
  hostname: foo.com
  protocol: HTTP
```

But the following would not be valid:

```yaml
- port: 80
  hostname: foo.com
  protocol: HTTP
- port: 80
  hostname: foo.com
  protocol: HTTP
```

I think what you have is slightly different than that, requiring that for the same hostname (foo.com), both port _and_ protocol must be different for each listener. The spec as it's written today only requires that one of them is different.

A simple way to accomplish this would be to use something like `fmt.Sprintf("%s:%s:%d", protocol, hostname, port)` for each listener and then use k8s sets util to see if the value for that listener duplicated any listeners you'd already looked at within the same Gateway. Hopefully that makes sense, sorry for the delay on this!

robscott:
Thanks for the work on this @gyohuangxin! Sorry I lost track of this, don't hesitate to ping me if I can help explain anything better.

apis/v1beta1/util/validation/gateway.go
```diff
@@ -0,0 +1,41 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
hzxuzhonghu, line 2: 2023

apis/v1beta1/util/validation/gateway.go
```diff
@@ -0,0 +1,41 @@
...
+		if eachItem == *item {
+			return true
+		}
+	}
+	return false
+}
+
+// ContainsInPortSlice checks whether the provided Port
+// is in the target Port slice.
+func ContainsInPortSlice(items []gatewayv1b1.PortNumber, item *gatewayv1b1.PortNumber) bool {
```
hzxuzhonghu, line 34: pass value type of PortNumber

apis/v1beta1/util/validation/gateway.go
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+
+package validation
+
+import gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+// ContainsInProtocolSlice checks whether the provided Protocol
+// is in the target Protocol slice.
+func ContainsInProtocolSlice(items []gatewayv1b1.ProtocolType, item *gatewayv1b1.ProtocolType) bool {
```
hzxuzhonghu, line 23: value type instead of pointer

apis/v1beta1/validation/gateway.go
arkodg: should this webhook attempt to reject one of these tuples (80, http, *.com) and (80, http, *.foo.com) :) ?

gyohuangxin:
@robscott @shaneutt Can you review this PR? Thanks

apis/v1beta1/validation/gateway.go
```diff
@@ -133,3 +135,24 @@ func ValidateListenerNames(listeners []gatewayv1b1.Listener, path *field.Path) f
...
+		if listener.Hostname != nil {
+			hostname = listener.Hostname
+		}
+		protocol := listener.Protocol
+		port := listener.Port
+		hostnameProtocolPort := fmt.Sprintf("%s:%s:%d", *hostname, protocol, port)
+		if hostnameProtocolPortSets.Has(hostnameProtocolPort) {
+			errs = append(errs, field.Forbidden(path.Index(i), fmt.Sprintln("combination of port, protocol, and name must be unique for each listener")))
+		}
+		hostnameProtocolPortSets.Insert(hostnameProtocolPort)
```
LiorLieberman, line 155: small nit: consider not calling Insert if the set has the combination already.

gyohuangxin: @LiorLieberman Thanks for your comments, updated.

apis/v1beta1/validation/gateway.go
LiorLieberman: @arkodg  I think it will very hard to accomplish. How would you decide if regexA (*.com in this case) is overlapping with regexB(*foo.com).

LiorLieberman:
Thanks @gyohuangxin ! This Looks good to me.
We probably want to change the PR description to match the new logic - ensuring Listener's port-protocol-hostname combinations are unique within an individual Gateway.

over to @robscott for a final review and approval.

gyohuangxin:
> Thanks @gyohuangxin ! This Looks good to me. We probably want to change the PR description to match the new logic - ensuring Listener's port-protocol-hostname combinations are unique within an individual Gateway.
> 
> over to @robscott for a final review and approval.

Sorry for missing the PR description, updated. Thanks for reminding 💯

apis/v1beta1/validation/gateway.go
arkodg: yeah its hard, but I wanted to highlight the case.
Since the API calls out any regex match to be implementation specific, even adding language in the API for this might be tricky
cc @robscott

arkodg (approved):
LGTM thanks !

dprotaso:
Based on this comment it seems like we should take into consideration the tls properties attached to a listener

https://github.com/kubernetes-sigs/gateway-api/issues/1842#issuecomment-1484615419

'

gyohuangxin:
Thanks for your comments @dprotaso . As said in @robscott 's https://github.com/kubernetes-sigs/gateway-api/issues/1842#issuecomment-1496560365, the "Hostname" can be different for shared Port and Protocol, like the TLS example on the website: https://gateway-api.sigs.k8s.io/guides/tls/#wildcard-tls-listeners

So, I think it can cover the use cases. Please let me know if there are some use cases that use the same combination of port, protocol and hostname but use distinct `tls` fields.

apis/v1beta1/validation/gateway_test.go
```diff
@@ -136,14 +136,37 @@ func TestValidateGateway(t *testing.T) {
...
+					},
 				)
 			},
 			expectErrsOnFields: []string{"spec.listeners[1].name"},
 		},
+		"combination of port, protocol, and name are not unique for each listener": {
+			mutate: func(gw *gatewayv1b1.Gateway) {
+				hostnameFoo := gatewayv1b1.Hostname("foo.com")
+				gw.Spec.Listeners[0].Name = "foo"
+				gw.Spec.Listeners[0].Hostname = &hostnameFoo
```
robscott, line 156: Since hostname is optional can you add one more test case that shows that the same setup fails when both listeners have separate hostnames?

robscott: Similarly, it would be helpful to have additional test cases where everything was the same except for one of the fields (port, protocol, or hostname) to ensure those are all still valid.

robscott (commented):
Thanks @gyohuangxin! Would like a few more test cases but otherwise LGTM.

apis/v1beta1/validation/gateway_test.go
gyohuangxin: > Since hostname is optional can you add one more test case that shows that the same setup fails when both listeners have separate hostnames?

Do you mean to test the combination of port and protocol are unique when hostnames not set? If yes, I've add one: https://github.com/kubernetes-sigs/gateway-api/pull/1457/commits/d75ab5690eab043ce5e3dd3e4f495ec5e398b2b0#diff-e3c0a139e14037364db162e955588e8153e795f4cbe84b6fce32149d62ca2a73R170-R184



> Similarly, it would be helpful to have additional test cases where everything was the same except for one of the fields (port, protocol, or hostname) to ensure those are all still valid.

Updated, thank you!

apis/v1beta1/validation/gateway_test.go
```diff
@@ -136,14 +136,117 @@ func TestValidateGateway(t *testing.T) {
...
+						Name:     "bar",
+						Hostname: &hostnameFoo,
+						Protocol: gatewayv1b1.HTTPProtocolType,
+						Port:     8080,
+					},
+				)
+			},
+			expectErrsOnFields: nil,
+		},
+		"hostname is unique when protoco and port are the same": {
```
robscott, line 203: Tiny nit:
```suggestion
		"hostname is unique when protocol and port are the same": {
```

apis/v1beta1/validation/gateway.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
robscott, line 2: Generally this date represents when the file was created, not when it was updated, I'd leave it unchanged.
```suggestion
Copyright 2021 The Kubernetes Authors.
```

robscott (commented):
Thanks @gyohuangxin! Just a couple tiny nits but otherwise LGTM.

apis/v1beta1/validation/gateway_test.go
gyohuangxin: Fixed.

apis/v1beta1/validation/gateway.go
gyohuangxin: Fixed.

robscott:
Thanks @gyohuangxin!

/approve
/ok-to-test

PR #1461, opened 2022-10-17
Use correct reason in docs when no ReferenceGrant for cert ref

nathancoleman:
**What type of PR is this?**

/kind documentation

#1297 and #1362 have clarified that a cross-namespace certificate reference without a `ReferenceGrant` permitting it should use `RefNotPermitted` instead of `InvalidCertificateRef`; however, there is one remaining reference to `InvalidCertificateRef` for this purpose in the docs.

```release-note
NONE
```


mlavacca (approved):
/lgtm for what it may be worth.

PR #1460, opened 2022-10-17
Webhook: validate certificateRefs when tls is set and mode is Terminate.

gyohuangxin:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

When the TLSModeType is set to terminate, there is always the need for a valid certificate to terminate the TLS connection.

**Which issue(s) this PR fixes**:

Fixes #1441

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1463, opened 2022-10-18
conformance: fix GatewayUnsupportedRouteKind HTTPS listeners without TLS config

michaelbeaumont:
**What type of PR is this?**
/kind bug

The resources created by the `GatewayUnsupportedRouteKind` test aren't valid because if protocol is HTTPS, the TLS field must be set. I think #1460 would have caught this.

They don't need to be HTTPS for this test however.

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-unsupported-route-kind.go
+++ b/conformance/tests/gateway-unsupported-route-kind.go
@@ -39,7 +39,7 @@ var GatewayUnsupportedRouteKind = suite.ConformanceTest{
 		t.Run("Gateway listener should have a false ResolvedRefs condition with reason InvalidRouteKinds and no supportedKinds", func(t *testing.T) {
 			gwNN := types.NamespacedName{Name: "gateway-only-unsupported-route-kind", Namespace: "gateway-conformance-infra"}
 			listeners := []v1beta1.ListenerStatus{{
-				Name:           v1beta1.SectionName("https"),
+				Name:           v1beta1.SectionName("http"),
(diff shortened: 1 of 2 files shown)


gyohuangxin:
@michaelbeaumont Thanks for raising this PR, but I don't think we have a validation to check TLS config must be set when the protocol is HTTPS so far. My PR https://github.com/kubernetes-sigs/gateway-api/pull/1460 is to check the certificateRefs must be set and not empty when tls is set and tls mode is terminate.

michaelbeaumont:
@gyohuangxin From reading the docs, i.e. the field comments, I would say you could add to your validation that TLS _must_ be set when HTTPS is the protocol. `Terminate` is the default mode and when `Terminate` is set, `certificateRefs` must be set, right?

> TLS is the TLS configuration for the Listener. This field is required if the Protocol field is “HTTPS” or “TLS”.

michaelbeaumont:
> @michaelbeaumont Thanks for raising this PR, but I don't think we have a validation to check TLS config must be set when the protocol is HTTPS so far. My PR #1460 is to check the certificateRefs must be set and not empty when tls is set and tls mode is terminate.

Ah ok, I commented on that PR

mlavacca (approved):
I think you are right @michaelbeaumont, thanks for catching this!

gyohuangxin (approved):
LGTM after talking in another PR.

shaneutt (approved):
/lgtm

PR #1476, opened 2022-10-25
conformance: all cases for which a TLS secret can be invalid are tested

### LINKED ISSUE #1437: Improve the Gateway `ResolvedRefs` condition conformance tests

#1419 improved the Gateway `ResolvedRefs` condition documentation by specifying in which cases the reason `InvalidCertificateRef` must be used:

https://github.com/kubernetes-sigs/gateway-api/blob/38a295bb0ad65b1b8cfc3610dad89754259f6623/apis/v1beta1/gateway_types.go#L748-L759

To ensure full compliance with this specification, the conformance tests related to the gateway `ResolvedRefs` condition should be improved to ensure that all the following cases are tested:

* the object referenced by the TLS configuration does not exist (already tested [here](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/gateway-secret-missing-referenced-secret.go))
* the object referenced by the TLS configuration is invalid because of:
  * unsupported group (to be tested)
  * unsupported kind (to be tested)
  * the data within the resource is malformed (to be tested)

**Why this is needed**:

When all the above conditions are tested, all the reasons that can lead the Gateway condition `ResolvedRefs` to be false can be considered thoroughly tested.

mlavacca:
**What type of PR is this?**

/kind feature

All the cases for which a TLS secret can be invalid are tested.

Fixes #1437 

```release-note
NONE
```


conformance/tests/gateway-invalid-tls-certificateref.go
```diff
@@ -0,0 +1,81 @@
...
+				Reason: string(v1beta1.ListenerReasonInvalidCertificateRef),
+			}},
+		}}
+
+		testCases := []struct {
+			name                  string
+			gatewayNamespacedName types.NamespacedName
+		}{
+			{
+				name:                  "Inexistent secret referenced as CertificateRef in a Gateway listener",
```
shaneutt, line 57: While "inexistent" is understandable, I think it's just more common to use "nonexistent". For instance you'll find in this repo several instances of "nonexistent" to refer to this state but no instances of "inexistent" so far.

conformance/tests/gateway-invalid-tls-certificateref.yaml
```diff
@@ -0,0 +1,89 @@
...
+            name: malformed-certificate
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: malformed-certificate
+  namespace: gateway-conformance-infra
+data:
+  tls.crt: SGVsbG8gd29ybGQK
+  tls.key: SGVsbG8gd29ybGQK
```
shaneutt, line 88: For future readers it may be helpful to put a comment here explaining _how_ this certificate is malformed, e.g.:

```suggestion
  # certificate malformed due to identical cert and key
  tls.crt: SGVsbG8gd29ybGQK
  tls.key: SGVsbG8gd29ybGQK
```

conformance/utils/suite/suite.go
```diff
@@ -139,6 +139,8 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	t.Logf("Test Setup: Applying programmatic resources")
 	secret := kubernetes.MustCreateSelfSignedCertSecret(t, "gateway-conformance-web-backend", "certificate", []string{"*"})
 	suite.Applier.MustApplyObjectsWithCleanup(t, suite.Client, suite.TimeoutConfig, []client.Object{secret}, suite.Cleanup)
+	secret = kubernetes.MustCreateSelfSignedCertSecret(t, "gateway-conformance-infra", "valid-certificate", []string{"*"})
+	suite.Applier.MustApplyObjectsWithCleanup(t, suite.Client, suite.TimeoutConfig, []client.Object{secret}, suite.Cleanup)
```
shaneutt, line 143: So with this change we'll have `certificate` and `valid-certificate`. Technically both of them are valid. Should we maybe update the naming here to indicate what's different about these and what they're for?

Perhaps we can even just use the existing `certificate` resource rather than creating a new one?

mlavacca: I changed the name to highlight the usage. We need another certificate in `gateway-conformance-infra` to be able to check intra-namespace certificate reference. If we use the certificate in `gateway-conformance-web-backend`, we would also need a `ReferenceGrant`, but these tests are not about it.

shaneutt (approved):
/lgtm

Issue #1483, opened 2022-10-27
Define if or how TCPRoute might be used for GAMMA configuration


mikemorris:
**What would you like to be added**:

[GEP-1426](https://github.com/kubernetes-sigs/gateway-api/pull/1426) focuses exclusively on how to configure Layer 7 routing rules between services in a mesh using `HTTPRoute`. We should specify if or how `TCPRoute` should also be supported for Layer 4 routing in GAMMA.

**Why this is needed**:

Clarity is needed on whether it should be possible to use `TCPRoute` for GAMMA routing configuration (e.g. to weight traffic between different backends, but only at L4), what should happen if both an `HTTPRoute` and `TCPRoute` would attempt to target the same service as a `parentRef`, and if `TCPRoute` should be a preferred UX for some use cases or mesh implementations (e.g. if L7 traffic management is not supported).

howardjohn:
Fixed by https://github.com/kubernetes-sigs/gateway-api/pull/1492 - don't believe github UI, this merged

shaneutt:
Sounds like we can close this then, if that's not the case please feel free to re-open. :vulcan_salute:

PR #1492, opened 2022-10-28
Mesh: allow non-HTTP types

howardjohn:
**What type of PR is this?**

/kind gep

This expands the GEP to cover non-HTTP types

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1483

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1426.md
+++ b/site-src/geps/gep-1426.md
@@ -133,7 +133,7 @@ A `Service` may only be specified as a `parentRef` of an `xRoute` if it is:
     * Resolving conflicts from multiple `HTTPRoutes` attempting to target the same `Service` as a `parentRef` should follow the existing [conflict guidelines](https://gateway-api.sigs.k8s.io/concepts/guidelines/#conflicts).
       * When an existing `xRoute` targeting the `Service` as a `parentRef` has already been accepted, any `xRoute` with a newer creation timestamp should be rejected, because the match target considered for the specificity rules (the `Service`) is identical.
       * If creation timestamps are identical (from applying both routes simultaneously), only the `xRoute` with a `name` appearing first in alphabetical order should be accepted.
-      * Any rejected `xRoute` should set an `Accepted` condition with `status: "False"`.
+      * Any rejected `xRoute` must set an `Accepted` condition with `status: "False"`.
     * This requirement may be relaxed later to follow the existing [precedence rules for merging configuration](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule), but is intended to simplify initial implementation.
     * The namespace colocation requirement effectively forecloses using this strategy to delegate route management, but there is service mesh [user interest](https://github.com/istio/istio/issues/22997) in merging separately-defined routes together to manage large configurations, multiple versions or canary deployments for a single mesh service.
 * NOT already specified as a `backendRef` of any _other_ `xRoute` with a `Service` `parentRef`
--- a/site-src/geps/gep-1426.md
+++ b/site-src/geps/gep-1426.md
@@ -150,13 +150,14 @@ The logic described here is a bit verbose, but could be validated with conforman
 
 All types currently defined in the gateway-api core (`HTTP`, `GRPC`, `TCP`, `TLS`, and `UDP`) are available for use in a Mesh implementation.
 
-If multiple routes with of different type both bind to the same Service and Port pair, only a single route type should be applied. The rejected routes should be ignored and have the `RouteConditionAccepted` status set to the (new) reason `RouteReasonConflicted`.
+If multiple routes with different types both bind to the same Service and Port pair, only a single route type should be applied. The rejected routes should be ignored and have the `RouteConditionAccepted` status set to the (new) reason `RouteReasonConflicted`.
 
 Route type specificity is defined in the following order (first one wins):
 
-1. HTTPRoute
-2. TLSRoute
-3. TCPRoute
+1. GRPCRoute
+2. HTTPRoute
+3. TLSRoute
+4. TCPRoute
 
 Because UDP is its own protocol, it is orthogonal to these precedence order. Since there is only one UDP-based route, there is currently no conflicts possible; if other UDP-based routes are added a similar ordering will be defined.
 
--- a/site-src/geps/gep-1426.md
+++ b/site-src/geps/gep-1426.md
@@ -1,11 +1,11 @@
-# GEP-1426: Routes Mesh Binding
+# GEP-1426: xRoutes Mesh Binding
 
 * Issue: [#1294](https://github.com/kubernetes-sigs/gateway-api/issues/1294)
 * Status: Provisional
 
 ## Overview
 
-Similar to how `Routes` bind to `Gateways` and manage North/South traffic flows in Gateway API’s ingress use-case, it would be natural to adopt a similar model for traffic routing concerns in service mesh deployments. The purpose of this GEP is to add a mechanism to the Gateway API spec for the purpose of associating the various `Route` types to a service mesh and offering a model for service owners to manage traffic splitting configurations.
+Similar to how `xRoutes` bind to `Gateways` and manage North/South traffic flows in Gateway API’s ingress use-case, it would be natural to adopt a similar model for traffic routing concerns in service mesh deployments. The purpose of this GEP is to add a mechanism to the Gateway API spec for the purpose of associating the various `xRoute` types to a service mesh and offering a model for service owners to manage traffic splitting configurations.
 	
 This GEP is intended to establish an implementable, but experimental, baseline for supporting basic service mesh traffic routing functionality through the Gateway API spec.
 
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-1426.md
```diff
@@ -114,37 +113,44 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
+The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `Route` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `Route` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
+
+### Route types
+
+All types currently defined in the gateway-api core (`HTTP`, `GRPC`, `TCP`, `TLS`, and `UDP`) are available for use in a Mesh implementation.
+
+If multiple routes with of different type both bind to the same Service and Port pair, the same resolution rules as defined in [`listener.allowedRoutes`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.AllowedRoutes) apply.
```
howardjohn, line 153: This is actually ambiguous in Gateway as far as I can tell. I asked for clarity and will update either the `allowedRoutes`, this sentence, or both once I understand the desired intent.

site-src/geps/gep-1426.md
```diff
@@ -114,37 +113,44 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
 
-The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
+The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `Route` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `Route` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
+
+### Route types
+
+All types currently defined in the gateway-api core (`HTTP`, `GRPC`, `TCP`, `TLS`, and `UDP`) are available for use in a Mesh implementation.
```
mikemorris, line 151: Do any reference implementations exist for UDP routes in service mesh, or something approximating a TLS route that differs in behavior from HTTP or TCP routes?

howardjohn: TLSRoute largely only makes sense if the parentRef doesn't have a VIP _or_ we allow `hostnames` to be set as well.

UDP - Not sure, perhaps Cilium could.. Istio is planning to but it will be 1-2 years probably still

youngnick: UDP is tricky because it's hard to make sure that the client address is preserved, and most UDP protocols assume that they have the correctly client IP (they don't have mechanisms like x-Forwarded-For in HTTP).

youngnick:
Overall changes look fine, but we'll need to make sure we write down rules for what happens when you have multiple routes that do and don't conflict selecting the same Service as a parent (since Service doesn't have a `spec.listener[].protocol` field to give you a hint).

site-src/geps/gep-1426.md
howardjohn: I agree but its also orthogonal to using UDPRoute IMO. If a mesh supports UDP, they ought to be able to support UDPRoute as well. Maybe we call it experimental since there are no implementations yet, but the whole thing is experimental for now anyways?

keithmattix (commented):
Agree with @youngnick's comment above; we need to figure out how to deal with xRoutes of different types referencing the same service. To add more complexity to it, port numbers need not be unique (only names do); it's fairly common for a service to listen to two protocols on the same port (e.g. DNS). At minimum, we need some notes here about the semantics of this scenario

site-src/geps/gep-1426.md
mikemorris: Would narrowing the scope to just HTTPRoute, GRPCRoute and TCPRoute lets us at least start making this more general and get this merged, and punt on TLSRoute and UDPRoute until later?

youngnick: That makes sense to me, yeah.

howardjohn:
I updated the GEP to specify that in the face of conflicting route types, the 'most specific' wins and other protocols are rejcted.

keithmattix (approved):

site-src/geps/gep-1426.md
```diff
@@ -114,37 +113,54 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
+
+If multiple routes with of different type both bind to the same Service and Port pair, only a single route type should be applied. The rejected routes should be ignored and have the `RouteConditionAccepted` status set to the (new) reason `RouteReasonConflicted`.
+
+Route type specificity is defined in the following order (first one wins):
+
+1. HTTPRoute
+2. TLSRoute
+3. TCPRoute
```
mikemorris, line 159:
```suggestion
1. GRPCRoute
2. HTTPRoute
3. TLSRoute
4. TCPRoute
```
I think gRPC would layer above HTTP rather than being treated entirely separately like UDP?

site-src/geps/gep-1426.md
```diff
@@ -114,37 +113,54 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
+The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `Route` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `Route` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
+
+### Route types
+
+All types currently defined in the gateway-api core (`HTTP`, `GRPC`, `TCP`, `TLS`, and `UDP`) are available for use in a Mesh implementation.
+
+If multiple routes with of different type both bind to the same Service and Port pair, only a single route type should be applied. The rejected routes should be ignored and have the `RouteConditionAccepted` status set to the (new) reason `RouteReasonConflicted`.
```
mikemorris, line 153:
```suggestion
If multiple routes with different types both bind to the same Service and Port pair, only a single route type should be applied. The rejected routes should be ignored and have the `RouteConditionAccepted` status set to the (new) reason `RouteReasonConflicted`.
```

mikemorris (approved):
Nit on whether this should use `Route` (code style may be awkward given it's no longer a specific type) vs "xRoute" vs "route", and there's handful of `an` -> `a` preposition tweaks needed from the renaming, but overall LGTM with the addition of explicit conflict resolution logic.

site-src/geps/gep-1426.md
mikemorris: Refs https://gateway-api.sigs.k8s.io/geps/gep-1016/#cross-serving re potential for merging with HTTPRoute as discussed in 12/13/2022 GAMMA meeting.

howardjohn:
I addressed the comments + resolved @shaneutt

site-src/geps/gep-1426.md
```diff
@@ -114,37 +113,55 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
-      * When an existing `HTTPRoute` targeting the `Service` as a `parentRef` has already been accepted, any `HTTPRoute` with a newer creation timestamp should be rejected, because the match target considered for the specificity rules (the `Service`) is identical.
-      * If creation timestamps are identical (from applying both routes simultaneously), only the `HTTPRoute` with a `name` appearing first in alphabetical order should be accepted.
-      * Any rejected `HTTPRoute` should set an `Accepted` condition with `status: "False"`.
+      * When an existing `xRoute` targeting the `Service` as a `parentRef` has already been accepted, any `xRoute` with a newer creation timestamp should be rejected, because the match target considered for the specificity rules (the `Service`) is identical.
+      * If creation timestamps are identical (from applying both routes simultaneously), only the `xRoute` with a `name` appearing first in alphabetical order should be accepted.
+      * Any rejected `xRoute` should set an `Accepted` condition with `status: "False"`.
```
youngnick, line 136: Nit: this needs to be a "must".
```suggestion
      * Any rejected `xRoute` must set an `Accepted` condition with `status: "False"`.
```

youngnick (commented):
One small comment, but LGTM with the usual "This is not final" caveats.

shaneutt (approved):
This seems reasonable as a provisional GEP that we're going to experiment with to provide feedback back into the GEP before implementable :+1:

youngnick:
🤦 
/approve

howardjohn:
This merged but github is confused

Issue #1480, opened 2022-10-26
Allow GAMMA HTTPRoutes to direct traffic to backendRefs in a different namespace


mikemorris:
**What would you like to be added**:

Define expectations or requirements for an `HTTPRoute` with a `Service` `parentRef` to direct traffic to `backendRefs` in other namespaces.

**Why this is needed**:

For North/South implementations, Gateway API requires a `ReferenceGrant` to be present to allow traffic from a Gateway to reach a `backendRef` in a different namespace, but service mesh implementations typically rely on mesh-implemented AuthZ policy to determine whether traffic flows like this should be allowed. We need to define how this use case should be governed to allow us to safely remove this restriction from `HTTPRoute` mesh routing.

Issue #1479, opened 2022-10-26
Remove "one HTTPRoute per Service" GAMMA limitation if safe


mikemorris:
**What would you like to be added**:

Allow more than one `HTTPRoute` to target the same `Service` as a `parentRef`, removing the existing limitation in [GEP-1426: HTTPRoute Mesh Binding](https://github.com/kubernetes-sigs/gateway-api/pull/1426).

To remove this restriction, we would need to confirm that this can be allowed safely (wouldn't be susceptible to circular references or confusing conflict resolution), specifying expected behavior or more granular constraints if the existing [route rule merging guidelines](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule) aren't sufficient.

**Why this is needed**:

There is service mesh [user interest](https://github.com/istio/istio/issues/22997) in merging separately-defined routes together to manage large configurations, multiple versions or canary deployments for a single mesh service.

PR #1493, opened 2022-10-28
Mesh: define namespace boundaries for parent and backend

howardjohn:
**What type of PR is this?**

/kind gep
This also clarifies parentRef namespace boundries which were under-specified.

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1480

```release-note
NONE
```

### DIFF
--- a/site-src/geps/gep-1426.md
+++ b/site-src/geps/gep-1426.md
@@ -115,40 +115,6 @@ An alternate pattern additionally supported by this approach would be to target 
 
 `ServiceImport` would remain a valid backend option for `xRoute` resources (but _not_ currently a requirement for core conformance), and could be specified alongside a `Service` `backendRef` to split traffic across clusters within a `ClusterSet` (as defined in the [Multi-cluster Service (MCS) APIs project](https://github.com/kubernetes-sigs/mcs-api)). This could be a way to solve the need described in [A use case for using Gateway APIs in Multi-Cluster](https://docs.google.com/document/d/1bjr0uAVMmEtTX4mpU_aGXXapQFsEz_Qt0OnDoVswEEI/edit).
 
-### Ports
-
-By default, a `Service` attachment applies to all ports in the service. Users may want to attach routes to only a *specific* port in a Service. To do so, the `parentRef.port` field should be used.
-
-If `port` is set, the implementation MUST associate the route only with that port.
-If `port` is not set, the implementation MUST associate the route with all ports defined in the Service.
-
-### `hostnames` field
-
-GAMMA implementations SHOULD NOT infer any functionality from the `hostnames` field on `xRoute`s (currently, `TLSRoute`, `HTTPRoute`, and `GRPCRoute` have this field) due to current under-specification and reserved potential for future usage or API changes.
-
-For the use case of filtering incoming traffic from selected HTTP hostnames, it is recommended to guide users toward configuring [`HTTPHeaderMatch`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderMatch) rules for the [`Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) header. Functionality to be explored in future GEPs may include supporting concurrent usage of an `xRoute` traffic configuration for multiple North/South `Gateways` and East/West mesh use cases or redirection of egress traffic to an in-cluster `Service`.
-
-### Limiting graph depth, preventing conflicts and cycles
-
-A `Service` may only be specified as a `parentRef` of an `xRoute` if it is:
-
-* NOT already specified as a `parentRef` of another `xRoute` [in the same namespace]
-    * Resolving conflicts from multiple `HTTPRoutes` attempting to target the same `Service` as a `parentRef` should follow the existing [conflict guidelines](https://gateway-api.sigs.k8s.io/concepts/guidelines/#conflicts).
-      * When an existing `xRoute` targeting the `Service` as a `parentRef` has already been accepted, any `xRoute` with a newer creation timestamp should be rejected, because the match target considered for the specificity rules (the `Service`) is identical.
-      * If creation timestamps are identical (from applying both routes simultaneously), only the `xRoute` with a `name` appearing first in alphabetical order should be accepted.
-      * Any rejected `xRoute` must set an `Accepted` condition with `status: "False"`.
-    * This requirement may be relaxed later to follow the existing [precedence rules for merging configuration](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule), but is intended to simplify initial implementation.
-    * The namespace colocation requirement effectively forecloses using this strategy to delegate route management, but there is service mesh [user interest](https://github.com/istio/istio/issues/22997) in merging separately-defined routes together to manage large configurations, multiple versions or canary deployments for a single mesh service.
-* NOT already specified as a `backendRef` of any _other_ `xRoute` with a `Service` `parentRef`
-   * This restriction is intended to prevent route "inclusion/delegation" patterns or circular references from multiple tiers of routing rules _within mesh configuration_.
-   * A `Service` MAY still be listed as a `backendRef` on a different `xRoute` which only targets a `Gateway` as a `parentRef`. The specific behavior of how Gateways might interact with mesh routing rules will be defined in a future GEP.
-    * A `Service` MAY still be listed as a `backendRef` on the _same_ `xRoute` where it is specified as a `parentRef`, e.g. to only split off some subset of traffic from the default backend endpoints.
-    * This could be computationally expensive to validate.
-
-This should be enforced by controller validation, which should set an `Accepted: { status: False }` condition with a corresponding new `ParentRefConflict` `RouteConditionReason`.
-
-The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `xRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `xRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
-
 ### Route types
 
 All types currently defined in the gateway-api core (`HTTP`, `GRPC`, `TCP`, `TLS`, and `UDP`) are available for use in a Mesh implementation.
@@ -166,6 +132,19 @@ Because UDP is its own protocol, it is orthogonal to these precedence order. Sin
 
 Note: these conflicts only occur when multiple *different* route types apply to the same Service+Port pair. Multiple routes of the same type are valid, and merge according to the route-specific merging semantics.
 
+### Ports
+
+By default, a `Service` attachment applies to all ports in the service. Users may want to attach routes to only a *specific* port in a Service. To do so, the `parentRef.port` field should be used.
+
+If `port` is set, the implementation MUST associate the route only with that port.
+If `port` is not set, the implementation MUST associate the route with all ports defined in the Service.
+
+### `hostnames` field
+
+GAMMA implementations SHOULD NOT infer any functionality from the `hostnames` field on `xRoute`s (currently, `TLSRoute`, `HTTPRoute`, and `GRPCRoute` have this field) due to current under-specification and reserved potential for future usage or API changes.
+
+For the use case of filtering incoming traffic from selected HTTP hostnames, it is recommended to guide users toward configuring [`HTTPHeaderMatch`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderMatch) rules for the [`Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host) header. Functionality to be explored in future GEPs may include supporting concurrent usage of an `xRoute` traffic configuration for multiple North/South `Gateways` and East/West mesh use cases or redirection of egress traffic to an in-cluster `Service`.
+
 ### Namespace boundaries
 
 In a mesh, routes can be configured by two personas:
@@ -173,7 +152,7 @@ In a mesh, routes can be configured by two personas:
 * Service producers, who want to modify behavior of *inbound* requests to their services
 * Service consumers, who want to modify behavior of *outbound* requests to other services.
 
-While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these rolls and behaves differently depending on the role.
+While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these roles and behaves differently depending on the role.
 
 A route is a producer route when the `parentRef` refers to a service in the *same namespace*. This route SHOULD apply to all incoming requests to the service, including from clients in other namespaces.
 
@@ -181,7 +160,26 @@ Note: Some implementations may only be able to apply routes on client-side proxi
 
 A route is a consumer route when the `parentRef` refers to a service in *another namespace*. Unlike producer routes, consumer routes are scoped only the same namespace. This ensures that for traffic between two namespaces, another unrelated namespace cannot modify their traffic.
 
-Routes of either type can send traffic to `backendRefs` in any namespace. Unlike `Gateway` bound routes, this is allowed without a `ReferenceGrant`. In `Gateway`, routes are opt-in; by default, no Services are exposed (often to the public internet), and a service producer must explicitly opt-in by creating a route themselves, or allowing another namespace to via `ReferenceGrant`. For mesh, routes *augment* existing Services, rather than exposing them to a broader scope. As a result, a `ReferenceGrant` is not required. Access control, if desired, is handled by authorization policies.
+Routes of either type can send traffic to `backendRefs` in any namespace. Unlike `Gateway` bound routes, this is allowed without a `ReferenceGrant`. In `Gateway`-bound routes (North-South), routes are opt-in; by default, no Services are exposed (often to the public internet), and a service producer must explicitly opt-in by creating a route themselves, or allowing another namespace to via `ReferenceGrant`. For mesh, routes *augment* existing Services, rather than exposing them to a broader scope. As a result, a `ReferenceGrant` is not required in most mesh implementations. Access control, if desired, is handled by other mechanism such as `NetworkPolicy`. While uncommon, if a mesh implementation *does* expose the ability to access a broader scope than would otherwise be reachable, then `ReferenceGrant` must be used for cross namespaces references.
+
+### Multiple routes for a Service
+
+A service may be used as a `parentRef` (where we attach to the "Service Frontend") or as a `backendRef` (where we attach to the "Service Backend").
+
+In general, when a request is sent to a Service frontend (ex: `curl svc`), it should utilize a Route bound to that Service.
+However, when sent to a Service backend (ex: `curl pod-ip`), it would not.
+
+Similarly, if we have multiple "levels" of Routes defined, only the first will be used, as that is the only one that accesses the Service frontend.
+
+Consider a cluster with routes for a Service in both a Gateway, consumer namespace, and producer namespace:
+
+* Requests from the Gateway will utilize the (possibly merged) set of routes attached to the Gateway
+* Requests from a namespace with consumer routes will utilize the (possibly merged) set of routes in the consumer namespace
+* Requests from other namespaces will utilize the (possibly merged) set of routes in the producer namespace
+
+The merging of routes occurs only within groups of the same type of routes (Gateway bound, producer, or consumer), and follows the standard route merging behavior already defined.
+
+Note: a possible future extension is to allow `backendRefs` to explicitly target a "frontend" or "backend". This could allow chaining multiple routes together. However, this is out of scope for the current GEP.
 
 ### Drawbacks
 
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
 
 The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
 
+### Namespace boundaries
+
+In a mesh, routes can be configured by two personas:
+
+* Service producers, who want to modify behavior of *inbound* requests to their services
```
rsalmond, line 148: Suggestion: unless this term is explicitly defined somewhere, “providers” might be clearer than “producers”. Or maybe “request receivers” and “request makers”? Not sure if that’s much better but something about “service producers” feels a little off in this context.

howardjohn: Maybe I can add a link

mikemorris: The service producer/consumer roles are described in https://gateway-api.sigs.k8s.io/geps/gep-1324/#producer-and-consumer, I can open a separate PR linking that in the Personas section of this GEP.

site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
...
+* Service consumers, who want to modify behavior of *outbound* requests to other services.
+
+While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these rolls and behaves differently depending on the role.
+
+A route is a producer route when the `parentRef` refers to a service in the *same namespace*. This route SHOULD apply to all incoming requests to the service, including from clients in other namespaces.
+
+Note: Some implementations may only be able to apply routes on client-side proxies. As a result, these will likely only apply to requests from clients who are also in the mesh.
+
+A route is a consumer route when the `parentRef` refers to a service in *another namespace*. Unlike producer routes, consumer routes are scoped only the same namespace. This ensures that for traffic between two namespaces, another unrelated namespace cannot modify their traffic.
```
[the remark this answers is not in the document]
howardjohn: I changed the title; I think they are interconnected enough its a bit hard to split out.

site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
...
 The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
 
+### Namespace boundaries
+
+In a mesh, routes can be configured by two personas:
+
+* Service producers, who want to modify behavior of *inbound* requests to their services
+* Service consumers, who want to modify behavior of *outbound* requests to other services.
+
+While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these rolls and behaves differently depending on the role.
```
robscott, line 151:
```suggestion
While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these roles and behaves differently depending on the role.
```

site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
...
+
+A route is a producer route when the `parentRef` refers to a service in the *same namespace*. This route SHOULD apply to all incoming requests to the service, including from clients in other namespaces.
+
+Note: Some implementations may only be able to apply routes on client-side proxies. As a result, these will likely only apply to requests from clients who are also in the mesh.
```
robscott, line 155: How can we represent this when it comes to conformance? Do we need a new term for this?

site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
...
+### Namespace boundaries
+
+In a mesh, routes can be configured by two personas:
+
+* Service producers, who want to modify behavior of *inbound* requests to their services
+* Service consumers, who want to modify behavior of *outbound* requests to other services.
+
+While these concepts are not directly exposed in the API, a route is implicitly fulfilling one of these rolls and behaves differently depending on the role.
+
+A route is a producer route when the `parentRef` refers to a service in the *same namespace*. This route SHOULD apply to all incoming requests to the service, including from clients in other namespaces.
```
robscott, line 153: What happens when both a consumer and producer route apply to the same Service?

site-src/geps/gep-1426.md
```diff
@@ -146,6 +141,23 @@ This should be enforced by controller validation, which should set an `Accepted:
...
+
+Note: Some implementations may only be able to apply routes on client-side proxies. As a result, these will likely only apply to requests from clients who are also in the mesh.
+
+A route is a consumer route when the `parentRef` refers to a service in *another namespace*. Unlike producer routes, consumer routes are scoped only the same namespace. This ensures that for traffic between two namespaces, another unrelated namespace cannot modify their traffic.
+
+Routes of either type can send traffic to `backendRefs` in any namespace. Unlike `Gateway` bound routes, this is allowed without a `ReferenceGrant`. In `Gateway`, routes are opt-in; by default, no Services are exposed (often to the public internet), and a service producer must explicitly opt-in by creating a route themselves, or allowing another namespace to via `ReferenceGrant`. For mesh, routes *augment* existing Services, rather than exposing them to a broader scope. As a result, a `ReferenceGrant` is not required. Access control, if desired, is handled by authorization policies.
```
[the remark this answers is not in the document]
robscott: This paragraph is confusing to me. I think the "In Gateway" sentence is intended to refer to north-south traffic coming from a Gateway, but the current wording is a bit confusing to me. 

Is it always true that a mesh implementation is not going to be subject to a confused deputy attack? It seems like at least some mesh implementations may be able to route traffic to locations that would otherwise be inaccessible, and thus ReferenceGrant would still be an important mechanism.

When you reference "authorization policies", I think it's important to state that those are not currently defined by Gateway API and are out of scope for this GEP.

site-src/geps/gep-1426.md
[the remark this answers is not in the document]
youngnick: I get what these consumer routes is trying to achieve, but I think that there's one big thing that needs further explanation (which Rob raised above, but I think is most relevant to this bit):

What happens when a consumer route's backendRef to another namespace targets a producer Service in that namespace that is itself a parentRef to a Route?

To put this another way, does consumer-side traffic management override producer-side traffic-management? That seems like a very fraught idea, as it would make it so that consumer routes could end up preferring old versions etc, without the producer knowing about it.

If it doesn't, is it that a producer Service is not a valid target for the backendRef of a consumer Service? If so, how do you know that it's a consumer Service? There's nothing on the producer Service to indicate it, and it seems likely that the case would arrive that a consumer rule targets a Service that _later_ becomes the parentRef of a Route, which would mean that it would become invalid _when the owner of a producer service made a change_, which also seems... not ideal.

I think it's worth considering if maybe consumer behavior modification is worth putting into a metaresource or Policy-style object instead of re-using the Service for yet another use case.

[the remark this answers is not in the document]
howardjohn: In my head this diagram made a lot more sense... but here it is. The curvy lines are the logical flow of requests.

The idea is when you attach a route to a Service (frontend), you are _augmenting_ that frontend. When requests are sent to it, instead, the route rules are applied. In the example above, we hit the frontend from consumer land, so we use the consumer route in place of the service frontend. Then the route sends to the service _backend_, which doesn't have a route associated (since we don't have per-backend routes). So in effect, the consumer route takes precedence.

While its not ideal, I think its the least bad option. See alternatives below.

One thing to note, as well, is this is assuming we are sending to the service _backend_. In theory, we could have `backendRef: ServiceFrontend` and then have it go through both routes. However, we don't have a way to express this in the API and sidecars cannot implement it.

Alternatives:
* Do not have consumer routes: this is a fairly important feature for mesh users, and we still will need to solve it for Gateways at least (which are the ~same problem)
* Merge routes across the two using precedence rules: doesn't really solve any use cases; the routes become highly coupled and small changes to one may result in drastic changes in behavior
* Apply any matching consumer rules, then any matching producer rules. This actually could be viable, but doesn't make sense at all for gateways
* Apply *both* routes. Most implementations cannot do this

site-src/geps/gep-1426.md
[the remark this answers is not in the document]
howardjohn: In Gateway the guidance is basically: "You need Grant unless you have some exceptional implementation that handles it otherwise". I'll change the wording here to "You *don't* need Grant, unless you have some exceptional implementation that *doesn't* handle it otherwise".

Essentially the same but most/all meshes will not needed it, so its just flipping the default

site-src/geps/gep-1426.md
howardjohn: Added a (lot) of discussion about this. A bit worried I am creeping the scope of this PR, but its all connected so has to be done

site-src/geps/gep-1426.md
howardjohn: I would imagine conformance tests just are all about everything within the mesh. Otherwise, we have to define what is the behavior of out-of-mesh clients which is not something we should standardize IMO

site-src/geps/gep-1426.md
howardjohn: Compositing is still difficult with 1 impl, as doing the merge is quite complex. I take it that is how its done in consul today, right?

One thing I based this on is that Gateway route+producer route should behave like consumer route+producer route. So if we are doing merging for consumer routes, we ought to do it for Gateway, which brings its own can of worms.

I wrote up a bit on this a long time ago, might be helpful to guide discussion:
[Public Gateway API Mesh Route Merging (1).pdf](https://github.com/kubernetes-sigs/gateway-api/files/9923961/Public.Gateway.API.Mesh.Route.Merging.1.pdf)

site-src/geps/gep-1426.md
```diff
@@ -125,26 +120,41 @@ An alternate pattern additionally supported by this approach would be to target
...
 
-The logic described here is a bit verbose, but could be validated with conformance tests, and is intended to allow an `HTTPRoute` configuring mesh traffic rules to additionally specify a `Gateway` `parentRef` (or `HTTPRoute` `parentRef` if [GEP-1058: Route Inclusion and Delegation](https://github.com/kubernetes-sigs/gateway-api/pull/1085) moves forward) to apply the same routing rules when exposing a mesh service outside the cluster.
+A route is a producer route when the `parentRef` refers to a service in the *same namespace*. This route SHOULD apply to all incoming requests to the service, including from clients in other namespaces.
+
+Note: Some implementations may only be able to apply routes on client-side proxies. As a result, these will likely only apply to requests from clients who are also in the mesh.
+
+A route is a consumer route when the `parentRef` refers to a service in *another namespace*. Unlike producer routes, consumer routes are scoped only the same namespace. This ensures that for traffic between two namespaces, another unrelated namespace cannot modify their traffic.
```
michaelbeaumont, line 136: Some thoughts on this from a "how Kuma does routes right now" perspective, sorry for the wall. Some of this probably belonged on the original GEP PR but it's still relevant to this PR especially:

One way to think about an `HTTPRoute` is by being explicit about its effects on a request based on the source and the destination of that request.

When an `HTTPRoute` attaches to a `Gateway` via `parentRef`, this attachment expresses both the source as well as the destination of a request. The source being "somewhere, I don't know this is the edge" and the destination being the listener or the listener's `hostname`. And there's the option of further restricting this in the HTTPRoute `hostname`.

`Gateway` also gives us `.allowedRoutes.namespaces` separately to answer the question "who can attach `HTTPRoutes` and affect requests that come to my listener?".

With this GEP as is, using a `Service` as `parentRef` only tells us something about the destination of the requests we want to alter.

In the mesh case, it's important to be able to talk about the _source_ of a request when doing traffic routing. That's what the idea of consumer/producer routes in this PR expresses. A consumer route is "only requests from this namespace" and a producer route is "requests from anywhere in the mesh". 

The PR gives the `Namespace` of an `HTTPRoute` the ability to scope a route based on source of a request. But doing this also answers the question of who can attach to a `Service` as a consumer/producer.

Is there a way to answer these two questions less implicitly/separately?

I introduce the following as an illustration of what I vaguely have in mind, not a concrete suggestion. The `Mesh` resource has already been discussed and shows up in some of the GEP alternatives. Some of the GEP alternatives may even address the following but I couldn't find this source & destination aspect mentioned explicitly in the GEP.

Let's just imagine a different `parentRefs` and specify exactly two object references. The first to scope the route based on the request's _source_ and the other on the request's _destination_. Theoretically this is compatible with `parentRefs` as is, using a tuple as a list. I'll illustrate with a producer and a consumer route:

```
parentRefs:
  from:
    kind: Mesh
  to:
    kind: Service
    name: my-service
    namespace: my-namespace
```

This is a producer route which would affect requests _from_ anywhere in the mesh _to_ the `my-service` `Service`.

Translated to this PR IIUC:

```
metadata:
  namespace: my-service
parentRefs:
  - kind: Service
    name: my-service
    namespace: my-namespace
```

And then a consumer route:

```
parentRefs:
  from:
    kind: Namespace
    name: other-namespace
  to:
    kind: Service
    name: my-service
    namespace: my-namespace
```

that would affect requests _from_ services in `other-namespace` _to_ the `my-service` `Service`.

Translated to this PR:

```
metadata:
  namespace: other-namespace
parentRefs:
  - kind: Service
    name: my-service
    namespace: my-namespace
```

Coming from Kuma at least, the "object refs as source and destination" style feels more natural when thinking about which requests are affected by a route. "Which namespace is my route in plus an object ref" feels a little implicit/subtle.

If we were designing mesh routing as a Policy Attachment policy, we might have come up with something like

```
spec:
  targetRef:
    kind: Mesh
  to:
    kind: Service
  ...
```

Explicit refs would also give the option to specify something like:

```
parentRefs:
  from:
    kind: Service
    name: other-service
    namespace: other-namespace
  to:
    kind: Service
    name: my-service
    namespace: my-namespace
```

to scope a route to apply to requests from and to single services.

If the producer route only ever needs to be created in the same `Namespace` as the `Service`, a consumer route _never_ needs to be created in the same `Namespace` as the `Service` and `Namespace` is the only answer to "who's allowed to be a producer and who a consumer", then this degenerates to what the PR proposes.

It's different and I don't mean to open old discussion but it could be a useful perspective on how to deal with the consumer vs producer dichotomy.

howardjohn: I do generally agree with ~everything you said. Some initial reactions:

* IMO producer routes ought to always be in the same namespace as the service. Otherwise they are not really the "producer"?
* Wanting "same namespace consumer" or "more specific than namespace" is definitely valid though
* `From=specific Service` is suspicious, since requests come from workloads not Services. But can be made to work or slightly tweaked of course

In general I would love to have a reasonable way to be more specific, I am just not sure the best schema to represent that

michaelbeaumont: * Yeah I think I mostly agree about producer routes. Though if we consider that producer routes come from the producer role, it makes me wonder if there's a use case for GEP-1324 mesh administrator routes, which might reside in a different "administrator namespace". The only cases I can think of are especially, but also only, useful with `backendRefs` omitted. And probably also need more generality in the `to` sense than a `Service` `parentRef`.

  Looking at GEP-1426 I see under goals:

  > SHOULD allow control by the cluster operator (mesh administrator) to grant permission for whether HTTPRoute resources in a given namespace are allowed to configure mesh traffic routing.

  This doesn't feel covered yet, right?

* `Service` in `from` is a little suspicious, you're right. I guess it'd make sense only if a workload is pointed to by exactly one service backend `Service`? But yeah, like you said, `Service` in particular is probably not that important, it's about having some way to refer to "a workload" or "something more specific than namespace."

If tweaking `HTTPRoute` isn't in the cards, which is totally valid given it's primary use case not being mesh, I guess the too late for this PR, concrete suggestion would be a vote to reconsider/also support a new resource as `parentRef`. One that holds the `from` and `to` references. I see both `ServiceBinding` and `ServiceProjection` under GEP-1426 but I'm not totally sure either of those covers it.

mateiidavid: @michaelbeaumont thanks for the concise write-up! I wanted to comment here since the conversation in this thread has been slightly on-point with some of the concerns we've had in the Linkerd team.

On the whole, we're pretty much in agreement with the GEP, we would like to discuss something similar to the `to` and `from` idea you wrote down though. On the Linkerd side, it will be necessary in the future to limit this configuration to only a subset of the workloads in the consumer namespace -- we already know that there are situations where users of Linkerd will need to enforce different policies for consumers in the same namespace.

We were thinking this could be addressed for a resource that captures the Consumer-to-Service binding relationship. Something between `ServiceBinding` and `ServiceProjection` (although projection probably fits better as a model and name). For example:

```
spec:
  consumerRefs:
    - group: apps
      kind: Deployment
      name: web
  providerRef:
    kind: Service
    namespace: backend
    name: emoji
status:
  conditions: ...
```

The resource describes traffic originating from a consumer workload (`deploy/web`) to a producer service (`service/emoji`). The status is added in for conflict resolution. Wdyt about an approach similar to this?

cc: @kflynn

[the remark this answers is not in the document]
kflynn: That's how I read this, @keithmattix, FTR.

site-src/geps/gep-1426.md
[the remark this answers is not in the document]
kflynn: I think that having the consumer route take precedence over the producer route is fine for now – I think it's a great way to get to something implementable, and I think that's the really critical thing at the moment.

On the other hand, I suspect that the moment we try to use this bit in anger, we'll find that we really, really want some variety of merge behavior. 🙂 It feels like there are a lot of use cases where what the consumer really wants is just to slightly restrict or modify something, not to replace everything wholesale. An example from @howardjohn's PDF jumps out here: perhaps the only thing the consumer needs to do is to rewrite a no-longer-supported HTTP path into its modern equivalent, to let an old client workload keep functioning until it can be replaced. It's easy to imagine a world where the producer route is complex and perhaps changing, and the consumer route really, really doesn't want that burden.

(It's also easy to imagine a world in which, say, the producer specifies a 5-second timeout, and the consumer really wants a 1-second timeout because the client workload has special requirements. In my head, though, that's policy rather than routing, and policy is its own thing entirely.)

So: I agree with @howardjohn that consumer-route-wins is far from ideal, but I'd vote to land this as-is and get to implementing so that we can learn more about what we really need.

site-src/geps/gep-1426.md
kflynn: I think the thread starting with https://github.com/kubernetes-sigs/gateway-api/pull/1493/files#r1009709390 more completely covers this, so I commented there.

keithmattix (approved):
I'm approving this PR as-is for a couple of reasons:

1. It seems like we're pretty strongly in favor of having consumer routes take precedence over producer routes _for now_ and then we can introduce some kind of merging behavior later if there's need for it
2. These rules feel sufficient enough to move an experimental GEP to implementable which is our goal right now

site-src/geps/gep-1426.md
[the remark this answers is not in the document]
howardjohn: Seems like consensus is to have "consumer wins" for short term, at least.

robscott:
Based on discussion in today's meeting and another read through the GEP, I think this is good to go. Going to approve, but will defer to someone else (cc @youngnick @shaneutt @mikemorris for example) to add final LGTM.

/approve

site-src/geps/gep-1426.md
howardjohn: I think this is a great discussion and I don't want to bury it, so I opened an issue in https://github.com/kubernetes-sigs/gateway-api/issues/1640 to track this further. Based on the meeting it sounds like the desire is to move forward with this as-is to check something in as experimental, but we should follow up on exploring this further.

linsun, line 136: Does this mean users can't do consumer route if consumer and producer are in the same namespace? e.g. user would have to put the consumer in a different ns.

howardjohn: Yes

howardjohn: But the consumer is the producer... so its kind of moot. Whether it is a "consumer" or "producer" route is not really relevant, it will end up the same behavior. Unless you want 2 clientsin the same namespace to have different routes, but this is not supported in general (route granularity is namespace)

shaneutt (approved):
I think this reads well, and given that it's still provisional I don't have any reason to block this iteration. There's a few comments which I think we should just double check and resolve before we merge.

keithmattix:
Fixes #1479

shaneutt (approved):
/lgtm

site-src/geps/gep-1426.md
```diff
@@ -37,10 +37,6 @@ This GEP uses the [roles and personas](https://gateway-api.sigs.k8s.io/concepts/
 * MUST set a status field on `xRoute` to show if the routing configuration has been applied to the mesh.
 * MUST only be allowed to configure "producer" traffic rules for a `Service` in the same namespace as the `xRoute`.
```
mikemorris, line 38:
```suggestion
* MUST be allowed to configure "producer" traffic rules for a `Service` in the same namespace as the `xRoute`.
```

mikemorris (approved):
Apologies for taking so long to get through this! I'm happy with where all the discussions have ended up, and this feels like a good step forward to unblock important use cases while minimizing complexity for initial implementations.

/approve
/lgtm

PR #1497, opened 2022-10-31
Add and test single filter action per header restriction

### LINKED ISSUE #480: HTTPRequestHeaderFilter should specify the order of operations
**What happened**:

HTTPRequestHeaderFilter has add/set/remove operations that manipulate the set of request headers. Depending on the order that a proxy applies these operations, the final result can be different.

Header manipulations should be consistent across proxies.

rainest:
**What type of PR is this?**
/kind documentation

Documents restrictions on header filters and how implementations should handle invalid filters. Header filters cannot modify the same header twice. When users require multiple values for the same header, they must specify the comma-separated values in configuration.

~Adds a conformance test confirming that invalid headers result in a false `Accepted` Condition with reason `RouteReasonUnsupportedValue`.~

Adds a webhook rule that rejects header filters that violate the single action per header rule.

Fix #480. Order of operations is a moot point if there is only one action for a given header, and there are no outcomes that actually require multiple actions.

```release-note
HTTPRequestHeaderFilter and HTTPResponseHeaderFilter forbid configuring multiple actions for the same header.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -722,9 +722,10 @@ type HTTPHeader struct {
 
 // HTTPHeaderFilter defines a filter that modifies the headers of an HTTP
 // request or response. Only one action for a given header name is permitted.
-// Filters MUST NOT specify multiple actions of the same or different type for
-// any one header name. Configuration to set or add multiple values for a
-// header MUST use RFC 7230 header value formatting, separating each value with
+// Filters specifying multiple actions of the same or different type for
+// any one header name are invalid and will be rejected by the webhook if
+// installed. Configuration to set or add multiple values for a
+// header must use RFC 7230 header value formatting, separating each value with
 // a comma.
 type HTTPHeaderFilter struct {
 	// Set overwrites the request with the given header (name, value)
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -720,8 +720,12 @@ type HTTPHeader struct {
 	Value string `json:"value"`
 }
 
-// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP request
-// or response.
+// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP
+// request or response. Only one action for a given header name is permitted.
+// Filters MUST NOT specify multiple actions of the same or different type for
+// any one header name. Configuration to set or add multiple values for a
+// header MUST use RFC 7230 header value formatting, separating each value with
+// a comma.
 type HTTPHeaderFilter struct {
 	// Set overwrites the request with the given header (name, value)
 	// before the action.
@@ -756,12 +760,11 @@ type HTTPHeaderFilter struct {
 	// Config:
 	//   add:
 	//   - name: "my-header"
-	//     value: "bar"
+	//     value: "bar,baz"
 	//
 	// Output:
 	//   GET /foo HTTP/1.1
-	//   my-header: foo
-	//   my-header: bar
+	//   my-header: foo,bar,baz
 	//
 	// +optional
 	// +listType=map
--- a/apis/v1beta1/validation/httproute_test.go
+++ b/apis/v1beta1/validation/httproute_test.go
@@ -475,6 +475,82 @@ func TestValidateHTTPRoute(t *testing.T) {
 							Name:  gatewayv1b1.HTTPHeaderName("x-grain"),
 							Value: "wheat",
 						},
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("x-spice"),
+							Value: "coriander",
+						},
+					},
+				},
+			}},
+		}},
+	}, {
+		name:     "multiple actions for the same request header with inconsistent case (invalid)",
+		errCount: 1,
+		rules: []gatewayv1b1.HTTPRouteRule{{
+			Filters: []gatewayv1b1.HTTPRouteFilter{{
+				Type: gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
+				RequestHeaderModifier: &gatewayv1b1.HTTPHeaderFilter{
+					Add: []gatewayv1b1.HTTPHeader{
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("x-fruit"),
+							Value: "apple",
+						},
+					},
+					Set: []gatewayv1b1.HTTPHeader{
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("X-Fruit"),
+							Value: "watermelon",
+						},
+					},
+				},
+			}},
+		}},
+	}, {
+		name:     "multiple of the same action for the same request header (invalid)",
+		errCount: 1,
+		rules: []gatewayv1b1.HTTPRouteRule{{
+			Filters: []gatewayv1b1.HTTPRouteFilter{{
+				Type: gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
+				RequestHeaderModifier: &gatewayv1b1.HTTPHeaderFilter{
+					Add: []gatewayv1b1.HTTPHeader{
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("x-fruit"),
+							Value: "apple",
+						},
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("x-fruit"),
+							Value: "plum",
+						},
+					},
+				},
+			}},
+		}},
+	}, {
+		name:     "multiple actions for different request headers",
+		errCount: 0,
+		rules: []gatewayv1b1.HTTPRouteRule{{
+			Filters: []gatewayv1b1.HTTPRouteFilter{{
+				Type: gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
+				RequestHeaderModifier: &gatewayv1b1.HTTPHeaderFilter{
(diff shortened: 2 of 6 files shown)


rainest:
https://github.com/Kong/kubernetes-ingress-controller/pull/3119 demonstrates a draft implementation in Kong's implementation. Not quite sure what the expected chicken-before-egg order is for demonstrating changes that aren't actually in the spec yet.

https://github.com/Kong/kubernetes-ingress-controller/actions/runs/3365550934/jobs/5581123044 is a simplified conformance run that runs this test only.

I originally included a test for multiple of the same action, but it turns out this is already forbidden by the schema. Something like:

```
    - type: RequestHeaderModifier
      requestHeaderModifier:
        set:
        - name: X-Header-Set
          value: one
        - name: X-Header-Set
          value: two
```
is rejected by the API server with
```
The HTTPRoute "invalid-request-header-modifier-same-action" is invalid: spec.rules[0].filters[0].requestHeaderModifier.set[1]: Duplicate value: map[string]interface {}{"name":"X-Header-Set"}
```

apis/v1beta1/httproute_types.go
```diff
@@ -720,8 +720,15 @@ type HTTPHeader struct {
 	Value string `json:"value"`
 }
 
-// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP request
-// or response.
+// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP
+// request or response. Implementations MUST NOT accept more than one action
+// for the same header name (including multiple actions of the same type). If
```
robscott, line 725: This seems like something we should add to webhook validation?

conformance/tests/httproute-request-header-modifier.go
```diff
@@ -192,3 +194,28 @@ var HTTPRouteRequestHeaderModifier = suite.ConformanceTest{
...
+
+var HTTPRouteInvalidRequestHeaderModifier = suite.ConformanceTest{
+	ShortName:   "HTTPRouteInvalidRequestHeaderModifier",
+	Description: "An HTTPRoute with invalid request header modifiers is not accepted",
+	Manifests:   []string{"tests/httproute-invalid-request-header-modifier.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+
+		t.Run("HTTPRoutes that do intersect with listener hostnames", func(t *testing.T) {
```
robscott, line 206: This probably was accidentally left over from another test

conformance/tests/httproute-request-header-modifier.go
```diff
@@ -192,3 +194,28 @@ var HTTPRouteRequestHeaderModifier = suite.ConformanceTest{
...
+				{Namespace: ns, Name: "invalid-request-header-modifier-multiple-actions"},
+			}
+			for _, route := range routes {
+				headerConflictCond := metav1.Condition{
+					Type:   string(v1beta1.RouteConditionAccepted),
+					Status: metav1.ConditionFalse,
+					Reason: string(v1beta1.RouteReasonUnsupportedValue),
+				}
+
+				kubernetes.HTTPRouteMustHaveCondition(t, suite.Client, suite.TimeoutConfig, route, gwNN, headerConflictCond)
```
robscott, line 217: If we do solve this with webhook validation, we either may not need this conformance test, or we should explicitly require that conformance tests are run without the validating webhook present so we can ensure that controllers still behave well in its absence. I tend to think we should just start with webhook validation, but open to other opinions here.

/cc @youngnick

robscott (commented):
Thanks for catching this @rainest!

apis/v1beta1/httproute_types.go
rainest: Yeah, probably better insofar as it's possible and we'd rather reject bad changes up front rather than let them go in and then have to invalidate a route that was previously working.

conformance/tests/httproute-request-header-modifier.go
rainest: Moot, removed test.

conformance/tests/httproute-request-header-modifier.go
rainest: I agree that we probably only want the webhook--we may as well avoid the additional implementation burden. I've updated the docstring to remove references to the condition as such, since these should never reach implementations.

apis/v1beta1/validation/httproute_test.go
```diff
@@ -445,6 +445,58 @@ func TestValidateHTTPRoute(t *testing.T) {
...
+	}, {
+		name:     "multiple actions for the same request header (invalid)",
+		errCount: 2,
+		rules: []gatewayv1b1.HTTPRouteRule{{
+			Filters: []gatewayv1b1.HTTPRouteFilter{{
+				Type: gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
+				RequestHeaderModifier: &gatewayv1b1.HTTPHeaderFilter{
+					Add: []gatewayv1b1.HTTPHeader{
+						{
+							Name:  gatewayv1b1.HTTPHeaderName("x-fruit"),
```
robscott, line 457: I think we're still missing the following test cases:

1. Repeat names within the same list (Add, Set, Remove)
2. Case insensitive matching ("x-fruit" and "X-fruit" should still be invalid)

robscott (commented):
Thanks @rainest!

apis/v1beta1/validation/httproute_test.go
rainest: You would think so for the first one, but it's [apparently a schema violation](https://github.com/kubernetes-sigs/gateway-api/pull/1497#issuecomment-1297827635). I can run it through `Validate()`, but it won't ever occur in practice. Should the tests include configuration they should never receive?

apis/v1beta1/httproute_types.go
```diff
@@ -720,8 +720,12 @@ type HTTPHeader struct {
 	Value string `json:"value"`
 }
 
-// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP request
-// or response.
+// HTTPHeaderFilter defines a filter that modifies the headers of an HTTP
+// request or response. Only one action for a given header name is permitted.
+// Filters MUST NOT specify multiple actions of the same or different type for
+// any one header name. Configuration to set or add multiple values for a
```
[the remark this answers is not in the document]
rainest: ??? the previous comment said to go with the webhook alone, I thought--I wouldn't think we'd remove the conformance test but leave the controller behavior in the spec.

I'd lean towards de facto requiring the webhook (if you don't run it, you acknowledge that you may encounter undefined behavior) and placing checks that can operate on a single resource alone in it as much as possible. Consistent validation across all implementations, less validation code needed in implmentations, and rejecting resources before they're committed (saving you from bad edits invalidating a previously working resource) are all big upsides. Are there significant reasons why users may not want to run it?

If something does have the potential to cause a serious issue (Gateways not starting would qualify) we may want to validate in both. I wouldn't expect this to have much of a blast radius, as it should, at worst, just result an end header set that isn't what you wanted, or that varies across implementations.

I don't like the split UX much either--a single method of indicating a given failure is just less users need to think about, so we should try for that if possible.

apis/v1beta1/validation/httproute_test.go
rainest: Repeats of the same are now in, but note that you shouldn't see these in practice as they [violate the schema](https://github.com/kubernetes-sigs/gateway-api/pull/1497#issuecomment-1297827635)

apis/v1beta1/validation/httproute_test.go
```diff
@@ -445,6 +445,58 @@ func TestValidateHTTPRoute(t *testing.T) {
 				},
 			}},
 		}},
+	}, {
+		name:     "multiple actions for the same request header (invalid)",
```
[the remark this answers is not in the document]
rainest: They were sorta mixed into the request header test, to confirm that the error count did not tally actions that were fine, but there are now additional 0-error tests.

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: Sorry for the confusion here, I was intending to say in that [previous comment](https://github.com/kubernetes-sigs/gateway-api/pull/1497#discussion_r1009983880) that we still need to have some guidance for a controller if the webhook is not present. 

It's probably fine to leave controller guidance out of the spec and rely on https://gateway-api.sigs.k8s.io/concepts/guidelines/#limitations-of-crd-and-webhook-validation. We're just kinda leaving more parts of the spec unspecified though - ie 10 different implementations may take 10 different approaches if the webhook doesn't catch this. This is not something we've solved elsewhere in the API though, and does not need to block this change.

In general it just feels weird to say that a field MUST NOT do something. We generally reserve that kind of language for implementations, not users. So maybe we can just stick with something like the first 2 sentences I suggested?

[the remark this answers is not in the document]
youngnick: I created #1514 to track work on clarifying the expectations around the webhook. I think the first two lines of @robscott's suggestion cover it.

youngnick:
Aside from the minor change @robscott asked for, and the followups about clarifying the webhook's position, this LGTM.

youngnick:
Oops, I meant to approve, so I can leave the final LGTM to someone else.

/approve

robscott:
Thanks @rainest!

/lgtm
/approve

PR #1499, opened 2022-11-01
Add "Programmed" Gateway condition, move "Ready" to extended conformance

### LINKED ISSUE #1454: Add "Programmed" Gateway condition, move "Ready" to extended conformance
- [ ] Add a new `"Programmed"` status condition and reason to:
  - [ ] [ListenerConditionType](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerConditionType)
  - [ ] [ListenerConditionReason](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerConditionReason)
  - [ ] [GatewayConditionType](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayConditionType)
  - [ ] [GatewayConditionReason](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayConditionReason)
- [ ] Update the description for the `"Ready"` condition to note the stricter behavior and that it is now extended conformance on:
  - [ ] [ListenerConditionType](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerConditionType)
  - [ ] [GatewayConditionType](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.GatewayConditionType)
  - [ ] Update GatewayConditionReason `"Pending"` to note it should be used with the `"Ready"` condition when status is `"Unknown"`, refs https://github.com/kubernetes-sigs/gateway-api/pull/1453#discussion_r994885740
- [ ] Update all conformance tests that check the `"Ready"` Gateway or Listener condition to check `"Programmed"` instead.

Adding a `"Ready"` reason to `HTTPRoute` is not in scope for this, and is tracked separately in https://github.com/kubernetes-sigs/gateway-api/issues/1156

**Why this is needed**:

[GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/) clarified that the behavior currently expressed by the `"Ready"` condition should instead be conveyed through a new `"Programmed"` condition, and `"Ready"` should move to extended conformance and become stricter, indicating that the data plane is able to handle requests NOW, rather than "soon".

LCaparelli:
**What type of PR is this?**
/kind feature
/kind api-change

**What this PR does / why we need it**: Make necessary changes to Gateway and Listener conditions in order to implement [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/).

Fixes #1454 

```release-note
Added "Programmed" Gateway and Listener conditions, moved "Ready" to extended conformance
```

### DIFF
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -493,7 +493,7 @@ const (
 )
 
 const (
-	// This condition indicates whether a Listener have generated some
+	// This condition indicates whether a Listener has generated some
 	// configuration that will soon be ready in the underlying data plane.
 	//
 	// It is a positive-polarity summary condition, and so should always be
--- a/apis/v1alpha2/gateway_types.go
+++ b/apis/v1alpha2/gateway_types.go
@@ -196,6 +196,44 @@ type GatewayConditionType = v1beta1.GatewayConditionType
 type GatewayConditionReason = v1beta1.GatewayConditionReason
 
 const (
+	// This condition indicates whether a Gateway have generated some
+	// configuration that will soon be ready in the underlying data plane.
+	//
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource.
+	//
+	// It should be set to Unknown if the controller performs updates to the
+	// status before it has all the information it needs to be able to determine
+	// if the condition is true.
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "Programmed"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Invalid"
+	// * "Pending"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Pending"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	GatewayConditionProgrammed GatewayConditionType = "Programmed"
+
+	// This reason is used with the "Programmed" condition when the condition is
+	// true.
+	GatewayReasonProgrammed GatewayConditionReason = "Programmed"
+
+	// This reason is used with the "Programmed" condition when the Listener is
+	// syntactically or semantically invalid.
+	GatewayReasonInvalid GatewayConditionReason = "Invalid"
+)
+
+const (
 	// This condition is true when the controller managing the Gateway is
 	// syntactically and semantically valid enough to produce some configuration
 	// in the underlying data plane, though it has not yet produced such
@@ -232,8 +270,8 @@ const (
 	// Deprecated: use the "Accepted" condition with reason "Accepted" instead.
 	GatewayReasonScheduled GatewayConditionReason = "Scheduled"
 
-	// This reason is used with the "Accepted" condition when no controller has
-	// reconciled the Gateway.
+	// This reason is used with the "Accepted", "Programmed" and "Ready"
+	// conditions when no controller has reconciled the Gateway.
 	GatewayReasonPending GatewayConditionReason = "Pending"
 
 	// Deprecated: Use "Pending" instead.
@@ -246,11 +284,9 @@ const (
 )
 
 const (
-	// This condition is true when the Gateway is expected to be able
-	// to serve traffic. Note that this does not indicate that the
-	// Gateway configuration is current or even complete (e.g. the
-	// controller may still not have reconciled the latest version,
-	// or some parts of the configuration could be missing).
+	// Ready is an optional Condition that has Extended support. When it's set,
+	// the condition indicates whether the Gateway has been completely configured
+	// and traffic is ready to flow through the data plane immediately.
 	//
 	// If both the "ListenersNotValid" and "ListenersNotReady"
 	// reasons are true, the Gateway controller should prefer the
@@ -456,10 +492,45 @@ const (
 )
 
 const (
-	// This condition indicates whether the Listener has been
-	// configured on the Gateway.
+	// This condition indicates whether a Listener have generated some
+	// configuration that will soon be ready in the underlying data plane.
 	//
-	// Possible reasons for this condition to be true are:
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource.
+	//
+	// It should be set to Unknown if the controller performs updates to the
+	// status before it has all the information it needs to be able to determine
+	// if the condition is true.
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "Programmed"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Invalid"
+	// * "Pending"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Pending"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ListenerConditionProgrammed ListenerConditionType = "Programmed"
+
+	// This reason is used with the "Programmed" condition when the condition is
+	// true.
+	ListenerReasonProgrammed ListenerConditionReason = "Programmed"
+)
+
+const (
+	// Ready is an optional Condition that has Extended support. When it's set,
+	// the condition indicates whether the Listener has been configured on the
+	// Gateway and traffic is ready to flow through the data plane immediately.
+	//
+	// Possible reasons for this condition to be True are:
 	//
 	// * "Ready"
 	//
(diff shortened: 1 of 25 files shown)


apis/v1alpha2/gateway_types.go
```diff
@@ -196,9 +196,48 @@ type GatewayConditionType = v1beta1.GatewayConditionType
 type GatewayConditionReason = v1beta1.GatewayConditionReason
 
 const (
-	// This condition is true when the controller managing the
-	// Gateway has scheduled the Gateway to the underlying network
-	// infrastructure.
+	// This condition indicates whether a Gateway have generated some
```
LCaparelli, line 199: Is there a reason why we don't alias these to the ones defined in `v1beta1` like we did with the types? Should help reducing duplication, wdyt?

shaneutt: Yes, in general unless there's a reason a split needs to exist, add to beta and use in alpha.

LCaparelli: Cool, I haven't dug deeper to check if all of them are exactly the same, but I think there are some. I can open an issue after I have a closer look and send in a follow-up. Thanks!

apis/v1alpha2/gateway_types.go
```diff
@@ -196,9 +196,48 @@ type GatewayConditionType = v1beta1.GatewayConditionType
...
+	// configuration that will soon be ready in the underlying data plane.
+	//
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource.
```
robscott, line 203: Since we're saying it should always be present, we should probably also require "ObservedGeneration" be set with this condition.

apis/v1alpha2/gateway_types.go
```diff
@@ -196,9 +196,48 @@ type GatewayConditionType = v1beta1.GatewayConditionType
...
+	// This reason is used with the "Programmed" condition when the Listener is
+	// syntactically or semantically invalid.
+	GatewayReasonInvalid GatewayConditionReason = "Invalid"
+)
+
+const (
+	// This condition is true when the controller managing the Gateway is
+	// syntactically and semantically valid enough to produce some configuration
+	// in the underlying data plane, though it has not yet produced such
+	// configuration.
```
robscott, line 240: Maybe out of scope for this, but given the explicit requirement that "Programmed" always be set, it seems like we should add the same requirement here. 

Also if my understanding of this condition is correct, we want this to be to be set regardless of if the underlying dataplane has been configured. The current spec seems to suggest that the condition should only be set if the underlying dataplane has not been configured yet.

apis/v1alpha2/gateway_types.go
```diff
@@ -231,8 +270,8 @@ const (
 	// Deprecated: use the "Accepted" condition with reason "Accepted" instead.
 	GatewayReasonScheduled GatewayConditionReason = "Scheduled"
 
-	// This reason is used with the "Accepted" condition when no controller has
-	// reconciled the Gateway.
+	// This reason is used with the "Accepted", "Programmed" and "Ready"
+	// conditions when no controller has reconciled the Gateway.
 	GatewayReasonPending GatewayConditionReason = "Pending"
```
robscott, line 275: I think we may have gotten this wrong in the GEP. It seems like the current recommendation in API Conventions is "True/False/Unknown", where "Unknown" is used similarly to how we're recommending "Pending" being used: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties.

/cc @youngnick

apis/v1alpha2/gateway_types.go
```diff
@@ -455,10 +492,45 @@ const (
...
 
 const (
-	// This condition indicates whether the Listener has been
-	// configured on the Gateway.
+	// This condition indicates whether a Listener have generated some
+	// configuration that will soon be ready in the underlying data plane.
 	//
-	// Possible reasons for this condition to be true are:
+	// It is a positive-polarity summary condition, and so should always be
+	// present on the resource.
```
robscott, line 499: Same comment as above, we should probably require ObservedGeneration here as well.

apis/v1alpha2/gateway_types.go
```diff
@@ -455,10 +492,45 @@ const (
 )
 
 const (
-	// This condition indicates whether the Listener has been
-	// configured on the Gateway.
+	// This condition indicates whether a Listener have generated some
```
robscott, line 495: Nit: I think "has" is better than "have" here

apis/v1alpha2/gateway_types.go
```diff
@@ -480,12 +552,12 @@ const (
...
 	// Listener is syntactically or semantically invalid.
 	ListenerReasonInvalid ListenerConditionReason = "Invalid"
 
-	// This reason is used with the "Accepted" and "Ready" conditions when the
-	// Listener is either not yet reconciled or not yet not online and ready to
-	// accept client traffic.
+	// This reason is used with the "Accepted", "Ready" and "Programmed"
+	// conditions when the Listener is either not yet reconciled or not yet not
+	// online and ready to accept client traffic.
 	ListenerReasonPending ListenerConditionReason = "Pending"
```
robscott, line 562: Same comment as above re: "Pending" vs. "Unknown"

robscott (commented):
Thanks @LCaparelli! This mostly LGTM, would appreciate a review from @youngnick as well though.

apis/v1alpha2/gateway_types.go
youngnick: It could use clarification, but I thought that this was for the defaults, so the Type should be `Accepted`, `Programmed`, or `Ready`, the status is `Unknown`, and then the Reason is `Pending`.

Maybe like this?
```suggestion
	// This reason is used with the "Accepted", "Programmed" and "Ready"
	// conditions when the status is "Unknown" and no controller has reconciled the Gateway.
	GatewayReasonPending GatewayConditionReason = "Pending"
```

apis/v1alpha2/gateway_types.go
youngnick: Again, I think that the _status_ should be `Unknown`, but the _reason_ can be `Pending`.

youngnick (commented):
The API changes look great, but I think there's quite a bit more work to do to update all the conformance tests.

@LCaparelli, did you want to round out the conformance test updates in this PR, or should one of us do it in a follow-up?

LCaparelli:
@youngnick, @robscott thanks for the review!

> @LCaparelli, did you want to round out the conformance test updates in this PR, or should one of us do it in a follow-up?

Oh, I kinda just searched for usages of `ListenerReasonReady` and `GatewayConditionReady` consts and replaced them, it's what I understood from the original issue's:

> Update all conformance tests that check the "Ready" Gateway or Listener condition to check "Programmed" instead.


Am I missing other places where this is checked but is not using the consts? Or did you mean something else?

If you mean other tests that were not initially in the issue's scope I'm happy to help and can either do it here or in a follow-up.

youngnick:
Thanks for the update @LCaparelli!

Hmm, I just went and checked, and I was thinking of this function call:
https://github.com/kubernetes-sigs/gateway-api/blob/9568c673895dd3ff0c2b1d50029702af619c7f2c/conformance/utils/kubernetes/helpers.go#L135-L139

But it turns out that's actually _already_ checking for `Accepted` instead, so is misnamed. Would you be able to tidy that up please?

Additionally, it looks like this function needs an update:
https://github.com/kubernetes-sigs/gateway-api/blob/9568c673895dd3ff0c2b1d50029702af619c7f2c/conformance/utils/kubernetes/helpers.go#L92-L95
It's looking for the Gateway to be `Ready` , but we're not using the constant, just the string `Ready`. May be worth a case-sensitive search for `Ready` as well to find any other uses in the `conformance/` directory.

Sorry to be picky, but I want to make sure we squash all the `Ready` references in one go.

LCaparelli:
Yup, no worries at all @youngnick! Thanks for the additional pointers, I'll take a more thorough look

LCaparelli:
@youngnick @robscott I think I got it all covered this time, please take a look at the last 4 commits when you can. Thanks again for the guidance!

youngnick:
I can't see anything else this time, nice work @LCaparelli. I'll approve and let @robscott or @shaneutt do the final LGTM.

This is the one that will break v0.5.1 conformance though, i think.

/approve

apis/v1alpha2/gateway_types.go
```diff
@@ -196,9 +196,47 @@ type GatewayConditionType = v1beta1.GatewayConditionType
 type GatewayConditionReason = v1beta1.GatewayConditionReason
 
 const (
-	// This condition is true when the controller managing the
-	// Gateway has scheduled the Gateway to the underlying network
-	// infrastructure.
+	// This condition indicates whether a Gateway have generated some
```
shaneutt, line 199:
```suggestion
	// This condition indicates whether a Gateway has generated some
```

apis/v1beta1/gateway_types.go
```diff
@@ -516,9 +516,47 @@ type GatewayConditionType string
 type GatewayConditionReason string
 
 const (
-	// This condition is true when the controller managing the
-	// Gateway has scheduled the Gateway to the underlying network
-	// infrastructure.
+	// This condition indicates whether a Gateway have generated some
```
shaneutt, line 519:
```suggestion
	// This condition indicates whether a Gateway has generated some
```

shaneutt (approved):
Looks good, couple small language nits.

LCaparelli:
Thanks @shaneutt, fixed!

shaneutt (approved):
/lgtm

PR #1502, opened 2022-11-03
docs: mark GEP-735 as declined

shaneutt:
**What type of PR is this?**

/kind gep

At Kubecon 2022 in our working sessions, we decided to explicitly mark this GEP as declined so as to not cause any confusion whether we have any specific plans to implement it in an upcoming release, and to document the reasoning why it has been declined.

NONE


robscott:
Thanks @shaneutt!

/lgtm

PR #1507, opened 2022-11-04
Conformance: Merging SupportedFeatures and ExemptFeatures, adding TCPRoute feature

### LINKED ISSUE #1273: Conformance: Add option to run tests against experimental CRDs
With the recent v0.5.0 release, some conformance tests now depend on a resource that is only available in the experimental channel (ReferenceGrant). To improve this experience we should:

1. Add a new "MinChannel" string option to [Suite](https://github.com/kubernetes-sigs/gateway-api/blob/9537337d838cfebe3775c51be83c486a127c0d88/conformance/utils/suite/suite.go#L69) that defaults to "standard"
1. Add a new "MinChannel" option to [ConformanceTest](https://github.com/kubernetes-sigs/gateway-api/blob/9537337d838cfebe3775c51be83c486a127c0d88/conformance/utils/suite/suite.go#L153) - I don't think we should have a default here
1. Update all [conformance tests](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/tests) to specify `MinChannel: Standard` unless they include ReferenceGrant
1. Add logic to [conformanceTest.Run()](https://github.com/kubernetes-sigs/gateway-api/blob/9537337d838cfebe3775c51be83c486a127c0d88/conformance/utils/suite/suite.go#L166) that skips tests where `test.minChannel < suite.minChannel`

robscott:
**What type of PR is this?**
/kind cleanup
/area conformance

This attempts to simplify our conformance feature gating so we only have one mechanism. It also adds a feature gate for tests covering TCPRoute. Although I originally thought a "MinChannel" param would be the best way to handle this, I think feature gates will provide much more flexibility.

Fixes #1273

```release-note
Conformance: ExemptFeatures have been merged into SupportedFeatures.
```


conformance/tests/gateway-unsupported-route-kind.yaml
```diff
@@ -30,5 +30,5 @@ spec:
         namespaces:
           from: All
         kinds:
-        - kind: UDPRoute
+        - kind: TCPRoute
```
mlavacca, line 33: Should having a `TCPRoute` carrying HTTP traffic be allowed? What would be the use-case in using a `TCPRoute` instead of an `HTTPRoute`?

mikemorris (approved):
Mostly LGTM, I think the `gateway-unsupported-route-kind` and `gateway-supported-and-unsupported-route-kind` tests might have slightly misleading names though, and should instead be `-allowed-and-disallowed`, as they appear to be testing the intersection of `protocol` and `allowedRoutes` for known/supported Kinds, not an actually unknown/unsupported Group/Kind listed in `allowedRoutes` (I'm not sure if behavior is well-defined for that yet or tested anywhere else).

+1 on keeping feature gates over something like `MinChannel`

/approve

conformance/tests/gateway-unsupported-route-kind.yaml
robscott: Yeah this was more confusing than helpful, I refactored this test to more clearly be about an Invalid kind since that seemed to be the original goal. I'd been getting this one confused with the related "disallowed" kind tests.

robscott:
Thanks for the reviews! Updated to address everything, PTAL.

conformance/tests/gateway-invalid-route-kind.yaml
```diff
@@ -30,5 +30,5 @@ spec:
         namespaces:
           from: All
         kinds:
-        - kind: UDPRoute
+        - kind: InvalidRoute
```
mlavacca, line 33: This test is about checking whether the route is compatible with the protocol type. The test description states 
```
A Gateway in the gateway-conformance-infra namespace should fail to become ready if explicitly supports any route type incompatible with the protocol type, even if there are types which are valid as well.
```
And for this reason, having a `UDPRoute` with protocol `HTTP` is considered invalid. 
While the test of an invalid Route kind is something we do not have yet as far as I see from the code. Can you confirm this was your intention @robscott? I guess we can create an issue for this missing test and implement it in another PR. WDYT?

robscott: Good question, I'd wanted to take a step back from that as we're not prescriptive in the spec about which protocols are compatible with which route types. I think the most important takeaway here is that we want implementations to have consistent behavior if someone tries to attach an unsupported/invalid route kind to a Gateway. This slightly alters the purpose of the original test but I think it results in a clearer definition that more closely matches the spec.

robscott: I've updated the description of this test to more clearly match what it's actually doing now, hopefully the approach makes sense. 

It would probably be good to create a follow up issue like you're suggesting, but I think that issue should be focused on which route types we consider compatible with which Gateway protocols. Until we define that, I don't think we should have conformance tests that make assumptions about which combinations are valid/invalid.

mlavacca: It makes sense. I created #1515 to track it.

mlavacca (approved):

shaneutt (approved):
/lgtm

PR #1508, opened 2022-11-04
webhook: validates that if ParentRefs includes 2 or more references to the same parent, parentRefs must specify unique sectionName values

### LINKED ISSUE #1481: webhook: validate HTTPRoute parentRefs

The `ParentRefs` field in the `CommonRouteSpec` states:
 https://github.com/kubernetes-sigs/gateway-api/blob/21bba43681b3db3a62d7ca9f108bf3be4c0cea71/apis/v1beta1/shared_types.go#L130-L132
For his reason, the `HTTPRoute` webhook should be improved in a way the above condition is always verified. Hence, the following check should be added to the webhook:

if an `HTTPRoute` references more than once the same Gateway as the parent, the `Parentref`'s `SectionName`s must be set and different, otherwise, the validation returns an error.

**Why this is needed**:

This is needed to ensure that an `HTTPRoute` does not reference the same Gateway (or the same listener) more than once.

gyohuangxin:
**What type of PR is this?**

/kind feature

This PR add a validation to validate that if ParentRefs includes 2 or more references to the same parent, parentRefs must specify unique sectionName values

Fixes #1481

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/util/validation/http_route.go
+++ b/apis/v1beta1/util/validation/http_route.go
@@ -1,0 +1,30 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+// containsInSectionNameSlice checks whether the provided SectionName
+// is in the target SectionName slice.
+func ContainsInSectionNameSlice(items []gatewayv1b1.SectionName, item *gatewayv1b1.SectionName) bool {
+	for _, eachItem := range items {
+		if eachItem == *item {
+			return true
+		}
+	}
+	return false
+}
--- a/apis/v1beta1/validation/httproute.go
+++ b/apis/v1beta1/validation/httproute.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	utils "sigs.k8s.io/gateway-api/apis/v1beta1/util/validation"
 )
 
 var (
@@ -68,6 +69,7 @@ func validateHTTPRouteSpec(spec *gatewayv1b1.HTTPRouteSpec, path *field.Path) fi
 		}
 	}
 	errs = append(errs, validateHTTPRouteBackendServicePorts(spec.Rules, path.Child("rules"))...)
+	errs = append(errs, validateHTTPParentRefs(spec.ParentRefs, path.Child("spec"))...)
 	return errs
 }
 
@@ -286,4 +288,40 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
 	}
 
 	return true
+}
+
+// validateHTTPParentRefs validates that if ParentRefs includes 2 or more references
+// to the same parent (based on kind, name, and namespace), those ParentRefs must
+// specify unique SectionName values.
+func validateHTTPParentRefs(parentRefs []gatewayv1b1.ParentReference, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if len(parentRefs) <= 1 {
+		return nil
+	}
+	type sameKindParentRefs struct {
+		name      gatewayv1b1.ObjectName
+		namespace gatewayv1b1.Namespace
+		kind      gatewayv1b1.Kind
+	}
+	m := make(map[sameKindParentRefs][]gatewayv1b1.SectionName)
+	for i, p := range parentRefs {
+		if p.Namespace == nil {
+			p.Namespace = new(gatewayv1b1.Namespace)
+		}
+		if p.Kind == nil {
+			p.Kind = new(gatewayv1b1.Kind)
+		}
+		if p.SectionName == nil {
+			p.SectionName = new(gatewayv1b1.SectionName)
+		}
+		t := &sameKindParentRefs{name: p.Name, namespace: *p.Namespace, kind: *p.Kind}
+		if s, ok := m[*t]; ok {
+			if utils.ContainsInSectionNameSlice(s, p.SectionName) {
+				errs = append(errs, field.Invalid(path.Index(i).Child("parentRefs").Child("sectionName"), p.SectionName, "must be set and unique when ParentRefs includes 2 or more references to the same parent"))
+				return errs
+			}
+		}
+		m[*t] = append(m[*t], *p.SectionName)
+	}
+	return errs
 }
--- a/apis/v1beta1/validation/httproute_test.go
+++ b/apis/v1beta1/validation/httproute_test.go
@@ -982,3 +982,137 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateHTTPParentRefs(t *testing.T) {
+	namespace := gatewayv1b1.Namespace("example-namespace")
+	kind := gatewayv1b1.Kind("Gateway")
+	sectionA := gatewayv1b1.SectionName("Section A")
+	sectionB := gatewayv1b1.SectionName("Section B")
+	sectionC := gatewayv1b1.SectionName("Section C")
+
+	tests := []struct {
+		name       string
+		parentRefs []gatewayv1b1.ParentReference
+		errCount   int
+	}{{
+		name: "valid HTTPRouteParentRefs includes 1 reference",
+		parentRefs: []gatewayv1b1.ParentReference{
+			{
+				Name:        "example",
+				Namespace:   &namespace,
+				Kind:        &kind,
+				SectionName: &sectionA,
+			},
+		},
+		errCount: 0,
+	}, {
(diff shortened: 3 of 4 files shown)


gyohuangxin:
@robscott @mlavacca Could you help review this?

apis/v1beta1/util/validation/http_route.go
```diff
@@ -0,0 +1,30 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import gatewayv1b1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+// containsInSectionNameSlice checks whether the provided SectionName
```
mlavacca, line 21:
```suggestion
// ContainsInSectionNameSlice checks whether the provided SectionName
```

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,39 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+	var errs field.ErrorList
+	if len(parentRefs) <= 1 {
+		return nil
+	}
+	type sameKindParentRefs struct {
+		name      gatewayv1b1.ObjectName
+		namespace gatewayv1b1.Namespace
+		kind      gatewayv1b1.Kind
+	}
+	m := make(map[sameKindParentRefs][]gatewayv1b1.SectionName)
```
mlavacca, line 306: can you use a meaningful name for this variable, please?

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,39 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+	for i, p := range parentRefs {
+		if p.Namespace == nil {
```
mlavacca, line 308: Since we need to validate the ParentRefs, I'd prefer not to touch the original resource but instead, to rely only on additional variables; it'd look cleaner and easier to understand. WDYT?

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,39 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+			p.Namespace = new(gatewayv1b1.Namespace)
+		}
+		if p.Kind == nil {
+			p.Kind = new(gatewayv1b1.Kind)
+		}
+		if p.SectionName == nil {
+			p.SectionName = new(gatewayv1b1.SectionName)
+		}
+		t := &sameKindParentRefs{name: p.Name, namespace: *p.Namespace, kind: *p.Kind}
```
mlavacca, line 317: What's the point of getting a pointer to the struct if it is always dereferenced in the subsequent usages?

mlavacca (commented):
Thanks for this PR @gyohuangxin! In addition to the comment below, I'd ask you to add this validation to `v1alpha2` as well. Furthermore, The port check should be added, as discussed in the issue.

apis/v1beta1/validation/httproute.go
gyohuangxin: My mistake.. fixed.

apis/v1beta1/validation/httproute.go
gyohuangxin: Do you mean that we only check the addition/incremental ParentRefs instead of checking all ParentRefs? Can you explain more about the reason? And IMO, there is no way to only check incremental resource in Kubernetes webhook validation, because it cannot tell the incremental or original resource, correct me if I'm wrong.

gyohuangxin:
@mlavacca Thanks for your comments, I've added it to v1alpha2 as well. And about port checks, as @robscott's [comments](https://github.com/kubernetes-sigs/gateway-api/issues/1481#issuecomment-1297873260) said, the ParentRefs' SectionNames are always different, and ports can be the same, so in my opinion, we don't need port checks anymore. What do you think?

apis/v1beta1/validation/httproute.go
mlavacca: my point is only related to the way this code is written: you are initializing the nil fields in the `parentRefs`. I'm just suggesting rewriting this portion of the code to leave the original `parentRefs` untouched and rely only on the additional data structure. E.g.,
```go
t := sameKindParentRefs{name: p.Name, namespace: *p.Namespace, kind: *p.Kind}
if t.namespace == nil {
    t.namespace = new(gatewayv1b1.Namespace)
}
...
```

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,39 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+		if p.Namespace == nil {
+			p.Namespace = new(gatewayv1b1.Namespace)
+		}
+		if p.Kind == nil {
+			p.Kind = new(gatewayv1b1.Kind)
+		}
+		if p.SectionName == nil {
+			p.SectionName = new(gatewayv1b1.SectionName)
+		}
+		t := sameKindParentRefs{name: p.Name, namespace: *p.Namespace, kind: *p.Kind}
```
mlavacca, line 317: Can you use a meaningful name for this variable as well, please?

mlavacca:
> @mlavacca Thanks for your comments, I've added it to v1alpha2 as well. And about port checks, as @robscott's [comments](https://github.com/kubernetes-sigs/gateway-api/issues/1481#issuecomment-1297873260) said, the ParentRefs' SectionNames are always different, and ports can be the same, so in my opinion, we don't need port checks anymore. What do you think?

You are right, I didn't see that message :)

apis/v1beta1/validation/httproute_test.go
```diff
@@ -982,3 +982,137 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateHTTPParentRefs(t *testing.T) {
```
mlavacca, line 986: can you please add a couple more test cases like 
* different listeners having the same `sectionName`. (it should pass)
* the same listener referenced twice; one reference does have the sectionName, but the other doesn't. (it shouldn't pass)

gyohuangxin: > the same listener referenced twice; one reference does have the sectionName, but the other doesn't. (it shouldn't pass)

In my current implementation, this case will pass because it assumes their sectionNames are different. 
But I think this case is correct, I will add some checks to make sure it happens.

gyohuangxin: I add more checks when the same listener referenced twice:
```
		if s, ok := parentRefsSectionMap[targetParentRefs]; ok {
			if len(s[0]) == 0 || len(*section) == 0 {
				errs = append(errs, field.Required(path, "ParentRefs section names must all be set when ParentRefs includes 2 or more references to the same parent"))
				return errs
			}
			if utils.ContainsInSectionNameSlice(s, section) {
				errs = append(errs, field.Invalid(path.Index(i).Child("parentRefs").Child("sectionName"), section, "must be unique when ParentRefs includes 2 or more references to the same parent"))
				return errs
			}
		}
```
It means if the first ParentRefs section name is not set, or current ParentRefs section name is not set, it will also raise an error.

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,46 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+		if p.Namespace != nil {
+			namespace = p.Namespace
+		}
+		if p.Kind != nil {
+			kind = p.Kind
+		}
+		if p.SectionName != nil {
+			section = p.SectionName
+		}
+		targetParentRefs := sameKindParentRefs{name: p.Name, namespace: *namespace, kind: *kind}
```
mlavacca, line 320:
```suggestion
		targetParentRefs := sameKindParentRefs{name: p.Name, namespace: new(gatewayv1b1.Namespace), kind: new(gatewayv1b1.Kind)}
		if p.Namespace != nil {
			targetParentRefs.namespace = *p.Namespace
		}
		if p.Kind != nil {
			targetParentRefs.kind = *p.Kind
		}
		if p.SectionName != nil {
			targetParentRefs.section = *p.SectionName
		}
```
I meant something like this, I'm not sure it compiles, but still. Does it make sense to you?

mlavacca (commented):
Apart from a minor nitpick, lgtm

apis/v1beta1/validation/httproute.go
gyohuangxin: Yes, it makes sense. I've updated like that.

mlavacca (approved):
Thanks for your work, @gyohuangxin!

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,44 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
 
 	return true
 }
+
+// ValidateHTTPParentRefs validates that if ParentRefs includes 2 or more references
+// to the same parent (based on kind, name, and namespace), those ParentRefs must
+// specify unique SectionName values.
```
shaneutt, line 295: Nice and clear `godoc` comment, thanks.

apis/v1beta1/validation/httproute.go
```diff
@@ -287,3 +289,44 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
...
+func ValidateHTTPParentRefs(parentRefs []gatewayv1b1.ParentReference, path *field.Path) field.ErrorList {
```
shaneutt, line 296: So it seems to me that while we name this with `HTTP` we could just as easily apply this across other APIs. Is it worth making this validation more generic and plugging it in to validation for all routes kinds?

gyohuangxin: Yes, I agree. I will create a generic validation than can be used for all routes kinds.

apis/v1beta1/validation/httproute.go
gyohuangxin: Updated, thanks.

PR #1518, opened 2022-11-08
apis: update Group default docs and examples for *Reference types

mikemorris:
**What type of PR is this?**
/kind documentation

Clarifies the current default behavior when the `Group` field is unspecified for a `ParentReference`, and how to set the core API group for a `Service` referent (which will be relevant for GAMMA implementations).

For `BackendObjectReference`, `SecretObjectReference` and `LocalObjectReference`, disambiguates the behavior of an unspecified `Group` field from an empty string.

Updates the example `Group` values to `gateway.networking.k8s.io` (from `networking.k8s.io`) - this may not be necessary, but felt like it would be a more common value when the `Group` field is set explicitly.

N/A

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/object_reference_types.go
+++ b/apis/v1beta1/object_reference_types.go
@@ -25,8 +25,8 @@ package v1beta1
 // be rejected by the implementation, with appropriate Conditions set
 // on the containing object.
 type LocalObjectReference struct {
-	// Group is the group of the referent. For example, "networking.k8s.io".
-	// When unspecified (empty string), core API group is inferred.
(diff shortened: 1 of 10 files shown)


howardjohn (approved):

robscott (commented):
This change LGTM. One tiny nit is that "inferred" feels different than "default" and it could be useful to somehow include "default" in the language here. I'm not exactly sure how to integrate that though, and it doesn't need to block this.

/approve

shaneutt (approved):
/lgtm

Issue #1517, opened 2022-11-08
docs: TLS doc showing 'Macro rendering err'


xtineskim:
**What happened**:
The page here https://gateway-api.sigs.k8s.io/guides/tls/ is erroring. Template is not found 😲 
(update) and this page https://gateway-api.sigs.k8s.io/blog/2021/introducing-v1alpha2/
<img width="1245" alt="Screen Shot 2022-11-08 at 9 52 02 AM" src="https://user-images.githubusercontent.com/35876108/200596723-83381077-57c0-4f39-b49f-0277e632dbf0.png">

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

xtineskim:
https://github.com/kubernetes-sigs/gateway-api/pull/1513 fixes it!

PR #1513, opened 2022-11-08
Fix Broken Links in Docs Causing Rendering Failures

spencerhance:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:

* These broken links are causing rendering issues on the published docs and also show up when runnning `make docs`.
* For everything except the v1alpha2 blog I simply updated the link
* For the v1alpha2 blog I copied the actual yaml from when the blog was posted.
* Example of https://gateway-api.sigs.k8s.io/blog/2021/introducing-v1alpha2/ right now:
Screenshot 2022-11-07 7 16 31 PM
* `make docs` now reports no errors when I run it locally.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #1519, opened 2022-11-08
docs: Change yaml paths to standard

xtineskim:
**What type of PR is this?**

/kind documentation

Fixes the issue I created https://github.com/kubernetes-sigs/gateway-api/issues/1517 where the `experimental` reference to some yaml files generated errs on the gateway-api.sigs.k8s.io (for example:  https://gateway-api.sigs.k8s.io/guides/tls/)

Fixes  https://github.com/kubernetes-sigs/gateway-api/issues/1517

```release-note
NONE
```


robscott:
Thanks for reporting and fixing this @xtineskim! It looks like @spencerhance got #1513 in first which I think includes all the fixes here as well.

PR #1522, opened 2022-11-10
Fix job problem of build docker image

wilsonwu:
**What type of PR is this?**
/kind bug

I saw the job failed and try to fix.

The main branch job failed.

### DIFF
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -67,10 +67,11 @@ fi
 # And, we add an extra version tag - either :latest or semver.
 # The buildx integrate build and push in one line.
 docker buildx build \
-    --platform linux/amd64,linux/arm64 \
+    -f Dockerfile \
+    -t ${REGISTRY}/admission-server:${GIT_TAG} \
+    -t ${REGISTRY}/admission-server:${VERSION_TAG} \
     --build-arg "COMMIT=${COMMIT}" \
     --build-arg "TAG=${BINARY_TAG}" \
     --platform linux/amd64,linux/arm64 \
-    -t ${REGISTRY}/admission-server:${GIT_TAG} \
-    -t ${REGISTRY}/admission-server:${VERSION_TAG} \
-    --push
+    --push \
+    .
--- a/hack/build-and-push.sh
+++ b/hack/build-and-push.sh
@@ -66,7 +66,7 @@ fi
 # Note that an image will *always* be built tagged with the GIT_TAG, so we know when it was built.
 # And, we add an extra version tag - either :latest or semver.
 # The buildx integrate build and push in one line.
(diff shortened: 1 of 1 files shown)


youngnick (changes requested):
Thanks for the PR, @wilsonwu.

However, I don't think that just removing the build tags is the right option. If you look at line 23 of the Dockerfile, you can see that those build tags are used in the `go build` command to set versions in the binary.

If the build tags are causing issues, I think that it's better to either figure out what's causing the issues, or remove all trace of them from the Dockerfile as well.

wilsonwu:
> Thanks for the PR, @wilsonwu.
> 
> However, I don't think that just removing the build tags is the right option. If you look at line 23 of the Dockerfile, you can see that those build tags are used in the `go build` command to set versions in the binary.
> 
> If the build tags are causing issues, I think that it's better to either figure out what's causing the issues, or remove all trace of them from the Dockerfile as well.

I searched that, the `buildx` can support `--build-arg`, not sure the job base docker image `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90` could support buildx or not?

wilsonwu:
> > Thanks for the PR, @wilsonwu.
> > However, I don't think that just removing the build tags is the right option. If you look at line 23 of the Dockerfile, you can see that those build tags are used in the `go build` command to set versions in the binary.
> > If the build tags are causing issues, I think that it's better to either figure out what's causing the issues, or remove all trace of them from the Dockerfile as well.
> 
> I searched that, the `buildx` can support `--build-arg`, not sure the job base docker image `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90` could support buildx or not?

I checked that the `gcb-docker-gcloud` image support `buildx`: https://github.com/kubernetes/test-infra/tree/master/images/gcb-docker-gcloud

wilsonwu (commented):
Updated, how can I run the job to test it?

shaneutt:
> Updated, how can I run the job to test it?

You should be able to use your own registry to run the script and test it locally, but if you're asking whether we can just run the job in CI as part of this PR to test it I don't think so, right @robscott ?

wilsonwu:
> > Updated, how can I run the job to test it?
> 
> You should be able to use your own registry to run the script and test it locally, but if you're asking whether we can just run the job in CI as part of this PR to test it I don't think so, right @robscott ?

Got it, I will test it ASAP.

wilsonwu:
> > > Updated, how can I run the job to test it?
> > 
> > 
> > You should be able to use your own registry to run the script and test it locally, but if you're asking whether we can just run the job in CI as part of this PR to test it I don't think so, right @robscott ?
> 
> Got it, I will test it ASAP.

Fixed and tested in local successful:
image

PTAL @shaneutt

shaneutt (approved):
/lgtm

Thanks for testing :+1:

PR #1524, opened 2022-11-10
conformance: fix invalid `for` loop in InvalidTls test

howardjohn:
/kind bug

Makes the test actually run all 4 tests instead of run 1 test 4 times

NOTE: Istio is actually not conformant here so I didn't test that the new tests actually work as expected

```release-note
NONE
```


shaneutt (approved):
Yep, bad loop closure in there. Thanks for finding and fixing this :+1: 

/lgtm

PR #1526, opened 2022-11-10
Mark multiple `GatewayTLSConfig.CertificateRefs` as Extended support

### LINKED ISSUE #1330: Support and test multiple TLS certificates per Gateway listener

ksankeerth:
**What type of PR is this?**
/kind documentation

As per the https://github.com/kubernetes-sigs/gateway-api/issues/1330#issuecomment-1306430866 , the multiple certificates field has to be marked as "Extended Support"

Fixes #1330

cc : @youngnick 

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -327,7 +327,7 @@ type GatewayTLSConfig struct {
 	//
 	// Support: Core - A single reference to a Kubernetes Secret of type kubernetes.io/tls
 	//
-	// Support: Implementation-specific (More than one reference or other resource types)
+	// Support: Extended (More than one reference or other resource types)
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=64
--- a/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
@@ -344,8 +344,7 @@ spec:
                             Kubernetes resources, i.e. Secret, or implementation-specific
                             custom resources. \n Support: Core - A single reference
                             to a Kubernetes Secret of type kubernetes.io/tls \n Support:
-                            Implementation-specific (More than one reference or other
-                            resource types)"
+                            Extended (More than one reference or other resource types)"
                           items:
                             description: "SecretObjectReference identifies an API
                               object including its namespace, defaulting to Secret.
@@ -1038,8 +1037,7 @@ spec:
                             Kubernetes resources, i.e. Secret, or implementation-specific
                             custom resources. \n Support: Core - A single reference
                             to a Kubernetes Secret of type kubernetes.io/tls \n Support:
-                            Implementation-specific (More than one reference or other
-                            resource types)"
+                            Extended (More than one reference or other resource types)"
                           items:
                             description: "SecretObjectReference identifies an API
                               object including its namespace, defaulting to Secret.
--- a/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
+++ b/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
@@ -344,8 +344,7 @@ spec:
                             Kubernetes resources, i.e. Secret, or implementation-specific
                             custom resources. \n Support: Core - A single reference
                             to a Kubernetes Secret of type kubernetes.io/tls \n Support:
-                            Implementation-specific (More than one reference or other
-                            resource types)"
+                            Extended (More than one reference or other resource types)"
                           items:
                             description: "SecretObjectReference identifies an API
                               object including its namespace, defaulting to Secret.
@@ -1038,8 +1037,7 @@ spec:
                             Kubernetes resources, i.e. Secret, or implementation-specific
                             custom resources. \n Support: Core - A single reference
                             to a Kubernetes Secret of type kubernetes.io/tls \n Support:
-                            Implementation-specific (More than one reference or other
-                            resource types)"
+                            Extended (More than one reference or other resource types)"
                           items:
                             description: "SecretObjectReference identifies an API
                               object including its namespace, defaulting to Secret.


apis/v1beta1/gateway_types.go
```diff
@@ -327,7 +327,7 @@ type GatewayTLSConfig struct {
 	//
 	// Support: Core - A single reference to a Kubernetes Secret of type kubernetes.io/tls
 	//
-	// Support: Implementation-specific (More than one reference or other resource types)
+	// Support: Extended (More than one reference or other resource types)
```
robscott, line 330: I think this is going to need some more discussion. If we say that supporting multiple TLS certificates is "Extended" support, we'll need to define how multiple TLS certs should be handled. I'm not entirely sure what that looks like today, but it could be worth getting some input from anyone currently supporting that.

One concrete change here would be to limit the scope of "Extended" support to Kubernetes TLS secrets. That would result in the following:

```
	// Support: Core - A single reference to a Kubernetes Secret of type kubernetes.io/tls
	//
	// Support: Extended - Multiple references to Kubernetes Secrets of type kubernetes.io/tls
	//
	// Support: Implementation-specific - References or other resource types
```

robscott (commented):
Thanks for starting this @ksankeerth! I think this is a necessary change, but it may take a bit of time to spec out how we want to handle multiple TLS certs.

/cc @youngnick @shaneutt @howardjohn @mikemorris

youngnick:
@ksankeerth, remember to run `make generate` to clear the `pull-gateway-api-verify` job each time you make a change to the spec.

apis/v1beta1/gateway_types.go
shaneutt: I've added an agenda item for next week's community meeting to discuss this and bump this old PR.

robscott:
Hey @ksankeerth, do you still have time to work on this?

ksankeerth:
@robscott @shaneutt  apologies for the late reply. I can work on this.  A small confusion on this,(sorry if i missed any discussion).

>I think this is going to need some more discussion. If we say that supporting multiple TLS certificates is "Extended" support, we'll need to define how multiple TLS certs should be handled. I'm not entirely sure what that looks like today, but it could be worth getting some input from anyone currently supporting that.

>One concrete change here would be to limit the scope of "Extended" support to Kubernetes TLS secrets. That would result in the following:


>
https://github.com/kubernetes-sigs/gateway-api/pull/1526#discussion_r1019755797
It was mentioned that some discussions are needed about marking multi-TLS support as "implementation-specific" or "extended but with limited scope". Were there any discussions/conclusions on this?

robscott:
Hey @ksankeerth, sorry for the delay and overall confusion on this one! I've been struggling to figure out what makes sense going forward. Generally we don't mark features as "Extended" support unless we've clearly defined how they'd work. I'm not really sure what level of consistency we can guarantee across implementations when multiple certs are attached to the same listener. Unfortunately I think this is going to require some high level knowledge of how major underlying implementations (Envoy, NGINX, HAProxy, etc) handle multiple certs. If there's a sufficient path to consistent implementations we can likely move forward with this, but we _may_ need a GEP for this if it becomes complex. 

Would appreciate some additional ideas on what kind of requirements we could add to the spec when attaching multiple certs to a listener. Without looking into it more, I'm not sure what the common ground is here.

/cc @youngnick @shaneutt @sunjayBhatia @howardjohn @arkodg

shaneutt:
I appreciate your work here @ksankeerth. I agree with @robscott that we appear to be in GEP territory to discuss and build consensus on some of the finer details on how this _should_ work in the usual format for that (doing it with a code PR can be troublesome, and GEPs are good for posterity about decisions made).

arkodg:
from a capability perspective Envoy does support [multiple server tls certificates](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extensions-transport-sockets-tls-v3-commontlscontext)
and this is also an ask in Envoy Gateway https://github.com/envoyproxy/gateway/issues/393

apis/v1beta1/gateway_types.go
arkodg: I really like @robscott's suggestion

arkodg:
relevant links to show
* NGINX supports multiple certs - https://github.com/kubernetes/ingress-nginx/issues/6707
* HAProxy supports multiple certs - https://www.haproxy.com/blog/haproxy-ssl-termination/#choosing-the-certificate-with-sni

but as other suggested, probably needs some more clarification on which subset of the certs will we used for which traffic flow, to ensure interop amongst implementations

shaneutt:
@ksankeerth since we haven't heard back from you in a couple of weeks, and since it sounds like we're kinda getting into GEP territory (please see above comments) we're going to mark this one as draft for the time being to signal that we're waiting on some updates and it doesn't need active review.

ksankeerth:
> @ksankeerth since we haven't heard back from you in a couple of weeks, and since it sounds like we're kinda getting into GEP territory (please see above comments) we're going to mark this one as draft for the time being to signal that we're waiting on some updates and it doesn't need active review.

Yes, I saw the discussion. GEP seems to be good. If any help needed, I'm happy to help

robscott:
Hey @arkodg, thanks for following up on this! I think it's clear that this concept is broadly supportable, I'm just a bit concerned that there may be some nuanced differences in how things like conflicts or overlapping certificates are handled. I'm not quite sure how specific we need to be in the API Spec, but I want to make sure we're not missing an opportunity to ensure more consistent implementation details. At a minimum we likely need to require that implementations attach certificates in the order they're specified, but maybe there are more details I'm missing.

I think it would be worth getting a broader discussion around this, so I've added this to the [agenda for our next meeting](https://docs.google.com/document/d/1eg-YjOHaQ7UD28htdNxBR3zufebozXKyI28cl2E11tU/edit#).

Issue #1514, opened 2022-11-08
conformance: Test that webhook validations are performed


youngnick:
**What would you like to be added**:
This issue covers adding conformance tests that test that the implementation is implementing the same blocks as are implemented in the webhook. The simplest way to do this is to run the webhook we provide, but the conformance tests should be a bit agnostic about this.

However, I believe that the conformance test should check that invalid objects are never persisted into the cluster by trying to apply them, and checking that the apply fails. That flow will require some sort of webhook, or possibly the extensions to CRD validation api-machinery is considering (which I can't remember the name of right now).

A secondary part of this effort should also clarify some guidelines about what controllers should do with invalid objects - hopefully we can end up with this being as easy as "run a Validate function on the object, and drop the object if it fails" or something.

**Why this is needed**:
This will make it _very_ clear that the webhook is _required_ for having a conformant implementation.

This issue came up during discussion on #1497, but I think I've mentioned it before, and haven't written it down. Sorry all.

rainest:
If there is a case where we expect that objects are accepted and reach controllers, how to test that behavior isn't immediately obvious. For example, in the case of #1497, if guidance was:

> <Header filters that do bad things> are invalid and will be rejected by the webhook is installed. Implementations MUST exclude these resources from configuration and add an `Accepted=false` condition with a `HeaderFilterNotValid` reason if the webhook is not running.

We can only test one or the other in a given run. If the webhook _is_ running, the resource simply won't exist, and can't have conditions as such. If it _is not_ running, tests can check conditions, but any test to confirm the webhook behavior (that the resource is rejected at creation time) would then fail. This would basically require two suites of conformance tests for behavior with or without the webhook, correct?

youngnick:
Sigh, yes, you're right.

I think that if we're going to do this, we should lean in and say "we're not going to mandate anything about what happens if the webhook is not running, sorry".

shaneutt:
> Sigh, yes, you're right.
> 
> I think that if we're going to do this, we should lean in and say "we're not going to mandate anything about what happens if the webhook is not running, sorry".

This is inherently a problem with being out of tree, yeah?

[...]

PR #1534, opened 2022-11-14
[WIP] conformance: require that GatewayClass controller name is immutable

markmc:
Relates to #1514

This is a proof-of-concept conformance test that checks for the validation behavior provided by the admission webhook. In this case we're just checking that an update to a GatewayClass controller name is rejected.

We use a temporary GatewayClass rather than the one specified by the --gateway-class, to avoid messing up installations that do allow this update through.

This does raise an interesting question though - isn't it strange that we're saying an implementation is required to validate objects that it should otherwise ignore? In this case, we would would otherwise say an implementation should ingore this object because it doesn't recognize the controller name.

Failure looks like this:

```
=== RUN   TestConformance/GatewayClassAdmissionValidation/GatewayClass_controllerName_is_immutable
    gatewayclass-admission-validation.go:62:
        	Error Trace:	gatewayclass-admission-validation.go:62
        	Error:      	Should be in error chain:
        	            	expected: %!q(**errors.StatusError=0xc000014008)
        	            	in chain:
        	Test:       	TestConformance/GatewayClassAdmissionValidation/GatewayClass_controllerName_is_immutable
        	Messages:   	updating gatewayclass-immutable GatewayClass.Spec.ControllerName should not be permitted
```

### DIFF
--- a/conformance/tests/gatewayclass-admission-validation.go
+++ b/conformance/tests/gatewayclass-admission-validation.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


shaneutt:
This is marked as a WIP, so I'm converting it to draft for organizational reasons. If this is now ready for review, please convert it back from draft and remove the WIP and we'll take a look, thank you! :vulcan_salute:

PR #1529, opened 2022-11-10
chore: cleanup litespeed section on implementations page

rperper:
Added a section for the LiteSpeed Ingress Controller

**What type of PR is this?**

**What this PR does / why we need it**:

Announce LiteSpeed Ingress Controller Gateway API support.

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

yes
```release-note
Announce LiteSpeed Ingress Controller Gateway API support.
```

PR #1535, opened 2022-11-14
fix: cherry pick to re-apply google implementations page

shaneutt:
**What type of PR is this?**

/kind documentation

In https://github.com/kubernetes-sigs/gateway-api/pull/1529 we accidentally undid some changes @robscott had made to the Google section of the implementations page. This PR cherry-picks the original commit back on top to fix that mistake while also keeping authorship and attribution intact.


robscott:
Thanks @shaneutt!

/lgtm

PR #1540, opened 2022-11-17
Add PartiallyInvalid Route condition

### LINKED ISSUE #1521: webhook/conformance: Handle incompatible HTTP Filters on a rule
Re: The last two bullets in this section of GEP-1364: https://gateway-api.sigs.k8s.io/geps/gep-1364/#accepted

```
* A HTTPRoute with one rule that specifies a HTTPRequestRedirect filter and a HTTPURLRewrite filter. Accepted must be false,
  because there's only one rule, and this configuration for the rule is invalid (see [reference](https://gateway
  api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRequestRedirectFilter)) The error condition in this
  case is undefined currently - we should define it, thanks @sunjayBhatia.

* A HTTPRoute with two rules, one valid and one which specifies a HTTPRequestRedirect filter _and a HTTPURLRewrite filter.
  Accepted is true, because the valid rule can produce some config in the data plane. We'll need to raise the more specific error
  condition for an incompatible filter combination as well to make the partial validity clear.
```

For incompatible filters in an HTTPRoute, seems like we can do somethings similar to this change that validates HTTP header filters: https://github.com/kubernetes-sigs/gateway-api/pull/1497

We could do this instead of/in addition to the status changes described in the GEP

### LINKED ISSUE #1696: Define how to reflect partial validity of routing rules in the HTTPRoute status

If we have an HTTPRoute resource with some if its routing rules valid, some -- not valid (for example, [PathMatchType](https://github.com/kubernetes-sigs/gateway-api/blob/89b8079e775ba01336c3bbf18868cdda69ae11ec/apis/v1beta1/httproute_types.go#L303) in HTTPPathMatch is not one of Exact;PathPrefix;RegularExpression), define how to report that in the HTTPRoute status.

**Why this is needed**:

Current documentation is a bit contradictory:

The HTTPRoute docs [say](https://github.com/kubernetes-sigs/gateway-api/blob/89b8079e775ba01336c3bbf18868cdda69ae11ec/apis/v1beta1/httproute_types.go#L585):
```
	// Unknown values here must result in the implementation setting the
	// Accepted Condition for the Route to `status: False`, with a
``` 

However, in [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/), it is mentioned:

> HTTPRoute with two rules, one valid and one which specifies a HTTPRequestRedirect filter _and a HTTPURLRewrite filter. Accepted is true, because the valid rule can produce some config in the data plane. We'll need to raise the more specific error condition for an incompatible filter combination as well to make the partial validity clear.

This comes from https://github.com/kubernetes-sigs/gateway-api/discussions/1674 and a discussion on Kubernetes Gateway API Slack sig-network-gateway-api https://kubernetes.slack.com/archives/CR0H13KGA/p1675122817360569

sunjayBhatia:
**What type of PR is this?**
/kind cleanup
/kind documentation

~~Fixes documentation on HTTPRoute filters for when incompatible filters are present.~~

Adds PartiallyInvalid Route condition with one example Reason, "IncompatibleFilters"

This condition can be used in cases where cases where some configuration is valid but some is not (so Accepted=True, but not everything is valid/supported)

Fixes #1521
Fixes #1696 

```release-note
Add PartiallyInvalid Route condition to reflect Route status that is partially unsupported or invalid for an implementation.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -205,8 +205,14 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// document that limitation. A Route containing a single rule with incompatible
+	// filters must result in the implementation setting the Accepted Condition for
+	// the Route to `status: False`, with a Reason of `UnsupportedValue`. A Route
+	// containing multiple rules with a subset of them containing incompatible
+	// filters must result in the implementation setting the Accepted Contition for
+	// the Route to `status: True`. In all cases where incompatible or unsupported
+	// filters are specified, implementations MUST add the `IncompatibleFilters`
+	// condition to status.
 	//
 	// Support: Core
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -205,14 +205,9 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. A Route containing a single rule with incompatible
-	// filters must result in the implementation setting the Accepted Condition for
-	// the Route to `status: False`, with a Reason of `UnsupportedValue`. A Route
-	// containing multiple rules with a subset of them containing incompatible
-	// filters must result in the implementation setting the Accepted Contition for
-	// the Route to `status: True`. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add the `IncompatibleFilters`
-	// condition to status.
+	// document that limitation. In all cases where incompatible or unsupported
+	// filters are specified, implementations MUST add the `PartiallyInvalid` error
+	// condition with status `True` and reason `IncompatibleFilters`.
 	//
 	// Support: Core
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -205,9 +205,10 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add the `PartiallyInvalid` error
-	// condition with status `True` and reason `IncompatibleFilters`.
+	// document that limitation. In cases where incompatible or unsupported
+	// filters are specified on subset of the rules on a route, implementations
+	// MUST add the `PartiallyInvalid` error condition with status `True` and
+	// reason `IncompatibleFilters`.
 	//
 	// Support: Core
 	//
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -290,10 +290,15 @@ const (
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
 
 	// This condition indicates whether configuration present in a Route is
-	// partially invalid. This may occur in the form of a combination of
-	// incompatible filters present in an individual route rule that an
-	// implementation wishes to call out. Controllers should only add this
-	// condition with status true when such an error is present.
+	// partially invalid. This may occur when matches or filters in a routing
+	// rule has any of the following errors:
+	//
+	// * an unsupported value;
+	// * an incompatible set of filters;
+	// * an unpermitted reference in an ExtensionRef filter.
+	//
+	// Controllers should only add this condition with status true when
+	// such an error is present.
 	//
 	// Possible reasons for this condition to be true are:
 	//
@@ -348,7 +353,16 @@ type RouteParentStatus struct {
 	// Gateway, and why.
 	//
 	// A Route MUST be considered "Accepted" if at least one of the Route's
-	// rules is implemented by the Gateway.
+	// rules is implemented by the Gateway. An "Accepted" Route may include some
+	// rules that are invalid (e.g. include unsupported configuration values,
+	// unsupported combinations of Filters, etc.) and MUST have the "PartiallyInvalid"
+	// error condition set to "True" to reflect this.
+	//
+	// If no Route rules are implementable by the Gateway, the "Accepted" condition
+	// MUST be set with status "False". The reason accompanying this condition may
+	// vary depending on the source of the configuration invalidity. For example if
+	// caused by an unsupported value the controller MUST set the "Accepted" condition
+	// with status "False" and reason "UnsupportedValue".
 	//
 	// There are a number of cases where the "Accepted" condition may not be set
 	// due to lack of controller visibility, that includes when:
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -292,6 +292,17 @@ const (
 	// This condition indicates whether incompatible filters are present on
 	// a Route. Controllers should only add this condition with status true
 	// when an incompatible combination is present.
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "IncompatibleFilters"
+	//
+	// This condition is an error condition and should not be set with status
+	// false.
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
 	RouteConditionIncompatibleFilters RouteConditionType = "IncompatibleFilters"
 
 	// This reason is used with the "IncompatibleFilters" condition when the
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -298,6 +298,8 @@ const (
 	// Possible reasons for this condition to be true are:
 	//
 	// * "IncompatibleFilters"
+	// * "RefNotPermitted"
+	// * "UnsupportedValue"
 	//
 	// This condition is an error condition and should not be set with status
 	// false.
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -289,9 +289,11 @@ const (
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
 
-	// This condition indicates whether incompatible filters are present on
-	// a Route. Controllers should only add this condition with status true
-	// when an incompatible combination is present.
+	// This condition indicates whether configuration present in a Route is
+	// partially invalid. This may occur in the form of a combination of
+	// incompatible filters present in an individual route rule that an
+	// implementation wishes to call out. Controllers should only add this
+	// condition with status true when such an error is present.
 	//
 	// Possible reasons for this condition to be true are:
 	//
@@ -303,10 +305,11 @@ const (
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
-	RouteConditionIncompatibleFilters RouteConditionType = "IncompatibleFilters"
+	RouteConditionPartiallyInvalid RouteConditionType = "PartiallyInvalid"
 
-	// This reason is used with the "IncompatibleFilters" condition when the
-	// condition is true.
+	// This reason is used with the "PartiallyInvalid" condition when the
+	// there are incompatible filters present on a route rule (for example if
+	// the URLRewrite and RequestRedirect are both present on an HTTPRoute).
 	RouteReasonIncompatibleFilters RouteConditionReason = "IncompatibleFilters"
 )
 
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -288,6 +288,15 @@ const (
 	// This reason is used with the "ResolvedRefs" condition when one of the
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+
+	// This condition indicates whether incompatible filters are present on
+	// a Route. Controllers should only add this condition with status true
+	// when an incompatible combination is present.
+	RouteConditionIncompatibleFilters RouteConditionType = "IncompatibleFilters"
+
+	// This reason is used with the "IncompatibleFilters" condition when the
+	// condition is true.
+	RouteReasonIncompatibleFilters RouteConditionReason = "IncompatibleFilters"
 )
 
 // RouteParentStatus describes the status of a route with respect to an
--- a/site-src/api-types/httproute.md
+++ b/site-src/api-types/httproute.md
@@ -130,8 +130,14 @@ implementation-specific conformance.
 All filters are expected to be compatible with each other except for the
 URLRewrite and RequestRedirect filters, which may not be combined. If an
 implementation can not support other combinations of filters, they must clearly
-document that limitation. In all cases where incompatible or unsupported
-filters are specified, implementations MUST add a warning condition to status.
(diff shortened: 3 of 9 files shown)


apis/v1beta1/shared_types.go
```diff
@@ -275,6 +275,15 @@ const (
...
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+
+	// This condition indicates whether incompatible filters are present on
+	// a Route. Controllers should only add this condition with status true
+	// when an incompatible combination is present.
+	RouteConditionIncompatibleFilters RouteConditionType = "IncompatibleFilters"
+
+	// This reason is used with the "IncompatibleFilters" condition when the
+	// condition is true.
+	RouteReasonIncompatibleFilters RouteConditionReason = "IncompatibleFilters"
```
robscott, line 286: Should this just be another reason for the "Accepted" condition that is defined beside UnsupportedValue? I'm not sure we need a new condition for this, but I could be missing some context.

apis/v1beta1/shared_types.go
```diff
@@ -275,6 +275,15 @@ const (
 	// This reason is used with the "ResolvedRefs" condition when one of the
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+
+	// This condition indicates whether incompatible filters are present on
+	// a Route. Controllers should only add this condition with status true
+	// when an incompatible combination is present.
+	RouteConditionIncompatibleFilters RouteConditionType = "IncompatibleFilters"
```
robscott, line 282: If we decide this does require a new condition, I think we'd need to have the full structure we usually attach to a condition definition:

https://github.com/kubernetes-sigs/gateway-api/blob/5fdab20c7a591ad914b45a2d62e855f167192278/apis/v1beta1/shared_types.go#L193-L212

apis/v1beta1/httproute_types.go
```diff
@@ -198,8 +198,14 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// document that limitation. A Route containing a single rule with incompatible
+	// filters must result in the implementation setting the Accepted Condition for
+	// the Route to `status: False`, with a Reason of `UnsupportedValue`. A Route
```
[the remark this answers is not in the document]
robscott: It could be useful to give an example of what an incompatible filter might look like. Is this referring to unsupported filters, unsupported combinations of filters, or something else entirely?

In any case, the condition and reason we require here should line up with what we decide below. I think I'd personally prefer just adding a new "IncompatibleFilters" reason.

As one last question, are there any combination of filters that we know will be incompatible for everyone? Would combining response header modification and redirects ever be possible and/or desired?

robscott (commented):
Thanks @sunjayBhatia!

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
sunjayBhatia: > Is this referring to unsupported filters, unsupported combinations of filters, or something else entirely

this is mainly in response to the fact that URLRewrite and RequestRedirect filters are mentioned above and in GEP-1364 as incompatible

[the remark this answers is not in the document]
sunjayBhatia: > Would combining response header modification and redirects ever be possible and/or desired?

Yeah I'm not sure, seems possible a controller might have some use for this

Worth adding a conformance test for this combination so we capture feedback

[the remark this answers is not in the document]
mikemorris: I think a reason more like `FilterConflict` may be more clear than the nuance between "Unsupported" and "Incompatible", and has better parity with existing similar reasons like `HostnameConflict` and `ProtocolConflict` https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerConditionType

apis/v1beta1/shared_types.go
mikemorris: Does this perhaps imply adding a `Conflicted` RouteConditionType similar to https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io%2fv1beta1.ListenerConditionType to properly communicate the partial acceptance case?

shaneutt:
@sunjayBhatia just wanted to check in on the status of this one: is there any assistance we can provide to help move this one forward, as it's been sitting for some time?

sunjayBhatia:
> @sunjayBhatia just wanted to check in on the status of this one: is there any assistance we can provide to help move this one forward, as it's been sitting for some time?

just lost track of it, should make some more changes soon 👍🏽

shaneutt:
> just lost track of it, should make some more changes soon 👍🏽 

Ok! Thanks Sunjay 🤜🤛

pleshakov:
would it make sense to also consider how reporting incompatible filters would work when filters are part of an HTTPBackendRef? Will conditions and reasons be the same?

shaneutt:
Just checking in again @sunjayBhatia, do you need any help here to move this one forward?

apis/v1beta1/shared_types.go
sunjayBhatia: This reason is now the only initial option available on the new "PartiallyInvalid" condition

apis/v1beta1/shared_types.go
```diff
@@ -288,6 +288,29 @@ const (
 	// This reason is used with the "ResolvedRefs" condition when one of the
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+
+	// This condition indicates whether configuration present in a Route is
+	// partially invalid. This may occur in the form of a combination of
+	// incompatible filters present in an individual route rule that an
+	// implementation wishes to call out. Controllers should only add this
+	// condition with status true when such an error is present.
```
sunjayBhatia, line 296: I kinda wanted to add something to the effect of "use this when Accepted=true but something is still wrong", lmk if people think that would be useful

apis/v1beta1/httproute_types.go
```diff
@@ -199,7 +199,8 @@ type HTTPRouteRule struct {
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
 	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// filters are specified, implementations MUST add the `PartiallyInvalid` error
+	// condition with status `True` and reason `IncompatibleFilters`.
```
robscott, line 203: What would implementations be expected to do with a partially invalid route? Generally we've had the high level idea of failing fast when we encounter a broken/invalid state. This seems to indicate that implementations would continue to implement the rest of the route without these filters?

Edit: On further reading it's clearer to me that this condition would be used when there was at least one valid and one invalid rule present in the route. It would be helpful to clarify that here and potentially elsewhere in the spec.

apis/v1beta1/shared_types.go
robscott: @sunjayBhatia thanks for the reference to the text of that GEP! It seems like what that's saying is:

1) When this happens and there's still at least one valid Route rule, accept the valid route rules but don't program the invalid one.
2) When this happens and there are no valid Route rules, don't accept the Route at all.

In that case it seems like there would be a subset of reasons that could be added to `Accepted` in 2) or `PartiallyInvalid` in 1). For example the following reasons seem like they would also be equally valid for both conditions (in addition to the proposed `IncompatibleFilters` reason).

* RefNotPermitted
* UnsupportedValue

/cc @youngnick

robscott (commented):
Thanks for the work on this @sunjayBhatia!

apis/v1beta1/shared_types.go
```diff
@@ -288,6 +288,31 @@ const (
...
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "IncompatibleFilters"
```
pleshakov, line 300: do we need to clarify somewhere what the data plane should return in any of those cases?

pleshakov (commented):
Hi @sunjayBhatia  Please see my review.

In addition to the inline comments:

Currently, in `httproute_types.go`, there are a number of comments like below:

> // Unknown values here must result in the implementation setting the
// Accepted Condition for the Route to `status: False`, with a

Considering the new `PartiallyInvalid` condition, would it make sense to update them?

shaneutt:
@sunjayBhatia it appears to me in order to move this one forward there's some comments you'll need to look over. Sorry about how long it's taking, I'm doing a roundup of older PRs to see if we can start shoring some of these up. Let us know how we can best help you in this :vulcan_salute:

sunjayBhatia:
> Hi @sunjayBhatia Please see my review.
> 
> In addition to the inline comments:
> 
> Currently, in `httproute_types.go`, there are a number of comments like below:
> 
> > // Unknown values here must result in the implementation setting the
> > // Accepted Condition for the Route to `status: False`, with a
> > // Reason of `UnsupportedValue`.
> 
> Considering the new `PartiallyInvalid` condition, would it make sense to update them?

Feels like we should find a place to abstract this sort of thing out in the top-level Route documentation or otherwise rather than having to say this on each individual field, similar for the existing `UnsupportedValue` docs

apis/v1beta1/shared_types.go
sunjayBhatia: I've been torn on what to do here, since this may end up being somewhat implementation-specific, depending on the field

sunjayBhatia:
> > Hi @sunjayBhatia Please see my review.
> > In addition to the inline comments:
> > Currently, in `httproute_types.go`, there are a number of comments like below:
> > > // Unknown values here must result in the implementation setting the
> > > // Accepted Condition for the Route to `status: False`, with a
> > > // Reason of `UnsupportedValue`.
> > 
> > 
> > Considering the new `PartiallyInvalid` condition, would it make sense to update them?
> 
> Feels like we should find a place to abstract this sort of thing out in the top-level Route documentation or otherwise rather than having to say this on each individual field, similar for the existing `UnsupportedValue` docs

for reviewers, take a look at https://github.com/kubernetes-sigs/gateway-api/pull/1540/commits/9b37a3e6da23b5dfe07ae7f3becce8f54085d2c5, put the documentation about partial invalidity in the route status section doc

apis/v1beta1/shared_types.go
sunjayBhatia: should be resolved from previous commits, so resolving this thread

apis/v1beta1/shared_types.go
sunjayBhatia: we could recommend a 500, similar to what happens when ResolvedRefs is set due to missing Service/ReferenceGrant (since a lot of these are covered by apiserver validation not sure how we'll write conformance tests etc.)

apis/v1beta1/httproute_types.go
sunjayBhatia: I think this is covered, please reopen if not!

apis/v1beta1/shared_types.go
```diff
@@ -288,6 +288,31 @@ const (
 	// This reason is used with the "ResolvedRefs" condition when one of the
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+
+	// This condition indicates whether configuration present in a Route is
```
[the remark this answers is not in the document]
sunjayBhatia: used this content 👍🏽

sunjayBhatia:
One thing for reviewers/participants on this issue, do we have a document where we state explicitly that implementations "MUST" try to reconcile/program/process a whole resource, regardless of whether they encounter an invalid value etc?

Seems like that is missing from https://gateway-api.sigs.k8s.io/concepts/guidelines/ but could be a worthy addition

apis/v1beta1/shared_types.go
youngnick: Yeah, the exact behavior here depends on the field - for _most_ HTTPRoute things, the field value tells you what to do for an invalid _rule_ (return 500), and what to do if your whole Route is invalid (I think it's reject the whole route, but aren't sure).

youngnick:
I'm happy with this as it stands today, so I'll approve and hold.

/approve
/hold

apis/v1beta1/httproute_types.go
```diff
@@ -205,8 +205,10 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// document that limitation. In cases where incompatible or unsupported
+	// filters are specified on subset of the rules on a route, implementations
+	// MUST add the `PartiallyInvalid` error condition with status `True` and
+	// reason `IncompatibleFilters`.
```
robscott, line 211: This PR feels like it's combining 2 improvements that are related but different:

1. A new `PartiallyInvalid` reason when only a subset or rules in a Route are invalid
2. A new `IncompatibleFilters` reason

It seems like the following reasons could apply to an individual rule or the entire route:

* IncompatibleFilters
* UnsupportedValue (Accepted)
* BackendNotFound (ResolvedRefs)
* InvalidKind (ResolvedRefs)
* RefNotPermitted (ResolvedRefs)

Currently `ResolvedRefs` is only true if 100% of refs are resolved, that's _probably_ fine to leave as is, but it may be confusing for `ResolvedRefs` to be true and `PartiallyInvalid` to be false.

I think the simplest path forward would be to add `IncompatibleFilters` as a possible reason for `Accepted`. Adding a `PartiallyInvalid` condition to reflect that a portion of rules are invalid makes sense, but I think we'd need to make broader changes to the other conditions on the route to have that make sense.

Another option would be to say that both `IncompatibleFilters` and `UnsupportedValue` reasons could be used for `Accepted` and `PartiallyInvalid` conditions. When used with the `Accepted` condition, nothing in the Route would be reconciled, but when used with the `PartiallyInvalid` condition, the valid subset of the Route would be reconciled. We could potentially leave `ResolvedRefs` and the associated reasons as a special case that always applies to the entire route since I think our recommended behavior is just to return 404s in those cases and not actually to stop reconciling anything. I'm less sure about this approach though, so maybe it's better as a follow up.

site-src/api-types/httproute.md
```diff
@@ -131,7 +131,8 @@ All filters are expected to be compatible with each other except for the
 URLRewrite and RequestRedirect filters, which may not be combined. If an
 implementation can not support other combinations of filters, they must clearly
 document that limitation. In all cases where incompatible or unsupported
-filters are specified, implementations MUST add a warning condition to status.
+filters are specified, implementations MUST add the `PartiallyInvalid` error
+condition with status `True` and reason `IncompatibleFilters`.
```
robscott, line 135: Related to my comment above, this guidance doesn't really add up because the `PartiallyInvalid` condition is only intended for cases where a subset of rules are invalid, but here we're saying that it needs to be set _anytime_ there are incompatible filters.

robscott (commented):
Thanks @sunjayBhatia! Sorry I missed this one, added a suggestion for how this could be split up into two smaller pieces.

apis/v1beta1/httproute_types.go
sunjayBhatia: added https://github.com/kubernetes-sigs/gateway-api/pull/2150 to start over and not have to disentagle the changes from this PR

shaneutt:
@sunjayBhatia from what I can tell, there's a couple open comments holding this up and we need a rebase in order to move forward. Let us know how we can be helpful with this one to keep it progressing? It's more than half a year old now and if there are some ways we can help pull this one over the finish line let us know? :thinking:

sunjayBhatia:
https://github.com/kubernetes-sigs/gateway-api/pull/2150 supercedes this so can close this one

sunjayBhatia:
re: this comment: https://github.com/kubernetes-sigs/gateway-api/pull/1540#discussion_r1239173818

PR #1542, opened 2022-11-17
Remove -f to use default value for docker build job fix

wilsonwu:
**What type of PR is this?**
/kind bug

Build job failed, not sure what compatibility problem for this, after changed and tested successful now in local:
image


shaneutt (approved):
Let's give it another try :joy: 

/lgtm

PR #1544, opened 2022-11-17
Rename httproute-invalid-reference-grant to reflect partial validity

### LINKED ISSUE #1143: conformance: add test for partial acceptance of a route with one allowed and one unpermitted backend reference

Update `conformance/tests/httproute-invalid-reference-policy.go` to add assertions that a route containing both a valid and an unpermitted backend reference is accepted by the gateway. This should:

- [x] The route should set an `Accepted {status: true, reason: Accepted}` route status condition
- [x] ~~Ensure the route is counted in the listener's `attachedRoutes`~~ This is a larger issue and is now covered by #1543.
- [x] HTTP requests should successfully reach `app-backend-v1` at path "/"
- [x] Existing behavior of setting a `ResolvedRefs {status: false, reason: RefNotPermitted}` route status condition and checking that traffic does not reach `app-backend-v2` and is rejected with an HTTP 503 response should be preserved
- [ ] Update the test name and description to clarify the intent of checking for partial acceptance

**Why this is needed**:

Refs #1112 and discussion in https://github.com/kubernetes-sigs/gateway-api/pull/1081#pullrequestreview-938455345. It's currently not quite clear from the documentation if partial acceptance of a route with at least one unpermitted backend reference is intended behavior, but it feels like a reasonable implementation to limit the blast radius of changes or configure routing prior to actually deploying an app or allowing traffic. If we can reach consensus among implementations on this approach it should be codified in the conformance tests for clarity, in addition to updating the documentation.

youngnick:
Signed-off-by: Nick Young <nick@isovalent.com>

**What type of PR is this?**
/kind cleanup
/area conformance

Renames the invalid reference grant test to make it clear that it's also checking partial HTTPRoute validity. (It was already doing this, but this makes it clearer.

Fixes #1143

```release-note
NONE
```


conformance/tests/httproute-partially-invalid-via-reference-grant.go
```diff
@@ -29,12 +29,12 @@ import (
...
-	ConformanceTests = append(ConformanceTests, HTTPRouteInvalidReferenceGrant)
+	ConformanceTests = append(ConformanceTests, HTTPRoutePartiallyInvalidViaInvalidReferenceGrant)
 }
 
-var HTTPRouteInvalidReferenceGrant = suite.ConformanceTest{
-	ShortName:   "HTTPRouteInvalidReferenceGrant",
-	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should fail to attach to a Gateway in the same namespace if the route has a backendRef Service in the gateway-conformance-app-backend namespace and a ReferenceGrant exists but does not grant permission to route to that specific Service",
+var HTTPRoutePartiallyInvalidViaInvalidReferenceGrant = suite.ConformanceTest{
+	ShortName:   "HTTPRoutePartiallyInvalidViaInvalidReferenceGrant",
+	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should attach and be partially failed for a Gateway in the same namespace if the route has a backendRef Service in the gateway-conformance-app-backend namespace and a ReferenceGrant exists but does not grant permission to route to that specific Service",
```
robscott, line 37: Tiny nit: "be partially failed" was a bit confusing and I don't think we need it, open to alternatives though.
```suggestion
	Description: "A single HTTPRoute in the gateway-conformance-infra namespace should attach to a Gateway in the same namespace if the route has a backendRef Service in the gateway-conformance-app-backend namespace and a ReferenceGrant exists but does not grant permission to route to that specific Service",
```

robscott (commented):
Thanks @youngnick!

conformance/tests/httproute-partially-invalid-via-reference-grant.go
youngnick: Nah, that makes sense.

robscott:
Thanks @youngnick!

/lgtm

PR #1552, opened 2022-11-21
fix: multi-push ci failed

### LINKED ISSUE #1514: conformance: Test that webhook validations are performed

Xunzhuo:
Signed-off-by: bitliu <bitliu@tencent.com>

**What type of PR is this?**

/kind bug

Fixes #1514

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,7 +7,7 @@ timeout: 1200s
 options:
   substitution_option: ALLOW_LOOSE
 steps:
-  - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90'
+  - name: gcr.io/cloud-builders/docker
     entrypoint: make
     env:
     - DOCKER_CLI_EXPERIMENTAL=enabled
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -7,7 +7,7 @@ timeout: 1200s
 options:
   substitution_option: ALLOW_LOOSE
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/cloud-builders/docker:20.10.14
     entrypoint: make
     env:
     - DOCKER_CLI_EXPERIMENTAL=enabled


Xunzhuo:
CC @wilsonwu, could you have a look and review on this?

I am not 100% sure, because I am get used to use GitHub Actions to build/push multi-arch image, not cloudbuild. This is kind of weird for me : (

cloudbuild.yaml
```diff
@@ -7,7 +7,7 @@ timeout: 1200s
 options:
   substitution_option: ALLOW_LOOSE
 steps:
-  - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20211118-2f2d816b90'
+  - name: gcr.io/cloud-builders/docker
```
robscott, line 10: I think it probably makes sense to switch to this base image since it appears to be update more frequently. With that said, I think we should probably pin to a version. This is the latest release:

```suggestion
  - name: gcr.io/cloud-builders/docker:20.10.14
```

In case we need to fall back to the k8s one, I'd recommend using the latest released version of that (from Oct 7). As far as I can tell, the main reason to use the k8s one is that it includes go, but with the way our Dockerfile is structured, I don't think that helps us:

```
gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20221007-69e0da97ef
```

robscott (commented):
Thanks @Xunzhuo! Just one comment but otherwise LGTM

wilsonwu:
> CC @wilsonwu, could you have a look and review on this?
> 
> I am not 100% sure, because I am get used to use GitHub Actions to build/push multi-arch image, not cloudbuild. This is kind of weird for me : (

I viewed the doc of this image, but I also can't run it in my local env, and from the docs I think we can have a try.

robscott:
Thanks!

/lgtm
/approve

Issue #1496, opened 2022-10-31
Conformance RoundTripper should allow for custom http.RoundTripper implementation


briantkennedy:
**What would you like to be added**:

Allow for optionally constructing the `roundtripper.DefaultRoundTripper` with a custom `http.RoundTripper`.

**Why this is needed**:

Users may need to run conformance tests through a SSH tunnel or Proxy where a custom `http.RoundTripper` is required to connect to the load balancer.

https://github.com/kubernetes-sigs/gateway-api/blob/0adb8a8bd14672db6ea16ab6191a828295fe8d04/conformance/utils/roundtripper/roundtripper.go#L70

LiorLieberman:
curios if there is any progress with it?

akash-gautam:
@LiorLieberman Not yet, this one slipped out of my radar unfortunately. Will start on this soon.

ashutosh887:
@akash-gautam Should I give it a try now?
Please assign me @LiorLieberman

robscott:
Hey @ashutosh887, I think you're welcome to take this one on unless @akash-gautam is actively working on this. Thanks for volunteering!

ashutosh887:
Thanks @robscott 
I'll get started with it
Please guide me a bit what is needed to be fixed

shaneutt:
Just keep in mind @ashutosh887 that you've assigned multiple issues to yourself within a short period of time: it's often best to take one or two and focus in on those according to priority (priority labels, plus milestones). If you want to focus in more on #1515 or #1639 it would be OK to leave this one unassigned for now and come back to it OR open it up for other contributors to grab in the meantime (as this one is considered more of a long-term maintenance gain than a milestone deliverable).

As for your question: this issue is requesting that a `RoundTripper` option be made available in https://github.com/kubernetes-sigs/gateway-api/blob/0adb8a8bd14672db6ea16ab6191a828295fe8d04/conformance/utils/roundtripper/roundtripper.go#L70 to override the defaults, which would enable running the test suite when a SSH tunnel or proxy is in place. If you do not have an implementation to test changes to this on, OR more specifically an implementation that would need to utilize an SSH tunnel or proxy, it may be better to leave this one for someone with an implementation with that need who can actively test it.

ashutosh887:
Thanks @shaneutt 
Sure Sir... I respect your suggestions.

akash-gautam:
Starting work on this one today. 
/assign

[...]

PR #1556, opened 2022-11-21
added redirect tests for core redirect features

LiorLieberman:
**What type of PR is this?**
/kind feature
/area conformance

Add tests for core redirect features 

Issue #1103 

NONE

### DIFF
--- a/conformance/tests/httproute-request-redirect.go
+++ b/conformance/tests/httproute-request-redirect.go
@@ -43,8 +43,8 @@ var HTTPRouteRequestRedirect = suite.ConformanceTest{
 
 		testCases := []http.ExpectedResponse{{
 			Request: http.Request{
-				Path:             "/hostname",
-				UnFollowRedirect: true,
+				Path:             "/hostname-redirect",
+				UnfollowRedirect: true,
 			},
 			Response: http.Response{
 				StatusCode: 302,
@@ -57,7 +57,7 @@ var HTTPRouteRequestRedirect = suite.ConformanceTest{
 		}, {
 			Request: http.Request{
 				Path:             "/status-code-301",
-				UnFollowRedirect: true,
+				UnfollowRedirect: true,
 			},
 			Response: http.Response{
 				StatusCode: 301,
@@ -67,7 +67,7 @@ var HTTPRouteRequestRedirect = suite.ConformanceTest{
 		}, {
 			Request: http.Request{
 				Path:             "/host-and-status",
-				UnFollowRedirect: true,
+				UnfollowRedirect: true,
 			},
 			Response: http.Response{
 				StatusCode: 301,
--- a/conformance/tests/httproute-request-redirect.go
+++ b/conformance/tests/httproute-request-redirect.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRequestRedirect)
+}
+
+var HTTPRouteRequestRedirect = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRequestRedirect",
+	Description: "An HTTPRoute with hostname and statusCode redirect filters",
+	Manifests:   []string{"tests/httproute-request-redirect.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+		routeNN := types.NamespacedName{Name: "request-redirect", Namespace: ns}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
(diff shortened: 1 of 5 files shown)


LiorLieberman:
This is a draft working implementation to check core features for redirect.
@robscott I am not a fan of my own implementation 😅 for reasons we discussed offline. 
mainly around the fact that that we initilaize the http client inside `CaptureRoundTrip()` and custom roundtripper implementations would have to customize their logic in order to be able to test conformance for redirects.

As per #1496 I think that we should add a roundTripper under `DefaultRoundTripper` to implementations to override it.

Perhaps we should continue with this lean implementation and amend accordingly once #1496 is finished?

a few more nits; 
I do not like adding `RedirectRequest` type and `IsRedirect()` func under the roundtripper package but I did have an import cycle so had to do it. strong thoughts on it ?

conformance/utils/http/http.go
```diff
@@ -57,10 +59,11 @@ type ExpectedResponse struct {
...
 type Request struct {
-	Host    string
-	Method  string
-	Path    string
-	Headers map[string]string
+	Host             string
+	Method           string
+	Path             string
+	Headers          map[string]string
+	UnFollowRedirect bool
```
robscott, line 66: Tiny nit
```suggestion
	UnfollowRedirect bool
```

conformance/tests/httproute-request-redirect.yaml
```diff
@@ -0,0 +1,40 @@
...
+        type: PathPrefix
+        value: /host-and-status
+    filters:
+    - type: RequestRedirect
+      requestRedirect:
+        statusCode: 301
+        hostname: example.org
+    backendRefs:
+    - name: infra-backend-v1
+      port: 8080
```
robscott, line 40: Tiny nit: Don't forget newline

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -153,3 +179,14 @@ func formatDump(data []byte, prefix string) string {
 	data = startLineRegex.ReplaceAllLiteral(data, []byte(prefix))
 	return string(data)
 }
+
+func IsRedirect(statusCode int) bool {
+	// Gateway allows only 301, and 302
```
robscott, line 184: Even though we only allow 301 and 302, I'd make this logic more future-proof to include the full set of redirect codes. There's a chance we'll add more in the future, but even if not, it will be more helpful to show a more useful error message than if we didn't capture the redirect info.

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -57,12 +58,19 @@ type CapturedRequest struct {
 	Pod       string `json:"pod"`
 }
 
+type RedirectRequest struct {
```
robscott, line 61: Nit: Should have some kind of comment here for the type.

conformance/tests/httproute-request-redirect.yaml
```diff
@@ -0,0 +1,40 @@
...
+  name: request-redirect
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /hostname
```
robscott, line 13: For some reason "hostname" as the path matcher feels a bit confusing to me, maybe a slight tweak would make it clearer.
```suggestion
        value: /hostname-redirect
```

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -153,3 +179,14 @@ func formatDump(data []byte, prefix string) string {
 	data = startLineRegex.ReplaceAllLiteral(data, []byte(prefix))
 	return string(data)
 }
+
+func IsRedirect(statusCode int) bool {
```
robscott, line 183: Comment would be good for this func

robscott (commented):
Thanks @LiorLieberman, this LGTM, just a few minor nits.

conformance/utils/roundtripper/roundtripper.go
LiorLieberman: Can you elaborate? my thought was to put it in a function so we will have a one central place for adding more codes in the future. I think now we want to fail conformance if someone returns something other than 301/302 because we do not support it, correct?

conformance/tests/httproute-request-redirect.go
```diff
@@ -0,0 +1,92 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRequestRedirect)
+}
+
+var HTTPRouteRequestRedirect = suite.ConformanceTest{
```
LiorLieberman, line 34: @robscott Also what do you think about the naming, both here and the file name, considering we want to add two more tests for extended and experimental features

conformance/utils/roundtripper/roundtripper.go
robscott: Let's say an implementation is actually returning a 307 or 308. With the current code, we wouldn't actually populate the rest of the redirect info for the rest of the test logic, so it would just look like _everything_ failed instead of just having the wrong status code set. I think the `IsRedirect` function is primarily a check on if we should populate the new redirect struct. The roundtripper itself should not care about what is conformant or not, just in making sure it populates as much information as possible so the tests can determine if the response was conformant.

conformance/tests/httproute-request-redirect.go
robscott: Good question - I think it's probably simplest to start with this and then rename in follow up PRs.

conformance/utils/roundtripper/roundtripper.go
LiorLieberman: Thats a good point, however we do check for wrong status code https://github.com/kubernetes-sigs/gateway-api/blob/c1c4952732d0acbcf294ce93c511ff86a6380c4d/conformance/utils/http/http.go#L194-L196

I agree the roundtripper should not care about conformance, would you prefer to construct the redirect struct always? Whats the value you think it'll give us? it wont change the behavior of the above lines - we will have a failure with a wrong status code still.

robscott: I think we should always construct the redirect struct for any 3xx status code. Although the test will still fail, it will likely provide more useful information to the person debugging the test. Instead of simply knowing the code is wrong, they'll also be able to see if the location they are redirecting to has any issues.

LiorLieberman: Makes sense, I have added all the codes but if you prefer I will do something like this instead let me know
```golang
func IsRedirect(statusCode int) bool {
	switch {
	case statusCode >= 300 && statusCode < 400:
		return true
	}
	return false
}
```

shaneutt (approved):
+1 to Rob's comments but they appear to be resolved already.

/lgtm

robscott:
🤔 looks like @k8s-ci-robot got stuck on the LGTM, adding another.

/lgtm

PR #1560, opened 2022-11-29
Update GEP-1364 to Standard implementation

### LINKED ISSUE #1364: GEP: Conditions and Status update

youngnick:
**What type of PR is this?**
/kind documentation
/kind gep

Updates GEP-1364 (#1364) to implemented in the Standard channel.

Fixes #1364.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -75,7 +75,6 @@ nav:
         - geps/gep-1324.md
         - geps/gep-1282.md
       - Implementable:
-        - geps/gep-1364.md
       - Experimental:
         - geps/gep-1323.md
         - geps/gep-1016.md
@@ -83,6 +82,7 @@ nav:
         - geps/gep-726.md
         - geps/gep-713.md
       - Standard:
+        - geps/gep-1364.md
         - geps/gep-922.md
         - geps/gep-917.md
         - geps/gep-851.md
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -75,6 +75,7 @@ nav:
         - geps/gep-1324.md
         - geps/gep-1282.md
       - Implementable:
+        - Nothing
       - Experimental:
         - geps/gep-1323.md
         - geps/gep-1016.md
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -75,7 +75,7 @@ nav:
         - geps/gep-1324.md
         - geps/gep-1282.md
       - Implementable:
-        - Nothing
+        - DummyEntry
       - Experimental:
         - geps/gep-1323.md
         - geps/gep-1016.md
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -75,14 +75,15 @@ nav:
         - geps/gep-1324.md
         - geps/gep-1282.md
       - Implementable:
-        - geps/gep-1364.md
+        - DummyEntry
```
robscott, line 78: I think this was a temporary change for the sake of CI?
```suggestion
        - DummyEntry
```

youngnick: no, `mkdocs` will permanently crash if the object is empty. Actually, let me see if I can add the empty list instead...

robscott: Maybe just remove "Implementable" altogether until we have a GEP that moves to that status?

robscott:
Thanks!

/lgtm

PR #1565, opened 2022-11-29
Policy attachment update

youngnick:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind gep

This updates the Policy Attachment documentation as per the discussion in https://docs.google.com/document/d/1p_kmAOmqebX8zyHnXW5I58cD8ZhNX4QNSa_gw8uAYxw/edit?usp=sharing

There's an update to the original GEP, and to the site documentation.

### DIFF
--- a/geps/gep-713.md
+++ b/geps/gep-713.md
@@ -1,29 +1,49 @@
-# GEP-713: Policy Attachment
+# GEP-713: Metaresources and Policy Attachment_metaresource_
 
 * Issue: [#713](https://github.com/kubernetes-sigs/gateway-api/issues/713)
 * Status: Experimental
 
 ## TLDR
 
-This GEP aims to standardize policy attachment to resources associated with
-Gateway API by establishing a pattern which defines how `Policy` API types can
-have their relevant effects applied to network traffic. Individual policy APIs
-(e.g. `TimeoutPolicy`, `RetryPolicy`, etc) will include a common `TargetRef`
-field in their specification to identify how and where to apply that policy.
-This will be important for providing a consistent experience across
-implementations of the API, even for configuration details that may not be fully
-portable.
+This GEP aims to standardize terminology and processes around using one Kubernetes
+object to modify the functions of one or more other objects.
+
+This GEP defines some terms, firstly: _Metaresource_.
+
+A Kubernetes object that that _augments_ the behavior of an object
+in a standard way is called a _Metaresource_.
+
+This document proposes controlling the creation of configuration in the underlying
+Gateway data plane using two types of Policy Attachment.
+A "Policy Attachment" is a specific type of _metaresource_ that can affect specific
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Inherited Policy Attachment").
+
+Individual policy APIs:
+- must be their own CRDs (e.g. `TimeoutPolicy`, `RetryPolicy` etc),
+- can be included in the Gateway API API group and installation or be defined by
+  implementations
+- and must include a common `TargetRef` struct in their specification to identify
+  how and where to apply that policy.
+- _may_ include either a `defaults` section, an `overrides` section, or both. If
+  these are included, the Policy is an Inherited Policy, and should use the
+  inheritance rules defined in this document.
+
+For Inherited Policies, this GEP also describes a set of expected behaviors
+for how settings can flow across a defined hierarchy.
+
 
 ## Goals
 
-* Establish a pattern for policy attachment which will be used for any policies
+* Establish a pattern for Policy resources which will be used for any policies
   included in the Gateway API spec
-* Establish a pattern for policy attachment which should be used for any
-  implementation specific policies used with Gateway API resources
+* Establish a pattern for Policy attachment, whether Direct or Inherited,
+  which must be used for any implementation specific policies used with
+  Gateway API resources
 * Provide a way to distinguish between required and default values for all
   policy API implementations
 * Enable policy attachment at all relevant scopes, including Gateways, Routes,
-  Backends
+  Backends, along with how values should flow across a hierarchy if necessary
 * Ensure the policy attachment specification is generic and forward thinking
   enough that it could be easily adapted to other grouping mechanisms like
   Namespaces in the future
@@ -38,24 +58,210 @@ portable.
 * Define all potential policies that may be attached to resources
 * Design the full structure and configuration of policies
 
+## Background and concepts
+
+When designing Gateway API, one of the things we’ve found is that we often need to be
+able change the behavior of objects without being able to make changes to the spec
+of those objects. Sometimes, this is because we can’t change the spec of the object
+to hold the information we need ( ReferenceGrant, from
+[GEP-709](https://gateway-api.sigs.k8s.io/geps/gep-709/), affecting Secrets
+and Services is an example, as is Direct Policy Attachment), and sometimes it’s
+because we want the behavior change to flow across multiple objects
+(this is what Inherited Policy Attachment is for).
+
+To put this another way, sometimes we need ways to be able to affect how an object
+is interpreted in the API, without representing the description of those effects
+inside the spec of the object.
+
+This document describes the ways we design objects to meet these two use cases,
+and why you might choose one or the other.
+
+We use the term “metaresource” to describe the class of objects that _only_ augment
+the behavior of another Kubernetes object, regardless of what they are targeting.
+
+“Meta” here is used in its Greek sense of “more comprehensive”
+or “transcending”, and “resource” rather than “object” because “metaresource”
+is more pronounceable than “metaobject”. Additionally, a single word is better
+than a phrase like “wrapper object” or “wrapper resource” overall, although both
+of those terms are effectively synonymous with “metaresource”.
+
+A "Policy Attachment" is a metaresource that affects the fields in existing objects
+(like Gateway or Routes), or influences the configuration that's generated in an
+underlying data plane.
+
+"Direct Policy Attachment" is when a Policy object references a single object _only_,
+and only modifies the fields of or the configuration associated with that object.
+
+"Inherited Policy Attachment" is when a Policy object references a single object
+_and any child objects of that object_ (according to some defined hierarchy), and
+modifies fields of the child objects, or configuration associated with the child
+objects.
+
+In either case, a Policy may either affect an object by controlling the value
+of one of the existing _fields_ in the `spec` of an object, or it may add
+additional fields that are _not_ in the `spec` of the object.
+
+### Direct Policy Attachment
+
+A Direct Policy Attachment is tightly bound to one instance of a particular
+Kind within a single namespace (or to an instance of a single Kind at cluster scope),
+and only modifies the behavior of the object that matches its binding.
+
+As an example, one use case that Gateway API currently does not support is how
+to configure details of the TLS required to connect to a backend (in other words,
+if the process running inside the backend workload expects TLS, not that some
+automated infrastructure layer is provisioning TLS as in the Mesh case).
+
+A hypothetical TLSConnectionPolicy that targets a Service could be used for this,
+using the functionality of the Service as describing a set of endpoints. (It
+should also be noted this is not the only way to solve this problem, just an
+example to illustrate Direct Policy Attachment.)
+
+The TLSConnectionPolicy would look something like this:
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: TLSConnectionPolicy
+metadata:
+  name: tlsport8443
+  namespace: foo
+spec:
+  targetRef: # This struct is defined as part of Gateway API
+    group: "" # Empty string means core - this is a standard convention
+    kind: Service
+    name: fooService
+  tls:
+    certificateAuthorityRefs:
+      - name: CAcert
+    port: 8443
+
+```
+
+All this does is tell an implementation, that for connecting to port `8443` on the
+Service `fooService`, it should assume that the connection is TLS, and expect the
+service's certificate to be validated by the chain in the `CAcert` Secret.
+
+Importantly, this would apply to _every_ usage of that Service across any HTTPRoutes
+in that namespace, which could be useful for a Service that is reused in a lot of
+HTTPRoutes.
+
+With these two examples in mind, here are some guidelines for when to consider
+using Direct Policy Attachment:
+
+* The number or scope of objects to be modified is limited or singular. Direct
+  Policy Attachments must target one specific object.
+* The modifications to be made to the objects don’t have any transitive information -
+  that is, the modifications only affect the single object that the targeted
+  metaresource is bound to, and don’t have ramifications that flow beyond that
+  object.
+* In terms of status, it should be reasonably easy for a user to understand that
+  everything is working - basically, as long as the targeted object exists, and
+  the modifications are valid, the metaresource is valid, and this should be
+  straightforward to communicate in one or two Conditions. Note that at the time
+  of writing, this is *not* completed.
+* Direct Policy Attachment _should_ only be used to target objects in the same
+  namespace as the Policy object. Allowing cross-namespace references brings in
+  significant security concerns, and/or difficulties about merging cross-namespace
+  policy objects. Notably, Mesh use cases may need to do something like this for
+  consumer policies, but in general, Policy objects that modify the behavior of
+  things outside their own namespace should be avoided unless it uses a handshake
+  of some sort, where the things outside the namespace can opt–out of the behavior.
+  (Notably, this is the design that we used for ReferenceGrant).
+
+### Inherited Policy Attachment: It's all about the defaults and overrides
+
+Because a Inherited Policy is a metaresource, it targets some other resource
+and _augments_ its behavior.
+
+But why have this distinct from other types of metaresource? Because Inherited
+Policy resources are designed to have a way for settings to flow down a hierarchy.
+
+Defaults set the default value for something, and can be overridden by the
+“lower” objects (like a connection timeout default policy on a Gateway being
+overridable inside a HTTPRoute), and Overrides cannot be overridden by “lower”
+objects (like setting a maximum client timeout to some non-infinite value at the
+Gateway level to stop HTTPRoute owners from leaking connections over time).
+
+Here are some guidelines for when to consider using a Inherited Policy object:
+
+* The settings or configuration are bound to one containing object, but affect
+  other objects attached to that one (for example, affecting HTTPRoutes attached
+  to a single Gateway, or all HTTPRoutes in a GatewayClass).
+* The settings need to able to be defaulted, but can be overridden on a per-object
+  basis.
+* The settings must be enforced by one persona, and not modifiable or removable
+  by a lesser-privileged persona. (The owner of a GatewayClass may want to restrict
+  something about all Gateways in a GatewayClass, regardless of who owns the Gateway,
+  or a Gateway owner may want to enforce some setting across all attached HTTPRoutes).
+* In terms of status, a good accounting for how to record that the Policy is
+  attached is easy, but recording what resources the Policy is being applied to
+  is not, and needs to be carefully designed to avoid fanout apiserver load.
+  (This is not built at all in the current design either).
+
+When multiple Inherited Policies are used, they can interact in various ways,
+which are governed by the following rules, which will be expanded on later in
+in this document.
+
+* If a Policy does not affect an object's fields directly, then the resultant
+  Policy should be the set of all distinct fields inside the relevant Policy objects,
+  as set out by the rules below.
+* For Policies that affect an object's existing fields, multiple instances of the
+  same Policy Kind affecting an object's fields will be evaluated as
+  though only a single Policy "wins" the right to affect each field. This operation
+  is performed on a _per-distinct-field_ basis.
+* Settings in `overrides` stanzas will win over the same setting in a `defaults`
+  stanza.
+* `overrides` settings operate in a "less specific beats more specific" fashion -
+  Policies attached _higher_ up the hierarchy will beat the same type of Policy
+  attached further down the hierarchy.
+* `defaults` settings operate in a "more specific beats less specific" fashion -
+  Policies attached _lower down_ the hierarchy will beat the same type of Policy
+  attached further _up_ the hierarchy.
+* For `defaults`, the _most specific_ value is the one _inside the object_ that
+  the Policy applies to; that is, if a Policy specifies a `default`, and an object
+  specifies a value, the _object's_ value will win.
+* Policies interact with the fields they are controlling in a "replace value"
+  fashion.
+  * For fields where the `value` is a scalar, (like a string or a number)
+    should have their value _replaced_ by the value in the Policy if it wins.
+    Notably, this means that a `default` will only ever replace an empty or unset
+    value in an object.
+  * For fields where the value is an object, the Policy should include the fields
+    in the object in its definition, so that the replacement can be on simple fields
+    rather than complex ones.
+  * For fields where the final value is non-scalar, but is not an _object_ with
+    fields of its own, the value should be entirely replaced, _not_ merged. This
+    means that lists of strings or lists of ints specified in a Policy will overwrite
+    the empty list (in the case of a `default`) or any specified list (in the case
+    of an `override`). The same applies to `map[string]string` fields. An example
+    here would be a field that stores a map of annotations - specifying a Policy
+    that overrides annotations will mean that a final object specifying those
+    annotations will have its value _entirely replaced_ by an `override` setting.
+* In the case that two Policies of the same type specify different fields, then
+  _all_ of the specified fields should take effect on the affected object.
+
+Examples to further illustrate these rules are given below.
+
 ## API
 
 This approach is building on concepts from all of the alternatives discussed
-below. This is very similar to the existing BackendPolicy resource in the API,
+below. This is very similar to the (now removed) BackendPolicy resource in the API,
 but also borrows some concepts from the [ServicePolicy
 proposal](https://github.com/kubernetes-sigs/gateway-api/issues/611).
 
 ### Policy Attachment for Ingress
-Attaching policy to Gateway resources for ingress use cases is relatively
-straightforward. A policy can reference the resource it wants to apply to.
+Attaching a Directly Attached Policy to Gateway resources for ingress use cases
+is relatively straightforward. A policy can reference the resource it wants to
+apply to.
+
 Access is granted with RBAC - anyone that has access to create a RetryPolicy in
 a given namespace can attach it to any resource within that namespace.
 
 ![Simple Ingress Example](images/713-ingress-simple.png)
 
-To build on that example, it’s possible to attach policies to more resources.
-Each policy applies to the referenced resource and everything below it in terms
-of hierarchy. Although this example is likely more complex than many real world
+An Inherited Policy can attach to a parent resource, and then each policy
+applies to the referenced resource and everything below it in terms of hierarchy.
+Although this example is likely more complex than many real world
 use cases, it helps demonstrate how policy attachment can work across
 namespaces.
 
@@ -204,7 +410,7 @@ precedence over Routes and Services below it. On the other hand, an app owner
 may want to set a default timeout for their Service. That would have precedence
 over defaults attached at higher levels such as Route or Gateway.
 
-If using defaults and overrides, each policy resource MUST include 2 structs
+If using defaults _and_ overrides, each policy resource MUST include 2 structs
 within the spec. One with override values and the other with default values.
 
 In the following example, the policy attached to the Gateway requires cdn to
@@ -212,7 +418,7 @@ be enabled and provides some default configuration for that. The policy attached
 to the Route changes the value for one of those fields (includeQueryString).
 
 ```yaml
-kind: GKEServicePolicy # Example of implementation specific policy name
+kind: CDNCachingPolicy # Example of implementation specific policy name
 spec:
   override:
     cdn:
@@ -227,13 +433,14 @@ spec:
     kind: Gateway
     name: example
 ---
-kind: GKEServicePolicy
+kind: CDNCachingPolicy
 spec:
   default:
     cdn:
       cachePolicy:
         includeQueryString: false
   targetRef:
+    type: direct
     kind: HTTPRoute
     name: example
 ```
@@ -243,7 +450,10 @@ precedence over the default drainTimeout value attached to the Route. At the
 same time, we can see that the default connectionTimeout attached to the Route
 has precedence over the default attached to the Gateway.
 
-![Hierarchical Policy Example](images/713-policy-hierarchy.png)
+Also note how the different resources interact - fields that are not common across
+objects _may_ both end up affecting the final object.
+
+![Inherited Policy Example](images/713-policy-hierarchy.png)
 
 #### Supported Resources
 It is important to note that not every implementation will be able to support
@@ -266,7 +476,7 @@ used to set defaults and requirements for an entire GatewayClass.
 ### Targeting External Services
 In some cases (likely limited to mesh) we may want to apply policies to requests
 to external services. To accomplish this, implementations can choose to support
-a refernce to a virtual resource type:
+a reference to a virtual resource type:
 
 ```yaml
 apiVersion: networking.acme.io/v1alpha1
@@ -282,13 +492,59 @@ spec:
     name: foo.com
 ```
 
+### Merging into existing `spec` fields
+
+It's possible (even likely) that configuration in a Policy may need to be merged
+into an existing object's fields somehow, particularly for Inherited policies.
+
+When merging into an existing fields inside an object, Policy objects should
+merge values at a scalar level, not at a struct or object level.
+
+For example, in the `CDNCachingPolicy` example above, the `cdn` struct contains
+a `cachePolicy` struct that contains fields. If an implementation was merging
+this configuration into an existing object that contained the same fields, it
+should merge the fields at a scalar level, with the `includeHost`,
+`includeProtocol`, and `includeQueryString` values being defaulted if they were
+not specified in the object being controlled. Similarly, for `overrides`, the
+values of the innermost scalar fields should overwrite the scalar fields in the
+affected object.
+
+Implementations should not copy any structs from the Policy object directly into the
+affected object, any fields that _are_ overridden should be overridden on a per-field
+basis.
+
+In the case that the field in the Policy affects a struct that is a member of a list,
+each existing item in the list in the affected object should have each of its
+fields compared to the corresponding fields in the Policy.
+
+For non-scalar field _values_, like a list of strings, or a `map[string]string`
+value, the _entire value_ must be overwritten by the value from the Policy. No
+merging should take place. This mainly applies to `overrides`, since for
+`defaults`, there should be no value present in a field on the final object.
+
+This table shows how this works for various types:
+
+|Type|Object config|Override Policy config|Result|
+|----|-------------|----------------------|------|
+|string| `key: "foo"` | `key: "bar"`  | `key: "bar"` |
+|list| `key: ["a","b"]` | `key: ["c","d"]` | `key: ["c","d"]` |
+|`map[string]string`| `key: {"foo": "a", "bar": "b"}` | `key: {"foo": "c", "bar": "d"}` | `key: {"foo": "c", "bar": "d"}` |
+
+
 ### Conflict Resolution
-It is possible for multiple policies to target the same resource. When this
-happens, merging is the preferred outcome. If multiple policy resources target
+It is possible for multiple policies to target the same object _and_ the same
+fields inside that object. If multiple policy resources target
 the same resource _and_ have an identical field specified with different values,
 precedence MUST be determined in order of the following criteria, continuing on
 ties:
 
+* Direct Policies override Inherited Policies. If preventing settings from
+  being overwritten is important, implementations should only use Inherited
+  Policies, and the `override` stanza that implies. Note also that it's not
+  intended that Direct and Inherited Policies should overlap, so this should
+  only come up in exceptional circumstances.
+* Inside Inherited Policies, the same setting in `overrides` beats the one in
+  `defaults`.
 * The oldest Policy based on creation timestamp. For example, a Policy with a
   creation timestamp of "2021-07-15 01:02:03" is given precedence over a Policy
   with a creation timestamp of "2021-07-15 01:02:04".
@@ -308,28 +564,37 @@ the API structure defined above and add a `gateway.networking.k8s.io/policy:
 true` label to the CRD.
 
 ### Status
-In the future, we may consider adding a new `Policies` field to status on
-Gateways and Routes. This would be a list of `PolicyTargetReference` structs
-with the fields instead used to refer to the Policy resource that has been
-applied.
 
-Unfortunately, this may create more confusion than it is worth, here are some of
-the key concerns:
+In the current iteration of this GEP, metaresources and Policy objects don't
+have any standard way to record what they're attaching to, or applying settings
+to in the case of Policy Attachment. There are some recommended Condition types
+defined below, but further work on the status design is required to ensure that
+some problems are resolved:
 
 * When multiple controllers are implementing the same Route and recognize a
-  policy, it would be difficult to determine which controller should be
-  responsible for adding that policy reference to status.
-* For this to be somewhat scalable, we'd need to limit the status entries to
-  policies that had been directly applied to the resource. This could get
-  confusing as it would not provide any insight into policies attached above or
-  below.
+  policy, it must be possible to determine which controller was
+  responsible for adding that policy reference to status. Adding Conditions to
+  status on the Policy instead can be helpful here, but we're still lacking a way
+  for the Route or Gateway owner to find _all_ the Policies that are influencing
+  their object.
+* For this to be somewhat scalable, we must limit the number of status updates
+  that can result from a metaresource update.
 * Since we only control some of the resources a policy might be attached to,
-  adding policies to status would only be possible on Gateway API resources, not
-  Services or other kinds of backends.
+  adding policies to status would only be possible on the policy objects themselves
+  or on Gateway API resources, not Services or other kinds of backends.
 
-Although these concerns are not unsolvable, they lead to the conclusion that
-a Kubectl plugin should be our primary approach to providing visibility here,
-with a possibility of adding policies to status at a later point.
+Previous experience in the Kubernetes API has made it clear that having a single
+object that can cause status updates to occur across many other objects can have
+a big performance impact, so the status design must be very carefully done to
+avoid these kind of fanout problems.
+
+However, the whole purpose of having a standardized Policy API structure and
+patterns is intended to make this problem solvable both for human users and with
+tooling.
+
+This is currently a _very_ open question. A discussion is ongoing at
+[#1531](https://github.com/kubernetes-sigs/gateway-api/discussions/1531), and this
+GEP will be updated with any outcomes.
 
 ### Conditions
 Controllers using the Gateway API policy attachment model SHOULD populate the 
@@ -396,14 +661,14 @@ controller implementation:
 
 2. Although it's possible that arbitrary fields could be supported by custom
    policy, custom route filters, and core/extended fields concurrently, it is
-   strongly recommended that implementations not use multiple mechanisms for
-   representing the same fields. A given field should only be supported through
-   a single extension method. An example of potential conflict is policy
-   precedence and structured hierarchy, which only applies to custom policies.
-   Allowing a field to exist in custom policies and also other areas of the API,
-   which are not part of the structured hierarchy, breaks the precedence model.
-   Note that this guidance may change in the future as we gain a better
-   understanding for extension mechanisms of the Gateway API can interoperate.
+   recommended that implementations only use multiple mechanisms for
+   representing the same fields when those fields really _need_ the defaulting
+   and/or overriding behavior that Policy Attachment provides. For example, a
+   custom filter that allowed the configuration of Authentication inside a
+   HTTPRoute object might also have an associated Policy resource that allowed
+   the filter's settings to be defaulted or overridden. It should be noted that
+   doing this in the absence of a solution to the status problem is likely to
+   be *very* difficult to troubleshoot.
 
 ### Conformance Level
 This policy attachment pattern is associated with an "EXTENDED" conformance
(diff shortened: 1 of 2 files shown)


site-src/geps/gep-713.md
```diff
@@ -1,25 +1,35 @@
...
-(e.g. `TimeoutPolicy`, `RetryPolicy`, etc) will include a common `TargetRef`
-field in their specification to identify how and where to apply that policy.
+This GEP aims to standardize attachment of one Kubernetes object to another in
+Gateway API by establishing a pattern which defines how metaresources such as
+`Policy` API types can have their relevant effects applied to network traffic.
+Metaresources are a type of object that _augments_ the behavior of an object
+in a standard way.
+
+"Policy Attachment" is a specific type of metaresource that controls how policy
+can flow _across_ multiple objects in a hierarchy.
```
shaneutt, line 15: I think the definition of "metaresource" above is pretty solid, but I'm thinking there might be some way to be more descriptive here... I know it's the TLDR, but maybe there's some way we can sneak in a brief example of what we mean? Just a thought.

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
 * Define all potential policies that may be attached to resources
 * Design the full structure and configuration of policies
 
+## Background and concepts
+
+One of the things we’ve found when making Gateway API is that we often need to be
+able to represent information _about_ objects _outside of_ those objects. That is,
+sometimes we need ways to be able to affect how an object is interpreted in the
+API, without representing the description of those effects inside the spec of the
+object.
```
shaneutt, line 57: :+1:

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
...
+
+Sometimes, this is because we can’t change the spec of the object to hold the
+information we need ( ReferenceGrant, from GEP-709, is an example), and sometimes
+it’s because we want the information we need to flow across multiple objects
+(this is what Policy Attachment is for).
```
shaneutt, line 62: This might be better as the opening statement, e.g. "Sometimes we can change spec ... e.t.c." and then lead into "and so sometimes we need to affect how an object is interpreted without changing the spec"?

youngnick: I changed the wording of these first two paragraphs around a little. PTAL.

shaneutt (approved):
Overall, especially considering we still consider this experimental, I'm :+1: for this in its current state. Would like to have Rob take a look too.

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
...
+  object. ReferenceGrant and BackendCapabilities are good examples here because
+  they don’t expect the metaresource to modify the behavior of more than one
+  object. (Even in the ReferenceGrant case, the ability to set a Kind rather
+  than an individual object is syntactic sugar for having to prevent needing to
+  have lots of ReferenceGrants if you have lots of wrapped objects).
+* In terms of status, it should be reasonably easy for a user to understand that
+  everything is working - basically, as long as the targeted object exists, and
+  the modifications are valid, the metaresource is valid, and this should be
+  straightforward to communicate in one or two Conditions. Note that at the time
+  of writing, this is *not* completed.
```
dprotaso, line 112: > straightforward to communicate in one or two Conditions.

On the policy or the targeted object?

How would you also signal that the policy is still being actively worked on? In other resources we have `generation/observedGeneration` pairing.

youngnick: This part is still TBD, and needs to be finished, yes.

site-src/geps/gep-713.md
```diff
@@ -1,25 +1,35 @@
-# GEP-713: Policy Attachment
+# GEP-713: Metaresources and Policy Attachment
```
robscott, line 1: I'm a little torn on this, but I'm wondering if we should have a new GEP to describe this new form of policy attachment. That would allow either the hierarchical or direct versions to graduate independently if necessary. In any case we should have unified docs, and if the base API(s) are nearly identical, they probably do belong in the same GEP, just figured I'd least mention the alternative.

site-src/geps/gep-713.md
```diff
@@ -1,25 +1,35 @@
...
 * Status: Experimental
 
 ## TLDR
 
-This GEP aims to standardize policy attachment to resources associated with
-Gateway API by establishing a pattern which defines how `Policy` API types can
-have their relevant effects applied to network traffic. Individual policy APIs
-(e.g. `TimeoutPolicy`, `RetryPolicy`, etc) will include a common `TargetRef`
-field in their specification to identify how and where to apply that policy.
+This GEP aims to standardize attachment of one Kubernetes object to another in
```
robscott, line 8: I think this may be too vague. Probably the most used forms of attachment in Gateway API will be Route -> Parent and Route -> Backend, neither of which use the models described in this doc.

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
...
+
+### Targeted metaresources
+
+ReferenceGrant is an example of a targeted metaresource. It’s tightly bound to a
+particular Kind within a single namespace, and only modifies the behavior of the
+objects that match its binding (by allowing references to them from outside their
+namespace).
+
+We’ve been discussing the idea of having a new object that holds extra capabilities
+for a backend Service (something like BackendCapabilities or similar, name TBD),
```
robscott, line 86: I'm hesitant to make too many references to resources that have not yet been defined. Generally once a GEP is merged, we tend to forget about it - and references to terms like this may soon become outdated. Maybe it would be useful to make this a bit more generic and future proof by focusing on the general concept of extending a Service.

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
...
+  namespace should be avoided unless it uses a handshake of some sort, where the
+  things outside the namespace can opt–out of the behavior. (Notably, this is the
+  design that we used for ReferenceGrant).)
+
+### Policy Attachment: It's all about the defaults and overrides
+
+Because a Policy is a metaresource, it targets some other resource and _augments_
+its behavior.
+
+But why have this distinct from other types of metaresource? Because Policy
```
robscott, line 133: In my opinion, there are still valid cases of policy attachment that could look more like a "Targeted Metaresource" (or "Direct Policy Attachment"). In GKE, we're using HealthCheckPolicy to configure custom health checks, but of course this can only be attached to a Service, so the direct attachment model would have actually been useful for us in this case. https://cloud.google.com/kubernetes-engine/docs/how-to/configure-gateway-resources#configure_health_check.

robscott (commented):
Thanks @youngnick! Sorry it took me so long to get to this!

site-src/geps/gep-1364.md
```diff
@@ -419,7 +419,7 @@ an Extended condition.
 It will have the following behavior:
 
 * `Ready` is an optional Condition that has Extended support, with conformance
-tests to verify the behavior.
+tests to verify the behvaior.
```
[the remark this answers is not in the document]
youngnick: 🤦 Yeah, reverted.

site-src/geps/gep-713.md
youngnick: I've thought about this, and I think that having them both in the same document is better - I think that the uses of direct policy attachment will be much less that hierarchical policy attachment.

site-src/geps/gep-713.md
youngnick: I've updated here, PTAL.

site-src/geps/gep-713.md
youngnick: I've updated this, PTAL.

site-src/geps/gep-713.md
youngnick: Updated.

geps/gep-713.md
```diff
@@ -38,24 +56,161 @@ portable.
...
+
+All this does is tell an implementation, that for connecting to port `8443` on the
+Service `fooService`, it should assume that the connection is TLS, and expect the
+service's certificate to be validated by the chain in the `CAcert` Secret.
+
+Importantly, this would apply to _every_ usage of that Service across any HTTPRoutes
+in that namespace, which could be useful for a Service that is reused in a lot of
+HTTPRoutes.
+
+With this two examples in mind, here are some guidelines for when to consider
```
robscott, line 142:
```suggestion
With these two examples in mind, here are some guidelines for when to consider
```

geps/gep-713.md
```diff
@@ -38,24 +56,161 @@ portable.
...
+using Direct Policy Attachment:
+* The number or scope of objects to be modified is limited or singular. Direct
```
robscott, line 144: Sadly our markdown renderer needs an empty line before the start of lists: https://deploy-preview-1565--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/

geps/gep-713.md
```diff
@@ -38,24 +56,161 @@ portable.
...
+  the modifications are valid, the metaresource is valid, and this should be
+  straightforward to communicate in one or two Conditions. Note that at the time
+  of writing, this is *not* completed.
+* Direct Policy Attachment _should_ only be used to target objects in the same
+  namespace as the Policy object. Allowing cross-namespace references brings in
+  significant security concerns, and/or difficulties about merging cross-namespace
+  policy objects. Notably, Mesh use cases may need to do something like this for
+  consumer policies, but in general, Policy objects that modify the behavior of
+  things outside their own namespace should be avoided unless it uses a handshake
+  of some sort, where the things outside the namespace can opt–out of the behavior.
```
robscott, line 162: The merged version of this GEP currently limits cross-namespace references to the following:

>Policy can be applied to any resource in any namespace but it can only apply to requests from the same namespace if the target is in a different namespace

This feels like an expansion of that, is that intentional?

geps/gep-713.md
```diff
@@ -38,24 +56,161 @@ portable.
...
+  (Notably, this is the design that we used for ReferenceGrant).)
```
robscott, line 163: Nit: Extra closing paren here

geps/gep-713.md
robscott: It looks like you've limited cross-namespace references to direct policy attachment? If so, I think that's a great additional limitation as long as we're not breaking anyone along the way.

geps/gep-713.md
```diff
@@ -100,6 +255,13 @@ The `targetRef` field MUST have the following structure:
 ```go
 // PolicyTargetReference identifies an API object to apply policy to.
 type PolicyTargetReference struct {
+    // Type is the type of the policy attachment.
+    // This can be either Direct or Hierarchical.
+    //
+    Type string `json:"type"`
```
robscott, line 263: I'd actually assumed we'd need this to be something like a standard label on the CRD. It seems like we wouldn't be able to reuse the same APIs for direct and hierarchical because the latter would also have `overrides` and `defaults` built in to the CRD.

site-src/references/policy-attachment.md
```diff
@@ -1,25 +1,38 @@
...
-
-* A standardized means of attaching policy to resources.
-* Support for configuring both default and override values within policy
-  resources.
-* A hierarchy to illustrate how default and override values should interact.
-
-This kind of standardization not only enables consistent patterns, it allows
-future tooling such as kubectl plugins to be able to visualize all policies that
-have been applied to a given resource.
+# Metaresources and Policy Attachment
```
robscott, line 1: This is for a separate PR, but reviewing this page reminded me of a broader idea I'd had recently. I think we may want to introduce a broader document/set of documents that cover all the ways the API can be extended. That would include both forms of Policy Attachment as well as ExtensionRefs and GatewayClass Params.

site-src/references/policy-attachment.md
```diff
@@ -160,6 +173,7 @@ spec:
         includeProtocol: true
         includeQueryString: true
   targetRef:
+    type: hiearchical
```
robscott, line 176: This feels like it's going to be the most commonly misspelled word in a k8s API :). If we can avoid this as an input altogether, I expect we'll avoid some frustration. 
```suggestion
    type: hierarchical
```

site-src/references/policy-attachment.md
```diff
@@ -219,45 +235,37 @@ the same resource _and_ have an identical field specified with different values,
 precedence MUST be determined in order of the following criteria, continuing on
 ties:
 
+* Hierarchical Policies override Direct Policies.
```
robscott, line 238: I'd actually thought the opposite would be true 🙃. Ideally I think there would be no overlap in what you could configure in a hierarchical and direct policy though. Are there any scenarios where you'd want these kinds of policy attachment to interact with each other?

robscott (commented):
Thanks @youngnick! Added lots of comments but really like this direction, lots of great ideas here.

geps/gep-713.md
```diff
@@ -1,29 +1,47 @@
...
-have their relevant effects applied to network traffic. Individual policy APIs
-(e.g. `TimeoutPolicy`, `RetryPolicy`, etc) will include a common `TargetRef`
-field in their specification to identify how and where to apply that policy.
-This will be important for providing a consistent experience across
-implementations of the API, even for configuration details that may not be fully
-portable.
+This GEP aims to standardize terminology and process around using one Kubernetes
+object modify the functions of one or more other objects.
+
+Generally, a Kubernetes object that that _augments_ the behavior of an object
```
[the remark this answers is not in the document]
youngnick: You're right, this is a new term, so I'll update.

geps/gep-713.md
youngnick: Yeah, that's the intent. For hierarchical policy attachment, the targeted object must be in the same namespace as the policy object.

For direct reference, you _can_ do it, although it's generally not a good idea except in specific use cases (ie mesh consumer policies).

geps/gep-713.md
youngnick: The idea here is to make the `TargetRef` struct be _the_ thing that defines a Policy object, and that way having this field in here lets tooling duck-type a Policy object and make some assumptions about it.

site-src/references/policy-attachment.md
youngnick: Yeah, I don't like the name either, I've been noodling some others that I'll suggest in the meeting today.

geps/gep-713.md
```diff
@@ -1,29 +1,47 @@
...
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Hierarchical Policy Attachment").
+
+Individual policy APIs must
+- be their own CRDs (e.g. `TimeoutPolicy`, `RetryPolicy` etc),
+- can be included in the Gateway API API group and installation or be defined by
+  implementations
+- and must include a common `TargetRef` struct in their specification to identify how and where to
+  apply that policy, along with a field that indicates if the reference is `Direct`
+  or `Hierarchical`. (Shorter name suggestions welcomed)
```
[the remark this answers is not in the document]
youngnick: Fixed, thanks.

site-src/geps/gep-713.md
```diff
@@ -38,10 +48,116 @@ portable.
...
+object (or objects). “Meta” here is used in its Greek sense of “more comprehensive”
+or “transcending”, and “resource” rather than “object” because “metaresource”
+is more pronounceable than “metaobject”. Additionally, a single word is better
+than a phrase like “wrapper object” or “wrapper resource” overall, although both
+of those terms are effectively synonymous with “metaresource”.
+
+All ReferenceGrants are metaresources, and all objects that use Policy Attachment
+are metaresources, but metaresources can be more than just those two things.
+
+### Targeted metaresources
```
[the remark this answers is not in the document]
youngnick: I've started this work, but I think the naming needs work. PTAL, keeping in mind we'll probably need to change `Hierarchical Policy` to something else.

youngnick:
I've pushed another update here, I think the biggest outstanding thing is, of course, naming.

Currently, there are two types of Policy Attachment:
- `Hierarchical`: the original form, with `defaults` and `overrides`.
- `Direct`: a simplified form that can only attach to one object (or a small set).

I think that the `Hierarchical` name is _technically correct_ but too complex and hard to write. (I've spelled it wrong _so many_ times already).
So, we need another name.

Some thoughts:
- `Direct` and `Indirect`: Simple, basically accurate, but I don't know if it indicates the hierarchical aspect enough.
- `Direct` and `Tree`: Not great, but it really pushes the hierarchy.
- Something else: I'm out of ideas.

I'll reword the site documentation once we've settled on this name.

TODO for the site docs:
- [ ] Direct attachment example
- [ ] Diagrams that include direct and hierarchical policy attachment
- [ ] Better explanation of the differences

geps/gep-713.md
```diff
@@ -38,24 +58,162 @@ portable.
...
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: TLSConnectionPolicy
+metadata:
+  name: tlsport8443
+  namespace: foo
+spec:
+  targetRef: # This struct is defined as part of Gateway API
+    type: Direct
```
howardjohn, line 125: I had expected Direct vs Hierarchical was a property of the CRD - whether it supports the more complex form or only the simplified one.

I don't really understand how a user would know to set either field. Can you show an example of the same policy type (TLSConnectionPolicy) in both modes and explain how they differ?

shaneutt (approved):
Looking this over since there's been significant changes since my previous approval (as this PR has been "stuck" now for quite some time), I do see language and things I would _like_ to tweak, _however_ it's predominantly for reasons of brevity which (since we're not moving to standard yet) I feel I could just as easily contribute changes _myself_ in a follow up PR where I feel strongly.

This is `Experimental` I want to operate within the spirit of that: **we're in the experimental stage**, the **core concepts in here make sense to me**, I imagine we'll have **plenty more iterations to come** before we move it _out_ of experimental but I don't feel any of it's worth holding up _this_ PR. Therefore I approve.

site-src/references/policy-attachment.md
youngnick: Good point, I've updated.

geps/gep-713.md
youngnick: I'm actually kind of using this field as a way for the CRD to indicate in a field inside the object which type it is - I'm thinking about how we can conceivably write tooling to generically parse an arbitrary Policy object, _from any implementation_, and we need a way for it to know what sort of object we're dealing with. At a minimum, I'd like an
arbitrary tool to have relatively-simple rule to allow it to guess if the Policy object is hierarchical or direct.

I guess I'm also trying to lean in to the statement `targetRef` is what makes a Policy object.

You're right that this doesn't make sense as a user-settable field, but neither does an annotation, I think, so I'm at a loss.

Maybe a better option would be to say "a Policy is an object with a `targetRef`, and a Hierarchical Policy is a Policy object with a `targetRef` and one or more of `defaults` or `overrides`." So, unless you use the standard way of doing Hierarchical Policy, then your policy must only affect that one object. I'll think about that some more, but hopefully this explains a little better what I was aiming at.

howardjohn: Defaulting in the CRD could set it (and enforce it isn't set to anything else) perhaps?

Or the bit about defaults/overrides

youngnick: I though about this some more, and updated to make the defining feature of a hierarchical policy that it possesses a `targetRef` stanza, and then either a `defaults`, an `overrides`, or both. Not having at least one of `defaults` or `overrides` means that your Policy is a Direct Attached Policy.

Essentially, this reserves two special keys in the top level of every Policy CRD, `defauls`, and `overrides`, and uses their presence to imply behavior.

I don't think I can do much clearer than that without also being too clever. 😄

youngnick:
Okay, I think I've got all the comments now. PTAL.

geps/gep-713.md
```diff
@@ -289,6 +457,7 @@ the same resource _and_ have an identical field specified with different values,
 precedence MUST be determined in order of the following criteria, continuing on
 ties:
 
+* Hierarchical Policies override Direct Policies.
```
howardjohn, line 460: You say the opposite in site-src/references/policy-attachment.md L254

youngnick: Ah, thanks, having two places to record this stuff is not ideal. I'll fix.

arkodg:
@youngnick any guidance on merge semantics for non scalar type fields within hierarchical policies ?

geps/gep-713.md
maleck13: I like this way (default and overrides vs just targetRef) of distinguishing direct vs Hierarchical so +1 from me

site-src/references/policy-attachment.md
```diff
@@ -170,6 +199,7 @@ spec:
       cachePolicy:
         includeQueryString: false
   targetRef:
+    type: direct
```
sunjayBhatia, line 202: this line seems removable now based on other changes?

youngnick:
@arkodg I think that, for hierarchical policies, the merge semantics depend on whether the settings are in `defaults` or `overrides`. However, for non-scalar fields, the merge semantics should be more context-dependent than just merging an entire struct into the same struct in another object.

I'll add in what I think here, and see if that makes sense.

geps/gep-713.md
```diff
@@ -1,29 +1,49 @@
...
+one.
+
+This document proposes controlling the creation of configuration in the underlying
+Gateway data plane using two types of Policy Attachment.
+A "Policy Attachment" is a specific type of _metaresource_ that can affect specific
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Hierarchical Policy Attachment").
+
+Individual policy APIs:
+- must be their own CRDs (e.g. `TimeoutPolicy`, `RetryPolicy` etc),
```
hzxuzhonghu, line 25: Could gateway-api provide such policy api?

geps/gep-713.md
```diff
@@ -1,29 +1,49 @@
...
+- can be included in the Gateway API API group and installation or be defined by
+  implementations
+- and must include a common `TargetRef` struct in their specification to identify how and where to
```
hzxuzhonghu, line 28: Why does bi-directional reference need?

geps/gep-713.md
```diff
@@ -277,18 +444,46 @@ spec:
...
+should merge the fields at a scalar level, with the `includeHost`,
+`includeProtocol`, and `includeQueryString` values being defaulted if they were
+not specified in the object being controlled. Similarly, for `overrides`, the
+values of the innermost scalar fields should overwrite the scalar fields in the
+affected object.
+
+Implementations should not copy the higher-level structs directly into the
+affected object.
+
+In the case that the field in the Policy affects a struct that is a member of a list,
```
arkodg, line 473: lets take an example of a Hierarchical policy
* Only specify defaults
Applied at Gateway
```
RetryPolicy
default:
  retryOn:
  - "501"
  - "502"
  - "503"
```
Applied at xRoute
```
RetryPolicy
default:
  retryOn:
  - "504"
```
is this the final state (appending/merging `defaults`) ?
```
retryOn:
  - "501"
  - "502"
  - "503"
  - "504"
 ``` 
 
* Top level specifies a default and child level specifies a override
Applied at Gateway
```
RetryPolicy
default:
  retryOn:
  - "501"
  - "502"
  - "503"
```
Applied at xRoute
```
RetryPolicy
override:
  retryOn:
  - "504"
```
is this the final state (bottom level `override` wins) ?
```
retryOn:
  - "504"
 ``` 
 * Top level specifies a override and child level specifies a default
Applied at Gateway
```
RetryPolicy
override:
  retryOn:
  - "501"
  - "502"
  - "503"
```
Applied at xRoute
```
RetryPolicy
default:
  retryOn:
  - "504"
```
is this the final state (top level `override` wins) ?
```
retryOn:
  - "501"
  - "502"
  - "503"
 ```

Comment (howardjohn) [reply geps/gep-713.md:473]: Just my 2c, it probably makes sense to be implementation specific how merging is done. For this case, I would recommend implementations replace.

If we do prescribe replace vs merge it should be replace, IMO, since that covers more use cases (you can always add more with replace, but cannot take away with merge).

Comment (arkodg) [reply geps/gep-713.md:473]: yeah agreed `replace` for all cases isa good solution, would prefer if this was a recommendation from this project

File: geps/gep-713.md [hunk 27]
```diff
@@ -38,24 +58,162 @@ portable.
...
+and only modifies the fields of or the configuration associated with that object.
+
+"Hierarchical Policy Attachment" is when a Policy object references a single object
+_and any child objects of that object_ (according to some defined hierarchy), and
+modifies fields of the child objects, or configuration associated with the child
+objects.
+
+### Direct Policy Attachment
+
+A Direct Policy Attachment is tightly bound to one or more instances of a particular
```
sunjayBhatia, line 101: above we say `"Direct Policy Attachment" is when a Policy object references a single object _only_` but `one or more`

since `PolicyTargetReference` currently has the `name` field as required I guess the former is currently true (unless we plan on changing that?)

geps/gep-713.md
sunjayBhatia: +1 to how it's described [here](https://github.com/kubernetes-sigs/gateway-api/pull/1565/files#r1133422862) (and looks like the line this thread is attached to can be removed)

sunjayBhatia (approved):
left some small comments but I'm generally on board with the changes here 👍🏽

geps/gep-713.md
youngnick: Okay, there's some confusion here about the _intent_ of `defaults` and `overrides`, and it's also intersecting with a particular type of element that makes merging very difficult.

For the latter, merging lists is always going to be hard, that's why Kubernetes (and kubebuilder) have the listMergeType directives, so that the apiserver and other api machinery get some clues about how to merge things. A list of strings is probably close to the worst-case for merge complexity, because there are no clues about priority _in the string itself_, unlike if there was a structured object there.

That said, I can see that in the example here, it does make sense. But we need to understand that we should _try_ to avoid lists-of-strings in the same way we should avoid `map[string]string` - the lack of structure makes doing things with the field harder in the long term. (That isn't to say we should never do it, just that it has a cost we should consider).

For this specific setting, you could achieve the same aim by using a single string that contained comma-separated values. To that end, when a Policy changes the `retryOn` field, it will _replace_ the field, not be merged into it. Effectively, the _value_ of the list is the _set of strings_ included in the list, and the Policy will replace the value, the same as it would on a `key: value` field inside a struct. The same applies for `value: map[string]string` fields.

In this case, the other rule that's important is the differences between `defaults` and `overrides`.

The overall rule is that only one instance of a Policy should be in effect on the final object at any one time. If two Policy objects of the same type end up affecting the same leaf node, then only one can win.

`defaults` are a "more specific wins" construct, so if multiple Policy objects are attached, the one that's _closer to the thing being modified_ will win.

`overrides` are a "less specific wins" construct, so if multiple Policy objects are attached, the ones that's _higher in the hierarchy_ will win.

In the event a default and an override both apply, no matter what level they've come from, then the override will win.

This means that in the first example, the defaults from the two Policy objects should _not_ be merged. The defaults should be merged _into_ the thing they're modifying, but only one instance of the same Policy will be the winner.
That means that the final result should look like

```
RetryPolicy
default:
  retryOn:
  - "504"
```

This is the same as if the actual Route had specified the setting.

For the second example, top level specifies a default and a lower an override, the override wins, because overrides beat defaults.

For the third, top level specifies an override, and a lower a default, the override wins, because overrides beat defaults.

Some other examples:
Top level specifies an override, and a lower a different override, the top-level override wins, because overrides are "less specific beats more specific".

This comment was me working through my thoughts, I'll take it and fold it into the GEP.

youngnick:
I started writing some updates to address @arkodg's feedback, and it got a bit out of hand. I've added a bunch of extra description and examples to further describe:
- how interactions between overrides and defaults should work
- how fields should be overridden, including for non-scalar fields
- when and how merging should be done (it's complicated).

I'm open to discussing if we should simplify this by only allowing one Policy of each Kind to be in effect on any object at once, but I think that it would substantially decrease the usefulness of this pattern.

I've got pretty detailed to try and eliminate edge cases, but you can sum up the rules for Hierarchical Policy interactions like this:
- `overrides` override everything from the top down
- `defaults` set defaults, with more specific beating less specific defaults
- If a Policy affects a _field_ inside an object, then that field is the most specific possible value, and so will beat _any_ default from anywhere.

geps/gep-713.md
youngnick: We will probably end up supplying some Policy CRDs, I suspect. But it's not certain yet.

geps/gep-713.md
youngnick: I'm not sure what you mean, sorry. There's no bidirectional reference here, just a mandated struct.

geps/gep-713.md
youngnick: Ah yes, sorry, updated.

geps/gep-713.md
youngnick: I'll leave this thread here for a bit so everyone can see it, but I've folded this language into both the GEP and the doc page now, so I consider this thread resolved.

youngnick:
Oh, I've also folded most of the new changes into the doc page as well as the GEP, as I think that it should make the doc page clearer.

Also, I've specifically left the section about adding a `sectionName` to `targetRef` alone for now, although I do think we should change it, this PR is big enough already.

geps/gep-713.md
```diff
@@ -525,13 +783,248 @@ type RouteRule struct {
...
+apiVersion: networking.example.io/v1alpha1
+kind: TLSMinimumVersionPolicy
+metadata:
+  name: minimum12
+  namespace: default
+spec:
+  defaults:
+    minimumTLSVersion: 1.2
+  targetRef:
+    name: default
```
robscott, line 827: The use of "default" as the namespace name here feels confusing, maybe some other more generic term would be helpful so there's no chance that someone thinks that setting "default" as a name has some kind of special meaning.

robscott (commented):
Thanks for all the work on this @youngnick! There are some really great improvements in this. I ran out of time to make it all the way through this, but added a bunch of small comments. Overall, I'm in agreement with all of the major changes/additions I've read so far.

geps/gep-713.md
```diff
@@ -1,29 +1,49 @@
...
+A "Policy Attachment" is a specific type of _metaresource_ that can affect specific
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Hierarchical Policy Attachment").
+
+Individual policy APIs:
+- must be their own CRDs (e.g. `TimeoutPolicy`, `RetryPolicy` etc),
+- can be included in the Gateway API API group and installation or be defined by
+  implementations
+- and must include a common `TargetRef` struct in their specification to identify how and where to
+  apply that policy, along with a field that indicates if the reference is `Direct`
```
howardjohn, line 29: I think we discussed removing the "field indicating the reference type" part?

geps/gep-713.md
```diff
@@ -38,24 +58,210 @@ portable.
...
+objects.
+
+In either case, a Policy may either affect an object by controlling the value
+of one of the existing _fields_ in the `spec` of an object, or it may add
+additional fields that are _not_ in the `spec` of the object.
+
+### Direct Policy Attachment
+
+A Direct Policy Attachment is tightly bound to one or more instances of a particular
+Kind within a single namespace, and only modifies the behavior of the object or
```
howardjohn, line 106: nit: can it not apply to a cluster scoped resource?

geps/gep-713.md
howardjohn: This was resolved but I don't think the example was updated

geps/gep-713.md
```diff
@@ -525,13 +783,248 @@ type RouteRule struct {
...
+This means that, for HTTPRoutes that do _NOT_ explicitly set this field to something
+else, (in other words, they contain an empty list), then the field will be set to
+a list containing `501`, `502`, and `503`. (Notably, because of Go zero values, this
+would also occur if the user explicitly set the value to the empty list.)
+
+However, if a HTTPRoute owner sets any value other than the empty list, then that
+value will remain, and the Policy will have _no effect_. These values are _not_
+merged.
+
+If the Policy used `overrides` instead:
```
howardjohn, line 906: I think this may be missing the conclusion?

geps/gep-713.md
```diff
@@ -525,13 +783,248 @@ type RouteRule struct {
...
+spec:
+  overrides:
+    retryOn:
+      - "501"
+      - "502"
+      - "503"
+  targetRef:
+    kind: Gateway
+    group: gateway.networking.k8s.io
+    name: YouMustRetry
```
howardjohn, line 922: very nit: 
```suggestion
    name: you-must-retry
```

I don't think the other is a legal name

howardjohn (commented):
Overall looks good, just left a few nitpicks. Nice!

arkodg (approved):
LGTM, thanks for incorporating the "Direct" policy type and clarifying the semantics for Hierarchical policies

geps/gep-713.md
```diff
@@ -1,29 +1,49 @@
...
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Hierarchical Policy Attachment").
+
+Individual policy APIs:
+- must be their own CRDs (e.g. `TimeoutPolicy`, `RetryPolicy` etc),
+- can be included in the Gateway API API group and installation or be defined by
+  implementations
+- and must include a common `TargetRef` struct in their specification to identify how and where to
+  apply that policy, along with a field that indicates if the reference is `Direct`
+  or `Hierarchical`. (Shorter name suggestions welcomed)
```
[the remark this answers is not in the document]
youngnick: I really like "Inherited", I think that's actually both easier to type and understand. I'll update.

geps/gep-713.md
```diff
@@ -1,29 +1,49 @@
...
+
+For Hierarchical Policies, this GEP also describes a set of expected behaviors
+for how settings can flow across a defined hierarchy.
+
 
 ## Goals
 
-* Establish a pattern for policy attachment which will be used for any policies
+* Establish a pattern for Policy objects which will be used for any policies
```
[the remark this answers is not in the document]
youngnick: I think in this case it is, but I've been avoiding the term "resources" because it has such a specific meaning in api-machinery.

geps/gep-713.md
```diff
@@ -38,24 +58,210 @@ portable.
...
+
+To put this another way, sometimes we need ways to be able to affect how an object
+is interpreted in the API, without representing the description of those effects
+inside the spec of the object.
+
+This document describes the ways we design objects to meet these two use cases,
+and why you might choose one or the other.
+
+We use the term “metaresource” to describe the class of objects that _only_ augment
+the behavior of another Kubernetes object, regardless of what they are targeting.
```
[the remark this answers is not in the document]
youngnick: I'd argue that you're still augmenting the _behavior_, because the underlying object is _part of_ the behavior.

Also, the word `augment` took a lot of arguing to get to, so I'm reluctant to mess with this too much. 😄

geps/gep-713.md
```diff
@@ -38,24 +58,210 @@ portable.
...
+
+A "Policy Attachment" is a metaresource that affects the fields in existing objects
+(like Gateway or Routes), or influences the configuration that's generated in an
+underlying data plane.
+
+"Direct Policy Attachment" is when a Policy object references a single object _only_,
+and only modifies the fields of or the configuration associated with that object.
+
+"Hierarchical Policy Attachment" is when a Policy object references a single object
+_and any child objects of that object_ (according to some defined hierarchy), and
```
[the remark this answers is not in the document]
youngnick: I'd agree with that, but we actually use at least two different hierarchies in the examples... (I'm also using Namespace -> Gateway, since we suggest _not_ using GatewayClass as a way to apply Policy except in certain circumstances).

I'd prefer to leave this as "some defined" because then it covers something like having a Policy that controls a custom HTTP Filter, which is not _exactly_ the same as the GatewayClass -> Gateway -> HTTPRoute -> Backend hierarchy we will _mainly_ be using.

geps/gep-713.md
```diff
@@ -38,24 +58,210 @@ portable.
...
+modifies fields of the child objects, or configuration associated with the child
+objects.
+
+In either case, a Policy may either affect an object by controlling the value
+of one of the existing _fields_ in the `spec` of an object, or it may add
```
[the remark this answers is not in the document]
youngnick: Yes, exactly. If we added a request timeout, for example, an override policy _may_ override that specific value.

geps/gep-713.md
```diff
@@ -277,18 +495,53 @@ spec:
...
+a `cachePolicy` struct that contains fields. If an implementation was merging
+this configuration into an existing object that contained the same fields, it
+should merge the fields at a scalar level, with the `includeHost`,
+`includeProtocol`, and `includeQueryString` values being defaulted if they were
+not specified in the object being controlled. Similarly, for `overrides`, the
+values of the innermost scalar fields should overwrite the scalar fields in the
+affected object.
+
+Implementations should not copy the higher-level structs directly into the
+affected object.
```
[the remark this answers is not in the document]
youngnick: Clarified, PTAL.

geps/gep-713.md
```diff
@@ -308,28 +561,33 @@ the API structure defined above and add a `gateway.networking.k8s.io/policy:
...
+  that can result from a metaresource update.
 * Since we only control some of the resources a policy might be attached to,
-  adding policies to status would only be possible on Gateway API resources, not
-  Services or other kinds of backends.
+  adding policies to status would only be possible on the policy objects themselves
+  or on Gateway API resources, not Services or other kinds of backends.
 
-Although these concerns are not unsolvable, they lead to the conclusion that
-a Kubectl plugin should be our primary approach to providing visibility here,
-with a possibility of adding policies to status at a later point.
```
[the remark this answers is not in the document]
youngnick: This section is still present, just lower.

geps/gep-713.md
youngnick: Thanks, removed.

geps/gep-713.md
youngnick: Updated.

geps/gep-713.md
```diff
@@ -38,24 +58,210 @@ portable.
...
+    of an `override`). The same applies to `map[string]string` fields. An example
+    here would be a field that stores a map of annotations - specifying a Policy
+    that overrides annotations will mean that a final object specifying those
+    annotations will have its value _entirely replaced_ by an `override` setting.
+* In the case that two Policies of the same type specify different fields, then
+  _all_ of the specified fields should take effect on the affected object.
+
+Examples to further illustrate these rules are given below.
+
 ## API
```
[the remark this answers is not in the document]
youngnick: If they can be applied on multiple levels, then they should be an Inherited (renamed from Hierarchical) Policy. I agree there's no good way to represent this exact use case for now though, so can we come back and follow this one up? Maybe we can add another reserved top-level key that makes something an Inherited Policy?

geps/gep-713.md
```diff
@@ -277,18 +495,53 @@ spec:
...
+affected object.
+
+Implementations should not copy the higher-level structs directly into the
+affected object.
+
+In the case that the field in the Policy affects a struct that is a member of a list,
+each existing item in the list in the affected object should have each of its
+fields compared to the corresponding fields in the Policy.
+
+For non-scalar field _values_, like a list of strings, or a `map[string]string`
```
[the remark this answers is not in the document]
youngnick: I added a small table, which hopefully helps. The key is that there's no merging in the non-scalar case anyway.

geps/gep-713.md
```diff
@@ -525,13 +783,248 @@ type RouteRule struct {
...
+resources in a namespace. Multiple hierarchies are possible, even within Gateway
+API, for example Gateway -> Route, Gateway -> Route -> Backend, Gateway -> Route
+-> Service. GAMMA Policies could conceivably use a hierarchy of Service -> Route
+as well.
+
+Note that this will not be very discoverable for Gateway owners in the absence of
+a solution to the Policy status problem. This is being worked on and this GEP will
+be updated once we have a design.
+
+Conceivably, a security or admin team may want to _force_ Gateways to have at least
```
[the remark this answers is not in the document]
youngnick: Yeah, I know, but I was running out of good-example brain juice. PRs to improve this welcomed once this goes in? 😄

youngnick:
Okay, I think I've got both those rounds, thanks @robscott and @howardjohn!

The big change in this last revision is that I picked "Inherited" for the other Policy name, so we now have "Direct" or "Inherited" Policy.

howardjohn:
LGTM, since this was a pretty popular PR and just changed I'll let someone else give the official approval though to give some time to review. Ping me if that never happens :slightly_smiling_face:

geps/gep-713.md
```diff
@@ -38,24 +58,211 @@ portable.
...
+Policy resources are designed to have a way for settings to flow down a hierarchy.
+
+Defaults set the default value for something, and can be overridden by the
+“lower” objects (like a connection timeout default policy on a Gateway being
+overridable inside a HTTPRoute), and Overrides cannot be overridden by “lower”
+objects (like setting a maximum client timeout to some non-infinite value at the
+Gateway level to stop HTTPRoute owners from leaking connections over time).
+
+Here are some guidelines for when to consider using a Inherited Policy object:
+* The settings or configuration are bound to one containing object, but affect
```
robscott, line 188: Newline needed for markdown list rendering: https://deploy-preview-1565--kubernetes-sigs-gateway-api.netlify.app/references/policy-attachment/

```suggestion

* The settings or configuration are bound to one containing object, but affect
```

geps/gep-713.md
```diff
@@ -38,24 +58,211 @@ portable.
...
+* In terms of status, a good accounting for how to record that the Policy is
+  attached is easy, but recording what resources the Policy is being applied to
+  is not, and needs to be carefully designed to avoid fanout apiserver load.
+  (This is not built at all in the current design either).
+
+When multiple Inherited Policies are used, they can interact in various ways,
+which are governed by the following rules, which will be expanded on later in
+in this document.
+
+* If an Policy does not affect an object's fields directly, then the resultant
```
robscott, line 206:
```suggestion
* If a Policy does not affect an object's fields directly, then the resultant
```

geps/gep-713.md
```diff
@@ -38,24 +58,211 @@ portable.
...
+
 Access is granted with RBAC - anyone that has access to create a RetryPolicy in
 a given namespace can attach it to any resource within that namespace.
 
 ![Simple Ingress Example](images/713-ingress-simple.png)
 
-To build on that example, it’s possible to attach policies to more resources.
-Each policy applies to the referenced resource and everything below it in terms
-of hierarchy. Although this example is likely more complex than many real world
+A Inherited Policy can attach to a parent resource, and then each policy
```
robscott, line 263:
```suggestion
An Inherited Policy can attach to a parent resource, and then each policy
```

geps/gep-713.md
```diff
@@ -227,13 +434,14 @@ spec:
     kind: Gateway
     name: example
 ---
-kind: GKEServicePolicy
+kind: CDNCachingPolicy
```
robscott, line 437: Thanks for cleaning this up!

geps/gep-713.md
```diff
@@ -525,13 +791,251 @@ type RouteRule struct {
...
+|Gateway override| Gateway override | Namespace override | Gateway override<br />first created wins<br />otherwise first alphabetically | Gateway override |
+|HTTPRoute override| HTTPRoute override | Namespace override | Gateway override | HTTPRoute override<br />first created wins<br />otherwise first alphabetically|
+
+#### Defaults interacting with other defaults for RetryOnPolicy, empty list in HTTPRoute
+||No default|Namespace default|Gateway default|HTTPRoute default|
+|----|-----|-----|----|----|
+|No default|Empty list|Namespace default| Gateway default| HTTPRoute default|
+|Namespace default| Namespace default| Namespace default<br />first created wins<br />otherwise first alphabetically | Gateway default | HTTPRoute default |
+|Gateway default| Gateway default | Gateway default | Gateway default<br />first created wins<br />otherwise first alphabetically | HTTPRoute default |
+|HTTPRoute default| HTTPRoute default | HTTPRoute default | HTTPRoute default | HTTPRoute default<br />first created wins<br />otherwise first alphabetically|
```
robscott, line 1001: Maybe just me, but I find these tables pretty confusing. https://deploy-preview-1565--kubernetes-sigs-gateway-api.netlify.app/references/policy-attachment/. The values in the cells seem to largely be just repeating the headers?

robscott (commented):
Thanks @youngnick! This is a really great improvement. Had a few areas where I'm still confused but agree with direction of everything here.

site-src/references/policy-attachment.md
```diff
@@ -1,25 +1,176 @@
...
+settings across either one object (this is "Direct Policy Attachment"), or objects
+in a hierarchy (this is "Inherited Policy Attachment").
+
+In either case, a Policy may either affect an object by controlling the value
+of one of the existing _fields_ in the `spec` of an object, or it may add
+additional fields that are _not_ in the `spec` of the object.
+
+### Direct Policy Attachment
+
+A Direct Policy Attachment is tightly bound to one or more instances of a particular
```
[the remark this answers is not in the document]
youngnick: No, I'll remove all the "one or more" things now.

geps/gep-713.md
youngnick: That's kind of the idea, they're intended to be like truth tables representing the various interactions that make it clear that the `override` value always wins. I thought it might be better to be saying this in more than one way.

youngnick: I updated the truth tables with labels for rows and columns, hopefully that makes it a bit clearer.

site-src/references/policy-attachment.md
```diff
@@ -69,9 +220,10 @@ struct included in the Gateway API. Where possible, it is recommended to use
 that struct directly instead of duplicating the type.
 
 ### Policy Boilerplate
-The following structure MUST be used as for any Policy resource using this API
-pattern. Within the spec, policy resources may omit `Override` or `Default`
-fields, but at least one of them MUST be present.
+The following (or something like it) SHOULD be used as for any Policy resource using this API
+pattern. Within the spec, policy resources _may_ omit `Override` or `Default`
```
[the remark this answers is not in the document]
youngnick: Thanks, I clarified.

site-src/references/policy-attachment.md
```diff
@@ -204,60 +365,106 @@ kind: RetryPolicy
...
 ### Conflict Resolution
-It is possible for multiple policies to target the same resource. When this
-happens, merging is the preferred outcome. If multiple policy resources target
+It is possible for multiple policies to target the same object _and_ the same
+fields inside that object. If multiple policy resources target
 the same resource _and_ have an identical field specified with different values,
 precedence MUST be determined in order of the following criteria, continuing on
 ties:
 
+* Direct Policies override Inherited Policies. If preventing settings from
```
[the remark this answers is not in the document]
youngnick: Yeah, agreed. I've changed this.

site-src/references/policy-attachment.md
```diff
@@ -303,3 +510,241 @@ the same behavior and semantics, although they may not be able to support
...
+spec:
+  minimumTLSVersion: 1.2
+  targetRef:
+    name: internet
+    group: gateway.networking.k8s.io
+    kind: Gateway
+```
+
+Note that because there is no version controlling the minimum TLS version in the
+Gateway `spec`, this is an example of a non-field Policy.
```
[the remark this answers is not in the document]
youngnick: Updated.

site-src/references/policy-attachment.md
```diff
@@ -303,3 +510,241 @@ the same behavior and semantics, although they may not be able to support
...
+Then this will have the same effect as applying that Policy to every Gateway in
+the `default` namespace - namely that every HTTPRoute that attaches to every
+Gateway will have its `retryOn` field set to `501`, `502`, `503`, _if_ there is no
+other setting in the HTTPRoute itself.
+
+With two layers in the hierarchy, we have a more complicated set of interactions
+possible.
+
+Let's look at some tables for a particular HTTPRoute, assuming that it does _not_
+configure the `retryOn` field, for various types of Policy at different levels.
```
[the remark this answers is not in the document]
youngnick: Sigh. I guess you're right. I always anticipated that it would apply to fields, but it make sense for it not to. I'll update and remove the references to `retryOn` being a field.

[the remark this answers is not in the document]
youngnick: If you feel really strongly about it, I can change it, but that will involve quite a bit of change to the example (to use a Filter instead or something).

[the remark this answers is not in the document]
robscott: Up to this point I've thought of policy attachment as purely an extension mechanism, not something that actually overrides behavior specified by the resource. I think right now the UX would be pretty awful if this actually happened. It's one thing to say "some other resource may add config on top of this route to extend it", but it's an entirely different thing to say "some other resource may completely override the config you've specified in this route." If we want policy attachment to be able to solve the latter I think we need more than just status on the policy resource, but actually on the target resource. 

It's not obvious to me that we _need_ this today, so I'd rather leave it out of scope until we do because I'm quite concerned about the UX issues this would introduce. The lack of status is bad enough when we're extending functionality, it would be even more pronounced if we were overriding functionality specified by the API. Of course I can see the usefulness of a capability like this, but so far the best use cases seem to be concepts we haven't included in the API (timeouts, retries, etc), and when we include them, they may just be policies.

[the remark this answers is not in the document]
youngnick: Yeah, okay. I'll update the example to be a RetryFilter that's controlled by a RetryPolicy.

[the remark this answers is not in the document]
youngnick: I've updated. I actually think it makes this a bit harder to understand, but I get that we don't want to imply you can do something we don't know if we want to allow.

robscott:
Thanks for the massive amount of work on this @youngnick! This is a huge step forward for our policy attachment model. 

/lgtm

PR #1576, opened 2022-12-03
v0.6.0 review cleanup

robscott:
**What type of PR is this?**
/kind cleanup

This is a follow up to #1538, addressing many of the comments from that review. (There are still plenty more for anyone that has time to help out).

```release-note
NONE
```


apis/v1alpha2/grpcroute_types.go
```diff
@@ -90,7 +90,7 @@ type GRPCRouteStatus struct {
 type GRPCRouteSpec struct {
 	CommonRouteSpec `json:",inline"`
 
-	// Hostnames defines a set of hostname to match against the GRPC
```
shaneutt, line 93: Wait, why did we change this? This is godoc, the godoc for the attribute should be the same case as the code?

apis/v1alpha2/grpcroute_types.go
```diff
@@ -391,10 +392,31 @@ type GRPCHeaderMatch struct {
 	Value string `json:"value"`
 }
 
+// GRPCHeaderMatchType specifies the semantics of how GRPC header values should
+// be compared. Valid GRPCHeaderMatchType values are:
+//
+// * "Exact" * "RegularExpression"
+//
+// Note that values may be added to this enum, implementations must ensure that
```
shaneutt, line 400: I can see this language potentially being confusing for non-native english speakers, what do you think about:

```suggestion
// Note that new values may be added to this enum in future releases, implementations MUST ensure that
```

apis/v1alpha2/grpcroute_types.go
```diff
@@ -391,10 +392,31 @@ type GRPCHeaderMatch struct {
...
+// unknown values will not cause a crash.
+//
+// Unknown values here must result in the implementation setting the Accepted
```
shaneutt, line 403:
```suggestion
// Unknown values here MUST result in the implementation setting the Accepted
```

apis/v1alpha2/validation/httproute.go
```diff
@@ -107,8 +108,8 @@ func validateHTTPRouteFilters(filters []gatewayv1a2.HTTPRouteFilter, matches []g
 		delete(counts, key)
 	}
 
-	if counts[gatewayv1a2.HTTPRouteFilterRequestRedirect] > 0 && counts[gatewayv1a2.HTTPRouteFilterURLRewrite] > 0 {
-		errs = append(errs, field.Invalid(path.Child("filters"), gatewayv1a2.HTTPRouteFilterRequestRedirect, "Redirect and Rewrite filters cannot be defined in the same list of filters"))
+	if counts[gatewayv1b1.HTTPRouteFilterRequestRedirect] > 0 && counts[gatewayv1b1.HTTPRouteFilterURLRewrite] > 0 {
+		errs = append(errs, field.Invalid(path.Child("filters"), gatewayv1b1.HTTPRouteFilterRequestRedirect, "Redirect and Rewrite filters cannot be defined in the same list of filters"))
```
shaneutt, line 112: Given other examples in kube core and networking, perhaps this language would be better?

```suggestion
		errs = append(errs, field.Invalid(path.Child("filters"), gatewayv1b1.HTTPRouteFilterRequestRedirect, "may specify httpRouteFilterRequestRedirect or httpRouteFilterRequestRewrite, but not both"))
```

shaneutt (changes requested):
Looking good thanks! Just a couple smaller comments :+1:

apis/v1alpha2/grpcroute_types.go
```diff
@@ -220,7 +220,8 @@ type GRPCRouteRule struct {
 	//
 	// Conformance-levels at this level are defined based on the type of filter:
 	//
-	// - ALL core filters MUST be supported by all implementations.
+	// - ALL core filters MUST be supported by all implementations which support
```
youngnick, line 223: Nit:
```suggestion
	// - ALL core filters MUST be supported by all implementations that support
```

apis/v1alpha2/grpcroute_types.go
```diff
@@ -391,10 +392,31 @@ type GRPCHeaderMatch struct {
 	Value string `json:"value"`
 }
 
+// GRPCHeaderMatchType specifies the semantics of how GRPC header values should
+// be compared. Valid GRPCHeaderMatchType values are:
+//
+// * "Exact" * "RegularExpression"
```
youngnick, line 398: I think this will need this formatting to render correctly?
```suggestion
// * "Exact"
// * "RegularExpression"
```

Also, there's no clarification about regex support - elsewhere we have it as ImplementationSpecific because of flavor choices, should we replicate that here?

youngnick (commented):
Couple of small things from me, but I think with @shaneutt's changes and those things, this should be good to go.

apis/v1alpha2/grpcroute_types.go
robscott: Good catch! This looks like a broader issue, filed https://github.com/kubernetes-sigs/gateway-api/issues/1581 to follow up.

apis/v1alpha2/grpcroute_types.go
robscott: Reverted the change for now, I think there may have been an upstream effort to make godocs text match the user-facing capitalization, but a quick run through k/k shows that's not very widespread right now. @thockin might have more context.

apis/v1alpha2/grpcroute_types.go
robscott: Updated, thanks!

apis/v1alpha2/grpcroute_types.go
robscott: Updated, thanks!

apis/v1alpha2/validation/httproute.go
robscott: Updated, thanks!

apis/v1alpha2/grpcroute_types.go
robscott: Updated, thanks!

PR #1578, opened 2022-12-05
feat: conformance tests opt-out option

### LINKED ISSUE #1525: Conformance: allow opting out of subtests
Currently, we have some logic like this:

```go
			for _, ct := range tests.ConformanceTests {
				t.Run(ct.ShortName, func(t *testing.T) {
					if reason, f := skippedTests[ct.ShortName]; f {
						t.Skip(reason)
					}
					ct.Run(t, csuite)
				})
			}
```

To execute the conformance tests from our own Go runner. The issue is this is not very granular as the conformance has subtests in plenty of places.

It would be nice to be able to be more specific so we could opt out of something like `GatewayInvalidTLSConfiguration/Nonexistent_secret_referenced_as_CertificateRef_in_a_Gateway_listener` instead of just `GatewayInvalidTLSConfiguration/`

mlavacca:
**What type of PR is this?**

/kind feature

A new "SkipTests" field has been added to the conformance test options to opt out of specific tests.

Fixes #1525 

```release-note
A new SkipTests field has been added to the conformance test options to opt-out of specific tests.
```

### DIFF
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -91,7 +91,7 @@ type ConformanceTestSuite struct {
 	Applier           kubernetes.Applier
 	SupportedFeatures map[SupportedFeature]bool
 	TimeoutConfig     config.TimeoutConfig
-	SkipTests         []string
+	SkipTests         sets.Set[string]
 }
 
 // Options can be used to initialize a ConformanceTestSuite.
@@ -152,7 +152,7 @@ func New(s Options) *ConformanceTestSuite {
 		},
 		SupportedFeatures: s.SupportedFeatures,
 		TimeoutConfig:     s.TimeoutConfig,
-		SkipTests:         s.SkipTests,
+		SkipTests:         sets.New(s.SkipTests...),
 	}
 
 	// apply defaults
@@ -226,7 +226,7 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 	}
 
 	// check that the test should not be skipped
-	if sets.New(suite.SkipTests...).Has(test.ShortName) {
+	if suite.SkipTests.Has(test.ShortName) {
 		t.Logf("Skipping %s", test.ShortName)
 		return
 	}


mlavacca:
@howardjohn, @robscott, @shaneutt I created this PR to opt-out of specific tests or test cases. Since it requires a lot of refactoring along all the test suites, I'd like to ask for your feedback to validate the approach. I refactored three different test suites to give an idea of how the tests will look after this PR.

shaneutt (commented):
To give some early feedback, and at a very high level because this is a draft I think this looks like an OK way to go. I do think others should check it over and chime in as well.

mlavacca:
@howardjohn @robscott can you please take a look at this one? It is waiting for early feedback before getting finished 🙏

mlavacca:
After discussing with @robscott, we agreed upon scoping down the PR to allow users to skip only tests (and not test cases), as it would have required a vast (and scary) refactoring along all the conformance tests.

conformance/utils/suite/suite.go
```diff
@@ -90,6 +90,7 @@ type ConformanceTestSuite struct {
 	Applier           kubernetes.Applier
 	SupportedFeatures map[SupportedFeature]bool
 	TimeoutConfig     config.TimeoutConfig
+	SkipTests         []ConformanceTest
```
robscott, line 93: Nit: Can we use k8s sets util to avoid looping through this list every time to check for the presence of a test name? (https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/validation/common.go#L56). Related comment, it would probably be more efficient to just track the test name instead of the full test definition.

robscott (commented):
Thanks @mlavacca! This looks like a great simplification.

conformance/utils/suite/suite.go
mlavacca: 👍 Done

conformance/utils/suite/suite.go
```diff
@@ -219,6 +225,12 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 		}
 	}
 
+	// check that the test should not be skipped
+	if sets.New(suite.SkipTests...).Has(test.ShortName) {
```
robscott, line 229: Completely understand why you did this because it's exactly what I linked to, but actually intended something slightly different. Initializing this in every test run is equivalent to looping through the list each time. Instead I'd recommend using `sets.String` as the type of the `SkipTests` field so this line can become `if suite.SkipTests.Has(test.ShortName)`. (May also need a nil check).

mlavacca: Yeah, I understand. I've left the slice of strings in the options and used the sets in the `ConformanceTestSuite`, so that the set is initialized only once (at `ConformanceTestSuite` initialization time), but the user can specify the set of tests to skip as a slice of strings through the options.

mlavacca: There is no need to check nil. Even passing a nil slice doesn't cause panic.

robscott:
Thanks @mlavacca!

/lgtm
/approve

PR #1582, opened 2022-12-06
chore: remove mesh leads from * reviews

shaneutt:
This change makes it so that mesh leads wont be pulled in for
every review, but instead just for GEPs and conformance tests
which for the moment are the most relevant to them. This was
done in response to a plea from mesh leads in the community
sync to reduce the amount of reviews they get assigned to for
subjects that might not be relevant.

### DIFF
--- a/conformance/OWNERS
+++ b/conformance/OWNERS
@@ -7,3 +7,4 @@ approvers:
 
 reviewers:
   - gateway-api-maintainers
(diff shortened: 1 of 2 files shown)


keithmattix (approved):

robscott:
Thanks @shaneutt!

/lgtm

Issue #1584, opened 2022-12-07
Update conformance test to assert observedGeneration is bumped


dprotaso:
As a follow up to https://github.com/kubernetes-sigs/gateway-api/issues/1364 we should update conformance tests to assert that each condition's `observeredGeneration` is incremented to match the resource's `metadata.generation`

This is important because as updates occur to a resource's `spec` block (which increments `metadata.generation`) it implies the conditions are stale until the controller acknowledges them.

Also the language in the GEP is a bit confusing

> observedGeneration is an optional field that sets what the metadata.generation field was when the controller last saw a resource. Note that this is optional in the struct, but is required for Gateway API conditions. This will be enforced in the conformance tests in the future.

I would probably just state that `observedGeneration` is required to be set by gateway implementations and ignore mentioning that there's an `+optional` marker

youngnick:
Yeah, reading it back, that is confusing, thanks. I'll do a small PR to update that. But agreed that we should check observedGeneration matches correctly as well.

PR #1586, opened 2022-12-07
Assert observedGeneration is incremented in Status.Conditions.

dprotaso:
**What type of PR is this?**

/kind cleanup
/kind api-change

This PR ensures Gateway Resources and their conditions update observedGeneration as described in the GEP.

Current it tests
- [x] Gateway
- [x] GatewayClass
- [x] HTTPRoute

Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/1584

```release-note
Conformance tests now assert observedGeneration is incremented in Status.Conditions (GatewayClass, Gateway, HTTPRoute)
```

### DIFF
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -82,9 +82,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			// Sanity check
 			kubernetes.GatewayMustHaveLatestConditions(t, updated)
 
-			if original.Generation == updated.Generation {
-				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
-			}
+			require.NotEqual(t, original.Generation, updated.Generation, "generation should change after an update")
 		})
 	},
 }
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -39,7 +39,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 	Manifests:   []string{"tests/gateway-observed-generation-bump.yaml"},
 	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
 
-		gwNN := types.NamespacedName{Name: "observed-generation-bump", Namespace: "gateway-conformance-infra"}
+		gwNN := types.NamespacedName{Name: "gateway-observed-generation-bump", Namespace: "gateway-conformance-infra"}
 
 		t.Run("observedGeneration should increment", func(t *testing.T) {
 			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -48,17 +48,19 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			namespaces := []string{"gateway-conformance-infra"}
 			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
 
-			existing := &v1beta1.Gateway{}
-			err := s.Client.Get(ctx, gwNN, existing)
+			original := &v1beta1.Gateway{}
+			err := s.Client.Get(ctx, gwNN, original)
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// Sanity check
-			kubernetes.GatewayMustHaveLatestConditions(t, existing)
+			kubernetes.GatewayMustHaveLatestConditions(t, original)
 
 			all := v1beta1.NamespacesFromAll
 
+			mutate := original.DeepCopy()
+
 			// mutate the Gateway Spec
-			existing.Spec.Listeners = append(existing.Spec.Listeners, v1beta1.Listener{
+			mutate.Spec.Listeners = append(mutate.Spec.Listeners, v1beta1.Listener{
 				Name:     "alternate",
 				Port:     8080,
 				Protocol: v1beta1.HTTPProtocolType,
@@ -67,7 +69,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 				},
 			})
 
-			err = s.Client.Update(ctx, existing)
+			err = s.Client.Update(ctx, mutate)
 			require.NoErrorf(t, err, "error updating the Gateway: %v", err)
 
 			// Ensure the generation and observedGeneration sync up
@@ -80,12 +82,12 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			// Sanity check
 			kubernetes.GatewayMustHaveLatestConditions(t, updated)
 
-			if existing.Generation == updated.Generation {
+			if original.Generation == updated.Generation {
 				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
 			}
 
 			for _, uc := range updated.Status.Conditions {
-				for _, ec := range existing.Status.Conditions {
+				for _, ec := range original.Status.Conditions {
 					if ec.Type == uc.Type && ec.ObservedGeneration == uc.ObservedGeneration {
 						t.Errorf("Expected status condition %q observedGeneration to change - remained at %v", uc.Type, uc.ObservedGeneration)
 					}
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -85,14 +85,6 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			if original.Generation == updated.Generation {
 				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
 			}
-
-			for _, uc := range updated.Status.Conditions {
-				for _, ec := range original.Status.Conditions {
-					if ec.Type == uc.Type && ec.ObservedGeneration == uc.ObservedGeneration {
-						t.Errorf("Expected status condition %q observedGeneration to change - remained at %v", uc.Type, uc.ObservedGeneration)
-					}
-				}
-			}
 		})
 	},
 }
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayObservedGenerationBump)
+}
+
+var GatewayObservedGenerationBump = suite.ConformanceTest{
+	ShortName:   "GatewayObservedGenerationBump",
+	Description: "A Gateway in the gateway-conformance-infra namespace should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
+	Manifests:   []string{"tests/gateway-observed-generation-bump.yaml"},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+
+		gwNN := types.NamespacedName{Name: "observed-generation-bump", Namespace: "gateway-conformance-infra"}
+
+		t.Run("observedGeneration should increment", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+			defer cancel()
+
+			namespaces := []string{"gateway-conformance-infra"}
+			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
+
+			existing := &v1beta1.Gateway{}
+			err := s.Client.Get(ctx, gwNN, existing)
+			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+
+			// Sanity check
+			if kubernetes.ConditionsHaveLatestObservedGeneration(existing, existing.Status.Conditions) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			all := v1beta1.NamespacesFromAll
+
+			// mutate the Gateway Spec
+			existing.Spec.Listeners = append(existing.Spec.Listeners, v1beta1.Listener{
+				Name:     "alternate",
+				Port:     80,
+				Protocol: v1beta1.HTTPProtocolType,
+				AllowedRoutes: &v1beta1.AllowedRoutes{
+					Namespaces: &v1beta1.RouteNamespaces{From: &all},
+				},
+			})
+
+			err = s.Client.Update(ctx, existing)
+			require.NoErrorf(t, err, "error updating the Gateway: %v", err)
+
+			// Ensure the generation and observedGeneration sync up
+			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
+
+			updated := &v1beta1.Gateway{}
+			err = s.Client.Get(ctx, gwNN, updated)
+			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+
+			// Sanity check
+			if kubernetes.ConditionsHaveLatestObservedGeneration(updated, updated.Status.Conditions) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			if existing.Generation == updated.Generation {
+				t.Errorf("expected generation to change because of spec change - remained at %v", updated.Generation)
+			}
+
+			for _, uc := range updated.Status.Conditions {
+				for _, ec := range existing.Status.Conditions {
+					if ec.Type == uc.Type && ec.ObservedGeneration == uc.ObservedGeneration {
+						t.Errorf("expected status condition %q observedGeneration to change - remained at %v", uc.Type, uc.ObservedGeneration)
+					}
+				}
+			}
+		})
+	},
+}
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -53,7 +53,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// Sanity check
-			if kubernetes.ConditionsHaveLatestObservedGeneration(existing, existing.Status.Conditions) {
+			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(existing) {
 				t.Fatal("Not all the condition's observedGeneration were updated")
 			}
 
@@ -80,7 +80,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// Sanity check
-			if kubernetes.ConditionsHaveLatestObservedGeneration(updated, updated.Status.Conditions) {
+			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(updated) {
 				t.Fatal("Not all the condition's observedGeneration were updated")
 			}
 
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -53,9 +53,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// Sanity check
-			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(existing) {
-				t.Fatal("Not all the condition's observedGeneration were updated")
-			}
+			kubernetes.GatewayMustHaveLatestConditions(t, existing)
 
 			all := v1beta1.NamespacesFromAll
 
@@ -80,9 +78,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// Sanity check
-			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(updated) {
-				t.Fatal("Not all the condition's observedGeneration were updated")
-			}
+			kubernetes.GatewayMustHaveLatestConditions(t, updated)
 
 			if existing.Generation == updated.Generation {
 				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -42,7 +42,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 		gwNN := types.NamespacedName{Name: "observed-generation-bump", Namespace: "gateway-conformance-infra"}
 
 		t.Run("observedGeneration should increment", func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 			defer cancel()
 
 			namespaces := []string{"gateway-conformance-infra"}
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -62,7 +62,7 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			// mutate the Gateway Spec
 			existing.Spec.Listeners = append(existing.Spec.Listeners, v1beta1.Listener{
 				Name:     "alternate",
-				Port:     80,
+				Port:     8080,
 				Protocol: v1beta1.HTTPProtocolType,
 				AllowedRoutes: &v1beta1.AllowedRoutes{
 					Namespaces: &v1beta1.RouteNamespaces{From: &all},
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -85,13 +85,13 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 			}
 
 			if existing.Generation == updated.Generation {
-				t.Errorf("expected generation to change because of spec change - remained at %v", updated.Generation)
+				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
 			}
 
 			for _, uc := range updated.Status.Conditions {
 				for _, ec := range existing.Status.Conditions {
 					if ec.Type == uc.Type && ec.ObservedGeneration == uc.ObservedGeneration {
-						t.Errorf("expected status condition %q observedGeneration to change - remained at %v", uc.Type, uc.ObservedGeneration)
+						t.Errorf("Expected status condition %q observedGeneration to change - remained at %v", uc.Type, uc.ObservedGeneration)
 					}
 				}
 			}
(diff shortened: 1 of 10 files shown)


dprotaso:
Also question: How come there are no existing conformance tests for GatewayClass mutations?

robscott:
>Also question: How come there are no existing conformance tests for GatewayClass mutations?

What would that look like? The conformance tests accept a GatewayClass to run tests against, but I can't think of anything meaningful to actually test on a GatewayClass other than that it is accepted.

dprotaso:
> but I can't think of anything meaningful to actually test on a GatewayClass other than that it is accepted.

Until now with ensuring `observedGeneration` gets bumped.

Thus changing the `parameterRef` or even `description` should trigger this

conformance/utils/kubernetes/helpers.go
```diff
@@ -37,7 +37,7 @@ import (
 	"sigs.k8s.io/gateway-api/conformance/utils/config"
 )
 
-// GatewayRef is a tiny type for specifying an HTTP Route ParentRef without
+// GatewayRef is a tiny type for specifying an HTTP Route ParentRef withouthttps://www.cdw.ca/product/apple-airpods-pro-2nd-generation-true-wireless-earphones-with-mic/7171387?pfm=srh
```
youngnick, line 40: Feels like a paste error maybe? 😂

youngnick (commented):
The approach LGTM. This could go into v0.6.0, or a v0.6.1 update later if need be.

conformance/utils/kubernetes/helpers.go
dprotaso: lol that's hilarious

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,100 @@
...
+			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(existing) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			all := v1beta1.NamespacesFromAll
+
+			// mutate the Gateway Spec
+			existing.Spec.Listeners = append(existing.Spec.Listeners, v1beta1.Listener{
+				Name:     "alternate",
+				Port:     80,
```
dprotaso, line 65: nit: this listener port already exists - should use `8080`

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,100 @@
...
+var GatewayObservedGenerationBump = suite.ConformanceTest{
+	ShortName:   "GatewayObservedGenerationBump",
+	Description: "A Gateway in the gateway-conformance-infra namespace should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
+	Manifests:   []string{"tests/gateway-observed-generation-bump.yaml"},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+
+		gwNN := types.NamespacedName{Name: "observed-generation-bump", Namespace: "gateway-conformance-infra"}
+
+		t.Run("observedGeneration should increment", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
dprotaso, line 45: nit: timeout of 20 seconds might be too short

conformance/tests/gatewayclass-observed-generation-bump.go
```diff
@@ -0,0 +1,90 @@
...
+var GatewayClassObservedGenerationBump = suite.ConformanceTest{
+	ShortName:   "GatewayClassObservedGenerationBump",
+	Description: "A GatewayClass should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
+	Manifests:   []string{"tests/gatewayclass-observed-generation-bump.yaml"},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+
+		gwc := types.NamespacedName{Name: "gatewayclass-observed-generation-bump"}
+
+		t.Run("observedGeneration should increment", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
dprotaso, line 45: increase timeout

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
+
+		acceptedCondition := metav1.Condition{
+			Type:   string(v1beta1.RouteConditionAccepted),
+			Status: metav1.ConditionTrue,
+			Reason: "", // any reason
+		}
+
+		t.Run("observedGeneration should increment", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
dprotaso, line 54: Increase timeout

conformance/tests/httproute-observed-generation-bump.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: HTTPRoute
+metadata:
+  name: observed-generation-bump
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - backendRefs:
+    - name: infra-backend-old
```
dprotaso, line 11: non-existent backed

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+			existing := &v1beta1.HTTPRoute{}
+			err := s.Client.Get(ctx, routeNN, existing)
+			require.NoErrorf(t, err, "error getting HTTPRoute: %v", err)
+
+			// Sanity check
+			if kubernetes.HTTPRouteConditionsHaveLatestObservedGeneration(existing) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			existing.Spec.Rules[0].BackendRefs[0].Name = "infra-backend-new"
```
dprotaso, line 69: non-existent backed

dprotaso:
This is ready for review - I haven't tested this against any implementations yet. I'll do my own testing but I don't think that should get the review.

If any implementors are willing to help test this change that would be much appreciated.

conformance/tests/gateway-observed-generation-bump.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: gateway-observed-generation-bump
```
robscott, line 4: Any way to use an existing Gateway for this?

conformance/tests/gatewayclass-observed-generation-bump.go
```diff
@@ -0,0 +1,90 @@
...
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayClassObservedGenerationBump)
+}
+
+var GatewayClassObservedGenerationBump = suite.ConformanceTest{
```
robscott, line 36: The GatewayClass is part of the environment provided by the person running the conformance tests. I personally don't think we should modify it. Maybe we can ensure that ObservedGeneration in status of the provided GatewayClass matches the generation of the resource though?

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			if existing.Generation == updated.Generation {
+				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
+			}
+
+			for _, up := range updated.Status.Parents {
+				existing := parentStatusForRef(existing.Status.Parents, up.ParentRef)
+				if existing == nil {
+					t.Logf("Observed unexpected new parent ref %#v", up.ParentRef)
```
robscott, line 91: This feels like it should cause the test to fail

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteObservedGenerationBump)
+}
+
+var HTTPRouteObservedGenerationBump = suite.ConformanceTest{
```
robscott, line 38: Can we make this sufficiently generic that the included code can work for every route type? This can be a follow up, just would be nice to have consistency across all Route types.

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+			namespaces := []string{"gateway-conformance-infra"}
+			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
+
+			existing := &v1beta1.HTTPRoute{}
+			err := s.Client.Get(ctx, routeNN, existing)
+			require.NoErrorf(t, err, "error getting HTTPRoute: %v", err)
+
+			// Sanity check
+			if kubernetes.HTTPRouteConditionsHaveLatestObservedGeneration(existing) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
```
robscott, line 66: This applies to all the comments like this - the wording/pluralization feels a bit off here. Maybe `Expected observedGeneration to be updated for all conditions, only x/y were updated.`

robscott (commented):
Thanks @dprotaso!

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,100 @@
...
+	ConformanceTests = append(ConformanceTests, GatewayObservedGenerationBump)
+}
+
+var GatewayObservedGenerationBump = suite.ConformanceTest{
+	ShortName:   "GatewayObservedGenerationBump",
+	Description: "A Gateway in the gateway-conformance-infra namespace should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
+	Manifests:   []string{"tests/gateway-observed-generation-bump.yaml"},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+
+		gwNN := types.NamespacedName{Name: "observed-generation-bump", Namespace: "gateway-conformance-infra"}
```
sunjayBhatia, line 42: ran tests w/ Contour, so far ran into the fact that this name is incorrect, should be `gateway-observed-generation-bump` (or some existing Gateway if possible, similar to other comments)

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,100 @@
...
+
+			namespaces := []string{"gateway-conformance-infra"}
+			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
+
+			existing := &v1beta1.Gateway{}
+			err := s.Client.Get(ctx, gwNN, existing)
+			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+
+			// Sanity check
+			if kubernetes.GatewayConditionsHaveLatestObservedGeneration(existing) {
```
sunjayBhatia, line 56: shouldn't this be as below? (here and in all other similar places)
```suggestion
			if !kubernetes.GatewayConditionsHaveLatestObservedGeneration(existing) {
```

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+			updated := &v1beta1.HTTPRoute{}
+			err = s.Client.Get(ctx, routeNN, updated)
+			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+
+			// Sanity check
+			if kubernetes.HTTPRouteConditionsHaveLatestObservedGeneration(updated) {
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			if existing.Generation == updated.Generation {
```
sunjayBhatia, line 84: I dont think we can do it this way, `Update` will modify the generation of the local variable

if you print the generation before/after the update calls, e.g.

```
			fmt.Println("original route", existing.GetGeneration())

			// Sanity check
			if !kubernetes.HTTPRouteConditionsHaveLatestObservedGeneration(existing) {
				t.Fatal("Not all the condition's observedGeneration were updated")
			}

			existing.Spec.Rules[0].BackendRefs[0].Name = "infra-backend-v2"
			err = s.Client.Update(ctx, existing)
			require.NoErrorf(t, err, "error updating the HTTPRoute: %v", err)

			fmt.Println("original route after update", existing.GetGeneration())

			kubernetes.HTTPRouteMustHaveCondition(t, s.Client, s.TimeoutConfig, routeNN, gwNN, acceptedCondition)

			updated := &v1beta1.HTTPRoute{}
			err = s.Client.Get(ctx, routeNN, updated)
			require.NoErrorf(t, err, "error getting Gateway: %v", err)

			fmt.Println("updated route", updated.GetGeneration())
```

I get 

```
original route 1
original route after update 2
obj generation 2
```

so we should capture the generation of the original resource before we update it and not rely on fetching it from `existing` later

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+				t.Errorf("Expected generation to change because of spec change - remained at %v", updated.Generation)
+			}
+
+			for _, up := range updated.Status.Parents {
+				existing := parentStatusForRef(existing.Status.Parents, up.ParentRef)
```
sunjayBhatia, line 89: This check never works when I run it because I always get an empty list of parent statuses coming from `existing`, I think we need to examine exactly what state of the route ends up in the `expected` object since `Update` will update the content that `expected` points to

```
	// Update updates the given obj in the Kubernetes cluster. obj must be a
	// struct pointer so that obj can be updated with the content returned by the Server.
	Update(ctx context.Context, obj Object, opts ...UpdateOption) error
```

can we change this to just assert on the exact set of parent refs we expect? since we know what the route is we also know what the parentrefs should be from the fixture

conformance/tests/httproute-observed-generation-bump.go
```diff
@@ -0,0 +1,114 @@
...
+				if existing == nil {
+					t.Logf("Observed unexpected new parent ref %#v", up.ParentRef)
+					continue
+				}
+				for _, uc := range up.Conditions {
```
sunjayBhatia, line 94: would it be easier to just modify `HTTPRouteMustHaveCondition` to check generations?

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,100 @@
...
+				t.Fatal("Not all the condition's observedGeneration were updated")
+			}
+
+			all := v1beta1.NamespacesFromAll
+
+			// mutate the Gateway Spec
+			existing.Spec.Listeners = append(existing.Spec.Listeners, v1beta1.Listener{
+				Name:     "alternate",
+				Port:     8080,
```
sunjayBhatia, line 65: at the moment, contour will not mark this update as accepted, we don't support a port other than 80 or 443

sunjayBhatia: can we do something trivial here like you've done with gw class, or add an annotation etc.?

sunjayBhatia: or maybe change an existing listeners allowedroutes section?

conformance/tests/gatewayclass-observed-generation-bump.go
```diff
@@ -0,0 +1,90 @@
...
+	ShortName:   "GatewayClassObservedGenerationBump",
+	Description: "A GatewayClass should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
```
sunjayBhatia, line 38: currently Contour is not reconciling changes to GatewayClasses, just taking their state when they are first created/observed by the Contour Gateway API provisioner due to:

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayClass

```
It is recommended that this resource be used as a template for Gateways. This means that a Gateway is based on the state of the GatewayClass at the time it was created and changes to the GatewayClass or associated parameters are not propagated down to existing Gateways. This recommendation is intended to limit the blast radius of changes to GatewayClass or associated parameters. If implementations choose to propagate GatewayClass changes to existing Gateways, that MUST be clearly documented by the implementation.
```

We haven't yet finished our thinking/implementation on this area so this fails for us

conformance/utils/kubernetes/helpers.go
```diff
@@ -89,6 +95,38 @@ func GWCMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.Timeo
...
+	for _, p := range r.Status.Parents {
+		if !ConditionsHaveLatestObservedGeneration(r, p.Conditions) {
+			return false
+		}
+	}
+	return true
+}
+
+func GatewayClassConditionsHaveLatestObservedGeneration(gc *v1beta1.GatewayClass) bool {
+	if !ConditionsHaveLatestObservedGeneration(gc, gc.Status.Conditions) {
```
sunjayBhatia, line 108: this and below can be simplified to `return ConditionsHaveLatestObservedGeneration(gc, gc.Status.Conditions)`

conformance/tests/httproute-observed-generation-bump.go
sunjayBhatia: similar for status, probably capture the status before we update the resource to use later

conformance/tests/gateway-observed-generation-bump.go
dprotaso: > we don't support a port other than 80 or 443

This restriction is odd - the API doesn't specify any constraints on the port numbers

conformance/tests/gateway-observed-generation-bump.yaml
dprotaso: I would prefer to keep tests isolated - I don't want to change the base gateway if other tests assume it's static.

conformance/tests/gatewayclass-observed-generation-bump.go
dprotaso: > Maybe we can ensure that ObservedGeneration in status of the provided GatewayClass matches the generation of the resource though?

We do this already - this test is to ensure the observedGeneration changes when the `spec` changes

conformance/tests/gateway-observed-generation-bump.go
sunjayBhatia: well I may be misspeaking a little, but at the very least Contour doesn't let you configure multiple HTTP or HTTPS Listeners, only one port is allowed for each at the moment

conformance/tests/httproute-observed-generation-bump.go
dprotaso: Will do as a follow up

conformance/tests/httproute-observed-generation-bump.go
dprotaso: I already did but `HTTPRouteMustHaveCondition` polls until the conditions are in an acceptable state - I don't want to poll here.

conformance/tests/gatewayclass-observed-generation-bump.go
dprotaso: Given that `GatewayClass` has an `Accepted` condition I would expect the observedGeneration of that condition to increment when the GWC spec changes.

conformance/tests/gateway-observed-generation-bump.go
dprotaso: Created an issue to move this discussion off the PR - https://github.com/kubernetes-sigs/gateway-api/issues/1607

conformance/tests/httproute-observed-generation-bump.go
dprotaso: Resolved - thanks for pointing this out

dprotaso:
Hey folks - this is ready for another review.

conformance/utils/kubernetes/helpers.go
```diff
@@ -386,9 +495,14 @@ func HTTPRouteMustHaveCondition(t *testing.T, client client.Client, timeoutConfi
 		}
 
 		parents := route.Status.Parents
-
 		var conditionFound bool
 		for _, parent := range parents {
+
+			if err := ConditionsHaveLatestObservedGeneration(route, parent.Conditions); err != nil {
+				t.Logf("HTTPRoute(controller=%v,ref=%#v) %v", parent.ControllerName, parent, err)
```
sunjayBhatia, line 502: when looking at the test logs, these lines end up being hard to read, e.g.

```
=== RUN   TestConformance/HTTPRouteObservedGenerationBump/observedGeneration_should_increment
    helpers.go:209: Gateways and Pods in gateway-conformance-infra namespaces ready
    helpers.go:502: HTTPRoute(controller=projectcontour.io/gateway-controller,ref=v1beta1.RouteParentStatus{ParentRef:v1beta1.ParentReference{Group:(*v1beta1.Group)(0xc00041e310), Kind:(*v1beta1.Kind)(0xc00041e350), Namespace:(*v1beta1.Na
mespace)(nil), Name:"same-namespace", SectionName:(*v1beta1.SectionName)(nil), Port:(*v1beta1.PortNumber)(nil)}, ControllerName:"projectcontour.io/gateway-controller", Conditions:[]v1.Condition{v1.Condition{Type:"Accepted", Status:"True",
 ObservedGeneration:1, LastTransitionTime:time.Date(2022, time.December, 16, 17, 46, 3, 0, time.Local), Reason:"Accepted", Message:"Accepted HTTPRoute"}}}) expected observedGeneration to be updated for all conditions, only 0/1 were update
d. stale conditions are: Accepted
```

if we want to keep these logs might be good to pare this down a bit (and in other places the same pattern is used)

conformance/tests/gateway-observed-generation-bump.go
```diff
@@ -0,0 +1,90 @@
...
+			kubernetes.NamespacesMustBeAccepted(t, s.Client, s.TimeoutConfig, namespaces)
+
+			updated := &v1beta1.Gateway{}
+			err = s.Client.Get(ctx, gwNN, updated)
+			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+
+			// Sanity check
+			kubernetes.GatewayMustHaveLatestConditions(t, updated)
+
+			if original.Generation == updated.Generation {
```
sunjayBhatia, line 85: could use `require.NotEqual` here

conformance/tests/gatewayclass-observed-generation-bump.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: GatewayClass
+metadata:
+  name: gatewayclass-observed-generation-bump
+spec:
+  controllerName: "{GATEWAY_CONTROLLER_NAME}"
```
robscott, line 6: This seems like a significant change. As I understand it, this requires implementations to watch GatewayClasses matching their controller name and then update status on those. I think in some cases GatewayClass is part of the deployment model and each controller supports exactly one class. I could be wrong, but need to get some more input from other implementations before moving forward with this.

conformance/utils/suite/suite.go
```diff
@@ -145,8 +145,11 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	t.Logf("Test Setup: Ensuring GatewayClass has been accepted")
 	suite.ControllerName = kubernetes.GWCMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.GatewayClassName)
 
+	suite.Applier.GatewayClass = suite.GatewayClassName
+	suite.Applier.ControllerName = suite.ControllerName
```
robscott, line 149: Seems like we need a docs change to go along with this if we're going to move forward with this approach.

robscott (commented):
Thanks @dprotaso! I think the changes for Gateway and HTTPRoute are very reasonable/non-controversial and could likely merge in soon. I think we need to get more feedback from implementations to see if the proposed GWC test is actually implementable though. The alternative would be to modify the provided GatewayClass, but that does not seem ideal either. It _may_ be worth splitting out the GWC side of this since it could take longer to get in, but completely up to you.

robscott:
Adding a hold until we can confirm the proposed GatewayClass tests are broadly implementable.

/hold

conformance/tests/gatewayclass-observed-generation-bump.yaml
dprotaso: I can `t.Skip()` this test for now as not to hold the entire PR

conformance/utils/suite/suite.go
dprotaso: done

conformance/tests/gatewayclass-observed-generation-bump.yaml
dprotaso: I put it behind a support feature flag for now

conformance/utils/kubernetes/helpers.go
dprotaso: Done - just using `namespace/name` of the parent ref

dprotaso:
I created a support featuere `GatewayClassObservedGenerationBump` to allow implementations to opt into that test while it's being discussed.

conformance/tests/gateway-observed-generation-bump.go
sunjayBhatia: this test is passing now on Contour since observed generation is set correctly, the status we get is:

```
status:
  addresses:
  - type: IPAddress
    value: 172.24.255.203
  conditions:
  - lastTransitionTime: "2022-12-16T19:15:41Z"
    message: Gateway is accepted
    observedGeneration: 2
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2022-12-16T19:15:41Z"
    message: Listeners are not valid
    observedGeneration: 2
    reason: ListenersNotValid
    status: "False"
    type: Programmed
  listeners:
  - attachedRoutes: 0
    conditions:
    - lastTransitionTime: "2022-12-16T19:15:41Z"
      message: Only one HTTP port is supported
      observedGeneration: 2
      reason: PortUnavailable
      status: "False"
      type: Accepted
    - lastTransitionTime: "2022-12-16T19:15:41Z"
      message: Invalid listener, see other listener conditions for details
      observedGeneration: 2
      reason: Invalid
      status: "False"
      type: Programmed
    name: alternate
    supportedKinds: []
  - attachedRoutes: 0
    conditions:
    - lastTransitionTime: "2022-12-16T19:15:41Z"
      message: Valid listener
      observedGeneration: 2
      reason: Programmed
      status: "True"
      type: Programmed
    name: http
    supportedKinds:
      kind: HTTPRoute
```

The Gateway as a whole should still be `Accepted=true` since some part of the config is valid, per: https://gateway-api.sigs.k8s.io/geps/gep-1364/#accepted

observedGenerations are all correct, though the status/validity isn't all there

conformance/tests/gatewayclass-observed-generation-bump.go
sunjayBhatia: ah nvm, found where we were short circuiting the status update logic, this should work with Contour we have a small fix to make

sunjayBhatia: https://github.com/projectcontour/contour/pull/4932

youngnick:
This seems good to me, especially with the opt-out for the GatewayClass for now, so I'm going to hit the approve button. I'll leave it to one of the other approvers to give it the tick.

/approve

shaneutt (approved):
In terms of implementability, @pmalek @mlavacca and myself all talked about this and don't have any concerns for the multiple GWAPI implementations @ Kong :+1:

conformance/utils/suite/suite.go
robscott: To clarify, I was trying to say that we need an update to our user-facing docs here: https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/concepts/conformance.md#running-tests

robscott: Actually misunderstood this, we're already reading this from GatewayClass resource so users don't need to configure this.

robscott:
Hey @dprotaso, thanks for all your work on this! The tests LGTM, can you squash this PR into a single commit? (See https://kubernetes.io/docs/contribute/new-content/open-a-pr/#squashing-commits for more info). Once that's done, ping any of us for an LGTM and we'll get it merged.

dprotaso:
Weird isn’t your prow configured to squash on merge?

shaneutt:
> Weird isn’t your prow configured to squash on merge?

Yes, but I think the suggestion here is since this PR is quite large we would like to give you the opportunity to squash into _less_ commits, if not necessarily _one_ commit. That is to say: if you feel it's important for there to be more than one commit to help keep the history more clear and trackable, do a rebase and squash to a history you feel is solid and we can merge that _without_ simply squashing it all into one.

dprotaso:
I have no preference - if you can squash and merge into a single commit
it’ll save me the trouble of rebasing

dprotaso:
The test waits for all gateways to be `Accepted` prior to getting the Gateway and updating it.

The update can fail if the resource is modified after we fetch it _but_ before we perform the update - this is just regular K8s semantics.

My intention of the test is that the gateway resource is stable prior to doing the `Get`/`Update` - it may make more sense that we wait on another condition rather than `Accepted`.

Can you confirm what the exact mutation is?

Issue #1579, opened 2022-12-05
Conformance Tests: TLSRoute Passthrough


candita:
This issue is intended to track conformance test development for TLSRoute Passthrough. Comment below if you're interested in working on covering any of these areas.

Core Capabilities:

- [ ] Valid TLSRoute with 1 backendRef/rule 
- [ ] Valid TLSRoute with 1 backendRef/Rule and matching SNI hostname
- [ ] Valid TLSRoute with 1 backendRef/Rule and matching SNI hostname with wildcard prefix
- [ ] Valid TLSRoute with 1 backendRef/Rule and matching SNI hostname for both Listener and TLSRoute with intersecting hostname
- [ ] Invalid TLSRoute  - no backendRef/rule doesn't become Ready
- [ ] Invalid TLSRoute - invalid backendRef/rule doesn't become Ready (some or all from sublist?)
  - [ ] Missing Rule
  - [ ] Missing Port
  - [ ] Kind, if not nil, is also not Service
  - [ ] Multi ParentRef sections names not set
  - [ ] Multi ParentRef sections names not unique
  - [ ] SNI hostname invalid - IP address
  - [ ] TLSRoute must not be Accepted for any invalid hostnames; ‘Accepted’ condition False in RouteParentStatus
  - [ ] SNI hostname invalid - not RFC1123, not wildcard prefix
  - [ ] Matching SNI hostname for both Listener and TLSRoute without intersecting hostname
  - [ ] TLSRoute hostname that doesn’t match the Listener hostname is ignored
  - [ ] More than 16 hostnames
  - [ ] More than 16 rules
  - [ ] More than 16 backendRef
- [ ] Invalid TLSRoute with 1 backendRef/Rule that doesn’t exist performs no default forwarding
- [ ] Invalid TLSRoute with invalid BackendObjectReference performs no default forwarding (some or all from sublist?)
  - [ ] Name not set
  - [ ] Kind not  set
  - [ ] Namespace (of backend) not set 
  - [ ] Namespace (of backend) missing ReferenceGrant
  - [ ] Port missing when Kind is Service
- [ ] Invalid TLSRoute with filters that provide response must REJECT connection or return 500 status (some or all from sublist?)
  - [ ] Rejections must observe weight, e.g. if 50% might have been delivered, then 50% should be rejected instead

robscott:
[...]

PR #1587, opened 2022-12-08
Issue #1579 TLSRoute Passthrough Conformance Test (normative)

candita:
Issue #1579 TLSRoute Passthrough Conformance Test (normative)

**What type of PR is this?**
/area conformance

This is the first test specified in the TLSRoute Passthrough Conformance Test suite, the normative test.  This test may be used as reference to learn more about designing conformance tests, particularly for TLSRoute Passthrough.

This addresses part of Issue #1579 

```release-note
NONE
```


conformance/base/manifests.yaml
```diff
@@ -201,6 +201,68 @@ spec:
...
+  selector:
+    matchLabels:
+      app: infra-backend-v4
+  template:
+    metadata:
+      labels:
+        app: infra-backend-v4
+    spec:
+      containers:
+      - name: infra-backend-v4
```
howardjohn, line 234: Do we need a 4th backend (with 2 replicas)? Or can we just add TLS to the existing ones?

howardjohn (commented):
Did you test this on any implementations?

conformance/tests/tlsroute-simple-same-namespace.go
```diff
@@ -0,0 +1,89 @@
...
+	Manifests:   []string{"tests/tlsroute-simple-same-namespace.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := v1beta1.Namespace("gateway-conformance-infra")
+		routeNN := types.NamespacedName{Name: "gateway-conformance-infra-test", Namespace: string(ns)}
+		gwNN := types.NamespacedName{Name: "gateway-tlsroute", Namespace: string(ns)}
+		certNN := types.NamespacedName{Name: "tls-passthrough-checks-certificate", Namespace: string(ns)}
+
+		gwAddr, server := kubernetes.GatewayAndTLSRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		if len(server) != 1 {
+			fmt.Errorf("one and only one server required for TLS")
```
robscott, line 53: This error message is a bit confusing. Why do we have this requirement? If this just to make the test simpler? If it's required, I'd fail the test here.

robscott (commented):
Thanks for all the work on this @candita! Looks like a great start.

candita:
> Did you test this on any implementations?

Yes, I tested on Istio, but had to tweak it to make it get past some of the condition naming updates in Gateway API.

conformance/base/manifests.yaml
```diff
@@ -201,6 +201,68 @@ spec:
...
+            value: /etc/secret-volume/crt
+          - name: TLS_SERVER_PRIVKEY
+            value: /etc/secret-volume/key
+        resources:
+          requests:
+            cpu: 10m
+      volumes:
+      - name: secret-volume
+        secret:
+          secretName: tls-passthrough-checks-certificate
```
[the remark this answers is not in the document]
candita: This secret is created in Setup: https://github.com/kubernetes-sigs/gateway-api/blob/1d54add89d6f33f96b6649d6ab69acc2c26b1f07/conformance/utils/suite/suite.go#L156

conformance/base/manifests.yaml
candita: To conveniently use the existing `echoserver` image, I created a new backend with the environment variables `TLS_SERVER_CERT` and `TLS_SERVER_PRIVKEY` for this image to serve certificates.

conformance/tests/tlsroute-simple-same-namespace.go
candita: It is possible to add more than one hostname in the route spec, but this code doesn't handle that scenario.  I can fail the test here to make that clear.

conformance/base/manifests.yaml
candita: Please let me know if this is ok.

conformance/base/manifests.yaml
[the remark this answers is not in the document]
candita: For testing TLS Passthrough, I wanted to set a specific secret and test against it.

conformance/tests/tlsroute-simple-same-namespace.go
candita: Done

conformance/base/manifests.yaml
robscott: I think this could likely be added to an existing backend without breaking any conformance tests. Would be worth trying at least.

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -18,21 +18,23 @@ package roundtripper
...
 	"regexp"
-
 	"sigs.k8s.io/gateway-api/conformance/utils/config"
 )
 
 // RoundTripper is an interface used to make requests within conformance tests.
 // This can be overridden with custom implementations whenever necessary.
 type RoundTripper interface {
 	CaptureRoundTrip(Request) (*CapturedRequest, *CapturedResponse, error)
+	CaptureTLSRoundTrip(Request, []byte, []byte, string) (*CapturedRequest, *CapturedResponse, error)
```
robscott, line 37: Any chance we can reuse CaptureRoundTrip just with a few more options in the Request? At first glance it looks like CaptureTLSRoundTrip includes a lot of the same code.

conformance/utils/kubernetes/helpers.go
```diff
@@ -408,6 +410,36 @@ func HTTPRouteMustHaveParents(t *testing.T, client client.Client, timeoutConfig
 	require.NoErrorf(t, waitErr, "error waiting for HTTPRoute to have parents matching expectations")
 }
 
+// TLSRouteInfo waits for the specified TLSRoute to have parents
+// in status that match the expected parents, and also returns the assigned
+// hostnames of the TLSRoute.  This will cause the test to halt if the
+// specified timeout is exceeded.
+func TLSRouteInfo(t *testing.T, client client.Client, timeoutConfig config.TimeoutConfig, routeName types.NamespacedName, parents []v1beta1.RouteParentStatus, namespaceRequired bool) []v1beta1.Hostname {
```
robscott, line 417: Nit: This feels like it needs another name. It seems like the primary purpose of this function is to wait for the TLSRoute to meet the expected conditions.  It sounds similar to the above `HTTPRouteMustHaveParents`, maybe this could be called `TLSRouteMustHaveParents` and it could just return the TLSRoute?

conformance/utils/kubernetes/helpers.go
```diff
@@ -408,6 +410,36 @@ func HTTPRouteMustHaveParents(t *testing.T, client client.Client, timeoutConfig
...
+	t.Helper()
+
+	var actual []v1beta1.RouteParentStatus
+	var hostnames []v1beta1.Hostname
+
+	waitErr := wait.PollImmediate(1*time.Second, timeoutConfig.HTTPRouteMustHaveParents, func() (bool, error) {
```
robscott, line 423: It seems like we either need to introduce a new timeout name here or make the existing one more generic. At this point I think I'd be tempted to just have a single "RouteMustHaveParents" timeout config, but don't feel strongly.

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -191,6 +193,106 @@ func IsRedirect(statusCode int) bool {
...
+		return nil, nil, fmt.Errorf("unexpected error adding trusted CA: %w", err)
+	}
+
+	if server == "" {
+		return nil, nil, fmt.Errorf("unexpected error, server name required for TLS")
+	}
+
+	// Create the Transport for this provided host, cert, and trusted CA
+	client.Transport = &http.Transport{
+		TLSClientConfig: &tls.Config{
```
robscott, line 222: Not sure if this is strictly necessary, but in ingress conformance we also specify a VerifyPeerCertificate func:

https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/7ee2f3e2d283acb203be862cacd593c71afb1311/test/http/http.go#L75-L87

robscott (commented):
Thanks @candita! I haven't had a chance to run this test yet, but this looks promising from what I can tell.

conformance/conformance_test.go
```diff
@@ -44,14 +44,30 @@ func TestConformance(t *testing.T) {
...
+			}
+		}
+		tests.ConformanceTests = saved
+		if len(saved) == 0 {
+			t.Fatalf("error, no such test: %s", *flags.RunOneTest)
+		}
+	}
+
+	t.Logf("Running conformance tests with %s GatewayClass\n cleanup: %t\n debug: %t\n supported features: [%v]\n exempt features: [%v]\n run one test: %s",
+		*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug, *flags.SupportedFeatures, *flags.ExemptFeatures, *flags.RunOneTest)
```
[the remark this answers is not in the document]
candita: No, I just wasn't familiar with the `TestConformance` prefix so I had to make up a way to do this.  I can remove it and document the prefix.

skriss:
JFYI, I was able to run this test and have it pass against Contour's implementation which has support for TLS passthrough. I haven't reviewed the code yet but will try to do so soon. Thanks!

conformance/utils/kubernetes/helpers.go
candita: RouteMustHaveParents it is.

conformance/utils/roundtripper/roundtripper.go
candita: I did it this way because I didn't want to change the Request, but I'll look into it.

conformance/conformance_test.go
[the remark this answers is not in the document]
candita: Done

conformance/base/manifests.yaml
candita: The TLS backend, infra-backend-v4, is the only one using port/targetPort 443/8443.  It is also the only one setup with env variables TLS_SERVER_CERT and TLS_SERVER_PRIVKEY, which as mentioned [directs it to validate certs in the echoserver app](https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/master/images/echoserver/echoserver.go#L109.).  

It works fine with 1 replica for now, and will be reused for the rest of the TLSRoute Passthrough conformance tests.   I think putting all of this together, that TLSRoute Passthrough should have its own configuration with non-standard ports and env variables.

I will rename it from infra-backend-v4 to tls-backend so that it is clear that this is a specially configured backend.

conformance/utils/roundtripper/roundtripper.go
candita: Done.

conformance/utils/roundtripper/roundtripper.go
candita: I don't think it is strictly necessary unless you have custom cert verification that you want done on the peer.  https://github.com/golang/go/issues/16363 seems to indicate that was VerifyPeerCertificate's intended purpose.  I would expect this to be overkill unless it was also in the implementations for which we are testing conformance.

robscott:
Thanks @candita! This change LGTM, but would like someone else to sign off on this as well.

/cc @skriss @arkodg @shaneutt

Do any of you have time to run through this as well?

/approve

conformance/base/manifests.yaml
robscott: I'm OK with a new backend here, but can we add this TLS listener to an existing Gateway instead of adding a new Gateway just for this test?

conformance/utils/tls/tls.go
```diff
@@ -0,0 +1,70 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/config"
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+)
+
+// requiredConsecutiveSuccesses is the number of requests that must succeed in a row
+// for MakeRequestAndExpectEventuallyConsistentResponse to consider the response "consistent"
+// before making additional assertions on the response body. If this number is not reached within
+// maxTimeToConsistency, the test will fail.
+const requiredConsecutiveSuccesses = 3
```
arkodg, line 32: nit: should this knob live closer to where `maxTimeToConsistency` is defined since its applies to the conformance suite in general

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -43,6 +46,9 @@ type Request struct {
 	Method           string
 	Headers          map[string][]string
 	UnfollowRedirect bool
+	CertPem          []byte
+	KeyPem           []byte
+	Server           string
```
arkodg, line 51: can the Host field be reused ?

site-src/concepts/conformance.md
```diff
@@ -76,13 +76,26 @@ capabilities.
...
+[conformance flags](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/flags/flags.go).
+For example, if you'd like to examine the objects in Kubernetes after your test runs, you can pass a flag to
+suppress cleanup:
+```shell
+go test ./conformance/... -args -gateway-class=istio -cleanup-base-resources=false
+```
+If you'd like to run a single test instead of the entire conformance suite, find your test name
+`(suite.ConformanceTest.ShortName)` and use it like this:
+```shell
+go test ./conformance/... --run TestConformance/YOURTESTNAME --gateway-class=istio
```
gyohuangxin, line 97: We can use flag `supported-features` to specify the tests to run, for example:
```
go test ./conformance/... -supported-features=TLSRoute -gateway-class=my-class
```

sunjayBhatia (commented):
test works with contour and from a quick look seems correct 👍🏽

conformance/utils/tls/tls.go
candita: Sure.  It was used in http.go as well, so I made it a global parameter of timeoutConfig in timeout.go

site-src/concepts/conformance.md
```diff
@@ -76,13 +76,26 @@ capabilities.
...
 to be installed in the cluster and tests will be run against that. A different
-class can be specified with the `--gateway-class` flag along with the
-corresponding test command. For example:
+class can be specified with the `-gateway-class` flag along with the
+corresponding test command. For example, to run a conformance test against Istio,
+run:
 
 ```shell
-go test ./conformance --gateway-class my-class
+go test ./conformance/... -args -gateway-class=istio
```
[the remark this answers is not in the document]
candita: You're right.  When I began this, I wasn't sure if I needed to create a new gatewayClass, and it was helpful for me to get the example for one implementation.  I reworded the rest of this in an attempt to make it clearer to newcomers.  Maybe at some point we could list all the gateway-class names.

conformance/utils/roundtripper/roundtripper.go
candita: I considered `Host` and `Server` to be different.  `Server` is one of the `TLSRoute.Spec.Hostnames` that aligns with `CertPem` and `KeyPem`, and is passed as the `TLSClientConfig.ServerName`.   Since `TLSRoute.Spec.Hostnames` can contain wildcards, it might not match the host.  Ie, we could have a TLS server *.example.com but a request host abc.example.com.

site-src/concepts/conformance.md
candita: In this case, I wanted to explain how to run a single test case by name.  I haven't quite figured out how to use the supported-features flag to my advantage, so I would ask for someone else to document that.

candita:
@robscott @arkodg thanks for the review.  I made some more changes and responded to your comments on changes I didn't make.  Could you PTAL when you get a chance ?

arkodg (approved):
LGTM thanks !

candita:
@skriss @shaneutt this has been /approved by @robscott and LGTM by @arkodg.  It passes for Istio and Contour.  Could one of you please give it the /lgtm and allow to merge?

shaneutt (approved):
/lgtm

PR #1594, opened 2022-12-12
Update ParentRef wording

### LINKED ISSUE #973: Route ParentRef field does not mention ReferencePolicy

youngnick:
**What type of PR is this?**
/kind cleanup
/kind documentation

Notes that ParentRefs that cross namespace boundaries must have some way for the referent to accept/allow the routes.

Fixes #973

```release-note
Clarified that AllowedRoutes or ReferenceGrant must be used by a parent in a cross-namespace ParentRef.
```


youngnick:
I didn't update the v1alpha2 docs; happy to if we think that's worthwhile for this change.

apis/v1beta1/shared_types.go
```diff
@@ -140,6 +140,13 @@ type CommonRouteSpec struct {
 	// case, the list of routes attached to those resources should also be
 	// merged.
 	//
+	// Note that for ParentRefs that cross namespace boundaries, the parent
+	// being referred to must have some mechanism to allow Routes to attach.
+	// Gateway has the AllowedRoutes field, for example. Any other resource
+	// being used as a parent across namespaces MUST either include an AllowedRoutes
+	// field or equivalent functionality, OR use a ReferenceGrant (if the spec
+	// can't have a field added).
```
robscott, line 148: This is a great clarification, a few nits though:

1. This feels like something that belongs on the Namespace field within the ParentRef type definition.
2. As you mentioned, I think it would be good to include this in both v1a2 and v1b1 type definitions. They should be kept identical.
3. Maybe reword this to say something like "Cross-namespace references are only valid if they are explicitly allowed by something in the namespace they are referring to. For example, Gateway has the AllowedRoutes field, and ReferenceGrant provides a generic way to enable any other kind of cross-namespace reference."

robscott (commented):
Thanks @youngnick!

apis/v1beta1/shared_types.go
youngnick: Great wording, thanks. I'll update.

youngnick:
Rebased and updated, PTAL.

robscott:
Thanks @youngnick! LGTM once tests pass.

apis/v1alpha2/shared_types.go
```diff
@@ -25,6 +25,12 @@ import (
 // with "Core" support is Gateway. This API may be extended in the future to
 // support additional kinds of parent resources, such as HTTPRoute.
 //
+// Note that for ParentRefs that cross namespace boundaries, there are specific
+// rules. Cross-namespace references are only valid if they are explicitly
+// allowed by something in the namespace they are referring to. For example,
+// Gateway has the AllowedRoutes field, and ReferenceGrant provides a
+// generic way to enable any other kind of cross-namespace reference.
```
shaneutt, line 32:
```suggestion
// Note that for ParentRefs that cross namespace boundaries there are specific
// rules. Cross-namespace references are only valid if they are explicitly
// allowed by something in the namespace they are referring to. For example:
// Gateway has the AllowedRoutes field, and ReferenceGrant provides a
// generic way to enable any other kind of cross-namespace reference.
```

PR #1596, opened 2022-12-13
gateway.listeners: make optional

howardjohn:
**What type of PR is this?**
/kind api-change

This changes `listeners` field to be optional (i.e. allow 0 length list).

The motivation here is that a Gateway actually provisions underlying infrastructure. This can take a long time, too, depending on the set.
 It may be useful to initially provision a Gateway with
*zero* listeners, and then later add them.

This PR loosens validation to allow not configuring any listeners, with the expectation that underlying infra would be provisioned, then a user would add specific listeners.

```release-note
Gateway.Spec.Listeners field is now optional
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -109,7 +109,7 @@ type GatewaySpec struct {
 	//
 	// +listType=map
 	// +listMapKey=name
-	// +kubebuilder:validation:MinItems=1
+	// +optional
 	// +kubebuilder:validation:MaxItems=64
 	Listeners []Listener `json:"listeners"`
 
--- a/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
@@ -438,14 +438,12 @@ spec:
                   - protocol
                   type: object
                 maxItems: 64
-                minItems: 1
                 type: array
                 x-kubernetes-list-map-keys:
                 - name
                 x-kubernetes-list-type: map
             required:
             - gatewayClassName
-            - listeners
             type: object
           status:
             default:
@@ -1132,14 +1130,12 @@ spec:
                   - protocol
                   type: object
                 maxItems: 64
-                minItems: 1
                 type: array
                 x-kubernetes-list-map-keys:
                 - name
                 x-kubernetes-list-type: map
             required:
             - gatewayClassName
-            - listeners
             type: object
           status:
             default:
--- a/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
+++ b/config/crd/standard/gateway.networking.k8s.io_gateways.yaml
@@ -438,14 +438,12 @@ spec:
                   - protocol
                   type: object
                 maxItems: 64
-                minItems: 1
                 type: array
                 x-kubernetes-list-map-keys:
                 - name
                 x-kubernetes-list-type: map
             required:
             - gatewayClassName
-            - listeners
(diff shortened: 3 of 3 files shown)


hzxuzhonghu:
>The motivation here is that a Gateway actually provisions underlying infrastructure.

Isn't gatewayClass that does provision?

howardjohn:
Gateway class just says what controllers are installed (more or less). Gateway is an actual instance of the infra

robscott:
Need more time to think about this, but I think it's too big of a change for v0.6.0, adding a hold on this until we get that released.

/hold

shaneutt:
I'm not opposed to this, but I do have some general questions about the use case: it sounds like the intention here is to allow a CREATE followed by an UPDATE to add the `Listeners` after the `Gateway` becomes `Ready=True`, is that accurate? Can you help me to better understand the dilemma? What would be the downside of requiring the `Listeners` but then having the controller implementation wait until the underlying provisioning reaches a point where it can accept listeners before _eventually_ applying those listeners, and _then_ calling the `Gateway` `Ready=True` last? It seems to me that it would overall be better to not have the `Gateway` marked as `Ready=True` until it has listeners ready?

hzxuzhonghu:
That can make the `ready` not really ready if a listener can be updated later

howardjohn:
observedGeneration handles the fact that status and spec may be out of
sync, I am not sure this changes that? The same applies when changing from
1->2 listeners

robscott:
@howardjohn I think I see the value of pre-warming infra whenever it's possible, but have a few questions:

1. Why not just provision a Gateway with a dummy listener and change it later? If the infra you're provisioning is a Service type=LB, wouldn't you need to specify ports there?
2. What would this mean for conformance? Not every implementation is going to be able to provision infrastructure without ports/protocols specified?
3. Are there other use cases a Gateway without listeners would enable?

shaneutt:
@howardjohn looks like we have some outstanding questions here, let us know if you need anything from us in order to move this one forward :vulcan_salute:

howardjohn:
> 1. Why not just provision a Gateway with a dummy listener and change it later? If the infra you're provisioning is a Service type=LB, wouldn't you need to specify ports there?

Exposing a dummy port on public internet doesn't seem like a good idea and maybe conflict with security guidelines, etc. There is also a difference between "No listener" and "Listener without routes" (not to mention risk of accidentally attached routes)

> 2. What would this mean for conformance? Not every implementation is going to be able to provision infrastructure without ports/protocols specified?

I think like many other fields - implementation specific. We can declare regardless of support they should report status properly ("scheduled" but 0 listeners, or "not scheduled") .

> 3. Are there other use cases a Gateway without listeners would enable?

Yes, we want to use Gateway to deploy instances of waypoint proxies in ambient mesh. These have an implicit listener that isn't reliable to represent as a listener ("all protocols on all ports"), but also allow users adding explicit listeners using standard GW mechanism (ie port 443 HTTPS with some cert). Requiring a listener is fitting a square peg in a round hole and adds complexity for the user.

youngnick:
Coming back to this one, I should note that changing a previously required field to optional is a breaking API change, sadly. In this case, controllers can currently expect the slice to have at least one entry, changing that is a breaking change. And breaking changes are supposed to have an API revision bump.

I don't necessarily think this is impossible, but I think it would need something to run through the use cases in different deployment models (what should an ingress controller like Contour or Emissary do with an empty list? Should it be accepted? And so on).

@howardjohn What would you like to do here?

howardjohn:
I think at this point it may make sense to close this

PR #1599, opened 2022-12-14
Tightening GRPCRoute validation, fixing gaps in HTTPRoute and Gateway webhook validation

robscott:
**What type of PR is this?**
/kind cleanup

This PR started with a single goal - address feedback from #1538 by filling in missing GRPCRoute webhook validation. As I started that process, I realized that we'd introduced some significant gaps between v1a2 and v1b1 validation among the resources that had graduated. This also updates that logic to share code where possible and provide a consistent experience.

```release-note
- GRPCRoute: Regex validation for Method and Service has been tightened to match GRPC spec.
- GRPCRoute: Webhook validation of GRPCRoute has been expanded to closely match HTTPRoute validation.
- HTTPRoute and Gateway: Gaps between webhook validation for v1alpha2 and v1beta1 have been closed.
- HTTPRoute: Validating webhook now ensures that Exact and Prefix path match values can now only include valid path values per RFC-3986. (RegularExpression path matches are not affected by this change).
```


apis/v1alpha2/validation/grpcroute.go
```diff
@@ -53,9 +72,166 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
+	if m.Service != nil {
+		for _, invalidSeq := range invalidPathSequences {
+			if strings.Contains(*m.Service, invalidSeq) {
+				errs = append(errs, field.Invalid(path.Child("service"), *m.Service, fmt.Sprintf("must not contain %q", invalidSeq)))
+			}
+		}
+
+		for _, invalidSuff := range invalidPathSuffixes {
+			if strings.HasSuffix(*m.Service, invalidSuff) {
+				errs = append(errs, field.Invalid(path.Child("service"), *m.Service, fmt.Sprintf("cannot end with '%s'", invalidSuff)))
```
howardjohn, line 102: should this be a constraint on the CRD instead of the webhook? I think its preferred to have as much in the CRDs as possible?

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -53,9 +72,166 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
+			}
+		}
+	}
+
+	if m.Method != nil {
```
howardjohn, line 107: This is already validated in the CRD:

```

// MethodMatchType specifies the semantics of how gRPC methods and services are compared.
// Valid MethodMatchType values are:
//
// * "Exact"
// * "RegularExpression"
//
// Exact paths MUST be syntactically valid:
//
// - Must not contain `/` character
//
// +kubebuilder:validation:Enum=Exact;PathPrefix;RegularExpression
// +kubebuilder:validation:Enum=Exact;RegularExpression
type GRPCMethodMatchType string
```

(note: it is listed 2x for some reason)

apis/v1alpha2/validation/gateway.go
```diff
@@ -51,40 +31,5 @@ var (
...
-func validateListenerHostname(listeners []gatewayv1a2.Listener, path *field.Path) field.ErrorList {
-	var errs field.ErrorList
-	for i, h := range listeners {
-		if isProtocolInSubset(h.Protocol, protocolsHostnameInvalid) && h.Hostname != nil {
-			errs = append(errs, field.Forbidden(path.Index(i).Child("hostname"), fmt.Sprintf("should be empty for protocol %v", h.Protocol)))
-		}
-	}
-	return errs
+	return gatewayv1b1validation.ValidateGatewaySpec(&gw.Spec, field.NewPath("spec"))
 }
```
shaneutt, line 35: :+1:

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -17,11 +17,26 @@ limitations under the License.
...
+	"net/http"
+	"strings"
+
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
+var (
+	// repeatableGRPCRouteFilters are filter types that can are allowed to be
```
shaneutt, line 30:
```suggestion
	// repeatableGRPCRouteFilters are filter types that are allowed to be
```

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -17,11 +17,26 @@ limitations under the License.
...
+	// repeated multiple times in a rule.
+	repeatableGRPCRouteFilters = []gatewayv1a2.GRPCRouteFilterType{
+		gatewayv1a2.GRPCRouteFilterExtensionRef,
+	}
+
+	invalidPathSequences = []string{"//", "/./", "/../", "%2f", "%2F", "#"}
+	invalidPathSuffixes  = []string{"/..", "/."}
```
shaneutt, line 37: For future readers, it might be helpful to add comments to these explaining what, and maybe even some of the why.

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -53,9 +72,166 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
+		}
+		if m.Headers != nil {
+			errs = append(errs, validateGRPCHeaderMatches(m.Headers, path.Index(i).Child("headers"))...)
+		}
+	}
+	return errs
+}
+
+// validateGRPCMethod validates that no header name
+// is matched more than once (case-insensitive).
```
shaneutt, line 86:
```suggestion
// validateGRPCMethod validates that no header name is matched more than once (case-insensitive),
// and that at least one of service or method was provided.
```

apis/v1alpha2/validation/grpcroute.go
shaneutt: ++

youngnick: The existing regex for method is `^[^\/]*$` (blocking only slash characters), so I guess we could add some stuff in there as well. But the idea here is to be belt-and-suspenders for this.

apis/v1alpha2/validation/grpcroute.go
youngnick: I _think_ the first will be immediately overridden by the second, but it's incorrect anyway, so we should just remove it.

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -53,9 +72,166 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
+		}
+	}
+	for i, action := range filter.Remove {
+		if needsErr, ok := singleAction[strings.ToLower(action)]; ok {
+			if needsErr {
+				errs = append(errs, field.Invalid(path.Child("remove"), filter.Remove[i], "cannot specify multiple actions for header"))
+			}
+			singleAction[strings.ToLower(action)] = false
+		} else {
+			singleAction[strings.ToLower(action)] = true
```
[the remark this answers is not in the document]
youngnick: Yes, seems like this one should also be `string(action.Name)` like the above.

robscott:
Adding a hold since I want to wait for a couple reviews on this one.

/hold

apis/v1alpha2/validation/grpcroute.go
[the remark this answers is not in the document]
robscott: This is actually copied verbatim from HTTPRoute. Although it looks very confusing, it's because the other modifiers are setting a header to a value so there's a name and value whereas here we're just removing headers so only the name is necessary. To make this easier to understand, I've named this var `name` instead of `action` for this section in both GRPCRoute and the original HTTPRoute validation.

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -53,9 +72,166 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
+	}
+
+	for filterType, count := range counts {
+		if count > 1 {
+			errs = append(errs, field.Invalid(path.Child("filters"), filterType, "cannot be used multiple times in the same rule"))
+		}
+	}
+	return errs
+}
+
```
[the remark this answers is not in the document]
robscott: Added, thanks!

apis/v1alpha2/validation/grpcroute.go
robscott: Added, thanks!

apis/v1alpha2/validation/grpcroute.go
robscott: Added, thanks!

apis/v1alpha2/validation/grpcroute.go
robscott: Fixed here and in HTTPRoute, thanks!

apis/v1alpha2/validation/grpcroute.go
robscott: I can't find that line doubled up, but let me know if it's still in main somewhere that I can't find. @howardjohn which specific part are you saying is validated in the CRD? That m.Method can't be nil?

apis/v1alpha2/validation/grpcroute.go
robscott: I've been chatting with @gnossen about the best approach here and for now have tightened up CRD based validation since it appears that a safe regex is simple enough to write. Will hold off until I can get a clear reference to GRPC spec to confirm validations I've proposed here are accurate.

apis/v1alpha2/validation/grpcroute.go
youngnick: Yeah, I can't find the double up either.

youngnick (commented):
/lgtm

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -54,8 +69,129 @@ func validateRuleMatches(matches []gatewayv1a2.GRPCRouteMatch, path *field.Path)
 	var errs field.ErrorList
 	for i, m := range matches {
 		if m.Method != nil && m.Method.Service == nil && m.Method.Method == nil {
-			errs = append(errs, field.Required(path.Index(i).Child("methods"), "one or both of `service` or `method` must be specified"))
-			return errs
+			errs = append(errs, field.Required(path.Index(i).Child("method"), "one or both of `service` or `method` must be specified"))
```
gyohuangxin, line 72: There are two different methods to handle `return errs` in webhook validation now.
1. If the condition matches, return errors immediately to avoid returning too many errors :
https://github.com/kubernetes-sigs/gateway-api/blob/df03bd58002fcf01ee3ffeba97edc1f0c67fb386/apis/v1beta1/validation/common.go#L52-L54
2. Return errors at the end of the function like your code.

I think we should unify the ways to return errors to users, returning all errors or returning the first error. What do you think?

apis/v1alpha2/validation/grpcroute.go
```diff
@@ -54,8 +69,129 @@ func validateRuleMatches(matches []gatewayv1a2.GRPCRouteMatch, path *field.Path)
...
+		// Header names are case-insensitive.
+		counts[strings.ToLower(string(match.Name))]++
+	}
+
+	for name, count := range counts {
+		if count > 1 {
+			errs = append(errs, field.Invalid(path, http.CanonicalHeaderKey(name), "cannot match the same header multiple times in the same rule"))
+		}
+	}
+
```
gyohuangxin, line 98: Please remove the blank line.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -322,7 +322,7 @@ type GRPCMethodMatch struct {
 	// At least one of Service and Method MUST be a non-empty string.
 	// +optional
 	// +kubebuilder:validation:MaxLength=1024
-	// +kubebuilder:validation:Pattern=`^[^\/]*$`
+	// +kubebuilder:validation:Pattern=`^(?i)[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
gnossen, line 325: Pulling from the spec at protobuf.com. method is an `Identifier`. Translating from the BNF, I think it is correct.

Service is a [`TypeName`](https://protobuf.com/docs/language-spec#type-references), so I think there are two minor details off with this service regex though.

- It's underscores that are allowed, not dashes.
- We allow an optional leading dot

So I think the full regex should be the following ([regex101](https://regex101.com/r/bXTWuy/1)):

`^(?i)\.?[a-z_][a-z_0-9]+(\.[a-z_][a-z_0-9]+)*$`

This matches:
- `foo.bar.Baz`
- `.foo.bar.Baz`
- `.foo.bar.Bar_Baz`
- `foo`

robscott: Thanks @gnossen! Updated to match this, PTAL

apis/v1alpha2/validation/grpcroute.go
robscott: If needed we can file a follow up for this since it doesn't feel directly tied to this PR. In general I think we should return as many errors as we reasonably can so users don't have take multiple iterations to fix a set of problems. I think that largely matches what we're already doing, but your specific example may be an example where it makes sense to return both errors.

apis/v1alpha2/validation/grpcroute.go
robscott: I think this may just be a preference thing but I actually prefer a bit more whitespace like this. In this case, this func is copied almost verbatim from HTTPRoute and maintains the same spacing as the original function:

https://github.com/kubernetes-sigs/gateway-api/blob/df03bd58002fcf01ee3ffeba97edc1f0c67fb386/apis/v1beta1/validation/httproute.go#L180-L196

If you feel strongly about where we should have new lines in our functions it may be worth a follow up issue to discuss more.

gnossen (approved):

robscott:
Still need one more LGTM, but removing hold since I think we have consensus on the changes here.

/hold cancel

shaneutt (approved):
/lgtm

jackstine:
I had a question about the validation [here](https://github.com/kubernetes-sigs/gateway-api/commit/cd3a10073bc8243fbc49dd242acdac376323cd53#commitcomment-94626827)

PR #1602, opened 2022-12-14
use Programmed instead of Ready for Gateway print column

skriss:
**What type of PR is this?**
/kind cleanup

Replaces the Ready condition with the Programmed
condition as a print column for Gateways since
Programmed is core conformance and Ready is not.

This seems like a desirable change to me but others
should obviously weigh in too.

```release-note
Display the Programmed condition instead of the Ready condition in the output of `kubectl get gateways`.
```


robscott:
Thanks @skriss! I agree this will be more broadly useful, but will wait for @shaneutt or @youngnick to weigh in.

/approve

youngnick (approved):
Thanks @skriss, nice catch.

shaneutt (approved):
It's going to feel a little weird to see this change for a while, but this is definitely more correct than continuing to have an extended conformance condition be in the short view for all implementations. Nice catch thank you. :+1: 

/lgtm

PR #1603, opened 2022-12-14
docs: dev notes about prio labels

shaneutt:
**What type of PR is this?**

/kind documentation

It was requested by several members in the community in our recent syncs that we start utilizing priority labels more to help members understand better the expected priority of various issues.

I also took a second to rework how the guidance on milestones read to be more inline with how we use milestones today.

### DIFF
--- a/site-src/contributing/devguide.md
+++ b/site-src/contributing/devguide.md
@@ -42,11 +42,26 @@ for this project:
 
 Issues labeled `good first issue` and `help wanted` are especially good for a
 first contribution.
-We use [milestones][gh-milestones] to track our progress towards releases. Looking at our current milestone can help identify our highest priority issues.
+
+We use [milestones][gh-milestones] to track our progress towards releases.
+These milestones are generally labeled according to the [semver][semver]
+release version tag that they represent, meaning that in general we only focus
+on the next release in the sequence until it is closed and the release is
+finished. Only Gateway API maintainers are able to create and attach issues to
+milestones.
+
+We use [priority labels][prio-labels] to help indicate the timing importance of
+resolving an issue, or whether an issue needs more support from its creator or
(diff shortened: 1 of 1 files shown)


robscott (commented):
Thanks @shaneutt! This is great. Maybe also add information about how priority labels are set + a note that only Gateway API Maintainers are able to set milestones.

shaneutt:
> Thanks @shaneutt! This is great. Maybe also add information about how priority labels are set + a note that only Gateway API Maintainers are able to set milestones.

Sounds good, please see the most recent updates.

site-src/contributing/devguide.md
```diff
@@ -42,11 +42,25 @@ for this project:
...
+These milestones are generally labeled according to the [semver][semver]
+release version tag that they represent, meaning that in general we only focus
+on the next release in the sequence until it is closed and the release is
+finished. Only Gateway API maintainers are able to create and attach issues to
+milestones.
+
+We use [priority labels][prio-labels] to help indicate the timing importance of
+resolving an issue, or whether an issue needs more support from its creator or
+the community to be prioritized. These labels can be set with the [/label
+command in PR and issue comments][issue-cmds].
```
robscott, line 56: Although what you have is correct, I think it's slightly easier to use the `/priority` command.
```suggestion
the community to be prioritized. These labels can be set with the [/priority
command in PR and issue comments][issue-cmds]. For example, `/priority important-soon`.
```

youngnick (commented):
/lgtm

site-src/contributing/devguide.md
shaneutt: sg, changed in 65ad53cf92cb8cea74f6d2c65b29238469c136b9

PR #1611, opened 2022-12-18
added conformance checks for port, scheme and path (extended + experimental features)

LiorLieberman:
**What type of PR is this?**
/kind feature
/area conformance

Add tests for extended redirect features (port and scheme) and path experimental feature 

Issue #1103 

NONE


LiorLieberman:
@robscott

conformance/tests/httproute-redirect-port copy.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
```
gyohuangxin, line 1: Is this filename `conformance/tests/httproute-redirect-port copy.yaml` correct?

conformance/utils/http/http.go
```diff
@@ -286,8 +290,27 @@ func CompareRequest(cReq *roundtripper.CapturedRequest, cRes *roundtripper.Captu
...
+			expected.RedirectRequest.Host = cRes.RedirectRequest.Host
+		}
+
+		if expected.RedirectRequest.Port == "" {
+			expected.RedirectRequest.Port = "80"
+		}
+
+		if expected.RedirectRequest.Scheme == "" {
+			expected.RedirectRequest.Scheme = "http"
+		}
```
LiorLieberman, line 303: what do you think about these lines?
I am not 100% sure it is best to take these assumptions

conformance/tests/httproute-redirect-port copy.yaml
LiorLieberman: Fixed. Thanks for catching this!

conformance/tests/httproute-redirect-host-and-status.go
```diff
@@ -50,7 +50,7 @@ var HTTPRouteRequestRedirect = suite.ConformanceTest{
 				StatusCode: 302,
 			},
 			RedirectRequest: &roundtripper.RedirectRequest{
-				Hostname: "example.org",
+				Host: "example.org",
```
shaneutt, line 53: The `HTTPRequestRedirectFilter` still includes the related field `Hostname`, but this is now `Host` so we have two different names for the same concept in similar areas. Can you help me to better understand the decision to change this attribute?

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRedirectPort)
```
shaneutt, line 31: Port is `extended` conformance, but I don't see in this diff where we're adding a `SupportedFeature` for it? :thinking: 

See 0ddb8e68bcdbcb14df6674d20a58d270212963e8 for an example

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+var HTTPRouteRedirectPort = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRedirectPort",
+	Description: "An HTTPRoute with port redirect filter",
```
shaneutt, line 36:
```suggestion
	Description: "An HTTPRoute with a port redirect filter",
```

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,90 @@
...
+				UnfollowRedirect: true,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Port: "8083",
+			},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request: http.Request{
+				Path:             "/host-and-port",
```
shaneutt, line 56: This is a bit of a nit, but for consistency with the other tests and because this area of the tests is meant to be focused on `Port` support, should we use this name instead?

```suggestion
				Path:             "/port-and-host",
```

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,90 @@
...
+			},
+			Response: http.Response{
+				StatusCode: 301,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Port: "8083",
+			},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		},
```
shaneutt, line 78: For completion, should we not also add a `/port-and-host-and-status` test here?

conformance/tests/httproute-redirect-scheme.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRedirectScheme)
```
shaneutt, line 31: Scheme is extended, but similar to port I'm not seeing where we've added a `SupportedFeature` flag for it :thinking: 

See 0ddb8e68bcdbcb14df6674d20a58d270212963e8 for an example

conformance/tests/httproute-redirect-scheme.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+var HTTPRouteRedirectScheme = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRedirectScheme",
+	Description: "An HTTPRoute with scheme redirect filter",
```
shaneutt, line 36:
```suggestion
	Description: "An HTTPRoute with a scheme redirect filter",
```

conformance/tests/httproute-redirect-scheme.go
```diff
@@ -0,0 +1,90 @@
...
+				UnfollowRedirect: true,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Scheme: "https",
+			},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request: http.Request{
+				Path:             "/host-and-scheme",
```
shaneutt, line 56: For consistency, and since this is `Scheme` focused, should this be:

```suggestion
				Path:             "/scheme-and-host",
```

?

conformance/tests/httproute-redirect-scheme.go
```diff
@@ -0,0 +1,90 @@
...
+			},
+			Response: http.Response{
+				StatusCode: 301,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Scheme: "https",
+			},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		},
```
shaneutt, line 78: For completion, should we have `/scheme-and-host-and-status`?

conformance/utils/http/http.go
```diff
@@ -286,8 +290,27 @@ func CompareRequest(cReq *roundtripper.CapturedRequest, cRes *roundtripper.Captu
...
+		if expected.RedirectRequest.Host == "" {
+			expected.RedirectRequest.Host = cRes.RedirectRequest.Host
+		}
+
+		if expected.RedirectRequest.Port == "" {
+			expected.RedirectRequest.Port = "80"
+		}
+
+		if expected.RedirectRequest.Scheme == "" {
+			expected.RedirectRequest.Scheme = "http"
```
shaneutt, line 302: Just for pure convention I would expect this to default to `http`, no?

conformance/utils/http/http.go
shaneutt: The default for scheme and port made me pause slightly. Unless I'm misunderstanding things we're setting defaults here, conventionally you would (almost?) never redirect _to_ `http` so I'm just wondering if we should flip those :thinking:

shaneutt (changes requested):
Looking good!

I do have a few things I would like to see addressed before we merge, but they are minor.

conformance/utils/http/http.go
LiorLieberman: The thing is that I have the checks down there. For example the line 312 - I am checking for scheme. 
so when you would run for example a redirect test for Port when `expected.RedirectRequest.Scheme` is an empty string (because the implementations should not bother to set Scheme in their expected struct unless they want to check it) then we would fail the scheme. Hence I added it. 
My understanding is the by default it is http.

For port it is interesting, I will probably need to change it. Now I have tested Path redirect (changes are not in this PR) with envoyGateway other than istio because istio does not yet support it and I noticed that the response is an empty port, so "80" is not good.

If we decide to set the default to "80" we need to also fill `cRes.RedirectRequest.Port` to "80" but not sure thats desired.

LiorLieberman: The other option I see is setting it to what we get from the expected field if it is an empty string but that also feels a bit iffy - something like this:
```go
if expected.RedirectRequest.Port == "" {
  expected.RedirectRequest.Port = cRes.RedirectRequest.Port
}
if expected.RedirectRequest.Scheme == "" {
  expected.RedirectRequest.Scheme = cRes.RedirectRequest.Scheme
}
```

I do have the same this for expected.RedirectRequest.Path which I think to add something similar:
```go
if expected.RedirectRequest.Path == "" {
  expected.RedirectRequest.Path = cRes.RedirectRequest.Path
}
```

LiorLieberman:
@shaneutt Thanks very much for the feedback, I did miss some important things.
PTAL

conformance/tests/httproute-redirect-host-and-status.go
shaneutt: I still feel a bit odd about changing this name here, but I wont consider it a blocker.

conformance/utils/http/http.go
shaneutt: discussion of this appears to be happening in https://github.com/kubernetes-sigs/gateway-api/pull/1611#discussion_r1051991728 instead, considering this resolved.

conformance/utils/http/http.go
shaneutt: @youngnick and @robscott curious as to your thoughts here.

shaneutt (approved):
Approving so as to no longer block as most of the issues I brought up were resolved, but would still like a review from @youngnick or @robscott for the final lgtm.

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,101 @@
...
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+
+		testCases := []http.ExpectedResponse{{
+			Request: http.Request{
+				Path:             "/port",
+				UnfollowRedirect: true,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Port: "8083",
```
robscott, line 51: It looks like we default to a status code of 302 in HTTPRoute spec: 
https://github.com/kubernetes-sigs/gateway-api/blob/e097dff9f32a3edd0e9465158fa7b48028e3dc86/apis/v1beta1/httproute_types.go#L923-L925

We should expect that code to be returned here.
```suggestion
				Port: "8083",
				StatusCode: 302,
```

conformance/tests/httproute-redirect-port.go
```diff
@@ -0,0 +1,101 @@
...
+			},
+			Namespace: ns,
+		}, {
+			Request: http.Request{
+				Path:             "/port-and-host",
+				UnfollowRedirect: true,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Host: "example.org",
+				Port: "8083",
```
robscott, line 61: Same as above:
```suggestion
				Port: "8083",
				StatusCode: 302,
```

conformance/tests/httproute-redirect-port.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /port-and-host-and-status
+    filters:
+    - type: RequestRedirect
+      requestRedirect:
+        port: 8083
+        hostname: example.org
+        statusCode: 301
```
robscott, line 45: Would be useful for one of these to explicitly set 302 to ensure that works as well:
```suggestion
        statusCode: 302
```

conformance/tests/httproute-redirect-scheme.go
```diff
@@ -0,0 +1,101 @@
...
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+
+		testCases := []http.ExpectedResponse{{
+			Request: http.Request{
+				Path:             "/scheme",
+				UnfollowRedirect: true,
+			},
+			RedirectRequest: &roundtripper.RedirectRequest{
+				Scheme: "https",
```
robscott, line 51: Same comments as above, we should expect a 302 here and elsewhere in the test where it has not been specified. Even if we could make that a default somewhere, I'd rather explicitly state that expectation in each test case.

conformance/tests/httproute-redirect-scheme.yaml
```diff
@@ -0,0 +1,46 @@
...
+        scheme: "https"
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /scheme-and-status
+    filters:
+    - type: RequestRedirect
+      requestRedirect:
+        scheme: "https"
+        statusCode: 301
```
robscott, line 35: One of the status codes in this test should probably be 302 for better coverage.

robscott (commented):
Thanks @LiorLieberman! A few nits but otherwise LGTM.

conformance/utils/http/http.go
robscott: My preference is to avoid defaults here, specify what we care about in the test, and only check for that. In this case that would mean lots of repetitive config, but I think it would more clearly communicate what we actually expect from an implementation.

LiorLieberman: I have made some refactoring - PTAL

LiorLieberman:
@shaneutt @robscott 

I have made a small refactor

* Removed 302 default as per Rob's preference.
* added tests for path redirect:
** tested both with envoygateway and istio (after I have contributed the path redirect support to them(https://github.com/istio/istio/pull/42538)

* taking defaults from the original request if they arent set.

PTAL

conformance/utils/http/http.go
```diff
@@ -286,8 +286,38 @@ func CompareRequest(cReq *roundtripper.CapturedRequest, cRes *roundtripper.Captu
...
+		}
+
+		if expected.RedirectRequest.Scheme == "" {
+			expected.RedirectRequest.Scheme = req.URL.Scheme
+		}
+
+		if expected.RedirectRequest.Path == "" {
+			expected.RedirectRequest.Path = req.URL.Path
+		}
+		if expected.RedirectRequest.Host != cRes.RedirectRequest.Host {
```
robscott, line 307: Tiny nit: All the conditions above this are setting reasonable defaults, all the conditions below are checking values. I'd expect there to be more separation, maybe even comments, but in this case, this transition is the only one without an extra newline.

conformance/utils/suite/suite.go
```diff
@@ -54,6 +54,15 @@ const (
...
 	// This option indicates GatewayClass will update the observedGeneration in it's conditions when reconciling
 	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
+
+	// This option indicates support for HTTPRoute port redirect (extended conformance).
+	SupportHTTPRoutePortRedirect SupportedFeature = "HTTPRoutePortRedirect"
+
+	// This option indicates support for HTTPRoute scheme redirect (extended conformance).
+	SupportHTTPRouteSchemeRedirect SupportedFeature = "HTTPRouteSchemeRedirect"
+
+	// This option indicates support for HTTPRoute scheme redirect (extended conformance).
```
robscott, line 64:
```suggestion
	// This option indicates support for HTTPRoute path redirect (extended conformance).
```

conformance/tests/httproute-redirect-port.yaml
robscott: I'd still like this to exist somewhere even though 302 is currently the default value, that could theoretically change in a future API version, just replacing one of the 301s with a 302 would be ideal.

robscott (commented):
Thanks @LiorLieberman, a few tiny nits but otherwise LGTM.

conformance/utils/http/http.go
LiorLieberman: added a separate function

robscott:
Thanks @LiorLieberman!

/lgtm

PR #1617, opened 2022-12-21
conformance: do not use (and mutate) DefaultClient

howardjohn:
This is a global variable. Since we import the tests and then run other tests, this broke our other tests which depend on DefaultClient.

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```


shaneutt (approved):
At one point I could have sworn there was a linter for this... I'm unable to find it now :thinking: 

Anyhow, yes thank you. :+1: 

/lgtm

robscott:
Thanks!

/lgtm

PR #1622, opened 2022-12-28
Http rewrite conformance

LiorLieberman:
**What type of PR is this?**
/kind feature
/area conformance

Add tests for http rewrite 

Issue #1103 

```
- Added: conformance tests for http rewrite host and path filters.
```

### DIFF
--- a/conformance/tests/httproute-rewrite-host.go
+++ b/conformance/tests/httproute-rewrite-host.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRewriteHost)
+}
+
+var HTTPRouteRewriteHost = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRewriteHost",
+	Description: "An HTTPRoute with hostname rewrite filter",
+	Manifests:   []string{"tests/httproute-rewrite-host.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+		routeNN := types.NamespacedName{Name: "rewrite", Namespace: ns}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+
+		testCases := []http.ExpectedResponse{
+			{
+				Request: http.Request{
+					Path: "/one",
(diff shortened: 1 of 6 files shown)


conformance/tests/httproute-rewrite-path.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: rewrite
```
robscott, line 4: Recommend using a unique resource name here so we don't run into race conditions in between tests where a route of the same name is being deleting around the same time this is being created.

conformance/tests/httproute-rewrite-path.go
```diff
@@ -0,0 +1,79 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRewritePath)
+}
+
+var HTTPRouteRewritePath = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRewritePath",
+	Description: "An HTTPRoute with path rewrite filter",
```
robscott, line 35: Need to add feature gates for this.

conformance/tests/httproute-rewrite-host.go
```diff
@@ -0,0 +1,82 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRewriteHost)
+}
+
+var HTTPRouteRewriteHost = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRewriteHost",
+	Description: "An HTTPRoute with hostname rewrite filter",
```
robscott, line 35: Need feature gates for this.

conformance/tests/httproute-rewrite-path.go
```diff
@@ -0,0 +1,79 @@
...
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+		routeNN := types.NamespacedName{Name: "rewrite", Namespace: ns}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+
+		testCases := []http.ExpectedResponse{
+			{
+				Request: http.Request{
+					Path: "/prefix/one/two",
```
robscott, line 46: It might be useful to still set the hostname in the request and confirm that it is not modified by the rewrite.

robscott (commented):
Thanks @LiorLieberman!

/ok-to-test

conformance/tests/httproute-rewrite-path.go
LiorLieberman: I have concerns about doing it. I understand it is more explicit but it adds unnecessary values to the test cases that are just confusing in my opinion. 

Also, there are endless things we can add here to check that they are not modified; headers, method etc etc.

I think we shouldn't add hostname for now since we want to check only path rewrite in this file

robscott: That's fair, will not block on this

robscott:
Thanks @LiorLieberman! Looks like this has a merge conflict now that other rewrite PR merged. One other note, I think we should actually add release notes for conformance test changes since we've consistently been including conformance test changes in our changelog and release notes. Will LGTM once you rebase and add release note.

/approve

LiorLieberman:
@robscott I have added a note in the PR description - this is how we add release notes right?

robscott:
Thanks @LiorLieberman!

/lgtm

conformance/tests/httproute-rewrite-path.yaml
```diff
@@ -0,0 +1,35 @@
...
+        type: PathPrefix
+        value: /full/one
+    filters:
+    - type: URLRewrite
+      urlRewrite:
+        path:
+          type: ReplaceFullPath
+          replaceFullPath: /one
+    backendRefs:
+    - name: infra-backend-v1
```
skriss, line 34: I believe this should be `infra-backend-v2` given https://github.com/LiorLieberman/gateway-api/blob/7dc2bcaed9f75a4cf7e6908c4b9262b6c3609fec/conformance/tests/httproute-rewrite-path.go#L66

LiorLieberman: good catch, thanks! 

opened https://github.com/kubernetes-sigs/gateway-api/pull/1628 to fix it

PR #1624, opened 2022-12-29
conformance: add attachedRoutes to conformance tests

### LINKED ISSUE #1543: conformance: Update conformance tests to check for correct `attachedRoutes` value
Currently, we mandate behavior for `attachedRoutes` in Listener status, but don't check for it in conformance tests. We should do that.

**Why this is needed**:
Conformance tests should test all mandated behaviors.

Todo list:
- [ ] Add a helper to check that a given Gateway Listener has a given value for `attachedRoutes`
- [ ] Update existing tests to check `attachedRoutes` - this one is a fair amount of work, I think,

/area conformance
/help-wanted

ChaningHwang:
**What type of PR is this?**
/kind feature
/area conformance

Add attachedRoutes to conformance tests.

Fixes #1543

NONE

### DIFF
--- a/conformance/tests/gateway-invalid-tls-certificateref.go
+++ b/conformance/tests/gateway-invalid-tls-certificateref.go
@@ -47,7 +47,7 @@ var GatewayInvalidTLSConfiguration = suite.ConformanceTest{
 				Status: metav1.ConditionFalse,
 				Reason: string(v1beta1.ListenerReasonInvalidCertificateRef),
 			}},
-			AttachedRoutes: 1,
+			AttachedRoutes: 0,
 		}}
 
 		testCases := []struct {
--- a/conformance/tests/gateway-invalid-tls-certificateref.go
+++ b/conformance/tests/gateway-invalid-tls-certificateref.go
@@ -47,6 +47,7 @@ var GatewayInvalidTLSConfiguration = suite.ConformanceTest{
 				Status: metav1.ConditionFalse,
 				Reason: string(v1beta1.ListenerReasonInvalidCertificateRef),
 			}},
+			AttachedRoutes: 1,
 		}}
 
 		testCases := []struct {
--- a/conformance/tests/gateway-invalid-route-kind.go
+++ b/conformance/tests/gateway-invalid-route-kind.go
@@ -46,6 +46,7 @@ var GatewayInvalidRouteKind = suite.ConformanceTest{
 					Status: metav1.ConditionFalse,
 					Reason: string(v1beta1.ListenerReasonInvalidRouteKinds),
 				}},
+				AttachedRoutes: 0,
 			}}
 
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
@@ -64,6 +65,7 @@ var GatewayInvalidRouteKind = suite.ConformanceTest{
 					Status: metav1.ConditionFalse,
 					Reason: string(v1beta1.ListenerReasonInvalidRouteKinds),
 				}},
+				AttachedRoutes: 0,
 			}}
 
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
--- a/conformance/tests/gateway-secret-invalid-reference-grant.go
+++ b/conformance/tests/gateway-secret-invalid-reference-grant.go
@@ -51,6 +51,7 @@ var GatewaySecretInvalidReferenceGrant = suite.ConformanceTest{
 					Status: metav1.ConditionFalse,
 					Reason: string(v1beta1.ListenerReasonRefNotPermitted),
 				}},
+				AttachedRoutes: 0,
 			}}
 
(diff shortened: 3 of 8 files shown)


robscott:
Thanks @ChaningHwang!

/ok-to-test

conformance/tests/gateway-invalid-tls-certificateref.go
```diff
@@ -47,6 +47,7 @@ var GatewayInvalidTLSConfiguration = suite.ConformanceTest{
 				Status: metav1.ConditionFalse,
 				Reason: string(v1beta1.ListenerReasonInvalidCertificateRef),
 			}},
+			AttachedRoutes: 1,
```
robscott, line 50: I actually would expect `AttachedRoutes` to be 0 in this case. Did you happen to test this on any implementation? If so, are there any that are currently setting the value to 1 here? If a TLS Cert is invalid, I would not expect Routes to be considered attached. Maybe there would be an exception in the case of a Gateway with a cert that was previously valid, but I'm not seeing that here.

robscott (commented):
Thanks @ChaningHwang! This looks like a good start. Although this doesn't have to block merging this PR, I'd like to have this kind of logic in place across more tests before marking the linked issue as fixed.

conformance/tests/gateway-invalid-tls-certificateref.go
ChaningHwang: Thanks @robscott for the feedback! I was testing this on istio implementation. I am not very sure if there is anything set the value to 1. I may need to look into the setting of istio implementation.

frankbu: > If a TLS Cert is invalid, I would not expect Routes to be considered attached.

@robscott I agree it should be 0 here, since there are no routes attached in this test case, but in general shouldn't the `attachedRoutes` be based on how many routes are connected to the gateway, independent of whether or not the gateway itself is currently ready/valid?

robscott: The current spec for `attachedRoutes` is pretty light here, but it does require the Routes to have been "successfully attached to this listener":

https://github.com/kubernetes-sigs/gateway-api/blob/46d40889ea70291da0614274931c2ffacdc0b612/apis/v1beta1/gateway_types.go#L671-L673

In this case I can think there's some ambiguity in terms of what happens when the Listener itself is invalid but nothing is invalid with the Listener->Route relationship. This can be worked out in a separate PR, but I've been thinking this is how we should interpret "AttachedRoutes:

>AttachedRoutes represents the total number of Routes that have been successfully attached to this Listener. For the purposes of this condition, Routes MUST only be considered "Attached" if the Listener has both the "Accepted" and "ResolvedRefs" conditions set to "True" AND the Route has the "Accepted" condition set to "True". If one or more  of these Listener conditions transitions from "True" to "False", this field SHOULD represent the number of Routes that have been attached to the Listener before this change. Routes that are intentionally detached from a Gateway (either through changes to AllowedRoutes on the Listener or ParentRefs on the Routes) MUST NOT be counted.

@howardjohn do you know how Istio is interpreting this today? At first glance, it looks like it's considering any Route that would be attached, regardless of the validity of the Listener itself.

howardjohn: IMO, the Listener<->Route attachment is valid even if the listener is not valid - or at least that was my original thought. The revised description sounds good from a user perspective but its not implementable (for us, anyways) because it requires looking back in history instead of reconciling the current state of the world

frankbu: @robscott It seems simpler to me to keep them independent. Otherwise, it's confusing when a route status says the route is "Accepted", but it's not acknowledged as attached in the gateway. Maybe the route should not be "Accepted" if the listener is not ready, but I think the doc would then need to explicitly say that. The current doc seems to imply that a route could be accepted/attached to a gateway, but still not "available" depending on the gateway/listener status, which seems like a clean and simple approach.
https://github.com/kubernetes-sigs/gateway-api/blob/46d40889ea70291da0614274931c2ffacdc0b612/apis/v1beta1/shared_types.go#L315-L316

robscott: x-ref https://github.com/kubernetes-sigs/gateway-api/issues/1626 for further discussion.

ChaningHwang: Thanks for the discussion! Updated to 0 here. Sorry for the late response. I was busy taking care of my newborn in the past month.

shaneutt (commented):
At the time I'm reading this, we're only testing for conditions where `AttachedRoutes` is expected to be `0`, but to cover `AttachedRoutes` in conformance we should in fact include some tests where it would be 1 and more.

frankbu (approved):
This doesn't address the question of whether or not the attachedRoutes count should include routes that are attached to invalid not ready Gateways, but maybe the spec isn't going to mandate that one way or another?

Otherwise LGTM

shaneutt (approved):
Thanks for adding the positive cases, and thanks for your contribution! Targeting this for `v0.7.0`.

/lgtm

PR #1625, opened 2023-01-04
Doc: SecretObjectReference use "Secret" as Kind

kehao95:
According to the doc:
> A single CertificateRef to a Kubernetes Secret has “Core” support. Implementations MAY choose to support attaching multiple certificates to a Listener, but this behavior is implementation-specific.

and the line below
```
+kubebuilder:default=Secret
```
It looks like Secret should be the default implementation of the SecretObjectReference. And `HTTPRoute` or `Service` is certainly not a secret store.

**What type of PR is this?**

/kind documentation

### DIFF
--- a/apis/v1beta1/object_reference_types.go
+++ b/apis/v1beta1/object_reference_types.go
@@ -53,7 +53,7 @@ type SecretObjectReference struct {
 	// +kubebuilder:default=""
 	Group *Group `json:"group"`
 
-	// Kind is kind of the referent. For example "HTTPRoute" or "Service".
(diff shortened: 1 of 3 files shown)


robscott:
Thanks @kehao95! I think you're also going to need to run `make generate` locally and commit the changes to get our tests to pass.

/ok-to-test

kehao95:
Hi @robscott, thank you for the guide. I've committed the generated changes. Should be good to merge now.

robscott:
Thanks @kehao95!

/lgtm
/approve

PR #1627, opened 2023-01-05
conformance: Add additional path matching edge cases

sunjayBhatia:
**What type of PR is this?**
/kind cleanup

Adds some test cases (from Contour e2e tests) for edge cases in path matching.

In consolidating this set of test cases: https://github.com/projectcontour/contour/blob/76521b7dbc562815083ec207abef686bc02614ce/test/e2e/gateway/path_condition_match_test.go#L69-L84

We found we can eliminate all but a few that do not have overlap with the upstream conformance tests, so added tests for the following cases:
- trailing slash on path prefix should match
- path prefixes are path segment matches not string prefix matches
- test to ensure a path containing the prefix is not matched

N/A

```release-note
NONE
```


skriss (approved):

robscott:
Great additions, thanks @sunjayBhatia!

/lgtm
/approve

PR #1642, opened 2023-01-11
Add `all-features` flag to add ability to enable all supported feature conformance tests.

### LINKED ISSUE #1638: Add ability to enable all Supported Feature Conformance tests
A way to enable all core and extended conformance tests without explictly specifying [each supported feature suite name ](https://github.com/kubernetes-sigs/gateway-api/blob/d950009b34786a2fda3f5e8837c2de31b54b483d/conformance/utils/suite/suite.go#L34)

**Why this is needed**:
To be able to run the complete and latest set of conformance tests in the implementation repo (e.g. in [envoy-gateway](https://github.com/envoyproxy/gateway/blob/8dd60b1a3aaa8a93504aae0efd4d50315a43bd96/test/conformance/conformance_test.go#L56)) w/o having to manually track which new suite was added to the gateway-api repo

Relates to https://kubernetes.slack.com/archives/CR0H13KGA/p1668094311024819

gyohuangxin:
**What type of PR is this?**

/kind feature

It adds `all-features` flag to add ability to enable all supported feature conformance tests. For example:

```
go test ./conformance --gateway-class istio --debug --all-features
```

Fixes #1638 

```release-note
1. Add `all-features` flag to enable all supported feature conformance tests.
```

### DIFF
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -24,9 +24,10 @@ import (
 )
 
 var (
-	GatewayClassName     = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
-	ShowDebug            = flag.Bool("debug", false, "Whether to print debug logs")
-	CleanupBaseResources = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
-	SupportedFeatures    = flag.String("supported-features", "", "Supported features included in conformance tests suites")
-	ExemptFeatures       = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
+	GatewayClassName            = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
+	ShowDebug                   = flag.Bool("debug", false, "Whether to print debug logs")
+	CleanupBaseResources        = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
+	SupportedFeatures           = flag.String("supported-features", "", "Supported features included in conformance tests suites")
+	ExemptFeatures              = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
+	EnableAllSupportedFeatures  = flag.Bool("all-features", false, "Whether to enable all supported feature conformance tests")
 )
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -29,5 +29,5 @@ var (
 	CleanupBaseResources       = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
 	SupportedFeatures          = flag.String("supported-features", "", "Supported features included in conformance tests suites")
 	ExemptFeatures             = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
-	EnableAllSupportedFeatures = flag.Bool("all-features", false, "Whether to enable all supported feature conformance tests")
+	EnableAllSupportedFeatures = flag.Bool("all-features", false, "Whether to enable all supported features for conformance tests")
 )
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -24,10 +24,10 @@ import (
 )
 
 var (
-	GatewayClassName            = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
-	ShowDebug                   = flag.Bool("debug", false, "Whether to print debug logs")
-	CleanupBaseResources        = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
-	SupportedFeatures           = flag.String("supported-features", "", "Supported features included in conformance tests suites")
-	ExemptFeatures              = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
-	EnableAllSupportedFeatures  = flag.Bool("all-features", false, "Whether to enable all supported feature conformance tests")
+	GatewayClassName           = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
+	ShowDebug                  = flag.Bool("debug", false, "Whether to print debug logs")
+	CleanupBaseResources       = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
+	SupportedFeatures          = flag.String("supported-features", "", "Supported features included in conformance tests suites")
+	ExemptFeatures             = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
+	EnableAllSupportedFeatures = flag.Bool("all-features", false, "Whether to enable all supported feature conformance tests")
 )
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -243,7 +243,7 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 	// Check that all features exercised by the test have been opted into by
 	// the suite.
 	for _, feature := range test.Features {
-		if suite.SupportedFeatures.Has(feature) {
+		if !suite.SupportedFeatures.Has(feature) {
 			t.Skipf("Skipping %s: suite does not support %s", test.ShortName, feature)
 		}
 	}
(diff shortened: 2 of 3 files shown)


gyohuangxin:
@robscott @arkodg Could you review this PR? Thanks.

conformance/utils/flags/flags.go
```diff
@@ -24,9 +24,10 @@ import (
...
-	ShowDebug            = flag.Bool("debug", false, "Whether to print debug logs")
-	CleanupBaseResources = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
-	SupportedFeatures    = flag.String("supported-features", "", "Supported features included in conformance tests suites")
-	ExemptFeatures       = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
+	GatewayClassName           = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
+	ShowDebug                  = flag.Bool("debug", false, "Whether to print debug logs")
+	CleanupBaseResources       = flag.Bool("cleanup-base-resources", true, "Whether to cleanup base test resources after the run")
+	SupportedFeatures          = flag.String("supported-features", "", "Supported features included in conformance tests suites")
+	ExemptFeatures             = flag.String("exempt-features", "", "Exempt Features excluded from conformance tests suites")
+	EnableAllSupportedFeatures = flag.Bool("all-features", false, "Whether to enable all supported feature conformance tests")
```
arkodg, line 32: flag name lgtm !

conformance/utils/suite/suite.go
```diff
@@ -77,6 +77,23 @@ var StandardCoreFeatures = map[SupportedFeature]bool{
 	SupportReferenceGrant: true,
 }
 
+// AllFeatures contains all the supported features and can be used to run all
+// conformance tests with `all-features` flag.
+var AllFeatures = map[SupportedFeature]bool{
```
arkodg, line 82: is there any way a contributor defines the suite name once ? I fear the contributor will add it the constant w/o adding it to this map, and the burden to ensure its in sync will fall on the maintainer

gyohuangxin: Yes, you are right, developers need to modify two places after the constant changes. But I can't find a simple way to define the suite name once in this codebase. Or I can add a comment to notice developer/code reviewer to ensure they are in sync?
@robscott Can you help with some advice?

gyohuangxin:
@robscott Could you review it? Thanks

conformance/utils/suite/suite.go
arkodg: I can't think of a better solution rn, can you add a comment to ensure they are in sync, TIA !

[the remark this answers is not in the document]
gyohuangxin: IMO, there is no native set structure in golang, do you mean the `map[SupportedFeature]struct{}`? Alsowe should pass `AllFeatures` to suite's `SupportedFeatures`, so we should make them the same type.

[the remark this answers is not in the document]
shaneutt: Sets are available in apimachinery:

https://github.com/kubernetes/apimachinery/tree/master/pkg/util/sets

e.g.:

```go
import "k8s.io/apimachinery/pkg/utils/sets"

var AllFeatures = sets.NewString(
	SupportReferenceGrant,
	SupportTLSRoute,
	SupportHTTPRouteQueryParamMatching,
	SupportHTTPRouteMethodMatching,
	SupportHTTPResponseHeaderModification,
	SupportRouteDestinationPortMatching,
	SupportGatewayClassObservedGenerationBump,
	SupportHTTPRoutePortRedirect,
	SupportHTTPRouteSchemeRedirect,
	SupportHTTPRoutePathRedirect,
	SupportHTTPRouteHostRewrite,
	SupportHTTPRoutePathRewrite,
)
```

```go
if AllFeatures.Has(SupportReferenceGrant) {
```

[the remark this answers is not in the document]
shaneutt: Yes the implementation ultimately uses a map, so the benefit here is for the developers which I think is a fair thing to trade for, however I don't consider this suggestion a blocker I'll let others weigh in.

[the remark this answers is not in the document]
gyohuangxin: Understood, if others also want the benefit, I can implement it.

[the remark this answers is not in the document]
robscott: Now that the k8s sets util supports generics, I think it would be best to move both this and AllFeatures over to use that approach.

robscott (commented):
Thanks @gyohuangxin! Agree with moving to k8s sets util, but otherwise this LGTM.

arkodg (approved):
lgtm, thanks for adding this !

conformance/utils/suite/suite.go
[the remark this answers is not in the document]
gyohuangxin: > Now that the k8s sets util supports generics, I think it would be best to move both this and AllFeatures over to use that approach.

@robscott @shaneutt I tried to move it to K8s sets utils: https://github.com/kubernetes-sigs/gateway-api/pull/1642/commits/b85174a868892238ff589643da738f492fda4e35, can you review if it is correct?

conformance/utils/suite/suite.go
```diff
@@ -74,8 +74,28 @@ const (
...
+	SupportHTTPRouteMethodMatching:            {},
+	SupportHTTPResponseHeaderModification:     {},
+	SupportRouteDestinationPortMatching:       {},
+	SupportGatewayClassObservedGenerationBump: {},
+	SupportHTTPRoutePortRedirect:              {},
+	SupportHTTPRouteSchemeRedirect:            {},
+	SupportHTTPRoutePathRedirect:              {},
+	SupportHTTPRouteHostRewrite:               {},
+	SupportHTTPRoutePathRewrite:               {},
 }
```
shaneutt, line 99: It should be possible to define these this way, right?
```suggestion
var StandardCoreFeatures = sets.New(
	SupportReferenceGrant,
)

// AllFeatures contains all the supported features and can be used to run all
// conformance tests with `all-features` flag.
//
// Note that the AllFeatures must in sync with defined features when the
// feature constants change.
var AllFeatures = sets.New(
	SupportReferenceGrant,
	SupportTLSRoute,
	SupportHTTPRouteQueryParamMatching,
	SupportHTTPRouteMethodMatching,
	SupportHTTPResponseHeaderModification,
	SupportRouteDestinationPortMatching,
	SupportGatewayClassObservedGenerationBump,
	SupportHTTPRoutePortRedirect,
	SupportHTTPRouteSchemeRedirect,
	SupportHTTPRoutePathRedirect,
	SupportHTTPRouteHostRewrite,
	SupportHTTPRoutePathRewrite,
}
```

gyohuangxin: @shaneutt Yes, I've changed like this. Thanks.

conformance/utils/suite/suite.go
```diff
@@ -129,12 +150,14 @@ func New(s Options) *ConformanceTestSuite {
...
+		s.SupportedFeatures = AllFeatures
+	} else if s.SupportedFeatures == nil {
 		s.SupportedFeatures = StandardCoreFeatures
 	} else {
-		for feature, val := range StandardCoreFeatures {
-			if _, ok := s.SupportedFeatures[feature]; !ok {
-				s.SupportedFeatures[feature] = val
+		for feature := range StandardCoreFeatures {
+			if !s.SupportedFeatures.Has(feature) {
+				s.SupportedFeatures.Insert(feature)
```
shaneutt, line 160: nit: now that we're using a set, you could just do an `Insert(feature)` here for every element in the `StandardCoreFeatures`, as duplicates wont matter. Not necessary though.

shaneutt (approved):
Would like @robscott to give the final LGTM since he's been reviewing this too, but does LGTM.

shaneutt (approved):

gyohuangxin:
@robscott Can you review it?

conformance/utils/suite/suite.go
```diff
@@ -222,7 +243,7 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 	// Check that all features exercised by the test have been opted into by
 	// the suite.
 	for _, feature := range test.Features {
-		if supported, ok := suite.SupportedFeatures[feature]; !ok || !supported {
+		if suite.SupportedFeatures.Has(feature) {
```
robscott, line 246: I think this we actually want this?
```suggestion
		if !suite.SupportedFeatures.Has(feature) {
```

robscott (commented):
Thanks @gyohuangxin!

conformance/utils/suite/suite.go
gyohuangxin: @robscott It's my mistake, updated.

shaneutt (approved):
/lgtm

Issue #1619, opened 2022-12-22
Add requirement for implementations to document session persistence behavior 


gcs278:
**What would you like to be added**:

See Goals/Non Goals section below. I originally created this as a Google Doc, so I'm keeping the source format below. It's a lot of information for a simple request, but hopefully it describes the context of our request sufficiently. Please let me know if you disagree with verbiage, something is incorrect, have feedback on our objective, or have a suggestion on the next steps we could take.

**Why this is needed**:
# Overview

At the [November 28th, 2022 SIG Gateway API meeting](https://www.youtube.com/watch?v=x9b31Bu10Bc&t=2664s), we discussed the idea of standardizing and documenting session persistence (also known as session stickiness). Our request, at a minimum, is that we'd like Gateway API implementations to be required to document how they achieve session persistence and what is their default functionality.

It was recommended that we first generate discussion for how we describe session persistence, agree on how it is achieved, and find out which implementers support session persistence. Once information is gathered, we can move forward with a conformance test for session persistence if deemed appropriate. Furthermore, this information can be used for developing a Gateway API spec to configure session persistence in the future.

### Goals

- Define session persistence to establish a common language
- Have the Gateway API specification require implementations to document session persistence behavior

### Non-Goals

- Create proposals for adding session persistence configuration to Gateway API object specification
- Mandate a default or expected session persistence functionality for implementations
- Require implementations to support session persistence or a specific form of it

### User Stories

**Story:** _As a user of some Gateway API implementation who depends on session persistence, I need to know whether an alternative Gateway API implementation supports session persistence, so that I know whether switching from one implementation to the other will break my application._

**Resolution:** Implementations should document how they achieve session persistence.

**Story:** _As the administrator for a load balancer/WAF/proxy in front of a Gateway API implementation, I need to know how a given Gateway API implementation implements session persistence, so that I know whether my LB/WAF/proxy will break it (for example, by filtering cookies or by obscuring source addresses)_

**Resolution:** Implementations should document how they achieve session persistence.

## Defining Session Persistence

Session persistence is the act of using application layer information to direct a client request to the same backend server for the duration of a "session". Session persistence, when properly configured, guarantees a user will be directed to the same server. Persistence is an exception to load balancing: a persistent client request bypasses the proxy's load balancing algorithm, going directly to a backend server it has previously established a session with.

[...]

PR #1643, opened 2023-01-12
GEP-1619 Session Persistence and Session Affinity

gcs278:
**What type of PR is this?**
/kind gep

Add a GEP for session persistence that establishes the foundation for describing it and establishing a common language.

Doesn't fix, but supports https://github.com/kubernetes-sigs/gateway-api/issues/1619

N/A

Markdown Preview: https://github.com/gcs278/gateway-api/blob/gep-1619/site-src/geps/gep-1619.md

### DIFF
--- a/site-src/geps/gep-1619.md
+++ b/site-src/geps/gep-1619.md
@@ -1,0 +1,129 @@
+# GEP-1619: Session Persistence and Session Affinity
+
+* Issue: [#1619](https://github.com/kubernetes-sigs/gateway-api/issues/1619)
+* Status: Provisional
+
+(See definitions in [GEP Status][/contributing/gep#status].)
+
+## TLDR
+
+This GEP proposes a definition of session persistence and session affinity as the foundation of future session persistence related improvements. This GEP also outlines the ways session persistence could be achieved by implementations.
+
+## Goals
+- Define session persistence and session affinity to establish a common language
+- Identify differences in session persistence functionality between implementations
+
+## Non-Goals
+- Mandate a default or expected session persistence functionality for implementations
+
+## Introduction
+
+### Defining Session Persistence
+
+Session persistence is when a client request is directed to the same backend server for the duration of a "session". It is achieved when a client directly provides information, such as a header, that a proxy uses as a reference to direct traffic to a specific server. Persistence is an exception to load balancing: a persistent client request bypasses the proxy's load balancing algorithm, going directly to a backend server it has previously established a session with.
+
+Session persistence enables more efficient application workflows:
+1. Better performance: Maintaining a single session allows a server to cache information about a client locally reducing the need for servers to exchange session data and overall storage needs.
+2. Seamless client experience: Clients can reconnect to the same server without re-authenticating or re-entering their information.
+
+Some of the concerns of session persistence are the duration and expiration of the session, security of the transaction stream, and storage of the context or state. 
+
+Session affinity, not to be confused with session persistence, uses an existing attribute of the request to consistently send to the same backend. Session affinity can be considered a weaker form of session persistence: it is not guaranteed to persist a connection to the same backend server if certain attributes of the request or the backends are changed.
+
+### Security and Privacy Implications
+
+Session persistence can introduce security and privacy vulnerabilities if not properly implemented. These vulnerabilities can include:
+
+1. Session hijacking: Attackers intercepting or predicting a valid session token to gain unauthorized access.
+2. Session fixation: Attackers setting a client's session ID to a known value, which they can then use to hijack the session.
+3. Session replay attacks: Attackers capturing and resending a client's message with a valid session ID.
+4. Data leakage: Attackers can exploit sensitive session information cached on servers if not properly secured.
+5. Denial of service attacks: Attackers can use up server resources by creating and maintaining large numbers of sessions.
+
+To mitigate these security concerns, it is important to implement session persistence using secure practices, such as using strong session ID generation algorithms, implementing session timeouts, encrypting sensitive data, and monitoring server resources for unusual activity.
+
+Session affinity introduces fewer security and privacy vulnerabilities since there are no session tokens to protect or exploit.
+
+### Achieving Session Persistence
+
+Session persistence is achieved using attributes residing in the application layer. The following are mechanisms for achieving session persistence:
+
+**1. Cookie-Based Session Persistence**
+
+The most common mechanism is by using cookies (described by [RFC6265](https://www.rfc-editor.org/rfc/rfc6265)) with the set-cookie HTTP response header. A client will use the provided value in the set-cookie response header in a cookie request header in subsequent requests. Proxies can use this cookie header to maintain a persistent connection to a single backend server on behalf of the client.
+
+**2. Header-Based Session Persistence**
+
+Header-based stateful sessions are achieved by a backend or gateway providing an HTTP response header and the client using the same header in subsequent HTTP requests. Proxies can use this header to maintain a persistent connection to a single backend server on behalf of the client.
+
+**3. URL-Encoded Session Persistence**
+
+Session information can be also encoded into the request URL to establish a persistent session. The server rewrites the client's URL to encode the new session information automatically. The server then decodes the session information from the URL to identify the session.
+
+**Session Initiation**
+
+For both header-based and cookie-based sessions, either the gateway or the backend server can initiate establishing the session via the appropriate header or set-cookie attributes. The following rules apply based on who initiates the session:
+
+- If the backend initiates the session, the gateway should allow this and not force persistence connections, unless specifically configured to. The gateway may decode and alter the cookie established by the backend to achieve session persistence.
+- If the gateway initiates the session, the backend will be presented with session attributes regardless if it enabled them. 
+
+### Achieving Session Affinity
+
+While session persistence uses attributes in the application layer, session affinity often uses, but is not limited to, attributes below the application layer.
+
+Session affinity can be achieved by deterministic load balancing algorithms or a proxy feature that tracks IP-to-backend associations such as [HAProxy's stick tables](https://www.haproxy.com/blog/introduction-to-haproxy-stick-tables/) or [Cilium's session affinity](https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#id2). 
+
+### Implementations
+To help understand the scope of session persistence and session affinity, we've created a table describing how each implementation achieves it. Input from implementations is appreciated to complete this table.
+
+| **Proxy Type** 	| **Implementations Based on Proxy** 	| **Session Persistence** 	| **Session Affinity** 	| **Notes** 	|
+|---	|---	|---	|---	|---	|
+| Envoy 	| Acnodal EPIC, Contour, Emissary-Ingress (Ambassador API Gateway), Envoy Gateway, Gloo Edge 2.0, HashiCorp Consul, Istio, Kuma 	| [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto), [Header-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) 	| Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash) 	| Sessions are associated with a service; therefore, sessions can't cross service boundaries due to current envoy implementation. 	|
+| Nginx 	| Apache APISIX, NGINX Kubernetes Gateway 	| [Cookie-Based, Header-Based, or other request attributes](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#enabling-session-persistence) (Nginx Plus Only) 	| [Consistent Hashing](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash) via IP or other connection attributes (Nginx Open Source) 	| N/A 	|
+| eBPF 	| Cilium 	| No 	| Source IP Address 	| Cilium does not have a true session persistence feature, but does implement [session affinity by default](https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#id2) using source IP address. 	|
+| Google CloudRun 	| Google Kubernetes Engine 	| [Cookie-Based](https://cloud.google.com/run/docs/configuring/session-affinity) 	| No 	| Google Cloud Run uses the terminology "session affinity" for what we define as session persistence. 	|
+| Pipy 	| Flomesh Service Mesh 	| ? 	| ? 	| ? 	|
+| HAProxy 	| HAProxy Ingress 	| [Cookie-Based](https://docs.haproxy.org/2.6/configuration.html#4-option%20forwardfor) 	| [Stick Tables](https://cbonte.github.io/haproxy-dconv/2.4/configuration.html#4-stick-table) using connection attributes 	| HAProxy implements cookie-based stateful sessions via the [forwardfor option](https://docs.haproxy.org/2.6/configuration.html#4-option%20forwardfor). 	|
+| Kong 	| Kong 	| Cookie-Based 	| ? 	| Kong has the [session plugin](https://docs.konghq.com/hub/kong-inc/session/) which manages persistent sessions with cookie-based stateful sessions. 	|
+| Traefik Proxy 	| Traefik 	| [Cookie-Based](https://doc.traefik.io/traefik/routing/services/#sticky-sessions) 	| No 	| [Docs](https://traefik.io/glossary/what-are-sticky-sessions/) mention consistent hashing, but Traefik [only offers round robin](https://doc.traefik.io/traefik/routing/services/#load-balancing) (not consistent). 	|
+| Java Servlet 	| N/A 	| [Cookie-Based](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getSession--), [URL Encoded](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#isRequestedSessionIdFromURL--) 	| N/A 	| Java Servlets don't natively support proxy functions. 	|
+| Apache httpd 	| N/A 	| [Cookie-Based](https://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html), [URL Encoded](https://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html) 	| N/A 	|  N/A	|
+
+### Sessions in Java
+
+Java application servers such as Tomcat and Jetty, were the first to standardize the API around persistent sessions. These Java applications introduced the “jsessionid” cookie as well as more advanced features such as session migration, replication, and on demand session activation. It’s important for Gateway API to examine persistent session use cases and history from Java APIs to ensure the API is designed appropriately.
+
+### Open Questions
+
+TBD
+
+## TODO
+The following are items that we intend to resolve before we consider this GEP implementable:
+
+- We need to identify the needs and use cases from end users for specifying session persistence.
+- We need to identify and document requirements regarding session draining and migration. How do implementations drain established sessions during backend upgrades without disruption?
+- We need to document sessions with Java in greater detail. Java standardized the API and behavior of session persistence long ago and would be worth examining.
+- We need to add a small section on compliance regarding the browser and client relationship.
+- We need to design and document an API for session persistence and/or session affinity in the Gateway API spec.
+
+## API
+
+TBD
+
+## Alternatives
+
+### Alternate Naming
+
+This GEP describes session persistence and session affinity as the idea of strong and weak connection persistence respectively. Other technologies use different names or define persistence and affinity differently:
+
+- Envoy defines [stateful sessions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto) as what we've defined as session persistence
+- Google Cloud defines [session affinity](https://cloud.google.com/run/docs/configuring/session-affinity) as what we've defined as session persistence
+- Nginx defines [session persistence](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#enabling-session-persistence) as what we've defined as both session persistence and affinity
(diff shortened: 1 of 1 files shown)


site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Defining Session Persistence
+
+Session persistence is the act of using application layer information to direct a client request to the same backend server for the duration of a "session". Session persistence, when properly configured, guarantees a user will be directed to the same server. Persistence is an exception to load balancing: a persistent client request bypasses the proxy's load balancing algorithm, going directly to a backend server it has previously established a session with.
+
+Some of the concerns of session persistence are the duration and expiration of the session, security of the transaction stream, and storage of the context or state. Application developers use session persistence to improve the performance and reduce overall storage needs by aligning each transaction and its cached data with the same server.
+
+Session persistence is not to be confused with session affinity, which uses information below the application layer to maintain a client request to a single server. Session affinity can be considered a weaker form of session persistence: it is not guaranteed to persist a connection to the same backend server.
```
howardjohn, line 33: This seems to be saying the difference between Affinity and Persistence is L7 vs L4. Its not clear to me that is an important distinction

In my view, the primary difference between session types is whether the client _directly_ specifies the backend they want OR if some attribute of the request is just used to consistently send to a backend.

* Type 1 example: A cookie with `Backend=1.2.3.4`
* Type 2 example: Send to backend based on `backend[hash(SomeHeader) % BackendCount]`

In Envoy terms, 1 is "stick sessions" and 2 is "consistent hash".

To me, that distinction seems more relevant than L7 vs L4. I would agree that "affinity" does tend to be L4 based on source IP, as you noted in your examples.

gcs278: Good point, I agree. I reworded Persistence and Affinity verbiage with focus on what information/attributes are being used to achieve the connection consistency, let me know what you think.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+| Apache APISIX | Nginx | ? | ? | ? |
+| Cilium | eBPF | No | Source IP Address | Cilium does not have a true session persistence feature, but does implement [session affinity by default](https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#id2) using source IP address. |
+| Contour | Envoy | Cookie-Based | Consistent Hashing via Maglev and Ring Hash | Envoy proxy implements both [cookie-based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto) and [header-based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) stateful sessions. Envoy offers [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash) load balancing algorithms that provide consistent hashing. |
```
sunjayBhatia, line 61: Contour also supports using other request attributes other than cookies to route requests to a consistent backend: https://projectcontour.io/docs/v1.23.2/config/request-routing/#load-balancing-strategy (the `RequestHash` load balancing strategy)

gcs278: Thanks for the info - I added a note about this in the Contour row.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+## TLDR
+
+This GEP proposes a definition of session persistence and session affinity as the foundation of future session persistence related improvements. This GEP also outlines the ways session persistence could be achieved by implementations.
+
+## Goals
+- Define session persistence and session affinity to establish a common language
+- Identify differences in session persistence functionality between implementations
+
+## Non-Goals
+- Create proposals for adding session persistence configuration to Gateway API object specification
```
robscott, line 17: Although this may be a non-goal for this particular "provisional" stage of the GEP, I think it would need to be a goal of the GEP as a whole. Or maybe I'm misunderstanding the purpose of writing this up.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+- Mandate a default or expected session persistence functionality for implementations
+- Require implementations to support session persistence or a specific form of it
```
robscott, line 19: If there are values that are broadly implementable, we _may_ want to consider this "core" conformance which would conflict with this non-goal. (I don't actually know that this is broadly implementable, but it may be safer to leave this non-goal out).

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Introduction
+
+At the November 28th, 2022 SIG Gateway API meeting, we discussed the idea of standardizing and documenting session persistence (also known as session stickiness). It was recommended that we first generate discussion for how we describe session persistence, agree on how it is achieved, and find out which implementers support session persistence. 
+
+Conformance test for session persistence will be considered if deemed appropriate. Furthermore, this information can be used for developing a Gateway API spec to configure session persistence in the future.
```
robscott, line 25: Instead of providing a history, I think I'd just explain what session affinity means and why it should be configurable within Gateway API (assuming that's the goal). Edit: Looks like you've already done that well below, so maybe just leave this part out.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+**Header-Based Session Persistence**
+
+Header-based stateful sessions are achieved by a backend providing an arbitrarily-named HTTP response header and the client using the same arbitrarily-named header in subsequent HTTP requests. Proxies can use this arbitrarily-named header to maintain a persistent connection to a single backend server on behalf of the client.
+
+**Session Affinity**
+
+Session affinity can be achieved by deterministic load balancing algorithms or a proxy feature that tracks IP-to-backend associations such as [HAProxy's stick tables](https://www.haproxy.com/blog/introduction-to-haproxy-stick-tables/) or [Cilium's session affinity](https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#id2). 
+
+### Implementations
+To help understand the scope of session persistence and session affinity, we've created a table describing how each implementation achieves it. Input from implementations is appreciated to complete this table.
```
robscott, line 54: It may help to clarify that this table intends to provide the kinds of session persistence and affinity that are supported by each implementation. Just for the sake of maintaining this going forward, it may be easiest to just focus on underlying implementations such as Envoy, NGINX, HAProxy, etc. Even if an implementation based on one of those does not support a feature today, that may just be an omission of their APIs and not a limitation of their underlying dataplane.

robscott (commented):
Thanks @gcs278!

site-src/geps/gep-1619.md
shaneutt: Agreed, it seems that this should be an _eventual_ goal of this KEP, however it's reasonable for the first iteration to just be the "what and the why" without the how and still merge.

shaneutt (approved):
Overall I really like the approach of talking about the "what" and "why" first, rather than digging right into details.

I'm approving this GEP based on the fact that it is provisional, and I see no reason to block merging it in this provisional state as I expect we can iteratively update it afterwards. That said, I do have a few comments that I would like to see resolved before we merge please see what you think of them.

site-src/geps/gep-1619.md
shaneutt:
```suggestion
```

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+Conformance test for session persistence will be considered if deemed appropriate. Furthermore, this information can be used for developing a Gateway API spec to configure session persistence in the future.
+
+### Defining Session Persistence
+
+Session persistence is when a client request is directed to the same backend server for the duration of a "session". It is achieved when a client directly provides information, such as a header, that a proxy uses as a reference to direct traffic to a specific server. Persistence is an exception to load balancing: a persistent client request bypasses the proxy's load balancing algorithm, going directly to a backend server it has previously established a session with.
+
+Some of the concerns of session persistence are the duration and expiration of the session, security of the transaction stream, and storage of the context or state. Application developers use session persistence to improve the performance and reduce overall storage needs by aligning each transaction and its cached data with the same server.
```
costinm, line 31: Privacy too. The information used for establishing the session on the internet ( the main use of the gateway ), in particular in browser context is very sensitive, used for tracking and may include user-specific info.

costinm: I would also expand on the 'security' - maybe in separate section. Many systems use session as part of authentication - holding user info in a 'server side session' and using the session cookie as a bearer token ( Java but not only)

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+Session affinity, not to be confused with session persistence, uses an existing attribute of the request to consistently send to the same backend. Session affinity can be considered a weaker form of session persistence: it is not guaranteed to persist a connection to the same backend server if certain attributes of the request or the backends are changed.
```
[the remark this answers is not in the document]
costinm: I agree, affinity is quite different - far fewer security/privacy and operational implications ( draining / lame duck / etc).

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+Here are a some mechanisms to achieving both:
+
+**Cookie-Based Session Persistence**
+
+The most common mechanism is by using cookies (described by [RFC6265](https://www.rfc-editor.org/rfc/rfc6265)) with the set-cookie HTTP response header. A client will use the provided value in the set-cookie response header in a cookie request header in subsequent requests. Proxies can use this cookie header to maintain a persistent connection to a single backend server on behalf of the client.
+
+**Header-Based Session Persistence**
+
+Header-based stateful sessions are achieved by a backend providing an arbitrarily-named HTTP response header and the client using the same arbitrarily-named header in subsequent HTTP requests. Proxies can use this arbitrarily-named header to maintain a persistent connection to a single backend server on behalf of the client.
```
costinm, line 47: I'm not sure 'arbitrarily named' is right. We may support custom names, but I think Java made the right choice to define a 'default' name ( while allowing implementations to customize using their own settings). 

I would also use plural - for example in envoy-based persistence it is critical to use a second header to guarantee the same cluster ( aka Service/subset ) is used. 

I'm also not sure outside of envoy how many Internet proxies use header based persistence - would be nice having a survey. 

Using request parameter ( jsessionid=xxx in java ) is common, using the auth info is even more common. Until envoy - I haven't seen arbitrary header-based persistence.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Achieving Session Persistence and Session Affinity
+
+Session persistence is achieved with the application layer, while session affinity often uses, but is not limited to, attributes below the application layer.
+
+Here are a some mechanisms to achieving both:
+
+**Cookie-Based Session Persistence**
+
+The most common mechanism is by using cookies (described by [RFC6265](https://www.rfc-editor.org/rfc/rfc6265)) with the set-cookie HTTP response header. A client will use the provided value in the set-cookie response header in a cookie request header in subsequent requests. Proxies can use this cookie header to maintain a persistent connection to a single backend server on behalf of the client.
```
costinm, line 43: In the next section you have "backend providing an arbitrary name". This section doesn't specify who provides the cookie. 

I think in both cases we need to make it a bit more clear if we want to be compatible with broadly used patterns.

1. A backend may initiate the session as needed ( Java getHttpSession(), etc) - the Gateway API should allow this and not force but allow a persistent session. In this case - the contract needs to be very clear and we should mention that the gateway implementation may alter the cookie ( to add backend info ) or have ability to decode part of the cookie (servlets and many others use a suffix to identify the node).
2. The gateway implementation itself may generate the header/cookie - in this case backend no longer has a choice to use or not sessions. 

In Java spec, a lot of discussion was around on the 'shopping cart' use case - a user may browse the store searching ( using proper load balancing ), but when he adds to cart the persistent session is created. The other important use case was auth/SSO - a page may be accessed as unauthenticated, but once user logs in usually
a cookie is used and in most implementations it creates a session. 

Having persistent sessions tied to the user is still broadly used.

site-src/geps/gep-1619.md
costinm: +Java Servlets please. Persistent sessions since ~2000 encoded in the spec, multiple implementations and support in major proxies.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+| HashiCorp Consul | Envoy | [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto), [Header-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) | Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash)| N/A |
+| Istio | Envoy | [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto), [Header-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) | Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash)| N/A |
```
costinm, line 69: https://github.com/istio/istio/blob/master/pilot/pkg/features/pilot.go#L123

Current implementation is based on a Service label - since sessions are associated with a Service ( can't have sessions cross service due to current envoy impl ).

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+| Flomesh Service Mesh | Pipy | ? | ? | ? |
+| Gloo Edge 2.0 | Envoy | [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto), [Header-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) | Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash)| N/A |
+| Google Kubernetes Engine | ? | ? | ? | ? |
```
costinm, line 66: Google CloudRun - cookie based, set on service. https://cloud.google.com/run/docs/configuring/session-affinity

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+| Kong | Kong | Cookie-Based | ? | Kong has the [session plugin](https://docs.konghq.com/hub/kong-inc/session/) which manages persistent sessions with cookie-based stateful sessions. |
+| Kuma | Envoy | [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto), [Header-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/header/v3/header.proto) | Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash)| N/A |
+| LiteSpeed Ingress Controller | ? | ? | ? | ? |
+| NGINX Kubernetes Gateway | Nginx | ? | ? | ? |
```
pleshakov, line 73:
```suggestion
| NGINX Kubernetes Gateway | Nginx | [Cookie-Based, Header-based  or other request attributes](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#enabling-session-persistence) (NGINX Plus only) |  [Hashing](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash) (incl. consistent hashing) based on client IP address and any other connection/request attributes. | N/A |
````

File: site-src/geps/gep-1619.md [hunk 14]
```diff
@@ -0,0 +1,89 @@
...
+| Traefik |  Traefik Proxy | [Cookie-Based](https://doc.traefik.io/traefik/routing/services/#sticky-sessions) | No | [Docs](https://traefik.io/glossary/what-are-sticky-sessions/) mention consistent hashing, but Traefik [only offers round robin](https://doc.traefik.io/traefik/routing/services/#load-balancing) (not consistent). |
+
+### Open Questions
+- What are the needs and use cases from end users for specifying session persistence?
```
pleshakov, line 77: Do we need to consider how to drain established sessions?  For example, an application developer wants to upgrade backends without disrupting any established sessions but rather wait for the sessions to expire.


example of this use case -  https://github.com/jcmoraisjr/haproxy-ingress/issues/137

costinm: Yes, Istio implementation also includes draining ( using k8s new status ) - for the rollout of a new version it is better to gradually set traffic to 0% on old version and let the connection count drop over how long you need. 

However for autoscale down - drain support and a lot of tricks are needed ( stop hook, etc). Would be great to have a common doc, I assume all vendors will have the same problems - but may be a separate discussion.

gcs278:
Thanks for all of the feedback. I have some high priority items I'm trying to sort through at the moment, but I am planning to get back here in the next couple weeks and respond to the comments.

gcs278:
Almost done with addressing comments. Had some urgent items come up this week. Will have a new rev posted early next week.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+| LiteSpeed Ingress Controller | ? | ? | ? | ? |
+| NGINX Kubernetes Gateway | Nginx | ? | ? | ? |
+| Traefik |  Traefik Proxy | Cookie-Based | No | Traefik implements [cookied-based](https://doc.traefik.io/traefik/routing/services/#sticky-sessions) stateful sessions. [Docs](https://traefik.io/glossary/what-are-sticky-sessions/) mention consistent hashing, but Traefik [only offers round robin](https://doc.traefik.io/traefik/routing/services/#load-balancing) (not consistent). |
+
+### Open Questions
+- What are the needs and use cases from end users for specifying session persistence?
```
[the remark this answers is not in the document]
gcs278: Did you mean a new section dedicated to talking about Java? Or table entries? I started a section, but will probably need your advice on expanding on the content.

Thanks for the doc, it's very thorough. I'll leverage that.

For privacy, I think I addressed that in the new `Security and Privacy Implications` section. For compliance, could you expand? What compliance requirements are we referring to? Are you saying we should create requirements in this GEP?

gcs278 (commented):
Thanks for the detailed reviews. Sorry for the delay.

gcs278:
@shaneutt Thanks - marked back as ready for review, posted responses, and updates to the GEP.

site-src/geps/gep-1619.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Introduction
+
+At the November 28th, 2022 SIG Gateway API meeting, we discussed the idea of standardizing and documenting session persistence (also known as session stickiness). It was recommended that we first generate discussion for how we describe session persistence, agree on how it is achieved, and find out which implementers support session persistence. 
+
+Conformance test for session persistence will be considered if deemed appropriate. Furthermore, this information can be used for developing a Gateway API spec to configure session persistence in the future.
+
+### Defining Session Persistence
+
+Session persistence is when a client request is directed to the same backend server for the duration of a "session". It is achieved when a client directly provides information, such as a header, that a proxy uses as a reference to direct traffic to a specific server. Persistence is an exception to load balancing: a persistent client request bypasses the proxy's load balancing algorithm, going directly to a backend server it has previously established a session with.
```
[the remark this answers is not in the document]
shaneutt: This is sufficient, we can always improve in future PRs too if readers are asking for more clarifications :+1:

site-src/geps/gep-1619.md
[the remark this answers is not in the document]
shaneutt: This one is kinda up to you, if to get a whole working thing you need to cover both topics, then let's do it. If not, maybe another GEP and we just make reference here? I don't consider this a blocker, so I'll consider this resolved for the purposes of this iteration, we can continue to iterate on this PR with the current lagnuage I think.

site-src/geps/gep-1619.md
shaneutt: In [GEP 1709](https://gateway-api.sigs.k8s.io/geps/gep-1709/) we added an explicit [TODO Section](https://gateway-api.sigs.k8s.io/geps/gep-1709/#todo) as a list of things to follow up on in later PRs, and also as a list of things we felt would need to block moving the GEP to `implementable`. I would recommend using that same approach here to reduce how much more we need to cover in this single PR.

shaneutt (approved):
I think this is a great starting point, albeit we are already starting to get quite big for a first `provisional` GEP. So my suggestion is to wrap up any unresolved comments or thoughts into a TODO list (similar to https://gateway-api.sigs.k8s.io/geps/gep-1709/#todo) to mark them as things that can be picked up for follow-up PRs and that we want to resolve before `implementable`. Once that's done I think we're in a fine place to merge this as `Provisional` and start thinking about the next iterative steps.

sunjayBhatia (approved):
lgtm given the stated goals

site-src/geps/gep-1619.md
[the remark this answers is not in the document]
gcs278: Sounds good. I'm going to say lets keep it as is, and I think we can always diverge into another GEP in the future if it becomes too disconnected.

site-src/geps/gep-1619.md
gcs278: Adding to a TODO section.

gcs278:
I've created a `TODO` section and added TODOs from the comments from the review, mostly from @costinm's comments. Let me know if this looks good.

shaneutt (approved):
With the TODO list in place for things we need to resolve before `implementable`, the time this has been up for review, the fact that it's provisional, and that nobody has any blocking comments it seems like now is a good time to get this first pass landed and we can build off from here.

Thanks @gcs278!

/lgtm

PR #1644, opened 2023-01-12
update regex to comply with RFC-3986

jackstine:
# Description
Adding `[%][0-9a-fA-F]{2}` as an option to find Hex values. Also adding `@` as a value to find as well.
The subset of single characters can be selected in addition to the set of all Hex values, hence the `(?:<pattern1>|<pattern2>)`

**What type of PR is this?**
/kind bug

Complies with RFC-3986 "p-char" characters.

Updates this PR [#1599](https://github.com/kubernetes-sigs/gateway-api/pull/1599)

It might, but not really.
```release-note
update route validation to comply with RFC-3986 "p-char" characters.
```

### DIFF
--- a/apis/v1beta1/validation/httproute.go
+++ b/apis/v1beta1/validation/httproute.go
@@ -39,7 +39,7 @@ var (
 	invalidPathSuffixes  = []string{"/..", "/."}
 
 	// All valid path characters per RFC-3986
-	validPathCharacters = "^[A-Za-z0-9\\/\\-._~%!$&'()*+,;=:]+$"
+	validPathCharacters = "^(?:[A-Za-z0-9\\/\\-._~!$&'()*+,;=:@]|[%][0-9a-fA-F]{2})+$"
 )
 
 // ValidateHTTPRoute validates HTTPRoute according to the Gateway API specification.


jackstine:
Adding @robscott here because he initially worked on the PR https://github.com/kubernetes-sigs/gateway-api/pull/1599

robscott:
Thanks for catching this @jackstine! 

Just want to make sure I'm understanding both changes correctly:

1. Per [RFC-3986](https://www.rfc-editor.org/rfc/rfc3986). `pchar  = unreserved / pct-encoded / sub-delims / ":" / "@"`. My initial regex missed both `:` and `@`, this fixes that. 
2. Per the same RFC, `%` is only valid as part of `pct-encoded = "%" HEXDIG HEXDIG`, this enforces that restriction.

Does that line up with what you're thinking? Did I miss anything?

/ok-to-test

jackstine:
Hey @robscott that is exactly right.   Also `:` was already included.

robscott:
Makes sense, thanks @jackstine! Will defer to someone else to LGTM.

/approve

dprotaso:
Any chance we could add a unit test for the regexp change?

robscott:
@dprotaso that seems like a good idea. So far the only way we've tested CRD validation is by adding valid examples to `/examples` and invalid examples to `/hack/invalid-examples`. We then rely on `hack/verify-examples-kind.sh` to check that the valid examples pass validation and the invalid ones don't. Probably worth filing an issue to track this.

PR #1645, opened 2023-01-13
fix: reuse gc name in GatewayClassObservedGenerationBump conformance

Xunzhuo:
Signed-off-by: bitliu <bitliu@tencent.com>

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/conformance/tests/gatewayclass-observed-generation-bump.yaml
+++ b/conformance/tests/gatewayclass-observed-generation-bump.yaml
@@ -1,7 +1,7 @@
 apiVersion: gateway.networking.k8s.io/v1beta1
 kind: GatewayClass
 metadata:
-  name: gatewayclass-observed-generation-bump
+  name: "{GATEWAY_CLASS_NAME}"
 spec:
   controllerName: "{GATEWAY_CONTROLLER_NAME}"
   description: "old"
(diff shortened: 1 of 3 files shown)


conformance/tests/gatewayclass-observed-generation-bump.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: gateway.networking.k8s.io/v1beta1
 kind: GatewayClass
 metadata:
-  name: gatewayclass-observed-generation-bump
+  name: "{GATEWAY_CLASS_NAME}"
```
robscott, line 4: There's some previous context here: https://github.com/kubernetes-sigs/gateway-api/pull/1586#discussion_r1051027236. There was also more discussion in the community meeting around that time.

I had 2 main concerns:

1. I didn't want to allow an existing resource to be mutated by conformance tests
2. I also wasn't convinced that every implementation would be able to support changes to arbitrary GatewayClasses

Making this change would go against 1 thought it would help with 2.

Eventually we agreed that these could be added if they were guarded by a feature gate and not part of the default test suite. @dprotaso can likely add some more context here as well.

robscott (commented):
Thanks for the work on this @Xunzhuo, added a comment to provide some more context. If I'm understanding correctly, it sounds like a single Envoy Gateway controller can't support multiple GatewayClasses?

/cc @dprotaso 
/hold for consensus

conformance/tests/gatewayclass-observed-generation-bump.yaml
Xunzhuo: Thanks @robscott! I think EG should support the multiple-GC model, and it makes sence : )

dprotaso: Yeah this test purposely mutates a second gateway class to test the observed generation is bumped in the status conditions.

I believe changing the name to the default name `GATEWAY_CLASS_NAME` will (may?) break other tests since it's clearing params etc.

youngnick:
I've opened #1650 to cover updating the implementer's guide with some more info to clarify some of these issues.

arkodg:
this shouldn't be needed if https://github.com/kubernetes-sigs/gateway-api/pull/1655 is accepted

robscott:
Closing this out now that #1655 has merged, feel free to reopen if I missed anything.

/close

PR #1652, opened 2023-01-16
Fix GEP template - status link was wrong

dprotaso:
Status definitions were pointing to the wrong file


shaneutt (approved):
I followed the link in the preview, and this seems to be correct.

/lgtm

Thank you

PR #1655, opened 2023-01-18
Loosen `Accepted` check in GWC Obs Generation test

arkodg:
* enhance `GWCMustBeAccepted` to pass a status value along, an empty string indicates, any status value is matched.

* Loosen the check within `GatewayClassObservedGenerationBump` to wait for the `Accepted` condition to have any status value

Relates to https://kubernetes.slack.com/archives/CR0H13KGA/p1673836949189379?thread_ts=1673599220.709819&cid=CR0H13KGA

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -65,7 +65,7 @@ func NewGatewayRef(nn types.NamespacedName, listenerNames ...string) GatewayRef 
 
 // GWCMustBeAcceptedConditionTrue waits until the specified GatewayClass has an Accepted condition set with a status value equal to True.
 func GWCMustHaveAcceptedConditionTrue(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName string) string {
-	return gwcMustBeAccepted(t, c, timeoutConfig, gwcName, metav1.ConditionTrue)
+	return gwcMustBeAccepted(t, c, timeoutConfig, gwcName, string(metav1.ConditionTrue))
 }
 
 // GWCMustBeAcceptedConditionAny waits until the specified GatewayClass has an Accepted condition set with a status set to any value.
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -63,11 +63,21 @@ func NewGatewayRef(nn types.NamespacedName, listenerNames ...string) GatewayRef 
 	}
 }
 
-// GWCMustBeAccepted waits until the specified GatewayClass has an Accepted
-// condition set. Passing an empty status string means that any Status will do.
-// It also returns the ControllerName for the GatewayClass.
+// GWCMustBeAcceptedConditionTrue waits until the specified GatewayClass has an Accepted condition set with a status value equal to True.
+func GWCMustHaveAcceptedConditionTrue(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName string) string {
+	return gwcMustBeAccepted(t, c, timeoutConfig, gwcName, metav1.ConditionTrue)
+}
+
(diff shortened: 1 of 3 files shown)


conformance/utils/kubernetes/helpers.go
```diff
@@ -64,10 +64,10 @@ func NewGatewayRef(nn types.NamespacedName, listenerNames ...string) GatewayRef
 }
 
 // GWCMustBeAccepted waits until the specified GatewayClass has an Accepted
-// condition set to true. It also returns the ControllerName for the
-// GatewayClass. This will cause the test to halt if the specified timeout is
-// exceeded.
-func GWCMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName string) string {
+// condition set. Passing an empty status string means that any Status will do.
```
shaneutt, line 67: I think this language may be a little bit more clear:

```suggestion
// condition set. Passing an empty status string means that any value will be accepted.
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -64,10 +64,10 @@ func NewGatewayRef(nn types.NamespacedName, listenerNames ...string) GatewayRef
...
+// It also returns the ControllerName for the GatewayClass.
+// This will cause the test to halt if the specified timeout is exceeded.
+func GWCMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName, expectedStatus string) string {
```
shaneutt, line 70: With this taking a status value now, I'm not sure `GWCMustBeAccepted` feels like the right name. I'm slightly in favor of `GWCMustHaveAcceptedCondition` as I feel like "being accepted" seems to indicate the `True` state, whereas now that's optional.

Alternatively: I'm not really a fan of functions where arguments can be "optional" by providing their default value. I'm not going to push to hard for it, but tell me what you think of:

```go
func GWCMustHaveAcceptedConditionTrue(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName) string
func GWCMustHaveAcceptedConditionFalse(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName) string
func GWCMustHaveAcceptedConditionAny(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName) string
```

conformance/utils/suite/suite.go
```diff
@@ -161,7 +161,7 @@ func New(s Options) *ConformanceTestSuite {
 // in the cluster. It also ensures that all relevant resources are ready.
 func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 	t.Logf("Test Setup: Ensuring GatewayClass has been accepted")
-	suite.ControllerName = kubernetes.GWCMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.GatewayClassName)
+	suite.ControllerName = kubernetes.GWCMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.GatewayClassName, "True")
```
shaneutt, line 164: I know this was like this before, but it seems for completeness we should use the provided constant while we're here:

```suggestion
	suite.ControllerName = kubernetes.GWCMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.GatewayClassName, metav1.ConditionTrue)
```

conformance/utils/kubernetes/helpers.go
arkodg: thanks for the review ! going with the latter suggestion

shaneutt (approved):
I like the new functional approach, thank you. This LGTM but I'll let one other put the final label on it. :+1:

PR #1657, opened 2023-01-19
Fix invalid example for http redirect

nak3:
The example which redirects HTTP to HTTPS uses scheme `HTTPS` but emits the port. However, docs[1] mentions as "When empty, port (if specified) of the request is used.", so the access always is forwared to the HTTPS but `80` port. 

* The result of the test with Istio 1.17-alpha.ebe08a1368b65d93cf51c24675c29409567ca77c. Please find `location: https://redirect.example.com:80/get`.
```
< HTTP/1.1 301 Moved Permanently
< location: https://redirect.example.com:80/get
< date: Thu, 19 Jan 2023 10:12:52 GMT
< server: istio-envoy
< content-length: 0
```

This patch fixes it.

[1] https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRequestRedirectFilter
> Port is the port to be used in the value of the Location header in the response. When empty, port (if specified) of the request is used.

/kind bug
/kind documentation

### DIFF
--- a/examples/experimental/http-redirect-rewrite/httproute-redirect-https.yaml
+++ b/examples/experimental/http-redirect-rewrite/httproute-redirect-https.yaml
@@ -11,6 +11,7 @@ spec:
     - filters:
       - type: RequestRedirect
         requestRedirect:
+          port: 443
           scheme: https
           statusCode: 301
       backendRefs:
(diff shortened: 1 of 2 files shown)


nak3:
Ah, I overlooked the doc "port __(if specified)__ of the request" . So, if there are no `:80` port on the request, it works and example assumed that `:80` is not specified...
I close this.

PR #1659, opened 2023-01-20
test: path characters regexp validation

### LINKED ISSUE #1658: Test Invalid/Valid Path Characters
@dprotaso that seems like a good idea. So far the only way we've tested CRD validation is by adding valid examples to `/examples` and invalid examples to `/hack/invalid-examples`. We then rely on `hack/verify-examples-kind.sh` to check that the valid examples pass validation and the invalid ones don't. Probably worth filing an issue to track this.

_Originally posted by @robscott in https://github.com/kubernetes-sigs/gateway-api/issues/1644#issuecomment-1386468540_

em-r:
**What type of PR is this?**
/kind bug

adds a unit test for the regular expression used to validate route paths. 

Fixes #1658

```release-note
NONE
```


gyohuangxin:
Thanks for your first PR, good job!
/ok-to-test

robscott:
Hey @em-r, thanks for the PR! This is actually different than what I was hoping for though. This tests webhook validation, but we actually want to test the CRD validation itself. From the issue:

>So far the only way we've tested CRD validation is by adding valid examples to /examples and invalid examples to /hack/invalid-examples. We then rely on hack/verify-examples-kind.sh to check that the valid examples pass validation and the invalid ones don't.

That means instead of writing Go code, we just need some valid examples in `/examples` and invalid examples in `/hack/invalid-examples`. For the valid example, I'd just define a single HTTPRoute that included a variety of path matches that were all valid, and then for invalid examples, you'd probably want several different examples of invalid routes, all with slightly different reasons for being invalid path matches. Does that make sense?

em-r:
Hi @robscott, thank you for the feedback. Yes, that makes sense, I'll update the PR accordingly and let you know, thank you!

em-r:
Hi @robscott, I have added examples for valid and invalid paths as suggested.
Let me know if any change is needed.
Thank you!

robscott:
Thanks @em-r! Sorry for the delay on this one!

/lgtm
/approve

PR #1668, opened 2023-01-23
conformance: HTTPRoute resolvedRefs condition always checked

### LINKED ISSUE #1314: conformance: always check `resolvedRefs` condition in `HTTPRoute` conformance tests

In the `HTTPRoute` conformance tests, the `RouteConditionResolvedRefs` is checked only in those tests expecting it to have status`false` (e.g., `HTTPRouteInvalidNonExistentBackendRef`, `HTTPRouteInvalidBackendRefUnknownKind`, ...), but it is never checked in those tests who expect the routing to be successful. This leads to never checking the `ResolvedRefs` reason with `ConditionStatus == true`.

**Why this is needed**:

This is needed to ensure that the `RouteConditionResolvedRefs` is always set, no matter if the `ConditionStatus` is `true`, or `false`.

mlavacca:
**What type of PR is this?**

/kind feature

In the conformance tests, the ResolvedRefs condition existence is always checked, even when true.

Fixes #1314 

```release-note
The conformance tests always check that the HTTPRoute ResolvedRefs condition is enforced, even when the status is true.
```

### DIFF
--- a/conformance/tests/httproute-cross-namespace.go
+++ b/conformance/tests/httproute-cross-namespace.go
@@ -38,6 +38,7 @@ var HTTPRouteCrossNamespace = suite.ConformanceTest{
 		routeNN := types.NamespacedName{Name: "cross-namespace", Namespace: "gateway-conformance-web-backend"}
 		gwNN := types.NamespacedName{Name: "backend-namespaces", Namespace: "gateway-conformance-infra"}
 		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
 
 		t.Run("Simple HTTP request should reach web-backend", func(t *testing.T) {
 			http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, http.ExpectedResponse{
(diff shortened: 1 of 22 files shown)


conformance/tests/tlsroute-simple-same-namespace.go
```diff
@@ -53,6 +54,11 @@ var TLSRouteSimpleSameNamespace = suite.ConformanceTest{
 			t.Fatalf("unexpected error in test configuration, found %d hostnames", len(hostnames))
 		}
 		serverStr := string(hostnames[0])
+		kubernetes.HTTPRouteMustHaveCondition(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN, metav1.Condition{
+			Type:   string(v1beta1.RouteConditionResolvedRefs),
+			Status: metav1.ConditionTrue,
+			Reason: string(v1beta1.RouteReasonResolvedRefs),
+		})
```
shaneutt, line 61: With this PR this check suddenly becomes very common, and is now repeated in several places with the same `Type`, `Status`, and `Reason`. Given that, have you considered making a wrapper to capture the specific state we're checking for? e.g. `HTTPRouteMustHaveResolvedRefsConditionsTrue()` or something like that? :thinking:

mlavacca: Yep, it makes total sense. I'll introduce an helper for that 👍

mlavacca:
Converting it to draft as it needs some additional work and likely I won't be back to it until next week.

shaneutt:
How's it going @mlavacca? Need any help to move this one forward?

mlavacca:
> How's it going @mlavacca? Need any help to move this one forward?

I'm sorry for the delay. I didn't manage to work on this one lately. I'll do my best to complete it soon

shaneutt:
No worries at all, just let us know if there's anything we can do in support of you :vulcan_salute:

mlavacca:
@shaneutt it's ready to be reviewed now. I also addressed your comment from the early review phase

shaneutt (approved):
Thanks for the refactor, looks clean. Looks good to me, but will wait to have at least one other reviewer give it a double check. :+1:

robscott (commented):
Thanks @mlavacca!

/lgtm

Issue #1670, opened 2023-01-23
Cross-namespace `parentRef` reason for routes accepted condition 


mlavacca:
**What would you like to be added**:

A new `RouteConditionReason` should be added to state that the route is not accepted because a cross-namespace reference is not granted by a `ReferenceGrant`.

We already have a specular reason for the `ResolvedRefs` condition:
https://github.com/kubernetes-sigs/gateway-api/blob/7d408be7fc5389002858712fd7e7c0f14260274c/apis/v1beta1/shared_types.go#L276-L280

That should be set in such a case, jointly with a failed `accepted` condition, but no reason for the `accepted` condition is suitable to express it.

**Why this is needed**:

As of now, the *route `Accepted` condition can fail with the following reasons :
- `NotAllowedByListeners`
- `NoMatchingListenerHostname`
- `NoMatchingParent`
- `UnsupportedValue`

When a route references a gateway (through a `ParentRef`) in another namespace, and no `ReferenceGrant` allows such a reference, there is no suitable reason to use  for the `Accepted` condition to state it as failed.

Sajiyah-Salat:
This is somewhat confusing for me as a beginner. Can you please break it down or provide me some resources to get started  in this. Thank you.

mlavacca:
> This is somewhat confusing for me as a beginner. Can you please break it down or provide me some resources to get started in this. Thank you.

This issue is about adding a new reason to the set of already existing reasons for the `accepted` condition of the *route:
https://github.com/kubernetes-sigs/gateway-api/blob/7d408be7fc5389002858712fd7e7c0f14260274c/apis/v1beta1/shared_types.go#L205-L252
 This reason should be used to express that the Route has not been accepted because it has a `ParentRef` that references a Gateway in another namespace, but no ReferenceGrant allows such a reference.

Therefore, simply adding the new reason to the API and providing the documentation is the scope of the issue. Such a reason will likely be used in the conformance tests in subsequent PRs.

Sajiyah-Salat:
https://github.com/kubernetes-sigs/gateway-api/blob/7d408be7fc5389002858712fd7e7c0f14260274c/apis/v1beta1/shared_types.go#L212

In this block there are only three condititons while @mlavacca  says that there are four reasons and the extra reason is "NoMatchingParent"

[...]

PR #1672, opened 2023-01-26
Add condition in Failed accepted reasons

Sajiyah-Salat:
Issue #1670

The condition needs to be added in order to prevent crossname-space parentref

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -264,7 +264,7 @@ const (
 	// * "RefNotPermitted"
 	// * "InvalidKind"
 	// * "BackendNotFound"
-	// * "Cross-namespace ParentRef"
+	// * "ParentRefNotPermitted"
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
@@ -289,6 +289,12 @@ const (
 	// This reason is used with the "ResolvedRefs" condition when one of the
 	// Route's rules has a reference to a resource that does not exist.
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
+	
+	//This reason is used with the "ResolvedRef" condition when one of
+	// route has not been accepted because a ParentRef references a Gateway 
+	//in another namespace, but no ReferenceGrant allows such a reference.
+       RouteReasonRefNotPermitted RouteConditionReason = "ParentRefNotPermitted"
+
 )
 
 // RouteParentStatus describes the status of a route with respect to an
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -216,7 +216,7 @@ const (
 	// * "NoMatchingParent"
 	// * "UnsupportedValue"
 	// * "ParentRefNotPermitted"
-
+	//
 	// Possible reasons for this condition to be Unknown are:
 	//
 	// * "Pending"
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -264,6 +264,7 @@ const (
 	// * "RefNotPermitted"
 	// * "InvalidKind"
 	// * "BackendNotFound"
+	// * "Cross-namespace ParentRef"
 	//
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -293,7 +293,7 @@ const (
 	//This reason is used with the "ResolvedRef" condition when one of
 	// route has not been accepted because a ParentRef references a Gateway 
 	//in another namespace, but no ReferenceGrant allows such a reference.
-       RouteReasonRefNotPermitted RouteConditionReason = "ParentRefNotPermitted"
+       RouteReasonParentRefNotPermitted RouteConditionReason = "ParentRefNotPermitted"
 
 )
 
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -248,7 +248,7 @@ const (
 	// This reason is used with the "Accepted" condition when a value for an Enum
 	// is not recognized.
 	RouteReasonUnsupportedValue RouteConditionReason = "UnsupportedValue"
-	
+
 	// This reason is used with the "Accepted" condition when the route has not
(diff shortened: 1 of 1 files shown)


mlavacca (changes requested):
Well, the actual reason is missing here. You should add the Reason AND the documentation. Furthermore:
- the reason needs to be CamelCase
- just as a recommendation, I'd use something like `ParentRefNotPermitted`.

Sajiyah-Salat:
I will make changes accordingly. Thank you for your review.

Sajiyah-Salat:
I will add documentation and reason help me with programming line.

Sajiyah-Salat:
Why my build fails. let me know the reason also help me adding that one line. I am kinda confused in that.

shaneutt:
> Why my build fails. let me know the reason also help me adding that one line. I am kinda confused in that.

The reason I'm seeing is:

```console
apis/v1beta1/shared_types.go:296:8: RouteReasonRefNotPermitted redeclared in this block
```

Sajiyah-Salat:
I added this line just to have a demo

Sajiyah-Salat:
What to add instead of this

shaneutt:
That error suggests you've made a duplicate `RouteReasonRefNotPermitted` attribute on the struct, which in the diff I can see is true, there are two of them. Attributes on structs are unique, so the code wont compile if there's a duplicate, based on what I'm seeing I think you just need to rename the attribute you created. Perhaps `RouteReasonRefNotPermitted` -> `RouteReasonParentRefNotPermitted`? :thinking:

Sajiyah-Salat:
> That error suggests you've made a duplicate `RouteReasonRefNotPermitted` attribute on the struct, which in the diff I can see is true, there are two of them. Attributes on structs are unique, so the code wont compile if there's a duplicate, based on what I'm seeing I think you just need to rename the attribute you created. Perhaps `RouteReasonRefNotPermitted` -> `RouteReasonParentRefNotPermitted`? thinking

I have made changes as you suggested still its failing. its complex for me to understand

mlavacca (changes requested):
Your code is not properly formatted. Just run `go fmt ./...`

shaneutt:
> > That error suggests you've made a duplicate `RouteReasonRefNotPermitted` attribute on the struct, which in the diff I can see is true, there are two of them. Attributes on structs are unique, so the code wont compile if there's a duplicate, based on what I'm seeing I think you just need to rename the attribute you created. Perhaps `RouteReasonRefNotPermitted` -> `RouteReasonParentRefNotPermitted`? thinking
> 
> I have made changes as you suggested still its failing. its complex for me to understand

No worries! As @mlavacca mentioned it looks like you'll need to run `go fmt` (this is a tool that comes with Golang that automatically formats Go code for you). There's a blog post which describes it here:

https://go.dev/blog/gofmt

Please let us know how we can continue to be helpful, we really appreciate your contribution! :smiley:

Sajiyah-Salat:
I have formatted it as suggested. PTAL.

mlavacca (changes requested):
I just noticed you added the reason to the `ResolvedRefs` condition. Actually, it should be added to the list of the `Accepted` condition reasons. Furthermore, I would suggest rephrasing with something like:

```go
// This reason is used with the "Accepted" condition when the route has not
// been accepted by a Gateway because it has a cross-namespace parentRef,
// but no ReferenceGrant in the other namespace allows such a reference.
```

This is just a suggestion. Feel free to reword as you prefer while keeping the same meaning.

Sajiyah-Salat:
I have done the required changes. Please review this.

shaneutt:
> I have done the required changes. Please review this.

:+1: 

We do have a CI issue, looks like more `gofmt` changes are needed:

```
 apis/v1beta1/shared_types.go:255: File is not `gofmt`-ed with `-s` (gofmt) 
```

mlavacca:
@Sajiyah-Salat, the reason's comment is still the old one, please update it as I suggested in a few comments above (or with something similar but with the same meaning).

Sajiyah-Salat:
I run go fmt as suggested. still its failing. looks like its not my good day

apis/v1beta1/shared_types.go
```diff
@@ -215,7 +215,8 @@ const (
 	// * "NoMatchingListenerHostname"
 	// * "NoMatchingParent"
 	// * "UnsupportedValue"
-	//
+	// * "ParentRefNotPermitted"
+
```
shaneutt, line 219:
```suggestion
	// * "ParentRefNotPermitted"
	//
```

apis/v1beta1/shared_types.go
```diff
@@ -252,6 +253,11 @@ const (
 	// reconciled the route.
 	RouteReasonPending RouteConditionReason = "Pending"
 
+	//This reason is used with the "ResolvedRef" condition when one of
+	// route has not been accepted because a ParentRef references a Gateway
+	//in another namespace, but no ReferenceGrant allows such a reference.
+	RouteReasonParentRefNotPermitted RouteConditionReason = "ParentRefNotPermitted"
```
shaneutt, line 259:
```suggestion
	// This reason is used with the "ResolvedRef" condition when one of
	// route has not been accepted because a ParentRef references a Gateway
	// in another namespace, but no ReferenceGrant allows such a reference.
	RouteReasonParentRefNotPermitted RouteConditionReason = "ParentRefNotPermitted"
```

shaneutt (commented):
Looks like tests are passing now :+1: 

I have a couple of small suggestions, also @mlavacca if you wouldn't mind taking another look it would be appreciated.

mlavacca (commented):
I have a final request, then LGTM: can you please move the reason upward between `UnsupportedValue` and `pending`? Because `ParentRefNotPermitted` is used when the condition is false, while `pending` is when unknown.

Sajiyah-Salat:
after making any changes to code base we have to run ``go fmt ./...`` in order to format the code, right? Also Thanks a lot @mlavacca and @shaneutt. I have made many silly mistakes and you have helped me happily.

mlavacca:
> after making any changes to code base we have to run go fmt ./... in order to format the code, right?

Ideally yes. Doing it ensures that no `go fmt` errors will be triggered in the CI.

shaneutt (approved):
Thank you for your contribution!

/lgtm

Sajiyah-Salat:
This pr is special for me as its my first pr. Thank you @mlavacca and @shaneutt for being a great member.

PR #1694, opened 2023-02-02
`ParentRefNotPermitted` route reason in conformance test

### LINKED ISSUE #1688: Use the `ParentRefNotPermitted` reason in conformance tests

#1670 is going to add a new reason for the accepted condition that aims at stating that the route is not accepted because a cross-namespace reference is not granted by a `ReferenceGrant`.

Such a new reason should be checked in the conformance test `HTTPRouteInvalidCrossNamespaceParentRef`, where at the moment the condition is checked without caring about the reason: https://github.com/kubernetes-sigs/gateway-api/blob/8043dfb17a6d01526d7e4425db3e12081fb440f8/conformance/utils/kubernetes/helpers.go#L385-L388

**Why this is needed**:

This is needed to state that the `HTTPRoute` is not accepted because its parent reference is not explicitly granted by a specific `ReferenceGrant`.

mlavacca:
**What type of PR is this?**

When the `HTTPRoute` is not accepted because the cross-namespace `parentRef` is not granted, the specific reason `ParentRefNotPermitted` is used.

/kind feature

Fixes #1688 

```release-note
The conformance test HTTPRouteInvalidCrossNamespaceParentRef now requires the HTTPRoute accepted condition to be failing with the ParentRefNotPermitted reason.
```

### DIFF
--- a/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
+++ b/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
@@ -40,7 +40,7 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
 
 		// The Route must have an Accepted Condition with a ParentRefNotPermitted Reason.
-		t.Run("HTTPRoute with a cross-namespace ParentRef and no ReferenceGrants allows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
+		t.Run("HTTPRoute with a cross-namespace ParentRef where no ReferenceGrants allows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
 			acceptedCond := metav1.Condition{
 				Type:   string(v1beta1.RouteConditionAccepted),
 				Status: metav1.ConditionFalse,
--- a/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
+++ b/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
@@ -19,8 +19,10 @@ package tests
 import (
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
 	"sigs.k8s.io/gateway-api/conformance/utils/suite"
 )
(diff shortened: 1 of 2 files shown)


conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
```diff
@@ -37,6 +39,17 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
 
+		// The Route must have an Accepted Condition with a ParentRefNotPermitted Reason.
+		t.Run("HTTPRoute with a cross-namespace ParentRef and no ReferenceGrants allows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
```
shaneutt, line 43:
```suggestion
		t.Run("HTTPRoute with a cross-namespace ParentRef and no ReferenceGrants disallows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
```

shaneutt (changes requested):

conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
mlavacca: Why disallows? Isn't the `ReferenceGrant` intended to explicitly allow a reference?

shaneutt: I think I was reading this wrong, my bad.

shaneutt: What about this instead:

```suggestion
		t.Run("HTTPRoute with a cross-namespace ParentRef where no ReferenceGrants allows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
```

mlavacca: yeah, it sounds better :)

shaneutt (approved):
/lgtm

conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
```diff
@@ -37,6 +39,17 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
 
+		// The Route must have an Accepted Condition with a ParentRefNotPermitted Reason.
+		t.Run("HTTPRoute with a cross-namespace ParentRef where no ReferenceGrants allows such a reference, has an Accepted Condition with status False and Reason ParentRefNotPermitted", func(t *testing.T) {
```
skriss, line 43: Sorry I'm late here -- aren't cross-namespace Route->Listener/Gateway attachments governed by the Listener's [AllowedRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.AllowedRoutes), not ReferenceGrant? The test name could use an update to clarify that, if I'm not mistaken.

mlavacca: Honestly, I didn't realize that the grant was given by the allowedRoutes field, you are right. Other than changing the test name, the reason's documentation should also be updated, as it currently states https://github.com/kubernetes-sigs/gateway-api/blob/4ff6208a6a4238792e7477a03400933f5ae26c42/apis/v1beta1/shared_types.go#L252-L255

mlavacca: Concerning this, are you aware if there is any specific reason not to use the `ReferenceGrants` in this case as well, in favor of `AllowedRoutes`?

skriss: Actually, looking at this a bit more, could you clarify why the [NotAllowedByListeners](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteConditionReason) reason was not appropriate here? It seems to me like it was intended for this scenario.

skriss: ref. https://gateway-api.sigs.k8s.io/api-types/referencegrant/#exceptions

mlavacca: > Actually, looking at this a bit more, could you clarify why the [NotAllowedByListeners](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteConditionReason) reason was not appropriate here? It seems to me like it was intended for this scenario.

Yeah, this reason is not needed at all now that you point me out. The `NotAllowedByLysteners` reason should be used in this case.

shaneutt: Thanks for catching this. I admittedly looked at this the first time like "yep!" and now I'm like "oh..." :fried_egg: . Thank you for all your contributions @mlavacca, stuff happens! And thank you @skriss for identifying it!

I'm reverting this and the other relevant PR, appreciate everyone's help :vulcan_salute:

shaneutt: https://github.com/kubernetes-sigs/gateway-api/pull/1704 to follow up. Thank you!

PR #1695, opened 2023-02-03
Validate admission Kind first for efficiency

kerthcet:
Signed-off-by: Kante Yin <kerthcet@gmail.com>

**What type of PR is this?**

/kind cleanup

What commits in this PR:
1. Remove redundant `return` who conflicts with go-staticcheck
2. Validate admission Kind first for saving performance in Unmarshal. Maybe bit improves but better than not.

Fixes #

```release-note
None
```

### DIFF
--- a/pkg/admission/server.go
+++ b/pkg/admission/server.go
@@ -35,6 +35,8 @@ import (
 	v1b1Validation "sigs.k8s.io/gateway-api/apis/v1beta1/validation"
 )
 
+const admissionReview = "AdmissionReview"
+
 var (
 	scheme = runtime.NewScheme()
 	codecs = serializer.NewCodecFactory(scheme)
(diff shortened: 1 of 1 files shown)


gyohuangxin (approved):
LGTM, thanks for your improvement.

pkg/admission/server.go
```diff
@@ -96,25 +96,20 @@ var (
...
-	if err != nil {
-		return false
-	}
-	if msg.Kind != "AdmissionReview" {
-		return false
+	if err := json.Unmarshal(req, &msg); err != nil {
+		return false, err
 	}
-	return true
+	return msg.Kind == "AdmissionReview", nil
```
shaneutt, line 112: I think we should be able to reference a var or constant elsewhere for `"AdmissionReview"` rather than using a string literal for it right? :thinking:

pkg/admission/server.go
```diff
@@ -137,17 +132,24 @@ func ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	review := admission.AdmissionReview{}
-	err = json.Unmarshal(data, &review)
+	res, err := ensureKindAdmissionReview(data)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	if !ensureKindAdmissionReview(data) {
+	if !res {
```
shaneutt, line 140: It looks like we're superfluously running `json.Unmarshal` _twice_ for any particular data :thinking: 

I'm inclined to say we should just do it once around (what is currently) line 135, and then here just check the kind?

e.g. something like:

```go
	msg, err := marshal(data)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 	if msg.Kind != "AdmissionReview" {
```

shaneutt (commented):
Looking good, thank you for the contribution! :+1: 

Have a couple of smaller comments.

shaneutt (approved):
/lgtm

Thanks!

pkg/admission/server.go
kerthcet: Make sense, updated.

pkg/admission/server.go
kerthcet: Yeah, I thought about this, updated.

PR #1701, opened 2023-02-07
Gateway default conditions updated

mlavacca:
**What type of PR is this?**

/kind bug

The Gateway default conditions have been updated to always use `accepted` with `pending` reason and `programmed` with `pending` reason by default. Before this, the deprecated `NotReconciled` reason was still used by default when the `Gateway` status was not set.

Fixes #

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -41,7 +41,7 @@ type Gateway struct {
 
 	// Status defines the current state of Gateway.
 	//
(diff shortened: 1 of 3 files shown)


shaneutt (approved):
Generally this seems right to me, but I do think that there's a possibility this could trip up existing implementation, e.g. if they have a hard expectation somewhere on the previous default to start a reconciliation workflow, so if nothing else it may be better to release this as part of v0.7.0 or some other minor release. Otherwise LGTM but I'll let Rob or Nick give the final check.

youngnick (commented):
This looks fine to me now that we're effectively working on v0.7.0.

/lgtm

Issue #1703, opened 2023-02-07
Use of ApiMachinery Sets with structs containing references


shaneutt:
@skriss noticed while we were evaluating https://github.com/kubernetes-sigs/gateway-api/pull/1702 that some of the conformance test code wasn't working for him, particularly https://github.com/kubernetes-sigs/gateway-api/pull/1692#discussion_r1099010300 where the `sets` functionality from apimachinery was used with a struct with nested references. Sets does not currently support accurate `Has(obj)` functionality for structs with nested pointers, and so this test can not proceed properly.

Thank you @skriss! :partying_face:

shaneutt:
(@arkodg is working on this :+1: )

PR #1707, opened 2023-02-07
Add explicit equality checks for SupportedKinds

arkodg:
Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/1703

Signed-off-by: Arko Dasgupta <arko@tetrate.io>

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -32,7 +32,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -636,12 +635,20 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
 		// Ensure that the expected Listener.SupportedKinds items are present in actual Listener.SupportedKinds
 		// Find the items instead of performing an exact match of the slice because the implementation
 		// might support more Kinds than defined in the test
-		eSupportedKindsSet := sets.New(eListener.SupportedKinds...)
-		aSupportedKindsSet := sets.New(aListener.SupportedKinds...)
-		if !aSupportedKindsSet.IsSuperset(eSupportedKindsSet) {
-			t.Logf("Expected %v kinds to be present in SupportedKinds", eSupportedKindsSet.Difference(aSupportedKindsSet))
-			return false
+		for _, eKind := range eListener.SupportedKinds {
+			found := false
+			for _, aKind := range aListener.SupportedKinds {
+				if *eKind.Group == *aKind.Group && eKind.Kind == aKind.Kind {
+					found = true
+					break
+				}
+			}
+			if !found {
+				t.Logf("Expected Group:%s Kind:%s to be present in SupportedKinds", *eKind.Group, eKind.Kind)
+				return false
+			}
 		}
+
 		if aListener.AttachedRoutes != eListener.AttachedRoutes {
 			t.Logf("Expected AttachedRoutes to be %v, got %v", eListener.AttachedRoutes, aListener.AttachedRoutes)
 			return false
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -637,7 +637,16 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
 		// might support more Kinds than defined in the test
 		for _, eKind := range eListener.SupportedKinds {
 			found := false
+
 			for _, aKind := range aListener.SupportedKinds {
+				if eKind.Group == nil {
+					eKind.Group = (*v1beta1.Group)(&v1beta1.GroupVersion.Group)
+				}
+
+				if aKind.Group == nil {
+					aKind.Group = (*v1beta1.Group)(&v1beta1.GroupVersion.Group)
+				}
+
 				if *eKind.Group == *aKind.Group && eKind.Kind == aKind.Kind {
 					found = true
 					break
--- a/conformance/utils/kubernetes/helpers_test.go
+++ b/conformance/utils/kubernetes/helpers_test.go
@@ -1,9 +1,12 @@
 /*
-Copyright 2023 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 )
--- a/conformance/utils/kubernetes/helpers_test.go
+++ b/conformance/utils/kubernetes/helpers_test.go
@@ -1,0 +1,160 @@
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+func TestSupportedKindsMatch(t *testing.T) {
+	tests := []struct {
+		name     string
+		expected []v1beta1.ListenerStatus
+		actual   []v1beta1.ListenerStatus
+		want     bool
+	}{
+		{
+			name: "expected empty and actual is non empty",
+			expected: []v1beta1.ListenerStatus{
+				{
+					Name:           v1beta1.SectionName("https"),
+					SupportedKinds: []v1beta1.RouteGroupKind{},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionResolvedRefs),
+						Status: metav1.ConditionFalse,
+						Reason: string(v1beta1.ListenerReasonRefNotPermitted),
+					}},
+				},
+			},
+			actual: []v1beta1.ListenerStatus{
+				{
+					Name: v1beta1.SectionName("https"),
+					SupportedKinds: []v1beta1.RouteGroupKind{{
+						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+						Kind:  v1beta1.Kind("HTTPRoute"),
+					}},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionResolvedRefs),
+						Status: metav1.ConditionFalse,
+						Reason: string(v1beta1.ListenerReasonRefNotPermitted),
+					}},
+				},
+			},
+			want: false,
+		},
+		{
+			name: "expected and actual are equal",
+			expected: []v1beta1.ListenerStatus{
+				{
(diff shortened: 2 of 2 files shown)


shaneutt:
Hey :wave: I had assigned myself this and was working on an implementation, however I'm very happy to differ to you.

Two extra things I was doing:

1. adding unit tests. This whole problem would have been avoided if we had unit tests (which, to be clear, is on us as maintainers)
2. add an error condition for duplicate SupportedKinds: technically it's never "correct" to have duplicates in this list, so this is a place we can helpfully let implementations know if they're doing something weird.

Looks like you already started writing some unit tests, perhaps to test the original issue. I don't mean to put extra work on you, but if you wouldn't mind adding unit tests for the relevant functionality I think that can help us from having problems here again. If you're not up for this, please do say so I'm happy to take this on.

In any case, very much appreciate the help, thank you! Let me know how I can be helpful to you :vulcan_salute:

skriss:
Can confirm this fixes the issue for Contour, thanks @arkodg!

arkodg:
> Hey 👋 I was working on an implementation of this as well, however I'm very happy to differ to you.
> 
> Two extra things I was doing:
> 
> 1. adding unit tests. This whole problem would have been avoided if we had unit tests (which, to be clear, is on us as maintainers)
> 2. add an error condition for duplicate SupportedKinds: technically it's never "correct" to have duplicates in this list, so this is a place we can helpfully let implementations know if they're doing something weird.
> 
> Looks like you already started writing some unit tests, perhaps to test the original issue. I don't mean to put extra work on you, but if you wouldn't mind adding unit tests for the relevant functionality I think that can help us from having problems here again. If you're not up for this, please do say so I'm happy to take this on.
> 
> In any case, very much appreciate the help, thank you! Let me know how I can be helpful to you 🖖

hi @shaneutt , ive added tests to this PR, haven't handled the duplicate case in this PR, please feel free to take care of it in a future PR

conformance/utils/kubernetes/helpers.go
```diff
@@ -636,12 +635,20 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
...
+		for _, eKind := range eListener.SupportedKinds {
+			found := false
+			for _, aKind := range aListener.SupportedKinds {
+				if *eKind.Group == *aKind.Group && eKind.Kind == aKind.Kind {
+					found = true
+					break
+				}
+			}
+			if !found {
+				t.Logf("Expected Group:%s Kind:%s to be present in SupportedKinds", *eKind.Group, eKind.Kind)
```
shaneutt, line 647: It's a bit of a nit, but in multiple places here we're de-referencing `Group` and the attribute is technically optional and _could_ be `nil`.

We colloquially expect any `RouteGroupKinds` that come through here will have been initialized via the API first so generally we wouldn't expect this to be an issue because of CRD defaults that would be applied. That said, some situation where someone later tries to use these helper functions in a more derived way would leave them with a segmentation fault that might be little terse.

One thing you could do would be just initialize them with the default value before using them if needed:

```diff
diff --git a/conformance/utils/kubernetes/helpers.go b/conformance/utils/kubernetes/helpers.go
index a0926f9d..33a0c7a8 100644
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -638,6 +638,14 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
                for _, eKind := range eListener.SupportedKinds {
                        found := false
                        for _, aKind := range aListener.SupportedKinds {
+                               if eKind.Group == nil {
+                                       eKind.Group = (*v1beta1.Group)(&v1beta1.GroupVersion.Group)
+                               }
+
+                               if aKind.Group == nil {
+                                       aKind.Group = (*v1beta1.Group)(&v1beta1.GroupVersion.Group)
+                               }
+
                                if *eKind.Group == *aKind.Group && eKind.Kind == aKind.Kind {
                                        found = true
                                        break
```

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+func TestSupportedKindsMatch(t *testing.T) {
```
shaneutt, line 12:
```suggestion
func Test_listenersMatch(t *testing.T) {
```

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,160 @@
...
+			name: "expected empty and actual is non empty",
+			expected: []v1beta1.ListenerStatus{
+				{
+					Name:           v1beta1.SectionName("https"),
+					SupportedKinds: []v1beta1.RouteGroupKind{},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionResolvedRefs),
+						Status: metav1.ConditionFalse,
+						Reason: string(v1beta1.ListenerReasonRefNotPermitted),
+					}},
```
robscott, line 29: Do these ever actually vary? Can we either omit them entirely or use something like a `commonConditions` var?

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,160 @@
...
+				},
+			},
+			actual: []v1beta1.ListenerStatus{
+				{
+					Name: v1beta1.SectionName("https"),
```
robscott, line 34: Same question for name - is this actually needed?

robscott (commented):
Thanks @arkodg!

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,160 @@
...
+				},
+			},
+			want: true,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			assert.Equal(t, test.want, listenersMatch(t, test.expected, test.actual))
```
shaneutt, line 157: One thing I noticed is that because we only test for `bool` return values, and we don't have a mechanism here to verify the logging output our checking of results is a bit limited. That said however, this isn't something I think we need to resolve right here and now because it would grow the scope of this PR considerably to refactor this now, and this is blocking the `v0.6.1` release. As such, as per @mikemorris' comments throughout, I'll make a follow-up issue to do some refactoring in here in a later iteration.

shaneutt (commented):
No hard blockers, though I do have a few comments to be resolved before we merge.

Thank you so much for jumping in and adding this fix @arkodg we appreciate you! :vulcan_salute: 

cc @robscott / @youngnick to take a look as well.

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+func TestSupportedKindsMatch(t *testing.T) {
+	tests := []struct {
```
shaneutt, line 13: Somehow I had a comment here that got dropped from my previous review.

While looking at coverage for this, there are a few cases we're not covering:

```golang
        if len(expected) != len(actual) {
                t.Logf("Expected %d Gateway status listeners, got %d", len(expected), len(actual))
                return false
        }
```

```golang
                if aListener.Name != eListener.Name {
                        t.Logf("Name doesn't match")
                        return false
                }
```

```golang
                if aListener.AttachedRoutes != eListener.AttachedRoutes {
                        t.Logf("Expected AttachedRoutes to be %v, got %v", eListener.AttachedRoutes, aListener.AttachedRoutes)
                        return false
                }
```

```golang
                if !conditionsMatch(t, eListener.Conditions, aListener.Conditions) {
                        t.Logf("Expected Conditions to be %v, got %v", eListener.Conditions, aListener.Conditions)
                        return false
                }
```

e.g.:

`go test -coverprofile cover.out ./conformance/utils/kubernetes/...`
`go tool cover -html=cover.out -o cover.html`
`firefox cover.html`

conformance/utils/kubernetes/helpers_test.go
shaneutt: https://github.com/kubernetes-sigs/gateway-api/issues/1708 was created to resolve this comment :+1:

conformance/utils/kubernetes/helpers_test.go
arkodg: yah, these should be tracked with a GH issue and fixed in a follow up PR :)

conformance/utils/kubernetes/helpers_test.go
arkodg: can make this change, but hope it doesnt mean all test cases associated with `listenersMatch` need to be handled in this PR :)

conformance/utils/kubernetes/helpers.go
arkodg: since `Group` is optional, and if you believe GAPI code might skip populating `Group`, then it makes more sense for the `Group` check to be skipped imho

conformance/utils/kubernetes/helpers_test.go
arkodg: sure, will rm

conformance/utils/kubernetes/helpers_test.go
arkodg: sure will rm

arkodg:
tested this PR in https://github.com/envoyproxy/gateway/pull/990

conformance/utils/kubernetes/helpers.go
shaneutt: I don't think _skipping_ the group check would be right, because if you did that you could have two kinds of the same name, with two different group names that would be considered matching right?

conformance/utils/kubernetes/helpers_test.go
shaneutt: If this is pushing it for your available time, then I can agree to making it a follow-up as what we have here would appear to be a significant improvement over the previously broken logic. Thank you for your time so far :bow:

shaneutt: Added to https://github.com/kubernetes-sigs/gateway-api/issues/1708 :+1:

conformance/utils/kubernetes/helpers.go
arkodg: also previously the check was an exact Match and all implementations abided by it, will make this change if thats your preference

shaneutt: Unless I've missed something, this just means we'll provide the already configured default behavior in edge cases rather than a segfault, so yeah if it's not a problem I would appreciate it thank you.

arkodg: sg, will add

shaneutt (approved):
Looking good.

cc @robscott or @youngnick for the final LGTM please.

robscott:
Thanks @arkodg!

/lgtm
/ok-to-test

conformance/utils/kubernetes/helpers_test.go
```diff
@@ -0,0 +1,121 @@
+package kubernetes
```
robscott, line 1: This file needs a boilerplate header, for example https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1alpha2/doc.go#L1-L15C3 (but with 2023 as the year)

arkodg: ack

shaneutt:
@arkodg looks like imports need cleanup:

```
conformance/utils/kubernetes/helpers_test.go:19: File is not `goimports`-ed with -local sigs.k8s.io/gateway-api (goimports)
```

conformance/utils/kubernetes/helpers_test.go
arkodg: the tool still thinks its 2022, will raise a GH issue for it

arkodg:
> @arkodg looks like imports need cleanup:
> 
> ```
> conformance/utils/kubernetes/helpers_test.go:19: File is not `goimports`-ed with -local sigs.k8s.io/gateway-api (goimports)
> ```

thanks fixed it, there was also an issue around the date mentioned in LICENSE - `hack/verify-boilerplate.sh` didn't like it when I set the year to `2023`, set it to `2022` for now

robscott:
Thanks @arkodg!

/lgtm

PR #1710, opened 2023-02-08
add additional unit test case for #1703

skriss:

**What type of PR is this?**
/kind cleanup

Adds one more test case to specifically test the bug from #1703.

This test case fails with the old code, and passes with the new code from #1707.

Fixes #

```release-note
NONE
```

cc @shaneutt


arkodg (approved):
LGTM, thanks !

shaneutt (approved):
Thanks Skriss!

/lgtm

Issue #1709, opened 2023-02-07
Conformance Levels


shaneutt:
We've referred to the concept of adding "conformance levels" to Gateway API for quite some time.

The concept is effectively that we would have named groupings of conformance tests that can be easily selected for implementations, and correspond (perhaps somewhat loosely) with the "category" of implementation.

While in time we may have many "conformance levels", these are high level ones we want initially (details intentionally sparing right now):

- Layer 7
- Layer 4
- Mesh

The purpose of this issue is to track this desire and at least get a GEP created for the details.

Riaankl:
@shaneutt this would be a topic of SIG Arch meeting where we discuss Conformance topics. Next [meeting is tomorrow](http://bit.ly/sig-architecture).

shaneutt:
@Riaankl sounds good, I will bring this up in that forum as we progress. However I will not be available for the meeting tomorrow, so I'll intend on a later date meeting which will provide some time for this to grow anyhow.

shaneutt:
My intention right now is to show up at the March 16th "Enhancements Subproject" meeting to talk about Gateway API conformance profiles, if anyone else wants to join me there.

shaneutt:
I'm going to try and make it to the March 9th meeting, but I don't know for sure whether I can make it to that one yet.

Riaankl:
@shaneutt IMO any SIG Arch meeting would be good for this.
To get more focus on your topic I would also suggest to share in [SIG Arch mailing list](https://groups.google.com/forum/#!forum/kubernetes-sig-architecture) before you attend the meeting

shaneutt:
Sounds good :+1:

shaneutt:
At this point we've achieved everything that we considered a _requirement_ for our GA release, so we no longer need to track this for GA.

We will continue working on the relevant sub-issues, there are just no remaining GA blockers.

PR #1714, opened 2023-02-08
conformance: check reason in HTTPRouteInvalidCrossNamespaceParentRef test

### LINKED ISSUE #1705: HTTPRouteInvalidCrossNamespaceParentRef check for NotAllowedByListeners

Noted by @skriss, the `HTTPRouteInvalidCrossNamespaceParentRef` test should be checking for `NotAllowedByListeners`.

skriss:
Checks for the NotAllowedByListeners reason on
the HTTPRoute's Accepted: false condition in
the HTTPRouteInvalidCrossNamespaceParentRef
test.

**What type of PR is this?**
/area conformance

Checks for the NotAllowedByListeners reason on
the HTTPRoute's Accepted: false condition in
the HTTPRouteInvalidCrossNamespaceParentRef
test.

Fixes #1705

```release-note
Checks for the NotAllowedByListeners reason on
the HTTPRoute's Accepted: false condition in
the HTTPRouteInvalidCrossNamespaceParentRef
conformance test.
```

### DIFF
--- a/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
+++ b/conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
@@ -19,8 +19,10 @@ package tests
 import (
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
 	"sigs.k8s.io/gateway-api/conformance/utils/suite"
 )
@@ -36,6 +38,20 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
 		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
+
+		// When running conformance tests, implementations are expected to have visibility across all namespaces, and
+		// must be setting this condition on routes that are not allowed. However, outside of conformance testing,
+		// it's also valid for implementations to run in modes where they only have access to a limited subset of
+		// namespaces, in which case they are not obligated to populate this condition on routes they cannot access.
+		t.Run("HTTPRoute should have an Accepted: false condition with reason NotAllowedByListeners", func(t *testing.T) {
+			acceptedCond := metav1.Condition{
+				Type:   string(v1beta1.RouteConditionAccepted),
+				Status: metav1.ConditionFalse,
+				Reason: string(v1beta1.RouteReasonNotAllowedByListeners),
+			}
+
+			kubernetes.HTTPRouteMustHaveCondition(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN, acceptedCond)
+		})
 
 		t.Run("Route should not have Parents set in status", func(t *testing.T) {
 			kubernetes.HTTPRouteMustHaveNoAcceptedParents(t, suite.Client, suite.TimeoutConfig, routeNN)


conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
```diff
@@ -37,8 +39,14 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
 
-		t.Run("Route should not have Parents set in status", func(t *testing.T) {
-			kubernetes.HTTPRouteMustHaveNoAcceptedParents(t, suite.Client, suite.TimeoutConfig, routeNN)
+		t.Run("HTTPRoute should have an Accepted: false condition with reason NotAllowedByListeners", func(t *testing.T) {
```
shaneutt, line 42: So we've removed the check for `kubernetes.HTTPRouteMustHaveNoAcceptedParents(t, suite.Client, suite.TimeoutConfig, routeNN)` but isn't that check still valid, even if we also add the check to ensure the status was updated correct? :thinking:

skriss: Fair enough, I thought it was mostly redundant since it was checking for a parent with an "Accepted: false" condition (with any reason), but I guess it does also validate that no _other_ parents accept the route. Will add back

shaneutt: Sounds good, thank you :+1:

shaneutt (approved):
Looks good thank you :+1:

robscott:
I _think_ this is OK, but we need to be careful here. As I recall, we don't require implementations to populate conditions on HTTPRoutes making cross-namespace references because some implementations may be running with limited access to namespace(s). We want to allow implementations to run in a restricted access mode, or even a single-namespace mode, and still allow them to be conformant.

With that said, I think it's reasonable to require implementations running conformance tests to have cluster-wide visibility, so I think this LGTM, just need to be _very_ clear that it is valid not to set this condition if you're running in a mode where you don't have visibility to all namespaces.

conformance/tests/httproute-invalid-cross-namespace-parent-ref.go
```diff
@@ -37,6 +39,16 @@ var HTTPRouteInvalidCrossNamespaceParentRef = suite.ConformanceTest{
 		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: "gateway-conformance-infra"}
 		routeNN := types.NamespacedName{Name: "invalid-cross-namespace-parent-ref", Namespace: "gateway-conformance-web-backend"}
 
+		t.Run("HTTPRoute should have an Accepted: false condition with reason NotAllowedByListeners", func(t *testing.T) {
```
robscott, line 42: Maybe add a comment above this that clarifies that:

1. We expect all implementations running conformance tests to have full namespace visibility and when they are running in that mode, this condition needs to be set.
2. It is valid for controllers to run in modes where they have access to a subset of namespace(s). In that case, they are not obligated to populate this condition on routes they can't see.

skriss: Ah, I hadn't considered that use case. If you think it's better to not merge this in order to continue to accommodate that use case, that's OK with me for now; otherwise, I'll add the comment you suggest. I imagine at some point we might need to incorporate that into a conformance profile (#1709)?

robscott: Thanks @skriss! I think it's good to add this comment, just as long as we have a clarifying comment. I think even the implementations that were deployed with limited access by default would likely have a way to watch all namespaces for the sake of conformance tests, just want to make sure it's clear that's not required for every use of Gateway API.

skriss: Comment added!

robscott:
Thanks @skriss!

/lgtm

dprotaso:
Was just testing to see if the plugin was active

robscott:
@dprotaso x-ref https://github.com/kubernetes-sigs/gateway-api/issues/1800, probably want to enable some more plugins, agree that cherry-pick would be very helpful.

shaneutt:
> @dprotaso x-ref #1800, probably want to enable some more plugins, agree that cherry-pick would be very helpful.

We talked about this a bit in Slack: https://kubernetes.slack.com/archives/CR0H13KGA/p1678891960116929

The manual cherry-picking is fast and doesn't really bother me, but I too have been thinking about this plugin as well as some others (like triage) that we should be turning on to make things just that little bit more convenient.

shaneutt:
I've updated https://github.com/kubernetes-sigs/gateway-api/issues/1800 to include this.

PR #1732, opened 2023-02-15
Requiring "Programmed" condition in conformance tests

### LINKED ISSUE #1722: Add missing GatewayConditionType conformance test
As per https://github.com/kubernetes-sigs/gateway-api/issues/1454, 

> - [ ] Update the description for the "Ready" condition to note the stricter behavior and that it is now extended conformance on:
>  - [ ] ListenerConditionType
>  - [ ] GatewayConditionType

`ListenerConditionType` and `GatewayConditionType` updated the description but there are no conformance test yet.

Here is the requet for the conformance test of `GatewayConditionType`.

robscott:
**What type of PR is this?**
/kind test
/area conformance

This adds a requirement in conformance tests that Gateways must be "Programmed" (previously this was just "Accepted"). Note that this only covers the top level "Programmed" condition, not each Listener.

Fixes #1722

```release-note
Conformance: Gateways must publish the "Programmed" condition.
```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -174,10 +174,10 @@ func FilterStaleConditions(obj metav1.Object, conditions []metav1.Condition) []m
 	return stale
 }
 
-// NamespacesMustBeAccepted waits until all Pods are marked ready and all Gateways
-// are marked accepted in the provided namespaces. This will cause the test to
-// halt if the specified timeout is exceeded.
-func NamespacesMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, namespaces []string) {
+// NamespacesMustBeReady waits until all Pods are marked Ready and all Gateways
+// are marked Accepted and Programmed in the specified namespace(s). This will
+// cause the test to halt if the specified timeout is exceeded.
+func NamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, namespaces []string) {
 	t.Helper()
 
 	waitErr := wait.PollImmediate(1*time.Second, timeoutConfig.NamespacesMustBeReady, func() (bool, error) {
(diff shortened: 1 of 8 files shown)


robscott:
I believe that every time we make this check in conformance tests, we're also expecting the Gateway(s) to be programmed. If I've missed anything, let me know. I'm going to put a hold on this because I'd like to get sign off from a few people before merging. I know this will _completely_ break conformance tests for some implementations because it's a change at such a high level.

/hold
/cc @arkodg @howardjohn @mlavacca @skriss @sunjayBhatia

arkodg:
@robscott from a quick glance it looks like this case is being addressed in Envoy Gateway [here](https://github.com/envoyproxy/gateway/blob/9785b4192ae47f5ed60d591deb654a2ca4155294/internal/status/conditions.go#L75) 
you could try testing it out using steps outlined [here](https://github.com/kubernetes-sigs/gateway-api/issues/1734#issuecomment-1432310808)

shaneutt (approved):
This seems like the right change to me, but definitely would like to invite some more people to check this out before we merge it.

conformance/utils/kubernetes/helpers.go
```diff
@@ -203,6 +203,12 @@ func NamespacesMustBeAccepted(t *testing.T, c client.Client, timeoutConfig confi
 					t.Logf("%s/%s Gateway not ready yet", ns, gw.Name)
```
skriss, line 203: This might be slightly more clear:
```suggestion
					t.Logf("%s/%s Gateway not accepted yet", ns, gw.Name)
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -203,6 +203,12 @@ func NamespacesMustBeAccepted(t *testing.T, c client.Client, timeoutConfig confi
...
 					return false, nil
 				}
+
+				// Passing an empty string as the Reason means that any Reason will do.
+				if !findConditionInList(t, gw.Status.Conditions, string(v1beta1.GatewayConditionProgrammed), "True", "") {
+					t.Logf("%s/%s Gateway not ready yet", ns, gw.Name)
```
skriss, line 209: Similarly:
```suggestion
					t.Logf("%s/%s Gateway not programmed yet", ns, gw.Name)
```

mlavacca (approved):
This change looks good to me 👍

skriss (commented):
Thanks @robscott, this mostly looks good from Contour's perspective; unfortunately it's causing us to start failing the gateway-observed-generation-bump test because we don't yet support the addition of the second HTTP Listener port, but that's an issue for us rather than the conformance test.

conformance/utils/kubernetes/helpers.go
robscott: Updated, thanks!

conformance/utils/kubernetes/helpers.go
robscott: Updated, thanks!

robscott:
Thanks everyone! Sounds like this is safe enough to remove the hold. Next LGTM will merge this.

/hold cancel

shaneutt (approved):
/lgtm

PR #1740, opened 2023-02-17
Clarify semantics when port_redirect is unset

arkodg:
Relates to https://github.com/kubernetes-sigs/gateway-api/discussions/1725

/kind documentation

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -901,7 +901,8 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, port (if specified) of the request is used.
+	// When empty, port (if specified) in the `Host` header
+	// of the request is used.
 	//
 	// Support: Extended
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -902,8 +902,10 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, the Gateway listener port is used (except
-	// when it is a default port - 80 or 443).
+	// When empty, the Gateway Listener port is used.
+	// In all cases, when the protocol is http and the port is port 80,
+	// then the port must be omitted. Similarly, when the protocol is
+	// https and the port is port 443, then the port must also be omitted.
 	//
 	// Support: Extended
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -882,7 +882,8 @@ type HTTPRequestRedirectFilter struct {
 
 	// Hostname is the hostname to be used in the value of the `Location`
 	// header in the response.
-	// When empty, the hostname of the request is used.
+	// When empty, the hostname in the `Host` header
+	// of the request is used.
 	//
 	// Support: Core
 	//
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -902,8 +902,8 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, port (if specified) in the `Host` header
-	// of the request is used.
+	// When empty, the Gateway listener port is used (except
+	// when it is a default port - 80 or 443).
 	//
 	// Support: Extended
 	//
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -571,9 +571,13 @@ spec:
                                     port:
                                       description: "Port is the port to be used in
                                         the value of the `Location` header in the
-                                        response. When empty, the Gateway listener
-                                        port is used (except when it is a default
-                                        port - 80 or 443). \n Support: Extended"
+                                        response. When empty, the Gateway Listener
+                                        port is used. In all cases, when the protocol
+                                        is http and the port is port 80, then the
+                                        port must be omitted. Similarly, when the
+                                        protocol is https and the port is port 443,
+                                        then the port must also be omitted. \n Support:
+                                        Extended"
                                       format: int32
                                       maximum: 65535
                                       minimum: 1
@@ -1192,8 +1196,11 @@ spec:
                               port:
                                 description: "Port is the port to be used in the value
                                   of the `Location` header in the response. When empty,
-                                  the Gateway listener port is used (except when it
-                                  is a default port - 80 or 443). \n Support: Extended"
+                                  the Gateway Listener port is used. In all cases,
+                                  when the protocol is http and the port is port 80,
+                                  then the port must be omitted. Similarly, when the
+                                  protocol is https and the port is port 443, then
+                                  the port must also be omitted. \n Support: Extended"
                                 format: int32
                                 maximum: 65535
                                 minimum: 1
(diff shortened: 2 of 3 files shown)


apis/v1beta1/httproute_types.go
```diff
@@ -901,7 +901,8 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, port (if specified) of the request is used.
+	// When empty, port (if specified) in the `Host` header
```
howardjohn, line 904: Copying my comment from discussion here, I would push for this to be the parent (gateway listener) port personally. 

On a complex multi-hop network path, there is no way we can know the "right" port to use. IMO, using the most stable thing we have - parent port - is a reasonable compromise. We should indicate to users that setting port explicitly is a best practice in these scenarios.

I don't feel too strongly if everyone else likes Host and envoy implements it.

arkodg: when the `hostname` setting is empty, the hostname from the `Host` header is used, so imho
when the `port` setting is empty, the port from the `Host` header should be used :)

arkodg: regarding the current implementation limitation - this should be fixed in envoy soon https://github.com/envoyproxy/envoy/pull/25573 (thanks @cpakulski ! )

howardjohn: Here is some prior art in the "L4" space:

```
$ curl -s wikipedia.org -H "Host: wikipedia.org:8080" -v |& grep Location
< Location: https://wikipedia.org/
$ curl -s google.com -H "Host: google.com:8080" -v |& grep Location
< Location: http://www.google.com/
$ curl -s gateway-api.org -H "Host: gateway-api.org:8080" -v |& grep Location
< Location: http://gateway-api.sigs.k8s.io
$ curl httpbin.org/relative-redirect/1 -H "Host: httpbin.org:8080" -v -s |& grep Location
< Location: /get
$ curl httpbin.org/redirect-to?url=http://a -H "Host: httpbin.org:8080" -v -s |& grep Location
< Location: http://a
```

No one cares about the port in Host that I have found.

robscott: I tend to agree with @howardjohn here. As far as I can tell, defaulting to the port of the Gateway listener (or Service for GAMMA) seems like the most broadly implementable solution. The current spec is unfortunately too vague, so a big thanks for @arkodg for highlighting this and working to clarify the spec.

arkodg: thanks for running these tests, if we go with above, then the default semantics should be that the Location header contains the stripped hostname (no port) when port redirect is unset

robscott:
Thanks @arkodg! We definitely need to clarify the spec here. If we're not careful, we may end up breaking some implementations along the way, so need to make sure we get as much feedback as possible before merging. Adding a hold until we've had more implementers weigh in.

/cc @skriss @shaneutt @youngnick 
/hold
/ok-to-test

apis/v1beta1/httproute_types.go
arkodg: @robscott from the current docs, its much easier to imply that the original author was referring to L7 Host Header due to this phrase `port (if specified)`, hence this PR simply makes it more clear
any other change makes it an API change (which is fine, but needs more scrutiny)

shaneutt (changes requested):

apis/v1beta1/httproute_types.go
arkodg: are you suggesting that when `port-redirect` is unset, and the Host header contains `example.com` but the `port` within the `listener` in the `Gateway` resource is set to `10080`, the Location header should be `example.com:10080` ?
this conflicts with https://github.com/kubernetes-sigs/gateway-api/pull/1740#discussion_r1113268023 as well as the suggested PR change

arkodg: also would be great if @hbagdi ([author](https://github.com/kubernetes-sigs/gateway-api/pull/683)) can help clarify the default semantics

shaneutt: So far, I do believe that's what's being suggested by John, Rob and myself. I can be persuaded, but the listenerPort fallback does seem more broadly implementable.

arkodg:
@shaneutt updated based on the guidance received by @youngnick yesterday

arkodg:
cross linking [this ](https://kubernetes.slack.com/archives/CR0H13KGA/p1677596579333419)slack thread where a user was unable to redirect correctly because the default port  443 was getting appended to the `Location` header.

apis/v1beta1/httproute_types.go
arkodg: updated the language to point to listener port, thanks for the discussion !

apis/v1beta1/httproute_types.go
```diff
@@ -901,7 +902,8 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, port (if specified) of the request is used.
+	// When empty, the Gateway listener port is used (except
+	// when it is a default port - 80 or 443).
```
howardjohn, line 906: I don't think this "except ..." part is clear. I actually don't know what it means

arkodg: I was trying to say this in a succinct way - When the Port is unspecified, The port from the Gateway Listener will be appended to the Location Header, except when the Gateway Listener port is a default/standard port (80 for HTTP and 443 for HTTPS ).
Any suggestions here ?

youngnick: I'd say something like:
```
In all cases, when the protocol is http and the port is port 80, then the port must be omitted.
Similarly, when the protocol is https and the port is port 443, then the port must also be omitted.
```

arkodg: hey @youngnick are you suggesting that even when the redirect port is explicitly set to `80` by the user, it is not appended to the `Location` header ?

youngnick: Yes, the Location header should, like all URLs, only show the port when it's relevant, which is when it's _not_ 80 for http, or 443 for https.

So if a request comes in on `http://foo.bar.com:8080/baz` and is rewritten to `http://quux.com/baz`, then you can tell the port is 80 because it's not present.

arkodg: thanks for clarification, will update the PR !

apis/v1beta1/httproute_types.go
```diff
@@ -901,7 +902,10 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
-	// When empty, port (if specified) of the request is used.
+	// When empty, the Gateway Listener port is used.
+	// In all cases, when the protocol is http and the port is port 80,
+	// then the port must be omitted. Similarly, when the protocol is
+	// https and the port is port 443, then the port must also be omitted.
```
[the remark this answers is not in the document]
arkodg: should I change `must` to `will` so it caters to both parties ?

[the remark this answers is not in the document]
robscott: I think what we're trying to say is the following:

```
If the port is not specified in the Redirect Filter, implementations should use the port from the listener for redirects _unless_ the Listener or Redirect Filter protocol is HTTP or HTTPS, in which case it should not specify a port in the redirect configuration.
```

I'm not actually sure that's what we're trying to say, and what I wrote above is still rather confusing, but maybe it's at least clear that the guidance is intended for implementers?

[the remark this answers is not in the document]
robscott: I think that overall idea makes sense, it's just not clear to me in the current variation of the spec, moving this to https://github.com/kubernetes-sigs/gateway-api/issues/1806 so we don't lose track of the discussion.

PR #1745, opened 2023-02-20
chore: cleanup err check scoped down

mlavacca:
The error triggered by the test cleaner makes the test fail only if the error is !NotFound.

**What type of PR is this?**

The cleanup might be failing because some resources were already previously deleted. For example, the tests `gateway-secret-reference-grant-specific` and `gateway-secret-reference-grant-all-in-namespaces` create a gateway named in the same way (i.e., the second apply is an update), but the cleaner is called twice. When the cleaner is called the 2nd time on that resource, the cleanup fails because the gateway does not exist anymore, making the test fail.
This PR introduces a check on the error triggered by the cleanup function. If `NotFound`, the cleanup error can be tolerated.

Fixes #

```release-note
NONE
```


shaneutt (approved):
Sounds good, thanks!

/lgtm

PR #1753, opened 2023-02-22
drop default method matches from GRPCRoute

### LINKED ISSUE #1751: GRPCRoute match defaulting prevents matching all requests
**What happened**:

Created a `GRPCRoute` like the following:

```yaml
kind: GRPCRoute
metadata:
  name: grpcroute
spec:
  parentRefs:
  - name: gateway
  rules:
  - backendRefs:
    - name: grpc-echo
      port: 9000
```
With the intention of matching all requests, [per the spec](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1alpha2.GRPCRouteRule):

> If no matches are specified, the implementation MUST match every gRPC request.

This GRPCRoute then ended up not being Accepted by my implementation, because [the API defaulting added a default method match](https://github.com/kubernetes-sigs/gateway-api/blame/main/apis/v1alpha2/grpcroute_types.go#L208), with no service or method specified, which is [not a valid configuration](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.GRPCMethodMatch):
> At least one of Service and Method MUST be a non-empty string.

I think this default should be removed; specifying no matches is valid configuration for matching all requests and should be left as-is.

If folks agree this should be changed, I'll submit a PR for it.

skriss:
**What type of PR is this?**
/kind bug

GRPCRoute: drops method match defaults to allow for either no matches to be specified in order to match all requests, or header-only matches.

Fixes #1751 

```release-note
Removes GRPCRoute method match defaulting to allow for matching all requests, or matching only by header.
```

### DIFF
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -205,7 +205,6 @@ type GRPCRouteRule struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{method: {type: "Exact"}}}
 	Matches []GRPCRouteMatch `json:"matches,omitempty"`
 
 	// Filters define the filters that are applied to requests that match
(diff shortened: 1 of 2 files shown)


robscott:
Thanks @skriss! This change LGTM but want @gnossen to confirm since it would represent a behavioral change.

/assign @gnossen 
/approve

gnossen (approved):
Thanks for the fix!

shaneutt (approved):
/lgtm

youngnick:
Updates #1016

Issue #1329, opened 2022-08-15
Document How to Report Conformance Test Results


robscott:
**What would you like to be added**:
A standard process for reporting conformance results, similar to what already exists for upstream Kubernetes https://github.com/cncf/k8s-conformance. After chatting with @hh about how this works in upstream, there are a few details that may need some tweaking for our use cases (please correct me if I misunderstood something).

* Upstream conformance is all or nothing, they don't have the same concept of conformance profiles
* Upstream conformance is per minor version, at least initially we may want to track conformance per patch version

As a general note, each upstream conformance report includes:

- A product.yaml file that describes the implementation, and version of the implementation that was used ([example](https://github.com/cncf/k8s-conformance/blob/master/v1.24/kops/PRODUCT.yaml))
- A readme that provides a step by step process describing how conformance tests were run to get the passing result ([example](https://github.com/cncf/k8s-conformance/blob/master/v1.24/kops/README.md))
- A test log that shows conformance tests passing ([example](https://github.com/cncf/k8s-conformance/blob/master/v1.24/kops/e2e.log))

In addition to the above, we'll likely want some ability to describe partial conformance, or at a minimum, to indicate the extended features that conformance tests pass for with the specified implementation.

This may eventually transition to CNCF control, but it likely makes sense to start this structure in this repo.

**Why this is needed**:
As we get an increasing number of implementations of this API, it is more important than ever to ensure that each provides a consistent experience. In the future we'll likely want to translate the reports here into an easy to read format on https://gateway-api.sigs.k8s.io/implementations/.

shaneutt:
#1709 is now covering the scope of this issue and adding reporting of conformance test results.

PR #1754, opened 2023-02-22
GEP 1709: initial implementation ideas

shaneutt:
**What type of PR is this?**

/kind gep

This contributes to #1709 and #1329 by adding the initial ideas for an implementation of conformance profiles to the relevant GEP.

Please note that the GEP remains in `provisional` state for this iteration so nothing that merges here is necessarily set in stone. Therefore if you have some lingering questions or concerns there will be more time to voice those concerns, and ideally even make your own PR making the changes that you want to see and helping co-author the GEP to pull in your ideas!

### DIFF
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -193,8 +193,9 @@ cSuite, err := suite.New(suite.Options{
     ),
     // other options
 })
+```
 
-> **NOTE**: The `UnsupportedFeatures` and `SupportedFeatures` fields are
+> **NOTE**: The `UnsupportedFeatures` and `SupportedFeatures` fields are 
 > mutually exclusive.
 
 Once an implementation has integrated with the conformance test suite, they can
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -213,49 +213,64 @@ repositories using our [certification process](#certification-process).
 
 #### Reporting Process
 
-When conformance tests complete a log message will be emitted:
+When conformance tests are executed an argument can be provided to the test
+suite to emit `ConformanceReport` resource with the test results. This resource
+can be configured to emit to the test output itself, or to a specific file.
 
-```console
-Success! Conformance profiles "Layer7" and "Layer4" passed for version v0.7.0.
-The following extended features were flagged as not supported for "Layer7Extended":
-- HTTPResponseHeaderModification
-- HTTPRouteQueryParamMatching
-```
+The following is an example report:
 
-For the above output upstream Gateway API can manually generate or update a report
-in the following format:
-
-```json
+```yaml
 apiVersion: v1alpha1
 kind: ConformanceReport
-implementation: acmeorg-acme
-reports:
+implementation:
+  name: acmeorg-acme
+  url: https://acmeorg.com
+  version: v1.0.0
+  contact:
+  - @acme-org/maintainers
+date: "2023-02-28 20:29:41+00:00"
+gatewayAPIVersion: v0.7.0
+profiles:
   layer4:
-    releases:
-    - tag: v0.7.0
-    - tag: v0.6.2
-    - tag: v0.6.1
+    core:
+      status: success
+      summary: "all core functionality passed"
+      stats:
+        passed: 4
+        skipped: 0
+        failed: 0
+    extended:
+      status: skipped
+      summary: "no extended features supported"
+      stats:
+        passed: 0
+        skipped: 6
+        failed: 0
+      unsupportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
   layer7:
-    releases:
-    - tag: v0.7.0
-      extended:
-        unsupportedFeatures:
-        - HTTPResponseHeaderModification
-        - HTTPRouteQueryParamMatching
-    - tag: v0.6.2
-      extended:
-        unsupportedFeatures:
-        - HTTPResponseHeaderModification
-        - HTTPRouteQueryParamMatching
-        - HTTPRouteMethodMatching
-    - tag: v0.6.1
-      extended:
-        unsupportedFeatures:
-        - HTTPResponseHeaderModification
-        - HTTPRouteQueryParamMatching
-        - HTTPRouteMethodMatching
-maintainers:
-- @kubernetes-sigs/gateway-api-maintainers
+    core:
+      status: success
+      summary: "all core functionality passed"
+      stats:
+        passed: 20
+        skipped: 0
+        failed: 0
+    extended:
+      status: success
+      summary: "all extended features supported"
+      stats:
+        passed: 8
+        skipped: 0
+        failed: 0
+      supportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
+      - ExtendedFeature4
+      - ExtendedFeature5
 ```
 
 > **NOTE**: In the above the `implementation` field is a combination of
@@ -263,41 +278,177 @@ maintainers:
 > an individual, a company, e.t.c.. Organizations can theoretically have more
 > than one `<project>` name and submit separate reports for each of them.
 
-> **NOTE**: The above demonstrates that the `acmeorg-acme` project was already
-> passing for versions prior to `v0.7.0`. This is how we will track history of
-> conformance for releases across implementations, but at some point might
-> require some archival process (which we probably don't have to worry about
-> for this iteration). You can also see that there's been progress to add the
-> `HTTPRouteMethodMatching` for the latest release.
-
-> **NOTE**: The `maintainers` field indicates the Github usernames or team
+> **NOTE**: The `contact` field indicates the Github usernames or team
 > names of those who are responsible for maintaining this file, so they can be
 > easily contacted when needed (e.g. for relevant release announcements
 > regarding conformance, e.t.c.).
 
-TODO: need to add testing flags to emit the `ConformanceReport` here automatically
+The above report describes an implemenation that just released `v1` and has
+`Core` support for Layer4 functionality and full supports both `Core` and
+`Extended` Layer7 functionality.
 
-The implementation can compile its `ConformanceReport` and upload it to the
-Gateway API project by creating a pull request. The file should be uploaded
-to the following sub-directory:
+`ConformanceReports` can be stored as a list of reports in chronological order.
+The following shows previous releases of the `acmeorg-acme` implementation and
+its feature progression:
+
+```yaml
+apiVersion: v1alpha1
+kind: ConformanceReport
+implementation:
+  name: acmeorg-acme
+  url: https://acmeorg.com
+  version: v0.91.0
+  contact:
+  - @acme-org/maintainers
+date: "2022-09-28 20:29:41+00:00"
+gatewayAPIVersion: v0.6.2
+profiles:
+  layer4:
+    core:
+      status: partial
+      summary: "some tests were manually skipped"
+      stats:
+        passed: 2
+        skipped: 2
+        failed: 0
+      skippedTests:
+      - TCPRouteBasics
+      - UDPRouteBasics
+    extended:
+      status: skipped
+      summary: "no extended features supported"
+      stats:
+        passed: 0
+        skipped: 4
+        failed: 0
+      unsupportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
+  layer7:
+    core:
+      status: success
+      summary: "all core functionality passed"
+      stats:
+        passed: 20
+        skipped: 0
+        failed: 0
+    extended:
+      status: success
+      summary: "all extended features supported"
+      stats:
+        passed: 5
+        skipped: 3
+        failed: 0
+      supportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
+      unsupportedFeatures:
+      - ExtendedFeature4
+      - ExtendedFeature5
+---
+apiVersion: v1alpha1
+kind: ConformanceReport
+implementation:
+  name: acmeorg-acme
+  url: https://acmeorg.com
+  version: v0.90.0
+  contact:
+  - @acme-org/maintainers
+date: "2022-08-28 20:29:41+00:00"
+gatewayAPIVersion: v0.6.1
+profiles:
+  layer4:
+    core:
+      status: failed
+      summary: "all tests are failing"
+      stats:
+        passed: 0
+        skipped: 0
+        failed: 4
+  layer7:
+    core:
+      status: success
+      summary: "all core functionality passed"
+      stats:
+        passed: 20
+        skipped: 0
+        failed: 0
+    extended:
+      status: skipped
+      summary: "no extended features supported"
+      stats:
+        passed: 2
+        skipped: 6
+        failed: 0
+      supportedFeatures:
+      - ExtendedFeature1
+      unsupportedFeatures:
+      - ExtendedFeature2
+      - ExtendedFeature3
+      - ExtendedFeature4
+      - ExtendedFeature5
+---
+apiVersion: v1alpha1
+kind: ConformanceReport
+implementation:
+  name: acmeorg-acme
+  url: https://acmeorg.com
+  version: v0.89.0
+  contact:
+  - @acme-org/maintainers
+date: "2022-07-28 20:29:41+00:00"
+gatewayAPIVersion: v0.6.0
+profiles:
+  layer7:
+    core:
+      status: partial
+      summary: "some tests were skipped"
+      stats:
+        passed: 16
+        skipped: 2
+        failed: 0
+      skippedTests:
+      - HTTPRouteTestExample1
+      - HTTPRouteTestExample2
+    extended:
+      status: skipped
+      summary: "no extended features supported"
+      stats:
+        passed: 0
+        skipped: 8
+        failed: 0
+      unsupportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
+      - ExtendedFeature4
+      - ExtendedFeature5
+```
+
+> **NOTE**: In the above you can see the `acme` implementation's progression. In
+> their release `v0.89.0` they had started adding `Layer7` support and added the
+> conformance tests to CI, but they were still skipping some core tests. In
+> their next release `v0.90.0` they completed adding `Layer7` `Core`
+> functionality (and even added one extended feature), and also starting adding
+> `Layer4` functionality during `v0.90.0` (but it was failing at that time). In
+> `v0.91.0` they had completed core `Layer7` supported and added two more
+> `Extended` features, and also started to get their `Layer4` functionality to
+> partially pass.
+
+Implementers can submit their reports upstream by creating a pull request to
+the Gateway API repository and adding new reports to a file specific to their
+implementation's name:
 
 ```console
 conformance/results/<organization>-<project>.yaml
 ```
 
-> **NOTE**: For this iteration we're focusing on getting these reports created
-> but at first they will probably be created by hand. In future iterations and
-> after having some time to experience the system, tooling to handle reporting
-> must be created.
-
-This will start the certification process which will result in the
-implementation receiving a badge they can use on their project pages and/or
-repositories which indicates their conformance levels and link to their
-conformance reports in the upstream Gateway API website.
+Creating a pull request to add the `ConformanceReport` will start the
+[certification process](#certification-process).
 
 TODO: need to add some linting and validation for `ConformanceReports`
-
-TODO: need to add a merge tool to merge old reports into new ones
 
 > **NOTE**: No verification process (to prevent intentionally incorrect
 > conformance results) will be implemented at this time. We expect that this wont
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -276,6 +276,8 @@ maintainers:
 > easily contacted when needed (e.g. for relevant release announcements
 > regarding conformance, e.t.c.).
 
+TODO: need to add testing flags to emit the `ConformanceReport` here automatically
+
 The implementation can compile its `ConformanceReport` and upload it to the
 Gateway API project by creating a pull request. The file should be uploaded
 to the following sub-directory:
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -7,7 +7,9 @@
 
 Add selectable profiles for conformance tests which implementations can
 subscribe to. Also add the ability to report conformance results back to the
-Gateway API project and receive recognition.
+Gateway API project and receive recognition. Conformance test reports will
+add conformance data to the implementations page for the reporter which can
+be linked to with badges from projects and repositories.
 
 ## Goals
 
@@ -24,6 +26,12 @@ Gateway API project and receive recognition.
 
 - We want to avoid adding any infrastructure for the reporting mechanism if
   feasible.
+- For this iteration we don't want to add configuration files for conformance
+  tests, instead leaving that to future iterations and working on the raw
+  machinery here (see [alternatives considered](#alternatives-considered)).
+- For this iteration we don't want to add container images for conformance test
+  runs, instead leaving that to future iterations (see
+  [alternatives considered](#alternatives-considered).
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,300 @@ which implementations can prove they satisfy and be recognized for.
...
+    -conformance-profile=Layer7Extended \
+    -unsupported-features=HTTPResponseHeaderModification,HTTPRouteMethodMatching,HTTPRouteQueryParamMatching,
+```
+
+Or the equivalent configuration using the Golang library:
+
+```go
+cSuite, err := suite.New(suite.Options{
+    GatewayClassName: "acme",
+    Profile: Layer7ExtendedProfile,
```
howardjohn, line 176: can I be L7 and L4 at the same time?

howardjohn: And mesh eventually

shaneutt: There's a note about this concept here: https://github.com/kubernetes-sigs/gateway-api/blob/c6b5417042119e6d3e1b023c9ffd92935706f2f3/site-src/geps/gep-1709.md?plain=1#L103

I'll consider this one resolved, but if you feel like this is insufficient and you want to see more change I would definitely invite you to mark it unresolved and let me know what changes you feel are needed.

howardjohn: I am not worried about the reporting but about the test running. Unless the tests for each profile are completely isolated so running all of them together vs running each profile in sequence is the ~same?

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
-TODO
+The API for conformance profiles will be a pipeline that implementations can
+opt into. The workflow is effectively:
+
+1. select a [profile](#profiles)
+2. [integrate](#integration) tests in the downstream project
+3. [report results and get certified](#certification)
+
+The goal is to make selecting a conformance profile as simple of a process as
+feasible and support both the existing command line integration approach as well
```
LiorLieberman, line 56: what is the existing command line approach? 
`go test  ......`?

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
+    -unsupported-features=HTTPResponseHeaderModification,HTTPRouteMethodMatching,HTTPRouteQueryParamMatching,
+```
+
+Or the equivalent configuration using the Golang library:
+
+```go
+cSuite, err := suite.New(suite.Options{
+    GatewayClassName: "acme",
+    Profile: Layer7ExtendedProfile,
+    UnsupportedFeatures: sets.New(
```
LiorLieberman, line 177: Feels like we can combine `UnsupportedFeatures` and `SkipTests` to one field and have additional logic that checks whether this combined field is used with `Core` profile to check validity.

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
+The conformance report will emit a heading that displays the latest release
+version which conformance has been reported for, along with a note about any
+newer releases which have yet to be reported.
+
+CI will provide [badges](https://shields.io) to contributors upon completion.
+
+> **NOTE**: Badges which cover `Extended` support (such as `Layer7Extended`)
+> will look the same regardless of whether the implementation only fulfills
+> the extended feature set partially. The badge links will direct to the full
+> report, which will include a list of any feature exceptions.
```
LiorLieberman, line 315: maybe a different colour for semi implemented Extended support?

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
+When conformance tests complete a log message will be emitted:
+
+```console
+Success! Conformance profile "Layer7Extended" passed for version v0.7.0
+The following extended features were not supported:
+- HTTPResponseHeaderModification
+- HTTPRouteQueryParamMatching
+```
+
+For the above output upstream Gateway API can be provided the following report:
```
LiorLieberman, line 227: One thing I have not found in the doc is who is responsible for generating `ConformanceReport` from the log?

LiorLieberman: gw-api go library? 
a shell script?

I am thinking it will probably be the first one but probably worth calling it out

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
+repositories which indicates their conformance levels and link to their
+conformance reports in the upstream Gateway API website.
+
+> **NOTE**: No verification process (to prevent intentionally incorrect
+> conformance results) will be implemented at this time. We expect that this wont
+> be an issue in our community and even if someone were to try and "cheat" on
+> the reporting the reputation loss for being caught would make them look very
+> bad and would not be worth it.
+
+#### Certification Process
```
LiorLieberman, line 293: for this iteration, should we consider any kind of visualisation of conformance reports and/or badges other than looking into `conformance/results/<organization>-<project>.yaml` files?

LiorLieberman: I think its better to keep it for future iterations but maybe worth adding a reference here

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,295 @@ which implementations can prove they satisfy and be recognized for.
...
+
+- `Layer4`
+- `Layer7`
+- `Mesh`
+
+> **NOTE**: these are simply the initial categories we're going to start with,
+> it's entirely reasonable for there to be more in the future.
+
+Support levels will be limited to `Core` and `Extended`, since
+`ImplementationSpecific` isn't meaningful in conformance at the time of
```
LiorLieberman, line 75: There are some impls that support experimental features, like istio and envoy-gw. 
Whats the additional cost of providing the profile for them to be able to include these tests?

site-src/geps/gep-1709.md
shaneutt: Yes that's correct, we can add that here:

```suggestion
feasible and support both the existing command line integration approach (e.g. `go test`) as well
```

site-src/geps/gep-1709.md
shaneutt: I'm open to doing something like this in the future if implementers complain about it... but I feel like extended is a completely reasonable place to have some features but not all.

site-src/geps/gep-1709.md
shaneutt: I was actually thinking for the first iteration of this, these could just be manually edited as they have straightforward contents and need to go through a review process anyway. Let's make that more explicit:

```suggestion
For the above output upstream Gateway API can manually generate or update a report
in the following format:
```

site-src/geps/gep-1709.md
shaneutt: I think `UnsupportedFeatures` and `SkipTests` need to be kept separate as `SkipTests` is particularly helpful _during_ the development process. E.g. someone could have partial support for a _feature_ but not all tests passing just yet and use `SkipTests` to disable the tests they're unable to fulfill just yet while they otherwise plan the work to fix it. We did literally this downstream at Kong and it was pretty convenient.

I'll consider this one resolved, but if you still feel there's more to do here please don't hesitate to unresolve and let me know if I've missed something or there's something else you feel needs changing.

site-src/geps/gep-1709.md
shaneutt: I think that would be good, however I'm not sure _what_ that should look like yet. I'm actually hoping that someone else will swoop in and think up a display layer using the raw data laid out so far.

site-src/geps/gep-1709.md
```diff
@@ -37,8 +45,296 @@ which implementations can prove they satisfy and be recognized for.
...
+
+for i := 0; i < len(tests.ConformanceTests); i++ {
+    test := tests.ConformanceTests[i]
+    test.Run(t, cSuite)
+}
+```
+
+> **NOTE**: In the `suite.Options` above it's still possible to add `SkipTests`
+> but when used in conjunction with `Profile` this will result in a report that
+> the profile is not valid for reporting. Implementations in this state may be
```
arkodg, line 156: why not add a section in the report highlighting what was `Skipped` ?

site-src/geps/gep-1709.md
shaneutt: In my first pass thinking through this I honestly didn't really think experimental features needed any special recognition, as they have a special nature that makes them off the normal course of the project's feature scope. That said, if there's some strong feelings we need to represent _experimental feature support_ visibly at a high level with conformance profiles I am open to that, I'm just not quite convinced yet :thinking:

site-src/geps/gep-1709.md
arkodg: what about adding the ability to input a list of supported Profiles ?

site-src/geps/gep-1709.md
shaneutt: My intention here was to _never_ report conformance when tests are explicitly skipped, instead making `SkipTests` a tool for the development phase. Can you help me to understand the value you see in allowing reporting where entire tests are skipped, even if they might have been required? In the case of extended where we know some _features_ may not be fully supported, wouldn't `UnsupportedFeatures` adequately cover things here?

arkodg: `SkipTests` are fewer, more fine grained tests, within `SupportedFeatures`, that the project cannot support either due to a bug in the project or due to API clarification. Here the setting in [Envoy Gateway](https://github.com/envoyproxy/gateway/blob/afe646b84d03cf4126f55721589850dc202dc130/test/conformance/conformance_test.go#L50). 

With above decision, Envoy Gateway would continue to run all tests, except some with SkipTests (temporarily until issues are fixed), but would not be able to generate any report

site-src/geps/gep-1709.md
arkodg: looks good, thanks !

site-src/geps/gep-1709.md
shaneutt: Right, I think we're kinda saying the same thing. But where I was kinda coming from, is that if you're _skipping tests_ then this basically means you're **not** conformant. I'm a little confused about having a reporting option for implementations which are not yet conformant, but I guess I could be persuaded? :thinking:

arkodg: Sure from a reporting standpoint, you can mark it as `not` conformant :)
From an implementation perspective, we prefer this approach to ensure more test coverage and minimal regression
TLDR, would prefer if the reporting tool marks the specific features (whose tests are skipped) as non conformant rather than fail reporting for other features too

geps/gep-1709.md
```diff
@@ -37,8 +45,304 @@ which implementations can prove they satisfy and be recognized for.
...
+
+```console
+Success! Conformance profiles "Layer7Extended" and "Layer4Core" passed for version v0.7.0
+The following extended features were flagged as not supported for "Layer7Extended":
+- HTTPResponseHeaderModification
+- HTTPRouteQueryParamMatching
+```
+
+For the above output upstream Gateway API can manually generate or update a report
+in the following format:
```
sunjayBhatia, line 231: worth versioning this schema? in case of future changes that might need a migration etc.?

site-src/geps/gep-1709.md
arkodg: The conformance tests run in CI in EG to prevent regression as well as report conformance (e.g. X feature tests pass). Earlier I was thinking we could this plug into the CI and run the tests once , so the report is generated on every PR run, so EG would replace this suite flag `EnableAllSupportedFeatures: true,` with `Profile: Layer4CoreProfile,Layer4ExtendedProfile,Layer7CoreProfile,Layer7ExtendedProfile`. 

But thinking more on this, the conformance profile tests might need to run as a separate task who's job is to only report conformance rather than use the exit code of the process to decide whether CI is passing or not (catch regression).

geps/gep-1709.md
shaneutt: Yes, I was just thinking about this today too!

Added in b783fcac4db87482cc0f9f43e19828e237d01de3 :+1:

site-src/geps/gep-1709.md
shaneutt: I added a TODO comment to the document that says we must figure this out _BEFORE_ we consider this `implementable`. That way we don't have to hold up this iteration just for that bit. :+1:

site-src/geps/gep-1709.md
shaneutt: After talking with @LiorLieberman in a video call, the entire concept of how you refer to profiles has been reworked to incorporate providing a positive (or negative) list of additional features so this should be set now.

site-src/geps/gep-1709.md
shaneutt: So it just occurred to me a couple days ago, I had already added a note about being able to explicitly report "in progress" conformance via this process:

https://github.com/kubernetes-sigs/gateway-api/pull/1754/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R130

Also I've removed the support levels on the profiles in favor of a nested reporting structure. If I'm understanding what you were asking for, I think this will provide the facility you were looking for. If not, please mark this unresolved and let me know!

geps/gep-1709.md
```diff
@@ -37,8 +45,303 @@ which implementations can prove they satisfy and be recognized for.
...
+of support are opt-in.
+
+[feat]:https://github.com/kubernetes-sigs/gateway-api/blob/c61097edaa3b1fad29721e787fee4b02c35e3103/conformance/utils/suite/suite.go#L33
+
+### Integration
+
+Integrating the test suite into your implementation can be done using one of
+the following methods:
+
+- The [go test][go-test] command line interface which enables projects of any
```
LiorLieberman, line 95: typo?

geps/gep-1709.md
```diff
@@ -37,8 +45,303 @@ which implementations can prove they satisfy and be recognized for.
...
+  language to run the test suite on any platform [Golang][go] supports.
+- Using the conformance test suite as a [Golang library][lib] within an already
```
LiorLieberman, line 97: `[Golang library][lib]` typo?

geps/gep-1709.md
```diff
@@ -37,8 +45,304 @@ which implementations can prove they satisfy and be recognized for.
...
+The implementation can compile its `ConformanceReport` and upload it to the
+Gateway API project by creating a pull request. The file should be uploaded
+to the following sub-directory:
+
+```console
+conformance/results/<organization>-<project>.yaml
+```
+
+> **NOTE**: For this iteration we're focusing on getting these reports created
+> but at first they will probably be created by hand. In future iterations and
```
sunjayBhatia, line 288: might want to add somewhere that we should have CI in this repo to do some YAML linting and/or schema validation of these uploaded reports

implementation detail but maybe good to point out as a requirement if we're starting to do this with "hand crafted" YAML

geps/gep-1709.md
```diff
@@ -37,8 +45,304 @@ which implementations can prove they satisfy and be recognized for.
...
+The following extended features were flagged as not supported for "Layer7Extended":
+- HTTPResponseHeaderModification
+- HTTPRouteQueryParamMatching
+```
+
+For the above output upstream Gateway API can manually generate or update a report
+in the following format:
+
+```json
+apiVersion: v1alpha1
```
sunjayBhatia, line 230: wdyt about each project/implementation's report file being a list of these sort of report objects? e.g.

```yaml
---
kind: ConformanceReport
apiVersion: v1alpha1
implementation:
  name: acmeorg-acme
  url: https://acmeorg.com
  version: v9.99.0
  contact:
  - @acme-org/maintainers
date: "2023-02-28 20:29:41+00:00"
gatewayAPIVersion: v0.6.1
profiles:
- name: layer4
  core:
    supportedFeatures:
    - ...
  extended:
    supportedFeatures:
    - ...
    unsupportedFeatures:
    - ...
- name: layer7
  core:
    supportedFeatures:
    - ...
  extended:
    supportedFeatures:
    - ...
    unsupportedFeatures:
    - ...
---
kind: ConformanceReport
apiVersion: v1alpha1
implementation:
  name: acmeorg-acme
  url: https://acmeorg.com
  version: v9.98.0
  contact:
  - @acme-org/maintainers
date: "2023-02-28 20:29:41+00:00"
gatewayAPIVersion: v0.6.0
profiles:
...
```

this would allow us to add the fields like date, version, etc. to each report as well since that detail is often relevant when assessing an implementation

geps/gep-1709.md
```diff
@@ -37,8 +45,304 @@ which implementations can prove they satisfy and be recognized for.
...
+> require some archival process (which we probably don't have to worry about
+> for this iteration). You can also see that there's been progress to add the
+> `HTTPRouteMethodMatching` for the latest release.
+
+> **NOTE**: The `maintainers` field indicates the Github usernames or team
+> names of those who are responsible for maintaining this file, so they can be
+> easily contacted when needed (e.g. for relevant release announcements
+> regarding conformance, e.t.c.).
+
+The implementation can compile its `ConformanceReport` and upload it to the
```
shaneutt, line 279:
```suggestion
TODO: need to add testing flags to emit the `ConformanceReport` here automatically

The implementation can compile its `ConformanceReport` and upload it to the
```

geps/gep-1709.md
shaneutt: I don't think so, this just provides a hyperlink which is handled further down. Please un-resolve and let me know if there's still a problem though? :thinking:

geps/gep-1709.md
shaneutt: Similar to above, this just adds a hyperlink which appears to be working? Please un-resolve and let me know if there's still a problem though? :thinking:

geps/gep-1709.md
shaneutt: sg, added a TODO for this in cfcc08351432dfe239ff18dd295ff357f55716ef, we will need to resolve this and other TODO's prior to going `implementable` :+1:

geps/gep-1709.md
```diff
@@ -37,8 +45,304 @@ which implementations can prove they satisfy and be recognized for.
...
+### Certification
+
+Implementations will be able to report their conformance testing results using
+our [reporting process](#reporting-process). Implementations will be able to
+visibly demonstrate their conformance results on their downstream projects and
+repositories using our [certification process](#certification-process).
+
+#### Reporting Process
+
+When conformance tests complete a log message will be emitted:
```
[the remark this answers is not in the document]
shaneutt: We may have been thinking this separately and simultaneously, because I just recently added a TODO for this:

75ec95f8e573dc6602020c3b8ecb99781bed75df

Let me know if that seems adequate :+1:

geps/gep-1709.md
shaneutt: I can see advantages to this, particularly for the purpose of the human reporting via a PR part, and also making it really straightforward to build an archive several years down the road when these reports get significantly bigger. :thinking: 

Only thing I would say is that you've combined `supportedFeatures` and `unsupportedFeatures` in the same report for `extended`, but there is a note that had attempted to establish these as mutually exclusive earlier. We'll have to discuss more on the subject of how we represent extended features (and maybe even implementation-specific ones) but that doesn't need to hold this particular comment up.

@LiorLieberman and @arkodg had also expressed interest in contributing to this doc, so I would like to hear from there.

sunjayBhatia: ah yeah feel free to disregard that part on the reporting of what is actually passed/failed etc. that was just a quick idea without considering what may have already been updated here

but good to hear some of that metadata additions and the idea that we could generate a report per-run of the suite sounds useful 👍🏽

geps/gep-1709.md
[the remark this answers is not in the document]
sunjayBhatia: yep that sounds great 👍🏽

geps/gep-1709.md
sunjayBhatia: I also considered suggesting something of the form of what sonobuoy or some other tool might output, but not sure if we want to go that far in reporting test result details etc., I don't remember exactly but I imagine using sonobuoy was discussed when starting the conformance framework

geps/gep-1709.md
```diff
@@ -37,8 +45,309 @@ which implementations can prove they satisfy and be recognized for.
...
+This will start the certification process which will result in the
+implementation receiving a badge they can use on their project pages and/or
+repositories which indicates their conformance levels and link to their
+conformance reports in the upstream Gateway API website.
+
+TODO: need to add some linting and validation for `ConformanceReports`
+
+TODO: need to add a merge tool to merge old reports into new ones
+
+> **NOTE**: No verification process (to prevent intentionally incorrect
```
mlavacca, line 302: I think we could at least check that if an implementor sets a specific profile, all the tests belonging to such a profile are successfully executed. So, for example, if not running specific tests belonging to a profile through the `SkipTests` field, we might want to invalidate the certification as some required tests have not been run. WDYT?

geps/gep-1709.md
shaneutt: I wouldn't be opposed to using that in time, if we found there were advantages. But at least for this very specific PR I'm not ready to completely switch things around yet. I would encourage you to make a PR after this one merges _just_ for pushing this change so that we can hyper-focus on it.

geps/gep-1709.md
shaneutt: > I think we could at least check that if an implementor sets a specific profile, all the tests belonging to such a profile are successfully executed.

Yes, but more specifically all the _core_ tests. We are already making _core_ a minimum requirement:

https://github.com/kubernetes-sigs/gateway-api/pull/1754/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R169

> So, for example, if not running specific tests belonging to a profile through the SkipTests field, we might want to invalidate the certification as some required tests have not been run. WDYT?

Yes. There's a note specifically about this present here:

https://github.com/kubernetes-sigs/gateway-api/pull/1754/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R127

We're not 100% sure exactly how it will _look_ yet, but we are already trying to be clear that `SkipTests` usage will invalidate the `ConformanceReport` (or at least mark conformance as "in-progress") or something like that.

mlavacca: Thanks for pointing out the notes mentioning this aspect, I somehow lost them :)

geps/gep-1709.md
arkodg: my high level comment here is, we should separate out `what tests to run` with `what to report`.
* under `what tests to run` , we should continue to support the existing knobs of `SupportedFeatures`, `UnsuportedFeatures`, `SkipTests` and `SupportAllFeatures`. The user can either choose to use these historic knobs **OR** use the new `ConformanceProfile` knob. These flags should be validated.
* under `what to report`, it could be as granular as which tests passed or more coarse as which features passed (if all tests within that feature passed), but this logic is not tied to the input knobs, and is tied to - which tests ran, which passed, which failed. And the report is an explicit add on data output enabled via some `--report` like flag

shaneutt: what you're suggesting seems to at least be close to where we're headed already @arkodg, do you feel there's anything specific in the doc that needs to be changed?

arkodg: yeah the doc says `Profiles` flag, in conjunction with `SkipTests` will result in an invalid report. Prefer if this tied to `what tests to run` instead, and said either use `Profile`  or use a mix of `SupportedFeatures, UnsuportedFeatures, SkipTests and SupportAllFeatures` to decide which tests to run. If not adhered, the tests wont start/or fail....

arkodg: ^ can be added later as well as a follow up

sunjayBhatia: regardless of the exact reporting format of test results, seems useful to generally summarize and agree on somewhere in this doc what is useful for gw api maintainers/implementers/users to see in the conformance test report

my two cents:
- the project/version/test run metadata above to identify a particular test result ^
- conformance profile test run results overall summary for each "profile"
  - implementation is deemed fully conformant with the profile if the tests all passed 
  - partial if some passed/some were skipped (IMO this is useful to know for esp. the extended profile)
  - no conformance if none passed or none were run (i.e. an L4 only controller might skip all L7 tests)
  - include summary stats, e.g. total number of tests, how many passed, skipped, failed
- conformance profile test run detailed results (in case of partial conformance etc.)
  - for each test, test name, short description like we have [here](https://github.com/kubernetes-sigs/gateway-api/blob/69fabd994640debc2d9c00d4a086cdee0bbe9a22/conformance/tests/gateway-invalid-route-kind.go#L35-L36)
  - did it pass, fail, or was it skipped
  - can include test filename for people to do closer inspection if needed
  - can add detailed failure messages, etc. if we want too

sunjayBhatia: (IMO seems ok to skip tests, that shouldn't necessarily make the output "invalid" but will mean you end up with a report that your implementation is not fully conformant)

sunjayBhatia: I'm on board with the notion of separating what to test from what to report conceptually as @arkodg mentions above, we could e.g. start with no changes to the test suite invocation and iteratively add convenience flags to let you run subsets of tests that correspond to "profiles"

shaneutt:
Marking this to not automatically merge, I will merge this manually once it's got the final LGTM.

geps/gep-1709.md
shaneutt: After thinking it over I was persuaded by the list of separate `ConformanceReports` idea, and did a reworking of the `ConformanceReport` API which includes that and somewhat follows the bullet points you listed above, @sunjayBhatia.

My hope is that this will be satisfactory for all stakeholders. I'm going to consider this comment thread resolved, as it's gotten quite long and hard to follow, but if any of you _don't_ find the new version satisfactory please create some new comments **ideally with explicitly and prescriptive suggestions about what specifically you would like to see changed** as that can be incredibly helpful to keeping up velocity on these GEPs. Also keep in mind that we're still in a `provisional` state so nothing that merges here is going to be implemented yet.

arkodg (approved):
this is a good start, prefer to get this in, and iteratively improve this doc, thanks !

shaneutt:
@sunjayBhatia I'm marking this one as a merge hold for the moment to make sure we hear back from you as a stakeholder before we merge it.

LiorLieberman:
I like the changes. Thanks!

/lgtm
/approve

sunjayBhatia (approved):
looks great to me 👍🏽 

I think regardless of how the implementation details etc. shake out (in terms of reporting schema, command line flags, etc.) I think the core ideas of what we want to get out of the centralized profiles/reporting are there!

shaneutt:
@youngnick as I understand it you wanted to give this a pass too, so keeping the hold on it until you've had a chance. Please feel free to remove the hold now once you're happy with it :+1:

youngnick (commented):
This LGTM, I think a couple of things I consider imperative are done:
- Using `SkipTests` on any Core features means you're not conformant
- The output is machine parsable.

The one thing I'd add, non-blocking though, is an example of what the YAML would look like with failed tests. Not that I think anyone would submit that, but it will make it clearer for implementors.

shaneutt:
> * Using `SkipTests` on any Core features means you're not conformant
> * The output is machine parsable.
> * The one thing I'd add, non-blocking though, is an example of what the YAML would look like with failed tests.

These things are already accounted for in the document:

https://github.com/kubernetes-sigs/gateway-api/pull/1754/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R127
https://github.com/kubernetes-sigs/gateway-api/pull/1754/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R369

> Oh, and it would be good to add that the descriptions are free-form, we shouldn't require specific values there.

This one wasn't accounted for yet, I've added this in 6489c36356345789354402103cfe9976a466690a :+1:

PR #1755, opened 2023-02-22
Removes field validation from GRPCRoute Method/Service fields

### LINKED ISSUE #1746: The validation for service and method fields in GRPCMethodMatch prohibits the usage of regular expressions.
**What happened**:
The current validation restricts users from using regular expressions in service/method fields.

https://github.com/kubernetes-sigs/gateway-api/blob/4aa03846c769cf4624930fa2d58b65cac33b88aa/apis/v1alpha2/grpcroute_types.go#L329
https://github.com/kubernetes-sigs/gateway-api/blob/4aa03846c769cf4624930fa2d58b65cac33b88aa/apis/v1alpha2/grpcroute_types.go#L342

User is able to use regex in service/method fields when GRPCMethodMatchType ==  RegularExpression.

```
kind: GRPCRoute
metadata:
  name: example
  labels:
    example: grpc-routing
spec:
  parentRefs:
    - name: example-gateway
  hostnames:
    - "grpc-example.com"
  rules:
    - matches:
      - method:
          method: Pi.*
          type: RegularExpression
      - method:
          service: example.*
          type: RegularExpression
      backendRefs:
        - group: ""
          kind: Service
          name: yages
          port: 9000
          weight: 1
```
```
The GRPCRoute "example" is invalid:
* spec.rules[0].matches[0].method.method: Invalid value: "Pi.*": spec.rules[0].matches[0].method.method in body should match '^[A-Za-z_][A-Za-z_0-9]*$'
* spec.rules[0].matches[1].method.service: Invalid value: "example.*": spec.rules[0].matches[1].method.service in body should match '^(?i)\.?[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*$'
* ```

### PR DESCRIPTION
**Pull request by sbko**
**What type of PR is this?**
/kind bug

Removes validation for service and method fields in GRPCMethodMatch to allow use of regular expressions.

Fixes #1746

### DIFF
--- a/apis/v1alpha2/validation/grpcroute.go
+++ b/apis/v1alpha2/validation/grpcroute.go
@@ -17,7 +17,9 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -31,6 +33,8 @@ var (
 	repeatableGRPCRouteFilters = []gatewayv1a2.GRPCRouteFilterType{
 		gatewayv1a2.GRPCRouteFilterExtensionRef,
 	}
+	validServiceName = `^(?i)\.?[a-z_][a-z_0-9]*(\.[a-z_][a-z_0-9]*)*$`
+	validMethodName  = `^[A-Za-z_][A-Za-z_0-9]*$`
 )
 
 // ValidateGRPCRoute validates GRPCRoute according to the Gateway API specification.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by gnossen** [comment]
CC @robscott since he advocated for and implemented this validation.

I think this is the only valid way to go here. What was missed with this validation was that we need to accept regexes, but there are two problems here:

- [The language consisting of all regular expressions is not itself regular](https://stackoverflow.com/questions/19533804/is-there-a-regular-language-to-represent-regular-expressions) and therefore cannot be captured by a standard regular expression
- According to the spec, the particular regex engine used is implementation-defined. Therefore, even if we could construct a regular expression to capture the valid regexes, we would not be able to agree on a single implementation's flavor to use in validation.

**Review by gnossen — APPROVED** [review]

**Comment by robscott** [comment]
Hey @sbko, I agree with @gnossen that this change is unfortunately likely required, but I'd like to make sure we still try to capture this validation somewhere. Can you mirror what we're doing for HTTPRoute here?

https://github.com/kubernetes-sigs/gateway-api/blob/ac16d2358125458a8f0ffba4020440d45944b0ed/apis/v1beta1/validation/httproute.go#L157-L189

Of course we'd still need to use the GRPCRoute-specific regexes we're using here, and add the validation to https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1alpha2/validation/grpcroute.go, but I think the general logic and structure would be very similar.

**Comment by sbko** [comment]
@robscott I'll take a look

**Comment by sbko** [comment]
@robscott PTAL

File: apis/v1alpha2/validation/grpcroute.go [hunk 1]
```diff
@@ -63,13 +67,38 @@ func validateGRPCRouteRules(rules []gatewayv1a2.GRPCRouteRule, path *field.Path)
...
 func validateRuleMatches(matches []gatewayv1a2.GRPCRouteMatch, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
 	for i, m := range matches {
-		if m.Method != nil && m.Method.Service == nil && m.Method.Method == nil {
-			errs = append(errs, field.Required(path.Index(i).Child("method"), "one or both of `service` or `method` must be specified"))
+		if m.Method != nil {
+			if m.Method.Service == nil && m.Method.Method == nil {
+				errs = append(errs, field.Required(path.Index(i).Child("method"), "one or both of `service` or `method` must be specified"))
+			}
+			// GRPCRoute method matchers admits two types: Exact and RegularExpression.
```
gyohuangxin, line 78:
```suggestion
			// GRPCRoute method matcher admits two types: Exact and RegularExpression.
```

youngnick (commented):
/hold for multiple reviews

/approve
though, this LGTM.

robscott:
Thanks @sbko!

/lgtm
/hold cancel

youngnick:
Updates #1016

PR #1760, opened 2023-02-27
conformance: use Patch to make ObservedGeneration tests robust against conflicts

michaelbeaumont:
**What type of PR is this?**
/kind feature
/area conformance

Using `Update` can lead to Conflict errors that fail the test:

```
the object has been modified; please apply your changes to the latest version and try again
```

Another option would be to use something like https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict but just using `Patch` seems a simpler option for this case.

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-observed-generation-bump.go
+++ b/conformance/tests/gateway-observed-generation-bump.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
@@ -69,8 +70,8 @@ var GatewayObservedGenerationBump = suite.ConformanceTest{
 				},
 			})
 
-			err = s.Client.Update(ctx, mutate)
-			require.NoErrorf(t, err, "error updating the Gateway: %v", err)
+			err = s.Client.Patch(ctx, mutate, client.MergeFrom(original))
+			require.NoErrorf(t, err, "error patching the Gateway: %v", err)
 
 			// Ensure the generation and observedGeneration sync up
 			kubernetes.NamespacesMustBeReady(t, s.Client, s.TimeoutConfig, namespaces)
--- a/conformance/tests/gatewayclass-observed-generation-bump.go
+++ b/conformance/tests/gatewayclass-observed-generation-bump.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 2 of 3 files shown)


arkodg:
we in Envoy Gateway might have this same issue, our tests are flaking as well https://github.com/envoyproxy/gateway/issues/1016. Will test this PR out and report back

sunjayBhatia:
would have to run this in CI for a little while, but Contour generally runs into this type of issue as well

arkodg:
didnt seem to help in EG, here's the draft PR I tested with https://github.com/envoyproxy/gateway/pull/1106

michaelbeaumont:
> didnt seem to help in EG, here's the draft PR I tested with [envoyproxy/gateway#1106](https://github.com/envoyproxy/gateway/pull/1106)

I may be missing it in the output but it looks like the `*ObservedGenerationBump` tests are passing in the checks.

mlavacca (approved):
The Kong Ingress controller did not experience this flake, but the change looks legit and did not trigger any error in our CI.

arkodg:
> > didnt seem to help in EG, here's the draft PR I tested with [envoyproxy/gateway#1106](https://github.com/envoyproxy/gateway/pull/1106)
> 
> I may be missing it in the output but it looks like the `*ObservedGenerationBump` tests are passing in the checks.

@michaelbeaumont you're right, the `ObservedGenerationBump` tests pass, but some newer changes that were pulled in with your fork cause another unrelated test to fail, my bad

arkodg (approved):

dprotaso:
Changes looks good - but a question since I'm not familiar so much with the `controller-runtime` client.

When calling `Patch` if the supplied resource includes a `resourceVersion` in the `metadata` is that included in the patch request?

It's not really documented but if the `resourceVersion` is present in a PATCH it will return a 409 conflict if it doesn't match the `resourceVersion` on the server - https://github.com/knative/serving/issues/1992

michaelbeaumont:
> Changes looks good - but a question since I'm not familiar so much with the `controller-runtime` client.
> 
> When calling `Patch` if the supplied resource includes a `resourceVersion` in the `metadata` is that included in the patch request?
> 
> It's not really documented but if the `resourceVersion` is present in a PATCH it will return a 409 conflict if it doesn't match the `resourceVersion` on the server - [knative/serving#1992](https://github.com/knative/serving/issues/1992)

Looking at `controller-runtime`, the [`resourceVersion` isn't included by default but can be](https://github.com/kubernetes-sigs/controller-runtime/blob/c304e7ec2ee7879f53ce69c12a7aa133853fccdb/pkg/client/patch.go#L104) by passing in a [`client.MergeFromWithOptimisticLock{}`](https://github.com/kubernetes-sigs/controller-runtime/blob/c304e7ec2ee7879f53ce69c12a7aa133853fccdb/pkg/client/patch.go#L57-L66) to [`client.MergeFromWithOptions`](https://github.com/kubernetes-sigs/controller-runtime/blob/c304e7ec2ee7879f53ce69c12a7aa133853fccdb/pkg/client/patch.go#L156) instead of `client.MergeFrom` as called in the PR.

robscott:
Thanks @michaelbeaumont!

/lgtm
/approve

PR #1768, opened 2023-03-03
GEP 1709: small cleanups and clarifications

shaneutt:
**What type of PR is this?**

/kind cleanup
/kind gep

This adds some cleanup and clarifications for GEP 1709.

It **does not add or change concepts or context**, just an attempt to improve how it reads.

In particular, we had mentioned `implementation-specific` conformance in the doc previously, but that was mostly because we wanted to point out we consider that a future possibility. This clarifies that by adding it to the alternatives considered section as a note for follow-up iterations.

Relates to #1709

### DIFF
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -50,16 +50,21 @@ submitting those reports and displaying the results.
 
 ## API
 
-The API for conformance profiles will be a pipeline that implementations can
-opt into. The workflow is effectively:
+The API for conformance profiles will include an API resource called
+`ConformanceReport` which will be at the center of a workflow that
+implementations can opt into to generate and submit those resources.
+
+The workflow implementers will follow will include the following high-level
+steps:
 
 1. select a [profile](#profiles)
 2. [integrate](#integration) tests in the downstream project
 3. [report results and get certified](#certification)
 
-The goal is to make selecting a conformance profile as simple of a process as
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):

LiorLieberman:
Thanks @shaneutt 
One thing I am thinking about now is `experimental`. I think it is important to have reports for these features as still have some really basic feature in this channel. I can think of path redirect, host rewrite etc.

If I am a user who wants to pick up a mesh OR switch off from implementation specific CRDs to core gateways I would like to know whether the gw-api impl supports these features.

We do not necessarily should address this concern now but it maybe worth a bullet under the TODOs

shaneutt:
> Thanks @shaneutt One thing I am thinking about now is `experimental`. I think it is important to have reports for these features as still have some really basic features in this channel. I can think of path redirect, host rewrite etc.
> 
> If I am a user who wants to pick up a mesh OR switch off from implementation specific CRDs to core gateways I would like to know whether the gw-api impl supports these features.
> 
> We do not necessarily should address this concern now but it maybe worth a bullet under the TODOs

Sounds good. I've made a formal pre-`implementable` TODO list at the top of the document and added this in 6de78d57503af2b86b32c124c8bd794e285944ce :+1:

LiorLieberman (approved):
Lgtm

PR #1781, opened 2023-03-07
`GatewayClassObservedGenerationBump` as a core feature

### LINKED ISSUE #1780: `GatewayClassObservedGenerationBump` feature type
**What happened**:

`GatewayClassObservedGenerationBump` is marked as an additional feature, but it Is marked neither as extended conformance nor as experimental conformance: https://github.com/kubernetes-sigs/gateway-api/blob/f6d4eb7a31506be144d5288526eb5f3cceb1bcf1/conformance/utils/suite/suite.go#L56-L57
My guess is that it is extended conformance, is it correct? If so, we should add `(extended conformance)` at the end of the comment line as all the other features above and below.

mlavacca:
**What type of PR is this?**

The additional feature `GatewayClassObservedGenerationBump` has been removed, and the test is now executed as a core one.

Fixes #1780 

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gatewayclass-observed-generation-bump.go
+++ b/conformance/tests/gatewayclass-observed-generation-bump.go
@@ -38,7 +38,6 @@ var GatewayClassObservedGenerationBump = suite.ConformanceTest{
 	ShortName: "GatewayClassObservedGenerationBump",
 	Features: []suite.SupportedFeature{
 		suite.SupportGateway,
-		suite.SupportGatewayClassObservedGenerationBump,
 	},
 	Description: "A GatewayClass should update the observedGeneration in all of it's Status.Conditions after an update to the spec",
 	Manifests:   []string{"tests/gatewayclass-observed-generation-bump.yaml"},
--- a/conformance/utils/suite/features.go
+++ b/conformance/utils/suite/features.go
@@ -55,15 +55,6 @@ var StandardCoreFeatures = sets.New(
 // -----------------------------------------------------------------------------
 
 const (
-	// This option indicates GatewayClass will update the observedGeneration in
-	// it's conditions when reconciling (extended conformance).
-	//
-	// NOTE: we intend to make this core and require implementations to do it
-	//       as we expect this is something every implementation should be able
-	//       to do and it's ideal behavior.
-	//
-	//       See: https://github.com/kubernetes-sigs/gateway-api/issues/1780
-	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
 	// This option indicates that the Gateway can also use port 8080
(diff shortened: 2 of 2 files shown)


shaneutt (approved):
Indeed, thank you @mlavacca!

This matches the original intention as I understood it, _however_ let's make sure we let a few people take a look before we merge in case there were different takes (we wouldn't want to break someone).

LiorLieberman (approved):
Thanks @mlavacca ! 
This LGTM, but we do not have any way to check who does not conform with it yet. 
So perhaps we should ask in Slack if someone is not conforming with this?

arkodg (approved):

conformance/utils/suite/suite.go
```diff
@@ -53,9 +53,6 @@ const (
 	// This option indicates support for Destination Port matching (extended conformance).
 	SupportRouteDestinationPortMatching SupportedFeature = "RouteDestinationPortMatching"
 
-	// This option indicates GatewayClass will update the observedGeneration in it's conditions when reconciling
-	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
```
dprotaso, line 57: Maybe mark this `Deprecated` for a single release and then drop it in a subsequent one

shaneutt:
We've received some feedback in Slack that at least 1 implementation would like some time to make some adjustments for this. I think we should probably put this on the agenda for next weeks meeting so we can check in with others too.

robscott:
We've talked about this in some meetings without actually leaving an update on the PR. I think we're in agreement that this should _eventually_ happen, but this test makes a relatively large change, it requires implementations to support GatewayClasses with arbitrary names. I believe there are at least some in-cluster implementations of the API that are deployed with a hardcoded GatewayClass name. 

The intent of the API is for every implementation to be able to support GatewayClasses based on the controller name, not just the name of the class. We have not clearly specified that anywhere yet though. We likely need to first clarify that that is a requirement (ideally by v0.7.0) and then make this specific change in a follow up release.

robscott:
Filed https://github.com/kubernetes-sigs/gateway-api/issues/1822 as a blocker for this PR.

shaneutt:
As it stands we're blocked on https://github.com/kubernetes-sigs/gateway-api/issues/1822 for this PR, so I'm going to mark it as `draft` for now just to indicate that it doesn't need active review (and we can mark it ready again whenever we're ready).

shaneutt:
#2228 added some language in support of this behavior to indicate that implementations MUST manage the status for any `GatewayClass` which carries their `controllerName`.

mlavacca:
I've revamped this one and put it in Ready for review as it has been unblocked.

/cc @shaneutt

robscott:
Thanks @mlavacca!

/lgtm

PR #1783, opened 2023-03-07
docs: GEP 1709 profiles at the *Route granularity level

shaneutt:
**What type of PR is this?**

/kind gep
/kind documentation

There's been a lot of feedback from the community in community meetings and on recent PRs that there's a preference for named profiles to correspond directly with `*Route` APIs, and move down a level in granularity from `Layer 4` and `Layer 7`. This PR makes the language changes necessary to support that desire.

### DIFF
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -508,7 +508,7 @@ Creating a pull request to add the `ConformanceReport` will start the
 > the reporting the reputation loss for being caught would make them look very
 > bad and would not be worth it.
 
-[crds]:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
+[crd]:https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
 
 #### Certification Process
 
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -29,11 +29,6 @@ The following are items that we intend to resolve before we consider this GEP
   this soak for some time, bring up the topic at SIG Arch community meetings, and
   generally make changes according to their feedback before considering this
   `implementable`.
-- There are lingering concerns from several community members that having high
-  level profiles like `Layer7` and `Layer4` might not be the best way to start,
-  but instead we should consider focusing on specific APIs, e.g. profiles that
-  we start with might be `HTTPRoute`, `GRPCRoute`, `TCPRoute`, `UDPRoute`
-  e.t.c. so that subscribing to profiles can be more à la carte.
 
 [sig-arch-feedback]:https://groups.google.com/g/kubernetes-sig-architecture/c/YjrVZ4NJQiA/m/7Qg7ScddBwAJ
 
@@ -97,24 +92,36 @@ conformance suite as a library.
 
 ### Profiles
 
-Initially there will be two profiles:
+"Profiles" are effectively categories which represent the high level grouping of
+tests related to some feature (or feature set) of Gateway API. When conformance
+is reported using one of these profiles extra features can be covered according
+to support levels:
 
-- `Layer4`
-- `Layer7`
-
-> **NOTE**: these are simply the initial profiles we're going to start with,
-> it's plausible for there to be more in the future.
-
-These named profiles are effectively categories which represent the high level
-grouping of tests. When conformance is reported using one of these profiles
-extra features can be covered according to support levels:
-
+- `core`
 - `extended`
 
 > **NOTE**: `implementation-specific` doesn't really have much in the way of
 > tests today, but it is something users want to be able to display. We leave
 > door open for it later and mention it in the [alternatives
 > considered](#alternatives-considered) section below.
+
+We will start with the following named profiles:
+
+- `HTTPRoute`
+- `TLSRoute`
+
+These profiles correspond with `*Route` type APIs that we currently have tests
+for. As the tests roll in, we'll also eventually have:
+
+- `UDPRoute`
+- `TCPRoute`
+- `GRPCRoute`
+
+> **NOTE**: In time we may have higher level groupings, like `Layer4` (which
+> would include at least `TCPRoute` and `UDPRoute`) but feedback from the
+> community has been strong for a preference on the `*Route` level (see the
+> [alternatives considered](#alternatives-considered) for some more notes on
+> this) for the moment.
 
 The technical implementation of these profiles is very simple: effectively a
 "profile" is a static compilation of existing [SupportedFeatures][feat] which
@@ -563,6 +570,22 @@ we believe the test suite for mesh could end up being it's own separate thing
 so it's not a part of this GEP for now, but we should make sure our tooling is
 modular and composable so if we do end up with an eventual mesh conformance
 test suite, it can employ conformance profiles and reporting as-is.
+
+### High Level Profiles
+
+We originally started with two high level profiles:
+
+- `Layer4`
+- `Layer7`
+
+However the overwhelming feedback from the community was to go a step down and
+define profiles at the level of each individual API (e.g. `HTTPRoute`,
+`TCPRoute`, `GRPCRoute`, e.t.c.). One of the main reasons for this was that we
+already have multiple known implementations of Gateway API which only support
+a single route type (`UDPRoute`, in particular as it turns out).
+
+We may consider in the future doing some of these higher level profiles if
+there's a technical reason or strong desire from implementers.
 
 ## References
 
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -180,7 +180,7 @@ above):
 ```console
 $ go test ./conformance/... -args \
     -gateway-class=acme \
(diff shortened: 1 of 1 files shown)


youngnick:
/hold for multiple reviews

but
/lgtm

geps/gep-1709.md
```diff
@@ -97,25 +92,37 @@ conformance suite as a library.
...
 > tests today, but it is something users want to be able to display. We leave
 > door open for it later and mention it in the [alternatives
 > considered](#alternatives-considered) section below.
 
+We will start with the following named profiles:
+
+- `HTTPRoute`
+- `TLSRoute`
+
+These profiles correspond with `*Route` type APIs that we currently have tests
```
sunjayBhatia, line 113: where would the existing tests corresponding to Gateway/GatewayClass validation/features fall here? would they also automatically be run?

sunjayBhatia (commented):
ultimately for Contour I don't think we're too picky about what the different conformance profiles are (so I don't hold too strong of an opinion on what we go with), but some thoughts/ideas that might spark discussion and some since people are enthusiastic about breaking up the profiles more granularly:

For implementers, having a little more granularity makes sense
  - Allows you to possibly pick and choose functionality to implement iteratively while not having to monitor every test that is added
  - Easy to see in a report what is passing/failing/skipped etc.

If we intend to use the conformance reports to present Gateway API implementations (and the API itself) to users, might it make more sense to break up the higher level (e.g. L7) groupings of tests along functionality lines rather than by resource?
- at least at first this might be more of a naming distinction tbh, since I think the intent of the per-*Route profiles is pretty similar
- e.g. subset of L7 tests that correspond to HTTP (basically HTTPRoute tests, maybe some TLSRoute tests for HTTP over TLS passthrough)
- e.g. subset of L7 tests that correspond to gRPC (basically GRPCRoute tests)
- as a user I might not know/care which resource I need to use for what purpose, especially when coming from Ingress v1, so evaluating implementation's conformance with that as the first "filter" is less useful?
- as someone who is totally new to Gateway API, if I see the available features expressed as behavior/functionality that might be more compelling to migrate to Gateway API?

I guess the above ^ is related to the idea of nailing down what the target audience for the conformance report is, if it is not just Gateway API maintainers/implementers, we might want to target how we present things (or decide that work is done in the website UX presentation of conformance reports, rather than the raw report format)

shaneutt:
> 

So it sounds like you're advocating for the possibility that we might _want_ the higher level ones eventually, which is a thought I share. Currently @sunjayBhatia I've placed this in the `alternatives considered` which is like the `TODO` area in that it's stuff to follow up on later, but unlike `TODO` this change would _not_ consider it a blocker for this KEP to move to `implementable`. As I see it, there's nothing stopping us from adding higher levels of composition at a later time given what we have so far, does that ring true with you as well? Basically: are there any specific actions you want to see taken here and now, as a part of this PR before you would be willing to approve this PR?

geps/gep-1709.md
shaneutt: That's a good point. Implicitly we understand they would run under the route named profiles, but we didn't actually explicitly spell that out yet. f9d04b965ece6074d52ae3988f8fb4cc704c1e53 takes care of adding a note about this :+1:

sunjayBhatia:
> > 
> 
> So it sounds like you're advocating for the possibility that we might _want_ the higher level ones eventually, which is a thought I share. Currently @sunjayBhatia I've placed this in the `alternatives considered` which is like the `TODO` area in that it's stuff to follow up on later, but unlike `TODO` this change would _not_ consider it a blocker for this KEP to move to `implementable`. As I see it, there's nothing stopping us from adding higher levels of composition at a later time given what we have so far, does that ring true with you as well? Basically: are there any specific actions you want to see taken here and now, as a part of this PR before you would be willing to approve this PR?

I think that TODO section is clear about wanting to leave open the possibility for "higher level" profiles 👍🏽, but my comment is more about seeing what people think about reframing how the granular breakdown of profiles is presented, i.e. from resource kind-centric to functional area/feature-centric (e.g. "HTTP" rather than "HTTPRoute") since new users of GW API might not know/care about what resources they need to use, just what functionality they want to mirror from Ingress (or some implementation-specific CR) for example

shaneutt:
> > > 
> > 
> > 
> > So it sounds like you're advocating for the possibility that we might _want_ the higher level ones eventually, which is a thought I share. Currently @sunjayBhatia I've placed this in the `alternatives considered` which is like the `TODO` area in that it's stuff to follow up on later, but unlike `TODO` this change would _not_ consider it a blocker for this KEP to move to `implementable`. As I see it, there's nothing stopping us from adding higher levels of composition at a later time given what we have so far, does that ring true with you as well? Basically: are there any specific actions you want to see taken here and now, as a part of this PR before you would be willing to approve this PR?
> 
> I think that TODO section is clear about wanting to leave open the possibility for "higher level" profiles 👍🏽, but my comment is more about seeing what people think about reframing how the granular breakdown of profiles is presented, i.e. from resource kind-centric to functional area/feature-centric (e.g. "HTTP" rather than "HTTPRoute") since new users of GW API might not know/care about what resources they need to use, just what functionality they want to mirror from Ingress (or some implementation-specific CR) for example

Ah I was misunderstanding before. Now I get what you're saying. That sounds good to me actually, let's try that out. 0526ef85d59aae73e8eb5e23f3d5dc89a7e2e1cb makes the change :+1:

sunjayBhatia (approved):
I'm on board with this change, interested in what others think!

geps/gep-1709.md
```diff
@@ -97,25 +92,42 @@ conformance suite as a library.
...
+
+- `HTTP`
+- `TLS`
+
+These profiles correspond with `*Route` type APIs that we currently have tests
+for. As the tests roll in, we'll also eventually have:
+
+- `UDP`
+- `TCP`
+- `GRP`
```
arkodg, line 118:
```suggestion
- `GRPC`
```

geps/gep-1709.md
```diff
@@ -97,25 +92,42 @@ conformance suite as a library.
...
 
 > **NOTE**: `implementation-specific` doesn't really have much in the way of
 > tests today, but it is something users want to be able to display. We leave
 > door open for it later and mention it in the [alternatives
 > considered](#alternatives-considered) section below.
 
+We will start with the following named profiles:
+
+- `HTTP`
+- `TLS`
```
arkodg, line 111: does this cover passthrough or termination ?

robscott: This is a good question, it's probably safest to just name these profiles `HTTPRoute`, `TLSRoute`, etc, to avoid any confusion with overall protocol support.

robscott (commented):
Thanks @shaneutt! This mostly LGTM.

geps/gep-1709.md
shaneutt: @sunjayBhatia any thoughts here?

shaneutt:
Adding a hold on merge so that I can do some squashing and a rebase prior to merge.

geps/gep-1709.md
sunjayBhatia: thinking about the `TLS` category name a little, it is a bit broad, you can do HTTP over TLS, gRPC over TLS, some other TCP protocol etc., and then you get the terminate/passthrough configuration option

I think as-written the `TLS` category is intended to cover [this existing test](https://github.com/kubernetes-sigs/gateway-api/blob/40a86dcad8f2dd307435b82317483248d484f853/conformance/tests/tlsroute-simple-same-namespace.go) and any other related additional tests. Given our stance on listener protocol/route type (https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls), TLSRoute is only really used for passthrough TLS, so conceptually I think we can express that.

IMO we could rename `TLS` here to `TLSPassthrough`

shaneutt: Let's go for that, again we're still in provisional/prototyping so we've got lots of runway to try this out and see how it feels.

```suggestion
- `TLSPassthrough`
```

Please anyone unresolve if you feel that there's more action to take here :+1:

shaneutt: `ReferenceGrant` is actually one that's been troubling me. We have a note about how `GatewayClass` and `Gateway` will be automatically included as they are referent in some featuresets, see here:

https://github.com/kubernetes-sigs/gateway-api/pull/1783/files#diff-7764ccb47f68a80348b14a2c6df6c6cadd80db02de7817dbcb8039a611bf1328R126

However I only put `GatewayClass` and `Gateway` there because those are easy and obvious, `ReferenceGrant` _might_ fit there but it's a little more subtle.

Since we'll need to make sure we account for this before we go GA, but I would prefer not to get this PR hung up for it I've added an explicit TODO item saying that we'll need to resolve this before `implementable`:

https://github.com/kubernetes-sigs/gateway-api/pull/1783/commits/d7bd41733e5f6ce40bbd45cb7f09ddac0863a71c

youngnick (commented):
I think this LGTM for provisional, given we've got a bunch of TODOs still.

/lgtm
/unhold

Issue #1779, opened 2023-03-07
HTTP Redirect Example is Invalid


robscott:
**What happened**:
https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/#redirects example includes a backendRef.

**What you expected to happen**:
* [ ] The included example should not include a backendRef
* [ ] Webhook validation should prevent an HTTPRoute that specifies a redirect filter from specifying a backendRef (until/unless we resolve https://github.com/kubernetes-sigs/gateway-api/issues/1185)

cc @rainest

Issue #1185, opened 2022-06-06
Add redirect loop protection to HTTPRequestRedirect


rainest:
**What would you like to be added**:
I want the HTTPRequestRedirect to include language indicating how implementations handle redirects that result in loops, where the request URL received is equal to the request URL returned by the filter.

I propose that the spec be amended to include something like
> If the result of applying an HTTPRequestRedirect would be a response whose `Location` header value matches the original request URL, implementations MUST NOT return the redirect request, and MUST instead proxy the request upstream.

Is this behavior desirable, and can implementations honor it easily?

**Why this is needed**:
https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRequestRedirectFilter defines an HTTPRoute Filter for issuing HTTP redirects in response that match a set of criteria. My strictest interpretation of the spec is that implementations must honor these absolutely: there are no matching criteria other than the HTTPRoute itself, and if a request matches that HTTPRoute, the implementation must issue the redirect.

Because the redirect filter applies to any matching request, you can easily configure HTTPRoutes that will result in a redirect loop: clients will send a request with some URL, receive a redirect response directing them to the same URL, send a new request to that URL, and receive another redirect to the same URL ad infinitum. For example, this HTTPRoute performs an HTTP->HTTPS redirect for any request to the `redirect.example` domain:

```
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: http-filter-redirect
spec:
  hostnames:
    - redirect.example
  rules:
    - filters:
      - type: RequestRedirect
        requestRedirect:
          scheme: https
          statusCode: 301
```
With the naive implementation above, attaching this HTTPRoute to a Gateway with both HTTP and HTTPS Listeners would result in a loop: `http://example.redirect/` will be redirected to `https://example.redirect` and `https://example.redirect` will also be redirected, back to itself, indefinitely.

AFAIK to avoid this you'd need to create separate HTTPRoutes, one with the redirect and one without, and either bind them to separate Gateways or use a combination of allowedRoutes configuration and namespaced backendRefs to bind them to only one Listener on a single Gateway. I may have overlooked something here, but I don't see another way to ensure they don't collide. I am not aware of any reason users would _want_ to configure a redirect loop, since HTTP-compliant clients will never receive a useful response from a looping URL, so I believe the strict interpretation of the current spec only needlessly complicates configuration.

[...]

PR #1788, opened 2023-03-08
fix: add webhook validation of redirect usage

Xunzhuo:
**What type of PR is this?**

/kind bug

Webhook validation should prevent a HTTPRoute that specifies a redirect filter from specifying a backendRef

Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/1779

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/validation/httproute_test.go
+++ b/apis/v1alpha2/validation/httproute_test.go
@@ -805,6 +805,7 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 	tests := []struct {
 		name        string
 		routeFilter gatewayv1a2.HTTPRouteFilter
+		backendRefs []gatewayv1a2.HTTPBackendRef
 		errCount    int
 	}{{
 		name: "valid HTTPRouteFilterRequestHeaderModifier route filter",
@@ -816,6 +817,14 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 				Remove: []string{"remove"},
 			},
 		},
+		backendRefs: []gatewayv1a2.HTTPBackendRef{{
+			BackendRef: gatewayv1a2.BackendRef{
+				BackendObjectReference: gatewayv1a2.BackendObjectReference{
+					Name: gatewayv1a2.ObjectName("test"),
+					Port: ptrTo(gatewayv1b1.PortNumber(8080)),
+				},
+			},
+		}},
 		errCount: 0,
 	}, {
 		name: "invalid HTTPRouteFilterRequestHeaderModifier type filter with non-matching field",
@@ -823,12 +832,28 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 			Type:          gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
 			RequestMirror: &gatewayv1a2.HTTPRequestMirrorFilter{},
 		},
+		backendRefs: []gatewayv1a2.HTTPBackendRef{{
+			BackendRef: gatewayv1a2.BackendRef{
+				BackendObjectReference: gatewayv1a2.BackendObjectReference{
+					Name: gatewayv1a2.ObjectName("test"),
+					Port: ptrTo(gatewayv1b1.PortNumber(8080)),
+				},
+			},
+		}},
 		errCount: 2,
 	}, {
 		name: "invalid HTTPRouteFilterRequestHeaderModifier type filter with empty value field",
 		routeFilter: gatewayv1a2.HTTPRouteFilter{
 			Type: gatewayv1b1.HTTPRouteFilterRequestHeaderModifier,
 		},
+		backendRefs: []gatewayv1a2.HTTPBackendRef{{
+			BackendRef: gatewayv1a2.BackendRef{
+				BackendObjectReference: gatewayv1a2.BackendObjectReference{
+					Name: gatewayv1a2.ObjectName("test"),
+					Port: ptrTo(gatewayv1b1.PortNumber(8080)),
+				},
+			},
+		}},
 		errCount: 1,
 	}, {
 		name: "valid HTTPRouteFilterRequestMirror route filter",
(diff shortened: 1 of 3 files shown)


robscott:
Hey @Xunzhuo, thanks for working on this! We really need more information in the PR description to help reviewers understand what the goals are. For example, https://github.com/kubernetes-sigs/gateway-api/pull/1753 has good detail for each section.

As far as the content of the PR, I think we need to have some broader discussion about this before merging. This PR would effectively prevent a case where redirect loop as proposed by @rainest in https://github.com/kubernetes-sigs/gateway-api/issues/1185 is possible. As I see it, we essentially have two options:

1. Say that redirect loop protection is likely not viable and instead require that both a redirect and backendRef cannot be specified in the same route rule (that's this PR).
2. Update the spec to define how redirect loop protection would work, enabling a single route rule to express something like "redirect to HTTPS, if already there, forward to backend foo". 

My gut tells me that 2 would be pretty challenging and that 1 is the most viable solution at this point, but we need to be very clear that this is a change to the API Spec and include a release note + API Spec update if that's the direction we go.

Since this would be a significant change, I want to hold off until we can discuss this at a community meeting + get some feedback on the high level direction of this PR first.

/hold
/cc @youngnick @rainest @sunjayBhatia @howardjohn

Xunzhuo:
Yes @robscott, I am also a bit concerned about this change, since this is a user-facing change. We definitely need more discussion before pushing this forward.

Take this as a draft PR, changes can make after the final approach is out.

sunjayBhatia (commented):
I may have missed some discussion here from a community meeting, but in Contour we currently ignore backendrefs if you've set a redirect filter so option 1 Rob describes seems fine

we could also specify that infinite redirect protection is recommended but implementation-specific support (or extended if we can write up a nice conformance test)

apis/v1alpha2/validation/httproute_test.go
```diff
@@ -816,19 +817,43 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
...
 			},
 		},
+		backendRefs: []gatewayv1a2.HTTPBackendRef{{
+			BackendRef: gatewayv1a2.BackendRef{
+				BackendObjectReference: gatewayv1a2.BackendObjectReference{
+					Name: gatewayv1a2.ObjectName("test"),
+					Port: ptrTo(gatewayv1b1.PortNumber(8080)),
+				},
+			},
+		}},
```
robscott, line 827: It looks like this is repeated a lot, can you move the standard set of backendRefs to a var so there's not quite so much repetition here?

robscott: Actually it looks like this isn't modified at all for any of the test cases, maybe just add this to the route populated in `t.Run` below?

apis/v1beta1/validation/httproute.go
```diff
@@ -74,6 +74,33 @@ func ValidateHTTPRouteSpec(spec *gatewayv1b1.HTTPRouteSpec, path *field.Path) fi
...
+				if len(rule.BackendRefs) != 0 {
+					errs = append(errs, field.Invalid(path.Index(ri).Child("filters").Index(fi), filter.RequestRedirect, "cannot specify backendRefs when using redirect filter"))
+				}
+				break
+			}
+		}
+		for bi, backendRef := range rule.BackendRefs {
+			for fi, filter := range backendRef.Filters {
+				if filter.RequestRedirect != nil {
+					errs = append(errs, field.Invalid(path.Index(ri).Child("backendRefs").Index(bi).Child("filters").Index(fi), filter.RequestRedirect, "cannot specify backendRef when using redirect filter"))
```
robscott, line 97:
```suggestion
					errs = append(errs, field.Invalid(path.Index(ri).Child("backendRefs").Index(bi).Child("filters").Index(fi), filter.RequestRedirect, "cannot specify RequestRedirect filter within a backendRef"))
```

apis/v1beta1/validation/httproute.go
```diff
@@ -74,6 +74,33 @@ func ValidateHTTPRouteSpec(spec *gatewayv1b1.HTTPRouteSpec, path *field.Path) fi
...
+
+// validateInvalidBackendWithRedirectFilter validates if an HTTPRoute that specifies a redirect filter and specifying a backendRef
+func validateInvalidBackendWithRedirectFilter(rules []gatewayv1b1.HTTPRouteRule, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+
+	for ri, rule := range rules {
+		for fi, filter := range rule.Filters {
+			if filter.RequestRedirect != nil {
+				if len(rule.BackendRefs) != 0 {
+					errs = append(errs, field.Invalid(path.Index(ri).Child("filters").Index(fi), filter.RequestRedirect, "cannot specify backendRefs when using redirect filter"))
```
robscott, line 89:
```suggestion
					errs = append(errs, field.Invalid(path.Index(ri).Child("filters").Index(fi), filter.RequestRedirect, "cannot specify backendRefs when using RequestRedirect filter"))
```

apis/v1beta1/validation/httproute.go
```diff
@@ -338,10 +365,6 @@ func hasExactlyOnePrefixMatch(matches []gatewayv1b1.HTTPRouteMatch) bool {
 	if len(matches) != 1 || matches[0].Path == nil {
 		return false
 	}
-	pathMatchType := matches[0].Path.Type
-	if *pathMatchType != gatewayv1b1.PathMatchPathPrefix {
-		return false
-	}
 
-	return true
+	return *matches[0].Path.Type == gatewayv1b1.PathMatchPathPrefix
```
robscott, line 369: This feels not very safe and also not relevant to the rest of the PR, maybe revert this change?

apis/v1beta1/validation/httproute_test.go
```diff
@@ -997,19 +999,43 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
 				Remove: []string{"remove"},
 			},
 		},
+		backendRefs: []gatewayv1b1.HTTPBackendRef{{
+			BackendRef: gatewayv1b1.BackendRef{
+				BackendObjectReference: gatewayv1b1.BackendObjectReference{
+					Name: gatewayv1b1.ObjectName("test"),
+					Port: ptrTo(gatewayv1b1.PortNumber(8080)),
+				},
+			},
```
robscott, line 1008: This also looks like it's repeated a lot, can we use a var for this?

robscott (commented):
Thanks for the work on this @Xunzhuo!

PR #1814, opened 2023-03-14
Release v0.6.2

shaneutt:
**What type of PR is this?**

/kind test

This adds the content for the `v0.6.2` release.

### DIFF
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -45,7 +45,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 					Kind:  v1beta1.Kind("HTTPRoute"),
 				}},
 				Conditions: []metav1.Condition{{
-					Type:   string(v1beta1.RouteConditionAccepted),
+					Type:   string(v1beta1.ListenerConditionAccepted),
 					Status: metav1.ConditionTrue,
 					Reason: "", //any reason
 				}},
@@ -64,7 +64,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 					Kind:  v1beta1.Kind("HTTPRoute"),
 				}},
 				Conditions: []metav1.Condition{{
-					Type:   string(v1beta1.RouteConditionAccepted),
+					Type:   string(v1beta1.ListenerConditionAccepted),
 					Status: metav1.ConditionTrue,
 					Reason: "", //any reason
 				}},
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -28,32 +28,16 @@ import (
 )
 
 func init() {
-	ConformanceTests = append(ConformanceTests, GatewayInvalidRouteKind)
+	ConformanceTests = append(ConformanceTests, GatewayWithAttachedRoutes)
 }
 
-var GatewayInvalidRouteKind = suite.ConformanceTest{
-	ShortName:   "GatewayInvalidRouteKind",
-	Description: "A Gateway in the gateway-conformance-infra namespace should fail to become ready an invalid Route kind is specified.",
-	Manifests:   []string{"tests/gateway-invalid-route-kind.yaml"},
+var GatewayWithAttachedRoutes = suite.ConformanceTest{
+	ShortName:   "GatewayWithAttachedRoutes",
+	Description: "A Gateway in the gateway-conformance-infra namespace should be attached to routes.",
+	Manifests:   []string{"tests/gateway-with-attached-routes.yaml"},
 	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
-		t.Run("Gateway listener should have a false ResolvedRefs condition with reason InvalidRouteKinds and no supportedKinds", func(t *testing.T) {
-			gwNN := types.NamespacedName{Name: "gateway-only-invalid-route-kind", Namespace: "gateway-conformance-infra"}
-			listeners := []v1beta1.ListenerStatus{{
-				Name:           v1beta1.SectionName("http"),
-				SupportedKinds: []v1beta1.RouteGroupKind{},
-				Conditions: []metav1.Condition{{
-					Type:   string(v1beta1.ListenerConditionResolvedRefs),
-					Status: metav1.ConditionFalse,
-					Reason: string(v1beta1.ListenerReasonInvalidRouteKinds),
-				}},
-				AttachedRoutes: 0,
-			}}
-
-			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
-		})
-
-		t.Run("Gateway listener should have a false ResolvedRefs condition with reason InvalidRouteKinds and HTTPRoute must be put in the supportedKinds", func(t *testing.T) {
-			gwNN := types.NamespacedName{Name: "gateway-supported-and-invalid-route-kind", Namespace: "gateway-conformance-infra"}
+		t.Run("Gateway listener should have one valid http routes attached", func(t *testing.T) {
+			gwNN := types.NamespacedName{Name: "gateway-with-one-attached-route", Namespace: "gateway-conformance-infra"}
 			listeners := []v1beta1.ListenerStatus{{
 				Name: v1beta1.SectionName("http"),
 				SupportedKinds: []v1beta1.RouteGroupKind{{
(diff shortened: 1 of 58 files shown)


shaneutt:
Please review, but hold until I'm ready:

/hold

mlavacca (approved):
After having run this branch with the Kong Ingress controller, it turned out we have conformance issues with:
- `HTTPRouteInvalidCrossNamespaceParentRef` (we are not setting the Reason `NotAllowedByListeners`) https://github.com/kubernetes-sigs/gateway-api/pull/1714
- `GatewayWithAttachedRoutes` (we are not updating the `AttachedRoutes` https://github.com/kubernetes-sigs/gateway-api/pull/1624)

Since these changes look legit and we need to fix these issues in our implementation, LGTM.

/lgtm

robscott:
Thanks @shaneutt!

/lgtm

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Run("Gateway listener should have one valid http routes attached", func(t *testing.T) {
+			gwNN := types.NamespacedName{Name: "gateway-with-one-attached-route", Namespace: "gateway-conformance-infra"}
+			listeners := []v1beta1.ListenerStatus{{
+				Name: v1beta1.SectionName("http"),
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
+				}},
+				Conditions: []metav1.Condition{{
+					Type:   string(v1beta1.RouteConditionAccepted),
```
sunjayBhatia, line 48: nit: should this be `ListenerConditionAccepted`

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Run("Gateway listener should have two valid http routes attached", func(t *testing.T) {
+			gwNN := types.NamespacedName{Name: "gateway-with-two-attached-routes", Namespace: "gateway-conformance-infra"}
+			listeners := []v1beta1.ListenerStatus{{
+				Name: v1beta1.SectionName("http"),
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
+				}},
+				Conditions: []metav1.Condition{{
+					Type:   string(v1beta1.RouteConditionAccepted),
```
sunjayBhatia, line 67: same nit as above

conformance/tests/gateway-with-attached-routes.go
shaneutt: Sounds good, good catch! Changed in b9d5351349d0ce76ab13b872e03fc5d847b1f859

sunjayBhatia:
running conformance with contour now, will update here on what goes on

shaneutt:
Will consider @sunjayBhatia's upcoming report along with the reviews from @mlavacca and @robscott enough to unhold :+1:

/unhold

conformance/tests/httproute-disallowed-kind.go
```diff
@@ -43,6 +43,7 @@ var HTTPRouteDisallowedKind = suite.ConformanceTest{
 
 		routeNN := types.NamespacedName{Name: "disallowed-kind", Namespace: "gateway-conformance-infra"}
 		gwNN := types.NamespacedName{Name: "tlsroutes-only", Namespace: "gateway-conformance-infra"}
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
```
sunjayBhatia, line 46: should this be ResolvedRefs = True?

```
	// Possible reasons for this condition to be false are:
	//
	// * "RefNotPermitted"
	// * "InvalidKind"
	// * "BackendNotFound"
```

False and reason `InvalidKind` seems more appropriate here?

sunjayBhatia: i guess ResolvedRefs in this case is just for backendRefs?

sunjayBhatia: hm i guess in Contour's case we stop processing a Route once it has an invalid parent reference so we don't continue on to validate the rest of its references

sunjayBhatia: I would maybe push back a little on this line in the test, just because this is an error case so prescribing how an implementation handles error cases like this (that you still have to process the rest of the resource) is a little odd

shaneutt: I think the original intention here was that:

1. the reference itself was permitted
2. the kind was valid
3. the backend was found

These three aspects currently comprise the list of things that would make the condition false:

```go
        // This condition indicates whether the controller was able to resolve all
        // the object references for the Route.
        //
        // Possible reasons for this condition to be true are:
        //
        // * "ResolvedRefs"
        //
        // Possible reasons for this condition to be false are:
        //
        // * "RefNotPermitted"
        // * "InvalidKind"
        // * "BackendNotFound"
        //
        // Controllers may raise this condition with other reasons,
        // but should prefer to use the reasons listed above to improve
        // interoperability.
        RouteConditionResolvedRefs RouteConditionType = "ResolvedRefs"
```

But I can see how saying the references are resolved, but the route is not accepted feels a little bit odd.

After pondering it some more I'm kinda leaning towards sticking with this as it seems pretty reasonable to inform end-users of a situation where "yes I can resolve the references you provided and they are allowed, but I can't accept the route because there's no valid listener". :thinking: 

I'm open to being persuaded however, let me know your thoughts?

sunjayBhatia: yeah my bad i saw kind in the test name and immediately thought this line was referencing the "parent" kind but it isn't, it's referencing the backendrefs kinds, so that part of my comment can be ignored

this change https://github.com/kubernetes-sigs/gateway-api/pull/1668 seems to have added this line everywhere but really seems like it should be part of the `GatewayAndHTTPRoutesMustBeAccepted` helper (or something similar) since it is part of validating the base set of assumptions for a happy-path test?

making an assertion about ResolvedRefs in every test feels a *little* off, esp. for error cases where IMO we want to test some specific behavior but don't necessarily care about other things

similar to this test `HTTPRouteInvalidParentRefNotMatchingListenerPort` does not check for `ResolvedRefs`, just the `Accepted` condition

in Contour's case, we don't bother with processing the rest of the Route if it can't actually result in any traffic due to not attaching to a Gateway properly

regardless of what we end up going with, some agreed upon negative/error case handling would be a good outcome (e.g. the full set of conditions should still be present etc.) since we have some inconsistencies here

shaneutt: Are you suggesting we should drop the #1668 changes for this release so we can follow up, or do you think we can move forward?

shaneutt: @mlavacca interested to get your thoughts here? :thinking:

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -0,0 +1,77 @@
...
+		t.Run("Gateway listener should have one valid http routes attached", func(t *testing.T) {
+			gwNN := types.NamespacedName{Name: "gateway-with-one-attached-route", Namespace: "gateway-conformance-infra"}
+			listeners := []v1beta1.ListenerStatus{{
+				Name: v1beta1.SectionName("http"),
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
+				}},
+				Conditions: []metav1.Condition{{
+					Type:   string(v1beta1.ListenerConditionAccepted),
```
sunjayBhatia, line 48: coincidentally this is the first conformance test that actually asserts on `ListenerConditionAccepted` so it might cause a little friction, there are a few tests using `Programmed` for similar assertions (e.g. `GatewaySecretReferenceGrantAllInNamespace`)

conformance/tests/httproute-disallowed-kind.go
mlavacca: AFAIK, the `Accepted` and `ResolvedRefs` conditions are unrelated, a Route can be accepted/not accepted and can have the `ResolvedRefs` condition either set to true or false, depending on the outcome of the references. For this reason, moving it into the `GatewayAndHTTPRoutesMustBeAccepted` is misleading in my opinion.  

> similar to this test HTTPRouteInvalidParentRefNotMatchingListenerPort does not check for ResolvedRefs, just the Accepted condition

It looked like The ResolvedRefs condition was checked [here](https://github.com/kubernetes-sigs/gateway-api/blob/26080e900c1bf159e36b9d70c2ada5ae8ecf2838/conformance/tests/httproute-invalid-parentref-not-matching-listener-port.go#L45-L51), but the name was wrong, it actually checks the accepted condition. I prepared [this PR](https://github.com/kubernetes-sigs/gateway-api/pull/1819) to fix that and add the `resolvedRefs` check there as well as in the `httproute-invalid-cross-namespace-parent-ref` test.

shaneutt: As I understand it abdc042057572a1a85919629a7a3a564cb367b9a should resolve this issue as it adds the missing consistency. @sunjayBhatia if you feel it _doesn't_ resolve it, please unresolve the thread and let us know what prescriptive actions you feel need to be taken next.

Putting this PR back on hold to ensure we resolve @sunjayBhatia's needs and stake in this.

/hold

conformance/tests/gateway-with-attached-routes.go
shaneutt: As I understand it this is more of a calling out than a request for change. Let me know if I'm misunderstood, however.

sunjayBhatia: yeah just a callout in case anyone else sees an issue with it

sunjayBhatia (approved):

conformance/tests/httproute-disallowed-kind.go
shaneutt: Sounds good to me. Do you mind making an issue for this as a follow-up or would you prefer me to?

PR #1816, opened 2023-03-15
docs: some clarity for Service type ExternalName

### LINKED ISSUE #1070: Explicitly define ExternalName expectations
Currently, Service is a supported backendRef. A service can be of type ExternalName, which has non-obvious semantics.

In the Ingress world, I previously did some research into how different controllers handle differently; I don't think I was able to find two implementations doing the same thing.

It would be great to explicitly define how this should be treated. Depending on the answer, we may need to more broadly address how we access external sites.

shaneutt:
**What type of PR is this?**

/kind documentation

There's been a historical lack of clarity on how `ExternalName` type `Services` would be handled. This patch adds documentation which points out the hazards of them, and marks them as being under "Implementation-specific" with implementations discouraged from supporting them unless _they are certain they need to_.

Fixes #1070

```release-note
Services which don't have selectors (e.g. ExternalName) are now considered supported at an implementation-specific level
```


apis/v1beta1/object_reference_types.go
```diff
@@ -99,9 +99,33 @@ type BackendObjectReference struct {
...
+	// Services that utilize selectors to be considered supported when used as
+	// a BackendObjectReference at a "Core" level of support.
+	//
+	// Services which do not use Selectors (e.g. ExternalName, and headless
+	// Services without selectors) have problems with clarity about what
+	// Namespaces traffic would be destined for (and in many cases like
+	// ExternalName they can be destined in places outside the cluster
+	// entirely) and as such are considered inherently unsafe. Implementations
+	// should avoid supporting these types of Services unless they are
+	// VERY certain that they need to handle them. Due to the ambiguity
```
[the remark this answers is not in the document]
robscott: Also tempted to say something stronger like this and only add exceptions if/when they're requested.
```suggestion
	// should avoid supporting these types of Services.
```

apis/v1beta1/object_reference_types.go
```diff
@@ -99,9 +99,33 @@ type BackendObjectReference struct {
...
+	// described here, we consider Services which do not use Selectors to be
+	// supported under the "Implementation-specific" level of support for use as
+	// BackendObjectReferences.
+	//
+	// Support: Core (Services with Selectors)
+	//
+	// Support: Implementation-specific (Services without Selectors (e.g. ExternalName))
```
robscott, line 127: Whether or not a Service has a Selector actually does not matter here. The presence of a selector is used to determine if Endpoint(Slice) controller(s) should create Endpoint(Slice)s for the Service. With Endpoints, leaving the selector off was the only way to attach custom self-managed endpoints. With EndpointSlices, custom self-managed endpoints can be attached to any Service, regardless of the presence of a Selector. In my opinion, the only kind of Service that should get special treatment would be ExternalName. If implementations are forwarding to ClusterIPs instead of EndpointSlices we'd also need special treatment for Headless Services, but I think I'd rather just recommend that Services forward to individual endpoints.

apis/v1beta1/object_reference_types.go
```diff
@@ -99,9 +99,33 @@ type BackendObjectReference struct {
...
+	// Kind is the Kubernetes resource kind of the referent. For example
+	// "Service" or "HTTPRoute".
+	//
 	// Defaults to "Service" when not specified.
 	//
+	// Selectors in Kubernetes Services provide an abstraction around Pods,
+	// meaning that when a Service uses selectors Gateway API implementations
+	// can reason about what Namespaces the traffic will go to and can therefore
+	// enforce namespace boundaries (e.g. ReferenceGrant). As such we consider
+	// Services that utilize selectors to be considered supported when used as
```
robscott, line 111: Commented below, but I don't think this is the right thing to make the distinction on.

robscott (commented):
Thanks @shaneutt! We really needed to clarify this somewhere.

apis/v1beta1/object_reference_types.go
shaneutt: > Whether or not a Service has a Selector actually does not matter here. The presence of a selector is used to determine if Endpoint(Slice) controller(s) should create Endpoint(Slice)s for the Service.

Right, this is why it _does_ matter, because it means the Endpoint(Slice)s will be made _predictably_.

> With Endpoints, leaving the selector off was the only way to attach custom self-managed endpoints. With EndpointSlices, custom self-managed endpoints can be attached to any Service, regardless of the presence of a Selector. 

Right, this is why in this draft I was considered `Selectors` the "line in the sand", on the other side of that line we can't predict what A/AAAA/CNAME entries will actually point to.

> If implementations are forwarding to ClusterIPs instead of EndpointSlices we'd also need special treatment for Headless Services, but I think I'd rather just recommend that Services forward to individual endpoints.

Wouldn't it be better to just "draw the line in the sand" and say these kinds of use cases are `Implementation-specific`, as they are hard to reason about in a _generalized manner_ due to their inherent distinctiveness?

howardjohn: > Right, this is why it does matter, because it means the Endpoint(Slice)s will be made predictably.

You can have a Service with a selector and then attach user-defined EndpointSlices to it, so I don't think that is an accurate distinction.

Also EndpointSlice can also have a FQDN which behaves the same way as ExternalName. The only difference, AFAIK, is the EndpointSlice write access is considered "privileged", I think?

shaneutt: You're right that it wont be true in all cases, and while I considered trying to tease this out in the language further to get the effect I was heading for, I ultimately decided that for now the better thing would be to make a meaningful language change that I expect wouldn't be controversial and gives us a bit more clarity. As such, I'm dropping this language in favor of just focusing on calling `ExternalName` type `Services` supported as `Implementation-specific` as I think that's a better baseline for now. Thanks for the thoughts and conversation on this!

robscott: x-ref https://github.com/kubernetes/kubernetes/issues/114210 for FQDN EndpointSlices. TLDR, it's a mess. We added it without clearly defining the behavior. FQDN has been deprecated for a bit, but it may be hard to completely remove it.

apis/v1beta1/object_reference_types.go
```diff
@@ -99,9 +99,22 @@ type BackendObjectReference struct {
 	// +kubebuilder:default=""
 	Group *Group `json:"group,omitempty"`
 
-	// Kind is kind of the referent. For example "HTTPRoute" or "Service".
+	// Kind is the Kubernetes resource kind of the referent. For example
+	// "Service" or "HTTPRoute".
```
robscott, line 103: I know that HTTPRoute is not something that was added by you, but reading it now it really doesn't make sense that HTTPRoute could be a BackendRef unless we had defined route delegation.
```suggestion
	// "Service".
```

apis/v1beta1/object_reference_types.go
```diff
@@ -99,9 +99,22 @@ type BackendObjectReference struct {
...
+	//
 	// Defaults to "Service" when not specified.
 	//
+	// ExternalName services can refer to CNAME DNS records that may live
+	// outside of the cluster and as such are difficult to reason about in
+	// terms of conformance. Implementations are recommended to not support
+	// ExternalName Services when feasible. As such we consider ExternalName
+	// Services to be supported at an "Implementation-specific" level for usage
+	// as a BackendObjectReference.
```
robscott, line 112: Can we start with a stronger statement here? If someone really wants to support ExternalName services we can revisit this at that point.
```suggestion
	// terms of conformance. They also may not be safe to forward to (see 
	// CVE-2021-25740 for more information). Implementations SHOULD NOT
	// support ExternalName Services.
```

robscott (commented):
Thanks @shaneutt!

apis/v1beta1/object_reference_types.go
mlavacca: When we state "Implementations SHOULD NOT support ExternalName Service." I guess we need tests to assert such behavior in our conformance suite. But if we really mean to block this possibility, we should assert that all the ways to mimic `externalName` (as with `endpointslices` with FQDN entries) are blocked and declared as not conformant as well.

apis/v1beta1/object_reference_types.go
shaneutt: Sounds good, I had a similar thought. Changed in bef0c2087ba2c5e69b69a5689a9c7c1799b32b86 :+1:

apis/v1beta1/object_reference_types.go
shaneutt: > Can we start with a stronger statement here? If someone really wants to support ExternalName services we can revisit this at that point.

Sounds good, changed in bef0c2087ba2c5e69b69a5689a9c7c1799b32b86 :+1: 

> When we state "Implementations SHOULD NOT support ExternalName Service." I guess we need tests to assert such 
> behavior in our conformance suite. But if we really mean to block this possibility, we should assert that all the ways to
> mimic externalName (as with endpointslices with FQDN entries) are blocked and declared as not conformant as well.

MUST NOT is something that we would definitely want to add conformance assertions for, but not SHOULD NOT. SHOULD NOT is saying "here be dragons" and giving a clear sign to implementers that they travel to these dragon infested waters at their own risk, yarrrr

mlavacca: Gotcha, it makes sense! I think that if we'd write documentation in this way, instead of the usual boring one, more people would read it :)

mlavacca (approved):
Thanks @shaneutt!

/lgtm

PR #1834, opened 2023-03-17
Clarify HTTPRequestRedirectFilter.Port Semantics

### LINKED ISSUE #1806: Port Redirect Semantics for HTTP and HTTPS Are Confusing
**What happened**:
It's not clear to me if the following portion of the spec is targeted at implementers or users. I initially [commented on the PR](https://github.com/kubernetes-sigs/gateway-api/pull/1740#pullrequestreview-1335290763), but creating an issue so we don't lose track of it.
https://github.com/kubernetes-sigs/gateway-api/blob/227de0756d763b320cb4dcfa16a668867e326554/apis/v1beta1/httproute_types.go#L905-L908

Ideally the spec would also clarify the following:

* Are users, implementers, or both supposed to leave Port unspecified in the Redirect Filter?
* What protocol are we looking at to determine HTTP or HTTPS? Is this a runtime decision or is it something we determine by looking at either the Listener protocol or the Redirect filter protocol?

spencerhance:
**What type of PR is this?**
/kind documentation

This PR clarifies Port Redirect Semantics for HTTP and HTTPS in the `HTTPRequestRedirectFilter.Port` doc.

Fixes #1806 

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -902,10 +902,15 @@ type HTTPRequestRedirectFilter struct {
 
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
+	//
 	// When empty, the Gateway Listener port is used.
-	// In all cases, when the protocol is http and the port is port 80,
-	// then the port must be omitted. Similarly, when the protocol is
-	// https and the port is port 443, then the port must also be omitted.
+	//
+	// Implementers should _not_ include the port number in the 'Location' header in the following cases:
+	//
+	// * a Location header that will use HTTP (whether that is determined via the Listener protocol
+	//   or the Scheme field) _and_ use port 80.
+	// * a Location header that will use HTTPS (whether that is determined via the Listener protocol
+	//   or the Scheme field) _and_ use port 443.
 	//
 	// Support: Extended
 	//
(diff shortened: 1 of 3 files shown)


apis/v1beta1/httproute_types.go
```diff
@@ -902,10 +902,14 @@ type HTTPRequestRedirectFilter struct {
...
-	// In all cases, when the protocol is http and the port is port 80,
-	// then the port must be omitted. Similarly, when the protocol is
-	// https and the port is port 443, then the port must also be omitted.
+	//
+	// Implementers should _not_ include the port number in the 'Location' header in the following cases:
+	// - a Location header that will use HTTP (whether that is determined via the Listener protocol
+	// or the Scheme field) _and_ use port 80.
+	// - a Location header that will use HTTPS (whether that is determined via the Listener protocol
+	// or the Scheme field) _and_ use port 443.
 	//
```
robscott, line 913: Nit: I think the following changes are required to render well as a markdown list:
```suggestion
	// Implementers should _not_ include the port number in the 'Location' header in the following cases:
	//
	// * a Location header that will use HTTP (whether that is determined via the Listener protocol
	//   or the Scheme field) _and_ use port 80.
	// * a Location header that will use HTTPS (whether that is determined via the Listener protocol
	//   or the Scheme field) _and_ use port 443.
	//
```

robscott (commented):
Thanks @spencerhance!

/cc @youngnick 
/ok-to-test
/approve

shaneutt (approved):
I approve as well, but does seem like @youngnick should give the final lgtm

youngnick:
Thanks @spencerhance, that's great.

/lgtm

Issue #1748, opened 2023-02-21
GEP: Define Gateway API Interaction with Multicluster Services


robscott:
**What would you like to be added**:
We need to define via documentation how Gateway API interacts with Multicluster Services. This will likely be a docs-only change, but may also include a change to API Spec to note "ServiceImport" is supported for "extended" conformance.

**Why this is needed**:
There are already implementations of Multi-Cluster Gateway either released or in progress, it would be great to ensure that we're all providing a consistent experience.

costinm:
I assume this can also cover GAMMA ( both as backend reference, but also attaching to ServiceImport in addition to Service as a frontend).

robscott:
> I assume this can also cover GAMMA

Yep, that's the goal, hoping to discuss more at the GAMMA meeting tomorrow.

Shared this doc at today's community meeting: https://docs.google.com/document/d/1akwzBKtMKkkUV8tX-O7tPcI4BPMOLZ-gmS7Iz-7AOQE/edit. Will work on a follow up GEP once we get some feedback on the doc.

youngnick:
/remove-lifecycle stale

@robscott, with the move of this to Memorandum status, what are your thoughts on next steps here?

mlavacca:
/remove lifecycle-rotten

Should we keep this one open as it has been recently moved to memorandum @robscott?

mlavacca:
/remove-lifecycle rotten

Should we keep this one open as it has been recently moved to memorandum @robscott?

PR #1843, opened 2023-03-17
Adding GEP-1748: Gateway API Interaction with Multi-Cluster Services

robscott:
**_Recommend reviewing deploy preview so examples are inlined: https://deploy-preview-1843--kubernetes-sigs-gateway-api.netlify.app/geps/gep-1748/._**

**What type of PR is this?**
/kind gep

This formally defines the relationship between Gateway API and Multi-Cluster Services as part of #1748.

I think the changelog notes should be saved for actual implementation PR, but open to ideas:

```release-note
NONE
```

/hold for consensus
Want at least one LGTM each from SIG-MC, GW API maintainers, and GAMMA leads before removing hold. Feedback from everyone (including people not directly mentioned below) is very appreciated.

SIG-MC:
/cc @lauralorenz @JeremyOT @pmorie 

GW API:
/cc @bowei @shaneutt @youngnick

GAMMA
/cc @howardjohn @keithmattix @mikemorris

### DIFF
--- a/examples/standard/multicluster/httproute-gamma.yaml
+++ b/examples/standard/multicluster/httproute-gamma.yaml
@@ -1,0 +1,19 @@
+#$ Used in:
+#$ - geps/gep-1748.md
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: store
+spec:
+  parentRefs:
+  - group: multicluster.x-k8s.io
+    kind: ServiceImport
+    name: store
+  rules:
+  - matches:
+    - path:
+        value: "/cart"
+    backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: cart
--- a/geps/gep-1748.md
+++ b/geps/gep-1748.md
@@ -1,0 +1,139 @@
+# GEP-1748: Gateway API Interaction with Multi-Cluster Services
+
+* Issue: [#1748](https://github.com/kubernetes-sigs/gateway-api/issues/1748)
+* Status: Experimental
+
+## TLDR
+
+The Kubernetes Multi-Cluster Services API enables Services to span multiple
+clusters. Gateway API enables advanced traffic routing, serving as the next
+generation Ingress, Load Balancing, and Mesh API. This document describes how
+these APIs can be used together.
+
+## Goals
+
+* Define the interaction between Gateway API and Multi-Cluster Services
+* Define any situations where Gateway API may span multiple clusters without the
+  Multi-Cluster Services API
+
+## Non-Goals
+
+* Make significant changes to either API
+
+## Introduction
+
+Gateway API and the Multi-Cluster Services API represent two of the newest
+Kubernetes networking APIs. As they’ve been developed in parallel, there’s been
+some cross-project discussion about how they can interact, but that has never
+formally been written down. This GEP aims to change that.
+
+## Overview
+
+Multi-Cluster Services can be used within Gateway API wherever Services can be
+used. The difference is that Services refer only to cluster-local endpoints while
+Multi-Cluster Services can refer to endpoints throughout an entire ClusterSet.
+
+### ServiceImport as a Backend
+
+A Route can forward traffic to the endpoints attached to an imported Service.
+This behaves identically to how forwarding to Services work in Kubernetes, with
+the exception that the endpoints attached to a ServiceImport may span multiple
+clusters. For example, the following HTTPRoute would forward traffic to
+endpoints attached to the "store" ServiceImport:
+
+```yaml
+{% include 'standard/multicluster/httproute-simple.yaml' %}
+```
+
+#### Routing to Specific Clusters
+
+In some cases, it may be helpful to route certain paths to a specific cluster
+(or region). Similar to single-cluster Services, this can be accomplished by
+creating multiple Multi-Cluster Services, one for each subset of endpoints you
+would like to route to. For example, the following configuration will send
+requests with paths prefixed with “/west” to the store-west ServiceImport, and
+“/east” to the store-east ServiceImport. Requests that don’t match either of
+these paths will be routed to the “store” ServiceImport which represents a
+superset of the “store-west” and “store-east” ServiceImports.
+
+```yaml
+{% include 'standard/multicluster/httproute-location.yaml' %}
+```
+
+#### Advanced Routing With ServiceImports
+
+All Routing capabilities in Gateway API should apply equally whether the backend
+is a Service or ServiceImport. For example, when routing to a system with
+multiple read replicas, it could be beneficial to route requests based on HTTP
+Method. In the following example, requests with POST, PUT, and DELETE methods
+are routed to `api-primary` while the rest are routed to `api-replicas`:
+
+```yaml
+{% include 'standard/multicluster/httproute-method.yaml' %}
+```
+
+#### Routing to Both Services and ServiceImports
+
+There are some situations where it will be useful to split traffic between a
+Service and ServiceImport. In the following example, 90% of traffic would go to
+endpoints attached to the cluster-local "store" Service, and the remaining 10%
+would go to endpoints attached to the Multi-Cluster "store-global" Service:
+
+```yaml
+{% include 'standard/multicluster/httproute-hybrid.yaml' %}
+```
+
+#### Cross-Namespace References with ReferenceGrant
+
+It is possible to use ReferenceGrant to enable cross-namespace references to
+ServiceImports. For example, the following HTTPRoute would forward traffic to
+endpoints attached to the “bar” Multi-Cluster Service in the “bar” namespace:
+
+```yaml
+{% include 'standard/multicluster/httproute-referencegrant.yaml' %}
+```
+
+### Mesh: ServiceImport as Parent
+
+In some cases, you may want to override traffic destined for a Multi-Cluster
+Service with a mesh. As part of the broader GAMMA initiative, ServiceImport can
+be used in the same way that Service is used as a ParentRef. When a Service is
+specified as a parent, meshes will intercept traffic destined for the ClusterIP
+of the Service and apply any policies or routing decisions defined by the Route.
+Similarly, when a ServiceImport is specified as a parent, meshes will intercept
+traffic destined for the ClusterSetIP and apply any policies or routing
+decisions defined by the Route. In the following example, the mesh would
+intercept traffic destined for the store ClusterSetIP matching the `/cart` path
+and redirect it to the `cart` Multi-Cluster Service.
+
+```yaml
+{% include 'standard/multicluster/httproute-gamma.yaml' %}
+```
+
+### Services vs ServiceImports
+
+It is important that all implementations provide a consistent experience. That
+means that references to Services SHOULD always be interpreted as references to
+endpoints within the same cluster for that Service. References to ServiceImports
+MUST be interpreted as routing to Multi-Cluster endpoints across the ClusterSet
+for the given ServiceImport. In practice, that means that users should use
+“Service” when they want to reference cluster-local endpoints, and
+“ServiceImport” when they want to route to Multi-Cluster endpoints across the
+ClusterSet for the given ServiceImport. This behavior should be analogous to
+using `.cluster.local` versus `.clusterset.local` DNS for a given Service.
+
+## API Changes
+
+* ServiceImport is recognized as backend with “Extended” conformance
+* ServiceImport is included in GAMMA GEP(s) with “Extended” conformance
+* Clarification that Services refer to endpoints within the same cluster
+
+## Alternatives
+
+### Develop Custom Multi-Cluster Concepts Independently from Multi-Cluster Services
+
+We could theoretically develop an entirely new way to handle multi-cluster routing. We’re choosing not to do that because the existing APIs are sound and can work well together.
+
+## References
+
+* [Original Doc](https://docs.google.com/document/d/1akwzBKtMKkkUV8tX-O7tPcI4BPMOLZ-gmS7Iz-7AOQE/edit#)
--- a/examples/standard/multicluster/0-namespaces.yaml
+++ b/examples/standard/multicluster/0-namespaces.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: foo
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: bar
--- a/examples/standard/multicluster/httproute-hybrid.yaml
+++ b/examples/standard/multicluster/httproute-hybrid.yaml
@@ -1,0 +1,20 @@
+#$ Used in:
+#$ - geps/gep-1748.md
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: store
+spec:
+  parentRefs:
+  - name: external-http
+  rules:
+  - backendRefs:
+    - kind: Service
+      name: store
+      port: 8080
+      weight: 90
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: store-global
+      port: 8080
+      weight: 10
--- a/examples/standard/multicluster/httproute-location.yaml
+++ b/examples/standard/multicluster/httproute-location.yaml
@@ -1,0 +1,33 @@
+#$ Used in:
+#$ - geps/gep-1748.md
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: store
+spec:
+  parentRefs:
+  - name: external-http
+  rules:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /west
+    backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: store-west
+      port: 8080
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /east
+    backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: store-east
+      port: 8080
+  - backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: store
+      port: 8080
--- a/examples/standard/multicluster/httproute-method.yaml
+++ b/examples/standard/multicluster/httproute-method.yaml
@@ -1,0 +1,24 @@
+#$ Used in:
+#$ - geps/gep-1748.md
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: api
+spec:
+  parentRefs:
+  - name: api-gw
+  rules:
+  - matches:
+    - method: POST
+    - method: PUT
+    - method: DELETE
+    backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: api-primary
+      port: 8080
+  - backendRefs:
+    - group: multicluster.x-k8s.io
+      kind: ServiceImport
+      name: api-replicas
+      port: 8080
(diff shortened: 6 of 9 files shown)


geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Routing to Both Services and ServiceImports
+
+There are some situations where it will be useful to split traffic between a
+Service and ServiceImport. In the following example, 90% of traffic would go to
+endpoints attached to the cluster-local "store" Service, and the remaining 10%
+would go to endpoints attached to the Multi-Cluster "store-global" Service:
+
+```yaml
+{% include 'standard/multicluster/httproute-hybrid.yaml' %}
```
howardjohn, line 83: nit: The sample has the names backwards

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+endpoints for the local Service within the same cluster.
+
+```yaml
+{% include 'standard/multicluster/httproute-gamma.yaml' %}
+```
+
+### Services vs ServiceImports
+
+It is important that all implementations provide a consistent experience. That
+means that references to Services MUST always be interpreted as references to
```
keithmattix, line 117: I share John's concerns above; I'd be much more comfortable with the MUST being lessened to a SHOULD at least until MCS API is beta.

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+endpoints within the same cluster for that Service. References to ServiceImports
+MUST be interpreted as references to the full set of endpoints attached to that
+ServiceImport. In practice, that means that users should use “Service” when they
```
mikemorris, line 120: This is the biggest ⚠️ for me, and is a bit tricky at multiple levels:
- Reading this the first time, I had assumed `"full set of endpoints"` was intending to strictly define _the set of endpoints expected to be attached to a ServiceImport_ rather than just how they should be consumed, which would have been a significant expectation above the guidance in [KEP-1645](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services):
  > The `EndpointSlices` imported for a service are not guaranteed to exactly match the originally exported `EndpointSlices`, but each slice is guaranteed to map only to a single source cluster.

  I've heard guidance elsewhere that the _endpoints attached to a ServiceImport_ should be "representative" of the available multicluster endpoints (citation needed cc @JeremyOT) rather than necessarily the full set, which is important for allowing a ServiceImport to:
  - _directly_ expose only a subset of pod endpoints from other clusters, particularly in clustersets with a very large number of endpoints for a ServiceExport
  - _indirectly_ expose imported services through a transit gateway of some sort (rather than allowing direct pod <> pod connections across clusters), which may otherwise necessitate a permissive networking environment with egress and ingress allowed across a wide range of IPs and/or the assumption of a "flat" non-overlapping networking shared by all clusters (IIRC "reachability" of imported services is an assumption/requirement of MCS, but the exact mechanics of how to implement that are intentionally not proscribed)
- The phrase `"references to ... endpoints"` seems like it _might_ be understood as implying "direct" pod addressing, which could preclude using gateways between clusters as described above. Cloud provider MCS implementations do commonly enable this through VPCs, but some Gateway API implementations may want to reference off-cluster or multi-network services without relying on this substrate.
  - If this were clarified to more explicitly describe the contents of the EndpointSlices associated with a ServiceImport (whatever those may point to) this might be flexible enough.

There's some additional complexity around how cross-cluster "direct endpoint" routing might work for GAMMA, but I think that scope could be deferred if we can affirm that ServiceImport "indirect" routing through gateways in other clusters is a permissible implementation.

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+
+```yaml
+{% include 'standard/multicluster/httproute-gamma.yaml' %}
+```
+
+### Services vs ServiceImports
+
+It is important that all implementations provide a consistent experience. That
+means that references to Services MUST always be interpreted as references to
+endpoints within the same cluster for that Service. References to ServiceImports
```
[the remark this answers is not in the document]
robscott: >I wanted to flag that this seems to prohibit "transparent-proxy" mesh functionality that could automatically route traffic to a Service to viable multicluster endpoints unless explicitly encoded in Route configuration.

Yeah, that's the intent here. The idea would be that users should explicitly be able to request multicluster endpoints/routing, but never have that happen automatically. 

>For GAMMA configuration, this could be enabled per Service by writing a Route to direct all traffic intended for a Service to a corresponding ServiceImport backend to spread load across multicluster endpoints (as an alternative to explicitly updating application upstreams from *.cluster.local to *.clusterset.local), which seems reasonable enough, but does have some configuration overhead for enabling clusterset-wide routing for all multicluster services.

Agree that that's a potential workaround. I have updated this to a `SHOULD` for now based on @keithmattix's comment above, but would like to see Gateway API/GAMMA usage adopt the Kubernetes pattern that multicluster routing is something you opt in to. Maybe the more important concept would be that exporting a Service in one cluster is required for it to be accessible in other clusters. Open to input here, but if that's a workable solution, it would definitely be more consistent with the rest of Kubernetes.

I am interested in finding ways to lower the implementation overhead for MCS today, such as the current proposal that would remove the requirement to propagate EndpointSlices to each cluster. I'm hoping this can become just a configuration thing where users simply have to opt-in to exporting a Service and the rest is lightweight and automatic.

geps/gep-1748.md
robscott: Yeah I think that's a fair point. Hopefully MCS API is going to make it to beta sometime soon, but agree that we should not overly restrict Services until that point. Updated to "SHOULD".

geps/gep-1748.md
robscott: Woops, thanks!

geps/gep-1748.md
robscott: Good catch, thanks @mikemorris! I think you're right that the MCS Spec just requires a representative set, not the full set. I also agree that it allows for gateways between clusters or other intermediate steps. I've tried to reword this section to avoid requiring either direct pod addressing or routing to the full set of endpoints from each cluster. Not sure I've got it quite right yet, so let me know if you can think of a better way to state this requirement.

youngnick:
I will have comments on this, but I haven't had the bandwidth to review properly yet (requires me spending some more time on Cilium's existing Clustermesh to be able to talk about what it can do more intelligibly). Could we give me some time on this one please?

shaneutt:
> I will have comments on this, but I haven't had the bandwidth to review properly yet (requires me spending some more time on Cilium's existing Clustermesh to be able to talk about what it can do more intelligibly). Could we give me some time on this one please?

I think it's reasonable to make a review where you "request changes" and in the review comment just state that you are a stakeholder and need to review before it merges. This should take care of holding the review even if someone removes the hold tag until you've dismissed your review or approved.

lauralorenz (commented):
Little nits suggested. 

Representing SIG-Multicluster:
/lgtm

lauralorenz (approved):

lauralorenz:
Sorry I did the github-native approve thing and it seems to mean something nowadays 😅  but the hold stands!

keithmattix (approved):

geps/gep-1748.md
robscott: Resolving this for now, let me know if I missed anything @mikemorris

geps/gep-1748.md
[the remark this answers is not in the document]
robscott: Resolving this for now, let me know if I missed anything @mikemorris

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+```yaml
+{% include 'standard/multicluster/httproute-hybrid.yaml' %}
+```
+
+
+#### Cross-Namespace References with ReferenceGrant
+
+It is possible to use ReferenceGrant to enable cross-namespace references to
+ServiceImports. For example, the following HTTPRoute would forward traffic to
+endpoints attached to the “bar” Service in the “bar” namespace:
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

examples/standard/multicluster/httproute-gamma.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: store
+spec:
+  parentRefs:
+  - group: multicluster.x-k8s.io
+    kind: ServiceImport
+    name: store
+  rules:
+  - backendRefs:
+    - name: store
+      port: 8080
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
+# GEP-1748: Gateway API Interaction with Multi-Cluster Services
+
+* Issue: [#1748](https://github.com/kubernetes-sigs/gateway-api/issues/1748)
+* Status: Experimental
```
shaneutt, line 4: In just a general way I'm uncomfortable with things that would not _start_ in `Provisional` and spend some time there. Given how we use `Experimental` today, this would seem to indicate that most or all of the current questions and concerns about this initiative are more or less sorted out and we're ready to implement and try it. My suggestion would be:

```suggestion
* Status: Provisional
```

And then maybe to "experiment" (as distinct from what we call the "experimental" today) we could put it through the `Prototyping` status where we starting demoing an implementation? As far as I can see, this change in approach would result in a very similar trajectory, but with a more clear message that implementations should not necessarily start implementing this right away? I'm open to suggestion I just need some help to better understand the motivation.

shaneutt (changes requested):
On first review I don't have any specific immediate concerns with the contents of the GEP overall.

What I am concerned about is the launch straight to experimental, considering that we consider experimental to be something more like "next".

geps/gep-1748.md
robscott: I think this is different than other GEPs in that we're not actually adding any APIs here or making any changes to existing APIs, we're just defining how APIs should interact. In that sense, I'd consider it similar to how Policy Attachment is also in an experimental state. 

> we could put it through the Prototyping status where we starting demoing an implementation?

There are already 2 implementations that support Gateway API and MCS API (GKE and AWS Lattice) with at least one more on the way soon. @liwenwu-amazon and I will be [speaking about our implementations at KubeCon](https://sched.co/1HyZO) in a couple weeks and I believe both of our implementations meet the guidelines described here. 

I can't really think of any other possible interpretation of how these APIs could interact with each other. In my opinion this is really just formalizing something that is already being done in practice.

shaneutt: > In that sense, I'd consider it similar to how Policy Attachment is also in an experimental state.

Policy attachment has been a communicated concept for some time and something that's more of a suggestion in practice, so it seems a bit different from this which was only colloquially understood by some implementations implementing it independently, and is also somewhat niche.

> I can't really think of any other possible interpretation of how these APIs could interact with each other. In my opinion this is really just formalizing something that is already being done in practice.

I get this, and I personally agree with it which is why I said in my above review that I'm for the content overall. My main concern is that this comes in fast, we all think it's good, and then someone comes along later and actually does have some different interpretation but we're like "sorry we're in production with this already". We wouldn't want that scenario for anything, of course, I suppose the flip side is that we _always_ have the potential of running into this kind of problem even if things take a different path, if someone interested simply takes to long to show up :shrug: 

I'm getting the point that we're basically already in a scenario where "it's already a thing" as we're just formalizing something that people are already doing, so I perhaps the best thing for us to do is just make sure we get an approval from the 3 implementers we have so far and make sure we haven't missed anything. I'll consider this comment resolved thanks for the discussion.

shaneutt (approved):

liwenwu-amazon:
The GEP looks good!. In [AWS VPC Lattice controller gateway API/multicluster implementation](https://github.com/aws/aws-application-networking-k8s),  VPC Lattice is the federation.   

For serviceexport, the lattice controller register all of its endpoints in local cluster to VPC lattice

To reference a serviceimport as a backendRef of a HTTPRoute, the lattice controller will configure VPC lattice resource accordingly. So that the traffic will get routed to endpoints in that remote cluster(VPC) by VPC lattice dataplane.

In addition to K8S endpoints, serviceimport can references targets like EC2, lambda function and ECS container(s).

youngnick:
Because this is extended, this looks fine to me. I don't think we'll be doing this in Cilium for a while though.

/lgtm

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+endpoints attached to the “bar” Multi-Cluster Service in the “bar” namespace:
+
+```yaml
+{% include 'standard/multicluster/httproute-referencegrant.yaml' %}
+```
+
+### Mesh: ServiceImport as Parent
+
+In some cases, you may want to override traffic destined for a Multi-Cluster
+Service with a mesh. As part of the broader GAMMA initiative, ServiceImport can
```
zhiying-lin, line 100: do we have any doc to describe how mesh integrates with gateway api?

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+specified as a parent, meshes will intercept traffic destined for the ClusterIP
+of the Service and apply any policies or routing decisions defined by the Route.
+Similarly, when a ServiceImport is specified as a parent, meshes will intercept
+traffic destined for the ClusterSetIP and apply any policies or routing
+decisions defined by the Route. In the following example, the mesh would
+intercept traffic destined for the store ClusterSetIP and redirect it to
+endpoints for the local Service within the same cluster.
+
+```yaml
+{% include 'standard/multicluster/httproute-gamma.yaml' %}
```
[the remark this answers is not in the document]
mikemorris: Service as a parentRef for TCPRoute is a viable use case, and is included in [GEP-1426](https://gateway-api.sigs.k8s.io/geps/gep-1426/). HTTPRoute is important for some more advanced configuration, such as redirecting only a specific path (like `/admin`) to a different Service backendRef when doing something like splitting up a monolith application.

geps/gep-1748.md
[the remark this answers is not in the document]
mikemorris: > The idea would be that users should explicitly be able to request multicluster endpoints/routing, but never have that happen automatically.

Was talking with @howardjohn a bit about this yesterday, and the concern is that this approach can force multicluster load balancing or failover to be a concern of the service _consumer_ rather than the _producer_ who may best understand how to shape incoming traffic to a service under their management to ensure reliability and high availability.

Consumer opt-in can make sense for service discovery use cases when DNS is the primary interface to return workload addresses to dial directly (Consul's service discovery uses [prepared queries](https://developer.hashicorp.com/consul/docs/services/discovery/dns-dynamic-lookups) for similar client-defined routing behavior), but allowing producer-defined routing rules to direct `.cluster.local` requests to off-cluster endpoints is an important mesh capability.

I think we have enough wiggle room with the change to `SHOULD`, but definitely want to get this explored in more depth as a GAMMA+MCS user stories.

> would like to see Gateway API/GAMMA usage adopt the Kubernetes pattern that multicluster routing is something you opt in to

For me the key is which persona is opting in to this functionality. Agreed it would be dangerous for a cluster/mesh administrator to enable global multi-cluster routing (it's nice for demos though 😉), but we shouldn't have to rely on all downstream consumers of a service understanding when and how to configure multi-cluster routing for each upstream service they depend on.

> Maybe the more important concept would be that exporting a Service in one cluster is required for it to be accessible in other clusters.

Is this not already an assumption/requirement? **Definitely agreed** with this - meshes _may_ offer a higher-level API to do something like automatically create ServiceExports for every service in a given namespace, but having services explicitly exported in order for an MCS implementation to create corresponding ServiceImports on other clusters in the group/clusterset is definitely essential - some services should likely always stay cluster-local.

> I am interested in finding ways to lower the implementation overhead for MCS today

Same! I'm excited to see progress on conformance tests, and getting more clarity around some of these behaviors or edge cases is going to be important for MCS to be a useful building block for multi-cluster service mesh deployments.

examples/standard/multicluster/httproute-gamma.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  parentRefs:
+  - group: multicluster.x-k8s.io
+    kind: ServiceImport
+    name: store
+  rules:
+  - backendRefs:
+    - kind: Service
+      name: store
+      port: 8080
```
mikemorris, line 16: This appears to be configuring traffic to a `ServiceImport` ClusterSetIP to instead stay cluster-local, which feels a bit odd (and possibly more natural to configure by either not exporting the service in the first place or blocking `store.svc.clusterset.local` with egress policy). Would the inverse of this make more sense?
```suggestion
spec:
  parentRefs:
  - kind: Service
    name: store
  rules:
  - backendRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: store
```
**Alternatively**, splitting off one path from a `ServiceImport` to a _different_ `ServiceImport` would be an alternate viable pattern if we're explicitly trying to illustrate `ServiceImport` as a `parentRef`.
```suggestion
spec:
  parentRefs:
  - group: multicluster.x-k8s.io
    kind: ServiceImport
    name: store
  rules:
  - matches:
    - path:
      value: "/cart"
    backendRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: cart
```

geps/gep-1748.md
```diff
@@ -0,0 +1,140 @@
...
+In some cases, you may want to override traffic destined for a Multi-Cluster
+Service with a mesh. As part of the broader GAMMA initiative, ServiceImport can
+be used in the same way that Service is used as a ParentRef. When a Service is
+specified as a parent, meshes will intercept traffic destined for the ClusterIP
+of the Service and apply any policies or routing decisions defined by the Route.
+Similarly, when a ServiceImport is specified as a parent, meshes will intercept
+traffic destined for the ClusterSetIP and apply any policies or routing
+decisions defined by the Route. In the following example, the mesh would
+intercept traffic destined for the store ClusterSetIP and redirect it to
+endpoints for the local Service within the same cluster.
```
mikemorris, line 108:
```suggestion
intercept traffic destined for the store ClusterSetIP matching the `/cart` path
and redirect it to the `cart` Multi-Cluster Service.
```

examples/standard/multicluster/httproute-gamma.yaml
mikemorris, line 16:
```suggestion
spec:
  parentRefs:
  - group: multicluster.x-k8s.io
    kind: ServiceImport
    name: store
  rules:
  - matches:
    - path:
      value: "/cart"
    backendRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: cart
```

After reading through the updated content, I think splitting off one path from a `ServiceImport` to a _different_ `ServiceImport` would be a better example for illustrating `ServiceImport` as a `parentRef` than redirecting ClusterSetIP traffic to stay cluster-local.

Configuring traffic to a `ServiceImport` ClusterSetIP to instead stay cluster-local feels a bit odd (and possibly more natural to configure by either not exporting the service in the first place or blocking `store.svc.clusterset.local` with egress policy).

mikemorris: Would we additionally want to add an example for redirecting ClusterIP traffic to a Multi-Cluster Service, or should we wait on explicitly endorsing that?
```spec:
  parentRefs:
  - kind: Service
    name: store
  rules:
  - backendRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: store
```

mikemorris (approved):
Proposed two alternative suggestions for GAMMA examples, but generally supportive of this with the change to
> Services SHOULD always be interpreted as references to endpoints within the same cluster

/lgtm

geps/gep-1748.md
robscott: Updated, thanks!

examples/standard/multicluster/httproute-gamma.yaml
robscott: Updated, thanks! Opted to leave out the additional example since this is about to merge, and don't want to increase scope for others that have already approved. Probably worth a follow up though.

shaneutt (approved):
/lgtm

PR #1853, opened 2023-03-20
conformance: check HTTPRoute connectivity after deleting relevant ReferenceGrant

### LINKED ISSUE #1838: conformance: new test checking a standard condition on a Route that cannot route due to lacking `ReferenceGrant`

When an `HTTPRoute` (or route in general but only HTTPRoute is supported now, right?) lacks a `ReferenceGrant` in order to route traffic to designated `backendRef`s the status condition(s) should reflect that in a standard way. 

For instance: a condition with `type: ResolvedRefs`, `reason: RefNotPermitted` and `status` set to `"False"`.

**Why this is needed**:

To standardize route's statuses when rejected due to missing `ReferenceGrant`.

https://gateway-api.sigs.k8s.io/api-types/referencegrant/ doesn't mention this. On purpose?

pmalek:
**What type of PR is this?**

/kind test
/area conformance

Fixes #1838

```release-note
HTTPRoute connectivity is in now enforced in conformance tests if a relevant ReferenceGrant gets deleted.
```

### DIFF
--- a/conformance/tests/httproute-reference-grant.go
+++ b/conformance/tests/httproute-reference-grant.go
@@ -17,10 +17,14 @@ limitations under the License.
 package tests
 
 import (
+	"context"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance/utils/http"
 	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
 	"sigs.k8s.io/gateway-api/conformance/utils/suite"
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
This seems _correct_ ultimately, but I do want some feedback from other implementers before we merge so let's hold. I think the 503 is particularly going to draw some questions.

/hold

sunjayBhatia:
> That might have been an incorrect assumption on my side which just felt "right".
> 
> I believe this should indeed be specified just as you mentioned. What would be the best course of action for this then? Should this change include that mention or does it have to go through a different process where maintainers comment on this idea?

Not a maintainer, just a reviewer, but I am ok with making an API change along with the relevant test change as I think that keeps the justification for the test change bundled nicely. However, you may want to open an issue about the API spec change before making it just to make sure there's enough signal for people to discuss such a change

pmalek:
@sunjayBhatia Got it. Created  #1902 to track it.

shaneutt:
@pmalek I think to move forward with this one, given some time has passed we need to make the appropriate spec change as noted in #1902 and signal that one as in progress. Let us know if you need any assistance!

youngnick:
I just commented on #1902, but it's actually really important for security to ensure that the result code is the same if the service doesn't exist or if there's no ReferenceGrant so that you can't use that difference to enumerate service names.

pmalek:
@youngnick As per the discussion in #1902 I've updated the PR to use HTTP 500. PTAL

Issue #1842, opened 2023-03-17
Clarify Conformance Expectations for Gateway Ports and Listeners


robscott:
**What would you like to be added**:
As described in https://github.com/kubernetes-sigs/gateway-api/pull/1827#issuecomment-1474273702, it would be helpful to describe what we expect from implementations in terms of supported combinations of ports, protocols, and listeners. It seems like the minimum bar is that L7 implementations can handle one HTTP listener on 80 and one HTTPS listener on 443. 

What is less clear is what happens beyond that:

1) Is support for additional ports considered "Extended" conformance?
2) Is support for multiple listeners on the same port considered "Extended" as well?
3) What about non-standard combinations like "HTTPS" on 80 or "HTTP" on 443?

**Why this is needed**:
There's some variation among implementations in terms of what's supported today, it would be good to ensure expectations are clearly stated for conformance.

dprotaso:
> Is support for additional ports considered "Extended" conformance?

It would be nice if this was "Core" and didn't require unique manual configuration of each implementation to support.

> Is support for multiple listeners on the same port considered "Extended" as well?

A use case here is how easy it is to manage multiple certificates on a gateway. Adding multiple listeners with the same port but different `certificateRefs` could be a valid option.

An alternative that scales better in K8s clusters would be  https://github.com/kubernetes-sigs/gateway-api/issues/1713

> What about non-standard combinations like "HTTPS" on 80 or "HTTP" on 443?

Is there a list of standard combinations?

dprotaso:
I created a PR that includes a conformance tests for dynamic listener ports

dprotaso:
> Is support for multiple listeners on the same port considered "Extended" as well?

Note: the TLS examples on the website here show multiple listeners using the same port
https://gateway-api.sigs.k8s.io/guides/tls/#wildcard-tls-listeners

robscott:
>sharing a Port and Protocol is okay if the TLS details are distinct. I don't know if this is the common view however.

@youngnick To clarify, are you including "Hostname" in the general category of "TLS Details"? So far we've said that the combination of "Hostname", "Protocol", and "Port" must be unique for each listener. I think that would cover your use case if I'm understanding it correctly.

dprotaso:
FYI with #2065 merging we now have

[...]

Issue #1713, opened 2023-02-08
GEP: Standard Mechanism to Merge Multiple Gateways 


dprotaso:
**What would you like to be added**:

Support for merging Gateways should have standard & documented mechanic.

**Why this is needed**:

> Prior Discussions: https://github.com/kubernetes-sigs/gateway-api/discussions/1248, https://github.com/kubernetes-sigs/gateway-api/discussions/1246

Knative generates on demand per-service certificates using HTTP-01 challenges. There can be O(1000) Knative Services in the cluster which means we have O(1000) distinct certificates. The Gateway Resource is a contention point since is the only place to attach listeners to gateways with certificates.

The spec currently has language to indicate implemenations `MAY` merge Gateways resources but the mechanic isn't defined. 
https://github.com/kubernetes-sigs/gateway-api/blob/541e9fc2b3c2f62915cb58dc0ee5e43e4096b3e2/apis/v1beta1/gateway_types.go#L76-L78

Secondly, given similar use-cases/needs for control planes to manage listeners it might make sense that `merging` Gateways is part of `extended` conformance.

arkodg:
+1 to this proposal. 

* This has been asked by users of Envoy Gateway to make their runtime deployments more efficient (currently a  `Gateway` resource maps to a unique Data Plane (Envoy Proxy) deployment)

* The spec says compatible listeners can be collapsed, but since `Gateways` can be applied dynamically  
it is bound to cause traffic disruption in the data plane if the implementation collapses or separates listener implementations in runtime.

* If there is way to add a knob which sets constraints for the L3/L4 fields of the `Listener` such as ensuring a port must be unique across all `Listeners` within all `Gateways` within a `GatewayClass`, statically collapsing listeners in implementations should be doable. 
Below is the current spec
https://github.com/kubernetes-sigs/gateway-api/blob/0177808c57566502f8fb697abd6d42c06008ff6a/apis/v1beta1/gateway_types.go#L72

howardjohn:
FWIW how we do this in Istio is https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment. Basically the same way to apply 1 Gateway to a manually deployed Deployment/Service instead of automatically deployed one can be used to apply N Gateways and merge them.

If they want automated deployment and merging, they would make one "root" Gateway and attach all others. Actually that is another use case for https://github.com/kubernetes-sigs/gateway-api/pull/1596

[...]

PR #1854, opened 2023-03-20
Ensure listener name must be unique within a Gateway

### LINKED ISSUE #1845: Listener name must be unique within a Gateway
**What happened**:

According to https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Listener

> Name is the name of the Listener. This name MUST be unique within a Gateway.

However, it seems gateway-api doesn't check the name currently.

The listener with the duplicate name should be rejected.

A configuration like this can be incorrectly accepted:
```
  kind: Gateway
  metadata:
    namespace: envoy-gateway
    name: gateway-1
  spec:
    gatewayClassName: envoy-gateway-class
    listeners:
    - name: http-1
      protocol: HTTP
      port: 8080
    - name: http-1
      protocol: HTTP
      port: 80

```

The issue is originally raised at https://github.com/envoyproxy/gateway/issues/1180 and I have provided an initial fix in https://github.com/envoyproxy/gateway/pull/1181. I could rewrite the fix to adapt gateway-api's code if the report is accepted.

muyuan0:
Signed-off-by: muyuan0 <127020730+muyuan0@users.noreply.github.com>

**What type of PR is this?**

/kind bug

Check if the listener names are unique within a Gateway

Fixes #1845

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -69,6 +69,7 @@ func validateGatewayListeners(listeners []gatewayv1b1.Listener, path *field.Path
 	errs = append(errs, ValidateListenerTLSConfig(listeners, path)...)
 	errs = append(errs, validateListenerHostname(listeners, path)...)
 	errs = append(errs, ValidateTLSCertificateRefs(listeners, path)...)
+	errs = append(errs, ValidateListenerNames(listeners, path)...)
 	return errs
 }
 
@@ -115,6 +116,20 @@ func ValidateTLSCertificateRefs(listeners []gatewayv1b1.Listener, path *field.Pa
 				errs = append(errs, field.Forbidden(path.Index(i).Child("tls").Child("certificateRefs"), fmt.Sprintln("should be set and not empty when TLSModeType is Terminate")))
 			}
 		}
+	}
+	return errs
+}
+
+// ValidateListenerNames validates the names of the listeners
+// must be unique within the Gateway
+func ValidateListenerNames(listeners []gatewayv1b1.Listener, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	nameMap := map[gatewayv1b1.SectionName]struct{}{}
+	for i, c := range listeners {
+		if _, found := nameMap[c.Name]; found {
+			errs = append(errs, field.Forbidden(path.Index(i).Child("name"), fmt.Sprintln("must be unique within the Gateway")))
+		}
+		nameMap[c.Name] = struct{}{}
 	}
 	return errs
 }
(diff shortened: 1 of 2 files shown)


apis/v1beta1/validation/gateway.go
```diff
@@ -118,3 +119,17 @@ func ValidateTLSCertificateRefs(listeners []gatewayv1b1.Listener, path *field.Pa
 	}
 	return errs
 }
+
+// ValidateListenerNames validates the names of the listeners
+// must be unique within the Gateway
+func ValidateListenerNames(listeners []gatewayv1b1.Listener, path *field.Path) field.ErrorList {
```
danehans, line 125: This func should be called by `validateGatewayListeners()` to be included in the admission webhook.

shaneutt:
> I think we should sort out some of the discussions in the following related issues prior to merging this change
> 
> #1842 #1713

So just to be clear you're asking that this PR be put on _hold_ until both #1842 and #1713 are resolved?

dprotaso:
Yeah.

apis/v1beta1/validation/gateway.go
muyuan0: Added in the `validateGatewayListeners` above

shaneutt:
> Yeah.

I'm not feeling very compelled to block here, it seems unique listener name checking is something we would want in any case, no?

hzxuzhonghu (approved):

robscott:
@dprotaso I'm not sure I see the connection between this and the issues you linked, can you help clarify why they should block this?

dprotaso:
Yeah #1842 doesn't apply to `Name` (`SectionName`) - my mistake

For https://github.com/kubernetes-sigs/gateway-api/issues/1713 I wondered if we're merging Gateways would we allow listeners on different Gateways to merge with the same `Name`.  A concern I had was that the validation here would prevent that but I might have been overzealous. We can have that discussion in the other GEP

shaneutt (approved):
/lgtm

shaneutt:
@muyuan0 just need to change your commit message as per https://github.com/kubernetes-sigs/gateway-api/pull/1854#issuecomment-1475899793 and then we're all set :+1:

muyuan0:
@shaneutt 
Could I force push to update the commit message?

shaneutt:
> @shaneutt Could I force push to update the commit message?

Yes that's fine! Thank you!

robscott:
Thanks @muyuan0! This LGTM as well once the commit message is fixed.

robscott:
Thanks @muyuan0!

/lgtm

PR #1855, opened 2023-03-20
feat: add support for path matching precedence

### LINKED ISSUE #1770: Corner case in HTTPRouteRule match order when RegularExpression is used

We need to add test cases to ensure the RegularExpression match happens after the other matches.

**Why this is needed**:

According to the HTTPRouteRule definition:
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule

> Proxy or Load Balancer routing configuration generated from HTTPRoutes MUST prioritize matches based on the following criteria, continuing on ties. Across all rules specified on applicable Routes, precedence must be given to the match with the largest number of:
>
>    Characters in a matching path.
>    Header matches.
>    Query param matches.

The routes are sorted by the "Characters in a matching path." first.

However, when the RegularExpression in https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.PathMatchType is involved, the order is ambiguous as the regex can be very specific.

For example, a regex `/a/(.*)+` has more characters in a matching path than exact match `/a/b`, while it is similar to the path prefix `/a/`. If we strictly sort the route by the path, `/a/(.*)+` will be matched before `/a/b`.

An idea to sort this problem is to ensure the regex match happens after the other matches.

Xunzhuo:
**What type of PR is this?**

/kind bug
/area conformance

Path matches in following orders:
1. Exact
2. Prefix

Fixes #1770

```release-note
* Updated spec to clarify that Exact matches have precedence over Prefix matches and RegularExpression matches have implementation specific precedence.
* Added conformance test to verify that path matching precedence is implemented.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -156,9 +156,12 @@ type HTTPRouteRule struct {
 	// ties. Across all rules specified on applicable Routes, precedence must be
 	// given to the match with the largest number of:
 	//
-	// * Characters in a matching path.
+	// * Characters in a matching "Exact" path match
+	// * Characters in a matching "Prefix" path match
 	// * Header matches.
 	// * Query param matches.
+	//
+	// Note: The precedence of RegularExpression path matches are implementation-specific.
 	//
 	// If ties still exist across multiple Routes, matching precedence MUST be
 	// determined in order of the following criteria, continuing on ties:
--- a/conformance/tests/httproute-path-match-order.go
+++ b/conformance/tests/httproute-path-match-order.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRoutePathMatchOrder)
+}
+
+var HTTPRoutePathMatchOrder = suite.ConformanceTest{
+	ShortName:   "HTTPRoutePathMatchOrder",
+	Description: "An HTTPRoute where there are multiple matches routing to any given backend follows match order precedence",
+	Manifests:   []string{"tests/httproute-path-match-order.yaml"},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+		routeNN := types.NamespacedName{Namespace: ns, Name: "path-matching-order"}
+		gwNN := types.NamespacedName{Namespace: ns, Name: "same-namespace"}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		testCases := []http.ExpectedResponse{{
+			Request:   http.Request{Path: "/match/exact/one"},
+			Backend:   "infra-backend-v3",
+			Namespace: ns,
+		}, {
+			Request:   http.Request{Path: "/match/exact"},
+			Backend:   "infra-backend-v2",
+			Namespace: ns,
+		}, {
+			Request:   http.Request{Path: "/match"},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request:   http.Request{Path: "/match/prefix/one/any"},
+			Backend:   "infra-backend-v2",
+			Namespace: ns,
+		}, {
+			Request:   http.Request{Path: "/match/prefix/any"},
+			Backend:   "infra-backend-v1",
+			Namespace: ns,
+		}, {
+			Request:   http.Request{Path: "/match/any"},
+			Backend:   "infra-backend-v3",
+			Namespace: ns,
+		},
+		}
+
+		for i := range testCases {
+			tc := testCases[i]
+			t.Run(tc.GetTestCaseName(i), func(t *testing.T) {
+				t.Parallel()
+				http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, tc)
+			})
+		}
+	},
+}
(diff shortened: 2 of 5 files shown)


conformance/tests/httproute-path-match-order.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - matches:
+    - path:
+        type: Exact
+        value: /a/b
+    backendRefs:
+    - name: infra-backend-v1
+      port: 8080
+  - matches:
+    - path:
+        type: RegularExpression
```
sunjayBhatia, line 19: regex match is implementation-specific support so we probably can't use that here: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.PathMatchType

apis/v1beta1/httproute_types.go
```diff
@@ -156,7 +156,9 @@ type HTTPRouteRule struct {
 	// ties. Across all rules specified on applicable Routes, precedence must be
 	// given to the match with the largest number of:
 	//
-	// * Characters in a matching path.
+	// * Characters in a matching "Exact" path match
+	// * Characters in a matching "RegularExpression" path match
```
howardjohn, line 160: Regex between Exact and Prefix feels a bit odd...

TBH, using Characters at all in regex feels odd entirely. Characters has no meaning in regex. is there any proxy implementation that does this today? I worry we will see users inserting dummy characters

sunjayBhatia: contour uses regex between exact and prefix because a regex "could be" more specific than a prefix match, not a perfect heuristic but yeah

also yeah we do sort our regex matches on length just to enable consistent ordering, again not a perfect heuristic

robscott: Chatting with @briantkennedy about this and he pointed out that this would likely not be compatible with NGINX where RegEx matches always happen last. It's probably best to require "Exact", "Prefix", then "RegEx". @brianehlert @pleshakov are probably best to confirm from NGINX side though.

conformance/tests/httproute-path-match-order.yaml
Xunzhuo: Thanks

apis/v1beta1/httproute_types.go
Xunzhuo: I am not quite sure for putting the order of regex to where, since it is different in implementations. I just added conformance to test:
```
Characters in a matching "Exact" path match
Characters in a matching "Prefix" path match
```

conformance/tests/httproute-path-match-order.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRoutePathMatchOrders)
+}
+
+var HTTPRoutePathMatchOrders = suite.ConformanceTest{
+	ShortName:   "HTTPRoutePathMatchOrders",
+	Description: "A HTTPRoute with different path matches for different backends",
+	Manifests:   []string{"tests/httproute-path-match-orders.yaml"},
```
sunjayBhatia, line 36: typo here in the test yaml file name

sunjayBhatia (changes requested):

conformance/tests/httproute-path-match-order.go
sunjayBhatia: would suggest making sure "Orders/orders" -> "Order/order"

apis/v1beta1/httproute_types.go
robscott: I don't think we're going to be able to mandate any specific precedence for RegEx path matches in the API. Instead, we probably need to label that as implementation-specific. Looking at the [NGINX reference docs], it does not look like it would be possible for it to be conformant with any precedence based on the length of a regex.

>To find location matching a given request, nginx first checks locations defined using the prefix strings (prefix locations). Among them, the location with the longest matching prefix is selected and remembered. Then regular expressions are checked, in the order of their appearance in the configuration file. The search of regular expressions terminates on the first match, and the corresponding configuration is used. If no match with a regular expression is found then the configuration of the prefix location remembered earlier is used.

Since we already consider RegEx path matching "Implementation-Specific" we likely should also consider the precedence of RegEx path matches "Implementation-Specific".

I'd recommend removing the "RegEx" mention from the list and adding a note below the list that states that "The precedence of RegularExpression path matches are implementation-specific."

Xunzhuo: Thanks @robscott

conformance/tests/httproute-path-match-order.go
Xunzhuo: Thanks @sunjayBhatia !

sunjayBhatia (approved):
lgtm, found a little bug in contour as part of running this https://github.com/projectcontour/contour/pull/5217

nice!

sunjayBhatia (approved):

apis/v1beta1/httproute_types.go
```diff
@@ -156,10 +156,13 @@ type HTTPRouteRule struct {
 	// ties. Across all rules specified on applicable Routes, precedence must be
 	// given to the match with the largest number of:
 	//
-	// * Characters in a matching path.
+	// * Characters in a matching "Exact" path match
+	// * Characters in a matching "Prefix" path match
 	// * Header matches.
 	// * Query param matches.
 	//
+	// Note: The precedence of RegularExpression path matches are implementation-specific.
```
shaneutt, line 164: I agree with the decision to make precedence for RE implementation-specific, as I think we would likely have created a conundrum for several implementations.

conformance/tests/httproute-path-match-order.go
```diff
@@ -0,0 +1,79 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRoutePathMatchOrder)
+}
+
+var HTTPRoutePathMatchOrder = suite.ConformanceTest{
+	ShortName:   "HTTPRoutePathMatchOrder",
+	Description: "A HTTPRoute with different path matches for different backends",
```
shaneutt, line 35: Am i understanding the intention of this test to be something more like:

```suggestion
	Description: "An HTTPRoute can route multiple separate path matches to a single backend",
```

?

shaneutt: Made an edit above, lmkwyt

conformance/tests/httproute-path-match-order.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

conformance/tests/httproute-path-match-order.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - path:
+        type: PathPrefix
+        value: /match/prefix/
+    backendRefs:
+    - name: infra-backend-v1
+      port: 8080
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /match/prefix/one/
```
robscott, line 48: It may be helpful to have at least one of these prefix matches without a trailing slash.

robscott (commented):
This is really great, thanks @Xunzhuo! A couple tiny nits but otherwise LGTM.

conformance/tests/httproute-path-match-order.go
Xunzhuo: Thanks @shaneutt ans @sunjayBhatia !

conformance/tests/httproute-path-match-order.yaml
Xunzhuo: Done Thanks!

robscott:
Thanks @Xunzhuo! Will wait for someone else to LGTM in case I missed anything in the conformance tests.

/cc @arkodg @mlavacca 
/approve

arkodg:
hey @Xunzhuo have you tested this test against any implementation ?

Xunzhuo:
@sunjayBhatia tested it in Contour, I will have a test in EG.

arkodg (approved):

PR #1858, opened 2023-03-21
conformance: don't require k8s version v1.21

michaelbeaumont:
**What type of PR is this?**
/kind test
/area conformance

Makes it possible to pass the tests on k8s before `v1.21`, which is the first version the `kubernetes.io/metadata.name` label is set automatically. [Might be worth noting that use of arbitrary labels like this is somewhat unsafe](https://gateway-api.sigs.k8s.io/concepts/security-model/#risks-of-other-labels), but it serves to test implementations nevertheless.

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-with-attached-routes.yaml
+++ b/conformance/tests/gateway-with-attached-routes.yaml
@@ -16,9 +16,7 @@ spec:
         from: Selector
         selector:
           matchLabels:
-            # This label is added automatically as of K8s 1.22
-            # to all namespaces
-            kubernetes.io/metadata.name: gateway-conformance-infra
+            gateway-conformance: infra
 ---
 apiVersion: gateway.networking.k8s.io/v1beta1
 kind: HTTPRoute
@@ -53,9 +51,7 @@ spec:
         from: Selector
         selector:
           matchLabels:
-            # This label is added automatically as of K8s 1.22
-            # to all namespaces
-            kubernetes.io/metadata.name: gateway-conformance-infra
+            gateway-conformance: infra
 ---
 apiVersion: gateway.networking.k8s.io/v1beta1
 kind: HTTPRoute


dprotaso:
1.21 is pretty out of date - it was EOL almost a year ago 2022-06-28

https://kubernetes.io/releases/patch-releases/#non-active-branch-history

michaelbeaumont:
> 1.21 is pretty out of date - it was EOL almost a year ago 2022-06-28
> 
> https://kubernetes.io/releases/patch-releases/#non-active-branch-history

I won't disagree that it's out of date but [the FAQs](https://gateway-api.sigs.k8s.io/faq/) explicitly mention 1.16/1.19 and I don't think keeping this test as is a good reason to break that compatibility. It seems unintentional here more than anything else.

robscott:
Yeah we definitely need some docs updates here, thanks for catching this! The desired state is defined in our versioning guidelines within the "Supported Versions" section: https://gateway-api.sigs.k8s.io/concepts/versioning/#supported-versions.

In there we commit to supporting "a minimum of the most recent 5 Kubernetes minor versions". I'd interpret that as supporting the following versions today:

* 1.22
* 1.23
* 1.24
* 1.25
* 1.26

[Kubernetes 1.27 is going to be released in a few weeks](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.27) which would result in 1.22 dropping off the list of versions we're required to support. I've filed https://github.com/kubernetes-sigs/gateway-api/issues/1862 to help make our docs around this more consistent.

michaelbeaumont:
> Yeah we definitely need some docs updates here, thanks for catching this! The desired state is defined in our versioning guidelines within the "Supported Versions" section: https://gateway-api.sigs.k8s.io/concepts/versioning/#supported-versions.
> 
> In there we commit to supporting "a minimum of the most recent 5 Kubernetes minor versions". I'd interpret that as supporting the following versions today:
> 
> * 1.22
> * 1.23
> * 1.24
> * 1.25
> * 1.26
> 
> [Kubernetes 1.27 is going to be released in a few weeks](https://github.com/kubernetes/sig-release/tree/master/releases/release-1.27) which would result in 1.22 dropping off the list of versions we're required to support. I've filed #1862 to help make our docs around this more consistent.

Nice, yeah I couldn't find that in the docs either for some reason. It feels a bit of a shame to break compatibility for this but I'll close this PR

PR #1859, opened 2023-03-21
[conformance] Gateway Dynamic Listener Ports

dprotaso:
**What type of PR is this?**
/area conformance
/kind feature

This adds a conformance tests where  listeners with dynamic ports are added

Part of https://github.com/kubernetes-sigs/gateway-api/issues/1842

```release-note

```

### DIFF
--- a/conformance/tests/gateway-dynamic-listeners.go
+++ b/conformance/tests/gateway-dynamic-listeners.go
@@ -168,8 +168,9 @@ var GatewayListenerHTTPRouteDynamicPorts = suite.ConformanceTest{
 				addr := net.JoinHostPort(host, strconv.Itoa(int(listener.Port)))
 
 				if listener.TLS != nil {
-					expectedResponse.Request.Host = string(*listener.Hostname)
-					tls.MakeTLSRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, s.TimeoutConfig, addr, certBytes, keyBytes, host, expectedResponse)
+					listenerHost := string(*listener.Hostname)
+					expectedResponse.Request.Host = listenerHost
+					tls.MakeTLSRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, s.TimeoutConfig, addr, certBytes, keyBytes, listenerHost, expectedResponse)
 				} else {
 					http.MakeRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, s.TimeoutConfig, addr, expectedResponse)
 				}
@@ -177,86 +178,74 @@ var GatewayListenerHTTPRouteDynamicPorts = suite.ConformanceTest{
 		}
 
 		original := &v1beta1.Gateway{}
-		mutate := &v1beta1.Gateway{}
 
-		t.Run("should be able to add multiple HTTP listeners with dynamic ports that then become available for routing traffic", func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), s.TimeoutConfig.GetTimeout)
-			defer cancel()
+		t.Log("should be able to add multiple HTTP listeners with dynamic ports that then become available for routing traffic")
+		ctx, cancel := context.WithTimeout(context.Background(), s.TimeoutConfig.GetTimeout)
+		defer cancel()
 
-			err = s.Client.Get(ctx, gwNN, original)
-			require.NoErrorf(t, err, "error getting Gateway: %v", err)
+		err = s.Client.Get(ctx, gwNN, original)
+		require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
-			// verify that the implementation is tracking the most recent resource changes
-			kubernetes.GatewayMustHaveLatestConditions(t, s.TimeoutConfig, original)
-			mutate = original.DeepCopy()
-			mutate.Spec.Listeners = append(mutate.Spec.Listeners, listeners...)
+		// verify that the implementation is tracking the most recent resource changes
+		kubernetes.GatewayMustHaveLatestConditions(t, s.TimeoutConfig, original)
+		mutate := original.DeepCopy()
+		mutate.Spec.Listeners = append(mutate.Spec.Listeners, listeners...)
 
-			err = s.Client.Patch(ctx, mutate, client.MergeFrom(original))
-			require.NoErrorf(t, err, "error patching the Gateway: %v", err)
+		err = s.Client.Patch(ctx, mutate, client.MergeFrom(original))
+		require.NoErrorf(t, err, "error patching the Gateway: %v", err)
+		kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, expectedListeners)
 
-			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, expectedListeners)
+		successResponse := http.ExpectedResponse{
+			Namespace: gwNN.Namespace,
+			Request:   http.Request{Path: "/"},
+			Response:  http.Response{StatusCode: 200},
+		}
 
-			expectedResponse := http.ExpectedResponse{
-				Namespace: gwNN.Namespace,
-				Request:   http.Request{Path: "/"},
-				Response:  http.Response{StatusCode: 200},
-			}
+		sendRequestToEachListener(t, successResponse, mutate.Spec.Listeners)
 
-			sendRequestToEachListener(t, expectedResponse, mutate.Spec.Listeners)
-		})
+		t.Log("should be able to remove multiple HTTP listeners with dynamic ports")
+		ctx, cancel = context.WithTimeout(context.Background(), s.TimeoutConfig.GetTimeout)
+		defer cancel()
 
-		t.Run("should be able to remove multiple HTTP listeners with dynamic ports", func(t *testing.T) {
-			ctx, cancel := context.WithTimeout(context.Background(), s.TimeoutConfig.GetTimeout)
-			defer cancel()
+		err = s.Client.Get(ctx, gwNN, mutate)
+		require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
-			err = s.Client.Patch(ctx, original, client.MergeFrom(mutate))
-			require.NoErrorf(t, err, "error patching the Gateway: %v", err)
+		mutate.Spec.Listeners = original.Spec.Listeners
+		err = s.Client.Update(ctx, mutate)
+		require.NoErrorf(t, err, "error patching the Gateway: %v", err)
 
-			expectedListeners = []v1beta1.ListenerStatus{{
-				Name: "http",
-				SupportedKinds: []v1beta1.RouteGroupKind{{
-					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
-					Kind:  v1beta1.Kind("HTTPRoute"),
-				}},
-				Conditions:     expectedConditions,
-				AttachedRoutes: 1,
-			}}
+		expectedListeners = []v1beta1.ListenerStatus{{
+			Name: "http",
+			SupportedKinds: []v1beta1.RouteGroupKind{{
+				Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+				Kind:  v1beta1.Kind("HTTPRoute"),
+			}},
+			Conditions:     expectedConditions,
+			AttachedRoutes: 1,
+		}}
 
-			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, expectedListeners)
+		kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, expectedListeners)
 
-			// Original listener should work
-			expectedResponse := http.ExpectedResponse{
-				Namespace: gwNN.Namespace,
-				Request:   http.Request{Path: "/"},
-				Response:  http.Response{StatusCode: 200},
-			}
+		// Original listener should work
+		sendRequestToEachListener(t, successResponse, original.Spec.Listeners)
 
-			sendRequestToEachListener(t, expectedResponse, original.Spec.Listeners)
+		for _, listener := range listeners {
+			addr := net.JoinHostPort(host, strconv.Itoa(int(listener.Port)))
 
-			// Remove the first listener
-			listeners := mutate.Spec.Listeners[1:]
-
-			for _, listener := range listeners {
-				addr := net.JoinHostPort(host, strconv.Itoa(int(listener.Port)))
-
-				// Listeners that were removed should stop working
-				dial := func(elapsed time.Duration) bool {
-					conn, err := net.Dial("tcp", addr)
-
-					if conn != nil {
-						conn.Close()
-						return false
-					}
-
-					if err != nil && strings.Contains(err.Error(), "connection refused") {
-						return true
-					}
+			// Listeners that were removed should stop working
+			dial := func(elapsed time.Duration) bool {
+				conn, err := net.DialTimeout("tcp", addr, time.Second)
+				if conn != nil {
+					conn.Close()
 					return false
 				}
-
-				http.AwaitConvergence(t, s.TimeoutConfig.RequiredConsecutiveSuccesses, s.TimeoutConfig.MaxTimeToConsistency, dial)
+				if err != nil && strings.Contains(err.Error(), "connection refused") {
+					return true
+				}
+				return false
 			}
-		})
+			http.AwaitConvergence(t, s.TimeoutConfig.RequiredConsecutiveSuccesses, s.TimeoutConfig.MaxTimeToConsistency, dial)
+		}
 	},
 }
 
--- a/conformance/tests/gateway-dynamic-listeners.go
+++ b/conformance/tests/gateway-dynamic-listeners.go
@@ -179,11 +179,6 @@ var GatewayListenerHTTPRouteDynamicPorts = suite.ConformanceTest{
 			require.NoErrorf(t, err, "timed out waiting for Gateway address to be assigned")
 
 			for _, listener := range mutate.Spec.Listeners {
-				timeoutConfig := s.TimeoutConfig
-				// TODO - It takes much longer for listeners to be consistent
-				if timeoutConfig.MaxTimeToConsistency < timeoutConfig.GatewayStatusMustHaveListeners {
-					timeoutConfig.MaxTimeToConsistency = timeoutConfig.GatewayStatusMustHaveListeners
-				}
 				host, _, err := net.SplitHostPort(gwAddr)
 				require.NoErrorf(t, err, "unable to split gateway address %q", gwAddr)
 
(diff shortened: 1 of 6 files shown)


robscott:
Hey @dprotaso, thanks for the PR! I'm realizing my issue wasn't very clear, but the goal of that issue was to actually first ensure we'd reached consensus on requirements and added them to the API spec before we added conformance tests to exercise this. I unfortunately did not communicate that very well, sorry!

In general, I think we need to be careful anytime we're adding conformance tests that require behavior that isn't explicitly required by the spec. I think it's always much safer to agree on the requirements we want via API Spec updates and then to add conformance tests once we've reached an agreement.

/hold

dprotaso:
@robscott yeah no worries - I just forgot to mark this as a draft - this was done for the purposes of the discussion.

Also FWIW this test passes on Istio (cc @howardjohn) 🎉

dprotaso:
I marked this PR ready for review since this test is behind a flag

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,208 @@
...
+					http.MakeRequestAndExpectEventuallyConsistentResponse(t, s.RoundTripper, timeoutConfig, addr, expectedResponse)
+				}
+
+			}
+		})
+	},
+}
+
+func nextPort(start, end int, ports sets.Set[int]) int {
+	port := start + rand.Intn(end-start) //nolint:gosec
```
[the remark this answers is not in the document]
dprotaso: > then that isn't actually testing that a user could use a random port

I think I'd rather make this port range configurable.

Long term I wonder if it makes sense to list valid port ranges in a `status` block somewhere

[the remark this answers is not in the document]
dprotaso: port range is configurable now

shaneutt (dismissed):

conformance/utils/config/listener.go
```diff
@@ -0,0 +1,45 @@
...
+func SetupListenerConfig(c *ListenerConfig) {
+	defaultConfig := DefaultListenerConfig()
+
+	if c.DynamicPortRange.Start == 0 {
+		c.DynamicPortRange.Start = defaultConfig.DynamicPortRange.Start
+	}
+	if c.DynamicPortRange.End == 0 {
+		c.DynamicPortRange.End = defaultConfig.DynamicPortRange.End
+	}
+}
```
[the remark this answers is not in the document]
dprotaso: done!

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,209 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/tls"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayListenerDynamicPorts)
+}
+
+var GatewayListenerDynamicPorts = suite.ConformanceTest{
+	ShortName:   "GatewayListenerDynamicPorts",
+	Features:    []suite.SupportedFeature{suite.SupportGatewayListenerDynamicPorts},
```
mlavacca, line 47: I think we should add the `SupportGateway` feature as well here

conformance/utils/config/listener.go
```diff
@@ -0,0 +1,54 @@
...
+// DynamicPortRange specifies the starting and ending port of a
+// range
+type DynamicPortRange struct {
+	Start int
+	End   int
+}
+
+// ListenerConfig allow conformance test runners to configure
+type ListenerConfig struct {
+	DynamicPortRange DynamicPortRange
```
mlavacca, line 28: What do you think about setting this field inline to simplify customization via options?

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,209 @@
...
+			listeners  = make([]v1beta1.Listener, 0, portCount)
+			same       = v1beta1.NamespacesFromSame
+
+			expectedListeners = []v1beta1.ListenerStatus{{
+				Name: "http",
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
+				}},
+				Conditions: []metav1.Condition{{
```
mlavacca, line 74: I think we should check the `ResolvedRefs` condition for all the listeners we already check the accepted one (especially for the HTTPS ones).

conformance/utils/config/listener.go
dprotaso: I think I prefer the verbosity since there could be another `*PortRange` type for some other purpose.

conformance/tests/gateway-dynamic-listeners.go
dprotaso: >  all the listeners we already check the accepted one

What's the value asserting these on non-tls listeners?

robscott:
@dprotaso before we go too far with this conformance test can you actually propose an update to the spec? We absolutely can't let a conformance test merge unless it's based on something in the API spec.

dprotaso:
@robscott sound's good - done here https://github.com/kubernetes-sigs/gateway-api/pull/2065

conformance/tests/gateway-dynamic-listeners.go
dprotaso: done.

dprotaso:
@robscott #2065 has merged - PTAL

conformance/utils/suite/features.go
```diff
@@ -64,6 +64,9 @@ const (
 	//
 	//       See: https://github.com/kubernetes-sigs/gateway-api/issues/1780
 	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
+
+	// This option indicates support for Gateway to support dynamic ports in their listeners
+	SupportGatewayListenerDynamicPorts SupportedFeature = "GatewayListenerDynamicPorts"
```
robscott, line 69: Thinking about this in terms of the future when a GatewayClass publishes supported features in status, this particular feature name doesn't really provide much information. IE an implementation could potentially claim this if they supported ports 80, 443, and 8080 if they configured only those ports when running this test. That could be a completely separate set of ports compared to another implementations that also claimed to support this feature.

I'm not sure what the solution is here, just concerned that this doesn't really provide a good way for implementations to actually signal what they support.

/cc @mlavacca @shaneutt

conformance/utils/config/listener.go
```diff
@@ -0,0 +1,54 @@
...
+// SetupListenerConfig will apply defaults to the passed [ListenerConfig]
+func SetupListenerConfig(c *ListenerConfig) {
+	defaultConfig := DefaultListenerConfig()
+
+	if c.DynamicPortRange.Start == 0 {
+		c.DynamicPortRange.Start = defaultConfig.DynamicPortRange.Start
+	}
+	if c.DynamicPortRange.End == 0 {
+		c.DynamicPortRange.End = defaultConfig.DynamicPortRange.End
+	}
```
robscott, line 53: What about implementations that may want to run this test with non-contiguous port ranges? Does this test require that all protocols are supported on all ports? If an implementation only supports HTTPS on 443, can/should we represent that with config here?

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,223 @@
...
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayListenerDynamicPorts)
+}
+
+var GatewayListenerDynamicPorts = suite.ConformanceTest{
+	ShortName: "GatewayListenerDynamicPorts",
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportGatewayListenerDynamicPorts,
```
robscott, line 49: Do we need "supported features" or similar for protocols? This specific test would fail with a L4 Gateway implementation which I don't think we want. Maybe there's a way to make the protocol(s) used in this test configurable to avoid this?

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,223 @@
...
+				AllowedRoutes: &v1beta1.AllowedRoutes{
+					Namespaces: &v1beta1.RouteNamespaces{From: &same},
+				},
+			})
+
+			expectedListeners = append(expectedListeners, v1beta1.ListenerStatus{
+				Name: v1beta1.SectionName(strconv.Itoa(port)),
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
```
robscott, line 109: I don't think we want to assume that this is the only Route type that an implementation may support with the "HTTP" protocol.

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,223 @@
...
+						Name: v1beta1.ObjectName(certNN.Name),
+					}},
+				},
+			})
+
+			expectedListeners = append(expectedListeners, v1beta1.ListenerStatus{
+				Name: v1beta1.SectionName(strconv.Itoa(port)),
+				SupportedKinds: []v1beta1.RouteGroupKind{{
+					Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+					Kind:  v1beta1.Kind("HTTPRoute"),
```
robscott, line 143: Same comment as above.

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,223 @@
...
+			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, expectedListeners)
+
+			gwAddr, err := kubernetes.WaitForGatewayAddress(t, s.Client, s.TimeoutConfig, gwNN)
+			require.NoErrorf(t, err, "timed out waiting for Gateway address to be assigned")
+
+			for _, listener := range mutate.Spec.Listeners {
+				timeoutConfig := s.TimeoutConfig
+				if timeoutConfig.MaxTimeToConsistency < 2*time.Minute {
+					timeoutConfig.MaxTimeToConsistency = 2 * time.Minute
+				}
```
robscott, line 189: This feels problematic - why are we overriding this manually here?

robscott (commented):
Thanks @dprotaso! Appreciate you clarifying the spec related to this in https://github.com/kubernetes-sigs/gateway-api/pull/2065! Took a closer look at the PR and have some follow up questions. Would appreciate some reviews from conformance reviewers as well.

/cc @arkodg @sunjayBhatia @mlavacca @LiorLieberman @Xunzhuo

conformance/tests/gateway-dynamic-listeners.go
dprotaso: By default `MaxTimeToConsistency` is set to 30 seconds - which was too low when I tested against Istio.

I could add a new param.

conformance/tests/gateway-dynamic-listeners.go
dprotaso: > This specific test would fail with a L4 Gateway implementation which I don't think we want.

FWIW this doesn't seem isolated to this PR - there are a bunch of Gateway tests that are bundling HTTPRoutes in their fixtures.

> Do we need "supported features" or similar for protocols?  ... Maybe there's a way to make the protocol(s) used in this test configurable to avoid this?

I'm a bit mixed here - I think testing this across multiple protocols is good. I think making this test configurable isn't ideal since it's not exhaustively testing the different scenarios. Then it would require multiple invocations of the conformance tests to get a picture if this feature works for all protocols.

I think we want multiple tests and that for now I should add `SupportHTTPRoute` to the set of features here.

conformance/utils/config/listener.go
dprotaso: re: multiple protocols see this discussion thread https://github.com/kubernetes-sigs/gateway-api/pull/1859#discussion_r1234153687

> What about implementations that may want to run this test with non-contiguous port ranges?

I'd be great if they could comment on this PR and highlight this. Otherwise I don't want to increase the scope/complexity here.

> If an implementation only supports HTTPS on 443, can/should we represent that with config here?

No - they shouldn't say they support this feature. This really about users who want to create many listeners with different ports and whether gateways support that.

conformance/tests/gateway-dynamic-listeners.go
dprotaso: closing as dupe - let's talk about it in this thread https://github.com/kubernetes-sigs/gateway-api/pull/1859#discussion_r1234153687

dprotaso:
rebased

conformance/tests/gateway-dynamic-listeners.go
robscott: It seems like you should be using a different timeout here, maybe this one:
https://github.com/kubernetes-sigs/gateway-api/blob/a124c94771951bf1e96556ed18071385e6babffb/conformance/utils/config/timeout.go#L38-L40

conformance/tests/gateway-dynamic-listeners.go
robscott: Yeah agree with all of that, `SupportHTTPRoute` is probably the best short term solution.

conformance/utils/config/listener.go
robscott: So `DynamicPortRange` needs to be a contiguous set of ports that an implementation can support HTTP or HTTPS protocols on? Maybe we should further tighten the name and scope of this test since it's limited to those protocols and HTTPRoute?

conformance/utils/suite/features.go
robscott: > What would be useful is if a GatewayClass publishes the dynamic range of ports it supports

I think we're saying the same thing here, I also want to be able to represent this in GWC status. I'm just coming from the perspective of an implementation that does not support a contiguous set of ports, so we'd need a way to represent that.

conformance/utils/config/listener.go
dprotaso: I changed the name of the feature to include `HTTPRoute`

conformance/tests/gateway-dynamic-listeners.go
dprotaso: Added the `SupportHTTPRoute` feature to the `SupportedFeatures` of the test

conformance/tests/gateway-dynamic-listeners.go
dprotaso: So the issue here is the existing helper method to make HTTP(s) requests is only using `MaxTimeToConsistency`

I made two changes to maybe make this more paltable in the short term
- changed the `2*time.Minute` to be `GatewayStatusMustHaveListeners`
- ensure we wait for the Programmed=True condition prior to invoking the HTTP(s) requests

conformance/utils/suite/features.go
dprotaso: Seems like that would warrant a GEP no? and is out of scope for this PR?

robscott: I'm not sure. The work @shaneutt and @mlavacca are doing with Conformance Profiles is leading to a point where conformance tests will be run based on something like `supportedFeatures` populated in GatewayClass status. So far that's the only input we need for all existing conformance tests. This is the first conformance test I'm aware of that also _requires_ additional inputs, and if we're adding it without a plan to migrate to the GWC status-based approach, we're just creating a headache for ourselves in the future.

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,220 @@
...
+				AttachedRoutes: 1,
+			})
+		}
+
+		kubernetes.NamespacesMustBeReady(t, s.Client, s.TimeoutConfig, namespaces)
+		certBytes, keyBytes, err := GetTLSSecret(s.Client, certNN)
+		require.NoErrorf(t, err, "error getting certificate: %v", err)
+
+		t.Run("should be able to add multiple HTTP listeners with dynamic ports that then becomes available for routing traffic", func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
```
robscott, line 161: Why isn't this timeout configurable?

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,220 @@
...
+
+			gwAddr, err := kubernetes.WaitForGatewayAddress(t, s.Client, s.TimeoutConfig, gwNN)
+			require.NoErrorf(t, err, "timed out waiting for Gateway address to be assigned")
+
+			for _, listener := range mutate.Spec.Listeners {
+				timeoutConfig := s.TimeoutConfig
+				// TODO - It takes much longer for listeners to be consistent
+				if timeoutConfig.MaxTimeToConsistency < timeoutConfig.GatewayStatusMustHaveListeners {
+					timeoutConfig.MaxTimeToConsistency = timeoutConfig.GatewayStatusMustHaveListeners
+				}
```
robscott, line 187: I don't think we should change a `timeoutConfig` value in the middle of a test - IMO these are intended to be read but not modified by tests. If an implementation can't pass within the timeout, they should increase the timeout in their conformance test config, but we shouldn't be adding workarounds in the tests themselves.

This specific test feels like it should be able to reuse the same timeouts used elsewhere in conformance tests. Theoretically the steps required to set up a new listener should be a subset of the steps required to set up a new Gateway. It seems like if `status` is showing the listener as programmed, then `MaxTimeToConsistency` should apply here just as much as it does anywhere else. Maybe the problem is that although we're waiting for a number of listeners in status, we're not waiting for them to have programmed set to true?

robscott (commented):
Thanks @dprotaso!

conformance/tests/gateway-dynamic-listeners.go
dprotaso: > Maybe the problem is that although we're waiting for a number of listeners in status, we're not waiting for them to have programmed set to true?

I have Programmed=True set as an expected condition to wait on.

> Then MaxTimeToConsistency should apply here just as much as it does anywhere else

Yeah you're right - I'm probably encoutering something weird with Istio.

conformance/tests/gateway-dynamic-listeners.go
dprotaso: done

conformance/tests/gateway-dynamic-listeners.go
dprotaso: reverted the change

conformance/utils/suite/features.go
dprotaso: Ok make sense. 

I guess something else to discuss - do we feel the port ranges should appear on the GatewayClass Status or would it be better on the Gateway?

conformance/tests/gateway-dynamic-listeners.go
```diff
@@ -0,0 +1,214 @@
...
+				}}, expectedConditions...),
+				AttachedRoutes: 1,
+			})
+		}
+
+		kubernetes.NamespacesMustBeReady(t, s.Client, s.TimeoutConfig, namespaces)
+		certBytes, keyBytes, err := GetTLSSecret(s.Client, certNN)
+		require.NoErrorf(t, err, "error getting certificate: %v", err)
+
+		t.Run("should be able to add multiple HTTP listeners with dynamic ports that then becomes available for routing traffic", func(t *testing.T) {
```
mlavacca, line 159:
```suggestion
		t.Run("should be able to add multiple HTTP listeners with dynamic ports that then become available for routing traffic", func(t *testing.T) {
```

mlavacca, line 159: I'm not saying we need to have it in this PR, but I think it would be cool to have another test case where we remove listeners, and we check that traffic does not flow anymore through them. WDYT?

dprotaso: I think it makes sense to add - done

shaneutt:
Since this hasn't been updated in a while, and has some conflicts I'm going to mark it as a draft again for now as this helps us organize our reviews. Once we're updated and ready for review please feel free to set it back to ready again.

PR #1757, opened 2023-02-23
GEP-1762: In Cluster Gateway Deployments

howardjohn:
Review note: This GEP was split into 3 - the current GEP (focusing on in-cluster deployments), https://github.com/kubernetes-sigs/gateway-api/pull/1868 (infrastructure field we depend on), and https://github.com/kubernetes-sigs/gateway-api/pull/1863/files (gateway merging)

**What type of PR is this?**

/kind gep

**What this PR does / why we need it**:

This PR attempts to inject some opinions about how in-cluster deployments should work. Currently, there is a lot of different implementations behaving differently, leading to confusion and drifting user experiences.

As-is, this PR is fairly opinionated. I expect that some of the MUSTs become SHOULDs as we iterate on this PR. In its current state, this is largely meant as a discussion point. I expect there to be a lot of iteration as we learn different implementations requirements and perspectives.

Note: most of the names were just picked arbitrarily. I'd like to first focus on the concepts, then we can debate the best names (almost certainly not what the initial GEP has!)

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note

```

Issue #1817, opened 2023-03-15
remove minItems limit of listeners in gateway


woodgear:
**What would you like to be added**:
remove minItems limit of listeners in gateway. 
**Why this is needed**:
we are currently to impl a web ui for gateway-api, a common case could be
1. create a gateway (via admin)
2. add or delete arbitrary listener in it later.
when a gateway must have atleast  one listeners, it had to decide which listener shoud be add on createing gateway. and must delete the gateway when want to delete the last listener in it.

howardjohn:
Prev discussion in https://github.com/kubernetes-sigs/gateway-api/pull/1596

shaneutt:
Given #1596 (particularly, https://github.com/kubernetes-sigs/gateway-api/pull/1596#issuecomment-1439467862) it doesn't appear we could move forward doing this without a full API revision which would be painful. We can see why this is useful, but the requirements for the change make it difficult. @woodgear what do you think about us defining some sentinel values as a workaround (e.g. something that indicates a listener that is _meant_ to be ignored) would this be adequate for your web UI to provide the display you're hoping for?

sdjksdajshd:
@shaneutt If we decide to do this, should the sentinel values be created by the front end or generated by the back end? I hope this is clear, to avoid confusion when accessing different classes.

shaneutt:
> @shaneutt If we decide to do this, should the sentinel values be created by the front end or generated by the back end? I hope this is clear, to avoid confusion when accessing different classes.

If by front-end you mean the UI, then I _think_ I would assume the back-end (the actual Gateway API implementation) would make this? But it's certainly open to debate since this is just an early concept on how to solve the problem _without_ having to go through a full API revision.

robscott:
I'd echo other comments from the previous issue (#1596) that this would be a breaking change and I don't think we can justify that here. Any change like this would require significant work for implementations. All implementations I know of today would require at least one listener to be present due to mapping to underlying load balancing infrastructure or a Service LB. If we introduced a state that didn't have that, it would be pretty messy to handle for every implementation and would risk breaking implementations that did not handle it before updated CRDs were installed in the same cluster.

Is there any reason the UI can't just suggest a default listener? Maybe 80 on HTTP? Could always be modified later if needed.

sdjksdajshd:
[...]

Issue #1687, opened 2023-01-31
Static/Unmanaged Gateway Mode (vs Provisioning)


shaneutt:
(follow-up from GAMMA meeting 2023-01-31)

In the past we've had discussions about a mode of operation for Gateway API which we'll call "static mode" (previously called "unmanaged mode" in some old issues) wherein a `Gateway` resource refers to a gateway that has been provisioned by some other means (e.g. via Helm) but the implementation may or may not actually manage the lifecycle of that gateway.

Multiple implementations now have options for such a mode which can be particularly helpful for allowing legacy/pre-operator deployments to have Gateway API functionality. Implementations that do this include (but are not limited to):

- Contour
- Istio
- Kong
- NGinx

Today there are prospective implementations of GAMMA (such as LinkerD) which are in need of such a mode, but _currently that mode is a colloquialism, and not properly documented or accounted for here in upstream Gateway API_.

TODO:

- [ ] add documentation that prescribes implementing "static" or "unmanaged" `Gateway` mode
- [ ] make any provisions necessary in the conformance tests to accommodate such an implementation, and/or potentially add this as a separate conformance category

Prior art: #892, #925

howardjohn:
Here is how Istio does it: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#manual-deployment

The idea was to match how cloud LB do it, where you point to an existing  LB by address

shaneutt:
Thanks @howardjohn!

cc @kflynn

sunjayBhatia:
similarly here for Contour: https://projectcontour.io/docs/v1.23.2/guides/gateway-api/#option-1-statically-provisioned

mlavacca:
This is how we at Kong deal with it: https://docs.konghq.com/kubernetes-ingress-controller/latest/concepts/gateway-api/#binding-kong-gateway-to-a-gateway-resource

Instead of relying on static/unmanaged `gateways`, we use static/unmanaged `gatewayClasses`, i.e., a Gateway is considered unmanaged if uses an unmanaged `GatewayClass`.

kate-osborn:
[NGINX](https://github.com/nginxinc/nginx-kubernetes-gateway) only supports "static" mode. We support one Gateway resource that is provisioned and managed by the user. We'd love to be able to run the conformance tests against our implementation via a static mode option or something similar.

kflynn:
[...]

PR #1653, opened 2023-01-16
GEP-1651: Gateway Routability

dprotaso:
/kind gep

**What this PR does / why we need it**:
Support different types of routable Gateways

**Which issue(s) this PR fixes**:

Supports (but does not resolve) https://github.com/kubernetes-sigs/gateway-api/issues/1651

PR #1863, opened 2023-03-22
GEP-1713: Standard Mechanism to Merge Multiple Gateways

dprotaso:
Note - this refactors out the 'merging' aspect of Gateways from https://github.com/kubernetes-sigs/gateway-api/pull/1757 - but tweaks the approach

**What type of PR is this?**

/kind gep

Outlines a mechanism to merge Gateways

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1713

```release-note

```


geps/gep-1713.md
```diff
@@ -0,0 +1,103 @@
...
+
+We want to keep this API very simple so that the merging requirement level could increase from `MAY` to `MUST`
+
+## API
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: parent-gateway
```
arkodg, line 38: can the parent gateway choose which children gateways it wants to be linked with ?
imo this can be mitigated by an defining the intent to merge at a top level (or even sibling level) rather than going bottom up (child to parent)

hzxuzhonghu (commented):
How can validate the invalid gateways conflicts?

geps/gep-1713.md
```diff
@@ -0,0 +1,103 @@
+# GEP-1713: Standard Mechanism to Merge Multiple Gateways
+
+* Issue: [#1713/kubernetes-sigs/gateway-api/issues/1713)
+* Status: Provisional
+
+(See status definitions [here](overview.md#status).)
+
+## TLDR
+
+The Gateway Resource is a contention point since it is the only place to attach listeners with certificates. We propose a mechanism to allow distinct Gateway resources to be logically merged.
```
briantkennedy, line 10: I'm not sure I see how this will address issues with attaching certificates.  Can you elaborate a bit more on what the core issue is with certificate attachment only existing in the Gateway, the personas might need to attach certificates, and update the example to include certificate attachment?

dprotaso: The root issue here is that the Knative control planes can't set a certificate for an HTTPRoute. There are valid use cases for this - ie. HTTP01 challenges produce a non-wildcard cert. 

For Knative we produce a certificate for each Knative Service with the host name of the form `{service-name}.{namespace}.some-operator-defined.tld`

In the GatewayAPI world we need to attach this certificate to the Gateway resource. But there's a disconnect between the lifecycle of the certificate and the Gateway. The cert lifecycle aligns more with the HTTPRoute (ie. deleting the Knative Service should delete this certificate). We also support HA controllers and each of them manipulating a single resource will cause 409 conflicts in the K8s API.

More details are here: https://github.com/kubernetes-sigs/gateway-api/discussions/1248

If you think it's worth revisiting TLS setting on an HTTPRoute I'd be open to starting that discussion.

geps/gep-1713.md
arkodg: 1. have a `mergeable`/`combinable` field at the `GatewayClass` level that need to be set to `true` or some enum (which defines merge capability)
or
2. add a `group` / `scope` field at the `Gateway` level as suggested by you earlier. But @youngnick raised a good point earlier [here](https://github.com/kubernetes-sigs/gateway-api/pull/1653#issuecomment-1447523825) that this style makes grouping hard to comprehend.

Both the approaches would require a new status condition at the `Gateway` level to be created to signal outcome of the merge.

dprotaso: I'm considering the split between gateways to just be 'logical' - meaning how do I take a single gateway and describe it using N objects. In contrast to Jon's proposal about how N distinct gateways can merge into the same infrastructure (https://github.com/kubernetes-sigs/gateway-api/pull/1757)

Based on that regarding 1. - I would want to avoid defining things at the GatewayClass. I think that answers 2 as well since you wouldn't merge Gateways with different addresses/scopes.

> Both the approaches would require a new status condition at the Gateway level to be created to signal outcome of the merge.

So I added the following clause to help simplify this 
> Each `Gateway` will only list listeners in their status if there is a corresponding entry in their spec.

Do you envision other status conditions?

geps/gep-1713.md
dprotaso: > If you think it's worth revisiting TLS setting on an HTTPRoute I'd be open to starting that discussion.
Asked in slack - https://kubernetes.slack.com/archives/CR0H13KGA/p1679510731917299

hzxuzhonghu:
>Can you provide some examples that come to mind?

Just come to mind because we met this in istio, istio's gateway has server fields which is like listener here. If all the listners in one gateway, it can be validated at least. But if they are in at least two gateways, once they are created at the same time, we have no way to validate them. This is the mechanism of eventual consistence limitation, the only way is to detect any conflict on running

geps/gep-1713.md
```diff
@@ -0,0 +1,103 @@
...
+---
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: child-gateway
+  labels:
+    gateway.networking.k8s.io/parent-gateway: parent-gateway
+spec:
+  gatewayClassName: example
+  parentGateway: parent-gateway
```
robscott, line 54: This seems to conflict with @howardjohn's KEP that proposes introducing an AttachTo ObjectRef, https://github.com/kubernetes-sigs/gateway-api/pull/1868. It seems like these KEPs are intended to work together, which approach is intended here?

dprotaso: From John's GEP
> See GEP-1713 for details, which this depends on.

My assumption is John will update his GEP according to the result of this one.  I'm open to discussion on the mechanics here.

My motivation for not putting it under a `infrastructure` field is because I view merging to be a logical operation done independently on the infrastructure. We're just describing a single Gateway using N resources - in order to work around the limitations we have when updating many listeners.  This is why I asked in slack whether there may be narrow use cases for certs on HTTPRoutes (https://github.com/kubernetes-sigs/gateway-api/pull/1863#discussion_r1145355857)

youngnick: The idea I was thinking of by putting this in `infrastructure` is that the split-out Gateways are then saying "I use this Gateway for my infrastructure, and Listeners I specify (along with Routes that attach to them) should be merged into that Gateway's infrastructure."

In this model, you'd have a single Gateway A that represents the actual infrastructure, and then other Gateways that reference that Gateway as their infrastructure provider. All Gateways that select Gateway A as their infrastructure provider would then be merged.

I think it achieves the same goal, without needing constructs like a boolean field to enable merging.

dprotaso: OK I'm convinced.

Are we ok with `infrastructure.attachTo` or prefer a more explicit `infrastructure.parentGateway` property ?

dprotaso: @howardjohn is there a reason why you used a `LocalObjectReference` vs a gateway name?

dprotaso: Thanks for clarifying - I'm good with `attachTo => Gateway` mechanic. 

I just want to confirm with everyone that `infrastructure` and this specific use of `attachTo` is something that we could see in 'Core' long term? And that they're not just specific to in-cluster deployments.

howardjohn: I cannot speak for everyone but my intent is that infrastructure goes beyond "in cluster"

geps/gep-1713.md
```diff
@@ -0,0 +1,218 @@
...
+  parentRefs:
+  - name: acme-lb
+    sectionName: https
+```
+
+#### `parent-gateway` label
+
+Implementations MUST label "child" Gateways with the `gateway.networking.k8s.io/parent-gateway` label. The value of the label is the name of the "parent" Gateway and it MUST be equivalent to the value `spec.infrastructure.attachTo.name`
+
+Users MUST NOT be able to edit this field.
```
dprotaso, line 107: I think if the project's admission webhook could add this label it would make it easier for implementations to do a lookup of child Gateways when reconciling a parent one.

It would be nice if tha parent had the same label.

An outstanding question here is the lack of `gatewayclass` and I wonder if the value should be `{gateway-class}/{parent-name}`

dprotaso:
Can folks take another look at the GEP - I hope to have addresses the topics in the discussions and I've flushed out a lot of the semantics.

geps/gep-1713.md
youngnick: > I just want to confirm with everyone that `infrastructure` and this specific use of `attachTo` is something that we could see in 'Core' long term? And that they're not just specific to in-cluster deployments.

Yes, I can see this in Core in the long term. It will need to start in Extended until we are confident that it's supported everywhere though.

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+#### Gateway Attaching
+
+Gateways that _do not_ reference another Gateway (using `spec.infrastructure.attachTo`) are considered "primary" Gateways. A "child" Gateway is one where `spec.infrastructure.attachTo` field is set to a "primary" Gateway. A "sibling" is a Gateway that shares a "parent" with another "child" Gateway.
+
+
+Failure to attach a "child" Gateway to another MUST result in the "child" Gateway setting their `Accepted` condition to `False`. This MUST happen in the following scenarios:
+- Attaching to another "child" Gateway
+- Attaching to oneself (self-referential)
+- Attaching to a non-existing Gateway
+- Attaching to a Gateway with a different `GatewayClassName`
```
howardjohn, line 85: I would add "Setting fields only valid on parent Gateways".

At this time, I think that is only spec.addresses. In the future, other parts of `infrastructure`

This could be done in the webhook as well, though (same with self-referential)

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### GatewayClassName 
+
+The `GatewayClassName` MUST be set on "child" and "primary" Gateways and be equivalent.
+
+#### Listeners
+
+Implementations MUST treat the "primary" Gateway as having the concatenated list of all listeners from "child" Gateways.
+
+Validation of this list of listeners is subject to the discussion here: https://github.com/kubernetes-sigs/gateway-api/issues/1842
```
howardjohn, line 122: not sure the issue covers this; we need to specify precedence for conflicts, like we do with routes

geps/gep-1713.md
howardjohn: actually this is mentioned below, just not that it results in Accepted=False

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+### Status Fields
+
+
+#### Addresses
+
+The list of `Addresses` that appear in the status of the "child" Gateway MUST be the same as the "parent" Gateway.
+
+#### Listeners
+
+The "child" Gateways MUST provide a status for each Listener defined in their specification. Listeners defined in "sibling" and the "parent" Gateway SHOULD NOT appear on a "child" Gateway.
```
howardjohn, line 173: Do child listeners appear in the parent?

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+#### YAML
+```yaml
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: child-gateway
+  labels:
+    gateway.networking.k8s.io/parent-gateway: parent-gateway
```
howardjohn, line 59: I don't think this label holds its own weight in the current proposal. Without more context, I would remove it

howardjohn (commented):
A few small comments but LGTM!

geps/gep-1713.md
dprotaso: I'll drop this from the GEP. It was really an optimization for implementations to do lookups of related gateways via a list+label selector.

Similar to how the `k8s.io/service-name` label is added to `EndpointSlices`

geps/gep-1713.md
dprotaso: Dropping the requirement of this label
comment here: https://github.com/kubernetes-sigs/gateway-api/pull/1863#discussion_r1157814631

geps/gep-1713.md
dprotaso: I don't think it should. Otherwise edits on the children will mean the parent will need to be updated which I want to avoid

howardjohn: Agreed. Good to specify this in the doc though.

geps/gep-1713.md
dprotaso: > This could be done in the webhook as well, though

I split out the obvious invalid scenarios where a webhook should reject the create/update

> I would add "Setting fields only valid on parent Gateways". ... actually this is mentioned below, just not that it results in Accepted=False

Updated the lower section to mention `Accepted=False`

geps/gep-1713.md
dprotaso: Added a section for precedence and conflicts

geps/gep-1713.md
dprotaso: removed

shaneutt (commented):
One of the teams I work with does some merging of Gateways which _may_ be relevant here and I would be interested to have their review on the subject:

/cc @mlavacca @pmalek @rainest 

If we could give a little bit of time for them to check this out they might have some good thoughts on the matter.

/hold

pmalek (commented):
No substiantial comments from my side but I just want to ask this to be sure: (maybe I've missed this in the comments somewhere):

Can a child attach to a parent in a different namespace? My gut tells me that this is not meant to be allowed but it wasn't explicitly mentioned here.

geps/gep-1713.md
```diff
@@ -0,0 +1,222 @@
...
+kind: HTTPRoute
+metadata:
+  name: httproute-example
+spec:
+  parentRefs:
+  - name: child-gateway
+    sectionName: metrics
+```
+
+#### Merging Spec Fields
```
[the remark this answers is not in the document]
dprotaso: done - I indented the sections below

dprotaso:
> Can a child attach to a parent in a different namespace? My gut tells me that this is not meant to be allowed but it wasn't explicitly mentioned here.

I added a sentence to make this explicit. The API uses an `LocalObjectReference` which means it can only reference Gateway in the same namespace.

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+
+##### GatewayClassName
+
+The `GatewayClassName` MUST be set on "child" and "parent" Gateways and be equivalent.
```
briantkennedy, line 109: If I understand correctly, the Gateway will assume the properties of the specified GatewayClass at a given moment in time.  If the GatewayClass changes between the parent Gateway creation and the child gateway creation, this would lead to a mismatch between the parameters indicated by the parent and child.

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+      name: parent-gateway
+      kind: Gateway
+      group: gateway.networking.k8s.io
+  listeners:
+  - name: metrics
+    port: 8080
+    protocol: HTTP    
+```
+
+With this configuration, the realized "primary" Gateway should listen on port `80` and `8080`.
```
briantkennedy, line 147: It feels unusual that the child would be able to cause the parent to listen on additional ports since this would make it relatively difficult for a user to understand the entire set of ports the Gateway is listening on.  Is merging new ports from the child gateway filling a specific requirement?

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+- Define a mechanic to merge multiple Gateways (logically)
+- Define a set of acceptable properties that can be merged and their semantics
+
+## Non-Goals
+
+- Apply a Gateway resource onto N distinct gateways (one to many)
+
+## Introduction
+
+Knative generates on demand per-service certificates using HTTP-01 challenges. There can be O(1000) Knative Services in the cluster which means we have O(1000) distinct certificates. Thus updating a single Gateway resource with this many certificates is a contention point and inhibits horizontal scaling of our controllers.
```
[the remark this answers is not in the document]
dprotaso: I called this out as an alternative but in various GitHub discussions @youngnick specifically wanted to avoid this approach.

geps/gep-1713.md
dprotaso: > If the GatewayClass changes between the parent Gateway creation and the child gateway creation, this would lead to a mismatch between the parameters indicated by the parent and child.

I don't think it's valid to change the gatewayClass after a Gateway has been created.

geps/gep-1713.md
dprotaso: I thought the introduction to the GEP covers the _why_. After reading it do you have further questions or suggestions to reword it?

geps/gep-1713.md
dprotaso: I think it would be handy to drop the requirement on the "child" Gateway but it would make the validation a bit harder. Right now this field is required so the API server will reject it if it's not set. 

Changing it would require a webhook or dealing with it when reconciling (thus losing making the feedback loop async)

geps/gep-1713.md
dprotaso: hi @briantkennedy just following up - as it's been a while

geps/gep-1713.md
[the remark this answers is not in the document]
dprotaso: Going to resolve - as there isn't any followup

geps/gep-1713.md
mlavacca, line 147: What about the constraint already posed to the listeners? For example, two listeners from the same gateway cannot have the same port or the same name. Do these constraints apply to listeners coming from different gateways that are attached?

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+  - name: metrics
+    port: 8080
+    protocol: HTTP
+```
+
+### Semantics
+
+#### Gateway Attaching
+
+Gateways that _do not_ reference another Gateway (using `spec.infrastructure.attachTo`) are considered "primary" Gateways. A "child" Gateway is one where `spec.infrastructure.attachTo` field is set to a "primary" Gateway - making the "primary" Gateway a "parent". A "sibling" is a Gateway that shares a "parent" with another "child" Gateway. A "child" Gateway can only attach to Gateways in the same namespace.
```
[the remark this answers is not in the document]
howardjohn: It might be feasible... I was initially thinking it made things like cert references not work out, but I am not sure that's actually true. I think its plausible this could be extended to support cross namespace as well

geps/gep-1713.md
howardjohn: Yes, I will add more details around this. But generally the idea should be the result is roughly the same as if you inlined all of the listeners

geps/gep-1713.md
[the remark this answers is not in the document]
dprotaso: @mlavacca do you have a use case for attaching across namespaces? I wonder if `ReferenceGrant` is granular enough to make this scenario 'safe'.

My use case is more about sharding a single Gateway so you don't have multiple controllers having a single gateway as a point of contention.

geps/gep-1713.md
dprotaso: This is covered by

> Implementations MUST treat the "parent" Gateway as having the concatenated list of all listeners from itself and "child" Gateways.
>
> Validation of this list of listeners is subject to the discussion here: https://github.com/kubernetes-sigs/gateway-api/issues/1842

geps/gep-1713.md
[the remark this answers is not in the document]
mlavacca: I have no use case ready in my mind, but in this kind of scenario where we may want to allow cross-namespace reference, that's what the `ReferenceGrant`s have been created for. It should be granular enough, as you can also specify the `Gateways` you want to be explicitly referred to in the `ReferenceGrantTo` field.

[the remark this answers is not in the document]
mlavacca: sounds good 👍

geps/gep-1713.md
```diff
@@ -0,0 +1,220 @@
...
+  gatewayClassName: example
+  infrastructure:
+    attachTo:
+      name: parent-gateway
+      kind: Gateway
+      group: gateway.networking.k8s.io
+  listeners:
+  - name: metrics
+    port: 8080
+    protocol: HTTP
```
david-martin, line 144: How would you see a listener with the same name as one in the parent being merged (or not).
For example, in the parent:

```yaml
    - allowedRoutes:
        namespaces:
          from: All
      hostname: test.example.com
      name: api
      port: 443
      protocol: HTTPS
```
and in the child:

```yaml
    - allowedRoutes:
        namespaces:
          from: All
      hostname: test.example.com
      name: api
      port: 443
      protocol: HTTPS
      tls:
        certificateRefs:
        - group: ""
          kind: Secret
          name: test.example.com
          namespace: example
        mode: Terminate
```

The use case here is a parent Gateway managed by a developer, and child gateways that are automatically created & managed by a controller (in this case supplementing the config with a tls secret reference that my controller is managing).

shaneutt:
@dprotaso checking in on this one: seems that we've got some approvals, and generally in its current `Provisional` state I would be inclined to approve it as well, but it does seem some unresolved comments linger that need attention. Let us know if you need any help moving this one forward!

geps/gep-1713.md
dprotaso: > How would you see a listener with the same name as one in the parent being merged (or not).

You need distinct names as per the merging logic defined here - https://github.com/kubernetes-sigs/gateway-api/blob/f2b55c8eaf833163f751ff1b9b2abe61db8e7343/apis/v1beta1/gateway_types.go#L87


> The use case here is a parent Gateway managed by a developer, and child gateways that are automatically created & managed by a controller (in this case supplementing the config with a tls secret reference that my controller is managing).

That's exactly the use case I'm hoping to cover. Though you'll need a distinct listener name and host name as per the link I sent.

This GEP doesn't adjust the merge semantics of a list of listeners. This GEP is really just splits that list of listeners over N objects so they can be manage more easily (eg. creating and reading status etc)

dprotaso: I realize I left a lingering link in the GEP let me update it

dprotaso: done

dprotaso:
@shaneutt added the last bit - to resolve the last comment.

geps/gep-1713.md
mikemorris: > can the parent gateway choose which children gateways it wants to be linked with ?

If this is still needed, I'd imagine it could look similar to [`AllowedRoutes`](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.AllowedRoutes), adding an optional field to `Gateway.spec` named something like `MergeableGateways` with a `namespaces` field similar to [`RouteNamespaces`](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.RouteNamespaces).

The biggest difference is that the default value for `MergeableGateways` (to ensure allowing merge semantics remains opt-in, and to raise a status condition if configured in spec but not supported by the implementation) should probably be:
```
from:
  namespaces: "None"
```
which would necessitate the addition of an additional value for [`FromNamespaces`](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.FromNamespaces)

youngnick:

dprotaso:
Following up on the question

> Can a child attach to a parent in a different namespace? My gut tells me that this is not meant to be allowed but it wasn't explicitly mentioned here.

I now have a use case that would warrant having a parent in a different namespace. Imagine the cluster operator persona is split into two parts - the cluster operator and the application operator.

The cluster operator provisions a 'parent' Gateway in some system namespace. They're responsible for installing/upgrading the underlying package that implements the Gateway API. They want to ensure costs are kept low so they want their Gateway instance to be shared with app operators. Thus there's only a single LB for the cluster.

The application operator owns a namespace and sets up the namespace to be used by app developers. The app operator will want to define a child Gateway with the right listeners and attaches the 'child' Gateway to 'parent' Gateway mentioned above. 

The application developer creates HTTPRoute and attaches them to the child gateway in their namespace.

dprotaso:
/hold

Rebased - but I still need to update the GEP to allow the scenario mentioned https://github.com/kubernetes-sigs/gateway-api/pull/1863#issuecomment-1927794180

keithmattix (commented):
This looks right to me; will approve later once I can go through in more detail

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+##### Validation
+
+A parent and child MUST have the same `gatewayClassName`.
+This will be detected by the implementation and reported in [status](#status-fields).
+
+A child resource MUST not set any `spec.infrastructure` fields beyond `attachTo`, and cannot set `spec.address`.
+This can be validated in the CRD schema.
+
+A parent resource MUST not set `spec.infrastructure.attachTo`.
+That is, we do not allow multiple tiers of Gateways chaining to each other; there is only a single parent with children..
```
mikemorris, line 241:
```suggestion
That is, we do not allow multiple tiers of Gateways chaining to each other; there is only a single parent with children.
```

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+This can be validated in the CRD schema.
+
+A child resource cannot `attachTo` any resource that doesn't allow attachment (i.e. it does not specify `spec.infrastructure.allowedChildren` for `Gateway`s).
```
mikemorris, line 244: The spirit of this makes sense for Gateway attachment targets, which are gaining the ability to allow attachment explicitly - I'm not sure if this language may be slightly overbroad with implications for usage such as attachment to other resources without this capacity, such as the proposed example of attaching to Service in https://github.com/kubernetes-sigs/gateway-api/pull/1863#discussion_r1150833245 @howardjohn

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+
+If there are conflicts between these, this should be reported as `Conflicted=True` in the listener as usual.
+
+### Status Fields
+
+#### Addresses
+
+The list of `Addresses` that appear in the status of the "child" Gateway MUST be the same as the "parent" Gateway.
+
+#### Gateway Conditions
```
mikemorris, line 304: For purposes of discoverability, is there any way to determine just by examining status on a parent Gateway, which child Gateways have successfully attached to it? (Agreed that pushing up listener conditions from child to parent would be undesirable though.)

mikemorris (approved):
This feels pretty solid to me now, and I believe all the dependent issues this has spawned have been resolved. Left a few questions but overall happy with where this is at and excited to see if move forward!

/approve

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+This proposal has two aspects: configuration for the parent Gateway, and configuration for the child Gateway.
+
+A "parent" Gateway _does not_ reference another Gateway. A parent Gateway MUST explicitly opt into merging.
+This is done with a new field `allowedChildren` in the `spec.infrastructure` of a Gateway.
+If, and only if, they have done so, the same Gateway is also permitted to leave `listeners` empty (currently, there is a `MinItems=1` restriction).
+
+A "child" Gateway references a parent Gateway. 
+This is done with a new field `attachTo` in the `spec.infrastructure` stanza of a Gateway.
+The `attachTo` field is a new type `GatewayObjectReference`.
+Although the use of `GatewayObjectReference` allows users to attach to any `kind`, this GEP only defines the behavior of attaching a Gateway to another Gateway.
```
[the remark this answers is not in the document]
dprotaso: `sectionName` and `port` don't really mean anything in this context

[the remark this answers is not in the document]
dprotaso: We have a bunch of these here actually - https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/object_reference_types.go


Big thing is to have smart defaulting - eg. in this scenario we want the kind/type to be a Gateway.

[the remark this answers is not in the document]
howardjohn: I think the general advice in k8s apis is to not excessively reuse types when they are not the same thing. Past example was Ingress and Service shared a type ofr status and a field accidentally got added to both thinking it was just one (oops!).

In this case, there are _already_ irrelevant fields so I think its best to keep it a new type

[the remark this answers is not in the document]
mikemorris: In that case, would a more generalized name like `AttachmentReference` (still defaulting to `group: gateway.networking.k8s.io` and `kind: Gateway`) feel a bit less awkward if it were to ever be used for attachment to non-Gateway kinds?

geps/gep-1713/index.md
dprotaso: I'll make this more specific to say 'Gateway resource'

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: The reason why the type has a `Gateway` prefix wasn't to imply that it only attaches to Gateways but that it's the Gateway that wants to attach to something else.

I'm flexible here - I can see how it can be read as "Attach to this other Gateway"

If we go for `AttachmentReference` then I'd probably drop the defaulting of the gateway kind. This would open up the type to be re-used in the future. eg. HTTPRoute attaching to another one.

Getting more input here would be useful

geps/gep-1713/index.md
dprotaso: A list of children on the parent's status could be an option here. Alternatively, could follow what routes do and have a count on the parent status.

I'm flexible here - would solicit what people want

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+* If `child-1` references a parent that doesn't exist then `child-1` will report `Accepted=False`
+* If `child-1` references a parent that doesn't allow merging then `child-1` will report `Accepted=False`
+* If `child-1` references another child (eg. `child-2`) then `child-1` will report `Accepted=False` 
+* If `child-1` references itself then `child-1` will report `Accepted=False`
+* If `child-1` and `parent` have different gatewayClassNames then `child-1` will report `Accepted=False`
+
+When reporting status of a child, an implementation SHOULD be cautious about what information from the parent or siblings are reported
+to avoid accidentally leaking sensitive information that the child would not otherwise have access to.
+
+#### Listener Conditions
```
howardjohn, line 327: We should probably add something like "An implementation MAY reject Listeners with ListenerConditionAccepted=False and reason TooManyListeners" or something, to allow an implementation to not have to support infinite number of listeners?

geps/gep-1713/index.md
mikemorris: Another path may be through `gwctl`, I don't think we _need_ to add this if it's problematic due to potential churn and/or not helpful enough as just a count.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+  parentRefs:
+  - name: child-gateway
+    sectionName: section-in-parent
+```
+
+#### Merging Spec Fields
+
+##### Validation
+
+A parent and child MUST have the same `gatewayClassName`.
```
louiscryan, line 234: Is this required or can the child simply omit. Omission is likely better if the root gateway is infrastructure that is being managed centrally. The children don't actually need to know or care about the class.

dprotaso: I think it would great to not require it.

I'm not familiar if it's possible with CEL to make a field optional if some other property is set (eg. infrastructure.attachTo`

@robscott do you know - I'd consider you the CEL expert.

howardjohn: You can do it, you just need to attach the rule to some common root (here that would be `spec`), rather than on `gatewayClassName`

mikemorris: It does make sense that the class needs to be managed centrally (by the parent Gateway administrator), but`spec.gatewayClassName` is a required field on [Gateway](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.Gateway) and while the field doesn't appear to actually be immutable, it's [recommended that the specified GatewayClass be used as a template](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.GatewayClass) when creating a Gateway so while it's a bit of extra boilerplate I don't expect it's something that should ever be changing dynamically.

What we may want to add is a clarification that about the GatewayClass "template snapshotting" recommendation - this should likely be tied to the parent Gateway rather than attempting to reconcile potentially drifting GatewayClass configuration snapshots from multiple child Gateways.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+The spec currently has language to indicate implementations `MAY` merge Gateways resources but the mechanic isn't defined.
+https://github.com/kubernetes-sigs/gateway-api/blob/541e9fc2b3c2f62915cb58dc0ee5e43e4096b3e2/apis/v1beta1/gateway_types.go#L76-L78
+
+## API
+
+This proposal has two aspects: configuration for the parent Gateway, and configuration for the child Gateway.
+
+A "parent" Gateway _does not_ reference another Gateway. A parent Gateway MUST explicitly opt into merging.
+This is done with a new field `allowedChildren` in the `spec.infrastructure` of a Gateway.
+If, and only if, they have done so, the same Gateway is also permitted to leave `listeners` empty (currently, there is a `MinItems=1` restriction).
```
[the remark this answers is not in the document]
louiscryan: I'm definitely support of this but its worth pointing out that this does not represent a guarantee that the merged set produces MinItems=1 since there may be no children. Thus we cannot completely state that runtime merging is the exact equivalent of a single merged resource.  

Given that runtime handling of empty lists by controllers was given as a reason for rejecting

https://github.com/kubernetes-sigs/gateway-api/pull/1596
https://github.com/kubernetes-sigs/gateway-api/issues/1817

we may have to reconsider the logic of that decision given the desire to support merging. E.g.

"if controllers want to support merging as per the spec they _must_ tolerate empty listeners even if merging is not used"

It seems like we are making a similar constraint loosening decision around certs for TLS listeners
https://github.com/kubernetes-sigs/gateway-api/pull/2721

geps/gep-1713/index.md
louiscryan: While I get the templating recommendation in principle I think it presents a toil problem for users when you attempt to apply it to the specification of Gateway.gatewayClassName.  

Since the compositional model for routes & policies is back-reference the toil fanout of forcing Gateway immutability is high on users. In this situation any class transition (infrastructure, version etc.) would force a reference update on every inbound policy since the user would have to create a new Gateway resource. 

Even immutability is somewhat of an illusion since the recommendation can't prevent the deletion & recreation of the resource with the same name and a difference value. The user is simply left with the side-effects of their workaround.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+#### Gateway Conditions
+
+Gateway conditions currently supports the following condition types: `Accepted` and `Programmed`
+
+For parent gateways, `Accepted` should be set based on the entire set of merged listeners.
+For instance, if a child listener is invalid, `ListenersNotValid` would be reported.
+`Programmed` is not expected, generally, to depend on the children resources, but if an implementation does depend on these
+they should consider child resources when reporting this status.
+
+For child gateways, `Accepted` and `Programmed` should consider the overall merged Gateway status, but only the child's own listeners.
```
sunjayBhatia, line 313:
```suggestion
For child gateways, `Accepted` and `Programmed` should not consider the overall merged Gateway status, but only the child's own listeners.
```

sunjayBhatia: I'm assuming this was the intended sentence, but might be slightly at odds with the below: "If parent is entirely invalid (for example, an invalid `address`), all three Gateways will reported `Accepted=False`."

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+	// ...
+
+	// AllowedChildren allows child objects to attach to this Gateway.
+	// A common scenario is to allow other objects to add listeners to this Gateway.
+	AllowedChildren *AllowedChildren `json:"allowedChildren,omitempty"`
+
+	// AttachTo allows the Gateway to associate itself with another resource.
+	// A common scenario is to reference another Gateway which marks
+	// this Gateway a child of another.
+	AttachTo GatewayObjectReference `json:"attachTo"`
```
sunjayBhatia, line 66: Not sure if this was discussed before, but would there be any appetite for this being a list so a single Gateway could attach to multiple parents? One way this could be useful is if there were separate parent/root Gateways for say prod/staging and any delegated child Gateways could attach to both to avoid duplicating config.

I personally don't have a particular affinity for this, just an idea in case it is useful/something to consider, I don't want to make this resource relationship too complicated if it is not useful

Has implications for the gatewayclass name discussion below as well

geps/gep-1713/index.md
```diff
@@ -0,0 +1,377 @@
...
+`Programmed` is not expected, generally, to depend on the children resources, but if an implementation does depend on these
+they should consider child resources when reporting this status.
+
+For child gateways, `Accepted` and `Programmed` should consider the overall merged Gateway status, but only the child's own listeners.
+
+For example, if I have a `parent`, `child-1`, and `child-2`:
+* If parent is entirely invalid (for example, an invalid `address`), all three Gateways will reported `Accepted=False`.
+* If `child-1` has an invalid listener, `parent` and `child-1` will report `ListenersNotValid`, while `child-2` will not.
+* If `child-1` references a parent that doesn't exist then `child-1` will report `Accepted=False`
+* If `child-1` references a parent that doesn't allow merging then `child-1` will report `Accepted=False`
```
sunjayBhatia, line 319: just for completeness/to be explicit, a related scenario to add onto this point: If `child-1` references a parent whose  allowed child namespaces do not include `child-1`'s namespace, then `child-1` will report `Accepted=False`

sunjayBhatia (approved):
Left a few comments, there may be a few more status edge cases I've missed and other comments will cover but overall I think this sounds like it would be useful for a lot of implementations (and implementable) so I'm a fan of getting this GEP in!

geps/gep-1713/index.md
howardjohn: It would make status fairly challenging. I think this can be solved at higher levels as well (tooling like helm, etc) so I am not sure its worth it

geps/gep-1713/index.md
dprotaso: done

keithmattix (approved):
LGTM - will continue conversation in the other threads

geps/gep-1713/index.md
mikemorris: For clarity, I would be in favor of allowing an empty `gatewayClassName` contingent on `attachTo` being configured if that's permissible - I just think it's a nicety rather than a necessity because changing a `gatewayClassName` will often be a destructive action akin to deleting and recreating all child and parent resources.

geps/gep-1713/index.md
mikemorris: We considered something similar during the route delegation exploration in https://github.com/kubernetes-sigs/gateway-api/issues/1058, but I think trying to add this here could significantly scope creep this GEP.

robscott (commented):
Thanks @dprotaso! Sorry for the delay in getting to this review. Generally appreciate that so many people have shown support for this GEP and that the overall concept of merging Gateways is already widely implemented.

I've got some concerns about the risks to safety here and the overall complexity this introduces to the API. Specifically the idea of a complexity budget comes to mind, and this specific area feels like it could end up having a significant cost when considering all the various knock-on effects like policy attachment, conflict resolution, cross-namespace merging, etc. 

With all of that said, I recognize that Gateway merging is already _very_ common, and I would like to see something like this work, I just want to make sure we can find sufficient guardrails to keep the complexity and safety costs down.

geps/gep-1713/index.md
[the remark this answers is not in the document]
cornfeedhobo: If I may toss in an opinion here..  I've been watching this proposal because I like the use case of service-based teams and trusting them to manage any needed domains.  It would be ideal if these teams didn't have to disrupt core infrastructure resources (e.g. a top-level 80/443 gateway) before deploying their domain.

Could this be more specific about the way `allowedChildren` should be used?  Could it specify an intent to support regex and/or broader matching, e.g. all gateways in a namespace.?

geps/gep-1713/index.md
dprotaso: Going to resolve this - as this idea introduces _more_ complexity for a GEP which folks thing is already complex ;)

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+parent or sibling listeners.
+If the implementation cannot apply the policy to only specific listeners, it should reject the policy.
+
+## Future Goals
+
+### Requirement Level
+
+We want to keep this API very simple so that the merging requirement level could increase from `MAY` to `MUST`
+
+## Alternatives
```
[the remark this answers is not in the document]
dprotaso: > Move more of this logic into Routes since those already support ~infinite merging with Gateways

I was under the impression this was off the table - but if it's _on_ the table that could simplify Knative's use case tremendously.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+#### Gateway Conditions
+
+Gateway conditions currently supports the following condition types: `Accepted` and `Programmed`
+
+For parent gateways, `Accepted` should be set based on the entire set of merged listeners.
+For instance, if a child listener is invalid, `ListenersNotValid` would be reported.
+`Programmed` is not expected, generally, to depend on the children resources, but if an implementation does depend on these
+they should consider child resources when reporting this status.
+
+For child gateways, `Accepted` and `Programmed` should not consider the overall merged Gateway status, but only the child's own listeners.
```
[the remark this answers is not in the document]
dprotaso: I don't think so - that would be disruptive so I would expect a bad child gateway to not be accepted and the siblings/parent remain intact.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+- Define a set of acceptable properties that can be merged and their semantics
+
+## Non-Goals
+
+- Apply a Gateway resource onto N distinct gateways (one to many)
+
+## Introduction
+
+Knative generates on demand per-service certificates using HTTP-01 challenges. 
+There can be O(1000) Knative Services in the cluster which means we have O(1000) distinct certificates.
```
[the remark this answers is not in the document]
dprotaso: @bravenut curious in your use case what are you configuring for each listener? Also how are you delegating the requests?

geps/gep-1713/index.md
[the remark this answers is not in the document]
mikemorris: @cornfeedhobo I'd expect `allowedChildren` to look very similar to [AllowedRoutes](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.AllowedRoutes), and specifically the `namespaces` field type [RouteNamespaces](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.RouteNamespaces), which yes would support matching all gateways one or more namespaces.

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: @bravenut just following up again to my earlier question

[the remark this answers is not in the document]
dprotaso: >  Is there some kind of number where it makes sense to just have more than one Gateway and/or push some of this complexity down to Routes?

@robscott Introducing more gateways (thus more IPs and underlying infra) incurs a cost and it's not reasonable for some customers.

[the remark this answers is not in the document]
mikemorris: We didn't really touch on it in the call today but another use case here is for managed providers who may want to consolidate separate logical gateways onto multi-tenant infrastructure for efficiency reasons.

Uburro:
Could you please explain why you want to implement mergeGateway but have no plans to work with certificates at the HTTPRoute level as it was done previously in Ingress? Currently, I'm working with Envoy Gateway and using the mergeGateway option. Of course, it's great that the developers of Envoy Gateway provided such a mechanism, especially in multi-tenant environments. Working with native CRDs like Gateway has made things easier, but nonetheless, if working with certificates were moved to the HTTPRoute level, it would simplify working with API Gateway even further.

geps/gep-1713/index.md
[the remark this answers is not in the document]
bravenut: @dprotaso We delegate requests based on the host name on the edge gateway. Then there's one HTTP route per inner gateway applied to the edge gateway routing requests to the inner gateways based on host name + path prefix.
 
Example scenario: One edge gateway, 3 inner gateways.
The edge gateway has two host names, say some.domain, some.other.domain configured.
 
Then there's three HTTP routes applied to the edge gateway:
HTTPRoute 1: hostname "some.domain", pathPrefix: "/v1", targeting: service of inner gateway "some-domain-v1"
HTTPRoute 2: hostname "some.domain", pathPrefix: "/v2", targeting: service of inner gateway "some-domain-v2"
HTTPRoute 3: hostname "some.other.domain", pathPrefix: "/v1", targeting: service of inner gateway "some-other-domain-v3"

interpixelstudios:
How would I handle SSL on a Kubernetes Gateway for multiple clients? My gateway currently handles HTTP traffic for [example] 20 different domains, using HTTPRoutes. On my gateway, HTTP would realistically just forward to HTTPS. No additional listeners, just the two. 

However, on reading how to incorporate SSL into the gateway, I cannot quite figure it out. On my old Ingress Controller [HAProxy], I could just tie the cert-manager issuer to the ingress resource itself and it would issue and enforce the cert. When reading about Gateways and their deployment, it seems like you only attach SSL to the Gateway itself. Wouldn’t that lock down the SSL encompassment to only the domains defined in the cert? What would happen if I need to add domain 21, wouldn’t it make more sense to be able to attach an issuer to the HTTPRoute itself and have it resolve SSL in there? I’m just very confused on how to migrate my existing stack from Ingress to Gateway.

My Class is gke-l7-global-external-managed-mc, thanks!

geps/gep-1713/index.md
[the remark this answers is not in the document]
robscott: >I was under the impression this was off the table

IMO Gateway merging would add an additional layer of hierarchy to the API. Given the complexity that entails, we really need to be open to other similarly complex approaches as well. Would love to see these alternatives written out a bit more to see if any of them can solve the use cases we've defined in this GEP. If so, we might find a solution that is at least somewhat simpler than Gateway merging.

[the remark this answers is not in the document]
mikemorris: The abandoned effort for "route delegation" can be found at https://github.com/kubernetes-sigs/gateway-api/issues/1058 and cross-referenced links if there's any content you want to review or pull from there.

geps/gep-1713/index.md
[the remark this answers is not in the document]
youngnick: One thing that just occurred to me when rereading this is that changing `minItems` to 0 for Listeners means that we are moving Listeners from _required_ to _optional_, which is a breaking API change (because fields that you used to be able to assume would always have at least one entry now do not). Strictly, that would mean we'd have to do an API revision to v2, which does not seem likely.

Realistically, it _may_ not be too much of a problem for Go implementations, if they're doing a `range` across `spec.listeners`, since that will execute the code 0 times if the list is empty. But it is a _substantial_ UX change to be able to add Gateways with no Listeners. It's _very_ likely that code will break as a result.

geps/gep-1713/index.md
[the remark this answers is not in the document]
youngnick: I agree that this should be `GatewayObjectReference` or maybe `ParentGatewayReference` to make the purpose super clear.

geps/gep-1713/index.md
youngnick: Again, `gatewayClassName` is a required field, and changing it to optional is a breaking API change.

geps/gep-1713/index.md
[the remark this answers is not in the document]
youngnick: @bravenut, I don't understand why the inner service is a Gateway? What you're describing is very possible today with Gateway, HTTPRoute and Service (where the Service is the target as you say above).

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+spec:
+  gatewayClassName: example
+  listeners:
+  - name: common-monitoring
+    port: 8080
+    protocol: HTTP
+  - name: domain-a
+    hostname: a.example.com
+    protocol: HTTP
+    port: 80
```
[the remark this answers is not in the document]
youngnick: Agreed, given the following things, which seem on the face of it to be reasonable asks:
- Gateway merging merges the Listeners of each Gateway into one flat list, essentially
- Gateway merging can occur across namespaces

There's no way around the fact that conflicts are not predictable by looking at any single namespace plus the parent Gateway. There's no way to know if someone in a sibling Gateway in another namespace is already using the same port/protocol/hostname/TLS config you are, which will put those Listeners into conflict.

By the current rules, Listeners that are conflicting are _both_ rejected, because they must necessarily be created at the same time (since they are in the same object). Assuming that we use our usual conflict-resolution rules, then, as @robscott says above, the first Gateway to claim the any distinct Listener combination will be the winner forever.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,380 @@
...
+
+For example, if I have a `parent`, `child-1`, and `child-2`:
+* If `parent` is entirely invalid (for example, an invalid `address`), all three Gateways will reported `Accepted=False`.
+* If `child-1` has an invalid listener, `parent` and `child-1` will report `ListenersNotValid`, while `child-2` will not.
+* If `child-1` references a parent that doesn't exist then `child-1` will report `Accepted=False`
+* If `child-1` references a parent that doesn't allow merging then `child-1` will report `Accepted=False`
+* If `child-1` references another child (eg. `child-2`) then `child-1` will report `Accepted=False` 
+* If `child-1` references itself then `child-1` will report `Accepted=False`
+* If `child-1` and `parent` have different gatewayClassNames then `child-1` will report `Accepted=False`
+* If `child-1` references a parent whose allowed child namespaces do not include `child-1`'s namespace, then `child-1` will report `Accepted=False`
```
youngnick, line 324: One other scenario here:

`parent`, `child-1`, `child-2`, and `child-3`.

All three children have multiple listeners.
`child-2` and `child-3` have _one_ listener that conflicts.
Let's say `child-2` was created first.

What happens?

How do the owners of each child know what's happened? (Preferably without looking at the parent's status)
What actions do the owners of `child-2` and `child-3` need to take to fix this? Can they know that which other sibling Gateway has caused the problem? If `child-2` wins precedence because of age, how do they know that something tried to claim their config and failed?

Note that in this case, if `child-2` ever deletes and recreates their Gateway, they will now be in the same position as the owner of `child-3`.

How does the owner of the parent know if any children are bad, and is it possible for them to know which ones?

geps/gep-1713/index.md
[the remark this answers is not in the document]
mikemorris: FWIW I'm mostly thinking of future needs here related to attaching to static infrastructure (#1687), where attachment to a non-Gateway resource such as an Azure [ApplicationLoadBalancer](https://learn.microsoft.com/en-us/azure/application-gateway/for-containers/quickstart-create-application-gateway-for-containers-managed-by-alb-controller?tabs=new-subnet-aks-vnet) should be possible on this field.

shaneutt:
This seems very stuck. Is anyone actively trying to un-stick it, or have any new context or thoughts on the subject?

dprotaso:
It's not stale - I was waiting on reviews of the GEP prior to addressing the feedback.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,380 @@
...
+(See status definitions [here](overview.md#status).)
+
+## tl;dr
+
+The Gateway Resource is a contention point since it is the only place to attach listeners with certificates. We propose a mechanism to allow distinct Gateway resources to be logically merged.
+
+## Goals
+
+- Define a mechanic to merge multiple Gateways (logically)
+- Define a set of acceptable properties that can be merged and their semantics
```
candita, line 15: @dprotaso Overall the only properties I see defined here are based on Gateway parents and children.  Does this mean we don't consider whether listeners are "[compatible](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/gateway_types.go#L159)" or "[distinct](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/gateway_types.go#L95)" as described in the `GatewaySpec`?

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: I wonder if this could be solved conditionally for example we allow `minItems == 0 `if `spec.infrastructure.allowedChildren` is not nil which implies the Gateway will be a parent. @robscott is that something that is possible with CEL?

> But it is a substantial UX change to be able to add Gateways with no Listeners. It's very likely that code will break as a result.

This doesn't break existing users or implementations. I'd say this is more analogous to supporting a new feature in a newer version of the API.

When an implementation adds 1.2 support they can reject Gateways with no listeners if they do not support gateway merging etc.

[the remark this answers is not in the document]
howardjohn: CEL was the intent here and how we implement "If, and only if, they have done so..."

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: I like `ParentGatewayReference` because it doesn't imply the Gateway is the _only_  type of object. It would allow for other resources like an Azure ALB.

geps/gep-1713/index.md
dprotaso: I don't think we can remove `gatewayClassName` because if we try to attach to a non-existent parent then who would be responsible to update the status of the Gateway saying invalid references etc?

I'm going to resolve this thread end of week unless someone has an idea on solving the situation I just mentioned.

geps/gep-1713/index.md
dprotaso: Added a count field as requested.

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: Added more alternatives

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+## Non-Goals
+
+- Apply a Gateway resource onto N distinct gateways (one to many)
+
+## Introduction
+
+Knative generates on demand per-service certificates using HTTP-01 challenges. 
+There can be O(1000) Knative Services in the cluster which means we have O(1000) distinct certificates. 
+Thus updating a single Gateway resource with this many certificates is a contention point and inhibits horizontal scaling of our controllers.
+[Istio Ambient](https://istio.io/v1.15/blog/2022/introducing-ambient-mesh/), similarly, creates a listener per Kubernetes service.
```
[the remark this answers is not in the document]
dprotaso: > maybe there's a simpler solution than Gateway merging?

Care to provide some alternative solutions? I'm going to close this out end of week otherwise.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+    port: 80
+```
+
+### Semantics
+
+#### Route Attaching
+
+Routes MUST be able to specify a child Gateway as a `parentRef` and make use of the fields in `ParentReference` to help target a specific listener.
+If no listener is targeted (`sectionName`/`port` are unset) then the Route references all the listeners on the child Gateway. It MUST NOT attach
+to a listener on a parent or sibling Gateway.
```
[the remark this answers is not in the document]
dprotaso: I don't understand the question - what conflict are you referring to?

geps/gep-1713/index.md
mikemorris: > I don't think we can remove gatewayClassName because if we try to attach to a non-existent parent then who would be responsible to update the status of the Gateway saying invalid references etc?

That makes a lot of sense to me, hadn't considered that situation.

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: >  Unfortunately this doesn't provide any kind of controls for that so it seems like the first namespace to claim a domain would win. A key idea of Gateway listeners was that they allowed you to delegate domains to routes in different namespaces, this seems to make that distinction less clear.

Not really - it's really more about splitting up the listener list across resources.

> A key idea of Gateway listeners was that they allowed you to delegate domains to routes in different namespaces, this seems to make that distinction less clear.

This doesn't change that in my opinion.

> Are there any controls/guardrails we can offer here? Is cross-namespace delegation a key goal? If not, could/should we start without it?

@howardjohn @mikemorris you have comments here regarding the cross namespace need?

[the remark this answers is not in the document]
dprotaso: > There's no way to know if someone in a sibling Gateway in another namespace is already using the same port/protocol/hostname/TLS config you are, which will put those Listeners into conflict.

This isn't really any different than what we have now. eg. if Gateway's aren't visible to Application Developers then they wouldn't know. I agree with your point it becomes more difficult when the resource is split up.

> By the current rules, Listeners that are conflicting are both rejected, because they must necessarily be created at the same time (since they are in the same object). Assuming that we use our usual conflict-resolution rules, then, as @robscott says above, the first Gateway to claim the any distinct Listener combination will be the winner forever.

The GEP tweaks the rules such that child Gateways can be rejected and it doesn't affect siblings. So like you said conflict resolution rules apply.

> As @robscott says above, the first Gateway to claim the any distinct Listener combination will be the winner forever.

Yup.

geps/gep-1713/index.md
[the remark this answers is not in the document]
dprotaso: New alternative mentions multiple gateways incurs a cost when it might not be needed.

Closing this out as this answers the original question.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,381 @@
...
+  listeners:
+  - name: metrics
+    port: 8080
+    protocol: HTTP
+status: ...
+```
+
+#### Use of the `gateway.networking.k8s.io/parent-gateway` label
+
+Use of a label (ie. `gateway.networking.k8s.io/parent-gateway: name`) could be used to select child gateways vs using `spec.infrastructure.attachTo`
```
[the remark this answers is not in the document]
dprotaso: Done

dprotaso:
@interpixelstudios your comment is sorta off topic for this PR. It might be better to create a discussion here - https://github.com/kubernetes-sigs/gateway-api/discussions

geps/gep-1713/index.md
dprotaso: > What happens?

`child-3` would be rejected and `child-2` wouldn't change. I'll add a condition about said conflict - I thought this was covered elsewhere.

> How do the owners of each child know what's happened? (Preferably without looking at the parent's status)

The status on the child could state an appropriate error message.

> What actions do the owners of child-2 and child-3 need to take to fix this? 

Gateway owners (cluster operator) would have to address this by tweaking listeners.

> Can they know that which other sibling Gateway has caused the problem? 

I think it's worth the error message saying there's a conflict and providing the more detailed message about siblings on the parent's status message.

> If child-2 wins precedence because of age, how do they know that something tried to claim their config and failed? 

Do you have a use case why child-2 would _want_ to know this information?

> Note that in this case, if child-2 ever deletes and recreates their Gateway, they will now be in the same position as the owner of child-3.

Yup

> How does the owner of the parent know if any children are bad, and is it possible for them to know which ones?

It's mentioned that `ListenersNotValid` will appear on the parent's status. A more detailed message can be presented.

geps/gep-1713/index.md
dprotaso: I believe this is sort of covered in the `listener precedence` sections and the `gateway conditions` section in the GEP.

I'll close this out end of week. Let me otherwise if those sections doesn't answer your question.

dprotaso: Circling back this is covered under the section "Merging Spec Fields"

children can't  set `spec.address` and is limited to what they can set in `spec.infrastructure`

geps/gep-1713/index.md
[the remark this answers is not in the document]
mikemorris: > Is cross-namespace delegation a key goal? If not, could/should we start without it?

I think there's two aspects to this:
- Should a child Gateway be able to attach to a parent Gateway in a different namespace?
  - I think this capability was somewhat implied by the discussion around `allowedChildren` up in https://github.com/kubernetes-sigs/gateway-api/pull/1863#discussion_r1540041898, but we should probably consider this directly - it may add more complexity than value.
- When a Gateway with attached routes attaches to a parent Gateway, what permissions are required on the parent? Does `allowedRoutes` config on the _parent_ Gateway apply here, such as if the parent specified `allowedRoutes` with a selector for only a _different_ namespace? Should we specify `allowedRoutes` as scoped to only _directly_ attached routes?

geps/gep-1713/index.md
candita: Ok.  If you don't mind changing the word "mechanic" to "mechanism" at some point, I'll comment in those sections, and you can close this out.

dprotaso: done

geps/gep-1713/index.md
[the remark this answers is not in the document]
candita: I agree there are lots of other reasons we'd want to specify gateway merging - the main one would be having fewer resources to manage.  If multiple listeners shared an endpoint, may as well merge them to reduce traffic as well.  If this proposal isn't really motivated by making things more efficient for those reasons, maybe gateway merging isn't the right solution.

geps/gep-1713/index.md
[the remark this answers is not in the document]
candita: I'm not keen on enforcing the parent/child relationship for this.  It might not be a bad idea to consider a policy or other CRD to record the members of a group that can merge with each other.  Then you are free to design a solution without the restrictions of breaking API changes.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+```
+
+#### Merging Spec Fields
+
+##### Validation
+
+A parent and child MUST have the same `gatewayClassName`.
+This will be detected by the implementation and reported in [status](#status-fields).
+
+A child resource MUST not set any `spec.infrastructure` fields beyond `attachTo`, and cannot set `spec.address`.
```
candita, line 237: Would you consider a case where there can be **some** listeners in a Gateway that are never supposed to merge, and those would need to maintain a `spec.address` in that case.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+    port: 8080
+    protocol: HTTP
+```
+
+With this configuration, the realized "parent" Gateway should listen on port `80` and `8080`.
+
+###### Listener Precedence
+
+Gateway Listeners should be merged using the following precedence:
+- "parent" Gateway
```
candita, line 292: Can you go into more detail?  Do you mean that a set of listeners from the parent Gateway will "win" in a conflict with another listener not in the parent Gateway?

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+- "child" Gateway ordered by creation time (oldest first)
+- "child" Gateway ordered alphabetical by “{namespace}/{name}”.
```
candita, line 294: Does this mean that the oldest listeners from a set of child Gateways will "win" in a conflct of conflict listeners from the set of child Gateways?  And if all are created at the same name, then the ordering goes by alphabetical namespace/name?  So everything in one early-letter namespace wins over any other namespace?

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+
+If there are conflicts between these, this should be reported as `Conflicted=True` in the listener as usual.
```
candita, line 296: Existing reasons for Conflicted=True are HostnameConflict and ProtocolConflict, and they represent a conflict amongst listeners.  The new conflicts presented here are more like conflicts amongst Gateways, is that right?  If not, maybe you could list some example conflict types.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Status Fields
+
+#### Addresses
+
+The list of `Addresses` that appear in the status of the "child" Gateway MUST be the same as the "parent" Gateway.
```
candita, line 302: Again, I think this is problematic in a case where you want to exclude a subset of the listeners from a gateway merge.  They should be able to maintain their own addresses.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+
+#### AttachedGateways
+
+The property `attachedGateways` will appear on the status of the "parent" Gateway will indicate the total number of "child" Gateways that have are attached
```
candita, line 306:
```suggestion
The property `attachedGateways` will appear on the status of the "parent" Gateway and will indicate the total number of "child" Gateways that are attached.
```

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+
+#### Gateway Conditions
+
+Gateway conditions currently supports the following condition types: `Accepted` and `Programmed`
+
+For parent gateways, `Accepted` should be set based on the entire set of merged listeners.
+For instance, if a child listener is invalid, `ListenersNotValid` would be reported.
```
candita, line 313: Do we already mark Accepted=False for an invalid listener?  If not, then is this only reflecting a "conflicted" status for a listener?  Won't marking Accepted=False make it impossible to attach a xRoute?  I don't think we should make the whole Gateway Accepted=False to account for only one invalid listener.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+`Programmed` is not expected, generally, to depend on the children resources, but if an implementation does depend on these
+they should consider child resources when reporting this status.
+
+For child gateways, `Accepted` and `Programmed` should not consider the overall merged Gateway status, but only the child's own listeners.
+The exception to this is if the parent Gateway is not valid in any regard.
+
+For example, if I have a `parent`, `child-1`, and `child-2`:
+* If `parent` is entirely invalid (for example, an invalid `address`), all three Gateways will reported `Accepted=False`.
```
candita, line 321: I'm not in favor of marking all three Gateways as Accepted=False if the parent is entirely invalid.  Why not let the other Gateways act as backups?

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+* If `child-1` has an invalid listener, `parent` and `child-1` will report `ListenersNotValid`, while `child-2` will not.
+* If `child-1` references a parent that doesn't exist then `child-1` will report `Accepted=False`
+* If `child-1` references a parent that doesn't allow merging then `child-1` will report `Accepted=False`
+* If `child-1` references another child (eg. `child-2`) then `child-1` will report `Accepted=False`
```
candita, line 325: This should be disallowed to begin with, unless a parent Gateway can transform into a child gateway.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+For child gateways, `Accepted` and `Programmed` should not consider the overall merged Gateway status, but only the child's own listeners.
+The exception to this is if the parent Gateway is not valid in any regard.
+
+For example, if I have a `parent`, `child-1`, and `child-2`:
+* If `parent` is entirely invalid (for example, an invalid `address`), all three Gateways will reported `Accepted=False`.
+* If `child-1` has an invalid listener, `parent` and `child-1` will report `ListenersNotValid`, while `child-2` will not.
+* If `child-1` references a parent that doesn't exist then `child-1` will report `Accepted=False`
+* If `child-1` references a parent that doesn't allow merging then `child-1` will report `Accepted=False`
+* If `child-1` references another child (eg. `child-2`) then `child-1` will report `Accepted=False` 
+* If `child-1` references itself then `child-1` will report `Accepted=False`
```
candita, line 326: Also, disallow from the start.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+* If `child-1` and `parent` have different gatewayClassNames then `child-1` will report `Accepted=False`
```
candita, line 327: Disallow from the start.

geps/gep-1713/index.md
```diff
@@ -0,0 +1,408 @@
...
+#### Listener Conditions
+
+Listener conditions should only be set for listeners directly defined in a given Gateway.
+Parent gateways MUST NOT have children's resources in their listener conditions list.
+Children gateways MUST NOT have parent's or sibling's resources in their listener conditions list.
+An implementation MAY reject Listeners with ListenerConditionAccepted=False and Reason "TooManyListeners"
+
+#### Policy Attachment
+
+Policies attached to a parent Gateway apply to both the parent and all children listeners.
```
candita, line 343: It would be nice to consider or rule out an alternative that is defined in a policy that could be attached to the parent Gateway.

geps/gep-1713/index.md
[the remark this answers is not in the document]
robscott: I started a doc for brainstorming here: https://docs.google.com/document/d/1qj7Xog2t2fWRuzOeTsWkabUaVeOF7_2t_7appe8EXwA/edit. I've added this proposal + a couple alternatives there, interested in any additional ideas people have. (They all include some form of Gateway merging for the record).

geps/gep-1713/index.md
dprotaso: > Would you consider a case where there can be some listeners in a Gateway that are never supposed to merge, and those would need to maintain a spec.address in that case.

If you don't want the Gateways to merge then I wouldn't be use the Gateway merging feature.

>  those would need to maintain a spec.address in that case.

I'm assuming you mean different `spec.address` - in this scenario just use two different disjoint Gateways

geps/gep-1713/index.md
dprotaso: > where you want to exclude a subset of the listeners from a gateway merge

This can be done by using separate disjoint Gateways. This subset merge isn't being suggested.

geps/gep-1713/index.md
dprotaso: Yes. Let me number the list to make it obvious there's an order.

geps/gep-1713/index.md
dprotaso: Yes. Let me number the list to make it obvious there's an order.

This is similar to how route matching works

https://github.com/kubernetes-sigs/gateway-api/blob/e971a8db22e70b77775622c5ed613c1bb5067e62/apis/v1/gateway_types.go#L304-L320

dprotaso: updated GEP

geps/gep-1713/index.md
dprotaso: > The new conflicts presented here are more like conflicts amongst Gateways, is that right?

The conflicts described in the GEP are listener conflicts.

dprotaso: Let me update this last line of text. It's a bit ambiguous.

dprotaso: Ok done. https://github.com/kubernetes-sigs/gateway-api/pull/1863/commits/43074ed0bdcad98ec0aa7c04f93d71e955ea7847

geps/gep-1713/index.md
dprotaso: Gateway merging is logical. The parent gateway is what creates the underlying infrastructure and the other child gateways are just a sharded list of listeners.

If a parent gateway isn't valid then there's no way to fallback onto a child listener.

geps/gep-1713/index.md
dprotaso: > Do we already mark Accepted=False for an invalid listener? 

It depends on the implementation - it's here https://github.com/kubernetes-sigs/gateway-api/blob/e971a8db22e70b77775622c5ed613c1bb5067e62/apis/v1/gateway_types.go#L825-L836

>  I don't think we should make the whole Gateway Accepted=False to account for only one invalid listener.

The spec allows Accepted=True to be present and to indicate there is a conflict with some listeners.

Details here:
https://github.com/kubernetes-sigs/gateway-api/blob/e971a8db22e70b77775622c5ed613c1bb5067e62/apis/v1/gateway_types.go#L143-L145

I'm going to resolve this since the GEP here is suggesting implementations follow the existing spec.

geps/gep-1713/index.md
dprotaso: It is - hence why we're saying the child gateway shouldn't be accepted.

> unless a parent Gateway can transform into a child gateway.

It cannot.

geps/gep-1713/index.md
dprotaso: Yup - hence it's not Accepted

geps/gep-1713/index.md
dprotaso: Yup it wouldn't be accepted.

geps/gep-1713/index.md
dprotaso: Can you elaborate what you mean? I don't understand.

geps/gep-1713/index.md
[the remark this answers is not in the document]
youngnick: I think for at least some of the use cases, cross-namespace attachment seems important - which implies that we should have a two-way handshake for safety. Which is difficult with the current object. Sigh. Another reason that @robscott and I have been talking about a new object type instead.

geps/gep-1713/index.md
mikemorris: I actually didn't realized we explicitly allowed partially accepting a Gateway as long as Conflicted listeners are excluded already. Agreed we're good with spec alignment here and no further change seems needed.

geps/gep-1713/index.md
mikemorris: The details here may vary between implementations, but in general I view this pattern as parent creates/reserves the underlying infrastructure, and child gateways program listeners onto it.

If the infrastructure fails to be created, the behavior of accepting child gateways would at best be unpredictable (something like falling back to spinning up their own in-cluster proxy pods instead of attaching to a centralized resource, potentially cluster-external) and could easily run into issues with L3 policy rules, IP allocation, etc such that I think rejecting children of an invalid parent is likely the safer and more consistent option.

dprotaso:
Hey folks - feedback has been to change approaches in terms of the Gateway <> Gateway relationship this GEP proposed.

Given the large # of comments and commits on this PR it didn't make sense to continue further revisions here. This PR has now been usurped by https://github.com/kubernetes-sigs/gateway-api/pull/3213

Going to close this out - please review the new PR when you get a chance.

Issue #1651, opened 2023-01-16
GEP:  Gateway Routability


dprotaso:
### Discussed in https://github.com/kubernetes-sigs/gateway-api/discussions/1247

<div type='discussions-op-text'>

<sup>Originally posted by **evankanderson** June 30, 2022</sup>
Currently, the `GatewayAddress` field supports only external (LoadBalanced) `Gateway` definitions. I'd like to suggest adding a `ClusterLocal` or `ClusterIP` [AddressType](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1alpha2.AddressType) with Extended support (like the existing `IPAddress` and `Hostname` support).

One of the early feature requests for Knative was the ability to deploy an application using Knative's HTTP routing support, but make it only available within the cluster. I want to be able to specify both the "internal" (`service.namespace.svc`) and "external" (`service.namespace.example.com`) Gateways using the same `GatewayClass` on the cluster, but ensure that the "internal" service is only reachable within the cluster. This would greatly simplify deployment for users over the instructions we have today.</div>

robscott:
Although this was on pace to be released as experimental in v0.8.0 (thanks to lots of great work by @dprotaso), the initial API review with @thockin and @khenidak raised some important questions that I don't think we have sufficient time to resolve before the v0.8.0 release.

I think the most important points are:

1. "Public" and "Private" are hard to define
2. Those become especially hard to define when combined with the incoming Kubernetes multi-network initiative

Until we can answer these questions, we've decided to pull this GEP back to provisional with these specific questions marked as "unresolved". Unfortunately that means this will miss the v0.8.0 release.

Rycieos:
I think any abstraction we try to make here will inherently be too limiting. If we make some field that accepts either `Public` or `Private`, inherently someone will some different behavior.

For myself, I have needed to set these fields on the `Service.Spec`:
```yaml
- type
- loadBalancerClass
- loadBalancerIP
- ipFamilyPolicy
- loadBalancerSourceRanges
- sessionAffinity
```

While a `Gateway` now has the `Infrastructure.annotations` field, not all of these settings are supported as annotations, and those that are are implementation and cloud provider specific.

My current workaround is to let the `Gateway` create a `Service`, and then clone that service with the same `selector`, but set my own values for the other fields. Then I use that `Service` instead of the one the `Gateway` creates.

[...]

PR #1868, opened 2023-03-22
GEP-1867: Per-Gateway Infrastructure

### LINKED ISSUE #1867: GEP: Provide a standard way to configure per-Gateway Infrastructure
**What would you like to be added**: a standard way to configure per-Gateway Infrastructure

**Why this is needed**: Currently this is only configurable on a Cluster scoped resource, which has a variety of issues

GEP: https://github.com/kubernetes-sigs/gateway-api/pull/1868

howardjohn:
Reviewer note: this was split from https://github.com/kubernetes-sigs/gateway-api/pull/1757 

**What type of PR is this?**
/kind feature

This PR introduces a new GEP on Per-Gateway Infrastructure. Read the GEP for details :-)

Fixes #1867

```release-note
NONE
```

### DIFF
--- a/geps/gep-1867.md
+++ b/geps/gep-1867.md
@@ -36,7 +36,7 @@ This has been previously discussed in [this issue](https://github.com/kubernetes
 
 As a cluster scoped resource, `GatewayClass` does not meet this requirement.
 This restricts customization use cases to either a few pre-provisioned classes by the admin, or running in an environment where the "Infrastructure Provider" and "Cluster Operator" are the same roles.
-The distinction between these roles is explicitly called out on the homepage: https://gateway-api.sigs.k8s.io/#what-is-the-gateway-api.
+The distinction between these roles is explicitly called out on the [homepage](https://gateway-api.sigs.k8s.io/#what-is-the-gateway-api).
 
 ### Custom Resource
 
@@ -46,7 +46,10 @@ Neither of these have any consistency between implementations.
 While there will always be some vendor-specific requirements, there are also a number of configuration aspects of a Gateway that are common between implementations.
 However, these cannot currently be expressed in a vendor-neutral way.
 
-Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations. The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
+The original motivation behind `parametersRef` was for implementation specific concepts, while portable comments could be added into the API as first-class fields, but this has not been done (yet).
+
+Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations.
+The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
 
 ### Separation of concerns
 
@@ -126,3 +129,8 @@ type GatewayInfrastructure struct {
 The exact fields contained in `GatewayInfrastructure` are subject to change, and additional common fields can be added as use cases arise.
 However, this will suffer from the common problem of a tension between providing common fields against offering extension points, so we should ensure that we are cautious to prevent excessive bloating of the field.
 Fields in `GatewayInfrastructure` should be attributes of infrastructure that are relatively common across implementations.
+
+### Status
+
+The API should likely expose some status. However, it is not yet clear what that will look like.
+This will be addressed prior to promotion beyond "Provisional".
--- a/geps/gep-1867.md
+++ b/geps/gep-1867.md
@@ -24,7 +24,7 @@ Related discussions:
 * Provide the ability to configure arbitrary (implementation specific) attributes about a **specific Gateway**.
 * Provide the ability to configure a standardized set of attributes about a **specific Gateway**.
 
-## Challenges with GatewayClass
+## Why not GatewayClass parameters?
 
 `GatewayClass.spec.parametersRef` is the existing mechanism to configure arbitrary fields on a Gateway.
 However, this introduces operational challenges when configuring Gateways.
@@ -78,9 +78,12 @@ For example, if I wanted to represent a `version` field and change that to trigg
 In order to address the concerns above, I propose a standard `infrastructure` API is added to `Gateway` and `GatewayClass`.
 Note the important part of this is the `Gateway` change; the `GatewayClass` aspect is mostly for consistency.
 
-Note that many of the fields under `GatewayInfrastructure` refer to work-in-progress GEPs.
-These fields are not tied to this GEP, and will only be included if the respective GEPs are approved.
-However, they are shown below to give a sense of the use cases the field can support.
+The exact fields are out of scope for this GEP and will be handled by additional GEPs.
+Some example GEPs already depending on this are [GEP-1713](/geps/gep-1713.md), [GEP-1651](/geps/gep-1651.md), and [GEP-1762](/geps/gep-1762.md)
+
+The fields as defined below are, of course, not useful.
+This is intended as a basis for other PRs, not to provide value on its own.
+This GEP will remain in provisional until at least one field is ready to be promoted.
 
 ```go
 type GatewaySpec struct {
@@ -94,41 +97,20 @@ type GatewayClassSpec struct {
   // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
   // the merging logic between these is as well. However, the GatewayClass is generally expected to be providing defaults
   // rather than overrides
-  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  Infrastructure GatewayClassInfrastructure `json:"infrastructure"`
   // ...
 }
 
 type GatewayInfrastructure struct {
-  // AttachTo marks this Gateway as a child to the referenced parent Gateway.
-  // See GEP-1713 for details, which this depends on.
-  AttachTo LocalObjectReference
-
-  // Scope defines the "scope" of the Gateway's addresses.
-  // Valid options are ClusterLocal or Public
-  // See GEP-1651 for details, which this depends on.
-  // NOTE: that GEP is likely to be implemented in a different way, at which point this field can be removed from this struct.
-  Scope AddressScope
-
-  // Version defines the version of the Gateway infrastructure to use.
-  // This is an opaque string that is implementation specific.
-  // Some possible valid examples include "canary", "v1.2.3", "docker.io/example:latest".
-  Version string
-
-  // Size defines the size of the Gateway infrastructure.
-  // The meaning of this is implementation specific.
-  // GEP-1762 will recommend this to mean "Pod Replicas" for in-cluster deployments.
-  Size int
-
   // ParametersRef provides a arbitrary implementation-specific configuration for
   // fields not expressed directly in this struct.
   // This follows the same semantics as GatewayClass's ParametersRef, but lives on the Gateway.
   ParametersRef ParametersReference
 }
-```
 
-The exact fields contained in `GatewayInfrastructure` are subject to change, and additional common fields can be added as use cases arise.
-However, this will suffer from the common problem of a tension between providing common fields against offering extension points, so we should ensure that we are cautious to prevent excessive bloating of the field.
-Fields in `GatewayInfrastructure` should be attributes of infrastructure that are relatively common across implementations.
+type GatewayClassInfrastructure struct {
+}
+```
 
 ### Status
 
--- a/geps/gep-1867.md
+++ b/geps/gep-1867.md
@@ -96,7 +96,7 @@ type GatewayClassSpec struct {
   // Infrastructure defines infrastructure level attributes for all Gateways in this class.
   // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
   // the merging logic between these is as well. However, the GatewayClass is generally expected to be providing defaults
-  // rather than overrides
+  // rather than overrides.
   Infrastructure GatewayClassInfrastructure `json:"infrastructure"`
   // ...
 }
@@ -111,6 +111,22 @@ type GatewayInfrastructure struct {
 type GatewayClassInfrastructure struct {
 }
 ```
+
+### API Principles
+
+For any given field, we will need to make two decisions:
+* whether this should be a first-class field or a generic `parametersRef`.
+* whether this field should be configurable on a Gateway and/or GatewayClass level
+
+The choice to use an extension (`parametersRef`) or first-class field is a well known problem across the API, and the same logic will be used here.
+Fields that are generally portable across implementations and have wide-spread demand and use cases will be promoted to first-class fields,
+while vendor specific or niche fields will remain extensions.
+Because infrastructure is somewhat inherently implementation specific, it is likely most fields will be Extended or ImplementationSpecific.
+However, there are still a variety of concepts that have some meaning between implementations that can provide value to users.
+
+Introduction at Gateway or GatewayClass level will depend on the specific field and use cases for the field.
+In general, it makes sense to provide defaults (GatewayClass) and specific settings (Gateway) for most fields, but
+this will be evaluated on a case-by-case basis.
 
 ### Status
 
--- a/geps/gep-1867.md
+++ b/geps/gep-1867.md
@@ -1,0 +1,128 @@
+# GEP-1867: Per-Gateway Infrastructure
+
+* Status: Provisional
+* Issue: [#1876](https://github.com/kubernetes-sigs/gateway-api/issues/1876)
+
+## Overview
+
+`Gateway`s represent a piece of infrastructure implemented by cloud load balancers, in-cluster deployments, or other mechanisms.
+These often need vendor-specific configuration outside the scope of existing APIs (e.g. "size" or "version" of the infrastructure to provision).
+
+Today `GatewayClass.spec.parametersRef` is available to attach arbitrary configuration to a `GatewayClass`.
+
+This GEP will explain why that is not sufficient to meet common use cases, and introduce a new field - `infrastructure` - to address these cases.
+
+Related discussions:
+* [Support cluster-local Gateways](https://github.com/kubernetes-sigs/gateway-api/discussions/1247)
+* [Scaling Gateway Resources](https://github.com/kubernetes-sigs/gateway-api/discussions/1355)
+* [Manual deployments](https://github.com/kubernetes-sigs/gateway-api/issues/1687)
+* [Merging Gateways](https://github.com/kubernetes-sigs/gateway-api/pull/1863)
+* [In Cluster Gateway Deployments](https://github.com/kubernetes-sigs/gateway-api/pull/1757)
+
+## Goals
+
+* Provide the ability to configure arbitrary (implementation specific) attributes about a **specific Gateway**.
+* Provide the ability to configure a standardized set of attributes about a **specific Gateway**.
+
+## Challenges with GatewayClass
+
+`GatewayClass.spec.parametersRef` is the existing mechanism to configure arbitrary fields on a Gateway.
+However, this introduces operational challenges when configuring Gateways.
+
+### Scope
+
+As a `Gateway` manager (with RBAC permissions to a specific `Gateway`) I should be able to declaratively make changes to that `Gateway` without the need for access to cluster-scoped resources (`GatewayClass`) and without affecting other `Gateways` managed by the same `GatewayClass`.
+This has been previously discussed in [this issue](https://github.com/kubernetes-sigs/gateway-api/issues/567).
+
+As a cluster scoped resource, `GatewayClass` does not meet this requirement.
+This restricts customization use cases to either a few pre-provisioned classes by the admin, or running in an environment where the "Infrastructure Provider" and "Cluster Operator" are the same roles.
+The distinction between these roles is explicitly called out on the homepage: https://gateway-api.sigs.k8s.io/#what-is-the-gateway-api.
+
+### Custom Resource
+
+`parametersRef` is entirely a generic implementation-specific meaning.
+This means implementations will either need a custom CRD or use untyped resources like ConfigMap.
+Neither of these have any consistency between implementations.
+While there will always be some vendor-specific requirements, there are also a number of configuration aspects of a Gateway that are common between implementations.
+However, these cannot currently be expressed in a vendor-neutral way.
+
+Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations. The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
+
+### Separation of concerns
+
+While there is value out of providing class-wide options as defaults, there is also value in providing these options on the object (Gateway) directly.
+
+Some parallels in existing APIs:
+
+[Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment) offers a hierarchy of defaults and overrides, allowing attachment to GatewayClass and Gateway.
+This is similar to our needs here, but representing infrastructure configuration as a "Policy" is a bit problematic, and the existing mechanisms have no hierarchy.
+
+In core Kubernetes, Pods declare their requirements (for example, CPU requests) inline in the Pod resource; there is not a `ResourceClass` API that abstracts these further.
+These higher level abstractions are handled by layered APIs (whether this is a CRD, an admission webhook, CI/CD tooling, etc).
+This allows users the flexibility to easily configure things per-pod basis.
+If the infrastructure admin wants to impose defaults or requirements on this flexibility, they are able to do so (in fact, `LimitRanger` provides a built in mechanism to do so).
+
+### Dynamic Changes
+
+Currently, the spec recommends `GatewayClass` to be used as a *template*.
+Changes to it are not expected to change deployed `Gateway`s.
+
+This makes usage problematic in a declarative way.
+For example, if I wanted to represent a `version` field and change that to trigger an upgrade, I would need to create an entirely new `Gateway`.
+
+## API
+
+In order to address the concerns above, I propose a standard `infrastructure` API is added to `Gateway` and `GatewayClass`.
+Note the important part of this is the `Gateway` change; the `GatewayClass` aspect is mostly for consistency.
+
+Note that many of the fields under `GatewayInfrastructure` refer to work-in-progress GEPs.
+These fields are not tied to this GEP, and will only be included if the respective GEPs are approved.
+However, they are shown below to give a sense of the use cases the field can support.
+
+```go
+type GatewaySpec struct {
+  // Infrastructure defines infrastructure level attributes about this Gateway instance.
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+type GatewayClassSpec struct {
+  // Infrastructure defines infrastructure level attributes for all Gateways in this class.
+  // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
+  // the merging logic between these is as well. However, the GatewayClass is generally expected to be providing defaults
+  // rather than overrides
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+type GatewayInfrastructure struct {
+  // AttachTo marks this Gateway as a child to the referenced parent Gateway.
+  // See GEP-1713 for details, which this depends on.
+  AttachTo LocalObjectReference
+
+  // Scope defines the "scope" of the Gateway's addresses.
+  // Valid options are ClusterLocal or Public
+  // See GEP-1651 for details, which this depends on.
+  // NOTE: that GEP is likely to be implemented in a different way, at which point this field can be removed from this struct.
+  Scope AddressScope
+
+  // Version defines the version of the Gateway infrastructure to use.
+  // This is an opaque string that is implementation specific.
+  // Some possible valid examples include "canary", "v1.2.3", "docker.io/example:latest".
+  Version string
+
+  // Size defines the size of the Gateway infrastructure.
+  // The meaning of this is implementation specific.
+  // GEP-1762 will recommend this to mean "Pod Replicas" for in-cluster deployments.
+  Size int
+
+  // ParametersRef provides a arbitrary implementation-specific configuration for
+  // fields not expressed directly in this struct.
+  // This follows the same semantics as GatewayClass's ParametersRef, but lives on the Gateway.
+  ParametersRef ParametersReference
+}
+```
+
+The exact fields contained in `GatewayInfrastructure` are subject to change, and additional common fields can be added as use cases arise.
+However, this will suffer from the common problem of a tension between providing common fields against offering extension points, so we should ensure that we are cautious to prevent excessive bloating of the field.
+Fields in `GatewayInfrastructure` should be attributes of infrastructure that are relatively common across implementations.


site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,137 @@
...
+GatewayClass is a cluster scoped resource. This effectively limits customization to either:
+1. Cluster admin pre-provisions a few classes, lower-privilege Gateway admins select them.
+2. Gateway admin and Cluster admin are the same.
+
+Case 1 works great when there are only a handful of customizations needed.
+For example, you could have a "public" vs "private" class, or a "big" and "small" class.
+This doesn't scale beyond simple cases, however; even the example above requires 4 classes and exponentially grows as more use cases are needed.
+
+Case 2 works fine, but greatly limits use cases, such as [Support namespace scoped implementations](https://github.com/kubernetes-sigs/gateway-api/issues/567).
+Rather than having 1 big ingress for the entire cluster, it is common for clusters to have multiple smaller gateways ([example](https://istio.io/latest/docs/setup/additional-setup/gateway/#gateway-deployment-topologies).
```
arkodg, line 44: why not create multiple `GatewayClass` for this use case ?

howardjohn: This entire section is describing why not to do that. tl;dr - you need 1 GWC per usage anyways, RBAC permissions separation is broken,  the separation is not useful, and no dynamic changes

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,137 @@
...
+This can be A gateway per app, per namespace,  per "tenant", or any mix of these.
```
shaneutt, line 45: I'm on board with you that there are challenges of doing this stuff at the `GatewayClass` level and this "scope" argument here is one of the most compelling to me. It seems to me like we're saying a little bit too much about it though, perhaps we can simplify:

```suggestion
### Scope

As a `Gateway` manager (with RBAC permissions to a specific `Gateway`) I should be able to declaratively make changes to that `Gateway` without the need for access to cluster-scoped resources (`GatewayClass`) and without affecting other `Gateways` managed by the same `GatewayClass`.
```

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,137 @@
...
+### Custom Resource
+
+`parametersRef` is entirely a generic implementation-specific meaning.
+This means users will either need a customer CRD or use untyped resources like ConfigMap.
+Neither of these have any consistency between implementations.
+While there will always be some vendor-specific requirements, there are also a number of configuration aspects of a Gateway that are common between implementations.
+However, these cannot currently be expressed in a vendor-neutral way.
+
+Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations. The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
+
```
shaneutt, line 56: While I'm not saying this issue is without merit, I think it's the least compelling of the 4 issues brought up _to me_. I personally would just remove it, as I think the rest stands fine without it and this one is slightly more subjective:

```suggestion
```

However I don't consider this comment blocking if you feel really strongly it's an important part of the decision making.

shaneutt (approved):
Several comments which I would like to see resolved before we merge please, but otherwise as a `Provisional` first step I think this is an important issue you're helping to identify and fix and a good place to start.

I feel pretty strongly that we should drop the implementation details from this PR as commented below, but if you prefer not to do that then I just expect we'll need to invest more time here.

Hold for another maintainer review.

/hold

shaneutt:
Hold for another maintainer review, just so an `lgtm` doesn't push us through just yet.

/hold

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,137 @@
...
+	Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+
+type GatewayClassSpec struct {
+  // Infrastructure defines infrastructure level attributes for all Gateways in this class.
+  // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
+  // the merging logic between these is as well.
+	Infrastructure GatewayInfrastructure `json:"infrastructure"`
```
robscott, line 102: Nit: Indentation looks off here.

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,137 @@
...
+type GatewayInfrastructure struct {
+  // AttachTo marks this Gateway as a child to the referenced parent Gateway.
+  // See GEP-1713 for details, which this depends on.
+  AttachTo LocalObjectReference
+
+  // Scope defines the "scope" of the Gateway's addresses.
+  // Valid options are ClusterLocal or Public
+  // See GEP-1651 for details, which this depends on.
+  // NOTE: that GEP is likely to be implemented in a different way, at which point this field can be removed from this struct.
+  Scope AddressScope
```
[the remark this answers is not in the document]
howardjohn: If we do it in spec.addresses we should not do it in spec.infra. I meant to indicate that in my comment that we may remove this depending on how 1651 goes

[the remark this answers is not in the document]
robscott: I actually think I like it in `infra` since in many/most cases only a single address scope is going to be supported by a Gateway. Are there important use cases I'm missing where we'd want to be able to request multiple address scopes?

arkodg:
* the Gateway API docs mentions that the `GatewayClass` resource is used to provision infrastructure so imo it is a natural place to define intent saying - "dont provision infra but link to it using this config"
* if there is a need to link infrastructure at the `Gateway` level, imo there should be some intent defined at the GatewayClass level saying - "dont provision infra but allow the Gateways to link to infrastructure using config defined at the Gateway level"

howardjohn:
> * the Gateway API docs mentions that the `GatewayClass` resource should be used to provision infrastructure so imo it is a natural place to define intent saying - "dont provision infra but link to it using this config"
> * if there is a need to link infrastructure at the `Gateway` level, imo there should be some intent defined at the GatewayClass level saying - "dont provision infra but allow the Gateways to link to infrastructure using config defined at the Gateway level"

Where is that mentioned? All of my understanding is that Gatewayclass is a template for infra, and Gateway is the infra.

https://gateway-api.sigs.k8s.io/api-types/gatewayclass/?h=gateway

> This resource represents a class of Gateways that can be instantiated.

site-src/geps/gep-1867.md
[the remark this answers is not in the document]
howardjohn: Statically assign a LoadBalancer IP and a ClusterIP, I guess? not sure why you would actually do that, though

Probably a discussion for 1651

arkodg:
> > * the Gateway API docs mentions that the `GatewayClass` resource should be used to provision infrastructure so imo it is a natural place to define intent saying - "dont provision infra but link to it using this config"
> > * if there is a need to link infrastructure at the `Gateway` level, imo there should be some intent defined at the GatewayClass level saying - "dont provision infra but allow the Gateways to link to infrastructure using config defined at the Gateway level"
> 
> Where is that mentioned? All of my understanding is that Gatewayclass is a template for infra, and Gateway is the infra.
> 
> https://gateway-api.sigs.k8s.io/api-types/gatewayclass/?h=gateway
> 
> > This resource represents a class of Gateways that can be instantiated.

based on https://gateway-api.sigs.k8s.io/api-types/gateway/#deployment-models, creating a `Gateway` could either create new infra or create new config for an existing infra, but it is tied to `GatewayClass`
```
Depending on the GatewayClass, the creation of a Gateway could do any of the following actions:

Use cloud APIs to create an LB instance.
Spawn a new instance of a software LB (in this or another cluster).
Add a configuration stanza to an already instantiated LB to handle the new routes.
Program the SDN to implement the configuration.
Something else we haven’t thought of yet...
The API does not specify which one of these actions will be taken.
```

howardjohn:
> Depending on the GatewayClass, the creation of a Gateway could do any of the following actions:

None of those imply the GatewayClass actually provisions the infrastructure IMO, just that Gateways can be merged. if you want to pre-provision some infra when GatewayClass is created that is fine - it doesn't impact this PR IMO. The entire `infrastructure` API is implementation specific and can be rejected/ignored.

site-src/geps/gep-1867.md
howardjohn: I also do think its the least meaningful but it is important IMO. Most of gateway-api is about balancing extension points vs 1p fields, so this addresses that aspect

site-src/geps/gep-1867.md
howardjohn: Turn your github tabs to 2 spaces and it looks great :wink:

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+This allows users the flexibility to easily configure things per-pod basis.
+If the infrastructure admin wants to impose defaults or requirements on this flexibility, they are able to do so (in fact, `LimitRanger` provides a built in mechanism to do so).
+
+### Dynamic Changes
+
+Currently, `GatewayClass` is intended as a *template*.
+Changes to it are not expected to change deployed `Gateway`s.
+
+This makes usage problematic in a declarative way.
+For example, if I wanted to represent a `version` field and change that to trigger an upgrade, I would need to create an entirely new `Gateway`.
```
youngnick, line 70: One of the things I like about this proposal is that this still allows people to create a second Gateway if the implementation can handle two running side-by-side, which will provide for smooth-and-slow transitions.

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+  // Infrastructure defines infrastructure level attributes about this Gateway instance.
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+
+type GatewayClassSpec struct {
+  // Infrastructure defines infrastructure level attributes for all Gateways in this class.
+  // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
+  // the merging logic between these is as well.
```
youngnick, line 92: I'd add something in here to the effect that the GatewayClass is expected to be a _default_ rather than an override _in general_.

site-src/geps/gep-1867.md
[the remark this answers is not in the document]
youngnick: Yes, I think this raises the point that #1651 implies that it will be useful to have one Gateway with multiple address scopes.

Personally, I think it may be better to encourage users to have different Gateways with different scopes, to make decisions about where to expose things more explicit (you'll need to point your Routes at _both_ Gateways if you want your route available in both scopes).

youngnick (commented):
I'm liking how this looks, but I think the interaction with #1651 is very interesting.

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+  // Version defines the version of the Gateway infrastructure to use.
+  // This is an opaque string that is implementation specific.
+  // Some possible valid examples include "canary", "v1.2.3", "docker.io/example:latest".
+  Version string
+
+  // Size defines the size of the Gateway infrastructure.
+  // The meaning of this is implementation specific.
+  // GEP-1762 will recommend this to mean "Pod Replicas" for in-cluster deployments.
+  Size int
```
danehans, line 117: I'm concerned whether an `int` can properly define the size of the Gateway infra. Size could define the type and amount of compute resources to use for the Gateway. IMO, `Replicas` is a well know definition throughout the industry and is a commonly requested config knob.

danehans (commented):
In general, I'm in favor of exposing Gateway-specific configuration through `gateway.spec`. The semantics for doing so are still open for discussion. Primarily, I'm interested in better understanding how merging of GC/GW config parameters are handled.

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+This means users will either need a customer CRD or use untyped resources like ConfigMap.
+Neither of these have any consistency between implementations.
+While there will always be some vendor-specific requirements, there are also a number of configuration aspects of a Gateway that are common between implementations.
+However, these cannot currently be expressed in a vendor-neutral way.
+
+Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations. The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
+
+### Separation of concerns
+
+While there is value out of providing class-wide options as defaults, there is also value in providing these options on the object (Gateway) directly.
```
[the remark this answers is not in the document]
howardjohn: This is discussed in the API spec itself, see L92

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+Currently, `GatewayClass` is intended as a *template*.
+Changes to it are not expected to change deployed `Gateway`s.
+
+This makes usage problematic in a declarative way.
+For example, if I wanted to represent a `version` field and change that to trigger an upgrade, I would need to create an entirely new `Gateway`.
+
+## API
+
+In order to address the concerns above, I propose a standard `infrastructure` API is added to `Gateway` and `GatewayClass`.
```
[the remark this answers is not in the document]
howardjohn: This is an established pattern in the API, IMO. To have core fields and extensions. For example, in HTTPRoute we have HeaderModifierFilter and ExtensionRefFilter.

Infrastructure provides core fields, and parametersRef provides an extension mechanism for vendor-specific fields

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+Note the important part of this is the `Gateway` change; the `GatewayClass` aspect is mostly for consistency.
```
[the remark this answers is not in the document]
howardjohn: I am fine with any name that will get this approved :-)

site-src/geps/gep-1867.md
howardjohn: Just a note - the specific fields of struct here are explicitly marked as not in scope for this GEP and are delegated to other GEPs. So your comment is valid but I don't think it should impact this PR at all. These are merely examples of the types of things that could be in the field.

That being said I will incorporate this into GEP-1762

site-src/geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+  // Size defines the size of the Gateway infrastructure.
+  // The meaning of this is implementation specific.
+  // GEP-1762 will recommend this to mean "Pod Replicas" for in-cluster deployments.
+  Size int
+
+  // ParametersRef provides a arbitrary implementation-specific configuration for
+  // fields not expressed directly in this struct.
+  // This follows the same semantics as GatewayClass's ParametersRef, but lives on the Gateway.
+  ParametersRef ParametersReference
```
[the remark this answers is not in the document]
howardjohn: Lets consolidate on the thread above

youngnick:
This is ready to go for me at `Provisional`.
 
However, we should be able to move this to Implementable/Experimental pretty soon (probably after v0.7.0 final is cut). That will give the further GEPs that depend on this a place to put their config. I think the only things remaining for `Implementable` are:
- making sure the new fields are optional and pointers
- Making the defaulting behavior not implementation-specific (we should be moving any fields we put in to Extended as quickly as possible,  I think)

This is already held, so I'll leave it held until @robscott has a review or unholds.

/lgtm

howardjohn:
> seems reasonable motivation to me, I vaguely remember something (but could be mixing up with another GEP or a previous conversation) in here about the Gateway becoming the [owner](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) of the relevant infra resources provisioned to implement it, is that in scope for this GEP? (maybe it is part of the implementation/how iteration?)

I somewhat intentionally did not specify this in the GEP (although maybe it would be useful to explicitly call it out as unspecified). I think its a good idea, but I don't know that this low level resource management needs to be specced?

In Istio we do this, and I think its a good idea, so I certainly don't mind adding it if others think we should.

sunjayBhatia:
> but I don't know that this low level resource management needs to be specced?

yep that is fair, could be in the "guidelines for implementers" kind of documentation if not in the actual spec, since it is a practicality/interoperability/etc. type of thing

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+However, this introduces operational challenges when configuring Gateways.
+
+### Scope
+
+As a `Gateway` manager (with RBAC permissions to a specific `Gateway`) I should be able to declaratively make changes to that `Gateway` without the need for access to cluster-scoped resources (`GatewayClass`) and without affecting other `Gateways` managed by the same `GatewayClass`.
+This has been previously discussed in [this issue](https://github.com/kubernetes-sigs/gateway-api/issues/567).
+
+As a cluster scoped resource, `GatewayClass` does not meet this requirement.
+This restricts customization use cases to either a few pre-provisioned classes by the admin, or running in an environment where the "Infrastructure Provider" and "Cluster Operator" are the same roles.
+The distinction between these roles is explicitly called out on the homepage: https://gateway-api.sigs.k8s.io/#what-is-the-gateway-api.
```
robscott, line 39: Can you turn this into a link?

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+Additionally, there is hesitancy to use a CRD (which leads to CRD proliferation), which pushes users towards untyped ConfigMaps which are not much better than annotations. The scoping, as mentioned above, is also a bit awkward of a cluster scoped resource pointing to a namespaced object.
+
+### Separation of concerns
+
+While there is value out of providing class-wide options as defaults, there is also value in providing these options on the object (Gateway) directly.
+
+Some parallels in existing APIs:
+
+[Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment) offers a hierarchy of defaults and overrides, allowing attachment to GatewayClass and Gateway.
+This is similar to our needs here, but representing infrastructure configuration as a "Policy" is a bit problematic, and the existing mechanisms have no hierarchy.
```
robscott, line 58: Why would we not want defaults _and_ overrides for this at GatewayClass level? Are we sure that cluster admins are going to be fine with others creating custom variations of Gateways and overriding GatewayClass config?

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+  // Scope defines the "scope" of the Gateway's addresses.
+  // Valid options are ClusterLocal or Public
+  // See GEP-1651 for details, which this depends on.
+  // NOTE: that GEP is likely to be implemented in a different way, at which point this field can be removed from this struct.
+  Scope AddressScope
+
+  // Version defines the version of the Gateway infrastructure to use.
+  // This is an opaque string that is implementation specific.
+  // Some possible valid examples include "canary", "v1.2.3", "docker.io/example:latest".
+  Version string
```
robscott, line 112: Does this have a separate GEP in progress or is it just meant to be covered by this one?

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+This restricts customization use cases to either a few pre-provisioned classes by the admin, or running in an environment where the "Infrastructure Provider" and "Cluster Operator" are the same roles.
+The distinction between these roles is explicitly called out on the homepage: https://gateway-api.sigs.k8s.io/#what-is-the-gateway-api.
+
+### Custom Resource
+
+`parametersRef` is entirely a generic implementation-specific meaning.
+This means implementations will either need a custom CRD or use untyped resources like ConfigMap.
+Neither of these have any consistency between implementations.
+While there will always be some vendor-specific requirements, there are also a number of configuration aspects of a Gateway that are common between implementations.
+However, these cannot currently be expressed in a vendor-neutral way.
```
robscott, line 47: I think the idea had always been to add portable concepts to GatewayClass and leave implementation-specific concepts for parametersRef. It could be helpful to provide that context here.

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+Some parallels in existing APIs:
+
+[Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment) offers a hierarchy of defaults and overrides, allowing attachment to GatewayClass and Gateway.
+This is similar to our needs here, but representing infrastructure configuration as a "Policy" is a bit problematic, and the existing mechanisms have no hierarchy.
+
+In core Kubernetes, Pods declare their requirements (for example, CPU requests) inline in the Pod resource; there is not a `ResourceClass` API that abstracts these further.
+These higher level abstractions are handled by layered APIs (whether this is a CRD, an admission webhook, CI/CD tooling, etc).
+This allows users the flexibility to easily configure things per-pod basis.
+If the infrastructure admin wants to impose defaults or requirements on this flexibility, they are able to do so (in fact, `LimitRanger` provides a built in mechanism to do so).
```
robscott, line 63: It would be helpful to provide some guidelines for when we'd want Gateway-level config to be able to override class-level config. I personally think there are many cases where admins would not want to allow config to be overridden. This feels especially problematic if we have something parametersRef that could potentially allow _all_ GatewayClass params to be overridden at the Gateway level, including something like implementation-specific firewall config.

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Dynamic Changes
+
+Currently, the spec recommends `GatewayClass` to be used as a *template*.
+Changes to it are not expected to change deployed `Gateway`s.
+
+This makes usage problematic in a declarative way.
+For example, if I wanted to represent a `version` field and change that to trigger an upgrade, I would need to create an entirely new `Gateway`.
```
robscott, line 71: I think in most cases it's significantly safer to spin up a new Gateway when doing any kind of meaningful infra-level changes, but I'd agree that it's overkill for bumping the image version used. There are some simpler ways to solve this though:

1. Suggest that implementations document which fields will result in in-place upgrades vs only apply to future Gateways
2. Allow Gateway owners to specify if they want to subscribe to automatic updates from the GatewayClass

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+* [Support cluster-local Gateways](https://github.com/kubernetes-sigs/gateway-api/discussions/1247)
+* [Scaling Gateway Resources](https://github.com/kubernetes-sigs/gateway-api/discussions/1355)
+* [Manual deployments](https://github.com/kubernetes-sigs/gateway-api/issues/1687)
+* [Merging Gateways](https://github.com/kubernetes-sigs/gateway-api/pull/1863)
+* [In Cluster Gateway Deployments](https://github.com/kubernetes-sigs/gateway-api/pull/1757)
+
+## Goals
+
+* Provide the ability to configure arbitrary (implementation specific) attributes about a **specific Gateway**.
+* Provide the ability to configure a standardized set of attributes about a **specific Gateway**.
```
robscott, line 25: These goals feel incomplete because you're also proposing standard infra-level config at the GatewayClass level but these only focus on individual Gateways.

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+  // Size defines the size of the Gateway infrastructure.
+  // The meaning of this is implementation specific.
+  // GEP-1762 will recommend this to mean "Pod Replicas" for in-cluster deployments.
+  Size int
+
+  // ParametersRef provides a arbitrary implementation-specific configuration for
+  // fields not expressed directly in this struct.
+  // This follows the same semantics as GatewayClass's ParametersRef, but lives on the Gateway.
+  ParametersRef ParametersReference
```
robscott, line 122: I'm having trouble determining if some of these fields are covered by separate GEPs or meant to be included here. One quick comment - if infrastructure is included on GatewayClass, we could not include ParametersRef in the struct as it would conflict with the existing field.

robscott (commented):
Thanks @howardjohn! Having trouble keeping track of all the related GEPs around this so let me know if any of these comments either don't make sense or have been addressed elsewhere.

site-src/geps/gep-1867.md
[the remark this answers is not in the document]
robscott: I think this is a very important question because this GEP introduces a lot of overlap between the two resources. @howardjohn do you have a link to the API spec you're referring to? The closest I could find was the GatewayClass spec, but it doesn't really seem to cover merging.

https://github.com/kubernetes-sigs/gateway-api/blob/7b8f168c82a90522209a340bf291fcacab0fc5ef/apis/v1beta1/gatewayclass_types.go#L82-L94

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+type GatewayInfrastructure struct {
+  // AttachTo marks this Gateway as a child to the referenced parent Gateway.
+  // See GEP-1713 for details, which this depends on.
+  AttachTo LocalObjectReference
+
+  // Scope defines the "scope" of the Gateway's addresses.
+  // Valid options are ClusterLocal or Public
+  // See GEP-1651 for details, which this depends on.
+  // NOTE: that GEP is likely to be implemented in a different way, at which point this field can be removed from this struct.
+  Scope AddressScope
```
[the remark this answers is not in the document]
howardjohn: This GEP will have an empty set of fields added to infrastructure first. It should  go in first so other GEPs can use it to add fields. All fields here are just hypothetical placeholders.

geps/gep-1867.md
howardjohn: The implicit goal on the GEP is "Have consistency in the API". Providing `infrastructure` is purely to maintain consistency between the APIs; the entire goal of the GEP is to _not_ use GatewayClass.

geps/gep-1867.md
howardjohn: This is overlapping with https://github.com/kubernetes-sigs/gateway-api/pull/1868#discussion_r1198096967 so I am going to resolve this and address it there

geps/gep-1867.md
howardjohn: Good point. We can probably do a GatewayInfrastructure and GatewayClassInfrastructure, with GatewayInfrastructure just having additional ParametersRef that is not in GatewayClassInfrastructure to address this.

or we deprecate GatewayClass.ParametersRef, but that is probably not worth it

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+type GatewayClassSpec struct {
+  // Infrastructure defines infrastructure level attributes for all Gateways in this class.
+  // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
+  // the merging logic between these is as well. However, the GatewayClass is generally expected to be providing defaults
+  // rather than overrides
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
```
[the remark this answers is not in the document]
howardjohn: Yes they will all be Extended or Implementation-Specific likely. Nothing in this is final, its just a sample of the types of things that would be here. Most of these already have GEPs that flesh these out more appropriately (blocked on this GEP)

geps/gep-1867.md
howardjohn: it will be separate. As mentioned in the GEP, everything here is just an example for now. Note this GEP is `provisional`, its just to unblock other GEPs depending on it.

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+
+In order to address the concerns above, I propose a standard `infrastructure` API is added to `Gateway` and `GatewayClass`.
+Note the important part of this is the `Gateway` change; the `GatewayClass` aspect is mostly for consistency.
+
+Note that many of the fields under `GatewayInfrastructure` refer to work-in-progress GEPs.
+These fields are not tied to this GEP, and will only be included if the respective GEPs are approved.
+However, they are shown below to give a sense of the use cases the field can support.
+
+```go
+type GatewaySpec struct {
```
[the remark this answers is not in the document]
howardjohn: Any suggestions on what we actually put? Something like the same fields but with defaults filled in (ex, they specify version=latest and we set the status to version=1.2.3, or they leave size unset and we specify we default to "large" or whatever)?

I marked this as WIP for the GEP, blocking promotion from provisional. I would like to get this merged so we can unblock the growing number of GEPs that are blocked by it

geps/gep-1867.md
```diff
@@ -0,0 +1,128 @@
...
+  // Infrastructure defines infrastructure level attributes about this Gateway instance.
+  Infrastructure GatewayInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+type GatewayClassSpec struct {
+  // Infrastructure defines infrastructure level attributes for all Gateways in this class.
+  // A Gateway may provide configuration for the same values; as all fields in GatewayInfrastructure are implementation specific,
+  // the merging logic between these is as well. However, the GatewayClass is generally expected to be providing defaults
```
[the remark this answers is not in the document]
howardjohn: Responding on https://github.com/kubernetes-sigs/gateway-api/pull/1868#discussion_r1198096967 and resolving this, as my comment will overlap

geps/gep-1867.md
howardjohn: We don't want AttachTo in GatewayClassInfrastructure either

geps/gep-1867.md
howardjohn: Also responding to your other comment here
> Why would we not want defaults and overrides for this at GatewayClass level? Are we sure that cluster admins are going to be fine with others creating custom variations of Gateways and overriding GatewayClass config?
> Why would the merging logic be implementation specific?


I think there are a few options:
1. `default` and `overrides` in Gateway, aligning with policy attachment

Makes paramRefs awkward, since it doesn't have these concepts

Gives GatewayClass editor ability to decide what Gateway admins can do. I don't think this is useful in most cases, since most users are not actually editing GC but are using defaults provided by implementations during install.

2. Implementation specific merge

This aligns with paramRef, which already has implementation specific merging with Gateways (if they have per-gateway settings, of course -- its all impl specific so can be anything).

Gives the _controller_ author ability to decide what gateway admins can do. I suspect this is often actually the person writing the GC, anyways, though, so this ends up being similar to (1) but simpler.

3. No `infrastructure` field at GC

Not consistent with Gateway API. Users who want this at GC level would use a paramRef to something with the same fields, and then would have impl specific merging logic. So ends up the same as (2), roughly, but with an inconsistent UX I think.

---

So overall it seems like if we think GC owner wants overrides we should do (1), else we should do (2).

I don't really see why a GC would want to override TBH. For things like Version and Size, you probably don't want override -- you want a range. This is better addressed through other mechanisms like Gatekeeper (effectively what LimitRanger is in core). So I would lean towards (2).

---

I will note that the merging logic and the presence of the field are decoupled. Since the presence of the field is blocking 3 other GEPs, I can say the merging logic is WIP and a blocker for promotion beyond 'provisional' if it helps?

geps/gep-1867.md
robscott: Maybe it would help to retitle and refocus the current "Challenges with GatewayClass" section to "Why not GatewayClass params?" or similar. I don't think this GEP is saying that we shouldn't use GatewayClass at all, just that there are some cases where Gateway level config would be helpful. I think this GEP should focus on the specific situations where Gateway-level config is more useful than GWC-level config. 

I'm not even sure that either of my suggestions above need to be alternatives considered. I think either or both could be reasonable extensions of the GEP without significantly changing it.

I think this is really getting to an overall theme of my comments/perspective on this GEP: We need principles for when to introduce infra config at each level. I don't think every GWC-level param should have a GW-level equivalent, and of course the same is true in the other direction.

geps/gep-1867.md
robscott: >the entire goal of the GEP is to not use GatewayClass

This seems problematic. Not to use GatewayClass _at all_ or just for this specific set of config?

geps/gep-1867.md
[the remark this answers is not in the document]
howardjohn: Nothing. It would be an empty `Infrastructure` struct. Any additions will be followups/other GEPs.

geps/gep-1867.md
howardjohn: It is to be able to configure infrastructure options on a per-gateway basis. Its not to stop users from using GC, but they can already do that today. They cannot do it on a per-GW basis today, so this GEP aims to enable this.

geps/gep-1867.md
howardjohn: I removed all the fields to avoid confusion

geps/gep-1867.md
howardjohn: I updated this to provide principals

geps/gep-1867.md
howardjohn: I added a section on this

youngnick:
Okay, I caught up on this one, and I think it's easier to just bring my comments back to the top because they're about a few of these threads. I've previously approved, but I think that @robscott has brought up some great points.

On reflection, what we're doing with this is graduating some classes of config from requiring a `paramsRef` to standard fields, and `infrastructure` is the place to put these fields. Because the things we're talking about are standard, we can also include an `infrastructure` stanza on Gateway as well as the one on GatewayClass.

So, I think that we should be clear that's what we're doing, and we need a few rules for how this whole thing will work.

I propose:
- `paramsRef` always stays for implementation-specific support and early-mover experiments. `infrastructure` is the way in which we _may_ graduate things from `paramsRef` to full fields.
- GatewayClass `infrastructure` has `defaults` and `overrides` like a Policy Attachment, and any field we add in one, we add in both. Additionally, any field we add here, we add to Gateway `infrastructure` as well.
- Gateway `infrastructure` may have additional fields that GatewayClass `infrastructure` doesn't, which are things that don't make sense to be defaulted or overridden (like `attachTo` from the merging GEP).
- we include _one_ field in this GEP: `annotations`, which will be the annotations to be used on any resources generated to fulfill the Gateway contract. This will be included in the GatewayClass `infrastructure`, in both `defaults` and `overrides`, as well as the Gateway `infrastructure`. We can then _show_ how `defaults` and `overrides` work using this example. (For example, we'll need a Condition on Gateway to say if Gateway-level `infrastructure` has been overridden by GatewayClass-level `infrastructure`.)

This way, we can
- create the `infrastructure` stanza with clarity about how it works
- provide a path for `paramsRef` settings to graduate to full fields on GatewayClass
- have a concrete illustration of how `defaults` and `overrides` can work in practice

geps/gep-1867.md
```diff
@@ -0,0 +1,134 @@
...
+  // rather than overrides.
+  Infrastructure GatewayClassInfrastructure `json:"infrastructure"`
+  // ...
+}
+
+type GatewayInfrastructure struct {
+  // ParametersRef provides a arbitrary implementation-specific configuration for
+  // fields not expressed directly in this struct.
+  // This follows the same semantics as GatewayClass's ParametersRef, but lives on the Gateway.
+  ParametersRef ParametersReference
```
robscott, line 108: Can you copy the relevant spec from GWC here? I think we may need a bit more than that written to also clarify what this is intended to be used for. This could become a very generic Gateway extension mechanism for things that aren't exactly infra-related (timeouts, retries, auth, etc). I don't think that's intended here, but if we don't clearly define what is meant to be configured here, it may end up getting used in unintended ways.

https://github.com/kubernetes-sigs/gateway-api/blob/4738c6c6e981d4a0e38c8c6120dc0198e4128351/apis/v1beta1/gatewayclass_types.go#L83-L96

geps/gep-1867.md
```diff
@@ -0,0 +1,134 @@
...
+However, there are still a variety of concepts that have some meaning between implementations that can provide value to users.
+
+Introduction at Gateway or GatewayClass level will depend on the specific field and use cases for the field.
+In general, it makes sense to provide defaults (GatewayClass) and specific settings (Gateway) for most fields, but
+this will be evaluated on a case-by-case basis.
+
+### Status
+
+The API should likely expose some status. However, it is not yet clear what that will look like.
+This will be addressed prior to promotion beyond "Provisional".
```
robscott, line 134: Can you add an "unresolved" keyword somewhere in here just so we don't miss this prior to promotion?

robscott:
Thanks for all the updates @howardjohn! Mostly agree with @youngnick's comment, but have a couple slight disagreements:

>GatewayClass infrastructure has defaults and overrides like a Policy Attachment, and any field we add in one, we add in both. Additionally, any field we add here, we add to Gateway infrastructure as well.

I think that the options we expose on GWC-infrastructure should be specific to each concept. For example, I think giving a min and max may be useful in some cases. As long as we're clear that that's a possibility for fields in the future, I don't think we need to solve it here.

> we include one field in this GEP: annotations, which will be the annotations to be used on any resources generated to fulfill the Gateway contract. This will be included in the GatewayClass infrastructure, in both defaults and overrides, as well as the Gateway infrastructure. We can then show how defaults and overrides work using this example. 

I think I actually like the most recent iteration of this GEP which only adds infrastructure and paramsRef fields and leaves the rest to future GEPs. I think we'd need at least one of those future GEPs to merge before we could mark this one as implementable though. I think even something as simple as annotations could open another can of worms that doesn't necessarily need to delay the simplified version of this GEP.

youngnick:
Ah, so the structure you're suggesting would be more like (using the annotations example even though I agree it's probably too hard for inclusion here, as well as some numerical thing):
```yaml
  spec:
    infrastructure:
      annotations:
        defaults:
          <a map[string]string>
        overrides:
          <a map[string]string>
      numericalThing:
        max: 1000
        min: 1
        default: 30
```

Or something?

I can agree with that, although this PR might benefit from some arbitrary examples like the `numericalThing` one here to illustrate how this should work.

To be clear, I think I agree with leaving annotations out of this, it's probably more complicated than it seems, just like everythig else. Sigh.

howardjohn:
I updated this to include that we may want ranges or default+override, on a case by case basis

robscott:
Thanks @howardjohn! I'd still like to see a goal related to GatewayClass (https://github.com/kubernetes-sigs/gateway-api/pull/1868/files#r1204455976) but otherwise this LGTM.

/approve

youngnick:
This LGTM for provisional to me.

/lgtm

shaneutt (approved):
I'm OK with moving forward with this as `Provisional` :+1: 

/approve

mikemorris:
Should this be promoted to `Experimental` (it appears that the `Implementable` GEP status is either gone or has nothing in it currently) now that [GEP-1762: In Cluster Gateway Deployments](https://gateway-api.sigs.k8s.io/geps/gep-1762/) using this field has merged as `Experimental`?

robscott:
Good catch @mikemorris! It looks like this just needs a navigation update? https://gateway-api.sigs.k8s.io/geps/gep-1867/

PR #1869, opened 2023-03-22
conformance: listener status can be in any order

skriss:
**What type of PR is this?**
/area conformance

Allows listener statuses to be in any order.

```release-note
NONE
```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -636,12 +636,17 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
 		return false
 	}
 
-	// TODO(mikemorris): Allow for arbitrarily ordered listeners
-	for i, eListener := range expected {
-		aListener := actual[i]
-		if aListener.Name != eListener.Name {
-			t.Logf("Name doesn't match")
-			return false
+	for _, eListener := range expected {
+		var aListener *v1beta1.ListenerStatus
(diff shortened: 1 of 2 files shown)


conformance/utils/kubernetes/helpers.go
```diff
@@ -636,12 +636,17 @@ func listenersMatch(t *testing.T, expected, actual []v1beta1.ListenerStatus) boo
...
+		var aListener *v1beta1.ListenerStatus
+		for i := range actual {
+			if actual[i].Name == eListener.Name {
+				aListener = &actual[i]
+				break
+			}
+		}
+		if aListener == nil {
+			t.Logf("Expected status for listener %s to be present", eListener.Name)
+			continue
```
shaneutt, line 649: If we're looking at an expected listener, and the named listener is not even _present_ in the list of actual listeners, then there's no reason to continue as we can't possibly match, right?

```suggestion
			t.Logf("Expected status for listener %s to be present", eListener.Name)
			return false
```

shaneutt (changes requested):

conformance/utils/kubernetes/helpers.go
skriss: :+1: https://github.com/kubernetes-sigs/gateway-api/pull/1869/commits/12ce05f724f19d75d569b6a806057a673697d90d

shaneutt (approved):
/lgtm

PR #1878, opened 2023-03-25
Initial Mesh (GAMMA) conformance tests

### LINKED ISSUE #1340: conformance: support testing mesh implementations
Currently, the conformance tests use a pretty simple `echoserver` deployment and make HTTP calls from the go test runner to the Gateway IP (which must be reachable from the test runner).

For testing mesh implementations (and in general, but especially mesh), this may not be sufficient. Since we are testing calls to in-cluster Services, we typically cannot reach these from the test runner (out of the cluster). Even mechanism like port-forward typically do not work as they traverse a different network path than real in-cluster calls.

The conformance tests should be enhanced to support these.

howardjohn:
**What type of PR is this?**
/area conformance

This PR introduces the basic scaffolding for Mesh testing. Mesh testing is fundamentally different from ingress testing, as we need to send requests from pod-to-pod instead of "go test" to "reachable IP address".

This PR introduces the basic scaffolding to run this test. With this PR, you can run a test against GAMMA tests.

This will:
* Deploy a new namespace and a few applications
* Run tests against them. Currently, there is only a trivial test that ensures applications can talk to each other; this PR is just intended to get the base infra, not add useful tests (yet)

Architecturally, we are deploying `gcr.io/istio-testing/app:latest` to each app. This is a client+server pair. We then on-demand `exec` into each pod to run `client http://some-request`, and parse the result and validate against expectations.

Likely desired improvements:
* Move image into k8s org, do not use latest
* Move to use a long-lived port-forward to each application; exec has a high overhead. The echo app has a grpc server we can control it from (instead of exec, we send it a ForwardMessage request). However, this has a downside that it requires us to pull in gRPC.
* Make code a bit more robust, etc

Opening this PR to get early feedback before moving further

Fixes #1340

```release-note
NONE
```


shaneutt (changes requested):

shaneutt (changes requested):

conformance/conformance_test.go
```diff
@@ -63,7 +63,16 @@ func TestConformance(t *testing.T) {
 		EnableAllSupportedFeatures: *flags.EnableAllSupportedFeatures,
 	})
 	cSuite.Setup(t)
-	cSuite.Run(t, tests.ConformanceTests)
+
+	var toRun []suite.ConformanceTest
+	if supportedFeatures.Has(suite.SupportGateway) {
+		toRun = append(toRun, tests.ConformanceTests...)
+	}
```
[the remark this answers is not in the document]
howardjohn: I would prefer to go with (2) to decouple things. I don't mind churn in the Mesh test space much (currently only Istio has an impl anyways, so all the pain is consumed by myself - linkerd soon to follow I hope) but don't want to couple it w/ the rest. That being said, I can also help out there as a separate PR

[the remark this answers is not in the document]
shaneutt: That's perfectly fine. I've updated the title and description of #1891 to better reflect and include this context. I've also marked it for `v0.7.0` as a release blocker and I'll intend to get it done today or tomorrow (as it was something I was already working on as part of a bigger branch).

Please add a TODO comment indicating we expect this stuff to change soon with a link to #1891 and then I consider this comment resolved.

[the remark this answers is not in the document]
shaneutt: @howardjohn https://github.com/kubernetes-sigs/gateway-api/pull/1894 is the follow up we discussed in support of #1891 which will require features for ALL tests eliminating the need for separate test lists and in support of the upcoming #1709.

My intention is to rebase it after this merges as a follow up PR, but if you'd like to simply incorporate it and pull it into your PR that would be fine too and we can mark this as a rebase merge to make sure they don't get squashed together.

keithmattix (approved):
I personally don't have any reservations about this, but I'll let someone else be the final LGTM

conformance/mesh/manifests.yaml
```diff
@@ -0,0 +1,130 @@
...
+      version: v1
+  template:
+    metadata:
+      labels:
+        app: echo
+        version: v1
+    spec:
+      containers:
+      - name: echo
+        image: gcr.io/istio-testing/app:latest
```
robscott, line 30: I'd rather have this image owned by k8s/gateway-api infra so we can bundle versions of it with each release of our conformance tests. Can you create an issue to track following up on that? Shorter term can we at least use a pinned version of this image so a change in this image doesn't have an immediate impact on our conformance tests?

conformance/tests/mesh-split.go
```diff
@@ -0,0 +1,70 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2: A lot of these files have old copyright dates, are they just being copied from Istio codebase?

howardjohn: copied from other files in this repo

conformance/mesh/manifests.yaml
howardjohn: https://github.com/kubernetes-sigs/gateway-api/issues/1895 and moved to a fixed image tag

michaelbeaumont (approved):
Great! LGTM

shaneutt:
Alright, enough people have seen this now so we can

/unhold

howardjohn:
Thanks all. Need 1 lgtm to get this in

kflynn:
Looks good from here, FTR.

robscott:
Thanks @howardjohn!

/lgtm

Issue #1805, opened 2023-03-10
Update conformance tests to handle the update in port redirect semantics


arkodg:
**What would you like to be added**:
With an update in the API around port redirect semantics with https://github.com/kubernetes-sigs/gateway-api/pull/1740, the conformance tests pertaining to port redirect do not match this behavior and need to be updated [here](https://github.com/kubernetes-sigs/gateway-api/blob/227de0756d763b320cb4dcfa16a668867e326554/conformance/utils/http/http.go#L340)

Here are the additional cases that need to be addressed
* Port specified in the redirect, and it’s not 80 or 443
* Port specified in the redirect and it is 80 or 443
* No port specified, Listener is 80 or 443
* No port specified, Listener is not 80 or 443

**Why this is needed**:
To ensure conformance test are evaluating the API correctly

PR #1880, opened 2023-03-27
Clarify port redirect behaviour and add conformance tests for the same

gauravkghildiyal:
### **What type of PR is this?**
/kind documentation
/area conformance
/kind test

### **What this PR does / why we need it**:
Clarify port redirect behaviour and add conformance tests for the same:
```
// When empty, the redirect port MUST be derived using the following rules:
//
// * If redirect scheme is not-empty, the redirect port MUST be the well-known
//   port associated with the redirect scheme. Specifically "http" to port 80
//   and "https" to port 443. If the redirect scheme does not have a
//   well-known port, the listener port of the Gateway SHOULD be used.
// * If redirect scheme is empty, the redirect port MUST be the Gateway
//   Listener port.
```

Proposal doc: https://docs.google.com/document/d/1fx8kuWh2bmBge9IRGn7NlhGqTP0Kj21Cz5RZyUW6428

### **Which issue(s) this PR fixes**:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1805
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1909

### **Does this PR introduce a user-facing change?**:
```release-note
Port redirect when empty will depend on the configured Redirect scheme
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -855,6 +855,9 @@ type HTTPRequestRedirectFilter struct {
 	// Scheme is the scheme to be used in the value of the `Location` header in
 	// the response. When empty, the scheme of the request is used.
 	//
+	// Scheme redirects can affect the port of the redirect, for more information,
+	// refer to the documentation for the port field of this filter.
+	//
 	// Note that values may be added to this enum, implementations
 	// must ensure that unknown values will not cause a crash.
 	//
@@ -889,7 +892,15 @@ type HTTPRequestRedirectFilter struct {
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
 	//
-	// When empty, the Gateway Listener port is used.
+	// If no port is specified, the redirect port MUST be derived using the
+	// following rules:
+	//
+	// * If redirect scheme is not-empty, the redirect port MUST be the well-known
+	//   port associated with the redirect scheme. Specifically "http" to port 80
+	//   and "https" to port 443. If the redirect scheme does not have a
+	//   well-known port, the listener port of the Gateway SHOULD be used.
+	// * If redirect scheme is empty, the redirect port MUST be the Gateway
+	//   Listener port.
 	//
 	// Implementations SHOULD NOT add the port number in the 'Location'
 	// header in the following cases:
--- a/conformance/tests/httproute-redirect-port-and-scheme.go
+++ b/conformance/tests/httproute-redirect-port-and-scheme.go
@@ -1,0 +1,303 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/roundtripper"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+	"sigs.k8s.io/gateway-api/conformance/utils/tls"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteRedirectPortAndScheme)
+}
+
+var HTTPRouteRedirectPortAndScheme = suite.ConformanceTest{
+	ShortName:   "HTTPRouteRedirectPortAndScheme",
+	Description: "An HTTPRoute with port and scheme redirect filter",
+	Manifests:   []string{"tests/httproute-redirect-port-and-scheme.yaml"},
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportHTTPRoute,
+		suite.SupportHTTPRoutePortRedirect,
+	},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		ns := "gateway-conformance-infra"
+
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		routeNN := types.NamespacedName{Name: "http-route-for-listener-on-port-80", Namespace: ns}
+		gwAddr80 := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		gwNN = types.NamespacedName{Name: "same-namespace-with-http-listener-on-8080", Namespace: ns}
+		routeNN = types.NamespacedName{Name: "http-route-for-listener-on-port-8080", Namespace: ns}
+		gwAddr8080 := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		gwNN = types.NamespacedName{Name: "same-namespace-with-https-listener", Namespace: ns}
+		routeNN = types.NamespacedName{Name: "http-route-for-listener-on-port-443", Namespace: ns}
+		gwAddr443 := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		certNN := types.NamespacedName{Name: "tls-validity-checks-certificate", Namespace: string(ns)}
+		cPem, keyPem, err := GetTLSSecret(suite.Client, certNN)
+		if err != nil {
+			t.Fatalf("unexpected error finding TLS secret: %v", err)
+		}
+
+		// NOTE: In all the test cases, a missing value of expected Port within
+		//   RedirectRequest implies that it is still valid and acceptable for a
+		//   port to be specified in the redirect if it corresponds to the scheme
+		//   (80 and 443).
+
+		////////////////////////////////////////////////////////////////////////////
+		// Test cases that use http-route-for-listener-on-port-80
+		////////////////////////////////////////////////////////////////////////////
+
+		testCases := []http.ExpectedResponse{
+			{
+				Request: http.Request{
+					Path:             "/scheme-nil-and-port-nil",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "http",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-nil-and-port-80",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "http",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-nil-and-port-8080",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "http",
+					Port:   "8080",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-https-and-port-nil",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "https",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-https-and-port-443",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "https",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-https-and-port-8443",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "https",
+					Port:   "8443",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+		}
+
+		for i := range testCases {
+			tc := testCases[i]
+			t.Run("http-listener-on-80/"+tc.GetTestCaseName(i), func(t *testing.T) {
+				t.Parallel()
+				http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr80, tc)
+			})
+		}
+
+		////////////////////////////////////////////////////////////////////////////
+		// Test cases that use same-namespace-with-http-listener-on-8080
+		////////////////////////////////////////////////////////////////////////////
+
+		testCases = []http.ExpectedResponse{
+			{
+				Request: http.Request{
+					Path:             "/scheme-nil-and-port-nil",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "http",
+					Port:   "8080",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
+					Path:             "/scheme-nil-and-port-80",
+					UnfollowRedirect: true,
+				},
+				Response: http.Response{StatusCode: 302},
+				RedirectRequest: &roundtripper.RedirectRequest{
+					Scheme: "http",
+					Host:   "example.org",
+				},
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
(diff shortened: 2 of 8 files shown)


conformance/tests/httproute-redirect-port.go
```diff
@@ -43,57 +43,97 @@ var HTTPRouteRedirectPort = suite.ConformanceTest{
...
-			Response: http.Response{
-				StatusCode: 302,
+			{
+				// Port should not be expected in the Location header of redirect
+				// response since:
+				// 1. Scheme is defaulted to 'http', and
+				// 2. Redirect port is specified as 80, which is redundant when paired
+				// 		with Scheme.
+				Request: http.Request{
+					Path:             "/port-80",
```
howardjohn, line 108: It would be ideal to have all 4 combos of {http,https}x{80,443}

robscott:
Thanks @gauravkghildiyal!

/ok-to-test

conformance/tests/httproute-redirect-port.go
```diff
@@ -43,57 +43,97 @@ var HTTPRouteRedirectPort = suite.ConformanceTest{
...
 			},
-			RedirectRequest: &roundtripper.RedirectRequest{
-				Host: "example.org",
-				Port: "8083",
+			{
+				// Port should not be expected in the Location header of redirect
+				// response since:
+				// 1. Redirect port was not specified.
+				// 2. Scheme defaulted to 'http' which means that a port is already
+				// 		implied.
```
shaneutt, line 124: Extra comments to highly the purpose of the test are appreciated :+1:

conformance/utils/http/http.go
```diff
@@ -344,14 +344,27 @@ func setRedirectRequestDefaults(req *roundtripper.Request, cRes *roundtripper.Ca
...
 
-	if expected.RedirectRequest.Port == "" {
-		expected.RedirectRequest.Port = req.URL.Port()
-	}
-
 	if expected.RedirectRequest.Scheme == "" {
 		expected.RedirectRequest.Scheme = req.URL.Scheme
 	}
 
+	if expected.RedirectRequest.Port == "" {
```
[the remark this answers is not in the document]
youngnick: If it's redirecting to `https` scheme on port `80` it should definitely include the port though, since `https` should only skip the port if it's `443`.

conformance/utils/http/http.go
```diff
@@ -344,14 +344,27 @@ func setRedirectRequestDefaults(req *roundtripper.Request, cRes *roundtripper.Ca
...
+		func() {
```
youngnick, line 352: Also, why wrap this in an anonymous function? I don't see the advantage?

conformance/tests/httproute-redirect-port.go
gauravkghildiyal: Definitely. Changed.

conformance/utils/http/http.go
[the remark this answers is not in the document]
gauravkghildiyal: Agree with Nick's explanation. The test seems to be working as expected.

conformance/utils/http/http.go
gauravkghildiyal: Perfectly valid question. Maybe it was just a personal preference - I felt that it made the code more readable and matched the way in which we've described this condition in our API reference:
image


```
If http with port 80 -> do nothing

If https with 443 -> do nothing

If none of the previous conditions -> set default value.
```

But yeah, I've changed it to a different format while still trying to avoid complex/long if/else. Hope this is more generally appealing.

conformance/utils/http/http.go
```diff
@@ -344,14 +344,26 @@ func setRedirectRequestDefaults(req *roundtripper.Request, cRes *roundtripper.Ca
...
+	if expected.RedirectRequest.Port == "" {
+		skipSettingDefaultPort := false
+
+		// No port value should be set if the redirect location uses "http" scheme
+		// and the gateway port is "80".
+		skipSettingDefaultPort = skipSettingDefaultPort || strings.ToLower(expected.RedirectRequest.Scheme) == "http" && req.URL.Port() == "80"
+
+		// No port value should be set if the redirect location uses "https" scheme
+		// and the gateway port is "443".
+		skipSettingDefaultPort = skipSettingDefaultPort || strings.ToLower(expected.RedirectRequest.Scheme) == "https" && req.URL.Port() == "443"
```
youngnick, line 360: The `skipsettingDefaultPort ||` part has no effect, because `skipsettingDefaultPort` is always false. To be honest, I think it would be more readable and idiomatic to do this with two `ifs`:
```suggestion

		if (strings.ToLower(expected.RedirectRequest.Scheme) == "http" && req.URL.Port() == "80") {
			skipSettingDefaultPort = true
		}
		
		if (strings.ToLower(expected.RedirectRequest.Scheme) == "https" && req.URL.Port() == "443") {
			skipSettingDefaultPort = true
		}
```
That indicates pretty clearly what we're doing, at the cost of some `if` statements. We shouldn't ever hit both `ifs` in the same case anyway.

youngnick (changes requested):
One more small change from me, then this LGTM.

conformance/utils/http/http.go
gauravkghildiyal: Sounds good Nick. Changed. Thanks :)

conformance/utils/http/http.go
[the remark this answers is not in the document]
howardjohn: I think it has been updated since my original comment. But I am seeing other failures so I would like some time to validate if Istio impl is broken or if the tests are unexpected. Will update today

conformance/tests/httproute-redirect-port.go
howardjohn: Also if we can add with `Host: IP` and `Host: IP:port` could be helpful (saying this as an implementation that is broken for one but not the other)

arkodg:
> As far as I can tell this cannot currently be implemented by Envoy as is, meaning most implementations cannot be conformant. Envoy has 2 modes:
> 
> * Set explicit port
> * Derive port from Host
> 
> But we have a test that `Host: ip:80` results in `Location: http://ip`. Envoy doesn't have a way to say "set no port", as far as I can tell?
> 
> I think we need either [envoyproxy/envoy#25573](https://github.com/envoyproxy/envoy/pull/25573) or [envoyproxy/envoy#26516](https://github.com/envoyproxy/envoy/issues/26516)

The simplest way for to make some progress now would be to set `Host: ip` instead of `Host: ip:80` when sending requests for this test ?

arkodg:
> @arkodg @howardjohn It sounds like what we actually want here is to modify the spec to say that port should not be added if it was not already present in the host header, does that sound right? Current spec for reference:
> 
> https://github.com/kubernetes-sigs/gateway-api/blob/8dde50805b8010f6ec52ae52e7d47d799cdcaabf/apis/v1beta1/httproute_types.go#L897-L902
> 
> Given that the spec says "should" and we avoid conformance testing "should", we should probably update the tests to either allow both options (port specified or not) or use different ports. I think the safest option here is to allow both options.

Today the spec is being specific about adding or not appending the port to the existing `Host` string i.e.
Location = Host(all of it) + ":" + Port (not always/depends on above statement).
To make some progress I propose
* Update the conformance client  to not append `port` into the `Host` header
* Follow up with a future GH issue with -  should the `port` from the `Host` header be stripped before appending it to the `Location` header, and update spec if clarification is needed in API

pleshakov:
@robscott 

> It sounds like what we actually want here is to modify the spec to say that port should not be added if it was not already present in the host header, does that sound right?

Some perspective from a non-Envoy implementation, NGINX: I don't think it is possible to extract the port from the host header, so as a result, we can't include it in the resulting `Location` header.

gauravkghildiyal:
> To make some progress I propose: Update the conformance client to not append port into the Host header

Is everyone fine with this for the time being? Just so I understood things correctly, given we make the request with "Host: IP" only (without any port), envoy based implementations will be able to skip setting any port in the `Location` header in both the cases where: (?)
1. `redirectPort` is not set
2. `redirectPort` is set to 80

Aside: I'm assuming istio still needs to configure things to make the above work -- since when I ran a test against the latest master from istio today, I could still observe the port being set even when the `Host` did not include any port. Precisely:
Given the HTTPRequestRedirectFilter
```yaml
  - matches:
    - path:
        type: PathPrefix
        value: /http-listener-on-80-and-redirect-port-not-set
    filters:
    - type: RequestRedirect
      requestRedirect:
        statusCode: 301
```
And the request made was as below, I still see the port in the response.
```bash
❯ curl 34.136.70.31/http-listener-on-80-and-redirect-port-not-set -v -H "Host: abc"

*   Trying 34.136.70.31:80...
* Connected to 34.136.70.31 (34.136.70.31) port 80 (#0)
> GET /http-listener-on-80-and-redirect-port-not-set HTTP/1.1
> Host: abc
> User-Agent: curl/7.86.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< location: http://abc:80/http-listener-on-80-and-redirect-port-not-set
< date: Tue, 04 Apr 2023 07:42:55 GMT
< server: istio-envoy
< content-length: 0
<
* Connection #0 to host 34.136.70.31 left intact
```

---

Or, should we instead update the spec to something along the lines of?

```
Implementations should make sure that:
* If redirect port is explicitly specified, then the same port value should also
  be present in the Location header of the response.
* If redirect port is _not_ specified, presence of port value in the Location
  header of response is optional. If the implementation decides to include a
  port value in the Location header, then this port should correspond to a
  well-known port number derived from the Location header scheme.
```

michaelbeaumont:
> As far as I can tell this cannot currently be implemented by Envoy as is, meaning most implementations cannot be conformant. Envoy has 2 modes:
> 
> * Set explicit port
> * Derive port from Host
> 
> But we have a test that `Host: ip:80` results in `Location: http://ip`. Envoy doesn't have a way to say "set no port", as far as I can tell?
> 
> I think we need either [envoyproxy/envoy#25573](https://github.com/envoyproxy/envoy/pull/25573) or [envoyproxy/envoy#26516](https://github.com/envoyproxy/envoy/issues/26516)

I can confirm Kuma will have the same problem.

Interestingly enough, I _was_ able to get rid of the port in that case and get all tests to pass but only because we use `strip_any_host_port` in our `http_connection_manager` and [there is probably buggy behavior currently where `strip_any_host_port` also strips ports from redirected URLs](https://github.com/envoyproxy/envoy/issues/19589).

> Is everyone fine with this for the time being? Just so I understood things correctly, given we make the request with "Host: IP" only (without any port), envoy based implementations will be able to skip setting any port in the Location header in both the cases where: (?)

> 1. `redirectPort` is not set
> 1. `redirectPort` is set to 80

Yeah, this should be possible with Kuma.

howardjohn:
> strip_any_host_port

This, IMO, cannot be forced (and possible should be non-conformant). It modifies the actual user's requests. This may not matter for redirect, but if you are using it you are very likely applying it to all traffic not just redirects.

> Aside: I'm assuming istio still needs to configure things to make the above work 

Yeah I have a local fork with some changes

gauravkghildiyal (commented):
I've updated things as per latest discussion (and enroute changes in https://github.com/kubernetes-sigs/gateway-api/pull/1908) . I feel the tests now at least cover the "well defined" part of the port redirect behaviour. There's still some ambiguities (like https://github.com/kubernetes-sigs/gateway-api/issues/1909) and we'll take that up separately.

conformance/tests/httproute-redirect-port.go
```diff
@@ -38,70 +39,172 @@ var HTTPRouteRedirectPort = suite.ConformanceTest{
...
-			Namespace: ns,
-		}, {
-			Request: http.Request{
-				Path:             "/port-and-host-and-status",
-				UnfollowRedirect: true,
+			{
+				// When requestRedirect.Port is not set and the interpreted scheme is
+				// http, port 80 should not be ADDED to the response Location header.
+				// (The Location header MAY still result in having a port if the request
+				// Host header contains one.)
```
robscott, line 146: In general we avoid adding conformance tests for "SHOULD" because we're using [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) terminology here where should is a recommendation and not a requirement. In this case I think it's acceptable to have a conformance test here if it accepts the port either being specified or unspecified, but more specifically, it must not be a different value like 81 or something.

If I'm understanding this specific test case correctly, it's only checking the status code. I think we probably want to check for more than that here.

conformance/tests/httproute-redirect-port.go
```diff
@@ -38,70 +39,172 @@ var HTTPRouteRedirectPort = suite.ConformanceTest{
...
+					Namespace:       ns,
+				},
 			},
-			Response: http.Response{
-				StatusCode: 302,
+			{
+				// When requestRedirect.Port is set to 443 and the interpreted scheme is
+				// https, port 443 should not be ADDED to the response Location header.
+				// (The Location header MAY still result in having a port if the request
+				// Host header contains one.)
```
robscott, line 163: This comment makes sense but I'm either misunderstanding the test case or it's just not testing/verifying the same level of detail as the comment suggests. The same comment applies to several test cases, though I could just be missing something obvious.

conformance/tests/httproute-redirect-port.go
```diff
@@ -38,70 +39,172 @@ var HTTPRouteRedirectPort = suite.ConformanceTest{
...
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		certNN := types.NamespacedName{Name: "tls-validity-checks-certificate", Namespace: string(ns)}
+		cPem, keyPem, err := GetTLSSecret(suite.Client, certNN)
+		if err != nil {
+			t.Fatalf("unexpected error finding TLS secret: %v", err)
+		}
+
+		testCases := []struct {
+			expectedResponse http.ExpectedResponse
```
robscott, line 60: I'm not sure I understand this, does this mean the request is now embedded in the expected response?

robscott (commented):
Thanks @gauravkghildiyal!

gauravkghildiyal:
/hold

Lets get the spec well defined first. Will come back to this after tackling https://github.com/kubernetes-sigs/gateway-api/issues/1909

shaneutt:
Considering draft for now as active review is not needed until https://github.com/kubernetes-sigs/gateway-api/issues/1909 is complete.

conformance/tests/httproute-redirect-port.go
gauravkghildiyal: The comment may be obsolete but yes I may share that opinion.

If we're talking about the same thing, then yes I do find the existing naming of the struct `ExpectedResponse` confusing -- it does embed the request to be made (which is counterintuitive from the name)

conformance/tests/httproute-redirect-port.go
gauravkghildiyal: Obsolete

conformance/tests/httproute-redirect-port.go
gauravkghildiyal: Thanks. I've tried to follow this with the newer changes.

conformance/tests/httproute-redirect-port.yaml
```diff
@@ -43,3 +43,67 @@ spec:
...
+      requestRedirect:
+        port: 443
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /https-listener-on-443-and-redirect-port-not-set
+    filters:
+    - type: RequestRedirect
+      requestRedirect:
+        statusCode: 301
```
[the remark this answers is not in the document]
gauravkghildiyal: Obsolete.

conformance/tests/httproute-redirect-port.yaml
```diff
@@ -43,3 +43,67 @@ spec:
...
+    protocol: HTTPS
+    allowedRoutes:
+      namespaces:
+        from: All
+    tls:
+      certificateRefs:
+      - group: ""
+        kind: Secret
+        name: tls-validity-checks-certificate
+        namespace: gateway-conformance-infra
```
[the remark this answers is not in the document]
gauravkghildiyal: I've taken a note of this with the newer changes.

conformance/tests/httproute-redirect-port-and-scheme.go
```diff
@@ -0,0 +1,298 @@
...
+		cPem, keyPem, err := GetTLSSecret(suite.Client, certNN)
+		if err != nil {
+			t.Fatalf("unexpected error finding TLS secret: %v", err)
+		}
+
+		////////////////////////////////////////////////////////////////////////////
+		// Test cases that use http-route-for-listener-on-port-80
+		////////////////////////////////////////////////////////////////////////////
+
+		testCases := []http.ExpectedResponse{
```
robscott, line 72: Nit: Somewhere in here, maybe above this section, it would be helpful to add a note clearly stating that wherever port is not specified by a test case, it is still valid and acceptable for a port to be specified in the redirect if it corresponds to the scheme (80 and 443). (Essentially the work you did in https://github.com/kubernetes-sigs/gateway-api/pull/1880/commits/fa0691a28576bdad7a163dff31d7d278747bbced).

robscott (commented):
Thanks @gauravkghildiyal! Mostly LGTM, just a few nits.

youngnick:
This LGTM, thanks everyone for patiently working through the detail here.

/approved

youngnick (approved):

howardjohn (approved):
/lgtm

robscott:
Will wait until tomorrow to remove the hold, but this LGTM as well. Thanks for all the work on this @gauravkghildiyal!

conformance/utils/http/http.go
```diff
@@ -135,6 +137,25 @@ func MakeRequest(t *testing.T, expected *ExpectedResponse, gwAddr, protocol, sch
 	return req
 }
 
+// calculateHost will calculate the Host header as per [HTTP spec]. To
+// summarize, host will not include ay port if it is implied from the scheme. In
```
zirain, line 141:
```suggestion
// summarize, host will not include any port if it is implied from the scheme. In
```

gauravkghildiyal: Thanks. Corrected.

apis/v1beta1/httproute_types.go
```diff
@@ -889,7 +892,14 @@ type HTTPRequestRedirectFilter struct {
 	// Port is the port to be used in the value of the `Location`
 	// header in the response.
 	//
-	// When empty, the Gateway Listener port is used.
+	// When empty, the redirect port MUST be derived using the following rules:
```
kflynn, line 895: Maybe change "When empty" to "If no port is specified", just to be that much more explicit?

This looks good to me, with that minor nit -- thanks!

gauravkghildiyal: Sounds good. Thanks for the suggestion :)

conformance/utils/http/http.go
```diff
@@ -135,6 +137,25 @@ func MakeRequest(t *testing.T, expected *ExpectedResponse, gwAddr, protocol, sch
...
+// summarize, host will not include any port if it is implied from the scheme. In
+// case of any error, the input gwAddr will be returned as the default.
+//
+// [HTTP spec]: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23
+func calculateHost(gwAddr, scheme string) string {
+	host, port, err := net.SplitHostPort(gwAddr)
+	if err != nil {
+		return gwAddr
+	}
+	if strings.ToLower(scheme) == "http" && port == "80" {
```
arkodg, line 150: shouldnt this be OR ?

arkodg: sure its fine from a request URL perspective, wanted to make sure that this field is not being used to compare response Location header, guess im mixing this with https://github.com/kubernetes-sigs/gateway-api/pull/1980 which I shouldnt, my bad, resolving, thanks

robscott:
Thanks to everyone for the quick reviews and feedback on this and to @gauravkghildiyal for the great work here! This was discussed at both Gateway API and GAMMA meetings this week and no significant blockers came up. Going to go ahead and merge and transition to RC2 prep.

/lgtm
/hold cancel

PR #1882, opened 2023-03-27
add option to specify http roundtripper while running conformance tests

### LINKED ISSUE #1496: Conformance RoundTripper should allow for custom http.RoundTripper implementation

akash-gautam:
**What type of PR is this?**
/area conformance

Fixes #1496 

No
```release-note
NONE

```

### DIFF
--- a/conformance/utils/roundtripper/roundtripper.go
+++ b/conformance/utils/roundtripper/roundtripper.go
@@ -84,8 +84,9 @@ type CapturedResponse struct {
 // DefaultRoundTripper is the default implementation of a RoundTripper. It will
 // be used if a custom implementation is not specified.
 type DefaultRoundTripper struct {
-	Debug         bool
-	TimeoutConfig config.TimeoutConfig
+	Debug            bool
+	TimeoutConfig    config.TimeoutConfig
+	HTTPRoundTripper http.RoundTripper
 }
 
 // CaptureRoundTrip makes a request with the provided parameters and returns the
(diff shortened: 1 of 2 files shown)


robscott:
Thanks @akash-gautam!

cc @briantkennedy

briantkennedy:
It looks like the original Issue (#1496) is no longer feasible due to the need for [setting a trusted CA certificate](https://github.com/kubernetes-sigs/gateway-api/blob/78ab14e85845c249cf5b014b2f4e769c67b128db/conformance/utils/roundtripper/roundtripper.go#L128).  I think we could still benefit from being able to configure the `Dial` member on the `http.Transport` resource ([reference](https://pkg.go.dev/net/http#Transport)).

akash-gautam:
@briantkennedy So should we close this PR and the related issues and open a new issue for configuring Dial member?

briantkennedy:
Yes @akash-gautam, that would be a simpler way to resolve.  Thanks!

PR #1886, opened 2023-03-27
Reserve the Ready condition for GW/Listener

### LINKED ISSUE #1877: Remove GatewayConditionReady
Related to https://github.com/kubernetes-sigs/gateway-api/issues/1832 but split out to be more clear

`GatewayConditionReady` 

> 	// Ready is an optional Condition that has Extended support. When it's set,
> 	// the condition indicates whether the Gateway has been completely configured
> 	// and traffic is ready to flow through the data plane immediately.

It was changed in https://gateway-api.sigs.k8s.io/geps/gep-1364/#ready, when we added a new Programmed condition and moved Ready to extended.

`GatewayConditionProgrammed` is a weaker form of `Ready`, indicating we have configured the data plane, but it may be eventually consistent.

The problem, IMO, is it is not useful to know a _Gateway_ is Ready. Gateway's do not provide functionality on their own; they are an attachment point for functionality (Route).

The typical use case for Ready is something like knative, where we `kubectl wait --for=ready` and then start sending traffic. But that cannot be used here, because all we know is the "Gateway" is ready - its not even clear what this means without routes.

We cannot change the spec to mean "Gateway and all attached routes are ready", as there is no way to verify this is up-to-date due to the cross-object references (typically this is solved via observedGeneration, but that applies only to single route).

As a result, I propose we remove the Ready condition entirely.

Alternative: we instead move Ready to a route condition

cc @youngnick @dprotaso @frankbu

howardjohn:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1877

**What type of PR is this?**

/kind documentation

Reviewers note: this PR used to take a different approach but has changed.

This PR "reserves" Ready, clarifies the language of Programmed, and allows it to be "Not Programmed because Addresses are not allocated".

This is not expected to cause churn on status, as no implementations currently implement Ready (at least not correctly). 

By marking it as Reserved, we are able to go to GA without a condition that has no usage (generally, no usage @ GA is not a great combo), but still have the ability to add it back if we find compelling reasons to in the future.

This leaves 2 conditions:
* Accepted: config is correct and control plane saw it
* Programmed: data plane is programmed and as "ready" as we can determine, but no guarantees 

Note: I intentionally didn't run the codegen to keep the PR small while we discuss, will obviously run it if we approve this.

Fixes #1877

```release-note
The `Ready` condition on listeners and gateways has been reserved. Users and implementations should use the `Programmed` condition.
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -517,9 +517,12 @@ type GatewayConditionType string
 type GatewayConditionReason string
 
 const (
-	// This condition indicates whether a Gateway has generated some
-	// configuration that is assumed to be ready soon in the underlying data
-	// plane.
+	// This condition indicates whether a Gateway has generated some configuration and programmed the
+	// underlying data plane implementation. Implementations should wait until resources are fully
+	// provisioned, to the best of their knowledge. This may include waiting for load balancers to be
+	// provisioned, waiting for addresses to be allocated, or waiting until configuration has been
+	// received by the data plane. However, as implementations may be eventually consistent, this does
+	// not guarantee the gateway is *immediately* able to recieve traffic.
 	//
 	// It is a positive-polarity summary condition, and so should always be
 	// present on the resource with ObservedGeneration set.
@@ -536,6 +539,7 @@ const (
 	//
 	// * "Invalid"
 	// * "Pending"
+	// * "AddressNotAssigned"
 	//
 	// Possible reasons for this condition to be Unknown are:
 	//
@@ -553,6 +557,13 @@ const (
 	// This reason is used with the "Programmed" condition when the Listener is
 	// syntactically or semantically invalid.
 	GatewayReasonInvalid GatewayConditionReason = "Invalid"
+
+	// This reason is used with the "Programmed" condition when none of the requested
+	// addresses have been assigned to the Gateway. This reason can be used to
+	// express a range of circumstances, including (but not limited to) IPAM
+	// address exhaustion, invalid or unsupported address requests, or a named
+	// address not being found.
+	GatewayReasonAddressNotAssigned GatewayConditionReason = "AddressNotAssigned"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


robscott:
I don't think we can justify switching status conditions _again_. I personally appreciate the distinction that "Programmed" is not necessarily "Ready" and that we're leaving room in the API for underlying implementations to be able to support a true "Ready" condition. I agree that we can/should update our docs to make this clearer though.

If we were to make the change proposed here I can guarantee that we'd get issues that "My Gateway is never becoming Programmed". Every transition/rename has a cost, I don't think we can justify the significant cost of switching everyone back to the old terms.

/hold

apis/v1beta1/gateway_types.go
```diff
@@ -607,9 +581,18 @@ const (
 )
 
 const (
-	// Ready is an optional Condition that has Extended support. When it's set,
-	// the condition indicates whether the Gateway has been completely configured
-	// and traffic is ready to flow through the data plane immediately.
+	// This condition indicates whether a Gateway has generated some
+	// configuration and programmed the underlying data plane.
```
arkodg, line 585:
```suggestion
	// configuration, programmed it in the underlying data plane and is ready to be used.
```

howardjohn:
> I don't think we can justify switching status conditions _again_. I personally appreciate the distinction that "Programmed" is not necessarily "Ready" and that we're leaving room in the API for underlying implementations to be able to support a true "Ready" condition. I agree that we can/should update our docs to make this clearer though.
> 
> If we were to make the change proposed here I can guarantee that we'd get issues that "My Gateway is never becoming Programmed". Every transition/rename has a cost, I don't think we can justify the significant cost of switching everyone back to the old terms.
> 
> /hold

That leaves 2 options (that satisfies my desires anyhow):
* Remove Ready (this PR but invert which one we deprecate)
* Keep Ready but allow it to not mean " has programmed the proxy immediately " which no implementation supports. But then I don't know when you would set a different value than Programmed.

Are you suggesting the latter?

apis/v1beta1/gateway_types.go
```diff
@@ -607,9 +581,18 @@ const (
...
+	//
+	// Most implementations are eventually consistent, providing different
```
arkodg, line 587: prefer if this more prescriptive, for e.g.
```
Here is an example of some of the guarantees an implementation can provide when the Ready condition is set
* Address has been allocated
* Configuration has been programmed and acknowledged by the data plane.
```

howardjohn:
@shaneutt @robscott I updated the PR entirely, I think it addresses your concerns. Please take another look.

howardjohn:
will make a new PR

howardjohn:
I moved this to https://github.com/kubernetes-sigs/gateway-api/pull/1888

Issue #1832, opened 2023-03-16
Consistent status reasons


howardjohn:
GatewayConditionReady is extended. It has a set of reasons:
	// * "ListenersNotValid"
	// * "ListenersNotReady"
	// * "AddressNotAssigned"

ListenerConditionProgrammed is core, but it only has a few reasons:
	// * "Invalid"
	// * "Pending"

Given these represent almost the same thing, it seems like these should be the same set, or GatewayConditionReady should be a superset 

similar concern on Listener status

robscott:
Good catch, agree that these should all be present for both conditions.

/help

howardjohn:
Also I think the spec is pretty clear but while I am here to double check... if I cannot say "traffic is ready to flow immediately" I should just not set Ready at all, right? And only use Programmed?

arkodg:
> Also I think the spec is pretty clear but while I am here to double check... if I cannot say "traffic is ready to flow immediately" I should just not set Ready at all, right? And only use Programmed?

+1 to this, helps users understand the difference between CP programmed and DP ready
cc @dprotaso

howardjohn:
So my interpretation means:

* Accepted: I got far enough to start sending *some* data to the data plane
* Programmed: I got far enough to start send *all* the data to the data plane
* Ready (optional): I sent *all* the data to the data plane *and* it fully processed it and is ready to go

robscott:
>Also I think the spec is pretty clear but while I am here to double check... if I cannot say "traffic is ready to flow immediately" I should just not set Ready at all, right? And only use Programmed?

Yep, the requirements for setting ready have become pretty strict:

https://github.com/kubernetes-sigs/gateway-api/blob/1c4ae619312754c52e1ff130439caed87132dc54/apis/v1beta1/gateway_types.go#L609-L611

If there were some kind of delay after which you could guarantee readiness that may also be a good reason to set "Ready", but not sure if that's a good idea.

robscott:
@gauravkghildiyal has volunteered to work on this one

arkodg:
would be good to add conformance tests to make sure if Route reports status as `Ready`, all requests through it must succeed

robscott:
@arkodg completely agree, mind creating an issue to track that?

gauravkghildiyal:
@howardjohn The issue says that "**Listener**ConditionProgrammed" should be subset of "**Gateway**ConditionReady". 

[...]

PR #1888, opened 2023-03-28
Update Gateway/Listener status for consistency

howardjohn:
**What type of PR is this?**

/kind documentation

Supercedes https://github.com/kubernetes-sigs/gateway-api/pull/1886
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1832
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1877

This PR attempts to introduce a bit more consistency into Gateway conditions, while reducing churn. There are 4 changes here:

* The "Ready" Gateway and Listener condition has been reserved for future use.

This condition has no implementations; while it may have some future value, we cannot justify the additional reason going GA without usage. Additionally, it has caused confusion by the subtle distinction between Programmed and Ready, and has "stolen" some useful Reasons to be not Programmed.

This is expected to have no user impact, as no implementation supports this.

* The UnsupportedAddress Listener condition reason has been moved to a Gateway condition reason

This does not make sense on a Listener, as Listeners do not have addresses

* The AddressNotAssigned Gateway condition reasons has moved from Accepted to Programmed

Accepted is about static analysis of the configuration; assigning an address is about actually programming the resources, so it fits more in Programmed

* The NoResources Gateway condition reasons has moved from Ready to Programmed

Lack of resources is a valid reason to not program the data plane; with the reservation of Ready this is needed in Programmed.

```release-note
The "Ready" Gateway and Listener condition has been reserved for future use.
The UnsupportedAddress Listener condition reason has been moved to a Gateway condition reason
The AddressNotAssigned Gateway condition reasons has moved from Accepted to Programmed
The NoResources Gateway condition reasons has moved from Ready to Programmed
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -536,6 +536,8 @@ const (
 	//
 	// * "Invalid"
 	// * "Pending"
+	// * "NoResources"
+	// * "AddressNotAssigned"
 	//
 	// Possible reasons for this condition to be Unknown are:
 	//
@@ -550,9 +552,20 @@ const (
 	// true.
 	GatewayReasonProgrammed GatewayConditionReason = "Programmed"
 
-	// This reason is used with the "Programmed" condition when the Listener is
+	// This reason is used with the "Programmed" and "Accepted" condition when the Gateway is
 	// syntactically or semantically invalid.
 	GatewayReasonInvalid GatewayConditionReason = "Invalid"
+
+	// This reason is used with the "Programmed" condition when the
+	// Gateway is not scheduled because insufficient infrastructure
+	// resources are available.
+	GatewayReasonNoResources GatewayConditionReason = "NoResources"
+
+	// This reason is used with the "Programmed" condition when none of the requested
+	// addresses have been assigned to the Gateway. This reason can be used to
+	// express a range of circumstances, including (but not limited to) IPAM
+	// address exhaustion, address not yet allocated, or a named address not being found.
+	GatewayReasonAddressNotAssigned GatewayConditionReason = "AddressNotAssigned"
 )
 
 const (
@@ -564,11 +577,13 @@ const (
 	// Possible reasons for this condition to be True are:
 	//
 	// * "Accepted"
+	// * "ListenersNotValid"
 	//
 	// Possible reasons for this condition to be False are:
 	//
+	// * "Invalid"
 	// * "NotReconciled"
-	// * "NoResources"
+	// * "UnsupportedAddress"
 	//
 	// Possible reasons for this condition to be Unknown are:
 	//
@@ -579,12 +594,27 @@ const (
 	// interoperability.
 	GatewayConditionAccepted GatewayConditionType = "Accepted"
 
-	// Deprecated: use "Accepted" instead.
-	GatewayConditionScheduled GatewayConditionType = "Scheduled"
-
 	// This reason is used with the "Accepted" condition when the condition is
 	// True.
 	GatewayReasonAccepted GatewayConditionReason = "Accepted"
+
+	// This reason is used with the "Accepted" and "Programmed"
+	// conditions when the status is "Unknown" and no controller has reconciled
+	// the Gateway.
+	GatewayReasonPending GatewayConditionReason = "Pending"
+
+	// This reason is used with the "Accepted" condition when the Gateway could not be configured
+	// because the requested address is not supported. This reason could be used in a number of
+	// instances, including:
+	//
+	// * The address is already in use.
+	// * The type of address is not supported by the implementation.
+	GatewaReasonUnsupportedAddress GatewayConditionReason = "UnsupportedAddress"
+)
+
+const (
+	// Deprecated: use "Accepted" instead.
+	GatewayConditionScheduled GatewayConditionType = "Scheduled"
 
 	// This reason is used with the "Scheduled" condition when the condition is
 	// True.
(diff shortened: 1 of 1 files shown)


apis/v1beta1/gateway_types.go
```diff
@@ -579,76 +594,54 @@ const (
...
 
-	// This reason is used with the "Ready" condition when the condition is
-	// true.
+	// Deprecated: Ready is reserved for future use
 	GatewayReasonReady GatewayConditionReason = "Ready"
 
-	// This reason is used with the "Ready" condition when one or
-	// more Listeners have an invalid or unsupported configuration
-	// and cannot be configured on the Gateway.
+	// Deprecated: Ready is reserved for future use
```
robscott, line 640: This feels like it should move to "Programmed"?

apis/v1beta1/gateway_types.go
```diff
@@ -579,76 +594,54 @@ const (
...
-	// Possible reasons for this condition to be False are:
-	//
-	// * "ListenersNotValid"
-	// * "ListenersNotReady"
-	// * "AddressNotAssigned"
-	//
-	// Controllers may raise this condition with other reasons,
-	// but should prefer to use the reasons listed above to improve
-	// interoperability.
+	// Deprecated: Ready is reserved for future use
```
shaneutt, line 634: I think this might be a good compromise. I was generally starting to be in favor of removing the condition anyhow, but reserving it in this way seems like it might be reasonable.

apis/v1beta1/gateway_types.go
```diff
@@ -579,76 +594,54 @@ const (
...
-	// and traffic is ready to flow through the data plane immediately.
-	//
-	// If both the "ListenersNotValid" and "ListenersNotReady"
-	// reasons are true, the Gateway controller should prefer the
-	// "ListenersNotValid" reason.
-	//
-	// Possible reasons for this condition to be true are:
+	// "Ready" is a reserved condition type for future use. It should not be used by implementations.
+	// Note: its not really "deprecated", but rather "reserved"; however, deprecated triggers Go linters
+	// to alert about usage.
```
robscott, line 632: The rationale from the PR description here makes sense to me:

>This condition has no implementations; while it may have some future value, we cannot justify the additional reason going GA without usage. Additionally, it has caused confusion by the subtle distinction between Programmed and Ready, and has "stolen" some useful Reasons to be not Programmed.

I think it would be helpful to provide a bit more context in the type definition though. Maybe describe that we want to reserve this term to represent that all configuration has fully propagated. Although we recognize that this is not possible for most implementations today, we are reserving this term for that purpose if it ever becomes broadly implementable.

I'm not really sure what the best mechanism is for this. I like the idea of a deprecation warning because that should show up in most linters. At some point in the future we may be able to move this out of type definitions altogether and just into something like implementation guidelines.

robscott (commented):
Thanks @howardjohn! Lots of great cleanup in here. Adding a hold because this feels like it should have several approvals before merging.

/hold

apis/v1beta1/gateway_types.go
arkodg: cc @liwenwu who mentioned they were using this condition in the call yesterday

apis/v1beta1/gateway_types.go
howardjohn: I mentioned this in https://github.com/kubernetes-sigs/gateway-api/issues/1832 but left it out since it seemed controversial. 

I will add it if folks want

apis/v1beta1/gateway_types.go
liwenwu-amazon: AWS VPC Lattice Gateway Controller creates a AWS VPC lattice service network (SN) object for each K8S gateway object.   After this VPC lattice service network(SN) is fully provisioned, the controller set the status as follow:
```
status:
  conditions:
  - lastTransitionTime: "2023-03-28T04:43:40Z"
    message: 'aws-gateway-arn: arn:aws:vpc-lattice:us-west-2:xxxxxxxx:servicenetwork/sn-06aa98be0c9c95af6'
    reason: Reconciled
    status: "True"
    type: Accepted
```

This means, all Pods inside the cluster(VPC) can access all HTTPRoutes behind the K8S Gateway.  TLDR, through K8S Gateway and able to access all Lattice service(s) in that Lattice Service Network

youngnick: > Maybe describe that we want to reserve this term to represent that all configuration has fully propagated. Although we recognize that this is not possible for most implementations today, we are reserving this term for that purpose if it ever becomes broadly implementable.

This wording is pretty close, how about:
```suggestion
	// "Ready" is a reserved condition type for future use. It should not be used by implementations.
	// Note: This condition is not really "deprecated", but rather "reserved"; however, deprecated triggers Go linters
	// to alert about usage.
	//
	// If used in the future, "Ready" will represent the final state where all configuration is confirmed good
	// _and has completely propagated to the data plane_. That is, it is a _guarantee_ that, as soon as something
	// sees the Condition as `true`, then connections will be correctly routed _immediately_.
	// 
	// This is a very strong guarantee, and to date no implementation has satisfied it enough to implement it.
	// This reservation can be discussed in the future if necessary.
```

apis/v1beta1/gateway_types.go
youngnick: I liked the way you specified it in https://github.com/kubernetes-sigs/gateway-api/issues/1832#issuecomment-1487167378. I think it could be fine as long as we have the "can be present when Accepted is True" caveat.

youngnick:
LGTM, I'll leave someone else to add the official one though.

apis/v1beta1/gateway_types.go
frankbu: I think these definitions of the two conditions "Programmed" and "Ready" are fine, but the names should be reversed. Intuitively, programmed would be this future "final confirmed routed connections" state. I would think most people would think "Programmed" means having routes configured (i.e., the gateway has been programmed), not just that the gateway is up and running with addresses that you can send requests to. I would call that state "Ready", since you could argue that the gateway is ready to accept requests, it's just not yet programmed (i.e., it doesn't know where to route them, so it will return 404s).

apis/v1beta1/gateway_types.go
howardjohn: Added it back

apis/v1beta1/gateway_types.go
howardjohn: Thanks @youngnick , that looks good.

shaneutt (approved):

apis/v1beta1/gateway_types.go
youngnick: @frankbu that makes sense in isolation, however a bunch of other Kubernetes objects already use `Ready` to mean something like what we're using it for in this PR, so we're trying to keep the language consistent with other Kubernetes usage.

Issue #1784, opened 2023-03-07
Initial conformance profiles and reporting prototype


shaneutt:
This is a sub-issue of https://github.com/kubernetes-sigs/gateway-api/issues/1709 to start prototyping the GEP in it's current step so that we can test and try out some of the ideas we have so far and see how they work. The result that should be produced are non-released (not accessible by end-users of the go library unless _explicitly_ opted in to) tools for running profiles and producing a `ConformanceReport`.

PR #1890, opened 2023-03-28
Conformance Tests: remove implicit HTTPRoute support

shaneutt:
**What type of PR is this?**

/kind cleanup
/kind test
/area conformance

Up until this point we had been implicitly enabling `HTTPRoute` tests for conformance test runs. We now have multiple implementations that _do not_ implement `HTTP` support at all (Stunner, Blixt) and so in support of the upcoming work for #1709 we we need to make this an opt in going forward.

This is a blocker for #1709 and #1784 which require this change to enable conformance profiles for the test suite (but it doesn't by itself _resolve_ either).


conformance/utils/suite/features.go
```diff
@@ -0,0 +1,163 @@
...
+
+// -----------------------------------------------------------------------------
+// Features - Compilations
+// -----------------------------------------------------------------------------
+
+// AllFeatures contains all the supported features and can be used to run all
+// conformance tests with `all-features` flag.
+//
+// NOTE: as new feature sets are added they should be inserted into this set.
+var AllFeatures = sets.New[SupportedFeature]().
```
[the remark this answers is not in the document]
shaneutt: It's covered by `StandardExtended`, which [includes it on line 76](https://github.com/kubernetes-sigs/gateway-api/pull/1890/files#diff-f42c32db470693cfe1bf5989b0c4b73928ae7b0598404059d44938e88472161eR76)

conformance/utils/suite/features.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+// StandardCoreFeatures are the features that are required to be conformant with
+// the Core API features (e.g. GatewayClass, Gateway, e.t.c.).
+//
+// TODO: rename this from "Standard" as that might be confusing with our
+// "Standard" release channel.
+var StandardCoreFeatures = sets.New(
+	SupportReferenceGrant,
+)
```
robscott, line 47: Maybe we should just remove this for the time being and add it back in when we have a clearer picture of what it would look like? Just have a separate set with ReferenceGrant doesn't seem very helpful.

conformance/utils/suite/features.go
```diff
@@ -0,0 +1,163 @@
...
+
+const (
+	// This option indicates GatewayClass will update the observedGeneration in
+	// it's conditions when reconciling (extended conformance).
+	//
+	// TODO: we intend to make this core and require implementations to do it
+	//       as we expect this is something every implementation should be able
+	//       to do and it's ideal behavior.
+	//
+	//       See: https://github.com/kubernetes-sigs/gateway-api/issues/1780
```
robscott, line 61: Maybe just make this a note instead of a TODO and then add some kind of issue to track removing this?

conformance/utils/suite/features.go
```diff
@@ -0,0 +1,163 @@
...
+
+// StandardExtendedFeatures are extra generic features that implementations may
+// choose to support as an opt-in.
+//
+// TODO: rename this from "Standard" as that might be confusing with our
+// "Standard" release channel.
+var StandardExtendedFeatures = sets.New(
+	SupportGatewayClassObservedGenerationBump,
+	SupportRouteDestinationPortMatching,
+).Insert(StandardCoreFeatures.UnsortedList()...)
```
robscott, line 76: Similar to my comment above, I'd rather punt on this standard categorization until we have a clearer picture of the common buckets. I think both of these features could be closely tied to individual resources instead (GatewayClass and Gateway respectively). It could help with our test output if we could show that an implementation supports 80% of extended features on Gateway for example.

robscott (commented):
Thanks @shaneutt! This is a great improvement.

conformance/utils/suite/features.go
shaneutt: You made another comment about this below with more prescription about the changes you'd actually like to see, I'll consider this thread resolved in favor of continuing the conversation there.

conformance/utils/suite/features.go
shaneutt: So in case it wasn't clear this isn't exactly a "new" thing, something similar exists currently:

```go
// StandardCoreFeatures are the features that are required to be conformant with
// the Core API features that are part of the Standard release channel.
var StandardCoreFeatures = sets.New(
	SupportReferenceGrant,
)
```

https://github.com/kubernetes-sigs/gateway-api/blob/004eb1b5f290431dd746ee8b1ed21d63d80ac5c7/conformance/utils/suite/suite.go#L77

So I'm not clear on how we would "punt" on this, which seems to insinuate we don't add it at all without completely removing something that existed before in any fashion. Is that what you're advocating for? You just want us to completely remove it for now?

robscott: Yeah you're right, I missed that `StandardCoreFeatures` already exists. It feels weird to have `StandardCoreFeatures` be so empty, but I guess this is really serving as a list of things we're allowing implementations to skip when running conformance tests, even though they're core. I guess this all makes sense, it just feels especially weird with the overlap with standard channel. I hate to merge code that has open TODOs without issues to track them, so if we want to stick with the `standard` name here I think we at least need a follow up issue to come up with a better one.

conformance/utils/suite/features.go
robscott: I guess the existing issue is fine here, I'd just prefer to replace "TODO:" with "Note:".

shaneutt: Ah, I gotcha now. Changed!

conformance/utils/suite/features.go
shaneutt: Sounds good, https://github.com/kubernetes-sigs/gateway-api/issues/1891 was created and the TODO items have been updated to link it.

shaneutt:
Holding just to make sure a couple implementers have the chance to see this.

/hold

CHANGELOG.md
```diff
@@ -23,6 +23,21 @@
...
+
+## Breaking Changes
+
+The following changes only affect implementers which run the conformance test
+suite:
+
+- In preparation for our upcoming [conformance profiles feature][#1709] a new
+  `SupportedFeature` named `HTTPRoute` has been added. Previously there was an
+  implicit assumption in the test suite that implementations supported the
+  `HTTPRoute` API, but this change makes it explicit. All implementations
```
arkodg, line 36:
```suggestion
  `HTTPRoute` API, but this change makes it explicit. All implementations previously running HTTP conformance tests
```

conformance/utils/suite/features.go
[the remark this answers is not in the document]
arkodg: yeah I like the pattern introduced here of adding in one place, is reflected in other superset suites :)

arkodg (approved):
LGTM with minor nit

CHANGELOG.md
shaneutt: Updated, thanks!

skriss (approved):
LGTM -- we're using the `-all-features` flag so looks like this should be a no-op for us for now.

CHANGELOG.md
```diff
@@ -23,6 +23,22 @@
...
+  previously running conformance tests **MUST** add the `SupportHTTPRoute`
+  feature to their test suite options if running the tests via the Go library
+  or the command line.
+  (#1890, @shaneutt)
+
```
robscott, line 41: Any reason not just to include this in the release note block? This does seem like a useful idea for incoming major changes, but technically anything can be reverted before we release so I'd hate for people to invest too much in these kinds of notices.

conformance/utils/suite/features.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+// StandardExtendedFeatures are extra generic features that implementations may
+// choose to support as an opt-in.
+//
+// TODO: we need clarity for standard vs experimental features.
+// See: https://github.com/kubernetes-sigs/gateway-api/issues/1891
+var StandardExtendedFeatures = sets.New(
+	SupportGatewayClassObservedGenerationBump,
+	SupportRouteDestinationPortMatching,
```
robscott, line 75: Nit: This is not in standard channel, may need a separate `ExperimentalExtendedFeatures` set for this.

robscott (commented):
Thanks @shaneutt! A couple nits but otherwise LGTM.

conformance/utils/suite/features.go
shaneutt: Already done in https://github.com/kubernetes-sigs/gateway-api/pull/1890/commits/991b7a647bde9f62475f7a6225a38c1bb3e2f378 :+1:

CHANGELOG.md
shaneutt: Yeah that's fine. Removed and added to PR description.

robscott:
Thanks @shaneutt!

/lgtm

Issue #1891, opened 2023-03-29
Features for every API in conformance tests


shaneutt:
**What would you like to be changed**:

Right now we use `StandardCoreFeatures` as a public symbol in our conformance test suite to refer to the features which are required generically across all implementations, and `StandardExtendedFeatures` for the extended support level version of that. While "standard" may be correct in this case we need more expressiveness for other feature options:

- we need to be able to indicate features which are specific to "common" APIs like `GatewayClass` and `Gateway`
- we also should be able to indicate features that are only present in the experimental channel

Ultimately we should be in a position where no tests run implicitly, and features must be opted into to run any tests.

**Why this is needed**:

`StandardCoreFeatures` doesn't feel quite right standing alone by itself today and we expect a need for more expressiveness as we complete the work in #1709 as we already have experimental features and will have more in the future.

shaneutt:
Between https://github.com/kubernetes-sigs/gateway-api/pull/1890, https://github.com/kubernetes-sigs/gateway-api/pull/1878 and https://github.com/kubernetes-sigs/gateway-api/pull/1894 we are now all set :+1:

PR #1894, opened 2023-03-31
Conformance Tests: No implicit test runs without features

shaneutt:
**What type of PR is this?**

/kind cleanup
/kind test
/area conformance

In #1890 we removed implicit support for `HTTPRoute`, and created #1891 as a follow up to do this for all types. With #1878 bringing Mesh conformance tests soon this patch makes it cleaner for implementations which ONLY support mesh to simply not run any unrelated conformance tests.

This doesn't completely fix, but does support #1891.

```release-note
No conformance tests run by default anymore, including tests for GatewayClass and Gateway. A new
SupportGateway feature must be opted into in order to run those tests (similar to what we've done
previously for ReferenceGrant and HTTPRoute). Also with this release, `EnableAllSupportedFeatures`
enables all Gateway AND Mesh features (where previously that was just Gateway).
```

### DIFF
--- a/conformance/conformance_test.go
+++ b/conformance/conformance_test.go
@@ -70,7 +70,7 @@ func TestConformance(t *testing.T) {
 	if supportedFeatures.Has(suite.SupportGateway) {
 		toRun = append(toRun, tests.ConformanceTests...)
 	}
-	if supportedFeatures.Has(suite.SupportGAMMA) {
+	if supportedFeatures.Has(suite.SupportMesh) {
 		toRun = append(toRun, tests.MeshConformanceTests...)
 	}
 
--- a/conformance/conformance_test.go
+++ b/conformance/conformance_test.go
@@ -64,17 +64,7 @@ func TestConformance(t *testing.T) {
 	})
 	cSuite.Setup(t)
 
-	// TODO(https://github.com/kubernetes-sigs/gateway-api/issues/1891) this will likely changed;
-	// for now, we use this mechanism to split mesh and gateway types
-	var toRun []suite.ConformanceTest
-	if supportedFeatures.Has(suite.SupportGateway) {
-		toRun = append(toRun, tests.ConformanceTests...)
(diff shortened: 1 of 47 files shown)


arkodg (approved):
LGTM !

candita:
@shaneutt could you add a quick update to the conformance docs in https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/concepts/conformance.md ?

arkodg:
curious will `EnableAllSupportedFeatures` run `Gateway` and `Mesh` tests going forward ?

arkodg:
> EnableAllSupportedFeatures already runs mesh before this change. If you don't want mesh, then you don't want **all** features I think. maybe we should have AllGatewayFeatures helper though?

thanks for the clarification, in that case a release note mentioning this would be good to have.
and maybe that `EnableAllSupportedFeatures` can be replaced with just enabled `SupportGateway` feature suite, to retain previous intent ?

shaneutt:
> @shaneutt could you add a quick update to the conformance docs in https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/concepts/conformance.md ?

Good call! changed in 2acae499bc45d585bb240cfe48fdd1047bbde763 :+1:

shaneutt:
> > EnableAllSupportedFeatures already runs mesh before this change. If you don't want mesh, then you don't want **all** features I think. maybe we should have AllGatewayFeatures helper though?
> 
> thanks for the clarification, in that case a release note mentioning this would be good to have.

I had already had a release note in the PR description, let me know if you feel it needs to be expanded upon.

michaelbeaumont (approved):
LGTM! +1 on making the release notes very explicit

shaneutt:
Release notes updated as requested, thank you! :+1:

robscott (approved):
Thanks @shaneutt!

/lgtm

PR #1899, opened 2023-04-01
make embed.FS configurable

zirain:
**What type of PR is this?**

/kind test
/area conformance

make embed.FS configurable, let downstream project override manifests if needed.

xref: https://github.com/envoyproxy/gateway/pull/1239

Fixes #

```release-note
The embedded filesystem used in the conformance test suite is now configurable,
and required as a `suite.Option` (if you use `suite.New()` however, the
default will be provided).
```


arkodg (approved):
LGTM ! thanks
ptal @robscott @youngnick @shaneutt 
this is a no-op for existing projects leveraging the conformance tests and allows some downstream projects like Envoy Gateway to reuse the conformance test framework for implementation specific filter extension conformance tests.

arkodg:
> Why not just `kubectl apply` (or equivilent) them before running the test?

that has been abstracted into the `suite` noun (and implemented in it) which we are trying to leverage (existing ways to apply, run clients, validate results).

shaneutt (approved):
Looks pretty simple, seems helpful, and I understand from @arkodg that this has been manually tested and is working well. It's also just generally low threat since it's conformance test suite only functionality. Looks good to me, I did update the release notes in the description with a little bit more explanation of effect on conformance test users as this will be landing in `v0.7.0`.

/lgtm

PR #1906, opened 2023-04-03
GEP-1897: TLS from Gateway to Backend for ingress

### LINKED ISSUE #1897: GEP: TLS from Gateway to Backend for ingress (backend TLS termination)
Add a document which specifically addresses the topic of conveying HTTPS from the Gateway dataplane to the backend (backend TLS termination), and intends to satisfy the single use case “As a client implementation of Gateway API, I need to know how to connect to a backend pod that has its own certificate”. TLS configuration can be a nebulous topic, so in order to drive resolution this GEP focuses only on this single piece of functionality. 

Given that the current ingress solution specifies **edge** TLS termination (from the edge to the gateway), and how to handle **passthrough** TLS (from the edge to the backend pod), this proposed ingress solution specifies **backend** TLS termination (from the gateway to the backend pod).  As mentioned, this solution satisfies the use case in which the backend pod has its own certificate and the gateway client needs to know how to connect to the backend pod.

**Why this is needed**:

The document is _very_ tightly scoped because we have tried and failed to address this well-known gap in the API specification. The lack of support for this fundamental concept is holding back Gateway API adoption by users that require a solution to the use case. One of the recurring themes that has held up the prior art has been interest related to service mesh and as such this proposal focuses only on the ingress use case to avoid contention there.  Another reason for the tight scope is that we have been too focused on a generic representation of everything that TLS can do, which covers too much ground to address in a single GEP.

candita:
**What type of PR is this?**
/kind gep

There is a well-known gap in the API specification regarding TLS from Gateway to Backend for ingress.

This GEP is _very_ tightly scoped because we have tried and failed to address this well-known gap in the API specification. The lack of support for this fundamental concept is holding back Gateway API adoption by users that require a solution to the use case. 

Fixes #1897

```release-note
NONE
```


geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+
+This document specifically addresses the topic of conveying HTTPS from the Gateway
+dataplane to the backend (backend TLS termination), and intends to satisfy the single
+use case “As a client implementation of Gateway API, I need to know how to connect to
+a backend pod that has its own certificate”. TLS configuration can be a nebulous topic,
+so in order to drive resolution this GEP focuses only on this single piece of functionality.
+
+## Goals
+
+1. The solution should satisfy the following use case: the backend pod has its own
```
bowei, line 16: We should mention the approach to the x-ns reference between Route and Service.

As mentioned on the call -- it would help to say explicitly that we are going to use the standard rules wrt to direct reference configuration.

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+8. Termination of TLS for non-HTTP TCP streams (#3 in [Gateway API TLS Use Cases](#references))
+9. Controlling certificates used by more than one workload (#6 in [Gateway API TLS Use Cases](#references))
+10. Client certificate settings used in the TLS handshake **from external clients to the
+Listener** (#7 in [Gateway API TLS Use Cases](#references))
+
+## Overview - what do we want to do?
+
+Given that the current ingress solution specifies **edge** TLS termination (from the edge to
+the gateway), and how to handle **passthrough** TLS (from the edge to the backend pod), this
+proposed ingress solution specifies **backend** TLS termination (from the gateway to the
```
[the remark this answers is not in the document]
hzxuzhonghu: IIRC, only simple tls origination is covered here and mTLS is not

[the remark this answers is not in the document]
bowei: FWIW -- I do think this is important to call out -- the service is not in scope for the GW API.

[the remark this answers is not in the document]
shaneutt: I have a suggestion which I think may help to resolve the concerns here:

```suggestion
proposed ingress solution specifies TLS origination to the **backend** (from the gateway to the
```

The comment in parenthesis might need a bit more tweaking.

shaneutt (approved):
I think this is a great example of a "what and why without the how" initial GEP to start back up the conversation on this topic without having too many details to comb through right away. As a maintainer I particularly appreciate the iterative approach in the face of our [focus on GA](https://docs.google.com/document/d/1GcPK4cCkbyW6iGwSmBk6Q1dOSA_hGtq02aUx7TwjlTw) as it helps to spread the cognitive load over smaller bits of time, and makes it more feasible for me to review in between other priorities.

The suggestion that this is an important feature for multiple implementations rings true with me, the goals and non-goals make sense, and I very much appreciate you meticulously laying out the long and troubled history of this effort's past.

As I have no blockers and this is `Provisional` with more iterations to come, I approve and I'm looking forward to seeing the follow-ups.

As has become customary with GEPs recently I'll place a temporary hold so that we can allow some time for several people to review without it merging too fast.

/hold

Thanks @candita!

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+3. The solution should consider client certificate settings used in the TLS handshake **from
+Gateway to backend**, such as TLS versions and cipher suites. (Use case #5 in [Gateway API TLS Use Cases](#references))
+
+## Non-Goals
+
+1. Changes to the existing mechanisms for edge or passthrough TLS termination
+2. Providing a mechanism to decorate multiple route instances
+3. Service mesh use cases
+4. Mutual TLS use cases
+5. TLSRoute, TCPRoute, or UDPRoute use cases
```
howardjohn, line 31: While I acknowledge that it is critical to keep the scope reduced in order to make forward progress, I am a bit concerned with making these "Non-goals" (these 2 in particular, but to some extent egress and mesh as well).

I worry that by only considering the narrowest case, we will pigeon-hole ourselves into a non-flexible API. I don't want us to implement "TLSSettingsForHTTPFromIngress" (logically or literally) now, and realize we actually wanted it to be a more general "TLSSettingsFromIngress" (support TCPRoute) or "TLSSettings" (support non-ingress) cases and not be able to move.

Its possible that we will just get lucky and this won't happen, but relying on that seems unlikely to work out well.

I would feel more comfortable if the GEP was more like https://github.com/kubernetes-sigs/gateway-api/pull/1643, which I view as the golden standard for a "What and why" GEP, which explores the problem space and prior art.

Without this, I worry we are just going to solve one very very narrow case and create an awkward long term API.

I would also like to remind folks that the easiest way to experiment with these type of customizations is to _just go implement it_. Gateway API is designed around extensibility; there is nothing stopping a custom implementation from implementing TLS origination today. This serves as an incredibly valuable reference point for discussion - at the end of the day, we get far more info from a real implementation than debating on a doc. That isn't to say that this isn't valuable in core - it absolutely is. But its important that we do it right in core than that we do it fast (or even at all!).

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Goals
+
+1. The solution should satisfy the following use case: the backend pod has its own
+certificate and the gateway implementation client needs to know how to connect to the
+backend pod. (Use case #4 in [Gateway API TLS Use Cases](#references))
+2. In terms of the Gateway API personas, the application developer should control the
+gateway to backend TLS settings, not the cluster operator, as requiring a cluster operator
+to manage certificate renewals and revocations would be extremely cumbersome.
+3. The solution should consider client certificate settings used in the TLS handshake **from
```
costinm, line 22: I don't think TLS versions or cipher suites are appropriate - should be out of scope as well. 

They are typically restricted for compliance reasons  - an app owner should not be able to override. Also platform owner may enforce this by having infastructure restrict this ( compile only the compliant subset, etc). 

Another reason is that they require expertise that app owners typically don't have - very few people understand the compliance space.

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+so in order to drive resolution this GEP focuses only on this single piece of functionality.
+
+## Goals
+
+1. The solution should satisfy the following use case: the backend pod has its own
+certificate and the gateway implementation client needs to know how to connect to the
+backend pod. (Use case #4 in [Gateway API TLS Use Cases](#references))
+2. In terms of the Gateway API personas, the application developer should control the
+gateway to backend TLS settings, not the cluster operator, as requiring a cluster operator
+to manage certificate renewals and revocations would be extremely cumbersome.
```
costinm, line 21: On client certificate - I don't think an API is appropriate. App code can ask for them ( using TLS protocol ), the API can't influence this. And the Gateway - if it can - should present its identity. App owner can't choose a random client cert.

costinm: Also - Gateway SHOULD authenticate with either a JWT with audience or client cert or some other mean - so gateway added headers can be trusted, etc.

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+This proposal is _very_ tightly scoped because we have tried and failed to address this well-known
+gap in the API specification. The lack of support for this fundamental concept is holding back
+Gateway API adoption by users that require a solution to the use case. One of the recurring themes
+that has held up the prior art has been interest related to service mesh and as such this proposal
+focuses only on the ingress use case to avoid contention there.  Another reason for the tight scope
+is that we have been too focused on a generic representation of everything that TLS can do, which
+covers too much ground to address in a single GEP.
+
+## Prior Art - the history of backend TLS
+
```
costinm, line 60: I think Prior Art is about existing implementations - like Istio DR, Ingress annotations, APIs supported by various load balancers. 

At least that's what we do for persistent sessions - previous discussions are not prior art.

costinm (commented):
I think it's a valid use case and it is ok to start with a narrow set of use cases ( as long as we believe the API choice wouldn't prevent or make difficult the other use cases). 

However I would requiest that "Prior art" section is fixed and we include the actual 
APIs in current use by different LB solutions. Past internal discussions are not prior art.
Once we have a clear picture of what various implementations currently do and support - we will have a much easier time defining our API.

geps/gep-1897.md
```diff
@@ -0,0 +1,113 @@
...
+This proposal is _very_ tightly scoped because we have tried and failed to address this well-known
+gap in the API specification. The lack of support for this fundamental concept is holding back
+Gateway API adoption by users that require a solution to the use case. One of the recurring themes
+that has held up the prior art has been interest related to service mesh and as such this proposal
+focuses only on the ingress use case to avoid contention there.  Another reason for the tight scope
+is that we have been too focused on a generic representation of everything that TLS can do, which
+covers too much ground to address in a single GEP.
+
+## Prior Art - the history of backend TLS
```
shaneutt, line 59: Based on feedback from @costinm in https://github.com/kubernetes-sigs/gateway-api/pull/1906/files#r1157820085

```suggestion
## The history of backend TLS
```

geps/gep-1897.md
shaneutt: It seems fine to me to change the language, Github suggestions can be really helpful for this because it shows the contributor a clear and prescriptive action they can take to consider your comment resolved, as such I created this one:

https://github.com/kubernetes-sigs/gateway-api/pull/1906/files#r1157838147

(I had to create a new comment rather than commenting here because you targeted line 60 which was the line below)

Let me know if you're fine with how that's worded, or want something different, or feel free to make a suggestion in the same thread and I'll rescind mine.

I'll consider this resolved so we can follow up over there.

liwenwu-amazon:
The GEP looks good!. In [AWS VPC Lattice controller gateway API/multicluster implementation](https://github.com/aws/aws-application-networking-k8s),  VPC Lattice is the federation.   

For serviceexport, the lattice controller register all of its endpoints in local cluster to VPC lattice

To reference a serviceimport as a backend of a HTTPRoute, the lattice controller will configure VPC lattice resource accordingly. So that the traffic will get routed to endpoints in that remote cluster(VPC) by VPC lattice dataplane.

In addition to K8S endpoints, serviceimport can references targets like EC2, lambda function and ECS container(s).

geps/gep-1897.md
candita: Thanks @bowei, I will be sure to mention this in the implementation section, as I think the description will be dependent on the implmentation details.

geps/gep-1897.md
candita: I am referring to the application developer as a Gateway API persona, and thus the person who creates the HTTPRoute.  The second comment, iiuc is implementation dependent.

geps/gep-1897.md
candita: They are in scope for this discussion, as TLS version and cipher suites can be application dependent.  Whether the app owner can override the platform owner is an implementation detail too.

howardjohn: My understanding:
* Apps will set allowed version/suites - this is not something we decide, that is how things work.
* Per costin, platform owners _should_ have a way to enforce gateway-api driven defaults.
* TLS version/cipher is a handshake

Where we may want a per-app/per-service gateway-api config override is if the platform admin says (for example) "TLS 1.3 is required", but an app says "I only speak TLS 1.2". As is, that would fail. However, we could allow a per-service opt-out of the platform rules and say "for only svc-legacy, allow 1.2".

This seems a bit strange to me, though. If we want TLS 1.3 everywhere, we probably _really_ want it everywhere. Its not clear why we would be ok with less secure methods for one service but not another.

If we just want to default to more secure options, that is very different - as mentioned, TLS is a handshake negotiation on these, so we can say at admin level "Support 1.3,1.2 and suites [best,2ndbest,3rdbest]" or equivilent.

candita:
@howardjohn @costinm @bowei I have highlighted some goals as immediate and longer term, as well as added some details on prior art in several implementations. PTAL when you get a chance.

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+## Immediate Goals
+
+1. The solution must satisfy the following use case: the backend pod has its own
+certificate and the gateway implementation client needs to know how to connect to the
+backend pod. (Use case #4 in [Gateway API TLS Use Cases](#references))
+2. In terms of the Gateway API personas, only the application developer persona in this
+solution. The application developer should control the gateway to backend TLS settings,
+not the cluster operator, as requiring a cluster operator to manage certificate renewals
+and revocations would be extremely cumbersome.
+3. The solution should consider client certificate settings used in the TLS handshake **from
```
howardjohn, line 23: sorry to nitpick but version and cipher suites are TLS handshake properties, not "certificate settings"

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+
+These are worthy goals, but will not be covered by this GEP.
+
+1. Changes to the existing mechanisms for edge or passthrough TLS termination
+2. Providing a mechanism to decorate multiple route instances
+3. TLSRoute use cases
+4. UDPRoute use cases 
+5. Termination of TLS for HTTP routing (#1 in [Gateway API TLS Use Cases](#references))
+6. HTTPS passthrough use cases (#2 in [Gateway API TLS Use Cases](#references))
+7. Termination of TLS for non-HTTP TCP streams (#3 in [Gateway API TLS Use Cases](#references))
```
howardjohn, line 44: Mentioning these as "already solved" may be more clear than "Non-goals".

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+8. Controlling certificates used by more than one workload (#6 in [Gateway API TLS Use Cases](#references))
+9. Client certificate settings used in the TLS handshake **from external clients to the
```
howardjohn, line 46: I don't understand this. Its not possible for us to control and external client?

I get its a Non-Goal; but checking my understanding here

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+Listener** (#7 in [Gateway API TLS Use Cases](#references))
+10. The cluster operator should be allowed to override any gateway to backend TLS settings, to
```
howardjohn, line 48: Its a bit confusing to say "Should be allowed to" in "non-goals".

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Overview - what do we want to do?
+
+Given that the current ingress solution specifies **edge** TLS termination (from the client to
+the gateway), and how to handle **passthrough** TLS (from the client to the backend pod), this
+proposed ingress solution specifies TLS origination to the **backend** (from the gateway to the
+backend pod).  As mentioned, this solution satisfies the use case in which the backend pod
+has its own certificate and the gateway client needs to know how to connect to the backend pod.
+
+![image depicting TLS termination types](images/1897-TLStermtypes.png "TLS termination types")
```
howardjohn, line 59: very nit, would be nice to have one showing terminating but not originating.

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+Details deferred until we reach consensus on what we want to do, and why we want to do this.
+
+## Prior Art
+
+TLS from gateway to backend for ingress exists in several implementations, and was developed independently.
+
+### Istio Gateway supports this with a DestinationRule to encrypt the gateway->backend hop:
+
+* A secret representing a certificate/key pair, where the certificate is valid for the route host
+* Set Gateway spec.servers[].port.protocol: HTTPS, spec.servers[].tls.mode=SIMPLE, spec.servers[].tls.credentialName:
```
howardjohn, line 133: nit: this is technically not "encrypt the gateway->backend", this is terminating client->gateway.

Only `Set DestinationRule spec.trafficPolicy.tls.mode: SIMPLE` is "encrypt the gateway->backend"

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+* An Envoy client certificate
+* A CA certificate and SubjectName which are both used to verify the backend endpoint’s identity.
+* Kubernetes Service annotation: projectcontour.io/upstream-protocol.tls
+
+Ref: [Upstream TLS](https://projectcontour.io/docs/v1.21.1/config/upstream-tls/)
+
+### GKE supports a way to encrypt traffic to the backend pods using:
+
+* `AppProtocol` on Service set to HTTPS
+* Certificate verification by pod
```
howardjohn, line 158: Not sure what this means but I don't think GKE verifies anything

geps/gep-1897.md
```diff
@@ -0,0 +1,172 @@
...
+In January 2023 we closed GEP-1282 and began a new discussion on enumerating TLS use cases in
+[Gateway API TLS Use Cases](#references), for the purposes of a clear definition and separation of concerns.
+This GEP is the outcome of the TLS use cases #4 and #5 in
+[Gateway API TLS Use Cases](#references) as mentioned in the Goals section above.
+
+## API
+
+Details deferred until we reach consensus on what we want to do, and why we want to do this.
+
+## Prior Art
```
howardjohn, line 126: So:

* Client decides TLS: Istio, Ambassador, OpenShift
* Server (service) decides TLS: Contour, GKE

Is that a fair representation?

geps/gep-1897.md
candita: I agree that an override or default should be possible.  However, I will leave this discussion for another GEP or the implementation details.  Every cluster has its use cases, and we may need to address a cluster-level override or default with a different set of use cases.  Sound fair?

howardjohn: I don't think this is an "overrides and defaults" scenario. We need that when we are talking about something like Retry or Timeout. TLS is different; it is an existing protocol which already has a negotiation aspect. We cannot and should not invent another negotiation on top of that.

geps/gep-1897.md
candita: Updated

geps/gep-1897.md
candita: Updated

geps/gep-1897.md
candita: I think Ambassador relies on the service configuration, so I would put it with the second bullet.

geps/gep-1897.md
candita: Hm, this is the exact opposite of the first feedback I got on this diagram.

geps/gep-1897.md
candita: This is [use case 7](https://docs.google.com/document/d/17sctu2uMJtHmJTGtBi_awGB0YzoCLodtR6rUNmKMCs8/edit#heading=h.y0ofkfsm5t0c) , which @youngnick mentioned is "particularly useful for JWT use cases, and can make doing some types of authentication and authorization easier".

geps/gep-1897.md
candita: Fixed

geps/gep-1897.md
candita: I prefer to cover all the use cases in the referenced TLS use cases document.  Sound fair?

geps/gep-1897.md
costinm: +1

Also UX is important - FIPS compliance and what cipher suites are safe or required is a completely different persona than Jane the app developer.  It is something that may be implemented by compile options and deep into platform - and if a cipher or TLS version becomes unsafe it is not up to the developers to make changes, but platform owner.

geps/gep-1897.md
```diff
@@ -0,0 +1,171 @@
...
+7. Termination of TLS for non-HTTP TCP streams (#3 in [Gateway API TLS Use Cases](#references))
+8. Controlling certificates used by more than one workload (#6 in [Gateway API TLS Use Cases](#references))
+9. Client certificate settings used in TLS **from external clients to the
+Listener** (#7 in [Gateway API TLS Use Cases](#references))
+10. Providing a mechanism for the cluster operator to override gateway to backend TLS settings.
+
+## Overview - what do we want to do?
+
+Given that the current ingress solution specifies **edge** TLS termination (from the client to
+the gateway), and how to handle **passthrough** TLS (from the client to the backend pod), this
```
costinm, line 53: Note that 'passthrough' is covered by TLSRoute ( routing via SNI without edge termination ). 

Gateway communicating with a backend using TLS has nothing to do with edge TLS termination - it is just a property of the backend that it requires TLS. Even if the original internet request was plain HTTP - the gateway would still have to use TLS to the backend ( because typically that's all that the backend accepts ).

That's why 're-encryption' or 'passthrough' are not right terms here.

geps/gep-1897.md
```diff
@@ -0,0 +1,171 @@
...
+proposed ingress solution specifies TLS origination to the **backend** (from the gateway to the
+backend pod).  As mentioned, this solution satisfies the use case in which the backend pod
+has its own certificate and the gateway client needs to know how to connect to the backend pod.
+
+![image depicting TLS termination types](images/1897-TLStermtypes.png "TLS termination types")
+
+Gateway API is missing a mechanism for separately providing the details for the backend TLS handshake,
+including:
+* use of TLS
+* destination CA (certificate authority) or CA bundle
```
costinm, line 63: Not sure if defining K8S root CAs is really in scope for this WG - my understanding was that there are separate WG working on certificate provisioning ( cert request, etc) and root CA configurations. 

Mutating/Validating webhooks are a major pain - I would be very careful around this UX, it is extremely hard to automate or use in helm charts.

geps/gep-1897.md
```diff
@@ -0,0 +1,171 @@
...
+* SANs for validating upstream service (server authentication)
```
costinm, line 64: This is another area where I'm not sure the HttpRoute author should make decisions ( egress is a slightly different case that I believe is not in scope here ). 

Currently Spiffee (URL) and DNS SANs are the majority of SANs in use. There are use cases where the DNS SAN may not be the same as authority - but not clear that the HttpRoute author is the best persona for configuring this ( as opposed to the service owner ).

geps/gep-1897.md
costinm: AFAIK Istio is both - with the new auto-mtls clearly in 'server decide' category.

And the service ultimately decides everywhere - if the service has a TLS port, there is nothing a client can do to change the fact that only TLS connections will work.

geps/gep-1897.md
```diff
@@ -0,0 +1,171 @@
...
+
+## API
+
+Details deferred until we reach consensus on what we want to do, and why we want to do this.
+
+## Prior Art
+
+TLS from gateway to backend for ingress exists in several implementations, and was developed independently.
+
+### Istio Gateway supports this with a DestinationRule:
```
costinm, line 129: DestinationRule is mainly using this for egress destinations. 

For Gateway -> backend, Istio (now) uses auto-mtls, with the root CAs distributed as a config map and not explicitly configured by users.

Using tls.mode=SIMPLE and the rest for internal traffic is a rarely used config - and typically breaks Istio telemetry and policy features.

geps/gep-1897.md
shaneutt: As this comment stands there's a lot of room for interpretation and there's some ambiguity in the suggestions for action. In this state I could see it hanging around for some time without being resolved, so I want to try and get it distilled. The "just go implement it" suggestion is not something we the maintainers feel we should be pushing for in this case, but for everything else we would recommend creating prescriptive and explicit suggestions (ideally with Github's review suggestions feature when feasible, as these are clear and actionable). Are you OK with resolving this comment in favor of reworking it into more explicit suggestions for change?

geps/gep-1897.md
youngnick: @howardjohn and @costinm, everything you've said is correct, but you are missing the function that end users end up using an ingress proxy for - fixing broken apps. It's _very_ common that the TLS version policy will be "Everything must be TLS 1.3, except for this list of exceptions". Having a way to describe that list of exceptions is imperative.

Yes, this can absolutely be a footgun, but it is _definitely_ required in real-world use cases. We had to add similar functionality to Contour for exactly this reason.

Does this create effectively another negotiation over the top of the TLS one? Yes. Is it still unfortunately necessary? Also yes.

This is just one of those things that people who are stuck working in a service-mesh free world are stuck with.

geps/gep-1897.md
youngnick: This section is saying that TLS termination at the _Listener_ is already done, as is TLS Passthrough (with TLSRoute). It's entirely appropriate to use those terms here in those two cases.

geps/gep-1897.md
```diff
@@ -0,0 +1,171 @@
...
+9. Client certificate settings used in TLS **from external clients to the
+Listener** (#7 in [Gateway API TLS Use Cases](#references))
+10. Providing a mechanism for the cluster operator to override gateway to backend TLS settings.
+
+## Overview - what do we want to do?
+
+Given that the current ingress solution specifies **edge** TLS termination (from the client to
+the gateway), and how to handle **passthrough** TLS (from the client to the backend pod), this
+proposed ingress solution specifies TLS origination to the **backend** (from the gateway to the
+backend pod).  As mentioned, this solution satisfies the use case in which the backend pod
```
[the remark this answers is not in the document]
youngnick: Again, the point of this initial cut of this GEP is to agree that _the problem is worth solving_. Determining who owns the config and what the config looks like are part of the implementation, which will be a separate PR. This is not the time for this feedback.

geps/gep-1897.md
youngnick: This is _not_ about defining K8S root CAs. Read the text.

This is about _being able to specify a CA bundle for the connection from the Gateway to the backend_. It's common that a service owner that is doing their own certificate management will manage their own root of trust, which will necessitate a separate CA bundle. It has nothing to do with Mutating or Validating webhooks.

geps/gep-1897.md
shaneutt: Considering this resolved as there hasn't been an update in some time. If you feel there's still action that needs to be taken here please unresolve and let us know in a comment.

geps/gep-1897.md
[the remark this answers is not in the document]
shaneutt: Indeed, the implementation details will come and most certainly will take a while to tease through.

geps/gep-1897.md
shaneutt: Indeed, but if you feel we've missed something and there's some changes to the language needed here please do drop a specific suggestion for change and/or let us know.

geps/gep-1897.md
shaneutt: Indeed, we're just trying to acknowledge what the problem is and if it's worth working on. That said, if you feel there's a way we can better capture this and other comments which aren't clearly actionable at this stage, we've used a `TODO` list inside the GEP elsewhere it may be reasonable to suggest TODO items that need to be resolved before we move out of `Provisional` for some of your concerns, do feel free to suggest adding that to this PR.

geps/gep-1897.md
shaneutt: What action would you like to see taken here?

geps/gep-1897.md
robscott: Agree that a `TODO` and/or Unresolved section about this so we don't lose track of it. Let me know if this already happened and I missed it.

geps/gep-1897.md
robscott: Agree with @howardjohn here that it would be better to distinguish between "already solved" and "not yet done but not in scope".

robscott:
Would like to see a couple small follow ups on comment threads (https://github.com/kubernetes-sigs/gateway-api/pull/1906#discussion_r1176961243 and https://github.com/kubernetes-sigs/gateway-api/pull/1906#discussion_r1176959862), but otherwise this LGTM. Thanks @candita!

geps/gep-1897.md
candita: I will add an Open Questions section and try to record anything having to do with implementation details.

geps/gep-1897.md
candita: I will add a section "Already Solved TLS Use Cases" to isolate these 3 cases.

geps/gep-1897.md
candita: Removed the word "handshake".

geps/gep-1897.md
candita: I originally had the three termination types separated in a diagram, and got feedback from @robscott that it made more sense to combine these two in one as a contrast to the passthrough route. I can't please both of you, but hope it's okay as is.

geps/gep-1897.md
candita: @howardjohn @costin If there's something you want me to add to summarize this better, or you want me to point out the two categorizations, please let me know.

geps/gep-1897.md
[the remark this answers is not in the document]
candita: Added to Open Questions.

geps/gep-1897.md
candita: Added to Open Questions.

geps/gep-1897.md
candita: Added to Open Questions.

candita:
@howardjohn @costinm @robscott @bowei @shaneutt I have made the following changes:

- Removed already-solved use cases from non-goals
- Added these uses cases to a new section called Already Solved TLS Use Cases
- Added a new section called Open Questions (TODO) listing comments from @bowei and @costinm

geps/gep-1897.md
robscott: Yeah for a bit more context it was very difficult to show all possible forms here. We'd essentially need to cover all of the following options:

1. Gateway termination, no origination
2. Gateway termination, origination
3. No Gateway termination, origination
4. No Gateway termination, no origination
5. Passthrough

The diagram that we ended up with feels simpler than anything that tried to represent all possible cases, and representing a slightly larger subset could just lead to confusion.

robscott:
Will leave this for someone else to add the final LGTM, thanks for all the work on this @candita!

/approve

shaneutt (approved):
/lgtm

PR #1908, opened 2023-04-04
Clarifying Port Redirect Behavior

robscott:
**What type of PR is this?**
/kind cleanup

This is a follow up to address some of the limitations of the current spec discovered in https://github.com/kubernetes-sigs/gateway-api/pull/1880. Specifically this adjust the spec to say that port shouldn't be added in these cases, not that it needs to be unset/removed.

```release-note
Clarification that port redirects should not add port number to Location header for HTTP and HTTPS requests on 80 and 443.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -894,12 +894,13 @@ type HTTPRequestRedirectFilter struct {
 	//
 	// When empty, the Gateway Listener port is used.
 	//
-	// Implementers should _not_ include the port number in the 'Location' header in the following cases:
+	// Implementations SHOULD NOT add the port number in the 'Location'
+	// header in the following cases:
 	//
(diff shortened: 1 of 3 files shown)


arkodg (approved):
LGTM thanks !

howardjohn:
/hold

have a quick question

kflynn (approved):
Looks good from here, FWIW!

pleshakov (approved):
👍

robscott:
@howardjohn Based on chat I think I captured your question in https://github.com/kubernetes-sigs/gateway-api/issues/1909. I think that's something that's probably best covered by a follow up, but will leave this PR open until tomorrow in case someone wants to include the resolution to that issue here as well.

michaelbeaumont (approved):

PR #1912, opened 2023-04-05
Request mirroring conformance

LiorLieberman:
This PR is a adds a working version of http request mirror conformance tests.
It still needs slight improvements like;
 - Add more logs in `ExpectMirroredRequest()`
 - Add more test cases

Feedback is most welcomed
/cc @robscott
**What type of PR is this?**
/kind feature
/area conformance

Add tests for request mirroring 

Issue #1103 

```
- Added: conformance tests for http request mirroring.
```


conformance/tests/httproute-request-mirror.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+		}
+		for i := range testCases {
+			// Declare tc here to avoid loop variable
+			// reuse issues across parallel tests.
+			tc := testCases[i]
+			t.Run(tc.GetTestCaseName(i), func(t *testing.T) {
+				t.Parallel()
+				http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, tc)
+				if tc.MirroredTo != "" {
```
arkodg, line 66: is this `if tc.MirroredTo != "" {` required, we want to ensure we are validating mirrored backends in this test/file(httproute-request-mirror.go)

conformance/utils/http/http.go
```diff
@@ -51,6 +51,9 @@ type ExpectedResponse struct {
 	Backend   string
 	Namespace string
 
+	// MirroredTo is the destination pod of the mirrored request.
+	MirroredTo string
```
robscott, line 55: How would we know a specific Pod? I think most of our Services have at least 2 Pods. We could theoretically change one of our base services to only have 1 Pod to simplify testing this, but in any case this would probably need to be a Service name.

conformance/utils/kubernetes/logs.go
```diff
@@ -0,0 +1,64 @@
...
+	podListOptions := &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{"app": name}),
+		Namespace:     ns,
+	}
+	if err := c.List(context.TODO(), pods, podListOptions); err != nil {
+		return nil, err
+	}
+
+	podLogOptions := &corev1.PodLogOptions{
+		Container: name,
```
robscott, line 44: How do we know container name here? I thought we were only passing Service name in?

conformance/utils/http/mirror.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	clientset "k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
```
robscott, line 28:
```suggestion
	"sigs.k8s.io/controller-runtime/pkg/client"

	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
```

Without this separation between local and external imports golint will fail.

robscott (commented):
Thanks @LiorLieberman!

conformance/utils/http/http.go
LiorLieberman: I read logs from all pods with a label `app: <tc.MirroredTo value>`. (which is similar to how the service pick the pods it expose) 

I could maybe query the service and find which pods are attached to it if you think thats better.

LiorLieberman: We could also add a service with 1 Pod to the base manifests to make it clearer.

conformance/utils/kubernetes/logs.go
LiorLieberman: I took it from the base manifests, it is the same name.

conformance/tests/httproute-request-mirror.go
LiorLieberman: right, fail if `tc.MirroredTo = ""` sounds good ? or you might have a better idea?

arkodg: I'd recommend removing the check altogether, and directly using the val

LiorLieberman: @arkodg  I feel like using the val when its nil will just send unwanted requests to the api server asking for logs from `""` pod name

arkodg: that shouldnt be happening, you could move the check into `ExpectMirroredRequest` and then fail the test if `MirroredTo` is unset

conformance/utils/http/http.go
LiorLieberman: I am gonna resolve it but feel free to open it again in case needed

LiorLieberman:
@robscott @shaneutt PTAL

LiorLieberman:
@robscott @arkodg can you take another look?

arkodg (approved):
LGTM, thanks !

conformance/conformance_test.go
```diff
@@ -62,6 +62,7 @@ func TestConformance(t *testing.T) {
 		Client:                     client,
 		RESTClient:                 clientset.CoreV1().RESTClient().(*rest.RESTClient),
 		RestConfig:                 cfg,
+		Clientset:                  clientset,
```
robscott, line 65: Can you add a comment here stating that this is only intended to be used as a stopgap until controller-runtime client supports commands like logs? Maybe link to https://github.com/kubernetes-sigs/controller-runtime/issues/452.

conformance/base/manifests.yaml
```diff
@@ -413,3 +413,49 @@ spec:
         resources:
           requests:
             cpu: 10m
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: mirrored-backend
```
robscott, line 420: Why do we need a new Service and Deployment for this test? Could we reuse one of the existing ones?

robscott (commented):
Thanks @LiorLieberman! Still need to actually run these tests, but the code looks reasonable to me.

conformance/base/manifests.yaml
LiorLieberman: I initially reused one and changed it based on your comment here https://github.com/kubernetes-sigs/gateway-api/pull/1912#discussion_r1158957007

Would you prefer that I change one of the existing backends to have only 1 replica instead of having a new one? 

Anyway I cover cases where there is more than one pod (I read the logs from all the pods and break the loop once I found the line I want) so in theory I could remove this new one and not doing any change to the base manifests.

robscott: >Anyway I cover cases where there is more than one pod (I read the logs from all the pods and break the loop once I found the line I want) so in theory I could remove this new one and not doing any change to the base manifests.

Since you're already doing this, maybe just use an existing Service/Deployment.

robscott:
Thanks @LiorLieberman!

/lgtm
/approve

robscott:
@LiorLieberman do you mind manually creating a cherry-pick PR to the release-0.7 branch for this?

LiorLieberman:
Sure, https://github.com/kubernetes-sigs/gateway-api/pull/2034

Just make sure I did it correctly

PR #1914, opened 2023-04-05
Fix mermaid diagram rendering

dprotaso:
Seems like GitHub doesn't render the flow chart when the mermaid block is right under the div tag

/kind cleanup
/kind documentation


robscott:
Thanks @dprotaso!

/lgtm
/approve

PR #1917, opened 2023-04-05
Loosen requirement for GC finalizer

### LINKED ISSUE #1411: Consider the MUST requirement for gatewayclass finalizer

The `GatewayClass` [doc](https://github.com/kubernetes-sigs/gateway-api/blob/2738bba07d9cc98f4f0334a638c26d2159ae12b0/apis/v1beta1/gatewayclass_types.go#L45-L50) states that 
```
// Whenever one or more Gateways are using a GatewayClass, implementations MUST
// add the `gateway-exists-finalizer.gateway.networking.k8s.io` finalizer on the
// associated GatewayClass. This ensures that a GatewayClass associated with a
// Gateway is not deleted while in use.
```

Since this is described as a MUST in the API, it should be checked by a conformance test that creates a `GatewayClass` and a `Gateway` that uses it. Once the `Gateway` is created, the existence of such a finalizer should be checked on the `GatewayClass`.

**Why this is needed**:

This is needed because the existence of such a finalizer is stated as mandatory.

howardjohn:
**What type of PR is this?**

/kind documentation

Addresses https://github.com/kubernetes-sigs/gateway-api/issues/1411#issuecomment-1498253632

Fixes #1411

```release-note

```

### DIFF
--- a/apis/v1alpha2/gatewayclass_types.go
+++ b/apis/v1alpha2/gatewayclass_types.go
@@ -44,7 +44,7 @@ import (
 // If implementations choose to propagate GatewayClass changes to existing
 // Gateways, that MUST be clearly documented by the implementation.
 //
-// Whenever one or more Gateways are using a GatewayClass, implementations MUST
+// Whenever one or more Gateways are using a GatewayClass, implementations SHOULD
 // add the `gateway-exists-finalizer.gateway.networking.k8s.io` finalizer on the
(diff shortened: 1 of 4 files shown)


robscott:
Thanks @howardjohn! May need codegen, but LGTM beyond that.

/approve

shaneutt (approved):
needs `make` and then lgtm

shaneutt (approved):
/lgtm

PR #1926, opened 2023-04-07
Clarify parentRef binding for Mesh

### LINKED ISSUE #1824: Clarify GAMMA binding semantics
Current GEPs for mesh functionality explicitly mention binding to the Service resource as the only mechanism for expressing mesh traffic. However, as discussed in the latest GAMMA meeting, this limitation is not intended as we want to allow implementations to bind to their own CRDs, or ServiceImport, as examples. We should create a follow-up GEP that goes into more detail with respect to this binding mechanism, defining the qualities that a parentRef must have to be a valid target for mesh.

keithmattix:
**What type of PR is this?**

/kind gep
/area mesh

Clarifies parentRef binding

Fixes #1824 

```release-note
Add clarification for parentRef binding
```


geps/gep-1924.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Extended Conformance
+
+In addition to Service, there are other optional parentRef resources that, if used by implementations, MUST adhere to the spec’s prescriptions. At the time of writing (v0.70), there are two resources in extended conformance: ServiceImport (part of the mcs-api, currently in alpha) and Gateway . The semantics of ServiceImport parentRef binding can be found in GEP 1748 (Note: Headless ServiceImports are out of scope and not currently a part of the spec). The semantics for Gateways in the context of GAMMA implementations are laid out below:
+
+### `Gateways`
+
+There has been much discussion around cluster local Gateways (i.e. Gateways not associated with a traditional load balancer). While there are various potential UX impairments (e.g. what’s the difference between a GAMMA HTTPRoute with a Gateway parentRef and an ingress implementation’s HTTPRoute?), there is no technical reason why a Gateway cannot be a valid GAMMA parentRef. In this case, the mesh implementation MUST use the ClusterIP of the Gateway for traffic matching (but not necessarily to the exclusion of the External IP).
```
howardjohn, line 33: This started out being concise and then turned to me rambling:

There is something about this that doesn't quite sit right with me around "MUST use the ClusterIP of the Gateway for traffic matching".

A Gateway doesn't have a ClusterIP, that is a Service construct. It has an Address, which may be local to the cluster (and it may even literally be a ClusterIP, copied from a Service into the Gateway status).

The spec declares that when you connect to the Gateway's address, the configuration on the GW (listeners + routing rules + policy + etc) are actuated.

If we happen to have a GatewayClass in the future that generates a "cluster local IP" (and maybe even a DNS name!) and puts them in the gateway status, we would of course use them and should apply config to them.

---

Really nothing in GAMMA is actually mesh specific. The Gateway API just says a parent is something we are attaching a route to. A Gateway is something with an address we can call. A Service is also something with an address we can call, so of course we can bind to it.

If we were to have a "ClusterIP GatewayClass", a "GAMMA implementation" is really just the same as "A controller than implements the 'ClusterIP GatewayClass'". That happens to be something only really feasible by products that are "service meshes", but you can explain all of Gateway - including 'GAMMA' - without the word "mesh" IMO.

geps/gep-1924.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Why not `IPAddress`
+
+In Kubernetes 1.27, there will be a new IPAddress resource added to networking.k8s.io/v1alpha1 as part of [KEP 1880](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1880-multiple-service-cidrs#proposal). Naturally, it makes sense to examine whether or not this new resource makes sense as a GAMMA aware parentRef. At first glance, IPAddress seems to be an appropriate abstraction for the “frontend” role we’ve been discussing; every Kubernetes Service is accessed over the network via one of its ip addresses. Furthermore, the fact that the Service resource auto-creates an IPAddress is encouraging. However, the fact that the name of the IPAddress is simply the decimal/hex ip address and not a human-readable Service name makes the UX untenable as a parentRef.
```
howardjohn, line 37: I agree with this, generally. however, I would say that it makes a _horrendous UX_ to be a parent parentRef, but its not inherently _wrong_. It meets all the properties we need.

In fact, a hypothetical alternative implemented where the `name` was something stable/human readable, it would be perfectly fine.

So I think we should not make it an official thing supported, but we also don't need to do anything like say "You MUST not use IPAddress as a parentRef". If you can call it and attach routes to it, its a possible parentRef.

howardjohn (commented):
Looks good overall, just a few ramble-y comments

geps/gep-1924.md
```diff
@@ -0,0 +1,76 @@
+# GEP-1926: Mesh parentRef binding
+
+* Issue: [#1824](https://github.com/kubernetes-sigs/gateway-api/issues/1824)
+* Status: Provisional
+
+## Overview
+
+[GEP-1426](https://gateway-api.sigs.k8s.io/geps/gep-1426/) defines how xRoute resources should be bound (via parentRef) for service mesh implementations. However, the existing spec only explicitly mentions binding to the Service resource which is not the desired end state for the spec’s mesh capabilities. This GEP clarifies what kind of resources xRoutes can bind to for mesh implementations and goes into more detail what qualities xRoute parentRefs must have.
```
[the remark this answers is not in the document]
keithmattix: I thought about it, and if others agree I'm happy to modify the 1426. The thing that stopped me was that 1426 is already so long, I worry these points would get buried

geps/gep-1924.md
brianehlert: > Really nothing in GAMMA is actually mesh specific.
I think that is the struggle.  Somehow it has to be possible to say, I have a Mesh-y thing servicing this or I have an ingress-y thing servicing that.  Without precluding that there can't be a future thing that can do all the things and how do we not make that feel unnatural.

To me the Service remains the logical thing.  My ingress-y thing needs to know that in this case it should send to the Service cluster IP for a mesh-y thing to then take over.  And for this other service there is no mesh-y thing and traffic can be steered direct to the Service pods and the ingress-y thing simply does more work.
I don't think I have an answer, but feel a need to recognize the historic assumptions that I think we are stuck on.

geps/gep-1924.md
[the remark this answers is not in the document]
shaneutt: I think it boils down to whether these are meaningfully different things, and if they're not and we converge can we make some clear subsections such that nagivating to the parentRef binding specific details is straightforward.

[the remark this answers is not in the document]
keithmattix: Looking back through it, most of the length of 1426 is "alternatives considered", so I will just make these points there

geps/gep-1924.md
keithmattix: > A Gateway doesn't have a ClusterIP, that is a Service construct. It has an Address, which may be local to the cluster (and it may even literally be a ClusterIP, copied from a Service into the Gateway status).

This is a fantastic call-out and an oversight on my part. I agree 100% that Service constructs shouldn't scope creep in the Gateway spec. 

> Somehow it has to be possible to say, I have a Mesh-y thing servicing this or I have an ingress-y thing servicing that.

Would the controllerName in GatewayClass suffice?

geps/gep-1924.md
keithmattix: Agreed - I don't think anything here says "you MUST" not use it, but I can preemptively clear this up

geps/gep-1924.md
costinm: I don't know what 'mesh-y' is ( and I've been worked on this for 5 years). 

There is a DNS name that resolves to an IP, and requests happen to go where HttpRoute decides. Nothing 'mesh-y' or 'ingress-y' about it - that's how internet has worked long before Istio/meshes and K8S. 

Some Gateway implementation already support 'internal gateway' model ( i.e. IP visible only in the VPC). Service itself can be of LoadBalancer type - also external or internal. Will a HttpRoute attached to a Service with type LoadBalancer/external be any different from a gateway ? How and why ? 

Will a Gateway of type 'internal' be different from a mesh-y ? 

I think the main difference between a Service/LoadBalancer and a Gateway is that Gateway allow additional 
L7 settings and controls. But there is nothing in Gateway that requires 'ingress' ( i.e. internet traffic going into the VPC).

geps/gep-1924.md
costinm: Agreed. Anything with a FQDN or IP is a valid parentRev - if the WG chooses to.

That doesn't mean we MUST - we can take UX and practical considerations into account.

geps/gep-1924.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Implementation-specific `parentRef`s
+
+The remainder of this GEP will detail considerations and guidelines for mesh implementations if they wish to enable an implementation-specific resource as a parentRef. Recall that the frontend role of a (generic) service is how one calls the service. In the service mesh transparent proxy context, the frontend role (and parentRef by extension) is effectively the matching mechanism for the specified route. For the Service parentRef, this means that the mesh should apply a particular xRoute’s configuration if the destination ip address for a given connection is the ClusterIP of that parentRef Service. If a mesh wishes to use an implementation-specific resource for parentRef, that resource MUST contain layer-appropriate information suitable for traffic matching (e.g. no Host header capture in TCPRoute). For example, consider the following HTTPRoute with an Istio ServiceEntry as a parentRef:
+
+```yaml
+apiVersion: networking.istio.io/v1beta1
+kind: ServiceEntry
```
costinm, line 45: I am trying to convince people to push the "hosts" portion of ServiceEntry to MCS ServiceExport - making it possible to define 'bring your own' names. This may allow us to clean up ServiceEntry, which is now 
a mix of egress, internal and deprecated fields ( WorkloadGroup/etc).

Note that your example should be location: MESH_INTERNAL - the MESH_EXTERNAL example is an egress use case which we agreed to postpone and decouple for now.

geps/gep-1924.md
howardjohn: I am not sure I grok how a ClusterIP Gateway and Service are different. Are they not just syntax differences for the exact same thing?

Given its not right now a "vague idea" and not an implementation, maybe we just have different ideas of what it is/will be, though.

youngnick: They're not different in the functions they provide, I'm arguing that they're different in the implied usage, because a Gateway is intended to be a thing that sits at the logical edge of a cluster and translates. It's possible (and will almost certainly be the case for Cilium) that the act of passing through a Gateway may end up with your traffic being classified differently (security policy wise) that if it passes through a GAMMA Service binding.

We evidently haven't done a good enough job of communicating that this is the main differentiator of a _Gateway_ to other types of networking construct - the _translation_ between "inside" and "outside". The fact that "inside" and "outside" are murky when talking about ClusterIP Gateways doesn't change the fact that this is the intended behavior.

We've also had this come up with Cilium when people go from a pod _inside_ the cluster to the _outside_ of an Ingress entry - because of the Ingress contract, we're defining that to mean that the original source inside the cluster is stripped. Instead the traffic will pick up the "ingress" identity. If this behavior is undesirable, then we need to ensure that there's another construct available - and that's the point of the Service binding - it doesn't have the implied translation that a Gateway does.

keithmattix: Thanks for the discussion! For now, I removed mention of ClusterIP from Gateway discussion

geps/gep-1924.md
keithmattix: Changed everything to internal

geps/gep-1924.md
keithmattix: Added `However, `IPAddress` is NOT disallowed; implementations may use it if they wish.`

youngnick:
This looks fine for its current state to me.

/lgtm

/hold for more review though.

geps/gep-1426.md
```diff
@@ -43,11 +43,15 @@ This GEP uses the [roles and personas](https://gateway-api.sigs.k8s.io/concepts/
...
+
+### Why Service?
+
+The GAMMA initiative has been working to bring service mesh use-cases to the Gateway API spec, taking the best practices and learnings from mesh implementations and codifying them in a spec. Most mesh users are familiar with using the Kubernetes `Service` resource as the foundation for traffic routing. Generaly, this architecture makes perfect sense; unfortunately, `Service` is far too coupled of a resource. It orchestrates IP address allocation, DNS, endpoint collection and propagation, load balancing, etc. For this reason, it **cannot** be the right long-term answer for parentRef binding; however, it is the only feasible option that Kubernetes has for mesh implementations today. We expect this to change (indeed, we hope to be a part of that change), but in the interest of developing a spec now, we must once again lean on the Service resource. However, we will provide provisions to support additional resources as a parentRef.
```
mikemorris, line 50:
```suggestion
The GAMMA initiative has been working to bring service mesh use-cases to the Gateway API spec, taking the best practices and learnings from mesh implementations and codifying them in a spec. Most mesh users are familiar with using the Kubernetes `Service` resource as the foundation for traffic routing. Generaly, this architecture makes perfect sense; unfortunately, `Service` is far too coupled of a resource. It orchestrates IP address allocation, DNS, endpoint collection and propagation, load balancing, etc. For this reason, it **cannot** be the right long-term answer for `parentRef` binding; however, it is the only feasible option that Kubernetes has for mesh implementations today. We expect this to change (indeed, we hope to be a part of that change), but in the interest of developing a spec now, we must once again lean on the `Service` resource. However, we will provide provisions to support additional resources as a `parentRef`.
```

geps/gep-1426.md
```diff
@@ -65,7 +69,7 @@ spec:
...
+In the example above, routing rules have been configured to direct 90% of traffic for the `foo` `Service` to the default "backend" endpoints specified by the `foo` `Service` [`selector`](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) field, and 10% to the `foo-v2` `Service`. This is determined based on the `ClusterIP` (for `Service`) and `ClusterSetIP` (for `ServiceImport`) matching, and for "transparent proxy" mesh implementations would match all requests to `foo.store.svc.cluster.local` (or arbitrary custom suffix, as the hostname is not specified manually) from within the same namespace, all requests to `foo.store.svc.cluster.local` from other namespaces, and all requests to `foo.store.svc.clusterset.local` for multicluster services, within the scope of the service mesh.
```
mikemorris, line 72: I had intentionally omitted the namespace to demonstrate that using the shorter-form DNS within the same namespace should work too.

```suggestion
In the example above, routing rules have been configured to direct 90% of traffic for the `foo` `Service` to the default "backend" endpoints specified by the `foo` `Service` [`selector`](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) field, and 10% to the `foo-v2` `Service`. This is determined based on the `ClusterIP` (for `Service`) and `ClusterSetIP` (for `ServiceImport`) matching, and for "transparent proxy" mesh implementations would match all requests to `foo.svc.cluster.local` (or arbitrary custom suffix, as the hostname is not specified manually) from within the same namespace, all requests to `foo.store.svc.cluster.local` from other namespaces, and all requests to `foo.store.svc.clusterset.local` for multicluster services, within the scope of the service mesh.
```

geps/gep-1426.md
```diff
@@ -109,11 +113,60 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
+
+Currently (v0.7.0), this spec only considers the `Service` resource to be under Core conformance as a `parentRef`. However, Service is not the only resource that can fulfill the frontend role. While the Gateway API spec couldn’t possibly enumerate every existing (and future) frontend-like resource, it can specify a subset of resources that implementations MUST support as parentRefs under as a part of core conformance. Meshes MAY support other implementation-specific resources as parentRefs. The spec maintainers also reserve the right to add additional resources to core conformance as the spec evolves.
+
+#### Extended Conformance
+
+In addition to Service, there are other optional parentRef resources that, if used by implementations, MUST adhere to the spec’s prescriptions. At the time of writing (v0.70), there are two resources in extended conformance: ServiceImport (part of the mcs-api, currently in alpha) and Gateway . The semantics of ServiceImport parentRef binding can be found in GEP 1748 (Note: Headless ServiceImports are out of scope and not currently a part of the spec). The semantics for Gateways in the context of GAMMA implementations are laid out below:
```
mikemorris, line 122:
```suggestion
In addition to Service, there are other optional parentRef resources that, if used by implementations, MUST adhere to the spec’s prescriptions. At the time of writing (v0.7.0), there is one resource in extended conformance: `ServiceImport` (part of the [MCS API](https://github.com/kubernetes-sigs/mcs-api), currently in alpha). The semantics of `ServiceImport` `parentRef` binding can be found in [GEP-1748](https://gateway-api.sigs.k8s.io/geps/gep-1748/) (Note: Headless `ServiceImport` is out of scope and not currently a part of the spec).
```

geps/gep-1426.md
```diff
@@ -109,11 +113,60 @@ Services supported as `backendRefs` SHOULD be consistent with expectations for N
...
+
+##### `Gateways`
+
+There has been much discussion around cluster local Gateways (i.e. Gateways not associated with a traditional load balancer). While there are various potential UX impairments (e.g. what’s the difference between a GAMMA HTTPRoute with a Gateway parentRef and an ingress implementation’s HTTPRoute?), there is no technical reason why a Gateway cannot be a valid GAMMA parentRef if an implementation wishes to do so.
```
mikemorris, line 126: I think the content of this section is good but it should move to the implementation-specific section because it's still mostly theoretical and not something we have immediate plans to include conformance tests for AFAIK.

mikemorris (commented):
Overall LGTM with minor formatting nits.

/lgtm

geps/gep-1426.md
keithmattix: Fair enough

keithmattix:
@mikemorris I think I've addressed all your feedback. When you're ready, you can remove the hold and give the last lgtm

howardjohn (approved):

robscott:
Thanks for all the work on this! Let's aim for lazy consensus here and plan on removing the hold tomorrow unless there are any additional comments.

/lgtm
/approve

robscott:
(This GEP is in a provisional state and therefore merging before v0.7.0 does not affect release)

PR #1935, opened 2023-04-14
GEP-1619: Enumerate common configuration for session persistence and affinity

gcs278:
**What type of PR is this?**
/kind gep

This update adds concrete configuration for session persistence and affinity to help lay a framework for future potential API design.

We are building iteratively on GEP-1619 by taking small steps towards a workable solution. This PR does not introduce concrete API design yet, but that is expected to come in the near future.

Review this GEP PR with Markdown Rendered: https://github.com/gcs278/gateway-api/blob/gep-1619-cookie-configuration/geps/gep-1619.md

NONE

```release-note
NONE
```

### DIFF
--- a/geps/gep-1619.md
+++ b/geps/gep-1619.md
@@ -42,6 +42,8 @@ Session persistence can introduce security and privacy vulnerabilities if not pr
 
 To mitigate these security concerns, it is important to implement session persistence using secure practices, such as using strong session ID generation algorithms, implementing session timeouts, encrypting sensitive data, and monitoring server resources for unusual activity.
 
+IP address reuse may also be a security or privacy concern when using session persistence or session affinity. If Kubernetes reuses an IP address of previously shutdown pod, the new pod may receive session persistent traffic meant for the old pod.
+
 Session affinity introduces fewer security and privacy vulnerabilities since there are no session tokens to protect or exploit.
 
 ### Achieving Session Persistence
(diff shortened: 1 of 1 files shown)


geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
-
+In this section, we will describe how implementations achieve session persistence and affinity, along with a breakdown of related configuration options. Input from implementations is appreciated to complete this information.
+
+In the following tables, we will example two types of APIs:
+1. Dataplane APIs
+2. Implementation APIs
+
+Generally, the implementation API programs the dataplane API; however these two are not always clearly separated. The two types of APIs can use different API structures for configuring the same feature. Examining the dataplane APIs helps to remove the layer of API abstraction that implementations provide. Removing this layer avoids situations where implementations don’t fully implement all capabilities of a dataplane API or obfuscate certain configuration around session persistence. On the other hand, examining implementation APIs provides valuable data points in what implementations are interested in configuring.
+
+**Session Persistence**
```
gcs278, line 85: Review the tables here: https://github.com/gcs278/gateway-api/blob/gep-1619-cookie-configuration/geps/gep-1619.md

youngnick (commented):
This is an _outstanding_ first pass at a GEP, and I will be pointing to it as such in the future.

I'll approve here with a hold so we can get some more LGTMs though.

/approve
/hold

sanjaypujare:
@robscott asked me to take a look at this PR since I have similar ideas for an API for stateful session affinity that is currently modeled as a vendor extension.

sanjaypujare:
For the "Security and Privacy Implications" section we may also want to mention the impact of IP reuse - say in Kubernetes a new pod gets the IP of a recently shut down pod and hence receives session persistent traffic meant for the old pod.

sanjaypujare:
For the "Session Initiation" section: I think there is a 3rd case of the client initiating a session - by not including a cookie or a header in a request. A cookie or header is always present for an "in-session" request but for a request initiating a new session this will not be present. In fact this case is probably more common than a "gateway" initiating session.

Also I am wondering if we have to mandate that a client has to use a cookie or header returned in a previous response. I think there might be cases where a client may use a locally generated value for a header (or even a cookie but less likely) which is then used by a gateway/proxy to figure out if this is for an existing session or a new session.

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
-| Traefik Proxy 	| Traefik 	| [Cookie-Based](https://doc.traefik.io/traefik/routing/services/#sticky-sessions) 	| No 	| [Docs](https://traefik.io/glossary/what-are-sticky-sessions/) mention consistent hashing, but Traefik [only offers round robin](https://doc.traefik.io/traefik/routing/services/#load-balancing) (not consistent). 	|
-| Java Servlet 	| N/A 	| [Cookie-Based](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getSession--), [URL Encoded](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#isRequestedSessionIdFromURL--) 	| N/A 	| Java Servlets don't natively support proxy functions. 	|
-| Apache httpd 	| N/A 	| [Cookie-Based](https://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html), [URL Encoded](https://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html) 	| N/A 	|  N/A	|
-
+In this section, we will describe how implementations achieve session persistence and affinity, along with a breakdown of related configuration options. Input from implementations is appreciated to complete this information.
```
sanjaypujare, line 77: I would like to include a reference to "stateful session affinity" implemented in gRPC described at https://github.com/grpc/proposal/blob/master/A55-xds-stateful-session-affinity.md . However the distinction between "persistence" and "affinity" is not made in this doc or elsewhere. So it might be useful to mention that the 2 terms are used interchangeably elsewhere in the industry.

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+
+**Session Persistence**
+
+| **Technology** 	| **Technology Type** 	| **Session Persistence Type** 	| **Configuration Options** 	| **Configuration Association (Global, Gateway, Route, or Backends)** 	| **Notes** 	|
+|---	|---	|---	|---	|---	|---	|
+| Envoy 	| Dataplane 	| [Cookie-Based](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto) 	| Name=name<br>Path=path<br>TTL=duration 	| [HttpConnectionManager](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto) (Route) 	| Envoy does not natively support cookie modification, but Contour has implemented it via Lua ([reference](https://github.com/projectcontour/contour/blob/main/design/cookie-rewrite-design.md)). Lua scripts can be added per route in Envoy via the [Lua envoy filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter) and can modify. [Related Issue](https://github.com/envoyproxy/envoy/issues/15612). 	|
```
sanjaypujare, line 89: Not sure what you mean by "Envoy does not natively support cookie modification" but according to this code https://github.com/envoyproxy/envoy/blob/main/source/extensions/http/stateful_session/cookie/cookie.cc#L11 Envoy does set a generated cookie in the response.

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+| HAProxy 	| Dataplane 	| [Cookie-Based](https://docs.haproxy.org/2.6/configuration.html#4.2-cookie) 	| name=name<br>[rewrite \| insert \| prefix ]<br>indirect<br>nocache<br>postonly<br>preserve<br>httponly<br>secure<br>domain=domain<br>maxidle=idle<br>maxlife=life<br>dynamic<br>attr=value 	| [Default or Backends](https://docs.haproxy.org/2.6/configuration.html#4.2-cookie) (Global or Backends) 	| HAProxy allows for operational cookie strategy configuration (i.e. when/how HAProxy should inject cookies) 	|
+| Acnodal EPIC 	| Implementation (Envoy) 	| N/A 	| Supports Gateway API Only* 	| N/A 	| *Acnodal Epic solely uses Gateway API; therefore, it doesn’t yet have a way to configure session persistence. [Acnodal EPIC Docs](https://www.epick8sgw.io/docs/) 	|
+| Istio 	| Implementation (Envoy) 	| [Cookie-Based](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB-HTTPCookie) 	| Name=name<br>Path=path<br>TTL=duration 	| [ConsistentHashLB](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB) (Backends) 	| N/A 	|
```
sanjaypujare, line 94: Istio also has this label based support for persistent sessions : see https://github.com/costinm/istio/blob/cb3745d1f97e5da47a3a5fef6f6f60a1abaa6165/pilot/pkg/features/pilot.go#L117 . This is based on the Envoy's sessionAffinity filter described elsewhere and not the ConsistentHashLB. CC @costinm

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+|---	|---	|---	|---	|
+| Envoy 	| Dataplane 	| Consistent Hashing via [Maglev](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#maglev) and [Ring Hash](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#ring-hash) 	| N/A 	|
+| Nginx 	| Dataplane 	| [Consistent Hashing](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash) via IP or other connection attributes (Nginx Open Source) 	| N/A 	|
+| HAProxy 	| Dataplane 	| [Stick Tables](https://cbonte.github.io/haproxy-dconv/2.4/configuration.html#4-stick-table) using connection attributes 	| N/A 	|
+| Acnodal EPIC 	| Implementation (Envoy) 	| Supports Gateway API Only* 	| *Acnodal Epic solely uses Gateway API; therefore, it doesn’t yet have a way to configure session affinity. [Acnodal EPIC Docs](https://www.epick8sgw.io/docs/) 	|
+| Istio 	| Implementation (Envoy) 	| Consistent Hashing via [ConsistentHashLB](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB) with Maglev or Ring Hash  	| N/A 	|
```
sanjaypujare, line 125: Looks like the same as the corresponding Istio entry under persistence.

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+Session cookie strategy is configuration specific to the creation of the cookie by the dataplane. These are not specific configuration parameters of the cookie itself, but instead the operation to take when inserting the cookie such as:
+- Rewriting an existing cookie value
+- Inserting a new cookie
+- Prefixing an existing cookie
+- Preserving an existing cookie value
+- Only adding cookies on specific HTTP methods (POST/GET)
+
+### Session Affinity API
+
+TBD - See open question about session affinity
```
sanjaypujare, line 188: Link to the section/text? (Most probably line 192)

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+
 ### Open Questions
 
-TBD
+- Should we include both session persistence and session affinity API configurations in this GEP or just focus on session persistence?
```
sanjaypujare, line 192: Another option is to not differentiate between persistence and affinity and combine both into a single concept.

geps/gep-1619.md
```diff
@@ -106,10 +200,6 @@ The following are items that we intend to resolve before we consider this GEP im
 - We need to add a small section on compliance regarding the browser and client relationship.
 - We need to design and document an API for session persistence and/or session affinity in the Gateway API spec.
 
-## API
```
sanjaypujare, line 109: For line 198 (session draining) please check out https://github.com/grpc/proposal/blob/master/A55-xds-stateful-session-affinity.md#background (last para).

sanjaypujare: For line 197 (needs and use-cases) check out https://static.sched.com/hosted_files/kccnceu2023/c4/Autoscaling%20Elastic%20Kubernetes%20Infrastructure%20for%20Stateful%20Applications%20using%20Proxyless%20gRPC%20and%20Istio.pdf (slide 17 onwards).

geps/gep-1619.md
```diff
@@ -106,10 +200,6 @@ The following are items that we intend to resolve before we consider this GEP im
...
-
-TBD
-
 ## Alternatives
 
 ### Alternate Naming
```
sanjaypujare, line 205: My suggestion is to not make a big distinction between persistence and affinity and mention that these 2 terms are used interchangeably or with various meanings in the industry. So we could just use the term session persistence through out to also include the session affinity (as described here).

One more aspect to be explored: what happens when the affinity/persistence is broken because the given backend is not up/healthy? If that's an error case how should that be handled? Should the API dictate the http error code? Or should the API dictate fall back behavior? One fallback behavior is that when a backend for a session is not up or healthy the load balancer finds a new backend (using the applicable load balancing policy) and create a new cookie/header thereby creating a new session for this otherwise failing RPC.

geps/gep-1619.md
gcs278: At a minimum, I agree with redefining "session affinity". It's not immediately clear to me, based on the words themselves, what the difference between persistence and affinity is, and lots of implementations use them interchangeably and I think this would be confusing if not changed. But I've heard differing thoughts on whether we should still keep the idea of strong session persistence (cookies/headers) and weak session persistence (connection attribute hashing). 

John mentioned this in the first PR review: https://github.com/kubernetes-sigs/gateway-api/pull/1643#discussion_r1068250615. Per the comment, I developed the definition with his comment, but kept the labels of persistence/affinity:

> In my view, the primary difference between session types is whether the client directly specifies the backend they want OR if some attribute of the request is just used to consistently send to a backend.

Here are my thoughts I'm working through right now:
- If our future API for persistence and affinity is designed in such as way that the configurations live together, such as Istio's `ConsistentHashLB`, the to me, it makes sense we can redefine the concept of persistence and affinity into the same "session persistence" bucket.
  - However, grouping loadbalancing algorithms in with session persistence feel awkward to me. Some loadbalancing algorithms provide consistent hashing, others don't. Thinking ahead, GWAPI could potentially support configuring the algorithm in the future, but if some algorithms are selected in the session persistence API object, then how would you configure the non-consistent hashing algorithms? I think we'd have to separate to into something like [SimpleLB](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-SimpleLB) and  [ConsistentHashLB](https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings-ConsistentHashLB) like Istio. I think these concepts of "LBs" might be hard to translate to GWAPI. Though, we'd still have this problem if we had a API specifically for session affinity.
  - The other downside, is that we lose the distinction of guaranteed session persistence (headers/cookies) vs. best attempt (using connection properties, or hashing that may change or rebalance upon new backend)
- If our future API will keep persistence and affinity as separate objects, then it makes sense to define them differently (current state of this GEP).
  - May be more confusing vs. the value added for the separation.

So I'd like to start a discussion, involving @costinm @howardjohn and whoever has input, on how we should define these concepts. I lean towards the simplicity of calling everything session persistence.

gcs278 (commented):
Still working on responses, but wanted to get my discussion on persistence/affinity out.

geps/gep-1619.md
sanjaypujare: Thinking more about it I agree that there are 2 concepts and I would call them session persistence and affinity (without "session").

- session persistence: there is a concept called session where a client creates a session by sending the first request in a session without any special attribute (e.g. cookie). The infrastructure (proxy/LB/gateway etc) routes this "first request" based on applicable policies. The response contains the attribute (cookie) which the client includes in all the subsequent requests in the session thereby explicitly indicating the session semantics. The infrastructure respects the cookie to route the requests to the same backend as the first one thereby maintaining session persistence.

- affinity: this just involves the infrastructure (LB/proxy etc) using the request attributes (e.g. a header) to select a backend thereby maintaining affinity based on those attributes. Although the client application can simulate a "session" by injecting that header with a specific value, there is no special first request initiating a session.

So I am okay with keeping the existing terminology (session persistence and session affinity)

geps/gep-1619.md
```diff
@@ -106,10 +200,6 @@ The following are items that we intend to resolve before we consider this GEP im
 - We need to add a small section on compliance regarding the browser and client relationship.
 - We need to design and document an API for session persistence and/or session affinity in the Gateway API spec.
```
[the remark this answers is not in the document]
gcs278: Great reference, thanks. I'll add this as a reference in the reference section of the GEP. After this PR, I think we will be ready to propose an API.

geps/gep-1619.md
gcs278: That's valid. Just to be clear, are you saying to drop the "session" from "session affinity" for just "affinity"? Or keep it as "session affinity"?

I now realize Envoy defines them as [stateful sessions](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/http/stateful_session/cookie/v3/cookie.proto) as so does your [gRPC proposal](https://github.com/grpc/proposal/blob/master/A55-xds-stateful-session-affinity.md#background). The word "stateful" is used to distinguish between cookie-based session persistence and consistent hashing. That's another idea.

geps/gep-1619.md
gcs278: Yea this is a great reference. Added to the "references" section. Once we settle on the naming, I can mention the naming deltas somewhere.

geps/gep-1619.md
gcs278: Ack - added both links as references, but I'll keep this in mind as I work on this GEP.

geps/gep-1619.md
gcs278: Discussion for naming is here: https://github.com/kubernetes-sigs/gateway-api/pull/1935/files#r1177654213

Great question. Added as an open question for now to address later, as it seems like an implementation detail we can address in the near future.

Though, I don't fully understand the implications of Gateway API dictating failure modes and edge cases such as this error scenario or even something like connection draining. Is there precedent for this in the API today? If an implementation handled this error case differently, would they be non-compliant? How prescriptive is Gateway API for something like session persistence?

costinm:
> For the "Session Initiation" section: I think there is a 3rd case of the client initiating a session - by not including a cookie or a header in a request. A cookie or header is always present for an "in-session" request but for a request initiating a new session this will not be present. In fact this case is probably more common than a "gateway" initiating session.
> 
> Also I am wondering if we have to mandate that a client has to use a cookie or header returned in a previous response. I think there might be cases where a client may use a locally generated value for a header (or even a cookie but less likely) which is then used by a gateway/proxy to figure out if this is for an existing session or a new session.

I don't think we should mandate - there are use cases for using a cookie or header obtained in other ways to
bootstrap the session. I also don't see what mandating would do - there is no way to enforce that. 

A client may also decide to stop passing the cookie - for example to start a new session.

costinm:
One lesson from Istio is that mixing layers and too many things in one API is not ideal...
There are plenty of load balancing policies - some ring hash based, etc. We may need a way for users to pick one. But I don't think 'cookie or header based strong affinity' should be mixed with this.

I mentioned in one of the comments that K8S already has an equivalent in Service.sessionAffinity=ClientIP - 

"If you want to make sure that connections from a particular client are passed to the same Pod each time, you can select the session affinity based on the client's IP addresses by setting .spec.sessionAffinity to ClientIP for a Service (the default is None).

Session stickiness timeout
You can also set the maximum session sticky time by setting .spec.sessionAffinityConfig.clientIP.timeoutSeconds appropriately for a Service. (the default value is 10800, which works out to be 3 hours)."

Obviously we can't extend Service - and the cookie/header session affinity works in a broader context, but we should
take this into account too. It is not a 'load balancing' option, but session affinity option.

sanjaypujare:
> ...
> I don't think we should mandate - there are use cases for using a cookie or header obtained in other ways to bootstrap the session. I also don't see what mandating would do - there is no way to enforce that.
> 
> A client may also decide to stop passing the cookie - for example to start a new session.

Yes, we are saying the same thing (I think) so I am in agreement.

sanjaypujare:
> One lesson from Istio is that mixing layers and too many things in one API is not ideal... There are plenty of load balancing policies - some ring hash based, etc. We may need a way for users to pick one. But I don't think 'cookie or header based strong affinity' should be mixed with this.
> ...
> . It is not a 'load balancing' option, but session affinity option.

Actually session affinity/persistence is kind of related to "load balancing" - it is the exact opposite of load balancing. The goal is to keep sending requests (that match certain criteria) to a single backend instead of trying to balance load. So if all of the requests in a certain load balancer/proxy case have to obey the session persistence/affinity semantics then there is no (or very little) load balancing happening.

Given this, I think session persistence/affinity is an attribute of a load balancing policy. What it basically says is that if a request has to be routed based on session affinity/persistence, then route it but if that does not apply (e.g. there is no cookie or header for it) then use the load balancing algorithm dictated by the policy.

geps/gep-1619.md
sanjaypujare: > implications of Gateway API dictating failure modes and edge cases...
> How prescriptive is Gateway API for something like session persistence?

Good questions and I guess there are no good answers other than saying "it depends". My thoughts are as follows: to make the API truly portable and useful it will be good to have a tight spec that includes failures modes, fallback behavior and interaction with (on dependency on) other APIs such as the things we have talked about.

geps/gep-1619.md
gcs278: Okay - got it. Costin mentioned https://kubernetes.io/docs/reference/networking/virtual-ips/#session-affinity. Looks like Kubernetes refers to clientIP-based consistent sessions as "session affinity", which technically aligns to our naming conventions, but Kubernetes doesn't go as far to define that persistence != affinity as we are trying to.

I think the designations of persistence and affinity make sense, but I will most likely bring up in one our meetings to get feedback. Sounds like we agree on defining two separate things, but it's probably worth making sure there isn't a better naming convention (say stateful sessions vs stateless sessions or strong session persistence vs. weak session persistence). I worry "session affinity" is overloaded.

If we end up designing the persistence and affinity APIs to be tightly coupled, say LoadbalancerPolicy-centric design (like you are advocating for in https://github.com/kubernetes-sigs/gateway-api/pull/1935#issuecomment-1525438073), we might never need to distinguish between the two (they all be in the same bucket).

gcs278:
> One lesson from Istio is that mixing layers and too many things in one API is not ideal...

> Given this, I think session persistence/affinity is an attribute of a load balancing policy. What it basically says is that if a request has to be routed based on session affinity/persistence, then route it but if that does not apply (e.g. there is no cookie or header for it) then use the load balancing algorithm dictated by the policy.

@sunjayBhatia @costinm I think this is a really good back and forth. In my other comments, I think we've come to the conclusion it's best to keep separate definitions of persistence and affinity, regardless of our future API design. It will be more important to keep these definitions if we keep the API structures separate, but it doesn't hurt if we decide to bundle them together in a LoadBalancing-oriented API structure (like istio).

This API design debate seems like a really good conversation to bring up in the next PR in which we start designing a spec.

youngnick:
This has been a great discussion, but I agree that it's probably time to round out this initial cut, get the Provisional version merged, and leave implementation for later.

geps/gep-1619.md
```diff
@@ -42,6 +42,8 @@ Session persistence can introduce security and privacy vulnerabilities if not pr
 
 To mitigate these security concerns, it is important to implement session persistence using secure practices, such as using strong session ID generation algorithms, implementing session timeouts, encrypting sensitive data, and monitoring server resources for unusual activity.
 
+IP address reuse may also be a security or privacy concern when using session persistence or session affinity. If Kubernetes reuses an IP address of previously shutdown pod, the new pod may receive session persistent traffic meant for the old pod.
```
gcs278, line 45: @sunjayBhatia regarding https://github.com/kubernetes-sigs/gateway-api/pull/1935#issuecomment-1523022313

geps/gep-1619.md
gcs278, line 1: @sanjaypujare regarding:
> For the "Session Initiation" section: I think there is a 3rd case of the client initiating a session - by not including a cookie or a header in a request. A cookie or header is always present for an "in-session" request but for a request initiating a new session this will not be present. In fact this case is probably more common than a "gateway" initiating session.

I'm not sure if I completely understand. Are you saying, in addition to the gateway and backend server being able to initiate a session, the client can as well? The client would send a cookie before the gateway or backend server ever initiated it?

va7ish:
> Actually session affinity/persistence is kind of related to "load balancing" - it is the exact opposite of load balancing. The goal is to keep sending requests (that match certain criteria) to a single backend instead of trying to balance load. So if all of the requests in a certain load balancer/proxy case have to obey the session persistence/affinity semantics then there is no (or very little) load balancing happening.

I wouldn't say `its the exact opposite of load balancing`. Rather its a two tier decision logic where we say 1) if we have a session identity provided in the request then route according to previous state/contract otherwise 2) Load balance as per LB configuration

sanjaypujare:
> ...
> I wouldn't say `its the exact opposite of load balancing`. Rather its a two tier decision logic where we say 1) if we have a session identity provided in the request then route according to previous state/contract otherwise 2) Load balance as per LB configuration

I was trying to take a high level view in terms of the goal of each feature: a load balancer's goal is to spread (or "balance") the load among a set of backends (equally or based on some other criteria such as capacity) whereas the goal of a persistence/affinity feature is to *not* spread it but rather keep it stuck to one backend as far as possible.

We can model the behavior as you have described for the purpose of defining the API: if the policy dictates routing based on session identity then do it otherwise load balance based on some LB configuration.

geps/gep-1619.md
sanjaypujare: The way I see it the client application which is generating HTTP requests (say) decides whether a given request belongs to an existing session or not. So the client application includes a cookie if the request belongs to an existing session but if a new session is intended, then it does not include a cookie. So the intent to create a new session is signaled by the client but the actual session creation happens in the gateway or backend by creating a new cookie so I think we might be saying the same thing but saying it differently.

sanjaypujare:
> This has been a great discussion, but I agree that it's probably time to round out this initial cut, get the Provisional version merged, and leave implementation for later.

I would like to see a proposal for the API but it still says TBD in the "Session Affinity API" section. Will it be part of this PR?

sanjaypujare:
> 
> This API design debate seems like a really good conversation to bring up in the next PR in which we start designing a spec.

So this PR won't include the API definition? Just clarifying.

geps/gep-1619.md
sanjaypujare: Got it.

You probably tagged the wrong person - assuming you were trying to tag me (sanjaypujare vs sunjayBhatia)

geps/gep-1619.md
gcs278: Ah okay I see. The client can "signal" that it wants a new session, but the session (the cookie) is created by the gateway or the backend. 

Added a small paragraph to describe this concept.

shaneutt (approved):
As this continues to be `Provisional` I'm personally comfortable with adding what's here and continuing to iterate from this point.

We should capture any outstanding concerns that aren't easily rectified yet as part of a `TODO`-type section.

gcs278:
> We should capture any outstanding concerns that aren't easily rectified yet as part of a TODO-type section.

Added an additional open question in https://github.com/kubernetes-sigs/gateway-api/compare/df8eba0fa6ee7747822b0a6381016be541a6e21f..37d18517ea87f4b2271bad98e5f3436bf993b4af regarding discussion on API design.

Otherwise, I think our TODO section captures what we need todo. The biggest one being `We need to design and document an API for session persistence and/or session affinity in the Gateway API spec`. We'll tackle that in the next PR.

geps/gep-1619.md
gcs278: Update: sounds like we are continuing with separate definitions for affinity and persistence, so leaving this as is.

geps/gep-1619.md
robscott: @gcs278 can you capture some of this context in the GEP before merging and link to this thread?

robscott (commented):
Thanks @gcs278, this is really great. A few small requests but mostly LGTM.

geps/gep-1619.md
```diff
@@ -74,28 +74,122 @@ While session persistence uses attributes in the application layer, session affi
...
+|  	| Implementation (Nginx) 	| [Header-Based](https://apisix.apache.org/docs/apisix/terminology/upstream/#header) 	| hash_on=[vars \| header \| cookie \| consumer]<br>key=header_name 	| [Upstream](https://apisix.apache.org/docs/apisix/admin-api/#upstream) (Route or Backends) 	| N/A 	|
+| NGINX Kubernetes Gateway 	| Implementation (Nginx) 	| N/A 	| Supports Gateway API Only* 	| N/A 	| *NGINX Kubernetes Gateway solely uses Gateway API; therefore, it doesn’t yet have a way to configure session persistence. [Nginx Kubernetes Gateway Docs](https://github.com/nginxinc/nginx-kubernetes-gateway) 	|
+| Java Servlet 	| Web Server 	| [Cookie-Based / URL-Encoded](https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpSession.html) 	| invalidate()<br>setAttribute(String name, Object value)<br>setMaxInactiveInterval(int interval) 	| N/A 	| Java Servlets do not natively support proxy functions. 	|
+| Apache httpd 	| Web Server 	| [Cookie-Based / URL-Encoded](https://httpd.apache.org/docs/2.4/mod/mod_proxy_balancer.html) 	| Cookie Attributes 	| N/A 	| N/A 	|
+
+**Session Affinity**
```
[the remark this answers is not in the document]
gcs278: I have one for "Google Cloud Run". I don't remember where I got that from, but is that the same as GKE? Should these be two separate implementations?

See: https://github.com/kubernetes-sigs/gateway-api/pull/1935/files#diff-afa0e827962fe48152635217c6bc954020d5fec3578af4a9e5f744a2b4568fffR112

shaneutt:
I like how Rob's review is from 14 hours ago, but the individual comments say last week. He wasn't kidding when he said he had a review in progress for a while that he just hadn't submitted yet :joy:

geps/gep-1619.md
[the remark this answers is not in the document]
gcs278: Done

geps/gep-1619.md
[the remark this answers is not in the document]
gcs278: Added Google GKE, but left Google Cloud Run, let me know if I am conflating the two

geps/gep-1619.md
gcs278: Done: https://github.com/kubernetes-sigs/gateway-api/pull/1935/files#diff-afa0e827962fe48152635217c6bc954020d5fec3578af4a9e5f744a2b4568fffR80-R89

There were a couple ideas floating in this thread, but I added a section on the relationship of session persistence and session affinity as we have discussed it in the PR.

costinm:
CloudRun is different - and I'm sure google has a few more :-)

geps/gep-1619.md
```diff
@@ -67,35 +69,147 @@ For both header-based and cookie-based sessions, either the gateway or the backe
...
+|  	|  	| [Header-Based](https://docs.solo.io/gloo-edge/latest/reference/api/envoy/api/v2/route/route.proto.sk/#hashpolicy) 	| Name=name 	| [Route](https://docs.solo.io/gloo-edge/latest/reference/api/envoy/api/v2/route/route.proto.sk/#route) (Route) 	| N/A 	|
+| Google CloudRun 	| Implementation / Dataplane 	| [Cookie-Based](https://cloud.google.com/run/docs/configuring/session-affinity) 	| Enabled / Disabled 	| [Service](https://cloud.google.com/run/docs/configuring/session-affinity) (Backends) 	| Only allowed to turn off or on, no other configuration items 	|
+| Google GKE 	| Implementation / Dataplane 	| [Cookie-Based](https://cloud.google.com/load-balancing/docs/backend-service#session_affinity) 	| GENERATED_COOKIE or HTTP_COOKIE=name<br>cookieTtlSec 	| [Backend Policy](https://cloud.google.com/kubernetes-engine/docs/how-to/configure-gateway-resources#session_affinity) (Backends) 	| GKE [lists](https://cloud.google.com/load-balancing/docs/backend-service#bs-session-affinity) the products that can do persistence/affinity mode. 	|
```
robscott, line 118: Tiny NIt: Either "Google Kubernetes Engine" or "GKE" would be preferred to avoid repetitive names.

geps/gep-1619.md
```diff
@@ -67,35 +69,147 @@ For both header-based and cookie-based sessions, either the gateway or the backe
...
+| Flomesh Service Mesh 	| Implementation / Dataplane (Pipy) 	| ? 	| ? 	|
+| Gloo Edge 2.0 	| Implementation (Envoy) 	| Consistent Hashing via HashPolicy of [SourceIP](https://docs.solo.io/gloo-edge/latest/reference/api/envoy/api/v2/route/route.proto.sk/#connectionproperties) 	| N/A 	|
+| Google CloudRun	 	| Implementation / Dataplane 	| None 	| [Google CloudRun Docs](https://cloud.google.com/run/docs) 	|
+| Google GKE 	| Implementation / Dataplane 	| [CLIENT_IP_NO_DESTINATION](https://cloud.google.com/load-balancing/docs/backend-service#client-ip-no-destination-affinity) and [CLIENT_IP](https://cloud.google.com/load-balancing/docs/backend-service#client_ip_affinity) 	| GKE [lists](https://cloud.google.com/load-balancing/docs/backend-service#bs-session-affinity) the products that can do persistence/affinity mode. Note: Session affinity settings are fulfilled only if the load balancing locality policy (LocalityLbPolicy) is set to RING_HASH or MAGLEV. 	|
```
robscott, line 147: Same tiny nit on name here, also might be worth noting that GKE's options are exclusive - you can't configure both affinity and persistency at the same time (and in GKE everything is labeled as affinity).

robscott (commented):
Thanks @gcs278! A couple very tiny nits but these don't need to block merging.

/approve

geps/gep-1619.md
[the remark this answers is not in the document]
robscott: LGTM, they are different products.

youngnick:
/lgtm

I'll leave the hold though until @gcs278 has a chance to address those nits.

geps/gep-1619.md
gcs278: Whoops, "Google Google Kubernetes Engine".

Will do.

geps/gep-1619.md
gcs278: Done.

robscott:
Thanks @gcs278!

/lgtm

youngnick:
Noone ever mentioned #1619 to create the link, sigh.

PR #1956, opened 2023-04-20
Bump k8s.io to v0.27.1

zirain:
**What type of PR is this?**

/kind test

Bump k8s.io to v0.27.1

Fixes #1942

```release-note
Bump k8s.io to v0.27.1
```


zirain:
@shaneutt seems everything is pass now.

cmd/admission/main.go
```diff
@@ -70,7 +71,8 @@ func main() {
 	}
 
 	server := &http.Server{
-		Addr: ":8443",
+		Addr:        ":8443",
+		ReadTimeout: 5 * time.Second, // for Potential Slowloris Attack
```
shaneutt, line 75: So keep in mind that patching for G112 only requires modification of the `ReadHeaderTimeout` option, whereas you've configured the higher level and more encompassing `ReadTimeout` here. I think 5 seconds is too aggressive for an overall read timeout, particularly because this is our webhook and is used for production systems which at times may be under significant load that would impact that webhook. It might be slightly too aggressive for a header read timeout too. I think I would be in favor of specifically calling out the header read option, and to provide slightly more time as a first step here:

```suggestion
		ReadHeaderTimeout: 10 * time.Second, // for Potential Slowloris Attack (G112)
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -82,11 +82,11 @@ func GWCMustHaveAcceptedConditionAny(t *testing.T, c client.Client, timeoutConfi
 func gwcMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, gwcName, expectedStatus string) string {
 	t.Helper()
 
-	var controllerName string
-	waitErr := wait.PollImmediate(1*time.Second, timeoutConfig.GWCMustBeAccepted, func() (bool, error) {
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-		defer cancel()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
```
shaneutt, line 86: Would you mind please pointing out which linter was tripped that led to this change?

shaneutt (changes requested):
Thanks for taking this on. I have some comments relevant to timeouts set, and in general it would be helpful if the lint failure which was the impetus for a change here was commented along with the diff i you could please.

conformance/utils/kubernetes/helpers.go
zirain: `wait.PollImmediate` is deprecated

go.mod
```diff
@@ -6,14 +6,14 @@ require (
...
+	k8s.io/api v0.27.1
+	k8s.io/apiextensions-apiserver v0.27.1
+	k8s.io/apimachinery v0.27.1
+	k8s.io/client-go v0.27.1
+	k8s.io/code-generator v0.27.1
 	k8s.io/klog/v2 v2.90.1
-	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
-	sigs.k8s.io/controller-runtime v0.14.6
+	k8s.io/utils v0.0.0-20230209194617-a36077c30491
+	sigs.k8s.io/controller-runtime v0.13.1-0.20230417175046-71885c9b8a6b
```
sunjayBhatia, line 16: just pointing out since we're in the same boat in contour, this is an unreleased version

zirain: this will be upgraded by bot when controller-runtime release new version.

conformance/utils/kubernetes/helpers.go
howardjohn: PollUntilContextTimeout already sets a timeout on the context , Ithink the one you set is ignored

zirain: yes, you're right, fix it now.

zirain:
@shaneutt anything left?

shaneutt:
Sorry for the slow responses, last week was Kubecon week and it was incredibly busy. I'll take another pass at this soon.

go.mod
shaneutt: That sounds alright, I've created a follow-up issue as well so that we can't forget (and also includes some other maintenance items to check before GA): https://github.com/kubernetes-sigs/gateway-api/issues/1970

shaneutt:
merge conflict needs fixing when you have a second, I resolved a couple of open comments :+1:

conformance/utils/roundtripper/roundtripper.go
```diff
@@ -163,7 +163,7 @@ func (d *DefaultRoundTripper) CaptureRoundTrip(request Request) (*CapturedReques
 		fmt.Printf("Sending Request:\n%s\n\n", formatDump(dump, "< "))
 	}
 
-	resp, err := client.Do(req)
+	resp, err := client.Do(req) // nolint:bodyclose
```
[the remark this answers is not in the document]
zirain: yes, I think this's error of `bodyclose`, open an issue on upstream https://github.com/timakin/bodyclose/issues/49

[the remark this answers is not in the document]
zirain: @shaneutt PTAL

shaneutt (approved):

shaneutt:
Looks good, would like a double check from another maintainer

/cc @youngnick @robscott

conformance/utils/kubernetes/helpers.go
```diff
@@ -289,10 +283,7 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, timeoutConfig con
 	t.Helper()
 
 	var ipAddr, port string
-	waitErr := wait.PollImmediate(1*time.Second, timeoutConfig.GatewayMustHaveAddress, func() (bool, error) {
-		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
robscott, line 293: I'm not that familiar with this change, but why aren't we passing this `context.WithTimeout()` instead of `context.Background()` into all the `wait.PullUntilContextTimeout()` funcs?

zirain: `wait.PollImmediate` is marked as deprecated

zirain: kindly ping @robscott

robscott: Right, I get that `wait.PollImmediate` is deprecated, I was just expecting that we needed to pass a context that actually had a timeout to the function based on the name of the function (for example the context removed from line 293 above). At least to me `PollUntilContextTimeout` suggests that you should pass a context with a timeout to the function.

After looking at the actual function, it looks like the current interpretation in this PR is correct, and the name of the function is just confusing. Honestly based on that comment I find the usage and naming of PollUntilContextCancel more sensible, but that's not particularly relevant to this PR.

https://github.com/kubernetes/kubernetes/blob/ff5a582031108cc30d6a00032d1c50e99f7eaee4/staging/src/k8s.io/apimachinery/pkg/util/wait/poll.go#L36-L45C1

robscott (commented):
Thanks @zirain! This change LGTM, but given the scale of the chance and that we're about to  cut our final v0.7.0 release, I'd rather hold off on merging this until that releases.

/hold

shaneutt (approved):
Still looks good, just still waiting on the `v0.7.0` hold for branching.

/lgtm

zirain:
kindly ping @robscott , can this get in now?

robscott:
Thanks @zirain!

/lgtm
/hold cancel

PR #1957, opened 2023-04-21
Add conformance test flag to set implementation labels on mesh namespaces

frankbu:
**What type of PR is this?**

/test
/area conformance

Some mesh implementations require namespaces that are part of the mesh to include implementation-specific labels. For example, Istio requires an `istio-injection=enabled` label on mesh namespaces to turn on sidecar injection.

This PR adds a new flag that can be used to pass in labels that should be added to namespaces where mesh-enabled services are deployed (e.g., `--namespace-labels "istio-injection=enabled"` can be set to run on Istio).

This new flag is implemented using the existing NamespaceLabels field in suite.Options. This PR is only making it settable via the command line.

Note that this PR also includes a small fix in pod.go to check that the expected backends are being called. Without this fix, the split-mesh test isn't actually confirming that the correct backends are called, just that some backend has been called.

Fixes #

```release-note
NONE
```


conformance/utils/suite/suite.go
```diff
@@ -62,6 +62,11 @@ type Options struct {
 	BaseManifests    string
 	MeshManifests    string
 	NamespaceLabels  map[string]string
+
+	// MeshNamespaceLabels are implentation-specific labels that should be added
+	// to namespaces where mesh-enabled services will be deployed.
+	MeshNamespaceLabels map[string]string
```
howardjohn, line 68: Why do we need this when we already have `NamespaceLabels`? It only applies to "mesh workload" namespaces not everywhere?

frankbu: >  It only applies to "mesh workload" namespaces not everywhere?

Yes, that is the difference.

It wasn't clear to me what the intended use of `NamespaceLabels` is, so I added another field. What kind of labels would we want to add to ALL namespaces, mesh or not?

frankbu: @shaneutt, yes, that is my assumption. Mesh egress tests, for example.

frankbu: Actually the more general problem is that if one wants to run both the Gateway and Mesh tests, there is no way to label only the mesh namespace(s), but not the ones hosting plain k8s services that the Gateway/Route tests use. Maybe that's what we want? When running Mesh and Gateway tests together, the Gateway tests switch to mesh-ingress (instead of k8s-ingress) tests?

robscott: @frankbu This seems like something that's maybe a bit too implementation-specific to include in our conformance test helpers unless I'm missing something. Would it be possible to just have mesh tests runs separately from ingress tests?

frankbu: @robscott, yes, absolutely, I think that would be fine. The only potential issue is if we have mesh tests that want to call services outside of the mesh (egress testing). In Istio, we use a "non-mesh" namespace to run services that simulate external services for this kind of testing.

frankbu: @robscott @shaneutt I removed `MeshNamespaceLabels` and just use the existing `NamespaceLabels` for now. If in the future we need to differentiate mesh/not-mesh namespaces we can figure out a way to do it then.

So now this PR is only adding a flag to set `NamespaceLabels` and fixing up pod.go to properly check that the version routing test is actually calling the expected backends.

robscott:
@frankbu can you update the PR description to match what we've settled on? It looks like one of your latest commits caused CLA to fail, so that will also need to be fixed before this can be merged.

/ok-to-test

frankbu:
@robscott description updated and cla fixed.

conformance/mesh/manifests.yaml
```diff
@@ -4,7 +4,7 @@ kind: Namespace
 metadata:
   name: gateway-conformance-mesh
   labels:
-    gateway-conformance: infra
+    gateway-conformance: mesh
```
robscott, line 7: Do we need this change?

site-src/concepts/conformance.md
```diff
@@ -126,6 +126,10 @@ feature and any relevant API features, e.g:
 go test ./conformance/... -args -supported-features=Mesh,Gateway,HTTPRoute
 ```
 
+If the mesh you are testing requires labels on namespaces that host mesh workloads,
```
robscott, line 129: Mind making this less mesh-specific? (I agree that this is probably most relevant for mesh implementations, but this could be useful more broadly, for example to simplify manual cleanup of resources or if a different kind of implementation also benefited from namespace labels).

conformance/conformance_test.go
```diff
@@ -86,3 +89,15 @@ func parseSupportedFeatures(f string) sets.Set[suite.SupportedFeature] {
 	}
 	return res
 }
+
+func parseNamespaceLabels(f string) map[string]string {
```
robscott, line 93: Mind adding a simple unit test for this?

robscott (commented):
Thanks @frankbu!

site-src/concepts/conformance.md
frankbu: done.

conformance/conformance_test.go
robscott: Agree these are trivial unit tests, but we've had small things sit broken in conformance for longer than I'd like due to lack of test coverage. I'd rather lean towards spending the extra time here to avoid future pain.

frankbu: done

frankbu:
@robscott I added the unit tests you requested. Are there any other issues holding it up?

robscott:
Hey @frankbu we're effectively frozen right now until v0.7.0 releases, but we're expecting that to happen on Monday.

conformance/utils/suite/suite_test.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2: Tiny nit:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

robscott (commented):
Thanks @frankbu! This LGTM other than one tiny nit + needing a rebase.

/hold cancel

frankbu:
@robscott should be good to go now, thanks.

robscott:
Thanks @frankbu!

/lgtm
/approve

PR #1974, opened 2023-04-25
GEP(1958): TLS ALPN Routing

### LINKED ISSUE #1958: GEP: TLS ALPN Routing
A way to configure routing decisions based on the application protocols present in the TLS client hello of a TLS connection.

The `TLSRoute` is a good candidate for supporting this as it is a similar feature to SNI routing which is already handled by `TLSRoute`. Both SNI and ALPN are present in the TLS client hello and allow for TLS passthrough.

The initial discussion on this subject is [here](https://github.com/kubernetes-sigs/gateway-api/discussions/1683).

**Why this is needed**:

This would allow service owners to route to different backends behind a single domain based on the protocol the client supports.

For example if you have http/2 and mqtt clients connecting to `example.com:443` you could use the application protocols in the TLS client hello to route these clients to different backends that are able to handle the respective protocol.

dboslee:
**What type of PR is this?**
/kind gep

This PR kicks off the discussion of goals and non-goals for the GEP around TLS ALPN routing. I have left out any implementation details until there is agreement on the goals.

Fixes #1958 
[Initial discussion](https://github.com/kubernetes-sigs/gateway-api/discussions/1683)

### DIFF
--- a/geps/gep-1958.md
+++ b/geps/gep-1958.md
@@ -2,7 +2,7 @@
 Issue: #1958
 Status: Provisional
 ## TLDR
-Add support for routing traffic based on the application protcols present in the TLS client hello of a TLS connection.
+Add support for routing traffic based on the application protocols present in the "client hello" step of a TLS handshake.
 
 ## Goals
 
--- a/geps/gep-1958.md
+++ b/geps/gep-1958.md
@@ -1,0 +1,44 @@
+# GEP-1958: TLS ALPN Routing
+Issue: #1958
+Status: Provisional
+## TLDR
+Add support for routing traffic based on the application protcols present in the TLS client hello of a TLS connection.
+
+## Goals
+
+- Provide a way to route a TLS connection to a specific backend based on application protocols.
+- Define expected behavior for routing based on application protocols with respect to ordering and ambiguous matches.
+
+## Non-Goals
+- Match or detect protocols outside of the context of the application protocols present in the TLS client hello.
+- Provide routing based on any other TLS extensions.
+
+## Introduction
+ALPN (application-layer protocol negotiation) is a TLS extension which is present in the TLS client hello.
+It allows a client to specify a list of supported protocols when initiating a TLS connection.
+
+Similar to SNI (server name indication), another TLS extension present in the TLS client hello, ALPN can be used
+to route connections to different backends based on what application protocols the client supports.
+
+With this feature service owners may specify a list of application protocols to match and route TLS connections to a specific backend.
+
+This is a feature supported by the following load balancers:
+
+- [Envoy](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/tls_inspector#config-listener-filters-tls-inspector)
+- [Nginx](http://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html)
+- [Traefik](https://doc.traefik.io/traefik/routing/routers/#rule_1)
+
+## API
+
+(... details, can point to PR with changes)
+
+
+## Alternatives
+
+(List other design alternatives and why we did not go in that
+direction)
+
+## References
+
+(Add any additional document links. Again, we should try to avoid
+too much content not in version control to avoid broken links)


geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
+# GEP-1958: TLS ALPN Routing
+Issue: #1958
+Status: Provisional
+## TLDR
+Add support for routing traffic based on the application protcols present in the TLS client hello of a TLS connection.
+
+## Goals
```
howardjohn, line 7: One thing we may need to take into account is not just routing based on the ALPN, but what ALPNs we should offer in the first place.

It doesn't help to have a ALPN route for `foobar` if the Gateway doesn't advertise it

dboslee: Good point, I think that depends on whether TLS termination or TLS passthrough is being used.

I was only considering the passthrough case where the gateway is not responsible for the response. In this case the gateway is only making routing decisions, its up to the service owner to ensure the given protocol is actually selected by the upstream service.

For the TLS termination case I suppose the gateway would need to be configured to route and select ALPN based on the values configured by the service owner.

howardjohn: Ah right, forgot that. I think TLS route is only for passthrough today (?) so that would solve that concern

geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
+# GEP-1958: TLS ALPN Routing
+Issue: #1958
+Status: Provisional
+## TLDR
+Add support for routing traffic based on the application protcols present in the TLS client hello of a TLS connection.
```
shaneutt, line 5:
```suggestion
Add support for routing traffic based on the application protocols present in the "client hello" step of a TLS handshake.
```

geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Goals
+
+- Provide a way to route a TLS connection to a specific backend based on application protocols.
+- Define expected behavior for routing based on application protocols with respect to ordering and ambiguous matches.
```
shaneutt, line 10: :+1:

geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Non-Goals
+- Match or detect protocols outside of the context of the application protocols present in the TLS client hello.
+- Provide routing based on any other TLS extensions.
```
shaneutt, line 14: :+1:

shaneutt (approved):
I very much appreciate you focusing on the "what" and "why" for this initial iteration. As `Provisional` and because I see the value in the goals I approve and support moving this forward into the next steps.

Give it a bit of time to let other community members take a look:

/hold

And make sure we get at least one other maintainer to take a look:

/cc @robscott @youngnick

geps/gep-1958.md
costinm: I suspect we need ALPN both for TLS passthrough and for termination, and likely the APIs will be quite different. 

For passthrough - would be good to have some example use cases, I have not seen any server doing this and I'm not sure what will the user do - for example if alpn is 'h2', would the stream be routed to a different backend ? Doesn't seem a common enough use case.

 For 'terminate' - beyond an ordered list of protocols that the server accepts, not sure what else we can do. Server doesn't advertise protocols - it needs to selects one from the list sent by the client.

Besides h2 - I have not seen a lot of broadly used ALPNs on the internet.

arkodg (approved):
looks good, thanks !

geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Introduction
+ALPN (application-layer protocol negotiation) is a TLS extension which is present in the TLS client hello.
+It allows a client to specify a list of supported protocols when initiating a TLS connection.
+
+Similar to SNI (server name indication), another TLS extension present in the TLS client hello, ALPN can be used
+to route connections to different backends based on what application protocols the client supports.
+
+With this feature service owners may specify a list of application protocols to match and route TLS connections to a specific backend.
```
howardjohn, line 23: When we are routing it's based on the *client provided list of accepted ALPN*,right?

This may be odd for passthrough. Cliejt may present h2, http1. Route matches "h2" and routes. Backend selects "http1" .

Now our route that matched "h2" ends up serving http1/1

dboslee: This seems fine IMO as long as long as long as its clear how routing decisions are being made.

geps/gep-1958.md
dboslee: I think the two use cases and APIs are different enough that it makes sense to limit the scope this GEP to only include the tls passthrough use case.

shaneutt:
@dboslee checking in on this one: where are we at? do you need any assistance to move this one forward?

dboslee:
> @dboslee checking in on this one: where are we at? do you need any assistance to move this one forward?

@shaneutt The only clarification needed is on https://github.com/kubernetes-sigs/gateway-api/pull/1974#discussion_r1210574157. I would like to add the tls termination use case as a non-goal.

costinm:
I think I'm more ok with it for tls routing - termination cases are far more complicated.

But I'm still not convinced this belong here - and not in a vendor extension. It's such an arcane and rare use case.
I assume it'll be optional - how many implementations do you expect to bother to implement it ( or have the ability 
to do so) ? Even plain SNI routing is not so common - and in some cases it gets miss-used. 

We really need a repository for extensions and uncommon features - keeping the core APIs simple would 
be better for users. 

But I'm not against it - I can see some use cases and in the case of pass-through there are very few things we can 
use in ClientHello. I can think of few ways to take advantage of this feature, after all it's a list of opaque strings...

youngnick:

geps/gep-1958.md
tokers, line 1: That's an interesting idea, but could you share some real cases that we do need route connections according to the ALPN?

geps/gep-1958.md
shaneutt: Given the lack of response, we'll consider this comment resolved for the moment. Please mark it unresolved if more conversation is needed.

geps/gep-1958.md
shaneutt: @dboslee just wanted to ping on this, as I think capturing a couple "user stories" or user-story-like things in the GEP could be helpful for future readers.

geps/gep-1958.md
```diff
@@ -0,0 +1,44 @@
+# GEP-1958: TLS ALPN Routing
+Issue: #1958
+Status: Provisional
+## TLDR
+Add support for routing traffic based on the application protocols present in the "client hello" step of a TLS handshake.
+
+## Goals
+
+- Provide a way to route a TLS connection to a specific backend based on application protocols.
+- Define expected behavior for routing based on application protocols with respect to ordering and ambiguous matches.
```
howardjohn, line 10: This PR does not do this (or not completely, rather). Is it intended to achieve this goal in this PR or in the future?

Specifically, the handling of the fact a request has multiple ALPNs which are negotiated by the backends, and we may have listeners.

For instance, what is the behavior of:

Client ALPN: `[a, b, c]`
Listeners: `[no ALPN match, match ALPN=b, match ALPN=c]`
Server selects ALPN `c`

And many similar cases

Issue #1945, opened 2023-04-17
Conformance RoundTripper should allow for custom dialer


briantkennedy:
**What would you like to be added**:

A way to inject a custom dialer into the conformance round tripper.

**Why this is needed**:

The existing round tripper requires the user to use the built in Go `http.Client` which directly dials endpoints from the test process.  Adding a custom dialer will allow communicating with the load balancer via a tunneled TCP connection which is useful if the load balancer is behind a firewall or not routable from the test process.

shaneutt:
This seems reasonable to add:

/triage accepted

Though we would consider this a low priority in general:

/priority backlog

If it's blocking you however we would definitely welcome PRs to get people unstuck:

/help

briantkennedy:
@robscott can we reopen this?  We need to fix up the change just slightly since we want to add a `DialContext` function as a member of the conformance RoundTripper instead of putting a `http.RoundTripper` on the request.

PR #1959, opened 2023-04-22
Support custom dialer in conformance test RoundTripper

panslava:
**What type of PR is this?**
/kind test
/area conformance

**What this PR does / why we need it**:
Adds a way to specify custom dialer in RoundTripper

**Which issue(s) this PR fixes**:
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1945

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #1975, opened 2023-04-25
Extend DefaultRoundTripper with CustomDialContext

panslava:
**What type of PR is this?**
/kind test
/area conformance

This slightly changes previous approach, just merged in https://github.com/kubernetes-sigs/gateway-api/pull/1959
Instead of extending Request struct, extend DefaultRoundTripper. Extending Request was problematic, because Request is constructed in other function https://github.com/kubernetes-sigs/gateway-api/blob/ef75c8a50f70736bb3489e13a1357d89de8c282f/conformance/utils/http/http.go#L99 and we don't want to extend it. Plus, we still want tlsConfig to be added even for custom dialers

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1945

```release-note
NONE
```

/cc @briantkennedy

### DIFF
--- a/conformance/utils/roundtripper/roundtripper.go
+++ b/conformance/utils/roundtripper/roundtripper.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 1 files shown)


robscott:
Thanks @panslava!

/ok-to-test

robscott:
Thanks @panslava! This seems reasonable to me and results in an even smaller change than the original PR. 

/lgtm
/approve

PR #1977, opened 2023-04-26
Add validation for `gateway.spec.addresses` to validate whether fields of addresses are set according to the Gateway API specification

### LINKED ISSUE #1960: Add validation for `Gateway.Spec.Addresses`

In current validation server, it does not have validations for `Gateway.Spec.Addresses` . 

**Why this is needed**:

It's odd that this field does not have any validations, though we can do this in downstream implementations, still feels more logical to do it here.

refer to: https://github.com/envoyproxy/gateway/pull/1322#discussion_r1174349868

shawnh2:
**What type of PR is this?**

/kind feature

Add validation for `Gateway.Spec.Addresses` field.

Fixes #1960 

no

### DIFF
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -45,7 +45,8 @@ var (
 		gatewayv1b1.TLSProtocolType:   {},
 	}
 
-	validHostnameAddress = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+	validHostnameAddress   = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
+	validHostnameRegexp, _ = regexp.Compile(validHostnameAddress)
 )
 
 // ValidateGateway validates gw according to the Gateway API specification.
@@ -163,22 +164,30 @@ func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.
 	return errs
 }
 
-// validateGatewayAddresses validates whether required fields of addresses are set according
+// validateGatewayAddresses validates whether fields of addresses are set according
 // to the Gateway API specification.
 func validateGatewayAddresses(addresses []gatewayv1b1.GatewayAddress, path *field.Path) field.ErrorList {
 	var errs field.ErrorList
+	ipAddrSet, hostnameAddrSet := sets.Set[string]{}, sets.Set[string]{}
 	for i, address := range addresses {
 		if address.Type != nil {
 			if *address.Type == gatewayv1b1.IPAddressType {
 				if _, err := netip.ParseAddr(address.Value); err != nil {
 					errs = append(errs, field.Invalid(path.Index(i), address.Value, fmt.Sprintln("invalid ip address")))
 				}
+				if ipAddrSet.Has(address.Value) {
+					errs = append(errs, field.Duplicate(path.Index(i), address.Value))
+				} else {
+					ipAddrSet.Insert(address.Value)
+				}
 			} else if *address.Type == gatewayv1b1.HostnameAddressType {
-				r, err := regexp.Compile(validHostnameAddress)
-				if err != nil {
-					errs = append(errs, field.InternalError(path.Index(i), fmt.Errorf("could not compile hostname matching regex: %w", err)))
-				} else if !r.MatchString(address.Value) {
+				if !validHostnameRegexp.MatchString(address.Value) {
 					errs = append(errs, field.Invalid(path.Index(i), address.Value, fmt.Sprintf("must only contain valid characters (matching %s)", validHostnameAddress)))
+				}
+				if hostnameAddrSet.Has(address.Value) {
+					errs = append(errs, field.Duplicate(path.Index(i), address.Value))
+				} else {
+					hostnameAddrSet.Insert(address.Value)
 				}
 			}
 		}
--- a/apis/v1beta1/validation/gateway.go
+++ b/apis/v1beta1/validation/gateway.go
@@ -18,6 +18,8 @@ package validation
 
 import (
 	"fmt"
+	"net/netip"
+	"regexp"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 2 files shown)


apis/v1beta1/validation/gateway.go
```diff
@@ -157,3 +162,26 @@ func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.
...
+func validateGatewayAddresses(addresses []gatewayv1b1.GatewayAddress, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	for i, address := range addresses {
+		if address.Type != nil {
+			if *address.Type == gatewayv1b1.IPAddressType {
+				if _, err := netip.ParseAddr(address.Value); err != nil {
+					errs = append(errs, field.Invalid(path.Index(i), address.Value, fmt.Sprintln("invalid ip address")))
+				}
+			} else if *address.Type == gatewayv1b1.HostnameAddressType {
+				r, err := regexp.Compile(validHostnameAddress)
```
howardjohn, line 177: should compile it once and reuse

apis/v1beta1/validation/gateway.go
```diff
@@ -157,3 +162,26 @@ func validateHostnameProtocolPort(listeners []gatewayv1b1.Listener, path *field.
 	}
 	return errs
 }
+
+// validateGatewayAddresses validates whether required fields of addresses are set according
+// to the Gateway API specification.
+func validateGatewayAddresses(addresses []gatewayv1b1.GatewayAddress, path *field.Path) field.ErrorList {
```
arkodg, line 168: can you also add a duplicate check in here

gyohuangxin:
@shawnh2 Thanks for your first PR, can you update the PR title and description to describe what the validation does, like "Add validation for `gateway.spec.addresses` to validate whether addresses are set according to the Gateway API specification."

/ok-to-test

apis/v1beta1/validation/gateway.go
shawnh2: fixed

apis/v1beta1/validation/gateway.go
shawnh2: great idea! added

apis/v1beta1/validation/gateway.go
```diff
@@ -42,6 +44,9 @@ var (
 		gatewayv1b1.HTTPSProtocolType: {},
 		gatewayv1b1.TLSProtocolType:   {},
 	}
+
+	validHostnameAddress   = `^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`
```
gyohuangxin, line 48: Is this regular expression pattern specifically copied from a reference somewhere?

gyohuangxin: I think we should replace it with the pattern we already have. I think we should use this pattern: https://github.com/kubernetes-sigs/gateway-api/blob/16c54cdd2546bce917c1d045a5dc6a49e51625ae/apis/v1beta1/shared_types.go#L381-L382

shawnh2: sure, replaced

apis/v1beta1/validation/gateway.go
```diff
@@ -42,6 +44,9 @@ var (
 		gatewayv1b1.HTTPSProtocolType: {},
 		gatewayv1b1.TLSProtocolType:   {},
 	}
+
+	validHostnameAddress   = `^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	validHostnameRegexp, _ = regexp.Compile(validHostnameAddress)
```
shawnh2, line 49: could `regexp.MustCompile()` be better?

shaneutt: Yes, if you have a constant/literal string holding your regex that will ALWAYS compile the same way, use `MustCompile`.

shawnh2: great, thanks

gyohuangxin (approved):
LGTM, thanks for your contribution.

apis/v1beta1/validation/gateway_test.go
```diff
@@ -247,6 +247,63 @@ func TestValidateGateway(t *testing.T) {
...
+			mutate: func(gw *gatewayv1b1.Gateway) {
+				gw.Spec.Addresses = []gatewayv1b1.GatewayAddress{
+					{
+						Type:  ptrTo(gatewayv1b1.IPAddressType),
+						Value: "1.2.3.4",
+					},
+					{
+						Type:  ptrTo(gatewayv1b1.HostnameAddressType),
+						Value: "foo.bar",
+					},
```
shaneutt, line 260: Just for the sake of completion, we should add a valid ipv6 address to this test.

apis/v1beta1/validation/gateway_test.go
```diff
@@ -247,6 +247,63 @@ func TestValidateGateway(t *testing.T) {
...
+						Type:  ptrTo(gatewayv1b1.HostnameAddressType),
+						Value: "*foo/bar",
+					},
+					{
+						Type:  ptrTo(gatewayv1b1.HostnameAddressType),
+						Value: "12:34:56::",
+					},
+				}
+			},
+			expectErrsOnFields: []string{"spec.addresses[0]", "spec.addresses[1]", "spec.addresses[2]"},
```
shaneutt, line 282: While I suspect there may be some precedent somewhere else, I'm not a huge fan of this method of checking errors because it just checks _that_ there are errors and doesn't validate the consistency of the actual error produced. We should verify that the correct error type is produced (e.g. `Invalid`, `Duplicate`, e.t.c.) across all these tests, and also test the error text context. This will help to avoid future developers accidentally introducing subtle changes to error output on accident.

shaneutt (changes requested):
Mostly looks good, just a couple thoughts on the tests.

apis/v1beta1/validation/gateway_test.go
shawnh2: not a fan of this method either, and I can improve that

shaneutt: Thank you, very much appreciated :bow:

apis/v1beta1/validation/gateway_test.go
shawnh2: added

apis/v1beta1/validation/gateway_test.go
shawnh2: hi, the changes has been made, PTAL

shaneutt (approved):
I very much appreciate the extra time and care you put in on the tests to check the errors, looks good to me thank you.

shaneutt (approved):
/lgtm

Issue #1911, opened 2023-04-05
GEP: Backend Protocol Selection


dprotaso:
/kind gep

This is an attempt to separate out the backend protocol selection from [GEP-1282](https://github.com/kubernetes-sigs/gateway-api/issues/1282). This is similar to how TLS use-cases are being split out into [GEP-1897](https://github.com/kubernetes-sigs/gateway-api/issues/1897)

**What would you like to be added**:

The ability to express the protocols supported by backends.

**Why this is needed**:

Not all implementation support automatic protocol selection. Even in some cases protocols are disabled without an explicit opt-in (ie. websockets). Thus application developers need the ability specify the protocol of their applications.

dprotaso:
For those following GEP PR is out - https://github.com/kubernetes-sigs/gateway-api/pull/1979

shaneutt:
Since the GEP already exists and _because_ it's a GEP it seems reasonable to consider this accepted:

/triage accepted

However, it doesn't seem clear to me that this is something we _must_ sort out prior to GA (though I invite arguments to the contrary) so at least for now we wont consider this a blocker for GA and it will have a lower priority than GA-oriented work:

/priority backlog

dprotaso:
@shaneutt the GA backlog has #1282 -  you could say this GEP and #1897 are part the fallout of that original GEP

shaneutt:
ACK

Ultimately it's just not clear that this is something we _must_ have for GA :thinking: 

Since you're working on it anyway, let's go ahead and track it on the board. But we should probably take another look at what it's gonna take to implement and do some planning and prioritization then.

robscott:
Leaving this open until this GEP either graduates to standard or is removed from the API.

/kind gep
/reopen

Issue #205, opened 2020-06-04
WebSocket protocol support


jpeach:
This issue is closely related to #204.

There should be a way for applications to indicate that they want WebSocket streams for [HTTP/1.1](https://tools.ietf.org/html/rfc8441) or [HTTP/2](https://tools.ietf.org/html/rfc6455).

One way to do this could be a `WebSocketRoute` as a peer to `HTTPRoute`. Another way cold be to add a field somewhere under the `HTTPRoute` API.

Some ingress controllers implement this by a flag on the HTTP virtual host or  forwarding route (e.g. Contour puts it on the [Route](https://projectcontour.io/docs/v1.5.0/api/#projectcontour.io/v1.Route)).

jpeach:
/remove-lifecycle stale

Not stale, but also not for v1alpha1.

masus04:
Any progress on this? Seems like an essential feature.

shaneutt:
There's been no progress that I'm personally aware of. If you have a strong interest and use case it might be worth joining the monday night community meeting and bringing this one up for discussion :thinking:

shaneutt:
So as it turns out, there's some discussion about this going on in #1244. I'm going to mark this as `help wanted` as we need an owner to drive this forward (hopefully as a direct result of that discussion).

shaneutt:
Given that this has been marked help wanted for a long time without anyone with a strong interest to champion it and move it forward, it doesn't seem like we've got any momentum behind this currently. As such we'll consider this closed as "not planned" for the moment. This doesn't mean we'll never do it it just means it doesn't currently have any priority: if you're interested in taking this issue on and moving it forward please comment here and we would be happy to re-open it.

dprotaso:
@shaneutt this falls under the backend properties

https://gateway-api.sigs.k8s.io/geps/gep-1282/

I would reopen this

arkodg:
some implementations like Envoy Gateway and Istio have web socket upgrade enabled by default for HTTP 1.1

shaneutt:
@dprotaso are you specifically interested in driving this forward, may we assign you?

dprotaso:
Yup please do - I'm going to try tackling specifying backend protocol which would cover this

shaneutt:
Awesome, thank you! Let us know how we can support you in your efforts :vulcan_salute: 

/assign @dprotaso

shaneutt:
[...]

PR #1979, opened 2023-04-27
GEP-1911 - Backend Protocol Selection

dprotaso:
**What type of PR is this?**

/kind gep

This PR flushes out how app developers can specify that their application supports different protocols

Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/1911 (Backend Protocol GEP)
Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/205 (Websocket Support)

```release-note
Allow end-users to specify backend protocols
```


geps/gep-1911.md
```diff
@@ -0,0 +1,187 @@
...
+metadata:
+  name: some-service
+  namespace: some-namespace
+spec:
+  ports:
+  - port: 80
+    http2: 
+      encrypted: false # enable h2c
+  - port: 80
+    http1: {}          # We also listen to http1 on port 80
```
dprotaso, line 108: Probably worth calling out you could pack multiple protocol entries in a single `BackendPort` struct - but you would lose priority between protocols (unless we added another mechanic)

```yaml
- port: 80
  http2: {}
  http1: {}
```

If we choose to limit `BackendPort` struct to a single protocol I would need to spell that out in the GEP

arkodg:
+1 for adding an API / API field to be explicit about backend/upstream protocol

since this PR suggests a `HOW` , will share my preference, which would be to keep it close to the
[backendRef](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendRef)

```
protocols: []Protocol
```
e.g.
```
- backendRefs:
    - name: foo-v1
      port: 8080
      protocols:
      - HTTP
      - PROXYPROTOCOL
      weight: 90
```

relates to https://github.com/envoyproxy/gateway/issues/1328

dprotaso:
> since this PR suggests a HOW , will share my preference, which would be to keep it close to the
[backendRef](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendRef)

The only thing with the list of protocol strings is that it doesn't enable the goal I set out - `Extensible - additional protocol specific fields should be easy to add in the future`

This is why rather than having an `h2c` constant I added a `Cleartext` field under `HTTP2PortProtocol`.

I'm not super opinionated about this and welcome discussion.

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+### Extending Kubernetes Service
+
+This is considered untenable due to the 'the turnaround time for those changes can be years.' ([ref-1282](geps/gep-1282.md#non-goals):
+
+### AppProtocol
+
+From [GEP-1282](geps/gep-1282.md#non-goals):
+> v1.Service’s appProtocol field is not fit for purpose, because it is defined as accepting values either from the IANA Service Name registry, or domain-prefixed values and we need more flexibility than that. 
+
+See further discussion [here](https://github.com/kubernetes-sigs/gateway-api/discussions/1244#discussioncomment-3151197)
```
howardjohn, line 158: I am not sure this is true since https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols#summary?

There was talk of allowing multiple appProtocols as well - I don't think it landed but I don't think its in-plausible

robscott: Yep I think multiple app protocols is inevitable - help wanted though. @LiorLieberman has been helping move AppProtocols forward upstream.

arkodg:
> > since this PR suggests a HOW , will share my preference, which would be to keep it close to the
> > [backendRef](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendRef)
> 
> The only thing with the list of protocol strings is that it doesn't enable the goal I set out - `Extensible - additional protocol specific fields should be easy to add in the future`
> 
> This is why rather than having an `h2c` constant I added a `Cleartext` field under `HTTP2PortProtocol`.
> 
> I'm not super opinionated about this and welcome discussion.

good point, a `struct` is better if we are envisioning customizing the backend protocol functionality.
I was taking the concept of `upstream` protocol https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.ProtocolType, converting it into a `list` and reusing it here

geps/gep-1911.md
arkodg: this would be needed for all backend types including ServiceImport
https://github.com/kubernetes-sigs/mcs-api/blob/11eea5e71266160cc900f319b3f080c24dbea4f1/pkg/apis/v1alpha1/serviceimport.go#L99

dprotaso: > There was talk of allowing multiple appProtocols as well - I don't think it landed but I don't think its in-plausible

Is that this https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1435-mixed-protocol-lb ? It's GA in K8s 1.26. Do we know why the KEP is scoped to type=LoadBalancer?

> I am not sure this is true since https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols#summary?

The examples are confusing - it suggests `HTTP2` has two different meanings (one is http2 over tls in GKE and the other is h2c with Istio). Not a fan of that.

I think we need to decided do we want the ability to customize the backend in any way - https://github.com/kubernetes-sigs/gateway-api/pull/1979#issuecomment-1526585087

robscott: >The examples are confusing - it suggests HTTP2 has two different meanings (one is http2 over tls in GKE and the other is h2c with Istio). Not a fan of that.

To clarify, that was trying to note that the same values were being interpreted different ways by different implementations. It was not an example of something to do, more something this KEP was trying to solve.

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+
+* Issue: [#1911](https://github.com/kubernetes-sigs/gateway-api/issues/1911)
+* Status: Provisional
+
+(See status definitions [here](overview.md#status).)
+
+## TLDR
+
+Not all implementation support automatic protocol selection. Even in some cases protocols are disabled without an explicit opt-in (ie. websockets). Thus application developers need the ability specify the protocol(s) that their application supports.
+
```
candita, line 11:
```suggestion
Not all implementations support automatic protocol selection. Even in some cases protocols are disabled without an explicit opt-in (e.g. websockets). Thus application developers need the ability to specify the protocol(s) that their application supports.

```

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+    http1: {}
+  targetRef:
+    group: "" # corev1
+    kind: Service
+    name: some-service
+```
+
+### Default Protocols
+
+If a backend protocol isn't specified an implementation MAY infer the backend protocol through it's own means.
```
candita, line 126:
```suggestion
If a backend protocol isn't specified an implementation MAY infer the backend protocol through its own means.
```

geps/gep-1911.md
dprotaso: I shouldn't skim KEPs during dinner ;)

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+then it must set it's condition to `Accepted` `False`.
+
+## Alternatives
+
+### Meta-resource for each protocol
+
+Rather than bundle protocol details into a single resource an alternative would be to create distinct meta resources.
+ie. `HTTP2Backend`, `GPRCBackend`, `WebsocketBackend`.
+
+Although this has the advantage of seperate it doesn't allow
```
candita, line 140: Is there more to this? :)

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+See further discussion [here](https://github.com/kubernetes-sigs/gateway-api/discussions/1244#discussioncomment-3151197)
+
+An option would be to have a specific vendor prefix but unfortunately this limits the protocol to one per port.
+
+### Unstructured Data/Special Values
+
+Unstructured data refers to using labels and annotations.
+
+From [GEP-1282](geps/gep-1282.md#non-goals):
+> these are very sticky and hard to get rid of once you start using them.
```
candita, line 167: Annotations are also not self-documenting, there is no `kubectl explain` for annotations.

geps/gep-1911.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Non-Goals
+
+- Backend TLS (covered in [GEP-1897](https://github.com/kubernetes-sigs/gateway-api/issues/1897)
+
+## Introduction
+
+## API
+
+We propose adding a new meta-resource [GEP-713](geps/gep-713.md) called `BackendProtocol`.
```
candita, line 27: Is #713 the right GEP?

dprotaso: This is open for discussion. Technically since I'm using a `targetRef` it could be _anything_. But if we want to scope this to just `Service` then we could simplify this.

As is,  _we_ could target a multi cluster ServiceImport - but I'm not to familiar with that. cc @robscott

geps/gep-1911.md
dprotaso: whoops

dprotaso: flushed this out a bit

geps/gep-1911.md
dprotaso: added this to the section

geps/gep-1911.md
dprotaso: One observation I have about appProtocol (or `lists of protocols` in general) is how can a backend _disable_ a protocol.

ie. disable websockets & http

geps/gep-1911.md
```diff
@@ -0,0 +1,199 @@
...
+type WebsocketPortProtocol struct {
+  // Future knobs can go here
+}
+```
+
+### YAML API
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1alpha1
+kind: BackendProtocol
```
hzxuzhonghu, line 101: The service is k8s core/v1 has already include ports and protocol,, why not use  it

geps/gep-1911.md
dprotaso: Added some comments in the GEP from this discussion

Following up on[KEP-1435 - Mixed Protocol](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1435-mixed-protocol-lb) - it only applies to the `protocol` field. Thus trying to specify a K8s Service with multiple `appProtocol` to a single port won't work

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    app.kubernetes.io/name: proxy
spec:
  containers:
  - name: nginx
    image: nginx:stable
    ports:
      - containerPort: 80
        name: http-web-svc
---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app.kubernetes.io/name: proxy
  ports:
  - name: http
    protocol: TCP
    appProtocol: http
    port: 80
    targetPort: http-web-svc
  - name: h2c
    appProtocol: h2c
    protocol: TCP
    port: 80
    targetPort: http-web-svc
```

geps/gep-1911.md
```diff
@@ -0,0 +1,238 @@
...
+## Goals
+
+- Support protocols that can have a Gateway '*Route' resource as a frontend
+- Consistent way of declaring backend protocols using a meta-resource
+- Support backends with multiple protocols on the same port (ie. h2c/http1)
+- Extensible - additional protocol specific fields should be easy to add in the future
+
+## Non-Goals
+
+- Backend TLS (covered in [GEP-1897](https://github.com/kubernetes-sigs/gateway-api/issues/1897)
```
shaneutt, line 21:
```suggestion
- Backend TLS (covered in [GEP-1897](https://github.com/kubernetes-sigs/gateway-api/issues/1897))
```

robscott:
@dprotaso We discussed this at the community meeting yesterday. Here are some rough notes:

1) I think we should really consider the list of AppProtocols proposed for upstream Kubernetes as a solution to this specific problem
2) If/when Service includes a list of AppProtocols, that would likely be propagated to ServiceImport since that API is just mirroring Service.

@LiorLieberman has been working on the upstream AppProtocol changes and has also been commenting here. Would be helpful to understand what we'd still be missing if AppProtocol became a list on the Service API.

dprotaso:
Sounds good - I'm glad there's consensus around using the `appProtocol` as the mechanic for backend protocol selection.

I'll update the GEP

geps/gep-1911.md
```diff
@@ -0,0 +1,238 @@
+# GEP-1911: Backend Protocol Selection
+
+* Issue: [#1911](https://github.com/kubernetes-sigs/gateway-api/issues/1911)
+* Status: Provisional
+
+(See status definitions [here](overview.md#status).)
+
+## TLDR
+
+Not all implementations support automatic protocol selection. Even in some cases protocols are disabled without an explicit opt-in (eg. websockets). Thus application developers need the ability to specify the protocol(s) that their application supports.
```
LiorLieberman, line 10: > Even in some cases protocols are disabled without an explicit opt-in (eg. websockets)

What do you mean here ? can you give an example?

LiorLieberman (commented):
Left some comments, I think you probably still have not updated it yet since https://github.com/kubernetes/enhancements/pull/3996 opened. But still good to better understand the usecases. (Apologies if you already elaborated on the usecases on a GW meeting and I was not there - if so do you have a link or notes to what was brought up during the meeting? )

Finally, would https://github.com/kubernetes/enhancements/pull/3996 be enough for your usecases now?

LiorLieberman (commented):
Left some comments, I think you probably still have not updated it yet since https://github.com/kubernetes/enhancements/pull/3996 opened. But still good to better understand the usecases. (Apologies if you already elaborated on the usecases on a GW meeting and I was not there - if so do you have a link or notes to what was brought up during the meeting? )

Finally, would https://github.com/kubernetes/enhancements/pull/3996 be enough for your usecases now?

geps/gep-1911.md
dprotaso: Contour by default when using their HTTPProxy resource doesn't support websockets out of the box. You need to set a bool attribute to true - details here: https://projectcontour.io/docs/1.25/config/websockets/

geps/gep-1911.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Goals
+
+- Support protocols that can have a Gateway '*Route' resource as a frontend
+- Consistent way of declaring backend protocols using a meta-resource
+- Support backends with multiple protocols on the same port (ie. h2c/http1)
```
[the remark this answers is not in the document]
dprotaso: > Can you elaborate on your specific case ? do you need a backend that supports multiple protocols on the same port?

Yeah this is fairly common. For example I'm aware of users running TensorFlow Serving on CloudRun. It exposes HTTP and gRPC protocols for model prediction and it's being run on a single port that's multiplexing between them.

geps/gep-1911.md
```diff
@@ -0,0 +1,238 @@
...
+Originally the use of this field in the Gateway API was rejected in [GEP-1282](geps/gep-1282.md#non-goals):
+> v1.Service’s appProtocol field is not fit for purpose, because it is defined as accepting values either from the IANA Service Name registry, or domain-prefixed values and we need more flexibility than that. 
+
+Further discussion [here](https://github.com/kubernetes-sigs/gateway-api/discussions/1244#discussioncomment-3151197)
+
+Since then a Kubernetes enhancement proposal was created [KEP-3726](https://github.com/kubernetes/enhancements/issues/3726)
+to repurpose `appProtocol` to include a convention for protocols that are not IANA service names. This would involve
+prefixing protocol names with `kubernetes.io/`. At the moment there exists two defined protocols `kubernetes.io/h2c` and `kubernetes.io/grpc`.
+
+It is not possible to have different `appProtocol` entries for the same port. A related [KEP-1435](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1435-mixed-protocol-lb) mentions different `protocol` entries for the same port. Finally it's unclear how one would disable certain protocol behaviours (ie. HTTP/1 Upgrades).
```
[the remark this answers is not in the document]
dprotaso: I know envoy has options to disable this in various places - https://www.envoyproxy.io/docs/envoy/v1.26.1/intro/arch_overview/http/upgrades.html

I don't have a use case really - just added this clause since `appProtocol` mechanic allows enabling but not disabling.

geps/gep-1911.md
dprotaso: Re-wrote the GEP so it's not a meta-resource anymore

geps/gep-1911.md
dprotaso: Switched the GEP to consider `appProtocol` resolving - please review

geps/gep-1911.md
dprotaso: Mentioned nginx and contour in this tl;dr where the websocket protocols were off by default

geps/gep-1911.md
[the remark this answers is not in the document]
dprotaso: I made this a non-goal for the GEP

geps/gep-1911.md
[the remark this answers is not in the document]
dprotaso: Also K8s supports multiple Service `protocol` on the same port.

I added a section about this - please re-review.

dprotaso:
Hey everyone, the GEP has been re-written with the intent of supporting `appProtocol` as the mechanic for users to specify the backend protocol.

Please take a look and thanks for reviewing the GEP again/

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+
+A Gateway implementation MUST recognize the Kubernetes Standard Application Protocols ([KEP-3726](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols)) for specifying the protocol for a backend reference in a Gateway API `*Route` resource
+
+Thus when a `*Route` points to a Kubernetes Service resource the backend protocol for each port can be specified by
+- setting the `appProtocol` field on the Kubernetes `Service`
+- setting the `appProtocol` field on an `Endpoint`/`EndpointSlice` object associated with a Kubernetes `Service`
+
+At the moment there exists two defined constants:
+- `kubernetes.io/h2c` - HTTP/2 over cleartext as described in https://www.rfc-editor.org/rfc/rfc7540
+- `kubernetes.io/raw` - Instructs the implementation to look at the `Service`'s `protocol` field which supports `TCP`, `UDP`, and `SCTP`. Default is `TCP`.
```
LiorLieberman, line 45: `kubernetes.io/raw` - was not recognized  as a standard (yet). I know the current situation is confusing because https://github.com/kubernetes/enhancements/pull/3912 was not merge yet to reflect the actual implementation. 
I will ping again there to ask for approval.

dprotaso: PR merged so going to resolve this

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+
+  // Ports contains a list of protocols for a specific ports
+  // When multiple protocols reference the same port the 
+  // entries earlier in the list should have higher priority
+  //
+  // ie. attempt to use h2c otherwise fallback on http1
+  //  ports: 
+  //   - port: 80
+  //     http2: 
+  //       encrypted: false # enable h2c
```
costinm, line 106: Seems a bit strange - h2c has pretty clear semantics. 

Also differentiating between false and not set can be confusing

costinm: Also if encrypted = true  - we need to be clear ALPN takes priority and any setting here is ignored.

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+
+type BackendPort struct {
+  // Number or name of the port who's protocol is being defined
+  Port intstr.IntOrString
+
+  PortProtocol `json:",inline"`
+}
+
+type PortProtocol struct {
+  HTTP2     *HTTP2PortProtocol    `json:"http2"`
```
costinm, line 121: How about adding a separate 'TLS' protocol - which indicates the backend handles TLS and will negotiate the protocol as defined in the RFC.

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+  HTTP1     *HTTP1PortProtocol    `json:"http1"`
+  GRPC      *GRPCPortProtocol     `json:"grpc"`
```
costinm, line 123: Is it encrypted ? How is it different from h2c from gateway perspective ?

geps/gep-1911.md
dprotaso: you're reviewing an alternative proposal - this isn't what's being recommended anymore

geps/gep-1911.md
dprotaso: you're reviewing an alternative proposal - this isn't what's being recommended anymore

geps/gep-1911.md
dprotaso: you're reviewing an alternative proposal - this isn't what's being recommended anymore

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+  WebSocket *WebsocketPortProtocol`json:"websocket"`
+}
```
costinm, line 125: Not sure what websocket means. 

WS works over h2 - as CONNECT method - and on http1.1, where the protocol upgrade happens. I assume this is the http/1.1 websocket ? But even on 1.1, WS is based on path - a backend may handle both regular POST/GET methods and have WS on one path. 

Same for http2, where CONNECT is just one of the methods that a h2 server supports, so can't be just ws.

There is also the question of what will the gateway do when receiving a h1 websocket request and the backend is h2, or the other way around.

dprotaso: you're reviewing an alternative proposal - this isn't what's being recommended anymore

geps/gep-1911.md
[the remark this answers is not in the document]
LiorLieberman: > Yeah this is fairly common. For example I'm aware of users running TensorFlow Serving on CloudRun. It exposes HTTP and gRPC protocols for model prediction and it's being run on a single port that's multiplexing between them.

I read a little bit about TensorFlow Service, it looks like it is two different ports. 8500 for GRPC and 8501 for HTTP.

The reason I am bringing this up is that I am trying to collect real uses cases for converting `AppProtocol` to a list.

geps/gep-1911.md
LiorLieberman: can you remove `kubernetes.io/raw`? it is not a standard so it may confuse other people.

geps/gep-1911.md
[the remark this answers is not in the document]
LiorLieberman: Also, it is still a goal on this GEP, not sure I found how this GEP suggests to support it?
few service protocols?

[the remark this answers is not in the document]
dprotaso: > Also, it is still a goal on this GEP, not sure I found how this GEP suggests to support it?

It was buy not anymore. I'll remove it

geps/gep-1911.md
dprotaso: What's the alternative mechanic to say we support TCP etc. ?

geps/gep-1911.md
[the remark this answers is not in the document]
dprotaso: Actually, K8s supports TCP/UDP protocol on the same port - so I think this requirement stands.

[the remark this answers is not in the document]
dprotaso: > The reason I am bringing this up is that I am trying to collect real uses cases for converting AppProtocol to a list.

An alternative to a list could be allowing multiple entries like how the mixed protocol KEP suggested it

https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1435-mixed-protocol-lb#proposal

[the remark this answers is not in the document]
LiorLieberman: lets say we support a list, how would it help for your use case?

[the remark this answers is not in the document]
dprotaso: Knative OSS and Cloud Run support the same portable API. It has a restrictions to listen only on one port since we need to capture request metrics to scale and we also enforce request concurrency.

So if there was a TensorFlow user which wanted to bring  that workload to Knative in order to get elastic autoscaling and scale to zero etc. they wouldn't be able to host both GRPC and HTTP using the same instance.

The TensorFlow Serving isn't a random use case - there are competing products doing the same thing - https://kserve.github.io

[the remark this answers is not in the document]
howardjohn: We run Istio on CloudRun and do the same - gRPC and http on the same port

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+
+If a backend protocol isn't specified an implementation MAY infer the backend protocol through its own means. 
+
+### Unsupported/Incompatible Protocols
+
+If a Route attached to a Gateway is not able to sending traffic to the backend using the specified protocol then it must set it's RouteConditionType `Accepted` to `False`, with a RouteConditionReason `UnsupportedProtocol`
+
+### Supporting new protocols
+
+To add support for a new protocol it should first become a Kubernetes Standard Application Protocol by updating the [KEP-3726](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols). If the suggested protocol is not suited to have a `kubernetes.io/*` prefix then the Gateway API MAY support the new protocol using it's own prefix `gateway-api.kubernetes.io/*`.  Please make a PR to this GEP.
```
youngnick, line 65: Can you please split this on multiple lines?

Aside from that, love this, this is a great section to have in a GEP, and we should consider adding it to other GEPs in general - "How to add new bits to this GEP" is not clear in many cases.

youngnick (commented):
I'm broadly in support of this, but I think that we'll need approvals from most if not implementations for something that will be Core conformance.

I mentioned some examples below, but laying out a more complete conformance testing plan might also help, because implementations can then check against that plan, even in the absence of the actual conformance testing framework.

geps/gep-1911.md
LiorLieberman: The current situation is the if the protocol is `TCP` and `appProtocol` is not set, implementations do whatever they want, one can take is "raw" TCP other can try to sniff and guess the traffic.

We wanted to introduce an appProtocol value that will basically be "raw" not only for TCP but for all/most port protocols (TCP/UDP/SCTP) hence the name `raw` and not `tcp-raw` or `tcp`.  That said, it was last minute and we did not added it eventually. (https://github.com/kubernetes/kubernetes/pull/115433#discussion_r1136267401)

You could add it to your current open enhancement and/or start a discussion in slack/github to see what other people think.

geps/gep-1911.md
```diff
@@ -0,0 +1,260 @@
...
+
+Currently Kubernetes `Service` API does not allow different `appProtocol` values for the same port. At this time there seems to be interest in changing `appProtocol` to be a list in order to faciliate this use-case.
+
+### Default Protocols
+
+If a backend protocol isn't specified an implementation MAY infer the backend protocol through its own means. 
+
+### Unsupported/Incompatible Protocols
+
+If a Route attached to a Gateway is not able to sending traffic to the backend using the specified protocol then it must set it's RouteConditionType `Accepted` to `False`, with a RouteConditionReason `UnsupportedProtocol`
```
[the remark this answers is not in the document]
LiorLieberman: To clarify, `kubernetes.io/raw` is still not supported. (see [here](https://github.com/kubernetes-sigs/gateway-api/pull/1979#discussion_r1214081474))

I do think it is useful to have one but it would need an small enhancement.

Aside from that, I am not sure I understand the examples.  Why would HTTPRoute not get https, http or `kubernetes.io/h2c` ?

geps/gep-1911.md
dprotaso: I added a PR to add support for raw - https://github.com/kubernetes/enhancements/pull/4106

geps/gep-1911.md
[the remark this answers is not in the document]
dprotaso: I created at able showing some recommended compatability - it does raise more questions

shaneutt (approved):
As a `Provisional` proposal I think we've covered enough ground to merge this and iterate from here.

/approve
/lgtm

Holding for one other maintainer

/cc @robscott @youngnick 
/hold

dprotaso:
Let me update the GEP since upstream k8s doesn't want to define a `raw` constant

Details in the discussion here: https://github.com/kubernetes/enhancements/pull/4106

geps/gep-1911.md
dprotaso: Resolving as raw is not supported upstream - https://github.com/kubernetes/enhancements/pull/4106

Will remove it from the GEP

LiorLieberman:
I know this is waiting for approval for a while, but didn't we say we gonna take it to the sig net meeting this week first?

dprotaso:
> I know this is waiting for approval for a while, but didn't we say we gonna take it to the sig net meeting this week first?

Yeah - I'll be doing that

dprotaso:
Did some minor cleanup

geps/gep-1911.md
```diff
@@ -0,0 +1,332 @@
...
+
+Implementations MUST recognize support for `TCP` and `UDP` multiplexing on the same port.
+
+Currently Kubernetes `Service` API does not allow different `appProtocol` values for the same port. At this time there seems to be interest in changing `appProtocol` to be a list in order to faciliate this use-case.
+
+Open Question: Can SCTP be multiplexed with TCP and UDP on the same port?
+
+### Unsupported/Incompatible Protocols
+
+If a Route attached to a Gateway is not able to sending traffic to the backend using the specified protocol then it must set its RouteConditionType `Accepted` to `False`, with a `RouteConditionReason` `UnsupportedProtocol`
```
pleshakov, line 71: considering that there could be multiple backends in the HTTPRoute, is it reasonable to invalidate the whole resource if there is a problem with just one backend?

geps/gep-1911.md
```diff
@@ -0,0 +1,332 @@
...
+
+## API Semantics
+
+A Gateway implementation MUST recognize the Kubernetes Standard Application Protocols ([KEP-3726][kep-3726]) for specifying the protocol for a backend reference in a Gateway API `*Route` resource
+
+Thus when a `*Route` points to a Kubernetes Service resource the backend protocol for each port can be specified by
+- setting the `appProtocol` field on the Kubernetes `Service`
+- setting the `appProtocol` field on an `Endpoint`/`EndpointSlice` object associated with a Kubernetes `Service`
+
+At the moment there exists two defined constants:
```
pleshakov, line 43:
```suggestion
At the moment there exists three defined constants:
```

geps/gep-1911.md
dprotaso: Good catch. There already is precendence for invalid backend. I'll follow that.

dprotaso: https://github.com/kubernetes-sigs/gateway-api/blob/d4e08e888f3d19ddbb5242324a6505899fc29e9c/apis/v1beta1/httproute_types.go#L1061-L1064

liwenwu-amazon:
AWS [aws-application-networking-k8s gateway, a K8S gateway API implementation](https://github.com/aws/aws-application-networking-k8s/blob/main/config/crds/bases/application-networking.k8s.aws_targetgrouppolicies.yaml#L109) have defined A CRD to allow user to define backend protocol

robscott:
GKE supports an annotation like this on Services `cloud.google.com/app-protocols: '{"my-https-port":"HTTPS","my-http-port":"HTTP"}'`. More info in: https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#https_tls_between_load_balancer_and_your_application

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+`TCP`  |`kubernetes.io/ws`  | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/ws`  | `TLSRoute`  | Conditional [1]
+`TCP`  |`kubernetes.io/wss` | `HTTPRoute` | Conditional [1]
+`UDP`  |`kubernetes.io/wss` | `HTTPRoute` | TBD [3]
+`TCP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`UDP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`TCP`  |`kubernetes.io/wss` | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/wss` | `TLSRoute`  | Conditional [1]
+
+1. Only if there is a corresponding `TLSConnectionPolicy` - see [GEP-1897](TODO)
```
youngnick, line 124: This doesn't make sense to me. If the backend protocol is `kubernetes.io/h2c` or `kubernetes.io/ws`, that says that the service is expecting unencrypted communication, which would rule out what's now called a `BackendTLSPolicy` (since that specifies that the service is expecting encrypted communication).

I'm still not a fan of allowing TLSRoutes with `tls.mode: terminate`, but I've lost that one, so _only because_ we allow that, I think that unencrypted protocols with TLSRoute makes sense.

So, I guess my requested change here is that the condition be updated to:

Only if the TLSRoute is attached to a Listener that has `tls.mode: terminate` set.

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+2. GRPC works over h2c
```
youngnick, line 125: Is it worth calling out that doing unencrypted gRPC over an external connection like a Gateway is pretty weird though?

youngnick (commented):
Some thoughts on the table caveats, but aside from that this looks pretty good to me.

I've specifically _not_ addressed the open questions at this time because I think we should all take a bit and think about them before doing anything. Let's get this merged at `Provisional` first.

geps/gep-1911.md
dprotaso: > Is it worth calling out that doing unencrypted gRPC over an external connection like a Gateway is pretty weird though?

I just wanted to highlight GRPC can happen over clear text. I updated the clause to be more specific that the clear text connection is from the gateway to the backend.

geps/gep-1911.md
dprotaso: updated please take a look

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+`TCP`  |`kubernetes.io/ws`  | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/ws`  | `TLSRoute`  | Conditional [5]
+`TCP`  |`kubernetes.io/wss` | `HTTPRoute` | Conditional [1]
+`UDP`  |`kubernetes.io/wss` | `HTTPRoute` | TBD [3]
+`TCP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`UDP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`TCP`  |`kubernetes.io/wss` | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/wss` | `TLSRoute`  | Yes
+
+1. Only if there is a corresponding `TLSConnectionPolicy` - see [GEP-1897](geps/gep-1897)
```
youngnick, line 124: This name got changed - it's now `BackendTLSPolicy` again...

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+`TCP`  |`kubernetes.io/ws`  | `GRPCRoute` | No
+`UDP`  |`kubernetes.io/ws`  | `GRPCRoute` | No
+`TCP`  |`kubernetes.io/ws`  | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/ws`  | `TLSRoute`  | Conditional [5]
+`TCP`  |`kubernetes.io/wss` | `HTTPRoute` | Conditional [1]
+`UDP`  |`kubernetes.io/wss` | `HTTPRoute` | TBD [3]
+`TCP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`UDP`  |`kubernetes.io/wss` | `GRPCRoute` | No
+`TCP`  |`kubernetes.io/wss` | `TCPRoute`  | TBD [4]
+`TCP`  |`kubernetes.io/wss` | `TLSRoute`  | Yes
```
robscott, line 122: This is pretty complex, I think it might be simpler to just have a smaller list of the combinations that we think should always be valid (`Yes`) in this table, and a separate list of the combinations that we think _may_ be valid (`Conditional`) in this table. I think that would mean we could also replace the tables above this one as well with just focusing on combinations that may be valid.

robscott (commented):
Thanks for all the work on this @dprotaso!

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Default Protocols
+
+If a Service `appProtocol` isn't specified an implementation MAY infer the backend protocol through its own means. Implementations MAY infer the protocol from the `Route` type referring to the backend Service.
+
+### Multiple Protocols on the same port
+
+Only the Kubernetes `Service` `protocol` field supports multiple protocols on the same port. See the details in [KEP-1435][kep-1435].
+
+Implementations MUST recognize support for `TCP` and `UDP` multiplexing on the same port.
```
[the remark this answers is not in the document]
dprotaso: > What does this mean? 

Gateways support K8s Service backends with different `protocol` values on the same port. Thus in theory I could point a TCPRoute and UDPRoute the referenced backend and they should be Accepted & Programmed.

>  In upstream Kubernetes the only requirement is to surface an error message if you can't support more than one protocol on the same port.

Are you suggesting we do the same?

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+
+Currently Kubernetes `Service` API does not allow different `appProtocol` values for the same port. At this time there seems to be interest in changing `appProtocol` to be a list in order to faciliate this use-case.
```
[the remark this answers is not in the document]
dprotaso: Good catch - I can note the exception when the `protocol` is different

geps/gep-1911.md
dprotaso: I like this approach - this leaves other situations 'undefined' for now

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+
+Open Question: Can SCTP be multiplexed with TCP and UDP on the same port?
+
+### Unsupported/Incompatible Protocols
+
+If a Route attached to a Gateway is not able to send traffic to the backend using the specified protocol then the backend is considered invalid. Implementations MUST set `ResolvedRefs` condition to `False` with the Reason `UnsupportedProtocol`.
+
+It is RECOMMENDED that implementaions follow the supported combinations below:
```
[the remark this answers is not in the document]
dprotaso: I changed the table to be MUST for obvious things and MAY for some conditional things

geps/gep-1911.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Multiple Protocols on the same port
+
+Only the Kubernetes `Service` `protocol` field supports multiple protocols on the same port. See the details in [KEP-1435][kep-1435].
+
+Implementations MUST recognize support for `TCP` and `UDP` multiplexing on the same port.
+
+Currently Kubernetes `Service` API does not allow different `appProtocol` values for the same port. At this time there seems to be interest in changing `appProtocol` to be a list in order to faciliate this use-case.
+
+Open Question: Can SCTP be multiplexed with TCP and UDP on the same port?
```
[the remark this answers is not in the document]
dprotaso: i'll drop the question since we inverted the table to be an allow list only

geps/gep-1911.md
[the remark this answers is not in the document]
robscott: >Are you suggesting we do the same?

I'm suggesting that we shouldn't have stricter requirements than Kubernetes upstream. I don't think we can require that every implementation of this API support multiplexing, so I think that would mean that implementations would have to surface some kind of error if they can't support it. Maybe we can standardize on what that is at least?

geps/gep-1911.md
```diff
@@ -0,0 +1,181 @@
...
+
+If a Route attached to a Gateway is not able to send traffic to the backend using the specified protocol then the backend is considered invalid. Implementations MUST set `ResolvedRefs` condition to `False` with the Reason `UnsupportedProtocol`.
+
+Implementations MUST supported the following combinations below:
+ServicePort Protocol | ServicePort AppProtocol | Route Type | Supported
+-|-|-|-
+`TCP` | `kubernetes.io/h2c` | `GRPCRoute` | Yes [1]
+`TCP` | `kubernetes.io/h2c` | `HTTPRoute` | Yes
+`TCP` | `kubernetes.io/ws`  | `HTTPRoute` | Yes
+`TCP` | `kubernetes.io/wss` | `TLSRoute`  | Yes
```
robscott, line 78: Does the absence of this AppProtocol fields mean that we shouldn't enable any websocket upgrades? What about implementations that already have this upgrade enabled by default in their underlying dataplane?

geps/gep-1911.md
```diff
@@ -0,0 +1,181 @@
...
+
+Implementations MUST recognize support for `TCP` and `UDP` multiplexing on the same port.
+
+Currently Kubernetes `Service` API only allows different `appProtocol` values for the same port when `protocol` fields differs. At this time there seems to be interest in changing `appProtocol` to be a list in order to faciliate this use-case.
+
+### Unsupported/Incompatible Protocols
+
+If a Route attached to a Gateway is not able to send traffic to the backend using the specified protocol then the backend is considered invalid. Implementations MUST set `ResolvedRefs` condition to `False` with the Reason `UnsupportedProtocol`.
+
+Implementations MUST supported the following combinations below:
```
robscott, line 72: I don't think that we can require implementations to support these combinations. We can recommend that they do, or even just state these combinations have well-defined interactions with Gateway API and should be supported wherever possible.

geps/gep-1911.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. GRPC works over h2c - so a GRPCRoute should be able to connect to an h2c backend
+
+Implementations MAY support the following combinations below:
+
+ServicePort Protocol | ServicePort AppProtocol | Route Type | Supported
+-|-|-|-
+ServicePort Protocol | ServicePort AppProtocol | Route Type | Supported
+`TCP`  | `kubernetes.io/h2c` | `TLSRoute`  | Conditional [1]
+`TCP`  | `kubernetes.io/ws`  | `TLSRoute`  | Conditional [1]
```
robscott, line 88: These feel a little wonky to me. Now that we're moving to "allow-list" only, I think I'd be in favor of omitting these unless we have a clear need for them. For example, we've said that [HTTPRoute can't be used with a TLS listener protocol](https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls). With that context, it feels weird to say that you can use h2c or ws with TLSRoute. (I recognize that we haven't actually directly said anything that would prevent the use of HTTP from a Gateway to a Backend with TLS as the listener protocol, just HTTPRoute, but it all feels awfully unsettled and underspecified right now and I'd hate to add to the potential confusion here).

robscott (commented):
Thanks @dprotaso, didn't go all the way through it, but I think I saw most of your updates.

geps/gep-1911.md
dprotaso: Should I change the entire GEP to be `RECOMMENDED` ? line 37 has a `MUST`

geps/gep-1911.md
dprotaso: > Does the absence of this AppProtocol fields mean that we shouldn't enable any websocket upgrades? 

No - this

> What about implementations that already have this upgrade enabled by default in their underlying dataplane?

No change - ideally users wouldn't need to set these `appProtocol` values and things just work :) 

I'll add some language to the `Default Protocol` section

dprotaso: Added

> Absence of the `appProtocol` field does not imply the implementation should disable any features (eg. websocket upgrades).

geps/gep-1911.md
sunjayBhatia: this section has some language that might help, talking about Listener protocols so a similar idea: https://github.com/kubernetes-sigs/gateway-api/blob/ea382e7a989d2849ce1a76d47a60bd5f5ea754e5/apis/v1beta1/gateway_types.go#L260-L266

(though the line referencing a table in the linked godoc might need to be removed now)

I think in other places `MAY` is used in similar contexts

If we anticipate protocol/kind combos in the tables to be tested in conformance then could add a mention of that? For standardization these combinations will be verified, but as extended conformance?

dprotaso: >  but as extended conformance so not a MUST?

OK sounds good

dprotaso: Circling back - I opt'd to leave the `MUST` requirement that implementations look at the `appProtocol` field to infer backend protocol.

And I changed the specific `support {x} protocol with {y} route` an optional (MAY) requirement

dprotaso:
@youngnick @robscott Can you take another look? 🙇

youngnick:
/lgtm

/hold

for @robscott's review though.

geps/gep-1911.md
```diff
@@ -0,0 +1,181 @@
...
+Rather than bundle protocol details into a single resource an alternative would be to create distinct meta resources.
+ie. `HTTP2Backend`, `GPRCBackend`, `WebsocketBackend`.
+
+The advantages of this approach are:
+- Easy to introduce new protocols
+- Definitions/types would be simpler
+
+The disadvantages of this approach are:
+- N resources for N protocols need to be created to describe a single backend
+- No easy mechanic to specify priority of protocols
```
robscott, line 128: New lines needed for list formatting.
```suggestion
The advantages of this approach are:

- Easy to introduce new protocols
- Definitions/types would be simpler

The disadvantages of this approach are:

- N resources for N protocols need to be created to describe a single backend
- No easy mechanic to specify priority of protocols
```

robscott (commented):
Thanks @dprotaso! One non-blocking nit. Feel free to remove hold and catch this in a follow up.

/lgtm

Issue #1936, opened 2023-04-15
Allow `host:port` in `HTTPRouteSpec.hostnames`


tommie:
**What happened**: The [`HTTPRouteSpec.hostnames`](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteSpec) does not allow HTTP authority that includes a port number suffix.

**What you expected to happen**: Either

1. The [`Hostname` validation pattern](https://github.com/kubernetes-sigs/gateway-api/blob/b751bbdce6245c24556ac8ee13b6de0797515bb8/apis/v1beta1/shared_types.go#L381) allows full authority,
2. the `HTTPRouteSpec.hostnames` is changed to a separate `HTTPAuthority` string with an updated validation pattern, or
3. it is clarified that HTTPRoutes are matched against the hostname only, ignoring the port.

**Edit:** Since there is interaction between `Listener.hostname` and these hostnames, that needs to be expanded on in solutions 1-2.

**Anything else we need to know?**: The `HTTPRouteSpec.hostnames` field is of type `[]Hostname`, which is specified to only use a simple domain name. However, in [HTTP/1](https://www.rfc-editor.org/rfc/rfc7230#section-5.4), the `Host` header optionally includes the port. In [HTTP/2](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2), the `:authority` pseudo-header carries the same syntax, but disambiguating the name. I.e.both use authority from [RFC 3986, sec 3.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2), without userinfo

Based on https://github.com/envoyproxy/gateway/pull/1311#issuecomment-1509429314.

howardjohn:
Prefer (3) personally (ignore port)

robscott:
+1 to @howardjohn on this one.

/help

PR #1980, opened 2023-04-28
Add clarification for ignoring port while matching hostnames and update conformance

gauravkghildiyal:
**What type of PR is this?**
/kind documentation
/kind test
/area conformance

Add clarification for ignoring port while matching HTTPRouteSpec.Hostnames. Update conformance test to add test case covering this scenario

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1936

```release-note
Implementations MUST ignore any port value specified in the HTTP Host header while performing a match against HTTPRoute.Hostnames
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -56,9 +56,13 @@ type HTTPRouteList struct {
 type HTTPRouteSpec struct {
 	CommonRouteSpec `json:",inline"`
 
-	// Hostnames defines a set of hostname that should match against the HTTP
-	// Host header to select a HTTPRoute to process the request. This matches
-	// the RFC 1123 definition of a hostname with 2 notable exceptions:
+	// Hostnames defines a set of hostname that should match against the HTTP Host
+	// header to select a HTTPRoute used to process the request. Implementations
+	// MUST ignore any port value specified in the HTTP Host header while
+	// performing a match.
+	//
+	// Valid values for Hostnames are determined by RFC 1123 definition of a
+	// hostname with 2 notable exceptions:
 	//
 	// 1. IPs are not allowed.
 	// 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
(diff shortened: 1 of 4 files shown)


howardjohn (commented):
LGTM, should we add a conformance test?

gauravkghildiyal:
Updated with test case

/kind test
/area conformance

arkodg:
Wanted to share the implications of this change, the spec proposes
```
 MUST ignore any port value specified in the HTTP Host header while performing a match
```

This makes `Host` header matches more loose. So if in the Gateway API resources,  the hostname is `example.com` and the listener port is `8080` , an incoming request with a Host header `example.com:9090` can be accepted by implementations. Ideally only `example.com` and `example.com:8080` should be accepted.
Having shared this, its also much easier for implementations to implement this proposed spec, so its a SGTM from my end.

howardjohn:
> Wanted to share the implications of this change, the spec proposes
> 
> ```
>  MUST ignore any port value specified in the HTTP Host header while performing a match
> ```
> 
> This makes `Host` header matches more loose. So if in the Gateway API resources, the hostname is `example.com` and the listener port is `8080` , an incoming request with a `Host` header value of `example.com:9090` can be accepted by implementations. Ideally only `example.com` and `example.com:8080` should be accepted. Having shared this, its also much easier for implementations to implement this proposed spec, so its a SGTM from my end.

I disagree with this. example.com:9090 SHOULD be allowed. The reason is things are almost always layered. The port the client called and the port the gw impl got are very often not the same. For example, maybe its an in-cluster deployment and they reached via the NodePort. Or they have a LB in front accepting connections on port 9090 and sending to the gw impl on port 8080, etc.

We saw lots of users broken when we had strict port matching like this in Istio.

youngnick:
I think @howardjohn raises a good point here that I certainly hadn't thought of. With that in mind, I think this change LGTM.

I'll approve and hold for further review though.

/approve
/hold

gauravkghildiyal:
I think this one is ready to go. Requires LGTM :)

robscott:
Thanks @gauravkghildiyal!

/lgtm
/hold cancel
/cherry-pick release-0.7

PR #1983, opened 2023-04-30
Added HTTP/2 support in conformance test roundtripper

### LINKED ISSUE #1982: Request for information/support on http2 transport enablement in conformance test
Need support for http2 in conformance test specifically the "roundtripper" package

**Why this is needed**:
To overcome limitation of just using http1 in the conformance test as the upstream backend host might just support only http2
https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/roundtripper/roundtripper.go#L111

**Additional Details **:
From the initial investigation, it looks like we cannot pass the custom HTTP client into the default round tripper implementation. If there is an alternate way to achieve this, I would like to be redirected there.

ibilalkayy:
What type of PR is this?

/kind feature
/area conformance

**What this PR does / why we need it:**

This PR adds support for HTTP/2 in the conformance test **roundtripper** package. The changes include updating the **Request** struct in **roundtripper.go** to include an HTTP2Transport field and modifying the **defaultTransport** function in **roundtripper.go** to create an HTTP/2 transport if required, using the http2.ConfigureTransport function from the **golang.org/x/net/http2** package.

These changes will enable users to set the **HTTP2Transport** field when creating a **Request** in the conformance test, allowing for HTTP/2 support. This addition helps overcome the limitation of just using HTTP/1 in the conformance test, especially when the upstream backend host might only support HTTP/2.

**Which issue(s) this PR fixes:**

Fixes #1982

**Does this PR introduce a user-facing change?:**

`Added support for HTTP/2 in the conformance test roundtripper package.
`

### DIFF
--- a/conformance/utils/roundtripper/roundtripper.go
+++ b/conformance/utils/roundtripper/roundtripper.go
@@ -28,6 +28,7 @@ import (
 	"net/url"
 	"regexp"
 
+	"golang.org/x/net/http2"
 	"sigs.k8s.io/gateway-api/conformance/utils/config"
 )
 
@@ -49,6 +50,7 @@ type Request struct {
 	KeyPem           []byte
 	Server           string
 	CustomDialer     http.RoundTripper
+	HTTP2Transport   http.RoundTripper
 }
 
 // String returns a printable version of Request for logging. Note that the
@@ -214,9 +216,23 @@ func (d *DefaultRoundTripper) CaptureRoundTrip(request Request) (*CapturedReques
 }
 
 func defaultTransport(request Request) (http.RoundTripper, error) {
+	// Use the provided HTTP2Transport if it's not nil
+	if request.HTTP2Transport != nil {
+		return request.HTTP2Transport, nil
+	}
+
 	// Setup TLS transport if there are CertPem, KeyPem, and Server in the request
 	if request.Server != "" && len(request.CertPem) != 0 && len(request.KeyPem) != 0 {
 		return tlsTransport(request.Server, request.CertPem, request.KeyPem)
+	}
+
+	// Create an HTTP/2 transport if required
+	if request.Protocol == "http2" {
+		transport := &http.Transport{}
+		if err := http2.ConfigureTransport(transport); err != nil {
+			return nil, err
+		}
+		return transport, nil
 	}
 
 	return http.DefaultTransport, nil


robscott:
Agree with @sunjayBhatia here, this would be best to save until we have test(s) that need this. @ibilalkayy was there a specific test you were hoping this would enable?

gauravkghildiyal:
Unless, I'm missing something, I believe one can already achieve the requirements of the issue. Refer comment https://github.com/kubernetes-sigs/gateway-api/issues/1982#issuecomment-1528727074

Please correct me if I misunderstood something.

sunjayBhatia:
> Unless, I'm missing something, I believe one can already achieve the requirements of the issue. Refer comment [#1982 (comment)](https://github.com/kubernetes-sigs/gateway-api/issues/1982#issuecomment-1528727074)
> 
> Please correct me if I misunderstood something.

All the existing conformance tests use a known backend that is deployed as part of the tests (the ingress-conformance echo server) so there isn't anything in the tests I'm aware of that an HTTP/2 backend is relevant for yet (the still-draft conformance tests for gRPC route look like they will have their own gRPC echo server but will use a gRPC client)

pramodrj07 (commented):
LGTM. I was thinking of passing the client. But this seems like a better solution.

aojea:
> gRPC echo server but will use a gRPC client

just important note for e2e tests and something that is a terrible headache is that grpc uses http2 under the hood https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md, so the only way to differentiate between http2 and grpc at the transport level is by checking the "content-type" header  that should begin with "application/grpc"

pramodrj07:
> t to save until we have test(s) that need thi

We are using conformance tests to test specific scenarios where we need http2 support. Currently we have a wrapper around the existing round tripper. But this native support will be much helpful going forward. 

@sunjayBhatia - Agree to your point on non-existence of any http2 related conformance test at the moment. Going forward do you see any support for http2 support in the near future?

sunjayBhatia:
> @sunjayBhatia - Agree to your point on non-existence of any http2 related conformance test at the moment. Going forward do you see any support for http2 support in the near future?

The outcome of this GEP should lead to some conformance tests being developed: https://github.com/kubernetes-sigs/gateway-api/pull/1979

shaneutt:
As this PR hasn't received an update in some time, and needs some conflict resolution I'm going to mark this as "draft" for the time being as this helps us to better organize our reviews. Please let us know if you're blocked!

Issue #1962, opened 2023-04-23
Conformance Test: Combinations of Extended HTTPRoute Matches


robscott:
**What would you like to be added**:
We currently have test coverage of Path, Header, Method, and Query Param matches. We even have tests that [cover the combination of Path and Header matches](https://github.com/kubernetes-sigs/gateway-api/blob/b8b01612a0924a34efadeedd544118e5557ac85a/conformance/tests/httproute-matching.go). To complete our test coverage we need tests that combine Query Param and Method matches with other kinds of HTTPRoute matches. 

**Why this is needed**:
This is the very last step of https://github.com/kubernetes-sigs/gateway-api/issues/1102, closing that one out in favor of a more focused bug.

PR #1984, opened 2023-05-01
Add conformance test for combination of extended match types with core types

gauravkghildiyal:
/kind test
/area conformance

**What type of PR is this?**

Add conformance test for combination of extended match types with core types

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1962

```release-note
NONE
```


conformance/tests/httproute-matching-combinations.go
```diff
@@ -0,0 +1,134 @@
...
+			{
+				Request:  http.Request{Headers: map[string]string{"version": "one"}},
+				Response: http.Response{StatusCode: 404},
+			},
+			{
+				Request:  http.Request{Path: "/?animal=whale"},
+				Response: http.Response{StatusCode: 404},
+			},
+			{
+				Request:  http.Request{Method: "POST"},
```
sunjayBhatia, line 113: seems like we need a test case that has all the expected content that would match a particular rule except the method to make sure all the match criteria is respected, I don't think there is any rule that matches just `/` (w/ no header or query params) so might be good to add another test case that e.g. sends a GET to `/path3?animal=whale` (I don't think the above request to `/?animal=whale` satisfies this b/c I don't think there is a rule that expects to match on that with a different method)

gauravkghildiyal: Sure I've added that. We do have cases which "matched all except one" but none where the "one" which didn't match was an HTTP method.

sunjayBhatia (commented):
test passes w/ Contour main 👍🏽

arkodg (approved):
LGTM thanks !

conformance/tests/httproute-matching-combinations.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: matching-combinations
+  namespace: gateway-conformance-infra
+spec:
+  parentRefs:
+  - name: same-namespace
+  rules:
```
mlavacca, line 9: Thank you @gauravkghildiyal! WDYT about adding another rule with all the possible `matches` set (path, header, method, queryparam)? If I see it correctly, we have many combinations, but none of them specifies all the possible `match` types.

gauravkghildiyal: Sure thing! Done. Thanks for the review

conformance/tests/httproute-matching-combinations.yaml
```diff
@@ -0,0 +1,99 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: matching-combinations
```
robscott, line 4: I think it would be very helpful to have some tests covering a use case where multiple rules match the same request. For example

Rule A:
- Path match on /foo

Rule B:
- Query param match on animal=whale

Test: Request to `/foo?animal=whale`. 

You may already have some similar test cases in here that I'm missing, but if not, I think it would be really helpful to have something like this. This is conceptually similar to https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-matching-across-routes.go, but would cover extended concepts like query param and method matching.

conformance/tests/httproute-matching-combinations.go
```diff
@@ -0,0 +1,144 @@
...
+var HTTPRouteMatchingCombinations = suite.ConformanceTest{
+	ShortName:   "HTTPRouteMatchingCombinations",
+	Description: "An HTTPRoute containing combinations of multiple match types",
+	Manifests:   []string{"tests/httproute-matching-combinations.yaml"},
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportHTTPRoute,
+		suite.SupportHTTPRouteQueryParamMatching,
+		suite.SupportHTTPRouteMethodMatching,
+	},
```
robscott, line 42: Similar to how the Redirect tests were split up into separate tests for each extended feature combined with all core features, it could be helpful to have subsets of this available. For the record, I don't think that needs to block this PR from merging, but could be a good follow up.

robscott (commented):
Thanks @gauravkghildiyal! I've got a couple non-blocking comments that don't need to block this PR, but could make for good follow ups here or in a future PR.

/approve

conformance/tests/httproute-matching-combinations.yaml
gauravkghildiyal: Interesting. I completely missed this. Thanks Rob! I've added some to get started

conformance/tests/httproute-matching-combinations.go
gauravkghildiyal: I went ahead and split it up here itself, since we're anyway waiting for LGTM. I wasn't aware we try to maintain such a pattern. But yes it does make sense. 

@robscott -- I guess this implies that if one wishes to add tests that cover ALL match types, that would go in it's own file/test. Will see if that is needed in a separate PR.

gauravkghildiyal (commented):
Thanks for the review Rob!

conformance/tests/httproute-matching-combinations.go
robscott: Yep, I think separate files are _probably_ sufficient here, but don't feel too strongly either way.

robscott:
Thanks @gauravkghildiyal!

/lgtm
/cherry-pick release-0.7

conformance/tests/httproute-query-param-matching.go
```diff
@@ -84,6 +84,68 @@ var HTTPRouteQueryParamMatching = suite.ConformanceTest{
...
+		// For requests that satisfy multiple matches, ensure precedence order
+		// defined by the Gateway API spec is maintained.
+		testCases = append(testCases, []http.ExpectedResponse{
+			{
+				Request:   http.Request{Path: "/path5?animal=hydra"},
+				Backend:   "infra-backend-v1",
+				Namespace: ns,
+			},
+			{
+				Request:   http.Request{Headers: map[string]string{"version": "four"}, Path: "/?animal=hydra"},
```
sunjayBhatia, line 143: Looks like this test case added after my review doesn't work with Contour

this assumes header matching gets precedence over method matching, but I'm not sure that is expressed anywhere explicitly in the spec, see https://github.com/kubernetes-sigs/gateway-api/issues/1993

gauravkghildiyal: Yes that was how I interpreted it -- given method matching is not in the list (and the only thing not in the list), it must have the least priority.

I'm sorry for causing any issues. I've sent https://github.com/kubernetes-sigs/gateway-api/pull/2052 as a temporary solution to help unblock (if this is the path we wish to take)

sunjayBhatia: thanks! yeah not a super strong opinion on where it lands in terms of priority, but should probably be something explicitly debated/documented 👍🏽

PR #1994, opened 2023-05-04
mesh: mark GEPs as Implementable

howardjohn:
**What type of PR is this?**

/kind gep

This has now met the bar to be promoted to Implementable, per https://gateway-api.sigs.k8s.io/geps/overview/?h=#process, as the GEP doc is complete. In fact, it has already been implemented by 2 vendors and has (some) conformance tests.

Fixes #

```release-note

```

### DIFF
--- a/geps/gep-1686.md
+++ b/geps/gep-1686.md
@@ -1,7 +1,7 @@
 # GEP-1686: Mesh conformance testing plan
 
 - Issue: [#1686](https://github.com/kubernetes-sigs/gateway-api/issues/1686)
-- Status: Provisional
+- Status: Implementable
 
 ## TLDR
 
--- a/geps/gep-1426.md
+++ b/geps/gep-1426.md
@@ -1,7 +1,7 @@
 # GEP-1426: xRoutes Mesh Binding
 
 * Issue: [#1294](https://github.com/kubernetes-sigs/gateway-api/issues/1294)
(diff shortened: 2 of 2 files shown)


robscott:
I think this is really just catching the GEPs up to the current state where we have multiple meshes that have either implemented or are implementing these proposals. Will leave formal approval for someone else, but this LGTM.

keithmattix (approved):
Agree with @robscott above. LGTM but will allow others to do the honors

michaelbeaumont (approved):
Makes sense. Implemented by Kuma and passing current conformance test suite

geps/gep-1686.md
```diff
@@ -1,7 +1,7 @@
 # GEP-1686: Mesh conformance testing plan
 
 - Issue: [#1686](https://github.com/kubernetes-sigs/gateway-api/issues/1686)
-- Status: Provisional
+- Status: Implementable
```
frankbu, line 4: At the same time, it would be good to merge #1957 which fixes a couple of issues with running the initial "Mesh" conformance tests.

robscott: That change mostly LGTM and should merge soon but don't think that needs to block this change.

robscott:
Thanks @howardjohn!

/lgtm
/approve

Issue #1742, opened 2023-02-20
GEP: Timeouts


youngnick:
**What would you like to be added**:

Gateway API needs whatever support we can do for timeouts. This GEP is to consider what timeouts are available, find some common ground, and then to implement that common ground somehow. (Very likely a Policy object, but not certain).

**Why this is needed**:
Timeouts are a critical part of any proxy infrastructure, and since many Gateway implementations use a proxy, they're important here as well.

youngnick:
The first step here is a preliminary GEP PR that talks about the Why and What we need to do; in this case this includes a review of all the implementations, what dataplanes they use, and what information we have about the data plane's timeout capabilities. That should set us up for what we _can_ do, which can then help us define what we _will_ do.

I've already done the comparison, and am in the process of writing this PR up.

shaneutt:
@arkodg provided several details relevant to this in https://github.com/kubernetes-sigs/gateway-api/issues/1741 which we should reference while working here.

candita:
Would like to request discussion of:
- tunnel timeouts (for websocket connections, maybe the same use case as @liangyuanpeng mentioned above
- server-side connection timeouts

youngnick:
With the merge of #2013, this GEP is actually implemented, so it should be moved to Experimental. I'll open a PR to do that.

rainest:
With https://gateway-api.sigs.k8s.io/geps/gep-1742/#timeout-values and #2013 there's no option that NGINX-based implementations would support, correct? I don't know of any that add timeout behaviors other than the idle/time between  individual reads behavior offered by ngx_http_proxy_module (https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#set_timeouts just configures those for the lua-nginx-module-based implementations).

[...]

Issue #1731, opened 2023-02-15
GEP: Configurable Retries in HTTPRoute


robscott:
**What would you like to be added**:
I would like to be able to configure the following in HTTPRoute:

1. The max number of times to retry a request
2. The reason(s) and/or status codes a request should be retried
3. The timeout for each retry attempt

I believe all 3 of these would be implementable for Envoy based implementations, the first 2 would be implementable for HAProxy based implementations, unclear what would be implementable for NGINX or others (cc @pleshakov @shaneutt).

**Why this is needed**:
This is a common feature request and represents a concept that would likely get tied to a variety of custom policies if we did not include it in the main API.

kflynn:
@robscott, on the mesh front, Linkerd actually doesn't support the fixed-number-of-retries concept. Instead there's a retry budget: if your retry budget is e.g. 20%, then as long as not more than 20% of your request volume is retries, Linkerd can continue retrying.

It would be lovely to be able to configure that form of retry, too, without having to resort to crazy custom stuff. Just to complicate your life. 😉

bowei:
That's a very interesting approach. Is the 20% calculated "globally" or on a-per proxy basis? This may be difficult for more distributed proxy systems to do for a global percentage.

dprotaso:
@kflynn does setting 0% disable retries?

robscott:
@ramaraochavali thanks for the reference! I'd assumed this would be implemented by [RetryPolicy in xDS](https://www.envoyproxy.io/docs/envoy/v1.26.1/api-v3/config/route/v3/route_components.proto#config-route-v3-retrypolicy) which does not seem to require retry budgets, but I'm also far from an Envoy expert so may be missing some nuance here.

frankbu:
Although configuring retry budgets and other circuit breaking mechanisms would be useful to support, I don't think they are typically configurable on the individual route (vs service) level. Since the title of this issue is about configuring retries on HTTPRoute specifically, I wonder if it make sense to start with a GEP that only addresses retry configuration on an HTTPRoute and then handle retry budgets, etc. (probably using policy attachments instead of explicit fields in the API) in a separate GEP.

@robscott If you want to assign the issue to me, I can put together a first pass GEP to discuss this further.

robscott:
Thanks @frankbu!

>Although configuring retry budgets and other circuit breaking mechanisms would be useful to support, I don't think they are typically configurable on the individual route (vs service) level. 

[...]

PR #1997, opened 2023-05-04
GEP-1742: HTTPRoute Timeouts API

frankbu:
**What type of PR is this?**

/kind gep

Adds API proposal to GEP-1742.
Related issues: #1742

```release-note

```


geps/gep-1742.md
```diff
@@ -301,8 +301,87 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+
+type HTTPRetry struct {
+	// Number of retries to be allowed for a given request.
+	Attempts int32 `json:"attempts,omitempty"`
+	// Timeout per attempt for a given request, including the initial call and any retries.
+	// Default is the value of the HTTPRouteRule request Timeout.
+	PerTryTimeout *time.Duration `json:"per_try_timeout,omitempty"`
+	// Specifies the conditions under which retry takes place.
+	// One or more conditions can be specified using a ‘,’ delimited list.
+	RetryOn string `json:"retry_on,omitempty"`
```
howardjohn, line 332: If its a list it should be `[]string` IMO

geps/gep-1742.md
```diff
@@ -301,8 +301,87 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// Flag to specify whether the retries should retry to other localities.
```
howardjohn, line 333: This seems under-specified given there is no mention of "localities" anywhere else in the repo. Can we expand on this a bit.

geps/gep-1742.md
```diff
@@ -301,8 +301,87 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+addressing issue: [#1731](https://github.com/kubernetes-sigs/gateway-api/issues/1731)
+by adding another field for retry conifguration in an HTTPRoute rule. This would also
+benefit many users and seems to be implementable by many (most?) Gateway implementations.
+
+### GO
+
+```go
+type HTTPRouteRule struct {
+	// Timeout for HTTP requests, disabled by default.
+	Timeout *time.Duration `json:"timeout,omitempty"`
```
howardjohn, line 318: Can we expand the specification for this a bit? Given the GEP already documents probably 20 types of "timeout" in https://gateway-api.sigs.k8s.io/geps/gep-1742/?h=1742#flow-diagrams-with-available-timeouts, having just a single field as "timeout" with no further explanation seems a bit ambiguous.

geps/gep-1742.md
```diff
@@ -301,8 +301,87 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+  // ...
+}
+
+type HTTPRetry struct {
+	// Number of retries to be allowed for a given request.
+	Attempts int32 `json:"attempts,omitempty"`
+	// Timeout per attempt for a given request, including the initial call and any retries.
+	// Default is the value of the HTTPRouteRule request Timeout.
+	PerTryTimeout *time.Duration `json:"per_try_timeout,omitempty"`
+	// Specifies the conditions under which retry takes place.
```
howardjohn, line 330: What conditions are valid?

geps/gep-1742.md
```diff
@@ -301,8 +301,87 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    timeout: 10s
+    retries:
+      attempts: 3
+      perTryTimeout: 2s
+      retryOn: 503
+```
+
+### Timeout values
+
+Timeout and PerTryTimeout values are formatted as 1h/1m/1s/1ms and MUST BE >= 1ms.
```
howardjohn, line 360: Is this used anywhere else in gateway-api or core k8s? Is there a more formal spec?

is 1h5m valid?

geps/gep-1742.md
frankbu: I explain below that HTTP error codes are valid and must be supported by implementations. Other implementation-specific identifier are also valid, depending on the implementation. 

1. Does that sound reasonable?
2. Do you think I need to explain it more clearly in the below text?

geps/gep-1742.md
frankbu: Updated

geps/gep-1742.md
frankbu: Added another section below.

geps/gep-1742.md
frankbu: I couldn't find any other use of duration in gateway-api. I couldn't find it in core k8s either, but maybe not looking in the right place? Otherwise, I think this is just the Go time.Duration string serialization that we'd use?

geps/gep-1742.md
frankbu: @arkodg Good question. I think it depends on whether implementations support configuring other kinds of timeouts at the HTTPRoute level. Based on the diagrams in the previous sections, I think Envoy does not, but unclear about the others.

@youngnick do you have more insight on this?

geps/gep-1742.md
```diff
@@ -301,8 +301,99 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// Timeout for HTTP requests, disabled by default.
+	Timeout *time.Duration `json:"timeout,omitempty"`
+	// Retry policy for HTTP requests.
+	Retries *HTTPRetry `json:"retries,omitempty"`
+	// ...
+}
+
+type HTTPRetry struct {
+	// Number of retries to be allowed for a given request.
+	Attempts int32 `json:"attempts,omitempty"`
```
kflynn, line 332: I have quite a few comments on retries, but I really think that the retry portion here needs to be pulled out into a different GEP. 🙂

kflynn (changes requested):
I'd vote for pulling retries into a separate GEP, but for pulling the `perTryTimeout` up into the main timeouts section. Thanks for diving into this!

geps/gep-1742.md
frankbu: @kflynn This suggestion sounds quite good to me. The reason that I dragged retries into this GEP in the first place is because perTryTimeout was so tightly coupled to retries, but moving it out and generalizing it like you suggest allows for better separation of concerns and to only focus on Timeouts in this GEP. I'm happy to make this change but will give others some time to comment first.

youngnick:
Just a note to say I have seen this, but need a bit more time to give it a full review. I should hopefully be able to do this later on tomorrow.

geps/gep-1742.md
sunjayBhatia: +1 to this

frankbu:
@youngnick @howardjohn @kflynn @arkodg and other reviewer:

I have made a number of updates to address reviewer comments and removed retry configuration from this proposal. This PR is now only addressing timeout configuration for GEP issue #1742. Will move discussion of retries to a separate GEP for issue #1731.

geps/gep-1742.md
```diff
@@ -301,8 +301,100 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    timeouts:
+      request: 10s
+      backendRequest: 2s
+```
+
+### Timeout values
+
+There a 2 kinds of timeouts that can be configured in an `HTTPRouteRule`:
+
+1. `timeouts.request` is an end-to-end timeout for a single request as measured by the client.
```
howardjohn, line 390: Pedantic: we cannot measure the client end-to-end time, as there is latency between the client and us.

kflynn: Since I originally wrote that description... uh... maybe describe it as "an end-to-end timeout for a single request from the client to the Gateway API implementation", contrasted with "a single backend request from the Gateway API implementation to a workload"?

This would probably benefit from a diagram.

geps/gep-1742.md
```diff
@@ -301,8 +301,100 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+most valuable to clients.
+
+From the above analysis, it appears that most (all?) implementations are capable of
+supporting the configuration of simple client downstream request timeouts on HTTPRoute
+rules. This is a relatively small addition that would benefit many users.
+
+A client HTTP request to a gateway may result in more than one call to the destination
+backend service, for example, if automatic retries are supported. Some implementations
+support configuring a timeout for the individual backend requests, seperate from the
+overall client request timeout. Adding None-Core support for this would also benefit
```
kflynn, line 317: s/None-Core/non-Core/ ?

kflynn (approved):
@frankbu Thanks much!! I left a couple of small comments about wordsmithing but honestly, I'd be OK merging this. Others may have other thoughts, of course. 🙂

geps/gep-1742.md
frankbu: Actually, this is not so pedantic. At a high level, the "useful" timeout that we're trying to support here is basically the time it takes to get a response. In the various diagrams above, it seems like there is a timeout that corresponds to the `Finishes Response` arrow, which is more-or-less what we want, but the question of when the timeout starts measuring is important. John pointed out to me that in the case of Envoy, the implementation route timeout (`R timeout` in the above diagram) does not start `until the entire downstream request stream has been received`. Not sure about Nginx and other impls, but seems we need to be more flexible with this definition, to make it implementable.

Maybe something like this: 

`timeouts.request` is the timeout for the Gateway API implementation to send a response to a client HTTP request. Whether the gateway starts the timeout before or after the entire client request stream has been received, is implementation dependent.

Maybe we could first suggest the timeout begins after the entire client request stream has been received, and see if that is a problem for non-Envoy implementations?

Thoughts?

youngnick: Looking at the diagrams I did earlier (needed to recover context myself), I think that it's going to be difficult to define this timeout in such a way that _all_ the implementations can implement it.

I was going to say something like "from when the proxy has finished receiving the request from the client to when it starts sending the response back to the client", but some proxies can't do that, I think.

I'd definitely encourage @frankbu to take the initial source diagram and try and mark where the request timeout would cover on a diagram for this section, and for the `backendRequest` timeout as well.

It _may_ be that we need to be more vague here and say something like "this timeout covers as close to the whole request->response transaction as is possible for the proxy implementation, given its configurable timeouts." and then leave some of it up to the implementations. This would have the downside of being difficult to test though.

youngnick (commented):
Overall, the structure looks okay to me - although I think that as it stands, this proposal will _not_ be able to be combined with a TimeoutPolicy to _default_ or _override_ these fields, as we currently don't allow Policy fields to mess with the values of _fields_ within an object.

That would just mean that this setting would only be configurable on two levels: a default in the implementation somehow, configured via runtime config, environment variables, or similar, and then a HTTPRoute-specific value which _must_ be configured on _every_ HTTPRoute which needs a different timeout to the default. (The additional boilerplate adds up over lots of separate objects over time, in my experience).

geps/gep-1742.md
frankbu: I think that the common timeout we want to handle is for when it's taking too long to get the backend response. We should be able to test that, even if we make the spec a little vague about whether the timeout value includes other request overhead or not.

I'll update the wording to make it more clear, including adding the diagram with the proposed timeouts marked.

frankbu:
> I think that as it stands, this proposal will not be able to be combined with a TimeoutPolicy to default or override these fields, as we currently don't allow Policy fields to mess with the values of fields within an object.

@youngnick I understand and agree that a TimeoutPolicy shouldn't be able `override` the fields in the route rule, but why couldn't we (in the future) allow a TimeoutPolicy to set the `default` values at the namespace, gateway, etc. level?

geps/gep-1742.md
frankbu: Updated wording and added diagram, PTAL.

https://github.com/kubernetes-sigs/gateway-api/blob/90606a864a21c81f9db9c4d99b63e3fa25c37895/geps/gep-1742.md#timeout-values

geps/gep-1742.md
```diff
@@ -301,8 +301,133 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    P->>U: Finishes Headers
+    U->>P: Starts Response
+    U->>P: Finishes Headers
+    note right of P: timeouts.backendRequest timeout
+    note left of P: timeouts.request timout
+    U->>P: Finishes Response
+    note right of P: Repeat if retry
+    P->>C: Starts Response
+    P->>C: Finishes Headers
+    P->>C: Finishes Response
```
kflynn, line 425: I think that `note left of P: timeouts.request timeout` should move from line 420 to here, as

```
note left of P: timeouts.request end time`
```

geps/gep-1742.md
```diff
@@ -301,8 +301,133 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    note left of P: timeouts.request start time (max)
+    C->>P: Finishes request
+    P->>U: Connection Started
+    note right of P: timeouts.backendRequest start time
+    P->>U: Starts sending Request
+    P->>U: Finishes request
+    P->>U: Finishes Headers
+    U->>P: Starts Response
+    U->>P: Finishes Headers
+    note right of P: timeouts.backendRequest timeout
```
kflynn, line 419: s/timeouts.backendRequest timeout/timeouts.backendRequest end time/ ?

kflynn (changes requested):
Hey @frankbu, thanks! I left a note for one thing in the diagram that I think needs to be moved -- the diagram helps enormously, thanks!

geps/gep-1742.md
frankbu: I thought about that but put it where it is for 2 reasons:

1. All the implementations seem to have a timeout that corresponds to the point of starting to send the response, so it shouldn't be a problem to support.
2. Seems to me that from a client's perspective, once the response starts coming back it's reasonable to expect that the requested timeout no longer applies.

So from a practical point of view the diagram should stay as is IMO.

geps/gep-1742.md
```diff
@@ -301,8 +301,133 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    U->>P: Finishes Response
+    note right of P: Repeat if retry
+    P->>C: Starts Response
+    P->>C: Finishes Headers
+    P->>C: Finishes Response
+    Note right of P: Repeat if connection sharing
+    U->>C: Connection ended
+```
 
+Both timeout fields are string duration values as specified by ISO 8601 and MUST be >= 1ms.
```
howardjohn, line 430: So I would set

```
timeout: PT1M
```

for a 1 minute timeout?

geps/gep-1742.md
```diff
@@ -301,8 +301,133 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+metadata:
+  name: timeout-example
+spec:
+  ...
+  rules:
+  - backendRefs:
+    - name: some-service
+      port: 8080
+    timeouts:
+      request: 10s
```
howardjohn, line 383: This is not `ISO 8601` as specified

geps/gep-1742.md
frankbu: Yeah, I noticed that too. It's actually the Golang `time.Duration` string format, which somewhere I read is "based on ISO 8601", or something like that. In the comment in front of the go struct I have this:
```
// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
```
Not sure what I should reference where it has the complete syntax documented? The Golang time.ParseDuration doc is pretty vague: https://pkg.go.dev/time#ParseDuration

geps/gep-1742.md
frankbu: Maybe, but I'm still waiting to see how well the min/max proposal for start time goes over with other reviewers 😃

kflynn: 😂 Tough to argue with that. 😂

geps/gep-1742.md
frankbu: Changed to say format is as time.ParseDuration

sunjayBhatia: thats probably my fault, I added a link to the POC PR mentioning ISO 8601 but didnt read much further into the actual format

apimachinery metav1.Duration is pretty light on documentation too, can maybe just rip the field documentation from go if we need to be specific: https://pkg.go.dev/maze.io/x/duration#ParseDuration

geps/gep-1742.md
sunjayBhatia: ah looks like this was already discussed, so maybe disregard my comment below

sunjayBhatia:
> @youngnick I understand and agree that a TimeoutPolicy shouldn't be able to override the fields in the route rule, but why couldn't we (in the future) allow a TimeoutPolicy to set the default values at the namespace, gateway, etc. level?

I think the documented recommendation now (have to find it in a GEP etc. can't remember where it is at the moment) is to have Route fields and Policies be concerned with different concepts just for ease of understanding, but this doesn't sound unreasonable, similar to the idea of having a Policy targeting a Gateway that sets some global request/response headers to add/remove/etc.

frankbu:
@shaneutt and other reviewers,

I think this PR and the corresponding Go Interfaces (PR #2013) are ready now. What is the process to finish the review and get sign-off so they can merge?

geps/gep-1742.md
```diff
@@ -301,8 +301,134 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
 
 ## API
 
-TBD.
+The above diagrams show that there are many different kinds of configurable timeouts
+supported by Gateway implementations: connect, idle, request, upstream, downstream.
+Although there may be opportunity for the specification of a common API for more of
+them in the future, this GEP will focus on the L7 timeouts in HTTPRoutes that are
+most valuable to clients.
```
robscott, line 308: I think one of the most important questions here is _where_ the timeout config belongs. This proposes HTTPRouteRule, but I think that both HTTPBackendRef and/or a new policy to extend Service would be compelling alternatives. It would be helpful if this GEP described why HTTPRouteRule is proposed and why the others alternatives were not chosen.

geps/gep-1742.md
```diff
@@ -301,8 +301,134 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// For example, the following rule will timeout if a client request is taking
+	// longer than 10 seconds to complete:
+	//
+	// ```
+	// rules:
+	// - timeouts:
+	//     request: 10s
+	//   backendRefs:
+	//     ...
+	// ```
```
robscott, line 351: Although we've done it in some places throughout the API, I'd prefer to avoid adding new YAML snippets in our godocs because they're quite painful to support when generating docs.

geps/gep-1742.md
```diff
@@ -301,8 +301,134 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+    - name: some-service
+      port: 8080
+    timeouts:
+      request: 10s
+      backendRequest: 2s
+```
+
+### Timeout values
+
+There a 2 kinds of timeouts that can be configured in an `HTTPRouteRule`:
```
robscott, line 389:
```suggestion
There are 2 kinds of timeouts that can be configured in an `HTTPRouteRule`:
```

geps/gep-1742.md
```diff
@@ -301,8 +301,134 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+
+1. `timeouts.request` is the timeout for the Gateway API implementation to send a
+    response to a client HTTP request. Whether the gateway starts the timeout before
+    or after the entire client request stream has been received, is implementation dependent.
+    This field is required `Core` support.
```
robscott, line 394: Usually fields marked as `Core` support have tighter requirements than this and don't have implementation-specific components. I think this might be OK, but we need to be confident that we can write a reliable conformance test for this that would be compatible with all implementations. Are we sure that _all_ implementations can support this?

geps/gep-1742.md
```diff
@@ -301,8 +301,134 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+
+1. `timeouts.backendRequest` is a timeout for a single request from the gateway to a backend.
+    This field is optional `Extended` support. Typically used in conjuction with retry configuration,
+    if supported by an implementation.
+    Note that retry configuration will be the subject of a separate GEP (GEP-1731).
```
robscott, line 399: Any reason not to leave this for a GEP focused on retry configuration?

robscott (commented):
Thanks for the work on this @frankbu!

geps/gep-1742.md
frankbu: My first thought was to make it a requirement that the timeout start after the client request has been fully received, @youngnick had the same thought apparently (https://github.com/kubernetes-sigs/gateway-api/pull/1997#discussion_r1195445362, but unfortunately it seems that some proxies can't do that so this is the compromise.

I think a conformance test compatible with all instances is one that simply checks that a request times out when a backend is very slow, which btw is the main motivating use case for this feature.

geps/gep-1742.md
frankbu: This is just another timeout, so IMO belongs in this GEP. Even though it's most commonly used in conjunction with retries, it could legitimately be used on its own, for example if a client is only interested in timing out if the backend is slow, but wants to be tolerant of delays on the client side. @kflynn was the one who suggested clearly separating it from retries, which I agree was a great idea.

frankbu:
@robscott Thanks for the review! I made several changes to address your comments, PTAL

geps/gep-1742.md
```diff
@@ -1,7 +1,7 @@
 # GEP-1742: Timeouts
 
 * Issue: [#1742](https://github.com/kubernetes-sigs/gateway-api/issues/1742)
-* Status: Provisional
+* Status: Implementable
```
shaneutt, line 4: I've been considering if we should let this one soak for just a _bit_ longer in Provisional after this PR: what are your thoughts and feelings on changing the status in a separate follow-up PR to allow just a little bit more time for things to settle, would that be painful?

shaneutt, line 4: The more I think about it the more I feel like we should merge this PR in a `Provisional` state to allow just a bit more soak time, and to invite others to bring their own PRs to the table prior to when we would start implementation. We could bring the GEP up at the next couple of weekly syncs just to let people know now's a good time to chime in, and maybe add a few commits and ideas of their own.

Let me know your thoughts?

shaneutt (changes requested):

geps/gep-1742.md
shaneutt, line 4: The more I think about it the more I feel like we should merge this PR in a `Provisional` state to allow just a bit more soak time, and to invite others to bring their own PRs to the table prior to when we would start implementation. We could bring the GEP up at the next couple of weekly syncs just to let people know now's a good time to chime in, and maybe add a few commits and ideas of their own.

```suggestion
* Status: Provisional
```

Let me know your thoughts?

shaneutt (changes requested):

geps/gep-1742.md
frankbu: That's a good idea. Done.

shaneutt: Thank you, I'm adding this GEP to the next couple weeks of agenda to provide reminders that we're just about ready to move to `Implementable` and now is a critical time for anyone interested who hasn't yet jump in to jump in. :+1:

shaneutt (approved):

geps/gep-1742.md
```diff
@@ -301,13 +301,167 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+supported by Gateway implementations: connect, idle, request, upstream, downstream.
+Although there may be opportunity for the specification of a common API for more of
+them in the future, this GEP will focus on the L7 timeouts in HTTPRoutes that are
+most valuable to clients.
+
+From the above analysis, it appears that, at a minimum, implementations are capable of
+supporting the configuration of simple client downstream request timeouts on HTTPRoute
+rules. This is a relatively small addition that would benefit many users.
+
+A client HTTP request to a gateway may result in more than one call to the destination
```
sunjayBhatia, line 314: nit: I would move the `Some implementations support...` sentence to the beginning of this block to make it clearer its starting a new idea

geps/gep-1742.md
```diff
@@ -301,13 +301,167 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+type HTTPRouteTimeouts struct {
+	// Request specifies a timeout for the Gateway to send a response to a client HTTP request.
+	// Whether the gateway starts the timeout before or after the entire client request stream
+	// has been received, is implementation dependent.
+	//
+	// For example, setting the `rules.timeouts.request` field to the value `10s` in an
+	// `HTTPRoute` will cause a timeout if a client request is taking longer than 10 seconds
+	// to complete.
+	//
+	// Request timeouts are disabled by default.
```
robscott, line 347: This may be problematic. I think many implementations likely already set default timeouts (for example GKE Gateway uses 30 seconds by default). 

Unfortunately we can't set a default on a new field in a backwards compatible way as that would change the behavior of the API during an upgrade.

I think we may need to loosen this up to say something like "When this field is unspecified, request timeout behavior is implementation-specific."

Would be helpful to see if other implementations are already setting defaults here.

/cc @arkodg @howardjohn @sunjayBhatia @mlavacca @shaneutt @youngnick

geps/gep-1742.md
frankbu: reworded, thanks

geps/gep-1742.md
```diff
@@ -301,13 +301,167 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// to a backend service. Typically used in conjuction with retry configuration,
+	// if supported by an implementation.
+	//
+	// Default is the value of Request timeout.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	BackendRequest *metav1.Duration `json:"backendRequest,omitempty"`
```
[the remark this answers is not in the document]
frankbu: > it should be LET

Oh, did you mean to say LESS? If so, good idea. Done.

geps/gep-1742.md
sunjayBhatia: As far as I remember Envoy doesn't have an exact analogy for this but we might approximate to use something like the "response timeout" field which measures the time taken for the backend to return a response and has a default of 15s: https://projectcontour.io/docs/1.25/config/api/#projectcontour.io/v1.TimeoutPolicy

geps/gep-1742.md
sunjayBhatia: > I think a conformance test compatible with all instances is one that simply checks that a request times out when a backend is very slow, which btw is the main motivating use case for this feature.

rereading this intent, I'm not sure the description and naming of this field really line up? may be nitpicky and discussed before but did we consider naming this something like `backendRequest` if the intent is to cover the case that a backend is slow seems like the naming should reflect that

`the timeout for the Gateway API implementation to send a response to a client HTTP request` sounds like something the client itself should be enforcing (as it is written here)

frankbu: > did we consider naming this something like `backendRequest`

This proposal actually has an optional non-Core field named `backendRequest` 😃 It's used to set a timeout for a single request from the gateway to the backend. The `request` timeout is controlling how long the client should wait for a response given that the gateway may be making multiple backend requests (e.g., retries) before replying. Note that without retries or other fancy things going on, the two timeouts degenerate to more-or-less the same thing.

geps/gep-1742.md
robscott: > there will be ways to override the default (e.g., a policy attachment on the Gateway

Note that policy attachment GEP intentionally avoids configuring anything that overlaps with config in Gateway API resources. In its current form, it's meant to extend Gateway API, not override existing values. If we ever extend the scope of policy attachment to include this, it will only be after we can make it _very easy_ for users to understand what's happening, primarily to avoid issues like @kflynn described in #2015.

frankbu: Understood, but I would argue that the default values do not overlap with config in Gateway API resources unless we also provide some API to set them. Regardless, I think your point is correct that we unfortunately need to say that the default is implementation-specific.

geps/gep-1742.md
sunjayBhatia: ah right, confused that field since it was talking about retries as well so forgot about it, but that field sounds to me to cover more like what the intent of your comment about testing above is

the language `The request timeout is controlling how long the client should wait for a response...` similarly `timeouts.request is the timeout for the Gateway API implementation to send a response to a client HTTP request` sounds like a timeout that a client would enforce

If the field is documented like this but really is implemented differently/means something else I think this will be confusing for users, implementers, and those trying to give advice on supporting an implementation (we already have trouble with users of Contour getting confused about timeouts we expose, which are a lot more specific in that they tie in to Envoy timeouts and cover specific hops in the request flow)

(this and other threads may be rehashing why this wasn't done earlier, generalizing timeouts feels like it will lead to confusion)

sunjayBhatia: another general option as opposed to trying to generalize timeouts would be to make timeouts for all hops of a proxied request "extended" support and implementations could advertise what they support via conformance profile reporting as well as documentation

frankbu: I was trying to describe `timeout.request` in a way to be clear that it is `for the Gateway API implementation` to timeout doing it's job (which mostly means while waiting for 1 or more backend requests to finish). Maybe there's a better way to word it? I'll give it some thought, and suggestions are welcome 😃

frankbu: > which has advantages but also a lot of disadvantages tbh

Yeah, it would be nice to make at least something Core.

kflynn (approved):
Still looks good to me, FWIW. 🙂

kflynn (approved):
Still looks good to me, FWIW. 🙂

geps/gep-1742.md
```diff
@@ -301,13 +301,167 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service. Typically used in conjuction with retry configuration,
+	// if supported by an implementation.
+	//
+	// The value of BackendRequest defaults to and must be <= the value of Request timeout.
+	//
+	// Support: Extended
```
costinm, line 361: What happens if you have 2 routes with same backend, and define 2 different backend timeouts ? 

Will it mean that the backend request will have different timeouts depending on path ? Are all implementations capable of doing this ?

geps/gep-1742.md
frankbu: @robscott changed default to implementation-dependent.

geps/gep-1742.md
frankbu: Updated the comment in the Go struct for more clarity.

geps/gep-1742.md
candita, line 1: When I first read this I wondered why it only focused on HTTPRoute, but later I realized the original GEP specified that focus.  It would be a good idea to change the title to "HTTPRoute Timeouts API", or else mention in Non-Goals that other route types are not covered.

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// `HTTPRoute` will cause a timeout if a client request is taking longer than 10 seconds
+	// to complete.
+	//
+	// When this field is unspecified, request timeout behavior is implementation-dependent.
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
```
candita, line 354:
```suggestion
	ClientRequest *metav1.Duration `json:"clientRequest,omitempty"`
```

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+From the above analysis, it appears that, at a minimum, implementations are capable of
+supporting the configuration of simple client downstream request timeouts on HTTPRoute
+rules. This is a relatively small addition that would benefit many users.
+
+Some implementations support configuring a timeout for individual backend requests,
+separate from the overall client request timeout. This is particularly useful if a
+client HTTP request to a gateway can result in more than one call from the gateway
+to the destination backend service, for example, if automatic retries are supported.
+Adding non-Core support for this would also benefit many users.
+
```
candita, line 319: Can you clarify if the client request covers the entire connection from client to backend, or just the downstream request from client to gateway?   There is also at least one tunnel timeout (HAProxy) and perhaps some other timeouts that cover the entire connection, not just the downstream and upstream halves of the connection.

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+
+	// ...
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms.
+type HTTPRouteTimeouts struct {
```
candita, line 338: I would also add a name and/or type to the HTTPRouteTimeouts in case there are multiple needed for any given route.  For example, maybe an idle timeout would be a different duration than a connect timeout.

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+(e.g., `HTTPBackendRef`) or course (e.g. `HTTPRoute`) level of granularity.
+
+The downside, however, is complexity introduced for the most common use case, adding a simple
+timeout for an HTTP request. Setting a single field in the route rule, instead of needing to
+create a policy resource, for this simple case seems much better.
+
+Note that in the future, we could use policy attachments to configure other kinds of less common
+timeouts that may be needed. The default values of the proposed timeout fields could also be overriden
+using policy attachments in the future. For example, a policy attachment could be used to set the
+default value of `rules.timeouts.request` for all routes under an `HTTPRoute` or `Gateway`.
```
candita, line 452: Using Policy Attachments to override is a great idea, but I would remove the suggestion that they be used to supplement the existing timeouts with less common timeouts.  I think adding a name and type to the timeouts and allowing multiples is better.

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+
+### Other API Objects
+
+The new timeouts field could be added to a different API struct, instead of `HTTPRouteRule`.
+
+Putting it on an `HTTPBackendRef`, for example, would allow users to set different timeouts for different
+backends. This is a feature that we believe has not been requested by existing proxy or service mesh
+clients and is also not easily implementable.
```
candita, line 460: I would omit the comment about "not easily implementable" unless you explain why.  Maybe also add this to Non-Goals?
```suggestion
clients and is not being considered for this proposal.
```

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// +optional
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+
+	// ...
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms.
```
candita, line 337: I suggest having a meaningful zero-value for the timeouts, [like Envoy does](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-timeout).  "A value of 0 will completely disable the .... timeout", even if other timeouts are configured.

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service. Typically used in conjuction with retry configuration,
+	// if supported by an implementation.
+	//
+	// The value of BackendRequest defaults to and must be <= the value of Request timeout.
```
candita, line 360: Does the Request timeout encompass the BackendRequest timeout, or are they separate halves of the total timeout?

geps/gep-1742.md
frankbu: Title changed

geps/gep-1742.md
```diff
@@ -301,13 +301,168 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms.
+type HTTPRouteTimeouts struct {
+	// Request specifies the duration for processing an HTTP client request after which the
+	// gateway will time out if unable to send a response.
+	// Whether the gateway starts the timeout before or after the entire client request stream
+	// has been received, is implementation-dependent.
```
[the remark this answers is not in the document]
frankbu: I agree that it would be better to not have this implementation dependent start time. The problem is that (based on the evaluation that @youngnick did) we believe that it can't be supported by **all** implementations. So, the alternative is to make the feature "Extended", but we instead are trying to use this compromise so that we can at least have something be part of "Core" support. FYI, here is some of the background discussion on this: https://github.com/kubernetes-sigs/gateway-api/pull/1997#discussion_r1195375173

geps/gep-1742.md
frankbu: There is a diagram in the "Timeout values" section that shows the relationships, but you need to look at the rendered markdown to see it. https://github.com/kubernetes-sigs/gateway-api/blob/7e20c3d25368b2c91d0952dfbd842f46e3324b36/geps/gep-1742.md#timeout-values

geps/gep-1742.md
frankbu: This sounds like a good idea to me. I'll add that 0 means disable (no timeout). Thanks.

geps/gep-1742.md
frankbu: Yes, it encompasses backendRequest. It's shown in the [diagram](https://github.com/kubernetes-sigs/gateway-api/blob/7e20c3d25368b2c91d0952dfbd842f46e3324b36/geps/gep-1742.md#timeout-values) but maybe I should explain it better in the API comment too?

geps/gep-1742.md
frankbu: I'll leave it in since it is an alternative, but I'll add something like "but it would probably be better to extend the proposed API in the future for those timeouts as well".

geps/gep-1742.md
frankbu: Clarified not easily implementable

geps/gep-1742.md
frankbu: I tend to prefer keeping it short-and-sweet (Request) but will switch if others also think ClientRequest would be better.

geps/gep-1742.md
candita: Ah, ok, so maybe you need to clarify in Goal and Non-Goals which timeout values you're addressing.   If you call it "request", that doesn't mean much.  What transactions are covered in "request"?

geps/gep-1742.md
candita: I did see the diagram, yet still wanted to get the explanation in words and in the godoc of the implementation section for `Request` and `BackendRequest`.  I asked about something similar in that section.

I don't see any timeout that would cover the entire transaction, unless `Request` does.

geps/gep-1742.md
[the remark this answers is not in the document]
candita: I interpret that discussion a little differently than saying it is implementation-dependent.  It doesn't look like there was a conclusion on it, but I think @youngnick is asking for a better clarification on what exactly we're measuring here.  It's still just a little too vague.

geps/gep-1742.md
candita: That would be helpful!

geps/gep-1742.md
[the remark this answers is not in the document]
frankbu: @candita, I added more detailed doc in the comments of both timeout fields. PTAL.

geps/gep-1742.md
youngnick: I generally prefer greater specificity too, but this timeout as designed is broadly similar to what's called a "request timeout" by a few of the proxy implementations at least, so I think in this case, the more general term is okay.

geps/gep-1742.md
```diff
@@ -301,13 +301,179 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+  rules:
+  - backendRefs:
+    - name: some-service
+      port: 8080
+    timeouts:
+      request: 10s
+      backendRequest: 2s
+```
+
+### Timeout values
```
youngnick, line 395: I think this section should go before the `API` section, to further explain the timeouts that are being added.

geps/gep-1742.md
```diff
@@ -301,13 +301,179 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service.
+	//
```
youngnick, line 361: I can see the definition in the diagram below, but this struct needs to have a full definition of what BackendRequest means.

That is, there should be something like this here:

The BackendRequest timeout covers the time from when the first request starts being sent from the Gateway to when the full response has been received from the backend.

geps/gep-1742.md
```diff
@@ -301,13 +301,179 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// An entire client HTTP transaction with a gateway, covered by the Request timeout,
+	// can result in more than one call from the gateway to the destination backend service,
```
youngnick, line 363: This should be "may".

youngnick (commented):
I think that this will end up being annoying to use in practice, so I expect that we'll be following up in the not-too-distant future with Policy to allow _defaulting_ this, and probably something to allow setting bounds on the maximum and minimum value as well. I anticipated that we would do this all in one step originally.

That said, it's probably better to start with a more straightforward proposal like this one, that at least hits the basic use case, and see what people actually need. Maybe I'm wrong about how annoying configuring this in _every_ rule where it's needed will be.

Nice work @frankbu, I'm +1 on this. I'll check the comments and labels before putting any commands in here though.

geps/gep-1742.md
frankbu: done

geps/gep-1742.md
frankbu: done

geps/gep-1742.md
```diff
@@ -298,16 +298,182 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+	// to complete.
+	//
+	// This timeout is intended to cover as close to the whole request-response transaction
+	// as possible although an implementation MAY choose to start the timeout after the entire
+	// request stream has been received instead of immediately after the transaction is
+	// initiated by the client.
+	//
+	// When this field is unspecified, request timeout behavior is implementation-dependent.
+	//
+	// Support: Core
```
pleshakov, line 399: NGINX doesn't support request timeouts. However, not sure how many other implementations are in the same situation. 

Is it possible to have this feature as Extended?

frankbu: I thought NGINX's `proxy_read_timeout` aligned with `Request` timeout, is it a `BackendRequest` timeout, instead? If so, maybe `BackendRequest` should be Core, and `Request` should be Extended? Is there an NGINX timeout that takes retries into account?

pleshakov: > I thought NGINX's proxy_read_timeout aligned with Request timeout, is it a BackendRequest timeout, instead?

unfortunately `proxy_read_timeout` applies not to the duration of a request, but to successive read operations from the proxied server :(

frankbu:
@youngnick @robscott @shaneutt

Based on feedback over the last few weeks, it seems that both new timeout fields can be supported by most implementations but neither can be supported by ALL implementations. Therefore, I think the prudent thing to do is to make both fields `Extended` support, at least for now. I would think that we can promote one or the other to `Core` in the future, if it turns out that one of them is implementable by ALL.

I also agree with @youngnick that we will soon want to follow up with a GEP for controlling the default values. If we do that with a policy attachment, I'm wondering if that would be a `DefaultTimoutsPolicy` or if default values could be a more generic feature, e.g., `DefaultValuesPolicy`, similar to the way Istio has a global mesh config resource that includes all kinds of default values. I can open a new issue to begin the discussion, if you'd like.

I believe I have also addressed all of the other issues that have been brought forward over the course of the last several weeks and given that timeouts are now totally optional extended support, I've change the GEP from `Provisional` to `Implementable`.

Can we go ahead an merge this now? If not, please let me know what else is needed from me. Thanks.

shaneutt (dismissed):

geps/gep-1742.md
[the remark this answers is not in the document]
frankbu: @shaneutt, it's been a couple of weeks now, and we've changed it to be Extended support, so I'm suggesting that we can make it Implementable now, so that we can also merge the related interfaces PR (#2013).

See https://github.com/kubernetes-sigs/gateway-api/pull/1997#issuecomment-1578812611

Let me know if you think it's still too soon, and I'll switch it back to Provisional.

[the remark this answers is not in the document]
shaneutt: Ok I think I just missed the comment. I can be convinced that we should move forward with `Implementable` at this point, I just wanna make sure nobody else has any confusion about that.

/cc @robscott @youngnick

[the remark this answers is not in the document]
robscott: I think going direct to implementable makes sense here, but you'll also need to make a corresponding change to update the status in the navigation as well:

https://github.com/kubernetes-sigs/gateway-api/blob/8afce4ea7223d85aef23b93e212a115b41d06a72/mkdocs.yml#L82

mkdocs.yml
```diff
@@ -79,12 +79,13 @@ nav:
...
       - Provisional:
-        - geps/gep-1742.md
-        - geps/gep-1426.md
         - geps/gep-1324.md
         - geps/gep-1282.md
       - Prototyping:
         - geps/gep-1709.md
+      - Implementable:
+        - geps/gep-1742.md
+        - geps/gep-1426.md
```
youngnick, line 88: Nice catch that we forgot to move this one as well previously, thanks.

youngnick (commented):
This LGTM overall, but one thing I think we need to start including for Extended features is what we're going to call the feature - in this case we're adding two additional Extended features, so we should note what the feature will be called for conformance test and conformance profile purposes.

I suggest `HTTPRouteRequestTimeout` and `HTTPRouteBackendTimeout`.

frankbu:
@youngnick

> I suggest `HTTPRouteRequestTimeout` and `HTTPRouteBackendTimeout`.

SGTM. Should that be included in this document? If so where?

Maybe it should be included in the comments of the Go APIs? https://github.com/kubernetes-sigs/gateway-api/pull/2013

youngnick:
I think we should add a new standard section to GEPs - "Conformance Details", that's a subsection of the "API" section, that lists the feature name(s) so that they're agreed on as part of this, and then can be used in standard ways as part of the conformance tests and profiles (this is _particularly_ important for Extended features.)

@shaneutt @mlavacca and anyone else who's working on the conformance profiles stuff.

frankbu:
@youngnick @shaneutt @mlavacca and others - Added Nick's suggested "Conformance Details" section. PTAL.

geps/gep-1742.md
```diff
@@ -298,16 +298,190 @@ Could not find any HTTP specific timeouts. PRs welcomed. 😊
...
+      backendRequest: 2s
+```
+
+### Conformance Details
+
+Gateway implementations can indicate support for the optional behavior in this GEP using
+the following feature names:
+
+- `HTTPRouteRequestTimeout`: supports `rules.timeouts.request` in an `HTTPRoute`.
+- `HTTPRouteBackendTimeout`: supports `rules.timeouts.backendRequest` in an `HTTPRoute`.
```
spacewander, line 448: Should we call it `HTTPRouteBackendRequestTimeout`? What if we add other backend timeouts in the future?

frankbu: I don't think it's necessary if we think of `request` as the "default" backend timeout. If we add other ones in the future, e.g., `rules.timeouts.backendIdle`, they need to include the kind explicitly: `HTTPRouteBackendIdelTimeout`.

spacewander: Just want to let the feature name matches the field name. But it's fine to keep the current name.

youngnick:
Thanks for the work @frankbu, this LGTM. I'll leave someone else to unhold though.

/lgtm

robscott:
Thanks @frankbu!

/hold cancel

PR #1991, opened 2023-05-03
mesh: "build" test image in tree

howardjohn:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
This isn't complete, but ensures we have ownership and versioning of the testing image.

**Which issue(s) this PR fixes**:

Fixes #

For https://github.com/kubernetes-sigs/gateway-api/issues/1895 (does not close)

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1999, opened 2023-05-05
Updating mesh conformance images to point to k8s registry

### LINKED ISSUE #1895: Move mesh test app into k8s org
I'd rather have this image owned by k8s/gateway-api infra so we can bundle versions of it with each release of our conformance tests. Can you create an issue to track following up on that? Shorter term can we at least use a pinned version of this image so a change in this image doesn't have an immediate impact on our conformance tests?

_Originally posted by @robscott in https://github.com/kubernetes-sigs/gateway-api/pull/1878#discussion_r1154742350_

robscott:
**What type of PR is this?**
/kind cleanup
/area conformance

This is the logical next step now that #1991 has merged. I considered just pointing to `:latest` but that would require for us to flip versions with every release to ensure conformance tests at a given release were consistent. Instead pinning to a version seems fine and matches what we're already doing for the other image we're using for conformance tests.

Fixes #1895

```release-note
NONE
```

### DIFF
--- a/conformance/mesh/manifests.yaml
+++ b/conformance/mesh/manifests.yaml
@@ -27,7 +27,7 @@ spec:
     spec:
       containers:
       - name: echo
-        image: gcr.io/istio-release/app:1.17.1
+        image: gcr.io/k8s-staging-gateway-api/echo-server:v20230505-v0.7.0-rc1-10-g497e67da
         imagePullPolicy: IfNotPresent
         args:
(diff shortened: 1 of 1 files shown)


Xunzhuo (approved):
LGTM

howardjohn (approved):

shaneutt (approved):
/lgtm

PR #2007, opened 2023-05-09
Conformance: Removing status-only redirect test

### LINKED ISSUE #1928: Conformance test HTTPRouteRedirectHostAndStatus has an unusual test case
**What happened**:

The test `HTTPRouteRedirectHostAndStatus` configures a redirect filter with only the [status code](https://github.com/kubernetes-sigs/gateway-api/blob/c27cdfc7c63f6f6f80e51a520b6bade7cf813acb/conformance/tests/httproute-redirect-host-and-status.yaml#L25).  According to the [spec](https://github.com/kubernetes-sigs/gateway-api/blob/c27cdfc7c63f6f6f80e51a520b6bade7cf813acb/apis/v1beta1/httproute_types.go#L852), any fields omitted will be identical to the request, so this is instructing the controller to configure the LB to return 301 to the requested URL.  Most browsers / clients will eventually error out when given such a response, and this is typically something that should be avoided.

It's possible people might want to do this for unconventional reasons, but it seems like such a configuration would be mostly broken, so I'd expect the controller to return an error in the event redirect is configured to the request URL.

robscott:
**What type of PR is this?**
/kind cleanup
/area conformance

This was not really a practical form of redirect as it would just result in a redirect loop or similar.

Fixes #1928

```release-note
Conformance: Remove a test that only covered redirect status without any other changes. 
```

### DIFF
--- a/conformance/tests/httproute-redirect-host-and-status.go
+++ b/conformance/tests/httproute-redirect-host-and-status.go
@@ -60,15 +60,6 @@ var HTTPRouteRedirectHostAndStatus = suite.ConformanceTest{
 			Namespace: ns,
 		}, {
 			Request: http.Request{
-				Path:             "/status-code-301",
-				UnfollowRedirect: true,
-			},
-			Response: http.Response{
-				StatusCode: 301,
-			},
-			Namespace: ns,
-		}, {
-			Request: http.Request{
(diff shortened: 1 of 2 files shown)


shaneutt (approved):
/lgtm

PR #2013, opened 2023-05-12
HTTPRoute timeout

SRodi:
**What type of PR is this?**

[HTTPRoute Timeouts](https://github.com/kubernetes-sigs/gateway-api/pull/2013/commits/e32217ea84af9d69817f835887949f005a9e1e2c) 

* update HTTPRoute interfaces according to latest proposal in GEP-1742
* add Duration type as per GEP-2257
* include Request and BackendRequest in HTTPRouteTimeouts
* add validation for HTTPRouteTimeouts
* add unit tests in httproute_test.go
* relates to https://github.com/kubernetes-sigs/gateway-api/pull/1997

Relates to: 

- https://gateway-api.sigs.k8s.io/geps/gep-1742/
- https://gateway-api.sigs.k8s.io/geps/gep-2257/
- https://github.com/kubernetes-sigs/gateway-api/issues/1742
- https://github.com/kubernetes-sigs/gateway-api/pull/1997

POC for HTTPRout timeout proposal 

Relates to https://github.com/kubernetes-sigs/gateway-api/issues/1742

```release-note
New fields have been added to configure timeouts on `HTTPRoute` rules:

- `rules.timeouts.request`: to specify a timeout for responding to client HTTP requests
- `rules.timeouts.backendRequest`: to specify a timeout for individual requests from the gateway to a backend service

See [GEP-1472](https://gateway-api.sigs.k8s.io/geps/gep-1742/) for more information.
```


config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
```diff
@@ -1692,6 +1692,21 @@ spec:
...
+                        request:
+                          description: "Request specifies an end-to-end timeout for
+                            client HTTP requests, disabled by default. \n Request
+                            timeout values are formatted as 1h/1m/1s/1ms and MUST
+                            BE >= 1ms. \n For example, the following rule will wait
+                            no longer than 2 secs for the client request to complete:
+                            \n ``` rules: - timeouts: request: 2s backendRefs: ...
+                            ``` \n Support: Core"
+                          format: int64
+                          type: integer
```
frankbu, line 1708: Not sure how to make it happen, but I think we want something like this?
```suggestion
                          format: duration
                          type: string
```

sunjayBhatia: in Contour we did the following: https://github.com/projectcontour/contour/blob/023f6075399213ef7a2c54f6b2853e07e7713c8c/apis/projectcontour/v1/httpproxy.go#L217-L218

the type is string, and we used a kubebuilder validation regex to ensure the right format is passed in

(`time.Duration` devolves down to an int64, the nice `1s` is just string representation)

frankbu: Thanks @sunjayBhatia, that's Interesting. I wonder how the Contour `// +kubebuilder:validation:Pattern=regex` compares to `// +kubebuilder:validation:Format=duration`? I tried to create a field with format set to duration and it seemed to be enforcing time.Duration string format, i.e., it accepted "2s", but didn't accept "hello". I'm not sure where these format values are documented though.

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +246,37 @@ type HTTPRouteRule struct {
...
+	//	- timeouts:
+	//	    request: 2s
+	//	  backendRefs:
+	//	    ...
+	// ```
+	//
+	// Support: Core
+	//
+	// +optional
+	Request *time.Duration `json:"request,omitempty"`
```
frankbu, line 279: Maybe something like this would work? I wonder if we would also need to override marshal/unmarshal if we want the API to return *time.Duration like this?

```suggestion
	// +kubebuilder:validation:Type=string
	// +kubebuilder:validation:Format=duration
	Request *time.Duration `json:"request,omitempty"`
```

config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
SRodi: @sunjayBhatia thanks for sharing the above. When defining the type as follows:

```go
...
	// +kubebuilder:validation:Type=string
	// +kubebuilder:validation:Format=duration
	Request *time.Duration `json:"request,omitempty"`
...
```

The generated CRD looks like this:

```yaml
...
                    timeouts:
                      description: "Timeouts defines the timeouts that can be configured
                        for an HTTP request. \n Support: Core \n <gateway:experimental>"
                      properties:
                        request:
                          allOf:
                          - format: int64
                          - format: duration
                          description: "Request specifies an end-to-end timeout for
                            client HTTP requests, disabled by default. \n Request
                            timeout values are formatted as 1h/1m/1s/1ms and MUST
                            BE >= 1ms. \n For example, the following rule will wait
                            no longer than 2 secs for the client request to complete:
                            \n ``` rules: - timeouts: request: 2s backendRefs: ...
                            ``` \n Support: Core"
                          type: integer
                      type: object
...
```

The generated `request.allOf` and `request.type: integer` are not what we need. Instead, we need `request.type: string` and `request.format: duration`. I am not familiar with kubebuilder markers and I could not find a clear example anywhere in the official docs https://book.kubebuilder.io/reference/markers/crd-validation.html 

Would you have any idea what I am doing wrong?

sunjayBhatia: you might want to try to use the `metav1.Duration` wrapper type rather than `time.Duration` https://github.com/kubernetes/apimachinery/blob/0ef2ec80f7a200acb85cc11636ab3a14a765b6f8/pkg/apis/meta/v1/duration.go#L24-L29

i think you should be able to remove the `validation:Type=string` line as well

SRodi: this worked as expected, thanks @sunjayBhatia

apis/v1beta1/httproute_types.go
SRodi: if we use `metav1.Duration` it seems marshal/unmarshal is already taken care of according to https://github.com/kubernetes/apimachinery/blob/0ef2ec80f7a200acb85cc11636ab3a14a765b6f8/pkg/apis/meta/v1/duration.go

frankbu:
@SRodi Can you also add a check for timeouts in the validator, e.g., check the value is >= 1ms?
https://github.com/kubernetes-sigs/gateway-api/blob/7b8f168c82a90522209a340bf291fcacab0fc5ef/apis/v1beta1/validation/httproute.go#L54

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,46 @@ type HTTPRouteRule struct {
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
 	BackendRefs []HTTPBackendRef `json:"backendRefs,omitempty"`
+	// Timeouts defines the timeouts that can be configured for an HTTP request.
```
frankbu, line 247:
```suggestion

	// Timeouts defines the timeouts that can be configured for an HTTP request.
```

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,46 @@ type HTTPRouteRule struct {
...
+	// - timeouts:
+	//     request: 10s
+	//   backendRefs:
+	//     ...
+	// ```
+	//
+	// Support: Core
+	//
+	// +optional
+	Request *metav1.Duration `json:"request,omitempty"`
```
frankbu, line 277:
```suggestion
	// +kubebuilder:validation:Format=duration
	Request *metav1.Duration `json:"request,omitempty"`
```

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,46 @@ type HTTPRouteRule struct {
...
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service.
+	// Default is the value of Request timeout.
+	//
+	// Support: Extended
+	//
+	// +optional
+	BackendRequest *metav1.Duration `json:"backendRequest,omitempty"`
```
frankbu, line 286:
```suggestion
	// +kubebuilder:validation:Format=duration
	BackendRequest *metav1.Duration `json:"backendRequest,omitempty"`
```

apis/v1beta1/httproute_types.go
SRodi: addressed with 2fad26d06fd5ee37486fca3a12e690fecaed3793

apis/v1beta1/httproute_types.go
SRodi: addressed with 2fad26d06fd5ee37486fca3a12e690fecaed3793

apis/v1beta1/httproute_types.go
SRodi: addressed with 2fad26d06fd5ee37486fca3a12e690fecaed3793

apis/v1beta1/validation/httproute.go
```diff
@@ -40,6 +40,9 @@ var (
 
 	// All valid path characters per RFC-3986
 	validPathCharacters = "^(?:[A-Za-z0-9\\/\\-._~!$&'()*+,;=:@]|[%][0-9a-fA-F]{2})+$"
+
+	// Minimum value of timeout in milliseconds
+	TimeoutMinValue = 1
```
frankbu, line 45: Should this be lower case?
```suggestion
	timeoutMinValue = 1
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,22 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", TimeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		if int(timeouts.BackendRequest.Duration.Milliseconds()) >= TimeoutMinValue {
```
frankbu, line 348:
```suggestion
		if int(timeouts.BackendRequest.Duration.Milliseconds()) < TimeoutMinValue {
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,22 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
+			errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, errorString))
+		}
+	}
+	if timeouts.Request != nil {
+		if int(timeouts.Request.Duration.Milliseconds()) >= TimeoutMinValue {
```
frankbu, line 353:
```suggestion
		if int(timeouts.Request.Duration.Milliseconds()) < TimeoutMinValue {
```

apis/v1beta1/validation/httproute.go
SRodi: addressed in 05f32daa4ee6b188d6f8d07c3777d8ececdb786d

apis/v1beta1/validation/httproute.go
SRodi: addressed in 05f32daa4ee6b188d6f8d07c3777d8ececdb786d

apis/v1beta1/validation/httproute.go
SRodi: addressed in 05f32daa4ee6b188d6f8d07c3777d8ececdb786d

frankbu (approved):
@SRodi You'll need to add a release note for this change (https://github.com/kubernetes-sigs/gateway-api/pull/2013#issuecomment-1554135756) but I think you should wait for #1997 to merge first and then you can reference the GEP in your release note.

Otherewise LGTM. Thanks.

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,28 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
```
frankbu, line 344: This still needs a unit test in httproute_test.go?

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,49 @@ type HTTPRouteRule struct {
...
+	// ```
+	//
+	// Support: Core
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service.
```
frankbu, line 282:
```suggestion
	// to a backend service. Typically used in conjunction with automatic retries,
        // if supported by an implementation.
```

SRodi: addressed in 0fb746498cddabddee5c70bb6d377ac163f10999

apis/v1beta1/validation/httproute.go
SRodi: addressed in 0fb746498cddabddee5c70bb6d377ac163f10999

config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
```diff
@@ -1692,6 +1692,26 @@ spec:
...
                       maxItems: 8
                       type: array
+                    timeouts:
+                      description: "Timeouts defines the timeouts that can be configured
+                        for an HTTP request. \n Support: Core \n <gateway:experimental>"
+                      properties:
+                        backendRequest:
+                          description: "BackendRequest specifies a timeout for an
+                            individual request from the gateway to a backend service.
+                            Default is the value of Request timeout. \n Support: Extended"
```
frankbu, line 1702: This needs another regen because not up to date with the comment in the go code?

SRodi: addressed in 2678d3cf1fa6774a4a405583e1d9e528f0a66209

robscott:
Thanks for the work on this @SRodi! I'll try to get back to review this, but adding a hold in the meantime because the corresponding GEP needs to both be approved and graduate to "implementable" before this can go in. More info here: https://gateway-api.sigs.k8s.io/geps/overview/#process.

/hold

frankbu (changes requested):
@SRodi Please replace the comments/doc with the latest updated interfaces in https://github.com/kubernetes-sigs/gateway-api/pull/1997. Also note that a timeout value of 0 is now valid, so you will need to update the validator to allow it.

apis/v1beta1/validation/httproute.go
```diff
@@ -40,6 +40,9 @@ var (
 
 	// All valid path characters per RFC-3986
 	validPathCharacters = "^(?:[A-Za-z0-9\\/\\-._~!$&'()*+,;=:@]|[%][0-9a-fA-F]{2})+$"
+
+	// Minimum value of timeout in milliseconds
+	timeoutMinValue = 0
```
frankbu, line 45: min timeout is 1 ms
```suggestion
	timeoutMinValue = 1
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,28 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", timeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		if int(timeouts.BackendRequest.Duration.Milliseconds()) < timeoutMinValue {
```
frankbu, line 348:
```suggestion
		backendTimeout := int(timeouts.BackendRequest.Duration.Milliseconds())
		if backendTimeout < timeoutMinValue && backendTimeout != 0 {
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,28 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
+			errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, errorString))
+		}
+		if timeouts.Request != nil {
+			if int(timeouts.BackendRequest.Duration.Milliseconds()) > int(timeouts.Request.Duration.Milliseconds()) {
+				errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, "backendRequest timeout cannot be longer than request timeout"))
+			}
+		}
+	}
+	if timeouts.Request != nil {
+		if int(timeouts.Request.Duration.Milliseconds()) < timeoutMinValue {
```
frankbu, line 358:
```suggestion
		timeout := int(timeouts.Request.Duration.Milliseconds())
		if timeout < timeoutMinValue && timeout != 0 {
```

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1140,3 +1142,45 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
...
+			Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+				Request: &metav1.Duration{
+					Duration: time.Duration(time.Millisecond * 1),
+				},
+				BackendRequest: &metav1.Duration{
+					Duration: time.Duration(time.Millisecond * 2),
+				},
+			},
+		},
+		}},
```
frankbu, line 1174: Maybe add a couple more tests?
1. 0 -> valid
2. 0.5 ms -> invalid

apis/v1beta1/validation/httproute.go
SRodi: addressed in 4056411a2ddccafe6f9eb40bb9b14fd685e1859e

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,28 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", timeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		if int(timeouts.BackendRequest.Duration.Milliseconds()) < timeoutMinValue {
+			errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, errorString))
+		}
+		if timeouts.Request != nil {
+			if int(timeouts.BackendRequest.Duration.Milliseconds()) > int(timeouts.Request.Duration.Milliseconds()) {
```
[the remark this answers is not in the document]
SRodi: addressed in 4056411a2ddccafe6f9eb40bb9b14fd685e1859e

apis/v1beta1/validation/httproute.go
SRodi: addressed in 4056411a2ddccafe6f9eb40bb9b14fd685e1859e

apis/v1beta1/validation/httproute_test.go
SRodi: addressed in 4056411a2ddccafe6f9eb40bb9b14fd685e1859e

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +341,30 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", timeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		backendTimeout := int(timeouts.BackendRequest.Duration.Milliseconds())
+		if backendTimeout < timeoutMinValue && backendTimeout != 0 {
+			errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, errorString))
+		}
+		if timeouts.Request != nil {
+			if backendTimeout > int(timeouts.Request.Duration.Milliseconds()) {
```
frankbu, line 353: Sorry, one more I missed
```suggestion
			timeout := int(timeouts.Request.Duration.Milliseconds())
			if backendTimeout > timeout && timeout != 0 {
```

SRodi: addressed in 2a3078bc8351b2d4b990811700971d2066694c62

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1140,3 +1142,65 @@ func TestValidateHTTPRouteTypeMatchesField(t *testing.T) {
...
+				},
+			},
+		},
+		}}, {
+		name:     "valid httpRoute Rules timeout set to 0 (disabled)",
+		errCount: 0,
+		rules: []gatewayv1b1.HTTPRouteRule{{
+			Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+				Request: &metav1.Duration{
+					Duration: time.Duration(time.Millisecond * 0),
```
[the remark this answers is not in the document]
SRodi: addressed in 2a3078bc8351b2d4b990811700971d2066694c62

apis/v1beta1/validation/httproute.go
```diff
@@ -40,6 +41,9 @@ var (
 
 	// All valid path characters per RFC-3986
 	validPathCharacters = "^(?:[A-Za-z0-9\\/\\-._~!$&'()*+,;=:@]|[%][0-9a-fA-F]{2})+$"
+
+	// Minimum value of timeout in milliseconds
```
frankbu, line 45:
```suggestion
	// Minimum value of timeout in microseconds
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +342,31 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", timeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		backendTimeout := timeouts.BackendRequest.Duration
+		if backendTimeout < timeoutMinValue && backendTimeout != 0 {
+			errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, errorString))
```
frankbu, line 351:
```suggestion
			errs = append(errs, field.Invalid(path.Child("backendRequest"), backendTimeout, errorString))
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +342,31 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
+		}
+		if timeouts.Request != nil {
+			timeout := timeouts.Request.Duration
+			if backendTimeout > timeout && timeout != 0 {
+				errs = append(errs, field.Invalid(path.Child("backendRequest"), timeouts.BackendRequest.Duration, "backendRequest timeout cannot be longer than request timeout"))
```
frankbu, line 356:
```suggestion
				errs = append(errs, field.Invalid(path.Child("backendRequest"), backendTimeout, "backendRequest timeout cannot be longer than request timeout"))
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +342,31 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
+			}
+		}
+	}
+	if timeouts.Request != nil {
+		timeout := timeouts.Request.Duration
+		if timeout < timeoutMinValue && timeout != 0 {
+			errs = append(errs, field.Invalid(path.Child("request"), timeouts.Request.Duration, errorString))
```
frankbu, line 363:
```suggestion
			errs = append(errs, field.Invalid(path.Child("request"), timeout, errorString))
```

apis/v1beta1/validation/httproute.go
SRodi: why microseconds @frankbu ?

frankbu: Golang Duration values are in microseconds. Notice that the value `of int(time.Millisecond)` is `1000000`

frankbu: Oh, wait, it's nanoseconds, actually.
```suggestion
	// Minimum value of timeout in nanoseconds
```

frankbu: I suppose you could avoid the detail by just saying:
```suggestion
	// Minimum value of timeout fields
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +342,31 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %d ms", timeoutMinValue)
```
frankbu, line 347:
```suggestion
	errorString := fmt.Sprintf("timeout value must be greater than or equal to %s", timeoutMinValue)
```

frankbu (approved):
Thanks @SRodi!

robscott:
@SRodi Thanks for working on this! Now that the GEP has merged I think you can remove "POC" from the title of this PR.

/ok-to-test

robscott:
Hey @SRodi and @frankbu thanks for getting this GEP so far along! In general we try to add at least some basic conformance test coverage along with a new API, do you mind adding a follow up PR with conformance tests? If it's possible to get those lined up this week, I think this might be able to make it into v0.8.0.

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,56 @@ type HTTPRouteRule struct {
...
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
+	Request *metav1.Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service. This covers the time from when the request first starts being
+	// sent from the gateway to when the full response has been received from the backend.
```
robscott, line 283: In general prefer to capitalize the names of Kubernetes resources - same comment applies elsewhere.
```suggestion
	// sent from the Gateway to when the full response has been received from the backend.
```

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,56 @@ type HTTPRouteRule struct {
...
+	// may result in more than one call from the gateway to the destination backend service,
+	// for example, if automatic retries are supported.
+	//
+	// Because the Request timeout encompasses the BackendRequest timeout,
+	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:validation:Format=duration
```
robscott, line 295: I'm pretty sure this won't work, but if it did, it would mean that we could avoid the webhook validation entirely.

```suggestion
	// +kubebuilder:validation:Format=duration
	// +kubebuilder:validation:Maximum=1ms
	// +kubebuilder:validation:Maximum=100y
```

apis/v1beta1/validation/httproute.go
```diff
@@ -334,6 +342,31 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
 	return errs
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	errorString := fmt.Sprintf("timeout value must be greater than or equal to %s", timeoutMinValue)
+	if timeouts.BackendRequest != nil {
+		backendTimeout := timeouts.BackendRequest.Duration
+		if backendTimeout < timeoutMinValue && backendTimeout != 0 {
```
robscott, line 350: As mentioned above, I think we should have a min and max value here and that 0 should be invalid.

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,56 @@ type HTTPRouteRule struct {
...
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms or 0 to disable (no timeout).
+type HTTPRouteTimeouts struct {
+	// Request specifies the duration for processing an HTTP client request after which the
+	// gateway will time out if unable to send a response.
```
robscott, line 262: Maybe something like this?
```suggestion
	// Request specifies the maximum duration for a Gateway to respond to a HTTP request. If the Gateway
	// has not been able to respond before this deadline is met, the Gateway MUST return a 504 timeout error.
```

robscott (commented):
Thanks @SRodi!

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,56 @@ type HTTPRouteRule struct {
...
+	// Support: Extended
+	//
+	// +optional
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms or 0 to disable (no timeout).
```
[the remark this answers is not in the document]
frankbu: @robscott

I think the upper bound is limited by the internal representation of `time.Duration` which is `int64`. So that's ~300 years? So the limit is controlled by `time.ParseDuration`. Not sure if it truncates or fails to parse string values that won't fit? But I was assuming that documenting this value is according to `time.ParseDuration` is all we really need to say. I suppose we could make up some smaller value and add validation for that, but I'm not sure that's worth it.

Also, if IIUC, the Rust implementation of duration is represented by an unsigned int, so the max value might be higher. Bottom line is I don't think this value really needs to be documented in the spec because the bottom line is that the max is huge, and nobody will ever use such huge values anyway.

The only exception is "no timeout", hang forever until the client kills the connection. Envoy does in fact use 0 for "no timeout". This was something I added to address reviewer feedback of the GEP from @candita.

apis/v1beta1/httproute_types.go
frankbu: Good idea to make the return code explicit because not everybody knows that "will time out" in HTTP really means return 504.

apis/v1beta1/httproute_types.go
frankbu: I agree with this, but only when we are referring to the configuration resource (Gateway), vs just talking about the proxy server implementing it. If we're talking about the runtime server, then I's better to use lower case (gateway) for that IMO.

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,56 @@ type HTTPRouteRule struct {
...
+	// BackendRequest specifies a timeout for an individual request from the gateway
+	// to a backend service. This covers the time from when the request first starts being
+	// sent from the gateway to when the full response has been received from the backend.
+	//
+	// An entire client HTTP transaction with a gateway, covered by the Request timeout,
+	// may result in more than one call from the gateway to the destination backend service,
+	// for example, if automatic retries are supported.
+	//
+	// Because the Request timeout encompasses the BackendRequest timeout,
+	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
```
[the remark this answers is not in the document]
frankbu: If we remove the "defaults to and" from here, I think we need to add another sentence to say the default value is implementation dependent.

That said, I thought that the result of the GEP review (where this comes from) was that defaulting to the value of the other field was sensible. Why do you think it's a problem? All this means is that if the `BackendRequest` field isn't set, then it default the the value that is set in the other fields (`Request`). The default value of `Request` in turn is implementation dependent (according to it's doc).

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
frankbu: Repeating for each field isn't very nice IMO, but I don't have a strong opinion on this so either way is fine with me.

apis/v1beta1/httproute_types.go
frankbu: Interesting. @SRodi can you try to see if this works?

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: I created #2155 to follow up on this - at a high level, I think we really need to start with the tightest possible validation here and only loosen if/when it's needed. Having clearly defined limits enables implementations to test against min and max values to ensure that they can support the full range of possible inputs from the API.

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: Unfortunately we don't have a means to implement variable defaulting unless we want to create and maintain a defaulting webhook, and I'd strongly prefer to avoid that maintenance cost. 

We _could_ suggest that if a value is not set here, implementations should consider this the same value as the overall request timeout, but I can't think of any way to validate that, and I'm also not sure that's what we want. Let's say that an implementation has a default of 2 retries with a 10s per-try timeout. If a user configures an overall request timeout of 20s, I think they'd probably prefer the implementation default behavior here instead of that 1 value actually affecting the per-try timeout value as well. In general I'd like to suggest working towards the principle of least surprise, and I think that means avoiding a variable default like this.

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
adleong: For reference, Linkerd also interprets 0 as "no timeout": https://github.com/linkerd/linkerd2/blob/main/policy-controller/k8s/index/src/outbound/index.rs#L341

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
adleong: The way that I think about these timeouts composing means that I think BackendRequest can have a default of "no timeout".

If there is e.g. a 20s overall request timeout, I would expect the entire request to timeout after 20s even if no BackendRequest timeout is set.

adleong (approved):

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: That's a great way of looking at this @adleong. I think we're all agreeing that there's no way for the value here to be greater than the overall request timeout value, even if it's not explicitly set. Maybe we should just focus on clearly stating that? That may have actually been what was originally intended here and I misunderstood it.

apis/v1beta1/httproute_types.go
spacewander: It would be better to use 408 (Request Timeout) for Request and 504 (Gateway Timeout) for BackendRequest?

spacewander: The server may also close the connection without sending 408 or any message, as the client may not response to the server anymore when the Request timeout hits.

frankbu:
> Hey @SRodi and @frankbu thanks for getting this GEP so far along! In general we try to add at least some basic conformance test coverage along with a new API, do you mind adding a follow up PR with conformance tests?

@robscott Yes, we can look into adding conformance test for timeouts, but not this week. I think it will take a little longer to figure out how to set up a slow service in the conformance framework. I've been thinking we could add the [httpbin service](https://github.com/istio/istio/tree/master/samples/httpbin) from Istio for this (and other HTTPRoute conformance tests in the future) where we could call `/delay` from a client with a timeout configured. I'm going to be on vacation next week, but if @SRodi or someone else can pick it up before I'm back, that would be great.

apis/v1beta1/httproute_types.go
frankbu: Hmm, maybe it is best to just keep it vague, like I had it, and just say it "will time out", and leave the exact return code up to the implementation?

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
frankbu: The intent was as @adleong said, the request timeout applies if no backend timeout is set. I realize now that there is a subtle difference on how it's worded. To illustrate say we have a 10s Request timeout and 3 retries configured and we call a backend that is hanging. With the current wording (BackendRequest defaults to the value of Request) there will be 1 call to the backend which will timeout in 10s (no retries). If we change the wording to say BackendDefault is implementation dependent, then it's possible that an implementation has a shorter default backend timeout (say 2s) so all three retries will happen before the request fails in 8s. So if we remove the "defaults to Request" part, the default behavior when Request is set but BackendRequest is not, will be undefined. It's not really a big deal IMO, so I'm fine with it either way. Btw, Istio will work the BackendRequest default to Request way regardless.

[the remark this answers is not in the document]
robscott: This became clearer to me in the other PR, but since implementations already have default behavior that is working with the API today, I don't think we can really add any kind of defaulting to these fields since it would result in backwards incompatible changes for implementations that already have defaults for per-try-timeout and/or retries.

So I think we should state that that this value can't be higher than request timeout since I don't think we can safely guarantee anything beyond that.

apis/v1beta1/httproute_types.go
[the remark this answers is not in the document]
robscott: Yeah it's not great but it does make it more clear when reading API spec for an individual field.

apis/v1beta1/httproute_types.go
robscott: I think in Gateway API we've merged the concepts together, but this is a good point. I don't feel very strongly about this one, feel free to disregard the suggestion.

apis/v1beta1/httproute_types.go
robscott: @spacewander has a good point that there are cases where it could be either 504 or 408. I've updated the guidance to simply require a "timeout error". 

My reading of the [relevant RFC](https://datatracker.ietf.org/doc/html/rfc7231#section-6.6.5) makes me think that 504 would be more common here:

> The 504 (Gateway Timeout) status code indicates that the server,
   while acting as a gateway or proxy, did not receive a timely response
   from an upstream server it needed to access in order to complete the
   request.

But it does seem possible that in some cases a [408](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.7) would be appropriate:

>The 408 (Request Timeout) status code indicates that the server did
   not receive a complete request message within the time that it was
   prepared to wait.  A server SHOULD send the "close" connection option
   ([Section 6.1 of [RFC7230]](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1)) in the response, since 408 implies that
   the server has decided to close the connection rather than continue
   waiting.  If the client has an outstanding request in transit, the
   client MAY repeat that request on a new connection.
   
So given the above, I think the most prescriptive we can be is that a timeout error should be returned by the Gateway.

spacewander: Yes, the `504 (Gateway Timeout)` section in the RFC emphasizes that this error code is used when `did not receive a timely response from an upstream server`. So it is more relative to the BackendRequest.

spacewander: Anyway, I also agree to use `timeout error` to avoid conflict with the existing implementations.

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,55 @@ type HTTPRouteRule struct {
...
+	//
+	// +optional
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
+// and MUST BE >= 1ms or 0 to disable (no timeout).
+type HTTPRouteTimeouts struct {
```
hzxuzhonghu, line 260:
```suggestion
type HTTPRouteTimeout struct {
```

apis/v1beta1/httproute_types.go
```diff
@@ -244,6 +244,55 @@ type HTTPRouteRule struct {
...
 	// +kubebuilder:validation:MaxItems=16
 	BackendRefs []HTTPBackendRef `json:"backendRefs,omitempty"`
+
+	// Timeouts defines the timeouts that can be configured for an HTTP request.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
```
hzxuzhonghu, line 254: why use plural here

frankbu: I see your point, but I think it's the right name unless it is against Kubernetes naming conventions? This is just implementing GEP-1742, so if you think it requires more discussion you should probably bring it up in https://github.com/kubernetes-sigs/gateway-api/pull/2155

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,55 @@ type HTTPRouteRule struct {
...
+	// request stream has been received instead of immediately after the transaction is
+	// initiated by the client.
+	//
+	// When this field is unspecified, request timeout behavior is implementation-dependent.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=^([0-9]{1,5}(h|m|s|ms)){1,4}$
+	Request *string `json:"request,omitempty"`
```
frankbu, line 297: This should use type Duration, which already includes the validation pattern.
```suggestion
	Request *Duration `json:"request,omitempty"`
```

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,55 @@ type HTTPRouteRule struct {
...
+	// for example, if automatic retries are supported.
+	//
+	// Because the Request timeout encompasses the BackendRequest timeout,
+	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
+	//
+	// Support: Extended
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=^([0-9]{1,5}(h|m|s|ms)){1,4}$
+	BackendRequest *string `json:"backendRequest,omitempty"`
```
frankbu, line 314:
```suggestion
	BackendRequest *Duration `json:"backendRequest,omitempty"`
```

apis/v1alpha2/shared_types.go
```diff
@@ -340,6 +340,11 @@ type AnnotationValue = v1beta1.AnnotationValue
 // +kubebuilder:validation:Pattern=`^Hostname|IPAddress|NamedAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type AddressType = v1beta1.AddressType
 
+// A duration is a ...
+//
+// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms))`
```
frankbu, line 345:
```suggestion
// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
```

apis/v1beta1/httproute_types.go
robscott: I think this might also be the source of your CRD generation problems - this and at least 1 other regex added in this PR are not surrounded by backticks which may be having some unintended consequences.

apis/v1beta1/validation/httproute.go
robscott: Resolving since we've allowed "0s" intentionally.

config/webhook/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+resources:
```
robscott, line 1: Assuming this was added for local dev purposes, please don't forget to remove it.

apis/v1alpha2/shared_types.go
```diff
@@ -340,6 +340,11 @@ type AnnotationValue = v1beta1.AnnotationValue
...
+type Duration string
```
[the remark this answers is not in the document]
SRodi: I have reverted this and added `Duration type` only in `v1beta1` since this is available only in this api. @robscott please correct me if I am wrong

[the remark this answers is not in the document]
frankbu: Do we also need to define it in `v1alpha2` as a clone of the `v1beta1` type, as is done for other shared types? For example https://github.com/kubernetes-sigs/gateway-api/blob/e32217ea84af9d69817f835887949f005a9e1e2c/apis/v1alpha2/shared_types.go#L176

^^^^ @robscott

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1176,3 +1176,67 @@ func TestValidateRequestRedirectFiltersWithNoBackendRef(t *testing.T) {
...
+					},
+				},
+			},
+		}, {
+			name:     "valid httpRoute Rules timeout set to 0 (disabled)",
+			errCount: 0,
+			rules: []gatewayv1b1.HTTPRouteRule{
+				{
+					Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+						Request: toDuration("0ms"),
```
frankbu, line 1206: Although "0ms" also works, the spec recommends using "0s" for 0 (infinite) timeout, so best to follow best practices here?
```suggestion
						Request: toDuration("0s"),
```

apis/v1beta1/shared_types.go
```diff
@@ -610,6 +610,11 @@ type AddressType string
 // +k8s:deepcopy-gen=false
 type HeaderName string
 
+// A duration is a ...
+//
+// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
+type Duration string
```
frankbu, line 616: Do we also need to define this in `v1alpha2` as a clone of the `v1beta1` type, as is done for other shared types? For example https://github.com/kubernetes-sigs/gateway-api/blob/e32217ea84af9d69817f835887949f005a9e1e2c/apis/v1alpha2/shared_types.go#L176

^^^^ @robscott, please advise

SRodi: addressed in 6281ef19931f78d873d920088ba4abb36055f346

apis/v1beta1/validation/httproute_test.go
SRodi: addressed in 6281ef19931f78d873d920088ba4abb36055f346

apis/v1beta1/httproute_types.go
SRodi: can we resolve this @hzxuzhonghu @frankbu or does it need further discussion ?

apis/v1alpha2/shared_types.go
```diff
@@ -340,6 +340,12 @@ type AnnotationValue = v1beta1.AnnotationValue
 // +kubebuilder:validation:Pattern=`^Hostname|IPAddress|NamedAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type AddressType = v1beta1.AddressType
 
+// Duration is a string value representing a duration in time. The format is as specified
+// in GEP-2257, a strict subset of the syntax parsed by Golang time.ParseDuration.
+//
+// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
+type Duration v1beta1.Duration
```
frankbu, line 347: This should be a type alias: https://stackoverflow.com/questions/61247864/what-is-the-difference-between-type-alias-and-type-definition-in-go
```suggestion
type Duration = v1beta1.Duration
```

apis/v1beta1/httproute_types.go
frankbu: Yes, this can be resolved since this PR is just implementing GEP-1742. Any further discussion would need to be in an issue with GEP-1742.

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,58 @@ type HTTPRouteRule struct {
...
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms, as specified in GEP-2257,
+// and MUST BE >= 1ms or "0s" to disable (no timeout).
+//
+// +kubebuilder:validation:XValidation:message="request timeout value must be greater than or equal to 1ms",rule="!(has(self.request) && duration(self.request) != duration('0s') && duration(self.request) < duration('1ms'))"
+// +kubebuilder:validation:XValidation:message="backendRequest timeout value must be greater than or equal to 1ms",rule="!(has(self.backendRequest) && duration(self.backendRequest) != duration('0s') && duration(self.backendRequest) < duration('1ms'))"
```
frankbu, line 281: I just realized that since GEP-2257 doesn't allow fractional values or any unit smaller than ms, the minimum value allowed by the syntax is 1ms now, so I don't think we need these validation rules anymore.

```suggestion
```

SRodi: addressed in d8f2b2bd1792c458eeb93fffee6c08f0392c05f6

apis/v1alpha2/shared_types.go
SRodi: addressed in d8f2b2bd1792c458eeb93fffee6c08f0392c05f6

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1176,3 +1176,67 @@ func TestValidateRequestRedirectFiltersWithNoBackendRef(t *testing.T) {
 		})
 	}
 }
+
+func toDuration(durationString string) *gatewayv1b1.Duration {
+	return (*gatewayv1b1.Duration)(&durationString)
+}
+
+func TestValidateHTTPTimeouts(t *testing.T) {
```
frankbu, line 1184: Now that Gateway API supports CEL validation, you need to add a similar CEL validation test function (i.e., `func TestHTTPRouteTimeouts`) in here:  https://github.com/kubernetes-sigs/gateway-api/blob/main/pkg/test/cel/httproute_test.go

apis/v1alpha2/shared_types.go
```diff
@@ -340,6 +340,12 @@ type AnnotationValue = v1beta1.AnnotationValue
 // +kubebuilder:validation:Pattern=`^Hostname|IPAddress|NamedAddress|[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*\/[A-Za-z0-9\/\-._~%!$&'()*+,;=:]+$`
 type AddressType = v1beta1.AddressType
 
+// Duration is a string value representing a duration in time. The format is as specified
+// in GEP-2257, a strict subset of the syntax parsed by Golang time.ParseDuration.
+//
+// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms)){1,4}$`
```
robscott, line 346: I don't think you need to repeat validation here, would be worth double checking if removing this has any effect on the generated CRDs.

apis/v1beta1/httproute_types.go
robscott: @SRodi I think the conclusion here was that my suggestion to start this thread was about as specific as we can get here, do you mind updating the spec to reflect that suggestion?

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+	// For example, setting the `rules.timeouts.request` field to the value `10s` in an
+	// `HTTPRoute` will cause a timeout if a client request is taking longer than 10 seconds
+	// to complete.
+	//
+	// This timeout is intended to cover as close to the whole request-response transaction
+	// as possible although an implementation MAY choose to start the timeout after the entire
+	// request stream has been received instead of immediately after the transaction is
+	// initiated by the client.
+	//
+	// When this field is unspecified, request timeout behavior is implementation-dependent.
```
robscott, line 294: Tiny nit: Just to match our wording elsewhere in the spec.
```suggestion
	// When this field is unspecified, request timeout behavior is implementation-specific.
```

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+	//
+	// Support: Extended
+	//
+	// +optional
+	Request *Duration `json:"request,omitempty"`
+
+	// BackendRequest specifies a timeout for an individual request from the gateway
```
robscott, line 301: Nit: Recommend capitalizing resource names throughout the spec, will not add duplicate comments everywhere though.
```suggestion
	// BackendRequest specifies a timeout for an individual request from the Gateway
```

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+	// to a backend service. This covers the time from when the request first starts being
```
robscott, line 302: Nit: I'd replace all references to "backend service" with just "backend" since we're usually just forwarding to backend of a Service, not a Service directly, and of course we also support different backend types.
```suggestion
	// to a backend. This covers the time from when the request first starts being
```

apis/v1beta1/validation/httproute.go
```diff
@@ -41,6 +42,9 @@ var (
 
 	// All valid path characters per RFC-3986
 	validPathCharacters = "^(?:[A-Za-z0-9\\/\\-._~!$&'()*+,;=:@]|[%][0-9a-fA-F]{2})+$"
+
+	// Minimum value of timeout fields
+	timeoutMinValue = time.Millisecond * 1
```
robscott, line 47: Why do we need this? "ms" is the smallest unit we support and we don't support fractional values. We also support "0" which makes this value particularly confusing.

apis/v1beta1/validation/httproute_test.go
robscott: +1 Would be great to have good coverage for our CEL validation tests. Note that these currently require a cluster with the latest CRDs installed. You can take a look at https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/verify-crds-kind.sh for an example of how we run these tests.

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1176,3 +1176,67 @@ func TestValidateRequestRedirectFiltersWithNoBackendRef(t *testing.T) {
...
+					},
+				},
+			},
+		}, {
+			name:     "valid httpRoute Rules timeout set to 0 (disabled)",
+			errCount: 0,
+			rules: []gatewayv1b1.HTTPRouteRule{
+				{
+					Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+						Request: toDuration("0s"),
```
robscott, line 1206: Would be nice to have a few more 0 values here, like 0ms, 0h, etc, just to ensure that we haven't just hardcoded support for 0s.

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms, as specified in GEP-2257,
+// and MUST BE >= 1ms or "0s" to disable (no timeout).
+//
+// +kubebuilder:validation:XValidation:message="backendRequest timeout cannot be longer than request timeout",rule="!(has(self.request) && has(self.backendRequest) && duration(self.request) != duration('0s') && duration(self.backendRequest) > duration(self.request))"
+type HTTPRouteTimeouts struct {
+	// Request specifies the duration for processing an HTTP client request after which the
```
robscott, line 282: Potentially a can of worms here, but seeing `Request` right beside `BackendRequest` feels a bit funny here. I know "Request Timeout" is a very common term so understand how we got there, and I think that both `Timeouts.Request` and `Timeouts.BackendRequest` independently make sense, but they feel a bit confusing when combined. Would any of the following pairs be clearer?

1. `ClientRequest` and `BackendRequest`
2. `FullRequest` and `BackendRequest`
3. `Request` and `PerTry`

This does not need to block, just would like to get a bit more feedback/discussion on potential name pairs before merging this.

/cc @frankbu @kflynn

robscott (commented):
Thanks @SRodi! I think this is very close to something we can merge, mostly have nits but a couple larger questions as well.

apis/v1beta1/httproute_types.go
frankbu: This is really a GEP1742 discussion, which we've already had: https://github.com/kubernetes-sigs/gateway-api/pull/1997#discussion_r1211967593

That said, I still prefer it as is, but could live with `ClientRequest` and `BackendRequest`.

robscott: Thanks for the link @frankbu! I'd assumed this was already covered in the GEP, just wanted to make sure we'd considered the alternatives here. I agree with what you and @youngnick said in that thread that "Request Timeout" is a well-known and broadly used term, and any of the alternatives for that feel inferior. I also can't come up with a great alternative to "Backend Request Timeout", so I'll just mark this as resolved.

apis/v1beta1/validation/httproute.go
frankbu: This is left over from before we changed to the restricted syntax where 1ms is the minimum value you can represent.

Seems like we can completely remove the min value checking from the validator now that anything less is not representable in the syntax. Only thing to check now is that BackendRequest <= Request timout and that 0 is allowed.

apis/v1beta1/httproute_types.go
frankbu: @robscott Are you sure about this? We discussed this before and I argue that if the doc is referring to the gateway instance/deployment, as opposed to the configuration resource, we use lower case. In this case it's not the Gateway configuration resource that is sending the request, it's the gateway instance.

robscott: Good catch, it feels really funny to see `gateway` instead of `Gateway` but I get your point here. Will go ahead and resolve this one.

apis/v1alpha2/shared_types.go
SRodi: addressed in e56b07bba434dcad3253d6ff03a07083f1aa198d

apis/v1beta1/httproute_types.go
SRodi: addressed in e56b07bba434dcad3253d6ff03a07083f1aa198d

apis/v1beta1/httproute_types.go
SRodi: addressed in e56b07bba434dcad3253d6ff03a07083f1aa198d

apis/v1beta1/validation/httproute.go
SRodi: addressed in e56b07bba434dcad3253d6ff03a07083f1aa198d

apis/v1beta1/validation/httproute_test.go
SRodi: addressed in e56b07bba434dcad3253d6ff03a07083f1aa198d

apis/v1beta1/validation/httproute_test.go
SRodi: addressed in e148233a0a32ebdd63d5d8c52468f0910ef345a2

apis/v1beta1/validation/httproute_test.go
```diff
@@ -1176,3 +1176,88 @@ func TestValidateRequestRedirectFiltersWithNoBackendRef(t *testing.T) {
...
+			name:     "valid httpRoute Rules timeout set to 0h (disabled)",
+			errCount: 0,
+			rules: []gatewayv1b1.HTTPRouteRule{
+				{
+					Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+						Request: toDuration("0h"),
+					},
+				},
+			},
+		}, {}, {}, {
```
frankbu, line 1230: Add another test 0s (infinite) is greater than 1ms is valid: Request "0s", BackendRequest "1ms")

SRodi: addressed in a9bd3a912f04a152b38034cdb56cd31b1dc41f51

pkg/test/cel/httproute_test.go
```diff
@@ -1391,3 +1391,105 @@ func validateHTTPRoute(t *testing.T, route *gatewayv1b1.HTTPRoute, wantErrors []
...
+			name: "valid timeout request infinite greater than backend request 1ms",
+			rules: []gatewayv1b1.HTTPRouteRule{
+				{
+					Timeouts: &gatewayv1b1.HTTPRouteTimeouts{
+						Request:        toDuration("0s"),
+						BackendRequest: toDuration("1ms"),
+					},
+				},
+			},
+		},
```
robscott, line 1480: Can you add a few more tests like this?

1. Request=1s,BackendRequest=200ms (should be valid)
2. Request=10s,BackendRequest=10s (should be valid)
3. Request=200ms,BackendRequest=1s (should not be valid) 

(Also would be great to replicate them in webhook tests while those still exist as well).

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+	// Support: Extended
+	//
+	// +optional
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are formatted like 1h/1m/1s/1ms, as specified in GEP-2257,
+// and MUST BE >= 1ms or "0s" to disable (no timeout).
```
robscott, line 278: To avoid being repetitive and risk this text getting out of sync I think it's best to leave the format description to the definition of that type. 
```suggestion
// Timeout values are represented with Gateway API Duration formatting.
// Specifying a zero value such as "0s" is interpreted as no timeout.
```

robscott (commented):
Thanks @SRodi! Found a couple more things but otherwise I think this is good to go.

apis/v1beta1/httproute_types.go
SRodi: addressed in ebfc098431820b4fb78d997c655ccfff7df23c62

pkg/admission/server_test.go
```diff
@@ -455,6 +455,144 @@ func TestServeHTTPSubmissions(t *testing.T) {
 				wantRespCode:       http.StatusInternalServerError,
 				wantFailureMessage: "unknown resource 'brokenroutes'\n",
 			},
+			{
+				name: "valid v1beta2 HTTPRoute resource with timeouts",
```
frankbu, line 459: Why are these timeout tests needed in server_test.go? Are they somehow needed to test that timeouts also work in v1alpha2 resources?

Anyway, if they are needed I think the name is wrong: v1beta2 -> v1alpha2?
```suggestion
				name: "valid v1alpha2 HTTPRoute resource with timeouts",
```

pkg/admission/server_test.go
```diff
@@ -455,6 +455,144 @@ func TestServeHTTPSubmissions(t *testing.T) {
...
+				reqBody: dedent.Dedent(`{
+						"kind": "AdmissionReview",
+						"apiVersion": "` + apiVersion + `",
+						"request": {
+							"uid": "03ca6fe0-4e17-11ee-be56-0242ac120002",
+							"resource": {
+								"group": "gateway.networking.k8s.io",
+								"version": "v1beta1",
```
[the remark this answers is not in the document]
SRodi: this was wrongly added, thanks @frankbu - addressed with 482ce1559fb95af38b191cef5eb30f9bfd638e54

pkg/admission/server_test.go
SRodi: this was wrongly added, thanks @frankbu - addressed with 482ce1559fb95af38b191cef5eb30f9bfd638e54

SRodi:
@robscott CEL validation tests are failing for the standard channel as this implementation is only available on the experimental channel - how does that work ?

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,56 @@ type HTTPRouteRule struct {
...
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are represented with Gateway API Duration formatting.
+// Specifying a zero value such as "0s" is interpreted as no timeout.
+//
+// +kubebuilder:validation:XValidation:message="backendRequest timeout cannot be longer than request timeout",rule="!(has(self.request) && has(self.backendRequest) && duration(self.request) != duration('0s') && duration(self.backendRequest) > duration(self.request))"
+type HTTPRouteTimeouts struct {
+	// Request specifies the maximum duration for a Gateway to respond to a HTTP request. If the Gateway
+	// has not been able to respond before this deadline is met, the Gateway MUST return a timeout error.
```
frankbu, line 283: nit: shorten lines to better align with the rest of the comments. Also "a HTTP" -> "an HTTP"
```suggestion
	// Request specifies the maximum duration for a Gateway to respond to an HTTP request.
	// If the Gateway has not been able to respond before this deadline is met, the Gateway
	// MUST return a timeout error.
```

frankbu: Also note that these 2 sentences have capitalized Gateway, even though they're referring to the gateway instance, not the configuration resource. This is different from other comments in this file that have gateway not capitalized. I prefer that lower case be used if not referring to the Gateway configuration resource, but more importantly it should be consistent, one way or the other.

cc @robscott WDYT?

kflynn: I agree with @frankbu here – these should be `gateway` without the capital `G`, since they refer to the controller rather than the resource.

apis/v1beta1/validation/httproute.go
```diff
@@ -348,6 +353,21 @@ func validateHTTPHeaderModifier(filter gatewayv1b1.HTTPHeaderFilter, path *field
...
 }
 
+func validateHTTPRouteTimeouts(timeouts *gatewayv1b1.HTTPRouteTimeouts, path *field.Path) field.ErrorList {
+	var errs field.ErrorList
+	if timeouts.BackendRequest != nil {
+		backendTimeout, _ := time.ParseDuration((string)(*timeouts.BackendRequest))
+		if timeouts.Request != nil {
+			timeout, _ := time.ParseDuration((string)(*timeouts.Request))
+			if backendTimeout > timeout {
+				errs = append(errs, field.Invalid(path.Child("backendRequest"), backendTimeout, "backendRequest timeout cannot be longer than request timeout"))
```
kflynn, line 363: If I'm reading this right - and I may not be - then

```yaml
timeouts:
    request: 0s
    backendRequest: 100ms
```

will fail this test, even though that's a valid timeout configuration. Am I reading this correctly?

kflynn (changes requested):
I had a couple of small questions – many thanks for doing this, @SRodi! 🙂

apis/v1beta1/validation/httproute.go
SRodi: good catch, I had mistakenly removed the `&& timeout != 0` on one of the last commits - re-adding that now. thank @kflynn

SRodi: addressed in d8d74726cc2f2a1c2e1e853865745ce3caee3af4

frankbu: @SRodi I thought there was a unit test for this case, but I see it's just in the CEL validation test. Please make sure that all the CEL tests are also replicated in the validator tests.

apis/v1beta1/httproute_types.go
SRodi: the capitalisation issue was fixed in e40ed42b80c4c98c3d5dd160a88a359d4bd4203d @frankbu @kflynn

apis/v1beta1/httproute_types.go
```diff
@@ -263,6 +263,57 @@ type HTTPRouteRule struct {
...
+	// <gateway:experimental>
+	Timeouts *HTTPRouteTimeouts `json:"timeouts,omitempty"`
+}
+
+// HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
+// Timeout values are represented with Gateway API Duration formatting.
+// Specifying a zero value such as "0s" is interpreted as no timeout.
+//
+// +kubebuilder:validation:XValidation:message="backendRequest timeout cannot be longer than request timeout",rule="!(has(self.request) && has(self.backendRequest) && duration(self.request) != duration('0s') && duration(self.backendRequest) > duration(self.request))"
+type HTTPRouteTimeouts struct {
```
frankbu, line 281: Do we also need `type HTTPRouteTimeouts = v1beta1.HTTPRouteTimeouts` in `apis/v1alpha2/httproute_types.go`?

frankbu:
Looks like the cel tests are failing?

I'm wondering why pull-gateway-api-verify says "Job succedded" even though several tests have failed?

^^^ @robscott any idea what's wrong here?

robscott:
@frankbu good catch, my best guess would be that https://github.com/kubernetes-sigs/gateway-api/pull/2362 may have introduced a regression here. Will put a hold on this until we can figure it out.

/hold

frankbu:
> Looks like the cel tests are failing?

Does `config/crd/standard/gateway.networking.k8s.io_httproutes.yaml` need to be regenerated to include the timeouts as well?

SRodi:
@frankbu @robscott as mentioned here https://github.com/kubernetes-sigs/gateway-api/pull/2013#issuecomment-1711732087 the CEL tests are failing on standard channel but passing on experimental. What should I do about that ?

robscott:
>@frankbu @robscott as mentioned here https://github.com/kubernetes-sigs/gateway-api/pull/2013#issuecomment-1711732087 the CEL tests are failing on standard channel but passing on experimental. What should I do about that ?

@SRodi you want to add the tests to this file instead: https://github.com/kubernetes-sigs/gateway-api/blob/main/pkg/test/cel/httproute_experimental_test.go. The test script actually loops through the channels, and installs CRDs and runs tests specifically for that channel (using build tags). Here's the actual test command that's being run: https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/verify-crds-kind.sh#L63.

>Does config/crd/standard/gateway.networking.k8s.io_httproutes.yaml need to be regenerated to include the timeouts as well?

Nope, experimental features are intentionally excluded from standard channel CRDs, so that part of this PR is working as intended.

apis/v1beta1/httproute_types.go
robscott: Yep, I think this is correct. I think this came from older feedback from me that suggested using `Gateway`, but I've since been convinced by @frankbu that this is the correct capitalization. Thanks for the update @SRodi!

apis/v1beta1/httproute_types.go
robscott: Let's leave it out. We're moving away from v1alpha2 types (they're no longer served in v0.8.0 CRDs) and I'd expect most/all implementations to have moved away from them before we hit v1.0.0. I haven't had time to figure out if we should leave v1alpha2 type definitions in place for one more release or drop them in v1.0, but in either case, I think I'd rather avoid adding anything new to them.

SRodi:
All tests are passing after moving them to httproute_experimental_test.go - thanks for clarifying this @robscott 

Is there anything else I should add to this PR?

youngnick:
This PR is just on hold while  I check the changes I introduced in #2362 - once I've either fixed any problems or confirmed it's not actually a problem, we can proceed. Until then, please sit tight.

robscott:
Fixes in #2362 were great, looks like we can remove the hold on this one.

/hold cancel
/approve

youngnick (commented):
/lgtm

Nice work @SRodi

PR #2015, opened 2023-05-12
GEP-2014: Declarative Policy

### LINKED ISSUE #2014: Users want a fully declarative policy mechanism
Managing policy is challenging. By definition, it's such a large, open-ended problem space that we cannot fully specify all possible policies. This means that must create extension mechanisms, which are in turn affected by the existing constraints on Kubernetes API design.

At the same time though, we need to ensure any mechanism we add for policy is broadly and easily usable by end users, and we should be willing to inconvenience implementers in order to gain clarity and usability for users. A critical aspect here is that we need to make sure that policy mechanisms fit into the fully-declarative model on which Kubernetes is built; policy mechanisms that don't line up with that model will cause a lot of usability headaches for anyone trying to use them.

kflynn:
This GEP is a follow-up to [GEP-713 Metaresources and Policy Attachment] to recommend that we consider changing the "attachment" part of "policy attachment" in favor of something that is declarative at the affected resource level.

(Fixes #2014)

[GEP-713 Metaresources and Policy Attachment]:https://gateway-api.sigs.k8s.io/geps/gep-713/

/kind gep

### DIFF
--- a/geps/gep-2014.md
+++ b/geps/gep-2014.md
@@ -1,20 +1,24 @@
 # GEP-X: Declarative Policy
 
-* Issue: TODO
+* Issue: [2014](https://github.com/kubernetes-sigs/gateway-api/issues/2014)
 * Status: Provisional
+* Authors: [Flynn](mailto:flynn@buoyant.io); [Shane Utt](mailto:shane@konghq.com)
 
 ## Definitions
 
-In this document we'll use `Policy` to refer to things that are specifically called policies
-as well as other "MetaResources" that follow similar patterns.
+In this document we'll use `policy` to refer to any resource whose purpose is
+setting policy around other resources. Notably, this could include either
+"policies" or "metaresources" as used in other documents: we're intentionally
+using the broader scope here.
 
-## TLDR
+## tl;dr:
 
-This proposal is a follow-up to [GEP-713 Metaresources and Policy Attachment][713] to recommend
-that we specifically remove the "attachment" part of "policy attachment" in favor of something
-that is declarative at the affected resource level.
+This proposal is a follow-up to [GEP-713 Metaresources and Policy Attachment]
+to recommend that we specifically remove the "attachment" part of "policy
+attachment" in favor of something that is declarative at the affected resource
+level.
 
-[713]:https://gateway-api.sigs.k8s.io/geps/gep-713/
+[GEP-713 Metaresources and Policy Attachment]:https://gateway-api.sigs.k8s.io/geps/gep-713/
 
 ## Goals
 
@@ -22,6 +26,10 @@ that is declarative at the affected resource level.
 - Retain `Policy` resource structure other than "attachment" semantics.
 - Provide new semantics to incorporate `Policy` resources at the level of the `Resource` that
   will be affected.
+
+## Non-Goals
+
+- To be clarified
 
 ## The Problem: A Parable of Jane
 
@@ -157,10 +165,16 @@ Given that the fundamental problem is that policy attachement isn't
 declarative as written and should be made declarative, there is only one
 fundamental answer: we need to modify the Kubernetes core resources to include
 extension points where a given object refers to its modifier, rather than
-having the modifying resource try to attach to its source. This is an ugly
-job, but it’s the only way to deal with this situation.
+having the modifying resource try to attach to its source. (For the record, we
+take no joy in this statement, but we do feel that it's the correct answer.)
 
 This GEP proposes to start this process with the Gateway API resources.
+
+A final note: while it's important to acknowledge that policy attachment is
+**not** the root cause of the application problems that Jane and Julian have
+in the parable above, it's also important to recognize that policy attachment
+makes understanding and fixing the problem much more difficult. That's the
+primary concern behind this GEP.
 
 ## API
 
@@ -168,9 +182,9 @@ TODO: future iteration
 
 ## Questions and Answers
 
-**Q**: _Why are you implying that there’s a problem with policy attachment?
-Isn’t your parable really just showing us that Jane and Julian work for a
-dysfunctional organization?_
+**Q**: _Isn’t your parable really just showing us that Jane and Julian work
+for a dysfunctional organization, rather than showing anything wrong with
+policy attachment?_
 
 **A**: As written, Evil Genius Cupcakes is _far_ from the most dysfunctional
 organization I’ve seen. Jane and Julian support each other, neither casts
--- a/geps/gep-2014.md
+++ b/geps/gep-2014.md
@@ -1,8 +1,8 @@
-# GEP-X: Declarative Policy
+# GEP-2014: Declarative Policy
 
 * Issue: [2014](https://github.com/kubernetes-sigs/gateway-api/issues/2014)
 * Status: Provisional
-* Authors: [Flynn](mailto:flynn@buoyant.io); [Shane Utt](mailto:shane@konghq.com)
+* Authors: [Flynn](https://github.com/kflynn); [Shane Utt](https://github.com/shaneutt)
 
 ## Definitions
 
--- a/geps/x.md
+++ b/geps/x.md
@@ -23,10 +23,257 @@ that is declarative at the affected resource level.
 - Provide new semantics to incorporate `Policy` resources at the level of the `Resource` that
   will be affected.
 
-## Introduction
+## The Problem: A Parable of Jane
 
-TODO
+It's a sunny Wednesday afternoon, and the lead microservices developer for
+Evil Genius Cupcakes is windsurfing. Work has been eating Jane alive for the
+past two and a half weeks, but after successfully deploying version 3.6.0 of
+the `baker` service that morning, she left early to try to unwind a bit.
+
+Her shoulders are just starting to unknot when her phone pings with a text
+from Julian, down in the NOC. Waterproof phones are a blessing, but also a
+curse.
+
+**Julian**: _Hey Jane. Things are still running, more or less, but latencies
+on everything in the baker namespace are crazy high after your last rollout,
+and baker itself has a weirdly high load. Sorry to interrupt you on the lake
+but can you take a look? Thanks!!_
+
+Jane stares at the phone for a long moment, then slumps and heads back to
+shore to dry off and grab her laptop.
+
+What she finds is strange. `baker` is taking a _lot_ of load, almost 4x what’s
+being reported by its usual clients, and its clients report that calls are
+taking much longer than they’d expect them to. She doublechecks the
+Deployment, the Service, and all the HTTPRoutes around `baker`; everything
+looks good. `baker`’s logs show her mostly failed requests... with a lot of
+duplicate requests? Jane checks her HTTPRoute again, though she's pretty sure
+you can't configure retries there, and finds nothing. But it definitely looks
+like a client is retrying when it shouldn’t be.
+
+She pings Julian.
+
+**Jane**: _Hey Julian. Something weird is up, looks like requests to `baker`
+are failing but getting retried??_
+
+A minute later he answers.
+
+**Julian**: 🤷 _Did you configure retries?_
+
+**Jane**: _Dude. I don’t even know how to._ 😂
+
+**Julian**: _You attach a RetryPolicy attached to your HTTPRoute?_
+
+**Jane**: _Nope. Definitely didn’t do that._
+
+She types `kubectl get retrypolicy -n baker` and gets a permission error.
+
+**Jane**: _Huh, I actually don’t have permissions for RetryPolicy._ 🤔
+
+**Julian**: 🤷 _Feels like you should but OK, guess that can’t be it._
+
+Minutes pass while both look at logs.
+
+**Jane**: _OK, it’s definitely retrying. Nearly every request fails the first
+few times, gets retried, and then finally succeeds?_
+
+**Julian**: _Are you sure? I don’t see the `mixer` client making duplicate requests…_
+
+**Jane**: _Check both logs for request ID
+6E69546E-3CD8-4BED-9CE7-45CD3BF4B889. `mixer` sends that once, but `baker`
+shows it arriving four times in quick succession. Only the fourth one
+succeeds. That has to be retries._
+
+Another pause.
+
+**Julian**: _I’m an idiot. There’s a RetryPolicy for the whole namespace –
+sorry, too many policies in the dashboard and I missed it. Deleting that since
+you don’t want retries._
+
+**Jane**: _Are you sure that’s a good–_
+
+Jane’s phone shrills while she’s typing, and she drops it. When she picks it
+up again she sees a stack of alerts. Quickly flipping through them, she feels
+the blood drain from her face: there’s one for every single service in the
+`baker` namespace.
+
+**Jane**: _PUT IT BACK!!_
+
+**Julian**: _Just did. Be glad you couldn't hear all the alarms here._ 😕
+
+**Jane**: _What the hell just happened??_
+
+**Julian**: _At a guess, all the workloads in the `baker` namespace actually
+fail a lot, but they seem OK because there are retries across the whole
+namespace?_ 🤔
+
+Jane’s jaw drops.
+
+**Jane**: _You’re saying that ALL of our services are broken??!_
+
+**Julian**: _That’s what it looks like. Guessing your `baker` rollout would
+have failed without retries turned on._
+
+There is a pause while Jane thinks through increasingly unpleasant possibilities.
+
+**Jane**: _I don't even know where to start here. How long did that
+RetryPolicy go in? Is it the only thing like it?_
+
+**Julian**: _I didn’t look closely before deleting it, but I think it said a
+few months ago. And there are lots of different kinds of policy and lots of
+individual policies, hang on a minute…_
+
+**Julian**: _Looks like about 47 for your chunk of the world, a couple hundred
+system-wide._
+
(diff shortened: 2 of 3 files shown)


howardjohn (commented):
Thanks, this is a useful perspective/doc/etc.

While I agree with a lot of the concerns I am not sure I agree with most of the conclusions. I think a lot of this is personal preference territory, so I am certainly not saying you are *wrong* - just that our preferences may differ. Some of the below is also just playing devils advocate, to some extent - I probably agree with you more than it seems like.

First, I would argue this is not a new problem. As long as there is more than 1 type in Kubernetes, this problem exists. Similar arguments could be made for almost any <insert 2 Kubernetes types>. Service and Deployment could have been one object (or HPA, PDB, ResourceQuota, ...) but they aren't and users either adopt the resource model or consume higher level abstractions over these resources or use simpler products (docker-compose for example). Its not clear to me this is fundamentally a new problem.

Julian should not be building tooling, that is our job as the gateway-api community (or extended community - but certainly not each user).

This may just be a language thing, but I am not sure what makes this not declarative and it isn't described in the GEP. Seems declarative to me - they declared their intent to add retries at the namespace level and that intent was actuated by the system.

All of this feels answerable by just making policy attachment easy to grok for users, rather than fundamentally changing the concepts.

In any case, without a concrete proposal I am not sure what the next steps here are. I/We can agree policy attachment sucks, but also that extensibility is essential to the API. Without a better option, the bad option is still better than no option... Thats not to say there isn't value in calling it out as bad, but without alternatives proposed there doesn't seem to be a path forward other than encoding this as "Informational - Policy attachment has problems but we are doing it anyways"?

sunjayBhatia (commented):
I'm curious if you feel any of the ideas from https://github.com/kubernetes-sigs/gateway-api/discussions/2012 help alleviate any concerns, IMO giving some visibility (via Status etc.) into what *Policies are in effect on particular resources would help a lot with the problems described here

robscott:
Thanks for writing this up @kflynn!

I'll be the first to agree that our existing policy attachment model complex. In my opinion, the primary issue with it is that it is difficult for users to understand which policies are available and what they effect. We've also had some previous discussions that covered the complexity involved here, for example https://github.com/kubernetes-sigs/gateway-api/discussions/1715. So I think we're roughly agreeing on the problem statement here, and many in the community likely have similar concerns. 

Having successful extension mechanism(s) is absolutely key to the success of the API. One of our key goals for this API has been to provide a better extension model than the Ingress and Service APIs so we don't end up with yet another explosion of annotations. 

With all of that said, I'm not sure what the proposal is here. It seems like this may be leading towards a suggestion that all policy needs to be directly included in or referenced by the resources it affects. For many cases, this is easier to reason with, and so we definitely tried to make this work in the initial policy attachment proposals. Ultimately we avoided it for the following reasons:

1. We can not reasonably introduce something like a `policyRefs` field to every API that could possibly be targeted by policy attachment (for example, Service, ServiceImport, Namespace, Pod, etc)
1. It would be unreasonably complex to have different ways to attach the same policy (forward and backward refs)
1. Directly referencing policies from resources would not work well for hierarchical policy when defaults and overrides will affect all resources below them in a tree

In the broader world of Gateway API extensions, we do already have custom HTTPRoute filters and GatewayClass params which are both direct refs from Gateway API resources to custom extensions. I think those have worked reasonably well. We may want to consider adding something like Gateway or Listener level (custom) filters to build on this model. This is decidedly not policy attachment but can be very useful for some extensions.

As far as policy attachment itself, I think there are a lot of relatively small things we can do that will dramatically improve the experience. We're collecting ideas here: https://github.com/kubernetes-sigs/gateway-api/discussions/2012. Wherever possible I'd like to build on what we have here with incremental improvements instead of introducing significant changes to the model.

youngnick:
Okay. I've obviously got a _lot_ of thoughts about this, so I'm going to take some time and address general things first, and then more specific things.

Firstly, I really appreciate the writing out of the story of how things can go wrong. It's useful to have a standard example for us to talk about. I think it's important to remember that Policy Attachment as it stands is _not complete_. It's still needing the solve the exact problems that this story raises, I definitely agree. We have been putting this off because it's a _very_ complex problem that has a lot of dimensions that aren't obvious at first glance (and are difficult to write down as well). So I think there is context that we've _discussed_ in the past missing from the _written_ documentation as well.

To get specific:

I _strongly_ disagree that the Policy Attachment we have is not _declarative_. It's definitely declarative in that _a_ user's intent is expressed, and the system _as a whole_ is ensuring that the intent is represented. It's not easily _discoverable_ on a single-resource level, which is where the action-at-a-distance feeling comes from.

There are multiple pieces of feedback that we _need_ but have not _defined_ for Policy Attachment as it stands:
- There needs to be a _standard_, _simple_ way for an owner of an object that is being affected by a Policy to know _that_ it's being affected by _some_ Policy and if possible _what_ Policy is affecting it. Based on the questions and goals, it seems like this is the main thrust of what this GEP will propose.
- There needs to be a _standard_ and _straightforward_ way to view the _resultant set of Policy_ for an object - the final config that's being set for it. Because Policy is arbitrarily extensible (by design), this is _hard_. Like, crypto hard.

Reading between the lines and from out-of-band discussions, I believe that the likely proposal is something like: "We're going to work on adding something to `metadata` that will allow _every_ Kubernetes resource to request other resources affect it in some way - which will include Policy". That certainly would _help_ with the discoverability problem, but because some Policies can have effects that span multiple objects, it doesn't completely solve it (@robscott alludes to this in his point 3 above). An example of a question here is "how do you specify defaults that effect all HTTPRoutes attached to a Gateway" or similar?

To talk about this more concretely in relation to the given story, the problem arises because Jane has no way to know that there's a `RetryPolicy` attached to the namespace , and so that everything is _already_ being retried. As I said earlier, this is not because the `RetryPolicy` is not declarative, it's because the _implications_ of the declarative config are not made clear to everyone that the Policy affects. And again, it's not just _that_ there is Policy attached that's the problem, it's also the _content_ of the Policy.

@robscott's discussion issue already has a couple of options here, and I have a couple of things I'd like to add, but I think the critical thing for us to discuss is about what makes something declarative and what doesn't. Because I think that's the critical impedance mismatch here.

kflynn:
Thanks for the comments, all! Rather than try to answer line-by-line or any such madness, let me instead try to tease out some commonalities across several responses and try to speak to them.

I also want to explicitly acknowledge that there's a _lot_ to unpack here. GEP-713 represents an enormous amount of work (thanks for all that, @nickyoung and @robscott), and it's revealed a lot of new problems that simply can’t be solved without many of us in the community working together to get the best results. So I very much appreciate the willingness of everyone to pitch in and try to constructively get to the bottom of these things -- many thanks, especially, to @nickyoung for taking the lion's share of the work and doing it gracefully and good-humoredly.

#### 1. _Policy attachment is definitely declarative; the issue is really one of discoverability._

This is both a really important thing to dig into _and_ a slippery slope where we could end up spending massive amounts of time on words about words. I’d really like to avoid the latter, so: the problem I see is that policy attachment can give Jane the sense that what she declares is being ignored and Kubernetes won’t tell her why. I used “not declarative” to summarize that, but I’m happy to use a different phrase if that’s more helpful; I care more about solving the problem than about being the one to give it a name. 🙂

If more explanation is useful: what we teach new folks coming to Kubernetes is that they declare what they want, Kube gives it to them, and they can always read the current state of the world directly from the cluster. There are boundaries, of course (RBAC, etc.), and in practice people understand that maybe they'll only get something close to what they want, but it's still presented as a contract that Kube will honor. If Jane declares that she wants a particular routing behavior and then gets something bewilderingly different, she will feel that this contract has been broken. If she can’t even read the cluster to find out why, she’ll feel it’s been broken twice.

I feel like honoring that contract is an important part of being “declarative”: expressing the API in YAML docs rather than function calls is only part of it. Let me quote @nickyoung for a moment:

> It's definitely declarative in that a user's intent is expressed, and the system as a whole is ensuring that the intent is represented. It's not easily discoverable on a single-resource level, which is where the action-at-a-distance feeling comes from.

There's a lot I agree with in here, but I want to call out explicitly that in the situation of the parable, Jane is going to feel like  _her_ intent is being ignored. That's why I think that this may be worse than a discoverability problem (though I absolutely agree that the discoverability issue is critical). As noted in the Gateway API meeting yesterday, policy attachment to date has largely focused on the cluster provider and below; this leaves Jane, as an application developer, in a difficult spot (especially as ops teams try to shift concerns closer to the developers).

#### 2. _This isn't a fundamentally new problem in Kubernetes. There are other areas where this is already a thing._

Definitely, and this was one of our key considerations while discussing this GEP: we think it’s possible for the work we do in Gateway API to have an effect on the Kubernetes project as a whole if we’re successful  here. These existing areas are actually some of the places where we see the most difficulty when teaching folks about Kubernetes: for example, think about teaching a novice about label selector linkage between Deployments, Pods, and Services. Even better, think back to learning it as a novice yourself. I would love to not make that worse.

#### 3. _Maybe the Gateway API can provide higher-level constructs to make things better._

I definitely agree that there’s value in exploring that idea, though I’m concerned that getting to consensus on what those higher-level constructs would look like will take a very long time. Higher-level tooling or abstractions necessarily involve opinionation that feels tricky to make portable.

(Tooling also has the interesting challenge that if we e.g. do a `kubectl` plugin, it won't necessarily help the mass of other tools already out there. That concern is a little out of scope, because I don't want to imply that we shouldn't try to create helpful tooling -- I just want to point out that it's less of a panacea than making the API easier for everyone to reason about.)

#### 4. _Policy attachment is complex, but we can't change the Kubernetes core so we're stuck with the complexity._

I have a lot of conflicting reactions to this one.

First, I do fully believe that changes to the core are worse than anyone ever expects, even when you take that fact into account. 😕

Second, and most importantly, I have nothing but respect and appreciation for all the work that @nickyoung and @robscott have put into this so far (again, props especially to Nick for being the one to take all the complaints!). I'm happy to defend them for the enormous amount of blood, sweat, and tears they've poured into this, and I'm deeply appreciative that they dove into this problem in the first place.

Bearing that in mind... it’s very clear that necessity mothered the invention of the current policy attachment mechanism, and also very clear that as we experiment with it, we’re seeing problems that necessity didn’t reveal at the start. (This is a _success_ story: I’m delighted and grateful that the experimentation has been happening!) I think we should dig into these things now, before we mark the API GA and people start throwing policies all over the place.

If we can't start with core modifications, maybe we can start with giving all the Gateway API resources a consistent extension mechanism (rather than just the GatewayClass paramsRef and custom HTTPRoute filters), and see how that goes. If it's NP-hard to fully tackle everything all at once, is there a way we can restrict things and do something simpler while not painting ourselves into horrible corners? (The discussions in #2012 are most definitely relevant here.)

Those are discussions that I think need careful attention before GA. (And as I've said before, I actually feel bad for raising all this stuff now, after not being around at the start of the Gateway API effort -- I'm sure that some of these have been explored in some context, and I apologize for that.)

youngnick:
Thanks for the comment @kflynn, I just edited it to fix my Github handle so that some other guy didn't get a bunch of notifications. Still processing, will respond tomorrow.

kflynn:
Thanks for that, @youngnick! 🤦‍♂️ I can't believe I did that. 🙁 🙁

kflynn:
/easycla

(since this should be fixed now... 🤞)

robscott:
Thanks for the great response @kflynn! 

> the problem I see is that policy attachment can give Jane the sense that what she declares is being ignored and Kubernetes won’t tell her why

Note that nothing that Jane declared was ignored, additional information/config was added on top though. I think https://github.com/kubernetes-sigs/gateway-api/discussions/2012 has some compelling options to help with the discoverability aspect of this statement though.

> If Jane declares that she wants a particular routing behavior and then gets something bewilderingly different, she will feel that this contract has been broken.

To me adding retries does not feel like the contract has been broken _unless_ Jane explicitly specified that there should not be retries.

> if we e.g. do a kubectl plugin, it won't necessarily help the mass of other tools already out there

Any tooling that we own as a project will be OSS and should be widely usable as a library for any other tools that want to build on top of these concepts.

> If we can't start with core modifications, maybe we can start with giving all the Gateway API resources a consistent extension mechanism

What would we do with other resources in the meantime? I think almost everyone, likely including the project itself with Gateway -> Backend TLS config, will need to extend Service and/or ServiceImport _somehow_. I'd hate to have one way for Gateway API resources and a separate way to attach policy to _everything else_. I also think that direct refs from target resource to policy/extension would prevent us from some important use cases:

* Organization admins that want to apply policy broadly
* Hierarchical policy attachment 
* Separation of concerns between policy owners (admins) and app owners

Although I'll be the first to admit that the current model was influenced by the limitations of the APIs we were targeting, it was also heavily influenced by the use cases described above. Even if we could have direct refs from all Kubernetes resources to Gateway API policies/extensions tomorrow, I'm not convinced that would actually be better. We would lose out on some significant and important use cases with that approach. With that said, it's a moot point because we're at least 16 months away (and a _lot_ of effort) from that even being a possibility. I personally would rather focus that time on improving our current model with some of the ideas in #2012.

youngnick:
Flynn and I talked about this last week, and I _think_ we agreed on a few things. Flynn, please correct me if I say anything we don't actually agree on. 😃

Also, I've tried to keep this focussed on the parable, and using the parable to draw out new requirements for Policy Attachment, in the interest of moving forward like @shaneutt suggested above.

The problem here is that Jane ends up having Policy affecting _her_ resources without any way to know that ahead of time.

_Also_, Julian had no way to know how many places the Namespace Policy was affecting (or even if it was affecting anything).

This is what Flynn called "not declarative" (Because Jane has her own services being affected by configuration she doesn't own and doesn't know about), and I called "not discoverable".

The naming is not as important as agreeing that it's a problem.

Another way to say this problem uses something I've talked about before - we want to keep the [troubleshooting distance](https://youngnick.net/post/troubleshooting-distance/) for a Route affected by Policy as close as possible to 1.

That is, Jane should be able to understand the state of her Route by doing a number of `kubectl` commands as close as possible to 1. Ideally, `kubectl get httproute` should tell you _everything_ you need to know. That _may_ not be actually achievable, but that should be the goal.

There are three things Jane needs to be able to know about her Route:

1. That a Policy is affecting a Route
1. Which Policy is affecting a Route
1. What settings the resultant set of Policy is configuring for the Route

There's a relatively obvious solution here, that's worth _showing_ and then talking about.

```yaml
   status:
    effectivePolicy:
    - group: example.com
      version: v1alpha1
      kind: RetryPolicy
      name: default-retry
      namespace: system
      attachedTo:
        version: v1beta1
        kind: Gateway
      policySettings:
        defaults:
          retries: 3
          delay: 5
          timeout: 10
```
Here `policySettings` would practically need to be a `map[string]string`, or an `unstructured.Unstructured` or something to allow it to contain the arbitrary data it would need to.

This is the most information we could possibly supply to Jane, and it would completely stop the parable from happening. It also meets the third goal above - the full resultant set of Policy for an object.

However, this creates significant scalability concerns in that this status would need to be updated every time the Policy or the Gateway it attaches to changed, and will need to be updated on _every_ relevant object (both the Gateway and any attached HTTPRoute). This means that an update in a single object (the `default-retry` Policy) could create a large number of status updates (as many HTTPRoutes as are attached to that Gateway.) If the Policy used a Namespace attachment, a similar problem ensues. This is the definition of fanout apiserver load, which historically has created significant issues, and in general should be avoided.

Some other problems I can see:
- No support for multiple controllers supporting the same Policy type (which may happen if we include some Policy with Gateway API)
- The more Policy that affects the single object, the larger the `effectivePolicy` stanza will get, running us up against etcd record size limits (again!)

This problem arises from the number of relationships there are between a single Policy and a number of objects it affects, not in the way those relationships are created. For Direct Attached Policy, this is only ever `1`, because it's _direct_. But for Inherited, it's arbitrarily large.

This is why I don't think that changing Policy so that affected objects reference the Policy that should affect them is useful enough to justify the effort, because while it improves the experience for direct-attached policy, it doesn't for inherited policy. I also suspect that over time, inherited Policy will be as used as Direct if not more so.

The above example is also only possible on objects that we control the `status` of - like objects in Gateway API. We couldn't add something like this to Service's `status` in the event that we add a TLSConnectionPolicy, for example. Because Policy can apply to _any_ object, we need a _selection_ of methods we can use to breadcrumb users towards the Policy that's affectingt their object.

This is why @robscott's discussion is important, I think we will need to use _multiple_ options out of there, based on things like scalability, what apigroup the object is in, and so on. We should be aiming to make it straightforward for every implementation to do at least _one_ of the pieces of information I listed above: _That_ there's a Policy, _which_ Policy is effective, and _what_ settings that Policy provides. We also need to accept that different target objects may end up with different available information (because they're different objects!), and we can't control _all_ the possible ways that this may be used.

I think we're okay here because Policy Attachment is a _pattern_, not an _API_. So we're trying to give implementations direction on ways to avoid the problems in the parable, since we can't be too prescriptive about rules, because in the end, many implementations of Policy will be ImplementationSpecific anyway.

So, to summarize this _very_ long comment:
- We all seem to agree that Jane and Julian's experience in this parable is bad
- I propose that update the deliverables of Policy Attachment as a pattern:
  - that implementations know that discoverability is a problem
  - provide possible solutions and direction on solving the problem
  - provide guidance on evaluating those solutions (that is, troubleshooting distance/troublshooting hops as a metric)
- I'm -1 on reversing the direction of attachment for Policy Attachment, because it doesn't solve enough of the broader discoverability problem, and would be a _lot_ of effort.

kflynn:
This has been subsumed by #2128.

PR #2024, opened 2023-05-15
conformance: fix ipv6 addresses

howardjohn:
**What type of PR is this?**

/kind test

This was a regression in 0.7.0-rc2. We generated invalid IPv6 address

Fixes #

```release-note
Fixed an issues causing conformance tests to fail when using IPv6 addresses
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -148,12 +148,21 @@ func calculateHost(gwAddr, scheme string) string {
 		return gwAddr
 	}
 	if strings.ToLower(scheme) == "http" && port == "80" {
-		return host
+		return ipv6SafeHost(host)
 	}
 	if strings.ToLower(scheme) == "https" && port == "443" {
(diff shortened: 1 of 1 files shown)


conformance/utils/http/http.go
```diff
@@ -148,12 +148,21 @@ func calculateHost(gwAddr, scheme string) string {
...
 	}
-	return host + ":" + port
+	return gwAddr
+}
+
+func ipv6SafeHost(host string) string {
+	// We assume that host is a literal IPv6 address if host has
+	// colons.
+	if strings.Contains(host, ":") {
+		return "[" + host + "]"
```
aojea, line 163: but you only maybe you should reference this https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2?

robscott:
Thanks @howardjohn!

/lgtm
/approve

PR #2029, opened 2023-05-17
conformance: add MeshConsumerRoute case for GAMMA

michaelbeaumont:
**What type of PR is this?**
/kind test
/area conformance

Tests the namespace-dependent behavior of HTTPRoute aka consumer routes.

```release-note
NONE
```

### DIFF
--- a/conformance/mesh/manifests.yaml
+++ b/conformance/mesh/manifests.yaml
@@ -6,7 +6,6 @@ metadata:
   labels:
     gateway-conformance: mesh
 ---
-
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -141,3 +140,33 @@ spec:
   - name: grpc
     port: 7070
     appProtocol: grpc
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gateway-conformance-mesh-consumer
+  labels:
+    gateway-conformance: mesh-consumer
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: echo-v1
+  namespace: gateway-conformance-mesh-consumer
+  labels:
+    app: echo
+spec:
+  selector:
+    matchLabels:
+      app: echo
+      version: v1
+  template:
+    metadata:
+      labels:
+        app: echo
+        version: v1
+    spec:
+      containers:
+      - name: echo
+        image: gcr.io/k8s-staging-gateway-api/echo-server:v20230505-v0.7.0-rc1-10-g497e67da
+        imagePullPolicy: IfNotPresent
--- a/conformance/tests/mesh-consumer-route.go
+++ b/conformance/tests/mesh-consumer-route.go
@@ -41,6 +41,7 @@ var MeshConsumerRoute = suite.ConformanceTest{
 		consumerClient := echo.ConnectToAppInNamespace(t, s, echo.MeshAppEchoV1, "gateway-conformance-mesh-consumer")
 		consumerCases := []http.ExpectedResponse{
 			{
+				TestCaseName: "request from consumer route's namespace modified by HTTPRoute",
 				Request: http.Request{
 					Host:   "echo-v1.gateway-conformance-mesh",
 					Method: "GET",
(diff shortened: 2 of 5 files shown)


shaneutt:
hold for some other implementations to test

/hold

conformance/mesh/manifests.yaml
```diff
@@ -132,3 +131,33 @@ spec:
     port: 9090
   - name: grpc
     port: 7070
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: gateway-conformance-mesh-consumer
```
howardjohn, line 138: namespace needs to be added here: https://github.com/istio/gateway-api/blob/a258900338d925af89116ef29d613bc6e84be032/conformance/utils/suite/suite.go#L175 (and maybe fix the log to say "Mesh" instead of "base" while we are there), so it will wait for the pod to be ready

conformance/tests/mesh-consumer-route.yaml
```diff
@@ -0,0 +1,17 @@
...
+  - kind: Service
+    name: echo-v1
+    namespace: gateway-conformance-mesh
+  rules:
+  - filters:
+    - type: ResponseHeaderModifier
+      responseHeaderModifier:
+        set:
+        - name: X-Header-Set
+          value: set
```
howardjohn, line 17: This relies on https://gateway-api.sigs.k8s.io/geps/gep-1426/?h=xro#omitting-backendrefs which is only SHOULD. Can we put the `backendRef` explicitly?

howardjohn (commented):
lgtm with two fixes. istio actually fails this, but its our issue - I will address it on our side

conformance/tests/mesh-consumer-route.go
```diff
@@ -0,0 +1,84 @@
...
+				},
+				Response: http.Response{
+					StatusCode:    200,
+					AbsentHeaders: []string{"X-Header-Set"},
+				},
+				Backend: "echo-v1",
+			},
+		}
+		for i, tc := range consumerCases {
+			t.Run(tc.GetTestCaseName(i), func(t *testing.T) {
```
howardjohn, line 74: can we name the cases? or we end up with

```
        --- FAIL: TestGatewayConformance/MeshConsumerRoute/0_request_to_'echo-v1.gateway-conformance-mesh/'_should_go_to_echo-v1 (30.00s)
        --- PASS: TestGatewayConformance/MeshConsumerRoute/0_request_to_'echo-v1.gateway-conformance-mesh/'_should_go_to_echo-v1#01 (0.13s)
```

howardjohn (commented):
also may be good to set both routes and ensure consumer 'wins'. No problem being a followup though

conformance/tests/mesh-consumer-route.yaml
```diff
@@ -0,0 +1,20 @@
...
+    backendRefs:
+    - name: echo-v1
```
howardjohn, line 19: did you intend for it to not have `namespace` set? The route is valid, just not what we had before

note the test is only checking `Backend: echo-v1` but we have it in both namespaces, so its untested if we go to the correct namespace?

howardjohn (approved):
/hold
/lgtm

LGTM, put a hold since there were open but optional comments. Feel free to remove hold

conformance/tests/mesh-consumer-route.yaml
michaelbeaumont: I did intend it to be the same as the omitted backendRefs route, I guess I intuitively assumed the omitted ns in a backendRef would imply the _parentRef_ namespace or something. Fixed

> note the test is only checking Backend: echo-v1 but we have it in both namespaces, so its untested if we go to the correct namespace?

Not sure that necessarily falls under this test, since I feel like this should just check consumer NS on, non-consumer NS off, but I suppose it should be tested somewhere.

michaelbeaumont: Looks like `Backend` is the hostname of the responding Pod so it will just have the Pod's name. So I suppose either we tweak [the echo server](https://github.com/istio/istio/blob/20bb4cad57d6abc5d30d9adb819321d47527909c/pkg/test/echo/response.go#L33) to put the Pod's namespace in the response or we use the k8s API to look up the Pod in the Namespace we expect it to be in using the hostname

howardjohn: Oh good point it is just the name. We can easily add namespace. Lets do a followup to do that?

shaneutt (approved):

howardjohn (approved):

PR #2031, opened 2023-05-17
gamma: drop provision to omit backendRef

howardjohn:
**What type of PR is this?**

/kind gep

IMO, this does not make sense as SHOULD. It should be a MUST or not in the spec.

I have proposed removal for the following reasons:
* Inconsistent with Gateway parent ref (which also means you cannot possibly have a service parent and gateway parent)
* Limits ability to have non-Service parents, as it implies the frontend and backend are the same object

```release-note
NONE
```

### DIFF
--- a/geps/gep-1426.md
+++ b/geps/gep-1426.md
@@ -71,30 +71,6 @@ spec:
 
 In the example above, routing rules have been configured to direct 90% of traffic for the `foo` `Service` to the default "backend" endpoints specified by the `foo` `Service` [`selector`](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service) field, and 10% to the `foo-v2` `Service`. This is determined based on the `ClusterIP` (for `Service`) and `ClusterSetIP` (for `ServiceImport`) matching, and for "transparent proxy" mesh implementations would match all requests to `foo.svc.cluster.local` (or arbitrary custom suffix, as the hostname is not specified manually) from within the same namespace, all requests to `foo.store.svc.cluster.local` from other namespaces, and all requests to `foo.store.svc.clusterset.local` for multicluster services, within the scope of the service mesh.
 
-### Omitting `backendRefs`
-
-Implementations SHOULD support a terse syntax which allows omitting `backendRefs` to avoid unnecessary redundancy for simple configurations. If no `backendRefs` are specified, implementations should direct traffic to the default endpoints of the `parentRef` service. The behavior should be exactly the same as if the `parentRef` service was explicitly defined as the only `backendRef`, so if the `parentRef` does not have any endpoints, implementations MUST return an HTTP 503 response code (see discussion in [#1210](https://github.com/kubernetes-sigs/gateway-api/pull/1210)).
-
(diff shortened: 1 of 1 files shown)


howardjohn:
@robscott @michaelbeaumont @keithmattix @mikemorris

geps/gep-1426.md
```diff
@@ -71,30 +71,6 @@ spec:
...
-    name: foo
-  rules:
-    matches:
-    - path:
-        value: "/bar"
-```
-
-The example above would drop all incoming traffic for HTTP paths other than `/bar` to the `foo` `Service`, following the existing spec to return HTTP 404 response codes for unmatched requests, and HTTP 500 response codes for requests excluded due to an [`HTTPRouteFilter`](https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter).
-
-When no `HTTPRoute` resources or AuthZ configuration are defined, all traffic should implicitly work - this is just how Kubernetes functions. When you create an `HTTPRoute` targeting a service as a `parentRef` you are replacing that implicit logic - not adding to it. Therefore, you may be reshaping or restricting traffic via an `HTTPRoute` configuration (which should be noted is distinct from disallowing traffic by AuthZ).
```
michaelbeaumont, line 96: I think this line should still be in the GEP

michaelbeaumont:
https://github.com/kubernetes-sigs/gateway-api/blob/a8826b8dec2f1ea40a4fa067a04d23c44c168916/geps/gep-1426.md?plain=1#L164-L169

Doesn't requiring `backendRefs` make the "all ports" option significantly less useful? Or is there a use case to direct multiple frontend ports to the same backend port?

howardjohn:
> Doesn't requiring `backendRefs` make the "all ports" option significantly less useful? Or is there a use case to direct multiple frontend ports to the same backend port?

I would say yes, and I think that is the most compelling reason to keep it if we do. But I would note this is a fairly narrow UX optimization; once they start using a different backend (example - canary) then they lose this enhancement. It seems a like a possible foot-gun that to change the backend from `svc` to `svc-v1=50%, svc-v2=50%` I need to change the parentRef _and_ backendRef, for example.

michaelbeaumont:
> > Doesn't requiring `backendRefs` make the "all ports" option significantly less useful? Or is there a use case to direct multiple frontend ports to the same backend port?
> 
> I would say yes, and I think that is the most compelling reason to keep it if we do. But I would note this is a fairly narrow UX optimization; once they start using a different backend (example - canary) then they lose this enhancement. It seems a like a possible foot-gun that to change the backend from `svc` to `svc-v1=50%, svc-v2=50%` I need to change the parentRef _and_ backendRef, for example.

👍 can we drop the optional port for Service parentRefs if we drop optional backendRefs? Then ideally "missing port & kind: Service" in a parentRef would be rejected by the webhook.

But the status quo of optional parentRef port and optional backendRefs works for me.

robscott:
@howardjohn I think if we do remove this we should clearly note why it was removed and provide very clear guidance for how the situation should (or must) be handled. Removing the "SHOULD" without actually leaving any guidance for how to handle this case could result in more confusion and inconsistency across providers.

howardjohn:
It should be done gateway-api wide then IMO. It's not gamma specific to
have no backendRef (if we merge this, anyways).
Which the answer will be a 500 error I think

michaelbeaumont (approved):
LGTM other than line 96

howardjohn:
Add back removed line and explained this was removed

keithmattix (approved):

Issue #1993, opened 2023-05-04
Method matching importance in matching priorities


mlavacca:
**What would you like to be added**:

When there are multiple HTTPRoute rules matching, the documentation [states](
https://github.com/kubernetes-sigs/gateway-api/blob/a56f84f6bffdb527c6796ed15c04440afd5a1ed6/apis/v1beta1/httproute_types.go#L154-L164):

```go
	// Proxy or Load Balancer routing configuration generated from HTTPRoutes
	// MUST prioritize matches based on the following criteria, continuing on
	// ties. Across all rules specified on applicable Routes, precedence must be
	// given to the match with the largest number of:
	//
	// * Characters in a matching "Exact" path match
	// * Characters in a matching "Prefix" path match
	// * Header matches.
	// * Query param matches.
	//
	// Note: The precedence of RegularExpression path matches are implementation-specific.
```

In this section, there is no mention of the method matching, that should be taken into account. A fifth line containing method matching could be added after the first four related to matching types.

sunjayBhatia:
I would like to bump the priority for 0.7.x for this since we've added a test that covers it without an explicit mention in the spec, see https://github.com/kubernetes-sigs/gateway-api/pull/1984/files#r1200873543

sunjayBhatia:
Another approach would be to comment out the test that is not covered by the spec and add a TODO until the next minor release

PR #2054, opened 2023-05-22
Add method matching to HTTPRoute matching precedence order

gauravkghildiyal:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind test
/area conformance

Clarify the precedency that Method matching has while comparing similar HTTPRouteRules.

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1993

```release-note
HTTPRoute Method matching precedence has been clarified
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -158,12 +158,13 @@ type HTTPRouteRule struct {
 	// Proxy or Load Balancer routing configuration generated from HTTPRoutes
 	// MUST prioritize matches based on the following criteria, continuing on
 	// ties. Across all rules specified on applicable Routes, precedence must be
-	// given to the match with the largest number of:
+	// given to the match having:
 	//
-	// * Characters in a matching "Exact" path match
-	// * Characters in a matching "Prefix" path match
-	// * Header matches.
-	// * Query param matches.
+	// * "Exact" path match.
+	// * "Prefix" path match with largest number of characters.
+	// * Method match.
+	// * Largest number of header matches.
+	// * Largest number of query param matches.
 	//
 	// Note: The precedence of RegularExpression path matches are implementation-specific.
 	//
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -1509,21 +1509,21 @@ spec:
                         request. \n Proxy or Load Balancer routing configuration generated
                         from HTTPRoutes MUST prioritize matches based on the following
                         criteria, continuing on ties. Across all rules specified on
-                        applicable Routes, precedence must be given to the match with
-                        the largest number of: \n * Characters in a matching \"Exact\"
-                        path match * Characters in a matching \"Prefix\" path match
-                        * Header matches. * Query param matches. \n Note: The precedence
-                        of RegularExpression path matches are implementation-specific.
-                        \n If ties still exist across multiple Routes, matching precedence
-                        MUST be determined in order of the following criteria, continuing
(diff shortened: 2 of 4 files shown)


robscott:
Thanks @gauravkghildiyal!

/ok-to-test
/cc @arkodg @mlavacca @sunjayBhatia

apis/v1beta1/httproute_types.go
```diff
@@ -164,6 +164,7 @@ type HTTPRouteRule struct {
 	// * Characters in a matching "Prefix" path match
 	// * Header matches.
 	// * Query param matches.
+	// * Method matches.
```
arkodg, line 167: curious why its after `Query param` ? is it to maintain backwards compatibility ?

howardjohn: +1. I would put it before "Headers" personally

sunjayBhatia: (also seems like the language before this list needs to be amended, currently it reads `precedence must be given to the match with the largest number of:`, and adding method matching in here doesn't really fit)

robscott: @gauravkghildiyal at today's meeting I think we had consensus that this should fit right above header matches. Can you update that and the corresponding conformance test?

gauravkghildiyal: Thanks for the info. Updated!

apis/v1beta1/httproute_types.go
```diff
@@ -158,12 +158,13 @@ type HTTPRouteRule struct {
 	// Proxy or Load Balancer routing configuration generated from HTTPRoutes
 	// MUST prioritize matches based on the following criteria, continuing on
 	// ties. Across all rules specified on applicable Routes, precedence must be
-	// given to the match with the largest number of:
+	// given to the rule having:
 	//
-	// * Characters in a matching "Exact" path match
```
gauravkghildiyal, line 163: Please pay attention to the rewrite here -- I felt the "number of characters in exact path match" may not have been entirely appropriate. Please do point out if that had some relevance, which I'm missing, and I'll correct the statements.

mlavacca: I think that this re-writing related to the exact path match is more accurate 👍

apis/v1beta1/httproute_types.go
```diff
@@ -158,12 +158,13 @@ type HTTPRouteRule struct {
 	// Proxy or Load Balancer routing configuration generated from HTTPRoutes
 	// MUST prioritize matches based on the following criteria, continuing on
 	// ties. Across all rules specified on applicable Routes, precedence must be
-	// given to the match with the largest number of:
+	// given to the rule having:
```
mlavacca, line 161: I think that `given to the match having` is more accurate than `given to the rule having` wording, as the field specifying the `HTTPRouteMatch` is a match actually, not a rule.

gauravkghildiyal: Thanks @mlavacca. Yes that is indeed more appropriate.

mlavacca (approved):
Thanks @gauravkghildiyal!

sunjayBhatia (approved):

apis/v1beta1/httproute_types.go
youngnick: The reason for the "number of characters" part is to handle the case of two rules with Exact path matching - so the longer one will win in that case.

youngnick: Ah right, I think I missed that nuance. I withdraw my objection. 😄

robscott:
Thanks @gauravkghildiyal!

/lgtm
/approve
/cherry-pick release-0.7

PR #2062, opened 2023-05-25
chore: conformance profiles API version and channel

mlavacca:
**What type of PR is this?**

/kind gep
/area conformance

In support of https://github.com/kubernetes-sigs/gateway-api/issues/1709, the GEP-1709 has been updated with content on how to detect the `gatewayAPIVersion` and the `gatewayAPIChannel` and how to check their consistency.

```release-note
NONE
```

### DIFF
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -257,7 +257,7 @@ move on to [certification](#certification) to report the results.
 
 The certification is related to a specific API version and a specific channel,
 therefore such information must be included in the final report.
-At `ExperimentalConformanceTestSuite.Setup` time, the conformance profile
+At test suite setup time, the conformance profile
 machinery gets all the CRDs with the field `.spec.group` equal to
 `gateway.networking.k8s.io`, and for each of them checks the annotations
 `gateway.networking.k8s.io/bundle-version` and
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -268,9 +268,28 @@ there are CRDs with different channels, the certification fails specifying
 that it's not possible to run the tests as there are different Gateway API
 channels installed in the cluster. If all the Gateway API `CRD`s have the same
 version and the same channel, the tests can be run and the detected version
-and channel will be set in the `GatewayAPIVersion` and `gatewayAPIChannel` fields of the
-final report. Furthermore, the suite must run all the experimental tests when
-the channel is `experimental`, and the related features are enabled.
+and channel will be set in the `GatewayAPIVersion` and `gatewayAPIChannel`
+fields of the final report. Furthermore, the suite must run all the
+experimental tests when the channel is `experimental`, and the related features
+are enabled.
+
+In addition to the `CRD`s version, the suite needs to check its version in
+relation to the `CRD`s one. To do so, a new `.go` file containing the current
+Gateway API version is introduced in the project and compiled with the
+conformance profile suite:
+
+```go
+const GatewayAPIVersion = "0.7.0"
+```
+
+At test suite setup time the conformance profile suite
+checks the `CRD`s version and the suite version; if the two versions differ,
+the certification fails.
+A new generator will be introduced in the project to generate the
+aforementioned `.go` file starting from a VERSION file contained in the root
+folder. Such a VERSION file contains the semver of the latest release and is
+manually bumped at release time. The script hack/verify-all.sh will be updated
+to ensure the generated `.go` file is up to date with the VERSION file.
 
 ### Certification
 
--- a/geps/gep-1709.md
+++ b/geps/gep-1709.md
@@ -253,6 +253,25 @@ move on to [certification](#certification) to report the results.
 [go]:https://go.dev
 [lib]:https://pkg.go.dev/sigs.k8s.io/gateway-api@v0.6.2/conformance/utils/suite
 
+### Gateway API version and channel
+
+The certification is related to a specific API version and a specific channel,
+therefore such information must be included in the final report.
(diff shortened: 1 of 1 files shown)


shaneutt (approved):
/approve
/hold to allow others to take a look
/cc @sunjayBhatia @arkodg @robscott

arkodg (approved):

shaneutt:
test suite version tracking is a good point.

shaneutt:
> one issue I could see arising is how to handle an outdated set of CRDs in the cluster in comparison w/ the test suite (e.g. v0.X.Y CRDs vs. running the v1.A.B tests)
> 
> I don't expect too much in the way of behavior _change_ between versions going forward since the API is stabilizing, but things that are considered experimental in older versions may no longer be as the tests/versions progress, which could maybe cause a test failure/reporting issue
> 
> I wonder if in addition to this we should consider writing in the report what version the _test suite_ is from and discouraging too much version skew (or maybe not allowing any for certified reports)

Test suite version tracking is a good point. I think we could incorporate that here, unless you'd like to do a follow up PR to add this and any other thoughts you have on the proposal so far @sunjayBhatia?

sunjayBhatia:
> Test suite version tracking is a good point. I think we could incorporate that here, unless you'd like to do a follow up PR to add this and any other thoughts you have on the proposal so far @sunjayBhatia?

If @mlavacca wants to add it here that's great, otherwise I'm happy to make a follow-up 👍🏽

mlavacca:
> If @mlavacca wants to add it here that's great, otherwise I'm happy to make a follow-up 👍🏽

Thanks for this good point @sunjayBhatia, I'll address it in this PR.

sunjayBhatia (approved):
looks great @mlavacca!

shaneutt:
Thanks for the review @arkodg and @sunjayBhatia!

I'll let one of my co-maintainers drop the final LGTM:

/cc @robscott @youngnick

geps/gep-1709.md
```diff
@@ -280,6 +318,7 @@ implementation:
   - @acme/maintainers
 date: "2023-02-28 20:29:41+00:00"
 gatewayAPIVersion: v0.7.0
+gatewayAPIChannel: standard
```
robscott, line 321: Does this mean that most implementations would need to submit separate reports for each Gateway API Channel? I'm not sure we gain any value from that since the experimental channel is always a superset of the standard channel, but I may be missing something here.

I agree that the version of tests that are being run is critical though, so I'm glad that's included here.

geps/gep-1709.md
```diff
@@ -253,6 +253,44 @@ move on to [certification](#certification) to report the results.
...
+that it's not possible to run the tests as there are different Gateway API
+channels installed in the cluster. If all the Gateway API `CRD`s have the same
+version and the same channel, the tests can be run and the detected version
+and channel will be set in the `GatewayAPIVersion` and `gatewayAPIChannel`
+fields of the final report. Furthermore, the suite must run all the
+experimental tests when the channel is `experimental`, and the related features
+are enabled.
+
+In addition to the `CRD`s version, the suite needs to check its version in
+relation to the `CRD`s one. To do so, a new `.go` file containing the current
```
robscott, line 277: What happens when there's some skew here? For example, most of our patch releases just include new conformance tests without meaningful changes to the API spec. Would these not run if the CRD version was v0.7.0 and the test version was either v0.7.1 or v0.7.0-dev?

geps/gep-1709.md
```diff
@@ -253,6 +253,44 @@ move on to [certification](#certification) to report the results.
...
+conformance profile suite:
+
+```go
+const GatewayAPIVersion = "0.7.0"
+```
+
+At test suite setup time the conformance profile suite
+checks the `CRD`s version and the suite version; if the two versions differ,
+the certification fails.
+A new generator will be introduced in the project to generate the
```
robscott, line 288: Tiny nit: The line lengths feel a bit weird here, I can't tell if 288 is intended to start a new paragraph, but it is not doing that when the markdown is rendered.

robscott (commented):
Thanks @mlavacca! Mostly LGTM, just some questions.

geps/gep-1709.md
mlavacca: Since the experimental channel is a superset of the standard one, we could treat the report in the same way, which means being conformant to the experimental channel implies being conformant to the standard one. So that you just need to submit the experimental report to declare the standard conformance as well. WDYT?

geps/gep-1709.md
mlavacca: That makes sense, we could allow a skew within the same minor version and block conformance where the CRDs and the suite have different minor versions.

shaneutt: Is this a situation where we should start as strict as possible, and ease off _if_ we get complaints?

geps/gep-1709.md
robscott: That's kinda what I'm thinking - we may just want to recommend that implementations that support experimental features run conformance tests with experimental CRDs. I'm not sure that we need to actually state that in the report, but don't feel strongly either way.

geps/gep-1709.md
robscott: I think that's also fine. I guess this doesn't prevent running conformance tests, just generating a conformance report, right? I think it's fine to limit conformance reports to tightly matching versions, but I don't want to prevent running the latest set of conformance tests based on this.

mlavacca: Yep, running conformance tests won't be affected, only generating reports.

robscott:
Thanks @mlavacca!

/lgtm

PR #2066, opened 2023-05-26
conformance profiles CLI

### LINKED ISSUE #1990: CLI for the experimental version of the conformance test suite

Add a CLI for the experimental version of the conformance test suite (which provides prototypical [Conformance Profiles](#1709) capabilities) such that non-Golang implementations can run the test suite. 

**Why this is needed**:

It helps to complete the prototype and makes it easier for other implementations to try the experimental version out and provide feedback.

mlavacca:
**What type of PR is this?**

/kind feature
/area conformance

Fixes #1990

```release-note
The experimental conformance profile suite can now be added as a stand-alone cli and by means of `go test`.
```


cmd/certification/cmd/root.go
```diff
@@ -0,0 +1,199 @@
...
+var (
+	k8sClientset *kubernetes.Clientset
+	restConfig   *rest.Config
+	configPath   string
+	labels       []string
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "certification",
```
shaneutt, line 50: So it looks like we're angling at having the `certification` as it's own CLI, separate from the test suite. I think this might be OK, but ultimately we might bake this right into the test suite right? We should decide if certification reporting stuff happens as a result of a `go test` or if we do in fact have a chained process.

mlavacca: Yeah, I started pursuing the separate CLI way, but actually, it seems that integrating it into the already existing suite suits our needs better, and it is a lower hanging fruit 👍

shaneutt:
/cc @sunjayBhatia @arkodg

Would appreciate your reviews of this as well if you have some time.

conformance/experimental_conformance_test.go
```diff
@@ -0,0 +1,130 @@
...
+		*flags.ImplementationVersion,
+		*flags.ImplementationContact,
+	)
+	if err != nil {
+		t.Fatalf("Error parsing implementation's details: %v", err)
+	}
+	conformanceProfiles := suite.ParseConformanceProfiles(*flags.ConformanceProfiles)
+
+	// if some conformance profiles have been set, then run the experimental conformance suite...
+	if conformanceProfiles.Len() > 0 {
```
arkodg, line 75: nit: maybe better if 
```
	} else { // ...otherwise run the standard conformance suite.
 		t.Run("standard conformance tests", TestConformance)
 	}
```
is brought to the top or this logic is moved to a separate function 	to improve readability ?

arkodg:
hey @mlavacca, this looks good !
would be great if you could share the o/p snippet when run again any implementation

conformance/experimental_conformance_test.go
mlavacca: Yeah, I agree it can be better in terms of readability. I'll do something to improve it 👍

mlavacca: I moved a bunch of lines in a separate function. Let me know if it looks better now

conformance/experimental_conformance_test.go
```diff
@@ -0,0 +1,151 @@
...
+	confv1a1 "sigs.k8s.io/gateway-api/conformance/apis/v1alpha1"
+	"sigs.k8s.io/gateway-api/conformance/tests"
+	"sigs.k8s.io/gateway-api/conformance/utils/flags"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+var (
+	cfg                 *rest.Config
+	clientset           *kubernetes.Clientset
+	k8sClient           client.Client
```
shaneutt, line 44: nit: these symbol names for these different clients might be _slightly_ confusing for future code readers. Perhaps we can provide names that speak more to the fact that these are different, and why they're different?

```suggestion
	k8sClientSet           *kubernetes.Clientset
	mgrClient           client.Client
```

Not sure if that's a lot better, but some more specificity in the name I think could help... OR maybe we don't actually need to have two separate clients? :thinking:

conformance/experimental_conformance_test.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func writeReport(log func(...any), report confv1a1.ConformanceReport, output string) error {
+	rawReport, err := yaml.Marshal(report)
+	if err != nil {
+		return err
+	}
+
+	if output != "" {
+		if err = os.WriteFile(output, rawReport, 0666); err != nil {
```
shaneutt, line 144: The file doesn't need to be _writable_ by group/world, we can probably just give write permissions to the file owner (current user for the test program):

```suggestion
		if err = os.WriteFile(output, rawReport, 0644); err != nil {
```

conformance/utils/flags/experimental_flags.go
```diff
@@ -0,0 +1,35 @@
...
+import "flag"
+
+var (
+	ImplementationOrganization = flag.String("organization", "", "Implementation's Organization to issue conformance to")
+	ImplementationProject      = flag.String("project", "", "Implementation's project to issue conformance to")
+	ImplementationUrl          = flag.String("url", "", "Implementation's url to issue conformance to")
+	ImplementationVersion      = flag.String("version", "", "Implementation's version to issue conformance to")
+	ImplementationContact      = flag.String("contact", "", "Comma-separated list of contact information for the maintainers")
+	ConformanceProfiles        = flag.String("conformance-profiles", "", "Comma-separated list of the conformance profiles to run")
+	ReportOutput               = flag.String("report-output", "STDOUT", "The file where to write the conformance report")
```
shaneutt, line 34: Passing the string `"STDOUT"` doesn't have the effect of actually writing to STDOUT, it will actually just print the output to a file called `STDOUT` which I assume we don't want. We need to add logic to handle the STDOUT case.

conformance/utils/suite/experimental_suite.go
```diff
@@ -282,3 +290,50 @@ func (suite *ExperimentalConformanceTestSuite) Report() (*confv1a1.ConformanceRe
...
+		return nil, errors.New("implementation's version can not be empty")
+	}
+	contacts := strings.SplitN(contact, ",", -1)
+	if len(contacts) == 0 {
+		return nil, errors.New("implementation's contact can not be empty")
+	}
+
+	// TODO: we may want to add validation for url
+	// TODO: we may want to add validation for version
+	// TODO: we may want to add validation for contact (e.g., @someone or someone@acme.org)
```
shaneutt, line 316: These would be a good thing to capture as follow up issues, and then link those issues here so they don't end up getting out of band.

conformance/utils/suite/experimental_suite.go
```diff
@@ -282,3 +290,50 @@ func (suite *ExperimentalConformanceTestSuite) Report() (*confv1a1.ConformanceRe
...
+	res := sets.Set[ConformanceProfileName]{}
+	if p == "" {
+		return res
+	}
+
+	for _, value := range strings.Split(p, ",") {
+		res.Insert(ConformanceProfileName(value))
+	}
+	return res
+}
```
shaneutt, line 339: Consider this completely optional, but FWIW for small functions like this I tend to like to declare the return value in the function signature:

```suggestion
func ParseConformanceProfiles(p string) (res sets.Set[ConformanceProfileName]) {
	if p == "" {
		return
	}

	for _, value := range strings.Split(p, ",") {
		res.Insert(ConformanceProfileName(value))
	}
	return
}
```

conformance/experimental_conformance_test.go
arkodg: looks good, thanks !

arkodg (approved):
LGTM thanks !

conformance/utils/flags/experimental_flags.go
mlavacca: Yeah, this is a stupid leftover from a previous iteration. I'm going to remove this default. The result is always printed on logs, and optionally on the given file.

conformance/experimental_conformance_test.go
mlavacca: It looks like we need both of them for this reason:

https://github.com/mlavacca/gateway-api/blob/af424c1dd38d8d8eed8674be54750943475b7d9a/conformance/conformance_test.go#L65

conformance/utils/suite/experimental_suite.go
mlavacca: https://github.com/kubernetes-sigs/gateway-api/issues/2178

shaneutt (approved):
Thanks @mlavacca!

/approve
/lgtm
/unhold

Issue #91, opened 2020-02-14
GEP: Client Certificate Verification for Gateway Listeners


jpeach:
**What would you like to be added**:

The ability for a HTTPS (or TLS generally) endpoint to require that the client present a certificate that can be validated according to some configurable policy.

**Why is this needed**:

As an application developer, I want to restrict access to my application to a certain audience of clients. The audience is defined by one or more of

- a collection of specific TLS certificates (maybe by hash)
- a collection of subject names in certificates
- a collection of certificates issued by a specific (unique) CA

I want the infrastructure to guarantee that I only receive client traffic that originates from this audience.

/kind user-story

bowei:
Is this mTLS? Can you add to the issue title, as that is a commonly known term around the K8S community.

youngnick:
It is mutual TLS in the sense that the client and the server perform mutual TLS, but 'mTLS' also has some connotations about speed and automation of certificate rotation that we probably don't want to bring in here.

I say this because when we had an issue to add 'mTLS' to Contour, people started asking straight away if Contour was now a service mesh. Which it is not. But the term is overloaded.

hbagdi:
/remove-lifecycle rotten

#268 should close this.

alanchrt:
Adding my comments from Slack regarding some configuration options that would be nice:

> either deny or don't deny requests from clients who don't present valid certs (useful when you want to show a "logged out state" based on application logic in your upstream)

> an option to pass through the subject/sans from the certificate to your upstream (the identity of the client), maybe in a header? not sure if there's a standard there. this could probably be the default.. not sure if it would matter if you still passed through the client identity and the upstream didn't care.

> this could let you configure things like "i'll take requests in my upstream even if they don't present a valid cert, but if they do present a valid cert, i want to know that the identity was alan so i can decide his access level. and if there's no cert, i want to show them logged out state."

The `ingress-nginx` annotations seem like a really good reference point for which configuration options might be nice, as well as the names and contents of headers to pass back to the upstream:

https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#client-certificate-authentication

I think this one largely gets at some of the use cases I was describing above:

```yaml
nginx.ingress.kubernetes.io/auth-tls-verify-client: optional_no_ca
```

youngnick:
It's also important to remember that, if the following are true:
- you have a single proxy routing multiple SNIs
[...]

PR #2080, opened 2023-05-31
GEP: Client Certificate Validation for TLS terminating at the Gateway Listener

arkodg:
/kind gep

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/91

### DIFF
--- a/geps/gep-91.md
+++ b/geps/gep-91.md
@@ -1,0 +1,22 @@
+# GEP-91: Client Certificate Validation for TLS terminating at the Gateway Listener
+
+* Issue: [#91](https://github.com/kubernetes-sigs/gateway-api/issues/91)
+* Status: Provisional
+
+(See definitions in [GEP Status][/contributing/gep#status].)
+
+## TLDR
+
+This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
+(Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).
+
+## Goals
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trusted anchor to validate the certificates presented by the client.
+
+## Non-Goals
+- Define other fields that can be used to verify the client certificate such as the Cerificate Hash or Subject Alt Name. 
+
+## References
+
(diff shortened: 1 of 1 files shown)


arkodg:
will continue adding more content in the GEP once Ive received broad consensus on the goals & non goals

geps/gep-91.md
```diff
@@ -0,0 +1,22 @@
...
+
+(See definitions in [GEP Status][/contributing/gep#status].)
+
+## TLDR
+
+This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
+(Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).
+
+## Goals
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trusted anchor to validate the certificates presented by the client.
```
frankbu, line 14: Isn't the existing `certificateRefs` field where the CA certificate would be specified? Is the idea that this GEP will add a field that can be used to specify something like a client validation cacert key in a `cacertificateRefs` secret, for example? Maybe adding an API section with a proposal would be helpful to start the discussion?

frankbu: fwiw, this goal of adding a way to config mTLS termination in a Gateway is valuable and very contained scope IMO.

robscott: Agree that feels like the inevitable placement here. Let's get this in and then we can work on the implementation details.

robscott:
Thanks @arkodg! Will give @shaneutt and @youngnick a bit more time to review this one, but don't think there's anything controversial here.

/approve

frankbu:
@robscott @shaneutt @youngnick 

Hey guys, since this is no more than a simple (I believe agreed to) problem statement that has been stagnant for a couple of months, can we either get this merged now or otherwise update it with an API proposal so we can start discussing the solution?

@argkodg do you have an API for this that you are waiting to propose?

robscott:
Thanks for the ping on this one @frankbu! I don't think there's anything controversial in the GEP at this point. I've already approved, would like @shaneutt or @youngnick to sign off on this though.

/release-note-none

youngnick:
Yeah, this is pretty tightly scoped as written, so I'm supportive.

/lgtm

PR #2084, opened 2023-06-01
Update kustomization.yaml to include reference grant

howardjohn:
**What type of PR is this?**

/kind bug

Fixes #

```release-note
Added the missing ReferenceGrant resource the kustomization.yaml for the standard channel
```


robscott:
Thanks @howardjohn!

Not sure if we'll have a v0.7.2, but if we do, this should be in it.

/cherry-pick release-0.7
/lgtm
/approve

PR #2092, opened 2023-06-05
apis/v1beta1: Host header must not be modified

### LINKED ISSUE #2091: Further clarification on hostname matching and forwarding when Host header includes a port
In https://github.com/kubernetes-sigs/gateway-api/pull/1980 we added clarification that implementations *MUST* ignore a port included in the HTTP Host header.

Implicitly, the test added also asserts that the backend will receive the original Host header with the port intact (e.g. the test/request with `Host: very.specific.com:1234` will forward that Host header unchanged to the backend).

Is this a desired behavior? If so, we should explicitly document it in the API spec so users/implementers expect it. If not, again we should document it.

**Why this is needed**:
Specificity in the API documentation/tests.

**Further Notes**:
- Contour at the moment (and I believe at least Envoy Gateway) strips any port from the host header and does not forward it to the backend.
- We added this port strip behavior in Contour to ensure we can do wildcard hostname matching in Envoy, Envoy only allows a wildcard prefix *or* suffix hostname match, `*.foo.com` or `foo.com:*` but not `*.foo.com:*`. We can likely come up with an implementation that fulfills whatever spec, but just noting this as Envoy-based implementations may not be conformant with this test for a bit depending on how they are implemented and what is decided. See [here](https://github.com/projectcontour/contour/pull/5353#issuecomment-1577203418)
-  It looks like Istio has a similar issue as described above, since Envoy does not allow double-ended wildcard hostname matching, from [here](https://github.com/istio/istio/blob/ef2bd93793f37190ad5c1f9d0efbde214faa683f/pilot/pkg/networking/core/v1alpha3/gateway.go#L1041-L1059)

sunjayBhatia:
**What type of PR is this?**
/kind cleanup

Adds clarification to the spec to go along with existing conformance test from #1980.

Fixes #2091

```release-note
Clarify that implementations must not modify HTTP Host header. Adds specificity alongside spec that port in Host header must be ignored when matching on host.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -59,7 +59,8 @@ type HTTPRouteSpec struct {
 	// Hostnames defines a set of hostname that should match against the HTTP Host
 	// header to select a HTTPRoute used to process the request. Implementations
 	// MUST ignore any port value specified in the HTTP Host header while
-	// performing a match.
+	// performing a match and (absent of any applicable header modification
+	// configuration) MUST forward this header unmodified to the backend.
 	//
 	// Valid values for Hostnames are determined by RFC 1123 definition of a
 	// hostname with 2 notable exceptions:
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -55,12 +55,14 @@ spec:
                 description: "Hostnames defines a set of hostname that should match
                   against the HTTP Host header to select a HTTPRoute used to process
                   the request. Implementations MUST ignore any port value specified
-                  in the HTTP Host header while performing a match. \n Valid values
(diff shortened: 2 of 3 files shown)


shaneutt (approved):
/approve

/cc @robscott @youngnick

howardjohn (approved):

apis/v1beta1/httproute_types.go
```diff
@@ -59,7 +59,8 @@ type HTTPRouteSpec struct {
 	// Hostnames defines a set of hostname that should match against the HTTP Host
 	// header to select a HTTPRoute used to process the request. Implementations
 	// MUST ignore any port value specified in the HTTP Host header while
-	// performing a match.
```
arkodg, line 62: suggest
```
// MUST ignore any port value specified in the HTTP Host header while
// performing a match, without modifying it.
```

robscott:
Thanks @arkodg!

/approve
/hold cancel

PR #2105, opened 2023-06-09
Use correct constant for listener reason "Programmed"

nathancoleman:
**What type of PR is this?**
/kind cleanup
/area conformance

Simply switches a constant used in conformance testing to the correct one. They both have the same value, so there's no functional impact.

N/A

```release-note
NONE
```


robscott:
Thanks @nathancoleman!

/lgtm
/approve
/ok-to-test

PR #2106, opened 2023-06-09
conformance: use distinct resource names for test isolation

nathancoleman:
**What type of PR is this?**
/kind bug
/kind flake
/area conformance

Two conformance tests are currently using the same name for their Gateway resource. This leads to contention when one test is actively deleting its Gateway and the other test is actively modifying the same Gateway. The second test then attempts to retrieve the Gateway to verify expected conditions; however, the deletion has completed by that time, causing the second test to fail as it can't find the Gateway in question.

The changes in this PR use distinct names for the Gateway in each test so that this contention does not occur. Tests reliably pass with these changes included using the Consul implementation of the Gateway API deployed on GKE.

```shell
$ cd ./conformance && gotest -v -timeout 10m ./ --gateway-class consul-api-gateway
```

Before:
CleanShot 2023-06-09 at 16 24 08@2x

After:
image

N/A

```release-note
NONE
```


robscott (commented):
Thanks @nathancoleman!

/approve

PR #2116, opened 2023-06-14
conformance: Ignore state of terminating pods

### LINKED ISSUE #2102: conformance: Improve run time of test suite by ignoring terminating pods
**What happened**:
Conformance tests that add or modify Gateways use the `NamespacesMustBeReady` helper to ensure all pods are ready before starting test logic. This is to ensure any test fixture apps and control/data plane pods are up and running before starting to make assertions.

However, a side effect of this is that as Gateways are deleted from prior tests in the same namespace (`gateway-conformance-infra` is a common one) as a running test and control/data plane components are spun down, `NamespacesMustBeReady` can get hung up waiting for *terminating* pods to be ready, significantly increasing the run time of the suite.

The `NamespacesMustBeReady` helper to ignore terminating pods to reduce the time it takes to run the conformance suite.

Another option would be to run each test in it's own namespace, but that seems against the pattern of where the tests have been going up till now.

Run the conformance test suite against an implementation like Contour, and you will see tests waiting for pods to be Ready that correspond to control/data plane instances from other tests.

Can use deletion timestamp on pods to check for those that are terminating, since it is not a status condition.

sunjayBhatia:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind test
/kind flake
/area conformance

In NamespacesMustBeReady helper that checks if all pods in a namespace are ready, used to gate test runs to ensure we're not making requests to backends that are still coming up, etc.

As tests create/delete Gateways (and associated infrastructure components, e.g. control/data planes) in commonly used namespaces, there are often Terminating pods that the helper waits for, which significantly increases test runtime and causes timeouts.

Instead, if a pod has a deletion timestamp, indicating it is being shut down, we ignore that when checking the pods in a namespace.

Fixes #2102

```release-note
NONE
```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -220,7 +220,8 @@ func NamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig config.T
 			}
 			for _, pod := range podList.Items {
 				if !findPodConditionInList(t, pod.Status.Conditions, "Ready", "True") &&
-					pod.Status.Phase != v1.PodSucceeded {
+					pod.Status.Phase != v1.PodSucceeded &&
+					pod.DeletionTimestamp == nil {
 					t.Logf("%s/%s Pod not ready yet", ns, pod.Name)
 					return false, nil
 				}


robscott:
This makes sense to me based on the discussion in #2102. Would like someone from conformance reviewers to LGTM though.

/cc @arkodg @michaelbeaumont @mlavacca @xunzhuo

arkodg (approved):

sunjayBhatia:
If we're doing another 0.7.x patch release, can we consider this for a backport?

PR #2128, opened 2023-06-19
Improve Policy discoverability

youngnick:
**What type of PR is this?**
/kind cleanup
/kind documentation
/kind gep

This PR updates GEP-713 - Policy Attachment - to clarify how implementations using the pattern should handle the discoverability problems it creates.

Some solutions need further GEPs - these are marked as EXPERIMENTAL DO NOT USE (to prevent similar problems from arising in the future).

Additionally, this PR removes most content from the existing "Policy Attachment" documentation page and adds a link to this GEP so that there's only one place to keep up to date.

Updates #2014 

```release-note
The Policy Attachment pattern has new requirements and recommendations to help with Policy discoverability.
```


youngnick:
Most of the new content is best viewed in the deployment preview:

https://deploy-preview-2128--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/#status-and-the-discoverability-problem

youngnick:
I've just removed the change to the site page about Policy Attachment, I'll split that into a seperate PR.

geps/gep-713.md
```diff
@@ -643,6 +1129,14 @@ const (
 )
 ```
 
+#### On targeted resources
+
+Implementations that use Policy objects MUST put a Condition into `status.Conditions` of any objects affected by a Policy.
```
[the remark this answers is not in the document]
youngnick: yes, I'll update this wording to make it clear that any implementation that has Policy affecting an object should ensure that the condition is set to `status: true`, and if it's already present with that status, leave it alone.

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+
+(There’s also another use case to consider, in that Charlie should have been able
+to see that the Policy on the namespace was in use in many places before deleting
+it.)
+
+To put this another way, Policy Attachment is effectively adding a fourth Persona,
+the Policy Admin, to Gateway API’s persona list, and without a solution to the
+discoverability problem, their actions are largely invisible to the Application
+Developer. Not only that, but their concerns cut across the previously established
+levels.
```
frankbu, line 719: I'm wondering if this problem would be less severe if there were certain restrictions on where and how policy attachments are used?

Seems like one thing that should always be possible is that the owner of a resource should be able to override any inherited policy attachments. Does it really make sense that a policy attachment can override an explicit setting in a resource? What is the use case for having explicit settings in the user's resource that get ignored and changed under the covers because of some policy attachment somewhere?

I'm not sure what properties of a resource the spec says can be controlled by inherited policy attachments, but I would say there "should be" only two things:
1. default values of fields in the resource (first class API fields)
2. fields that can be set with direct policy attachments (these are fields that presumably have been added to support things that are not directly supported in the resources API itself)

If that was the case, then the owner of a resource would always be able to explicitly override any inherited policy attachments that don't work, either in the resource itself or in a direct policy attachment sitting next to the resource.

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+
+In the parable, if Ana and Charlie had known that there were Policies affecting
+the relevant object, then they could have gone looking for the relevant Policies
+and things would have played out differently. If they knew which Policies, they
+would need to look less hard, and if they knew what the settings being applied
+were, then the parable would have been able to be very short indeed.
+
+(There’s also another use case to consider, in that Charlie should have been able
+to see that the Policy on the namespace was in use in many places before deleting
+it.)
```
frankbu, line 713: Shouldn't that be a default assumption if there are other resources in the namespace?

youngnick:
Thanks for the comments @frankbu, I'm about to go on leave until July 18, so I will address them when I'm back.

I'd encourage others to take a look at this proposal too in the meantime.

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+**Ana**: _Wish I could. Wind’s died down, though, and it'll be dark soon.
+Just gonna head home._
+
+**Charlie**: _Ouch. Sorry to hear that._ 😐
+
+One more look out at the lake.
+
+**Ana**: _Thanks for the help. Wish we’d found better answers._ 😢
+
+### The Problem, restated
```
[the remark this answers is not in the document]
youngnick: The problem isn't that the Policy object mitigated the effects, it's that Ana had no way of knowing that's what was happening.

As I say below, there are really three things that any owner of an object that's affected by a Policy wants to know (and the later ones mostly imply that you know the earlier ones:
- _That_ some Policy is affecting a particular object
- _Which_ Policy is (or Policies are) affecting a particular object
- _What_ settings in the Policy are affecting the object.

As we go on to discuss later, there are issues with just dropping these details into the status of objects, and so this is trying to give us a range of options based on how bad the need is, and what resources are available.

geps/gep-713.md
youngnick: Yes, but again, it's easy to miss in a stressful situation. Explicitly _telling_ people this is better than leaving it implicit.

geps/gep-713.md
youngnick: Being able to override specific settings can be _very_ useful in clusters with strict rules about things, although I agree that it's probably better in those cases to reject the create or change that would have an undesirable setting using a Validating Admission Controller - faster feedback is _much_ better than silently changing a value.

But, this override need comes from direct experience @robscott and I have had with previous iterations of similar ideas, sysadmins (in Gateway API's personas, Cluster Admins or even Infra Admins) sometimes want to override settings so that users can't take certain actions, and this is a mechanism to allow that to happen.

geps/gep-713.md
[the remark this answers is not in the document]
youngnick: I've just updated this, PTAL. Resolving this though.

david-martin (commented):
Great to see the latest updates here.

This is most definitely a separate GEP, but I'm wondering about the concept of a policy implementer/provider that is separate from the gateway provider.
Would that decoupling make sense? Perhaps this has been discussed before.

Asking in the context of seeing the GatewayClass extension types mentioned and how that might work.
For example, if I'm an expert in API caching, I'd like to write a controller for reconciling CachePolicy resources, but I don't want to/not able to provide a full gateway api impelmentation with my caching policy feature.
I'd like to register my controller in a standard way with a Gateway API provider, and then it lists the CachePolicy CRD in the GatewayClass status. 
The gateway provider then has a standard mechanism for integrating with my controller (or vice versa).

shaneutt (approved):
The reality of this effort is that we _must_ move forward because we are in a predicament for delivery. Given the reality and the need and the fact that this change is overall an improvement, I think we should accept this as the next iteration and keep iterating from here.

/approve

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+discoverability problem, their actions are largely invisible to the Application
+Developer. Not only that, but their concerns cut across the previously established
+levels.
+
+![Gateway API diagram with Policy Admin](images/713-the-diagram-with-policy-admin.png)
+
+
+From the Policy Admin’s point of view, they need to know across their whole remit
+(which conceiably could be the whole cluster):
+- _What_ Policy has been created
```
robscott, line 726: Needed for list formatting in our markdown parser
```suggestion
(which conceiably could be the whole cluster):

- _What_ Policy has been created
```

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+
+Which again, come down to discoverability, and can probably be addressed in similar
+ways at an API level to the Application Developer's concerns.
+
+An important note here is that a key piece of information for Policy Admins and
+Cluster Operators is “How many things does this Policy affect?”. In the parable,
+this would have enabled Charlie to know that deleting the Namespace Policy would
+affect many other people than just Ana.
+
+### Problems solutions need to solve
```
robscott, line 738: Open to other wording, but `Problems solutions` was a bit confusing for me

```suggestion
### Problems we need to solve
```

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+
+Minimizing the object updates is done by only having an update be necesary when
+the affected object starts or stops being affected by a Policy, rather than if
+the Policy itself has been updated.
+
+There is already a similar Condition to be placed on _Policy_ objects, rather
+than on the _targeted_ objects, so this solution is also being included in the
+Condiions section below.
+
+#### GatewayClass status Extension Types listing
```
robscott, line 915: It might be useful to differentiate between what's already been defined by the GEP and is already fairly solidified (the two solutions above this) with what is still provisional/in progress and will require future GEPs.

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+Policy, and other extension types, like paramsRef targets, filters, and so on). 
+
+This is going to be tracked in its own GEP, https://github.com/kubernetes-sigs/gateway-api/discussions/2118
+is the initial discussion. This document will be updated with the details once
+that GEP is opened.
+
+##### Design Considerations
+
+This solution:
+- is low cost in terms of apiserver updates (because it's only on the GatewayClass,
```
robscott, line 929:
```suggestion
This solution:

- is low cost in terms of apiserver updates (because it's only on the GatewayClass,
```

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+- provides a standard place for all users to look for relevant objects
+- ties in to the Conformance Profiles design and other efforts about GatewayClass 
+  status
+
+#### Standard status stanza
+
+This solution is IN PROGRESS and so is not binding yet.
+
+This solution should also be considered experimental and needs to be accepted by
+the community. (This comment will be removed before merge).
```
robscott, line 940: Don't forget to remove this line

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
+
+Ideally, these status updates SHOULD be in a separate, lower-priority queue than
+other status updates or similar solution.
+
+#### PolicyBinding resource
+
+This solution is IN PROGRESS and so is not binding yet.
+
+This solution should also be considered experimental and needs to be accepted
+by the community. (This comment will be removed before merge).
```
robscott, line 981: Don't forget to remove this.

geps/gep-713.md
```diff
@@ -649,6 +1135,29 @@ const (
...
+
+Implementations that use Policy objects MUST ensure that a Condition exists in
+the `status.Conditions` of any objects affected by a Policy. If that Condition is
+already set to `status: true`, then the implementation must ensure that
+the `observedGeneration` field matches the `metadata.generation` field of the
+object.
+
+Additionally, if an implemntation sees an update that changes the `status`` of this
+Condition from `false` to `true` _that it did not create_, it must not attempt
+to change the status back to `false`.
```
robscott, line 1148: Deploy preview struggles here, I think it will be fixed by this change, but worth double checking: https://deploy-preview-2128--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/#on-targeted-resources

```suggestion
Additionally, if an implementation sees an update that changes the `status` of this
Condition from `false` to `true` _that it did not create_, it must not attempt
to change the status back to `false`.
```

geps/gep-713.md
```diff
@@ -649,6 +1135,29 @@ const (
 )
 ```
 
+#### On targeted resources
+
+Implementations that use Policy objects MUST ensure that a Condition exists in
+the `status.Conditions` of any objects affected by a Policy. If that Condition is
+already set to `status: true`, then the implementation must ensure that
+the `observedGeneration` field matches the `metadata.generation` field of the
+object.
```
robscott, line 1144: I think this specific solution might be best represented by a follow-up GEP. There are a lot of edge cases here that are going to be hard to handle. For example it will be fairly straightforward for implementations to determine when they should _set_ this condition, but near impossible for them to determine when they should _unset_ the condition, assuming more than one implementation is active in the cluster.

geps/gep-713.md
```diff
@@ -560,50 +560,536 @@ ties:
...
-This is currently a _very_ open question. A discussion is ongoing at
-[#1531](https://github.com/kubernetes-sigs/gateway-api/discussions/1531), and this
-GEP will be updated with any outcomes.
+the API structure defined above and add a `gateway.networking.k8s.io/policy: true`
+label to the CRD.
 
 ### Conditions
-Controllers using the Gateway API policy attachment model SHOULD populate the 
+
+Implementations using Policy objects MUST include a `spec` and `status` stanza, and the `status` stanza MUST contain a `conditions` stanza, using the standard Condition format.
```
robscott, line 1088: I may just be missing it, but I'm not finding any requirement in this GEP for implementations to set the `Accepted` condition on policy objects they are implementing. 

https://github.com/kubernetes-sigs/gateway-api/blob/f638d3fe75a69354b4a15b6c9e9e475c4e7d8062/apis/v1alpha2/policy_types.go#L51-L64

With that said, this also becomes problematic if multiple controllers are implementing the same policy resource, so maybe we need to adjust the guidance here to leave room for `conditions` stanzas to be nested within a `parents` list similar to what we do for Routes?

robscott (commented):
Thanks @youngnick! Lots of great updates here.

geps/gep-713.md
youngnick: Fixed, I fixed the typo as well. ("conceiably")

geps/gep-713.md
youngnick: I've updated the wording here and in the Solution Cookbook section, but I'll also update both with the "this needs a followup GEP" wording and change to Experimental.

geps/gep-713.md
youngnick: It was already there, but I've updated to clarify a bit.

geps/gep-713.md
```diff
@@ -560,52 +578,567 @@ ties:
...
+Additionally, the status of each solution is noted at the beginning of the section.
+
+#### Standard label on CRD objects
+
+Status: Required
+
+Each CRD that defines a Policy object MUST include a label that specifies that
+it is a Policy object.
+
+The label is `gateway.networking.k8s.io/policy: true`.
```
robscott, line 900: Non-blocking: Maybe we should accept/encourage different values here like `inherited` or `direct` to more clearly define the type of policy.

geps/gep-713.md
```diff
@@ -560,52 +578,567 @@ ties:
...
+plugin.
+
+##### Design considerations
+
+This is already part of the API pattern, but is being lifted to more prominience
+here.
+
+#### Standard status Condition
+
+Support: Experimental
```
robscott, line 914: Maybe `Provisional` would be a better term here since we're already using `Experimental` to indicate something that is released and part of the API. In general, I think it's going to be confusing to add a lot of new/provisional ideas to a GEP that is already released + labeled as "experimental". I'd personally prefer adding a lot of the net new content in different files, issues, or discussions to avoid overloading this one.

```suggestion
Status: Provisional
```

geps/gep-713.md
```diff
@@ -1,4 +1,4 @@
-# GEP-713: Metaresources and Policy Attachment_metaresource_
+# GEP-713: Metaresources and Policy Attachment
```
robscott, line 1: A general note, this GEP is _massive_. I think the overall size of it makes an already complicated topic even more overwhelming to understand and/or contribute to. One potential way to mitigate this problem would be to have a separate page that collects and documents problems with policy attachment and links out to potential solutions and how we're trying to improve things here. For example, we could keep something like https://github.com/kubernetes-sigs/gateway-api/discussions/2012 as an interactive forum for proposing solutions/improvements that would eventually turn into GEPs of their own, but the problem statements could be clearly stated somewhere in our docs... 

This is not blocking feedback, I just think we're going to need find some ways to either split this GEP up and/or cut some of the content out so this can be a bit more manageable + approachable.

geps/gep-713.md
```diff
@@ -560,52 +578,567 @@ ties:
...
+For any given Policy type, whether Direct Attached or Inherited, implementations
+will need to be able to _calculate_ the resultant set of Policy to be able to
+apply that Policy to the correct parts of their data plane configuration.
+However, _displaying_ that resultant set of Policy in a way that is straightforward
+for the various personas to consume is much harder.
+
+The easiest possible option for Application Developers would be for the
+implementation to make the full resultant set of Policy available in the status
+of objects that the Policy affects. However, this runs into a few problems:
+- The status needs to be namespaced by the implementation
```
robscott, line 814: I wish our markdown parsing didn't require new lines here, but unfortunately it does and we constantly miss it because it's not intuitive. https://deploy-preview-2128--kubernetes-sigs-gateway-api.netlify.app/geps/gep-713/#evaluating-and-displaying-resultant-policy

```suggestion
of objects that the Policy affects. However, this runs into a few problems:

- The status needs to be namespaced by the implementation
```

geps/gep-713.md
```diff
@@ -560,52 +578,567 @@ ties:
...
+all the problems above, the strategy this GEP proposes is to define a range of
+options for increasing the discoverabilty of Policy resources, and provide
+guidelines for when they should be used.
+
+It's likely that at some stage, the Gateway API CRDs will include some Policy
+resources, and these will be designed with all these discoverabiity solutions
+in mind.
+
+
+### Solution cookbook
```
robscott, line 885: Instead of expanding an already massive GEP, I think I'd prefer to link out to GEPs, issues, discussions, etc for potential solutions.

geps/gep-713.md
robscott: Same comment applies to solutions listed below as well.

geps/gep-713.md
```diff
@@ -649,6 +1182,33 @@ const (
 )
 ```
 
+#### On targeted resources
```
robscott, line 1185: Given the size of this GEP and the provisional nature of this idea, maybe we can just link out to somewhere here instead of duplicating the content.

geps/gep-713.md
```diff
@@ -560,52 +578,567 @@ ties:
...
+Implemenations SHOULD create an instance of a new `gateway.networking.k8s.io/EffectivePolicy`
+object when one or more Policy objects become relevant to the target object.
+
+The `EffectivePolicy` object MUST be in the same namespace as the object targeted
+by the Policy, and must have the _same name_ as the object targeted like the Policy.
+This is intended to mirror the Services/Endpoints naming convention, to allow for
+ease of discovery.
+
+The `EffectivePolicy` object MUST set the following information:
+- The name, namespace, apiGroup and Kind of Policy objects affecting the targeted
```
robscott, line 1041:
```suggestion
The `EffectivePolicy` object MUST set the following information:

- The name, namespace, apiGroup and Kind of Policy objects affecting the targeted
```

robscott (commented):
Thanks @youngnick! You've got some great ideas here, really good to seem them all articulated. I've got a few comments, mostly non-blocking. The overall theme is that I think this GEP has gotten too big and I think it would be good to split it out ASAP or at least add new content in new places. That's more of an organizational thing, but I think I'm aligned with all the major points in here.

/approve

geps/gep-713.md
youngnick: I agree, but I'd prefer to get this merged, so that we have the text stored _somewhere_, and then move it around after. This PR does not finish the work to tidy up Policy Attachment, but it's become a bit unwieldy, so I'd like to concentrate on getting it merged and fixing afterwards.

geps/gep-713.md
youngnick: See a below comment, but I think that breaking this up should be a separate piece of work (that I'll pick up straight away).

geps/gep-713.md
youngnick: Updated.

geps/gep-713.md
youngnick: I've logged #2307, and assigned it to myself.

robscott (commented):
Thanks @youngnick!

/lgtm

PR #2132, opened 2023-06-20
Move GEP-1282 to declined

### LINKED ISSUE #1282: GEP: Describe Backend Properties

youngnick:
**What type of PR is this?**

/kind cleanup
/kind gep

We are not going to complete GEP-1282, so it's better to make this formal, and decline it.

Fixes #1282 

```release-note
GEP-1282, Backend Properties, has been declined.
```

### DIFF
--- a/geps/gep-1282.md
+++ b/geps/gep-1282.md
@@ -1,7 +1,14 @@
 # GEP-1282: Describing Backend Properties
 
 * Issue: [#1282](https://github.com/kubernetes-sigs/gateway-api/issues/1282)
-* Status: Provisional
+* Status: Declined
+
+
+## Explanation for Declined status
+
+This GEP is declined because, after spending a lot of time discussing it, we felt that it was too large, and had too many crosscutting concerns.
+
+It's superseded for TLS Backend Properties by [GEP-1897](https://github.com/kubernetes-sigs/gateway-api/issues/1897).
 
 ## TLDR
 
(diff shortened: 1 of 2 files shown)


dprotaso (approved):

shaneutt (approved):

robscott:
Thanks @youngnick!

/lgtm

PR #2138, opened 2023-06-22
ci(linter): enable several linters

mmorel-35:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,8 @@
 run:
   timeout: 10m
   issues-exit-code: 1
+  max-issues-per-linter: 0
+  max-same-issues: 0
   tests: true
   skip-dirs-use-default: true
   modules-download-mode: readonly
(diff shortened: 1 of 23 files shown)


.golangci.yml
```diff
@@ -9,13 +11,19 @@ run:
 linters:
   fast: false
   enable:
-  - gofmt
```
arkodg, line 12: why is `gofmt` removed

.golangci.yml
```diff
@@ -9,13 +11,19 @@ run:
...
-  - govet
   - gomodguard
+  - govet
   - misspell
-  - exportloopref
+  - revive
+  - unconvert
+  - unparam
+  - unused
+  - whitespace
```
arkodg, line 26: can we also add `gosec`
https://github.com/securego/gosec

.golangci.yml
arkodg: sgtm, thanks for the clarification, linking the repo is case others are curious https://github.com/mvdan/gofumpt

.golangci.yml
mmorel-35: It’s added !

arkodg (approved):

robscott:
Thanks @mmorel-35!

/release-note-none
/lgtm
/approve

PR #2143, opened 2023-06-23
Conformance - Fix Makefile 

dprotaso:
**What type of PR is this?**

/kind cleanup
/area conformance

- Running make conformance with a flag (eg. -gateway-class=istio) would fail since the package `./conformance/utils` doesn't import the conformance suite test flags
- Allow test runners to pass set `GO_TEST_FLAGS` to customize the test invocation (used for debugging)

```release-note
NONE
```


shaneutt (approved):
Seems reasonable, thanks @dprotaso :+1: 

/approve
/ok-to-test
/lgtm

PR #2146, opened 2023-06-26
apis: document ParentRef functionality for GAMMA

### LINKED ISSUE #2139: Add GAMMA references to API Spec
Before the v0.8.0 release, we need to update the API spec to include GAMMA-specific guidelines for usage. I believe the primary difference here is ParentRef on Routes, but I may be missing something.

**Why this is needed**:
GAMMA is on track to graduate to experimental in v0.8.0, that means we need to include guidance for the use of this pattern in the API Spec.

/kind documentation
/area mesh
/assign @mikemorris

mikemorris:
**What type of PR is this?**
/kind documentation
/area mesh

Updates API documentation support for ParentRef targeting a Kubernetes Service, as proposed in [GEP-1426](https://gateway-api.sigs.k8s.io/geps/gep-1426/).

Fixes #2139
Service port targeting refs #1995 

**Links:**
- https://kubernetes.io/docs/concepts/services-networking/service/

```release-note
Adds support for ParentRef targeting a Kubernetes Service resource for mesh implementations.
```

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -21,10 +21,14 @@ import (
 )
 
 // ParentReference identifies an API object (usually a Gateway) that can be considered
-// a parent of this resource (usually a route). The only kinds of parent resources
-// with "Core" support are Service (only for implementations supporting the Mesh
-// conformance profile) and Gateway. This API may be extended in the future to
-// support additional kinds of parent resources.
+// a parent of this resource (usually a route). There are two kinds of parent resources
+// with "Core" support:
+//
+// * Gateway (Gateway conformance profile)
+// * Service (Mesh conformance profile)
+//
+// This API may be extended in the future to support additional kinds of parent
+// resources.
 //
 // The API object must be valid in the cluster; the Group and Kind must
 // be registered in the cluster for this reference to be valid.
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -46,11 +46,12 @@ type ParentReference struct {
 
 	// Kind is kind of the referent.
 	//
-	// Support: Core (Gateway)
+	// There are two kinds of parent resources with "Core" support:
 	//
-	// Support: Core (Service - mesh conformance profile only)
+	// * Gateway (Gateway conformance profile)
+	// * Service (Mesh conformance profile)
 	//
-	// Support: Implementation-specific (Other Resources)
+	// Support for other resources is Implementation-Specific.
 	//
 	// +kubebuilder:default=Gateway
 	// +optional
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -85,10 +85,6 @@ type ParentReference struct {
 	// are specified, the name and port of the selected listener must match
 	// both specified values.
 	//
-	// * Service: Port Name. When both Port (experimental) and SectionName
-	// are specified, the name and port of the selected port must match
-	// both specified values.
-	//
 	// Implementations MAY choose to support attaching Routes to other resources.
 	// If that is the case, they MUST clearly document how SectionName is
 	// interpreted.
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -85,7 +85,7 @@ type ParentReference struct {
 	// are specified, the name and port of the selected listener must match
 	// both specified values.
 	//
-	// * Service: Named port. When both Port (experimental) and SectionName
+	// * Service: Port Name. When both Port (experimental) and SectionName
 	// are specified, the name and port of the selected port must match
 	// both specified values.
 	//
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -21,9 +21,10 @@ import (
 )
 
 // ParentReference identifies an API object (usually a Gateway) that can be considered
-// a parent of this resource (usually a route). The only kind of parent resource
-// with "Core" support is Gateway. This API may be extended in the future to
-// support additional kinds of parent resources, such as HTTPRoute.
+// a parent of this resource (usually a route). The only kinds of parent resources
+// with "Core" support are Service (only for implementations supporting the Mesh
+// conformance profile) and Gateway. This API may be extended in the future to
+// support additional kinds of parent resources.
 //
 // The API object must be valid in the cluster; the Group and Kind must
 // be registered in the cluster for this reference to be valid.
@@ -43,6 +44,8 @@ type ParentReference struct {
 	//
 	// Support: Core (Gateway)
 	//
+	// Support: Core (Service - mesh conformance profile only)
+	//
 	// Support: Implementation-specific (Other Resources)
 	//
 	// +kubebuilder:default=Gateway
@@ -57,6 +60,13 @@ type ParentReference struct {
 	// allowed by something in the namespace they are referring to. For example:
 	// Gateway has the AllowedRoutes field, and ReferenceGrant provides a
 	// generic way to enable any other kind of cross-namespace reference.
+	//
+	// ParentRefs from a Route to a Service in the same namespace are "producer"
+	// routes, which apply default routing rules to inbound connections from
+	// any namespace to the Service. ParentRefs from a Route to a Service in a
+	// different namespace are "consumer" routes, and these routing rules are
+	// only applied to outbound connections to the Service from the same
+	// namespace as the Route.
 	//
 	// Support: Core
 	//
@@ -73,6 +83,10 @@ type ParentReference struct {
 	//
 	// * Gateway: Listener Name. When both Port (experimental) and SectionName
 	// are specified, the name and port of the selected listener must match
+	// both specified values.
+	//
+	// * Service: Named port. When both Port (experimental) and SectionName
+	// are specified, the name and port of the selected port must match
 	// both specified values.
 	//
 	// Implementations MAY choose to support attaching Routes to other resources.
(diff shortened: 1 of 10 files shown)


apis/v1beta1/shared_types.go
```diff
@@ -75,6 +85,10 @@ type ParentReference struct {
 	// are specified, the name and port of the selected listener must match
 	// both specified values.
 	//
+	// * Service: Named port. When both Port (experimental) and SectionName
+	// are specified, the name and port of the selected port must match
+	// both specified values.
+	//
```
mikemorris, line 91: @howardjohn I wasn't sure if _named_ port support made it into any upstream conformance yet, or if that was just an Istio-specific implementation. Adding this feels reasonable to me, although I think it should be Extended conformance for the mesh profile, to align with how Port is supported (rather than Core as when used to target a Gateway Listener).

apis/v1beta1/shared_types.go
```diff
@@ -58,6 +61,13 @@ type ParentReference struct {
 	// Gateway has the AllowedRoutes field, and ReferenceGrant provides a
 	// generic way to enable any other kind of cross-namespace reference.
 	//
+	// ParentRefs from a Route to a Service in the same namespace are "producer"
+	// routes, which apply default routing rules to inbound connections from
+	// any namespace to the Service. ParentRefs from a Route to a Service in a
+	// different namespace are "consumer" routes, and these routing rules are
+	// only applied to outbound connections to the Service from the same
```
keithmattix, line 68: nit: the phrasing here is a bit ambiguous. Does "from the same namespace as the route" describe the connections or the `Service`?

mikemorris: This is intended to describe the connections - originating from a workload in the same namespace as the Route, with an intended destination of the Service [in a different namespace] targeted by the parentRef of the Route.

I tried a few other ways to write this, but it felt either less clear or too verbose 🙁

apis/v1beta1/shared_types.go
```diff
@@ -58,6 +61,13 @@ type ParentReference struct {
...
+	// namespace as the Route.
```
mikemorris, line 69:
```suggestion
	// different namespace are "consumer" routes, and these routing rules are
	// only applied to outbound connections originating from the same namespace
	// as the Route, for which the intended destination of the connections are a
	// Service targeted as a ParentRef of the Route.
```

@keithmattix More verbose, but is this perhaps more explicitly clear?

apis/v1beta1/shared_types.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 // ParentReference identifies an API object (usually a Gateway) that can be considered
-// a parent of this resource (usually a route). The only kind of parent resource
-// with "Core" support is Gateway. This API may be extended in the future to
-// support additional kinds of parent resources, such as HTTPRoute.
+// a parent of this resource (usually a route). The only kinds of parent resources
+// with "Core" support are Service (only for implementations supporting the Mesh
+// conformance profile) and Gateway. This API may be extended in the future to
+// support additional kinds of parent resources.
```
robscott, line 27: Maybe something like this would line up well with our conformance profiles approach? 

/cc @shaneutt @mlavacca 

```suggestion
// a parent of this resource (usually a route). There are two kinds of parent resources
// with "Core" support:
//
// * Gateway (Gateway conformance profile)
// * Service (Mesh conformance profile)
//
// This API may be extended in the future to support additional kinds of parent
// resources.
```

apis/v1beta1/shared_types.go
```diff
@@ -43,6 +44,8 @@ type ParentReference struct {
 	//
 	// Support: Core (Gateway)
 	//
+	// Support: Core (Service - mesh conformance profile only)
+	//
 	// Support: Implementation-specific (Other Resources)
```
robscott, line 49: Maybe something similar to what I suggested above.

```suggestion
	// There are two kinds of parent resources with "Core" support:
	//
	// * Gateway (Gateway conformance profile)
	// * Service (Mesh conformance profile)
	//
	// Support for other resources is Implementation-Specific.
```

apis/v1beta1/shared_types.go
robscott: This update makes sense to me, maybe the section about consumer routes could justify a new paragraph to help split it up a bit more?

apis/v1beta1/shared_types.go
```diff
@@ -75,6 +85,10 @@ type ParentReference struct {
 	// are specified, the name and port of the selected listener must match
 	// both specified values.
 	//
+	// * Service: Named port. When both Port (experimental) and SectionName
```
robscott, line 88: Tiny nit: I think this might be a bit clearer.

```suggestion
	// * Service: Port Name. When both Port (experimental) and SectionName
```

apis/v1beta1/shared_types.go
```diff
@@ -130,15 +148,19 @@ type CommonRouteSpec struct {
 	// to be attached to. Note that the referenced parent resource needs to
 	// allow this for the attachment to be complete. For Gateways, that means
 	// the Gateway needs to allow attachment from Routes of this kind and
-	// namespace.
+	// namespace. For Services, that means the Service must either be in the same
+	// namespace for a "producer" route, or the mesh implementation must support
+	// and allow "consumer" routes for the referenced Service.
```
robscott, line 153: I may have missed this, could someone use a ReferenceGrant to turn a Producer route into a Consumer route? Are Is a ReferenceGrant that grants access from `HTTPRoute` to `Service` _only_ for `BackendRefs` and not `ParentRefs`? Assuming that's the case, we probably need to very clearly state that somewhere.

apis/v1beta1/shared_types.go
```diff
@@ -130,15 +148,19 @@ type CommonRouteSpec struct {
...
 	//
-	// The only kind of parent resource with "Core" support is Gateway. This API
-	// may be extended in the future to support additional kinds of parent
+	// The only kinds of parent resources with "Core" support are Service (only
+	// for implementations supporting the Mesh conformance profile) and Gateway.
+	// This API may be extended in the future to support additional kinds of parent
 	// resources such as one of the route kinds.
```
robscott, line 158:
```suggestion
	// There are two kinds of parent resources with "Core" support:
	//
	// * Gateway (Gateway conformance profile)
	// * Service (Mesh conformance profile)
	//
	// This API may be extended in the future to support additional kinds of parent
	// resources.
```

apis/v1beta1/shared_types.go
```diff
@@ -150,7 +172,14 @@ type CommonRouteSpec struct {
...
 	// Gateway has the AllowedRoutes field, and ReferenceGrant provides a
-	// generic way to enable any other kind of cross-namespace reference.
+	// generic way to enable other kinds of cross-namespace reference.
+	//
+	// ParentRefs from a Route to a Service in the same namespace are "producer"
+	// routes, which apply default routing rules to inbound connections from
+	// any namespace to the Service. ParentRefs from a Route to a Service in a
+	// different namespace are "consumer" routes, and these routing rules are
+	// only applied to outbound connections to the Service from the same
+	// namespace as the Route.
```
robscott, line 182: Just so it doesn't get missed, this paragraph should be updated to match the related changes you're proposing above whenever we reach consensus on those.

robscott (commented):
Thanks @mikemorris!

apis/v1beta1/shared_types.go
keithmattix: Yeah I like this better, and I agree with Rob that at least a reference to some consumer route docs make sense

apis/v1beta1/shared_types.go
mikemorris: Yea, I struggled a bit figuring out how to introduce a docs pattern to express this well, would definitely appreciate more eyes on this to try to get it right.

robscott:
🤦  this failed with the same error that I thought I fixed with https://github.com/kubernetes-sigs/gateway-api/pull/2156. Will try again to see if this is a flake or more consistent failure.

/retest

apis/v1beta1/shared_types.go
mikemorris: Ah, I may have just been inferring more functionality than actually existed from your comments in https://github.com/kubernetes-sigs/gateway-api/pull/1995

apis/v1beta1/shared_types.go
robscott: I think what we need to say here is that ReferenceGrant is not applicable for ParentRefs to Services. Then in ReferenceGrant API Spec we should probably clarify that when a ReferenceGrant is used to grant access from a Route to a Service, that is _only_ for BackendRefs, not for ParentRefs.

apis/v1beta1/shared_types.go
mlavacca: yeah, I think this description fits very well 👍

hzxuzhonghu (commented):
lgtm.
  Just one question: if we have a CRD defining some rules functioning on service, should we use ParentRef or TargetRef?

mikemorris:
Resolved suggested changes for support level <> conformance profile documentation and clarifying ReferenceGrant non-applicability to Service as ParentRef, this should be ready for a final review pass now.

mikemorris:
@hzxuzhonghu If you're implementing your own CRD for defining some service policy rules, I believe that should follow the pattern in https://gateway-api.sigs.k8s.io/references/policy-attachment/ and use `targetRef` (although I do see how the distinction between `targetRef` and `parentRef` for similar use cases could be confusing).

hzxuzhonghu (commented):
/lgtm

robscott:
Thanks @mikemorris! Will add a hold for someone else to sign off on this as well, but LGTM.

/cc @howardjohn @keithmattix @kflynn
/approve

kflynn (approved):
Hit it! Thanks, @mikemorris. 🙂

keithmattix (approved):
LGTM as well. Thanks @mikemorris

keithmattix:
2/3 GAMMA leads approve, so removing the hold
/remove-hold

PR #2150, opened 2023-06-28
Add IncompatibleFilters reason for Accepted condition

### LINKED ISSUE #1521: webhook/conformance: Handle incompatible HTTP Filters on a rule

sunjayBhatia:
**What type of PR is this?**

/kind cleanup
/kind documentation

Adds new reason for configuration errors arising from invalid combination of route filters.

Pulled out of https://github.com/kubernetes-sigs/gateway-api/pull/1540 per @robscott 

Fixes #1521

```release-note
Add IncompatibleFilters reason for implementations to specify when a route is invalid due to an invalid combination of route filters.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -207,8 +207,10 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// document that limitation. In cases where incompatible or unsupported
+	// filters are specified and cause the `Accepted` condition to be set to status
+	// `False`, implementations may use the `IncompatibleFilters` reason to specify
+	// this configuration error.
 	//
 	// Support: Core
 	//
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -225,6 +225,13 @@ type GRPCRouteRule struct {
 	//
 	// Specifying a core filter multiple times has unspecified or
 	// implementation-specific conformance.
+	//
+	// If an implementation can not support a combinations of filters, they must clearly
+	// document that limitation. In cases where incompatible or unsupported
+	// filters are specified and cause the `Accepted` condition to be set to status
+	// `False`, implementations may use the `IncompatibleFilters` reason to specify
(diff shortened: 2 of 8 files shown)


apis/v1beta1/httproute_types.go
```diff
@@ -207,8 +207,10 @@ type HTTPRouteRule struct {
 	// All filters are expected to be compatible with each other except for the
 	// URLRewrite and RequestRedirect filters, which may not be combined. If an
 	// implementation can not support other combinations of filters, they must clearly
-	// document that limitation. In all cases where incompatible or unsupported
-	// filters are specified, implementations MUST add a warning condition to status.
+	// document that limitation. In cases where incompatible or unsupported
+	// filters are specified and cause the `Accepted` condition to be set to status
```
sunjayBhatia, line 211: this section `cause the Accepted condition to be set to status False` is because you might have a partially valid route, in that case it would still be `Accepted=True` so doesn't really make sense to set the `IncompatibleFilters` reason

robscott: Yeah, I ran into the same problem in https://github.com/kubernetes-sigs/gateway-api/pull/2155 where it would be _really_ useful to be able to communicate "something's not quite right, but we're still able to at least partially reconcile this route". In that case it was if the timeout specified was more precise (ie ms or ns) than the underlying implementation could support.

robscott (commented):
Thanks @sunjayBhatia! Will defer to someone else for final LGTM.

/cc @arkodg @mlavacca @howardjohn @shaneutt 
/approve

PR #2155, opened 2023-06-28
Updating GEP-1742 to clarify timeout formatting

robscott:
**What type of PR is this?**
/kind cleanup
/kind gep

While reviewing the implementation of this GEP, I realized that the upper and lower bounds of these timeout values were not what I expected. This makes the following changes:

**1. Introduces an upper bound of 10 years**
In Kubernetes APIs we _always_ try to have a clear upper and lower bound for fields. This enables implementations to write tests that ensure they can support the min and max values for any field in the API. As far as I can tell, 10 years is lower than all implementations support, but still likely sufficiently high to not be a limiting factor for anyone. Note that our [API versioning guidelines](https://gateway-api.sigs.k8s.io/concepts/versioning/#minor-version-eg-v040-v050) allow us to loosen validation in future releases, but we try to avoid tightening validation wherever possible. This means that we should aim to start with the tightest validation that is reasonable for a feature.

**2. Removes the possibility of setting 0 to mean "no timeout"**
Although both HAProxy and Envoy support 0 as "no timeout", I'm not sure how broadly supportable this concept of an "infinite" timeout would be beyond those two dataplanes. If 0 is really just intended to represent "max timeout", why not define a max timeout supported by the API and suggest users set that if that's really what they want? That feels both more specific and less likely to surprise users. If we ever determine that we really do need a "0" value here, we can loosen validation in the future to support it, but once we've allowed that value it's nearly impossible to undo that.

**3. Clarifies that values should be rounded up to the closest value an implementation can support**
There are likely some implementations that are not going to be able to support the level of granularity supported by Golang's concept of duration. This provides guidance to those implementations for how they should handle this. We likely also will want to add something to status for this in the future, but there's not an obvious condition to attach that to yet.

```release-note
NONE
```

/cc @frankbu @srodi @candita @shaneutt @howardjohn

### DIFF
--- a/geps/gep-1742.md
+++ b/geps/gep-1742.md
@@ -206,7 +206,7 @@ sequenceDiagram
     participant C as Client
     participant P as Proxy
     participant U as Upstream
-    
+
     C->>P: Connection Started
     activate U
     activate C
@@ -224,7 +224,7 @@ sequenceDiagram
     activate U
     note left of U: timeout queue<br/>(wait for available server)
     deactivate U
-		
+
     P->>U: Connection Started
     activate U
     P->>U: Starts sending Request
@@ -370,7 +370,7 @@ sequenceDiagram
 
 Both timeout fields are string duration values as specified by
 [Golang time.ParseDuration](https://pkg.go.dev/time#ParseDuration) and MUST be >= 1ms
-or 0 to disable (no timeout).
+and <= 100y.
 
 ### GO
 
@@ -388,8 +388,6 @@ type HTTPRouteRule struct {
 }
 
 // HTTPRouteTimeouts defines timeouts that can be configured for an HTTPRoute.
-// Timeout values are formatted like 1h/1m/1s/1ms as parsed by Golang time.ParseDuration
-// and MUST BE >= 1ms or 0 to disable (no timeout).
 type HTTPRouteTimeouts struct {
 	// Request specifies the duration for processing an HTTP client request after which the
 	// gateway will time out if unable to send a response.
@@ -403,13 +401,22 @@ type HTTPRouteTimeouts struct {
 	// request stream has been received instead of immediately after the transaction is
 	// initiated by the client.
 	//
+	// This value follows a subset of Golang and XML Duration formatting, for example 1m, 1s,
+	// or 1ms. This value MUST BE >= 0 and <= 9999h. When "0" is specified, it indicates that
+	// no timeout should be implemented. If an implementation cannot support the granularity
+	// specified, it MUST round up to the closest value it supports. For example, if 1ms is
+	// specified, and an implementation can only configure timeouts in seconds, it MUST configure
+	// 1 second as the timeout. In the special case of a "0" timeout, the implementation SHOULD
+	// NOT implement a timeout. If that is not possible it MUST configure the maximum timeout
+	// value it supports.
+	//
 	// When this field is unspecified, request timeout behavior is implementation-dependent.
 	//
 	// Support: Extended
 	//
 	// +optional
-	// +kubebuilder:validation:Format=duration
-	Request *metav1.Duration `json:"request,omitempty"`
+	// +kubebuilder:validation:Pattern=^(0|[1-9][0-9]{0,3}(m|s|ms))$
+  Request *metav1.Duration `json:"request,omitempty"`
 
 	// BackendRequest specifies a timeout for an individual request from the gateway
 	// to a backend service. This covers the time from when the request first starts being
@@ -419,14 +426,22 @@ type HTTPRouteTimeouts struct {
 	// may result in more than one call from the gateway to the destination backend service,
 	// for example, if automatic retries are supported.
 	//
-	// Because the Request timeout encompasses the BackendRequest timeout,
-	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
+	// This value follows a subset of Golang and XML Duration formatting, for example 1m, 1s,
+	// or 1ms. This value MUST BE >= 0 and <= 9999h. When "0" is specified, it indicates that
+	// no timeout should be implemented. If an implementation cannot support the granularity
+	// specified, it MUST round up to the closest value it supports. For example, if 1ms is
+	// specified, and an implementation can only configure timeouts in seconds, it MUST configure
+	// 1 second as the timeout. In the special case of a "0" timeout, the implementation SHOULD
+	// NOT implement a timeout. If that is not possible it MUST configure the maximum timeout
+	// value it supports.
+	//
+	// When this field is unspecified, request timeout behavior is implementation-dependent.
 	//
 	// Support: Extended
 	//
 	// +optional
-	// +kubebuilder:validation:Format=duration
-	BackendRequest *metav1.Duration `json:"backendRequest,omitempty"`
+	// +kubebuilder:validation:Pattern=^(0|[1-9][0-9]{0,3}(m|s|ms))$
+	BackendRequest *string `json:"backendRequest,omitempty"`
 }
 ```
 


robscott:
Adding a hold for consensus.

/hold

geps/gep-1742.md
```diff
@@ -364,7 +364,7 @@ sequenceDiagram
 
 Both timeout fields are string duration values as specified by
 [Golang time.ParseDuration](https://pkg.go.dev/time#ParseDuration) and MUST be >= 1ms
-or 0 to disable (no timeout).
+and <= 100y.
```
frankbu, line 367: I added some comments related to this in the API PR (https://github.com/kubernetes-sigs/gateway-api/pull/2013/files#r1245498559), but thinking about this more, I really don't see the value of specifying an upper limit.

The more I think about this, it seems that any value greater than a day or so, basically means no timeout. Certainly nobody would want to set a timeout to something like 5 years and really expect that the client and service will both be up and running in a paused state until it returns a value before the 5y deadline.

Unless I'm missing something, a user will want to configure the timeout to something relatively short, and any larger value really just mean "don't timeout", i.e., just hang until the client connection is terminated. So that's why I think the current design is right. User can set the timeout to some small value or 0 to disable timeouts. A user can also chose to set a pretty huge value which the implementation will likely support (up to the implementation limit of probably an int64 max value ns or ~300 years), but any specific number they chose that is large will work the same as 0 for all practical purposes. So documenting some arbitrary large max value is really meaningless IMO.

geps/gep-1742.md
```diff
@@ -413,8 +417,14 @@ type HTTPRouteTimeouts struct {
 	// may result in more than one call from the gateway to the destination backend service,
 	// for example, if automatic retries are supported.
 	//
-	// Because the Request timeout encompasses the BackendRequest timeout,
-	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
```
frankbu, line 417: I think defaulting BankendRequest to the value or Request is both easily implementable and a sensible approach. See my related comment here: https://github.com/kubernetes-sigs/gateway-api/pull/2013/files#r1245517080

geps/gep-1742.md
frankbu: One more related thought. If we want to say anything about max value it could be something like `the maximum value is implementation dependent, but MUST BE at least 1y`, or something like that.

That way the user can set it to any sensible value they'd like (>= 1ms) or 0 for infinite (don't timeout).

@robscott btw, you said:

> In Kubernetes APIs we always try to have a clear upper and lower bound for fields. This enables implementations to write tests that ensure they can support the min and max values for any field in the API.

How would an implementation test that a 10 year timeout works?

robscott: >How would an implementation test that a 10 year timeout works?

Yeah, you're right, it's not really possible to write a test that waits this long, and any value that we can feasibly test is likely lower than at least some users would want. I think the real win here is that implementations can know that they support the full range of possible values allowed by the API just by understanding their dataplane.

> If we want to say anything about max value it could be something like the maximum value is implementation dependent, but MUST BE at least 1y, or something like that.

I think that's reasonable, but I'm struggling to think of any use case that would benefit from > 1y, so distinguishing beyond that is likely not worthwhile (again I could be wrong here). My strong bias is to start with tighter limits and only loosen them if a clearly defined need is presented.

>The more I think about this, it seems that any value greater than a day or so, basically means no timeout. Certainly nobody would want to set a timeout to something like 5 years and really expect that the client and service will both be up and running in a paused state until it returns a value before the 5y deadline.

Yeah I think that's probably accurate, I'm not sure what the right upper limit is here, but lower is probably better as a starting point.

>So documenting some arbitrary large max value is really meaningless IMO.

There's a very strong precedent in Kubernetes API review to prevent unbounded values. Even if we were to push forward without an upper limit here, I can't imagine this passing the final level of API review before v0.8.0. Both @thockin and @khenidak have strongly pushed to avoid any unbounded values in this API.

geps/gep-1742.md
robscott: Commented in that thread first, copying over here for additional visibility:

Unfortunately we don't have a means to implement variable defaulting unless we want to create and maintain a defaulting webhook, and I'd strongly prefer to avoid that maintenance cost.

We could suggest that if a value is not set here, implementations should consider this the same value as the overall request timeout, but I can't think of any way to validate that, and I'm also not sure that's what we want. Let's say that an implementation has a default of 2 retries with a 10s per-try timeout. If a user configures an overall request timeout of 20s, I think they'd probably prefer the implementation default behavior here instead of that 1 value actually affecting the per-try timeout value as well. In general I'd like to suggest working towards the principle of least surprise, and I think that means avoiding a variable default like this.

geps/gep-1742.md
thockin: Two things:

1) String-encoded durations is not the norm in kube APIs, and I think we should question whether that's the right way to represent durations.  The norm is `int32` with the units encoded in the name - e.g. `timeoutSeconds int32`, and the unit is almost always seconds.

2) It's not really unbounded if it is a float32 - there is a natural bound.

robscott: That's a good point @thockin. I had thought of something like `timeoutSeconds int32` as well, it would certainly simplify the validation initially. A concern I'd have is that some users will want to be able to express more precise timeouts, for example GCP allows you to express timeouts in terms of nanoseconds, and most implementations support at least milliseconds as an input. That leaves us with a few options:

1. Only support one unit (probably seconds or milliseconds) - this matches the [approach of ingress-nginx](https://github.com/kubernetes/ingress-nginx/issues/6540) (seconds), [GCP BackendServices](https://cloud.google.com/load-balancing/docs/backend-service#timeout-setting) (seconds), [Kong](https://docs.konghq.com/kubernetes-ingress-controller/latest/references/annotations/#konghqcomconnect-timeout) (ms), and [HAProxy](https://github.com/haproxytech/kubernetes-ingress/blob/402c43c4698434d9d61b702772a2671ebf125987/documentation/custom-resources.md?plain=1#L11) (ms)
2. Support multiple units, combining the values (`timeout.seconds: 1` + `timeout.nanoseconds: 5000000` == `1.005s`). This roughly matches [Envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-routeaction-timeout) and [GCP URLMaps](https://cloud.google.com/compute/docs/reference/rest/v1/urlMaps#UrlMap.FIELDS.inlinedField_26) which both appear to use proto [duration](https://protobuf.dev/reference/protobuf/google.protobuf/#duration).
3. Support something like `metav1.Duration` as proposed by this GEP. This matches [Istio](https://istio.io/latest/docs/tasks/traffic-management/request-timeouts/), [Linkerd](https://linkerd.io/2.13/tasks/configuring-timeouts/), [Contour](https://projectcontour.io/docs/main/configuration/#timeout-configuration)

So with all of that said, we could start with 1 and leave room for 2. For example, that future could look like this:

GEP Today:
```yaml
timeouts:
  request: 10s
```

GEP Tomorrow:
```yaml
timeouts:
  request:
    seconds: 10
```

GEP Next Year:
```yaml
timeouts:
  request:
    seconds: 10
    nanoseconds: 500000
```

I'm not entirely sure that I prefer that over the more open-ended duration format, but it would be simpler to validate and more in line with other Kubernetes APIs. Does this make sense to you @thockin?

frankbu: @thockin @robscott 

> String-encoded durations is not the norm in kube APIs, and I think we should question whether that's the right way to represent durations. The norm is int32 with the units encoded in the name - e.g. timeoutSeconds int32, and the unit is almost always seconds.

I haven't worked much with kube APIs in the past, so I'm not sure I understand all the conventions and "rules", but given the use case here, where timeouts will typically be in milliseconds and sometimes seconds, the string serialization format used by time.Duration is really the best way to represent it IMO. I believe there is also an equivalent Rust duration type that also uses this same string format as well.

That said, I'm wondering if an exception from the usual kube rules makes sense, or is even possible, in this case?

thockin: > and more in line with other Kubernetes APIs

With regards to timeouts, we use int32 exclusively in the main kube API.  I didn't even realize `metav1.Duration` exists, and it seemonly used in component APIs (i.e. config files).  Don't get me wrong, I wish we had used Duration everywhere - I do find it easier to deal with, but inconsistency is a pain.

Let me discuss a bit.

geps/gep-1742.md
thockin: Can't you set static defaults through CRD, like `+default=` ?

robscott: Yep, static would be fine here, but what had been suggested here in the original GEP was that if this field was not specified but the overall request timeout was specified, this field should default to the value of the overall request timeout. I don't think we can/should do that. Also since this is a new field and many implementations already have default timeouts, I don't think we can add any static timeout in a backwards compatible way.

geps/gep-1742.md
robscott: OK this ended up being more complex than expected. @thockin started a [Slack thread](https://kubernetes.slack.com/archives/CJUQN3E4T/p1688063522998959) to discuss this more, that discussion will likely continue, but here's a high level summary:

1. There's not really any prior art for `metav1.Duration` in core Kubernetes APIs.
2. Open API v3 validation translates this to [xml dayTimeDuration](https://www.w3.org/TR/xmlschema11-2/#dayTimeDuration) ([source](https://github.com/OAI/OpenAPI-Specification/issues/607))
3. In Kubernetes, we'd be trying to parse this into a [Duration](https://pkg.go.dev/time#Duration), the formats look nearly identical, but there may be some differences in translation between types

I asked @howardjohn how Istio handles this - apparently they're using a plain `string` field for their timeouts and then using webhook validation to ensure the format lines up. That's theoretically an option for us, but does not work well if/when our webhook is not present.

I think we need more API reviewers to chime in here, but in general I'd prefer to move forward with a duration format as currently proposed if we can make it work. If not, I think our next best option would be to use something like `timeouts.request.milliseconds` as an `int32`, leaving room for additional units in the future if needed.

frankbu: Istio uses protobuf to define all its APIs. Here's a [duration field](https://github.com/istio/api/blob/f930ebfda37a76446f229bf7e465bbbf305501af/networking/v1beta1/virtual_service.proto#L618) example. In the generated GO binding that translates to [durationpb.Duration](https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb), similar API to `metav1.Duration`.

howardjohn:
I am sure you can make a regexp that matches the Go duration syntax? We
just happen to use a webhook, there are likely better options.

geps/gep-1742.md
robscott: OK, capturing some more context from the [Slack thread](https://kubernetes.slack.com/archives/CJUQN3E4T/p1688063522998959). @liggitt suggested that we could start with a simple regex that represented a small subset of these duration formats - something like this:

`^[1-9][0-9]*(h|m|s|ms)$`

As far as I can tell, that would be a subset of Go and XML/OpenAPI formats used here, likely many others as well. Importantly it would prevent specifying something like `1m30s` and instead you'd need to represent that as `90s`. Some potential modifications to make this more robust:

1. Allow a zero value to indicate "no timeout":

`^[0]|[1-9][0-9]*(h|m|s|ms)$`

2. Limit the number of digits to effectively enforce a max value:

`^[0]$|^[1-9][0-9]{0,3}(h|m|s|ms)$`

3. Remove hour from the allowed set of units since most timeouts can be expressed without them

`^[0]$|^[1-9][0-9]{0,3}(m|s|ms)$`

This would make the max value 9999 minutes which is ~167 hours or ~7 days.

Alternatively we could just choose a single value like milliseconds which would be easier to manage from both an API and implementation perspective, but decidedly more of a pain for users to configure.

robscott: I've updated this PR to use the approach I've described in the comment above, PTAL.

/cc @frankbu @srodi @shaneutt @thockin @howardjohn

thockin: If we do this, let's consider the precedent.  Hours seems required for anyone else who follows this pattern.

> As far as I can tell, that would be a subset of Go and XML/OpenAPI formats

I looked at the XML spec and I find 3 things that I can't answer for sure:

1) They define the units as "H", "M", "S", whereas go does "h", "m", "s".  I could not convince myself that case is ignored

2) They define all values as starting with literal "P"

3) They do not have a milliseconds unit.

As such, I think "compatible with XMLSchema" is not really true, and we're just defining a subset of Go's duration grammar.

candita: Regarding the 0 value, I would go a step further and regard a 0 timeout value as if the value was empty.  This indicates that the user doesn't care and will accept the default from the implementation.

geps/gep-1742.md
```diff
@@ -403,13 +401,22 @@ type HTTPRouteTimeouts struct {
 	// request stream has been received instead of immediately after the transaction is
 	// initiated by the client.
 	//
+	// This value follows a subset of Golang and XML Duration formatting, for example 1m, 1s,
+	// or 1ms. This value MUST BE >= 0 and <= 9999h. When "0" is specified, it indicates that
```
candita, line 405: Hour is not an accepted value in your kubebuilder validation pattern.  9999h is 599940m, so if you want to allow that amount, you would need to change the pattern to allow 6 digits.

geps/gep-1742.md
```diff
@@ -403,13 +401,22 @@ type HTTPRouteTimeouts struct {
...
+	// value it supports.
+	//
 	// When this field is unspecified, request timeout behavior is implementation-dependent.
 	//
 	// Support: Extended
 	//
 	// +optional
-	// +kubebuilder:validation:Format=duration
-	Request *metav1.Duration `json:"request,omitempty"`
+	// +kubebuilder:validation:Pattern=^(0|[1-9][0-9]{0,3}(m|s|ms))$
```
candita, line 418: 9999m is not long enough.

geps/gep-1742.md
```diff
@@ -403,13 +401,22 @@ type HTTPRouteTimeouts struct {
...
+  Request *metav1.Duration `json:"request,omitempty"`
```
candita, line 419:
```suggestion
      Request *metav1.Duration `json:"request,omitempty"`
```

geps/gep-1742.md
```diff
@@ -419,14 +426,22 @@ type HTTPRouteTimeouts struct {
 	// may result in more than one call from the gateway to the destination backend service,
 	// for example, if automatic retries are supported.
 	//
-	// Because the Request timeout encompasses the BackendRequest timeout,
-	// the value of BackendRequest defaults to and must be <= the value of Request timeout.
+	// This value follows a subset of Golang and XML Duration formatting, for example 1m, 1s,
```
candita, line 429: This applies to text above this line.  Can you add an explanation about whether the Request time encompasses/includes the BackendRequest time? Is Request = BackendRequest + other processing time, or is total time = Request + BackendRequest, implying that Request does NOT include the BackendRequest time?

geps/gep-1742.md
robscott: >Regarding the 0 value, I would go a step further and regard a 0 timeout value as if the value was empty.

@candita I think we need to avoid that here for backwards compatibility. We need the nil value to be implementation-specific and essentially map to whatever implementations are already doing. It's unfortunate to make 0 mean something different than nil, and I was initially against that, but it does seem like it's overwhelmingly used to represent "infinite" when it comes to timeout configuration in underlying implementations, so we should probably support that here as well.

geps/gep-1742.md
robscott: Good catch, thanks! This is an artifact of when I originally had hours as an allowed unit, will update pending the conversation above about whether or not we should add that back.

geps/gep-1742.md
frankbu: Another thought. I understand that we can't enforce that `BackendRequest timeout must be <= Request timeout`, but maybe we could/should still say something like this?
```suggestion
	// Because the Request timeout encompasses the BackendRequest timeout,
	// the value of BackendRequest should be <= the value of Request timeout to have any effect.
	// This value follows a subset of Golang Duration formatting, for example 1m, 1s,
```

geps/gep-1742.md
thockin: > Based on the feedback in yesterday's community meeting, I think the proposal itself of a simplified string format for duration is reasonable.

@liggitt are you OK with this?  It'll set a new precedent and impose client-side parsing.

If we do this, we should probably define a `type Duration string` and comment it with the full grammar.  We need to think about compat - if we don't allow `"1m30s"` or 'h' and 'd' suffixes or `"1.5m"` now, we may never be able to allow it (because client-side parsing).  Please don't take this lightly.

liggitt: > @liggitt are you OK with this? It'll set a new precedent and impose client-side parsing.

It doesn't seem worse than resource.Quantity or IntOrString we already have in other places.

> If we do this, we should probably define a type Duration string and comment it with the full grammar. We need to think about compat - if we don't allow "1m30s" or 'h' and 'd' suffixes or "1.5m" now, we may never be able to allow it (because client-side parsing). Please don't take this lightly.

As a general type, I'd definitely allow `h`, `0`, negative values, etc.

`d` is not defined in time.Duration, and "1 day" is not well-defined in the face of leap-seconds/daylight saving shifts/etc, so excluding that seems fine.

My instinct was to disallow fractional values, but maybe there's a case to be made for those?

thockin: > It doesn't seem worse than resource.Quantity or IntOrString we already have in other places.

I have been vocal about NOT further propagating IntOrString (or recreating the same hackery).  Many people think quantity "could have been done better".  I don't think "not worse than" is a great criteria here.  I think we need to say it is affirmatively better than the alternatives.

1) `TimeoutSeconds int32`
2) `TimeoutMillseconds int64`
3) `Timeout struct { Seconds int32, Milliseconds int32 }`
4) `TimeoutSeconds int32; TimeoutMilliSecinds int32;`
5) `Timeout DurationString`

> As a general type, I'd definitely allow h, 0, negative values, etc.

Agree

> d is not defined in time.Duration, and "1 day" is not well-defined 

SGTM

> My instinct was to disallow fractional values...as simple and unambiguous as possible

My instinct, too.  I don't want to specify FP behavior and precision limits.  I don't want to impose fixed-point math on clients.  I see why *simple* FP is attractive but the grammar is sufficiently complicated that I think we can live without it.  I just want to think it through.

`/-?([0-9]+\.)?[1-9][0-9]*(h|m|s|ms)/` is not so bad, but I have 50% confidence that I actually got it right.  Do we actually support fractional milliseconds?  Do we limit it to 3 decimal places?

thockin: We also need to NOT NORMALIZE it.

kflynn: While I'm as down on floats as the next person (🙂) , I feel like we should just go ahead and use `metav1.Duration` for the timeouts. It's already present and tested in apimachinery, it's starting to see use (KubeletConfiguration, for example), and it's well-supported in languages used in the ecosystem (Go is obvious, but a Rust parser is being added to [kube-rs](https://github.com/kube-rs/kube/pull/1224) as part of Buoyant's work implementing GEP-1742 timeouts in Linkerd, and there are implementations in Python as well).

Most importantly, though: metav1.Duration is easier for the user than other things. Sadly, my experience with Emissary-ingress is that having the unit in the name results in more errors than having it in the value -- I've seen far fewer cases of a user typoing `35s` when they meant `35ms` than I have of users e.g. putting a number of seconds into a field explicitly marked as milliseconds (e.g. `timeout_ms: 30` instead of `timeout_ms: 30000`). I think the Go folks did a nice job here of creating a thing that's simple for the user to use, even though it's harder on the implementers.

Extending @thockin's regex seems to work fine in quick tests:

```
^-?(([0-9]+\.)?[0-9]+(h|m|s|ms|us|ns))+$
```

We could simplify it if we want to disallow negative durations, floating-point values, or things smaller than milliseconds (I'd definitely be OK with the first two for timeouts). And we could expand it dramatically to catch out-of-order or repeating units, too, but I don't think that's worth it.

robscott: > I feel like we should just go ahead and use `metav1.Duration` for the timeouts

I don't mind this idea, but for reference I can't find a way to both use this type and add the kind of regex validation that would represent a subset of `metav1.Duration`. Very helpful to see that work is already underway to support this format in rust and python though. Ideally I think we should just aim to have a string format that is a subset of `metav1.duration`.

>I think we need to say it is affirmatively better than the alternatives.

I believe it clearly provides a better UX than the alternatives. Although it may be messier for non-golang implementations, it is also consistent with how so many implementations already handle timeouts, including Contour, Istio, and Linkerd.

Would it be fair to separate the following?

1. The range of values that should be supported by this broader format (may include positive and negative values + `h|m|s|ms|us|ns` units)
2. The range of values that we want to support in Gateway API (a subset of 1)

While I agree that the concept of a duration should include more units and _maybe_ negative values, I really don't want to overcomplicate the values supported by Gateway API. In general we've tried to support just the most common formats that are going to be broadly implementable. Concepts like negative timeouts and/or microseconds or nanoseconds really don't make sense in this context, and I'd hate to let validation allow them through.

thockin: I'm not convinced we need durations smaller than ms (famous last words, so I don't REALLY mind keeping `us` and `ns`, but not `µs`).

I think we can leave out unary `+` but keep unary `-`.

I think we can leave out compound values (e.g. no "1m30s").

I think we can leave out floating point values.  (my above regex was wrong.  `^-?([1-9][0-9]*|[0-9]*\.[0-9]{0,3})(h|m|s|ms)$` seems better, but let's not?)

That leaves us with something like `^-?[1-9][0-9]*(h|m|s|ms)$` - at least it's a) regexable and b) easy to parse by a human

Do we want to argue about FP or compound?

robscott: >I think we can leave out unary + but keep unary -.

I would strongly prefer to leave that out from valid Gateway API values. Although we can define a broader type that allows negative values, I don't want to unnecessarily expand the set of values Gateway API supports. In the specific case of negative values, I don't think there's any possible reasonable interpretation for timeouts so I'd rather avoid them.

I think we need to differentiate between a broader standard that we want to introduce, the subset of that standard that we want to allow for this specific use case.

robscott: One other variation that is critical for Gateway API but likely confusing for others. In most/all implementations I'm aware of, "0" is interpreted as no/infinite timeout. The variations of the regex I proposed allowed room for "0" without any units to communicate this state. If that's problematic, we could just tell users to configure something like "0s" to represent this concept.

thockin: I don't mind allowing 0, but in YAML it will have to be `"0"` because the field will be transported as a string.  `0s` does not suffer that problem.

thockin: >> I think we can leave out unary + but keep unary -.
>
> I would strongly prefer to leave that out from valid Gateway API values. Although we can define a broader type that allows negative values,

I'm talking about the difference between "Invalid value: timeout: must match regex ...." and "Invalid value: timeout: must be greater than or equal to 0ms"

thockin: > I don't mind allowing 0, but in YAML it will have to be "0" because the field will be transported as a string. 0s does not suffer that problem.

Also not that comparing `0s` to `0ms` to `0h` must succeed, as must "1000ms" and `1s`

robscott: >I'm talking about the difference between "Invalid value: timeout: must match regex ...." and "Invalid value: timeout: must be greater than or equal to 0ms"

Unfortunately I don't know of any way to do that with CRD validation. We could do that with webhook validation, but that is far less reliable so I'd rather cover as much as possible directly with validation defined on the CRD.

>Also not that comparing 0s to 0ms to 0h must succeed, as must "1000ms" and 1s

Agreed, we should likely have some conformance tests to confirm that.

thockin:
If we open this can of worms we will end up with a "standard" type, which
can eventually use all the declarative validation. Before we lock this
down, we should think about how this duration-as-string can be used in CEL.

jpbetz:
> If we open this can of worms we will end up with a "standard" type, which can eventually use all the declarative validation. Before we lock this down, we should think about how this duration-as-string can be used in CEL.

Any field in the openAPI schema of a CRD can `type: string, format:duration` today. In any x-kubernetes-validation strings, the duration is understood by CEL as a duration type making it trivial to validate against min/max limits,  and can be used in conjunction with datetime fields to do time math.

jpbetz:
You can also cast a string to a duration in CEL: `duration(self) < duration("1h")` for `type: string` (but no format: duration) if needed.

thockin:
Is there a spec for strings that can be casted to duration?

jpbetz:
> Is there a spec for strings that can be casted to duration?

Soo, I'm not thrilled with the current situation, but here it is:

From the [CEL spec](https://github.com/google/cel-spec/blob/master/doc/langdef.md): "Duration strings should support the following suffixes: "h" (hour), "m" (minute), "s" (second), "ms" (millisecond), "us" (microsecond), and "ns" (nanosecond). Duration strings may be zero, negative, fractional, and/or compound. Examples: "0", "-1.5h", "1m6s""

For comparison:

openapi: https://github.com/kubernetes/kube-openapi/blob/master/pkg/validation/strfmt/duration.go#L33-L42 (also supports "d" (day), "w" or "wk" (week)

Golang: https://pkg.go.dev/time#ParseDuration (almost identical but also supports "µs" as an alternative to "us")

I'm not aware of a recognized standard for any of this..

thockin:
This issue is proposing a kube standard for fields which represent duration.

We're trying to keep it small and easy, since it will require client-side
parsing to use (meaning we can't ever change it, really).

We described a subset of CEL and Go - no fractional or compound, and noting
smaller than ms or larger than h.

No literal 0, because YAML is too pervasive and will coerce to int.

Fractional would mean allowing things like "0.001us" and defining
floating-point precision.

Compound is just harder to spec, parse, and test.

All of those are potentially debatable.

As far as I can see, we don't have any actual duration-string fields in GA
APIs, yet. Tell me if I am wrong there?

I don't care if a CEL expression allows a broader set for casting, as long
as none of our APIs allow that to be saved. We would need to be careful
that CEL emits values in our format (e.g. always 90s and never 1m30s),
which could be able argument for adopting CEL's spec directly.

I think we need to either own our own fate or else ban these strings.

geps/gep-1742.md
youngnick: Catching up here, this is quite the rock to turn over!

I'm supportive of using a string based definition that's a subset of `metav1.Duration`, however we do it, and also agree with ensuring that we have conformance tests that validate as much behavior as possible.

frankbu:
@robscott What's the latest status on this duration format decision? I've added conformance tests for the new timeout features assuming we're going with the restricted string format: https://github.com/kubernetes-sigs/gateway-api/pull/2254

If this is going to continue to block the timeout features, maybe we should just switch the fields to uint32 millisecond values and live with the less nice API?

robscott:
@frankbu sorry this has taken so long! I think there's a rough consensus that we should aim for a format that is a subset of both OpenAPI/CEL and Go duration formats. That would include the following limitations:

1. No literal 0 ("0s" would be supported)
2. No fractional values
3. No compound values
4. Limited units (avoid day, week, and µs)

The remaining problem is if we can separate validation of formatting from min/max values. CEL should enable us to do that, but first we need to add CEL (x-ref https://github.com/kubernetes-sigs/gateway-api/pull/2253 and others)

frankbu:
> No literal 0 ("0s" would be supported)

This one doesn't really sound necessary to me. I understand that `0` in yaml will be assumed int (not string), but that just means you need to quote the value: `"0"`, which I'd rather do than add the s: `0s`. We could also do what Consul does and allow the value of "infinity" for no timeout, but personally, I prefer 0.

thockin:
I think a good first step would be to try to convince the openapi layer to validate fields of a specific type against the format we want.  I don't know where to do that, but I think we should shoot for a primitive field type.  E.g. `Timeout kubeapi.DurationString` such that it doesn't require any explicit validation in CEL (assuming any parseable value is OK).  Then you could use it with CEL validation to check the bounds, but not the format.

And if we think CEL is the path forward, maybe just adopting CEL's format is best?

kflynn:
@thockin, your suggestion of maybe starting with CEL's format sounded pretty sane to me. I spent some time yesterday looking over code for CEL and other extant formats, then spent some time ~argu~discussing things with @robscott (😇), and ultimately ended up wanting to recommend a slightly different answer for v0.8.0. 

tl;dr: Let's use the CEL duration type, restricted by a regex of `^([0-9]{1,5}(h|m|s|ms)){1,4}$`. That'll let us put some bounds on the input, while being compatible with a wide range of extant code and allowing us to use CEL for min/max validation.

Keep reading for all the details.

----

I found four different extant formats that seem obviously worth considering. (This starts out very Go-centric because Go is the implementation language for Kubernetes itself. Bear with me.)

### `metav1.Duration` (aka `time.ParseDuration`)

`k8s.io/apimachinery/pkg/apis/meta/v1` defines a `Duration` type. Its `Unmarshal` method calls `time.ParseDuration`. Per its docs:

    A duration string is a possibly signed sequence of decimal numbers, each with
    optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid
    time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".

So: negatives, multiple components, and floating-point are OK, but no units larger than hours are supported. Also note that where the doc says “µs”, both U+00B5 (`Micro Sign`) and U+03BC (`Greek Small Letter Mu`) are accepted, beautifully demonstrating the complexities of Unicode-enabled source code.

(Source: `https://github.com/golang/go/blob/master/src/time/format.go`)

### CEL (aka, again, `time.ParseDuration`)

CEL in the form of `google/cel-go` defines a `Duration` type that can be converted from a string using the `ConvertToType` function in `common/types/string.go`. That function simply calls `time.ParseDuration` and uses the result.

(Source: `https://github.com/google/cel-go/blob/master/common/types/string.go`)

### APIServer “OpenAPI” code (aka `strfmt.ParseDuration`)

Though the README for `k8s.io/apimachinery` says that the APIServer is one of its primary clients, `k8s.io/apiserver/pkg/cel/common/values.go` defines its own `UnstructuredToVal` that uses `strfmt.ParseDuration` to parse `string` values marked with the `duration` format.

`strfmt.ParseDuration` starts by calling `time.ParseDuration` on its input, and if that works, it uses the result – so everything supported by `time.ParseDuration` is also supported by `strfmt.ParseDuration`. If `time.ParseDuration` _doesn’t_ work, `strfmt.ParseDuration` will separately parse multiple components, but it **does not** support floating-point, and it **does** support “d” and “w” for days and weeks (neither of which take leap seconds or leap years into account). It also supports more spelled-out units (e.g. “milli” for “ms” and “hour” for “h”).

_(Why, if the file has `cel` in its name, do I call this “OpenAPI” code? Because this split between the `string`_ type _and the `duration`_ format _is an OpenAPI thing, not a CEL thing, and because the APIServer is described as using OpenAPI for CRD validation.)_

(Source: `https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/cel/common/values.go`)

### ISO8601/RFC3339

We would be remiss not to include ISO8601 here. It’s much less user-friendly than the Go duration types above, but it is _the_ standard representation of time for many use cases. An example ISO8601 duration is “P1DT1H1M1.1S” for 1 day, 1 hour, 1 minute, and 1.1 seconds: the leading “P” is required, no units smaller than seconds are supported (although fractional seconds are supported), and all the units are case-sensitive.

RFC3339 is mostly equivalent to ISO8601, but is case-insensitive. The leading “P” or “p” is still required.

## Recommendation

Given that everything above except ISO8601 has a common denominator of Go's `time.ParseDuration`, I think we should start with that format. It's already used in `apimachinery`, CEL, and the APIServer. It’s (obviously) supported in Go. It’s also supported in Rust in the `kube-rs` crate (as of [kube-rs pull 1224](https://github.com/kube-rs/kube/pull/1224), merged on 5 June 2023), and in Python in the `durationpy` package. These three languages should encompass the vast majority of the development in the Kubernetes ecosystem.

That being said, it definitely has downsides. Summarizing a lot of earlier discussion:

* Supporting floating-point values opens its own can of worms around precision, corner cases, normalization, etc., and `time.ParseDuration` doesn’t discuss these. Looking at the source, it uses `float64` (for everything, in fact, even if the number is actually written without a fractional part), and Go seems to just assume that saying “IEEE754” will cover the rest.

* Negative durations can easily come up when you’re doing calculations with time values, but it’s tough to think of a use case for specifying a negative duration. `time.ParseDuration`’s ability to handle a negative duration may never be relevant for us.

* `time.ParseDuration` doesn't support anything larger than hours. While timeouts are very unlikely to need days or weeks, the type is a `Duration`, not a `Timeout`, and days and weeks can be quite convenient when describing e.g. the lifetime of a TLS certificate (even though the disclaimer about leap seconds and days is critical).

* `time.ParseDuration` is extremely open-ended in what it accepts. For example, it doesn't limit the maximum length of the input or the number of individual components. It does have a maximum numeric value of 1<<63 for numeric values (both for individual components and for the overall sum), but that's not really a meaningful limit.

Given those downsides, some further restriction of `time.ParseDuration` seems in order. @robscott and I think that it would work well to only allow durations that match

    ^([0-9]{1,5}(h|m|s|ms)){1,4}$

which is to say:

* Negative values and floating-point numbers are not allowed.
* Each component must have a numeric value with a maximum of five digits.
* At most four components are allowed.
* Hours, minutes, seconds, and milliseconds are supported.
* Microseconds, days, and weeks are considered out of scope for, but could be added later.

That will allow us to put some bounds on the inputs while also just using the native CEL duration type (since anything that the regex matches can be parsed by `time.ParseDuration`, and that's what the CEL duration type uses). It should play nicely with using CEL for min/max validation, and it still allows "0s" for "no timeout".

Critically, it restricts the scope for v0.8.0 to something manageable. We can leave discussion of whether we want to support "d" and "w", for example, for v1.0.0; the only change we're talking about for v0.8.0 is adding a regex.

(For the curious, the maximum duration of `99999h99999m99999s99999ms` is a rather unwieldy way to express about 11.6 years.)

frankbu:
> (Dropping floating-point support already concerns me on this front, FTR.)

Personally, I think not supporting floating point is the much bigger chance of people getting burned with cut-and-paste. I think values like `1.5s` or `3.5h` are more common than `1s500ms` or `3h30m`.

> but going from 3h45m to 225m means I had to actually think about the math (🙂)

True, but once your talking more than a couple of hours (easy math), what's the chance that you still need finer granularity than an hour, e.g., `48h2m` 😃

Anyway, I'd be inclined to go with the more restrictive format, but it's not that big a deal either way IMO.

kflynn:
@frankbu Yeah, that is definitely my worry about floats. 😐 I'm more prepared to defend that, because [floats are pretty awful under the hood](https://floating-point-gui.de/), so I'm more invested in ditching it. (Even `0.1` can't actually be represented exactly in an IEEE754 float.)

So: [the proposal in GEP-2257](https://github.com/kubernetes-sigs/gateway-api/pull/2258) doesn't support floating-point at all, and I'd be delighted to land that and update GEP-1742 to use it... and if the screams about floating-point being gone are load, well, it's Experimental, we can figure out how to handle it. 🙂

robscott:
Thanks to everyone for the feedback here and to @kflynn for finalizing this with GEP #2257 and then replacing this with #2316. Going to close this one out.

/close

PR #2163, opened 2023-07-03
GEP 2162 - Supported features in GatewayClass Status

### LINKED ISSUE #2162: Gateway Class status should show supported features
Improve UX by enabling them to easily see what features their installed GatewayClass support. 

**Why this is needed**:
With the evolution of the Conformance tests suite and features, we want to;

1. Enable users easily see what features the implementation support
1. Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support

cc: @shaneutt

LiorLieberman:
/kind gep

This GEP proposes changes to GatewayClassStatus API to include the installed GatewayClass supported and unsupported features .
  
Fixes #2162 

```release-note
Allow end-users to see what features their installed GatewayClass support
```

### DIFF
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -61,11 +61,11 @@ type GatewayClassStatus struct {
     // +listType=map
     // +listMapKey=type
     // +kubebuilder:validation:MaxItems=8
-    // +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
     Conditions []metav1.Condition `json:"conditions,omitempty"`
 
     // SupportedFeatures is the features the GatewayClass support.
     // <gateway:experimental>
+    // +kubebuilder:validation:MaxItems=64
     SupportedFeatures []string `json:"supportedFeatures,omitempty"`
 }
 ```
@@ -132,7 +132,7 @@ Before we make the changes we need to;
 1. Move `ReferenceGrant`` to Extended
     * `ReferenceGrant` is not supported by all the implementations now and it will block them from reporting support for core route features.
     <!-- TODO(liorlieberman) find the link of sig-auth work -->
-    * The current plan is that `ReferenceGrant` is not going to go GA. sig-auth is working on something thats supposed to replace it.
+    * The current plan is that `ReferenceGrant` is not going to go GA. sig-auth is working on something thats supposed to replace it ([kubernetes/enhancements#3766](https://github.com/kubernetes/enhancements/issues/3766) for more info).
     * Even if it needs to go GA eventually, it is easier to move from Extended to Core than the other way around.
 
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -73,7 +73,7 @@ type GatewayClassStatus struct {
 
 ## Understanding SupportedFeatures field
 
-Its important to define how we read list of `SupportedFeatures` we report.
+Its important to define how we read the list of `SupportedFeatures` we report.
 
 We have no support features for core features. If an implementation reports a route name e.g `HTTPRoute` as a supportedFeature it means it supports all its core features.
 In other words, supporting the route's core features is a requirement for the implementation to say that it supports the route.
@@ -121,7 +121,7 @@ Every feature should:
 Conformance tests file names should try to follow the the `pascal-case-name.go` format.
 For example for `HTTPRoutePortRedirect` - the test file would be `httproute-port-redirect.go`.
 
-We should treat this requirement as "best effort" because we might have test files that checks combination of features and can not follow the same format.
+We should treat this requirement as "best effort" because we might have test files that checks combination of features and can't follow the same format.
 
 In any case, the conformance tests file names should be meaningful and easy to understand.
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -52,7 +52,6 @@ type GatewayClassFeatures struct {
 
     // ImplementationSpecific indicates the implementation-specific support level features.
     ImplementationSpecific FeaturesSupport `json:"extended,omitempty"`
-    
 }
 // GatewayClassStatus is the current status for the GatewayClass.
 type GatewayClassStatus struct {
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -37,9 +37,9 @@ This doc proposes to enhance the GatewayClassStatus API so implementations could
 
 Implementations **must** publish the supported features before Accepting the GatewayClass, or in the same operation.
 
-Implementations are free to decide how they manage this information - common approach could be to maintain static lists of supported features or using predefined sets.
+Implementations are free to decide how they manage this information. A common approach could be to maintain static lists of supported features or using predefined sets.
 
-Note: implementations must keep the published list sorted.
+Note: implementations must keep the published list sorted in ascending alphabetical order.
 
 ## API
 
@@ -74,8 +74,8 @@ type GatewayClassStatus struct {
 
 Its important to define how we read the list of `SupportedFeatures` we report.
 
-We have no support features for core features. If an implementation reports a route name e.g `HTTPRoute` as a supportedFeature it means it supports all its core features.
-In other words, supporting the route's core features is a requirement for the implementation to say that it supports the route.
+We have no supported features for core features. If an implementation reports a resource name e.g `HTTPRoute` as a supportedFeature it means it supports all its core features.
+In other words, supporting the resource's core features is a requirement for the implementation to say that it supports the resource.
 
 For Extended/Implementation-specific features we have the supported features names.
 
@@ -115,12 +115,12 @@ Every feature should:
 3. Not exceed 128 characters.
 4. Contain only letters and numbers
 
-#### Conformance tests names
+#### Conformance test names
 
 Conformance tests file names should try to follow the the `pascal-case-name.go` format.
 For example for `HTTPRoutePortRedirect` - the test file would be `httproute-port-redirect.go`.
 
-We should treat this requirement as "best effort" because we might have test files that checks combination of features and can't follow the same format.
+We should treat this guidance as "best effort" because we might have test files that check the combination of several features and can't follow the same format.
 
 In any case, the conformance tests file names should be meaningful and easy to understand.
 
@@ -129,7 +129,7 @@ In any case, the conformance tests file names should be meaningful and easy to u
 
 Before we make the changes we need to;
 
-1. Change the names of the support features and conformance tests that don't conform with the formatting rules.
+1. Change the names of the supported features and conformance tests that don't conform with the formatting rules.
 
 
 ## Alternatives
@@ -175,6 +175,6 @@ Once the GatewayClass features support are is published into the status we could
 
 We got some feedback that it will be useful to indicate what what Gateway API version the implementation supports. So when we have supported features published in the GatewayClass Status, users will also be able to understand that those are the supported features for a specific Gateway API version.
 
-This work is likely to require its own small GEP but ideally what this field would mean is that I support Max(vX.X). 
+This work is likely to require its own small GEP but ideally what this field would mean is that an implementation supports Max(vX.X). 
 
 The value of it is to provide a better user experience and also more foundation for tools to be able to warn for example when a GatewayClass and CRDs have mismatched versions.
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -9,9 +9,9 @@ This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kuberne
 
 ## Goals
 
-* Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+* Improve UX by enabling users to easily see what features the implementation (GatewayClass) support.
 
-* Automatically run conformance tests based on the supported features populated in GatewayClass status
+* Automatically run conformance tests based on the supported features populated in GatewayClass status.
 
 * Provide foundation for tools to block or warn when unsupported features are used.
 
@@ -38,7 +38,9 @@ Implementations are free to decide how they manage this information - common app
 ## API
 
 This GEP proposes API changes describes as follow:
-1. Update the `GatewayClassStatus` struct to include a string-represented list of `SupportedFeatures`.
+
+* Update the `GatewayClassStatus` struct to include a string-represented list of `SupportedFeatures`.
+
 
 ```go
 // GatewayClassStatus is the current status for the GatewayClass.
@@ -57,6 +59,8 @@ type GatewayClassStatus struct {
     Conditions []metav1.Condition `json:"conditions,omitempty"`
 
     // SupportedFeatures is the features the GatewayClass support.
+    // <gateway:experimental>
+    // +kubebuilder:validation:MaxItems=64
     SupportedFeatures []string `json:"SupportedFeatures,omitempty"`
 }
 ```
@@ -96,11 +100,9 @@ https://github.com/kubernetes-sigs/gateway-api/issues/1804
 
 ### Research the development of an unsupported feature warning/blocking mechanism
 Once the GatewayClass features support are is published into the status we could look into;
+
 1. Using the supported features in the webhook to validate or block attempts to apply manifests with unsupported features.
 
     * Developing such mechanism looks like it would have to include cross-resource validation. (checking the GatewayClass while trying to apply a Route for example). This comes with a lot of caveats and we will need consider it carefully.
 
 2. Build tooling to check and warn when unsupported features are used.
-
-## Open Questions
-N/A
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -23,7 +23,7 @@ The current  [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api
 With the evolution of the Conformance tests suite and features, we want to;
 
 1. Enable users easily see what features the implementation support
-1. Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+1. Improve UX and potentially provide clear errors when a user is trying to use an unsupported feature
 
 This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -20,7 +20,13 @@ This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kuberne
 
 ## Non-Goals
 
-* Validate correctness of supported features published by the implementation. Meaning we don't intend to verify whether the supported features reported by the implementation are indeed supported.
+* Validate correctness of supported features published by the implementation.
+    Meaning we don't intend to verify whether the supported features reported by
+    the implementation are indeed supported.
+
+    However, the supported features in the status of the GatewayClass should
+    make it very easy for any individual to run conformance tests against the
+    GatewayClass using our conformance tooling.
 
 ## Introduction
 
@@ -170,6 +176,8 @@ Once the GatewayClass features support are is published into the status we could
     * Developing such mechanism looks like it would have to include cross-resource validation. (checking the GatewayClass while trying to apply a Route for example). This comes with a lot of caveats and we will need consider it carefully.
 
 2. Build tooling to check and warn when unsupported features are used.
+
+3. Add a table with feature name and description to document what each feature means.
 
 ### Add Gateway API Version field to the GatewayClass Status
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -66,7 +66,6 @@ type GatewayClassStatus struct {
 
     // SupportedFeatures is the features the GatewayClass support.
     // <gateway:experimental>
-    // +kubebuilder:validation:MaxItems=64
     SupportedFeatures []string `json:"supportedFeatures,omitempty"`
 }
 ```
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -132,10 +132,6 @@ Before we make the changes we need to;
 1. Change the names of the support features and conformance tests that don't conform with the formatting rules.
 
 
-## Conformance Details
-
-(TBD)
-
 ## Alternatives
 
 ### Re-using ConformanceProfiles structs
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -27,6 +27,10 @@ With the evolution of the Conformance tests suite and features, we want to;
 
 This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
 
+Implementations **should** publish the supported/unsupported features before Accepting the GatewayClass.
+
+They can either maintain static lists of supported/unsupported features OR generate if from their conformance tests results/reports as part of their CI.
+
 ## API
 
 This GEP proposes API changes describes as follow:
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -40,12 +40,8 @@ This GEP proposes API changes describes as follow:
 ```go
 // FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
 type FeaturesSupport struct {
-	// SupportedFeatures indicates which extended features were flagged as
-	// supported by the implementation and tests will be attempted for.
 	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
 
-	// UnsupportedFeatures indicates which extended features the implementation
-	// does not have support for and therefore will not attempt to test.
 	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
 }
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -5,16 +5,15 @@
 
 ## TLDR
 
-This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class. 
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of Gateway API features supported by the installed GatewayClass. 
 
 ## Goals
 
-* Enable users easily see what features the implementation(gateway class) support
-* Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+* Improve UX by enable users to easily see what features the implementation(GatewayClass) support.
 
 ## Non-Goals
 
-* Validate correctness of supported features published by the implementation.
+* Validate correctness of supported features published by the implementation. Meaning we don't intend to verify whether the supported features reported by the implementation are indeed supported.
 
 ## Introduction
 
@@ -38,36 +37,37 @@ This GEP proposes API changes describes as follow:
 2. Update the `GatewayClassStatus` struct to include `GatewayClassFeatures`
 
 ```go
-// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+// FeaturesSupport contains the features supported by the installed GatewayClass.
 type FeaturesSupport struct {
-	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
-
-	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
+    SupportedFeatures []string `json:"supportedFeatures,omitempty"`
 }
 
 // GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
 type GatewayClassFeatures struct {
-	// Core indicates the core support level features. UnsupportedFeatures should be null for the implementation to be considered all conformant. 
+    // Core indicates the core support level features.
     Core FeaturesSupport `json:"core,omitempty"`
 
-	// Extended indicates the extended support level features.
+    // Extended indicates the extended support level features.
     Extended FeaturesSupport `json:"extended,omitempty"`
+
+    // ImplementationSpecific indicates the implementation-specific support level features.
+    ImplementationSpecific FeaturesSupport `json:"extended,omitempty"`
     
 }
 // GatewayClassStatus is the current status for the GatewayClass.
 type GatewayClassStatus struct {
-	// Conditions is the current status from the controller for
-	// this GatewayClass.
-	//
-	// Controllers should prefer to publish conditions using values
-	// of GatewayClassConditionType for the type of each Condition.
-	//
-	// +optional
-	// +listType=map
-	// +listMapKey=type
-	// +kubebuilder:validation:MaxItems=8
-	// +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
-	Conditions []metav1.Condition `json:"conditions,omitempty"`
+    // Conditions is the current status from the controller for
+    // this GatewayClass.
+    //
+    // Controllers should prefer to publish conditions using values
+    // of GatewayClassConditionType for the type of each Condition.
+    //
+    // +optional
+    // +listType=map
+    // +listMapKey=type
+    // +kubebuilder:validation:MaxItems=8
+    // +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
 
     // Features is the features supported by the GatewayClass broken down by support level.
     Features GatewayClassFeatures `json:"features,omitempty"`
@@ -93,7 +93,7 @@ That said, conformance profiles are still at experimental stage, we could explor
 The current plan for conformance profiles is to also include centralized reporting. (specific details don't matter now)
 We could wait for this to be implemented and instruct users to read from that source to determine what features their installed GatewayClass support.
 
-This does not cover a future piece of work we want to implement which is to warn/block users from applying a gw-api object if the installed GWC doesn't support it. (originally suggested in [#1804](https://github.com/kubernetes-sigs/gateway-api/issues/1804)). 
+This does not cover a future piece of work we want to implement which is to warn/block users from applying a Gateway API object if the installed GWC doesn't support it. (originally suggested in [#1804](https://github.com/kubernetes-sigs/gateway-api/issues/1804)). 
 
 
 ## References
@@ -103,8 +103,8 @@ https://github.com/kubernetes-sigs/gateway-api/issues/1804
 
 ## Future Work
 
-### Develop an unsupported feature warning/blocking mechanism
-Once the GatewayClass features support are is published into the status, we can use this to validate/warn/block offending configurations.
+### Research the development of an unsupported feature warning/blocking mechanism
+Once the GatewayClass features support are is published into the status, we might use this to validate/warn/block offending configurations.
 
 ## Open Questions
 N/A
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -9,7 +9,9 @@ This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kuberne
 
 ## Goals
 
-* Improve UX by enable users to easily see what features the implementation(GatewayClass) support.
+* Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+
+* Automatically run conformance tests based on the supported features populated in GatewayClass status
 
 ## Non-Goals
 
@@ -21,8 +23,9 @@ The current [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/
 
 Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
 
-1. Enable users to easily see what features an implementation supports
-1. Improve UX and potentially provide clear errors when a user is trying to use an unsupported feature
+1. Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+1. Automatically run conformance tests based on the supported features populated in GatewayClass status.
+1. Potentially provide clear errors when a user is trying to use an unsupported feature (more under [Future Work](#future-work)).
 
 This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -18,18 +18,18 @@ This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kuberne
 
 ## Introduction
 
-The current  [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
+The current [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
 
-With the evolution of the Conformance tests suite and features, we want to;
+Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
 
-1. Enable users easily see what features the implementation support
+1. Enable users to easily see what features an implementation supports
 1. Improve UX and potentially provide clear errors when a user is trying to use an unsupported feature
 
 This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
 
 Implementations **should** publish the supported/unsupported features before Accepting the GatewayClass.
 
-They can either maintain static lists of supported/unsupported features OR generate if from their conformance tests results/reports as part of their CI.
+They can either maintain static lists of supported/unsupported features OR generate them from their conformance tests results/reports as part of their CI.
 
 ## API
 
--- a/geps/gep-2162.md
+++ b/geps/gep-2162.md
@@ -1,0 +1,111 @@
+# GEP-2162: Supported features in GatewayClass Status
+
+[comment]: <> (change the issue reference)
+* Issue: [#2162](https://github.com/kubernetes-sigs/gateway-api/issues/2162)
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class. 
+
+## Goals
+
+* Enable users easily see what features the implementation(gateway class) support
+* Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+
+## Non-Goals
+
+* Validate correctness of supported features published by the implementation.
+
+## Introduction
+
+The current  [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
+
+With the evolution of the Conformance tests suite and features, we want to;
+
+1. Enable users easily see what features the implementation support
+1. Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+## API
+
+This GEP proposes API changes describes as follow:
+1. Adding `FeaturesSupport` and  `GatewayClassFeatures` structs. 
+2. Update the `GatewayClassStatus` struct to include `GatewayClassFeatures`
+
+```go
+// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+type FeaturesSupport struct {
+	// SupportedFeatures indicates which extended features were flagged as
+	// supported by the implementation and tests will be attempted for.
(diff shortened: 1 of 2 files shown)


geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class. 
+
+## Goals
+
+* Enable users easily see what features the implementation(gateway class) support
+* Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
```
shaneutt, line 13: :+1:

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Non-Goals
+
+* Validate correctness of supported features published by the implementation.
```
shaneutt, line 17: I think I get what you mean here, but would you expand on this a bit please just so I can be sure?

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Introduction
+
+The current  [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
```
shaneutt, line 21:
```suggestion
The current [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+```go
+// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+type FeaturesSupport struct {
+	// SupportedFeatures indicates which extended features were flagged as
+	// supported by the implementation and tests will be attempted for.
+	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
+
+	// UnsupportedFeatures indicates which extended features the implementation
+	// does not have support for and therefore will not attempt to test.
+	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
```
LiorLieberman, line 49: I chose to add both `SupportedFeatures` and `UnsupportedFeatures`, I think it is a better UX if the user also see what features are not supported in his GWC instead of comparing the supported ones with the gw-api documentation to see what's missing.  
Open to debate if anyone has a different opinion

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Non-Goals
+
+* Validate correctness of supported features published by the implementation.
+
+## Introduction
+
+The current  [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
+
+With the evolution of the Conformance tests suite and features, we want to;
```
shaneutt, line 23:
```suggestion
Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Enable users easily see what features the implementation support
```
shaneutt, line 25:
```suggestion
1. Enable users to easily see what features an implementation supports
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+1. Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
```
shaneutt, line 26:
```suggestion
1. Improve UX and potentially provide clear errors when a user is trying to use an unsupported feature
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // Features is the features supported by the GatewayClass broken down by support level.
+    Features GatewayClassFeatures `json:"features,omitempty"`
+}
+```
+
+## Conformance Details
+
+(TBD)
```
shaneutt, line 83: Is the intention to do this as a separate follow up PR?

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+With the evolution of the Conformance tests suite and features, we want to;
+
+1. Enable users easily see what features the implementation support
+1. Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **should** publish the supported/unsupported features before Accepting the GatewayClass.
+
+They can either maintain static lists of supported/unsupported features OR generate if from their conformance tests results/reports as part of their CI.
```
spacewander, line 32:
```suggestion
They can either maintain static lists of supported/unsupported features OR generate them from their conformance tests results/reports as part of their CI.
```

Is there any standard or guideline for the name of supported/unsupported features?

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+	// UnsupportedFeatures indicates which extended features the implementation
+	// does not have support for and therefore will not attempt to test.
+	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
+}
+
+// GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
+type GatewayClassFeatures struct {
+	// Core indicates the core support level features. UnsupportedFeatures should be null for the implementation to be considered all conformant. 
+    Core FeaturesSupport `json:"core,omitempty"`
```
spacewander, line 55: Bad indentations?

geps/gep-2162.md
spacewander: BTW, what is the granularity of the features?
For instance, if an implementation supports TLSRoute, does it mean the implementation needs to support both TLS Termination mode & TLS passthrough mode? Or each mode should have its own feature name?

LiorLieberman: > Is there any standard or guideline for the name of supported/unsupported features?
Constant values from [features.go](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go). for example - https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go#L118-L142

LiorLieberman: > BTW, what is the granularity of the features?
For instance, if an implementation supports TLSRoute, does it mean the implementation needs to support both TLS Termination mode & TLS passthrough mode? Or each mode should have its own feature name?

The [documentation](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig) and the [code](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go#L166) don't differentiate between support for TLS Termination and passthrough. They are both under TLSCore feature.

So to answer your question, for this specific q - yes, since the feature name would be TLSRoute is one feature.

A different example would be like HTTPRedirect where we have[ different feature names](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go#L126-L133) for different sub capabilities

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
+# GEP-2162: Supported features in GatewayClass Status
+
+* Issue: [#2162](https://github.com/kubernetes-sigs/gateway-api/issues/2162)
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class.
```
robscott, line 8: Tiny nit, I don't think we've used this abbreviation in our docs.

```suggestion
This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of Gateway API features supported by the installed GatewayClass. 
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+* Issue: [#2162](https://github.com/kubernetes-sigs/gateway-api/issues/2162)
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class. 
+
+## Goals
+
+* Enable users easily see what features the implementation(gateway class) support
```
robscott, line 12:
```suggestion
* Enable users easily see what features the implementation (GatewayClass) support
```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+* Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
+
```
robscott, line 14:
```suggestion
* Automatically run conformance tests based on the supported features populated in GatewayClass status

```

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+* Validate correctness of supported features published by the implementation.
+
+## Introduction
+
+The current [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
+
+Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
+
+1. Enable users to easily see what features an implementation supports
+1. Improve UX and potentially provide clear errors when a user is trying to use an unsupported feature
```
robscott, line 26: 1. Automatically run conformance tests based on the supported features populated in GatewayClass status

geps/gep-2162.md
robscott: I think it's going to be impossible for implementations to keep something like `UnsupportedFeatures` up-to-date. As we get more conformance profiles, I think the list of `UnsupportedFeatures` may become larger than the list of `SupportedFeatures` for some implementations - for example implementations that only support L4 or mesh would have a really long list here. Maybe more problematic, this is something where it's impossible to predict the future, so unless the latest version of an implementation is always deployed, and that implementation always keeps track of the full set of Gateway API features, there are going to be omissions here.

Ideally tooling such as gwctl, webhook, UX, etc should be able to calculate the list of UnsupportedFeatures based on the list of `SupportedFeatures` to help make this UX better, but I don't think we can expect individual implementations to be fully aware of every possible feature that could exist.

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+This GEP proposes API changes describes as follow:
+1. Adding `FeaturesSupport` and  `GatewayClassFeatures` structs. 
+2. Update the `GatewayClassStatus` struct to include `GatewayClassFeatures`
+
+```go
+// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+type FeaturesSupport struct {
+	// SupportedFeatures indicates which extended features were flagged as
+	// supported by the implementation and tests will be attempted for.
+	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
```
robscott, line 45:
```suggestion
	// SupportedFeatures indicates which extended and experimental features
	// are supported by this GatewayClass. When running conformance tests,
	// tests that include features not listed here will be skipped.
	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
```

robscott: Actually on further thought, maybe this is even broader than "extended and experimental"... @shaneutt and @mlavacca may have some thoughts here. I'm not really sure how we want to handle the broader categories/profiles like mesh implementations that don't support Gateways. Maybe we should just say "which features are supported by this GatewayClass"?

robscott (commented):
Thanks @LiorLieberman! A few comments but otherwise LGTM.

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+
+	// UnsupportedFeatures indicates which extended features the implementation
+	// does not have support for and therefore will not attempt to test.
+	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
+}
+
+// GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
+type GatewayClassFeatures struct {
```
pleshakov, line 53: because one of the goals is "Enable users easily see what features the implementation(gateway class) support", do we also need a way to report which implementation-specific features an implementation supports?

geps/gep-2162.md
```diff
@@ -0,0 +1,114 @@
...
+	// Core indicates the core support level features. UnsupportedFeatures should be null for the implementation to be considered all conformant. 
+    Core FeaturesSupport `json:"core,omitempty"`
```
pleshakov, line 55: We use `FeaturesSupport` type here, however, its documentation for `SupportedFeatures`  and `UnsupportedFeatures` fields is about "extended features", nothing about core features. Seems like an inconsistency

geps/gep-2162.md
LiorLieberman: I copy pasted the description, my bad. 

Essentially the way I envision it as I wrote the`GatewayClassFeatures` struct.

`SupportedFeatures` is just a list of strings but a higher is holding it. See `Core` and `Extended` under the `GatewayClassFeatures`

geps/gep-2162.md
LiorLieberman: right, the comment for `SupportedFeatures` and `UnsupportedFeatures` was wrong, I removed it.

LiorLieberman: resolving it but feel free to reopen if needed

geps/gep-2162.md
mlavacca: > I think it's going to be impossible for implementations to keep something like UnsupportedFeatures up-to-date. As we get more conformance profiles, I think the list of UnsupportedFeatures may become larger than the list of SupportedFeatures for some implementations - for example implementations that only support L4 or mesh would have a really long list here. Maybe more problematic, this is something where it's impossible to predict the future, so unless the latest version of an implementation is always deployed, and that implementation always keeps track of the full set of Gateway API features, there are going to be omissions here.

+1

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+type FeaturesSupport struct {
+	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
+
+	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
+}
+
+// GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
+type GatewayClassFeatures struct {
+	// Core indicates the core support level features. UnsupportedFeatures should be null for the implementation to be considered all conformant.
```
mlavacca, line 50: > UnsupportedFeatures should be null for the implementation to be considered all conformant. 

It actually depends, as an implementation can be conformant with `HTTPRoute`s (i.e., having the HTTP conformance profile badge), but not conformant with `TLSRoute`s.

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=8
+	// +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // Features is the features supported by the GatewayClass broken down by support level.
+    Features GatewayClassFeatures `json:"features,omitempty"`
```
mlavacca, line 73: Bad indentation here

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Conformance Details
+
+(TBD)
+
+## Alternatives
+
+### Re-using ConformanceProfiles structs
+
+We could use the same structs as we do in conformance profile object, more specifically, the [ProfileReport](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/apis/v1alpha1/profilereport.go#LL24C6-L24C19) struct.
```
mlavacca, line 85: conformance profiles rely on supported/unsupported features (they are at a higher level of abstraction), hence I think we should use the features for the GWC, as they can be considered as a source of truth.

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
+type GatewayClassFeatures struct {
+	// Core indicates the core support level features. UnsupportedFeatures should be null for the implementation to be considered all conformant. 
+    Core FeaturesSupport `json:"core,omitempty"`
+
+	// Extended indicates the extended support level features.
+    Extended FeaturesSupport `json:"extended,omitempty"`
+
```
mlavacca, line 55:
```suggestion
```

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of gw-api features supported by the installed Gateway Class. 
+
+## Goals
+
+* Enable users easily see what features the implementation(gateway class) support
+* Improving UX and potentially provide clear errors when a user is trying to use a feature the installed GWC doesn't support
```
howardjohn, line 13: How is this achieved by the GEP?

howardjohn (changes requested):

geps/gep-2162.md
LiorLieberman: I am not sure there is consensus around that yet. I think it needs more discussion. 

@shaneutt IIRC you had different ideas but do let me know if I am misremember and your are good to go with it.

geps/gep-2162.md
LiorLieberman: good question, I think so. 
Added `ImplementationSpecific`

geps/gep-2162.md
LiorLieberman: Removing `UnsupportedFeatures` as per the comment you and Rob left

geps/gep-2162.md
LiorLieberman: I agree here, it is also related to [Rob's comment](https://github.com/kubernetes-sigs/gateway-api/pull/2163#discussion_r1253441917). IIRC @shaneutt has different ideas in mind but if you both are good with this then I will add it.

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Instruct users to read from the future conformance profiles report
+
+The current plan for conformance profiles is to also include centralized reporting. (specific details don't matter now)
+We could wait for this to be implemented and instruct users to read from that source to determine what features their installed GatewayClass support.
+
+This does not cover a future piece of work we want to implement which is to warn/block users from applying a gw-api object if the installed GWC doesn't support it. (originally suggested in [#1804](https://github.com/kubernetes-sigs/gateway-api/issues/1804)). 
+
+
+## References
```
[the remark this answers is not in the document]
LiorLieberman: Not that I am aware of. 
@aojea pointed out this KEP https://github.com/kubernetes/enhancements/pull/2137/files which has similar stuff that was rejected. But it seems like the pushback was mainly for various other reasons which are not in scope here.

geps/gep-2162.md
```diff
@@ -0,0 +1,110 @@
...
+
+```go
+// FeaturesSupport contains the supported features and unsupported features by the installed GatewayClass.
+type FeaturesSupport struct {
+	SupportedFeatures []string `json:"supportedFeatures,omitempty"`
+
+	UnsupportedFeatures []string `json:"unsupportedFeatures,omitempty"`
+}
+
+// GatewayClassFeatures is the features supported by the GatewayClass broken down by support level.
```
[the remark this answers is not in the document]
LiorLieberman: I think the list can easily grow as we have more features in gw. especially when we have `Implementation-specific` features reported to status as well. 

My view is that it is a slightly better UX if the user know where to look for his Feature string instead of trying to find the feature he want to check within ~15+ features in one list.

open for feedback if you disagree

geps/gep-2162.md
LiorLieberman: This GEP is not achieving it now, thats why it is marked as potentially. Perhaps I should move it to Future work.
The ideas was to add it to the webhook validation.

But actually, after chatting with @robscott, I think it is going to be controversial as cross-resource validation is not super encouraged in k8s.

geps/gep-2162.md
LiorLieberman: not sure about that yet. I think we will not be able to test this but will revisit once the GEP is ready to merge

geps/gep-2162.md
LiorLieberman: done

geps/gep-2162.md
shaneutt: No I'm basically aligned with @mlavacca here, if we can just have one list of the actual features that can populate here, let's have one unified list.

LiorLieberman: wouldn't it be easier for the conformance profiles to have a list broken down by support level? One unified list would require another small bit of logic to determine support levels.

I am good with either

robscott: I think I lean towards one unified list. Although I don't expect it to be common, it's possible that some features currently labeled as "Extended" may eventually transition to "Core" or vice versa.

geps/gep-2162.md
```diff
@@ -0,0 +1,112 @@
...
+Though it would be nice to have only one place to update, these structs seems to include much more data relevant to the conformance report but not for our use case. 
+
+That said, conformance profiles are still at experimental stage, we could explore the option to create a shared struct that will be used both for the conformance reports and for the GatewayClass status.
+
+### Instruct users to read from the future conformance profiles report
+
+The current plan for conformance profiles is to also include centralized reporting. (specific details don't matter now)
+We could wait for this to be implemented and instruct users to read from that source to determine what features their installed GatewayClass support.
+
+This does not cover a future piece of work we want to implement which is to warn/block users from applying a Gateway API object if the installed GWC doesn't support it. (originally suggested in [#1804](https://github.com/kubernetes-sigs/gateway-api/issues/1804)).
```
robscott, line 98: I think this is now well defined in the Conformance Profiles GEP now, I'd recommend linking to that. I think the reasons for publishing this on GWC status are:

1. Possibility for automation to block or warn when unsupported features are used, regardless of GatewayClass. For example, the incoming gwctl tool could help with that and/or potentially a validating webhook.
2. Ability for users to select the right GatewayClass for their usage without needing to refer to documentation or conformance reports. 
3. Ability for implementers to be able to specify implementation-specific features to differentiate their GatewayClasses.

I think 1 is the primary use case, while 2 and 3 are more nice to have than anything.

geps/gep-2162.md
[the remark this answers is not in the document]
robscott: Mentioned in another comment as well, but I think I prefer a single list of all supported features here.

geps/gep-2162.md
```diff
@@ -0,0 +1,112 @@
...
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of Gateway API features supported by the installed GatewayClass. 
+
+## Goals
+
+* Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+
```
robscott, line 13:
```suggestion
* Provide foundation for tools to block or warn when unsupported features are used. 
```

robscott (commented):
Thanks @LiorLieberman! This mostly LGTM.

geps/gep-2162.md
mlavacca: > I think I lean towards one unified list. Although I don't expect it to be common, it's possible that some features currently labeled as "Extended" may eventually transition to "Core" or vice versa.

Yeah, it makes sense. I didn't consider it. +1

geps/gep-2162.md
LiorLieberman: Or you are talking about gwctl?

youngnick: By "foundation" here, do you mean a library @robscott? That seems like it could be a good place to start, especially if we put the caveats about cross-resource validation into the docs for the library.

geps/gep-2162.md
```diff
@@ -0,0 +1,112 @@
...
+Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
+
+1. Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+1. Automatically run conformance tests based on the supported features populated in GatewayClass status.
+1. Potentially provide clear errors when a user is trying to use an unsupported feature (more under [Future Work](#future-work)).
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **should** publish the supported/unsupported features before Accepting the GatewayClass.
```
youngnick, line 32:
```suggestion
Implementations **must** publish the supported/unsupported features before Accepting the GatewayClass.
```
"Should" implies optionality, which I don't think we want here. This feels like a "must" to me.

geps/gep-2162.md
```diff
@@ -0,0 +1,112 @@
...
+
+They can either maintain static lists of supported/unsupported features OR generate them from their conformance tests results/reports as part of their CI.
```
youngnick, line 34:
```suggestion
Implementations are free to decide how they manage this information - common approaches could be either to maintain static lists of supported/unsupported features OR generate them from their conformance tests results/reports as part of their CI.
```

We don't want to imply that people can't come up with some other way to do this, all we really care about is the end result (that is, that the information is in the status).

geps/gep-2162.md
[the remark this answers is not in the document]
youngnick: Yes, I think that we should ensure that the feature names indicate what object or objects they apply to, and only have a single list.

youngnick (commented):
This is looking pretty good to me aside from some small language clarifications and the already-discussed move to a single list of features.

geps/gep-2162.md
LiorLieberman: I see foundation as "populating the data" so for future tools(gwctl etc)/webhook it will have the features populated in the gwc status. Ad those features are the foundation for the gwctl or the webhook which might use them.

one example I can think of could be `gwctl get-supported-features  --class=<gwc-name>`

The webhook example would be, say a user try to apply a HttpRoute manifest with features that is not supported by the class. So the ideas was that the webhook would block/or warn. BUT the cross-resource validation has a lot of caveats especially about eventual consistency. So I would vote to avoid it but thats a future discussion

geps/gep-2162.md
LiorLieberman: done, PTAL

geps/gep-2162.md
LiorLieberman: Done but leaving this thread open in case you disagree or have questions @youngnick

youngnick: Yeah, I think that wording is fine for now.

geps/gep-2162.md
youngnick: I think not using the conformance structs themselves are fine, but we should put a requirement somewhere in this GEP that the feature names used for the supported features are the same or at least derivative of the conformance test names.

shaneutt (approved):

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
+# GEP-2162: Supported features in GatewayClass Status
```
robscott, line 1: Don't forget to make a corresponding change to mkdocs.yml

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of Gateway API features supported by the installed GatewayClass. 
+
+## Goals
+
+* Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+
+* Automatically run conformance tests based on the supported features populated in GatewayClass status
```
robscott, line 14: Tiny nit: Add trailing `.` to match other items in this list. 
```suggestion
* Automatically run conformance tests based on the supported features populated in GatewayClass status.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+* Issue: [#2162](https://github.com/kubernetes-sigs/gateway-api/issues/2162)
+* Status: Provisional
+
+## TLDR
+
+This GEP proposes to enhance the [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185) to include a list of Gateway API features supported by the installed GatewayClass. 
+
+## Goals
+
+* Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
```
robscott, line 12:
```suggestion
* Improve UX by enabling users to easily see what features the implementation (GatewayClass) support.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **must** publish the supported features before Accepting the GatewayClass.
+
+Implementations are free to decide how they manage this information - common approaches could be either to maintain static lists of supported features OR generate them from their conformance tests results/reports as part of their CI.
+
+## API
+
+This GEP proposes API changes describes as follow:
+1. Update the `GatewayClassStatus` struct to include a string-represented list of `SupportedFeatures`.
```
robscott, line 41: Nit: Our markdown parser requires an empty newline before the start of a list. On that note, do we need a list here? It seems like there's only one item?
```suggestion
This GEP proposes API changes describes as follows:

1. Update the `GatewayClassStatus` struct to include a string-represented list of `SupportedFeatures`.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+    //
+    // +optional
+    // +listType=map
+    // +listMapKey=type
+    // +kubebuilder:validation:MaxItems=8
+    // +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // SupportedFeatures is the features the GatewayClass support.
+    SupportedFeatures []string `json:"SupportedFeatures,omitempty"`
```
robscott, line 60:
```suggestion
    // <gateway:experimental>
    // +kubebuilder:validation:MaxItems=64
    SupportedFeatures []string `json:"SupportedFeatures,omitempty"`
```

geps/gep-2162.md
robscott: I don't think we need much here other than just a statement that conformance tests will read this information from GatewayClass unless it is specified in some other way.

geps/gep-2162.md
robscott: I think I'm agreeing with what @youngnick is saying above, but I'll state what I'm thinking just to make sure:

1. `GatewayClass.Status.SupportedFeatures` should use the same names verbatim as `SupportedFeatures` in conformance tests.
2. `SupportedFeatures` in conformance tests should be easy to understand, potentially even mapping to conformance test names.
3. As a result of 1 and 2, we should probably do an audit to be 100% sure we're happy with existing names for supported features because they're about to be part of our API Surface (and thus it will be near impossible to change them going forward).

Does that all line up with what others are thinking here?

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+* We could build a mechanism or a tool to block or warn when unsupported features are used.
+* Users will be able to select the GatewayClass that suits their needs without having to refer to documentation or conformance reports.
+
+This does not cover a future piece of work we want to implement which is to warn/block users from applying a Gateway API object if the installed GWC doesn't support it. (originally suggested in [#1804](https://github.com/kubernetes-sigs/gateway-api/issues/1804)). 
+
+
+## References
+
+https://github.com/kubernetes-sigs/gateway-api/discussions/2108
+https://github.com/kubernetes-sigs/gateway-api/issues/1804
```
robscott, line 93: Can you turn these into links? (Our markdown parser does not do that automatically).

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Future Work
+
+### Research the development of an unsupported feature warning/blocking mechanism
+Once the GatewayClass features support are is published into the status we could look into;
+1. Using the supported features in the webhook to validate or block attempts to apply manifests with unsupported features.
+
+    * Developing such mechanism looks like it would have to include cross-resource validation. (checking the GatewayClass while trying to apply a Route for example). This comes with a lot of caveats and we will need consider it carefully.
+
+2. Build tooling to check and warn when unsupported features are used.
```
robscott, line 103: The deploy preview is currently broken due to an issue parsing this file, but once that's resolved, I'd recommend ensuring that this section renders well here: https://deploy-preview-2163--kubernetes-sigs-gateway-api.netlify.app/geps/gep-2162/. I think you may need to add a new line and adjust some indentation, but not sure.

geps/gep-2162.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Open Questions
+N/A
```
robscott, line 106: Don't think we need this.

robscott (commented):
Thanks @LiorLieberman!

geps/gep-2162.md
youngnick: Agree with all three points here. As part of implementing this, we need to audit all the conformance test names, and probably even standardize a format for feature and/or conformance test names.

geps/gep-2162.md
LiorLieberman: changed it to "*"

geps/gep-2162.md
LiorLieberman: I looked at gep 1709 https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md?plain=1#L631C11-L632

The reference bit does look the same, what am I missing?

robscott: Looks like you found another bug 🙃: https://gateway-api.sigs.k8s.io/geps/gep-1709/#references

geps/gep-2162.md
```diff
@@ -0,0 +1,108 @@
...
+
+Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
+
+1. Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+1. Automatically run conformance tests based on the supported features populated in GatewayClass status.
+1. Potentially build tooling to block or warn when unsupported features are used (more under [Future Work](#future-work)).
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **must** publish the supported features before Accepting the GatewayClass.
```
youngnick, line 34: Maybe this? As written, this demands two GatewayClass updates, rather than one.
```suggestion
Implementations **must** publish the supported features before Accepting the GatewayClass, or in the same operation.
```

youngnick (commented):
Looks okay for `Provisional` to me, although I do think that the feature name and conformance test audit is probably necessary before we go too much further.

geps/gep-2162.md
LiorLieberman: I have opened https://github.com/kubernetes-sigs/gateway-api/discussions/2285 with some findings - PTAL !

geps/gep-2162.md
```diff
@@ -0,0 +1,108 @@
...
+    // +listType=map
+    // +listMapKey=type
+    // +kubebuilder:validation:MaxItems=8
+    // +kubebuilder:default={{type: "Accepted", status: "Unknown", message: "Waiting for controller", reason: "Pending", lastTransitionTime: "1970-01-01T00:00:00Z"}}
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+    // SupportedFeatures is the features the GatewayClass support.
+    // <gateway:experimental>
+    // +kubebuilder:validation:MaxItems=64
+    SupportedFeatures []string `json:"SupportedFeatures,omitempty"`
```
shaneutt, line 64: In a zoom, @LiorLieberman @robscott and myself discussed the possibility that this type would be something _other_ than a `[]string`. We want this list to be human readable in addition to machine readable, and so at minimum I think since features will be in the `<*Route><Feature>` format, it could be very helpful to human readers if this was sorted in lexicographical order.

```suggestion
    // SupportedFeatures is the features the GatewayClass support.
    //
    // Implements MUST sort this list in lexicographical order to help make it easier
    // to parse for human readers.
    //
    // <gateway:experimental>
    // +kubebuilder:validation:MaxItems=64
    SupportedFeatures []string `json:"SupportedFeatures,omitempty"`
```

youngnick: Having the features split by type probably does make sense, I think we might end up having some top-level features as well though ("Allow cross-namespace references using ReferenceGrant" feels like a likely candidate).

shaneutt: Wouldn't we just bucket that under `ReferenceGrant`? Some implementations today (I know of 2) explicitly do not support `ReferenceGrant` at all.

shaneutt: `ReferenceGrant` is a good point, _technically_ it's like a feature of every route type, but we wouldn't necessarily want to list it that way :thinking: would we?

geps/gep-2162.md
```diff
@@ -61,10 +67,76 @@ type GatewayClassStatus struct {
...
+In any case, the conformance tests file names should be meaningful and easy to understand.
+
+
+## Followups
+
+Before we make the changes we need to;
+1. Change the names of the support features and conformance tests that don't conform with the formatting rules.
+1. Move `ReferenceGrant`` to Extended
+    * `ReferenceGrant` is not supported by all the implementations now and it will block them from reporting support for core route features.
+    <!-- TODO(liorlieberman) find the link of sig-auth work -->
```
LiorLieberman, line 135: @robscott can you link it here?

robscott: https://github.com/kubernetes/enhancements/issues/3766

geps/gep-2162.md
LiorLieberman: Resolving this in favour of Robs PR here https://github.com/kubernetes-sigs/gateway-api/pull/2323 and the new formatting section I added. 

Feel free to reopen if you feel we need to discuss this more.

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+* Standardize features and conformance tests names.
+
+* Automatically run conformance tests based on the supported features populated in GatewayClass status.
+
+* Provide foundation for tools to block or warn when unsupported features are used.
+
+
+## Non-Goals
+
+* Validate correctness of supported features published by the implementation. Meaning we don't intend to verify whether the supported features reported by the implementation are indeed supported.
```
robscott, line 23: Non-blocking: Maybe could add a note that it should make it very easy for any individual to verify by running conformance tests against the GatewayClass.

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Introduction
+
+The current [GatewayClassStatus](https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/gatewayclass_types.go#L185 is only used to store conditions the controller publishes.
+
+Partnered with the [Conformance Profiles](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709.md) work, we want to:
+
+1. Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
```
robscott, line 31: Alongside this, maybe under future work, we should mention that we'll likely want to document what each feature means. Maybe this is actually just graduation criteria for graduation to standard?

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+1. Automatically run conformance tests based on the supported features populated in GatewayClass status.
+1. Potentially build tooling to block or warn when unsupported features are used (more under [Future Work](#future-work)).
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **must** publish the supported features before Accepting the GatewayClass, or in the same operation.
+
+Implementations are free to decide how they manage this information - common approach could be to maintain static lists of supported features or using predefined sets.
+
+Note: implementations must keep the published list sorted.
```
robscott, line 42:
```suggestion
Note: implementations must keep the published list sorted in ascending alphabetical order.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+1. Improve UX by enabling users to easily see what features the implementation(GatewayClass) support.
+1. Standardize features and conformance tests names.
+1. Automatically run conformance tests based on the supported features populated in GatewayClass status.
+1. Potentially build tooling to block or warn when unsupported features are used (more under [Future Work](#future-work)).
+
+This doc proposes to enhance the GatewayClassStatus API so implementations could publish a list of features they support/don't support.
+
+Implementations **must** publish the supported features before Accepting the GatewayClass, or in the same operation.
+
+Implementations are free to decide how they manage this information - common approach could be to maintain static lists of supported features or using predefined sets.
```
robscott, line 40:
```suggestion
Implementations are free to decide how they manage this information. A common approach could be to maintain static lists of supported features or using predefined sets.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+    // +kubebuilder:validation:MaxItems=64
+    SupportedFeatures []string `json:"supportedFeatures,omitempty"`
+}
+```
+
+## Understanding SupportedFeatures field
+
+Its important to define how we read the list of `SupportedFeatures` we report.
+
+We have no support features for core features. If an implementation reports a route name e.g `HTTPRoute` as a supportedFeature it means it supports all its core features.
```
robscott, line 77:
```suggestion
We have no supported features for core features. If an implementation reports a resource name e.g `HTTPRoute` as a supportedFeature it means it supports all its core features.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+In other words, supporting the route's core features is a requirement for the implementation to say that it supports the route.
```
robscott, line 78:
```suggestion
In other words, supporting the resource's core features is a requirement for the implementation to say that it supports the resource.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+#### Feature Names
+
+Every feature should:
+
+1. Start with the resource name. i.e HTTPRouteXXX
+2. Follow the PascalCase convention. Note that the resource name in the string should come as is and not be converted to PascalCase, i.e HTTPRoutePortRedirect and not HttpRoutePortRedirect.
+3. Not exceed 128 characters.
+4. Contain only letters and numbers
+
+#### Conformance tests names
```
robscott, line 118:
```suggestion
#### Conformance test names
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+
+Conformance tests file names should try to follow the the `pascal-case-name.go` format.
+For example for `HTTPRoutePortRedirect` - the test file would be `httproute-port-redirect.go`.
+
+We should treat this requirement as "best effort" because we might have test files that checks combination of features and can't follow the same format.
```
robscott, line 123:
```suggestion
We should treat this guidance as "best effort" because we might have test files that check the combination of several features and can't follow the same format.
```

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+
+In any case, the conformance tests file names should be meaningful and easy to understand.
+
+
+## Followups
+
+Before we make the changes we need to;
+
+1. Change the names of the support features and conformance tests that don't conform with the formatting rules.
```
robscott, line 132:
```suggestion
1. Change the names of the supported features and conformance tests that don't conform with the formatting rules.
```

Do we have any features or tests in mind? If not we may be able to remove this entirely.

geps/gep-2162.md
```diff
@@ -0,0 +1,180 @@
...
+
+    * Developing such mechanism looks like it would have to include cross-resource validation. (checking the GatewayClass while trying to apply a Route for example). This comes with a lot of caveats and we will need consider it carefully.
+
+2. Build tooling to check and warn when unsupported features are used.
+
+### Add Gateway API Version field to the GatewayClass Status
+
+We got some feedback that it will be useful to indicate what what Gateway API version the implementation supports. So when we have supported features published in the GatewayClass Status, users will also be able to understand that those are the supported features for a specific Gateway API version.
+
+This work is likely to require its own small GEP but ideally what this field would mean is that I support Max(vX.X).
```
robscott, line 178:
```suggestion
This work is likely to require its own small GEP but ideally what this field would mean is that an implementation supports Max(vX.X). 
```

robscott (commented):
Thanks @LiorLieberman! A few nits but mostly non blocking.

/approve

youngnick:
LGTM for provisional, but I won't do the command until @robscott's changes are done.

geps/gep-2162.md
LiorLieberman: not sure I understand. We should already have everything in our api documentation. 

Whats on your mind? a table with feature name and a link to our docs for this feature?

geps/gep-2162.md
youngnick: Yep, if you do this one, I will LGTM and this can go in.

geps/gep-2162.md
robscott: Yeah, something like a table with feature name and description would work well.

robscott:
Thanks @LiorLieberman!

/lgtm

PR #2170, opened 2023-07-07
feat: skip-tests flag added to conformance CLI

mlavacca:
**What type of PR is this?**

/kind feature
/area conformance

The `--skip-tests` flag has been added to the conformance CLI to enable tests opt-out when using it.

Fixes #

```release-note
The `--skip-tests` flag has been added to the conformance CLI to enable tests opt-out when using it.
```


shaneutt (approved):
While I do think at some point it would be nice if we were actually validating test names, that problem is a bigger scope and doesn't need to hold up this valuable, incremental improvement. :+1: 

/approve
/lgtm

PR #2179, opened 2023-07-11
Improve AttachedRoutes conformance test

### LINKED ISSUE #2058: Improve `AttachedRoutes` conformance test

At the moment, the [test cases](https://github.com/kubernetes-sigs/gateway-api/blob/4738c6c6e981d4a0e38c8c6120dc0198e4128351/conformance/tests/gateway-with-attached-routes.go) related to the `AttachedRoutes` are structured as follows:
- the first one has one `Gateway` with one listener and one `HTTPRoute` referring to it as a parent. The test verifies that the `status.listeners[0].AttachedRoutes` field gets the value 1 assigned.
- the second one has one `Gateway` with one listener and two `HTTPRoute`s referring to it as a parent. The test verifies that the `status.listeners[0].AttachedRoutes` field gets the value 2 assigned.

Alongside the above test cases, we should verify that a `gateway` with many listeners has the corresponding `AttachedRoutes` field assigned only if the Routes can really attach to them.

Example:
- create a `Gateway` with two listeners using two different protocols: HTTP and UDP
- create an `HTTPRoute`
- only the HTTP listener should have the `AttachedRoutes` field to 1, the UDP listener value should be 0.

levikobi:
**What type of PR is this?**

/kind test
/area conformance

Fixes #2058 

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -76,5 +76,39 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
 		})
+
+		t.Run("Gateway listener should not have route attached if route doesn't exist", func(t *testing.T) {
+			gwNN := types.NamespacedName{Name: "gateway-with-two-listeners-and-one-attached-route", Namespace: "gateway-conformance-infra"}
+			listeners := []v1beta1.ListenerStatus{
+				{
+					Name: v1beta1.SectionName("udp"),
+					SupportedKinds: []v1beta1.RouteGroupKind{{
+						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+						Kind:  v1beta1.Kind("UDPRoute"),
+					}},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionAccepted),
+						Status: metav1.ConditionTrue,
+						Reason: "", // any reason
+					}},
+					AttachedRoutes: 0,
+				},
+				{
+					Name: v1beta1.SectionName("http"),
+					SupportedKinds: []v1beta1.RouteGroupKind{{
+						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+						Kind:  v1beta1.Kind("HTTPRoute"),
+					}},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionAccepted),
+						Status: metav1.ConditionTrue,
+						Reason: "", // any reason
+					}},
+					AttachedRoutes: 1,
+				},
+			}
+
+			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
+		})
 	},
 }
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -77,7 +77,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
 		})
 
-		t.Run("Gateway listener should not have the route attached if the route is not allowed by the listener's protocol", func(t *testing.T) {
+		t.Run("Gateway listener should have attached route by specifying the sectionName", func(t *testing.T) {
 			gwNN := types.NamespacedName{Name: "gateway-with-two-listeners-and-one-attached-route", Namespace: "gateway-conformance-infra"}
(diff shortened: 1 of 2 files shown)


conformance/tests/gateway-with-attached-routes.go
```diff
@@ -76,5 +76,39 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
...
+			listeners := []v1beta1.ListenerStatus{
+				{
+					Name: v1beta1.SectionName("udp"),
+					SupportedKinds: []v1beta1.RouteGroupKind{{
+						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
+						Kind:  v1beta1.Kind("UDPRoute"),
+					}},
+					Conditions: []metav1.Condition{{
+						Type:   string(v1beta1.ListenerConditionAccepted),
+						Status: metav1.ConditionTrue,
```
sunjayBhatia, line 91: contour does not support UDPRoute or the UDP protocol on a Listener so this Listener will never be accepted, could we switch to a different protocol?

levikobi: I don't mind switching it to TCP
@shaneutt @mlavacca, are you guys ok with that?

shaneutt: Actually, I don't think we can be in a situation where `SupportGateway` as a feature implicit requires TCP or UDP awareness :thinking:

mlavacca: > Actually, I don't think we can be in a situation where SupportGateway as a feature implicit requires TCP or UDP awareness 🤔

I agree with you, I see two ways of getting over this problem:
1. move this test into a new file, add TLS as a feature, and use TLS as the protocol for the listener (I'm suggesting TLS as we already have some TLS tests).
2. change this test to check that the `SectionName` use is respected (one gateway with two listeners and one HTTPRoute that references only one of them by means the `SectionName` field, should result in only one listener with `AttachedRoutes` incremented)

WDYT?

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -76,5 +76,39 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
 		})
+
+		t.Run("Gateway listener should not have route attached if route doesn't exist", func(t *testing.T) {
```
mlavacca, line 80:
```suggestion
		t.Run("Gateway listener should not have the route attached if the route  is not allowed by the listener's protocol", func(t *testing.T) {
```

conformance/tests/gateway-with-attached-routes.go
levikobi: I can do both :)
Let me know what you prefer

levikobi: @mlavacca I've updated the test according to your 2nd suggestion, if you want the first one too please do let me know.

mlavacca: @shaneutt I'm curious to hear your thoughts on this one 🤔

sunjayBhatia: I like the sectionname approach since it is a core field 👍🏽

sunjayBhatia: seems like the test name will need to be updated since the protocol/route type isnt the differentiator now

levikobi: @sunjayBhatia I definitely agree. Changed the test name.
I'm resolving this conversion, but please do let me know if you think the new name is not descriptive enough.

robscott:
Thanks @levikobi! This looks reasonable to me, will defer to someone else for approval though.

/lgtm
/ok-to-test

conformance/tests/gateway-with-attached-routes.yaml
```diff
@@ -86,3 +86,54 @@ spec:
...
+  gatewayClassName: "{GATEWAY_CLASS_NAME}"
+  listeners:
+  - name: http-unattached
+    port: 8080
+    protocol: HTTP
+    allowedRoutes:
+      kinds:
+      - kind: HTTPRoute
+      namespaces:
+        from: Selector
```
mlavacca, line 105: Since this test is related to the `AttachedRoutes` and not to the `allowedRoutes` I'd lean toward using a more generic approach in relation to the `allowedRoutes` (`from: All` instead of `from: Selector`).

mlavacca (commented):
Apart from a minor nitpick, lgtm 👍

conformance/tests/gateway-with-attached-routes.yaml
levikobi: Sure mate, done

mlavacca (approved):
Thanks @levikobi!

/approve

LiorLieberman:
Thanks @levikobi ! 
/lgtm

Issue #2094, opened 2023-06-05
Conformance tests require a listener on port 8080


pleshakov:
**What happened**:

The Gateway spec states https://github.com/kubernetes-sigs/gateway-api/blob/c2d863bb06f63b069be58d6ae9dc79f22365f809/apis/v1beta1/gateway_types.go#L72-L77

However, one of the base conformance Gateways includes a listener on port 8080 -- https://github.com/kubernetes-sigs/gateway-api/blob/c2d863bb06f63b069be58d6ae9dc79f22365f809/conformance/base/manifests.yaml#L40

Since base Gateways always get deployed as part of running the conformance tests, that Gateway with port 8080 listener becomes a blocker for running conformance tests for implementations that do not support port 8080.

**What you expected to happen**:

Conformance tests only use ports 80 and 443 in base manifests.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

sunjayBhatia:
also note this test: https://github.com/kubernetes-sigs/gateway-api/blob/c2d863bb06f63b069be58d6ae9dc79f22365f809/conformance/tests/gateway-observed-generation-bump.go#L67-L74

(requires support for multiple ports for a single protocol at the moment, which Contour will support soon but does not now)

xtineskim:
i'd like to help w/ this!

PR #2184, opened 2023-07-12
Add SupportedFeature for port 8080 on Gateway

xtineskim:
**What type of PR is this?** /kind test /area conformance 

Conformance tests only use ports 80 and 443 in base manifests.  This PR removed the Gateway w/ port 8080 in the base.yaml manifest, adds it to the tests yaml that was using it, and adds a `SupportedFeature` for port 8080

Fixes #https://github.com/kubernetes-sigs/gateway-api/issues/2094

```release-note
NONE
```

### DIFF
--- a/conformance/utils/suite/features.go
+++ b/conformance/utils/suite/features.go
@@ -64,6 +64,8 @@ const (
 	//
 	//       See: https://github.com/kubernetes-sigs/gateway-api/issues/1780
 	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
+	// This option indicates that the Gateway can also use port 8080
+	SupportGatewayPort8080 SupportedFeature = "GatewayPort8080"
 )
 
 // StandardExtendedFeatures are extra generic features that implementations may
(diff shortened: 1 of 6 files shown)


conformance/utils/suite/features.go
```diff
@@ -64,6 +64,8 @@ const (
 	//
 	//       See: https://github.com/kubernetes-sigs/gateway-api/issues/1780
 	SupportGatewayClassObservedGenerationBump SupportedFeature = "GatewayClassObservedGenerationBump"
+	// This option indicates that the Gateway can also use port 8080
+	SupportGatewayPort8080 SupportedFeature = "GatewayPort8080"
```
xtineskim, line 68: not 100% sure if it should be added to this section - open to suggestions

robscott (commented):
Thanks @xtineskim!

/lgtm
/approve

mlavacca:
@robscott, is there any particular reason why this is not included in 0.8.0?

shaneutt:
I think the problem was we just didn't track it. We didn't branch for `v0.8.0` release candidates, so we should be all set in terms of inclusion, we'll just need to make sure we update the changelog.

PR #2185, opened 2023-07-14
docs(site-src): update the example for the matches

saltbo:
**What type of PR is this?**
/kind documentation

update the matches example to avoid misunderstanding

Fixes #

```release-note
NONE
```

### DIFF
--- a/site-src/api-types/httproute.md
+++ b/site-src/api-types/httproute.md
@@ -91,8 +91,8 @@ spec:
     - path:
         value: "/foo"
       headers:
-        values:
-          version: "2"
+      - name: "version"
+        value: "2"
     - path:
         value: "/v2/foo"
 ```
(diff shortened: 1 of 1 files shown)


site-src/api-types/httproute.md
```diff
@@ -85,14 +85,16 @@ Take the following matches configuration as an example:
...
-        version: "2"
-  - path:
-      value: "/v2/foo"
+spec:
+  rules:
+  - matches:
+    - path:
+        value: "/foo"
+      headers:
+        values:
```
sunjayBhatia, line 94: header match should be a list with at least header name and value as explicit fields (not a map like here), see https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPHeaderMatch

sunjayBhatia (changes requested):
left a comment on fixing the header match section

robscott:
Thanks @saltbo!

/lgtm
/approve

sunjayBhatia (approved):

PR #2199, opened 2023-07-18
adding support for more than one mirror backends + adding clarificati…

### LINKED ISSUE #2109: HttpRouteFilters Docs and webhook do not match.
**What happened**:
HttpRouteFilters Docs and webhook do not match. 

The docs/api field description say: 
https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/httproute_types.go#L204-L205

And it seems that the webhook does not allow multiple features from the same type (unless it is `HTTPRouteFilterExtensionRef`)
https://github.com/kubernetes-sigs/gateway-api/blob/f2cd9bb92b4ff392416c40d6148ff7f76b30e649/apis/v1beta1/validation/httproute.go#L137-L142

Expected both to match. 

Would vote for changing the webhook to allow and clarify in the docs that repeatable is implementation-specific.

cc: @howardjohn

LiorLieberman:
…on on mulitple same type filters within the same rule

**What type of PR is this?**

/kind bug

This adds permits HTTPRouteFilterRequestMirror filters within the same rule + improve documentation on multiple, same type filters. 

There is not reason to not permit multiple mirror and we can also add conformance tests for it 

It was also as requested in https://github.com/istio/api/pull/2805. 

Fixes #2109 

```release-note
enabling multiple HTTPRouteFilterRequestMirror filters in the same HTTP Rule
```

/assign @robscott 
/cc @howardjohn


apis/v1beta1/httproute_types.go
```diff
@@ -618,6 +617,8 @@ type HTTPRouteFilter struct {
 	// Requests are sent to the specified destination, but responses from
 	// that destination are ignored.
 	//
+	// This filter an be used multiple times within the same rule.
```
robscott, line 620:
```suggestion
	// This filter can be used multiple times within the same rule. Note that
	// not all implementations will be able to support mirroring to multiple
	// backends.
```

apis/v1beta1/httproute_types.go
```diff
@@ -201,8 +201,7 @@ type HTTPRouteRule struct {
 	// - Implementation-specific custom filters have no API guarantees across
 	//   implementations.
 	//
-	// Specifying a core filter multiple times has unspecified or
-	// implementation-specific conformance.
+	// Specifying the same filter multiple times is not supported unless explicitly indicated in the filter.
```
robscott, line 204: Nit: Mind wrapping line at 80 chars?

robscott (commented):
Thanks @LiorLieberman! Do you mind creating a follow up issue to add a conformance test + supported feature to cover this use case?

apis/v1beta1/httproute_types.go
LiorLieberman: thx!

LiorLieberman:
Thanks for reviewing @robscott. 

Done - https://github.com/kubernetes-sigs/gateway-api/issues/2210

LiorLieberman:
Is there an option for PR authors to set merge type squash?

robscott:
>Is there an option for PR authors to set merge type squash?

I think anyone should be able to do this:

/label tide/merge-method-squash

robscott:
Thanks @LiorLieberman! This LGTM, but since this is close to the territory of an API change I'd like @shaneutt or @youngnick to approve as well.

/lgtm

shaneutt:
Let's double check with some other community members:

/cc @sunjayBhatia @arkodg @mlavacca @michaelbeaumont @Xunzhuo

mlavacca (approved):

youngnick:
This generally looks good, but I think that we should definitely have a conformance test, with a feature name, for multiple mirror backends (`RouteMultipleMirrorBackends` maybe?).

Larger issue, I think that we may need to go through for each `Extended` field and add the feature name at some point, to make the link between conformance tests and field definitions more clear.

LiorLieberman:
Thanks @youngnick. Will definitely follow up with a conformance test and supported feature for it. There is already an opened issue. 

Once this is merged, @howardjohn will approve https://github.com/istio/api/pull/2805 and we will have it implemented in one implementation so we can check the conformance test validity as well.

Since we have a consensus, can we merge it ?

youngnick:
Thanks Lior, that sounds great. I'm happy for this to go ahead now.

Xunzhuo (approved):
/lgtm, this looks great : )

Issue #1883, opened 2023-03-27
GatewayObservedGenerationBump should have a configurable timeout


briantkennedy:
**What would you like to be added**:

A timeout for GatewayObservedGenerationBump that can be configured in the [TimeoutConfig](https://github.com/kubernetes-sigs/gateway-api/blob/e65388c2c25b8c2dd7bd7fd7a4ccf18b44e05bd7/conformance/utils/config/timeout.go#L21) resource.

**Why this is needed**:

This can cause test flakes if the controller is slow to respond.

PR #2203, opened 2023-07-19
conformance: fetch Gateway in GatewayMustHaveLatestConditions loop

skriss:
**What type of PR is this?**
/kind flake
/area conformance

Fetches the Gateway from the client each time through
the polling loop in GatewayMustHaveLatestConditions,
so updates to the conditions can be observed.

The current code is checking the same object over and
over again.

Updates #1883

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-modify-listeners.go
+++ b/conformance/tests/gateway-modify-listeners.go
@@ -55,7 +55,7 @@ var GatewayModifyListeners = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// verify that the implementation is tracking the most recent resource changes
-			kubernetes.GatewayMustHaveLatestConditions(t, s.TimeoutConfig, original)
+			kubernetes.GatewayMustHaveLatestConditions(t, s.Client, s.TimeoutConfig, gwNN)
 
 			all := v1beta1.NamespacesFromAll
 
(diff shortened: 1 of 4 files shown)


conformance/tests/gateway-modify-listeners.go
```diff
@@ -55,7 +55,7 @@ var GatewayModifyListeners = suite.ConformanceTest{
 			require.NoErrorf(t, err, "error getting Gateway: %v", err)
 
 			// verify that the implementation is tracking the most recent resource changes
-			kubernetes.GatewayMustHaveLatestConditions(t, s.TimeoutConfig, original)
+			kubernetes.GatewayMustHaveLatestConditions(t, s.Client, s.TimeoutConfig, gwNN)
```
sunjayBhatia, line 58: should we move this after the namespace ready check but before fetching the "original" gateway? (similar for the other test in this file)

skriss: It looks like `original` is only being used for spec fields, so shouldn't matter too much one way or another, but I'm fine changing the order for readability.

sunjayBhatia: yeah good point, the original Generation field shouldnt change since the resource shouldnt be updated from when we create it, so I'll defer to you if you think it makes the code more readable to fetch the Gateway after checking the status

sunjayBhatia (approved):
LGTM passing with Contour, but since Steve and I both are testing against the same implementation, should get a review from someone else as well

conformance/tests/gateway-modify-listeners.go
skriss: https://github.com/kubernetes-sigs/gateway-api/pull/2203/commits/991056d97d2d192866d08ee5dfa9e66abda2c047

robscott:
Thanks @skriss!

/lgtm
/approve

Issue #2198, opened 2023-07-18
Dynamic header matching in HTTPRoute


frankbu:
**What would you like to be added**:
A standard API for routing based on matching metadata derived from dynamically evaluated request headers.

**Why this is needed**:
Users want to be able to route HTTP requests based on a validated JWT associated with the request. \
Istio, for example, supports this by using header names with a special prefix (`@`) in a httproute header match configuration, to indicate that the matching is not against a static header value, but rather the metadata associated with the validated JWT: https://istio.io/latest/docs/tasks/security/authentication/jwt-route/#configuring-ingress-routing-based-on-jwt-claims

Although Istio could support this in Gateway API routes with some implementation specific (e.g., prefix) solution, it would be better to define an optional standard API for this kind of routing in Gateway API.

Maybe simply allowing some prefix (e.g., `@`) in a header name to allow for  implementation-specific dynamic header matching would be minimally sufficient? Beyond that, defining some standard dynamic header names (like the one to use to match JWT metadata, for example) would also be nice to have in the Gateway spec.

brianehlert:
This came up in a conversation between myself, @kflynn and @robscott a few weeks back and I have been slow at starting the write-up.
Looking across existing implementations (not all Gateway API of course) there are respectable differences that need to be surfaced.  
Not to mention there are the two use cases of validation and claims checking.
Let me see how quickly I can get my current (and not polished) write-up checked in for review and to start the broader process.

pleshakov:
I wonder if combined with here https://github.com/kubernetes-sigs/gateway-api/issues/2166 , a generalized case here is implementation specific routing - data plane provides some parameters which can be used to make a routing decision.

Perhaps something like this could work:

Client IP based routing:
```yaml
  - matches:
      proxyKeys:
      - name: $client_ip # proxy specific
        type: CIDR # proxy specific
        value: 192.168.0.0
    backendRefs:
    - name: backend-local
      port: 80
```
JWT claim based routing:
```yaml
  - matches:
      proxyKeys:
      - name: $jwt_claim_group # proxy specific
        type: Exact # proxy specific
        value: group-1
    backendRefs:
    - name: backend-group1
      port: 80
```
Cookie based routing:
`````yaml
  - matches:
      proxyKeys:
      - name: $cookie_version # proxy specific
        type: Exact # proxy specific
        value: v2
    backendRefs:
    - name: backend-v2
      port: 80
```

brianehlert:
[...]

PR #2213, opened 2023-07-21
initialize JWT GEP

brianehlert:
**What type of PR is this?**
/kind gep

This is the introduce the JWT Policy concept into Gateway API

Fixes [2198](https://github.com/kubernetes-sigs/gateway-api/issues/2198)

```release-note
NONE
```

### DIFF
--- a/geps/gep-2198.md
+++ b/geps/gep-2198.md
@@ -3,7 +3,7 @@
 ## TLDR
 
 This document is not intended to describe how a proxy implements JWT handling behavior, there is an RFC that addresses that.
-Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as par of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
+Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as part of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
 
 ## Goals
 
@@ -53,8 +53,8 @@ Here are some samples of JWT implementations in a similar problem space.  Each i
 
 - NGINX Ingress Controller:
 
-  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret)
-  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location)
+  - [https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret)
+  - [https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location)
   
 - Kong:
 
@@ -62,11 +62,13 @@ Here are some samples of JWT implementations in a similar problem space.  Each i
 
 - Envoy Gateway:
 
-  - [https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider)
+  - [https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider)
 
 - Istio:
 
   - [https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/](https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/)
+  - [https://istio.io/latest/docs/tasks/security/authentication/jwt-route/](https://istio.io/latest/docs/tasks/security/authentication/jwt-route/)
+  - [https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/](https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/)
 
 ## Properties of a JWT Validation Policy
 
--- a/geps/gep-2198.md
+++ b/geps/gep-2198.md
@@ -1,0 +1,103 @@
+# JWT Policy
+
+## TLDR
+
+This document is not intended to describe how a proxy implements JWT handling behavior, there is an RFC that addresses that.
+Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as par of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
+
+## Goals
+
+- Define how JWT is used to form a common understanding and establish a common language
+- Identify unique use cases for JWT with proxies
+- Propose some properties to form the basis for discussion
+
+## Non-Goals
+
+- Define how proxies implement the JWT standards
+
+## What is a JWT
+
+A [JSON web token](https://www.rfc-editor.org/rfc/rfc7519) is a common pattern of presenting authorization data (claims) between a front end or client system and a back end.
+A JWT by itself it is only one piece of the [authentication triad](https://csrc.nist.gov/glossary/term/Multi_Factor_Authentication#:~:text=The%20three%20authentication%20factors), which is "something you have".  We will assume that some other process satisfied "something you are" and "something you know".  And thus the JWT is generally considered the result of some type of authentication or formation of a trust.
+
+JWT's can be encrypted, have methods for validation, can contain a variety of claims, and be used in gate keeping (validation) and authorization (claims matching) workflows.
+
+## How is a JWT used
+
+In practice a JWT is presented to an authorization layer to make a decision of whether or not to grant access to a specific hostname or application path (backend service) or also possibly used by the backend service to decide if specific secured data can be presented.
+This bring forth a set of common workflows which in turn present the use cases. In this example we will walk through the life of an encrypted JWT (though not all JWTs are encrypted)
+
+- client obtains a JWT
+- client presents the JWT with the request to a Gateway
+- the Gateway decrypts the JWT and validates it
+- now that the JWT is proven valid it is implicitly trusted and the request processing continues
+
+At this point a number of and/or paths can take place, most commonly a combination of the following happens:
+
+- the token content is tested for the existence of some claim property or value
+- claim values may be extracted and saved as header value(s) to the request
+- the entire token or a sub-set of claims may be forwarded directly to backend services for evaluation
+
+## Persona(s)
+
+The Personas that can be involved in managing JWT Policy can be:
+
+- platform / cluster operations
+- application developer / admins
+- user account manager
+- security operations
+
+## Prior related art
+
+Here are some samples of JWT implementations in a similar problem space.  Each implementation differs in terminology and implementation.
+
+- NGINX Ingress Controller:
+
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret)
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location)
+  
+- Kong:
+
+  - [https://docs.konghq.com/hub/kong-inc/jwt/configuration/](https://docs.konghq.com/hub/kong-inc/jwt/configuration/)
+
+- Envoy Gateway:
+
+  - [https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider)
+
+- Istio:
+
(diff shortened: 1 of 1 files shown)


robscott:
Thanks @brianehlert!

/ok-to-test

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+
+name
+token
+jwksURI
+keycache
+
+## Properties of a JWT Claims (authorization) Policy
+
+- token - the token that needs to be evaluated (this is the result of the validation Policy or an OIDC workflow)
+- claimsMatch - an array of claim properties (exists, not exists) or match value that must match to allow the request to proceed.
```
frankbu, line 101: In addition to allowing the request to proceed, Istio provides access to the evaluated token in route rules, so that users can also do things like weighted routing based on the claims. See https://istio.io/latest/docs/tasks/security/authentication/jwt-route/

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+
+Processing claims, such as extracting a value from the token and setting it as a header for a backend should not be part of a policy.
```
frankbu, line 103: Istio does support this in an authorization policy. It would be good to provide optional support or an extension to allow it? https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
+# JWT Policy
+
+## TLDR
+
+This document is not intended to describe how a proxy implements JWT handling behavior, there is an RFC that addresses that.
+Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as par of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
```
frankbu, line 6:
```suggestion
Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as part of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
```

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+
+  - [https://docs.konghq.com/hub/kong-inc/jwt/configuration/](https://docs.konghq.com/hub/kong-inc/jwt/configuration/)
+
+- Envoy Gateway:
+
+  - [https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider)
+
+- Istio:
+
+  - [https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/](https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/)
```
frankbu, line 69:
```suggestion
  - [https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/](https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/)
  - [https://istio.io/latest/docs/tasks/security/authentication/jwt-route/](https://istio.io/latest/docs/tasks/security/authentication/jwt-route/)
  - [https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/](https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/)
```

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+
+## TLDR
+
+This document is not intended to describe how a proxy implements JWT handling behavior, there is an RFC that addresses that.
+Rather, this document is to form a common understanding or JWT use cases as it relates to configuring and handling JWTs as par of a workflow as it relates to a Gateway and establish a framework from which to build the common parts of the Policy.
+
+## Goals
+
+- Define how JWT is used to form a common understanding and establish a common language
+- Identify unique use cases for JWT with proxies
```
costinm, line 11: Is Oauth or OIDC in scope ?

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret)
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location)
+  
+- Kong:
+
+  - [https://docs.konghq.com/hub/kong-inc/jwt/configuration/](https://docs.konghq.com/hub/kong-inc/jwt/configuration/)
+
+- Envoy Gateway:
+
+  - [https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/v0.4.0/api/extension_types.html#jwtauthenticationfilterprovider)
```
spacewander, line 65:
```suggestion
  - [https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider)
```

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+- security operations
+
+## Prior related art
+
+Here are some samples of JWT implementations in a similar problem space.  Each implementation differs in terminology and implementation.
+
+- NGINX Ingress Controller:
+
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-local-kubernetes-secret)
+  - [](https://docs.nginx.com/nginx-ingress-controller/configuration/policy-resource/#jwt-using-jwks-from-remote-location)
```
spacewander, line 57: Should add link in `[]`?

geps/gep-2198.md
```diff
@@ -0,0 +1,103 @@
...
+## Non-Goals
+
+- Define how proxies implement the JWT standards
+
+## What is a JWT
+
+A [JSON web token](https://www.rfc-editor.org/rfc/rfc7519) is a common pattern of presenting authorization data (claims) between a front end or client system and a back end.
+A JWT by itself it is only one piece of the [authentication triad](https://csrc.nist.gov/glossary/term/Multi_Factor_Authentication#:~:text=The%20three%20authentication%20factors), which is "something you have".  We will assume that some other process satisfied "something you are" and "something you know".  And thus the JWT is generally considered the result of some type of authentication or formation of a trust.
+
+JWT's can be encrypted, have methods for validation, can contain a variety of claims, and be used in gate keeping (validation) and authorization (claims matching) workflows.
```
spacewander, line 23: Maybe we can also mention JWE.

geps/gep-2198.md
brianehlert: I see OIDC/OAuth2 as its own policy.
It is the "how you obtain a token" part of the process if the proxy is the point of auth.

geps/gep-2198.md
brianehlert: Does JWE need to be called out specifically?
Since it is part of the JWT spec.  But the Policy does not / should not care - IMHO

geps/gep-2198.md
brianehlert: I don't see claim-to-header as a Policy function.  To me that is a Header CRUD type of action.
Although I can see how an 'extract' type of definition.
<header>:<value from claim> would be an interesting ease of use in the Validation Policy.

"validate the claim, if it passes, copy these values to these headers for the backends"

It would need to handle discrete values as well as a library.  In the case that a client already authenticated with an OIDC/OAuth2 IdP and passes the JWT with the request.  It is not uncommon for a specific realm to be extracted into either another token or as part of a header value.
This could get messy as it is approaching programmability behavior.  In the case of nginx we start dealing with well known variables such as $http_, $arg_, and $cookie_

geps/gep-2198.md
brianehlert: The client can always authenticate before the proxy and include the JWT from the OIDC/OAuth2 IdP with the request rather than have the proxy handle the OIDC workflow.
If the OIDC workflow happens at the proxy then there would be no JWT validation Policy, but there could still be an authorization Policy.  Or an action to transpose token values into headers for backends.

geps/gep-2198.md
brianehlert: I have historically considered this a claims match modifier (Policy) on a path/route.
First the path/route must match, then the claim is tested for allow/deny
I am assuming the same implementation.
This is where I see the authorization Policy serving the function of defining the allow claim(s) match.

brianehlert: I do like the pattern with:
`    outputClaimToHeaders:
    - header: "x-jwt-claim-foo"
      claim: "foo"`
What I am thinking about is if there is a way to have the independent of the jwksUri, so that it stands alone as a transformation action pattern that could be used for both OIDC and JWT.  Since both can require the same claim-to-header, but both don't require the jwksUri.
Or the pattern is simply repeated when the OIDC Policy comes...

geps/gep-2198.md
spacewander: I see.

youngnick:

shaneutt:
@brianehlert this can be rebased now as per the GEP move we did a couple months back. Let us know if you run into any trouble.

geps/gep-2198.md
```diff
@@ -0,0 +1,105 @@
...
+
+## How is a JWT used
+
+In practice a JWT is presented to an authorization layer to make a decision of whether or not to grant access to a specific hostname or application path (backend service) or also possibly used by the backend service to decide if specific secured data can be presented.
+This bring forth a set of common workflows which in turn present the use cases. In this example we will walk through the life of an encrypted JWT (though not all JWTs are encrypted)
```
costinm, line 28: Not sure why it's focused on encrypted JWTs - they are quite rare. Maybe focus on the common case, and have a separate doc or section for the less common ?

geps/gep-2198.md
```diff
@@ -0,0 +1,105 @@
...
+- client obtains a JWT
+- client presents the JWT with the request to a Gateway
+- the Gateway decrypts the JWT and validates it
+- now that the JWT is proven valid it is implicitly trusted and the request processing continues
+
+At this point a number of and/or paths can take place, most commonly a combination of the following happens:
+
+- the token content is tested for the existence of some claim property or value
+- claim values may be extracted and saved as header value(s) to the request
+- the entire token or a sub-set of claims may be forwarded directly to backend services for evaluation
```
costinm, line 39: Important use: a different token signed by the gateway is forwarded to backends ( to prevent backends from accessing the original JWT which may have escalation risks). That's sometimes called 'mint'

geps/gep-2198.md
```diff
@@ -0,0 +1,105 @@
...
+
+  - [https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider](https://gateway.envoyproxy.io/latest/api/extension_types.html#jwtauthenticationfilterprovider)
+
+- Istio:
+
+  - [https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/](https://istio.io/latest/docs/tasks/security/authorization/authz-jwt/)
+  - [https://istio.io/latest/docs/tasks/security/authentication/jwt-route/](https://istio.io/latest/docs/tasks/security/authentication/jwt-route/)
+  - [https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/](https://istio.io/latest/docs/tasks/security/authentication/claim-to-header/)
+
+## Properties of a JWT Validation Policy
```
costinm, line 73: I think it would be ideal to decouple JWT validation - as in 'is signature correct' ( same as mTLS handshake checks ) and the authorization parts ( which claims, paths, etc need to be matched to authorize the user ). 

For the later - it is best to have authorization decisions independent of the authn, and not mix it here.

geps/gep-2198.md
```diff
@@ -0,0 +1,105 @@
...
+
+The validation policy is two fold, it establishes the relationship with the Token and how to validate it.
+According to [https://datatracker.ietf.org/doc/html/rfc7519#section-4.1](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) claims cannot be assumed mandatory.  At a minimum a claim should be able to be validated following the process outlined here which only requires a method to decrypt if necessary: [https://datatracker.ietf.org/doc/html/rfc7519#section-7.2](https://datatracker.ietf.org/doc/html/rfc7519#section-7.2)
+
+The statements made above point to the primary points of divergence between the JWT validation implementations in the prior art section.
+
+Some very high level proposed properties to open the conversation based on comparing the implementations and reviewing the RFC.
+
+- name / realm (required) - the name / realm of the JWT, this is commonly a string representing a specific application.  It is seen used in response and log messages.
```
costinm, line 82: Not sure why name/realm would be required. Log messages are important but not that important. 

Issuer field is probably more useful as a key. 

For jwksURI - it should be mentioned 'if missing, implementation should fetch using OIDC'. And perhaps have a way to specify the public keys directly. 

Not sure if keyCache is that important ( implementation detail ), and 'token' is a very bad name ( it usually means a secret ). If you mean the 'header' or 'parameter' - better be explicit.

geps/gep-2198.md
```diff
@@ -0,0 +1,105 @@
...
+
+name
+token
+jwksURI
+keycache
+
+## Properties of a JWT Claims (authorization) Policy
+
+- token - the token that needs to be evaluated (this is the result of the validation Policy or an OIDC workflow)
+- claimsMatch - an array of claim properties (exists, not exists) or match value that must match to allow the request to proceed.
```
costinm, line 103: I think claim match is already in the realm of authorization. 

It doesn't scale and doesn't work well in practice to have 'name=user1' ... or name=user100' in policies that specify which issuers to trust. 

And there are many other authn methods ( access tokens with Oauth, client certs, user:password, apikeys ) that result in 'claims' associated with a principal. We shouldn't have a duplicated or divergent authorization depending on how a user may choose or be able to authenticate.

PR #2228, opened 2023-07-24
Clarify necessity for implementations to handle GWC .status

### LINKED ISSUE #1822: Clarify GatewayClass Support Requirements
We need to clearly state that implementations MUST populate status on all GatewayClasses in a cluster that specify their controller name. This does not necessary require implementations to implement the config specified by arbitrary GatewayClasses, just to at least communicate the status via GatewayClass.

**Why this is needed**:
This was initially surfaced in https://github.com/kubernetes-sigs/gateway-api/pull/1781. We need to clarify our expected behavior here in the API Spec before adding a required conformance test.

shaneutt:
**What type of PR is this?**

/kind documentation

Clarifies that we expect implementations to handle the `.status` for `GatewayClass` resources which carry their `controllerName`.

Resolves #1822 in support of #1781.

### DIFF
--- a/apis/v1beta1/gatewayclass_types.go
+++ b/apis/v1beta1/gatewayclass_types.go
@@ -57,6 +57,9 @@ type GatewayClass struct {
 
 	// Status defines the current state of GatewayClass.
 	//
+	// Implementations MUST populate status on all GatewayClass resources which
+	// specify their controller name.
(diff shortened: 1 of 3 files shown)


robscott:
Thanks @shaneutt!

/lgtm

PR #2238, opened 2023-07-26
Initial GAMMA docs take

kflynn:
Initial cut of GAMMA website updates

Signed-off-by: flynn@buoyant.io

/kind documentation

```release-note
NONE
```


howardjohn (commented):
This is awesome :+1: :+1: 

I left a bunch of nitpicks, nothing blocking at all though

site-src/concepts/api-overview.md
```diff
@@ -279,29 +278,97 @@ relationships between the different resources:
...
+2. The data plane locates the Service matching `foo` in Namespace `ns`.
+3. The data plane locates Routes associated with the `foo` Service.
+4. If the request does not match any Route, it is rejected.
+5. The data plane uses the `backendRefs` of the highest-priority matching
+   Route to select a destination workload.
+6. The data plane routes the request on to the destination workload (most
+   likely using [endpoint routing], but it is allowed to use [Service
+   routing]).
+
+[east/west]:/concepts/glossary#east-west-traffic
```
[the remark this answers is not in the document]
kflynn: 🤦‍♂️ Thanks for that catch! Fixed everywhere.

site-src/concepts/api-overview.md
```diff
@@ -279,29 +278,97 @@ relationships between the different resources:
...
+- Different Namespaces <a name="consumer-routes"></a>
+
+    A Route in a different Namespace than its Service is called a [consumer
+    route]. Typically, this is a Route meant to refine how a consumer of a
+    given workload makes request of that workload (for example, configuring
+    custom timeouts for that consumer's use of the workload). This Route will
+    only affect requests from workloads in the same Namespace as the Route.
+
+    There is ongoing [GAMMA] work around the relationship between producer
+    routes and consumer routes. Note also that it is not currently possible to
```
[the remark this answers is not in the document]
kflynn: Thankfully, this is a minor language comment, thanks! 🙂

site-src/concepts/api-overview.md
```diff
@@ -279,29 +278,97 @@ relationships between the different resources:
...
+
+When one or more Routes are attached to a Service, requests that do not match
+at least one of the Routes will be rejected.
+
+### Request flow
+
+A typical [east/west] API request flow is:
+
+1. A client workload makes a request to <http://foo.ns.service.cluster.local>.
+2. The data plane locates the Service matching `foo` in Namespace `ns`.
```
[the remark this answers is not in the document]
kflynn: I like both these suggestions, yeah. Updated.

site-src/concepts/api-overview.md
```diff
@@ -4,16 +4,13 @@ This document provides an overview of Gateway API.
...
 ## Roles and personas
 
-There are 3 primary roles in Gateway API:
+There are 3 primary roles in Gateway API, as described in [roles and personas]:
 
-- Infrastructure Provider
-- Cluster Operator
-- Application Developer
+- **Ian** (he/him): Infrastructure Provider
+- **Charlie** (they/them): Cluster Operator
```
youngnick, line 10: @candita had the great suggestion to use "Chan" here, which is less western, but still gender-neutral. If we agree here, I'll update #2128 as well.

site-src/concepts/glossary.md
```diff
@@ -0,0 +1,66 @@
+# Gateway API Glossary
+
+### Consumer Route
+
+A Route bound to a workload's Service by a consumer of a given workload,
+refining the specific consumer's use of the workload.
+
+### Gateway Controller
```
youngnick, line 8: Maybe add a "Gateway Implementation" as a more general term that's anything that implements the spec, no matter where it runs?

(At least the GKE implementation doesn't run in the clusters that it manages, and it's definitely not a requirement, just common).

site-src/concepts/use-cases.md
```diff
@@ -0,0 +1,177 @@
...
+wheelhouse.
+
+1. Ana goes to Charlie to ask them to set up a cluster. Ana tells Charlie that
+   her clients will expect her APIs to be available using URLs rooted at
+   `https://ana.application.com/`.
+
+2. Charlie goes to Ian and requests a cluster.
+
+3. Ian provisions a cluster running a gateway controller with a [GatewayClass]
+   resource named `basic-gateway-class`. The gateway controller manages XXX.
```
youngnick, line 44: Comment to remind us all that this one needs to be finished up.

site-src/concepts/use-cases.md
```diff
@@ -0,0 +1,177 @@
...
+   distinct HTTPRoutes solely for the mesh, or they could apply single
+   HTTPRoutes that bind to both the Gateway and a Service.
+
+As always, the ultimate point of separating concerns in this way is that it
+permits Charlie to enforce centralized policies [such as
+TLS](/guides/tls#downstream-tls) at the Gateway, while allowing Ana and
+Allison to retain independent control of [routing
+logic](/guides/http-routing), [traffic splitting
+rollout](/guides/traffic-splitting), etc., both for [north/south] and for
+[east/west] routing.
```
youngnick, line 170: I think this section is great, but maybe it needs a larger callout about the details being a bit in flux? I don't want a repeat of Policy Attachment, where we make things seem like less experimental than they are by accident.

youngnick (commented):
This is looking great, nice work @kflynn.

The big question I have though is: do we have a backup plan for if we have to pivot away from Service binding? This PR effectively cements Service binding as the future, and if we change our mind, it will get fiddly.

I guess I'm really asking: at this point, is there any possibility we will use something other than Service binding as the main east/west approach?

If not, awesome, we're good.

If there is even a slim chance, we may need to put some more callouts and caveats that the service binding approach is still experimental and could change.

Like I said in one of my comments, I really want to avoid another Policy Attachment error - where the docs accidentally make something look more stable than it actually is.

site-src/concepts/use-cases.md
```diff
@@ -0,0 +1,177 @@
...
+advantage of the service mesh, with custom routing logic, without any
+bottlenecks in requests to Charlie or Ian.
+
+[east/west]:/concepts/glossary#east-west-traffic
+[GAMMA]:/contributing/gamma/
+[service mesh]:/concepts/glossary#service-mesh
+
+## Gateway and mesh use case
+
+This is effectively a combination of the `multiple applications behind a
```
robscott, line 140:
```suggestion
This is effectively a combination of the [multiple applications behind a
```

Was a broken link in deploy preview here: https://deploy-preview-2238--kubernetes-sigs-gateway-api.netlify.app/concepts/use-cases/

site-src/concepts/use-cases.md
robscott: Yeah really agree that we should be especially careful to add "experimental" notices/banners above all mesh-related sections (something like https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/api-types/grpcroute.md?plain=1#L3).

site-src/implementations.md
```diff
@@ -4,7 +4,7 @@ This document tracks downstream implementations and integrations of Gateway API
 
 Implementors and integrators of Gateway API are encouraged to update this document with status information about their implementations, the versions they cover, and documentation to help users get started.
 
-## Implementation Status
+## Gateway Controller Implementation Status <a name="gateways"></a>
```
robscott, line 7: Probably need to rebase to ensure Azure's implementation added today doesn't get lost in this change.

site-src/index.md
```diff
@@ -1,111 +1,183 @@
...
+how the Gateway API could also be used for inter-service or [_east/west_
+traffic][east/west traffic] within the same cluster.
+
+## Gateway API for Ingress <a name="for-ingress"></a>
+
+When using the Gateway API to manage ingress traffic, the [Gateway] resource
+defines a point of access at which traffic from outside clients can be routed
+into the cluster ([north/south traffic]). (If you're familiar with the older
+Ingress resource, you can think of a Gateway as analogous to a
+more-expressive, updated Ingress.)
```
robscott, line 33:
```suggestion
[Ingress API](https://kubernetes.io/docs/concepts/services-networking/ingress/), you can think of Gateway API as analogous to a more-expressive next-generation version of that API.
```

site-src/index.md
```diff
@@ -1,111 +1,183 @@
...
+[_north/south_][north/south traffic] case. Over time, interest from service
+mesh users prompted the creation of the [GAMMA initiative][gamma] to define
+how the Gateway API could also be used for inter-service or [_east/west_
+traffic][east/west traffic] within the same cluster.
+
+## Gateway API for Ingress <a name="for-ingress"></a>
+
+When using the Gateway API to manage ingress traffic, the [Gateway] resource
+defines a point of access at which traffic from outside clients can be routed
+into the cluster ([north/south traffic]). (If you're familiar with the older
```
robscott, line 31: Tiny nit: Always need to be very careful to avoid limiting the scope of the API here. I think Gateway API can and should be used to route traffic within a cluster (not necessarily just from outside to inside) and also that it should be used to route traffic to endpoints outside a cluster (ie multi-cluster Services or hybrid deployments that are only half on k8s). I don't actually know how to rephrase this in a way that makes sense but also doesn't exclude those use cases though, so for now just leaving a comment so we don't lose track of it.

site-src/concepts/glossary.md
```diff
@@ -0,0 +1,66 @@
...
+
+A _gateway controller_ is software that runs on a cluster to manage the
+infrastructure associated with routing traffic from outside the cluster to
+inside the cluster (north/south traffic), analogous to the earlier _ingress
```
robscott, line 12: Same comment as above, need to be careful to avoid limiting the scope of Gateways here.

site-src/concepts/api-overview.md
```diff
@@ -279,29 +278,97 @@ relationships between the different resources:
...
+[north/south]:/concepts/glossary#north-south-traffic
 
-## TLS Configuration
+### TLS Configuration
 
 TLS is configured on Gateway listeners, and may be referred to across namespaces.
 
 Please refer to the [TLS details](/guides/tls) guide for a deep dive on TLS.
 
+## Attaching Routes to Services
```
robscott, line 306: I think it would be very helpful to have a page about how GAMMA works (not just the contributing one). Maybe the content below here could be pulled out to separate GAMMA-focused page and linked from here? For example, it would be remarkably helpful if the GAMMA section on the main page could link to this new page, and the new GAMMA page could then link to the GAMMA contributing page.

robscott (commented):
Thanks for all the work on this @kflynn! This is really great!

site-src/concepts/api-overview.md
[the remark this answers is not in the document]
kflynn: This should be dealt with by d4047983ea70f1efd9ec851d928112a81e35b5ac, please resolve if you agree. 🙂

site-src/concepts/api-overview.md
```diff
@@ -279,29 +278,97 @@ relationships between the different resources:
...
+at least one of the Routes will be rejected.
+
+### Request flow
+
+A typical [east/west] API request flow is:
+
+1. A client workload makes a request to <http://foo.ns.service.cluster.local>.
+2. The data plane locates the Service matching `foo` in Namespace `ns`.
+3. The data plane locates Routes associated with the `foo` Service.
+4. If the request does not match any Route, it is rejected.
```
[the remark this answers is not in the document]
kflynn: This should be dealt with by d4047983ea70f1efd9ec851d928112a81e35b5ac, please resolve if you agree! 🙂

site-src/concepts/use-cases.md
kflynn: Oh this is just depressing, thanks for the catch! I went off and wrote the glossary definition, then forgot to update this. :man_facepalming:

site-src/concepts/glossary.md
```diff
@@ -0,0 +1,66 @@
...
+of the endpoints of the Service backend, bypassing routing decisions which
+might be made by the underlying network infrastructure. This is commonly
+necessary for advanced routing cases like sticky sessions, where the gateway
+will need to guarantee that every request for a specific session goes to the
+same endpoint.
+
+### Workload
+
+An instance of computation that provides a function within a cluster. A
+workload usually comprises a Service, some Pods, and often a Deployment.
```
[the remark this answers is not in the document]
kflynn: Doublechecking how we've been using it, you're right, the Service shouldn't be included. And, well, it'll be "standard" when we land this, right? 😂

site-src/concepts/glossary.md
kflynn: 3ec30329c95ce929fe7702b0b25de3d9deb98a72 updates the definition of a `gateway controller` to not require that it run in the cluster, let me know what you think there.

site-src/index.md
```diff
@@ -1,111 +1,183 @@
...
+from any traffic directed to that Service while preserving the role-oriented
+nature of the Gateway API.
+
+This use case is still rather new, and should be expected to evolve fairly
+quickly. One particular area that has rapidly become critical for GAMMA is the
+definition of the different [facets of the Service resource][service-facets].
+
+[gamma]:/contributing/gamma/
+[service-mesh]:/concepts/glossary#service-mesh
+[service-facets]:/concepts/service-facets.md
```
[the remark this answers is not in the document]
kflynn: Thanks for the catches on these links! I've fixed a bunch of things in 21eeb1d52945aa3fd7e4e4ce100c579303beb9e5.

site-src/index.md
```diff
@@ -1,111 +1,183 @@
...
 various backend targets (like Kubernetes Services, storage buckets, or
-functions). 
-
+functions).
+- **Service mesh support** with the GAMMA initiative - The Gateway API
+supports associating routing resources with Service resources, to configure
+service meshes as well as ingress controllers.
 
-## Why does a role-oriented API matter?
+## Why does a role-oriented API matter? <a name="role-oriented"></a>
```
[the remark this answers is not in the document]
kflynn: This particular one wasn't referenced anywhere, so I've just removed the anchor (I added it before deciding to split out the "Roles and Personas" page).

site-src/concepts/roles-and-personas.md
```diff
@@ -0,0 +1,80 @@
...
+personas.
+
+Note that, depending on the environment, a single human may end up taking on
+multiple roles, as discussed below.
+
+## Roles and Personas
+
+The Gateway API defines three roles and personas:
+
+* <a name="ian">**Ian**</a> (he/him) is an _infrastructure provider_,
```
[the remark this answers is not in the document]
kflynn: Nice catch – there's no real reason to have these contain the name, so I changed them to be the way I do other named anchors.

site-src/implementations.md
kflynn: Yup, done!

site-src/index.md
kflynn: Thanks!

site-src/concepts/use-cases.md
kflynn: Let me know if 41326713f314fed0abba6e759def365aceaa5f69 hits the mark here.

site-src/concepts/api-overview.md
kflynn: Great suggestion, I'll figure that out tomorrow.

site-src/concepts/glossary.md
```diff
@@ -0,0 +1,66 @@
+# Gateway API Glossary
+
+### Consumer Route
```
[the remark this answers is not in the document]
howardjohn: I am fine with that too but it's not alphabetical either (I think - you have me questioning 🙂 )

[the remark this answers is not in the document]
kflynn: 🤦‍♂️ Apparently I need to return to grade school or something. 😂 Fixed now. I think? 🙂

site-src/index.md
kflynn: Does 5783333f78307a4a037b61c111788454659a692b look good for this?

site-src/concepts/glossary.md
kflynn: Does 5783333f78307a4a037b61c111788454659a692b look good for this?

site-src/concepts/api-overview.md
kflynn: I really like the idea of bringing some less Western names in here. 🙂 At the same time, adopting names from languages and cultures that aren't our own has, historically, been fraught with peril for us Westerners (doubly so when Romanization rears its ugly head), so I'd like to be careful here.

As a given name, Chan comes from Khmer or Sanskrit. I'm just familiar enough with those two to be very concerned indeed that we will not treat that name correctly. 😐 Suppose we swap Charlie for Chihiro instead? That's a Japanese unisex name with a definitive Romanization, and it's likely to be pronounced (at least mostly) correctly by Anglophones when they see it. (And I was able to doublecheck all that with a friend from Japan who teaches Japanese for a living. 🙂)

With Chihiro, we'll have names from Spanish, Irish Gaelic, and Japanese. Hopefully that'll be a reasonable degree of diversity that won't get us into trouble for cultural misappropriation. 🤞🙂

site-src/concepts/api-overview.md
kflynn: Take a look at 2eb2361c074930ead4f435bdc505a026a5bcdf58 for this.

kflynn:
I think feedback other than renaming Charlie to Chihiro has been addressed in the current commits -- commenters (@youngnick, @robscott, @howardjohn) please check to see that things look good to you! Since I've pointed to specific commits in comments, I'll rebase _after_ folks take a look. 🙂

site-src/concepts/api-overview.md
candita: It's fine with me.  I just hadn't realized that Charlie was a unisex name, and first suggested Chris.  Later suggested Chan because it wasn't directly related to Christianity and had a more global meaning.

kflynn: Great! I'll go ahead and swap this one; @youngnick, if you can tackle #2128 that'd rock.

Thanks for the suggestion, @candita! 🙂

site-src/concepts/gamma.md
```diff
@@ -0,0 +1,234 @@
+# The GAMMA initiative
```
howardjohn, line 1: I really really hate the word "GAMMA". Maybe just a personal pet peave, but I would love to see this fade away.

IME its fine as "a name of a group of people". But the usage here is not great (not L1, I mean more like "How GAMMA works")

As a parallel, I feel like this is saying "How sig-network works" instead of "How `Service` works".

I would prefer something along the lines of "Gateway API for mesh", "Gateway for mesh", "Gateway for service mesh", "Gateway Service binding", ...

site-src/concepts/use-cases.md
kflynn: @youngnick, I'm going to go ahead and resolve this, yell if you want further edits!

site-src/concepts/gamma.md
kflynn: Yeah, something to that. I think I've cleaned this up now, take another look?

kflynn: (Resolving optimistically, comment further if you want other changes!)

kflynn:
I've marked this ready for review since I think we're pretty close at this point.

youngnick:
This LGTM, and is an excellent addition. Thanks @kflynn!

/approved
/hold
for further review

keithmattix (approved):

keithmattix:
Love this! Thanks flynn
/hold cancel

keithmattix:
I think @youngnick meant /approve

PR #2245, opened 2023-07-27
v0.8.0 API Review

robscott:
**What type of PR is this?**
/kind feature

This compares the `/apis` directory from the main branch (to be released soon as v0.8.0) with the `/apis` directory on the release-0.7 branch. This includes:

* API Type Definitions
* Webhook Validations

The intent of this PR is only to provide a mechanism for SIG Network review for the v0.8.0 API release. **_This PR will not merge._**

Note: As of the creation of this PR, there is still some ongoing work to translate webhook validations to CEL: https://github.com/kubernetes-sigs/gateway-api/pull/2226. This work will be included as soon as it is available. The CEL validation is not so much an API change as a different and more reliable way to represent the validation that already exists in webhook form.

```release-note
NONE
```

/assign @khenidak @mikezappa87 @thockin 
/hold (forever)


apis/v1beta1/gateway_types.go
```diff
@@ -450,6 +467,8 @@ type RouteGroupKind struct {
 }
 
 // GatewayAddress describes an address that can be bound to a Gateway.
+//
+// +kubebuilder:validation:XValidation:message="Hostname value must only contain valid characters (matching ^(\\*\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$)",rule="self.type == 'Hostname' ? self.value.matches('^(\\\\*\\\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'): true"
```
thockin, line 471: FYI: @jpbetz @alexzielenski Two thoughts:

1) This feels like a good candidate for a validation library function, e.g. "IsWildcardDNSDomain" or something.

2) For places that legit need to do a regex match, can we find a way not to repeat the regex twice (and with different levels of escape) ?

apis/v1beta1/httproute_types.go
```diff
@@ -59,7 +59,8 @@ type HTTPRouteSpec struct {
 	// Hostnames defines a set of hostname that should match against the HTTP Host
```
thockin, line 59: "set of hostname" -> "hostnames"

apis/v1beta1/httproute_types.go
```diff
@@ -303,6 +317,18 @@ const (
 )
 
 // HTTPPathMatch describes how to select a HTTP route by matching the HTTP request path.
+//
+// +kubebuilder:validation:XValidation:message="value must be an absolute path and start with '/' when type one of ['Exact', 'PathPrefix']",rule="(self.type == 'Exact' || self.type == 'PathPrefix') ? self.value.startsWith('/') : true"
+// +kubebuilder:validation:XValidation:message="must not contain '//' when type one of ['Exact', 'PathPrefix']",rule="(self.type == 'Exact' || self.type == 'PathPrefix') ? !self.value.contains('//') : true"
+// +kubebuilder:validation:XValidation:message="must not contain '/./' when type one of ['Exact', 'PathPrefix']",rule="(self.type == 'Exact' || self.type == 'PathPrefix') ? !self.value.contains('/./') : true"
+// +kubebuilder:validation:XValidation:message="must not contain '/../' when type one of ['Exact', 'PathPrefix']",rule="(self.type == 'Exact' || self.type == 'PathPrefix') ? !self.value.contains('/../') : true"
+// +kubebuilder:validation:XValidation:message="must not contain '%2f' when type one of ['Exact', 'PathPrefix']",rule="(self.type == 'Exact' || self.type == 'PathPrefix') ? !self.value.contains('%2f') : true"
```
thockin, line 325: no way to do caseless compare?  I guess you want to get the match value out, sort of like a regex...

apis/v1beta1/shared_types.go
```diff
@@ -104,6 +122,10 @@ type ParentReference struct {
 	// and SectionName are specified, the name and port of the selected listener
 	// must match both specified values.
 	//
+	// When the parent resource is a Service, this targets a specific port in the
+	// Service spec. When both Port (experimental) and SectionName are specified,
```
thockin, line 126: What does SectionName mean for a Service?

thockin (commented):
I Read comments in API.  I read some of the CEL but I can't verify that - I'm assuming that the tests which pass against manual validation still pass and the tests which fail still fail?

A few comments here warrant responses but otherwise OK

apis/v1beta1/shared_types.go
robscott: SectionName means Port name, will clarify.

apis/v1beta1/gateway_types.go
```diff
@@ -70,7 +70,13 @@ type GatewaySpec struct {
 	// At least one Listener MUST be specified.
 	//
 	// Each listener in a Gateway must have a unique combination of Hostname,
-	// Port, and Protocol.
+	// Port, and Protocol. Below combinations are considered Core and MUST be
+	// supported:
+	//
+	// 1. Port: 80, Protocol: HTTP
```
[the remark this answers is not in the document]
robscott: Yeah I think we're missing some nuance here. This was driven by our conformance tests and trying to define which ports we want to include in test coverage + what conformance level they would be considered. So this really means that 8080 on HTTP is only "Extended" support. I think we need to clarify that this is specific to our "HTTP" conformance profile, will work on improving the wording here.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -223,8 +223,15 @@ type GRPCRouteRule struct {
 	// - Implementation-specific custom filters have no API guarantees across
 	//   implementations.
 	//
-	// Specifying a core filter multiple times has unspecified or
-	// implementation-specific conformance.
+	// Specifying the same filter multiple times is not supported unless explicitly
+	// indicated in the filter.
+	//
+	// If an implementation can not support a combinations of filters, they must clearly
```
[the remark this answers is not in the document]
robscott: Covered in #2261

apis/v1beta1/gateway_types.go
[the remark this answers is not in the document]
robscott: Covered in #2261

apis/v1beta1/gateway_types.go
```diff
@@ -111,6 +117,11 @@ type GatewaySpec struct {
 	// +listMapKey=name
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:XValidation:message="tls must be set for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
```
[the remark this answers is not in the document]
robscott: Covered in #2261

apis/v1beta1/httproute_types.go
robscott: Covered in #2261

apis/v1beta1/httproute_types.go
robscott: Yeah I think these are all split up the way that they are just to provide more meaningful error messages

apis/v1beta1/shared_types.go
robscott: Covered in #2261

apis/v1alpha2/grpcroute_types.go
```diff
@@ -579,5 +590,7 @@ type GRPCBackendRef struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:XValidation:message="RequestHeaderModifier filter cannot be repeated",rule="self.exists(f, f.type == 'RequestHeaderModifier') ? self.exists_one(f, f.type == 'RequestHeaderModifier') : true"
```
jpbetz, line 593: `self.filter(f, f.type == 'RequestHeaderModifier').size() <= 1` ?

apis/v1beta1/gateway_types.go
```diff
@@ -111,6 +117,11 @@ type GatewaySpec struct {
...
+	// +kubebuilder:validation:XValidation:message="tls must be empty for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
+	// +kubebuilder:validation:XValidation:message="hostname must be empty for protocols ['TCP', 'UDP']",rule="self.all(l, l.protocol in ['TCP', 'UDP']  ? (!has(l.hostname) || l.hostname == '') : true)"
+	// +kubebuilder:validation:XValidation:message="Listener name must be unique within the Gateway",rule="self.all(l1, self.exists_one(l2, l1.name == l2.name))"
+	// +kubebuilder:validation:XValidation:message="Combination of port, protocol and hostname must be unique for each listener",rule="self.all(l1, self.exists_one(l2, l1.port == l2.port && l1.protocol == l2.protocol && (has(l1.hostname) && has(l2.hostname) ? l1.hostname == l2.hostname : true)))"
```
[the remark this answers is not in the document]
jpbetz: > maybe this is a candidate for some generalization? e.g. `self.unique(l, l.name)` or some simpler expression that means "evaluate this expression (`l.name`) on each element of self, accumulate and ensure unique".

We have a gap here for sure.  I think I favor a `.unique()` function that can be called on collection types. Callers can then do things like `self.map(l, l.name).unique()`.

> 
> Also, the repeated pattern of `has` checking. Does CEL have something like bash default values? `${optional:-}` means "the value of $optional, or else an empty value, but never undefined".

[CEL optionals](https://github.com/google/cel-spec/wiki/proposal-246) can do:

```
self.?someOptionalField.orValue("<desired default>")
```

Unfortunately, we just added optional support in 1.28 so it won't be usable here for a few releases...

youngnick:
Closed early, let's leave this open for now.

apis/v1beta1/gateway_types.go
jpbetz: > 1. This feels like a good candidate for a validation library function, e.g. "IsWildcardDNSDomain" or something.

I agree. As a general rule, I'd like CEL to have full parity with the formats we support, so when we add a format for wildcardDNSDomain, I'd like it to also be possible to validate that in CEL.

> 2. For places that legit need to do a regex match, can we find a way not to repeat the regex twice (and with different levels of escape) ?

We added variable support to ValidatingAdmissionPolicy but not CRD validation rules (with CEL). It would solve for this at the cost of added complexity.  I wonder if having named formats would be sufficient here?  We could document them well and then just reference them by name in the expression and message?

apis/v1beta1/shared_types.go
```diff
@@ -150,10 +187,22 @@ type CommonRouteSpec struct {
...
+	//
+	// ParentRefs from a Route to a Service in a different namespace are
+	// "consumer" routes, and these routing rules are only applied to outbound
+	// connections originating from the same namespace as the Route, for which
+	// the intended destination of the connections are a Service targeted as a
+	// ParentRef of the Route.
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=32
+	// <gateway:standard:validation:XValidation:message="sectionName must be unique when parentRefs includes 2 or more references to the same parent",rule="self.all(p1, self.exists_one(p2, p1.group == p2.group && p1.kind == p2.kind && ((!has(p1.__namespace__) && !has(p2.__namespace__)) || (!has(p1.__namespace__) && p2.__namespace__ == '') || (p1.__namespace__ == '' && !has(p2.__namespace__)) || (p1.__namespace__ == p2.__namespace__)) && p1.name == p2.name && ((!has(p1.sectionName) && !has(p2.sectionName)) || (!has(p1.sectionName) && p2.sectionName == '') || (p1.sectionName == '' && !has(p2.sectionName)) || (p1.sectionName == p2.sectionName))))">
```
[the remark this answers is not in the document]
jpbetz: CEL optionals will be a huge help here. I think this simplifies to [this](https://playcel.undistro.io/?content=H4sIAAAAAAAAA7WQwQrCMAyGXyX0YDeYg%2B04EN%2FAo14GpWxxFLuurB1OxHe33cqcoN7sKfn%2BJP2TO6m55aQgBuW5KBW4t4Wm7wZdAMWRt1pilnZ9Q2cR4CJU7bSTqBu0C1W8RUfb2%2FWdM%2BYVo3mFjLkCi%2BYlGqys6NRh7g1ZTv9g4%2BNP2SKThOCoezTG4XCNlEsZ6SyBKcFRGGtYpzDSeQK%2BUWfp5BB2O9B5iDcbz725gKdwpt5coFPoaJizfzuT2%2FPI5YARpXGo%2F1GwmrLa8tOML3Kp4pg8no%2B72AgKAgAA):

```
self.all(p1, self.exists_one(p2, 
  p1.group == p2.group && 
  p1.kind == p2.kind && 
  p1.name == p2.name && 
  p1.?__namespace__.orValue('') == p2.?__namespace__.orValue('') && 
  p1.?sectionName.orValue('') == p2.?sectionName.orValue('')
))
```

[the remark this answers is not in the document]
gauravkghildiyal: Great catch! Addresses this in https://github.com/kubernetes-sigs/gateway-api/pull/2320.

By the way, something like https://playcel.undistro.io/ is exactly what I had been looking for but failed to find. Thanks a lot for sharing

apis/v1beta1/httproute_types.go
gauravkghildiyal: Thanks Joe! Would you agree that an expression like:

```
!(self.type in ['Exact', 'PathPrefix']) || !self.value.contains('%2f')

v/s

(self.type in ['Exact','PathPrefix']) ? !self.value.contains('%2f') : true
```

although definitely (a bit) shorter, sacrifices the comprehensibility of the conditional here? (I think just using the `in` seems good enough)

khenidak (approved):

robscott:
Thanks everyone for the reviews! We've released v0.8.0 and are going to close this out.

/close

PR #2246, opened 2023-07-27
Completing CEL validation for Gateway

robscott:
**What type of PR is this?**
/kind cleanup

This is a follow up to @gauravkghildiyal's PR that added CEL validation (#2226). It contains a few things:

* IPAddress validation (using ~~https://regex101.com/r/Kzijkp/1~~ our custom CRD generation + `oneOf`)
* Validation for GatewayStatusAddress
* A few typo fixes

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -149,6 +149,7 @@ type GatewaySpec struct {
 	// Support: Extended
 	//
 	// +optional
+	// <gateway:validateIPAddress>
 	// +kubebuilder:validation:MaxItems=16
 	// +kubebuilder:validation:XValidation:message="IPAddress values must be unique",rule="self.all(a1, a1.type == 'IPAddress' ? self.exists_one(a2, a2.type == a1.type && a2.value == a1.value) : true )"
 	// +kubebuilder:validation:XValidation:message="Hostname values must be unique",rule="self.all(a1, a1.type == 'Hostname' ? self.exists_one(a2, a2.type == a1.type && a2.value == a1.value) : true )"
@@ -486,6 +487,8 @@ type GatewayAddress struct {
 }
 
 // GatewayStatusAddress describes an address that is bound to a Gateway.
+//
+// +kubebuilder:validation:XValidation:message="Hostname value must only contain valid characters (matching ^(\\*\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$)",rule="self.type == 'Hostname' ? self.value.matches('^(\\\\*\\\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'): true"
 type GatewayStatusAddress struct {
 	// Type of the address.
 	//
@@ -511,6 +514,7 @@ type GatewayStatus struct {
 	// assigns an address from a reserved pool.
 	//
 	// +optional
+	// <gateway:validateIPAddress>
 	// +kubebuilder:validation:MaxItems=16
 	Addresses []GatewayStatusAddress `json:"addresses,omitempty"`
 
--- a/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
@@ -73,10 +73,24 @@ spec:
                   manner, assigning an appropriate set of Addresses. \n The implementation
                   MUST bind all Listeners to every GatewayAddress that it assigns
                   to the Gateway and add a corresponding entry in GatewayStatus.Addresses.
-                  \n Support: Extended"
+                  \n Support: Extended \n "
                 items:
                   description: GatewayAddress describes an address that can be bound
                     to a Gateway.
+                  oneOf:
+                  - properties:
+                      type:
+                        enum:
+                        - IPAddress
+                      value:
+                        anyOf:
+                        - format: ipv4
+                        - format: ipv6
+                  - properties:
+                      type:
+                        not:
+                          enum:
+                          - IPAddress
                   properties:
                     type:
                       default: IPAddress
@@ -500,13 +514,27 @@ spec:
             description: Status defines the current state of Gateway.
             properties:
               addresses:
-                description: Addresses lists the IP addresses that have actually been
-                  bound to the Gateway. These addresses may differ from the addresses
+                description: "Addresses lists the IP addresses that have actually
+                  been bound to the Gateway. These addresses may differ from the addresses
                   in the Spec, e.g. if the Gateway automatically assigns an address
-                  from a reserved pool.
+                  from a reserved pool. \n "
                 items:
                   description: GatewayStatusAddress describes an address that is bound
                     to a Gateway.
+                  oneOf:
+                  - properties:
+                      type:
+                        enum:
+                        - IPAddress
+                      value:
+                        anyOf:
+                        - format: ipv4
+                        - format: ipv6
+                  - properties:
+                      type:
+                        not:
+                          enum:
+                          - IPAddress
                   properties:
                     type:
                       default: IPAddress
@@ -525,6 +553,11 @@ spec:
                   required:
                   - value
                   type: object
+                  x-kubernetes-validations:
+                  - message: Hostname value must only contain valid characters (matching
+                      ^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$)
+                    rule: 'self.type == ''Hostname'' ? self.value.matches(''^(\\*\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$''):
+                      true'
                 maxItems: 16
                 type: array
               conditions:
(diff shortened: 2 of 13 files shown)


apis/v1beta1/gateway_types.go
```diff
@@ -537,6 +537,7 @@ type RouteGroupKind struct {
...
...a-f]{1,4}|((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)(\\\\.(25[0-5]|2[0-4]\\\\d|1\\\\d\\\\d|[1-9]?\\\\d)){3}))|:)))(%.+)?\\\\s*$))'): true"
```
robscott, line 540: With all the escaping this gets to be pretty hard to parse - here's the original: https://regex101.com/r/Kzijkp/1

hack/cel-validation/gateway_test.go
```diff
@@ -446,14 +482,21 @@ func TestValidateGateway(t *testing.T) {
...
-			tc.mutate(gwc)
-			err := k8sClient.Create(ctx, gwc)
+			if tc.mutateStatus != nil {
+				tc.mutateStatus(gw)
+				err = k8sClient.Status().Update(ctx, gw)
+			}
 
 			if (len(tc.wantErrors) != 0) != (err != nil) {
-				t.Fatalf("Unexpected error while creating Gateway; got err=\n%v\n;want error=%v", err, tc.wantErrors != nil)
+				t.Fatalf("Expected error while creating Gateway but didn't get one; got err=\n%v\n;want error=%v", err, tc.wantErrors != nil)
```
gauravkghildiyal, line 499: Sorry it seems like the prior statement was also not correct. Would:
`Unexpected response while creating Gateway" be more correct?

Because there are tests that don't necessarily "expect an error"

gauravkghildiyal (commented):
Looks good to me! Thanks a lot for the corrections and additions!

apis/v1beta1/gateway_types.go
shaneutt, line 540: </img>

shaneutt: The escaping is making me wonder if we should hang out with the kubebuilder group a bit and see about options to get these rules from other files and locations.

howardjohn: @robscott thanks for the link, without the escaping it is super easy to understand!

howardjohn: Less sarcastic: can we use https://github.com/kubernetes/kubernetes/pull/85381/files? Seems we already use it in some other places (`date-time` - interestingly, the link I sent says it should be `datetime`.

Actually that may be tricky since we need this conditional... maybe there is some way to get these into the CEL... I can't help but think there must be a better way to do this (or if not, there should be)

robscott: Checked in with @jpbetz and sadly we're stuck with a regex for now. The good news is that this will get better with [KEP 3937](https://github.com/kubernetes/enhancements/pull/3938/files#diff-31dd3b3a600e79340c20ad455fd3e7ca2183f03a62defd473d2ea4ed85b90d1bR391-R401). Unfortunately in the best case this will be enabled by default in 1.31+, and with our 5 minor version support window, we'd need to wait for 1.36, or >2 years from now for this to be available.

The closest thing we have available now is `isURL("https://" + ip + "/")` which uses https://github.com/kubernetes/kubernetes/blob/97c7dbcd227ec253bb277a91575706fe30f519dc/staging/src/k8s.io/apiserver/pkg/cel/library/urls.go#L234, but of course that would still be too permissive for what we need here.

jpbetz: One possibility I had not considered earlier-

Could you use the OpenAPIv3 `format: ipv4` and `format: ipv6` to validate the field?  If the IP can either be v4 or v6, you might be able to use OpenAPIv3 `anyOf` to check if it's either.

gauravkghildiyal: Curious to know if we can have conditional logic on when we want to use those `format: ipv4` and `format: ipv6`.?Because the field itself could have string value as well.

jpbetz: Ah, I had missed the conditional. Thanks for pointing that out @gauravkghildiyal

jpbetz: One possible OpenAPIv3 way of handling the conditional:

```
type: object
   properties:
     type:
       type: string
     value:
       type: string
   oneOf:
     - properties:
         type:
           enum: ["IPAddress"]
         value:
           oneOf:
             - format: ipv4
             - format: ipv6
     - properties:
         type:
           not:
             enum: ["IPAddress"]
```

cc @Jefftree for any suggestions on making the OpenAPI cleaner/more idiomatic.

hack/cel-validation/gateway_test.go
robscott: Updated, thanks!

apis/v1beta1/gateway_types.go
robscott: Thanks @jpbetz! Added a _very_ ugly hack to our CRD generator to make this work for now, definitely seems preferable to the regex.

config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
```diff
@@ -500,13 +514,27 @@ spec:
             description: Status defines the current state of Gateway.
             properties:
               addresses:
-                description: Addresses lists the IP addresses that have actually been
-                  bound to the Gateway. These addresses may differ from the addresses
+                description: "Addresses lists the IP addresses that have actually
+                  been bound to the Gateway. These addresses may differ from the addresses
                   in the Spec, e.g. if the Gateway automatically assigns an address
-                  from a reserved pool.
+                  from a reserved pool. \n <gateway:validateIPAddress>"
```
howardjohn, line 520: We may want to strip "<gateway:validateIPAddress>"?

apis/v1beta1/gateway_types.go
Jefftree: Oops sorry I didn't see I was tagged earlier. Yeah i think Joe's suggestion would be fine, though having 

```
         type:
           not:
             enum: ["IPAddress"]
```

makes it look like any other value can be accepted if it doesn't fit the enum? If the fallback is string, why not keep that in the oneOf?

```
  oneOf:
     - properties:
         type:
           enum: ["IPAddress"]
         value:
           anyOf:
             - format: ipv4
             - format: ipv6
     - properties:
          type:
            type: string
          value:
            type: string
```

robscott: @Jefftree @jpbetz I think I'm misunderstanding the suggestion. It seems like that would be interpreted as 

```
One Of:
A) type == IPAddress && (format == ipv4 || format == ipv6)
OR
B) type == any string && value == any string
```

What we really want is:

```
A) type == IPAddress && (format == ipv4 || format == ipv6)
OR
B) type == anything other than IPAddress
```

I think that's closer to what we actually have.

We already have the following `x-kubernetes-validations` value at the same level to cover the hostname type:

```yaml
                  x-kubernetes-validations:
                  - message: Hostname value must only contain valid characters (matching
                      ^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$)
                    rule: 'self.type == ''Hostname'' ? self.value.matches(''^(\\*\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$''):
                      true'
```

Admittedly it would probably make sense to combine that all into the `oneOf` for a CRD that was a bit easier to understand. We're avoiding that because the `Hostname` validation can be expressed with a _relatively_ simple regex that fits in a `+kubebuilder` annotation and doesn't require customizing the CRD generation like we're doing for IP Addresses here.

config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
robscott: Good catch, removed this and a few other `<gateway:experimental>` tags that didn't need to be in CRDs.

apis/v1beta1/gateway_types.go
Jefftree: Ah right that will allow even invalid IPAddress types to pass validation via B. In that case would Joe's first suggestion should suffice then? It's already quite concise for what we want to express.

robscott: Yep, Joe's first suggestion seems to work well here, thanks for the help!

youngnick:
I don't think the generator hack is that bad. 😁 

/lgtm

PR #2247, opened 2023-07-28
Check ResolvedRefs condition in Gateway conformance tests

### LINKED ISSUE #1315: conformance: always check `ResolvedRefs` condition in `Gateway` conformance tests

In the `Gateway` conformance tests, the `ListenerConditionResolvedRefs` is checked only in those tests expecting it to have status`false` (e.g., `GatewaySecretMissingReferenceGrant`, `GatewaySecretInvalidReferenceGrant`, ...), but it is never checked in those tests who expect the routing to be successful. This leads to never checking the `ResolvedRefs` reason with `ConditionStatus == true`.

**Why this is needed**:

This is needed to ensure that the `ListenerConditionResolvedRefs` is always set, no matter if the `ConditionStatus` is `true`, or `false`.

levikobi:
**What type of PR is this?**

/kind test
/area conformance

Checks the `ListenerConditionResolvedRefs` value. This is needed to ensure that its value is always set, no matter if the `ConditionStatus` is true, or false.

Fixes #1315 

```release-note
None
```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -303,6 +303,11 @@ func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutCo
 						Status: metav1.ConditionTrue,
 						Reason: string(v1beta1.RouteReasonAccepted),
 					},
+					{
+						Type:   string(v1beta1.ListenerConditionResolvedRefs),
+						Status: metav1.ConditionTrue,
+						Reason: "", // any reason
+					},
 				},
 			})
 		}
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -267,9 +267,12 @@ func MeshNamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig conf
 	require.NoErrorf(t, waitErr, "error waiting for %s namespaces to be ready", strings.Join(namespaces, ", "))
 }
 
-// GatewayAndHTTPRoutesMustBeAccepted waits until the specified Gateway has an IP
-// address assigned to it and the Route has a ParentRef referring to the
-// Gateway. The test will fail if these conditions are not met before the
+// GatewayAndHTTPRoutesMustBeAccepted waits until:
+// 1. The specified Gateway has an IP address assigned to it.
+// 2. The route has a ParentRef referring to the Gateway.
+// 3. All the gateway's listeners have the ListenerConditionResolvedRefs set to true.
+//
+// The test will fail if these conditions are not met before the
 // timeouts.
 func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, controllerName string, gw GatewayRef, routeNNs ...types.NamespacedName) string {
 	t.Helper()
@@ -297,22 +300,23 @@ func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutCo
 					SectionName: listener,
 				},
 				ControllerName: v1beta1.GatewayController(controllerName),
-				Conditions: []metav1.Condition{
-					{
-						Type:   string(v1beta1.RouteConditionAccepted),
-						Status: metav1.ConditionTrue,
-						Reason: string(v1beta1.RouteReasonAccepted),
-					},
-					{
-						Type:   string(v1beta1.ListenerConditionResolvedRefs),
-						Status: metav1.ConditionTrue,
-						Reason: "", // any reason
-					},
-				},
+				Conditions: []metav1.Condition{{
+					Type:   string(v1beta1.RouteConditionAccepted),
+					Status: metav1.ConditionTrue,
+					Reason: string(v1beta1.RouteReasonAccepted),
+				}},
 			})
 		}
 		HTTPRouteMustHaveParents(t, c, timeoutConfig, routeNN, parents, namespaceRequired)
 	}
+
+	resolvedRefsCondition := metav1.Condition{
+		Type:   string(v1beta1.ListenerConditionResolvedRefs),
+		Status: metav1.ConditionTrue,
+		Reason: "", // any reason
+	}
+
+	GatewayListenersMustHaveCondition(t, c, timeoutConfig, gw.NamespacedName, resolvedRefsCondition)
 
 	return gwAddr
 }
@@ -350,6 +354,29 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, timeoutConfig con
 	})
 	require.NoErrorf(t, waitErr, "error waiting for Gateway to have at least one IP address in status")
 	return net.JoinHostPort(ipAddr, port), waitErr
+}
+
+// GatewayListenersMustHaveCondition checks if every listener of the specified gateway has a
+// certain condition.
+func GatewayListenersMustHaveCondition(t *testing.T, client client.Client, timeoutConfig config.TimeoutConfig, gwName types.NamespacedName, condition metav1.Condition) {
+	t.Helper()
+
+	waitErr := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, timeoutConfig.GatewayListenersMustHaveCondition, true, func(ctx context.Context) (bool, error) {
+		var gw v1beta1.Gateway
+		if err := client.Get(ctx, gwName, &gw); err != nil {
+			return false, fmt.Errorf("error fetching Gateway: %w", err)
+		}
+
+		for _, listener := range gw.Status.Listeners {
+			if !findConditionInList(t, listener.Conditions, condition.Type, string(condition.Status), condition.Reason) {
+				return false, nil
+			}
+		}
+
+		return true, nil
+	})
+
+	require.NoErrorf(t, waitErr, "error waiting for Gateway status to have a Condition matching expectations on all listeners")
 }
 
 // GatewayMustHaveZeroRoutes validates that the gateway has zero routes attached.  The status
--- a/conformance/tests/gateway-modify-listeners.go
+++ b/conformance/tests/gateway-modify-listeners.go
@@ -87,11 +87,18 @@ var GatewayModifyListeners = suite.ConformanceTest{
 						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
 						Kind:  v1beta1.Kind("HTTPRoute"),
 					}},
-					Conditions: []metav1.Condition{{
-						Type:   string(v1beta1.ListenerConditionAccepted),
-						Status: metav1.ConditionTrue,
-						Reason: "", // any reason
-					}},
+					Conditions: []metav1.Condition{
+						{
+							Type:   string(v1beta1.ListenerConditionAccepted),
+							Status: metav1.ConditionTrue,
+							Reason: "", // any reason
+						},
+						{
+							Type:   string(v1beta1.ListenerConditionResolvedRefs),
+							Status: metav1.ConditionTrue,
+							Reason: "", // any reason
+						},
+					},
 					AttachedRoutes: 1,
 				},
 				{
@@ -100,11 +107,18 @@ var GatewayModifyListeners = suite.ConformanceTest{
 						Group: (*v1beta1.Group)(&v1beta1.GroupVersion.Group),
 						Kind:  v1beta1.Kind("HTTPRoute"),
 					}},
-					Conditions: []metav1.Condition{{
-						Type:   string(v1beta1.ListenerConditionAccepted),
-						Status: metav1.ConditionTrue,
-						Reason: "", // any reason
-					}},
+					Conditions: []metav1.Condition{
+						{
+							Type:   string(v1beta1.ListenerConditionAccepted),
+							Status: metav1.ConditionTrue,
+							Reason: "", // any reason
+						},
+						{
+							Type:   string(v1beta1.ListenerConditionResolvedRefs),
+							Status: metav1.ConditionTrue,
+							Reason: "", // any reason
+						},
+					},
 					AttachedRoutes: 1,
 				},
 			}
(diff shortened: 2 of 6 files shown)


michaelbeaumont (approved):
/lgtm !

Xunzhuo (approved):
/lgtm thanks! 

defer to others, make sure nowhere is missed.

LiorLieberman (commented):
I might be missing something, but if are adding these conditions here and not changing https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/kubernetes/helpers.go#L708-L711

It must say that either our tests fail now or this change will cause them to fail.

What am I missing?

michaelbeaumont:
> I might be missing something, but if are adding these conditions here and not changing https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/kubernetes/helpers.go#L708-L711
> 
> It must say that either our tests fail now or this change will cause them to fail.
> 
> What am I missing?

[The check only asserts on conditions explicitly listed in the test](https://github.com/kubernetes-sigs/gateway-api/blob/3ae9fa78b2e64eb03c701a964bbcd199389f7776/conformance/utils/kubernetes/helpers.go#L725), so the existing or missing `ResolvedRefs` condition was ignored

LiorLieberman:
> [The check only asserts on conditions explicitly listed in the test](https://github.com/kubernetes-sigs/gateway-api/blob/3ae9fa78b2e64eb03c701a964bbcd199389f7776/conformance/utils/kubernetes/helpers.go#L725), so whether existing or missing, the ResolvedRefs condition was ignored

But it looks like it compares the number of `actual` conditions vs the `expected` conditions before.
So if we add another condition the length should be different

levikobi:
> > [The check only asserts on conditions explicitly listed in the test](https://github.com/kubernetes-sigs/gateway-api/blob/3ae9fa78b2e64eb03c701a964bbcd199389f7776/conformance/utils/kubernetes/helpers.go#L725), so whether existing or missing, the ResolvedRefs condition was ignored
> 
> But it looks like it compares the number of `actual` conditions vs the `expected` conditions before. So if we add another condition the length should be different

Have a look at https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/kubernetes/helpers.go#L721

Only if the number of expected conditions is larger than the actual the test will fail.
Up until now, the number of expected was lower, hence the tests passed.
Right now we are adding another condition, **_which should have already been in the actual conditions list_** (this is what the modification is verifying). So, len(actual) will equal len(expected) at most, and if not - the test should fail

LiorLieberman:
Right, thats what I was missing, long day ...😅

Thanks! 

/lgtm

mlavacca (commented):
Thanks for this PR @levikobi!

I think we should check that **all** the gateways used in the tests have actually set such a condition to true (Even the gateways used in the `HTTPRoute` tests). To do so, we could change [this helper](https://github.com/mlavacca/gateway-api/blob/1fe0c74b8ee2851ffe7117a9cbfe31026675bbd4/conformance/utils/kubernetes/helpers.go#L246) to check that the `resolvedRefs` is enforced as well.

/cc

levikobi:
> Thanks for this PR @levikobi!
> 
> I think we should check that **all** the gateways used in the tests have actually set such a condition to true (Even the gateways used in the `HTTPRoute` tests). To do so, we could change [this helper](https://github.com/mlavacca/gateway-api/blob/1fe0c74b8ee2851ffe7117a9cbfe31026675bbd4/conformance/utils/kubernetes/helpers.go#L246) to check that the `resolvedRefs` is enforced as well.

That's a great suggestion, thanks @mlavacca!

conformance/utils/kubernetes/helpers.go
```diff
@@ -275,6 +275,11 @@ func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutCo
 						Status: metav1.ConditionTrue,
 						Reason: string(v1beta1.RouteReasonAccepted),
 					},
+					{
```
mlavacca, line 278: When reading this helper more carefully, it looks like the Gateway conditions are not checked here (they are actually checked in no `HTTPRoute` test). The helper checks the `ParentConditions`, which by definition (from docs) are:

```
// Conditions describes the status of the route with respect to the Gateway.
// Note that the route's availability is also subject to the Gateway's own
// status conditions and listener status.
// ...
```

For this reason, I'd remove this `ResolvedRefs` condition from here (sorry for suggesting it :)) and seek other's opinion on checking/not checking gateway conditions in `HTTPRoute` tests. Is it something you deem we should take care of? Or just performing this check in the Gateway tests is enough?
I feel we should add such a check in all tests that create gateways, but it is not a really strong opinion.

@shaneutt @robscott @sunjayBhatia @michaelbeaumont @LiorLieberman @arkodg

robscott: Good catch @mlavacca! This led me to question what we actually _require_ here. At first I looked at the API Spec, and the only references I could find required it to be set when it was negative. Fortunately this is clarified in the [GEP that prompted this](https://gateway-api.sigs.k8s.io/geps/gep-1364/#should-conditions-always-be-added), and it appears that `ResolvedRefs` should indeed always be set. 

My next question is _where_ it should be set. For some reason, `ResolvedRefs` is a `Listener` condition and not a `Gateway` condition:

https://github.com/kubernetes-sigs/gateway-api/blob/f638d3fe75a69354b4a15b6c9e9e475c4e7d8062/apis/v1beta1/gateway_types.go#L833-L849

I _think_ that makes sense because 100% of the object references in a Gateway are defined within a Listener today. But it could get awfully noisy to have to specify a positive value for each listener in a Gateway here. I wonder if it would make more sense just to require that a top level positive condition is set on the Gateway, and if there are problems, they'll be surfaced on individual listeners as well?

In any case, to do a universal check for this condition, we'd likely want to add it somewhere like `GatewayAndHTTPRoutesMustBeAccepted` (along with some renaming) since that function appears to be pretty broadly used in our conformance tests.

Open to any/all ideas here, but particularly curious what @youngnick thinks since he's been so instrumental in defining how our status + conditions work.

/cc @youngnick

sunjayBhatia: seems like this carries over to Accepted and Programmed conditions as well, if I'm not mistaken we have inconsistent checks for those Listener conditions in tests as well

Also agreed `GatewayAndHTTPRoutesMustBeAccepted` could use a rename/rework to include some element of/call `GatewayStatusMustHaveListeners` (since as it is today it just checks on a Gateway for the latest generation on top-level conditions and for an address to be assigned)

sunjayBhatia: Thinking in terms of how a user finds out what is wrong with their Gateway today in Contour (without adding a having a top-level summary condition besides what exists), say in the case of an invalid TLS certificateRef:
- See Programmed=False with reason "ListenersNotValid"
- Drill down into Listeners to see what is wrong
- See a Listener is Programmed=False, and also ResolvedRefs=False with reason "InvalidCertificateRef"

Seems like there is enough of a flow/hierarchy to get the info you need currently

Right now we don't set ResolvedRefs on every Listener but we could, it would add some more info and clutter, but it feels like Accepted/Programmed are good enough summary conditions at the top-level and ListenerStatus level

Interested in others thoughts

some example status:
```
  status:
    addresses:
    - type: IPAddress
      value: 172.18.255.200
    conditions:
    - lastTransitionTime: "2023-08-10T19:41:11Z"
      message: Gateway is accepted
      observedGeneration: 3
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2023-08-10T19:41:11Z"
      message: Listeners are not valid
      observedGeneration: 3
      reason: ListenersNotValid
      status: "False"
      type: Programmed
    listeners:
    - attachedRoutes: 0
      conditions:
      - lastTransitionTime: "2023-08-10T19:41:11Z"
        message: 'Spec.VirtualHost.TLS.CertificateRefs "foo" referent is invalid:
          Secret not found'
        observedGeneration: 3
        reason: InvalidCertificateRef
        status: "False"
        type: ResolvedRefs
      - lastTransitionTime: "2023-08-10T19:41:11Z"
        message: Invalid listener, see other listener conditions for details
        observedGeneration: 3
        reason: Invalid
        status: "False"
        type: Programmed
      - lastTransitionTime: "2023-08-10T19:41:11Z"
        message: Listener accepted
        observedGeneration: 3
        reason: Accepted
        status: "True"
        type: Accepted
      name: site-invalid-https
      supportedKinds:
        kind: HTTPRoute
        kind: GRPCRoute
```

levikobi: So to my understanding, we need to check inside `GatewayAndHTTPRoutesMustBeAccepted` if every listener has the `ListenerConditionResolvedRefs` set to true.

I think the easiest way to do it is to add a new function (say we call it `GatewayListenersMustHaveCondition`) similar to `HTTPRouteMustHaveCondition`, which checks if every listener of the specified gateway has a certain condition (in our case - `ListenerConditionResolvedRefs`).

This new function will be called inside `GatewayAndHTTPRoutesMustBeAccepted`.

Let me know what you think about it @mlavacca @robscott @sunjayBhatia

shaneutt: @robscott

mlavacca: > So to my understanding, we need to check inside `GatewayAndHTTPRoutesMustBeAccepted` if every listener has the `ListenerConditionResolvedRefs` set to true.
> 
> I think the easiest way to do it is to add a new function (say we call it `GatewayListenersMustHaveCondition`) similar to `HTTPRouteMustHaveCondition`, which checks if every listener of the specified gateway has a certain condition (in our case - `ListenerConditionResolvedRefs`).
> 
> This new function will be called inside `GatewayAndHTTPRoutesMustBeAccepted`.

Yeah, it sounds good to me.

levikobi: The requested change has been added, so I resolved this conversation.
Please reopen if necessary.

conformance/utils/kubernetes/helpers.go
```diff
@@ -267,9 +267,12 @@ func MeshNamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig conf
 	require.NoErrorf(t, waitErr, "error waiting for %s namespaces to be ready", strings.Join(namespaces, ", "))
 }
 
-// GatewayAndHTTPRoutesMustBeAccepted waits until the specified Gateway has an IP
-// address assigned to it and the Route has a ParentRef referring to the
-// Gateway. The test will fail if these conditions are not met before the
+// GatewayAndHTTPRoutesMustBeAccepted waits until:
+// 1. The specified Gateway has an IP address assigned to it.
+// 2. The route has a ParentRef referring to the Gateway.
+// 3. All the gateway's listeners have the ListenerConditionResolvedRefs set to true.
```
robscott, line 273: Do we have any tests that currently expect ListenerConditionResolvedRefs to be false? Since this is such a wide-reaching change, I'm concerned we may break some implementations with this change. Have you run this test against any implementations?

robscott (commented):
Thanks @levikobi!

conformance/utils/kubernetes/helpers.go
mlavacca: Yes, we do have a bunch of tests that expect `ListenerConditionResolvedRefs` to be false:
- https://github.com/kubernetes-sigs/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-invalid-route-kind.go#L67
- https://github.com/kubernetes-sigs/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-invalid-tls-certificateref.go#L49
- https://github.com/kubernetes-sigs/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-secret-invalid-reference-grant.go#L53
- https://github.com/kubernetes-sigs/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-secret-missing-reference-grant.go#L53

I tested the changes against Kong's implementation and it works 👍

conformance/utils/kubernetes/helpers.go
```diff
@@ -297,18 +300,24 @@ func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutCo
...
 		HTTPRouteMustHaveParents(t, c, timeoutConfig, routeNN, parents, namespaceRequired)
 	}
 
+	resolvedRefsCondition := metav1.Condition{
+		Type:   string(v1beta1.ListenerConditionResolvedRefs),
+		Status: metav1.ConditionTrue,
+		Reason: "", // any reason
+	}
+
+	GatewayListenersMustHaveCondition(t, c, timeoutConfig, gw.NamespacedName, resolvedRefsCondition)
```
mlavacca, line 319: Since we are already adding the `ResolvedRefs` condition here, I think we should check all the positive-polarity conditions:
- `ResolvedRefs`
- `Accepted`
- `Programmed`

Ideally, `GatewayListenersMustHaveCondition` should check the three above conditions to be set and true, before going on. I've created https://github.com/kubernetes-sigs/gateway-api/issues/2415 for this purpose, as it looks like a lack of consistency in our suite. We can address it in another PR, as it is a bit out of scope for this PR, but I wanted to highlight this aspect here as well.

mlavacca (approved):
Thanks @levikobi!

/approve

leaving the lgtm to @robscott to be sure his concern is addressed.

michaelbeaumont (approved):
LGTM, passes with Kuma

conformance/utils/kubernetes/helpers.go
levikobi: @robscott  I tried running it against Contour's implementation but got some errors. It didn't seem like a problem in the conformance though, but missing assignments of the `ResolvedRefs`.

Then I discovered @sunjayBhatia had already opened https://github.com/projectcontour/contour/pull/5668 to address that 😅

Nevertheless I'm glad to hear that Kong and Kuma works.

robscott:
Thanks @levikobi!

/lgtm

PR #2253, opened 2023-07-31
Add initial set of CEL validations for HTTPRoute

gauravkghildiyal:
# What type of PR is this?
/kind feature

# What this PR does / why we need it:
Add initial set of CEL validations for HTTPRoute. 

This includes most (but not all) of the HTTPRoute validations offered by the webhook. 

## What remains?
**Only 2 kinds of validations remain.**

The validations which have not been ported along with the reason and details are available in https://github.com/gauravkghildiyal/gateway-api/blob/pending-httproute-validations/apis/v1beta1/validation/httproute.go. All validations which were successfully ported have been removed from this file.

### 1 "Namespace" is a reserved word in CEL.
Since namespace is a reserved word, a validation that involves a field named `namespace` seems to getting rejected. We need to figure out if there is a way around this. 
image

### 2 Estimated cost limit exceeds threshold
CEL attempts to estimate the [runtime cost of a validation](https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#estimated-cost-limit). Validations that require O(n^2) algorithms on larger cardinality fields end up crossing the allowed threshold. One such place is within `[]gatewayv1b1.HTTPHeaderMatch` where we want to ensure unique (case-insensitive) header names.

Fixes #

```release-note
Introduce validations through CEL for HTTPRoute.
```

### DIFF
--- a/apis/v1beta1/object_reference_types.go
+++ b/apis/v1beta1/object_reference_types.go
@@ -91,6 +91,8 @@ type SecretObjectReference struct {
 // References to objects with invalid Group and Kind are not valid, and must
 // be rejected by the implementation, with appropriate Conditions set
 // on the containing object.
+//
+// +kubebuilder:validation:XValidation:message="Must have port for Service reference",rule="(size(self.group) == 0 && self.kind == 'Service') ? has(self.port) : true"
 type BackendObjectReference struct {
 	// Group is the group of the referent. For example, "gateway.networking.k8s.io".
 	// When unspecified or empty string, core API group is inferred.
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -550,6 +550,10 @@ spec:
                                       required:
                                       - name
                                       type: object
+                                      x-kubernetes-validations:
+                                      - message: Must have port for Service reference
+                                        rule: '(size(self.group) == 0 && self.kind
+                                          == ''Service'') ? has(self.port) : true'
                                   required:
                                   - backendRef
                                   type: object
@@ -639,6 +643,23 @@ spec:
                                       required:
                                       - type
                                       type: object
+                                      x-kubernetes-validations:
+                                      - message: replaceFullPath must be set when
+                                          type is set to 'ReplaceFullPath'
+                                        rule: 'self.type == ''ReplaceFullPath'' ?
+                                          has(self.replaceFullPath) : true'
+                                      - message: type must be 'ReplaceFullPath' when
+                                          replaceFullPath is set
+                                        rule: 'has(self.replaceFullPath) ? self.type
+                                          == ''ReplaceFullPath'' : true'
+                                      - message: replacePrefixMatch must be set when
+                                          type is set to 'ReplacePrefixMatch'
+                                        rule: 'self.type == ''ReplacePrefixMatch''
+                                          ? has(self.replacePrefixMatch) : true'
+                                      - message: type must be 'ReplacePrefixMatch'
+                                          when replacePrefixMatch is set
+                                        rule: 'has(self.replacePrefixMatch) ? self.type
+                                          == ''ReplacePrefixMatch'' : true'
                                     port:
                                       description: "Port is the port to be used in
                                         the value of the `Location` header in the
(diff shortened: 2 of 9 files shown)


gauravkghildiyal:
Added another validation which was blocked on figuring out some regexp parsing behaviour. Now only two category of validations remain:

1. Seems like CEL might consider `namespace` field to be a reserved word. This seems to be blocking validations that involve the any field named `namespace`
image

2. The second category of validations remaining are still the "cost exceeded" kind.

(Will update the description with latest details)

apis/v1beta1/object_reference_types.go
```diff
@@ -91,6 +91,8 @@ type SecretObjectReference struct {
 // References to objects with invalid Group and Kind are not valid, and must
 // be rejected by the implementation, with appropriate Conditions set
 // on the containing object.
+//
+// +kubebuilder:validation:XValidation:message="Must have port for Service reference",rule="((!has(self.group) || size(self.group) == 0) && (!has(self.kind) || self.kind == 'Service')) ? has(self.port) : true"
```
robscott, line 95: Nit: In my experimenting I thought that defaults applied before CEL validation so it may be possible to simplify this. With that said, I could also be remembering wrong here. In any case this is not a blocker.

hack/cel-validation/httproute_test.go
```diff
@@ -0,0 +1,1393 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:       "invalid because multipler filters are repeated",
```
robscott, line 1155:
```suggestion
			name:       "invalid because multiple filters are repeated",
```

hack/cel-validation/httproute_test.go
```diff
@@ -0,0 +1,1393 @@
...
+		},
+		{
+			name:       "type must be 'ReplaceFullPath' when replaceFullPath is set",
+			wantErrors: []string{"type must be 'ReplaceFullPath' when replaceFullPath is set"},
+			pathModifier: gatewayv1b1.HTTPPathModifier{
+				ReplaceFullPath: ptrTo("foo"),
+			},
+		},
+		{
+			name: "valie ReplacePrefixMatch",
```
robscott, line 1325:
```suggestion
			name: "valid ReplacePrefixMatch",
```

config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
```diff
@@ -2991,12 +3310,97 @@ spec:
...
+                                          when replacePrefixMatch is set
+                                        rule: 'has(self.replacePrefixMatch) ? self.type
+                                          == ''ReplacePrefixMatch'' : true'
                                   type: object
                               required:
                               - type
                               type: object
+                              x-kubernetes-validations:
+                              - message: filter.RequestHeaderModifier must be nil
+                                  if the HTTPRouteFilter.Type is not RequestHeaderModifier
```
robscott, line 3336: I don't think we've been very consistent with this, but we should try to match the upstream pattern here where field names/paths match what a user would see when using YAML. For example, look at the core validation here: https://github.com/kubernetes/kubernetes/blob/2c6c4566eff972d6c1320b5f8ad795f88c822d09/pkg/apis/core/validation/validation.go#L1041-L1045.

This comment applies throughout the PR, but will not repeat it everywhere to reduce noise.

```suggestion
                              - message: filter.requestHeaderModifier must be nil
                                  if the HTTPRouteFilter.Type is not RequestHeaderModifier
```

robscott (commented):
Thanks @gauravkghildiyal, this is really great! A few small nits, but overall both the validation and corresponding tests look remarkably thorough.

hack/cel-validation/httproute_test.go
gauravkghildiyal: Thanks! Changed

hack/cel-validation/httproute_test.go
gauravkghildiyal: Thanks! Changed.

apis/v1beta1/object_reference_types.go
gauravkghildiyal: Yes you are absolutely right! Sorry didn't notice the defaults. Appreciate the closer look :)

config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
gauravkghildiyal: Sure can do! I've changed all such occurrences. The places which are still kept capitalized have the intent of referring to the enum (like `RequestRedirect`) instead of the field (`requestRedirect`). If in some place, the intent is not clear enough, I can try and reframe the statement.

gauravkghildiyal (commented):
Appreciate the thorough reviews!

robscott (commented):
Thanks @gauravkghildiyal!

/lgtm
/approve

PR #2254, opened 2023-07-31
Route timeout conformance tests

frankbu:
**What type of PR is this?**

/kind test
/area conformance

Conformance tests for HTTPRoute request and backend request timeouts.

Fixes #

```release-note
NONE
```


conformance/base/manifests.yaml
```diff
@@ -223,6 +223,53 @@ spec:
...
+      app: infra-backend-delay
+  template:
+    metadata:
+      labels:
+        app: infra-backend-delay
+    spec:
+      containers:
+      - name: infra-backend-delay
+        # TODO: can we enhance echo server to support a delayed response and use it here instead of httpbin?
+        image: docker.io/kong/httpbin
```
howardjohn, line 257: We already use Istoi's "echo" in mesh tests which support this - maybe can use that

conformance/tests/httproute-timeouts.go
```diff
@@ -0,0 +1,110 @@
...
+		ns := "gateway-conformance-infra"
+		routeNN := types.NamespacedName{Name: "request-timeout", Namespace: ns}
+		gwNN := types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		testCases := []http.ExpectedResponse{
+			{
+				Request:   http.Request{Path: "/echo"},
+				Backend:   "infra-backend-v1",
```
howardjohn, line 53: does no `Response` set imply 200? if not, we should add it

howardjohn (commented):
Mostly LGTM. The timing may be tricky though. Too small of numbers / differences and it may be flaky, too high and its slow..

conformance/base/manifests.yaml
frankbu: Great! I didn't know the Istio echo server supported delay.

Is this the Istio echo image you mean? `image: gcr.io/k8s-staging-gateway-api/echo-server:v20230505-v0.7.0-rc1-10-g497e67da`

Then we just call it with a path like this: `/?delay=1s`?

frankbu:
> Too small of numbers / differences and it may be flaky, too high and its slow..

@howardjohn I tried to find a sweet spot by making the timeout half a second and then calling a service delayed 1 sec and no delay, so we have a half second buffer in either direction and only a couple of slow tests with half second waits.

conformance/base/manifests.yaml
```diff
@@ -223,6 +223,42 @@ spec:
...
 apiVersion: v1
 kind: Service
+metadata:
+  name: infra-backend-delay
+  namespace: gateway-conformance-infra
+spec:
+  selector:
+    app: infra-backend-delay
+  ports:
+  - protocol: HTTP
```
SRodi, line 233: @frankbu It looks like this should be `TCP` - conformance tests fail with 

```
Service "infra-backend-delay" is invalid: spec.ports[0].protocol: Unsupported value: "HTTP": supported values: "SCTP", "TCP", "UDP"
```

SRodi: if this is set to `TCP` then tests are passing :

```bash
❯ go test ./conformance/conformance_test.go -args -gateway-class=istio -supported-features=Mesh --debug=true
ok  	command-line-arguments	51.791s
```

frankbu: Fixed. Thanks for testing it @SRodi!

tokers (approved):

conformance/base/manifests.yaml
```diff
@@ -223,6 +223,42 @@ spec:
 ---
 apiVersion: v1
 kind: Service
+metadata:
```
arkodg, line 226: do we need a new service and deployment for this case ?
the delay is being instrumented using the request query param

conformance/tests/httproute-timeouts.go
```diff
@@ -0,0 +1,108 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteTimeoutRequest)
+	ConformanceTests = append(ConformanceTests, HTTPRouteTimeoutBackendRequest)
+}
+
+var HTTPRouteTimeoutRequest = suite.ConformanceTest{
```
arkodg, line 34: any way to instrument a retry so we can differentiate b/w `HTTPRouteTimeoutRequest` & `HTTPRouteTimeoutBackendRequest` ?

conformance/base/manifests.yaml
frankbu: It's a different echo service (different image). It's the echo service used for mesh testing, which supports delay. The other services used for gateway testing are using a different echo service, which do not support the query param. It's also nice to have a specific named service for the delaying backend, for readability, IMO.

arkodg: maybe you just want to bump the echo sever used in the project which is actually the istio app
https://github.com/kubernetes-sigs/gateway-api/blob/main/docker/Dockerfile.echo

conformance/tests/httproute-timeouts.go
frankbu: I'm thinking that once we add retries to the API (future GEP-1731) we'll be able to add that kind of coverage. For now, just making sure both timeouts work.

arkodg: cool that works

arkodg: maybe you can add a TODO comment in there around this ?

conformance/base/manifests.yaml
arkodg: also most implementation are running conformance tests on CI with minimal resources such as GH Runners, so prefer less resources to be instantiated

frankbu: Maybe I'm missing something, but I thought that there were 2 different echo services with different implementations.

1. `gcr.io/k8s-staging-ingressconformance/echoserver:v...` used in `base/manfiest.yaml`
2. `gcr.io/k8s-staging-gateway-api/echo-server:v...` used in `mesh/manifest.yaml`

My understanding is that the second one comes from Istio and supports delay, but the first one does not.

Are you suggesting that we should switch to use the Istio one for everything in both manifest.yaml files? If so, then we don't need a separate service for the timeout test, we could just use backend-v1, for example, with the delay param?

I think that if I'm understanding you, I agree that would be better, but I'd rather leave that for a separate cleanup PR, instead of changing all of that as part of this one. WDYT?

conformance/tests/httproute-timeouts.go
frankbu: Good idea! Done.

arkodg (approved):
/approve

thanks @frankbu, hoping this has been manually tested in at least 1 implementation

frankbu:
> thanks @frankbu, hoping this has been manually tested in at least 1 implementation

Yep, it's tested and passing in Istio.

frankbu:
@robscott @shaneutt Is there anything else needed to finish this off? If not, can we merge it now?

conformance/base/manifests.yaml
robscott: Sorry I missed this! It probably is simpler to standardize around a single base image, so I'm fine with this change. I think it's definitely preferable over adding new Pods (and by extension increasing the resources required to run conformance tests). Maybe for this PR just swap one existing deployment over to the newer image and then we can have a follow up to get the rest.

conformance/tests/httproute-timeouts.go
```diff
@@ -0,0 +1,112 @@
...
+			tc := testCases[i]
+			t.Run(tc.GetTestCaseName(i), func(t *testing.T) {
+				t.Parallel()
+				http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, tc)
+			})
+		}
+	},
+}
+
+var HTTPRouteTimeoutBackendRequest = suite.ConformanceTest{
```
robscott, line 76: Nit: Prefer to have test file names directly match test names + manifest names, so would rather split this out to a separate file.

conformance/tests/httproute-timeouts.go
```diff
@@ -0,0 +1,112 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

robscott (commented):
Thanks @frankbu! Sorry I'd lost track of this one, just a few nits but otherwise should be good to go.

conformance/base/manifests.yaml
frankbu: Turns out that the two echo services are very different. The echo service used by the gateway conformance framework (in `base/manifest.yaml`) sends back `application/json` content containing all of the http request parameters. The Istio one just sends back whatever was in the request body. The gateway-api conformance framework is very much dependent on the `application/json` response, it's even a little difficult to write a test that uses a different target service. The test case needs to set `ExpectedRequest: &http.ExpectedRequest{}` to explicitly disable checking the response for the `application/json` that's expected from the echo service:
```
	testCases := []http.ExpectedResponse{
			{
				Request:         http.Request{Path: "/backend-timeout"},
				ExpectedRequest: &http.ExpectedRequest{},
				Response:        http.Response{StatusCode: 200},
			}, {
```

Anyway, I think that if we want to simplify things, we could add delay support to the gateway echo service, so that we can use it for all the gateway tests (i.e., `base/manifest.yaml`) but I think we'll probably still need to use the other (Istio) echo service for the mesh testing (`mesh/manifest.yaml`).

frankbu:
@robscott I believe I've addressed your review comments and confirmed that the new tests are passing in Istio. Please let me know if there's anything else needed before merging.

Regarding the two echo images used for conformance testing, I think we're stuck with them both, at least for now. See my comment here: https://github.com/kubernetes-sigs/gateway-api/pull/2254/files#r1340651689

conformance/base/manifests.yaml
robscott: Thanks for looking into this! Agree that it's likely out of scope to switch everything over to a different base image here. Not sure if you've got time, but it may be relatively straightforward to add support for a `delay` query param to the main image we're using, probably somewhere around here: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/master/images/echoserver/echoserver.go#L142. 

Longer term, I think it would be good to bring this into the Gateway API repo so we don't have to go to another mostly frozen repo to make small adjustments like this. If you want to give that a shot, I've created https://github.com/kubernetes-sigs/gateway-api/issues/2445 to track it.

frankbu: > it may be relatively straightforward to add support for a delay query param to the main image we're using, probably somewhere around here: https://github.com/kubernetes-sigs/ingress-controller-conformance/blob/master/images/echoserver/echoserver.go#L142.

Yes, I agree that's where to add it. Just need to add basically the same code that Istio has here: https://github.com/istio/istio/blob/da3bda9fd4bcbfc1873e0baab6eaddb4ad50aebe/pkg/test/echo/server/endpoint/http.go#L254

@SRodi or I will make sure to do this and #2445 relatively soon.

As far as this PR is concerned, I would like to get it merged now, if possible, since it works and I'd like to have it running in the Istio build integration testing. I'll definitely take the TODO to send a followup PR to clean it up to use the main image as soon as it's available with the delay function. WDYT?

robscott:
Thanks @frankbu!

/lgtm

PR #2263, opened 2023-08-05
Document the route kinds compatible for each protocolType

### LINKED ISSUE #1515: docs: specify the route kinds compatible with each protocolType

The listener spec field `allowedRoutes.kinds` documentation states:
```go
// Kinds specifies the groups and kinds of Routes that are allowed to bind
// to this Gateway Listener. When unspecified or empty, the kinds of Routes
// selected are determined using the Listener protocol.
//
// A RouteGroupKind MUST correspond to kinds of Routes that are compatible
// with the application protocol specified in the Listener's Protocol field.
// If an implementation does not support or recognize this resource type, it
// MUST set the "ResolvedRefs" condition to False for this Listener with the
// "InvalidRouteKinds" reason.
```

But there is no indication about the core kinds compatible with each `protocolType`. For this reason, an indication of which core routes are compatible for each protocol type is an important piece of information to add to the documentation.

**Why this is needed**:

This is needed to clearly state the core route kinds supported by each protocol type.

abs007:
**What type of PR is this?**
/kind documentation

This PR documents the route kinds compatible for each protocolType

Fixes #1515 

```release-note
NONE
```


shaneutt:
/cc @mlavacca 

Wanted to check if this is what you had in mind when you created #1512

mlavacca (commented):
Thanks for the PR, @abs007!

It looks like you need to run `make update-codegen` to fix the CI issue. 

> Wanted to check if this is what you had in mind when you created https://github.com/kubernetes-sigs/gateway-api/discussions/1512 

Yep, I meant something similar. However, I'm not 100% sure about the exhaustiveness and the correctness of the data. I think it makes sense to double-check it with the maintainers.

/cc @shaneutt @robscott @youngnick

youngnick:
Thanks for laying this out so clearly, @abs007!

Seeing it like this, makes me think about the fact that the `protocol` is currently effectively a way to limit the kinds of Routes that may attach to a Listener - it's almost syntactic sugar for specific `allowedRoutes` config. I seem to recall having discussions about whether to allow `protocol: TLS` listener to have HTTPRoutes attach (implying that TLS mode must also be set to `Terminate`), but it seems like it may be better to keep things at this very simple level.

It's probably also worth checking if we have any conformance tests that don't fit into these buckets too.

abs007:
Thank you @youngnick! 

Can you give a bit more context into the conformance tests bit?

apis/v1beta1/gateway_types.go
```diff
@@ -476,6 +476,14 @@ type AllowedRoutes struct {
...
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
+	//
+	// The following list states the Route kinds compatible for each protocolType
+	//
+	// HTTP: HTTPRoute, GRPCRoute
+	// HTTPS: HTTPRoutes, GRPCRoutes
+	// TCP: TCPRoute
+	// UDP:	UDPRoute
+	// TLS: TLSRoute
```
skriss, line 486: TLS protocol + TCPRoute should also be supported, right? (ref. https://gateway-api.sigs.k8s.io/guides/tls/ and https://gateway-api.sigs.k8s.io/concepts/api-overview/#route-summary-table)

mlavacca, line 486:
```suggestion
	// HTTP: HTTPRoutes, GRPCRoutes
	// HTTPS: HTTPRoutes, GRPCRoutes
	// TCP: TCPRoutes
	// UDP:	UDPRoutes
	// TLS: TLSRoutes
```

youngnick:
Thanks for the changes @abs007, I meant that it would be good to check that there's nothing else that we conformance test for - I suspect @skriss may have picked up one extra case, but it would be good if we could make sure there aren't any others.

abs007:
@youngnick and is this something that I verify by going through the code in the conformance folder? Asking as I'm a bit confused.

youngnick:
Yeah, I don't think there's any quicker way than to check the conformance tests to make sure we're not allowing any other types of traffic. Sorry.

abs007:
That's alright. I just needed confirmation on where to look at.

abs007:
I've pushed a protocol:route combination that I could find.

mlavacca:
> I've pushed a protocol:route combination that I could find.

Can you point out where did you find such a combination, please?

abs007:
From the table here under the TLS Route section: https://gateway-api.sigs.k8s.io/concepts/api-overview/#route-summary-table

Precisely, this part: "Routing of TLS protocols including HTTPS"

mlavacca (commented):
lgmt, thank you! Did you manage to check that our conformance tests are compliant with this table?

youngnick:
HI @abs007, sorry to have put you on the spot earlier. I think that this change is fine as it is, once it's been rebased. Are you still able to pick it up?

abs007:
@youngnick Let me know if the current state of the pr is good enough for merging

apis/v1/gateway_types.go
```diff
@@ -483,6 +483,14 @@ type AllowedRoutes struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
+	//
+	// The following list states the Route kinds compatible for each protocolType
```
youngnick, line 487: One last change, folks are allowed to make their own Route types, and their own protocolTypes, so let's update this wording to be about included ones.

```suggestion
	// The following list states the Route kinds compatible for each protocolType for
	// Route kinds and protocolTypes included in Gateway API
```

youngnick:
Sorry @abs007, you'll just need to run `make generate` and then commit and repush.

apis/v1/gateway_types.go
```diff
@@ -483,6 +483,15 @@ type AllowedRoutes struct {
...
+	// For route kinds and protocolTypes included in Gateway API
```
shaneutt, line 488:
```suggestion
	// for route kinds and protocolTypes included in Gateway API
```

apis/v1/gateway_types.go
```diff
@@ -483,6 +483,15 @@ type AllowedRoutes struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=8
+	//
+	// The following list states the Route kinds compatible for each protocolType
+	// for route kinds and protocolTypes included in Gateway API
+	//
+	// HTTP: HTTPRoutes, GRPCRoutes
+	// HTTPS: HTTPRoutes, GRPCRoutes, TLSRoute
```
mlavacca, line 491:
```suggestion
	// HTTPS: HTTPRoutes, GRPCRoutes, TLSRoutes
```

site-src/concepts/api-overview.md
```diff
@@ -256,6 +256,17 @@ following mechanisms:
 If none of the above are specified, a Gateway listener will trust Routes
 attached from the same namespace that support the listener protocol.
 
+The following table shows the **Route kinds** compatible for each **protocolType**
+
+|protocolType|core kind/s compatible|
+|------------|----------------------|
+| HTTP | HTTPRoutes, GRPCRoutes |
+| HTTPS |	HTTPRoutes, GRPCRoutes, TLSRoute |
```
mlavacca, line 264:
```suggestion
| HTTPS |	HTTPRoutes, GRPCRoutes, TLSRoutes |
```

mlavacca (commented):
Hi @abs007! this lgtm, if you can rebase it, I think we can merge it 👍

candita:
We haven't documented some of these changes completely.
/revert a518fef845afa9a73257684770022f9eeb29401d

PR #2267, opened 2023-08-06
Use '$(POD_NAMESPACE)' instead of hardcoded namespace

dihmandrake:
Properly reads the namespace from the actual deployment

**What type of PR is this?**

Do not hardcode the namespace for Pod arguments. The existing 'POD_NAMESPACE' variable suggests that the namespace should have actually been read at runtime. Similar to nginx-ingress in https://github.com/kubernetes/ingress-nginx/blob/d712dd9d92f1f46eab5f55d4643f3a45a7a97f90/charts/ingress-nginx/templates/admission-webhooks/job-patch/job-createSecret.yaml#L52C15-L52C43

Addtionally, this should enable easier deployments into non default namespaces with (i.e -) kustomize namespace setting

Fixes #

```release-note
* Enable deployments to non-default namespaces
```

### DIFF
--- a/config/webhook/certificate_config.yaml
+++ b/config/webhook/certificate_config.yaml
@@ -92,8 +92,8 @@ spec:
         imagePullPolicy: IfNotPresent
         args:
         - create
-        - --host=gateway-api-admission-server,gateway-api-admission-server.gateway-system.svc
-        - --namespace=gateway-system
+        - --host=gateway-api-admission-server,gateway-api-admission-server.$(POD_NAMESPACE).svc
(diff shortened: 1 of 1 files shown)


youngnick:
Hmm, neat, seems like a good idea to me.

/ok-to-test

youngnick:
Yes, this seems like a great improvement, thanks @dihmandrake.

/approve
/hold
for extra review

robscott:
Thanks @dihmandrake!

/lgtm
/hold cancel

PR #2272, opened 2023-08-07
docs: update signet in owner aliases

shaneutt:
While I was looking at this file today, I noticed we were out of date with the source at https://github.com/kubernetes/org/blob/main/OWNERS_ALIASES


shaneutt:
Closing for now until we get Zappas configuration fixed.

Issue #2110, opened 2023-06-12
Configure Mutual TLS Termination in a Gateway


frankbu:
**What would you like to be added**:

GatewayTLSConfig should support 3 variants of Terminate mode:

1. Simple TLS
2. Mutual (mTLS)
3. Mesh (Terminate mTLS managed by a mesh) - GAMMA only

**Why this is needed**:

The required certificateRef depends on which of the above Terminate modes is used.

This can currently be supported using an implementation-specific option. Istio, for example, has:

```
    tls:
      mode: Terminate
      certificateRefs:
      - name: example-credential
      options:
        gateway.istio.io/tls-terminate-mode: MUTUAL
```

Providing a standard API for this would be much better.

robscott:
I think this is at least partially addressed by https://github.com/kubernetes-sigs/gateway-api/pull/2080.

/cc @arkodg

arkodg:
yes 2. should be addressed by #2080, which should also enable the capability for 3. (mTLS for Gateways in a mesh)

robscott:
@frankbu I think this issue may need to be split up into smaller pieces. We're definitely making progress towards what I'd call "manually-configured frontend and backend mTLS" with both Arko's GEP (#2080) and BackendTLSPolicy. It's possible that some automation could be built on top of that as well. I _think_ what you're really focused on here is some kind of standardized way to sign up for the UX that is common among Service meshes - automatic mTLS everywhere. If that's the case, I'd recommend a separate issue focused on just that part.

PR #2113, opened 2023-06-14
GEP-1897 Explicit Backend TLS Connection Configuration (was TLS from Gateway to Backend...) Update with API details

candita:
**What type of PR is this?**

/kind gep

**What this PR does / why we need it**:

This PR adds the API details for GEP-1897 TLS from Gateway to Backend for ingress.  GEP-1897 has already had the questions answered - what do we want to do, and why do we want to do this.  This update attempts to answer - how do we do this.

**Which issue(s) this PR fixes**:

Fixes #1897

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2273, opened 2023-08-07
Adds API to GEP-91: Client Certificate Validation

arkodg:
**What type of PR is this?**

/kind gep

**What**
Defines an API in GEP-91 to support Client Certificate Validation for TLS Terminating at the Gateway
using a CA Certificate as a trusted anchor.

**Why**
Allows the platform admin configuring the Gateway resource to only select trusted clients to connect to the Gateway

Relates to https://github.com/kubernetes-sigs/gateway-api/issues/91 & https://github.com/kubernetes-sigs/gateway-api/issues/2110

### DIFF
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -72,11 +72,11 @@ spec:
       - kind: Secret
         group: ""
         name: foo-example-com-cert
-    clientValidation:
-      caCertificateRefs:
-      - kind: ConfigMap
-        group: ""
-        name: foo-example-com-ca-cert
+      clientValidation:
+        caCertificateRefs:
+        - kind: ConfigMap
+          group: ""
+          name: foo-example-com-ca-cert
 ```
 
 ## References
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -88,7 +88,7 @@ type GatewayTLSConfig struct {
 
 // FrontendTLSValidation holds configuration information that can be used to validate
 // the frontend initiating the TLS connection
-type FrontendTLSValidationContext struct {
+type FrontendTLSValidation struct {
     // CACertificateRefs contains one or more references to
     // Kubernetes objects that contain TLS certificates of
     // the Certificate Authorities that can be used
@@ -99,8 +99,8 @@ type FrontendTLSValidationContext struct {
     // Implementations MAY choose to support attaching multiple CA certificates to
     // a Listener, but this behavior is implementation-specific.
     //
-    // Support: Core - An optional single reference to a single Kubernetes Secret
-    // and to a single Kubernetes ConfigMap with the CA certificate in a key named `ca.crt`.
+    // Support: Core - A single reference to a single Kubernetes ConfigMap
+    // with the CA certificate in a key named `ca.crt`.
     //
     // Support: Implementation-specific (More than one reference, or other kinds
     // of resources).
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -71,7 +71,7 @@ type ClientValidationContext struct {
     //
     // +kubebuilder:validation:MaxItems=8
     // +optional
-    CACertificateRefs []corev1.ObjectReference `json:”caCertificateRefs,omitempty”`
+    CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
 }
 
 ```
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -51,7 +51,8 @@ type ClientValidationContext struct {
     // the Certificate Authorities that can be used
     // as a trust anchor to validate the certificates presented by the client.
     //
-    // A single CA certificate reference to a Kubernetes ConfigMap kind has "Core" support.
+    // A single CA certificate reference to a Kubernetes ConfigMap or Secret kind
+    // has "Core" support.
     // Implementations MAY choose to support attaching multiple CA certificates to
     // a Listener, but this behavior is implementation-specific.
     //
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -56,8 +56,8 @@ type ClientValidationContext struct {
     // Implementations MAY choose to support attaching multiple CA certificates to
     // a Listener, but this behavior is implementation-specific.
     //
-    // Support: Core - An optional single reference to a Kubernetes Secret and ConfigMap,
-    // with the CA certificate in a key named `ca.crt`.
+    // Support: Core - An optional single reference to a single Kubernetes Secret
+    // and to a single Kubernetes ConfigMap with the CA certificate in a key named `ca.crt`.
     //
     // Support: Implementation-specific (More than one reference, or other kinds
     // of resources).
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -7,7 +7,7 @@
 
 ## TLDR
 
-This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
+This GEP proposes a way to validate the TLS certificate presented by the frontend client to the server
 (Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
 ## Goals
@@ -35,26 +35,27 @@ This table highlights the support. Please feel free to add any missing implement
 
 ### API
 
-* Introduce a `FrontendValidation` field of type `FrontEndTLSValidationContext` within [GatewayTLSConfig][] that can be used to validate the peer(frontend) with which the TLS connection is being made.
-* Introduce a `caCertificateRefs` field within `FrontEndTLSValidationContext` that can be used to specify a list of CA Certificates that can be used as a trust anchor to validate the certificates presented by the client.
-* This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the frontend (downstream client).
+* Introduce a `FrontendValidation` field of type `FrontendTLSValidation` within [GatewayTLSConfig][] that can be used to validate the peer (frontend) with which the TLS connection is being made.
+* Introduce a `caCertificateRefs` field within `FrontendTLSValidation` that can be used to specify a list of CA Certificates that can be used as a trust anchor to validate the certificates presented by the client.
+* This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the backend peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the frontend client on the connection between the Gateway and the frontend. Both these configurations can coexist at the same time without affecting one another.
 
 #### GO
 
 ```go
 
 type GatewayTLSConfig struct {
-......
-    // FrontendValidation holds configuration for validating the frontend (client).
+    ......
+    // FrontendValidation holds configuration information for validating the frontend (client).
     // Setting this field will require clients to send a client certificate
-    // required for validation. In browsers this may result in a dialog appearing 
+    // required for validation during the TLS handshake. In browsers this may result in a dialog appearing 
     // that requests a user to specify the client certificate.
     // The maximum depth of a certificate chain accepted in verification is Implementation specific.
-    FrontendValidation *FrontEndTLSValidationContext `json:"frontEndValidation,omitempty"`
+    FrontendValidation *FrontendTLSValidation `json:"frontendValidation,omitempty"`
 }
 
-// FrontEndTLSValidationContext holds configuration that can be used to validate the frontend in the TLS connection
-type FrontEndTLSValidationContext struct {
+// FrontendTLSValidation holds configuration information that can be used to validate
+// the frontend initiating the TLS connection
+type FrontendTLSValidationContext struct {
     // CACertificateRefs contains one or more references to
     // Kubernetes objects that contain TLS certificates of
     // the Certificate Authorities that can be used
@@ -103,7 +104,7 @@ spec:
       - kind: Secret
         group: ""
         name: foo-example-com-cert
-      frontEndValidation:
+      frontendValidation:
         caCertificateRefs:
         - kind: ConfigMap
           group: ""
@@ -114,10 +115,10 @@ spec:
 
 This section highlights use cases that may be covered in a future iteration of this GEP
 
-* Using system CA certificates as the trust anchor to validate the certificates presented by the client.
+* Using system CA certificates as the trust anchor to validate the certificates presented by the frontend client.
 * Supporting a mode where validating client certficates is optional, useful for debugging and migrating to strict TLS.
-* Supporting an optional `subjectAltNames` field within `ClientValidationContext` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. This field falls under authorization and will be revisited when authorization is tackled as a whole in the project.
-* Specifying the verification depth in the client certificates chain
+* Supporting an optional `subjectAltNames` field within `FrontendTLSValidation` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. This field falls under Authorization, the initial focus here is on Client Authentication and will be revisited when Authorization is tackled as a whole in the project.
+* Specifying the verification depth in the client certificate chain. This is being deferred because the default verification depth differs across implementations.
 
 
 ## References
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -38,10 +38,43 @@ This table highlights the support. Please feel free to add any missing implement
 * Introduce a `FrontendValidation` field of type `FrontendTLSValidation` within [GatewayTLSConfig][] that can be used to validate the peer (frontend) with which the TLS connection is being made.
 * Introduce a `caCertificateRefs` field within `FrontendTLSValidation` that can be used to specify a list of CA Certificates that can be used as a trust anchor to validate the certificates presented by the client.
 * This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the backend peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the frontend client on the connection between the Gateway and the frontend. Both these configurations can coexist at the same time without affecting one another.
+* Also introduce a `ObjectReference` structure that can be used to specify `caCertificateRefs` references.
 
 #### GO
 
 ```go
+// ObjectReference identifies an API object including its namespace.
+//
+// The API object must be valid in the cluster; the Group and Kind must
+// be registered in the cluster for this reference to be valid.
+//
+// References to objects with invalid Group and Kind are not valid, and must
+// be rejected by the implementation, with appropriate Conditions set
+// on the containing object.
+type ObjectReference struct {
+	// Group is the group of the referent. For example, "gateway.networking.k8s.io".
+	// When unspecified or empty string, core API group is inferred.
+	Group Group `json:"group"`
+
+	// Kind is kind of the referent. For example "ConfigMap" or "Service".
+	Kind Kind `json:"kind"`
+
+	// Name is the name of the referent.
+	Name ObjectName `json:"name"`
+
+	// Namespace is the namespace of the referenced object. When unspecified, the local
+	// namespace is inferred.
+	//
+	// Note that when a namespace different than the local namespace is specified,
+	// a ReferenceGrant object is required in the referent namespace to allow that
+	// namespace's owner to accept the reference. See the ReferenceGrant
+	// documentation for details.
+	//
+	// Support: Core
+	//
+	// +optional
+	Namespace *Namespace `json:"namespace,omitempty"`
+}
 
 type GatewayTLSConfig struct {
     ......
@@ -80,7 +113,7 @@ type FrontendTLSValidationContext struct {
     //
     // +kubebuilder:validation:MaxItems=8
     // +kubebuilder:validation:MinItems=1
-    CACertificateRefs []SecretObjectReference `json:"caCertificateRefs,omitempty"`
+    CACertificateRefs []ObjectReference `json:"caCertificateRefs,omitempty"`
 }
 
 ```
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -99,7 +99,7 @@ type FrontendTLSValidation struct {
     // Implementations MAY choose to support attaching multiple CA certificates to
     // a Listener, but this behavior is implementation-specific.
     //
-    // Support: Core - A single reference to a single Kubernetes ConfigMap
+    // Support: Core - A single reference to a Kubernetes ConfigMap
     // with the CA certificate in a key named `ca.crt`.
     //
     // Support: Implementation-specific (More than one reference, or other kinds
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -10,6 +10,21 @@
 This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
 (Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
+### Existing support in Implementations
+
+This feature is widely supported in implementations that support Gateway API.
+This table highlights the support. Please feel free to add any missing implementations not mentioned below.
+
+| Implementation | Support       |
+|----------------|------------|
+| Apache APISIX  | [ApisixTls.Client.CASecret](https://apisix.apache.org/docs/ingress-controller/tutorials/mtls/#mutual-authentication)      |
+| Contour        | [HTTPProxy.Spec.VirtualHost.Tls.ClientValidation.CASecret](https://projectcontour.io/docs/v1.17.1/config/tls-termination/)      |
+| Emissary Ingress| [TlSContext.Spec.Secret](https://www.getambassador.io/docs/emissary/latest/topics/running/tls/mtls)     |
+| Gloo Edge      | [VirtualService.Spec.SSLConfig.SecretRef](https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls/#configuring-downstream-mtls-in-a-virtual-service)      |
+| Istio          | [Gateway.Spec.Servers.TLS.Mode](https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway)      |
+| Kong           | [mTLS Plugin](https://docs.konghq.com/hub/kong-inc/mtls-auth/)      |
+| Traefik        | [TLSOption.Spec.ClientAuth](https://doc.traefik.io/traefik/https/tls/#client-authentication-mtls)    |
+
 ## Goals
 - Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
 
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -10,7 +10,13 @@
 This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
 (Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
-### Existing support in Implementations
+## Goals
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
+
+## Non-Goals
+- Define other fields that can be used to verify the client certificate such as the Certificate Hash or Subject Alt Name. 
+
+## Existing support in Implementations
 
 This feature is widely supported in implementations that support Gateway API.
 This table highlights the support. Please feel free to add any missing implementations not mentioned below.
@@ -24,12 +30,6 @@ This table highlights the support. Please feel free to add any missing implement
 | Istio          | [Gateway.Spec.Servers.TLS.Mode](https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway)      |
 | Kong           | [mTLS Plugin](https://docs.konghq.com/hub/kong-inc/mtls-auth/)      |
 | Traefik        | [TLSOption.Spec.ClientAuth](https://doc.traefik.io/traefik/https/tls/#client-authentication-mtls)    |
-
-## Goals
-- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
-
-## Non-Goals
-- Define other fields that can be used to verify the client certificate such as the Certificate Hash or Subject Alt Name. 
 
 ### API
 
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -14,7 +14,7 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
 - Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
 
 ## Non-Goals
-- Define other fields that can be used to verify the client certificate such as the Certificate Hash or Subject Alt Name. 
+- Define other fields that can be used to verify the client certificate such as the Certificate Hash. 
 
 ## Existing support in Implementations
 
@@ -37,8 +37,6 @@ This table highlights the support. Please feel free to add any missing implement
 to the Gateway.
 * Introduce a `caCertificateRefs` field within `ClientValidationContext` that can be used to specify a list of CA Certificates that
 can be used as a trust anchor to validate the certificates presented by the client.
-* Add CEL validation to ensure that `caCertificateRefs` cannot be empty. This validation will be removed once more fields are added
-into `clientValidation`.
 * This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
 
 #### GO
@@ -70,8 +68,15 @@ type ClientValidationContext struct {
     // "RefNotPermitted" reason.
     //
     // +kubebuilder:validation:MaxItems=8
-    // +optional
+    // +kubebuilder:validation:MinItems=1
     CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
+    // SubjectAltNames contains one or more alternate names to verify
+    // the subject identity in the certificate presented by the client.
+    //
+    // Support: Core
+    //
+    // +optional
+    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
 }
 
 ```
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -33,9 +33,10 @@ This table highlights the support. Please feel free to add any missing implement
 
 ### API
 
-* Introduce a `clientValidation` field of type `ClientValidationContext` within [GatewayTLSConfig][] that can be used to validate the client initiating the TLS connection
+* Introduce a `Client` field of type `TLSClientContext` within [GatewayTLSConfig][] to hold TLS configuration specific to the client.
+* Introduce a `Validation` field of type `TLSValidationContext` within the `Client` field that can be used to validate the peer with which the TLS connection is being made.
 to the Gateway.
-* Introduce a `caCertificateRefs` field within `ClientValidationContext` that can be used to specify a list of CA Certificates that
+* Introduce a `caCertificateRefs` field within `ValidationContext` that can be used to specify a list of CA Certificates that
 can be used as a trust anchor to validate the certificates presented by the client.
 * This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
 * Introduce an optional `subjectAltNames` field within `ClientValidationContext` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. The maximum number of alternate names that can be specified is implementation defined.
@@ -43,10 +44,17 @@ can be used as a trust anchor to validate the certificates presented by the clie
 #### GO
 
 ```go
-// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
-// to the Gateway.
-// By default, no client specific configuration is validated.
-type ClientValidationContext struct {
+// TLSClientContext holds configuration specific to the client initiating the TLS connection.
+type TLSClientContext struct {
+    // Validation holds configuration around validating the client initiating the TLS connection.
+    //
+    // +optional
+    Validation *TLSValidationContext `json:"validation,omitempty"`
+}
+
+
+// TLSValidationContext holds configuration that can be used to validate the peer in the TLS connection
+type TLSValidationContext struct {
     // CACertificateRefs contains one or more references to
     // Kubernetes objects that contain TLS certificates of
     // the Certificate Authorities that can be used
@@ -71,7 +79,7 @@ type ClientValidationContext struct {
     //
     // +kubebuilder:validation:MaxItems=8
     // +kubebuilder:validation:MinItems=1
-    CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
+    CACertificateRefs []SecretObjectReference `json:"caCertificateRefs,omitempty"`
     // SubjectAltNames contains one or more alternate names to verify
     // the subject identity in the certificate presented by the client.
     //
@@ -103,11 +111,12 @@ spec:
       - kind: Secret
         group: ""
         name: foo-example-com-cert
-      clientValidation:
-        caCertificateRefs:
-        - kind: ConfigMap
-          group: ""
-          name: foo-example-com-ca-cert
+      client:
+        validation:
+          caCertificateRefs:
+          - kind: ConfigMap
+            group: ""
+            name: foo-example-com-ca-cert
 ```
 
 ## Deferred
@@ -115,6 +124,7 @@ spec:
 This section highlights use cases that may be covered in a future iteration of this GEP
 
 * Using system CA certificates as the trust anchor to validate the certificates presented by the client.
+* Supporting a permissive TLS mode where untrusted clients are allowed in, useful for debugging and migrating to strict TLS.
 
 ## References
 
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -8,20 +8,20 @@
 ## TLDR
 
 This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
-(Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).
+(Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
 ## Goals
-- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trusted anchor to validate the certificates presented by the client.
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client.
 
 ## Non-Goals
 - Define other fields that can be used to verify the client certificate such as the Certificate Hash or Subject Alt Name. 
 
 ### API
 
-* Introduce a `clientValidation` field with [Gateway.Tls][] that can be used to validate the client intiating the TLS connection
-to the Gateway
-* Introduce a `caCerficateRefs` field within `clientValidation` that can be used to specify a list of CA Certificates that
-can be used as a trusted anchor to validate the certificates presented by the client
+* Introduce a `clientValidation` field of type `ClientValidationContext` within [GatewayTLSConfig][] that can be used to validate the client initiating the TLS connection
+to the Gateway.
+* Introduce a `caCertificateRefs` field within `ClientValidationContext` that can be used to specify a list of CA Certificates that
+can be used as a trust anchor to validate the certificates presented by the client.
 
 #### GO
 
@@ -33,12 +33,17 @@ type ClientValidationContext struct {
     // CACertificateRefs contains one or more references to
     // Kubernetes objects that contain TLS certificates of
     // the Certificate Authorities that can be used to
-    // as a trusted anchor to validate the certificates presented by the client.
+    // as a trust anchor to validate the certificates presented by the client.
     //
-    // A single CACertificateRef to a Kubernetes ConfigMap with a key called `ca.crt`
-    // has "Core" support.
-    // Implementations MAY choose to support attaching multiple certificates to
+    // A single CACertRef to a Kubernetes ConfigMap kind has "Core" support.
+    // Implementations MAY choose to support attaching multiple CA certificates to
     // a Listener, but this behavior is implementation-specific.
+    //
+    // Support: Core - An optional single reference to a Kubernetes ConfigMap,
+    // with the CA certificate in a key named `ca.crt`.
+    //
+    // Support: Implementation-specific (More than one reference, or other kinds
+    // of resources).
     //
     // References to a resource in different namespace are invalid UNLESS there
     // is a ReferenceGrant in the target namespace that allows the certificate
@@ -46,7 +51,7 @@ type ClientValidationContext struct {
     // "ResolvedRefs" condition MUST be set to False for this listener with the
     // "RefNotPermitted" reason.
     //
-    // +kubebuilder:validation:MaxItems=64
+    // +kubebuilder:validation:MaxItems=8
     // +optional
     CACertificateRefs []corev1.ObjectReference `json:”caCertificateRefs,omitempty”`
 }
@@ -59,7 +64,7 @@ type ClientValidationContext struct {
 apiVersion: gateway.networking.k8s.io/v1beta1
 kind: Gateway
 metadata:
-  name: mtls-basic
+  name: client-validation-basic
 spec:
   gatewayClassName: acme-lb
   listeners:
@@ -83,4 +88,4 @@ spec:
 
 [TLS Handshake Protocol]: https://www.rfc-editor.org/rfc/rfc5246#section-7.4
 [Certificate Path Validation]: https://www.rfc-editor.org/rfc/rfc5280#section-6
-[Gateway.TLS]: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig
+[GatewayTLSConfig]: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig
--- a/geps/gep-91/index.md
+++ b/geps/gep-91/index.md
@@ -38,6 +38,7 @@ to the Gateway.
 * Introduce a `caCertificateRefs` field within `ClientValidationContext` that can be used to specify a list of CA Certificates that
 can be used as a trust anchor to validate the certificates presented by the client.
 * This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
+* Introduce an optional `subjectAltNames` field within `ClientValidationContext` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. The maximum number of alternate names that can be specified is implementation defined.
 
 #### GO
 
@@ -77,6 +78,7 @@ type ClientValidationContext struct {
     // Support: Core
     //
     // +optional
+    // +kubebuilder:validation:MinItems=1
     SubjectAltNames []string `json:"subjectAltNames,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


geps/gep-91.md
```diff
@@ -16,7 +16,69 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used to
+    // as a trusted anchor to validate the certificates presented by the client.
+    //
+	// A single CACertificateRef to a Kubernetes ConfigMap with a key called `ca.crt`
```
spacewander, line 38: Is there an indentation issue?

youngnick: Yeah, looks like the old tab/space indentation dilemma.

arkodg: thanks for catching this, fixed !

arkodg:
highlighting the key decisions that need to be made here
* naming - 
  * `clientValidation` to hold client specific information to validate
  * `caCertificateRefs` - to hold the CA certs
* `Secret` vs  `ConfigMap` to hold the CA Cert
* key name for the CA Cert, `ca.crt` is what this GEP proposes
* allowing these resources to exist in a separate namespace or not - GEP proposal allows this using `ReferenceGrant` similar to what is done for the `certificateRefs` field

spacewander:
> Secret vs ConfigMap to hold the CA Cert

IMHO, the CA cert only contains the public key part so a ConfigMap is safe enough to store it. Sometimes, people store the CA cert in the Secret because they also store the private key in a different entry of the same object.

spacewander:
> > Secret vs ConfigMap to hold the CA Cert
> 
> IMHO, the CA cert only contains the public key part so a ConfigMap is safe enough to store it. Sometimes, people store the CA cert in the Secret because they also store the private key in a different entry of the same object.

I think putting the CA cert in a separate object would be a better practice (so only the CA cert will be referenced). If I have to choose one from Secret and ConfigMap, I will vote for ConfigMap.

youngnick:
It's conceivable that we may want to add the ability to use a CA cert from an issues TLS Secret (using the not-quite-standard `ca.crt` key in a `kubernetes.io/tls` type Secret), but that's not a blocker, since we can always add that later.

Additionally, this change should move this GEP to `Implementable`, I think, and then once the types are present in the Experimental channel, this GEP will move to `Experimental`.

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+    tls:
+      certificateRefs:
+      - kind: Secret
+        group: ""
+        name: foo-example-com-cert
+      clientValidation:
+        caCertificateRefs:
+        - kind: ConfigMap
+          group: ""
+          name: foo-example-com-ca-cert
```
frankbu, line 79: I think Istio users would like to be able to point this to a Secret (instead of ConfigMap), specifically the same secret that `certificateRefs` point to, only with an additional `ca.crt` key containing the client's CA cert.

@howardjohn WDYT?

frankbu:
I think this also
Fixes #2110

geps/gep-91.md
howardjohn: Yes, secret reference is common practice in CNCF landscape, pioneered by cert-manager. Not in the core `tls` SEcret type, but its de-facto standard.

I don't mind config map as well of course.

arkodg:
@robscott @youngnick @shaneutt any edits I can make to this PR to help it move forward ?
I did notice that the [Backend TLS PR](https://github.com/kubernetes-sigs/gateway-api/pull/2113) has merged which defined the cert refs as a `ConfigMapObjectReference` .  This PR took the approach of using `ObjectReference` to give implementations the flexiblity to define the `ca.cert` in a ConfigMap or as a Secret (within the same TLS Cert used to define server certs). I can go ahead and make the change to this PR to keep the approach identical across the apis.

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+* Introduce a `caCerficateRefs` field within `clientValidation` that can be used to specify a list of CA Certificates that
+can be used as a trusted anchor to validate the certificates presented by the client
+
+#### GO
+
+```go
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
```
tokers, line 32: I think here the `ClientValidationContext` should claim some unmentioned behaviors:

1. What if the `caCertificateRefs` field is empty? It's implementation-specific, or all the clients will be rejected, or passed?
2. What's the verify depth of the client certificate? Only the server certificate and the intermediate CA, or the whole chain (Maybe we can add another field like `VerifyDepth`)?

frankbu:
> I did notice that the https://github.com/kubernetes-sigs/gateway-api/pull/2113 has merged which defined the cert refs as a ConfigMapObjectReference

I think the question for both this and #2113 is do we have a preferred default kind for certs, and if so, should it be ConfigMap or Secret? If Secret, maybe both PRs should be using SecretObjectReference?

candita:
> > I did notice that the #2113 has merged which defined the cert refs as a ConfigMapObjectReference
> 
> I think the question for both this and #2113 is do we have a preferred default kind for certs, and if so, should it be ConfigMap or Secret? If Secret, maybe both GEPs should be using SecretObjectReference?

For reference, we temporarily chose ConfigMapObjectReference because the data doesn't require specific tls.key and tls.data.  https://github.com/kubernetes-sigs/gateway-api/pull/2113#discussion_r1231584950 mentioned the use of `ClusterTrustBundle`, which seemed popular.

geps/gep-91.md
candita, line 1: > This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
> (Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).

Could you elaborate?  I'm not sure everyone would conclude that this is commonly referred to as mutual TLS.

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
 ## Non-Goals
 - Define other fields that can be used to verify the client certificate such as the Cerificate Hash or Subject Alt Name. 
 
+### API
+
+* Introduce a `clientValidation` field with [Gateway.Tls][] that can be used to validate the client intiating the TLS connection
```
candita, line 21:
```suggestion
* Introduce a `clientValidation` field within [GatewayTlsConfig][] that can be used to validate the client initiating the TLS connection
```

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+to the Gateway
+* Introduce a `caCerficateRefs` field within `clientValidation` that can be used to specify a list of CA Certificates that
```
candita, line 23:
```suggestion
* Introduce a `caCertificateRefs` field within `clientValidation` that can be used to specify a list of CA Certificates that
```

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+can be used as a trusted anchor to validate the certificates presented by the client
```
candita, line 24:
```suggestion
can be used as a trust anchor to validate the certificates presented by the client
```

geps/gep-91.md
candita: If you could also elaborate on what you mean by client and server in this proposal, that would be helpful.

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+
+#### GO
+
+```go
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
```
candita, line 29:
```suggestion
// ClientValidationContext holds configuration information that can be used to validate the client initiating the TLS connection
```

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used to
+    // as a trusted anchor to validate the certificates presented by the client.
```
candita, line 36:
```suggestion
    // the Certificate Authorities that can be used
    // as a trust anchor to validate the certificates presented by the client.
```

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+    // A single CACertificateRef to a Kubernetes ConfigMap with a key called `ca.crt`
+    // has "Core" support.
+    // Implementations MAY choose to support attaching multiple certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // References to a resource in different namespace are invalid UNLESS there
+    // is a ReferenceGrant in the target namespace that allows the certificate
+    // to be attached. If a ReferenceGrant does not allow this reference, the
+    // "ResolvedRefs" condition MUST be set to False for this listener with the
+    // "RefNotPermitted" reason.
```
candita, line 47: Can you also add docs on the behavior when the CACertificateRefs is empty but needed?  Echoing https://github.com/kubernetes-sigs/gateway-api/pull/2273/files#r1334951299 to some extent.

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
 ## Non-Goals
 - Define other fields that can be used to verify the client certificate such as the Cerificate Hash or Subject Alt Name. 
 
+### API
+
+* Introduce a `clientValidation` field with [Gateway.Tls][] that can be used to validate the client intiating the TLS connection
+to the Gateway
+* Introduce a `caCerficateRefs` field within `clientValidation` that can be used to specify a list of CA Certificates that
+can be used as a trusted anchor to validate the certificates presented by the client
+
```
candita, line 25: `GatewayTLSConfig` already has a `CertificateRefs` field.  Why not just add the `CACertificateRefs` field there?https://github.com/kubernetes-sigs/gateway-api/blob/800024e13f634545934ae10d9de82513f6f2798c/apis/v1beta1/gateway_types.go#L374

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
 ## Non-Goals
 - Define other fields that can be used to verify the client certificate such as the Cerificate Hash or Subject Alt Name.
```
candita, line 17: Can you point out which of the TLS (or mTLS) use cases in https://docs.google.com/document/d/17sctu2uMJtHmJTGtBi_awGB0YzoCLodtR6rUNmKMCs8/edit#heading=h.cxuq8vo8pcxm that are in your goal?

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
 ## Non-Goals
 - Define other fields that can be used to verify the client certificate such as the Cerificate Hash or Subject Alt Name. 
 
+### API
```
candita, line 19: I'd like to know for which route types (HTTP, TLS, TCP, UDP, GRPC, etc) , protocols, and termination types (edge, passthrough?) this might be applicable?

geps/gep-91.md
```diff
@@ -16,7 +16,71 @@ This GEP proposes a way to validate the TLS certificate presented by the downstr
...
+    // Implementations MAY choose to support attaching multiple certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // References to a resource in different namespace are invalid UNLESS there
+    // is a ReferenceGrant in the target namespace that allows the certificate
+    // to be attached. If a ReferenceGrant does not allow this reference, the
+    // "ResolvedRefs" condition MUST be set to False for this listener with the
+    // "RefNotPermitted" reason.
+    //
+    // +kubebuilder:validation:MaxItems=64
```
candita, line 49: It's easier to change this from smaller to larger if needed.
```suggestion
    // +kubebuilder:validation:MaxItems=8
```

geps/gep-91.md
frankbu: @candita can you clarify why you think the client and server here needs elaboration. The server is a Gateway listener (it says that) and the client is then whoever is making requests to the Gateway. Is there something more subtle that you're looking for?

geps/gep-91.md
frankbu: Use case number 7, I believe. Are you just asking or suggesting the GEP should include a link?

geps/gep-91.md
youngnick: Great questions - my thoughts:

1. We could copy the pattern we used for BackendTLSPolicy here, and have something that suggests whether or not to use the system CAs.
2. I think that unless we have a _very_ good reason, we should start with mandating the behavior that the full chain is checked, and then we can add configurability later if we need it. We should make the default case as secure as possible, and not have easily-accessible knobs that make things less secure unless we're _sure_ we need them.

geps/gep-91.md
youngnick: I think the issue here is that, while this is definitely a mutual handshake, calling something "mutual TLS" _can_ be taken to mean a very specific thing (dynamically-issued, frequently-rotated certificates mutually authenticated between services as a way to handle service-to-service authentication). This is definitely not that thing, but one thing that @candita and I have learned through _much longer than we would have liked_ experience is that you _really_ shouldn't use "mutual TLS" without some explanation of what sense you mean it in, or people will believe it's the wrong one.

Yes, I _know_ that "mutual TLS" in this case is technically correct, but it also can be misleading if, for example, a person is used to reading that phrase in a mesh context, not in a reverse-proxy-or-Ingress context.

geps/gep-91.md
tokers: @youngnick Fair enough, but we should clarify them in the GEP to reduce the ambiguity

jdoylei:
Hi, does anyone have a sense whether this PR is close to approval and merging?  We're trying to establish guidelines and a road map for our applications that use Gateway ingress with downstream client certificates.  We began using Contour as ingress controller for multiple applications, along with Gateway API, but a subset of those apps need client certificates from the end user.  We're currently able to support this with Contour, but that subset of apps has to deviate from the others in that they can't use Gateway API's HTTPRoute, and instead have to use Contour-specific APIs (HTTPProxy) to trigger client certificate requests.  It would be great to have an idea whether this is a good short-term strategy and whether it's likely that we could consolidate our apps around HTTPRoute in the future.  Thanks!

arkodg:
this PR was blocked until Gateway API went GA, now that it has, will address comments and push another commit this week

youngnick:

geps/gep-91.md
arkodg: removed the controversial mTLS term from the descriptions

geps/gep-91.md
arkodg: as frank mentioned this is 7, does the doc have to be linked into this GEP ?

geps/gep-91.md
arkodg: this is same as the downstream info mentioned in https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls, does this need to be added in the GEP ?

geps/gep-91.md
arkodg: because there are multiple attributes that can be used to validate the client, and this GEP starts off with CACertRef
here is an exhaustive list supported in Envoy https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-certificatevalidationcontext

geps/gep-91.md
candita: @arkodg please add a reference to the use case name that's in your goal ( and the doc ).  I strongly believe that tying GEPs in to existing work helps us realize what we've accomplished and what we're still missing from that original list.

geps/gep-91.md
candita: I would also remove the ambiguous phrase "downstream client", and replace it with something that clearly indicates what resource/s are connecting.

geps/gep-91.md
candita: I didn't see an update for this.

geps/gep-91.md
candita: In that case, I don't quite understand how this GEP ends up so brief compared to the GEP for BackendTLSPolicy.  There seems to be so much missing in terms of what you'd you need for working client certificate validation.

It would probably make sense to add at least:
- why we need this
- a description of use cases for mesh and non-mesh
- a section on how the request flow changes, using the GAMMA request flow in https://gateway-api.sigs.k8s.io/concepts/gamma/?h=request+flow#request-flow and/or the original request flow in https://gateway-api.sigs.k8s.io/concepts/api-overview/?h=request+flow#request-flow
- something that describes how the client certificate validation interacts with edge termination and backend TLS termination
- a section on prior art, e.g. how this is implemented already in a few implementations.

@youngnick do you agree?

geps/gep-91.md
arkodg: I dont think downstream client is ambiguous, can you share any other alternatives ?

geps/gep-91.md
arkodg: done

geps/gep-91.md
arkodg: im not sure I understand this, `caCaertificateRefs` is an optional field within `clientValidation`, if its empty it is assumed that is the intent and it is valid, because in the future `clientValidation` may have more fields within it such as `san`

geps/gep-91.md
youngnick: @candita yes, I think this GEP would be well served by adding some things:
- why we need this - "It's table stakes and widely supported", worded better than that is probably okay, but it should be recorded in the GEP. And the "widely supported" part would at least need a table listing some proxy implementations and their support. 
- I think this also needs a "Deferred" section that covers, at a minimum, "How do I use the system certs?". I think that the custom CA use case will be far more common, but I could be wrong.
- I agree that this GEP should specify what happens when you don't have other TLS details (it doesn't work or gets rejected), and what happens when BackendTLSPolicy is enabled (the client certificate will be used for the Listener connection only, and _not_ reused for the backend connection). This last item - reusing the Listener client cert on a backend connection - feels like something folks may ask for, but we should definitely have out of scope could be another one for the "Deferred" section.

geps/gep-91/index.md
```diff
@@ -1,22 +1,92 @@
...
+#### GO
+
+```go
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used to
```
frankbu, line 35:
```suggestion
    // the Certificate Authorities that can be used
```

geps/gep-91/index.md
```diff
@@ -1,22 +1,92 @@
...
+    // as a trust anchor to validate the certificates presented by the client.
+    //
+    // A single CACertRef to a Kubernetes ConfigMap kind has "Core" support.
```
frankbu, line 38:
```suggestion
    // A single CACertificateRefs to a Kubernetes ConfigMap kind has "Core" support.
```

geps/gep-91/index.md
```diff
@@ -1,22 +1,92 @@
...
+    // Implementations MAY choose to support attaching multiple CA certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // Support: Core - An optional single reference to a Kubernetes ConfigMap,
+    // with the CA certificate in a key named `ca.crt`.
+    //
+    // Support: Implementation-specific (More than one reference, or other kinds
+    // of resources).
+    //
+    // References to a resource in different namespace are invalid UNLESS there
```
frankbu, line 48:
```suggestion
    // References to resources in a different namespace are invalid UNLESS there
```

frankbu:
@arkodg @candita @youngnick Can we move this GEP forward soon?

geps/gep-91/index.md
```diff
@@ -1,22 +1,92 @@
...
+    // is a ReferenceGrant in the target namespace that allows the certificate
+    // to be attached. If a ReferenceGrant does not allow this reference, the
+    // "ResolvedRefs" condition MUST be set to False for this listener with the
+    // "RefNotPermitted" reason.
+    //
+    // +kubebuilder:validation:MaxItems=8
+    // +optional
+    CACertificateRefs []corev1.ObjectReference `json:”caCertificateRefs,omitempty”`
```
frankbu, line 56: Do we really want to use `corev1.ObjectReference` here? Seems that it should be either `SecretObjectReference` (same as GatewayTLSConfig.CertificateRefs) or `ConfigMapObjectReference` (same as gep-1897 BackendTLSPolicyConfig).

arkodg: pros for using `corev1.ObjectReference`
* this reference can be a `ConfigMap` with a `ca.crt` key or a `Secret` with a `ca.crt` (which can be the same secret that also has `tls.crt` and `tls.key` used in `CertificateRefs` )
* reference can live in a different namespace and `ResourceGrant` can be used to authorize the reference (similar to `CertificateRefs` )

geps/gep-91/index.md
```diff
@@ -1,22 +1,92 @@
...
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used to
+    // as a trust anchor to validate the certificates presented by the client.
+    //
+    // A single CACertRef to a Kubernetes ConfigMap kind has "Core" support.
+    // Implementations MAY choose to support attaching multiple CA certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // Support: Core - An optional single reference to a Kubernetes ConfigMap,
```
howardjohn, line 42: Should a Secret with ca.crt be core as well? its not uncommon for things like cert-manager etc to write these

arkodg: good point, we plan on adding support for it in Envoy Gateway, 

for now, its very close to the semantics of BackendTrafficPolicy (which defines a ConfigMap as Core)

adding another new `Core` constraint in this PR may slow down the completion of this PR imo, prefer
if its added incrementally. If the community gives a big thumbs up for this, happy to add it to this PR

arkodg:
updated the PR based on review comments ptal @youngnick @frankbu @candita

frankbu:
> another point for reviewers of this PR, the field is called CACertificateRefs

I prefer this longer name. Why isn't BackendTLSPolicy using the longer name also, to be consistent with the Gateway TLS spec?

cc @candita

geps/gep-91/index.md
```diff
@@ -1,22 +1,117 @@
...
+    // References to a resource in a different namespace are invalid UNLESS there
+    // is a ReferenceGrant in the target namespace that allows the certificate
+    // to be attached. If a ReferenceGrant does not allow this reference, the
+    // "ResolvedRefs" condition MUST be set to False for this listener with the
+    // "RefNotPermitted" reason.
+    //
+    // +kubebuilder:validation:MaxItems=8
+    // +optional
+    CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
+}
```
frankbu, line 75: Can we add SAN validation here? TLS without SAN verification isn't so good.

```suggestion

    // SubjectAltNames contains one or more alternate names to verify
    // the subject identity in the certificate presented by the client.
    //
    // Support: Core
    //
    // +optional
    SubjectAltNames []string `json:”subjectAltNames,omitempty”`
}
```

geps/gep-91/index.md
arkodg: Added Secret support to Core

geps/gep-91/index.md
arkodg: ive added SAN as a non goal to this GEP, only because it may delay the graduation of this GEP. Happy to add this diff if it doesnt

frankbu: I see the reverse problem wrt delay. If we don't add it here, it will delay the graduation of full/proper support for mTLS, because of all the extra overhead of starting up a second followup GEP to add SAN.

arkodg: cool, I'll add it, lets see how it goes ....

arkodg: added

arkodg:
seeing this failure locally
```
Verifying ./hack/../hack/verify-docker-build.sh
FAILED
```

as well as in the CI
```
Verifying docker images
ERROR: no builder "gateway-api-builder" found
gateway-api-builder
Building and pushing echo-advanced image (from Istio) ...linux/amd64,linux/arm64
#0 building with "gateway-api-builder" instance using docker-container driver
 #1 [internal] booting buildkit
181 skipped lines...
 #18 [linux/amd64 stage-1 2/3] COPY --from=builder /go/src/sigs.k8s.io/gateway-api/conformance/echo-basic/echo-basic /
#18 DONE 0.0s
 #17 [linux/arm64 builder 4/4] RUN go build -trimpath -ldflags="-buildid= -s -w" -o echo-basic .
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 2h0m0s timeout","severity":"error","time":"2024-02-06T18:37:12Z"}
++ early_exit_handler
++ '[' -n 183 ']'
++ kill -TERM 183
++ cleanup_dind
++ [[ true == \t\r\u\e ]]
++ echo 'Cleaning up after docker'
Cleaning up after docker
make: *** [Makefile:118: verify] Terminated
++ docker ps -aq
++ xargs -r docker rm -f
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
ERROR: failed to receive status: rpc error: code = Unavailable desc = error reading from server: EOF
4dbeda0d9a12
```

feels unrelated to this change

frankbu (approved):

geps/gep-91/index.md
```diff
@@ -1,22 +1,122 @@
...
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used
+    // as a trust anchor to validate the certificates presented by the client.
+    //
+    // A single CA certificate reference to a Kubernetes ConfigMap kind has "Core" support.
```
frankbu, line 54:
```suggestion
    // A single CA certificate reference to a Kubernetes ConfigMap or Secret kind
    // has "Core" support.
```

geps/gep-91/index.md
```diff
@@ -1,22 +1,122 @@
...
+    // Implementations MAY choose to support attaching multiple CA certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // Support: Core - An optional single reference to a Kubernetes Secret and ConfigMap,
+    // with the CA certificate in a key named `ca.crt`.
```
frankbu, line 59:
```suggestion
    // Support: Core - An optional references to a single Kubernetes Secret and to a
    // single Kuberentes ConfigMap, with the CA certificate in a key named `ca.crt`.
```

geps/gep-91/index.md
```diff
@@ -1,22 +1,122 @@
...
+    // +kubebuilder:validation:MaxItems=8
+    // +kubebuilder:validation:MinItems=1
+    CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
+    // SubjectAltNames contains one or more alternate names to verify
+    // the subject identity in the certificate presented by the client.
+    //
+    // Support: Core
+    //
+    // +optional
+    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```
frankbu, line 79: @arkodg See https://github.com/kubernetes-sigs/gateway-api/pull/2759#discussion_r1483295712
```suggestion
    // +kubebuilder:validation:MinItems=1
    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```

arkodg: afaik no implementation make SAN mandatory (only CA Cert is mandatory), curious why we want to do it here ?

frankbu: Maybe I'm mistaken, but IIUC we have `// +optional` which says the field is not required, but `MinItems=1` means that if the field is set then there must be at least 1 entry.

frankbu: @arkodg I tried it (`+optional` and `MinItems=1`) and it works as I said above, i.e., SAN is not mandatory.

arkodg: awesome thanks for checking, will update

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
+    tls:
+      certificateRefs:
+      - kind: Secret
+        group: ""
+        name: foo-example-com-cert
+      clientValidation:
+        caCertificateRefs:
+        - kind: ConfigMap
+          group: ""
+          name: foo-example-com-ca-cert
```
robscott, line 110: Have you considered making this a top level field instead of per listener? It's current placement requires SNI-matching prior to Client Cert validation which could be tricky. (I'm not necessarily advocating in one direction or the other here, just wanted to start a conversation).

frankbu: I'm not sure I understand, but I think all the TLS handshake info should be together like this.

Anyway, this PR is just implementing what @arkodg is proposing for GEP-91 in #2273, so design discussions should probably be there, and I'll keep this PR in sync with any changes that happen in that one.

arkodg: @robscott implementations are already doing SNI matching today to enforce listener isolation, `hostname` is defined within each `listener`, so this works fine, this even works when Gateway API listeners needs to be flattened out in the implementation data plane in case 2 listeners have the same `port`.
I didn't quite understand what the  top level definition would enable ?

robscott: > I think all the TLS handshake info should be together like this.

Jumping forward a bit, I think it's very likely that the client cert a Gateway uses to connect to backends will be configured per Gateway (not per Listener) based on the discussion at the community meeting earlier this week. Essentially the argument is that in that case, the client cert is representing the identity of the Gateway, not any individual listener.

If we were to take that argument further, would a Gateway admin rather define the set of clients it trusts in one place or individually for each Listener? 

> implementations are already doing SNI matching today to enforce listener isolation

Some may end up with a subset of this. For example, many cloud LBs prevent domain fronting but don't expose full SNI matching config. Preventing domain fronting is actually one way to meet the requirements of SNI/hostname matching in Gateways, but that logic may not extend to things like client cert validation.

One detail that's not clear to me from the list of implementations that support client cert validation in this GEP is how many support configuring this differently per SNI on the same VIP + Port. That would help show the viability of this. For what it's worth, in GCP this config is per VIP + Port, not per SNI (https://cloud.google.com/load-balancing/docs/mtls). Of course a single implementation does not need to impact the entire design here, my hunch is that GCP may not be the only one with this limitation.

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
 
 ## TLDR
 
 This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
-(Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).
+(Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
 ## Goals
-- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trusted anchor to validate the certificates presented by the client.
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
```
youngnick, line 14: TLS Use cases doc is
https://docs.google.com/document/d/17sctu2uMJtHmJTGtBi_awGB0YzoCLodtR6rUNmKMCs8/edit?usp=sharing

arkodg: yeah already linked on L125

frankbu:
> the naming of the field and its struct, is "clientValidation" the right one?

IMO, this is the right name, it's even consistent with the diagram in this section of the backend TLS doc: https://gateway-api.sigs.k8s.io/geps/gep-1897/#overview-what-do-we-want-to-do. I do agree that the backend doc could use some terminology improvements to be clear that there are 2 clients in the request flow: 1) the client of the gateway (the request client), and 2) the client of the backend (which happens to be the gateway iteself). Maybe just say `gateway` when referring to the backend client?

robscott (commented):
Thanks for all the work on this @arkodg!

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
+| Contour        | [HTTPProxy.Spec.VirtualHost.Tls.ClientValidation.CASecret](https://projectcontour.io/docs/v1.17.1/config/tls-termination/)      |
+| Emissary Ingress| [TlSContext.Spec.Secret](https://www.getambassador.io/docs/emissary/latest/topics/running/tls/mtls)     |
+| Gloo Edge      | [VirtualService.Spec.SSLConfig.SecretRef](https://docs.solo.io/gloo-edge/latest/guides/security/tls/server_tls/#configuring-downstream-mtls-in-a-virtual-service)      |
+| Istio          | [Gateway.Spec.Servers.TLS.Mode](https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-mutual-tls-ingress-gateway)      |
+| Kong           | [mTLS Plugin](https://docs.konghq.com/hub/kong-inc/mtls-auth/)      |
+| Traefik        | [TLSOption.Spec.ClientAuth](https://doc.traefik.io/traefik/https/tls/#client-authentication-mtls)    |
+
+### API
+
+* Introduce a `clientValidation` field of type `ClientValidationContext` within [GatewayTLSConfig][] that can be used to validate the client initiating the TLS connection
```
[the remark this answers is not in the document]
arkodg: @robscott the context here is `listener` && `tls` , so `clientValidation` shouldn't confuse the user with upstream TLS naming imo
@danehans what are the advantages of the top level `client` field here ? do you see more knobs being added to `client` . This location seems like the wrong home for Gateway Certs (needed for upstream TLS)

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
+* This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
+* Introduce an optional `subjectAltNames` field within `ClientValidationContext` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. The maximum number of alternate names that can be specified is implementation defined.
+
+#### GO
+
+```go
+// ClientValidationContext holds configuration that can be used to validate the client intiating the TLS connection
+// to the Gateway.
+// By default, no client specific configuration is validated.
+type ClientValidationContext struct {
```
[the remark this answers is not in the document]
arkodg: good point, I'd prefer if permissive mode is introduced later as an enhancement into this API

geps/gep-91/index.md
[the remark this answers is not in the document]
danehans: I agree that upstream TLS should stay as-is, e.g. `gateway.spec.listeners[].tls.certificateRefs`. If we wanted to set allowed ALPN protocols or a downstream client TLS connection timeout would that be considered client validation?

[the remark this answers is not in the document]
arkodg: ah I see where you're going with this, knobs like `alpn` should live under `gateway.spec.listeners[].tls`
here's an example of how we implemented in Envoy Gateway as a field within a policy
https://gateway.envoyproxy.io/latest/api/extension_types/#tlssettings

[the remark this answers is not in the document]
youngnick: Yeah, I think that ideally, we should be structuring `tls` stanzas similarly everywhere, something like this:

```yaml
tls:
  server:
    <existing TLS settings>
  client:
    validation:
      <some settings>
    othersetting:
```

Gateway is GA, so we can't change it now, but adding the `client` stanza in would really help.

Doing something like this now also means that we can move `BackendTLSPolicy` in that direction too - it's still experimental and alpha, so we can change things if necessary.

Additionally, this means that we could conceivably add a top-level `tls` stanza to Gateway that could hold Gateway-wide TLS settings (like, for example, a Gateway-wide TLS client certificate).

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
+    // CACertificateRefs contains one or more references to
+    // Kubernetes objects that contain TLS certificates of
+    // the Certificate Authorities that can be used
+    // as a trust anchor to validate the certificates presented by the client.
+    //
+    // A single CA certificate reference to a Kubernetes ConfigMap or Secret kind
+    // has "Core" support.
```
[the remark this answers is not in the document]
arkodg: hey @robscott is this a blocking comment ?

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: cool, I'll go ahead and make this change

geps/gep-91/index.md
```diff
@@ -1,22 +1,125 @@
...
+    // +kubebuilder:validation:MinItems=1
+    CACertificateRefs []SecretObjectReference `json:”caCertificateRefs,omitempty”`
+    // SubjectAltNames contains one or more alternate names to verify
+    // the subject identity in the certificate presented by the client.
+    //
+    // Support: Core
+    //
+    // +optional
+    // +kubebuilder:validation:MinItems=1
+    SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```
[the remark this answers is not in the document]
arkodg: @robscott CN is deprecated , highlighted in this RFC https://datatracker.ietf.org/doc/html/rfc2818#section-3.1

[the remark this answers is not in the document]
arkodg: pretty sure Envoy does not support CN validation https://github.com/envoyproxy/envoy/issues/4174, will need to cross check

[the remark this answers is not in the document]
arkodg: regarding 2. is the API you were hoping for ?

```
subjectAltNames:
- type: "DNS"
   value: client.example.com
```

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: ive gone ahead and made this change, @youngnick would be great if you can elaborate on which fields will call  `othersettings` their home

geps/gep-91/index.md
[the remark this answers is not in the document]
robscott: @costinm will follow up on this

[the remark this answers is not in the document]
youngnick: Yes, the `type` field should stay, and should be optional and defaulted to `DNS`. It at least also needs `URI` as an option.

geps/gep-91/index.md
```diff
@@ -1,22 +1,135 @@
...
+
+* Introduce a `Client` field of type `TLSClientContext` within [GatewayTLSConfig][] to hold TLS configuration specific to the client.
+* Introduce a `Validation` field of type `TLSValidationContext` within the `Client` field that can be used to validate the peer with which the TLS connection is being made.
+to the Gateway.
+* Introduce a `caCertificateRefs` field within `ValidationContext` that can be used to specify a list of CA Certificates that
+can be used as a trust anchor to validate the certificates presented by the client.
+* This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
+* Introduce an optional `subjectAltNames` field within `ClientValidationContext` that can be used to specify one or more alternate names to verify the subject identity in the certificate presented by the client. The maximum number of alternate names that can be specified is implementation defined.
```
costinm, line 42: I am not sure about this one - this is really authorization policy, and having a list of identities in the gateway itself is perhaps the worse for a gateway ( scale, personas doing the config, etc).

costinm: In Istio that's part of Authorization policy.

arkodg: @frankbu and @howardjohn from Istio requested for adding subjectAltName

geps/gep-91/index.md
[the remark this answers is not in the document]
costinm: To clarify: the deprecation is for CN use for _server_ identification, not client identification. 

Using DNS SAN for server identification is standard. Using DNS SAN for client identification is unusual - never saw it in use.

Email, URL SANs are used sometimes, different fields in the CN as well. I believe K8S client certs use the org or group.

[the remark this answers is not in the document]
costinm: Also as mentioned above - due to size of the yaml file and cost to update this, it would be far better to keep authz separate.

geps/gep-91/index.md
```diff
@@ -1,22 +1,135 @@
...
+
+#### GO
+
+```go
+// TLSClientContext holds configuration specific to the client initiating the TLS connection.
+type TLSClientContext struct {
```
costinm, line 48: Worth mentioning that adding this field will result in servers adding a 'certificate request' in the TLS server hello - which results in browsers asking for a cert in a dialog. This is pretty unfortunate UX for the browsers - many users are not aware of this. 

So best to use a separate hostname or port for the mtls if browsers are used.

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: we touched upon this in the call today, `client.Validation` is mainly being added to create a similar look and feel with the `BackendTLSPolicy` API

costinm (commented):
Looks good - see my other comments. 

One thing that is strange is having caCertificateRefs as a SecretObjectReference - but the example shows a ConfigMap, which is not secret. And in general the CA certs are not secrets. As long as users don't get confused and use config maps for private keys - it's fine, but not ideal naming. 

My top concern is adding a form of Authz in the Gateway resource - even for mesh it's not the best authz, but in the case of an ingress gateway with large number of clients it's really bad IMO. I understand we need authz - but that's not the best way to add it.

geps/gep-91/index.md
costinm: Well, Istio does have this field in istio Gateway - but we also have the authorization policy as a separate resource. 

I don't think this is the best UX for verifying client certificates - do we have any real world use case where client certificates are using SANs ( real internet clients - not mesh ), and real internet gateways authenticate by listing all the clients in the config file ? Even the old httpasswd was a separate config and didn't scale well...

geps/gep-91/index.md
```diff
@@ -1,22 +1,135 @@
...
 
 ## TLDR
 
 This GEP proposes a way to validate the TLS certificate presented by the downstream client to the server
-(Gateway Listener in this case) during a [TLS Handshake Protocol][], also commonly referred to as mutual TLS (mTLS).
+(Gateway Listener in this case) during a [TLS Handshake Protocol][].
 
 ## Goals
-- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trusted anchor to validate the certificates presented by the client.
+- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases][] document under point 7.
```
shaneutt, line 14:
```suggestion
- Define an API field to specify the CA Certificate within the Gateway Listener configuration that can be used as a trust anchor to validate the certificates presented by the client. This use case has been been highlighted in the [Gateway API TLS Use Cases] document under point 7.
```

?

geps/gep-91/index.md
danehans: @arkodg Without a solid use case, it's best to drop `subjectAltNames` from the API to move this PR forward.

geps/gep-91/index.md
danehans: @costinm is ^ the case even if `CACertificateRefs` refs a CA cert that resides as an anchor in the client's system trust bundle?

geps/gep-91/index.md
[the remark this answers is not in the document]
danehans: xref cert-manager secret ref: https://cert-manager.io/docs/usage/certificate/#target-secret

> The signed certificate and private key are stored in the specified Secret resource.

Although it might seem obvious, should the spec define the type of Secret, e.g. `type: kubernetes.io/tls`?

Since the field supports configmap or secret references, we should consider defining an equivalent type, e.g. `SecretOrConfigMapObjectReference`.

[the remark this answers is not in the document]
costinm: Isn't kubernetes.io/tls supposed to hold TLS private key and certificate ? 

IMO it is quite confusing to use Secret for public info like the root CAs to trust.

How about 'SYSTEM' ( i.e. client certs signed by the public roots of trust - many do sign email certs) ? 
Any prior art from kubeadmin or other k8s APIs dealing with setting the public certificates ? I know many support in-line certificates ( all webhooks ). Would be good to be consistent.

geps/gep-91/index.md
costinm: Yes. 

mTLS works by server asking the client for a cert in ServerHelo - if server doesn't ask, client won't send. And if server asks, a browser is required to show the cert selection dialog ( at least first time ? Haven't tried in a long time). 

Note that most servers have options to make client cert optional or required - both still ask for cert, but one will allow the request to proceed and use JWT or other forms of auth if client cert is not available. 

This should probably be covered in the 'prior art' and background section...

geps/gep-91/index.md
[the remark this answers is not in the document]
costinm: Please don't use the word 'permissive' - Istio has created a lot of confusion and mess on that area, really means 'allow insecure'.

Also there is nothing 'untrusted' about requests without client certs - JWTs are equivalent, and many systems allow multiple form of authentication. JWT bound to client certs are another form - that is different from client certs.

A section with the APIs used by Nginx, apache, etc would be helpful to get the terms right - I suspect what we're talking about is the 'optional' config, where server ask but client may not send client cert and use other forms of auth.

geps/gep-91.md
costinm: The term 'mTLS' is far older than meshes - and I don't think any recent implementation should 'take over' a well established term. Agree with 'upstream/downstream' - client and server are usually sufficient. 

My suggestion is to start with the survey of existing configs - Apache httpd, Nginx, servlets, vendor specific - and use the terms and keywords that are most common among them. There is no reason for Gateway API to invent new words or terms, and easier for users to have terms they are already using.

geps/gep-91/index.md
robscott: Note that both the original and suggestion have "been been", suggest removing one of those. I also generally prefer inline links to aliases to avoid confusion, unless a link is going to be reused a lot.

geps/gep-91/index.md
robscott: @howardjohn or @frankbu can you provide more info on why you'd like SANs to be configurable at this level?

frankbu: Good question about real world use cases. I was only requesting this feature because Istio supports it and we'd rather not add an Istio-specific gateway-api extension if we can avoid it. https://github.com/istio/istio/pull/48013.

@howardjohn?

howardjohn: I generally agree this probably makes sense to split out as a more general authz. However, there is no authz in gateway-api so its a bit tricky..

Generally any TLS usage ought to actually verify the identity in the certificates; rarely is it sufficient to just validate the issuing CA. So we would end up in a state where to properly use this feature, each vendor would need to have its own policy and users would need to use it, or be left in an insecure state.

arkodg: removing SAN for now, will add a comment in the GEP mentioning the reason - "it falls under the authz bucket so we can revisit this field once the project tackles authz as a whole"

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: I'm moving this back to only supporting `ConfigMap` as Core, this doesnt block implementations (such as Istio and Envoy Gateway) from also supporting Secret

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: removing SAN from this PR for now

geps/gep-91/index.md
arkodg: resolving this comment since its being addressed by Rob with https://github.com/kubernetes-sigs/gateway-api/pull/2910

geps/gep-91/index.md
dprotaso: Knative uses these features for the gateway=>backend. Since those requests can be handled by a revision instance or the activator component (when we scale up).

We don't want our activator component impersonating all revision instances and having to deal with so many certificates)

dprotaso: On the client=>gateway side wouldn't allowing subjectAltNames enable auth with certs?

eg. My internal corp site requires my machine's client cert to match the alternate names `*.internal.broadcom.net` or `*.internal.vmware.com` ?

geps/gep-91/index.md
```diff
@@ -1,22 +1,135 @@
...
+| Traefik        | [TLSOption.Spec.ClientAuth](https://doc.traefik.io/traefik/https/tls/#client-authentication-mtls)    |
+
+### API
+
+* Introduce a `Client` field of type `TLSClientContext` within [GatewayTLSConfig][] to hold TLS configuration specific to the client.
+* Introduce a `Validation` field of type `TLSValidationContext` within the `Client` field that can be used to validate the peer with which the TLS connection is being made.
+to the Gateway.
+* Introduce a `caCertificateRefs` field within `ValidationContext` that can be used to specify a list of CA Certificates that
+can be used as a trust anchor to validate the certificates presented by the client.
+* This new field is mutually exclusive with the [BackendTLSPolicy][] configuation which is used to validate the TLS certificate presented by the peer on the connection between the Gateway and the backend, and this GEP is adding support for validating the TLS certificate presented by the peer on the connection between the Gateway and the downstream client.
```
jdoylei, line 41: Would you mind clarifying "mutually exclusive"?  Let's say my use case is to have a Gateway that includes tls.client.validation.caCertificateRefs *and* has a backend Service with an attached BackendTLSPolicy and tls.caCertRefs. I'm indicating my trust for both the upstream connection (browser client cert) and downstream connection (backend server cert).  This use case - using both - is valid, right?

geps/gep-91/index.md
arkodg: thanks ! incorporated this in as a doc string

geps/gep-91.md
arkodg: I've add a doc string, mentioning that this is implementation specific today, because im seeing different default verifyDepth settings in different controller

geps/gep-91/index.md
arkodg: thanks fixed the typo

geps/gep-91/index.md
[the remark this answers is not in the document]
arkodg: cool, I'll update to be more specific - "support optional certificate validation" in the future

arkodg:
sorry for the force push 🙈

geps/gep-91/index.md
arkodg: yes, using both at the same time is valid, and mutually exclusive, let me update to make this clearer

geps/gep-91.md
arkodg: I agree with you costing, if you peek into the prior art section, most implementations call this mTLS, but since this is controversial, I've tried to stick with RFC terms as much as possible

geps/gep-91/index.md
```diff
@@ -1,22 +1,164 @@
...
+    // the Certificate Authorities that can be used
+    // as a trust anchor to validate the certificates presented by the client.
+    //
+    // A single CA certificate reference to a Kubernetes ConfigMap
+    // has "Core" support.
+    // Implementations MAY choose to support attaching multiple CA certificates to
+    // a Listener, but this behavior is implementation-specific.
+    //
+    // Support: Core - An optional single reference to a single Kubernetes Secret
+    // and to a single Kubernetes ConfigMap with the CA certificate in a key named `ca.crt`.
```
robscott, line 103: Nit: I'm not sure we really want to consider Secrets core support, that would mean we'd need to have conformance tests for both ConfigMaps and Secrets when we really just want people to use ConfigMaps and eventually ClusterTrustBundles.

robscott (commented):
Thanks @arkodg! One minor nit but otherwise LGTM as soon as #2910 merges.

/approve

robscott:
I don't think release notes are needed for GEP changes, can save them until the change that actually adds the API spec.

/release-note-none

geps/gep-91/index.md
arkodg: ah yes, thanks for finding this, fixed, was a left over, which I forgot to clean up

geps/gep-91/index.md
```diff
@@ -1,22 +1,164 @@
 # GEP-91: Client Certificate Validation for TLS terminating at the Gateway Listener
 
 * Issue: [#91](https://github.com/kubernetes-sigs/gateway-api/issues/91)
-* Status: Provisional
+* Status: Implementable
```
robscott, line 4: Before this formally moves to experimental, I think we need a corresponding metadata file for this GEP. Not going to block on this specific PR since it can easily be handled in a following.

robscott (commented):
Thanks @arkodg!

/lgtm

geps/gep-91.md
gawsoftpl: This issue is still live? 

In istio I can do that with:
```yaml
    tls:
      mode: Terminate
      certificateRefs:
      - name: httpbin-credential
      options:
        gateway.istio.io/tls-terminate-mode: MUTUAL
```

I try to setup gateway api with cilium (1.15.5) withtout istio and other sidecars, but MTLS not working
```yaml
    tls:
      certificateRefs:
      - kind: Secret
        name: demo-cert
      frontendValidation:
        caCertificateRefs:
        - kind: Secret
          group: ""
          name: demo-cert
```

PR #2280, opened 2023-08-09
fix: add check for unrecognized scheme in http redirect request scheme

shawnh2:
**What type of PR is this?**

/kind bug
/area conformance

running conformance test `HTTPRouteRedirectPortAndScheme`, will get a wired log in the output even if the redirect scheme is Well-Known. so add some check to ensure this log only be shown if the redirect scheme is a actual unrecognized one.

```
=== CONT  TestGatewayAPIConformance/HTTPRouteRedirectPortAndScheme/http-listener-on-80/1_request_to_'/scheme-nil-and-port-80'_should_receive_a_302
    httproute-redirect-port-and-scheme.go:159: Making GET request to http://172.18.255.200/scheme-nil-and-port-80
Sending Request:
< GET /scheme-nil-and-port-80 HTTP/1.1
< ...

Received Response:
< HTTP/1.1 302 Found
< ...

    http.go:348: Can't validate redirectPort for unrecognized scheme http  ### <<<=== wired log
    http.go:228: Request passed
```

None

```release-note
NONE
```

### DIFF
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -345,7 +345,7 @@ func CompareRequest(t *testing.T, req *roundtripper.Request, cReq *roundtripper.
 			if strings.ToLower(cRes.RedirectRequest.Scheme) == "https" && gotPort != "443" && gotPort != "" {
 				return fmt.Errorf("for https scheme, expected redirected port to be 443 or not set, got %q", gotPort)
 			}
-			if strings.ToLower(cRes.RedirectRequest.Scheme) != "http" || strings.ToLower(cRes.RedirectRequest.Scheme) == "https" {
+			if strings.ToLower(cRes.RedirectRequest.Scheme) != "http" || strings.ToLower(cRes.RedirectRequest.Scheme) != "https" {
 				t.Logf("Can't validate redirectPort for unrecognized scheme %v", cRes.RedirectRequest.Scheme)
 			}
 		} else if expected.RedirectRequest.Port != gotPort {
--- a/conformance/utils/http/http.go
+++ b/conformance/utils/http/http.go
@@ -345,7 +345,9 @@ func CompareRequest(t *testing.T, req *roundtripper.Request, cReq *roundtripper.
 			if strings.ToLower(cRes.RedirectRequest.Scheme) == "https" && gotPort != "443" && gotPort != "" {
 				return fmt.Errorf("for https scheme, expected redirected port to be 443 or not set, got %q", gotPort)
 			}
-			t.Logf("Can't validate redirectPort for unrecognized scheme %v", cRes.RedirectRequest.Scheme)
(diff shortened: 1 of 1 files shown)


LiorLieberman (commented):
Thanks @shawnh2 .
I am not sure if this change actually conforms with the definition here we introduced here https://github.com/kubernetes-sigs/gateway-api/pull/1880

Though looks like the wording has changed since then

/cc @gauravkghildiyal
/cc @arkodg

conformance/utils/http/http.go
```diff
@@ -345,7 +345,9 @@ func CompareRequest(t *testing.T, req *roundtripper.Request, cReq *roundtripper.
 			if strings.ToLower(cRes.RedirectRequest.Scheme) == "https" && gotPort != "443" && gotPort != "" {
 				return fmt.Errorf("for https scheme, expected redirected port to be 443 or not set, got %q", gotPort)
 			}
-			t.Logf("Can't validate redirectPort for unrecognized scheme %v", cRes.RedirectRequest.Scheme)
+			if strings.ToLower(cRes.RedirectRequest.Scheme) != "http" || strings.ToLower(cRes.RedirectRequest.Scheme) != "https" {
```
arkodg, line 348: shouldn't this be `&&` instead ?

arkodg:
reading the API
```
If no port is specified, the redirect port MUST be derived using the following rules:

If redirect scheme is not-empty, the redirect port MUST be the well-known port associated with the redirect scheme. Specifically “http” to port 80 and “https” to port 443. If the redirect scheme does not have a well-known port, the listener port of the Gateway SHOULD be used.
If redirect scheme is empty, the redirect port MUST be the Gateway Listener port.
```

https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter

can we just delete the existing log line since its not useful imo

shawnh2:
> reading the API
> 
> ```
> If no port is specified, the redirect port MUST be derived using the following rules:
> 
> If redirect scheme is not-empty, the redirect port MUST be the well-known port associated with the redirect scheme. Specifically “http” to port 80 and “https” to port 443. If the redirect scheme does not have a well-known port, the listener port of the Gateway SHOULD be used.
> If redirect scheme is empty, the redirect port MUST be the Gateway Listener port.
> ```
> 
> https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRequestRedirectFilter
> 
> can we just delete the existing log line since its not useful imo

i think we should keep this log, if the scheme of a redirect request is not well-known, we may able to recognize it through the log.

gauravkghildiyal:
+1 to still keeping the log line with the modified change

Yes that log line was not logically at the correct place. This change should fix it. Thanks @shawnh2

Issue #2147, opened 2023-06-26
Update TargetREf in Policy GEP


candita:
Discussion generated from GEP-1897 points out that TargetRef could use some updates.  
1. All policy resources must include TargetRef with the fields specified here: https://github.com/kubernetes-sigs/gateway-api/blob/a33a934af9ec6997b34fd9b00d2ecd13d143e48b/apis/v1alpha2/policy_types.go#L24-L41
2. Policy resources _may_ also choose to include `SectionName` and/or `Port` in TargetRef following the same mechanics of `ParentRef`

Then we can provide additional `TargetRef` types that extend the base `TargetRef` with the different possible combinations of these fields. That feels like it could be handled in a separate smaller update to the policy GEP.

_Originally posted by @robscott in https://github.com/kubernetes-sigs/gateway-api/pull/2113#discussion_r1239035806_

arkodg:
happy to pick this one up, @robscott does that warrant a discussion first, or can I start with a GEP ?

youngnick:
I think that #2128 needs to merge before we make any more Policy changes. After that, yes, this should update GEP-713.

shaneutt:
> I think that https://github.com/kubernetes-sigs/gateway-api/pull/2128 needs to merge before we make any more Policy changes. After that, yes, this should update GEP-713.

+1, let's get #2128 done and then update the existing doc.

arkodg:
looks like @zhaohuabing is driving this, thanks !

Issue #995, opened 2022-01-14
GEP: Add Name to HTTPRouteRule and HTTPRouteMatch


hzxuzhonghu:
**What would you like to be added**:

Add a name field to HTTPRouteRule and HTTPRouteMatch structs to allow users to identify different routes 

**Why this is needed**:

- Policy Attachment

- Route Attachment/Delegation: Ability to have more fine grained matches for policy references IF we add sectionName to the policy object

- Improved Debugging: Ability to add debugging information, such as adding the route matched to some access log

shaneutt:
If I'm understanding correctly based on the closed PR, we've decided not to move forward with this at this time. If that's not the case, please feel free to drop a comment and we can re-open.

robscott:
I think we decided this would be reasonable if we had a bigger use case associated with it. This will likely end up being included as part of route delegation in #1058.

shaneutt:
Ok. The project state also says "abandoned" so we should probably update this too so this issue doesn't look as though it is in fact abandoned.

shaneutt:
Ok, then let's consider this closed in favor of the implementation in https://github.com/kubernetes-sigs/gateway-api/issues/1058 :+1:

youngnick:
I think it makes sense in general, but will need careful writing of the spec - we can't add the new field as required, but what value should  be used in contexts where it's expected (like log messages or telemetry) if it's not set?

howardjohn:
`""`, `"-"`, `"unknown"`, `nil` - not sure. It may be usage specific, some systems have different ways to describe "not present" I would think, which makes it hard to prescribe against them

swathinsankaran:
@youngnick @howardjohn We would really appreciate having this in the next version of the HTTPRoute. We have a similar requirement to apply some policies per rule as mentioned in the description.

cc: @apalsule

maleck13:
Within the Kuadrant project, we also see value in adding this to support policy. Would really like to see it in the next version also

robscott:
One of the most common arguments against this feature is that we really shouldn't be encouraging users to have large Routes. Routes with lots of rules are remarkably hard to deal with in cases like partially invalid config where many Route rules are valid, but one is not. We still have not solved this problem so the API only provides guidance for when a Route is invalid.

[...]

PR #2283, opened 2023-08-10
Adding SectionName to PolicyTargetReference

zhaohuabing:
This allows attaching policies to a specific section within the targeted object.

Related issue: https://github.com/kubernetes-sigs/gateway-api/issues/2147

**What type of PR is this?**
/kind feature

What this PR does:

This PR allows attaching policies to a specific section within the targeted object.

Why we need it:

Lacking SectionName in PolicyTargetReference is blocking my current work. I'm implementing a direct policy attachment and need to target a specific listener inside a gateway. 

```release-note
NONE
```


robscott:
Hey @zhaohuabing, thanks for working on this! If I'm understanding correctly, it looks like you're trying to take on a couple of separate issues:

1. Adding SectionName to PolicyTargetReference (https://github.com/kubernetes-sigs/gateway-api/issues/2147)
2. Adding Name to Route Rules (https://github.com/kubernetes-sigs/gateway-api/issues/995)

I think we mostly have consensus on 1, but we're waiting for #2128 to merge before moving forward. On 2, I don't think we quite have enough consensus to move forward yet. Maybe once #2128 merges you can check in with @arkodg to see if he needs any help with the implementation of #2147.

It may also be helpful to take a look at our [documentation for the GEP process](https://gateway-api.sigs.k8s.io/geps/overview/). All API changes have to go through that process so we can't start directly with a PR to change the API itself without first having an approved GEP in an "implementable" state.

zhaohuabing:
@robscott What I really need is to have a `SectionName` field in the `PolicyTargetReference API`. Right now, its absence is causing a bit of a roadblock for me as I'm trying to set up a direct policy attachment and want to target a specific listener within a gateway.  

Since the community have already reached agreement on 1, could we consider adding this field as a first step? This matter is quite time-sensitive for me, and I'm a bit worried that it will take a while to get https://github.com/kubernetes-sigs/gateway-api/pull/2128 merged. Moreover, PR 2128 is mainly about the discoverability and doesn't related to the `SectionName` field at all, so it shouldn't block adding this field.

arkodg:
hey @zhaohuabing can you also edit https://github.com/kubernetes-sigs/gateway-api/blob/4651e916c370af99c87312de647adb00fe176b2a/geps/gep-713.md?plain=1#L685 to now account for `SectionName` . Imo, since this PR is in draft mode, it shouldn't distract others from reviewing the Policy Discoverability PR first

zhaohuabing:
Change to "ready to review". Since https://github.com/kubernetes-sigs/gateway-api/pull/2128 has already been merged, we can move on with this.

apis/v1alpha2/policy_types.go
```diff
@@ -38,6 +38,16 @@ type PolicyTargetReference struct {
 	//
 	// +optional
 	Namespace *Namespace `json:"namespace,omitempty"`
+
+	// SectionName is the name of a section within the target resource. When
+	// unspecified, this targets the entire resource. In the following
+	// resources, SectionName is interpreted as the following:
+	// * Gateway: Listener Name
+	// * Route: Rule Name
```
arkodg, line 46: Route Rule Name doesnt exist in the spec yet, so should be added only after it has been formally added into the spec

apis/v1alpha2/policy_types.go
```diff
@@ -38,6 +38,15 @@ type PolicyTargetReference struct {
 	//
 	// +optional
 	Namespace *Namespace `json:"namespace,omitempty"`
+
+	// SectionName is the name of a section within the target resource. When
+	// unspecified, this targets the entire resource. In the following
```
youngnick, line 43:
```suggestion
	// unspecified, this targetRef targets the entire resource. In the following
```
What "this" refers to is a little unclear, better to be specific.

geps/gep-713.md
```diff
@@ -1244,91 +1244,35 @@ level. The implementations that support this policy attachment model will have
...
-    // * Service: Port Name
-    //
-    // +kubebuilder:validation:MinLength=1
-    // +kubebuilder:validation:MaxLength=253
-    // +optional
-    SectionName string `json:"sectionName,omitempty"`
-    // ...
-}
-```
+The `sectionName` field could also be used to target a specific section of other resources, for example:
```
youngnick, line 1269:
```suggestion
The `sectionName` field can also be used to target a specific section of other resources, for example:
```
"Could" is a holdover from the previous language and is not as definite as "can".

geps/gep-713.md
```diff
@@ -1244,91 +1244,35 @@ level. The implementations that support this policy attachment model will have
...
-    // Support: Core
-    //
-    // +kubebuilder:validation:MinLength=1
-    // +kubebuilder:validation:MaxLength=253
-    // +optional
-    Name string `json:"name,omitempty"`
-    // ...
-}
-```
+* xRoute.Rules.Name
```
youngnick, line 1271: I'd move this one to below `Service.Ports`, since Service is already defined.

youngnick (commented):
This is looking great, thanks @zhaohuabing. Just a few small things to change.

geps/gep-713.md
zhaohuabing: Thanks! As a non-native speaker, I'm still trying to improve my English writing :-).

apis/v1alpha2/policy_types.go
```diff
@@ -38,6 +38,15 @@ type PolicyTargetReference struct {
 	//
 	// +optional
 	Namespace *Namespace `json:"namespace,omitempty"`
+
+	// SectionName is the name of a section within the target resource. When
+	// unspecified, this targetRef targets the entire resource. In the following
+	// resources, SectionName is interpreted as the following:
+	// * Gateway: Listener Name
+	// * Service: Port Name
+	//
```
youngnick, line 47: On further review, I think we need to be specific here about what happens if this is not found.

Would you be able to add some wording like:
```go
	// 
	// If a SectionName is specified, but does not exist on the targeted object,
	// the Policy must fail to attach, and the policy implementation should record
	// a `ResolvedRefs` or similar Condition in the Policy's status.
```

(We should specify something similar at the top level as well, but that's another issue).

geps/gep-713.md
```diff
@@ -1244,91 +1244,35 @@ level. The implementations that support this policy attachment model will have
...
-    // +optional
-    Name string `json:"name,omitempty"`
-    // ...
-}
-```
+* Service.Ports.Name
+* xRoute.Rules.Name
 
+This would require adding a `Name` field to those sub-resources that currently lack a name. For example,
+a `Name` field could be added to the `RouteRule` object:
```
youngnick, line 1275: If you could add a section summarizing what happens if the SectionName doesn't exist here as well please, that would be great.

youngnick:
Yep, this LGTM now. I'll approve and hold for other review though.

/approve
/hold

shaneutt (approved):
I think this looks helpful, but I think given the importance of policy attachment stuff with GA looming Rob should take a final look here as well.

/cc @robscott

robscott:
Thanks for working on this @zhaohuabing! Our policy attachment model is already quite complex so I want to ensure we've really thought out any potential addition before moving forward. Some questions that I'd like to see answered in the GEP:

1. How does this interact with inherited policy attachment? [Our chart here is already pretty complicated](https://gateway-api.sigs.k8s.io/geps/gep-713/#hierarchy), and as far as I can tell, this would nearly double the size of it by adding new "section" levels for nearly every level.
2. What are the specific use cases we're adding this for? I know Gateway -> Backend TLS may benefit from this, is there anything else?
3. In the case of Gateway -> Backend TLS it's possible that `port` would be a more natural extension than `sectionName`, adding both could result in an absolute explosion of complexity for policy attachment, especially "inherited" policy attachment.
4. The current PR adds `sectionName` to the existing PolicyTargetRef type, should we instead add a separate type so some implementations can choose to create Policy resources that intentionally do not support SectionName?
5. If this feature ends up getting paired with a name on Route rules, when would an implementation use a custom filter (ExtensionRef) instead of a Policy? (I'm assuming this is one of the long term goals here, but let me know if I'm wrong).
6. Is there anything we can do to reduce the added complexity here? We've relied pretty heavily on the idea that tooling and systems could generically visualize policy attachment (x-ref https://github.com/michaelvl/gateway-api-lens and https://github.com/gauravkghildiyal/gwctl), how would those integrate with the idea of SectionName? Is there a way for implementations to document the kinds of TargetRefs they support for a resource, and more specifically, where they support SectionName and what it means?

If any of these questions are too complicated, we can split up the work to answer them and merge some updates to the GEP with outstanding questions marked as unresolved, I just want to be very careful when adding complexity to this model.

arkodg:
hey @robscott, taking a stab at these questions

> 1.How does this interact with inherited policy attachment? [Our chart here is already pretty complicated](https://gateway-api.sigs.k8s.io/geps/gep-713/#hierarchy), and as far as I can tell, this would nearly double the size of it by adding new "section" levels for nearly every level.

* There is already language in the GEP around this
```
For Policies that affect an object's existing fields, multiple instances of the same Policy Kind affecting an object's fields will be evaluated as though only a single Policy "wins" the right to affect each field. This operation is performed on a per-distinct-field basis.
```

* Even without this change, the GEP could be expanded with examples on how multiple Inherited PolicyAttachment resources (lets say A, B, C ) affect outcome when applied to the same target resource (lets say Gateway G)

* For now, we can also add language in this PR, that says - `current recommendation is for Direct Policy Attachment types only` to limit scope and enhance in a future PR

> 2. What are the specific use cases we're adding this for? I know Gateway -> Backend TLS may benefit from this, is there anything else?

* Here's some of the many features that can be applied per Gateway Listener tied mostly to Downstream clients
* Authentication (JWT etc) ## used in the GEP as an example
* RateLimiting
* CORS
* Proxy Protocol

There are so many more that are relevant per Route Rule (tied to upstream) like Load Balancing, Retries, Timeouts, OutlierDetection, CircuitBreaker, TLS that can only be fine tuned if we add a `Name` to a route rule which doesnt exist today so the policy applied here will be coarse grained at the Route level.

> 3. In the case of Gateway -> Backend TLS it's possible that port would be a more natural extension than sectionName, adding both could result in an absolute explosion of complexity for policy attachment, especially "inherited" policy attachment.

* This PR follows the same design pattern as the [SectionName](https://github.com/kubernetes-sigs/gateway-api/blob/c3ceacfe8c302682b1f5b13a73c084cefb24926f/apis/v1beta1/shared_types.go#L117) in xRoute. This could be changed by maybe using a `Type` field making it look like
```
section:
  type: "Port"
  port: 8080
```
or
```
section:
  type: "Listener"
  listener: foo
```
but imo then similar treatment should also be done at the xRoute level 

> 4. The current PR adds sectionName to the existing PolicyTargetRef type, should we instead add a separate type so some implementations can choose to create Policy resources that intentionally do not support SectionName?

* sure this works, but for the case where an implementation doesnt use it, it could either ignore the field or use the `Status` to highlight the fact that it has ignored the field

> 5. If this feature ends up getting paired with a name on Route rules, when would an implementation use a custom filter (ExtensionRef) instead of a Policy? (I'm assuming this is one of the long term goals here, but let me know if I'm wrong).

* Here's how we think of policy vs ext filter in Envoy Gateway, and these opinions may vary across implementations
  * if the same resource can be applied to Gateway and xRoute, Policy makes sense here
  * if the resource only applies to xRoute and is authored by app team, ext filter makes sense here
  * if the resource needs to be overlayed/enforced on top of a xRoute, by another author / team, Policy makes sense here (transparent to the app team)

> 6. Is there anything we can do to reduce the added complexity here? We've relied pretty heavily on the idea that tooling and systems could generically visualize policy attachment (x-ref https://github.com/michaelvl/gateway-api-lens and https://github.com/gauravkghildiyal/gwctl), how would those integrate with the idea of SectionName? Is there a way for implementations to document the kinds of TargetRefs they support for a resource, and more specifically, where they support SectionName and what it means?

both these tools will be useful in helping a user visualize, which policy affects which resource/section/field.
In Envoy Gateway, we rely on the Policy Status to also provide more info to the user

please let us know which questions you'd like to see explicit verbiage, in the GEP, thanks

youngnick:
The more I think about @robscott's feedback here, the more I like it. I propose we do the following changes:
* We mandate that section names can only be used for Direct Policy, not Inherited. (The interactions are just too complicated if you allow sectionName + inheritance).
* We create a `DirectTargetRef` struct, that inlines the existing TargetRef struct, and adds `SectionName`.
* We update GEP-713 to with details about SectionName, and how it only works with Direct Attached Policy.

Note that the nice part here is that if we find we need it later, we can walk this back. But this allows us to have sub-object policy targeting for now, while keeping the scope and implications manageable.

I think that adding a `type` field is too far - I think that every field where we might want to do subsection binding already has a name field (Service ports have `name` already, for example).

youngnick:
Back to @robscott for another review, I think.

geps/gep-713.md
```diff
@@ -1244,91 +1244,35 @@ level. The implementations that support this policy attachment model will have
...
   name: foo
 spec:
-  maxRetries: 5
+  provider:
+    issuer: "https://accounts.google.com"
   targetRef:
     name: foo
     group: gateway.networking.k8s.io
-    kind: HTTPRoute
+    kind: Gateway
```
sanjaypujare, line 1265: Just curious why the example was changed from an HTTPRoute to Gateway (and from RetryPolicy to AuthenticationPolicy). Whatever is the reason, could we have both examples? Specially because you also have an example of adding a name to RouteRule (on line 1277)

zhaohuabing: @sanjaypujare The first example is changed to Gateway because the current use case I want to address is the Gateway Listener, and we already has the name field in the Listener object.  The second example is used to demonstrate what we need to change(adding the name field) to apply policy to RouteRule as well.

sanjaypujare: Thanks. With this change you included the RetryPolicy but not the HTTPRoute in the original doc which has the named section. From https://gateway-api.sigs.k8s.io/geps/gep-713/#apply-policies-to-sections-of-a-resource-future-extension I have the snippet as follows:

```
kind: HTTPRoute
metadata:
  name: http-app-1
  labels:
    app: foo
spec:
  hostnames:
  - "foo.com"
  rules:
  - name: bar
    matches:
    - path:
        type: Prefix
        value: /bar
    forwardTo:
    - serviceName: my-service1
      port: 8080
---
```

This is just above the RetryPolicy so it will be good to include (or rather restore) the above snippet before line 1277.

sanjaypujare (approved):
LGTM

sanjaypujare (approved):

apis/v1alpha2/policy_types.go
```diff
@@ -40,6 +40,31 @@ type PolicyTargetReference struct {
 	Namespace *Namespace `json:"namespace,omitempty"`
 }
 
+// DirectPolicyTargetReference identifies an API object to apply direct policy to. This
+// should be used as part of Policy resources that can target Gateway API
+// resources. For more information on how this policy attachment model works,
```
youngnick, line 45: Small nit, since it's not only Gateway API resources you can target.
```suggestion
// should be used as part of Policy resources that can target single
// resources. For more information on how this policy attachment model works,
```

geps/gep-713.md
```diff
@@ -1345,6 +1334,13 @@ type RouteRule struct {
 }
 ```
 
+If a `sectionName` is specified, but does not exist on the targeted object, the Policy must fail to attach,
+and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
+
+Note that the `sectionName` is currently intended to be used only for Direct Policy Attachment. Inherited Policies
+are always applied to the entire object. The `DirectPolicyTargetReference` API can be used to apply a direct Policy
+to a section of an object.
```
youngnick, line 1342: This is great, but we will probably also need to update the parts about Direct Attachment further up in this document with the details that you should use the DirectTargetRef instead of the more generic one.

(Non-blocking, just a note for myself for the next time I touch this GEP).

youngnick (commented):
This LGTM, I have no more objections.

arkodg:
a non specific vs specific `targetRef` applied to the same resource for a Direct Policy Attachment, does introduce the notion of overrides e.g. if `PolicyA` is applied to 
```
 targetRef:
     name: foo-gateway
     kind: Gateway
```

and if `PolicyB` is applied to 

```
 targetRef:
     name: foo-gateway
     kind: Gateway
     name: some-listener
```

what is the final setting for `foo-gateway.some-listener` ?
* implementation specific. # safest answer to add in this GEP
* most specific target wins

sanjaypujare:
>...
>
> * all settings from policy with most specific target wins

I vote for that.

zhaohuabing:
> * all settings from policy with most specific target wins

I prefer this one because it's simple and clear.

> * some settings from policy with most specific target are used as overrides and least specific policy is used as default for the rest of the unset settings

This one seems right but it would be very hard for users to figure out which settings are from which policy.

youngnick:
Yes, we use "most specific wins" in other places, so I am very much in favor.

@zhaohuabing, sorry, but you'll need to include some information in the SectionName part of GEP-713 about this behavior. Something like:

When multiple Policies of the same type target the same object, one with a SectionName specified, and one without, the one with a SectionName is more specific, and so will have all its settings apply. The less-specifc Policy will not attach to the target.

I'll also take a note to add or extend conflict-resolution rules for direct policy attachment further up, with the "more specific wins" rule. (I have more updates to do to GEP-713 already, this will add to the pile).

youngnick:
Thanks for that last change @zhaohuabing, this LGTM.

/lgtm

youngnick:
Ping @robscott for one more check, then we can remove the hold.

youngnick:
@zhaohuabing, you'll need to rerun `make generate` and commit the result for CI to pass.

robscott (commented):
Thanks for all the work on this @zhaohuabing! As much as I'm hesitant to add additional complexity here, this does feel inevitable. Have a few nits, but this looks good overall. In general just trying to at least make sure we have some guidance in here that ensures this is only used when it's actually necessary.

Would also appreciate some feedback from @kflynn on this because they've raised some important concerns about the complexity of our existing policy attachment model, so want to be careful when adding more.

apis/v1alpha2/policy_types.go
```diff
@@ -40,6 +40,30 @@ type PolicyTargetReference struct {
...
 }
 
+// DirectPolicyTargetReference identifies an API object to apply direct policy to.
+// This should be used as part of Policy resources that can target single resources.
+// For more information on how this policy attachment model works, and a sample
+// Policy resource, refer to the policy attachment documentation for Gateway API.
+//
+// Note: This should be used for direct policy attachment only.
+// <gateway:experimental>
+type DirectPolicyTargetReference struct {
```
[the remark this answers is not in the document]
zhaohuabing: I added comment below and hope it clarify the use of `DirectPolicyTargetReference`:

```
// Note: This should only be used for direct policy attachment when references
// to SectionName are actually needed. In all other cases, PolicyTargetReference
// should be used.
```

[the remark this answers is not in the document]
youngnick: I love the comment, but I agree that the naming doesn't help with possible confusion here. I think we should go with clarity over brevity here and call the thing `PolicyTargetReferenceWithSectionName`. Pretty long, but does exactly what it says.

robscott:
Thanks @zhaohuabing!

/lgtm
/hold cancel

Issue #2289, opened 2023-08-11
`cSuite.Report()` should support write report by checking if flags.ReportOutput is empty


Xunzhuo:
**What would you like to be added**:

We have experimental flag  `ReportOutput` to locate where the report is written to, but this flag is not used in suite, and if end-user tries to use experimental test suite, they need to write func like `writeReport` manually.

```go
func testExperimentalConformance(t *testing.T) {
	t.Logf("Running experimental conformance tests with %s GatewayClass\n cleanup: %t\n debug: %t\n enable all features: %t \n supported features: [%v]\n exempt features: [%v]",
		*flags.GatewayClassName, *flags.CleanupBaseResources, *flags.ShowDebug, *flags.EnableAllSupportedFeatures, *flags.SupportedFeatures, *flags.ExemptFeatures)

	cSuite, err := suite.NewExperimentalConformanceTestSuite(
		suite.ExperimentalConformanceOptions{
			Options: suite.Options{
				Client:     mgrClient,
				RestConfig: cfg,
				// This clientset is needed in addition to the client only because
				// controller-runtime client doesn't support non CRUD sub-resources yet (https://github.com/kubernetes-sigs/controller-runtime/issues/452).
				Clientset:                  k8sClientset,
				GatewayClassName:           *flags.GatewayClassName,
				Debug:                      *flags.ShowDebug,
				CleanupBaseResources:       *flags.CleanupBaseResources,
				SupportedFeatures:          supportedFeatures,
				ExemptFeatures:             exemptFeatures,
				EnableAllSupportedFeatures: *flags.EnableAllSupportedFeatures,
				NamespaceLabels:            namespaceLabels,
				SkipTests:                  skipTests,
			},
			Implementation:      *implementation,
			ConformanceProfiles: conformanceProfiles,
		})
	if err != nil {
		t.Fatalf("error creating experimental conformance test suite: %v", err)
	}

	cSuite.Setup(t)
	cSuite.Run(t, tests.ConformanceTests)
	report, err := cSuite.Report()
	if err != nil {
		t.Fatalf("error generating conformance profile report: %v", err)
	}
	writeReport(t.Logf, *report, *flags.ReportOutput)
}

func writeReport(logf func(string, ...any), report confv1a1.ConformanceReport, output string) error {
	rawReport, err := yaml.Marshal(report)
	if err != nil {
		return err
	}

	if output != "" {
		if err = os.WriteFile(output, rawReport, 0644); err != nil {
[...]

PR #2290, opened 2023-08-11
feat: write/log report in suite

Xunzhuo:
**What type of PR is this?**

/area conformance

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2289

```release-note
support automatically dump report in test suite
```

### DIFF
--- a/conformance/experimental_conformance_test.go
+++ b/conformance/experimental_conformance_test.go
@@ -20,7 +20,6 @@ limitations under the License.
 package conformance_test
 
 import (
-	"os"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -28,7 +27,6 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	"sigs.k8s.io/yaml"
 
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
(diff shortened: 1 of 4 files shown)


sunjayBhatia (approved):
Looks like it works as advertised 👍🏽, was able to get a report YAML out using the golang suite

conformance/experimental_conformance_test.go
```diff
@@ -119,32 +117,16 @@ func testExperimentalConformance(t *testing.T) {
...
-	}
-
-	if output != "" {
-		if err = os.WriteFile(output, rawReport, 0644); err != nil {
-			return err
-		}
-	}
-	logf("Conformance report:\n%s", string(rawReport))
-
-	return nil
```
youngnick, line 149: Why move all of this? You could also achieve this by passing `cSuite.ReportOutput` to the `writeReport` call on line 133, which would keep the output logic in the containing block, keeping the `Report` call more simple and closer to single-purpose.

conformance/utils/suite/suite.go
```diff
@@ -89,7 +89,7 @@ func New(s Options) *ConformanceTestSuite {
 	}
 
 	switch {
-	case s.EnableAllSupportedFeatures == true:
```
youngnick, line 92: Seems likely to be an IDE edit? It's fine in this case, but in general it's better to put this sort of change in a separate PR.

youngnick (changes requested):
I think that it's better to leave `writeReports` in place, and just change what we pass into it, rather than moving it inside the `Report` method.

conformance/utils/suite/suite.go
Xunzhuo: Yep: https://github.com/kubernetes-sigs/gateway-api/pull/2346

conformance/experimental_conformance_test.go
Xunzhuo: My idea is that, when gwapi implementations want to get the Report and use the experiment test suite, the flag `cSuite.Report` is not used inside the `experiment test suite` itself, we need to rewrite another writeReport to achieve this.

PR #2302, opened 2023-08-14
Add -buildvcs=false flag to fix golint script

### LINKED ISSUE #2071: Failed verifying golint while executing make
**What happened**:
Only verifying golint failed while executing `make`.
The output:
```bash
level=warning msg="[runner] Can't run linter goanalysis_metalinter: musttag: running `go list all`: exit status 1"
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: musttag: running `go list all`: exit status 1\n\n"
Test FAILED: hack/../hack/verify-golint.sh
```

It should work, not sure if there are any prerequisites I missed.

z1cheng:
**What type of PR is this?**

/kind bug

Got such errors while running `hack/verify-golint.sh`:
```
error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
error obtaining VCS status: exit status 128
        Use -buildvcs=false to disable VCS stamping.
```
So add `-buildvcs=false` to fix the errors

Fixes #2071

```release-note
NONE
```

### DIFF
--- a/hack/verify-golint.sh
+++ b/hack/verify-golint.sh
@@ -26,6 +26,6 @@ cd "${KUBE_ROOT}"
 # See configuration file in ${KUBE_ROOT}/.golangci.yml.
 mkdir -p cache
 
-docker run --rm -v $(pwd)/cache:/cache -v $(pwd):/app --security-opt="label=disable" -e GOLANGCI_LINT_CACHE=/cache -w /app "golangci/golangci-lint:$VERSION" golangci-lint run
+docker run --rm -v $(pwd)/cache:/cache -v $(pwd):/app --security-opt="label=disable" -e GOLANGCI_LINT_CACHE=/cache -e GOFLAGS="-buildvcs=false"  -w /app "golangci/golangci-lint:$VERSION" golangci-lint run
 
 # ex: ts=2 sw=2 et filetype=sh


z1cheng:
@shaneutt Please take a look at the PR when you're available :)

/assign @shaneutt

shaneutt:
I feel like I might be missing something.

https://github.com/golang/go/issues/52338 seems to suggest that this is the default since `v1.18.1` and the image we're using `golangci/golangci-lint:v1.52.2` contains Go `1.20.2`, so shouldn't this change in the flag be a no-opt?

robscott:
Hey @z1cheng, thanks for the PR! Can you help explain a bit more why this flag is necessary per @shaneutt's question above?

z1cheng:
@shaneutt @robscott Sorry for my delayed response, from https://pkg.go.dev/cmd/go@go1.20.2, the default value is "auto":

>
        -buildvcs
                Whether to stamp binaries with version control information
                ("true", "false", or "auto"). By default ("auto"), version control
                information is stamped into a binary if the main package, the main module
                containing it, and the current directory are all in the same repository.
                Use -buildvcs=false to always omit version control information, or
                -buildvcs=true to error out if version control information is available but
                cannot be included due to a missing tool or ambiguous directory structure.

robscott:
Thanks for clarifying @z1cheng! This LGTM with that context.

/lgtm
/approve

Issue #2277, opened 2023-08-08
CEL Validation: HTTPRoute Case-Insensitive Header Name Uniqueness


robscott:
**What happened**:
Unfortunately CEL is not currently capable of validating the case-insensitive uniqueness of header names in header modifier filters. This is the only standard channel validation we are still reliant on the validating webhook for. @gauravkghildiyal already developed a CEL rule for this:

```go
// +kubebuilder:validation:XValidation:message="Must not match the same header (case-insensitive) multiple times in the same rule",rule="self.all(h1, self.exists_one(h2, h1.name.lowerAscii() == h2.name.lowerAscii()))"
```

Unfortunately there is a problem with CEL cost estimation that prevents this https://github.com/kubernetes/kubernetes/issues/119749, @jpbetz is working on a fix for this.

**Anything else we need to know?**:
Given the Gateway API versioning policy to support the 5 latest minor Kubernetes versions, even when this is fixed upstream, it's going to be a long time before we can adopt this.

/lifecycle frozen

youngnick:
This should definitely be frozen, as it is blocked by upstream.

/remove-lifecycle stale
/lifecycle frozen

rikatz:
As the issue on k/k is closed, should we move with the validation here?

@snorwin as you have recently worked on header validations

snorwin:
Unfortunately, the cost-estimation issue related to `lowerAscii` did not resolve the issue we are facing here. Due to the high level of nesting, even without `lowerAscii` the validation exceeds the budget by a factor of more than 100.

rikatz:
[...]

PR #2306, opened 2023-08-15
Using "Set" list type for "Remove" in Header Modifier Filters

robscott:
**What type of PR is this?**
/kind cleanup

This fills one more gap in our CRD validation that we've historically relied on the validating webhook to cover. As documented in https://github.com/kubernetes-sigs/gateway-api/issues/2277, the webhook ensures that header names have case-insensitive uniqueness. The built-in CRD validation uses `+listType=map` and `+listMapKey=name` to ensure that names are unique (unfortunately still case-sensitive, but better than nothing). We were missing equivalent CRD validation on the `Remove` field, this adds that. 

As we're moving away from webhook-based validation, adding as much validation we can directly to the CRD will be helpful, so this takes one more step in that direction.

```release-note
HTTPRoute and GRPCRoute CRDs now provide built-in validation that ensures the uniqueness of names in Header Modifier "Remove" lists.
```

### DIFF
--- a/apis/v1beta1/httproute_types.go
+++ b/apis/v1beta1/httproute_types.go
@@ -839,6 +839,7 @@ type HTTPHeaderFilter struct {
 	//   my-header2: bar
 	//
 	// +optional
+	// +listType=set
 	// +kubebuilder:validation:MaxItems=16
 	Remove []string `json:"remove,omitempty"`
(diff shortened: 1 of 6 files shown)


youngnick (approved):
/lgtm

PR #2310, opened 2023-08-15
[conformance] Specify group: "" for mesh conformance parentRefs

### LINKED ISSUE #2309: Mesh conformance tests should be careful with parentRef groups
In many cases, the mesh conformance tests are applying HTTPRoutes including, for example:

```
parentRefs:
  - kind: Service
    name: echo-v2
```

The astute reader will note the absence of a `group`. Per [the parentRefs specification]:

> When [`group` is] unspecified, "gateway.networking.k8s.io" is inferred.

This is problematic, because `gateway.networking.k8s.io` doesn't contain a Service. Meshes which correctly interpret `parentRefs` will fail these tests.

[the parentRefs specification]:https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.ParentReference

Run the conformance tests with the `HTTPResponseHeaderModification` feature enabled.

kflynn:
Signed-off-by: flynn@buoyant.io

**What type of PR is this?**

/kind bug
/kind test
/area conformance

Many mesh conformance tests specify `parentRefs` with `kind: Service`, but don't specify the `group`. That will default the `group` to `gateway.networking.k8s.io`, which won't work for `kind: Service`.

Fixes #2309

```release-note
NONE
```

### DIFF
--- a/conformance/tests/mesh-consumer-route.yaml
+++ b/conformance/tests/mesh-consumer-route.yaml
@@ -5,7 +5,8 @@ metadata:
   namespace: gateway-conformance-mesh-consumer
 spec:
   parentRefs:
-  - kind: Service
+  - group: ""
+    kind: Service
     name: echo-v1
     namespace: gateway-conformance-mesh
   rules:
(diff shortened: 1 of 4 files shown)


arkodg (approved):
/ok-to-test

arkodg (approved):
/approve

robscott:
Thanks @kflynn!

/lgtm

PR #2305, opened 2023-08-15
Add CEL validation to GRPCRoute

gnossen:
/kind feature

**What this PR does / why we need it**:
CEL validation enables validation of GRPCRoute CRs without a webhook.

**Does this PR introduce a user-facing change?**:

```release-note
Add CEL validation for GRPCRoute.
```

PR #2311, opened 2023-08-16
Add CEL validation for GRPCMethodMatch

gauravkghildiyal:
**What type of PR is this?**
/kind feature

Add CEL validation for GRPCMethodMatch

Fixes https://github.com/kubernetes-sigs/gateway-api/pull/2305#pullrequestreview-1577991344

```release-note
GRPCRoute: The default match has been removed as it was invalid (it only specified a type of "Exact" without a corresponding Service or Method). Note that the match type still defaults to "Exact".
```

### DIFF
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -147,7 +147,6 @@ type GRPCRouteSpec struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
-	// +kubebuilder:default={{matches: {{method: {type: "Exact"}}}}}
 	Rules []GRPCRouteRule `json:"rules,omitempty"`
 }
 
@@ -313,6 +312,10 @@ type GRPCRouteMatch struct {
 // request service and/or method.
 //
 // At least one of Service and Method MUST be a non-empty string.
+//
+// +kubebuilder:validation:XValidation:message="One or both of 'service' or 'method' must be specified",rule="has(self.type) ? has(self.service) || has(self.method) : true"
+// +kubebuilder:validation:XValidation:message="service must only contain valid characters (matching ^(?i)\\.?[a-z_][a-z_0-9]*(\\.[a-z_][a-z_0-9]*)*$)",rule="(!has(self.type) || self.type == 'Exact') && has(self.service) ? self.service.matches(r\"\"\"^(?i)\\.?[a-z_][a-z_0-9]*(\\.[a-z_][a-z_0-9]*)*$\"\"\"): true"
+// +kubebuilder:validation:XValidation:message="method must only contain valid characters (matching ^[A-Za-z_][A-Za-z_0-9]*$)",rule="(!has(self.type) || self.type == 'Exact') && has(self.method) ? self.method.matches(r\"\"\"^[A-Za-z_][A-Za-z_0-9]*$\"\"\"): true"
 type GRPCMethodMatch struct {
 	// Type specifies how to match against the service and/or method.
 	// Support: Core (Exact with service and method specified)
--- a/pkg/test/cel/grpcroute_test.go
+++ b/pkg/test/cel/grpcroute_test.go
@@ -22,9 +22,10 @@ package main
 import (
 	"context"
 	"fmt"
-	"testing"
 	"strings"
+	"testing"
 	"time"
+
 	gatewayv1a2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


apis/v1alpha2/grpcroute_types.go
```diff
@@ -147,7 +147,7 @@ type GRPCRouteSpec struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
-	// +kubebuilder:default={{matches: {{method: {type: "Exact"}}}}}
+	// TODO(gauravkghildiyal): Discuss
```
gauravkghildiyal, line 150: There seems to be some conflict between what we are validating (_"One or both of 'service' or 'method' must be specified"_) and what we are setting as the default (_"neither of the two"_)

I'm not sure if we can set a sensible default here. Does it mean we remove the defaulting? Or do we skip the validation?

gauravkghildiyal: CC: @gnossen

apis/v1alpha2/grpcroute_types.go
```diff
@@ -147,7 +147,6 @@ type GRPCRouteSpec struct {
 	//
 	// +optional
 	// +kubebuilder:validation:MaxItems=16
-	// +kubebuilder:default={{matches: {{method: {type: "Exact"}}}}}
```
robscott, line 150: I'm guessing this was accidental removal? Should revert this to avoid API change here.

pkg/test/cel/grpcroute_test.go
```diff
@@ -297,6 +298,118 @@ func TestGRPCRouteRule(t *testing.T) {
...
+		{
+			name: "invalid GRPCRoute missing service or method in GRPCMethodMatch field",
+			method: gatewayv1a2.GRPCMethodMatch{
+				Service: nil,
+				Method:  nil,
+			},
+			wantErrors: []string{"One or both of 'service' or 'method"},
+		},
+		{
+			name: "GRPCRoute use regex in service and method with undefined match type",
```
robscott, line 332: Nit: recommend replacing "use" with "uses" in most test case names below this point
```suggestion
			name: "GRPCRoute uses regex in service and method with undefined match type",
```

robscott (commented):
Thanks @gauravkghildiyal!

pkg/test/cel/grpcroute_test.go
gauravkghildiyal: Yes definitely. Done

pkg/test/cel/grpcroute_test.go
```diff
@@ -297,6 +298,118 @@ func TestGRPCRouteRule(t *testing.T) {
...
+
+	tests := []struct {
+		name       string
+		method     gatewayv1a2.GRPCMethodMatch
+		wantErrors []string
+	}{
+		{
+			name: "valid GRPCRoute with 1 service in GRPCMethodMatch field",
+			method: gatewayv1a2.GRPCMethodMatch{
+				Service: &service,
```
gnossen, line 314: Could we also add a case that verifies that a leading dot is valid within a Service? Just want to make sure that this doesn't get lost in a future change?

apis/v1alpha2/grpcroute_types.go
gauravkghildiyal: This is being discussed in https://github.com/kubernetes-sigs/gateway-api/pull/2311#discussion_r1295584514

gnossen: Looks like this is actually a reasonable change. The default _behavior_ is to "match all traffic." In a previous state of the GEP, this default did exactly that. However, an empty matcher is now disallowed. However, per the spec, an _unspecified_ GRPCRoute should match all traffic. While this might be a literal API change, it is not a behavioral change. I think this is the right thing to do.

gnossen (approved):
Thanks for the PR!

apis/v1alpha2/grpcroute_types.go
gauravkghildiyal: This was discussed in https://github.com/kubernetes-sigs/gateway-api/pull/2311#discussion_r1296286374

robscott: Chatted with @gnossen and @gauravkghildiyal on this and came to the conclusion that removing this validation is correct. The behavior we want is "if someone specifies a match, the default type of that match should be exact." That's covered by this line: https://github.com/kubernetes-sigs/gateway-api/blob/ce5bf58ac9bf9101a09a9f64a9959a0f84802d8e/apis/v1alpha2/grpcroute_types.go#L325

The line we want to remove is this one: https://github.com/kubernetes-sigs/gateway-api/blob/ce5bf58ac9bf9101a09a9f64a9959a0f84802d8e/apis/v1alpha2/grpcroute_types.go#L150 That creates a default match of type "Exact" with no other config and is therefore failing validation and also not what we actually want here. Given that this default was never actually valid and this API is still experimental, I think it's good to remove the validation here - we should just add a release note to the PR description to document the change.

pkg/test/cel/grpcroute_test.go
gauravkghildiyal: Added testcase `GRPCRoute uses a valid service with a leading dot when match type is Exact`.

robscott:
Thanks @gauravkghildiyal!

/lgtm
/approve

youngnick:
Updates #1016

PR #2312, opened 2023-08-16
Conformance tweaks for Linkerd and other meshes

kflynn:
/kind bug
/area conformance

This PR makes some small changes to allow conformance to work for Linkerd and, most likely, other meshes that don't also provide gateway controllers. Huge props to @adleong for doing most of the heavy lifting on this, weeks ago. :slightly_smiling_face:

Like most nontrivial PRs of mine, it is best reviewed commit by commit.

- Allow `-namespace-annotations` to annotate namespaces created by the tests

    We already had `-namespace-labels` to supply labels on test namespaces. Linkerd needs annotations rather than labels, so we add a new parallel flag to set annotations on test namespaces.

- Document `-namespace-annotations`

    Documentation is in a separate commit.

- comment fix

    This just triggered me so I had to fix it. :slightly_smiling_face:
    
- Split out `MeshNamespacesMustBeReady`, since mesh namespaces don't have Gateways

    If testing a service mesh that doesn't also provide a gateway controller, there won't be any Gateways in the namespace, so this pulls out a separate function to check that the namespace is ready without needing Gateways, and uses it for the namespaces used for mesh testing.

- `MeshFrontendHostname` requires `HTTPResponseHeaderModification`

    The `MeshFrontendHostname` test actually relies on a header modification, but wasn't tagged as such.

```release-note
Better support mesh-only conformance testing
```

### DIFF
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -182,7 +182,7 @@ func (suite *ConformanceTestSuite) Setup(t *testing.T) {
 			"gateway-conformance-app-backend",
 			"gateway-conformance-web-backend",
 		}
-		kubernetes.NamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, namespaces)
+		kubernetes.MeshNamespacesMustBeReady(t, suite.Client, suite.TimeoutConfig, namespaces)
 	}
 }
 
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -55,15 +55,16 @@ type ConformanceTestSuite struct {
 
 // Options can be used to initialize a ConformanceTestSuite.
 type Options struct {
-	Client           client.Client
-	Clientset        clientset.Interface
-	RestConfig       *rest.Config
-	GatewayClassName string
-	Debug            bool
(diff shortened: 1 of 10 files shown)


keithmattix (approved):
Looks good; will wait for a bit for john to have a look

conformance/utils/suite/suite.go
```diff
@@ -261,6 +263,22 @@ func ParseNamespaceLabels(f string) map[string]string {
 	return res
 }
 
+// ParseNamespaceAnnotations parses flag arguments and converts the string to
+// map[string]string containing annotation key/value pairs.
+func ParseNamespaceAnnotations(f string) map[string]string {
```
frankbu, line 268: Why not just rename `ParseNamespaceLabels` something like `ParseKeyValuePairs` and use it for both flags, instead of having two functions with the same impl?

kflynn: Being honest, I think I was expecting to have to edit it after the clone, and then never needed to. Great suggestion, I'll change that.

howardjohn (approved):

robscott:
Thanks @kflynn!

/lgtm
/approve

PR #2316, opened 2023-08-17
Update GEP-1742 to refer to GEP-2257

kflynn:
**What type of PR is this?**

/kind gep

The point of GEP-2257 was to nail down the syntax used for Durations, such as appear in GEP-1742, so... we should probably update GEP-1742 to actually make use of that. 😂

```release-note
NONE
```

### DIFF
--- a/geps/gep-1742.md
+++ b/geps/gep-1742.md
@@ -206,7 +206,7 @@ sequenceDiagram
     participant C as Client
     participant P as Proxy
     participant U as Upstream
-    
+
     C->>P: Connection Started
     activate U
     activate C
@@ -224,7 +224,7 @@ sequenceDiagram
     activate U
     note left of U: timeout queue<br/>(wait for available server)
     deactivate U
-		
+
     P->>U: Connection Started
     activate U
     P->>U: Starts sending Request
--- a/geps/gep-1742.md
+++ b/geps/gep-1742.md
@@ -368,9 +368,11 @@ sequenceDiagram
     U->>C: Connection ended
 ```
 
-Both timeout fields are string duration values as specified by
-[Golang time.ParseDuration](https://pkg.go.dev/time#ParseDuration) and MUST be >= 1ms
-or 0 to disable (no timeout).
+Both timeout fields are [GEP-2257 Duration] values. A zero-valued timeout
+("0s") MUST be interpreted as disabling the timeout; a non-zero-valued timeout
+MUST be >= 1ms.
+
+[GEP-2257 Duration]:/geps/gep-2257/
 
 ### GO
 


geps/gep-1742.md
```diff
@@ -368,9 +368,11 @@ sequenceDiagram
...
 ```
 
-Both timeout fields are string duration values as specified by
-[Golang time.ParseDuration](https://pkg.go.dev/time#ParseDuration) and MUST be >= 1ms
-or 0 to disable (no timeout).
+Both timeout fields are [GEP-2257 Duration] values. A zero-valued timeout
+("0s") MUST be interpreted as disabling the timeout; a non-zero-valued timeout
+MUST be >= 1ms.
+
+[GEP-2257 Duration]:/geps/gep-2257/
```
robscott, line 375: Nit: I think you also need to change the corresponding types definition in this GEP to move from `metav1.Duration` to something like `gatewayv1alpha2.Duration`.

robscott (commented):
Thanks @kflynn!

geps/gep-1742.md
frankbu: I think GEP-2257 should also include the new Duration Go type? Something like this:

```
// A duration is a ...
//
// +kubebuilder:validation:Pattern=`^([0-9]{1,5}(h|m|s|ms))`
type Duration string
```

Where does this type belong? `apis/v1alpha2/shared_types.go`?

We will update #2013 with these changes, as soon as they are added.

robscott: Yep, that's exactly where I'd put it.

frankbu:
@robscott are you going to close #2155 in favor of this one?

robscott:
@frankbu thanks for the reminder! Closed that one out.

geps/gep-1742.md
frankbu: Also, IIUC https://github.com/kubernetes-sigs/gateway-api/pull/2155#issuecomment-1640891533 we should also add CEL validation something like this:
```
// +kubebuilder:validation:XValidation:message="request timeout value must be greater than or equal to 1ms",rule="!(has(self.request) && duration(self.request) != duration('0s') && duration(self.request) < duration('1ms'))"
// +kubebuilder:validation:XValidation:message="backendRequest timeout value must be greater than or equal to 1ms",rule="!(has(self.backendRequest) && duration(self.backendRequest) != duration('0s') && duration(self.backendRequest) < duration('1ms'))"
// +kubebuilder:validation:XValidation:message="backendRequest timeout cannot be longer than request timeout",rule="!(has(self.request) && has(self.backendRequest) && duration(self.request) != duration('0s') && duration(self.backendRequest) > duration(self.request))"
```

robscott: Agree on the general concept of CEL validation, particularly for max values, but I thought that we'd actually agreed to have "0" represent an infinite timeout since it's so widely implemented?

kflynn: @robscott That's what I was trying to say by "disabling the timeout" above -- do you think that

> A zero-valued timeout (e.g. "0s") MUST be interpreted as meaning "never time out"...

would be more clear?

kflynn: @robscott @frankbu Thanks for the catch and the pointer to the API definition! I'll try to do that sometime tomorrow. 🤞

frankbu: @kflynn I think the wording re: 0s in the GEF is fine, maybe @robscott was questioning the wording I proposed for the CEL validation?

`// +kubebuilder:validation:XValidation:message="request timeout value must be greater than or equal to 1ms",rule="!(has(self.request) && duration(self.request) != duration('0s') && duration(self.request) < duration('1ms'))"`

Notice that even though the message I've proposed just says the timeout must be >= 1ms, the rule does check that the value is also not 0. My thinking is that 0s is an infinite duration, so just saying the timeout needs to be >= 1ms covers the 0 case too 😃

kflynn:
Closing in favor of #2367; thanks @frankbu!

PR #2318, opened 2023-08-17
I can't take it any more. Sort the GEP list.

kflynn:
**What type of PR is this?**

/kind cleanup

I can't handle the randomly-sorted GEPs in the sidebar any more. :joy: Let's keep them numerically sorted, please? :slightly_smiling_face:

-->
```release-note
NONE
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -92,34 +92,34 @@ nav:
         - geps/gep-1282.md
       - Provisional:
         - geps/gep-1619.md
-        - geps/gep-1897.md
         - geps/gep-1867.md
+        - geps/gep-1897.md
         - geps/gep-2162.md
       - Implementable:
         - geps/gep-1742.md
       - Experimental:
+        - geps/gep-713.md
+        - geps/gep-957.md
+        - geps/gep-1016.md
         - geps/gep-1324.md
         - geps/gep-1426.md
-        - geps/gep-1686.md
-        - geps/gep-1748.md
         - geps/gep-1651.md
-        - geps/gep-1016.md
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -93,32 +93,32 @@ nav:
...
-        - geps/gep-1323.md
-        - geps/gep-922.md
-        - geps/gep-917.md
-        - geps/gep-851.md
-        - geps/gep-820.md
-        - geps/gep-746.md
-        - geps/gep-726.md
-        - geps/gep-724.md
-        - geps/gep-718.md
         - geps/gep-709.md
```
robscott, line 111: One nitpick: This was intentionally done in descending order so the most recent GEPs would be at the top of the list. I don't have a strong opinion here, and clearly other parts of this nav have it in ascending order so 🤷.

robscott (commented):
Thanks @kflynn! Don't care too much whether this is ascending or descending order, but completely agree that this should be consistent.

/approve

robscott:
@kflynn do you have time to follow up on this one?

mkdocs.yml
kflynn: I (eventually) decided that I'd prefer having them all ascending. Not a strong opinion here either, really -- anyone else feel strongly?

kflynn:
Rebased! Sorry for the delay.

mkdocs.yml
youngnick: I think consistency is better here. It's not a big enough deal to worry about too much - we can always change it later.

PR #2323, opened 2023-08-18
Removing concept of resource-level support levels

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation
/area conformance

The concept of conformance levels has gotten rather confusing as we've added both new features and resources. Some resources (GRPCRoute and ReferenceGrant) had conformance levels defined while the rest did not. This does some refactoring to try to make our documentation and lists of conformance features more consistent. This gets us to a point where:

1. Resources don't have support levels, implementations can choose to support whichever set of resources they want
2. Claiming the Support_Resource_ feature when running conformance tests means that you support the resource and all the core features it contains

As a result, the most notable change is that StandardCoreFeatures has been split into two: GatewayCoreFeatures and ReferenceGrantCoreFeatures, this more clearly matches the rest of our feature set. Hopefully this makes sense, but definitely let me know if this is just more confusing.

```release-note
Conformance Profiles: Some sets of features have been renamed to more closely match the resource they correspond to.
```

/hold 
/cc @LiorLieberman @mlavacca @shaneutt

### DIFF
--- a/conformance/utils/suite/features.go
+++ b/conformance/utils/suite/features.go
@@ -27,31 +27,23 @@ import "k8s.io/apimachinery/pkg/util/sets"
 type SupportedFeature string
 
 // -----------------------------------------------------------------------------
-// Features - Standard (Core)
+// Features - Gateway Conformance (Core)
 // -----------------------------------------------------------------------------
 
 const (
-	// This option indicates support for ReferenceGrant (core conformance).
-	// Opting out of this requires an implementation to have clearly implemented
-	// and documented equivalent safeguards.
-	SupportReferenceGrant SupportedFeature = "ReferenceGrant"
-	// This option indicates support for Gateway (core conformance).
+	// This option indicates support for Gateway.
 	// Opting out of this is allowed only for GAMMA-only implementations
 	SupportGateway SupportedFeature = "Gateway"
 )
 
-// StandardCoreFeatures are the features that are required to be conformant with
-// the Core API features (e.g. GatewayClass, Gateway, e.t.c.).
-//
-// TODO: we need clarity for standard vs experimental features.
-// See: https://github.com/kubernetes-sigs/gateway-api/issues/1891
-var StandardCoreFeatures = sets.New(
-	SupportReferenceGrant,
+// GatewayCoreFeatures are the features that are required to be conformant with
+// the Gateway resource.
+var GatewayCoreFeatures = sets.New(
 	SupportGateway,
 )
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


shaneutt (approved):
I think it makes sense to remove support levels at the resource level.

conformance/utils/suite/features.go
```diff
@@ -189,9 +189,10 @@ var MeshCoreFeatures = sets.New(
 //
 // NOTE: as new feature sets are added they should be inserted into this set.
 var AllFeatures = sets.New[SupportedFeature]().
-	Insert(StandardExtendedFeatures.UnsortedList()...).
-	Insert(ExperimentalExtendedFeatures.UnsortedList()...).
+	Insert(GatewayExtendedFeatures.UnsortedList()...).
+	Insert(ReferenceGrantCoreFeatures.UnsortedList()...).
```
arkodg, line 193: shouldnt the same logic also be applied to `HTTPCoreFeatures` , `HTTPExtendedFeatures` & `TLSCoreFeatures` ?

robscott: I'm not sure I understand the request here? Each of those is added to the set in the three lines below.

robscott: I think I understand it now - why I'm actually trying to do is have per-resource lists, so `GatewayCoreFeatures` would mean the core features included in just the `Gateway` resource and `HTTPRouteCoreFeatures` would include the core features that are part of the `HTTPRoute` resource. Does that approach make sense to you @arkodg?

arkodg: yeah, maybe this is not tied to this PR, but what I was trying to get at is `HTTPCore` is not needed in this list since it should be part of `GatewayCore`

robscott: But `GatewayCore` would only be core features that are part of the Gateway resource. I think we also want to separate the features of HTTPRoute from Gateway or Mesh use cases so implementations can claim support for HTTPRoute without Gateway. Maybe it would be clearer if we called the lists `HTTPRouteCoreFeatures` and `GatewayCoreFeatures`?

arkodg: yes :) until now, I was assuming `Gateway` is tied to `Ingress` features (for naming in this package/file)

robscott: Good call, updated, thanks for calling this out!

arkodg (changes requested):
thanks for adding the comments
/lgtm

arkodg (approved):

Issue #2242, opened 2023-07-27
Conformance Profiles: Remove experimental build tag


shaneutt:
Currently we require the `experimental` Golang build tag to be provided to use the CLI or Go library for the conformance profiles version of the conformance test suite.

We've [received feedback](https://github.com/kubernetes-sigs/gateway-api/pull/2180#discussion_r1276146601) that this is seemingly problematic for some community members, and they are in favor of dropping it.

As such, the purpose of this task is to drop that build tag and make the profiles enabled test suite available in a normal build.

PR #2327, opened 2023-08-20
conformance-profile: Remove experimental build tag

sayboras:
**What type of PR is this?**

/kind test
/area conformance

Relates: #2242

```release-note
NONE
```

Testing was done as part of https://github.com/cilium/cilium/pull/27596

### DIFF
--- a/conformance/apis/v1alpha1/doc.go
+++ b/conformance/apis/v1alpha1/doc.go
@@ -1,6 +1,3 @@
-//go:build experimental
-// +build experimental
-
 /*
 Copyright 2023 The Kubernetes Authors.
 
@@ -30,7 +27,7 @@ limitations under the License.
 // features, then you should not use them. If you would like to opt into these
 // unreleased features use Go build tags to enable them, e.g.:
 //
-//   $ GOFLAGS='-tags=experimental' go test ./conformance/... -args ${CONFORMANCE_ARGS}
+//   $ go test ./conformance/... -args ${CONFORMANCE_ARGS}
 //
 // Please note that everything here is considered experimental and subject to
 // change. Expect breaking changes and/or complete removals if you start using
--- a/conformance/apis/v1alpha1/conformancereport.go
+++ b/conformance/apis/v1alpha1/conformancereport.go
@@ -1,6 +1,3 @@
-//go:build experimental
-// +build experimental
-
 /*
 Copyright 2023 The Kubernetes Authors.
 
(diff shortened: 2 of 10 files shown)


conformance/apis/v1alpha1/doc.go
```diff
@@ -23,17 +20,10 @@ limitations under the License.
...
-// features, then you should not use them. If you would like to opt into these
-// unreleased features use Go build tags to enable them, e.g.:
-//
-//   $ GOFLAGS='-tags=experimental' go test ./conformance/... -args ${CONFORMANCE_ARGS}
+//   $ go test ./conformance/... -args ${CONFORMANCE_ARGS}
 //
-// Please note that everything here is considered experimental and subject to
-// change. Expect breaking changes and/or complete removals if you start using
-// them.
+// While the experiment build tag is removed, please note that everything here
```
mlavacca, line 25: I wonder if we really need to mention the experimental build tag here. From a user perspective, this preamble can be meaningless, as such a tag does not exist anymore.

mlavacca (commented):
Thanks for this PR, @sayboras!

conformance/apis/v1alpha1/doc.go
arkodg: +1, vote for keeping as is

kflynn: I'd probably also not mention the experimental build tag here.

kflynn (changes requested):
...hmmmm. When I tried a conformance run from the CLI with the `experimental` build tag enabled, it tried to run `TestConformance` and then `TestExperimentalConformance`, and `TestExperimentalConformance` failed because it was expecting an empty cluster to play in. 

This PR leaves both functions intact, and I would expect it to fail because of that. If the idea is that the stuff previously tagged with `experimental` is now OK to include everywhere, I'd expect to see `TestExperimentalConformance` go away?

conformance/apis/v1alpha1/doc.go
sayboras: good point, let me revert this change.

sayboras:
> I'd expect to see TestExperimentalConformance go away?

Yes, understand that cleaning up of similar/duplicated code is required to
have a clean conformance test framework, however, such work might delay
v1.0.0 GA release, hence, I only remove the build tag to remove the soft
blocker in this PR. Subsequent cleanup/optimization can be done later.

sayboras:
The previous CI failure was due to recent renaming in main (e.g. HTTPExtendedFeatures -> HTTPRouteExtendedFeatures). The current failure seems to be unrelated to this PR.

mlavacca:
@sayboras, the linter is complaining. It looks like you need to do some `go` and `yaml` linting.

shaneutt (approved):
/approve
/lgtm

arkodg (approved):
/approve
/lgtm

PR #2335, opened 2023-08-21
Bumping go version to 1.20 to match upstream Kubernetes

robscott:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
This bumps the go version to 1.20 to match upstream Kubernetes.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #2336, opened 2023-08-21
Adding changelog for v0.8.0-rc2

robscott:
**What type of PR is this?**
/kind documentation

This PR introduces the changelog for v0.8.0-rc2. We expect this to be the final RC before v0.8.0. By approving this PR, you're approving the release of v0.8.0-rc2.

```release-note
NONE
```

Note: This PR should not merge until #2323 and #2335 merge. Adding a hold for consensus + to wait for those to merge.

/hold


skriss:
Should this PR contain the webhook image tag bump? I'm not exactly sure where that should happen in the release process, just wanted to make sure it wasn't missed.

robscott:
Good catch @skriss! Somehow I'd thought that we were only promoting these images for final/non-RC releases, but it looks like we've also included some RCs as well (https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/images/k8s-staging-gateway-api/images.yaml), so probably best to do that here as well, updated to bump that as well.

PR #2337, opened 2023-08-22
WIP: Allow specifying NodePort for the Listener ports

### LINKED ISSUE #2321: Allow specifying NodePort for the Listener ports
I would like to be able to specify the node port to be used for each Listener port at the Gateway NodePort Svc.

**Why this is needed**:
We are working in a cluster where LoadBalancer svc is not available. We can only use NodePort. It will be much ever ux if the NodePort for a Listener port can be specified in the Gateway object. This will make things more predictable and usable.

Joffref:
**What type of PR is this?**
/kind feature

This PR allows users to specify a NodePort instead of a Port in the Listener.

Fixes #2321

```release-note
Add a NodePort field inside the Listener object.
```


youngnick:
Thanks for suggesting this change @Joffref, but this one will definitely need a GEP as per our GEP process, which you can see at https://gateway-api.sigs.k8s.io/geps/overview/.

The intent for Gateway API is generally to keep the Listener concept _very_ general, without a strong link to a particular implementation (which `nodePort` kind of is), so I'd be interested to hear your thoughts about why you need this and what other value it could add.

Joffref:
Thanks @youngnick for pointing out this! I'll make the GEP to propose these change. Until that, I'm closing this PR. 👍

Issue #2324, opened 2023-08-18
Conformance supported-features flag help string should include all valid values.


briantkennedy:
**What would you like to be added**:

https://github.com/kubernetes-sigs/gateway-api/blob/73f20f4ae4ab46a4a2ea52718fdbdef1a6726376/conformance/utils/flags/flags.go#L30C127-L30C127

The help string for `-supported-features` should include all valid feature values.

**Why this is needed**:

Users should not need to inspect source code to obtain the valid values.

pegasas:
Hi, @shaneutt , 

Do you mean that we should add these support features into description? https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/features.go#L27

Would you like to assign this task to me?

robscott:
Hey @pegasas, yep, I think the goal here is to add the full list of possible features (AllFeatures) as a comma separated list to the description of the flag here: https://github.com/kubernetes-sigs/gateway-api/blob/73f20f4ae4ab46a4a2ea52718fdbdef1a6726376/conformance/utils/flags/flags.go#L30. Thanks for volunteering!

/assign @pegasas

PR #2343, opened 2023-08-23
Conformance supported-features flag help string should include all va…

pegasas:
**What type of PR is this?**

/kind documentation
/area conformance

Add all support features for conformance for better reading & developing experience.

https://github.com/kubernetes-sigs/gateway-api/issues/2324

No.

### DIFF
--- a/conformance/utils/flags/flags.go
+++ b/conformance/utils/flags/flags.go
@@ -21,13 +21,16 @@ package flags
 
 import (
 	"flag"
+	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
 )
 
 var (
 	GatewayClassName           = flag.String("gateway-class", "gateway-conformance", "Name of GatewayClass to use for tests")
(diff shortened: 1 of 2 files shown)


pegasas:
try to resolve https://github.com/kubernetes-sigs/gateway-api/issues/2324

sunjayBhatia (changes requested):
we probably don't want to manually list out supported features since this list will have to be kept up to date and can easily become incorrect/incomplete

rather, lets try to use the `AllFeatures` set: https://github.com/kubernetes-sigs/gateway-api/blob/4ff1670a3b158c2baa83462ae246c1b876617bb5/conformance/utils/suite/features.go#L191

we can use the Set `UnsortedList` helper to get the contents, and then probably have to sort them so we ensure a stable order in the flag help string: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.UnsortedList

pegasas:
hi, @sunjayBhatia ,
I've make another version, 
it seems if we wants to use `suite` package on `flags`, then it will cause cycle import issue.
We may need to refactor the suite and put feature into an single subfolder.
Any suggestions on this?

PR #2350, opened 2023-08-25
Add handling for indistinct ParentRefs with conformance tests

### LINKED ISSUE #1925: Clarification on behavior for duplicate parent refs for routes.
An additional conformance test specifying the behavior when a gateway has only a single listener and an http route specifies the gateway twice: once in specifying the gateway with the listener and once without specifying the specific listener. According to the [docs](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.CommonRouteSpec) this should result in an invalid route configuration. 

This feels like a potentially uncovered edge case in the docs as it is neither two _identical_ parents (as could be detected by the webhook) nor two distinct sections, and _might_ be a valid configuration if the Gateway had a second listener which could match the parentRef not specifying a listener name.

**Why this is needed**:
To clarify the above behavior for implementers  of the gateway spec.

### LINKED ISSUE #2326: Clarify ParentRef Uniqueness with Port and SectionName
Currently the webhook and CEL validation consider the following valid:

```yaml
- name: "foo"
  sectionName: "bar"
- name: "foo"
  port: 80
```

I personally don't think these should be considered unique and instead we should only consider parentRefs unique if:

1. They both specify different section names
OR
2. They both specify different port numbers

Since the `port` field in `parentRef` is still experimental, we still have a window to tighten this validation if we want to. 

Thanks to @gauravkghildiyal for catching this in #2320!

youngnick:
/kind cleanup
/kind test

/area conformance
-->

This PR adds clarity around how ParentRefs must be distinct, with conformance tests to verify.

Fixes #2326
Fixes #1925

```release-note
Clarified that ParentRefs must be distinct.
```

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -172,13 +172,34 @@ type CommonRouteSpec struct {
 	// This API may be extended in the future to support additional kinds of parent
 	// resources.
 	//
-	// It is invalid to reference an identical parent more than once. It is
-	// valid to reference multiple distinct sections within the same parent
-	// resource, such as two separate Listeners on the same Gateway or two separate
-	// ports on the same Service.
+	// ParentRefs must be _distinct_. This means either that:
+	//
+	// * They select different objects.  If this is the case, then parentRef
+	//   entries are distinct. In terms of fields, this means that the
+	//   multi-part key defined by `group`, `kind`, `namespace`, and `name` must be
+	//   unique across all parentRef entries in the Route.
+	// * They do not select different objects, but for each optional field used,
+	//   each ParentRef that selects the same object must set the same set of
+	//   optional fields to different values. If one ParentRef sets a
+	//   combination of optional fields, all must set the same combination.
+	//
+	// Some examples:
+	//
+	// * If one ParentRef sets `sectionName`, all ParentRefs referencing the same
+	//   object must also set `sectionName`.
+	// * If one ParentRef sets `port`, all ParentRefs referencing the same
+	//   object must also set `port`.
+	// * If one ParentRef sets `sectionName` and `port`, all ParentRefs referencing
+	//   the same object must also set `sectionName` and `port`.
+	//
+	// It is invalid to reference non-distinct ParentRefs in the same Route object.
+	// In this case, the Route must have the `Accepted` Condition set to `false`,
+	// with a Reason of `ParentsNotDistinct`. Note that _all_ ParentRefs must be
+	// distinct - having one distinct ParentRef and two indistinct ParentRefs
+	// must result in the entire Route being marked not `Accepted`, for example.
 	//
 	// It is possible to separately reference multiple distinct objects that may
-	// be collapsed by an implementation. For example, some implementations may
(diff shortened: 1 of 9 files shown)


youngnick:
/hold

for multiple review if necessary.

apis/v1beta1/shared_types.go
```diff
@@ -295,6 +317,11 @@ const (
...
 	RouteReasonNoMatchingParent RouteConditionReason = "NoMatchingParent"
 
+	// This reason is used with the "Accepted" condition when there are
+	// multiple ParentRefs that are not distinct. See the ParentRef
+	// definition for the definition of distinct.
+	RouteReasonParentsNotDistinct RouteConditionReason = "ParentsNotDistinct"
+
 	// This reason is used with the "Accepted" condition when a value for an Enum
 	// is not recognized.
 	RouteReasonUnsupportedValue RouteConditionReason = "UnsupportedValue"
```
shaneutt, line 327: I'm pretty happy with the language changes in here :+1: 

We should make sure we hold for @robscott before merging.

/hold

conformance/tests/httproute-invalid-parentref-not-distinct.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
shaneutt, line 2: </img>

```suggestion
Copyright 2023 The Kubernetes Authors.
```

apis/v1beta1/shared_types.go
```diff
@@ -295,6 +317,11 @@ const (
 	// match any Listeners in the Gateway.
 	RouteReasonNoMatchingParent RouteConditionReason = "NoMatchingParent"
 
+	// This reason is used with the "Accepted" condition when there are
+	// multiple ParentRefs that are not distinct. See the ParentRef
+	// definition for the definition of distinct.
+	RouteReasonParentsNotDistinct RouteConditionReason = "ParentsNotDistinct"
```
robscott, line 323: What do we want to happen in the following case?

```yaml
parentRefs:
- name: "foo"
- name: "foo"
  sectionName: "bar"
```

Do we always want status to look something like this?

```yaml
status:
  parents:
  - parentRef:
    - name: foo
    conditions:
      type: Accepted
      status: True
  - parentRef:
    - name: foo
      sectionName: Bar
    conditions:
      type: Accepted
      status: False
      reason: ParentsNotDistinct
```

Or do we always want to consider both parents unaccepted in this state? I'd argue that it makes sense to accept one of the indistinct parents, preferably the oldest/first one.

robscott (commented):
Thanks @youngnick! Mostly LGTM, just one question on how this will work in practice.

apis/v1beta1/shared_types.go
shaneutt: We talked about this one on the meeting today, and decided we should see if CEL can provide validation to protect us from these situations.

robscott:
Closing in favor of #2433 (which copies some of the content from this one).

/close

PR #2362, opened 2023-08-29
Fixes for Verify CRDs script and invalid YAMLs

youngnick:
/kind cleanup
/kind test

This updates `verify-crds-kind.sh` with a few changes:
- Moves the apply operation to a bash for loop to make the checks easier to read and add
- Updates the logic so that anything that's not an allowlisted error will cause the job to fail (there were structural errors in the GRPCRoute invalid YAMLs that meant they were the _wrong type_ of invalid, but the check didn't pick that up. This version does).
- Runs the invalid YAML examples both with just CEL and with webhook (since the webhook will go away soon)
- Updates the output of the job to make finding the actual error easier.
- Removes the v1alpha2 examples that are no longer being served (`gateway`, `gatewayclass`, and `httproute` v1alpha2 examples). They were also invalid in the wrong way - this error case is also handled by the new script.

```release-note
NONE
```


youngnick:
/hold

for multiple review if necessary.

hack/verify-crds-kind.sh
```diff
@@ -67,6 +66,63 @@ for CHANNEL in experimental standard; do
...
 done
 
+# Temporary workaround for https://github.com/kubernetes/kubernetes/issues/104090
+sleep 8
+
+## Check using example YAMLs as well
+## with _only_ CEL validation
+
+
+for CHANNEL in experimental standard; do
```
youngnick, line 76: This whole block is duplicated, yes, but I didn't want to pull this into a bash function because it's more complex that way, and the webhook section has a limited lifespan.

hack/verify-crds-kind.sh
```diff
@@ -108,39 +165,43 @@ for check in {1..10}; do
...
+  # just CEL validation.
 
-  # Install all experimental example gateway-api resources when experimental mode is enabled
-  if [[ "${CHANNEL}" == "experimental" ]]; then
-    echo "Experimental mode enabled: deploying experimental examples"
-    kubectl apply --recursive -f examples/experimental || res=$?
-  fi
+  for file in $(find hack/invalid-examples -name "*.yaml"); do
+    # Don't check alpha resources in Standard checks
+    if [[ "$file" =~ "v1alpha2" && "$CHANNEL" == "standard" ]]; then
```
robscott, line 180: Maybe we should just be using `hack/invalid-examples/experimental` instead of `hack/invalid-examples/v1alpha2` to be more consistent with what we're doing with other examples.

hack/invalid-examples/v1alpha2/gateway/duplicate-listeners.yaml
```diff
@@ -1,13 +0,0 @@
-apiVersion: gateway.networking.k8s.io/v1alpha2
```
robscott, line 1: Why are so many of thee being removed?

robscott (commented):
Thanks @youngnick!

hack/invalid-examples/v1alpha2/gateway/duplicate-listeners.yaml
youngnick: Yes, there are actually more examples in v1beta1, but all the ones being removed are covered already.

hack/verify-crds-kind.sh
youngnick: Hmm, yeah, I can move the examples to `experimental` and `standard` directories instead of `v1alpha2` and `v1beta1` directories? That will be consistent with the working examples dir.

robscott:
Thanks @youngnick!

/lgtm

PR #2370, opened 2023-09-01
Fix CEL validation for listener uniqueness

### LINKED ISSUE #2369: GatewaySpec Listener CEL block empty Hostname if similar Listeners also present
**What happened**:

0.8.0 does not allow a Gateway with two same Port, same Protocol Listeners where one Listener has Hostname set and the other does not. This violates [a CEL rule](https://github.com/kubernetes-sigs/gateway-api/blob/v0.8.0/apis/v1beta1/gateway_types.go#L131) that ensures Listeners have a unique Port/Protocol/Hostname combination.

The Gateway passes validation. Per [the above spec comment](https://github.com/kubernetes-sigs/gateway-api/blob/v0.8.0/apis/v1beta1/gateway_types.go#L105-L107):

> one Listener within a group may omit Hostname, in which case this Listener matches when no other Listener matches.

With 0.8.0 installed:

```
$ echo "apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway                                                             
metadata:    
  name: prod-web
spec:           
  gatewayClassName: acme-lb
  listeners:               
  - protocol: HTTP
    port: 80      
    name: http
  - protocol: HTTP
    port: 80      
    name: httphostname
    hostname: http.example" | kubectl create -f -
The Gateway "prod-web" is invalid: spec.listeners: Invalid value: "array": Combination of port, protocol and hostname must be unique for each listener
```

It looks like

```
(has(l1.hostname) && has(l2.hostname) ? l1.hostname == l2.hostname : true
```
requires a set value to compare if Hostname is equal.

Setting Hostname to `""` or `"*"` does not work around this, as these values violate Hostname's `'^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'` validation regex.

frankbu:
**What type of PR is this?**

/kind bug

CEL validation not handling missing listener hostname corectly.

Fixes #2369 

```release-note
NONE
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -128,7 +128,7 @@ type GatewaySpec struct {
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="hostname must not be specified for protocols ['TCP', 'UDP']",rule="self.all(l, l.protocol in ['TCP', 'UDP']  ? (!has(l.hostname) || l.hostname == '') : true)"
 	// +kubebuilder:validation:XValidation:message="Listener name must be unique within the Gateway",rule="self.all(l1, self.exists_one(l2, l1.name == l2.name))"
-	// +kubebuilder:validation:XValidation:message="Combination of port, protocol and hostname must be unique for each listener",rule="self.all(l1, self.exists_one(l2, l1.port == l2.port && l1.protocol == l2.protocol && (has(l1.hostname) && has(l2.hostname) ? l1.hostname == l2.hostname : true)))"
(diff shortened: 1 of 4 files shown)


frankbu:
@robscott I added a test, but am wondering if the CEL tests getting run? I don't see them in https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_gateway-api/2370/pull-gateway-api-test/1697659277918867456

robscott:
@frankbu good question! This is decidedly not intuitive, but even though those are go tests, they require a k8s cluster with CRDs installed to be present, so these tests actually are installed in `hack/verify-crds-kind.sh`. The test results start around line 151 in the output of the verify test run: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_gateway-api/2370/pull-gateway-api-verify/1697659277851758592. All tests passed. Will defer to someone else for LGTM, but I think this should fix it.

/approve

howardjohn:
You can test the CRD validation locally fwiw, we have a package to do so in
Istio somewhere

gauravkghildiyal:
Thanks @frankbu.
LGTM

youngnick:
Thanks for this @frankbu.

/lgtm

@howardjohn, if you can find that package, we should probably create an issue to update this script-based arrangement to something better. (I also have #2362 open to make some changes to it for the YAML-based invalid example tests)

PR #2375, opened 2023-09-04
IPv6 conformance: Third time's the charm

keithmattix:
`net.SplitHostPort` errors if ipv6 addresses aren't in brackets; previous implementation fixed a different problem (SplitHostPort removes brackets if they are passed in)

```release-note
Fix IPv6 parsing in conformance tests
```

### DIFF
--- a/conformance/utils/echo/pod.go
+++ b/conformance/utils/echo/pod.go
@@ -118,7 +118,8 @@ func calculateHost(t *testing.T, reqHost, scheme string) string {
 	if strings.Contains(err.Error(), "too many colons in address") {
 		// This is an IPv6 address; assume it's valid ipv6
 		// Assume caller won't add a port without brackets
-		host, port, err = net.SplitHostPort("[" + reqHost + "]")
+		reqHost = "[" + reqHost + "]"
+		host, port, err = net.SplitHostPort(reqHost)
 	}
 	if err != nil {
 		t.Logf("Failed to parse host %q: %v", reqHost, err)
--- a/conformance/utils/echo/pod.go
+++ b/conformance/utils/echo/pod.go
@@ -58,7 +58,7 @@ func (m *MeshPod) MakeRequestAndExpectEventuallyConsistentResponse(t *testing.T,
 	t.Helper()
 
 	http.AwaitConvergence(t, timeoutConfig.RequiredConsecutiveSuccesses, timeoutConfig.MaxTimeToConsistency, func(elapsed time.Duration) bool {
-		req := makeRequest(exp.Request)
+		req := makeRequest(t, exp.Request)
 
 		resp, err := m.request(req)
 		if err != nil {
@@ -76,12 +76,12 @@ func (m *MeshPod) MakeRequestAndExpectEventuallyConsistentResponse(t *testing.T,
 	t.Logf("Request passed")
 }
 
-func makeRequest(r http.Request) []string {
+func makeRequest(t *testing.T, r http.Request) []string {
 	protocol := strings.ToLower(r.Protocol)
 	if protocol == "" {
 		protocol = "http"
 	}
-	host := calculateHost(r.Host, protocol)
+	host := calculateHost(t, r.Host, protocol)
 	args := []string{"client", fmt.Sprintf("%s://%s%s", protocol, host, r.Path)}
 	if r.Method != "" {
 		args = append(args, "--method="+r.Method)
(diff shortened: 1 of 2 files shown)


conformance/utils/echo/pod.go
```diff
@@ -113,32 +113,27 @@ func compareRequest(exp http.ExpectedResponse, resp Response) error {
...
+func calculateHost(t *testing.T, reqHost, scheme string) string {
 	host, port, err := net.SplitHostPort(reqHost)
+	if err != nil && strings.Contains(err.Error(), "too many colons in address") {
+		// This is an IPv6 address; assume it's valid ipv6
+		// Assume caller won't add a port without brackets
+		reqHost = "[" + reqHost + "]"
+		host, port, err = net.SplitHostPort(reqHost)
+	}
 	if err != nil {
+		t.Logf("Failed to parse host %q: %v", reqHost, err)
```
Xunzhuo, line 125: `t.Errorf` or `t.Logf` to print errors ?

conformance/utils/http/http.go
```diff
@@ -145,29 +145,25 @@ func MakeRequest(t *testing.T, expected *ExpectedResponse, gwAddr, protocol, sch
...
+	host, port, err := net.SplitHostPort(reqHost)
+	if err != nil && strings.Contains(err.Error(), "too many colons in address") {
+		// This is an IPv6 address; assume it's valid ipv6
+		// Assume caller won't add a port without brackets
+		reqHost = "[" + reqHost + "]"
+		host, port, err = net.SplitHostPort(reqHost)
+	}
 	if err != nil {
-		return gwAddr
+		t.Logf("Failed to parse host %q: %v", reqHost, err)
```
Xunzhuo, line 157: ditto

conformance/utils/echo/pod.go
keithmattix: These aren't fatal errors, nor are they being bubbled up and reported, so I opted for a generic log. Note that the test continues even if an error in parsing occurs. I would choose a warn log level if it existed

aojea:
> net.SplitHostPort errors if ipv6 addresses aren't in brackets

And that's the expectation , how can you get a host:port without brackets?

keithmattix:
> > net.SplitHostPort errors if ipv6 addresses aren't in brackets
> 
> 
> 
> And that's the expectation , how can you get a host:port without brackets? 

Indeed; our mesh conformance tests  sometimes pass a raw podIP, so we need to change up the logic a bit

aojea:
got it, thanks for clarifying 😄

howardjohn (approved):

robscott:
Thanks @keithmattix! This LGTM, but do you mind adding a release-note block?

keithmattix:
Ah thanks for the reminder @robscott. Just added one 👍🏾

conformance/utils/echo/pod.go
```diff
@@ -113,10 +113,16 @@ func compareRequest(exp http.ExpectedResponse, resp Response) error {
 	return nil
 }
 
-// copied from conformance/http to get the ipv6 matching
-func calculateHost(reqHost, scheme string) string {
-	host, port, err := net.SplitHostPort(reqHost)
+func calculateHost(t *testing.T, reqHost, scheme string) string {
```
frankbu, line 116: I realize that this PR is just following the already established pattern, but why do we need two copies of the same calculateHost function? Can't pod.go just use the one in http.go instead of duplicating the code?

keithmattix: Unexported functions in different packages I'm afraid. I could add it to a general utils package if that'll be helpful?

robscott: I think a shared package is a good idea here so we don't introduce accidental variations between packages. A couple ideas:

1. Introduce a new `conformance/utils/net` package that could house this and `ipv6SafeHost` functions
2. Export `calculateHost` from `conformance/utils/http` function since it's already there.

Since this package already depends on `conformance/utils/http`, I think I'd lean towards 2.

frankbu: > Since this package already depends on conformance/utils/http, I think I'd lean towards 2.

+1

keithmattix: Nice catch; I missed the existing dependency. Done!

robscott:
Thanks @keithmattix!

/lgtm

PR #2380, opened 2023-09-07
Conformance: update objects in MustApplyObjectsWithCleanup

### LINKED ISSUE #2230: conformance: update objects in MustApplyObjectsWithCleanup
MustApplyObjectsWithCleanup does not `Update` if the object exists. `MustApplyWithCleanup` does.

We should make these do the same thing, likely DRYing the code as well.

If we don't want to do this, we should rename MustApplyObjectsWithCleanup to MustCreateObjectsWithCleanup to be more clear.

In my case, I ran different versions of the test against the same cluster and got stale config as a result of this, which made me confused. Not a bit deal, but nice to fix

sadath-12:
**What type of PR is this?**

```MustApplyObjectsWithCleanup``` updates the objects if they are already available available

Fixes #2230 

```release-note
Bug Fixes : Updates the resource objects instead of keeping them same if the resource already exists
```


conformance/utils/kubernetes/apply.go
```diff
@@ -143,13 +143,24 @@ func (a Applier) MustApplyObjectsWithCleanup(t *testing.T, c client.Client, time
...
+			if !apierrors.IsNotFound(err) {
+				t.Logf("error checking resource: %s %s", resource.GetName(), resource.GetObjectKind().GroupVersionKind().Kind)
+				require.NoError(t, err, "error checking resource")
 			}
+
+			// Resource doesn't exist, create it
+			err = c.Create(ctx, resource)
+			require.NoError(t, err, "error creating resource")
+		} else {
+			// Resource exists, update it
```
LiorLieberman, line 161: same comment about logging

conformance/utils/kubernetes/apply.go
```diff
@@ -143,13 +143,24 @@ func (a Applier) MustApplyObjectsWithCleanup(t *testing.T, c client.Client, time
...
+			err = c.Update(ctx, resource)
```
LiorLieberman, line 162: if you update you might want to setResourceVersion
Like MustApplyWithCleanup is doing 

`uObj.SetResourceVersion(fetchedObj.GetResourceVersion())`

LiorLieberman (commented):
Thanks @sadath-12 , left a few comments

sadath-12:
@LiorLieberman  any further suggestions on the updated on code ?

conformance/utils/kubernetes/apply.go
```diff
@@ -161,6 +182,8 @@ func (a Applier) MustApplyObjectsWithCleanup(t *testing.T, c client.Client, time
 				require.NoErrorf(t, err, "error deleting resource")
 			})
 		}
+
+		require.NoErrorf(t, err, "error updating resource")
```
LiorLieberman, line 186: maybe we should do this directly after updating ? (I know the below function is doing it like this but I think thats not intentional.

conformance/mesh/manifests.yaml
LiorLieberman, line 1: I do not thing changes to this file are needed

conformance/utils/suite/suite.go
LiorLieberman, line 1: I do not think changes here are needed

LiorLieberman (commented):
Thanks @sadath-12 , apart from two small comments, looks good to me.
Did you test it locally ? I can help with explaining how if you need.

conformance/utils/kubernetes/apply.go
LiorLieberman: You can resolve this

sadath-12:
> Thanks @sadath-12 , apart from two small comments, looks good to me. Did you test it locally ? I can help with explaining how if you need.

Thank you @LiorLieberman , Yes I tried creating a secret in the cluster with same resource specifications but with different data and passed it to this function (which creates the secret with another data) and that seems to give me the modified data content on the secret

LiorLieberman:
Can you test that the conformance tests are passing?

Using the makefile I sent you for example in envoy-gw. (
https://github.com/envoyproxy/gateway/blob/554176b8cbbc5ffa627dc7109e1b3d2b59474e99/tools/make/kube.mk#L95
)

Remove the  “delete-cluster” from the end of the line, it will spin up a
kind cluster with meta lb and will leave it there.
Then you can run locally from the gateway-api repo folder (that contains
your changes).

one with cleanup  `go test -v  ./conformance
--gateway-class=envoy-gateway --cleanup-base-resources=true`

and one without  `go test -v  ./conformance --gateway-class=envoy-gateway
--cleanup-base-resources=false`

w

robscott:
Looks like this PR has gone stale so I'm going to close it out for now. Feel free to reopen if you find time to work on it again.

PR #2384, opened 2023-09-09
Adding new SupportedVersion condition to GatewayClass

### LINKED ISSUE #2077: Implementation Guidelines: Document Version Skew Between Implementations and CRDs
Update our Implementation Guidelines documentation to document what implementations should do when Gateway API CRDs are newer or older than expected.

**Why this is needed**:
As we release more versions of Gateway API and more implementations of the API are available, version skew is going to be increasingly common. This will be especially problematic when new features are added to the CRDs but one or more implementations in the cluster is not aware of them.

robscott:
**What type of PR is this?**
/kind documentation
/kind feature

This is intended to fix https://github.com/kubernetes-sigs/gateway-api/issues/2077 by adding a new SupportedVersion condition on GatewayClass and corresponding documentation.

Fixes #2077

```release-note
GatewayClass Status: A new SupportedVersion condition has been added that MUST be set when a GatewayClass is accepted.
```

/hold for consensus

### DIFF
--- a/apis/v1beta1/gatewayclass_types.go
+++ b/apis/v1beta1/gatewayclass_types.go
@@ -158,6 +158,7 @@ const (
 	// Possible reasons for this condition to be False are:
 	//
 	// * "InvalidParameters"
+	// * "UnsupportedVersion"
 	//
 	// Possible reasons for this condition to be Unknown are:
 	//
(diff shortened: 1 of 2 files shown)


youngnick (commented):
Love it, LGTM. Holding though in case we need further review.

/lgtm
/hold

apis/v1beta1/gatewayclass_types.go
```diff
@@ -186,6 +187,49 @@ const (
 	GatewayClassReasonWaiting GatewayClassConditionReason = "Waiting"
 )
 
+const (
+	// This condition indicates whether the GatewayClass supports the version(s)
+	// of Gateway API CRDs present in the cluster. This condition MUST be set by
```
pleshakov, line 192: Since this condition "MUST be set", would it make sense to update the conformance tests too?

robscott: Yes, I think we'll want to have corresponding updates to the conformance tests. I'd initially thought those should be a separate PR, but they may be simple enough to combine here.

shaneutt (approved):
I'm aware of the dilemma with generally not wanting to rely on information from CRDs, but I'm also aware we don't have an alternative to get this feature. Also if we ever needed to change it, it would be mostly on implementations. Generally, if we ever move off CRDs I think having to change this will be the least of problems we have to contend with.

youngnick:
I think that we have no option than to require something like this; I'm working on an implementer's guide that will lay out some suggestions for how to determine the required information in the right way.

I like the Condition idea because it allows implementations to mark the GatewayClass as `Accepted`, but show a clear warning that not everything may work.

I also agree that it would be nice to have conformance tests for this behavior, but I think that they may be harder to build than it seems (since we don't have any way to install different versions of the CRDs in the testing suite).

So I am going to merge this as-is, and create an issue for follow-up conformance testing of this Condition.

/unhold

PR #2386, opened 2023-09-11
Fix CEL tests in fix-verify-crds.sh

youngnick:
**What type of PR is this?**
/kind cleanup
/kind test

/kind failing-test

When I refactored the `fix-verify-crds.sh` script, I removed the `set -o errexit` directive, but missed that this meant that the `go test` invocation was now not updating the `$res` variable. This fixes that so that CEL test failures will now cause the tests to fail as expected.

```release-note
NONE
```


shaneutt (approved):
/approve
/lgtm

PR #2389, opened 2023-09-12
chore: remove `conformance/reports/latest` link

pmalek:
**What type of PR is this?**

/kind cleanup
/area conformance

Removes the `conformance/reports/latest`.

```release-note
NONE
```


shaneutt:
So I was thinking about this, and all things considered: I don't think we actually need this symlink. It would be _potentially_ a convenience if we were programmatically scanning the latest, but even then we could just use semver to figure that out. So rather than have to add a release step and all that, maybe we should just remove it entirely?

pmalek:
> So I was thinking about this, and all things considered: I don't think we actually need this symlink. It would be _potentially_ a convenience if we were programmatically scanning the latest, but even then we could just use semver to figure that out. So rather than have to add a release step and all that, maybe we should just remove it entirely?

👍 Removed

mlavacca (approved):
Thanks @pmalek!

/ok-to-test
/approve

shaneutt (approved):
/approve
/lgtm

PR #2391, opened 2023-09-13
Conformance profiles modes and versions

### LINKED ISSUE #2233: conformance profiles: how to handle multiple "modes" in a project
Slack context: https://kubernetes.slack.com/archives/CR0H13KGA/p1690238884857489?thread_ts=1690225970.833499&cid=CR0H13KGA

Many projects can operate in different modes. Some examples are Istio in sidecar mode vs ambient, GKE in internal LB mode or external LB mode (or multicluster vs single cluster), etc.

We should prescribe how these are reported in conformance profiles.

mlavacca:
**What type of PR is this?**

/kind gep
/area conformance

This PR updates GEP-1709 with the following goals:

- introduce a more structured approach to upload and store multiple conformance reports coming from different implementations.
- improve the `conformanceReport` API to include the implementation mode used during the certification process.
- introduce the `To reproduce` field in a README.md file with the steps to reproduce the claimed report. This improves the projects' accountability.

Fixes #2233 

```release-note
NONE
```

### DIFF
--- a/geps/gep-1709/index.md
+++ b/geps/gep-1709/index.md
@@ -242,7 +242,7 @@ OR internal load balancer, and have different capabilities depending on the mode
 Every certification report `modes` array contains only one element. Reports can be merged
 into one another to provide a single report that accounts for multiple modes, given the same
 Gateway API version.
-omni-comprehensive one, as described [later](#multiple-reports-merge).
+omni-comprehensive one, as described [below](#multiple-reports-merge).
 
 ### Implementation version
 
--- a/geps/gep-1709/index.md
+++ b/geps/gep-1709/index.md
@@ -235,15 +235,20 @@ move on to [certification](#certification) to report the results.
 
 The certification process runs against an implementation using a specific mode
 (e.g., `standard`, `enhanced`) that is defaulted to `standard`. The final
-report will contain such a mode in the modes field. Every certification report
-`modes` array contains only one element. Multiple reports can be merged into an
+report will contain such a mode in the modes field. A "mode" is intended to capture
+situations where a Gateway API implementation may have different features and capabilities
+depending on how it is deployed (e.g. an implementation might be deployed with an external,
+OR internal load balancer, and have different capabilities depending on the mode chosen).
+Every certification report `modes` array contains only one element. Reports can be merged
+into one another to provide a single report that accounts for multiple modes, given the same
+Gateway API version.
 omni-comprehensive one, as described [later](#multiple-reports-merge).
 
 ### Implementation version
 
 Every report can specify multiple results, dependent on the tested version of the
-certifying implementation. Each certification run uses a specific version of the
-implementation, multiple certification processes can be merged as explained [later](#multiple-reports-merge).
+implementation being certified. Each certification run uses a specific version of the
+implementation, multiple certification processes can be merged as explained [below](#multiple-reports-merge).
 
 ### Gateway API version and channel
 
@@ -282,16 +287,20 @@ generated `.go` file is up to date with the VERSION file.
 
 ### Multiple reports merge
 
-Every profile gives the possibility to specify multiple modes and
-implementation versions, but the testing process is related to a specific mode
-and a specific version. In order to produce reports with multiple modes and versions,
-a proper flag `merge-report-with` can be set when running the certification process.
-The flag value is taken, and if no file with such a name exists, it gets created
+In order to produce a single report which contains reports from multiple test
+runs each testing against distinct _modes_,
+the flag `merge-report-with` can be set when running the certification process.
+The flag value is expected to be a file name. If no file with exists with the given name, it gets created
 and the report is written therein. If the file already exists, the suite merges
 the current report with the given one. By proceeding this way, it is possible to
 iteratively run multiple times the certification suite with different setups
 (modes and versions) and provide a unique report.
 
+> **Warning**: By default trying to merge a new report that contains a mode that already exists in the 
+> existing report specified by `merge-report-with` will result in an error, to avoid automatically deciding
+> which report is the one the user actually intended to provide and then deleting data. Implementors
+> will need to manually remove a mode from an existing report if they wish to override, or they can use
+> the `merge-report-overwrite-existing` flag to explicitly indicate this is what they want.
 ### Certification
 
 Implementations will be able to report their conformance testing results using
--- a/geps/gep-1709/index.md
+++ b/geps/gep-1709/index.md
@@ -13,7 +13,7 @@ the Gateway API project and receive recognition (e.g. badges).
 
 ## Goals
 
-- Add high level profiles which downstream implementations can subscribe to in
+- Add high-level profiles which downstream implementations can subscribe to in
   order to run tests for the associated supported feature sets.
 - Add a reporting mechanism where conformance results can be reported back to
   the Gateway API project and provide "badges" to visibly decorate the
@@ -71,7 +71,7 @@ conformance suite as a library.
 
 ### Profiles
 
-"Profiles" are effectively categories which represent the high level grouping of
+"Profiles" are effectively categories which represent the high-level grouping of
 tests related to some feature (or feature set) of Gateway API. When conformance
 is reported using one of these profiles extra features can be covered according
 to support levels:
@@ -239,10 +239,7 @@ to capture situations where a Gateway API implementation may have different feat
 and capabilities depending on how it is deployed (e.g. an implementation might be
 deployed with an external, OR internal load balancer, and have different capabilities
 depending on the mode chosen). The modes are implementation-specific, and no upstream
-mode is defined, except for `standard`, the default one used in case no mode is specified.
-Every certification report `modes` array contains only one element. Reports can
-be merged into one another to provide a single report that accounts for multiple
-modes, given the same Gateway API version as described [below](#multiple-reports-merge).
+mode is defined, except for `plain`, the default one used in case no mode is specified.
 
 ### Implementation version
 
@@ -285,33 +282,12 @@ VERSION file contains the semver of the latest release and is manually bumped at
 release time. The script hack/verify-all.sh will be updated to ensure the
 generated `.go` file is up to date with the VERSION file.
 
-### Multiple reports merge
-
-In order to produce a single report which contains reports from multiple test
-runs each testing against distinct _modes_, the flag `merge-report-with` can be
-set when running the certification process. The flag value is expected to be a file
-name. If no file with the given name exists, it gets created and the report is
-written therein. If the file already exists, the suite merges the current report
-with the given one. By proceeding this way, it is possible to iteratively run multiple
-times the certification suite with different setups (modes and versions) and provide
-a unique report.
-
-> **Warning**: By default trying to merge a new report that contains a mode that
-> already exists in the existing report specified by `merge-report-with` will result
-> in an error, to avoid automatically deciding which report is the one the user
-> actually intended to provide and then deleting data. Implementors will need to
-> manually remove a mode from an existing report if they wish to override, or they
-> can use the `merge-report-overwrite-existing` flag to explicitly indicate this
-> is what they want.
-
-### Certification
-
 Implementations will be able to report their conformance testing results using
 our [reporting process](#reporting-process). Implementations will be able to
 visibly demonstrate their conformance results on their downstream projects and
 repositories using our [certification process](#certification-process).
 
-#### Reporting Process
+## Reporting process
 
 When conformance tests are executed an argument can be provided to the test
 suite to emit `ConformanceReport` resource with the test results. This resource
@@ -328,80 +304,54 @@ implementation:
   url: https://acme.com
   contact:
   - @acme/maintainers
+  version: v1.0.0
 date: "2023-02-28 20:29:41+00:00"
-reportResult:
-  - implementationVersion: v1.0.0
-    apiVersion: v0.8.0
-    apiChannel: standard
-    reports:
-    - modes:
-      - plain
-      profiles:
-        - name: http
-          core:
-            result: success
-            summary: "all core functionality passed"
-            statistics:
-              passed: 20
-              skipped: 0
-              failed: 0
-          extended:
-            result: success
-            summary: "all extended features supported"
-            statistics:
-              passed: 8
-              skipped: 0
-              failed: 0
-            supportedFeatures:
-            - ExtendedFeature1
-            - ExtendedFeature2
-            - ExtendedFeature3
-            - ExtendedFeature4
-            - ExtendedFeature5
-    - modes: 
-      - with-the-lot
-      profiles:
-        - name: tcp
-          core:
-            result: success
-            summary: "all core functionality passed"
-            statistics:
-              passed: 4
-              skipped: 0
-              failed: 0
-          extended:
-            result: skipped
-            summary: "some extended features supported"
-            statistics:
-              passed: 2
-              skipped: 0
-              failed: 0
-            supportedFeatures:
-            - ExtendedFeature1
-            - ExtendedFeature2
-            unsupportedFeatures:
-            - ExtendedFeature3
-        - name: http
-          core:
-            result: success
-            summary: "all core functionality passed"
-            statistics:
-              passed: 20
-              skipped: 0
-              failed: 0
-          extended:
-            result: success
-            summary: "all extended features supported"
-            statistics:
-              passed: 8
-              skipped: 0
-              failed: 0
-            supportedFeatures:
-            - ExtendedFeature1
-            - ExtendedFeature2
-            - ExtendedFeature3
-            - ExtendedFeature4
-            - ExtendedFeature5
+apiVersion: v0.8.0
+apiChannel: experimental
+mode: plain
+profiles:
+  profiles:
+  - name: http
+    core:
+      result: success
+      summary: "all core functionality passed"
+      statistics:
+        passed: 20
+        skipped: 0
+        failed: 0
+    extended:
+      result: success
+      summary: "all extended features supported"
+      statistics:
+        passed: 8
+        skipped: 0
+        failed: 0
+      supportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      - ExtendedFeature3
+      - ExtendedFeature4
+      - ExtendedFeature5
+  - name: tcp
+    core:
+      result: success
+      summary: "all core functionality passed"
+      statistics:
+        passed: 4
+        skipped: 0
+        failed: 0
+    extended:
+      result: skipped
+      summary: "some extended features supported"
+      statistics:
+        passed: 2
+        skipped: 0
+        failed: 0
+      supportedFeatures:
+      - ExtendedFeature1
+      - ExtendedFeature2
+      unsupportedFeatures:
+      - ExtendedFeature3
 ```
 
 > **WARNING**: It is an important clarification that this is NOT a full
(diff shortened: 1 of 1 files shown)


geps/gep-1709.md
```diff
@@ -231,6 +231,20 @@ move on to [certification](#certification) to report the results.
...
+The certification process runs against an implementation using a specific mode
+(e.g., `standard`, `enhanced`) that is defaulted to `standard`. The final
+report will contain such a mode in the modes field. Every certification report
+`modes` array contains only one element. Multiple reports can be merged into an
+omni-comprehensive one, as described [later](#multiple-reports-merge).
+
+### Implementation version
+
+Every report can specify multiple results, dependent on the tested version of the
+certifying implementation. Each certification run uses a specific version of the
```
shaneutt, line 245:
```suggestion
implementation being certified. Each certification run uses a specific version of the
```

geps/gep-1709.md
```diff
@@ -231,6 +231,20 @@ move on to [certification](#certification) to report the results.
...
+implementation, multiple certification processes can be merged as explained [later](#multiple-reports-merge).
```
shaneutt, line 246:
```suggestion
implementation, multiple certification processes can be merged as explained [below](#multiple-reports-merge).
```

geps/gep-1709.md
```diff
@@ -231,6 +231,20 @@ move on to [certification](#certification) to report the results.
 [go]:https://go.dev
 [lib]:https://pkg.go.dev/sigs.k8s.io/gateway-api@v0.6.2/conformance/utils/suite
 
+### Implementation mode
+
+The certification process runs against an implementation using a specific mode
+(e.g., `standard`, `enhanced`) that is defaulted to `standard`. The final
+report will contain such a mode in the modes field. Every certification report
+`modes` array contains only one element. Multiple reports can be merged into an
```
shaneutt, line 239: This patch tries to add more definition to what we mean by "mode".

```suggestion
report will contain such a mode in the modes field. A "mode" is intended to capture
situations where a Gateway API implementation may have different features and capabilities
depending on how it is deployed (e.g. an implementation might be deployed with an external,
OR internal load balancer, and have different capabilities depending on the mode chosen).
Every certification report `modes` array contains only one element. Reports can be merged
into one another to provide a single report that accounts for multiple modes, given the same
Gateway API version.
```

geps/gep-1709.md
```diff
@@ -266,6 +280,18 @@ VERSION file contains the semver of the latest release and is manually bumped at
 release time. The script hack/verify-all.sh will be updated to ensure the
 generated `.go` file is up to date with the VERSION file.
 
+### Multiple reports merge
+
+Every profile gives the possibility to specify multiple modes and
+implementation versions, but the testing process is related to a specific mode
+and a specific version. In order to produce reports with multiple modes and versions,
```
shaneutt, line 287:
```suggestion
In order to produce a single report which contains reports from multiple test
runs each testing against distinct _modes_,
```

geps/gep-1709.md
```diff
@@ -266,6 +280,18 @@ VERSION file contains the semver of the latest release and is manually bumped at
...
+a proper flag `merge-report-with` can be set when running the certification process.
```
shaneutt, line 288:
```suggestion
the flag `merge-report-with` can be set when running the certification process.
```

geps/gep-1709.md
```diff
@@ -266,6 +280,18 @@ VERSION file contains the semver of the latest release and is manually bumped at
...
+The flag value is taken, and if no file with such a name exists, it gets created
```
shaneutt, line 289:
```suggestion
The flag value is expected to be a file name. If no file with exists with the given name, it gets created
```

geps/gep-1709.md
```diff
@@ -266,6 +280,18 @@ VERSION file contains the semver of the latest release and is manually bumped at
...
+and the report is written therein. If the file already exists, the suite merges
+the current report with the given one. By proceeding this way, it is possible to
+iteratively run multiple times the certification suite with different setups
+(modes and versions) and provide a unique report.
+
```
shaneutt, line 294: I think it's a minor thing, but it might be helpful to add a default behavior of not overwriting data, and enable a flag:

```suggestion

> **Warning**: By default trying to merge a new report that contains a mode that already exists in the 
> existing report specified by `merge-report-with` will result in an error, to avoid automatically deciding
> which report is the one the user actually intended to provide and then deleting data. Implementors
> will need to manually remove a mode from an existing report if they wish to override, or they can use
> the `merge-report-overwrite-existing` flag to explicitly indicate this is what they want.
```

geps/gep-1709.md
```diff
@@ -355,9 +410,12 @@ profiles:
...
 
-The above report describes an implemenation that just released `v1` and has
-`Core` support for `TCP` functionality and fully supports both `Core` and
-`Extended` `HTTP` functionality.
+The above report describes an implementation that just released `v1`, uses gateway
+API `v0.8.0` `standard` channel, and has:
+
+* `HTTP` `core` and `extended` support in `standard` mode;
+* `HTTP` `core` and `extended` support in `enhanced` mode;
+* `TCP` `Core` support in `enhanced` mode.
```
shaneutt, line 418: From the perspective of this document, I don't feel like `enhanced` has enough meaning or context here.

geps/gep-1709.md
mlavacca: Yeah, it makes sense!

geps/gep-1709.md
mlavacca: Do you have other words in mind that could suit better our needs here?

geps/gep-1709.md
```diff
@@ -231,6 +231,25 @@ move on to [certification](#certification) to report the results.
...
+
+The certification process runs against an implementation using a specific mode
+(e.g., `standard`, `enhanced`) that is defaulted to `standard`. The final
+report will contain such a mode in the modes field. A "mode" is intended to capture
+situations where a Gateway API implementation may have different features and capabilities
+depending on how it is deployed (e.g. an implementation might be deployed with an external,
+OR internal load balancer, and have different capabilities depending on the mode chosen).
+Every certification report `modes` array contains only one element. Reports can be merged
+into one another to provide a single report that accounts for multiple modes, given the same
+Gateway API version.
```
youngnick, line 244: One thing that's not clear from this is where these modes are defined - if they are definable by each implementation, then we should say that explicitly here. (It would make sense to me that each implementation could define their own set of `mode`s.)

geps/gep-1709.md
```diff
@@ -231,6 +231,25 @@ move on to [certification](#certification) to report the results.
...
+omni-comprehensive one, as described [below](#multiple-reports-merge).
```
youngnick, line 245: This seems like a fragment?

geps/gep-1709.md
youngnick: I put a note about this above, but if I'm right and the implementation can define the modes, then maybe a note here as well that says something like "remember that the mode values are arbitrary and chosen by the implementation" or something.

We could also make that clearer by making the names more clearly arbitrary - like calling the two modes `plain` and `with-the-lot` or something silly.

youngnick (commented):
I love this approach, had a couple of language clarifications though.

geps/gep-1709.md
```diff
@@ -233,22 +233,22 @@ move on to [certification](#certification) to report the results.
...
-into one another to provide a single report that accounts for multiple modes, given the same
-Gateway API version.
-omni-comprehensive one, as described [below](#multiple-reports-merge).
+The certification process runs against an implementation using a specific mode.
+The final report will contain such a mode in the modes field. A "mode" is intended
+to capture situations where a Gateway API implementation may have different features
+and capabilities depending on how it is deployed (e.g. an implementation might be
+deployed with an external, OR internal load balancer, and have different capabilities
+depending on the mode chosen). The modes are implementation-specific, and no upstream
+mode is defined, except for `standard`, the default one used in case no mode is specified.
```
mlavacca, line 242: I'm not 100% sure we should have a default mode. In case an implementation has a unique mode and is not interested in this API functionality, should we just leave the `modes` array empty instead of forcing a `standard` mode?

The more I think about it, the more convinced I get about leaving the array empty and not having a default `standard` mode.

@shaneutt @youngnick any thoughts on this regard?

geps/gep-1709.md
mlavacca: yep, fixed 👍

geps/gep-1709.md
mlavacca: > I put a note about this above, but if I'm right and the implementation can define the modes, then maybe a note here as well that says something like "remember that the mode values are arbitrary and chosen by the implementation" or something.

I agree that we need to specify this aspect; I added it [here](https://github.com/kubernetes-sigs/gateway-api/blob/342075e387f4b6534921be5b172dad734b705e7b/geps/gep-1709.md#implementation-mode).

as suggested by Nick, I changed the modes to `plain` and `with-the-lot`, to better highlight this aspect.

geps/gep-1709.md
mlavacca: It should be addressed now (see the other comment).

geps/gep-1709.md
youngnick: Yeah, I think leaving the array empty by default is fine - but we just need to be clear that if there are multiple results with the same version, they must have different values for the `modes` array. (The same actually goes for having multiple versions with the same mode).

That means that we need to specify the `modes` array as optional with a defined behavior.

If we're going to do this though, we should put the compulsory field (`version`) first, and the optional field (`mode`) second in the docs.

For `mode`, we should use language like:

Implementations MAY have multiple entries for the same `version` if and only if each entry has a distinct `modes` array. Having two results with the same `version` and identical `modes` arrays is invalid.

geps/gep-1709.md
```diff
@@ -266,6 +285,25 @@ VERSION file contains the semver of the latest release and is manually bumped at
 release time. The script hack/verify-all.sh will be updated to ensure the
 generated `.go` file is up to date with the VERSION file.
 
+### Multiple reports merge
```
robscott, line 288: Merging seems like it could get pretty complex. If different `modes` have different supported features are we just going to take the intersection of them? What about failing tests?

For GKE, I think I'd rather have different files for each unique mode. In our case it would be per GatewayClass, and we already have [8 unique GatewayClasses that we support](https://cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities).

I think I'd rather take one step closer to [upstream k8s conformance reporting structure](https://github.com/cncf/k8s-conformance/tree/master/v1.28) here and end up with something like this:

```
- v1.0
  - gke
    - README.md
    - mode-1-report.yaml
    - mode-2-report.yaml
```

We could potentially take this one step further and pull some of the product specific information out of the report and into a separate `product.yaml` (eg: https://github.com/cncf/k8s-conformance/blob/master/v1.28/gke/PRODUCT.yaml).

I don't know, I'm probably missing some of the goals here, so let me know if these recommendations don't make sense.

geps/gep-1709.md
robscott: If we go with my suggestion below, I'd recommend that mode is not a list and just a single `string`, and we default to the Gateway Class Name used in the test run as `mode` unless a different value is specified.

robscott (commented):
Thanks for the work on this @mlavacca!

shaneutt:
@mlavacca where are we at with this one, is there any assistance required to keep it moving forward?

geps/gep-1709.md
mlavacca: > Merging seems like it could get pretty complex. If different modes have different supported features are we just going to take the intersection of them? What about failing tests?

The idea is to have a section for each mode running a specific implementation version. If different modes have different supported features, we create two different sections, one per mode, along with the `supportedFeatures` and the results (successful, failed, skipped tests).

>For GKE, I think I'd rather have different files for each unique mode. In our case it would be per GatewayClass, and we already have [8 unique GatewayClasses that we support](https://cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities).
>
>I think I'd rather take one step closer to [upstream k8s conformance reporting structure](https://github.com/cncf/k8s-conformance/tree/master/v1.28) here and end up with something like this:
>
>```- v1.0
>  - gke
>    - README.md
>    - mode-1-report.yaml
>    - mode-2-report.yaml
>```
> We could potentially take this one step further and pull some of the product specific information out of the report and into a separate product.yaml (eg: https://github.com/cncf/k8s-conformance/blob/master/v1.28/gke/PRODUCT.yaml).

I get your point, and I agree that with the proposed solution we can end up having a gigantic `yaml` file. I like the approach of having a "summary report file" and a set of reports for each mode/version pair (or something similar).
I'll put something together that takes into account these considerations.

mlavacca:
> @mlavacca where are we at with this one, is there any assistance required to keep it moving forward?

Sorry for the period of inactivity on this one. I've restarted working on it 👍

geps/gep-1709.md
mlavacca: @robscott it took some time, but eventually, I came up with a proposal very similar to what you suggested. Let me know what you think 😄

mlavacca:
@shaneutt @robscott @youngnick I made some significant changes to the GEP. Could you please take another look?

geps/gep-1709.md
```diff
@@ -555,7 +508,81 @@ Creating a pull request to add the `ConformanceReport` will start the
...
+General information about the Acme/operator project
+
+## Support Matrix
+
+| API channel | Implementation version | Mode | Profiles | Extended Features | Report |
+|-------------|------------------------|------|----------|-------------------|--------|
+|             |                        |      |          |                   |        |
+|             |                        |      |          |                   |        |
+|             |                        |      |          |                   |        |
+
```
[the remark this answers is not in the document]
mlavacca: The goal of the PR is to define different modes and versions for a specific report. This aspect comes with a challenge: how to organize multiple reports for an implementation against the same Gateway API version? For example, with Gateway API, acme/operator may have many reports, with different operator versions and modes. I'd like to introduce here a way of organizing reports (originally proposed by @robscott in a [comment](https://github.com/kubernetes-sigs/gateway-api/pull/2391#discussion_r1327857784)), along with a table of contents. I think the table of contents is something beneficial at this point, but if we want to keep it for a subsequent GEP update, I have no problem at all 👍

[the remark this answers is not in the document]
shaneutt: Alright, no that's fine let's leave it then. Thanks for the explanation :vulcan_salute:

shaneutt (approved):

geps/gep-1709/index.md
```diff
@@ -370,84 +389,58 @@ implementation:
...
-gatewayAPIVersion: v0.6.2
-gatewayAPIChannel: standard
+  version: v0.9.0
+date: "2023-02-28 20:29:41+00:00"
+apiVersion: v0.8.0
+apiChannel: standard
+mode: plain
 profiles:
-  - name: tcp
+  profiles:
```
tao12345666333, line 400: Are there nested profiles here?

```
profiles:
  profiles:
    - name: tcp
    - name: http
```

mlavacca: Nope, it's an error in the yaml. Good catch!

geps/gep-1709/index.md
```diff
@@ -231,6 +231,22 @@ move on to [certification](#certification) to report the results.
...
 
+### Implementation mode
+
+The certification process runs against an implementation using a specific mode.
+The final report will contain such a mode in the modes field. A "mode" is intended
+to capture situations where a Gateway API implementation may have different features
+and capabilities depending on how it is deployed (e.g. an implementation might be
+deployed with an external, OR internal load balancer, and have different capabilities
+depending on the mode chosen). The modes are implementation-specific, and no upstream
+mode is defined, except for `plain`, the default one used in case no mode is specified.
```
robscott, line 242: Why not just call the default mode "default"? It seems at least theoretically plausible that "plain" would have a real meaning for some implementations that was also non-default.

geps/gep-1709/index.md
```diff
@@ -245,8 +261,8 @@ there are CRDs with different channels, the certification fails specifying that
 it's not possible to run the tests as there are different Gateway API channels
 installed in the cluster. If all the Gateway API `CRD`s have the same version
 and the same channel, the tests can be run and the detected version and channel
-will be set in the `GatewayAPIVersion` and `gatewayAPIChannel` fields of the
-final report. Furthermore, the suite must run all the experimental tests when
+will be set in the `apiVersion` and `apiVersionChannel` fields of
```
robscott, line 264: I think I actually prefer `GatewayAPIVersion` and `GatewayAPIChannel` to avoid any confusion here.

geps/gep-1709/index.md
```diff
@@ -555,7 +508,81 @@ Creating a pull request to add the `ConformanceReport` will start the
...
+```
+
+The main folder `conformance/reports` contains a set of sub-folders, each for a
+Gateway API version. Implementors will create their project folder in all the Gateway
+API version folders for which they have a supported implementation. The implementors
+folder contains the following set of files:
+
+* `README.md`: contains general information about the implementation and a table
+  with some information, as shown below, to give users a quick overview of
+  the implementation details.
```
robscott, line 542: I think a key goal of `README.md` would be to provide information about how the conformance tests were run and how someone could replicate the results.

geps/gep-1709/index.md
```diff
@@ -555,7 +508,81 @@ Creating a pull request to add the `ConformanceReport` will start the
...
+
+## Support Matrix
+
+| API channel | Implementation version | Mode | Profiles | Extended Features | Report |
+|-------------|------------------------|------|----------|-------------------|--------|
+|             |                        |      |          |                   |        |
+|             |                        |      |          |                   |        |
+|             |                        |      |          |                   |        |
+
+```
```
robscott, line 564: I'm not sure that a table will scale well here, especially for implementations that have a very long list of extended features. It also feels like it's really just duplicating information from conformance reports which could be difficult to keep in sync.

I think I'd prefer modeling this file more closely off of what upstream, like https://github.com/cncf/k8s-conformance/tree/master/v1.29/k3s for example - a brief description of the implementation + instructions for how to recreate.

robscott (commented):
Thanks @mlavacca! This mostly LGTM. One nit - a high-level description on the PR would have helped me remember the context on this a bit faster.

geps/gep-1709.md
robscott: Thanks @mlavacca! Sorry for the delay on reviewing this!

geps/gep-1709/index.md
mlavacca: > I'm not sure that a table will scale well here, especially for implementations that have a very long list of extended features. It also feels like it's really just duplicating information from conformance reports which could be difficult to keep in sync.

yes, It makes sense, we'll end up having duplicate information from reports. Nevertheless, I think this kind of table would be pretty useful as a table of contents. What do you think about getting rid of the `profiles` and `extended features` columns? this way we have a single line for each report, with a quick link; something like this:

| API channel | Implementation version | Mode | Report |
|-------------|------------------------|-------------------|--------|
|             |                        |      |          |
|             |                        |      |          |

> I think I'd prefer modeling this file more closely off of what upstream, like https://github.com/cncf/k8s-conformance/tree/master/v1.29/k3s for example - a brief description of the implementation + instructions for how to recreate.

👍  I've added a "To reproduce" section in the README.md

geps/gep-1709/index.md
mlavacca: yep, I've replaced `plain` with `default`

mlavacca:
> Thanks @mlavacca! This mostly LGTM. One nit - a high-level description on the PR would have helped me remember the context on this a bit faster.

Sorry about that, I've added a high-level description in the PR 👍

geps/gep-1709/index.md
mlavacca: fixed

geps/gep-1709/index.md
robscott: I think that's great, thanks!

robscott:
Thanks @mlavacca!

/hold cancel
/lgtm

PR #2395, opened 2023-09-14
Fix multiple mirror conformance test

### LINKED ISSUE #2394: HTTPRouteRequestMultipleMirrors conformance test has invalid backendRef
**What happened**:

Running the `HTTPRouteRequestMultipleMirrors` conformance test fails because the second mirrored `backendRef` is in a different namespace: https://github.com/kubernetes-sigs/gateway-api/blob/759e09e3566ee51a2683e1d6f93edd28ef3c6a1e/conformance/tests/httproute-request-multiple-mirrors.yaml#L25

This failure is expected because:

1. there is no `another-namespace` namespace with the `infra-backend-v3` service
2. even if there was, there is no `ReferenceGrant` to allow the cross ns request

As per `BackendObjectRef` doc:

 If the referent cannot be found, this BackendRef is invalid and must be dropped from the Gateway. The controller must ensure the “ResolvedRefs” condition on the Route status is set to status: False and not configure this backend in the underlying implementation.
  
 If there is a cross-namespace reference to an existing object that is not allowed by a ReferenceGrant, the controller must ensure the “ResolvedRefs” condition on the Route is set to status: False, with the “RefNotPermitted” reason and not configure this backend in the underlying implementation.

Run the conformance test with any compliant gateway implementation.

frankbu:
**What type of PR is this?**

/kind bug
/area conformance

Fixes #2394

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-request-multiple-mirrors.go
+++ b/conformance/tests/httproute-request-multiple-mirrors.go
@@ -64,7 +64,7 @@ var HTTPRouteRequestMultipleMirrors = suite.ConformanceTest{
 					},
 					{
 						Name:      "infra-backend-v3",
-						Namespace: "another-namespace",
+						Namespace: ns,
 					},
 				},
 				Namespace: ns,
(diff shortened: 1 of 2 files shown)


robscott:
Thanks for catching and fixing this @frankbu!

/lgtm
/approve
/cherry-pick release-0.8

Issue #2400, opened 2023-09-15
Provide install manifests that do not contain the gateway-api admission server


arkodg:
**What would you like to be added**:
A release artifact similar to https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.8.1 which contains the installation manifests for Gateway API CRDs and not the Gateway API admission server and related resources

**Why this is needed**:

Installing the Gateway API admission server is optional and all its functionality has been replaced by CEL validations
https://kubernetes.io/blog/2023/08/29/gateway-api-v0-8/#cel-validation

shaneutt:
Is this kind of a part of https://github.com/kubernetes-sigs/gateway-api/issues/2319? We _could_ merge them as it would seem reasonable for one body of work to take care of both at the same time. We don't _have_ to though either.

arkodg:
@shaneutt yes its related to #2319, but imo having a smaller work item is much easier to track and assign to a community contributor

looks like https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/build-install-yaml.sh needs to be updated to accomplish this goal, the question is around naming and content - should we just create a `experimental-crds-install.yaml` & `standard-crds-install.yaml` and keep the existing `experimental-install.yaml` & `standard-install.yaml` as is ?

shaneutt:
> @shaneutt yes its related to #2319, but imo having a smaller work item is much easier to track and assign to a community contributor

:+1: 

> looks like https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/build-install-yaml.sh needs to be updated to accomplish this goal, the question is around naming and content - should we just create a `experimental-crds-install.yaml` & `standard-crds-install.yaml` and keep the existing `experimental-install.yaml` & `standard-install.yaml` as is ?

I'm not sure we should start having all these extras laying around, I think we should just remove the admission server bits and be done with it, `v0.8.0` (the last release intended to "support" the webhook) is branched and any future webhook development should probably be isolated to that branch.

arkodg:
cool @shaneutt, I can take care of this
/assign @arkodg

robscott:
>I'm not sure we should start having all these extras laying around, I think we should just remove the admission server bits and be done with it, v0.8.0 (the last release intended to "support" the webhook) is branched and any future webhook development should probably be isolated to that branch.

[...]

PR #2401, opened 2023-09-15
rm webhook resources from installation yaml

arkodg:
**What type of PR is this?**

Fixes # https://github.com/kubernetes-sigs/gateway-api/issues/2400

```release-note
Resources related to the validating webhook such as the `gateway-system` namespace and the `gateway-api-admission-server` deployment have been removed from the installation manifests, in favor of CEL based Validations that are built into the CRD definition. These are still available in `webhook-install.yaml` in case you would like to optionally install them.
```

### DIFF
--- a/hack/build-install-yaml.sh
+++ b/hack/build-install-yaml.sh
@@ -39,7 +39,15 @@ cat << EOF >> release/standard-install.yaml
 #
 EOF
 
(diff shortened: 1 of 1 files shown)


robscott:
Thanks @arkodg!

/lgtm
/approve

Issue #2403, opened 2023-09-15
Conformance Tests: Base manifests for ingress use cases assume HTTP support


shaneutt:
**What happened**:

When trying to enable the conformance profiles enabled test suite for an implementation that **does NOT** support `HTTP`, I discovered that we default to that protocol for the base manifests:

https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/base/manifests.yaml#L26
https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/utils/suite/suite.go#L131

This default doesn't make sense for an ingress implementation which doesn't support `HTTPRoute`.

**What you expected to happen**:

We need a way for the default `cSuite.Setup()` to work on ingress implementations that may not support the common `HTTPRoute` type. Perhaps it would be reasonable to template the protocol and allow implementations to override for a protocol that they _do_ support.

PR #2412, opened 2023-09-18
Conformance tests for static Gateway addresses

### LINKED ISSUE #2035: Addresses

shaneutt:
**What type of PR is this?**

/area conformance
/kind test

Going into GA we have an extended field for supplying static addresses for Gateways, but no conformance tests for that feature. This patch adds the test, while also tweaking some of the related condition reasons (see the explanation in the `CHANGELOG.md`).

Resolves #2035

```release-note
NONE
```

> **Warning**: The way we're injecting the address pools into the test is bad. I don't like it. However, I think at this point with GA in the next few weeks, trying to make this better would require some kind of fundamental change to the test suite which we don't have time for. If you have any great ideas for alternatives that are simple, that would be lovely. If your PR comments are going to be something like "man we really need proper templating in the test suite" please note that this will probably have to wait for later.

### DIFF
--- a/conformance/tests/gateway-static-addresses.go
+++ b/conformance/tests/gateway-static-addresses.go
@@ -1,0 +1,177 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayStaticAddresses)
+}
+
+// GatewayStaticAddresses tests the implementation's support of deploying
+// Gateway resources with static addresses, or in other words addresses
+// provided via the specification rather than relying on the underlying
+// implementation/network to dynamically assign the Gateway an address.
+//
+// Running this test against your own implementation is currently a little bit
+// messy, as at the time of writing we didn't have great ways to provide the
+// test suite with things like known good, or known bad addresses to run the
+// test with (as we obviously can't determine that for the implementation).
+//
+// As such, if you're trying to enable this test for yourself and you're getting
+// confused about how to provide addresses, you'll actually do that in the
+// conformance test suite BEFORE you even set up and run your tests. Make sure
+// you populate the following test suite fields:
+//
+//   - suite.UsableNetworkAddresses
+//   - suite.UnusableNetworkAddresses
+//
+// With appropriate network addresses for your network environment.
+var GatewayStaticAddresses = suite.ConformanceTest{
+	ShortName:   "GatewayStaticAddresses",
+	Description: "A Gateway in the gateway-conformance-infra namespace should be able to use previously determined addresses.",
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportGatewayStaticAddresses,
+	},
+	Manifests: []string{
+		"tests/gateway-static-addresses.yaml",
+	},
+	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
+		gwNN := types.NamespacedName{
+			Name:      "gateway-static-addresses",
+			Namespace: "gateway-conformance-infra",
+		}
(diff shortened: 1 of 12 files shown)


conformance/tests/gateway-static-addresses.go
```diff
@@ -0,0 +1,118 @@
...
+		err = s.Client.Patch(ctx, updatedGW, client.MergeFrom(currentGW))
+		require.NoError(t, err, "failed to patch Gateway: %v", err)
+		kubernetes.NamespacesMustBeReady(t, s.Client, s.TimeoutConfig, namespaces)
+		kubernetes.GatewayMustHaveLatestConditions(t, s.Client, s.TimeoutConfig, gwNN)
+
+		t.Logf("verifying that the Gateway %s/%s is accepted and programmed with the usable static address assigned", gwNN.Namespace, gwNN.Name)
+		err = s.Client.Get(ctx, gwNN, currentGW)
+		require.NoError(t, err, "error getting Gateway: %v", err)
+		kubernetes.GatewayAcceptedConditionMustEqual(t, currentGW, metav1.ConditionTrue, v1beta1.GatewayReasonAccepted)
+		kubernetes.GatewayProgrammedConditionMustEqual(t, currentGW, metav1.ConditionTrue, v1beta1.GatewayReasonProgrammed)
```
mlavacca, line 100: For completeness, we should check all the positive-polarity conditions, hence we need to check `ResolvedRefs` as well.

conformance/utils/kubernetes/apply.go
```diff
@@ -51,12 +53,84 @@ type Applier struct {
...
+		// relevant addresses (usable, or unusable ones) in the test suite, and those
+		// addresses will be injected into the Gateway and the placeholders removed.
+		//
+		// A special "test/fake-invalid-type" can be provided as well in the test to
+		// explicitly trigger a failure to support a type. If an implementation ever
+		// comes along actually trying to support that type, I'm going to be very
+		// cranky.
+		//
+		// Note: I would really love to find a better way to do this kind of
+		// thing in the future.
```
mlavacca, line 104: 100% agree, it would be great to have an established way of templating the resources.

conformance/utils/kubernetes/apply.go
```diff
@@ -266,3 +340,20 @@ func getContentsFromPathOrURL(fs embed.FS, location string, timeoutConfig config
 	}
 	return bytes.NewBuffer(b), nil
 }
+
+// convertGatewayAddrsToPrimitives converts a slice of Gateway addresses and
+// to a slice of primite types and then returns them as a []interface{} so that
```
mlavacca, line 345:
```suggestion
// convertGatewayAddrsToPrimitives converts a slice of Gateway addresses
// to a slice of primitive types and then returns them as a []interface{} so that
```

conformance/utils/kubernetes/helpers.go
```diff
@@ -239,6 +239,24 @@ func NamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig config.T
...
+
+// GatewayProgrammedConditionMustEqual requires that a provided Gateway resource includes the Programmed
+// condition present that matches the provided status and reason.
+func GatewayProgrammedConditionMustEqual(t *testing.T, gw *v1beta1.Gateway, status metav1.ConditionStatus, reason v1beta1.GatewayConditionReason) {
+	require.True(t,
+		findConditionInList(t, gw.Status.Conditions, string(v1beta1.GatewayConditionProgrammed), string(status), string(reason)),
+		"Programmed status of %s with reason %s was NOT present on Gateway %s/%s as required", status, reason, gw.Namespace, gw.Name,
+	)
+}
+
```
mlavacca, line 259: Similary to what we have for [HTTPRoute](https://github.com/kubernetes-sigs/gateway-api/blob/6d8927ddeba5f08a22fd27a1fa5aa0b8b28959ee/conformance/utils/kubernetes/helpers.go#L550), can't we create a `GatewayMustHaveCondition` and pass the `Programmed` and `Accepted` conditions?

conformance/tests/gateway-static-addresses.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: gateway-static-addresses
+  namespace: gateway-conformance-infra
+spec:
+  gatewayClassName: "{GATEWAY_CLASS_NAME}"
+  addresses:
+  # this address type is intentionally unsupported to test that implementations
+  # are indicating Gateways with unsupported addresses are not accepted.
+  # the test will remove this address once it confirms it triggers a status of
+  # Accepted==False.
+  - type: "test/fake-invalid-type"
```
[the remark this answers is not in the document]
shaneutt: You mean like, `IPAddress` but then don't add a valid address?

I believe validation should already cover us here, for instance with `v0.8.1` if I try to apply:

```yaml
---
kind: Gateway
metadata:
  name: gateway-static-addresses
  namespace: gateway-conformance-infra
spec:
  gatewayClassName: blixt
  addresses:
  - type: "IPAddress"
    value: "batman"
  listeners:
  - name: http
    port: 8080
    protocol: HTTP
```

I will receive:

```console
* spec.addresses[0].value: Invalid value: "batman": spec.addresses[0].value in body must be of type ipv4: "batman"
```

Let me know though if I'm misunderstanding something?

conformance/tests/gateway-static-addresses.go
shaneutt: Added :+1:

conformance/utils/kubernetes/helpers.go
shaneutt: I really liked these new ones because they had a simpler interface, however for consistency reasons I switched as requested. :+1:

shaneutt:
This is now passing in Blixt, albeit I had to add several hacks to deal with the fact that our test suite is currently expecting if you do `Gateway` you must support `HTTP` (filed an issue for this in #2403). Ready for review.

robscott:
Thanks @shaneutt!

/lgtm

PR #2426, opened 2023-09-25
conformance: Add v0.8.1 conformance report for Cilium

sayboras:
**What type of PR is this?**

/kind documentation

NA

```release-note
NONE
```

### DIFF
--- a/conformance/reports/v0.8.1/cilium-cilium.yaml
+++ b/conformance/reports/v0.8.1/cilium-cilium.yaml
@@ -1,0 +1,47 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+date: "2023-09-25T14:21:25Z"
+gatewayAPIVersion: v0.8.1
+implementation:
+  contact:
+    - https://github.com/cilium/community/blob/main/roles/Maintainers.md
+  organization: cilium
+  project: cilium
+  url: github.com/cilium/cilium
+  version: v1.15.0-pre.1
+kind: ConformanceReport
+profiles:
+  - core:
+      result: success
+      statistics:
+        Failed: 0
+        Passed: 29
+        Skipped: 0
+      summary: ""
+    extended:
+      result: success
+      statistics:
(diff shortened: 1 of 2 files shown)


conformance/reports/v0.8.1/cilium-cilium.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+date: "2023-09-25T14:21:25Z"
+gatewayAPIVersion: v0.8.1
+implementation:
+  contact:
+    - https://github.com/cilium/community/blob/main/roles/Maintainers.md
+  organization: cilium
+  project: cilium
+  url: github.com/cilium/cilium
+  version: main
```
shaneutt, line 10: So this version field is meant to indicate the version of YOUR implementation, just wanted to double check: did you actually want this to be `main` branch, or did you want to associate it with a specific cilium version?

sayboras: I am trying to keep Cilium main branch as closed as possible with Gateway API latest release. As Cilium is having different release cycles, it's only available in main branch. 

If it's not idea to have main version in conformance report, I am happy to wait till we cut the next Cilium release.

shaneutt: How long will that be? If that's like days away, then maybe yes it would be better to wait for that, but if that's months out or something maybe just pin to a specific SHA instead of a branch so it's more definitive?

sayboras: There are around 3 more months for Cilium 1.15 dev cycle. Thanks for your pointer, I will update to hash commit shortly.

mlavacca (approved):
Thanks @sayboras, this LGTM.

/approve

robscott:
Thanks @sayboras!

/lgtm
/approve

PR #2428, opened 2023-09-25
Add gwctl to gateway-api repo

gauravkghildiyal:
**What type of PR is this?**
/kind feature

Import `gwctl` packages from https://github.com/gauravkghildiyal/gwctl

Fixes #

```release-note
Add gwctl command-line tool which provides a better way to view and manage policies (https://gateway-api.sigs.k8s.io/geps/gep-713/#kubectl-plugin)
```


gwctl/pkg/policymanager/manager.go
```diff
@@ -0,0 +1,345 @@
...
+	allPolicies, err := fetchPolicies(ctx, p.dc, p.policyCRDs)
+	if err != nil {
+		return err
+	}
+	for _, unstrucutredPolicy := range allPolicies {
+		policy, err := PolicyFromUnstrucutred(unstrucutredPolicy, p.policyCRDs)
+		if err != nil {
+			return err
+		}
+		p.policies[unstrucutredPolicy.GetNamespace()+"/"+unstrucutredPolicy.GetName()] = policy
```
robscott, line 77: This may benefit from a separate function like `p.AddPolicy(unstructuredPolicy)`

gwctl/pkg/policymanager/manager.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+type ObjRef struct {
+	Group     string `json:",omitempty"`
+	Kind      string `json:",omitempty"`
+	Name      string `json:",omitempty"`
+	Namespace string `json:",omitempty"`
+}
+
+func PolicyFromUnstrucutred(u unstructured.Unstructured, policyCRDs map[PolicyCrdID]PolicyCRD) (Policy, error) {
```
robscott, line 205:
```suggestion
func PolicyFromUnstructured(u unstructured.Unstructured, policyCRDs map[PolicyCrdID]PolicyCRD) (Policy, error) {
```

gwctl/pkg/resources/backends/backends.go
```diff
@@ -0,0 +1,282 @@
...
+type describeView struct {
+	Group                    string                                                        `json:",omitempty"`
+	Kind                     string                                                        `json:",omitempty"`
+	Name                     string                                                        `json:",omitempty"`
+	Namespace                string                                                        `json:",omitempty"`
+	DirectlyAttachedPolicies []policymanager.ObjRef                                        `json:",omitempty"`
+	EffectivePolicies        map[string]map[policymanager.PolicyCrdID]policymanager.Policy `json:",omitempty"`
+}
+
+func PrintDescribeView(ctx context.Context, params *types.Params, backendsList []unstructured.Unstructured) {
```
robscott, line 240: Non-blocking nit: To simplify the use of this code as a library it may be best to separate code into different packages. For example, one package or set of packages intended to be used as a lib, and a separate package that is more unique to gwctl inputs and outputs (like this describe view here).

I'm not really sure what the ideal code structure is here yet, so this may be worth filing a follow up issue for more discussion once this one merges.

gwctl/pkg/resources/policies/policies.go
```diff
@@ -0,0 +1,120 @@
...
+
+func Print(params *types.Params, policies []policymanager.Policy) {
+	sort.Slice(policies, func(i, j int) bool {
+		a := fmt.Sprintf("%v/%v", policies[i].Unstructured().GetNamespace(), policies[i].Unstructured().GetName())
+		b := fmt.Sprintf("%v/%v", policies[j].Unstructured().GetNamespace(), policies[j].Unstructured().GetName())
+		return a < b
+	})
+
+	tw := tabwriter.NewWriter(params.Out, 0, 0, 2, ' ', 0)
+	row := []string{"POLICYNAME", "POLICYKIND", "TARGETNAME", "TARGETKIND"}
```
robscott, line 40:
```suggestion
	row := []string{"POLICY NAME", "POLICY KIND", "TARGET NAME", "TARGET KIND"}
```

gwctl/pkg/resources/policies/policies.go
```diff
@@ -0,0 +1,120 @@
...
+
+func PrintCRDs(params *types.Params, policyCRDs []policymanager.PolicyCRD) {
+	sort.Slice(policyCRDs, func(i, j int) bool {
+		a := fmt.Sprintf("%v/%v", policyCRDs[i].CRD().GetNamespace(), policyCRDs[i].CRD().GetName())
+		b := fmt.Sprintf("%v/%v", policyCRDs[j].CRD().GetNamespace(), policyCRDs[j].CRD().GetName())
+		return a < b
+	})
+
+	tw := tabwriter.NewWriter(params.Out, 0, 0, 2, ' ', 0)
+	row := []string{"CRD_NAME", "CRD_GROUP", "CRD_KIND", "CRD_INHERITED", "CRD_SCOPE"}
```
robscott, line 63:
```suggestion
	row := []string{"NAME", "GROUP", "KIND", "POLICY TYPE", "SCOPE"}
```

gwctl/pkg/resources/policies/policies.go
```diff
@@ -0,0 +1,120 @@
...
+	tw.Write([]byte(strings.Join(row, "\t") + "\n"))
+
+	for _, policyCRD := range policyCRDs {
+		row := []string{
+			policyCRD.CRD().Name,
+			policyCRD.CRD().Spec.Group,
+			policyCRD.CRD().Spec.Names.Kind,
+			fmt.Sprintf("%v", policyCRD.IsInherited()),
```
robscott, line 71: Would recommend using a string value here of "Inherited" or "Direct" instead

gwctl/pkg/types/params.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"sigs.k8s.io/gateway-api/gwctl/pkg/common"
+	"sigs.k8s.io/gateway-api/gwctl/pkg/policymanager"
+)
+
+type Params struct {
```
robscott, line 33: Nit: I'm not really sure what the best name for this is, but `Params` doesn't seem quite right. It seems like this is a mix of 3 k8s clients, an internal data structure, and an io writer. Are there other ways we can split this up? How broadly is this type used? Are there common combinations that only need a subset of what's in here?

robscott (commented):
This is very cool! Thanks for all the work on this @gauravkghildiyal! There's a lot here, but took a high level pass and had some nits + questions.

gwctl/pkg/common/consts.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+const (
+	GatewayPolicyLabelKey = "gateway.networking.k8s.io/policy"
```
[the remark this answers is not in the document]
gauravkghildiyal: Nice suggestion. Created https://github.com/kubernetes-sigs/gateway-api/pull/2434 to address this. Will update references once that merges.

gwctl/pkg/common/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"strings"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+type YamlString string
+
+var YamlStringTransformer = cmp.Transformer("YamlLines", func(s YamlString) []string {
```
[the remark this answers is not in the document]
gauravkghildiyal: No that's a good question. I should move the comment from within the function to outside the function. (Will update)

This is just a "good to have" thing which helps provide neat diffs during test. Idea being, split long blobs of yaml text by each line. When the diff is generated, it will clearly point out diffs per line (as oppposed to saying that: "somethings different in this large string, now go find what the difference is")

gwctl/pkg/policymanager/manager.go
gauravkghildiyal: Oh no sorry about that. Fixed. Wow how did you catch that amazing.

gwctl/pkg/policymanager/manager.go
gauravkghildiyal: Nice idea. Thanks changed.

gwctl/pkg/resources/policies/policies.go
gauravkghildiyal: Great idea. Changed.

gwctl/pkg/resources/backends/backends.go
gauravkghildiyal: I definitely resonate with your thoughts here. There's several improvements that could be made to make this more usable as a library. I created https://github.com/kubernetes-sigs/gateway-api/issues/2435 to track the work for this.

gwctl/pkg/cmd/describe.go
```diff
@@ -0,0 +1,148 @@
...
+		} else {
+			gwc, err := gatewayclasses.Get(context.TODO(), params, args[1])
+			if err != nil {
+				panic(err)
+			}
+			gwClasses = []gatewayv1beta1.GatewayClass{gwc}
+		}
+		gatewayclasses.PrintDescribeView(context.TODO(), params, gwClasses)
+	case "backend", "backends":
+		var backendsList []unstructured.Unstructured
```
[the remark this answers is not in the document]
gauravkghildiyal: Fair enough. Let me change to just "Services" for now.

gwctl/pkg/types/params.go
gauravkghildiyal: This is definitely a valid commment. I decided to not remain blocked on this initial friction and start with something simple like Params to get moving.
- At the moment, this type is kinda used like a global "context" which is consumed by majority of the functions.
- Initially, I just had three K8s clients here (and hence I initially called it Clients). But as things evolved, and now we have two more things.
- Similar to your other suggestion about `describeViews`, this too definitely requires some code refactor.

I would completely understand the point of view to hold this PR until we fix this (Just say so without any further explanation and I'll be glad to address it here). But I was planning on taking this up separately as part of some code refactoring which is focused. I'll need some time to think the solution here.

gwctl/pkg/common/consts.go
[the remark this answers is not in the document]
gauravkghildiyal: Done and rebased!

gwctl/pkg/resources/backends/backends.go
gauravkghildiyal: Refactored things around. This should be slightly better than the previous iteration. Function names need some refinement but the packages `effectivepolicy` and `policymanager` provide constructs that can be used as a library and are independent from the CLI part.

gwctl/pkg/types/params.go
gauravkghildiyal: Refactored. Now we have a `CmdParams` but that's limited in scope to only being required by the CLI interface (and not need when library packages are used)

robscott:
> You may consider naming the binary `kubectl-<something>` so its a "kubectl plugin"

Maybe `kubectl-gw` then?

robscott:
Thanks for all the work on this @gauravkghildiyal! I took some time to test this out today and have some high level non-blocking feedback.

1. I'd love to see this be a drop-in replacement for kubectl when reading (get/describe) Gateway API resources, so ideally this would be a superset of kubectl capabilities. With that said, if we end up having to recreate too many things, that's also not ideal. My main point of confusion here is that we're using `get` as a command, but it's not currently a drop-in replacement for `kubectl get`. If we wanted to get there, I think we'd need the following:
    * Support for all resources for both get and describe commands
    * Show at least the same set of columns, ideally more
    * Support for yaml and json output
    * _Maybe_ support for more advanced features like label selection, etc
    * The longer this list gets, the less I'm sure about the overall goal 😄 
2. It would be great to more gracefully handle errors. For example, if one or more Gateway API CRDs are not found, we could just tell the user they need to install the CRD(s) that are missing.
3. Similar to 1, but potentially more achievable in the short term, it would be great to include more information in our describe output. Right now the primary value seems to be policy connections, but it would be useful if it could include that + much of the other useful information included by `kubectl describe`.

I don't think any of these are blockers, but before we go too much further with this, it would probably be helpful to sketch out a user guide for the full set of functionality we want this to be able to accomplish, and then we can split up the work into smaller pieces so it can grow over time from this foundation.

I'm going to go ahead and approve what's here, because it works and looks like a good foundation, but would be very interested in a vision for how we want to build on top of this going forward.

/approve

gauravkghildiyal:
Thanks a lot for the review and great constructive feedback @robscott. Everything that you highlighted is definitely something that I can start addressing ASAP. 

I agree that kubectl doesn't natively provide an extension mechanism in a way in which we can reuse it for certain things which ends up with us having to duplicate things. I will create a user guide on what I think we can/should support so others can start contributing as well.

I had been working on some improvements on the "usage as a library" aspects which I'll send separately.

(Rebasing to resolve merge conflict)

shaneutt (approved):
Awesome, thanks for adding this!

I think adding it here in the repo is a good idea, as it will keep things in sync.

youngnick:
yeah, I really like this to get us started, but I think we should spend a little time post-GA and talk about what we want this to go - tbh the more I think about it, the more I think we may be better off with two similar but distinct tools - `gwctl` which is a pretty straight `kubectl` replacement for `get` and `describe`, and a separate `gw` kubectl plugin that does the same things but with a slightly different interface. Typing `kubectl gw get all` or something seems slightly more readable than `kubectl gwctl get all`.

/lgtm

Issue #2416, opened 2023-09-21
Listener isolation or cascading behavior


youngnick:
When talking about Listener acceptance and merging behavior as part of #2288, we discovered that there was a particularly important part of the spec that we had not clarified enough, and that this has created a behavior variance between implementations.

I (@youngnick) summarized this as follows:
> That concept is important because, for better or for worse, it has always been intended that the traffic-matching between Listeners and Routes is a two stage process. First you match a Listener, then you look for matching routes. In the case that there's a wildcard hostname, it was definitely not intended that a request could match both (for example, the `cafe.example.com` and `*.example.com` Listeners.) That is, having different values at all for hostname has been intended to make the Listeners distinct and unmergeable. We certainly have not done enough to explain that this is the intention, and I'm sorry for that.

That is, it was the original intention in the spec that, if there is a precise hostname match in one Listener, and a wildcard match in another, that traffic that matches the precise hostname will arrive _only at Routes bound to that Listener_.

Another way to put this is that if a precise hostname match is present, then a wildcard match should _never_ also match that precise hostname.

Because we were not clear about this in the spec originally, many implementations have implemented their rules such that requests that match a precise hostname may, if there are no matching Routes for the request, match Routes attached to the wildcard hostname.

To use an example, here is a Gateway that @pleshakov supplied in #2288.

```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gateway
spec:
  gatewayClassName: nginx
  listeners:
  - name: example
    port: 80
    protocol: HTTP
    hostname: "*.example.com"
  - name: cafe-http
    port: 80
    protocol: HTTP
    hostname: "cafe.example.com"
```

And here are some Routes that illustrate the config that _may_ result in cascading behavior. I've left out hostnames on the HTTPRoutes so that only the Listener hostname is relevant.

```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: coffee
spec:
  parentRefs:
  - name: gateway
    sectionName: cafe-http
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /coffee
    backendRefs:
    - name: backend
      port: 3000
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  name: tea
spec:
  parentRefs:
  - name: gateway
    sectionName: example 
  rules:
  - matches:
    - path:
[...]

PR #2429, opened 2023-09-25
Adding new PartiallyInvalid condition for Routes

### LINKED ISSUE #1696: Define how to reflect partial validity of routing rules in the HTTPRoute status

robscott:
**What type of PR is this?**
/kind feature

This adds a new condition along with corresponding guidance for how to handle Routes that are partially invalid.

Note: I can't think of any way to write conformance tests for this condition since the most likely way to get into this state is when an implementation can't support a feature configured in a Route. 

Fixes #1696.

```release-note
New "PartiallyInvalid" condition has been added for all Route types.
```

/hold for consensus

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -298,7 +298,7 @@ const (
 	// This condition indicates whether the route has been accepted or rejected
 	// by a Gateway, and why.
 	//
-	// Possible reasons for this condition to be true are:
+	// Possible reasons for this condition to be True are:
 	//
 	// * "Accepted"
 	//
@@ -349,15 +349,17 @@ const (
 	// are incompatible filters present on a route rule (for example if
 	// the URLRewrite and RequestRedirect are both present on an HTTPRoute).
 	RouteReasonIncompatibleFilters RouteConditionReason = "IncompatibleFilters"
+)
 
+const (
 	// This condition indicates whether the controller was able to resolve all
 	// the object references for the Route.
 	//
-	// Possible reasons for this condition to be true are:
+	// Possible reasons for this condition to be True are:
 	//
 	// * "ResolvedRefs"
 	//
-	// Possible reasons for this condition to be false are:
+	// Possible reasons for this condition to be False are:
 	//
 	// * "RefNotPermitted"
 	// * "InvalidKind"
@@ -392,6 +394,43 @@ const (
 	// Route's rules has a reference to a resource with an app protocol that
 	// is not supported by this implementation.
 	RouteReasonUnsupportedProtocol RouteConditionReason = "UnsupportedProtocol"
+)
+
+const (
+	// This condition indicates that the Route contains a combination of both
+	// valid and invalid rules.
+	//
+	// When this happens, implementations MUST take one of the following
+	// approaches:
+	//
+	// 1) Drop Rule(s): With this approach, implementations will drop the
+	//    invalid Route Rule(s) until they are fully valid again. The message
+	//    for this condition MUST start with the prefix "Dropped Rule" and
+	//    include information about which Rules have been dropped. In this
+	//    state, the "Accepted" condition MUST be set to "True" with the latest
+	//    generation of the resource.
+	// 2) Fall Back: With this approach, implementations will fall back to the
+	//    last known good state of the entire Route. The message for this
+	//    condition MUST start with the prefix "Fall Back" and include
+	//    information about why the current Rule(s) are invalid. To represent
+	//    this, the "Accepted" condition MUST be set to "True" with the
+	//    generation of the last known good state of the resource.
+	//
+	// Reverting to the last known good state should only be done by
+	// implementations that have a means of restoring that state if/when they
+	// are restarted.
+	//
+	// This condition MUST NOT be set if a Route is fully valid, fully invalid,
+	// or not accepted. By extension, that means that this condition MUST only
+	// be set when it is "True".
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "UnsupportedValue"
+	//
+	// Controllers may raise this condition with other reasons, but should
+	// prefer to use the reasons listed above to improve interoperability.
+	RouteConditionPartiallyInvalid RouteConditionType = "PartiallyInvalid"
 )
 
 // RouteParentStatus describes the status of a route with respect to an


shaneutt (commented):
I think having these two options and saying that you MUST do one or the other makes sense, because we don't want to have completely undefined behavior for partial invalidity between implementations.

I suppose conformance tests shouldn't be too hard, it will be like opting into a "feature" from a test triggering point of view, except you MUST opt for one or the other behaviors.

It seems like really late in the cycle for `v1.0.0` we're seeing a few more of these "do one or the other" kind of things crop up (also see #2416), which definitely gives me a bit of pause.

I think at a high level I'm on board because this is an improvement for users over the previously undefined behavior, I definitely think we should make sure a greater number of community members look it over too though.

/cc @arkodg @sunjayBhatia @LiorLieberman @candita @gcs278 @mlavacca @kflynn @michaelbeaumont @rainest @Xunzhuo

youngnick:
I would really have preferred that we could pick one or the other, but from discussion, it seems that we can't, so this is the next best thing. 

LGTM, but will leave it for others to review.

apis/v1beta1/shared_types.go
```diff
@@ -353,6 +355,46 @@ const (
...
+	//
+	// 1) Drop Rule(s): With this approach, implementations will drop the
+	//    invalid Route Rule(s) until they are fully valid again. The message
+	//    for this condition MUST start with the prefix "Dropped Rule" and
+	//    include information about which Rules have been dropped. In this
+	//    state, the "Accepted" condition MUST be set to "True" with the latest
+	//    generation of the resource.
+	// 2) Fall Back: With this approach, implementations will fall back to the
+	//    last known good state of the entire Route. The message for this
+	//    condition MUST start with the prefix "Fall Back" and include
```
howardjohn, line 373: Using message as a typed/specified thing seems a bit odd/unexpected/nonstandard?

apis/v1beta1/shared_types.go
```diff
@@ -353,6 +355,46 @@ const (
...
+	// This condition MUST NOT be set if a Route is fully valid, fully invalid,
+	// or not accepted.
+	//
+	// Possible reasons for this condition to be True are:
+	//
+	// * "UnsupportedValue"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "PartiallyInvalid"
```
howardjohn, line 391:
```
type: PartiallyInvalid
status: False
reason: PartiallyInvalid
```

Can you explain this? "Its not PartiallyInvalid because of reason PartiallyInvalid"...?

mikemorris:
My biggest concern with this proposal is that I haven't seen any implementations explicitly state their desire to support it yet - I agree that we need to do something to resolve the discrepancy between spec documentation and [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/), but I'm not sure if we want to add this to resolve a somewhat-abstract problem before determining this is the solution that some implementations actually want.

apis/v1beta1/shared_types.go
youngnick: I agree, but there we don't have another way to indicate this information, and we really need to have _some_ guidance around this before GA. Ideas welcomed, I guess.

apis/v1beta1/shared_types.go
youngnick: This is a good point - in this case, it probably makes sense to have the reason be "AllRulesValid" or something when the `status` is `False`, but we should also include guidance that the implementation MAY leave this Condition out if it would be set to `status: False`.

sunjayBhatia: Looks like above there is a statement:

```
	// This condition MUST NOT be set if a Route is fully valid, fully invalid,
	// or not accepted.
```

so maybe we can omit the guidance on `Possible reasons for this condition to be False are`?

sunjayBhatia (commented):
Looks like most concerns I would have are captured by comments from others 👍🏽 

Another outcome of this spec change/resolving the linked issue IMO would be agreeing on a plan on how to test this in conformance, especially given we have a `MUST` statement added (`When this happens, implementations MUST take one of the following approaches`). If we can come up with route rules that pass CEL validation but are universally invalid (not just implementation-specific invalid) my first thought is to add testing like:
- Add a new core feature, `SupportsPartiallyInvalidRoutes` (name tbd) and test that makes sure the status is set to true when there is a partially invalid route, likely ignoring reason/message
- Add a couple new extended features, `SupportsPartiallyInvalidRoutes[DropRule|Fallback]`
- Implementations would opt-in to the test they support
- We can modify a route to becoming partially invalid and test the invalid rule is dropped or the old configuration is respected, depending on what test is run
- Status checked for the right values depending on what test is run

I can include this comment in the original issue for this feature or a new one if that makes sense instead!

sunjayBhatia:
> My biggest concern with this proposal is that I haven't seen any implementations explicitly state their desire to support it yet - I agree that we need to do something to resolve the discrepancy between spec documentation and [GEP-1364](https://gateway-api.sigs.k8s.io/geps/gep-1364/), but I'm not sure if we want to add this to resolve a somewhat-abstract problem before determining this is the solution that at least some implementations actually want, especially given that [dropping invalid rules is not without risks](https://github.com/kubernetes-sigs/gateway-api/issues/1696#issuecomment-1679804122).

In Contour we haven't explicitly stated in our Gateway API support/in the community discussions about this, but representing partial validity has been a bit of an issue in our own CRD/status/etc., so I'm enthusiastic about trying to do this in a standardized way with representing in status what is going to happen to invalid routes.

apis/v1beta1/shared_types.go
```diff
@@ -353,6 +355,46 @@ const (
 	RouteReasonBackendNotFound RouteConditionReason = "BackendNotFound"
 )
 
+const (
+	// This condition indicates that the Route contains a combination of both
+	// valid and invalid rules.
+	//
+	// When this happens, implementations MUST take one of the following
```
sunjayBhatia, line 362: I like that the two options described below are explicitly stated. In Contour we've had a somewhat inconsistent mix in our own CRD throughout history of "dropping" invalid rules, programming 5xx responses for invalid rules while preserving valid ones, attempting to program as much of an invalid rule as we can but giving warnings/errors in status about invalid attributes of a rule, etc., which all make for a bit of an inconsistent UX. Explicitly giving users feedback in status about what is happening is a step forward IMO.

apis/v1beta1/shared_types.go
sunjayBhatia: Maybe adding new condition types with this info would work? Another condition that would only be set if PartiallyInvalid is also set

```
- type: PartiallyInvalid
  status: true
  reason: UnsupportedValue
  message: ...
- type: PartiallyInvalid[DropRouteRule|RouteFallBack]
  status: true
  reason: Implementation[DropsRulesOnPartialInvalidity|FallsBackOnPartialInvalidity]
  message: ...
```

apis/v1beta1/shared_types.go
robscott: Thanks, good catch, removed entirely because it should never be set when it's false as @sunjayBhatia pointed out.

robscott: (And also added a note that made it clear that the condition must not be set when it is false).

apis/v1beta1/shared_types.go
robscott: Yeah, agree this is a bit wonky, but I don't love the alternatives here. @sunjayBhatia's idea to add additional conditions to more clearly reflect this state may actually be something we need to do in the future, but I'd like to see if we can survive without adding more conditions first. If we try it out and determine that it would really be best to add additional conditions, we can do that later.

robscott:
> Another outcome of this spec change/resolving the linked issue IMO would be agreeing on a plan on how to test this in conformance, especially given we have a `MUST` statement added (`When this happens, implementations MUST take one of the following approaches`). If we can come up with route rules that pass CEL validation but are universally invalid (not just implementation-specific invalid) my first thought is to add testing like:
> 
> * Add a new core feature, `SupportsPartiallyInvalidRoutes` (name tbd) and test that makes sure the status is set to true when there is a partially invalid route, likely ignoring reason/message
> * Add a couple new extended features, `SupportsPartiallyInvalidRoutes[DropRule|Fallback]`
> * Implementations would opt-in to the test they support
> * We can modify a route to becoming partially invalid and test the invalid rule is dropped or the old configuration is respected, depending on what test is run
> * Status checked for the right values depending on what test is run

Agree that this would make a good follow up. I created #2458 based on this.

apis/v1beta1/shared_types.go
robscott: Yeah, completely agree, I think many implementations have inconsistent behavior here.

robscott:
Thanks to everyone for the feedback! I think I've resolved everything now, PTAL.

robscott:
I think we've got enough consensus here, going to remove the hold so the next LGTM merges this.

/hold cancel

PR #2320, opened 2023-08-18
Fix CEL validations for parentRefs and use 'in' operator in some CEL expressions

gauravkghildiyal:
**What type of PR is this?**
/kind feature
/kind bug
/kind cleanup
/kind test

**What this PR does / why we need it**:
Fix CEL validations for parentRefs and use 'in' operator in some CEL expressions

**Which issue(s) this PR fixes**:

Addresses comments:
* https://github.com/kubernetes-sigs/gateway-api/pull/2245#discussion_r1297452357
* https://github.com/kubernetes-sigs/gateway-api/pull/2245#discussion_r1284877438

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

---

# Explanation for future reference:
The validations in this PR are meant to cover the webhook validations: https://github.com/kubernetes-sigs/gateway-api/blob/73f20f4ae4ab46a4a2ea52718fdbdef1a6726376/apis/v1beta1/validation/common.go#L28

The validation is split apart into Standard and Experimental (as `Port` field is currently only available in `Experimental`). Each of them have two validations:

### 1. sectionName must be specified when parentRefs includes 2 or more references to the same parent

CEL Playground:
* [Standard](https://playcel.undistro.io/?content=H4sIAAAAAAAAA72Sy26EMAxFf8WTBQ9pikSWSKP%2BwWy7qYSiwaWoPCKSUacq%2FfcmARoS0lFXZeUc29c3Jp%2BkYpKRgghsX4rnHtT3APU4XHkBMd5Yx1vMs2Gs4zkJ8Nb0lco9NVWN8of2rENFu493l5elzgjOLliWqkCi0Mn%2FngRAjgRvfEQhmqFfLpyxtk14foS5yCK6orUbgOeZMQunE3C6xFGkufa5YBPOVLtZqAkVtWqJDb2DAYdXJpSxzLlTCtMEPtQT4jh11R0VGlKhQZWtgqm74zLsUNkID40itz10D7%2FLsWMP6caWjR5D0PP8%2B6IFXqR6F2c1fF3zBv1lyTsFGlBwu1NPzT1N071sEjI%2BP7zt2IMeu%2Fkru3K6L3dWXYAcr%2BjvVpWk5Osbr%2FEnjj0EAAA%3D)
[...]

PR #2433, opened 2023-09-27
Requiring SectionName or Port to be specified for distinct parents

robscott:
**What type of PR is this?**
/kind cleanup

Previously parents could be distinct if one specified a port and one specified a listener. Now they either both need to specify a distinct SectionName, both need to specify a distinct Port, or both. This PR replaces https://github.com/kubernetes-sigs/gateway-api/pull/2350.

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2326
Fixes https://github.com/kubernetes-sigs/gateway-api/issues/1925

```release-note
Experimental Channel: For ParentRefs to be considered distinct, they either both need to specify a distinct SectionName, both need to specify a distinct Port, or both.
```

**Note For Reviewers:**
CEL is complicated, and this took some time for me to fully understand. Fortunately @gauravkghildiyal pointed me back to his well documented PR that added the initial validation here https://github.com/kubernetes-sigs/gateway-api/pull/2320. 

Essentially, this requires 2 lines of validation per channel. The first one ensures that whenever ParentRefs refer to the same object (Group + Kind + Namespace + Name), they all must have the same fields set. The second one ensures that all values within ParentRefs are unique.

In this case we only need to change the first one, and we needed to ensure that `has(p1.sectionName)` must equal `has(p2.sectionName)` and similar for port in experimental channel. (For reference `p1` and `p2` are different parentRefs in the list in both this example and the CEL validation code). This change actually enables us to simplify our logic a bit so the key part of the condition in experimental channel looks like this:

```
((!has(p1.sectionName) || p1.sectionName == '') == (!has(p2.sectionName) || p2.sectionName == '') && (!has(p1.port) || p1.port == 0) == (!has(p2.port) || p2.port == 0))
```

Although the behavior in standard channel is unchanged, we can also simplify that logic similarly:

```
((!has(p1.sectionName) || p1.sectionName == '') == (!has(p2.sectionName) || p2.sectionName == '')
```

A huge thanks to @gauravkghildiyal for the help with this and also to @youngnick who contributed the actual changes to the guidance in the spec I'm including here.

/cc @youngnick @gauravkghildiyal

### DIFF
--- a/apis/v1beta1/shared_types.go
+++ b/apis/v1beta1/shared_types.go
@@ -177,10 +177,25 @@ type CommonRouteSpec struct {
 	// This API may be extended in the future to support additional kinds of parent
 	// resources.
 	//
-	// It is invalid to reference an identical parent more than once. It is
-	// valid to reference multiple distinct sections within the same parent
-	// resource, such as two separate Listeners on the same Gateway or two separate
-	// ports on the same Service.
+	// ParentRefs must be _distinct_. This means either that:
+	//
+	// * They select different objects.  If this is the case, then parentRef
+	//   entries are distinct. In terms of fields, this means that the
+	//   multi-part key defined by `group`, `kind`, `namespace`, and `name` must
+	//   be unique across all parentRef entries in the Route.
+	// * They do not select different objects, but for each optional field used,
+	//   each ParentRef that selects the same object must set the same set of
+	//   optional fields to different values. If one ParentRef sets a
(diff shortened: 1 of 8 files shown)


gauravkghildiyal:
Thanks @robscott . This LGTM (I verified as well)

That's a brilliant idea to compare boolean values. Definitely shortens thing up a lot.

youngnick:
/lgtm

Thanks @robscott for sorting this one out.

PR #2434, opened 2023-09-27
Add policy identifier label to policy API spec

gauravkghildiyal:
**What type of PR is this?**
/kind feature

Add the constant `gateway.networking.k8s.io/policy` used to identify a policy CRD to the policy related specs file. 

Fixes #

```release-note
NONE
```

/assign @robscott

### DIFF
--- a/apis/v1alpha2/policy_types.go
+++ b/apis/v1alpha2/policy_types.go
@@ -16,6 +16,19 @@ limitations under the License.
 
 package v1alpha2
 
+const (
+	// PolicyLabelKey is the label whose presence identifies a CRD that the
+	// Gateway API Policy attachment model. The value of the label SHOULD be one
+	// of the following:
+	//  - A label value of "Inherited" indicates that this Policy is inheritable.
+	//    An example of inheritable policy is one which if applied at the Gateway
+	//    level would affect all attached HTTPRoutes and their respective
+	//    Backends.
+	//  - A label value of "Direct" indicates that the policy only affects the
+	//    resource to which it is attached and does not affect it's sub resources.
+	PolicyLabelKey = "gateway.networking.k8s.io/policy"
+)
+
 // PolicyTargetReference identifies an API object to apply a direct or
 // inherited policy to. This should be used as part of Policy resources
 // that can target Gateway API resources. For more information on how this


apis/v1alpha2/policy_types.go
```diff
@@ -16,6 +16,18 @@ limitations under the License.
 
 package v1alpha2
 
+const (
+	// PolicyLabelKey is the label whose presence identifies a CRD as a policy
+	// CRD. The value of the label SHOULD be one of the following:
```
shaneutt, line 21: Are we just saying SHOULD here because we don't expect to be able to conformance test it?

robscott, line 21:
```suggestion
	// PolicyLabelKey is the label whose presence identifies a CRD that
	// the Gateway API Policy attachment model. The value of the label 
	// MUST be one of the following:
```

apis/v1alpha2/policy_types.go
```diff
@@ -16,6 +16,18 @@ limitations under the License.
...
+	//  - A label value of "inherited" indicates that this Policy is inheritable.
```
robscott, line 22: I've gotten this wrong before (I think I'm also responsible for this time), but in general Kubernetes prefers for values of annotations to be capitalized. I'd recommend only specifying the upper case values in the case but maybe we can still have a case-insensitive match in tools like gwctl.

```suggestion
	//  - A label value of "Inherited" indicates that this Policy is inheritable.
```

apis/v1alpha2/policy_types.go
```diff
@@ -16,6 +16,18 @@ limitations under the License.
...
+	//    An example of "inheritable" policy is one which if applied at the
+	//    Gateway level would affect all attached HTTPRoutes and their respective
+	//    Backends.
+	//  - A label value of "direct" indicates that the policy only affects the
```
robscott, line 26:
```suggestion
	//  - A label value of "Direct" indicates that the policy only affects the
```

apis/v1alpha2/policy_types.go
robscott: Yeah good catch, `MUST` is probably correct here.

robscott (commented):
Thanks for adding this @gauravkghildiyal! Just some nits but great to see this formally defined in our API types.

apis/v1alpha2/policy_types.go
gauravkghildiyal: Sure that sounds good. Changed.

apis/v1alpha2/policy_types.go
gauravkghildiyal: Applied. Thanks!

gauravkghildiyal: Thanks Shane and Rob. Changed to MUST

apis/v1alpha2/policy_types.go
gauravkghildiyal: Applied

robscott:
Thanks @gauravkghildiyal!

/lgtm
/approve

PR #2438, opened 2023-09-27
conformance: fix flakes in mirror tests

howardjohn:
**What type of PR is this?**

/kind test
/area conformance

The test in isolation is fine, but combined with the other mirror test we have problems.

1. Test 1 sets up /mirror -> v1. Passes, starts cleanup
2. Test 2 sets up /mirror -> v2. Sends request to /mirror, gets a 200. It sent to v1, though, since cleanup wasn't done.
3. We look for logs in v2 forever, never find them. :-(

This fixes things by making the two use isolated paths

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-request-multiple-mirrors.go
+++ b/conformance/tests/httproute-request-multiple-mirrors.go
@@ -49,11 +49,11 @@ var HTTPRouteRequestMultipleMirrors = suite.ConformanceTest{
 		testCases := []http.ExpectedResponse{
 			{
 				Request: http.Request{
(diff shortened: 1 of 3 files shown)


howardjohn:
> I think what's actually happening is when multi-mirror test runs the config from the previous mirror-test hasn't been cleaned up yet, so instead of sending the /mirror request to (v1, mirror v2, mirror v3), it's only sending the request to (v1, mirror v2) so the test fails waiting to see the mirror request to v3, which never happened.



yep, my bad.

PR #2442, opened 2023-09-28
clarify attachment by two policies to the same object when one uses a sectionName

maleck13:
**What type of PR is this?**

/kind documentation

Clarifies how two policies targeting the same resource should work when one uses a `sectionName` and the other does not

created based on Slack chat https://kubernetes.slack.com/archives/CR0H13KGA/p1695713117110159

None

**Release Note**
```release-note
clarify policy attachment by two of the same policy types when using section names.
```

### DIFF
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -1465,7 +1465,7 @@ and the policy implementation should record a `resolvedRefs` or similar Conditio
 
 When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
 the one with a `sectionName` is more specific, and so will have all its settings applied to the named target. The less-specific Policy will
-still be applied but MUST not affect the named target.
+still be applied but MUST not affect the named target i.e. it will be applied to all other sections which have not been targeted by another Policy 
 
 Note that the `sectionName` is currently intended to be used only for Direct Policy Attachment when references to
 SectionName are actually needed. Inherited Policies are always applied to the entire object. 
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -1463,7 +1463,7 @@ type RouteRule struct {
 If a `sectionName` is specified, but does not exist on the targeted object, the Policy must fail to attach,
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
-When multiple Policies of the same type target the same object, one with a `sectionName`, and one without, the policy with a `sectionName` is more specific, and so will have its entire `spec` applied to the named section. The less-specific Policy will still have its `spec` applied to the target but MUST not affect the named section i.e. the non specific policy will have its `spec` applied to all other sections of the target which have not been targeted by another Policy. 
+When multiple Policies of the same type target the same object, one with a `sectionName` and one without, the more specific policy (i.e., the one with a `sectionName`) will have its entire `spec` applied to the named section. The less specific policy will also have its `spec` applied to the target but MUST not affect the named section. The less specific policy will have its `spec` applied to all other sections of the target that are not targeted by any other more specific policies. 
 
 Note that the `sectionName` is currently intended to be used only for Direct Policy Attachment when references to
 SectionName are actually needed. Inherited Policies are always applied to the entire object. 
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -1463,9 +1463,7 @@ type RouteRule struct {
 If a `sectionName` is specified, but does not exist on the targeted object, the Policy must fail to attach,
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
-When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
-the one with a `sectionName` is more specific, and so will have all its settings applied to the named target. The less-specific Policy will
-still be applied but MUST not affect the named target i.e. it will be applied to all other sections which have not been targeted by another Policy 
+When multiple Policies of the same type target the same object, one with a `sectionName`, and one without, the policy with a `sectionName` is more specific, and so will have its entire `spec` applied to the named section. The less-specific Policy will still have its `spec` applied to the target but MUST not affect the named section i.e. the non specific policy will have its `spec` applied to all other sections of the target which have not been targeted by another Policy. 
 
 Note that the `sectionName` is currently intended to be used only for Direct Policy Attachment when references to
 SectionName are actually needed. Inherited Policies are always applied to the entire object. 
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -1464,8 +1464,8 @@ If a `sectionName` is specified, but does not exist on the targeted object, the 
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
 When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
-the one with a `sectionName` is more specific, and so will have all its settings apply. The less-specific Policy will
-not attach to the target.
+the one with a `sectionName` is more specific, and so will have all its settings applied to the named target. The less-specific Policy will
+still be applied but MUST not affect the named target.
 
 Note that the `sectionName` is currently intended to be used only for Direct Policy Attachment when references to
 SectionName are actually needed. Inherited Policies are always applied to the entire object. 


geps/gep-713.md
```diff
@@ -1338,8 +1338,8 @@ If a `sectionName` is specified, but does not exist on the targeted object, the
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
 When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
-the one with a `sectionName` is more specific, and so will have all its settings apply. The less-specific Policy will
-not attach to the target.
+the one with a `sectionName` is more specific, and so will have all its settings applied to the named target. The less-specific Policy will
+still be applied but should not affect the named target.
```
youngnick, line 1342:
```suggestion
still be applied but MUST not affect the named target.
```
This one isn't a recommendation, so I think we need to go with the stronger language.

youngnick (commented):
One small change, then this LGTM.

youngnick:
LGTM, will approve and hold for @robscott

/approve
/hold

geps/gep-713.md
```diff
@@ -1338,8 +1338,8 @@ If a `sectionName` is specified, but does not exist on the targeted object, the
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
 When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
-the one with a `sectionName` is more specific, and so will have all its settings apply. The less-specific Policy will
-not attach to the target.
+the one with a `sectionName` is more specific, and so will have all its settings applied to the named target. The less-specific Policy will
+still be applied but MUST not affect the named target.
```
robscott, line 1342: This is surprising complex to get right, although this is a great improvement, I think there's actually still some ambiguity here. Consider the following example:

* AcmePolicy A attaches to Gateway listener "http" and sets `foo` to `1`
* AcmePolicy B attaches to Gateway and sets `foo` to `2`, and `bar` to `3`

What should the computed policy for the "http" listener be?

1. {foo: 1}
2. {foo: 1, bar: 3}
3. {foo: 2, bar: 3}

1 might make sense if we say that each Directly Attached Policy can target exactly one level. IE AcmePolicy can only be used to target Gateway Listeners, other targets are invalid. The [current wording in GEP-713 ](https://gateway-api.sigs.k8s.io/geps/gep-713/#direct-policy-attachment)seems to imply that, but it's not clear how this interacts with SectionName:

>A Direct Policy Attachment is tightly bound to one instance of a particular Kind within a single namespace (or to an instance of a single Kind at cluster scope), and only modifies the behavior of the object that matches its binding.

>The number or scope of objects to be modified is limited or singular. Direct Policy Attachments must target one specific object.

2 might make sense if we say that each Directly Attached Policy can target either a full resource or a section within that resource. The GEP does not currently allow for that, but as noted above, it also doesn't have any wording that would prevent that.

One concern I have is that both 2 and 3 are starting to overlap with "Inherited Policy Attachment", with 2 effectively behaving as both policies set "defaults" and 3 behaving as both policies set "overrides". Importantly we've only allowed SectionName to be used with Direct Policy Attachment.

https://github.com/kubernetes-sigs/gateway-api/blob/3b87766a3708f02b1bf9b5a8da2305a0b3d3a1bd/apis/v1alpha2/policy_types.go#L62-L64

So all of the above actually leads me to think that 1 is the correct interpretation here (despite what I said in Slack). Interested in what others think here though. Looping in the people from the Slack thread for more discussion here:

/cc @arkodg @david-martin @frankbu @sanjaypujare @youngnick

robscott (commented):
Thanks for working on this @maleck13! Unfortunately I think this one ended up being more complex than I initially thought it would be, added a long comment with more detail.

geps/gep-713.md
arkodg: We've gone ahead with 2. In Envoy Gateway.
As you've pointed out 2. supports the case where user sets a default policy for the entire gateway and the same user can also set a policy for a specific section within the gateway to override behavior for that specific section.
This is a common use case and 2. makes this possible with the minimum number of resources ( policies) created to achieve this goal

arkodg, line 1342:
```suggestion
still be applied but MUST not affect the named target i.e. it will be applied to all other sections which have not been targeted by another Policy 
```

arkodg: @robscott is `foo` and `bar` part of the `AcmePolicy` spec ? or are they targets . (`foo` has also been used as targets as well as policy names in this doc, so im confused :) )

robscott: @arkodg good question, I updated my comment to be a bit more clear, they were both intended to be part of the spec.

arkodg: thanks @robscott , have to go with 1. entire `spec` is overridden for that section by `AcmePolicy A`

frankbu: I also tend to think that 1 is the intended behavior, but I think there are still some inconsistencies in the spec. For example:

> When multiple Policies of the same type target the same object, one with a sectionName specified, and one without

Note that it says `of the same type`. This implies they must both be Inherited Policy Attachments, but that goes against:

> Note that the sectionName is currently intended to be used only for Direct Policy Attachment

And there is also this:

> Note also that it's not intended that Direct and Inherited Policies should overlap, so this should only come up in exceptional circumstances.

So, it's really not very clear, how and when this applies, but I think the answer to the original question is still that a less specific policy applies to all other sections, just not the one that the more specific policy attaches to.

The question about whether the specific policy completely overrides the less specific one for the specific section is IMO yes (option 1).

robscott: Thanks for the responses! It sounds like we're on board with 1 here. Maybe to back out one more level, are we sure that we should allow Direct Attached Policies to concurrently target both a full resource and a section within that Resource? That seems to be the conclusion here, but I just want to make sure we're not adding unnecessary complexity here.

robscott: Also agree with @frankbu that we need to make it clear in the GEP and not just API Spec that SectionName is only for direct policies.

sanjaypujare:
> @robscott: GitHub didn't allow me to request PR reviews from the following users: david-martin, sanjaypujare.

Ack. Will take a look. (Hopefully this comment will add me as a reviewer.)

geps/gep-713.md
arkodg: > urce? That seems to be the conclusion here, but I just want to make sure we're not adding unnecessary complexity here.

@robscott yes can speak on behalf of Envoy Gateway, we think this is a common use case for a (default) Policy A to target the entire Gateway G (with Listeners X, Y, Z) and another Policy B to target only Listener X within Gateway G implicitly overriding the policy spec only for that specific section. This approach allows the user to achieve the goal of a global defaults with few overrides with minimal number of policy instantiations.

robscott: @arkodg That's starting to sound awfully similar to inherited policy attachment. Would it be better for us to open up SectionName to Inherited Policy Attachment or to make Direct Policy Attachment a bit less "Direct"?

arkodg: There is one use case where the DirectPolicyAttachment may want to only target a `sectionName`, which means the implementor may need to add an extra CEL validation in their CRD to make sure `sectionName` is not nil.

But for most cases, implementors may want to support policy hierarchy, but they might not want to support `Inherited Policy Attachment` which requires the CRD to have explicit `default` & `overrides` fields.

david-martin: +1 to option 1 being the computed policy, given it's direct policy attachment.

The wording in the PR supports that being the outcome.
If further clarification is helpful, the comment in https://github.com/kubernetes-sigs/gateway-api/pull/2442/files#r1346313279 could be added as an example.

Direct attached policy with an optional SectionName doesn't fit neatly with the idea of targeting a 'single object', given the potential for hierarchies within the object, but this wording clarifies the exception in my opinion.

> open up SectionName to Inherited Policy Attachment

If I understand this suggestion correctly, the inherited policy resource hierarchy would change to include SectionName versions of resource? 
e.g. GatewayClass > Gateway Namespace > Gateway > Gateway listener (via SectionName) > Route Namespace > Route > Route Rule (via SectionName) > etc...

I can see value in that.
It should play nicely with existing Direct attached policy with optional SectionName.
Kind of a 'best of both worlds', albeit with extra complexity on the inheritance side.
Perhaps it's something to be considered when a solid use case comes up that can't be solved with existing policy options?

youngnick: Thinking about this, I also support option 1, and think that we should include @robscott's example in the godoc.

maleck13: I may be missing this so looking to clarify. 

Option 1 looks correct to me also, and so the text suggested below is correct?
`still be applied but MUST not affect the named target i.e. it will be applied to all other sections which have not been targeted by another Policy `

arkodg: @maleck13 I think we should somehow insert `entire spec of the Policy` instead of `it` to add more clarity

maleck13: took another stab at this, let me know if it is clearer now

arkodg (approved):
thanks for adding this clarity !
/approve

geps/gep-713.md
sanjaypujare: I also support option 1. Also we should leave out Inherited Policy Attachment (with section names) to avoid the complexity. We can always add that later, right?

sanjaypujare:
> > @robscott: GitHub didn't allow me to request PR reviews from the following users: david-martin, sanjaypujare.
> 
> Ack. Will take a look. (Hopefully this comment will add me as a reviewer.)

Done. Looks good

sanjaypujare (approved):
LGTM

geps/gep-713.md
```diff
@@ -1337,9 +1337,7 @@ type RouteRule struct {
 If a `sectionName` is specified, but does not exist on the targeted object, the Policy must fail to attach,
 and the policy implementation should record a `resolvedRefs` or similar Condition in the Policy's status.
 
-When multiple Policies of the same type target the same object, one with a `sectionName` specified, and one without,
-the one with a `sectionName` is more specific, and so will have all its settings apply. The less-specific Policy will
-not attach to the target.
+When multiple Policies of the same type target the same object, one with a `sectionName`, and one without, the policy with a `sectionName` is more specific, and so will have its entire `spec` applied to the named section. The less-specific Policy will still have its `spec` applied to the target but MUST not affect the named section i.e. the non specific policy will have its `spec` applied to all other sections of the target which have not been targeted by another Policy.
```
frankbu, line 1340:
```suggestion
When multiple Policies of the same type target the same object, one with a `sectionName`, and one without, the more specific policy (i.e., the one with a `sectionName`) will have its entire `spec` applied to the named section. The less-specific Policy will also have its `spec` applied to the target but MUST not affect the named section. The non-specific policy will have its `spec` applied to all other sections of the target that are not targeted by any other more specific Policies. 
```

sanjaypujare: You say "less-specific Policy" in one place but "non-specific Policy" elsewhere. Should we be consistent and just say "less-specific" everywhere?

frankbu: Good point! "less-specific" should also not be hyphenated. I updated my suggestion, above.

frankbu (approved):
LGTM

robscott:
Thanks @maleck13!

/lgtm
/hold cancel

arkodg:
@maleck13 can you rebase ?

youngnick:
@maleck13 If you can rebase this one into the new format, we can merge - I'm about to start work on splitting up GEP-713, so it would be helpful to have this in first.

frankbu:
Why has thi suggested change has been ignored? https://github.com/kubernetes-sigs/gateway-api/pull/2442#discussion_r1367024696

maleck13:
@frankbu I seem to have missed that.  Updated now

robscott:
The current test failure looked like a flake. Adding another LGTM so we can get this one in.

/lgtm

robscott:
@maleck13 do you mind adding a release note to this PR? Once you do, I think this should merge.

/hold cancel

PR #2443, opened 2023-09-28
Check gateway positive-polarity conditions

### LINKED ISSUE #2415: conformance: always check gateway positive-polarity conditions

All the conformance tests that create a properly working Gateway, should always check the following three positive-polarity conditions are set to true, as stated [here](https://github.com/youngnick/gateway-api/blob/901c4d8d06d94aa4ef2ca3a11f2385cac7197e65/site-src/geps/gep-1364.md#proposed-changes-summary):
- `ResolvedRefs`
- `Accepted`
- `Programmed`

The `ResolvedRefs` is going to be always checked by #2247 .
Some tests check only `Accepted` (e.g., https://github.com/mlavacca/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-with-attached-routes.go#L114)
Some test check only `Programmed` (e.g., https://github.com/mlavacca/gateway-api/blob/8ab4e18ca73533793f8a5d4e8c4248626b4690bf/conformance/tests/gateway-secret-reference-grant-specific.go#L54)

**Why this is needed**:

We need to consistently check that all the valid gateways have all the three positive-polarity conditions always set to true.

levikobi:
**What type of PR is this?**

/kind test
/area conformance

Add a consistent check that all the valid gateways have all three positive-polarity conditions always set to true.

Fixes #2415 

```release-note
None
```

### DIFF
--- a/conformance/utils/kubernetes/helpers.go
+++ b/conformance/utils/kubernetes/helpers.go
@@ -24,6 +24,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -268,12 +269,14 @@ func MeshNamespacesMustBeReady(t *testing.T, c client.Client, timeoutConfig conf
 }
 
 // GatewayAndHTTPRoutesMustBeAccepted waits until:
-// 1. The specified Gateway has an IP address assigned to it.
-// 2. The route has a ParentRef referring to the Gateway.
-// 3. All the gateway's listeners have the ListenerConditionResolvedRefs set to true.
+//  1. The specified Gateway has an IP address assigned to it.
+//  2. The route has a ParentRef referring to the Gateway.
+//  3. All the gateway's listeners have the following conditions set to true:
+//     - ListenerConditionResolvedRefs
+//     - ListenerConditionAccepted
+//     - ListenerConditionProgrammed
 //
-// The test will fail if these conditions are not met before the
-// timeouts.
+// The test will fail if these conditions are not met before the timeouts.
 func GatewayAndHTTPRoutesMustBeAccepted(t *testing.T, c client.Client, timeoutConfig config.TimeoutConfig, controllerName string, gw GatewayRef, routeNNs ...types.NamespacedName) string {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


robscott:
Thanks @levikobi! This looks reasonable to me, have you run this against any implementation? Would be great to get some feedback from other conformance reviewers as well.

/cc @arkodg @mlavacca @sunjayBhatia @Xunzhuo

levikobi:
Hi @robscott
I ran it against envoy's gateway this time. Not all tests passed.
Curious about other implementations

conformance/utils/kubernetes/helpers.go
```diff
@@ -356,27 +370,39 @@ func WaitForGatewayAddress(t *testing.T, client client.Client, timeoutConfig con
...
+	var wg sync.WaitGroup
+	wg.Add(len(conditions))
 
-		for _, listener := range gw.Status.Listeners {
-			if !findConditionInList(t, listener.Conditions, condition.Type, string(condition.Status), condition.Reason) {
-				return false, nil
-			}
-		}
+	for _, condition := range conditions {
+		go func(condition metav1.Condition) {
```
sunjayBhatia, line 382: any particular reason to do this concurrently rather than just loop through the desired conditions and validate all 3 are present in one go?

sunjayBhatia (commented):
Modified Contour to set ResolvedRefs when everything is in fact resolved (not doing that on main at the moment) and looks like the tests are working/passing

conformance/utils/kubernetes/helpers.go
levikobi: Nothing special, honestly.
I guess I'll switch it to the way you suggested

levikobi:
> Modified Contour to set ResolvedRefs when everything is in fact resolved (not doing that on main at the moment) and looks like the tests are working/passing

Thanks @sunjayBhatia!

conformance/utils/kubernetes/helpers.go
robscott: I also would prefer the more efficient option that @sunjayBhatia suggested, but don't think it needs to block this.

robscott:
Thanks @levikobi!

/lgtm
/approve

conformance/utils/kubernetes/helpers.go
levikobi: Sure, I will follow up on it at issue #2459

PR #2444, opened 2023-09-28
[GEP-1911] Update BackendRef with GEP details

dprotaso:
/kind gep
/kind documentation

- This takes [GEP-1911](https://gateway-api.sigs.k8s.io/geps/gep-1911/) and updates the godoc for `BackendRef`
- Mention users can signal backend protocol using `ServicePort.protocol`

Fixes #

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -549,6 +549,41 @@ type GRPCRouteFilter struct {
 }
 
 // GRPCBackendRef defines how a GRPCRoute forwards a gRPC request.
+//
+// Note that when a namespace different than the local namespace is specified, a
+// ReferenceGrant object is required in the referent namespace to allow that
+// namespace's owner to accept the reference. See the ReferenceGrant
+// documentation for details.
+//
+// <gateway:experimental:description>
+//
+// When the BackendRef points to a Kubernetes Service, implementations SHOULD honor the
+// appProtocol field if it is set for the target Service Port.
+//
+// Implementations supporting appProtocol MUST recognize the Kubernetes Standard Application Protocols
+// defined in [KEP-3726]. This supports IANA standard service names and extra constants defined
+// in the KEP that have a prefix of "kubernetes.io/". Gateway API MAY define additional
+// constants with the prefix "gateway.networking.k8s.io/"
+//
+// If a Service appProtocol isn't specified an implementation MAY infer the backend
+// protocol through its own means. Implementations MAY infer the protocol from the
+// Route type referring to the backend Service.
+//
+// Implementations MAY support multiplexing TCP and UDP on the same port. Otherwise
+// implementations MUST set ResolvedRefs condition to False with the "UnsupportedProtocol"
+// Reason with a clear message that multiplexing is not supported.
+//
+// If a Route is not able to send traffic to the backend using the specified protocol then
+// the backend is considered invalid. Implementations MUST set ResolvedRefs condition to
+// False with the "UnsupportedProtocol".
+//
+// Implementations MAY support different combinations of protocol/appProtocol/Route Type.
+// See [GEP-1911] for a table.
+//
+// </gateway:experimental:description>
+//
+// [KEP-3726]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols
+// [GEP-1911]: https://gateway-api.sigs.k8s.io/geps/gep-1911/
 type GRPCBackendRef struct {
 	// BackendRef is a reference to a backend to forward matched requests to.
 	//
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -567,11 +567,6 @@ type GRPCRouteFilter struct {
 // protocol through its own means. Implementations MAY infer the protocol from the
 // Route type referring to the backend Service.
 //
-// Implementations MAY support Kubernetes Service BackendRefs that are multiplexing TCP
-// and UDP on the same port. Otherwise implementations MUST set the Route ResolvedRefs
-// condition to False with the "UnsupportedProtocol" Reason with a clear message that
-// multiplexing is not supported.
-//
 // If a Route is not able to send traffic to the backend using the specified protocol then
 // the backend is considered invalid. Implementations MUST set ResolvedRefs condition to
 // False with the "UnsupportedProtocol".
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -569,9 +569,10 @@ type GRPCRouteFilter struct {
 // protocol through its own means. Implementations MAY infer the protocol from the
 // Route type referring to the backend Service.
 //
-// Implementations MAY support multiplexing TCP and UDP on the same port. Otherwise
-// implementations MUST set ResolvedRefs condition to False with the "UnsupportedProtocol"
-// Reason with a clear message that multiplexing is not supported.
+// Implementations MAY support Kubernetes Service BackendRefs that are multiplexing TCP
+// and UDP on the same port. Otherwise implementations MUST set the Route ResolvedRefs
+// condition to False with the "UnsupportedProtocol" Reason with a clear message that
+// multiplexing is not supported.
 //
 // If a Route is not able to send traffic to the backend using the specified protocol then
 // the backend is considered invalid. Implementations MUST set ResolvedRefs condition to
--- a/apis/v1alpha2/grpcroute_types.go
+++ b/apis/v1alpha2/grpcroute_types.go
@@ -560,10 +560,8 @@ type GRPCRouteFilter struct {
 // When the BackendRef points to a Kubernetes Service, implementations SHOULD honor the
 // appProtocol field if it is set for the target Service Port.
 //
-// Implementations supporting appProtocol MUST recognize the Kubernetes Standard Application Protocols
-// defined in [KEP-3726]. This supports IANA standard service names and extra constants defined
-// in the KEP that have a prefix of "kubernetes.io/". Gateway API MAY define additional
-// constants with the prefix "gateway.networking.k8s.io/"
+// Implementations supporting appProtocol SHOULD recognize the Kubernetes Standard Application Protocols
+// defined in [KEP-3726].
 //
 // If a Service appProtocol isn't specified an implementation MAY infer the backend
 // protocol through its own means. Implementations MAY infer the protocol from the
@@ -578,13 +576,9 @@ type GRPCRouteFilter struct {
 // the backend is considered invalid. Implementations MUST set ResolvedRefs condition to
 // False with the "UnsupportedProtocol".
 //
-// Implementations MAY support different combinations of protocol/appProtocol/Route Type.
-// See [GEP-1911] for a table.
-//
 // </gateway:experimental:description>
 //
 // [KEP-3726]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols
-// [GEP-1911]: https://gateway-api.sigs.k8s.io/geps/gep-1911/
 type GRPCBackendRef struct {
 	// BackendRef is a reference to a backend to forward matched requests to.
 	//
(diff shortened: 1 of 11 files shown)


dprotaso:
cc @andrew-su

dprotaso:
What's the equivalent of `<gateway:experimental>` ? I didn't want to add it to the godoc otherwise it would have marked BackendRef experimental which it's not

robscott:
@dprotaso fortunately we actually added something for that recently:

https://github.com/kubernetes-sigs/gateway-api/blob/18506b70f36313bfa966e29d09971a6fe568ce18/apis/v1beta1/shared_types.go#L174-L176

dprotaso:
hmm.. it didn't update HTTPRoute in the experimental channel 🤔

dprotaso:
Random guess is the generator mutates the CRD - so we should be generating experimental types first (no mutations) then standard one)

dprotaso:
Oddly we call DeepCopy() so 🤔

dprotaso:
Following up - looks like the text is not updated on the HTTPRoute is because of how controller-tools works

Because `BackendRef` is inlined with `HTTPBackendRef` - the godoc present at the top level `HTTPBackendRef` is not present and neither is the godoc on `BackendRef`

For example this block doesn't appear in the CRD docs for HTTPRoute
https://github.com/kubernetes-sigs/gateway-api/blob/18506b70f36313bfa966e29d09971a6fe568ce18/apis/v1beta1/httproute_types.go#L1115-L1149

Instead what shows up in the CRD description is just the HTTPBackendRef godoc

https://github.com/kubernetes-sigs/gateway-api/blob/18506b70f36313bfa966e29d09971a6fe568ce18/apis/v1beta1/httproute_types.go#L1113-L1114


@robscott @shaneutt @youngnick not sure what we want here - do you have opinions?

If we are strict about wanting this documentation to appear on the crd descriptions we'll have to duplicate the godoc and add it to the top of `HTTPBackendRef`

dprotaso:
If we're otherwise good with the docs as they are on the types then feel free to drop the `/hold`

apis/v1beta1/shared_types.go
```diff
@@ -243,6 +243,40 @@ type PortNumber int32
...
 // documentation for details.
+//
+// <gateway:experimental:description>
+//
+// When the BackendRef points to a Kubernetes Service implementations MUST allow
+// the backend protocol to be specified by:
+//   - setting the protocol field on the Service's ServicePort
+//   - setting the protocol field on the Service's related Endpoint/EndpointSlice's EndpointPort
+//   - setting the appProtocol field on the Service's ServicePort
+//   - setting the appProtocol field on the Service's related Endpoint/EndpointSlice's EndpointPort
```
robscott, line 254: Note that at least in the case of Service, the values of the Protocol and AppProtocol fields are copied verbatim to Endpoints and EndpointSlices. Of course users could create custom EndpointSlices with different values but I don't think we want to honor or support that.

In general, I think this addition to the spec may be saying more than we need to. A lot of the content could potentially be replaced with something like this with the idea that implementers can read AppProtocol docs for more information:

```suggestion
// When a BackendRef points to a Kubernetes Service, implementations SHOULD
// honor the appProtocol field if it is set for the target Service Port.
```

I think that we're likely going to be stuck with "SHOULD" because I think AppProtocol has always been more of a "hint" than a requirement. I don't think we can change that retroactively.

apis/v1beta1/shared_types.go
```diff
@@ -243,6 +243,40 @@ type PortNumber int32
...
+//
+// If a Service appProtocol isn't specified an implementation MAY infer the backend
+// protocol through its own means. Implementations MAY infer the protocol from the
+// Route type referring to the backend Service.
+//
+// Implementations MAY support multiplexing TCP and UDP on the same port. Otherwise
+// implementations MUST set ResolvedRefs condition to False with the Reason UnsupportedProtocol
+// with a clear message that multiplexing is not supported.
+//
+// If a Route attached to a Gateway is not able to send traffic to the backend using
```
robscott, line 269: Why is this unique to a Gateway? Is the expected behavior different for mesh? (I don't actually know what the answer here should be)

/cc @howardjohn @keithmattix

geps/gep-1911.md
```diff
@@ -38,6 +38,8 @@ A Gateway implementation MUST recognize the Kubernetes Standard Application Prot
 
 Thus when a `*Route` points to a Kubernetes Service resource the backend protocol for each port can be specified by
 
+- setting the `protocol` field on the Kubernetes `Service`
+- setting the `protocol` field on the `Endpoint`/`EndpointSlice` object associated with a Kubernetes `Service`
 - setting the `appProtocol` field on the Kubernetes `Service`
 - setting the `appProtocol` field on an `Endpoint`/`EndpointSlice` object associated with a Kubernetes `Service`
```
robscott, line 44: Sorry I missed this the first time around, but I think we should just focus on the values that are set on a Service (longer comment above).

robscott (commented):
Thanks @dprotaso! I actually ended up with more questions this time around as I dug into it a bit more, hopefully these make sense.

apis/v1beta1/shared_types.go
```diff
@@ -243,6 +243,40 @@ type PortNumber int32
...
+// the specified protocol then the backend is considered invalid. Implementations
+// MUST set ResolvedRefs condition to False with the Reason UnsupportedProtocol.
+//
+// Implementations MAY support different combinations of protocol/appProtocol/Route Type.
+// See [GEP-1911] for a table.
+//
+// </gateway:experimental:description>
+//
+// [KEP-3726]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols
+// [GEP-1911]: https://gateway-api.sigs.k8s.io/geps/gep-1911/
```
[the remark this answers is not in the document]
dprotaso: FWIW - it renders properly in godoc and on the Gateway API website reference page.

apis/v1beta1/shared_types.go
dprotaso: Honestly forgot about mesh - I can drop the Gateway ref in this sentence

dprotaso: Updated the language - I would be suprised if we treated mesh and non-mesh differently here

apis/v1beta1/shared_types.go
```diff
@@ -243,6 +243,40 @@ type PortNumber int32
...
 type BackendRef struct {
```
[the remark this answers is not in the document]
dprotaso: I duplicated the paragraphs to these types

apis/v1beta1/shared_types.go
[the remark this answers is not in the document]
youngnick: I think it's fine in this case, it makes the additional context easily available.

youngnick:
LGTM so I'll hit the approve button, @robscott up to you after that.

/approve

apis/v1beta1/shared_types.go
```diff
@@ -243,6 +243,40 @@ type PortNumber int32
...
+// defined in [KEP-3726]. This supports IANA standard service names and extra constants defined
+// in the KEP that have a prefix of "kubernetes.io/". Gateway API MAY define additional
+// constants with the prefix "gateway.networking.k8s.io/"
+//
+// If a Service appProtocol isn't specified an implementation MAY infer the backend
+// protocol through its own means. Implementations MAY infer the protocol from the
+// Route type referring to the backend Service.
+//
+// Implementations MAY support multiplexing TCP and UDP on the same port. Otherwise
+// implementations MUST set ResolvedRefs condition to False with the Reason UnsupportedProtocol
```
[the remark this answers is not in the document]
robscott: @dprotaso I probably should have avoided combining these items, I think you may have accepted the suggestion without answering the main question I had here.

>Where would this condition be set? When I think of multiplexing TCP and UDP on the same port, I'm thinking Gateway, but this guidance is on the Route level.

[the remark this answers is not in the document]
dprotaso: I guess to be pedantic - this should read 

```
Implementations MAY support Kubernetes Services BackendRefs that are multiplexing TCP and UDP on the same port.
```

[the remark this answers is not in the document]
dprotaso: And I'd expect the condition to be set on the Route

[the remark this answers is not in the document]
dprotaso: Updated language - PTAL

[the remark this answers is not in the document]
robscott: Oh interesting, I hadn't even thought of the case where the same BackendRef Port number could actually target both a TCP and UDP port. Are there any Route types that would actually support this? IMO, this feels like a sufficiently complex issue to leave out of this PR and instead file a follow up issie.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -549,6 +549,42 @@ type GRPCRouteFilter struct {
...
+//
+// <gateway:experimental:description>
+//
+// When the BackendRef points to a Kubernetes Service, implementations SHOULD honor the
+// appProtocol field if it is set for the target Service Port.
+//
+// Implementations supporting appProtocol MUST recognize the Kubernetes Standard Application Protocols
+// defined in [KEP-3726]. This supports IANA standard service names and extra constants defined
+// in the KEP that have a prefix of "kubernetes.io/". Gateway API MAY define additional
+// constants with the prefix "gateway.networking.k8s.io/"
```
robscott, line 566: I'd rather defer to the contents of the KEP + only add references to Gateway prefixed values if/when they exist.
```suggestion
// defined in [KEP-3726].
```

apis/v1alpha2/grpcroute_types.go
```diff
@@ -549,6 +549,42 @@ type GRPCRouteFilter struct {
...
+//
+// If a Service appProtocol isn't specified an implementation MAY infer the backend
+// protocol through its own means. Implementations MAY infer the protocol from the
+// Route type referring to the backend Service.
+//
+// Implementations MAY support Kubernetes Service BackendRefs that are multiplexing TCP
+// and UDP on the same port. Otherwise implementations MUST set the Route ResolvedRefs
+// condition to False with the "UnsupportedProtocol" Reason with a clear message that
+// multiplexing is not supported.
```
robscott, line 575: I've generally thought that the Route implies the protocol that should be used to connect to a backend. I'd rather leave this for a follow up issue/discussion as I think there's a lot of nuance here that we might be missing.

apis/v1alpha2/grpcroute_types.go
```diff
@@ -549,6 +549,42 @@ type GRPCRouteFilter struct {
...
+//
+// If a Route is not able to send traffic to the backend using the specified protocol then
+// the backend is considered invalid. Implementations MUST set ResolvedRefs condition to
+// False with the "UnsupportedProtocol".
+//
+// Implementations MAY support different combinations of protocol/appProtocol/Route Type.
+// See [GEP-1911] for a table.
```
robscott, line 582: In general I don't like relying on GEPs or KEPs as documentation for concepts like this. GEPs + KEPs tend to get stale over time. I'd rather move this to @youngnick's incoming implementer's guide and out of the spec entirely. Mind creating an issue to track this? In the meantime I'd slightly prefer removing it from the spec, but that doesn't need to block.

robscott (commented):
Thanks @dprotaso! Just a few more nits but otherwise LGTM.

apis/v1alpha2/grpcroute_types.go
dprotaso: removed this line

dprotaso: Issue: https://github.com/kubernetes-sigs/gateway-api/issues/2453

apis/v1beta1/shared_types.go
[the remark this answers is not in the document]
dprotaso: > IMO, this feels like a sufficiently complex issue to leave out of this PR and instead file a follow up issie.


Sounds good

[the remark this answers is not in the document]
dprotaso: removed

apis/v1alpha2/grpcroute_types.go
dprotaso: removed

robscott:
Thanks @dprotaso!

/lgtm
/hold cancel

robscott:
One non-blocking thing I missed, do you mind adding a release note to the release note block of this PR @dprotaso? (Can happen post-merge, just helps us not miss anything when we're making a changelog).

PR #2448, opened 2023-10-01
Update GEP-713 and add BackendTLSPolicy implementation

youngnick:
**What type of PR is this?**
/kind feature
/kind gep

This updates GEP-713 with the information needed for BackendTLSPolicy's status (including making this a new recommendation for Policy status), and also implements BackendTLSPolicy.

Updates #713
Updates #1897 

```release-note
BackendTLSPolicy has been added to control TLS from Gateway to backends specified in HTTPRoutes.
```


youngnick:
This needs a little more work to add tests for the BackendTLSPolicy CEL rules, but I haven't done that before, not sure how long it will take. This will let folks review the rest while I work on that.

/hold

apis/v1alpha2/backendtlspolicy_types.go
youngnick, line 1: Note that this file is basically copied verbatim from GEP-1897.

Now's the time to carefully review though and see if we missed anything there.

youngnick:
BackendTLSPolicy also has the problem that support for it should be Extended, but it will be tricky to write conformance tests for. I probably also need to add something to GEP-1897 about how implementations should signal that they support the new object.

I'll work on that now.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Spec defines the desired state of BackendTLSPolicy.
+	Spec BackendTLSPolicySpec `json:"spec"`
+
+	// Status defines the current state of BackendTLSPolicy.
+	Status PolicyStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// BackendTLSPolicyList contains a list of TCPRoute
```
robscott, line 45:
```suggestion
// BackendTLSPolicyList contains a list of BackendTLSPolicies
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+type BackendTLSPolicyList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BackendTLSPolicy `json:"items"`
+}
+
+// BackendTLSPolicySpec defines the desired state of
+// BackendTLSPolicy.
+// Note: there is no Override or Default policy configuration.
```
robscott, line 54: I think we should just clearly state that this is a direct policy.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+//
+// Support: Extended
+type BackendTLSPolicySpec struct {
+	// TargetRef identifies an API object to apply policy to.
```
robscott, line 58:
```suggestion
	// TargetRef identifies an API object to apply the policy to.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// Services are the only valid API target references.
+	// Note that this config applies to the entire referenced resource
+	// by default, but this default may change in the future to provide
+	// a more granular application of the policy.
+	TargetRef PolicyTargetReference `json:"targetRef"`
```
robscott, line 63: We actually want https://github.com/kubernetes-sigs/gateway-api/blob/18506b70f36313bfa966e29d09971a6fe568ce18/apis/v1alpha2/policy_types.go#L65

We may also want to copy what we've said elsewhere in the spec where support for Service and maybe ServiceImport is extended, and everything else is implementation-specific.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// TLS contains backend TLS policy configuration.
+	TLS *BackendTLSPolicyConfig `json:"tls"`
+}
+
+// BackendTLSPolicyConfig contains backend TLS policy configuration.
+// +kubebuilder:validation:XValidation:message="must not contain both CertRefs and StandardCerts",rule="(has(self.certRefs) && size(self.certRefs) > 0 && has(self.standardCerts) && self.standardCerts != ”)"
+// +kubebuilder:validation:XValidation:message="must specify either CertRefs or StandardCerts",rule="!(has(self.certRefs) && size(self.certRefs) > 0 || has(self.standardCerts) && self.standardCerts != ”)"
```
robscott, line 71: I think these both look correct, but would appreciate new tests in `pkg/test/cel` to confirm that they're working as expected.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+type BackendTLSPolicyConfig struct {
+	// CertRefs contains one or more references to
+	// Kubernetes objects that contain PEM-encoded TLS certificates,
+	// which are used to establish a TLS handshake between the gateway
+	// and backend pod.
```
robscott, line 76: Nit: Generally prefer to have k8s resources capitalized (comment applies throughout)
```suggestion
	// which are used to establish a TLS handshake between the Gateway
	// and backend Pod.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// then a TLS connection must fail.
+	//
+	// References to a resource in a different namespace are
+	// invalid.
+	//
+	// A single CertRef to a Kubernetes ConfigMap kind has "Core"
+	// support.  Implementations MAY choose to support attaching
+	// multiple certificates to a backend, but this behavior is
+	// implementation-specific.  Also implementation-specific is
+	// a CertRef of other object kinds, e.g. Secret.
```
robscott, line 94: I think the lines below say the same thing in a slightly clearer way, would remove this.
```suggestion
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//
+	// Support: Core - An optional single reference to a Kubernetes
+	// ConfigMap.
```
robscott, line 97: Do we have a standard key that we're expecting the CA Cert to be set on?
```suggestion
	// Support: Core - A single reference to a Kubernetes ConfigMap.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//
+	// Support: Implementation-specific (No reference, more than one
+	// reference, or resource types other than ConfigMaps.
+	// Service mesh may ignore.)
```
robscott, line 101:
```suggestion
	// Support: Implementation-specific (More than one reference, or other kinds
	// of resources).
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// StandardCerts specifies whether system CA certificates may
+	// be used in the TLS handshake between the gateway and
+	// backend pod.
+	//
+	// If StandardCerts is unspecified or set to "", then CertRefs must
+	// be specified with at least one entry for a valid configuration.
+	// If StandardCerts is unspecified or set to "", then CertRefs must
+	// be specified.  Only one of CertRefs or StandardCerts may be
+	// specified, not both.
```
robscott, line 115:
```suggestion
	// If StandardCerts is unspecified or set to "", then CertRefs must
	// be specified with at least one entry for a valid configuration.
	// Only one of CertRefs or StandardCerts may be specified, not both.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//
+	// StandardCerts must be set to "System" when CertRefs is unspecified.
+	//
+	// If StandardCerts is set to "System", then the system trusted
+	// certificates should be used. If there are none, or the
+	// implementation doesn't define system trusted certificates,
+	// then a TLS connection must fail.
+	//
+	// Support: Core - An optional value to specify whether to use
+	// system certificates or not.
```
robscott, line 125:
```suggestion
	// Support: Core for "System"
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//
+	// Support: Implementation-specific (In the absence of support
+	// for usable system certs, may be ignored. Service mesh may ignore.)
```
robscott, line 128: Not sure we need this? Are there implementations that would not be able to support some concept of system certs?

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// Hostname is the Server Name Indication that the Gateway uses to
+	// connect to the backend.  It represents the fully qualified domain
+	// name of a network host, as defined by RFC1123 - except that numeric
+	// IP addresses are not allowed. Each label of the FQDN must consist
+	// of lower case alphanumeric characters or '-', and must start and
+	// end with an alphanumeric character.  No other punctuation is allowed.
+	// Wildcard domain names are specifically disallowed.
+	//
+	// It specifies the hostname that may authenticate, and must be in the
+	// certificate served by the matching backend.
```
robscott, line 142: Maybe this could be simplified to something like this:

```suggestion
	// Hostname is used for two purposes in the connection between Gateways and
	// backends:
	//
	// 1. Hostname MUST be used as the SNI to connect to the backend (RFC 3546).
	// 2. Hostname MUST be used for authentication and MUST be in the certificate
	//    served by the matching backend.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//
+	// Support: Core - A required value used by the Gateway to connect to
+	// the backend when a BackendTLSPolicy is specified.
```
robscott, line 145:
```suggestion
	// Support: Core
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+
+// +genclient
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:resource:categories=gateway-api,shortName=btlspolicy
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// BackendTLSPolicy provides a way to publish TLS configuration
+// that enables a gateway client to connect to a backend pod.
```
robscott, line 32:
```suggestion
// BackendTLSPolicy provides a way to configure how a Gateway
// connects to a Backend via TLS.
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+type BackendTLSPolicy struct {
```
robscott, line 33: I think we also need an addition somewhere on the BackendRef type that describes that when this policy is targeting a Backend matching the BackendRef, implementations SHOULD reject the backend if they can't implement the BackendTLSPolicy. We should probably also add a new Reason along with this and then transition the `SHOULD` to a `MUST` when this policy graduates to Standard channel.

https://github.com/kubernetes-sigs/gateway-api/blob/18506b70f36313bfa966e29d09971a6fe568ce18/apis/v1beta1/shared_types.go#L239-L246

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	Hostname v1beta1.PreciseHostname `json:"hostname"`
+}
+
+// StandardCertType is the type of CA certificate that will be used when
+// the TLS.certRefs is unspecified.
+// +kubebuilder:validation:Enum=System
+type StandardCertType string
+
+const (
+	StandardCertSystem StandardCertType = "System"
```
robscott, line 155:
```suggestion
	// Indicates that standard system CA certificates should be used.
	StandardCertSystem StandardCertType = "System"
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// If StandardCerts is unspecified or set to "", then CertRefs must
+	// be specified.  Only one of CertRefs or StandardCerts may be
+	// specified, not both.
+	//
+	// StandardCerts must be set to "System" when CertRefs is unspecified.
+	//
+	// If StandardCerts is set to "System", then the system trusted
+	// certificates should be used. If there are none, or the
+	// implementation doesn't define system trusted certificates,
+	// then a TLS connection must fail.
```
robscott, line 122: Recommend removing this in favor of top-level guidance for what to do when this policy can't be supported + a definition of the "System" type by the actual `const` definition.
```suggestion
```

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// Support: Core - An optional single reference to a Kubernetes
+	// ConfigMap.
+	//
+	// Support: Implementation-specific (No reference, more than one
+	// reference, or resource types other than ConfigMaps.
+	// Service mesh may ignore.)
+	//
+	// +kubebuilder:validation:MaxItems=8
+	// +optional
+	CertRefs []v1beta1.ConfigMapObjectReference `json:"certRefs,omitempty"`
```
robscott, line 105: Nit: I'm not convinced that ConfigMaps are the best long term solution here, would prefer not starting with a type that defaults to them.

robscott (commented):
Thanks for getting this one through @youngnick! I just focused on the TLS side of things, but the Policy side largely lines up with what I was hoping for, just did not have time to take a close look yet.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Okay, would you prefer no default here to start with then? That also doesn't seem ideal.

The idea we discussed was to make it easy to get started for now, and we can change the defaults later if we want - because they are defaults, they are added the first time the object is persisted, so existing resources will keep working even if we do.

robscott: My concern is that I think generally we're supposed to release a new API version when changing default values (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#static-defaults). We could theoretically say that it's experimental channel and anything goes, but in this particular case it seems very unlikely that ConfigMap as default is something we want to stick with through to GA, so I'd personally rather not start with it.

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
 This is already part of the API pattern, but is being lifted to more prominience
 here.
 
-#### Standard status Condition
+#### Standard status struct
+
+Status: Recommended
```
robscott, line 915: Maybe this would be clearer?
```suggestion
Status: Experimental
```

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+the `BackendTLSPolicy` object for detailed examples. Included here is a representative
+version.
+
+This is an `Ancestor` status rather than a `Parent` status, as in the Route status
+because for Policy attachment, the relevant object may or may not be the direct
+parent.
+
+For example, `BackendTLSPolicy` directly attaches to a Service, which may be included
+in multiple Routes, in multiple Gateways. However, the status of the `BackendTLSPolicy`
+will be different only at the Gateway level, so Gateway is the relevant Ancestor
```
robscott, line 928:
```suggestion
in multiple Routes, in multiple Gateways. However, for many implementations, 
the status of the `BackendTLSPolicy` will be different only at the Gateway level, 
so Gateway is the relevant Ancestor
```

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
-#### Standard status Condition
+#### Standard status struct
+
+Status: Recommended
+
+Policy objects SHOULD use the upstream `PolicyAncestorStatus` struct in their respective
+Status structs. Please see the included `PolicyAncestorStatus` struct, and its use in
+the `BackendTLSPolicy` object for detailed examples. Included here is a representative
+version.
+
```
robscott, line 921: Not exactly sure where this belongs, but I think some content roughly like this would help show that we want _everyone_ to migrate to this approach, and provide some rationale for why it will be beneficial.
```suggestion
This pattern enables different conditions to be set for different "Ancestors"
of the target resource. This is particularly helpful for policies that may be
implemented by multiple controllers or attached to resources with different
capabilities. This pattern also provides a clear view of what resources a
policy is affecting.

For the best integration with community tooling and consistency across
the broader community, we recommend that all implementations transition 
to Policy status with this kind of nested structure.
```

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+	Ancestors []PolicyAncestorStatus `json:"ancestors"`
+}
+```
+
+##### Design considerations
+
+This is recommended as the base for Policy object's status. As Policy Attachment
+is a pattern, not an API, "recommended" is the strongest we can make this, but
+we believe that standardizing this will help a lot with discoverability.
+
```
robscott, line 1020: Maybe add something like this?
```suggestion

Note that is likely that all Gateway API tooling will expect policy status to follow
this structure. To benefit from broader consistency and discoverability, we
recomment transitioning to this structure for all Gateway API Policies.
```

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Yeah, okay, I'll pull out ConfigMapObjectReference for now then, can always bring it back later if need be. This will be moved to just a straight ObjectReference then, which will mean that people will need to specify that this is a configmap to begin with (which I agree is not a huge deal).

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+```go
+// PolicyAncestorStatus describes the status of a route with respect to an
+// associated Ancestor.
+//
+// The object that a Policy attaches to (the object in its TargetRef) is its
+// _parent_, and any other objects above it in an object hierarchy are its
+// _ancestors_.
+// 
+// For some Policies, particularly Direct Attached Policies, the relevant object
+// that distinguishes its status is not necessarily the parent object.
```
robscott, line 949: Maybe something like this would also show the other use case here which is when different implementations are implementing the same policy.
```suggestion
// Policies targeting the same resource may have different effects depending on the 
// ancestors of those resources. For example, different Gateways targeting the same
// Service may have different capabilities, especially if they have different underlying
// implementations. 
```

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+
+[gep-1897]: /geps/gep-1897
+
+```go
+// PolicyAncestorStatus describes the status of a route with respect to an
+// associated Ancestor.
+//
+// The object that a Policy attaches to (the object in its TargetRef) is its
+// _parent_, and any other objects above it in an object hierarchy are its
+// _ancestors_.
```
robscott, line 946: Maybe something like this?
```suggestion
// Ancestors refer to objects that are either the Target of a policy or above it in terms
// of object hierarchy. For example, if a policy targets a Service, an Ancestor could be
// a Route or a Gateway. 

// In the context of policy attachment, the Ancestor is used to distinguish which
// resource results in a distinct application of this policy. For example, if a policy
// targets a Service, it may have a distinct result per attached Gateway.
```

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+// For some Policies, particularly Direct Attached Policies, the relevant object
+// that distinguishes its status is not necessarily the parent object.
+//
+// For example, in BackendTLSPolicy, the Policy attaches to a Service that is
+// used as a backend in a HTTPRoute that is itself attached to a Gateway.
+// In this case, the relevant object for status is the Gateway, and that is the
+// ancestor object referred to in this status.
+//
+// Note that a parent is also an ancestor, so for objects where the parent is the
+// relevant object for status, this struct SHOULD still be used.
```
robscott, line 957: For some reason it's confusing to me to consider the "Target" of a policy the "Parent" of that policy. Would this make sense?
```suggestion
// Note that a Target of a Policy is also a valid Ancestor, so for objects where the 
// Target is the relevant object for status, this struct SHOULD still be used.
```

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Paging @candita though for this one, this is a little different to the discussion we had on the GEP.

geps/gep-713.md
```diff
@@ -910,7 +910,115 @@ plugin.
...
+	// appropriate when the relevant ancestor is modified.
+	//
+	// Note that choosing the relevant ancestor is left to the Policy designers;
+	// an important part of Policy design is designing the right object level at
+	// which to namespace this status.
+	//
+	// Note also that parent references that cannot be resolved by an implementation
+	// of this API will not be added to this list. Implementations of this API
+	// can only populate ancestor status for the parent resources they are
+	// responsible for.
```
robscott, line 1005: I think this is roughly the same thing you're saying, feel free to reject if it doesn't make sense:

```suggestion
	// Note also that implementations MUST ONLY populate ancestor status for 
	// the Ancestor resources they are responsible for. Implementations MUST
	// use the ControllerName field to uniquely identify the entries in this list
	// that they are responsible for.
```

robscott (commented):
Thanks @youngnick! This review just covers the policy side of the PR. I think it all looks solid, just had some suggested tweaks + additions to spec here.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Added to the `HTTPBackendRef` struct in `v1beta1/httproute_types.go`.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Good point, updated.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: I agree, let's leave this out and add it back if we need it.

geps/gep-713.md
youngnick: Yeah, love it.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// multiple certificates to a backend, but this behavior is
+	// implementation-specific.
+	//
+	// Support: Core - An optional single reference to a Kubernetes
+	// ConfigMap, with the CA certificate in a key named `ca.crt`.
+	//
+	// Support: Implementation-specific (More than one reference, or other kinds
+	// of resources).
+	//
+	// +kubebuilder:validation:MaxItems=8
```
howardjohn, line 108: nit: MinItems=1? otherwise you can say `certRefs=[]` (I think)

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// BackendTLSPolicyConfig contains backend TLS policy configuration.
+// +kubebuilder:validation:XValidation:message="must not contain both CertRefs and StandardCerts",rule="(has(self.certRefs) && size(self.certRefs) > 0 && has(self.standardCerts) && self.standardCerts != \"\")"
+// +kubebuilder:validation:XValidation:message="must specify either CertRefs or StandardCerts",rule="!(has(self.certRefs) && size(self.certRefs) > 0 || has(self.standardCerts) && self.standardCerts != \"\")"
+type BackendTLSPolicyConfig struct {
+	// CertRefs contains one or more references to
+	// Kubernetes objects that contain PEM-encoded TLS certificates,
+	// which are used to establish a TLS handshake between the Gateway
+	// and backend Pod.
```
howardjohn, line 83: nitpicky: `PEM-encoded TLS certificates which are used **to establish** a TLS handshake`

To me this reads like its a client certificate. I can see how its correct, just a bit confusing to me personally.

In my mind "to establish a TLS handshake" implies its literally part of the handshake on the wire, which is not the case -- its just used in validation of the handshake.

Another example of the ~same config:

```
$ k explain mutatingwebhookconfigurations.webhooks.clientConfig.caBundle
GROUP:      admissionregistration.k8s.io
KIND:       MutatingWebhookConfiguration
VERSION:    v1

FIELD: caBundle <string>

DESCRIPTION:
    `caBundle` is a PEM encoded CA bundle which will be used to validate the
    webhook's server certificate. If unspecified, system trust roots on the
    apiserver are used.
```

Using some language there may be more clear

apis/v1alpha2/backendtlspolicy_types.go
howardjohn: +1 to no defaults

geps/gep-1897.md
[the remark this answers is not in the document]
youngnick: It is generally acceptable for GEPs to go stale, but I've updated this one to reflect the updated structs from the actual implementation. Most likely we won't update it after this though - the code is now the canonical definition.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// a more granular application of the policy.
+	//
+	// Support: Extended for Kubernetes Service
+	//
+	// Support: Implementation-specific for any other resource
+	//
+	TargetRef PolicyTargetReferenceWithSectionName `json:"targetRef"`
+
+	// TLS contains backend TLS policy configuration.
+	TLS *BackendTLSPolicyConfig `json:"tls"`
```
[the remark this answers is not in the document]
youngnick: Nice catch, thanks!

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Yeah, thanks, this is a good callout. I've updated to make clearer that these are CA bundles.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: `MinItems=1` makes this not optional though, so I think we need to stick with this. In any case, we can add this later if we need it.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: I've updated this to use a new `ObjectReference` type that has no defaults - this means that `group` and `kind` are now _required_.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Only one of CertRefs or StandardCerts may be specified, not both.
+	//
+	// StandardCerts must be set to "System" when CertRefs is unspecified.
+	//
+	//
+	// Support: Core for "System"
+	//
+	//
+	// +optional
+	StandardCerts *StandardCertType `json:"standardCerts,omitempty"`
```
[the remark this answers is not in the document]
youngnick: I really don't want to redo this discussion again at this point; I'd like to get this into review first and see what happens there. This is an alpha API, so we can bikeshed and come back if we need to.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Hostname is used for two purposes in the connection between Gateways and
+	// backends:
+	//
+	// 1. Hostname MUST be used as the SNI to connect to the backend (RFC 3546).
+	// 2. Hostname MUST be used for authentication and MUST be in the certificate
```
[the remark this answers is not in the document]
youngnick: "match" is a good idea, done.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	//    served by the matching backend.
+	//
+	// Support: Core
+	Hostname v1beta1.PreciseHostname `json:"hostname"`
+}
+
+// StandardCertType is the type of CA certificate that will be used when
+// the TLS.certRefs is unspecified.
+// +kubebuilder:validation:Enum=System
```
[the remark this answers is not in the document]
youngnick: Yes, this is essentially a boolean with room for later expansion if we need it. I think domain-prefix is not necessary in this case.

apis/v1alpha2/backendtlspolicy_types.go
candita: @youngnick we went in a few different directions in the original GEP, and first decided that ConfigMapReference would be the simplest at the time: https://github.com/kubernetes-sigs/gateway-api/pull/2113#discussion_r1231584950 

https://github.com/kubernetes-sigs/gateway-api/pull/2113#discussion_r1311009936 discusses why we went with ConfigMapReference.

ObjectReference seems even simpler, but perhaps we could add some examples for the Group and Kind if we don't default it (ClusterTrustBundled, ConfigMapReference)?

apis/v1beta1/httproute_types.go
```diff
@@ -1137,6 +1137,11 @@ type HTTPBackendRef struct {
 	// * It refers to a Kubernetes Service that has an incompatible appProtocol
 	//   for the given Route type
 	//
+	// * The BackendTLSPolicy object is installed in the cluster, a BackendTLSPolicy
+	//   is present that refers to the Service, and the implementation is unable
+	//   to meet the requirement. At the time of writing, BackendTLSPolicy is
+	//   experimental, but once it becomes
```
candita, line 1143: There's an early ending on the sentence.

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
 	// policy is attached to an invalid target resource.
 	PolicyReasonTargetNotFound PolicyConditionReason = "TargetNotFound"
 )
+
```
candita, line 125: Could you update L79 in`PolicyTargetReferenceWithSectionName` : "a `ResolvedRefs` or similar Condition in the Policy's status." to what it should be, either `TargetNotFound` or add a new condition `TargetSectionNotFound`?

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
...
+// PolicyAncestorStatus describes the status of a route with respect to an
+// associated Ancestor.
+//
+// Ancestors refer to objects that are either the Target of a policy or above it in terms
+// of object hierarchy. For example, if a policy targets a Service, an Ancestor could be
+// a Route or a Gateway.
```
candita, line 131: This seems to disagree with https://github.com/kubernetes-sigs/gateway-api/pull/2448/files#diff-4333085192c93205580c7dd9213f10fe061e77245adfa23310487ae0924c7dbaR144-R145, which claims the Gateway as the ancestor object.  Could you explain under what circumstance a Route would be an ancestor of a Service, but not the Gateway?

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
...
+// For example, in BackendTLSPolicy, the Policy attaches to a Service that is
+// used as a backend in a HTTPRoute that is itself attached to a Gateway.
+// In this case, the relevant object for status is the Gateway, and that is the
+// ancestor object referred to in this status.
+//
+// Note that a parent is also an ancestor, so for objects where the parent is the
+// relevant object for status, this struct SHOULD still be used.
+type PolicyAncestorStatus struct {
+	// AncestorRef corresponds with a ParentRef in the spec that this
+	// RouteParentStatus struct describes the status of.
```
candita, line 151:
```suggestion
	// PolicyAncestorStatus struct describes the status of.
```

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=8
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// PolicyStatus defines the common attributes that all Policies should include within
+// their status.
+type PolicyStatus struct {
+	// Ancestors is a list of ancestor resources (usually Gateways) that are
+	// associated with the route, and the status of the route with respect to
```
candita, line 182: This refers to route, shouldn't it refer to policy?

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
...
+	//
+	// Note also that implementations MUST ONLY populate ancestor status for
+	// the Ancestor resources they are responsible for. Implementations MUST
+	// use the ControllerName field to uniquely identify the entries in this list
+	// that they are responsible for.
+	//
+	// A maximum of 32 ancestors will be represented in this list. An empty list
+	// means the Policy is not relevant for any ancestors.
+	//
+	// +kubebuilder:validation:MaxItems=32
```
candita, line 200: 32 is probably too large for a v1alpha2 presence.  Maybe start with 8 and grow it later if needed.  Easy to grow, impossible to shrink.

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
 	// policy is attached to an invalid target resource.
 	PolicyReasonTargetNotFound PolicyConditionReason = "TargetNotFound"
 )
+
+// PolicyAncestorStatus describes the status of a route with respect to an
+// associated Ancestor.
+//
+// Ancestors refer to objects that are either the Target of a policy or above it in terms
```
candita, line 129: nit: `Ancestors` doesn't seem like the right phrase.  It represents parents and parents of parents ad infinitum.  A target is more akin to a predecessor.

apis/v1alpha2/policy_types.go
youngnick: A Route is always _an_ ancestor of the Service, but it is probably not _the relevant_ ancestor of the Service unless the Route does not attach to a Gateway, because all Routes that have Services with BackendTLSPolicies attached, that themselves attach to a Gateway that supports BackendTLSPolicy will have the same behaviors, in general.

Of course, this is only a _recommendation_, not a _specification_, so we can't _make_ people do anything.

I'll try to clarify this wording a bit.

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,81 @@ const (
...
+	// Example: "example.net/gateway-controller".
+	//
+	// The format of this field is DOMAIN "/" PATH, where DOMAIN and PATH are
+	// valid Kubernetes names
+	// (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names).
+	//
+	// Controllers MUST populate this field when writing status. Controllers should ensure that
+	// entries to status populated with their ControllerName are cleaned up when they are no
+	// longer necessary.
+	ControllerName GatewayController `json:"controllerName"`
```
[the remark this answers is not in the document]
youngnick: The whole point of PolicyAncestorStatus is that there must be one PolicyAncestorStatus _per controller name + AncestorRef combination_, the same as how there must be one RouteParentStatus per controller name and ParentRef combination. Those two fields are the key in a listTypeMap-style object.

I've added some information to the top of the godoc for the struct to this effect.

apis/v1alpha2/policy_types.go
youngnick: I've added some additional wording that I think clarifies why I've chosen this word.

apis/v1alpha2/backendtlspolicy_types.go
danehans: Regarding "standard", do we anticipate system certs other than standard, e.g. "optional" or "custom"?

apis/v1beta1/object_reference_types.go
```diff
@@ -145,3 +145,42 @@ type BackendObjectReference struct {
...
+//
+// References to objects with invalid Group and Kind are not valid, and must
+// be rejected by the implementation, with appropriate Conditions set
+// on the containing object.
+type ObjectReference struct {
+	// Group is the group of the referent.  For example, "gateway.networking.k8s.io".
+	// When unspecified or empty string, core API group is inferred.
+	//
+	// +kubebuilder:validation:Required
+	Group *Group `json:"group"`
```
danehans, line 163: According to [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required), this field should not be a pointer if it's required.

apis/v1beta1/object_reference_types.go
```diff
@@ -145,3 +145,42 @@ type BackendObjectReference struct {
...
+
+	// Kind is the kind of the referent.  For example, "ConfigMap".
+	//
+	// +kubebuilder:validation:Required
+	Kind *Kind `json:"kind"`
```
danehans, line 168: According to [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required), this field should not be a pointer if it's required.

apis/v1alpha2/backendtlspolicy_types.go
danehans: Note that GEP-1897 states "CertRefs contains one or more (up to 64) references to Kubernetes objects..." and should be updated accordingly.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// Support: Core for "System"
+	//
+	//
+	// +optional
+	StandardCerts *StandardCertType `json:"standardCerts,omitempty"`
+
+	// Hostname is used for two purposes in the connection between Gateways and
+	// backends:
+	//
+	// 1. Hostname MUST be used as the SNI to connect to the backend (RFC 3546).
```
louiscryan, line 132: 3546 has been obsoleted. I think you want to point to 6066

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// be specified.  Only one of CertRefs or StandardCerts may be
+	// specified, not both.
+	//
+	// If CertRefs is empty or unspecified, then system trusted
+	// certificates should be used. If there are none, or the
+	// implementation doesn't define system trusted certificates,
+	// then a TLS connection must fail.
+	//
+	// References to a resource in a different namespace are
+	// invalid.
```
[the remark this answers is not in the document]
youngnick: We left this like this for now so that we can add the cross-namespace reference language later if needed. I've added a note, thanks!

apis/v1beta1/object_reference_types.go
youngnick: Yeah, thanks, I hadn't updated since changing to required.

apis/v1alpha2/backendtlspolicy_types.go
youngnick: Thanks, updated.

kflynn:
I voiced concerns about this one at the v1.0 API review. Overall, those concerns kind of boil down to the backend TLS and ancestors feeling _even more_ experimental than GEP-713 itself, which worries me: landing this will fold those into GEP-713, and no one reading the GEP later will know the gradations between bits of 713.

But perhaps I should trust the experimental process more. @youngnick, I'm OK landing this as experimental for v1.0. Probably sometime after that we should take a serious look at graduation criteria for GEP-713 -- it's significant enough that we might not want to leave it exempt.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,149 @@
...
+	//
+	TargetRef PolicyTargetReferenceWithSectionName `json:"targetRef"`
+
+	// TLS contains backend TLS policy configuration.
+	TLS BackendTLSPolicyConfig `json:"tls"`
+}
+
+// BackendTLSPolicyConfig contains backend TLS policy configuration.
+// +kubebuilder:validation:XValidation:message="must not contain both CertRefs and StandardCerts",rule="(has(self.certRefs) && size(self.certRefs) > 0 && has(self.standardCerts) && self.standardCerts != \"\")"
+// +kubebuilder:validation:XValidation:message="must specify either CertRefs or StandardCerts",rule="!(has(self.certRefs) && size(self.certRefs) > 0 || has(self.standardCerts) && self.standardCerts != \"\")"
```
robscott, line 78: Added https://github.com/kubernetes-sigs/gateway-api/issues/2473 to track adding CEL tests for this. Does not need to block this PR, but probably needs to block v1.0.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,149 @@
...
+type BackendTLSPolicyConfig struct {
+	// CertRefs contains one or more references to
+	// Kubernetes objects that contain a PEM-encoded TLS CA certificate bundle,
```
robscott, line 81: Nit: Line length appears to have gotten a bit messed up here.

apis/v1alpha2/backendtlspolicy_types.go
robscott: I think CEL validation should be able to prevent invalid inputs like this, but I've filled #2473 to ensure we have good test coverage of our validation here before releasing this.

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,149 @@
...
+	//
+	// Support: Core - An optional single reference to a Kubernetes
+	// ConfigMap, with the CA certificate in a key named `ca.crt`.
+	//
+	// Support: Implementation-specific (More than one reference, or other kinds
+	// of resources).
+	//
+	// +kubebuilder:validation:MaxItems=8
+	// +optional
+	CertRefs []v1beta1.ObjectReference `json:"certRefs,omitempty"`
```
robscott, line 111: Discussed offline, but this may make the config both clearer and easier to extend this in the future.
```suggestion
	CACertRefs []v1beta1.ObjectReference `json:"caCertRefs,omitempty"`
```

apis/v1alpha2/backendtlspolicy_types.go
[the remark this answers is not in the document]
robscott: I also prefer slightly preferred `WellKnown`, but agree that it doesn't need to block. I would prefer to at least have `CA` in the name, so we end up with something like `StandardCACerts` and there's less ambiguity about what these are for.

apis/v1alpha2/policy_types.go
robscott: Good catch! I think this might actually be best for a follow up PR because we probably also want to have `TargetSectionNotFound` like you're suggesting, but I think it would be cleaner to add a new condition like that in a separate PR.

apis/v1alpha2/backendtlspolicy_types.go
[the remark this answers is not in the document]
robscott: Yeah, I think domain-prefixed values might actually be useful in the future, but I'm fine with leaving that until later since I can't think of any immediate use cases. IMO, the most important thing at this point is that we have room for expansion here.

robscott (commented):
Thanks @youngnick! A few nits but otherwise LGTM.

apis/v1alpha2/backendtlspolicy_types.go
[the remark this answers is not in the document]
youngnick: I've thought about this some more, and I think I agree with `WellknownCACerts`. I'm changing it now.

apis/v1alpha2/policy_types.go
```diff
@@ -120,3 +122,92 @@ const (
...
+	// the Ancestor resources they are responsible for. Implementations MUST
+	// use the ControllerName field to uniquely identify the entries in this list
+	// that they are responsible for.
+	//
+	// Note that to achieve this, the list of PolicyAncestorStatus structs
+	// MUST be treated as a map with a composite key, made up of the AncestorRef
+	// and ControllerName fields combined.
+	//
+	// A maximum of 8 ancestors will be represented in this list. An empty list
+	// means the Policy is not relevant for any ancestors.
```
robscott, line 209: Don't think it needs to block this PR, but we should address Tim's feedback here. We really should define what implementations should do when this is full. At a minimum we should clarify that they should not try to add to the list, but maybe we can think of some better options here longer term?

apis/v1beta1/httproute_types.go
```diff
@@ -1162,6 +1162,12 @@ type HTTPBackendRef struct {
 	// * It refers to a Kubernetes Service that has an incompatible appProtocol
 	//   for the given Route type
 	//
+	// * The BackendTLSPolicy object is installed in the cluster, a BackendTLSPolicy
+	//   is present that refers to the Service, and the implementation is unable
+	//   to meet the requirement. At the time of writing, BackendTLSPolicy is
+	//   experimental, but once it becomes standard, this will become a MUST
+	// requirement.
```
robscott, line 1169: Non-blocking nit:
```suggestion
	//   requirement.
```

geps/gep-713.md
```diff
@@ -910,7 +910,135 @@ plugin.
...
+	// appropriate when the relevant ancestor is modified.
+	//
+	// Note that choosing the relevant ancestor is left to the Policy designers;
+	// an important part of Policy design is designing the right object level at
+	// which to namespace this status.
+	//
+  // Note also that implementations MUST ONLY populate ancestor status for 
+  // the Ancestor resources they are responsible for. Implementations MUST
+  // use the ControllerName field to uniquely identify the entries in this list
+  // that they are responsible for.
```
robscott, line 1021: Non-blocking nit:
```suggestion
	// Note also that implementations MUST ONLY populate ancestor status for 
	// the Ancestor resources they are responsible for. Implementations MUST
	// use the ControllerName field to uniquely identify the entries in this list
	// that they are responsible for.
```

robscott (approved):
Thanks @youngnick! 3 non-blocking nits, feel free to remove hold if you'd rather cover them in a follow up.

/lgtm
/hold

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -0,0 +1,148 @@
...
+	// specified, not both.
+	//
+	// If CACertRefs is empty or unspecified, then system trusted
+	// certificates should be used. If there are none, or the
+	// implementation doesn't define system trusted certificates,
+	// then a TLS connection must fail.
+	//
+	// References to a resource in a different namespace are
+	// invalid for the moment, although we will revisit this
+	// in the future.
```
EItanya, line 95: No `ReferenceGrant` support?

youngnick: Not for now - we wanted to get this MVP in, and we can add ReferenceGrant support later. I agree it seems reasonable to add for this field though.

EItanya: Sounds good, at a high level is this the kind of decision an implementation could decide to implement on their own given the right documentation/reasoning, or would that be considered bad practice. Mostly just curious about how people should/have approach such a thing

youngnick: We've actually chosen the types such that you _can't_ at the moment - but the LocalObjectReference type and the ObjectReference type are unmarshal-equivalent, so when we do migrate, anything stored as JSON/YAML in etcd will deserialize correctly into the ObjectReference type.

So, I guess that's a long way of saying that changing this _right now_ will require forking BackendTLSPolicy, which I wouldn't recommend.

PR #2450, opened 2023-10-02
Update Gateway API URLRewrite tests

nxy7:
Signed-off-by: nxyt <lolnoxy@gmail.com>

**What type of PR is this?**
/kind test
/area conformance

Sometimes changing path prefix and stripping it are two different things. While using cilium Gateway API implementation I've found out that it cannot strip prefixes correctly and there are no tests for such edgecases. This change adds tests for stripping path prefixes.

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-rewrite-path.go
+++ b/conformance/tests/httproute-rewrite-path.go
@@ -61,6 +61,30 @@ var HTTPRouteRewritePath = suite.ConformanceTest{
 			},
 			{
 				Request: http.Request{
+					Path: "/strip-prefix/three",
+				},
+				ExpectedRequest: &http.ExpectedRequest{
+					Request: http.Request{
+						Path: "/three",
+					},
+				},
+				Backend:   "infra-backend-v1",
+				Namespace: ns,
+			},
+			{
+				Request: http.Request{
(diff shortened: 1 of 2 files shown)


robscott:
Thanks @nxy7! These additions look reasonable to me, but would be good to get some feedback from Cilium side as well.

/ok-to-test
/cc @sayboras @youngnick

nxy7:
@robscott this issue is not really cillium specific, I believe that any tool that uses envoy underneath can be affected (potentially more if they're not made for stripping whole paths). For additional context here's issue I've had https://github.com/cilium/cilium/issues/28270 and PR that fixes it https://github.com/cilium/cilium/pull/28294.
I think that there might be more things that could be tested (like empty replacements from [gateway api guidelines](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier)), but for now I've only made checks for things I've encountered myself.

youngnick:
This LGTM, thanks @nxy7 for tracking this down. These tests seem reasonable to me.

robscott:
Thanks @nxy7!

/approve

PR #2451, opened 2023-10-02
Add option to run single conformance test for test dev/debug

frankbu:
**What type of PR is this?**

/kind test
/area conformance

Make conformance test dev/debug a little easier.

Fixes #

```release-note
NONE
```


frankbu:
> Since the purpose of this new field is dev/debug, do you see any limitations in using the above?

@mlavacca I feel this new flag is just a trivial addition that gives one a very simple way to both setup and run a single test. Your suggested code is fine but the test env also needs to be set up first, which requires some additional hacking. It's not a big deal, but IMO seems convenient to provide the extra flag.

frankbu:
> Your suggested code is fine but the test env also needs to be set up first, which requires some additional hacking. 

@mlavacca Maybe you just meant to run the conformance test suite as usual, but first replace line 80 in confomance_test.go?
```
	//cSuite.Run(t, tests.ConformanceTests)
	cSuite.Run(t, []suite.ConformanceTest{tests.HTTPRouteRequestMirror})
```

If so, the only difference is that my suggestion doesn't require changing the code at all, which is slightly better IMO but maybe not worth adding a flag? I don't feel strongly, so could go either way.

frankbu:
@robscott @shaneutt What do you guys think about merging this one? It's a pretty trivial addtion that lets one run/debug a single test without hacking any of the framework code.

shaneutt (approved):
I actually kinda agree with Mattia's point, but I also appreciate that you're helping us improve the test suite _and_ thinking it over I think this is a particularly nice flag to have from the _CLI_ perspective. Let's go for it.

/ok-to-test
/approve
/lgtm

PR #2454, opened 2023-10-05
Add implementer's guide

youngnick:
/kind documentation

This adds an implementer's guide for Gateway API. The new guide is also now part of the specification, so MUST, SHOULD, and MAY there are treated the same as in the godoc or rendered Specification page.

```release-note
A guide for Gateway API implementers is now included in the specification.
```


site-src/concepts/guidelines.md
```diff
@@ -1,4 +1,4 @@
-# Implementation guidelines
```
youngnick, line 1: I started this PR reviewing this page, but it turned out that these guidelines are actually for us, the project contributors and maintainers, not for implementers. So "Design guidelines" felt more appropriate.

Joffref:
Hey @youngnick, fantastic idea. One thing that might be interesting to see in this document is a[ recent Slack question about filters and policies](https://kubernetes.slack.com/archives/CR0H13KGA/p1696347490523719). Seeing more context, including differences and the implementer PoV, would be nice.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+
+- Check if there are any Gateway API CRDs installed in the cluster.
+- If not, install its most compatible version of the CRDs.
+- If so, only install its version of the CRDs if the bundle version is higher
+  than the existing one.
+
+
+Because of our backwards compatibility guarantees, it's also safe for a controller
+to flip the install channel between "standard" and "experimental", although
+implementations MUST NOT do this without consulting the implementation owner.
```
howardjohn, line 159: Who is the "implementation owner"? In my reading, if I substitute phrases, this says "although
**Istio** MUST NOT do this without consulting the **Istio maintainers**" which seems odd? Or is it the user?

Anyhow, one risk with flipping the channel is another controller flipping it back endlessly

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+#### Manual CRD installation
+
+Manual CRD installation has the advantage that the implementer needs to maintain
+less code; however, it pushes the responsibility for correctly managing the
+Gateway API CRDs back to the cluster admin, who may not have as much context
+as is provided here.
+
+Implementations MAY require the installation to be done manually; if so, the
+installation instructions SHOULD include commands to check if there are any other
+CRDs installed already and verify that the installation will not be a downgrade.
```
howardjohn, line 174: As a followup maybe we should provide a common command to do this.

Istio, for instance, has an example of a suboptimal approach: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#setup

howardjohn (approved):
Nice!

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+fields are examples where each item in the slice identifies the Conditions
+associated with some subset of configuration.
+
+For the Gateway case, it's to allow Conditions per _Listener_, and in the Route
+case, it's to allow Conditions per _implementation_ (since Route objects can
+be used in multiple Gateways, and those Gateways can be reconciled by different
+implementations).
+
+In all of these cases, there are some relatively-common Condition types that have
+similar meanings:
```
robscott, line 221: I'm worried that the definitions we have here will become out of sync with the definitions we have in the spec, do we need to define these in both places?

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Resources details
+
+For each currently available conformance profile, there are a set of resources
+that implementations are expected to reconcile.
+
+The following section goes through each Gateway API object, indicates expected
+behaviors, and which conformance profiles that object is included in.
+
+#### GatewayClass
```
robscott, line 257: How do we decide what belongs in here vs https://gateway-api.sigs.k8s.io/api-types/gatewayclass/ vs API Spec? I'm worried that we're going to end up with subtle differences in each individual source if we're not _very_ careful.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+
+If something in the GatewayClass renders it incompatibie (at the time of writing,
+the only possible reason for this is that there is a pointer to a `paramsRef`
+object that is not supported by the implementation), then the implementation
+SHOULD mark the incompatible GatewayClass as not `Accepted`.
+
+Watched in profiles:
+
+- HTTP
+- TLS
```
robscott, line 283: I'd personally rather leave this out since I think we've already begun the transition towards per-resource conformance.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+installation before working. Both have tradeoffs.
+
+Either way has certain things that SHOULD be true, however:
+
+Whatever method is used, cluster admins SHOULD attempt to ensure that
+the Bundle version of the CRDs is not _downgraded_. Although we ensure that
+API changes are backwards compatible, changing CRD definitions can change the
+storage version of the resource, which could have unforseen effects. Most of the
+time, things will probably work, but if it doesn't work, it will most likely 
+break in weird ways.
```
robscott, line 114: Probably worth noting somewhere in here that older versions of the API may be missing fields or features which could be very disruptive for users.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
+# Gateway API Implementer's Guide
```
robscott, line 1: Nit: This location doesn't feel quite right. So far all our guides are targeted at users of the API, and this one is decidedly different than those other ones. Maybe this belongs under the "Reference" tab instead?

robscott (commented):
Thanks for all the work on this @youngnick!

youngnick:
> Hey @youngnick, fantastic idea. One thing that might be interesting to see in this document is a[ recent Slack question about filters and policies](https://kubernetes.slack.com/archives/CR0H13KGA/p1696347490523719). Seeing more context, including differences and the implementer PoV, would be nice.

That's a great idea, I'll make a note to come back and add that in a later PR.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+help implementers of this API to skip making common mistakes.
+
+It may not be very relevant if you are intending to _use_ this API as an end
+user as opposed to _building_ something that uses it.
+
+!!! note
+    This document is officially a part of the Gateway API specification.
+    Requirements in this document labelled as MUST, SHOULD, or MAY must be treated
+    the same as in the detailed specification page. (That is, the words MUST, SHOULD,
+    and MAY must be interpreted as described in RFC 2119.)
```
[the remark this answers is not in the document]
youngnick: Yeah, okay, I'll remove this and the associated change from the spec template file.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+
+The concepts of "bundle version" and "channel" (short for "release channel") are
+explained in our [versioning][versioning] documentation.
+
+Implementations may use these to determine what schema versions are installed in
+the cluster, if any.
+
+[versioning]: /concepts/versioning
+
+### Changes to the Gateway API CRDs are backwards compatible
```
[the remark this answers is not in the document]
youngnick: I understand, but I think that having a summary of the key parts is really important for implementers. I'm not sure how to trim this to tell people about the backward compatibility guarantees in any shorter way.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+Similarly, if an implementation was written with a _higher_ version, the newer
+values that it understands will simply _never be used_, as they are not present
+in the older version.
+
+A similar guarantee occurs between the "experimental" and "standard" channels for
+objects in the same API version, so an implementation may be written with the
+experimental API definitions, but work just fine with having only the standard
+definitions installed - there will be fields or values that will never be used.
+The same applies for an implementation written using the standard API definitions
+running in a cluster with the experimental definitions installed.
```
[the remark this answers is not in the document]
youngnick: Good point, removed.

site-src/guides/implementers-guide.md
youngnick: Thanks for pointing this out, I was intending to say "implementations shouldn't do this without asking a human first" - I've updated, PTAL.

site-src/guides/implementers-guide.md
youngnick: Agreed, this is a great followup. I'll create an issue a bit later for it.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+  than the existing one.
+
+
+Because of our backwards compatibility guarantees, it's also safe for a controller
+to flip the install channel between "standard" and "experimental", although
+implementations MUST NOT do this without consulting the implementation owner.
+
+Automatic CRD installation has the advantage that there is less for the
+implementation user to do; any required version checking can be performed by
+code instead of by the cluster admin.
```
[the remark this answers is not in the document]
youngnick: Added some further text below.

site-src/guides/implementers-guide.md
youngnick: I don't see these changing any time soon, so I'm not worried about this.

site-src/guides/implementers-guide.md
```diff
@@ -0,0 +1,367 @@
...
+associated with some subset of configuration.
+
+For the Gateway case, it's to allow Conditions per _Listener_, and in the Route
+case, it's to allow Conditions per _implementation_ (since Route objects can
+be used in multiple Gateways, and those Gateways can be reconciled by different
+implementations).
+
+In all of these cases, there are some relatively-common Condition types that have
+similar meanings:
+- `Accepted` - the resource or part thereof contains acceptable config that will
```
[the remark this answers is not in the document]
youngnick: Sigh, thanks.

site-src/guides/implementers-guide.md
youngnick: I've tried to keep this very general, with only the main status being captured here. I don't see us changing the `status` handling or `controllerName` behavior - we _can't_ once we go GA.

Maybe the guideline here is that we can only mention stable or core fields?

site-src/guides/implementers-guide.md
youngnick: I was trying to make this reflect what we actually have _right now_, but sure.

site-src/references/implementers-guide.md
```diff
@@ -0,0 +1,344 @@
...
+will indicate that any remaining status is out of date.
+
+
+#### HTTPRoute
+
+HTTPRoutes route HTTP traffic that is _unencrypted_ and available for inspection.
+This allows the HTTPRoute to use HTTP properties, like path, method, or headers
+in its routing directives.
+
+Watched in profiles:
```
robscott, line 315: Looks like this was leftover from earlier.
```suggestion
```

site-src/references/implementers-guide.md
```diff
@@ -0,0 +1,344 @@
...
+types for details, but this usually includes `Accepted`, `Programmed` and
+`ResovledRefs` Conditions.
+- Routes that fall out of scope SHOULD NOT have status updated, since it's possible
+that these updates may overwrite any new owners. The `observedGeneration` field
+will indicate that any remaining status is out of date.
+
+
+#### HTTPRoute
+
+HTTPRoutes route HTTP traffic that is _unencrypted_ and available for inspection.
```
robscott, line 311: I think this may be a bit confusing when combined with BackendTLSPolicy. My understanding is that HTTPRoute can/should be combined with HTTPRoute, but in that (and other) cases, traffic to the backend would be encrypted. I think it might be more accurate to say that this applies to HTTP(S) traffic that is terminated at the Gateway, but not TLS Passthrough traffic. I'm actually not sure how to word this, just need to be careful that we're not being unnecessarily restrictive here.

robscott (commented):
Thanks @youngnick!

/approve

site-src/references/implementers-guide.md
```diff
@@ -0,0 +1,344 @@
+# Gateway API Implementer's Guide
+
+aka
+
+Everything you wanted to know about building a Gateway API implementation
+but were too afraid to ask.
+
```
shaneutt, line 7: For those who don't speak english natively, the `aka` _might_ be a little confusing?

Not really a big deal I guess, but as a suggestion maybe you could just drop it and do this instead:

```suggestion
Everything you wanted to know about building a Gateway API implementation
but were too afraid to ask.

## Overview

```

Or something like that?

site-src/references/implementers-guide.md
youngnick: I updated this wording, PTAL.

robscott (approved):
A couple very tiny non-blocking nits, but this LGTM.

site-src/references/implementers-guide.md
```diff
@@ -0,0 +1,341 @@
...
+code instead of by the cluster admin.
+
+It should also be noted that many infra and cluster admins do this sort of
+automatic CRD management using external management systems that will not be
+visible to a Gateway implementation, so automatic installation MUST be able to
+be disabled by the installation owner (whether that is the infra or cluster
+admin).
+
+Because of all these caveats, we DO NOT recommend doing automatic CRD management
+at this time.
```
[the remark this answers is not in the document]
youngnick: I reworded this pretty substantially, PTAL.

shaneutt (approved):
Great to have something for this now, thanks @youngnick!

robscott:
Thanks @youngnick!

/lgtm
/retest

PR #2465, opened 2023-10-09
Documenting Listener Isolation in Gateway Spec

robscott:
**What type of PR is this?**
/kind cleanup
/kind documentation

This PR is a follow up to #2416. It introduces a new concept called "Listener Isolation" to describe the recommendation that at most one Listener matches a request, and only Routes attached to that Listener are used for routing.

Originally I'd planned to consider this an "Extended" feature (see https://github.com/kubernetes-sigs/gateway-api/issues/2416#issuecomment-1731747678) but generally the idea of "Extended" and "Core" is per-field, not really a description of behavior. So instead I'm using "SHOULD" which per [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) is a recommendation and not a requirement. This does add the requirement that implementations that do not support Listener Isolation must document that.

Related to #2416 but leaving open until conformance tests are present.

```release-note
Gateway: A new concept called "Listener Isolation" has been introduced to describe the recommendation that at most one Listener matches a request, and only Routes attached to that Listener are used for routing.
```

### DIFF
--- a/apis/v1beta1/gateway_types.go
+++ b/apis/v1beta1/gateway_types.go
@@ -132,6 +132,13 @@ type GatewaySpec struct {
 	// matches. For example, `"foo.example.com"` takes precedence over
 	// `"*.example.com"`, and `"*.example.com"` takes precedence over `""`.
 	//
+	// Note that requests SHOULD match at most one Listener. For example, if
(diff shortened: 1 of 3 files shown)


youngnick:
LGTM, but will leave it for @shaneutt to bring this one in.

PR #2467, opened 2023-10-10
Adding venv for verify-yamllint

robscott:
**What type of PR is this?**
/kind cleanup
/kind failing-test

When https://github.com/kubernetes/test-infra/pull/30973 merged it broken our verify-yamllint script with the following error when the script ran `pip3 install yamllint`:

```
× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
```

This is an attempt to resolve that, alternatives welcome. 

```release-note
NONE
```

### DIFF
--- a/hack/verify-yamllint.sh
+++ b/hack/verify-yamllint.sh
@@ -23,9 +23,14 @@ readonly KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
 
 cd "${KUBE_ROOT}"
 
+python3 -m venv .venv
+source .venv/bin/activate
+
 # See configuration file in ${KUBE_ROOT}/.yamllint.yaml.
 pip3 install yamllint==$VERSION
 
 yamllint -c .yamllint.yaml .
+
+deactivate
 
 # ex: ts=2 sw=2 et filetype=sh


robscott:
x-ref https://github.com/kubernetes/test-infra/pull/30998

robscott:
x-ref https://github.com/kubernetes/test-infra/pull/30999

youngnick:
/lgtm

once all the testing stuff is done

hack/verify-yamllint.sh
```diff
@@ -23,9 +23,14 @@ readonly KUBE_ROOT=$(dirname "${BASH_SOURCE}")/..
 
 cd "${KUBE_ROOT}"
 
+python3 -m venv env
+source env/bin/activate
```
BenTheElder, line 27: might need to .gitignore? also might want to either do this repo wide with a requirements.txt or scope an env specifically to this (call it "yamlenv" or something idk?)

BenTheElder (commented):
/lgtm
/hold
(holding in case you want further review, looks like a good starting point!)

hack/verify-yamllint.sh
robscott: 🤦 I actually had a related change locally that I forgot to push. I'm reusing the same name as we're using in our makefile so this should also be covered by .gitignore now. Not running the full thing because 100% of our requirements.txt is for docs generation and irrelevant for this specific yamllint script.

https://github.com/kubernetes-sigs/gateway-api/blob/335be6a634e733ac64ad47e17a7548600cec393e/Makefile#L160-L170

robscott:
Looks like current test runs kicked off before https://github.com/kubernetes/test-infra/pull/31001 merged, starting over.

/retest

robscott:
Thanks to @BenTheElder for all the help with this!

/hold cancel

PR #2472, opened 2023-10-11
add h2c server to echo-basic

dprotaso:
Pulls the image changes from: https://github.com/kubernetes-sigs/gateway-api/pull/2456

/kind test

```release-note
NONE
```


conformance/echo-basic/echo-basic.go
```diff
@@ -152,6 +161,38 @@ func delayResponse(request *http.Request) error {
...
+		Handler:           h2c.NewHandler(handler, &http2.Server{}),
+	}
+	fmt.Printf("Starting server, listening on port %s (h2c)\n", h2cPort)
+	err := h1s.ListenAndServe()
+	if err != nil {
+		errchan <- err
+	}
+}
+
+func h2cHandler(w http.ResponseWriter, r *http.Request) {
```
robscott, line 186: This doesn't look like it's used anymore?

conformance/echo-basic/echo-basic.go
```diff
@@ -152,6 +161,38 @@ func delayResponse(request *http.Request) error {
...
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.ProtoMajor != 2 {
+			w.WriteHeader(http.StatusBadRequest)
+			fmt.Fprint(w, "Expected h2(c) request")
+			return
+		}
+
+		echoHandler(w, r)
+	})
+	h1s := &http.Server{
```
robscott, line 174: Don't you mean `h2s` here?

robscott (commented):
Thanks @dprotaso!

conformance/echo-basic/echo-basic.go
dprotaso: so there's an http2 package in x/net that has a `Server` - so it's technically not an h2c server

dprotaso: I'll just call it h2c server

dprotaso:
Tested with the following commands

```sh
$ curl  -vvv --http2-prior-knowledge localhost:3001
succeeds

$ curl  -vvv --http2 localhost:3001
succeeds

$ curl  -vvv localhost:3001
fails (as expected)
```

robscott:
Thanks @dprotaso!

/lgtm

PR #2466, opened 2023-10-09
v1.0.0 API Review

robscott:
**What type of PR is this?**
/kind documentation
/kind api-change

**What this PR does / why we need it**:
This PR is intended to review the changes added to the API as part of the v1.0.0 release. This is effectively a diff of /apis in main vs /apis in the release-0.8 branch. This PR will not merge, it's just a means of reviewing this API changes.

/assign @khenidak @thockin @MikeZappa87

PR #2476, opened 2023-10-12
Cleanup related to v1.0.0 review

robscott:
**What type of PR is this?**
/kind cleanup

Took some time to review #2466 and ended up with this PR with lots of little bits of cleanup. Notable changes:

* Removed lingering references to `Standard` field in BackendTLSPolicy
* Changed BackendTLSPolicy to use LocalObjectReference for now since we're saying that cross-namespace refs are invalid for now
* Added `hack/supportedfeatures` util to simplify enum validation on new `SupportedFeatures` field in GatewayClass status
* Added enum validation to `SupportedFeatures`
* Clarified that a policy should be considered "unimplementable" of status is full and cannot be populated

```release-note
NONE
```


apis/v1beta1/gatewayclass_types.go
```diff
@@ -262,3 +262,8 @@ type GatewayClassList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []GatewayClass `json:"items"`
 }
+
+// SupportedFeature is used to describe distinct features that are covered by
+// conformance tests.
+// +kubebuilder:validation:Enum=Gateway;GatewayPort8080;GatewayStaticAddresses;HTTPRoute;HTTPRouteDestinationPortMatching;HTTPRouteHostRewrite;HTTPRouteMethodMatching;HTTPRoutePathRedirect;HTTPRoutePathRewrite;HTTPRoutePortRedirect;HTTPRouteQueryParamMatching;HTTPRouteRequestMirror;HTTPRouteRequestMultipleMirrors;HTTPRouteResponseHeaderModification;HTTPRouteSchemeRedirect;Mesh;ReferenceGrant;TLSRoute
```
youngnick, line 268: This is nice but doesn't feel very sustainable - I don't know what else to do here though.

apis/v1beta1/object_reference_types.go
```diff
@@ -145,42 +145,3 @@ type BackendObjectReference struct {
 	// +optional
 	Port *PortNumber `json:"port,omitempty"`
 }
-
-// ObjectReference identifies an API object including its namespace, with no
```
youngnick, line 149: Yeah, I guess taking this out for now is fine - I originally left it in because I think we will need to swap to it eventually, but  moving from LocalObjectReference to this is not too breaking later.

youngnick (commented):
LGTM with a couple of non-blocking things.

apis/v1beta1/gatewayclass_types.go
robscott: Agreed. Added a util that will generate this for us at least, but still not great. Decided it's safer to err towards more restrictive validation to start, we can always loosen later.

apis/v1beta1/shared_types.go
```diff
@@ -246,27 +246,25 @@ type PortNumber int32
...
+// protocol then the backend is considered invalid. Implementations MUST set the
+// "ResolvedRefs" condition to "False" with the "UnsupportedProtocol" reason.
 //
 // </gateway:experimental:description>
 //
 // Note that when the BackendTLSPolicy object is enabled by the implementation,
 // there are some extra rules about validity to consider here. See the fields
 // where this struct is used for more information about the exact behavior.
-//
-// [KEP-3726]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3726-standard-application-protocols
```
robscott, line 269: For the record, I initially tried to keep this link but just move it up above the `</gateway:experimental:description>` tag, but then `gofumpt` failed. I think not having a link here is better than this link mysteriously appearing in standard without context, so went with that, but open to any follow ups to add this back in only for experimental.

dprotaso: Yeah I had the same problem

hack/supportedfeatures/main.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
shaneutt, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

shaneutt (approved):

hack/supportedfeatures/main.go
robscott: 🤦 fixed, thanks!

PR #2477, opened 2023-10-12
Conformance: Adds Test to Exercise Gateway AttachedRoutes

### LINKED ISSUE #2402: Conformance: Add coverage for AttachedRoutes, ensuring count is accurate regardless of Listener or Route validity

Following on from https://github.com/kubernetes-sigs/gateway-api/pull/2396 we should add conformance that covers the exact semantics defined in the API for implementations when setting the Gateway Listener status `AttachedRoutes` count field.

Namely:
- `AttachedRoutes` should be set on Listeners that are valid or invalid
- The count of `AttachedRoutes` should include Routes that are valid or invalid

**Why this is needed**:

The spec now defines the `AttachedRoutes` field to document the number of Route resources that are successfully related to a Gateway, regardless of the validity of either side of the relationship.

danehans:
**What type of PR is this?**
/kind test

/area conformance

Fixes #2402

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -36,6 +36,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 	Description: "A Gateway in the gateway-conformance-infra namespace should be attached to routes.",
 	Features: []suite.SupportedFeature{
 		suite.SupportGateway,
+		suite.SupportHTTPRoute,
 	},
 	Manifests: []string{"tests/gateway-with-attached-routes.yaml"},
 	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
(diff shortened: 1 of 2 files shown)


robscott:
Thanks @danehans!

/cc @arkodg @mlavacca @sunjayBhatia

arkodg (approved):
/approve
thanks @danehans, hope this has been tested with at least one implementation

danehans:
> thanks @danehans, hope this has been tested with at least one implementation

@arkodg I have not tested this with an implementation yet but I'll start doing that today. Let's /hold until I complete this task.

arkodg:
> > thanks @danehans, hope this has been tested with at least one implementation
> 
> @arkodg I have not tested this with an implementation yet but I'll start doing that today. Let's /hold until I complete this task.

thanks, I hope in the future there are more automated ways of doing this in this project itself, but unfortunately for now this burden is falling on the contributor and drops the overall velocity of constributions

danehans:
I successfully tested this PR against Envoy Gateway. I had to use [this patch](https://github.com/envoyproxy/gateway/pull/2003) since Envoy Gateway does not pass this test until https://github.com/envoyproxy/gateway/pull/1918 is fixed. I also had to use v1beta1 APIs since https://github.com/envoyproxy/gateway/issues/1160 is required for Envoy Gateway to support v1 APIs.

arkodg (approved):
thanks !
/approve

mlavacca (approved):
Thanks @danehans!

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -90,6 +90,52 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 
 			kubernetes.GatewayStatusMustHaveListeners(t, s.Client, s.TimeoutConfig, gwNN, listeners)
 		})
+
+		t.Run("Gateway listener should have an accepted http route attached when unresolved refs exist", func(t *testing.T) {
```
robscott, line 94: Maybe something like this
```suggestion
		t.Run("Gateway listener should have AttachedRoutes set even when Gateway has unresolved refs", func(t *testing.T) {
```

robscott (commented):
Thanks for adding this @danehans! One other nit - I think the tests in this file should also depend on the `SupportHTTPRoute` feature.

conformance/tests/gateway-with-attached-routes.go
danehans: Commit 8b76391  implements ^ comment.

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -36,6 +36,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 	Description: "A Gateway in the gateway-conformance-infra namespace should be attached to routes.",
 	Features: []suite.SupportedFeature{
 		suite.SupportGateway,
+		suite.SupportHTTPRoute,
```
danehans, line 39: @robscott I added this line and L149 based on your comments in https://github.com/kubernetes-sigs/gateway-api/pull/2477#pullrequestreview-1690473146.

youngnick:
/lgtm

Thanks @danehans!

PR #2478, opened 2023-10-12
Checkout go.work and go.work.sum from the repository

gauravkghildiyal:
**What type of PR is this?**
/kind cleanup

The proposal for [Multi-Module go workspaces in go](https://go.googlesource.com/proposal/+/master/design/45713-workspace.md) suggests that we should avoid checking in go.work and related files into the repo:

_"go.work files should not be checked into the repositories so that they don‘t override the workspaces users explicitly define. Checking in go.work files could also lead to CI/CD systems not testing the actual set of version requirements on a module and that version requirements among the repository’s modules are properly incremented to use changes in the modules."_

Fixes #

```release-note
NONE
```


robscott:
Thanks @gauravkghildiyal!

/lgtm
/approve

gauravkghildiyal:
Somehow the commit to actually remove the file vanished. On it's way

dprotaso:
Although some guidance said it's useful for multi-module repos I did find the flow annoying. Since a change in say `echo-basic` submodule required me to then run `go test` or the equivalent at the top level in order to update the `go.work.sum`

If anyone knows any go team devs - it would be good to surface this friction with them

PR #2479, opened 2023-10-13
Include make target for bumping echo-basic & advanced in our conformance files

dprotaso:
**What type of PR is this?**
/kind cleanup
/area conformance
/kind test

This PR introduces a new make target - `update-conformance-image-refs` it will trigger a new bash script that will update the `echo-basic` & `echo-advanced` to the latest tag (sorted by semver)

I'm including the output of running the script - which is bumping both images to the latest semver tag

I also fixed the mesh yamls to use echo-advanced instead of basic - that was introduced in https://github.com/kubernetes-sigs/gateway-api/pull/2468/files

```release-note
Update mesh conformance to point to the correct images
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,6 +126,10 @@ verify:
 docs:
 	hack/make-docs.sh
 
+.PHONY: update-conformance-image-refs
+update-conformance-image-refs:
+	hack/update-conformance-image-refs.sh
+
 # Verify if support Docker Buildx.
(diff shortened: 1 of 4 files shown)


robscott:
Thanks @dprotaso!

/lgtm

dprotaso:
@robscott @shaneutt PTAL - I rebased and updated the images

PR #2491, opened 2023-10-17
BackendTLSPolicy cel tests

### LINKED ISSUE #2473: BackendTLSPolicy: Add CEL Tests
Tests to ensure the CEL validation added in #2448 is working as expected.

**Why this is needed**:
All other CEL validations have corresponding go tests in https://github.com/kubernetes-sigs/gateway-api/tree/main/pkg/test/cel. Without it, it's awfully hard to ensure that the CEL we've written covers all the inputs we want to allow or prevent.

Ygnas:
**What type of PR is this?**

/kind test

Without it, it's awfully hard to ensure that the CEL we've written covers all the inputs we want to allow or prevent.

Fixes #2473 

```release-note
NONE
```

### DIFF
--- a/apis/v1alpha2/backendtlspolicy_types.go
+++ b/apis/v1alpha2/backendtlspolicy_types.go
@@ -75,8 +75,8 @@ type BackendTLSPolicySpec struct {
 }
 
 // BackendTLSPolicyConfig contains backend TLS policy configuration.
-// +kubebuilder:validation:XValidation:message="must not contain both CertRefs and WellKnownCACerts",rule="(has(self.caCertRefs) && size(self.caCertRefs) > 0 && has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
-// +kubebuilder:validation:XValidation:message="must specify either CertRefs or WellKnownCACerts",rule="!(has(self.caCertRefs) && size(self.caCertRefs) > 0 || has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
+// +kubebuilder:validation:XValidation:message="must not contain both CACertRefs and WellKnownCACerts",rule="!(has(self.caCertRefs) && size(self.caCertRefs) > 0 && has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
+// +kubebuilder:validation:XValidation:message="must specify either CACertRefs or WellKnownCACerts",rule="(has(self.caCertRefs) && size(self.caCertRefs) > 0 || has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
 type BackendTLSPolicyConfig struct {
 	// CACertRefs contains one or more references to Kubernetes objects that
 	// contain a PEM-encoded TLS CA certificate bundle, which is used to
--- a/config/crd/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
@@ -178,12 +178,12 @@ spec:
                 - hostname
                 type: object
                 x-kubernetes-validations:
-                - message: must not contain both CertRefs and WellKnownCACerts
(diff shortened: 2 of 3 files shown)


robscott:
Thanks @Ygnas!

/ok-to-test
/approve
/cc @candita

config/crd/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
```diff
@@ -178,11 +178,11 @@ spec:
                 - hostname
                 type: object
                 x-kubernetes-validations:
-                - message: must not contain both CertRefs and WellKnownCACerts
-                  rule: (has(self.caCertRefs) && size(self.caCertRefs) > 0 && has(self.wellKnownCACerts)
-                    && self.wellKnownCACerts != "")
-                - message: must specify either CertRefs or WellKnownCACerts
-                  rule: '!(has(self.caCertRefs) && size(self.caCertRefs) > 0 || has(self.wellKnownCACerts)
+                - message: must not contain both CACertRefs and WellKnownCACerts
```
gauravkghildiyal, line 181: I think we would want to change this at the source of truth: https://github.com/kubernetes-sigs/gateway-api/blob/adbd9968d4ff7995af21cb9cd0913288e2b4ca27/apis/v1alpha2/backendtlspolicy_types.go#L78-L79 ...followed by running ./hack/update-codegen.sh to generate the CRD yaml.

pkg/test/cel/backendtlspolicy_test.go
```diff
@@ -0,0 +1,198 @@
...
+					},
+					TLS: tc.routeConfig,
+				},
+			}
+			validateBackendTLSPolicy(t, route, tc.wantErrors)
+		})
+	}
+}
+
+func TestPolicyTargetReferenceWithSectionName(t *testing.T) {
```
gauravkghildiyal, line 115: nit: 

More tests are definitely good to have, but maybe we should not mix non-CEL validations (`spec.targetRef.name in body should be at least 1 chars long` and others) and CEL validations errors in these tests?

I can be corrected, but I believe we have https://github.com/kubernetes-sigs/gateway-api/tree/main/hack/invalid-examples for non-CEL validation testing.

gauravkghildiyal (commented):
Thanks for the changes @Ygnas !

Ygnas:
I'm not 100% sure if we should test `//+kubebuilder:validation:MaxItems=8` here, also, should we be testing `LocalObjectReference` in this context?

gauravkghildiyal:
@Ygnas The immediate intent is to just verify the two validations defined through `+kubebuilder:validation:XValidation`. So I think what we have now should be sufficient. 

Thanks for taking this on.

LGTM (will defer approval to maintainers)

apis/v1alpha2/backendtlspolicy_types.go
```diff
@@ -75,8 +75,8 @@ type BackendTLSPolicySpec struct {
 }
 
 // BackendTLSPolicyConfig contains backend TLS policy configuration.
-// +kubebuilder:validation:XValidation:message="must not contain both CertRefs and WellKnownCACerts",rule="(has(self.caCertRefs) && size(self.caCertRefs) > 0 && has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
-// +kubebuilder:validation:XValidation:message="must specify either CertRefs or WellKnownCACerts",rule="!(has(self.caCertRefs) && size(self.caCertRefs) > 0 || has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
+// +kubebuilder:validation:XValidation:message="must not contain both CACertRefs and WellKnownCACerts",rule="!(has(self.caCertRefs) && size(self.caCertRefs) > 0 && has(self.wellKnownCACerts) && self.wellKnownCACerts != \"\")"
```
youngnick, line 78: And there's a great example of why tests are important! Thanks @Ygnas.

youngnick:
Yes, I think that testing the max items or LocalObjectReference are not desirable here - there is a good chance that both of those will change soon anyway.

/lgtm

PR #2495, opened 2023-10-19
[GEP-1911] websocket backend protocol conformance

dprotaso:
/kind test
/area conformance

This PR adds a conformance test validating implementations support websocket over cleartext when the target Kubernetes Service Service Port has a `kubernetes.io/ws` `appProtocol`

This PR includes echo-basic changes so it will need to bumped after this PR

Related https://github.com/kubernetes-sigs/gateway-api/issues/1911 https://github.com/kubernetes-sigs/gateway-api/issues/205

```release-note
NONE
```

### DIFF
--- a/conformance/base/manifests.yaml
+++ b/conformance/base/manifests.yaml
@@ -100,6 +100,11 @@ spec:
     appProtocol: kubernetes.io/h2c
     port: 8081
     targetPort: 3001
+  - name: third-port
+    protocol: TCP
+    appProtocol: kubernetes.io/ws
+    port: 8082
+    targetPort: 3000
 ---
 apiVersion: apps/v1
 kind: Deployment
--- a/conformance/echo-basic/echo-basic.go
+++ b/conformance/echo-basic/echo-basic.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"regexp"
@@ -31,6 +32,7 @@ import (
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
+	"golang.org/x/net/websocket"
 )
 
 // RequestAssertions contains information about the request and the Ingress
@@ -100,6 +102,7 @@ func main() {
 	httpMux.HandleFunc("/health", healthHandler)
 	httpMux.HandleFunc("/status/", statusHandler)
 	httpMux.HandleFunc("/", echoHandler)
+	httpMux.Handle("/ws", websocket.Handler(wsHandler))
 	httpHandler := &preserveSlashes{httpMux}
 
 	errchan := make(chan error)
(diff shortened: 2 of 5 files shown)


conformance/base/manifests.yaml
```diff
@@ -91,9 +91,15 @@ spec:
   selector:
     app: infra-backend-v1
   ports:
-  - protocol: TCP
+  - name: first-port
+    protocol: TCP
     port: 8080
     targetPort: 3000
+  - name: third-port
```
dprotaso, line 98: second port is going to be h2c

https://github.com/kubernetes-sigs/gateway-api/pull/2456/files#diff-21fe4de356049035b12df8b9efd2e0d29372189a548cd144caa363ca875e4f1dR98

conformance/echo-basic/echo-basic.go
```diff
@@ -130,6 +133,11 @@ func main() {
 	}
 }
 
+func wsHandler(ws *websocket.Conn) {
+	fmt.Println("established websocket connection", ws.RemoteAddr())
+	io.Copy(ws, ws)
```
arkodg, line 138: this is a one liner to ws.Read() and ws.Write(..) ?

dprotaso: yeah websocket.Conn is an `io.ReadWriter` so it's reading frames and writing them back

Got it from https://pkg.go.dev/golang.org/x/net/websocket#example-Handler

arkodg: cool, comment here would be great

arkodg:
this PR looks good, lets wait for https://github.com/kubernetes-sigs/gateway-api/pull/2456 to land first, since this PR will need a rebase

conformance/echo-basic/echo-basic.go
mlavacca: +1

conformance/tests/httproute-backend-protocol-ws.yaml
```diff
@@ -0,0 +1,18 @@
...
+  parentRefs:
+  - name: same-namespace
+  rules:
+  - backendRefs:
+    # This points to a Service with the following ServicePort
+    # - name: third-port
+    #   appProtocol: kubernetes.io/ws
+    #   protocol: TCP
+    #   port: 8082
+    # targetPort: 3000
```
mlavacca, line 16:
```suggestion
    #   targetPort: 3000
```

conformance/base/manifests.yaml
```diff
@@ -91,9 +91,15 @@ spec:
...
+    protocol: TCP
+    appProtocol: kubernetes.io/ws
+    port: 8082
+    targetPort: 3000
```
mlavacca, line 102: Do we really want to use the same targetPort for both TCP and WS traffic?

dprotaso: What is your concern about using the same targetPort?

mlavacca: I have no real concern, I was just wondering if this choice was backed by a specific reason. If not, I think we should prefer not to use the same back-side port for multiple front-side ports.

mlavacca: ok, sounds good to me 👍

mlavacca (approved):
Thanks @dprotaso, ltgm!

PR #2511, opened 2023-10-23
add v1 to various client schemes

dprotaso:
**What type of PR is this?**
/kind cleanup
/kind test
/area conformance

Gateway API v1 APIs need to be added the scheme. Discovered in https://github.com/kubernetes-sigs/gateway-api/pull/2456#pullrequestreview-1690743992 - thanks @sunjayBhatia 

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/conformance_test.go
+++ b/conformance/conformance_test.go
@@ -20,6 +20,7 @@ package conformance_test
 import (
 	"testing"
 
+	v1 "sigs.k8s.io/gateway-api/apis/v1"
 	"sigs.k8s.io/gateway-api/apis/v1alpha2"
 	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	"sigs.k8s.io/gateway-api/conformance/tests"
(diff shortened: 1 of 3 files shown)


sunjayBhatia (approved):

conformance/conformance_test.go
```diff
@@ -20,6 +20,7 @@ package conformance_test
 import (
 	"testing"
 
+	v1 "sigs.k8s.io/gateway-api/apis/v1"
```
sunjayBhatia, line 23: ha maybe don't need the alias for the files that don't already

maybe we should have a lint rule to make them consistent, but that can be done separately

dprotaso: >ha maybe don't need the alias for the files that don't already

goimports forces it - i'm gonna assume verify will pass. If we want to tweak it I'd say let's do this separate from this PR

sunjayBhatia (approved):

robscott:
Thanks @dprotaso!

/lgtm
/approve

Issue #2513, opened 2023-10-24
ExemptFeatures knob doesnt work for Experimental Conformance Profile


arkodg:
**What happened**:

Trying to run all the conformance tests related to Ingress by setting `SupportedFeatures: suite.AllFeatures` to run all tests and also setting `ExemptFeatures: suite.MeshCoreFeatures` to disable tests tied to mesh

Sample config
```
		suite.ExperimentalConformanceOptions{
			Options: suite.Options{
				Client:     mgrClient,
				RestConfig: cfg,
				// This clientset is needed in addition to the client only because
				// controller-runtime client doesn't support non CRUD sub-resources yet (https://github.com/kubernetes-sigs/controller-runtime/issues/452).
				Clientset:            k8sClientset,
				GatewayClassName:     *flags.GatewayClassName,
				Debug:                *flags.ShowDebug,
				CleanupBaseResources: *flags.CleanupBaseResources,
				SupportedFeatures:    suite.AllFeatures,
				ExemptFeatures: suite.MeshCoreFeatures,
			},
			Implementation:      *implementation,
			ConformanceProfiles: conformanceProfiles,
		})
```		

Running with this config causes the mesh tests to run, which is not respecting the `ExemptFeatures` knob
```
2023-10-24T00:29:27.1815127Z     --- PASS: TestExperimentalConformance/MeshBasic (0.28s)
2023-10-24T00:29:27.1816429Z         --- PASS: TestExperimentalConformance/MeshBasic/0_request_to_'echo'_should_receive_a_200 (0.27s)
2023-10-24T00:29:27.1817657Z     --- FAIL: TestExperimentalConformance/MeshConsumerRoute (30.13s)
2023-10-24T00:29:27.1819337Z         --- FAIL: TestExperimentalConformance/MeshConsumerRoute/request_from_consumer_route's_namespace_modified_by_HTTPRoute (30.00s)
2023-10-24T00:29:27.1821484Z         --- PASS: TestExperimentalConformance/MeshConsumerRoute/request_not_from_consumer_route's_namespace_not_modified_by_HTTPRoute (0.10s)
2023-10-24T00:29:27.1896361Z     --- FAIL: TestExperimentalConformance/MeshFrontendHostname (30.13s)
2023-10-24T00:29:27.1899950Z         --- FAIL: TestExperimentalConformance/MeshFrontendHostname/Send_to_service_with_wrong_hostname (30.00s)
[...]

PR #2515, opened 2023-10-24
implement exemptFeatures for conformance profile

arkodg:
Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/2513

**What type of PR is this?**

Implement the ExemptFeatures setting in Experimental Conformance Profile

Fixes # https://github.com/kubernetes-sigs/gateway-api/issues/2513

```release-note
Implemented the `ExemptFeatures` field for Experimental Conformance Profiles
```

### DIFF
--- a/conformance/utils/suite/experimental_suite.go
+++ b/conformance/utils/suite/experimental_suite.go
@@ -142,6 +142,10 @@ func NewExperimentalConformanceTestSuite(s ExperimentalConformanceOptions) (*Exp
 		}
 	}
 
+	for feature := range s.ExemptFeatures {
+		s.SupportedFeatures.Delete(feature)
+	}
+
 	if s.FS == nil {
 		s.FS = &conformance.Manifests
 	}
--- a/conformance/utils/suite/experimental_suite.go
+++ b/conformance/utils/suite/experimental_suite.go
@@ -138,6 +138,10 @@ func NewExperimentalConformanceTestSuite(s ExperimentalConformanceOptions) (*Exp
 					}
 					suite.extendedUnsupportedFeatures[conformanceProfileName].Insert(f)
 				}
+				// Add Exempt Features into unsupported features list
+				if s.ExemptFeatures.Has(f) {
+					suite.extendedUnsupportedFeatures[conformanceProfileName].Insert(f)
+				}
 			}
 		}
 	}


conformance/utils/suite/experimental_suite.go
```diff
@@ -142,6 +142,10 @@ func NewExperimentalConformanceTestSuite(s ExperimentalConformanceOptions) (*Exp
 		}
 	}
 
+	for feature := range s.ExemptFeatures {
+		s.SupportedFeatures.Delete(feature)
+	}
```
robscott, line 147: Are `ExemptFeatures` accounted for in display of conformance report? Just want to make sure we're not opening a way to have misleading conformance reports here.

robscott (commented):
Thanks for finding and fixing this @arkodg!

conformance/utils/suite/experimental_suite.go
arkodg: I can add another code block in here https://github.com/kubernetes-sigs/gateway-api/blob/6d0e6ccbccc73def46bce2e52ed50ca1e480a890/conformance/utils/suite/experimental_suite.go#L239 which sets the `testSkipped` if the test is an exempt features test

robscott: Maybe it would be fair to just ensure `exemptFeatures` are added to `unsupportedFeatures` list? Example: https://github.com/kubernetes-sigs/gateway-api/blob/9c830e50acbf0844370210c8f54354b4e062f768/conformance/reports/v0.8.0/nginxinc-nginx-gateway-fabric.yaml#L27-L37

I don't know if exemptFeatures are somehow different than unsupportedFeatures though. Maybe those concepts should just be merged altogether?

arkodg: I also think its not required, because the test is not supposed to run
cc @mlavacca

mlavacca: +1 on merging `exemptedFeatures` into `unsupportedFeatures`. A potential problem that comes to my mind is that the core features for the requested profiles should not be exempted. For example, when setting the HTTP profile, it should be disallowed exempting  the`HTTPRoute`, `Gateway`, and `ReferenceGrant` features.

arkodg: updated the PR with above suggestion, ptal @mlavacca @robscott

conformance/utils/suite/experimental_suite.go
```diff
@@ -138,10 +138,18 @@ func NewExperimentalConformanceTestSuite(s ExperimentalConformanceOptions) (*Exp
 					}
 					suite.extendedUnsupportedFeatures[conformanceProfileName].Insert(f)
 				}
+				// Add Exempt Features into unsupported features list
+				if s.ExemptFeatures.Has(f) {
+					suite.extendedUnsupportedFeatures[conformanceProfileName].Insert(f)
+				}
```
robscott, line 144: @arkodg can you add a follow up issue to prevent core features from being specified as "Exempt"?

robscott (commented):
Thanks @arkodg!

/lgtm

conformance/utils/suite/experimental_suite.go
arkodg: raised https://github.com/kubernetes-sigs/gateway-api/issues/2524

PR #2531, opened 2023-10-26
Go maintenance for v1.0.0

### LINKED ISSUE #1970: GA golang maintenance checklist
In addition to `v1.0.0` being the first GA release of the APIs where we will want to maintain backwards compatibility thereafter, this will also be the first release of the Go libraries that will be in the same situation.

The purpose of this issue is to double check a few Golang related maintenance items before we release GA:

- [x] we've updated our version of Go in all places (images, e.t.c.) to the latest appropriate major/minor release
- [x] `go.mod` doesn't have any pins to any `main`/`master` branches
- [x] we're up-to-date on our k8s go dependencies

shaneutt:
**What type of PR is this?**
/kind cleanup

This does some basic Go version and dependency maintenance prior to our release, as per #1970.

We already had updated to Go 1.21 for our container builds, but just not in the root `go.mod` files yet, so this fixes that.

Fixes #1970

```release-note
NONE
```


robscott:
Thanks @shaneutt!

/lgtm

PR #2534, opened 2023-10-26
Marking SupportedVersion as experimental

robscott:
**What type of PR is this?**
/kind documentation

Although this was only a new condition, it feels strange to add it immediately to a GA API without any soak period. This is not really going to have any affect on CRD generation or generated clients, this really just is meant to signal that details of the condition could still change and it does not yet have GA stabililty.

```release-note
GatewayClass "SupportedVersion" condition has been formally marked as experimental.
```


youngnick:
/lgtm

once the tests pass though.

Issue #2526, opened 2023-10-26
Conformance test: Expected accepted condition of HTTPRoute in the case GatewayWithAttachedRoutes/unresolved-gateway-with-one-attached-unresolved-route is not clear


randmonkey:
**What happened**:

https://github.com/kubernetes-sigs/gateway-api/blob/de8675f27a35bae354027a832b1335cc167a96c2/conformance/tests/gateway-with-attached-routes.go#L126-L128 
The name of the expected accepted condition is `notAccepted`, which indicate that the condition should have status `false`. However, the `Status` field is set to `true`, which is opposite to what the name indicates.

**What you expected to happen**:

Have a clear and consistent status of expected `Accepted` condition in status of testing `HTTPRoute`

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

arkodg:
/close
fixed with #2535

PR #2535, opened 2023-10-27
Fix bugs in GatewayWithAttachedRoutes

arkodg:
**What type of PR is this?**

* Set status to `False` for `notAccepted` Route condition

* Remove `Accepted` condition from `status.listeners[0].conditions` since it is not mandatory for the `Accepted` condition to be set. In this case the `ResolvedRefs=False` due to `InvalidCertificateRef` so an implementation may choose to not generate any listener config

Also Fixes # https://github.com/kubernetes-sigs/gateway-api/issues/2526

```release-note
Fixed the status condition requirements in the `GatewayWithAttachedRoutes` test
```

### DIFF
--- a/conformance/tests/gateway-with-attached-routes.go
+++ b/conformance/tests/gateway-with-attached-routes.go
@@ -102,11 +102,6 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 				}},
 				Conditions: []metav1.Condition{
 					{
-						Type:   string(v1.ListenerConditionAccepted),
-						Status: metav1.ConditionTrue,
-						Reason: "", // any reason
-					},
-					{
 						Type:   string(v1.ListenerConditionProgrammed),
 						Status: metav1.ConditionFalse,
 						Reason: "", // any reason
@@ -125,7 +120,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 			hrouteNN := types.NamespacedName{Name: "http-route-4", Namespace: "gateway-conformance-infra"}
 			notAccepted := metav1.Condition{
 				Type:   string(v1.RouteConditionAccepted),
-				Status: metav1.ConditionTrue,
+				Status: metav1.ConditionFalse,
 				Reason: "", // any reason
 			}
 			unresolved := metav1.Condition{


arkodg:
cc @danehans

youngnick:
Uh, where do we say that the `Accepted` condition is not mandatory? That is a mistake, it should be mandatory.

arkodg:
@youngnick based on the definition here https://github.com/kubernetes-sigs/gateway-api/blob/cb82fb525094bc4ae18e5150ec7b25de76c0877c/apis/v1/gateway_types.go#L918
a implementation should set the Listener Condition to `Accepted`, only if it generates any valid data plane config.
For this case, since `gateway-conformance-infra/does-not-exist` secret does not exist, the implementation may choose to not generate any data plane config, so imo the conformance test should not enforce the `Accepted=True` condition

youngnick:
I think that we may need to update that definition - as part of clarifying AttachedRoutes, we've effectively relaxed that so that Accepted means more "Accepted for processing" rather than "accepted for config".

@robscott, you need to see this one.

youngnick:
@robscott and I talked about this, and agreed that this change is good for now, but I've also logged #2536 to come back and clear this up a bit.

robscott:
Thanks @arkodg! Also thanks to @youngnick for all the thought put into this one!

/lgtm

conformance/tests/gateway-with-attached-routes.go
```diff
@@ -125,7 +120,7 @@ var GatewayWithAttachedRoutes = suite.ConformanceTest{
 			hrouteNN := types.NamespacedName{Name: "http-route-4", Namespace: "gateway-conformance-infra"}
 			notAccepted := metav1.Condition{
 				Type:   string(v1.RouteConditionAccepted),
-				Status: metav1.ConditionTrue,
+				Status: metav1.ConditionFalse,
```
sayboras, line 123: is it conflicting with https://github.com/kubernetes-sigs/gateway-api/blob/faba2585346ef0806585f8ce2ee458e9cd70ec7b/conformance/tests/httproute-invalid-backendref-nonexistent.go#L48 ?

michaelbeaumont: Yeah. I don't see how an implementation can pass both of these tests.
We have [this resource that should be `Accepted`](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-invalid-backendref-nonexistent.yaml) as @sayboras mentioned that's essentially identical to [the resource in the test from this PR](https://github.com/kubernetes-sigs/gateway-api/blob/93da4e4621151e72d1ef3b4515611bc5d1d31adf/conformance/tests/gateway-with-attached-routes.yaml#L118-L132) that should NOT be `Accepted`.

michaelbeaumont: Addressed in

Issue #1639, opened 2023-01-10
The client is missing Apply() functions, making use alongside other client-go packages difficult


therealmitchconnors:
Since K8s 1.22, client-go has supported server-side-apply through functions like Apply(), ApplyStatus(), etc.  Because the gateway api client library lacks these functions, it must be called in non-idiomatic ways, increasing friction for those who intend to automatically control gateways.

nikola-jokic:
Can I take this one :relaxed: ?

howardjohn:
Just FYI this is going to have a real world negative impact on all users: https://github.com/kubernetes/kubernetes/issues/101384

dprotaso:
Could you maintain the server-side apply client in a separate package - so it's opt-in?

Sajiyah-Salat:
Hello, I want to work on this issue. But I cant see the whole picture. A more detailed description and some resource links would be helpful. Thank you.

ashutosh887:
I would like to work on this @shaneutt @therealmitchconnors

shaneutt:
Sounds good @ashutosh887, but since I've noticed that you've taken some other issues on in the last couple days: please note that this one could take some significant time, let us know if you get stuck or need any help along the way! :vulcan_salute: 

/assign @ashutosh887

ashutosh887:
I'll try to work on it @shaneutt 
Sure I'll let you know if I face any difficulty

shaneutt:
@ashutosh887 do you still have time for this? Anything we can do to help support you?

therealmitchconnors:
@shaneutt what needs to happen to move this forward?  It was disappointing to see it removed from the GA tracker.  SSA is the standard that K8s seems to be trying to get everyone to move to, but lack of support in critical libraries makes doing the right thing very difficult...

dprotaso:
@howardjohn are the concerns in your [comment](https://github.com/kubernetes-sigs/gateway-api/issues/1639#issuecomment-1382166741) still relevant?

Bharadwajshivam28:
Hey @shaneutt I can take on this issue but I need some more details like steps to get started

dprotaso:
hey @Bharadwajshivam28 there's already a PR out of for this - https://github.com/kubernetes-sigs/gateway-api/pull/2537

Forgot to link it back to the issue.

Secondly, keep an eye on who's assigned to the issue cause if there's already someone assigned you don't want two people creating duplicate work.

PR #2537, opened 2023-10-27
Generate Gateway API ApplyConfiguration and update Clientsets to support ServerSideApply

dprotaso:
**What type of PR is this?**
/kind feature

[Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) is a Kubernetes feature that allows multiple controllers to manage the same single resource.

N/A

```release-note
Gateway API clientsets now support Server-Side Apply
```

**Additional Notes**:
To simplify review I suggest folks go through each commit

Fixes: https://github.com/kubernetes-sigs/gateway-api/issues/1639


robscott:
Thanks @dprotaso! While this looks reasonable, it's a large diff and doesn't seem like it's strictly necessary for v1.0.0. Since we've already passed RC2, I think we should wait for the next patch or minor release to include this.

/hold

therealmitchconnors:
Thank you, @dprotaso, I am really looking forward to using this!

It may be helpful to add Extract functions, similar to https://github.com/kubernetes/client-go/blob/master/applyconfigurations/core/v1/configmap.go#L62.

dprotaso:
It's there at the top-level - eg. https://github.com/kubernetes-sigs/gateway-api/pull/2537/files#diff-d518444c0b279f513a665c7bfc148619760c1e8fb8824dbde9896dbd03eb454bR61

Do you see it missing for a type?

robscott:
Hey @dprotaso, sorry I'd lost track of this one! Do you mind adding something to the [Implementer's Guide](https://gateway-api.sigs.k8s.io/guides/implementers/) showing how implementations could make use of this? 

/hold cancel

dprotaso:
rebased

cmd/modelschema/main.go
```diff
@@ -0,0 +1,96 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// From https://github.com/kubernetes/kubernetes/blob/0712728ee9caab4d04ec0a751855aca8c891b690/pkg/generated/openapi/cmd/models-schema/main.go
+
+package main
```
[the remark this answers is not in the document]
dprotaso: Created a slack thread - https://kubernetes.slack.com/archives/C0123CNN8F3/p1708650369536409

site-src/guides/implementers.md
```diff
@@ -27,6 +27,13 @@ like APIs delivered in the core Kubernetes binaries, each time a release happens
 the APIs have been reviewed by upstream Kubernetes reviewers, just like the APIs
 delivered in the core binaries.
 
+### API Semantics
+
+Gateway API MUST follow the Kubernetes API semantics. 
+The concepts are documented [here](https://kubernetes.io/docs/reference/using-api/api-concepts/).
```
robscott, line 33: I'm not sure how these are related to Server-Side Apply?

site-src/guides/implementers.md
```diff
@@ -27,6 +27,13 @@ like APIs delivered in the core Kubernetes binaries, each time a release happens
...
+[Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) should also be 
+supported as it allows multiple client appliers to collaborate on a single object.
```
robscott, line 35: May be useful to add a bit more context here.
```suggestion
[Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) should also
be  supported as it allows multiple client appliers to collaborate on a single object. This is particularly
relevant for resources like Routes and Policies that are likely to be implemented by multiple controllers.
For example, it's possible to attach a Route to multiple Gateways. If those Gateways were managed by
different controllers, both would need to update the Route status separately.
```

site-src/guides/implementers.md
```diff
@@ -27,6 +27,13 @@ like APIs delivered in the core Kubernetes binaries, each time a release happens
 the APIs have been reviewed by upstream Kubernetes reviewers, just like the APIs
 delivered in the core binaries.
 
+### API Semantics
+
+Gateway API MUST follow the Kubernetes API semantics. 
+The concepts are documented [here](https://kubernetes.io/docs/reference/using-api/api-concepts/).
+[Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) should also be
```
robscott, line 34: This link doesn't really provide a lot of information about what a controller author needs to do to use SSA, do you happen to have any more detailed references we could link to as well? If not, can we include some examples here?

robscott (commented):
Thanks @dprotaso!

site-src/guides/implementers.md
dprotaso: So I expect the client changes to be used by consumers of the Gateway API. Eg. two users managing different listeners on a single Gateway.


For implementors you get SSA for free from the k8s API. So I thought you wanted me to highlight what SSA is for those who aren't using the k8s API server for their control plane. A precursor to that is know the k8s concepts. 

I guess I'm not sure what you're looking for exactly. So let me know.

robscott: OK, I guess I'm just confused then. I thought the purpose of generating all this new code was so that implementations could use it. What is all the new generated code doing?

dprotaso: SSA requires separate go types (apply configurations) and codegen adds Apply methods to existing client sets 

eg. existing k8s APIs have their apply configurations here 

https://github.com/kubernetes/client-go/tree/master/applyconfigurations


The new types make all fields optional so the server can infer what's set from what's a zero value

dprotaso:
@robscott anything else for this PR?

dprotaso:
@robscott following up again

robscott:
Thanks @dprotaso!

/lgtm
/approve

dprotaso:
Let me rebase and run codegen again

dprotaso:
@robscott rebased needs another LGTM

therealmitchconnors:
Thank you!  I'm really looking forward to using this!

dprotaso:
@therealmitchconnors are you able to try it out from HEAD?

I know with some K8s resources the OpenAPI schema didn't have the correct `x-kubernetes-list-map-keys` so SSA didn't work as expected (eg. PodSpec.Containers).

I haven't done an audit of the Gateway API to ensure everything is annotated properly - I'll make an issue for that.

therealmitchconnors:
I'm on vacation for two weeks, but I might find time to kick these tires...

PR #2543, opened 2023-10-27
Fix conformance report missing tests

frankbu:
**What type of PR is this?**

/kind bug
/kind conformance

Reports generated by the conformance framework are missing information about mesh tests that are using features that are not part of the `MESH` core profile. This PR adds `HTTPRouteExtendedFeatures` to the MESH profile so that several of the mesh tests that weren't showing up in the report are now included. Generating the report (for istio) only showed 2 mesh tests, but now includes all 6 in the report.

Before this PR:
```
profiles:
- core:
    result: success
    statistics:
      Failed: 0
      Passed: 2
      Skipped: 0
    summary: ""
  name: MESH
```

After this PR:
```
profiles:
- core:
    result: success
    statistics:
      Failed: 0
      Passed: 2
      Skipped: 0
    summary: ""
  extended:
    result: partial
    skippedTests:
    - MeshFrontendHostname
    statistics:
      Failed: 0
      Passed: 3
      Skipped: 1
    summary: ""
  name: MESH
```

This PR also includes 2 additional minor changes:

1. Moved `SupportHTTPRouteBackendProtocolH2C` and `SupportHTTPRouteBackendProtocolWebSocket` constants to the correct place in the file.
2. Changed  check for external features to use `!profile.CoreFeatures.Has(supportedFeature)` because checking `profile.ExtendedFeatures.Has(supportedFeature)` doesn't work if `ExtendedFeatures` includes core features, which `GatewayExtendedFeatures` does seem to include.

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/utils/suite/experimental_reports.go
+++ b/conformance/utils/suite/experimental_reports.go
@@ -182,7 +182,7 @@ func isTestExtended(profile ConformanceProfile, test ConformanceTest) bool {
 	for _, supportedFeature := range test.Features {
 		// if ANY of the features needed for the test are extended features,
 		// then we consider the entire test extended level support.
-		if profile.ExtendedFeatures.Has(supportedFeature) {
+		if !profile.CoreFeatures.Has(supportedFeature) {
 			return true
 		}
 	}
--- a/conformance/utils/suite/features.go
+++ b/conformance/utils/suite/features.go
@@ -133,12 +133,6 @@ const (
 
 	// This option indicates support for HTTPRoute backendRequest timeouts (extended conformance).
 	SupportHTTPRouteBackendTimeout SupportedFeature = "HTTPRouteBackendTimeout"
-
-	// This option indicates support for HTTPRoute with a backendref with an appProtocol 'kubernetes.io/h2c'
-	SupportHTTPRouteBackendProtocolH2C SupportedFeature = "HTTPRouteBackendProtocolH2C"
-
-	// This option indicates support for HTTPRoute with a backendref with an appProtoocol 'kubernetes.io/ws'
-	SupportHTTPRouteBackendProtocolWebSocket SupportedFeature = "HTTPRouteBackendProtocolWebSocket"
 )
 
 // HTTPRouteExtendedFeatures includes all the supported features for HTTPRoute
@@ -166,6 +160,12 @@ var HTTPRouteExtendedFeatures = sets.New(
 const (
 	// This option indicates support for Destination Port matching.
 	SupportHTTPRouteDestinationPortMatching SupportedFeature = "HTTPRouteDestinationPortMatching"
+
+	// This option indicates support for HTTPRoute with a backendref with an appProtocol 'kubernetes.io/h2c'
+	SupportHTTPRouteBackendProtocolH2C SupportedFeature = "HTTPRouteBackendProtocolH2C"
+
+	// This option indicates support for HTTPRoute with a backendref with an appProtoocol 'kubernetes.io/ws'
+	SupportHTTPRouteBackendProtocolWebSocket SupportedFeature = "HTTPRouteBackendProtocolWebSocket"
 )
 
 // HTTPRouteExperimentalFeatures includes all the supported experimental features, currently only
--- a/conformance/tests/mesh-split.go
+++ b/conformance/tests/mesh-split.go
@@ -33,6 +33,7 @@ var MeshTrafficSplit = suite.ConformanceTest{
 	Description: "A mesh client can send traffic to a Service which is split between two versions",
 	Features: []suite.SupportedFeature{
 		suite.SupportMesh,
+		suite.SupportHTTPRoute,
 	},
 	Manifests: []string{"tests/mesh-split.yaml"},
 	Test: func(t *testing.T, s *suite.ConformanceTestSuite) {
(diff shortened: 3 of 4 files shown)


conformance/utils/suite/experimental_reports.go
```diff
@@ -182,7 +182,7 @@ func isTestExtended(profile ConformanceProfile, test ConformanceTest) bool {
 	for _, supportedFeature := range test.Features {
 		// if ANY of the features needed for the test are extended features,
 		// then we consider the entire test extended level support.
-		if profile.ExtendedFeatures.Has(supportedFeature) {
```
mlavacca, line 185: Why this change? Isn't it equivalent?

conformance/utils/suite/features.go
```diff
@@ -166,6 +160,12 @@ var HTTPRouteExtendedFeatures = sets.New(
 const (
 	// This option indicates support for Destination Port matching.
 	SupportHTTPRouteDestinationPortMatching SupportedFeature = "HTTPRouteDestinationPortMatching"
+
+	// This option indicates support for HTTPRoute with a backendref with an appProtocol 'kubernetes.io/h2c'
+	SupportHTTPRouteBackendProtocolH2C SupportedFeature = "HTTPRouteBackendProtocolH2C"
+
+	// This option indicates support for HTTPRoute with a backendref with an appProtoocol 'kubernetes.io/ws'
+	SupportHTTPRouteBackendProtocolWebSocket SupportedFeature = "HTTPRouteBackendProtocolWebSocket"
```
mlavacca, line 168: 👍

mlavacca (commented):
Thanks for this PR, @frankbu!

It generally looks good, I have a nit question though.

conformance/utils/suite/experimental_reports.go
frankbu: The checks aren't equivalent because for some reason I can't say I really understand `GatewayExtendedFeatures` also includes all `GatewayCoreFeatures`: https://github.com/kubernetes-sigs/gateway-api/blob/5fd8bec3bedc7c325bf2227da48d041a735b9b98/conformance/utils/suite/features.go#L64

frankbu:
Added work-in-progrss label because I see there are still more problems with the reporting. For example, there are only 4 TLS conformance tests but the framework is associating 10 (before this PR) and now 12 with the TLS profile, since the checking doesn't really confirm that the key feature (TLSRoute) is part of the test.

frankbu:
Turns out there is a bigger issue with how the report generator associates tests with conformance profiles. It currently decides based on the test.Features - if they are all part of a profile, it associates the test with that profile. There are 2 problems with this approach: 1) the tests are using the features to indicate what it needs to run the tests, not exactly what they're trying to test, 2) some tests are not correctly listing the features they use anyway. The result is that the same test run sometimes appear as a "passed/failed/skipped test" in more than one conformance profile, sometimes in profiles where they don't belong at all.

Fixing the above will require more significant changes in the framework including probably a way for tests to declare their conformance profile, not just the features they're using.

Since this PR fixes a different issue, i.e., adding missing tests that rightfully belong in the report plus a couple of other minor fixes, I suggest we merge this now and look at fixing the duplicate reporting issue in a future followup.

sunjayBhatia (approved):
This is definitely an improvement, though making sure the merged commit title/sub title are a bit more specific about what exactly is being fixed would be great since as you say, there are a lot of issues at play and this PR doesn't attempt to fix everything wrong with conformance test reporting

with a few tests skipped currently I see this report when running against contour `main` (looks like this branch might not be based on gateway-api `main`, so once that is rebased it would look a little different):

```yaml
profiles:
- core:
    result: partial
    skippedTests:
    - GatewayWithAttachedRoutes
    statistics:
      Failed: 0
      Passed: 28
      Skipped: 1
    summary: ""
  extended:
    result: partial
    skippedTests:
    - HTTPRouteTimeoutRequest
    - HTTPRouteTimeoutBackendRequest
    statistics:
      Failed: 0
      Passed: 10
      Skipped: 2
    summary: ""
    supportedFeatures:
    - HTTPRouteRequestMirror
    - HTTPRouteRequestTimeout
    - HTTPRouteResponseHeaderModification
    - HTTPRouteBackendTimeout
    - HTTPRouteQueryParamMatching
    - HTTPRoutePathRedirect
    - HTTPRoutePathRewrite
    - HTTPRouteRequestMultipleMirrors
    - HTTPRouteMethodMatching
    - HTTPRoutePortRedirect
    - HTTPRouteSchemeRedirect
    - HTTPRouteHostRewrite
  name: HTTP
```

frankbu:
> making sure the merged commit title/sub title are a bit more specific about what exactly is being fixed would be great

This PR started by adding `HTTPRouteExtendedFeatures` to the `MESH` profile, because several of the mesh tests weren't showing up in the report as they should. I then added `GatewayExtendedFeatures` to both the `TLS` and `HTTP` profiles, but that just seemed kind of confusing because it ended up reporting on tests that really aren't testing HTTP or TLS specific things. As I said in my previous comment, I think that identifying the profile that a test is for is a bigger issue that we need to work on in a future PR, so I removed the `GatewayExtendedFeatures` part from this PR so now it simply fixes the missed `MESH` tests plus a couple of other small things.

I'll update the title and description to say exactly what it's fixing now.

shaneutt (approved):
/lgtm

PR #2544, opened 2023-10-27
kind/documentation: concept/gamma: reword to clarify the namespace goes with clients and not the service

sanjaypujare:
**What type of PR is this?**
/kind documentation

This clarifies the wording since I was confused by the original wording about the "in the `fast-clients` Namespace" clause as to what it applies to

None

```release-note
None
```

### DIFF
--- a/site-src/concepts/gamma.md
+++ b/site-src/concepts/gamma.md
@@ -164,8 +164,8 @@ important:
     custom timeouts for that consumer's use of the workload). This Route will
     only affect requests from workloads in the same Namespace as the Route.
 
-    For example, this HTTPRoute would cause all clients of the `smiley`
-    workload in the `fast-clients` Namespace to have a 100ms timeout:
+    For example, this HTTPRoute would cause all clients in the `fast-clients`
+    Namespace of the `smiley` workload to have a 100ms timeout:
 
     ```yaml
     kind: HTTPRoute


sanjaypujare:
attn @robscott

site-src/concepts/gamma.md
```diff
@@ -164,8 +164,8 @@ important:
     custom timeouts for that consumer's use of the workload). This Route will
     only affect requests from workloads in the same Namespace as the Route.
 
-    For example, this HTTPRoute would cause all clients of the `smiley`
```
keithmattix, line 167: I think the original language is a bit clearer, but happy to defer to others

sanjaypujare: For me the qualifier ("in the `fast-clients` Namespace") should be close to the thing being qualified ("all clients"). The previous version implied the service "smiley" was in the "in the `fast-clients` Namespace".

But I admin this is subjective. I can put the "in the `fast-clients` Namespace" in parentheses so it will be:

```For example, this HTTPRoute would cause all clients (in the `fast-clients` Namespace) of the `smiley` workload  to have a 100ms timeout:```

michaelbeaumont: I do understand the potential confusion but the correction trips me up too. Can we just split the two phrases up?

> configure all clients in the `fast-clients` Namespace to have a 100ms timeout for requests to `smiley` workloads.

Issue #2557, opened 2023-11-02
Add v1.0 conformance profile of Envoy Gateway


Xunzhuo:
**What would you like to be added**:

Add v1.0 conformance profile of Envoy Gateway

/assign

shaneutt:
We don't usually open issues for adding these reports for individual implementations, as we don't track individual implementation progress in this repository. Not a big deal, but there's no reason to track this as an issue, we can just track the accompanying PR.

PR #2558, opened 2023-11-02
Add v1.0.0 Conformance Report of Envoy Gateway

Xunzhuo:
**What type of PR is this?**

/area conformance

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2557

```release-note
NONE
```

### DIFF
--- a/conformance/reports/v1.0.0/envoy-gateway.yaml
+++ b/conformance/reports/v1.0.0/envoy-gateway.yaml
@@ -1,0 +1,48 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+date: "2023-11-02T01:45:03Z"
+gatewayAPIVersion: v1.0.0
+implementation:
+  contact:
+  - https://github.com/envoyproxy/gateway/blob/main/GOVERNANCE.md
+  organization: envoyproxy
+  project: envoy-gateway
+  url: https://github.com/envoyproxy/gateway
+  version: v0.6.0
+kind: ConformanceReport
+profiles:
+- core:
+    result: success
+    statistics:
+      Failed: 0
+      Passed: 10
+      Skipped: 0
+    summary: ""
+  name: TLS
+- core:
+    result: success
+    statistics:
+      Failed: 0
+      Passed: 29
+      Skipped: 0
+    summary: ""
+  extended:
+    result: success
+    statistics:
+      Failed: 0
+      Passed: 12
+      Skipped: 0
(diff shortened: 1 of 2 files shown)


arkodg (approved):
/approve
thanks !

tao12345666333 (approved):

robscott:
I'm probably missing something obvious, but how is it possible to pass all tests with v1.0.0 per https://github.com/kubernetes-sigs/gateway-api/pull/2535/files#r1378650459?

arkodg:
@robscott the current code in v1.0.0 made an extra assertion that the Route is not `Accepted` . This is true for Envoy Gateway.
However there is nothing in the spec to enforce that, so that check was removed. ptal at https://github.com/kubernetes-sigs/gateway-api/pull/2548 for more context
tldr - Envoy Gateway will pass all tests with or without that assertion

robscott:
Thanks @arkodg! I guess the question was how that test could pass (Accepted == false) at the same time as a nearly identical test that requires Accepted to be true: https://github.com/kubernetes-sigs/gateway-api/blob/341c46e29f16501f301abf04313b8d2cf6e48e0f/conformance/tests/httproute-invalid-backendref-nonexistent.go#L48

Just trying to understand the difference here when this seems to be a blocker for many implementations to pass conformance on v1.0.0. Maybe Envoy Gateway is making a distinction between the two test cases that other implementations are not?

arkodg:
Envoy Gateway defines its own custom reason (for the Route) for this case/test- `NoReadyListeners`

      - lastTransitionTime: "2023-10-27T01:38:14Z"
        message: There are no ready listeners for this parent ref
        observedGeneration: 1
        reason: NoReadyListeners
        status: "False"
        type: Accepted
      - lastTransitionTime: "2023-10-27T01:38:14Z"

The listener is not ready in this test because the cert is invalid

      - lastTransitionTime: "2023-10-27T01:38:14Z"
        message: Secret gateway-conformance-infra/does-not-exist does not exist.
        observedGeneration: 2
        reason: InvalidCertificateRef
        status: "False"
        type: ResolvedRefs
      - lastTransitionTime: "2023-10-27T01:38:14Z"
        message: Listener is invalid, see other Conditions for details.
        observedGeneration: 2
        reason: Invalid
        status: "False"
        type: Programmed

robscott:
Got it, makes sense, thanks for the explanation @arkodg!

/hold cancel
/approve

PR #2565, opened 2023-11-04
GEP-1291: `Mesh` representation

### LINKED ISSUE #1291: GEP: Mesh Representation
As the GAMMA initiative seeks to create a model for service mesh implementations within Gateway API, it is necessary to have a mechanism for representing a [mesh](https://docs.google.com/document/d/1T_DtMQoq2tccLAtJTpo3c0ohjm25vRS35MsestSL9QU/edit#heading=h.fv9zxi8pjsiq) (follow link for definition) using the Gateway API.

This GEP covers the work to design and implement this mechanism.

As @shaneutt originally suggested, the first step will be implementing a provisional GEP that sets out the terms of what we're doing - the "What" and "Why" of the solution, and then once we're on the same page, we'll talk more about the "How".

**Why this is needed**:
Without a way to represent a mesh using Gateway API, no other work can continue.

michaelbeaumont:
**What type of PR is this?**
/kind gep

GAMMA implementations need a way to report conformance and CRD compatibility.
This GEP proposes a `Mesh` resource.

Fixes #1291

```release-note
Add `Mesh` resource for reporting by GAMMA implementations
```

### DIFF
--- a/geps/gep-1291.md
+++ b/geps/gep-1291.md
@@ -1,0 +1,132 @@
+# GEP-1291: `Mesh` resource
+
+- Issue: [#1291](https://github.com/kubernetes-sigs/gateway-api/issues/1291)
+- Status: Provisional
+
+## TLDR
+
+GAMMA implementations need a way to report conformance and CRD compatibility.
+This GEP proposes a `Mesh` resource.
+
+## Goals
+
+The initial `Mesh` resource should provide:
+
+1. Visibility into supported features
+1. CRD version compatibility
+
+It is intended to exist in a 1-1 correspondence to an implementation installation/control
+plane.
+
+This GEP also serves as a place to build consensus on the need for a `Mesh`
+resource at all. Note for clarity, this `Mesh` is essentially the `MeshClass`
+mentioned in [the doc from the linked issue](https://github.com/kubernetes-sigs/gateway-api/issues/1291#issuecomment-1209625419).
+
+## Current non-goals
+
+- The possibility of standardizing the `Mesh` as a `targetRef` resource for policy that should apply to all workloads
+  in a `Mesh` is put aside for a later iteration.
+- This GEP does not address a `Mesh` resource for use by Kubernetes operators
+  (the controller pattern, not the role).
+- Holding configuration for mesh installations (so no `parametersRef`)
+
+## Proposal
+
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
First, thanks for the GEP. It looks good and seems like a reasonable path for this if we decide to go with it.

In my subjective opinion this doesn't help the project meaningfully. We are adding complexity that doesn't hold it's own weight in user benefit. FWIW I think the same about GatewayClass, though, so that may be an unpopular opinion :-)

Basically we are making a new resource (high cost!) with the idea that maybe we want to do things with it in the future (this is not convincing at all; we shouldnt weight hypotheticals highly) and to report some information.

This can, and imo should, simply be a page on a projects' documentation. This is orders of magnitude simpler while still providing as much (if not more!) value to users

shaneutt:
> First, thanks for the GEP. It looks good and seems like a reasonable path for this if we decide to go with it.
> 
> In my subjective opinion this doesn't help the project meaningfully. We are adding complexity that doesn't hold it's own weight in user benefit. FWIW I think the same about GatewayClass, though, so that may be an unpopular opinion :-)
>
> Basically we are making a new resource (high cost!) with the idea that maybe we want to do things with it in the future (this is not convincing at all; we shouldnt weight hypotheticals highly) and to report some information.
> 
> This can, and imo should, simply be a page on a projects' documentation. This is orders of magnitude simpler while still providing as much (if not more!) value to users

We appreciate your thoughts as always @howardjohn. These comments do seem to unfortunately provide slight ambiguity as to the prescription or specific actions to take on your behalf, however. The best read I can come up with of what _action_ you may be prescribing is "don't proceed until we have stronger use cases / user stories". If that's the case I think asking for more and stronger emphasis on use cases and user stories seems legit, though I don't personally feel that it needs to hold us up from reaching a `Provisional` state.

In any case it would seem prudent for many of us who focus on GAMMA to give this some deep through and review it:

/cc @kflynn @keithmattix @mikemorris @whitneygriffith

Please review and share your thoughts and ideas about the `Mesh` resource concept.

mikemorris:
I like the idea here in the abstract (more than I like GatewayClass) and don't have any specific objections to the content of the GEP, but can agree we should hold off on introducing/requiring a `Mesh` resource until we have a specific functional need this will unblock.

PR #2582, opened 2023-11-13
conformance: missmatch `SectionName` and `Port` will trigger `NoMatchingParent` 

tao12345666333:
**What type of PR is this?**

/kind test
/area conformance
**What this PR does / why we need it**:

According to the description, when both Port and SectionName are specified, the name and port of the selected listener
	must match both specified values.

https://github.com/kubernetes-sigs/gateway-api/blob/e2f02aa0fc9dfa27be271397801f52f4e0a78acc/apis/v1/shared_types.go#L124C1-L130C38

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2589, opened 2023-11-17
chore(conformance): shorten the resource name

tao12345666333:
prevent it  from destroying other projects

**What type of PR is this?**

/area conformance
/kind test

> fails with contour since the gateway name has a really long name and that causes a label we add to some resources to be > 63 chars
> 
> selfishly it would be great to have a shorter gateway name here but otherwise the logic of the test looks good
>
> test passes with contour when i shorten the gateway name
> _Originally posted by @sunjayBhatia  https://github.com/kubernetes-sigs/gateway-api/pull/2582#pullrequestreview-1735319447

In Kubernetes, 63 is a very important length.
Although for Gateway resources, its name can exceed 63, different implementations may use it in different positions.

I think conformance tests should be as universal as possible to avoid encountering problems like this.
I also checked other test cases and currently, all lengths are below 50.

Fixes #

```release-note
None
```

### DIFF
--- a/conformance/tests/httproute-invalid-parentref-section-name-not-matching-port.go
+++ b/conformance/tests/httproute-invalid-parentref-section-name-not-matching-port.go
@@ -41,7 +41,7 @@ var HTTPRouteInvalidParentRefSectionNameNotMatchingPort = suite.ConformanceTest{
 	Manifests: []string{"tests/httproute-invalid-parentref-section-name-not-matching-port.yaml"},
 	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
 		routeNN := types.NamespacedName{Name: "httproute-listener-section-name-not-matching-port", Namespace: "gateway-conformance-infra"}
-		gwNN := types.NamespacedName{Name: "gateway-with-one-not-matching-port-and-section-name-route", Namespace: "gateway-conformance-infra"}
+		gwNN := types.NamespacedName{Name: "gw-with-one-section-name-not-matching-port-route", Namespace: "gateway-conformance-infra"}
 
 		// The Route must have an Accepted Condition with a NoMatchingParent Reason.
 		t.Run("HTTPRoute with sectionName does not match Port in ParentRef has an Accepted Condition with status False and Reason NoMatchingParent", func(t *testing.T) {
--- a/conformance/tests/httproute-invalid-parentref-section-name-not-matching-port.yaml
+++ b/conformance/tests/httproute-invalid-parentref-section-name-not-matching-port.yaml
@@ -1,7 +1,7 @@
 apiVersion: gateway.networking.k8s.io/v1
 kind: Gateway
 metadata:
-  name: gateway-with-one-not-matching-port-and-section-name-route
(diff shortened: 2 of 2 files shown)


mlavacca (commented):
I'm not sure about this change, to be honest. I think that implementations should deal with this kind of problem and they should not expect only gateway names which length is <= 50 chars. I get the point of not destroying other implementations, but since we just released v1.0 and there likely is some time before the next release, I am not sure about mitigating this issue, since it's an implementation issue rather than an API one. @sunjayBhatia How hard is it for you to fix this problem in Contour? Do you think you'll be able to fix it in a short time or do you foresee potential compatibility issues?

I want to underline I'm 100% down on "I do not want to break implementations", and if this can happen I'm okay with this change.

sunjayBhatia:
> I'm not sure about this change, to be honest. I think that implementations should deal with this kind of problem and they should not expect only gateway names which length is <= 50 chars. I get the point of not destroying other implementations, but since we just released v1.0 and there likely is some time before the next release, I am not sure about mitigating this issue, since it's an implementation issue rather than an API one. @sunjayBhatia How hard is it for you to fix this problem in Contour? Do you think you'll be able to fix it in a short time or do you foresee potential compatibility issues?
> 
> I want to underline I'm 100% down on "I do not want to break implementations", and if this can happen I'm okay with this change.

We have an issue to fix this in contour already: https://github.com/projectcontour/contour/issues/5970

I mainly just called it out for completeness since I had to modify the tests while reviewing to get them to pass properly with contour

mlavacca:
> > I'm not sure about this change, to be honest. I think that implementations should deal with this kind of problem and they should not expect only gateway names which length is <= 50 chars. I get the point of not destroying other implementations, but since we just released v1.0 and there likely is some time before the next release, I am not sure about mitigating this issue, since it's an implementation issue rather than an API one. @sunjayBhatia How hard is it for you to fix this problem in Contour? Do you think you'll be able to fix it in a short time or do you foresee potential compatibility issues?
> > I want to underline I'm 100% down on "I do not want to break implementations", and if this can happen I'm okay with this change.
> 
> We have an issue to fix this in contour already: [projectcontour/contour#5970](https://github.com/projectcontour/contour/issues/5970)
> 
> I mainly just called it out for completeness since I had to modify the tests while reviewing to get them to pass properly with contour

This is great! I guess this change is not needed anymore?

sunjayBhatia:
> This is great! I guess this change is not needed anymore?

Yeah I would say so 👍🏽 we can come back to this if for whatever reason we can't solve the issue in contour, but that is unlikely

mlavacca:
> > This is great! I guess this change is not needed anymore?
> 
> Yeah I would say so 👍🏽 we can come back to this if for whatever reason we can't solve the issue in contour, but that is unlikely

👍 Thanks anyway @tao12345666333 for promptly jumping on it.

/close

tao12345666333:
Make sense. Thanks for reply.

sunjayBhatia:
@tao12345666333 @mlavacca added https://github.com/projectcontour/contour/pull/5969 since this actually has wider implications beyond just this test

PR #2591, opened 2023-11-17
docs: `RouteReasonNoMatchingParent` docs improved

mlavacca:
**What type of PR is this?**

/area docs

The `RouteReasonNoMatchingParent` documentation has been improved to include all the possible cases for which a route can be rejected with such a reason.

Fixes #

```release-note
NONE
```

### DIFF
--- a/apis/v1/shared_types.go
+++ b/apis/v1/shared_types.go
@@ -334,9 +334,10 @@ const (
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
 	// This reason is used with the "Accepted" condition when there are
-	// no matching Parents. In the case of Gateways, this can occur when
-	// a Route ParentRef specifies a Port and/or SectionName that does not
-	// match any Listeners in the Gateway.
+	// no matching Parents. In the case of Gateways, this can occur when:
+	// * no referenced parent exists
+	// * a route ParentRef specifies a Port and/or SectionName that does not
+	//   match any Listeners in the Gateway.
 	RouteReasonNoMatchingParent RouteConditionReason = "NoMatchingParent"
 
 	// This reason is used with the "Accepted" condition when a value for an Enum


apis/v1/shared_types.go
```diff
@@ -334,9 +334,11 @@ const (
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
 	// This reason is used with the "Accepted" condition when there are
-	// no matching Parents. In the case of Gateways, this can occur when
-	// a Route ParentRef specifies a Port and/or SectionName that does not
-	// match any Listeners in the Gateway.
+	// no matching Parents. In the case of Gateways, this can occur when:
+	// * the route ParentRefs field is empty
+	// * no referenced parent exists
```
pmalek, line 339:
```suggestion
	// * the referenced parent does not exist
```

mlavacca: `parentRefs` is a slice of parents, so "no referenced parent exists" means that "none of the referenced parents exists", that is different than "the referenced parent does not exist", where we assume there is only one parent.

pmalek: Ah, yes. I forgot about this being a slice. Thanks!

apis/v1/shared_types.go
```diff
@@ -334,9 +334,11 @@ const (
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
 	// This reason is used with the "Accepted" condition when there are
-	// no matching Parents. In the case of Gateways, this can occur when
-	// a Route ParentRef specifies a Port and/or SectionName that does not
-	// match any Listeners in the Gateway.
+	// no matching Parents. In the case of Gateways, this can occur when:
+	// * the route ParentRefs field is empty
```
pleshakov, line 338: if there are no parent refs, can an implementation set any status at all? since status is per parentRef -- https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.RouteParentStatus ?

I wonder if this comment necessary

mlavacca: you are right 😮‍💨 removed that point.

apis/v1/shared_types.go
```diff
@@ -334,9 +334,10 @@ const (
 	RouteReasonNoMatchingListenerHostname RouteConditionReason = "NoMatchingListenerHostname"
 
 	// This reason is used with the "Accepted" condition when there are
-	// no matching Parents. In the case of Gateways, this can occur when
-	// a Route ParentRef specifies a Port and/or SectionName that does not
-	// match any Listeners in the Gateway.
+	// no matching Parents. In the case of Gateways, this can occur when:
+	// * no referenced parent exists
```
robscott, line 338: If "no referenced parent exists", who would populate this status? It's possible for a controller to populate the status if a Gateway they're implementing is referenced, but just a nonexistent port or section name, but it seems impossible to know who should populate this if a nonexistent Gateway/parent resource is referenced here.

mlavacca: At Kong (unmanaged gateway), we have a single ingress controller that watches all (or a subset of) namespaces and reconciles all `Gateway`s and all `*Route`s in those namespaces. In the contexts where there is no 1:1 (or n:1) relationship between gateway and gateway controller, but instead, a gateway controller reconciles everything, if no referenced gateway exists the unique gateway controller should be the one in charge of setting this status condition.

What is not immediately clear to me is: how this pattern is common among our implementations and if the Gateway API expects such a pattern.

youngnick: As written, Gateway API does not expect a pattern like that at all. It's expected that the only Gateway API objects touched by an implementation are those that are associated with a GatewayClass managed by the implementation - so there needs to be a tree of references that has a GatewayClass as a root node. Another thing we've never explicitly said in that way, but that's definitely what we intended.

This is the other reason why we're reluctant to allow people to skip having a GatewayClass object as well, but it seems clear that the only-have-the-implementation-watch-a-subset-of-namespaces use case needs to be addressed in some fashion - with recommendations if not specifications. I'm making a note about this one - I think that checking in with api-machinery about it may be essential.

mlavacca: > there needs to be a tree of references that has a GatewayClass as a root node.

I see, this is a good way of putting it. What do you think about adding a conformance test for this? We could create a fake `GatewayClass` and a chain of objects having that GWC as root (`Gateway`, `HTTPRoute`) and check that all the created objects are ignored by the controller and no status gets updated.

mlavacca:
Closing as not needed

/close

apis/v1/shared_types.go
youngnick: Hmm, that seems like a good idea to me.

Issue #2319, opened 2023-08-17
Document the upcoming deprecation and eventual removal of the Validating Webhook


robscott:
With the transition to CEL becoming a reality in v0.8.0, we are going to begin the process of spinning down support for the validating webhook. I expect the timeline will go as follows:

**v0.8.0:**

* CEL validation is included in all CRDs with only one missing feature from webhook validation: https://github.com/kubernetes-sigs/gateway-api/issues/2277
* Installation of the webhook is still recommended, but release notes state that this will be the last release it is recommended for
* Implementations are strongly encouraged to run validation code from webhook on the controller-side, primarily to cover the gap described in #2277 

**v1.0:**

* Validating webhook is formally deprecated, installation is only recommended when using Gateway API on clusters <= 1.24
* Implementations are _required_ to cover #2277 with controller-side validation (likely reusing the validation code from the webhook)

**v1.1:**

* The webhook is removed from the codebase
* Validation code remains as a library for controllers to run, covering any gaps in validation that can't be covered with CEL (currently limited to #2277)

PR #2595, opened 2023-11-17
Remove validating webhook

robscott:
**What type of PR is this?**
/kind cleanup

As described in https://github.com/kubernetes-sigs/gateway-api/issues/2319, it's time for the webhook to go away. The next minor release of Gateway API will _not_ include the webhook, so there's no need to keep it around. This should also dramatically speed up presubmits.

```release-note
The validating webhook has been removed. CEL validation is now built-in to CRDs and replaces the webhook.
```


robscott:
This is a pretty big PR, will add a hold so we can get at least a couple reviewers to sign off on this.

/hold

Xunzhuo (commented):
Generally LGTM, thanks for working on it Rob, since it is a big PR, defer to other forks to have a look, in case we do not miss anything.

shaneutt (changes requested):
We still have a section in `guidelines.md` entitled `Limitations of CRD and Webhook Validation` which we may want to reword some, as it seems the relevance will be somewhat changed and broken after this PR.

Also found a couple other documentation areas, you can probably just apply this diff to resolve them:

```diff
diff --git a/site-src/concepts/versioning.md b/site-src/concepts/versioning.md
index 4a254122..d9e11f48 100644
--- a/site-src/concepts/versioning.md
+++ b/site-src/concepts/versioning.md
@@ -2,12 +2,11 @@
 
 ## Overview
 Each new release of Gateway API is defined with a "bundle version" that
-represents the Git tag of a release, such as v0.8.0. This contains the
+represents the Git tag of a release, such as v1.0.0. This contains the
 following:
 
 * API Types (Go bindings for the resources)
 * CRDs (Kubernetes definitions of the resources)
-* Validating Webhook (Will be deprecated in v1.0 release)
 
 ### Release Channels
 Release channels are used to indicate feature stability within Gateway API. All
diff --git a/site-src/contributing/contributor-ladder.md b/site-src/contributing/contributor-ladder.md
index d498de5b..760fb7f7 100644
--- a/site-src/contributing/contributor-ladder.md
+++ b/site-src/contributing/contributor-ladder.md
@@ -85,7 +85,6 @@ reviewer or approver, this includes:
 * Conformance
 * Documentation
 * GEPs
-* Webhook Validation
 
 ## Maintainers and GAMMA Leads
 
```

If not part of THIS pr, then definitely of one in the future as we cut the release that removes the webhook.

robscott:
Thanks @shaneutt! I applied your suggestions + added the following note on the `guidelines.md` file:

```
    Webhook validation in Gateway API has been deprecated and will be fully
    removed in v1.1.0. With that said, all of this guidance will still apply for
    implementations as long as they support v1.0.0 or older releases of the API.
```

apis/v1/validation/common.go
youngnick, line 1: Hmm, for all the files in `/validation`, are breaking an API by removing this validation code? Is this code used directly by implementations, rather than just in the webhook? Should we do a separate deprecation for it?

hack/verify-crds-kind.sh
```diff
@@ -119,89 +114,5 @@ EOF
   kubectl delete -f "config/crd/${CHANNEL}/gateway*.yaml" || res=$?
 done
 
-###
-# This section and below can be REMOVED once the webhook is removed.
-###
```
youngnick, line 124: 🎉

youngnick:
LGTM with one caveat about if we can remove the `validation` packages yet.

shaneutt (approved):

apis/v1/validation/common.go
robscott: Good question. In my opinion, this is covered under "source code" within our versioning policy which has no compatibility guarantees: https://gateway-api.sigs.k8s.io/concepts/versioning/#source-code. I started to split this out to leave `validation` code in and just deprecate the packages, but I think that could lead to a worse outcome - stale code. Although we have some unit tests for the code, by removing the webhook, we'd be removing one of the primary ways we tested this code. I think the removal of this is adequately covered by the overall webhook deprecation announcements we made in the v0.8 and v1.0 release notes.

robscott:
I think this has sat long enough, been discussed at multiple community meetings, and now has multiple approvals, removing the hold.

/hold cancel

jwendell:
Should the validation go package be removed only from v1 and not from v1alpha1 and v1beta1?

Asking because Istio imports sigs.k8s.io/gateway-api/apis/v1alpha2/validation to perform some validation, and now they are gone. How can we support both 1.0 and older?

howardjohn:
> Should the validation go package be removed only from v1 and not from v1alpha1 and v1beta1?
> 
> Asking because Istio imports sigs.k8s.io/gateway-api/apis/v1alpha2/validation to perform some validation, and now they are gone. How can we support both 1.0 and older?

Istio doesn't support v1alpha2 anymore. I think the current logic is broken in Istio, and can probably be removed entirely given the removal of the webhook?

robscott:
Since CEL has been present in CRDs since v0.8 and this validation code will only be removed as part of the v1.1 release (when this PR will get released), I think it's likely safe to remove any code that ran these validations controller-side when you upgrade to importing v1.1. Note that the v1.1 release is still likely at least a couple of months away.

PR #2633, opened 2023-12-01
Sort items in conformance reports to maintain stability

programmer04:
**What type of PR is this?**

/kind feature
/area conformance

Using `UnsortedList()` on a set and iterating over a map gives a random order of items, thus reports are not stable (the order of items in those reports changes), for example, an update of one report that only differs in the app version.
- https://github.com/kubernetes-sigs/gateway-api/pull/2611

```release-note
The order of items in a conformance report is stable.
```


shaneutt (approved):
/lgtm

Issue #1778, opened 2023-03-06
Add support for Load Balancing Policy


arkodg:
**What would you like to be added**:
The ability for the user to specify the load balancing policy/algorithms used when load balancing requests from the gateway to the upstream / backend.

**Why this is needed**:
To improve performance/throughput in production.

Examples in Data Plane
* Envoy - https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers
* NGINX - https://nginx.org/en/docs/http/load_balancing.html

Examples in Control Plane
* Contour - https://projectcontour.io/docs/v1.10.0/config/request-routing/#load-balancing-strategy
* Istio - https://istio.io/latest/docs/reference/config/networking/destination-rule/#LoadBalancerSettings

shaneutt:
For posterity, this has been brought up a few times. Here's some of the prior art:

#1074 #992 #912 

Historically we have landed on policy attachment being the suggested method to handle this, however I'm personally OK with seeing where this goes. @youngnick and @robscott should weigh in.

arkodg:
thanks @shaneutt for surfacing all the prior issues ! would ideally like to see this as a native noun rather than a policy  attachment

shaneutt:
I'm with you that this would be good. To make it happen we're going to need someone who's dedicated to championing it, teasing out all the issues and pushing it forward as a GEP. I think getting this right across multiple implementations is going to be a little tricky.

arkodg:
thanks, I see how this feature can be related to the Backend Properties proposal outlined in https://github.com/kubernetes-sigs/gateway-api/pull/1333/files .

From an API perspective, the `HTTPBackendRef` https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPBackendRef could be a good place to introduce the notion of a backend filter/property to support a load balancing policy.

shaneutt:
We don't consider this a blocker for `v1.0.0`/GA because we have every expectation that it can be done post-GA. We do want this feature, and if any community member is interested in taking it on please do.

/help

shaneutt:
We want this feature, but we consider it low priority until `v1.0.0`/GA is complete and don't believe we will have bandwidth for it until then.

PR #2634, opened 2023-12-01
GEP-1619: Refactor SessionPersistencePolicy into BackendLBPolicy 

gcs278:
**What type of PR is this?**
/kind gep

N/A

The main goal of this PR is to get [GEP-1619](https://github.com/kubernetes-sigs/gateway-api/issues/1619) ready for experimental by answering the open questions in [Graduation Criteria for Implementable Status](https://gateway-api.sigs.k8s.io/geps/gep-1619/#graduation-criteria-for-implementable-status):
1. Should we leave room in this policy to add additional concepts in the future such as Session Affinity? If so, how would we adjust the naming and overall scope of this policy?
2. Should we leave room for configuring different forms of Session Persistence? If so, what would that look like?
 It also has variety of other clean up and clarification work.

The PR currently does the following (broken in to commits):
1. Fix links, formatting, metadata.yaml
2. Add new implementations with their session persistence APIs
3. Answer question on forms of Session Persistence
4. Distinguish between session vs. persistent cookies (see https://github.com/kubernetes-sigs/gateway-api/issues/2747)
5. Update timeouts to use Duration field
6. Adopt RFC8174 language
7. Answer open question about unhealthy backends
8. Refactor the API from `SessionPersistencePolicy` into `BackendLBPolicy` and attach only to services, and introduce a new inline route API

https://github.com/kubernetes-sigs/gateway-api/issues/2747
https://github.com/kubernetes-sigs/gateway-api/issues/1778 (Doesn't fix, but is related because `LoadBalancerPolicy` is the framework for adding algorithm selection)

```release-note
NONE
```


youngnick:

geps/gep-1619/index.md
```diff
@@ -317,71 +346,55 @@ In this section, we will explore the questions and design elements associated wi
...
 // Note: there is no Override or Default policy configuration.
-type SessionPersistencePolicySpec struct {
+type LoadBalancerPolicySpec struct {
     // TargetRef identifies an API object to apply policy to.
-    // The TargetRef may be a Service, HTTPRoute, GRPCRoute,
-    // or a HTTPRouteRule or GRPCRouteRule section.
-    // At least one of these targets must be supported for
-    // core-level compliance.
-    //
+    // Currently, Services are the only valid API target references.
```
costinm, line 369: Should it say 'backends' (i.e. anything that is allowed to be a backendRef - like clusterset or other vendor specific backends ) ?

geps/gep-1619/index.md
```diff
@@ -317,71 +346,55 @@ In this section, we will explore the questions and design elements associated wi
...
-    IdleTimeoutSeconds int64 `json:"idleTimeoutSeconds,omitempty"`
+    // SessionPersistence defines and configures session persistence.
+    SessionPersistence *SessionPersistence `json:"sessionPersistence"`
+}
 
+// SessionPersistence defines the desired state of
+// SessionPersistence.
+type SessionPersistence struct {
     // SessionName defines the name of the persistent session token
     // (e.g. a cookie name).
```
costinm, line 380: Can we add 'or a header name' - to make it more clear that it's not restricted to cookies ?

costinm:
Looks much better !

geps/gep-1619/metadata.yaml
```diff
@@ -7,3 +7,14 @@ authors:
...
+  seeAlso:
+    - number: 713
+      name: Metaresources and Policy Attachment
+      description: Defines metaresource and policy attachment API
+changelog:
+  - "https://github.com/kubernetes-sigs/gateway-api/pull/1643"
+  - "https://github.com/kubernetes-sigs/gateway-api/pull/1889"
+  - "https://github.com/kubernetes-sigs/gateway-api/pull/1935"
+  - "https://github.com/kubernetes-sigs/gateway-api/pull/2159"
+  - "https://github.com/kubernetes-sigs/gateway-api/pull/2634"
```
robscott, line 20: This is really helpful, thanks for filling it out!

geps/gep-1619/index.md
```diff
@@ -317,71 +346,55 @@ In this section, we will explore the questions and design elements associated wi
...
     // +optional
     // +kubebuilder:validation:MaxLength=4096
     SessionName String `json:"sessionName,omitempty"`
 }
 
-// SessionPersistencePolicyStatus defines the observed state of SessionPersistencePolicy.
-type SessionPersistencePolicyStatus struct {
-    // Conditions describe the current conditions of the SessionPersistencePolicy.
+// LoadBalancerPolicyStatus defines the observed state of LoadBalancerPolicy.
+type LoadBalancerPolicyStatus struct {
```
robscott, line 390: I think this needs to be expanded to allow for multiple controllers to write to this. I'd recommend reusing what @candita and @youngnick did for BackendTLSPolicy: https://github.com/kubernetes-sigs/gateway-api/blob/1f7b47f3acb137def7d2ebafc1e3a5b6c710252c/apis/v1alpha2/policy_types.go#L187

geps/gep-1619/index.md
```diff
@@ -3,22 +3,21 @@
 * Issue: [#1619](https://github.com/kubernetes-sigs/gateway-api/issues/1619)
```
robscott, line 3: Can't comment directly on the title, but given the rename of the API, we _may_ also want to rename the GEP to match.

geps/gep-1619/index.md
```diff
@@ -317,71 +346,55 @@ In this section, we will explore the questions and design elements associated wi
 ### GO
 
 ```go
-// SessionPersistencePolicy provides a way to define session persistence rules
-// for a service or route.
+// LoadBalancerPolicy provides a way to define load balancing rules
```
robscott, line 349: Naming 🧵 for anyone interested.

I agree with the idea to use a more generic/broad term here, but I'm not 100% sure what we want that name to be. To help frame this, here are some ideas that may fit into a policy attached to a Service:

1. Session Affinity
2. Timeout
3. Connection Draining
4. Logging
5. LB Algorithm

Of course this does get more complex if we want to also attach this to Route in the future, so we _probably_ don't want "Backend" in the name of this policy.

Some ideas beyond LoadBalancerPolicy:

* LBPolicy
* LoadBalancingPolicy
* TrafficPolicy
* Other ideas welcome

For the record, I don't hate LoadBalancerPolicy, just thought I'd throw out some other names in case any stuck.

robscott (commented):
Thanks @gcs278, this is a really well written GEP! I still haven't gotten through everything, but adding some initial comments since my review is already so late.

geps/gep-1619/index.md
```diff
@@ -705,21 +665,23 @@ spec:
...
+    name: split-route-cookie
 ```
 
 A potentially unexpected situation occurs when:
+
 1. Curl to `/a` which establishes a persistent session with `servicev1`
 2. Curl to `/b` routes to `servicev1` due to route persistence despite `weight: 0` configuration
 
-In this scenario, implementations should give precedence to session persistence, regardless of the `weight`
+In this scenario, implementations MUST give precedence to session persistence, regardless of the `weight`
```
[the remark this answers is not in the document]
costinm: In this case both cookies may work semantically - if the first request goes to A, the cookieA is set and future requests use it. Same for B. However not sure if envoy allows this - so validation may reject it.

IMO we don't have very strong use cases for allowing users to customize the cookie names - and some managed implementations don't support it, so at most it should be optional or 'hint'.

geps/gep-1619/index.md
costinm: I don't mind LBPolicy and LoadBalancerPolicy. Agree it's better to not have 'backend' in the name - but I don't mind it too much, setting load balancing policies per route is not common and  unlikely to support all options and semantics. So if/when that is desired - a RouteBalancerPolicy may be added with a small subset.

geps/gep-1619/index.md
```diff
@@ -265,20 +284,25 @@ Generally, the implementation API programs the dataplane API; however these two
 | **Technology** 	| **Technology Type** 	| **Session Persistence Type** 	| **Configuration Options** 	| **Configuration Association (Global, Gateway, Route, or Backends)** 	| **Notes** 	|
```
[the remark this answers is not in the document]
gcs278: Yea I agree, it will. Done

geps/gep-1619/index.md
gcs278: Yea, at the current moment, I don't see why we should prevent implementations from using a custom backend or `ServiceImport` for multi-cluster services. I'll include it.

geps/gep-1619/index.md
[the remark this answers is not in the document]
gcs278: So you are saying if the client presents two cookies simultaneously in a traffic splitting scenario? I think it should just be the cookie that is listed first in the client's HTTP request. That seems the simplest. Is that what you meant by "semantically" @costinm?

If I understand this correctly, it's a bit unlikely of a scenario. If you provide a cookie, the `set-cookie` response header _shouldn't_ be set, and if it was set for some reason, it _shouldn't_ be set with a name & value to for another backend. Am I understanding correctly?

And yes, `sessionName` is implementation-specific (because it's not testable) support-level and optional.

gcs278 (commented):
Thanks for the great comments. Latest push has fixes: https://github.com/kubernetes-sigs/gateway-api/compare/1817284c82da6c5e7618c3513654db8a8e8ad08e..4c903ae318df3625232c4df668a380ef7ed56b06

- Added back graduation criteria, but answered it. Also added criteria for Implementable about naming
- Added "Table Last Updated"
- Targets a backendRef instead of only a Service
- Fixed API status
- Added back in `HTTPCookie API Alternative` alternative with better context

ginayeh (commented):
Thanks for taking time on this.  This is super helpful.

shaneutt (approved):

site-src/api-types/httproute.md
```diff
@@ -43,6 +43,64 @@ spec:
 Note that the target Gateway needs to allow HTTPRoutes from the route's
 namespace to be attached for the attachment to be successful.
 
+You can also attach routes to specific sections of the parent resource.
+For example, let's say that the `acme-lb` Gateway includes the following
+listeners:
```
robscott, line 48: This looks like it was also accidentally included from another PR, some rebasing/git history cleanup may be needed here.

site-src/guides/tcp.md
```diff
@@ -49,8 +49,24 @@ In this way each `TCPRoute` "attaches" itself to a different port on the
 `Gateway` so that the service `my-foo-service` is taking traffic for port `8080`
 from outside the cluster and `my-bar-service` takes the port `8090` traffic.
 
+Note that you can achieve this same result by binding the Routes to the Gateway
+listeners using the `port` field in the `parentRefs`:
```
robscott, line 53: Same as above, changes in this file also look accidental.

geps/gep-1619/index.md
```diff
@@ -459,52 +438,46 @@ route in any given implementation.
...
-### API Attachment Points
+### API Attachment Point
 
-The `SessionPersistencePolicy` metaresource can target a service, route, and a route rule section (e.g. [HTTPRouteRule](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule)).
-The core functionality of attaching it to any of these is generally identical. The distinction lies in where the
-configuration gets propagated: when attached to a route or a route rule, it can define session persistence for multiple
-services, whereas attaching it to a service defines it for a single service. Enabling session persistence doesn't
-require configuration on the route, route rule, and service; configuring it on any one is sufficient.
+Attaching the `LoadBalancerPolicy` metaresource to a backend will be a core support level feature, while attaching
+it to a route or a route rule section may be addressed in a future update of this GEP. A backend can be a Service,
```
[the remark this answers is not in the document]
robscott: I'd echo @ginayeh's thoughts here. Attaching session persistence config via Route rule helps avoid many of the edge cases we deal with when it comes to backend attachment. For example, when traffic splitting it would be much easier/cleaner to have config that applied for the entire route rule instead of separately for each backend.

geps/gep-1619/index.md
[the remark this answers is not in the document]
robscott: Yeah, I think you're right. I'm just trying to find edge cases when persistence config is set differently for traffic splitting backends. It seems like there are lots of possible rough edges when combining backend-attached session persistence with traffic splitting.

geps/gep-1619/index.md
```diff
@@ -736,52 +669,139 @@ spec:
...
 
 ## Alternatives
 
-### Alternate Session Persistence API
+### Timeout Config Alternative
+
+Alternatively, timeouts such as `AbsoluteTimeoutSeconds` and `IdleTimeoutSeconds` could be added to the API. This would
+allow users to control the duration of their sessions. However, as described in [Expires / Max-Age](#expires-max-age),
+we currently prescribe that if cookies are used, then all cookies MUST be session cookies. If timeout configuration is
+added, then we also need to add an API to distinguish between session cookies and permanent cookies. This distinction is
```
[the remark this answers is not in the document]
robscott: Although historically I've been very much in favor of starting with an MVP for APIs, we're now starting to run into the pain of each relatively small addition to an API requiring yet another GEP. So if it's likely that we'll want to add something in the future, I think I'd rather just add it to start with and skip all the extra future work this would be to add as a separate GEP.

robscott (commented):
Thanks for all the work on this @gcs278! This is clearly a very complex topic and you've explained all the details along with the pros and cons of each possible approach very well.

geps/gep-1619/index.md
gcs278: This is an interesting pattern I hadn't considered yet. Policy for Service, and Inline for Route. I had some confusion on exactly what route attachment meant as far as inherited policy attachment design, but this approach seems simpler to me, i.e. it seems problematic to start with Direct Policy service attachment, then later having to redesign into Inherited Policy attachment to also support Routes.

The configuration, though common, would be in two different places (Policy & Route), so that may be considered a negative, but the positive is that is it gives us more independence to design LoadBalancerPolicy for services.

I'm in favor, but would like to hear other opinions.

gcs278 (commented):
Mostly comments trying to generate discussion.

Code updates : https://github.com/kubernetes-sigs/gateway-api/compare/4c903ae318df3625232c4df668a380ef7ed56b06..3cc17f04f64b8ab5133e710e3e69d244d98a05b7
- Rebased to fix odd diff issue
- Remove support level from the entire `LoadBalancerPolicy` resource

geps/gep-1619/index.md
```diff
@@ -442,9 +417,13 @@ obstruct the existing use of cookies while enabling session persistence. Enablin
 configurations, like allowing customization of the cookie name, could prevent certain implementations from conforming to
 the spec. In other words, opting for a higher-level API provides better interoperability among our implementations.
 
+Furthermore, this API spec doesn't allow for specifying particular forms or types of session persistence, such as
+cookie-based or header-based session persistence. It intentionally is generic to provide flexibility for implementations
+to use the mechanism that best fits their requirements.
```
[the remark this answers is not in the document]
gcs278: I went ahead and updated to add the `Type` field for session persistence as it makes sense to me, but please feel free to discuss.

geps/gep-1619/index.md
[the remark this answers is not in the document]
gcs278: Added it back in. I took a stab an API design that has cookie-specific configuration (everything thus far has been generic to sessions). I used a struct called `CookieConfig` to organize the new `LifetimeType` enum field (`Session` or `Permanent`).

I'm new to API design and just tried to copy exist GWAPI patterns as best I could and naming is hard, so please, chime in.

There is a bit of an awkward situation now that if `cookieConfig.lifetimeType` = `Permanent` then `absoluteTimeoutSeconds` needs to be set (we need a value for cookie attribute `Expire` or `Max-Age`). I feel like it's either:
1. A smell for poor API design
2. Or it's acceptable, but CEL can take care of this situation by enforcing a value in `absoluteTimeoutSeconds` when `cookieConfig.lifetimeType` = `Permanent`

gcs278 (commented):
Another update:

- Added back the timeout APIs, which also required a new API to specify `LifetimeType` as I called it (session vs. permanent cookie) per @ginayeh's suggestion.
- Added a way to configure the "Type" of session persistence per @robscott's suggestion. 

Diff: https://github.com/kubernetes-sigs/gateway-api/compare/3cc17f04f64b8ab5133e710e3e69d244d98a05b7..e5269bf28eee54550a3f5cf67e626e6040e849d7

geps/gep-1619/index.md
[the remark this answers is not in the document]
costinm: I am not against having route attachment ( in addition to Service) - besides persistent sessions there are few other things that could be set for route. 

I don't think we have any requirement to use the exact same CRD in both places - I think it is very reasonable to have a backend policy that specifies load balancing and related functions - with different kinds of persistence impacting load balancing. And we can have a slightly different resource per route:  some of the concerns with the route - for example hard to define behaviors if the cookie name or other settings are set on multiple routes - can be resolved by having a simpler per-route override, for example just ability to enable/disable persistence - while keeping the details on the load balancing policies. 

Envoy has the ability to do per-route - others may not, so the per route could be an advanced feature. I expect it may take a bit of time for implementations to identify corner cases ( routes in different namespaces setting different things) and other proxies deciding if they want to support per route, but if we start with a minimal per-route API, just like we reduced the per-backend API - it may be quite useful.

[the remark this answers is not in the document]
costinm: To clarify: I like the rename to LoadBalancerPolicy and ability to have multiple things impacting endpoint selection here, as a backend policy. 

For per-route - it seems useful to also have a policy with related overrides, but I don't think we should force LoadBalancerPolicy to be used for both.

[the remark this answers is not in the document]
gcs278: Updated to add back route support as an inline route-rule. Curious to hear thoughts. I kept the session persistence config independent like @costinm suggested.

gcs278:
@robscott @costinm @ginayeh [Updated](https://github.com/kubernetes-sigs/gateway-api/compare/ab3663ecda7e6c5e8e6c200b05cf8b541339e201..0a22f9585a939c052c45b8d8817b29825e2943cc) for:
1. Add back route support via inline route rule config
2. Separate the session persistence API structs for `BackendLoadBalancePolicy` and inline route rule...curious to hear thoughts on this.
3. Change name from `LoadBalancerPolicy` to `BackendLoadBalancePolicy`
4. Add some new edge cases 

Please let me know your thoughts so we can continue to keep iterating. My goal is to generate discussion and iterate again as somethings I am still uncertain.

geps/gep-1619/index.md
```diff
@@ -49,13 +61,26 @@ session affinity, as this design is expected to be addressed within a separate G
...
+
 1. Better performance: Maintaining a single session allows a server to cache information about a client locally reducing the need for servers to exchange session data and overall storage needs.
 2. Seamless client experience: Clients can reconnect to the same server without re-authenticating or re-entering their information.
 
 Some of the concerns of session persistence are the duration and expiration of the session, security of the transaction stream, and storage of the context or state.
 
 Session affinity, not to be confused with session persistence, uses an existing attribute of the request to consistently send to the same backend. Session affinity can be considered a weaker form of session persistence: it is not guaranteed to persist a connection to the same backend server if certain attributes of the request or the backends are changed.
 
+#### A Session vs. Session Persistence
+
```
costinm, line 73: This section is quite confusing ( at least to me ).

Are you trying to say that the 'user session' concept - 'login', 'do stuff', 'logout' is broader and during the user session multiple services will use the 'session id' with either storage-based and 'stickiness' implementations ?

"Session data must be accessbile to every backend" could be explained maybe as "a user session spans multiple subsystems, and can be implemented using a combination of storage - using session ID and user as key, replicating session data  and 'sticky' backends'

If that's what you're talking about - it's useful to emphasize the 'user session' spans multiple different subsystems and backends. But we don't do a very good job in this API to capture this use case, the current API is focused on a single backend in isolation.

geps/gep-1619/index.md
```diff
@@ -363,71 +445,116 @@ type SessionPersistencePolicySpec struct {
...
-    // SessionName defines the name of the persistent session token
-    // (e.g. a cookie name).
+    // Type defines the type of session persistence such as through
+    // the use a header or cookie. Defaults to cookie based session
+    // persistence.
     //
-    // Support: Extended
+    // Support: Core for "Cookie" type
+    //
+    // Support: Extended for "Header" type
```
costinm, line 454: How about query parameter ( jsessionid is supported as both cookie and query param by java) ? Should we mention it here as extended ? Forgot about that...

Also not sure how to say that this should be a hint.

gcs278: I left it out because there isn't an active GWAPI implementation that exposes it, so it feels like it may never be used. Is it better to keep it minimal and add it only when we need? 

> Also not sure how to say that this should be a hint.

Sorry - what hint? Am I missing something?

geps/gep-1619/index.md
gcs278: > Are you trying to say that the 'user session' concept - 'login', 'do stuff', 'logout' is broader and during the user session multiple services will use the 'session id' with either storage-based and 'stickiness' implementations ?

No, not intentionally. The goal for this paragraph was just to define what a "session" (or "user serssion" as you mentioned) is. I agree the current GEP is focused on a single backend in isolation, so I don't want to start something new regarding multiple subsystem sessions.

I'll save this commit/change for a future PR (I added it a bit on a whim anyways) and we can discuss it there.

gcs278 (commented):
Addressed @costinm comments
- Dropped session vs. session persistence section for a later PR
- `sessionName` value clarification

costinm:
My concerns are about the non trivial cases - for example in a browser page
different parts  from different subsystems ( possibly from different teams,
maybe not even all in the same k8s cluster for multi cluster gateways), but
use same cookie jar.

If the API has a requirement that cookie name is set by the user - it would
be useful to at least suggest that they coordinate and don't use same name
( like 'session' ).

Not a huge problem - but how to handle conflicts can become tricky. That's
why I was happy to not have a cookie name set by the user, but if we have
it - should say few words about this.

gcs278:
@costinm yea I understand the concern with `sessionName`. Originally, I added `sessionName` because every implementation (with an existing API) allowed for cookie name configuration. Now that I'm curious, I asked around internally why someone would want to set a cookie name (because OpenShift exposes cookie name configuration too).

Seems like we have a use case for explicitly setting your cookie name to `JSESSIONID` so that it enables compatibility with another Java-based product of ours.

So I'll document this `JSESSIONID`-type use case, and to your point, I actually added a [`Session Naming Collision`](https://github.com/kubernetes-sigs/gateway-api/pull/2634/files#diff-ca53575e03e20274304cb206a5c7a8017b31ee6564238012d5a7d092708e3488R797) section which should tell implementations how to handle such collisions, but I can go further and put something in the API go doc for the users.

robscott (commented):
Thanks @gcs278! A few nits and questions but this seems like it's really close to me.

geps/gep-1619/index.md
```diff
@@ -314,36 +346,74 @@ these two configurations.
...
+    // SessionName defines the name of the persistent session token
+    // which may be reflected in the cookie or the header.
+    //
+    // Support: Implementation-specific
+    //
+    // +optional
+    // +kubebuilder:validation:MaxLength=4096
+    SessionName *string `json:"sessionName,omitempty"`
+
     // AbsoluteTimeoutSeconds defines the absolute timeout of the
```
ginayeh, line 417: Is there a way to provide a default value to `AbsoluteTimeoutSeconds`?  I assume some implementations MAY not support this and would like to turn it off by default, so it would be nice if we can have a default value to say "no absolute timeout of the persistent session".

geps/gep-1619/index.md
```diff
@@ -457,58 +577,47 @@ considered cautiously, as making associations to Gateway API's notion of Gateway
...
+[`BackendObjectReference`](/reference/spec/#gateway.networking.k8s.io%2fv1.BackendObjectReference). Enabling session
+persistence for a backend enables subsequently enables it for any route directing traffic to this backend.
+
+On the other hand, configuring the `sessionPersistence` field in the route rule enables session persistence exclusively
+for the traffic directed to the `backendRefs` in this route rule. This means applying session persistence configuration
+to a route rule MUST NOT affect traffic for other routes or route rules. Designing the configuration for a specific
+route rule section rather than the route entirely, allows users to configure session persistence in a more granular
+fashion. This approach avoids the need to decompose routes if the configuration is specific to a route path.
+
+Session persistence configuration in `BackendLoadBalancePolicy` SHALL override equivalent configuration in route
```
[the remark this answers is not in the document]
ginayeh: Also, when Route rule field overrides BackendLoadBalancePolicy, implementations MAY want to show a warning/info log.

geps/gep-1619/index.md
```diff
@@ -363,71 +433,116 @@ type SessionPersistencePolicySpec struct {
...
+type HTTPRouteRule struct {
+    [...]
+
+    // SessionPersistence defines and configures session persistence
+    // for the route rule.
+    //
+    // Support: Extended
+    //
+    // +optional
+    SessionPersistence *RouteRuleSessionPersistence `json:"sessionPersistence"`
```
ginayeh, line 534: How about merging `RouteRuleSessionPersistence` and `BackendSessionPersistence`?  I think we probably need the same fields for both backend and route rule.

gcs278:
@ginayeh @costinm @robscott Next round of updates ([diff](https://github.com/kubernetes-sigs/gateway-api/compare/52ab943291795ab345db84f09085e6ce442cb44d..f0579f8fd87f0dac83c0c9a016399c2e2d3922b3)):
- Updated title of GEP
- Changed name from `BackendLoadBalancePolicy` to `BackendLBPolicy`
- Made the route API common with `BackendLBPolicy`
- Added a clause to route API to basically say, it's not done, reevaluate and added another `Implementable` criteria for this
- Changed Timeouts to Extended and use `Duration` field (added as new commit)

geps/gep-1619/index.md
```diff
@@ -457,58 +583,44 @@ considered cautiously, as making associations to Gateway API's notion of Gateway
...
+persistent session is established and traffic splitting is configured across services, the persistence to a single
+backend MUST be maintained across services. Consequently, a persistent session takes precedence over traffic split
+weights when selecting a backend after route matching. It's important to note that session persistence does not impact
+the process of route matching.
 
 When using multiple backends in traffic splitting, all backend services should have session persistence enabled.
-Nonetheless, implementations should also support traffic splitting scenarios in which one service has persistence
+Nonetheless, implementations MUST also support traffic splitting scenarios in which one service has persistence
 enabled while the other does not. This support is necessary, particularly in scenarios where users are transitioning
 to or from an implementation version designed with or without persistence.
```
ginayeh, line 621: Idealy, we should have all backend services enable session persistence in traffic splitting scenarios. However, I want to get some clarity on this traffic splitting scenario of one service (svc1) has persistence enabled while the other (svc2, svc3) does not.  In this scenario, if we keep session persistence only applied to svc1, all traffic will eventually goes to svc1 because the persistence should be maintained, resulting in traffic overloading svc1 and may break the system.  A more reasonable interpretation of this is that session persistence MUST be applied to all backends (svc1, svc2, svc3) involved in traffic splitting if any of them have session persistence enabled.

costinm:
I can confirm that at least in CloudRun session persistence ( which has
been around for quite some time), the persistence can be enabled on svc1
while others (svc2, svc3 ) do not have it enabled.
I don't think we can impose a MUST on the opposite behavior - probably
other implementations follow the same model.

Applying persistence on svc2 and svc3 may also break them - because the
traffic distribution will be different from what a backend not using
persistent sessions expect ( as well as draining, etc).

Traffic may move to svc1 - which is intended behavior if user is
transitioning to a new version that is using session persistence, and
should not cause any problems with auto-scalling,
Or if user is moving from persistent session to stateless - it is possible
to set svc1 with a weight=0, and no new request will go to it - so use can
transition to svc2.

I don't mind if we allow some implementations to treat one backend having
persistence as all backends having it - migration is a transitory thing and
pretty rare. Also that's the
current Istio behavior.

So far we are not aware of any customers having problems with either
approach.

geps/gep-1619/index.md
gcs278: I think you have a good point that it may be much more practical to apply session persistence to all backends given one backend has session persistence configured, but like @costinm mentioned, I think it might be too prescriptive to say it must be done that way.

Do you see any problem with leaving the spec open for implementations to do both?:
1. Maintain only session persistence to svc1 (eventually transitioning all traffic to svc1, possibly overloading it)
2. Apply session persistence to all svc1-3

costinm:
The reason svc1 would be overloaded is if it doesn't use auto-scaling. High
traffic can overload any service that doesn't
correctly auto-scale or shedds load.

Load balancers can (should) also use load reporting and track capacity of a
backend - regardless of traffic splitting
weights a properly configured workload with a good load balancer should
never overload - new sessions should
not be sent to an endpoint that is at capacity.

Without sessions - if you have a 10%->v1 90%->v2 split but v1 is
under-sized and can only handle 5% - I would
expect a load-aware LB to send 5% and prioritize the health/reliability of
the system over the API. At some
point we may need to discuss load shedding ( not in this GEP ), which is
different from draining and another thing
that may make an LB ignore the split settings.

geps/gep-1619/index.md
ginayeh: Let's leave the spec open and have implementations to decide which behavior they want since it's an edge case.

youngnick:
Updates #1619

geps/gep-1619/index.md
gcs278: Done in a new commit: https://github.com/kubernetes-sigs/gateway-api/pull/2634/commits/13eaa83fb171a836cf59496b57bbff31145351a3

geps/gep-1619/index.md
```diff
@@ -3,22 +3,33 @@
...
+### Standard
 
-1. Should we leave room in this policy to add additional concepts in the future
-   such as Session Affinity? If so, how would we adjust the naming and overall
-   scope of this policy?
-2. Should we leave room for configuring different forms of Session Persistence?
-   If so, what would that look like?
+Before this GEP graduates to the Standard channel, we must fulfill the following criteria:
+
+- Sign-off from the GAMMA leads to ensure service mesh gets fully considered.
```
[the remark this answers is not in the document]
robscott: /assign @mikemorris

gcs278:
Last [update](https://github.com/kubernetes-sigs/gateway-api/compare/d77629ab60580abaa02a2c99d3d874e8f24c686c..bbab28a8b094c0bbbf509d2e3de843eefb5a5265) just capitalized a `MUST`.

robscott:
Thanks for all the _persistence_ on this @gcs278! This all LGTM. Will defer to @shaneutt or @youngnick for final sign off. 

/approve

shaneutt (approved):
/ok-to-test
/unhold
/approve
/lgtm

PR #2652, opened 2023-12-07
Add validation for https tls mode

zhaohuabing:
/kind bug

This PR adds tls mode validation for https listeners.

In the current Gateway API, the tls mode can be configured as Passthrough for an https listener, which should not be allowed.  Rather than relying on the Gateway implementations to catch and handle this misconfiguration, it can be checked earlier at the Gateway creation/modification stage by the API server with a cel expression.

```release-note
NONE
```

### DIFF
--- a/apis/v1/gateway_types.go
+++ b/apis/v1/gateway_types.go
@@ -188,7 +188,7 @@ type GatewaySpec struct {
 	// +kubebuilder:validation:MaxItems=64
 	// +kubebuilder:validation:XValidation:message="tls must be specified for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
-	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol==HTTPS ? l.tls==Terminate)"
+	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol == 'HTTPS' ? (l.tls.mode == '' || l.tls.mode == 'Terminate') : true)"
 	// +kubebuilder:validation:XValidation:message="hostname must not be specified for protocols ['TCP', 'UDP']",rule="self.all(l, l.protocol in ['TCP', 'UDP']  ? (!has(l.hostname) || l.hostname == '') : true)"
 	// +kubebuilder:validation:XValidation:message="Listener name must be unique within the Gateway",rule="self.all(l1, self.exists_one(l2, l1.name == l2.name))"
 	// +kubebuilder:validation:XValidation:message="Combination of port, protocol and hostname must be unique for each listener",rule="self.all(l1, self.exists_one(l2, l1.port == l2.port && l1.protocol == l2.protocol && (has(l1.hostname) && has(l2.hostname) ? l1.hostname == l2.hostname : !has(l1.hostname) && !has(l2.hostname))))"
--- a/apis/v1/gateway_types.go
+++ b/apis/v1/gateway_types.go
@@ -188,7 +188,7 @@ type GatewaySpec struct {
 	// +kubebuilder:validation:MaxItems=64
 	// +kubebuilder:validation:XValidation:message="tls must be specified for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
-	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol == 'HTTPS' ? (l.tls.mode == '' || l.tls.mode == 'Terminate') : true)"
+	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol == 'HTTPS' ? (has(l.tls) && (l.tls.mode == '' || l.tls.mode == 'Terminate')) : true)"
 	// +kubebuilder:validation:XValidation:message="hostname must not be specified for protocols ['TCP', 'UDP']",rule="self.all(l, l.protocol in ['TCP', 'UDP']  ? (!has(l.hostname) || l.hostname == '') : true)"
 	// +kubebuilder:validation:XValidation:message="Listener name must be unique within the Gateway",rule="self.all(l1, self.exists_one(l2, l1.name == l2.name))"
 	// +kubebuilder:validation:XValidation:message="Combination of port, protocol and hostname must be unique for each listener",rule="self.all(l1, self.exists_one(l2, l1.port == l2.port && l1.protocol == l2.protocol && (has(l1.hostname) && has(l2.hostname) ? l1.hostname == l2.hostname : !has(l1.hostname) && !has(l2.hostname))))"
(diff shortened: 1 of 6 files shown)


gauravkghildiyal:
Unless I'm missing something, isn't this by design: 

/hold Until a maintainer can verify the same.

Ref: https://gateway-api.sigs.k8s.io/faq/
image

apis/v1/gateway_types.go
```diff
@@ -188,6 +188,7 @@ type GatewaySpec struct {
 	// +kubebuilder:validation:MaxItems=64
 	// +kubebuilder:validation:XValidation:message="tls must be specified for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
+	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol==HTTPS ? l.tls==Terminate)"
```
robscott, line 191: This is surprisingly tricky. I agree that if Listener protocol is "HTTPS", we've said that the only valid TLS Mode is "Terminate" ([docs](https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls)). Of course that is also the [default value for TLS mode](https://github.com/kubernetes-sigs/gateway-api/blob/a6c0e9ce5dd38cc3a065a672d4132f70a63cf087/apis/v1/gateway_types.go#L394), and I'm not sure whether CEL or defaulting comes first, I think it's defaulting, but we need to have some tests that cover that case.

For example, I think the following example should still be valid (relying on defaulting) if we remove the mode from the config: https://github.com/kubernetes-sigs/gateway-api/blob/a6c0e9ce5dd38cc3a065a672d4132f70a63cf087/examples/standard/simple-http-https/gateway.yaml#L1-L20

It would be good to add two variations of that example - one alongside this just without mode specified (this would be valid) and a separate copy in `hack/invalid-examples` that sets mode to "Passthrough" (invalid).

As far as the overall CEL, I think you add least need to check for TLS Mode specifically, but this may need further adjustments depending on defaulting behavior.

```suggestion
	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol==HTTPS ? l.tls.Mode==Terminate)"
```

It's also challenging to make a backwards incompatible change to validation on a GA API like this, but I believe this is the correct decision. In this case I believe it falls under "Correcting validation" in our [versioning guidelines](https://gateway-api.sigs.k8s.io/concepts/versioning/#patch-version-eg-v040-v041). Assuming we can get this right, this will only invalidate HTTPS listeners with TLS Mode set to "Passthrough" which is a state the spec has long considered invalid.

robscott (commented):
Thanks for catching this @zhaohuabing! Although this is going to be a tricky change, I think it's the correct one.

apis/v1/gateway_types.go
gauravkghildiyal: Thanks Rob for the clarification regarding supportability!

I wonder if we need the "else" clause of CEL ternary expression...hopefully the test run should reveal this.

robscott: Oh yep, good call, would also need an `:` in the ternary.

zhaohuabing:
I have no idea why this failed.  @robscott @gauravkghildiyal Can you suggest how can I fix this?

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_gateway-api/2652/pull-gateway-api-verify/1733011861806256128

```
No version requested with BASE_REF, nothing to do.
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/gateway-api'
Uncommitted changes in generated sources:
 M config/crd/experimental/gateway.networking.k8s.io_gateways.yaml
 M config/crd/standard/gateway.networking.k8s.io_gateways.yaml
[0;31mTest FAILED: hack/../hack/verify-codegen.sh 
go: downloading sigs.k8s.io/kind v0.20.0
go: downloading github.com/spf13/pflag v1.0.5
go: downloading github.com/spf13/cobra v1.4.0
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/alessio/shellescape v1.4.1
1151 skipped lines...
customresourcedefinition.apiextensions.k8s.io "referencegrants.gateway.networking.k8s.io" deleted
Deleting cluster "verify-gateway-api" ...
Deleted nodes: ["verify-gateway-api-control-plane"]
SUCCESS: hack/../hack/verify-crds-kind.sh 
Verifying docker images
ERROR: no builder "gateway-api-builder" found
gateway-api-builder
Building and pushing admission-server image...linux/amd64,linux/arm64
#0 building with "gateway-api-builder" instance using docker-container driver
#1 [internal] booting buildkit
347 skipped lines...
Successfully installed pathspec-0.11.2 pyyaml-6.0.1 yamllint-1.29.0
./conformance/tests/httproute-backend-protocol-h2c.yaml
  11:3      warning  comment not indented like content  (comments-indentation)
SUCCESS: hack/../hack/verify-yamllint.sh 
make: *** [Makefile:122: verify] Error 1
+ EXIT_VALUE=2
+ set +o xtrace
Cleaning up after docker in docker.
================================================================================
Cleaning up after docker
```

gauravkghildiyal:
@zhaohuabing Can you please try running `./hack/update-codegen.sh` and push the changed files?

apis/v1/gateway_types.go
```diff
@@ -188,6 +188,7 @@ type GatewaySpec struct {
 	// +kubebuilder:validation:MaxItems=64
 	// +kubebuilder:validation:XValidation:message="tls must be specified for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
```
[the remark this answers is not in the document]
gauravkghildiyal: Interesting. Does this mean that there's also a gap in our documentation? (I don't have an answer)
image

zhaohuabing:
@robscott @gauravkghildiyal @candita is there anything I can do to move this PR forward? Thanks.

apis/v1/gateway_types.go
```diff
@@ -188,6 +188,7 @@ type GatewaySpec struct {
...
+	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, l.protocol == 'HTTPS' ? (l.tls.mode == '' || l.tls.mode == 'Terminate') : true)"
```
gauravkghildiyal, line 191: Maybe we can also verify if `l.tls` actually exists to be more precise on when the error is returned.

```
l.protocol == 'HTTPS' && has(l.tls)
```

gauravkghildiyal (commented):
Sorry for the delay. Just one final suggestion and rest looks good to me.

apis/v1/gateway_types.go
gauravkghildiyal: @zhaohuabing -- Thanks for the update! The behaviour that I had in mind was to show "one exact error":
```
(l.protocol == 'HTTPS' && has(l.tls)) ? (l.tls.mode == '' || l.tls.mode == 'Terminate') : true 
```

As opposed to the following which will also display `tls mode must be Terminate for protocol HTTPS`, when the `tls` field is missing:
```
l.protocol == 'HTTPS' ? (has(l.tls) && (l.tls.mode == '' || l.tls.mode == 'Terminate')) : true
```

gauravkghildiyal:
Thanks a lot @zhaohuabing 

/lgtm

robscott:
Hey @zhaohuabing, thanks for checking on this! This PR LGTM, but I'd really appreciate some additional reviews from others here since I think we've actually been too strict in other related validation (x-ref https://github.com/kubernetes-sigs/gateway-api/issues/2668).

/cc @arkodg @mlavacca @sunjayBhatia @youngnick

apis/v1/gateway_types.go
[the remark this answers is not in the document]
candita: @zhaohuabing @gauravkghildiyal there is possibly a gap in our documentation: https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/guides/tls.md?plain=1#L31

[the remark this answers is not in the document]
candita: I also just noticed a related PR: https://github.com/kubernetes-sigs/gateway-api/pull/2263

arkodg (approved):
/approve

sunjayBhatia (approved):
/lgtm

I believe we already validate this in contour's implementation but would be good to have this at the CRD validation level

robscott:
Thanks @zhaohuabing!

/approve

PR #2676, opened 2023-12-13
Bump parent images for echo apps

jwendell:
In particular, the newer "advanced" echo no longer runs as root (it now has `USER` configured), which makes it possible to run in more strict environments like OpenShift.

```release-note
NONE
```


howardjohn (approved):

robscott:
Thanks @jwendell!

/ok-to-test
/lgtm
/approve

PR #2680, opened 2023-12-15
Adding new CRD Management Guide

### LINKED ISSUE #2319: Document the upcoming deprecation and eventual removal of the Validating Webhook

### LINKED ISSUE #2672: Docs: Include Guidance for Upgrading Storage Version
**What happened**:
In https://github.com/kubernetes-sigs/gateway-api/issues/2671, we saw our first reported case of a storage version issue. In general, this will happen to any users that have installed Gateway API v0.5 or earlier, and have some resources such as a GatewayClass that have not been updated since they upgraded to v0.6 or newer (the version that changed the storage version from v1alpha2 to v1beta1). In v1.0 we removed v1alpha2, so if a cluster has any Gateway API resources that have not been updated since v0.6 or newer was installed in a cluster, the upgrade to v1.0 CRDs will fail.

Ideally, we'd have documentation that clearly showed what could happen and how users can prevent this issue.

robscott:
**What type of PR is this?**
/kind documentation

This adds a new CRD Management Guide. This also moves other guides to the "guides" section of the documentation and makes some small tweaks to those in the process.

Fixes #2672
Fixes #2319 

```release-note
NONE
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -27,9 +27,11 @@ plugins:
   - redirects:
       redirect_maps:
         'guides/getting-started.md': 'guides/index.md'
+        'concepts/guidelines.md': 'guides/api-design.md'
         'contributing/community.md': 'contributing/index.md'
         'contributing/gamma.md': 'concepts/gamma.md#contributing'
-        'references/implementers-guide.md': 'reference/implementers-guide.md'
+        'reference/implementers-guide.md': 'guides/implementers.md'
+        'references/implementers-guide.md': 'guides/implementers.md'
         'references/spec.md': 'reference/spec.md'
         'references/policy-attachment.md': 'reference/policy-attachment.md'
   - mermaid2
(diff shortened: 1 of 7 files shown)


youngnick:
/lgtm
/hold in case someone else wants to look.

Thanks @robscott!

robscott:
Discussed at today's community meeting, no objections so removing the hold.

/hold cancel

PR #2687, opened 2023-12-18
Allow an implementation to use listed order for HTTPRouteFilter in a route-rule

### LINKED ISSUE #1598: Support for specifying the filter execution order in HTTPRoute

There are scenarios where the filters should be applied in a predefined order.  But the filters does not support a way to order individual filter since it's an unordered list.

For example following HTTPRoute has multiple filters,
```
kind: HTTPRoute
metadata:
  name: http-route-webhook
  namespace: apk
spec:
  hostnames:
  - my.gw.org
  parentRefs:
    kind: Gateway
    name: Default
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: web-hook-service
      port: 80
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /my-api/7.6.0
    filters:
      - type: ExtensionRef
        extensionRef:
            group: ""
            kind: MyPolicy
            name: my-policy-2
      - type: ExtensionRef
        extensionRef:
            group: ""
            kind: MyPolicy
            name: my-policy-1
      - type: URLRewrite
        urlRewrite:
          path:
            type: ReplacePrefixMatch
            replacePrefixMatch: /2457afc0-c3c0-46a0-a389-a84025b8eabe

```

How can someone specify the order like `my-policy-1`,  `my-policy-2`, and then `urlRewrite` filters are applied in order?

Docs says this is not supported yet:  https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteFilter

> "The effects of ordering of multiple behaviours are currently unspecified.". 

We would like to have the support for this.

Any thought on this?

Thanks!

sanjaypujare:
**What type of PR is this?**

/kind documentation
/area conformance

This PR fixes #1598 by allowing an implementation to optionally implement listed order for HTTPRouteFilters in a route-rule. There are implementations (as described in #1598 and discussion #2628) where the behavior depends on certain order for the filters so we would like to have this specified as an optional behavior in the spec.

Fixes #1598

```release-note
Possibly: users of certain implementations (which will conform to this new optional behavior) may need to reorder the filters they have listed in their routes.
```

### DIFF
--- a/apis/v1/httproute_types.go
+++ b/apis/v1/httproute_types.go
@@ -205,7 +205,7 @@ type HTTPRouteRule struct {
 	// that behavior.
 	//
 	// To reject an invalid combination or order of filters, implementations SHOULD
-	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
+	// consider the Route Rules with this configuration invalid. If all Route Rules
 	// in a Route are invalid, the entire Route would be considered invalid. If only
 	// a portion of Route Rules are invalid, implementations MUST set the
 	// "PartiallyInvalid" condition for the Route.
--- a/apis/v1/httproute_types.go
+++ b/apis/v1/httproute_types.go
@@ -196,8 +196,11 @@ type HTTPRouteRule struct {
 	// Filters define the filters that are applied to requests that match
 	// this rule.
 	//
-	// The effects of ordering of multiple behaviors are currently unspecified.
-	// This can change in the future based on feedback during the alpha stage.
+	// This spec does not mandate a specific ordering for the execution of filters.
+	// However an implementation may decide to execute the filters in the listed order.
+	// In case the implementation cannot support the listed order (due to its internal
+	// ordering limitation), the implementation will drop the route rule with such a
+	// conflict and set the status RouteConditionPartiallyInvalid for the Route.
 	//
 	// Conformance-levels at this level are defined based on the type of filter:
 	//
--- a/apis/v1/httproute_types.go
+++ b/apis/v1/httproute_types.go
@@ -196,11 +196,19 @@ type HTTPRouteRule struct {
 	// Filters define the filters that are applied to requests that match
 	// this rule.
 	//
-	// This spec does not mandate a specific ordering for the execution of filters.
-	// However an implementation may decide to execute the filters in the listed order.
-	// In case the implementation cannot support the listed order (due to its internal
-	// ordering limitation), the implementation will drop the route rule with such a
-	// conflict and set the status RouteConditionPartiallyInvalid for the Route.
+	// Wherever possible, implementations SHOULD implement filters in the order
+	// they are specified. 
+	//
+	// Implementations MAY choose to implement this ordering strictly, rejecting
+	// any combination or order of filters that can not be supported. If implementations
+	// choose a strict interpretation of filter ordering, they MUST clearly document
+	// that behavior.
+	//
+	// To reject an invalid combination or order of filters, implementations SHOULD
+	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
+	// in a Route are invalid, the entire Route would be considered invalid. In only
+	// a portion of Route Rules are invalid, implementations should set the 
+	// "PartiallyInvalid" condition for the Route.
 	//
 	// Conformance-levels at this level are defined based on the type of filter:
 	//
--- a/apis/v1/httproute_types.go
+++ b/apis/v1/httproute_types.go
@@ -206,7 +206,7 @@ type HTTPRouteRule struct {
 	//
 	// To reject an invalid combination or order of filters, implementations SHOULD
 	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
-	// in a Route are invalid, the entire Route would be considered invalid. In only
+	// in a Route are invalid, the entire Route would be considered invalid. If only
 	// a portion of Route Rules are invalid, implementations MUST set the
 	// "PartiallyInvalid" condition for the Route.
 	//
--- a/apis/v1/httproute_types.go
+++ b/apis/v1/httproute_types.go
@@ -197,7 +197,7 @@ type HTTPRouteRule struct {
 	// this rule.
 	//
 	// Wherever possible, implementations SHOULD implement filters in the order
-	// they are specified. 
+	// they are specified.
 	//
 	// Implementations MAY choose to implement this ordering strictly, rejecting
 	// any combination or order of filters that can not be supported. If implementations
@@ -207,7 +207,7 @@ type HTTPRouteRule struct {
 	// To reject an invalid combination or order of filters, implementations SHOULD
 	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
 	// in a Route are invalid, the entire Route would be considered invalid. In only
-	// a portion of Route Rules are invalid, implementations should set the 
+	// a portion of Route Rules are invalid, implementations MUST set the
 	// "PartiallyInvalid" condition for the Route.
 	//
 	// Conformance-levels at this level are defined based on the type of filter:
--- a/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_httproutes.yaml
@@ -1178,12 +1178,16 @@ spec:
                       type: array
                     filters:
                       description: "Filters define the filters that are applied to
-                        requests that match this rule. \n The effects of ordering
-                        of multiple behaviors are currently unspecified. This can
-                        change in the future based on feedback during the alpha stage.
-                        \n Conformance-levels at this level are defined based on the
-                        type of filter: \n - ALL core filters MUST be supported by
-                        all implementations. - Implementers are encouraged to support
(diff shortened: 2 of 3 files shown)


sanjaypujare:
gentle ping @youngnick @shaneutt

apis/v1/httproute_types.go
```diff
@@ -196,8 +196,11 @@ type HTTPRouteRule struct {
 	// Filters define the filters that are applied to requests that match
 	// this rule.
 	//
-	// The effects of ordering of multiple behaviors are currently unspecified.
-	// This can change in the future based on feedback during the alpha stage.
+	// This spec does not mandate a specific ordering for the execution of filters.
+	// However an implementation may decide to execute the filters in the listed order.
+	// In case the implementation cannot support the listed order (due to its internal
+	// ordering limitation), the implementation will drop the route rule with such a
+	// conflict and set the status RouteConditionPartiallyInvalid for the Route.
```
robscott, line 203: I'm not really sure how to word this, but at a high level, I think it makes sense for us to encourage all implementations to implement filters in the order they are specified wherever possible. The stricter mode also seems like it's worth mentioning as a potential option here. Note that the "SHOULD" and "MAY" are both using the [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) definitions of those words (as elsewhere in the spec).

```suggestion
	// Wherever possible, implementations SHOULD implement filters in the order
	// they are specified. 
	//
	// Implementations MAY choose to implement this ordering strictly, rejecting
	// any combination or order of filters that can not be supported. If implementations
	// choose a strict interpretation of filter ordering, they MUST clearly document
	// that behavior.
	//
	// To reject an invalid combination or order of filters, implementations SHOULD
	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
	// in a Route are invalid, the entire Route would be considered invalid. In only
	// a portion of Route Rules are invalid, implementations should set the 
	// "PartiallyInvalid" condition for the Route.
```

I'm not 100% sure on this particular variation, so would appreciate some additional feedback.

/cc @arkodg @howardjohn @mlavacca @sunjayBhatia @youngnick

sanjaypujare: I like the suggested wording by @robscott. Waiting to see the feedback from requested folks and after that I will accept it.

sanjaypujare:
Failure does not seem related to my change

/retest-required

robscott:
@sanjaypujare sadly our tooling is pretty limited so any changes to go types that result in changes to the CRDs also need to manually run `make generate` locally to generate the updates to the CRDs and other generated code.

sanjaypujare:
> @sanjaypujare sadly our tooling is pretty limited so any changes to go types that result in changes to the CRDs also need to manually run `make generate` locally to generate the updates to the CRDs and other generated code.

Not a problem. But in this case the changes are only in the comments so not changing any real Go code. I guess that's still considered a change to the Go type?

sanjaypujare:
> ...
> Let me try this `make generate` stuff.

Done.

apis/v1/httproute_types.go
youngnick: I like the wording as well, although I would replace the last "should" with either MUST or SHOULD, so it's clear we're using the RFC meaning.

That is:
```
		// a portion of Route Rules are invalid, implementations (MUST|SHOULD) set the 
		// "PartiallyInvalid" condition for the Route.
```

I can't remember if the PartiallyInvalid condition is MUST or SHOULD elsewhere, sorry.

sanjaypujare: Changed the `should` to `MUST` as suggested

apis/v1/httproute_types.go
```diff
@@ -196,8 +196,19 @@ type HTTPRouteRule struct {
...
+	// Wherever possible, implementations SHOULD implement filters in the order
+	// they are specified.
+	//
+	// Implementations MAY choose to implement this ordering strictly, rejecting
+	// any combination or order of filters that can not be supported. If implementations
+	// choose a strict interpretation of filter ordering, they MUST clearly document
+	// that behavior.
+	//
+	// To reject an invalid combination or order of filters, implementations SHOULD
+	// consider the Route Rule(s) with this configuration invalid. If all Route Rule(s)
```
robscott, line 208: On further thought, I think the `(s)` was unnecessary. 
```suggestion
	// consider the Route Rules with this configuration invalid. If all Route Rules
```

apis/v1/httproute_types.go
```diff
@@ -196,8 +196,19 @@ type HTTPRouteRule struct {
...
+	// in a Route are invalid, the entire Route would be considered invalid. In only
```
robscott, line 209: 🤦 
```suggestion
	// in a Route are invalid, the entire Route would be considered invalid. If only
```

robscott (commented):
Thanks @sanjaypujare! A couple nits (I think these were my mistakes), but otherwise LGTM.

apis/v1/httproute_types.go
```diff
@@ -196,8 +196,19 @@ type HTTPRouteRule struct {
 	// Filters define the filters that are applied to requests that match
 	// this rule.
 	//
-	// The effects of ordering of multiple behaviors are currently unspecified.
-	// This can change in the future based on feedback during the alpha stage.
+	// Wherever possible, implementations SHOULD implement filters in the order
```
arkodg, line 199: Id prefer if we removed `Wherever possible` and instead
* made ordering for `CORE` filters a `MUST`
* made ordering for `EXTENDED` and `IMPLEMENTATION SPECIFIC` filters a `SHOULD`

should be easy to a conformance test for `CORE` by adding combinations of - `set`, `add`, & `remove` settings of `requestHeaderModifier` filter

sanjaypujare: I don't understand why the ordering requirement is linked to CORE vs extended/implementation specific filters. Also if the filters list contains, say, both core and extended in a specific order the MUST requirement only applies to the core filters subset (or subsequence)?

robscott: It looks like we consider RequestHeaderModifier and RequestRedirect as our core filters and the rest are extended:

https://github.com/kubernetes-sigs/gateway-api/blob/8e62cc338e69bec9cb9fd01cf59af421e3aba879/apis/v1/httproute_types.go#L748-L799

If we _required_ implementations to have a strict interpretation of these filters, that would mean that RequestHeaderModifier would always have to be specified before RequestRedirect. Although this is probably reasonable in practice, it could end up breaking existing users which would not be good.

We do actually have validation that prevents the RequestHeaderModifier filter from being repeated though: https://github.com/kubernetes-sigs/gateway-api/blob/8e62cc338e69bec9cb9fd01cf59af421e3aba879/apis/v1/httproute_types.go#L225

I'm hesitant to add a new _requirement_ to the spec here. This could complicate adding future filters, but the more problematic aspect is that it could invalidate some existing configuration.

My rationale for suggesting `Wherever possible` is that I wanted to clearly recognize that many implementations simply cannot adjust the order of when certain filters are executed, I think that still applies here.

arkodg: I was trying to find wording that can be converted into conformance tests :), but this works, to an implementor and user, this reads as "best effort", which sgtm

sunjayBhatia: Yeah at least with the existing set of "core" filters, I'm not sure how much utility there would be in enforcing ordering for these filters, but maybe down the line depending on what is promoted to core. If we do intend to make respecting ordering for core filters a requirement, we should add a conformance test, e.g. if response header modification becomes core then asserting a redirect response does/does not have certain modified headers depending if the filters are in the right order.

My understanding of https://github.com/kubernetes-sigs/gateway-api/issues/1598 was that we just wanted to add some content to the spec saying that the list of filters might be interpreted as ordered, depending on the implementation. I might get rid of the `SHOULD` sentence that starts this diff and start off the change with something like `Implementations MAY choose to implement filters in the order they are specified. Such implementations MAY choose to implement this ordering strictly...`

sunjayBhatia (commented):
left a comment!

apis/v1/httproute_types.go
sunjayBhatia: I could go either way on the "wherever possible ... SHOULD" wording, if the intent is to intentionally be prescriptive then it makes sense, I think there are other instances of SHOULD being used to that end in the spec

sanjaypujare: @robscott and others what's the final decision here? If the existing wording is okay, can we go ahead with approval and merging etc? Thanks

youngnick: I think that the SHOULD already implies the "wherever possible" - in this case it's actually making the intent more clear. I think this wording is good to go.

sanjaypujare:
@robscott @youngnick is there anything else needed before this PR can be merged?

robscott:
Thanks @sanjaypujare! I think we've got consensus here.

/lgtm
/approve

sanjaypujare:
Thanks @robscott! What about apis/v1alpha2 and apis/v1beta1 ? Do they also need to be updated with this change or it doesn't matter?

robscott:
@sanjaypujare good question - they're all just using aliases to the types in v1, so this should cover all API versions.

PR #2689, opened 2023-12-19
Add GEP-2659 and updates to GEP overview

### LINKED ISSUE #2659: GEP: Document and improve the GEP process
(Alternative names: "The GEP GEP", "I heard you like GEPS, so I made a GEP about GEPs")

As we've merged more and more GEPs, a few things have become clear:
- Not all GEPs are the same - some GEPs document an agreement for future work instead of being a place to hold work themselves
- It seems likely that we will have future GEPs extend other GEPs or make older GEPs obsolete
- We haven't been super clear on what language to use in GEPs and the spec itself

So, this issue covers a new GEP that will lay out:
- The language we use when writing GEPs (that is, we use RFC2119, and will start using its update RFC8174)
- Define relationships between GEPs
- Define the possible states GEPs can be in, and a state table for transitions

Once this GEP is agreed on and merged, this issue will only be marked complete when the existing documentation is brought into line with the new GEP.

Note that the reason to do this as a GEP itself rather than just updating the existing documentation is that this keeps our decisions in the same place (the GEPs), and allows us to use the extension mechanisms introduced in this GEP ('Extends", "Obsoletes" and so on ) in the future _on this GEP itself_ if need be.

youngnick:
**What type of PR is this?**
/kind documentation
/kind gep

This adds GEP-2659 to update the GEP process.

It also makes some changes to GEP layout:

* GEPs are now stored in individual directories.
* Each GEP now has a YAML metadata file. The schema is documented in `pkg/gep`.
* All existing GEPs have had their YAML metadata fillled out.

Fixes #2659 

```release-note
NONE
```


geps/gep-1324.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - markmc
+  - kflynn
+relationships:
+  updatedBy:
+  - number: 1426
+    name: xRoutes Mesh Binding
+    description: Adds initial Service binding proposal
+  - number: 1686
+    name: Mesh conformance testing plan
+    description: Adds initial Mesh conformance testing plan
```
robscott, line 18: Nit: Some files either have no trailing new line or an extra trailing new line.

geps/gep-1867.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+kind: GEPDetails
+number: 1867
+name: Per-Gateway Infrastructure
+type: API
+status: Experimental
+authors:
+  - howardjohn
+relationships:
+  updates:
```
robscott, line 10: Somehow "updates" doesn't feel quite right here. It's like this GEP is building on top of 1762. Maybe `dependsOn` or `extends` or some other word would fit better here?

geps/gep-2257.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+kind: GEPDetails
+number: 2257
+name: Gateway API Duration Format
+type: API
+status: Experimental
+authors:
+  -
```
robscott, line 8:
```suggestion
  - @kflynn
```

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
+# GEP-2659: Document and improve the GEP process
+
+* Issue: [#2659](https://github.com/kubernetes-sigs/gateway-api/issues/2659)
+* Type: Memorandum
+* Status: Accepted
```
robscott, line 5: Would it be simpler to just have a `Memorandum` status instead of a separate type?

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+- Add relationships between GEPs
+- Add a metadata YAML schema for GEPs, including the new relationships
+- Add some new status types to cover some only recently noticed cases
+- Update existing documentation outside this GEP to support the new material
+
+
+## Non-Goals
+
+- Update all existing GEPs with relationships or other statuses. These will be
+  done later (for existing GEPs) or as they are created (for new GEPs).
```
robscott, line 27:
```suggestion
```

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+[GEP-1897: BackendTLSPolicy](https://gateway-api.sigs.k8s.io/geps/gep-1897/)
+which obsoletes the older GEP. But we have no way of representing this
+relationship (or any other, in fact) between GEPs at the moment.
+
+With these previous two changes, the addition of a metadata YAML file, similar
+to what the KEP process uses, seems like it is increasingly necessary. This GEP
+introduces a schema, which will be detailed in `.go` files rather than completely
+in this document.
+
+Lastly, I think that we should clarify our use of [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.txt)
```
robscott, line 57: Nit: Probably avoid "I" in here?
```suggestion
Lastly, we should clarify our use of [RFC2119](https://www.rfc-editor.org/rfc/rfc2119.txt)
```

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+the "Addition of GEP relationships" section for details, but in short, when an
+API GEP has been Obsoleted, it will be moved to a new "Obsolete" phase (which
+is functionally the same as "Declined").
+
+##### Memorandum GEPs
+
+These GEPs record an agreement on either the definition of a problem, its scope
+and a common language, or further process changes to the GEP process itself.
+The defining characteristic here is that the GEP MUST NOT result in any changes
+to the Gateway API spec, and SHOULD result in further GEPs to further clarify.
```
robscott, line 109: It seems like at least some of these GEPs aren't going to require follow up. For example, we have a GEP to define how Gateway API interacts with Multi-Cluster Services. I think that could _potentially_ make a case for being a Memorandum because it's not making any API changes. At the same time, I don't think that GEP will really need any follow-up GEPs. Perhaps some updates to the existing GEP, but I think that would be it.

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+For GEPs that _do_ make changes to the API, but also require further GEPs to
+clarify, they SHOULD use the new "Updated By" relationship instead.
+
+Memorandum GEPs should be used sparingly, and should form the umbrella for a
+significant amount of work, particularly work that may have parts that can
+move through the GEP phases at different speeds.
+
+The process for Memorandum GEPs is much more simple, and has really only one
+main phase - "Accepted".
```
robscott, line 118: Similar to my comment above, if there is only one phase, maybe "Memorandum" is the phase and we don't need to bother with both phase and type here.

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+
+This phase is reached when a Memorandum GEP is merged, although as we will document
+in the "Addition of GEP relationships" section, it can still be Updated
```
robscott, line 121: Speaking of updating GEPs, this is probably overkill, but should we have a section in GEP metadata that can track major updates to GEPs with a list of links to PRs with high level summaries of what was changed in that? It's basically just git history at that point, but could potentially filter out some noise + be easier to process for someone that's not using git.

geps/gep-2659.md
```diff
@@ -0,0 +1,199 @@
...
+for this `GEPDetails` object, which will not be included in the usual CRD
+definitions for Gateway API, nor will it be part of the regular API spec.
+
+The use of a CRD is just to make the schema more similar to the other schemas
+in this repository.
+
+With that said, here's a rough sample of what a GEPMetadata object will look like:
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1alpha1
```
robscott, line 147: I'm not sure we need an APIVersion here, but if we have one, I think it should clearly denote that this is both internal only and backwards incompatible changes are expected without a version rev. I don't think it really makes sense for us to maintain multiple versions of these structs since they're intended purely for internal tooling. I also have similar thoughts about conformance reports, so hopefully whatever we do here could translate there as well.

geps/gep-1016.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
```
robscott, line 1: Small procedural thing - I think I'd rather move to a directory per GEP so we end up with something like `gep-1234/metadata.yaml` and `gep-1234/index.md`, also corresponding to the upstream k/enhancements repo. Our GEPs directory is already massive and this current change would double it.

robscott (commented):
Thanks @youngnick! Didn't quite make it through everything, but lots of great things in here.

geps/gep-1867.yaml
youngnick: Sure, `extends` is probably better.

geps/gep-2659.md
youngnick: Hmmm, maybe? I'll have a think about that.

geps/gep-2659.md
youngnick: Agreed, that's why this was a SHOULD, but I guess I could downgrade to MAY?

geps/gep-1016.yaml
youngnick: Sigh, that makes sense. I'll see what I can do there.

geps/gep-2659.md
youngnick: I added `changelog`.

geps/gep-1016.yaml
youngnick: Okay, done. Whew, this is now a much bigger PR.

geps/overview.md
```diff
@@ -18,17 +20,85 @@ This diagram shows the state diagram of the GEP process at a high level, but the
...
+
+### Relationships between GEPs
+
+GEPs can have relationships between them. At this time, there are three possible
+relationships:
+
+* **Obsoletes** and its backreference **ObsoletedBy**: when a GEP is made obsolete
+  by another GEP, and has its functionality completely replaced. The Obsoleted
+  GEP is moved to the **Declined** state.
+* **Updates** and its backreference **UpdatedBy**: when a GEP has additional details
```
robscott, line 85: Nit: I think this got changed to `Extends` now

geps/overview.md
```diff
@@ -8,8 +8,10 @@ process for the main Kubernetes project:
 1. Make changes and proposals discoverable (current and future).
 1. Document design ideas, tradeoffs, decisions that were made for
   historical reference.
+1. Record the results of larger community discussions.
+1. Record changes to the GEP process itself.
 
-## Process
+## Process diagram
```
robscott, line 14: Tiny nit, I think I prefer the single word title because it's easier to link to + preserves any existing links that may already target this section.

geps/gep-2659/index.md
```diff
@@ -0,0 +1,167 @@
...
+changelog: {}
+```
+
+### Addition of GEP relationships
+
+As you can see in the previous section, this GEP adds three relationships between
+GEPs:
+- `Obsoletes` and its backreference `ObsoletedBy` - when a GEP is made obsolete
+  by another GEP, and has its functionality completely replaced.
+- `Updates` and its backreference `UpdatedBy` - when a GEP has additional details
```
robscott, line 149: Same comment as above - this got renamed.

geps/gep-2659/index.md
```diff
@@ -0,0 +1,167 @@
...
+latestRelease: v1.1
+# references is a list of hyperlinks to relevant external references.
+# It's intended to be used for storing Github discussions, Google docs, etc.
+references: {}
+# featureNames is a list of the feature names introduced by the GEP, if there
+# are any. This will allow us to track which feature was introduced by which GEP.
+featureNames: {}
+# changelog is a list of hyperlinks to PRs that make changes to the GEP, in
+# ascending date order.
+changelog: {}
```
robscott, line 140: Should you add a self-referential link here and in the corresponding metadata.yaml to provide an example?

robscott (commented):
Thanks @youngnick! A few nits but otherwise LGTM. I think it's great to get this in while there are very few changes in flight because this would otherwise result in a lot of merge conflicts. Will give a chance for others to weigh in over the next ~24 hours, but otherwise will plan to merge once that passes + this last bit of feedback has been addressed.

geps/gep-2659/index.md
youngnick: Oh, good point, yes I should, thanks.

robscott:
Thanks @youngnick! This LGTM, but will hold off on merging for another ~24 hours in case someone else has feedback here.

/cc @mikemorris @mlavacca

robscott:
Thanks @youngnick!

/lgtm
/hold cancel

PR #2695, opened 2023-12-28
Add RegularExpressionHTTPPathModifier in HTTPPathModifierType

### LINKED ISSUE #2694: GEP: HTTPRoute support match by RegularExpression

In [gep-726,](https://github.com/kubernetes-sigs/gateway-api/issues/726). it is mentioned that Regula......dsarExpression will be added to HttpRouter as a Future Extension. Hope to promote the implementation of RegularExpression.

**Why this is needed**:
I think RegularExpression is a function with a very wide range of usage scenarios. If route matching cannot be supported natively, we need to implement it in a very complicated way. For example, in istio, the redirect function cannot support regex matching. I can only use envoyfilter to achieve. Because the Gateway API is a standard, if the Gateway API does not support RegularExpression, the istio community's support for regex matching will also be affected. I believe other communities have similar problems.

Istio issue: https://github.com/istio/istio/issues/47500 for support RegularExpression.

rokkiter:
**What type of PR is this?**

/kind feature
RegularExpression is a function with a very wide range of usage scenarios, Because the Gateway API is a standard, if the Gateway API does not support RegularExpression, the istio community's support for regex matching will also be affected. I believe other communities have similar problems.

Fixes #2694 

```release-note

```


robscott:
Hey @rokkiter! This looks very reasonable to me, but I think we'll need a small GEP before we make the API change. Beyond getting consensus on design details, GEPs help us track things like stability (experimental, standard), conformance tests, and docs. Mind creating a new GEP for this (see https://gateway-api.sigs.k8s.io/geps/overview/) and then revisiting this PR once we've agreed on the GEP?

rokkiter:
Thanks, I will create a GEP for this

PR #2696, opened 2023-12-28
gep-713: remove the type field from targetRef. Also remove reference to undefined 'indirect' type in text

sanjaypujare:
**What type of PR is this?**

/kind documentation

One of the examples uses undefined "type" field in `targetRef` which I removed. Also removed reference to "indirect" since inherited seems to be the only indirect policy attachment and inherited was already mentioned so the additional reference to "indirect" was confusing.

```release-note
NONE
```

### DIFF
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -1226,8 +1226,7 @@ Thus its status as EXPERIMENTAL DO NOT USE YET.
 #### Validating Admission Controller to inform users about relevant Policy
 
 Implementations MAY supply a Validating Admission Webhook that will return a
-WARNING message when an applied object is affected by some Policy, which may be 
-an inherited one.
+WARNING message when an applied object is affected by some Policy.
 
 The warning message MAY include the name, namespace, apiGroup and Kind of relevant
 Policy objects.
--- a/geps/gep-713/index.md
+++ b/geps/gep-713/index.md
@@ -463,7 +463,6 @@ spec:
       cachePolicy:
         includeQueryString: false
   targetRef:
-    type: direct
     kind: HTTPRoute
     name: example
 ```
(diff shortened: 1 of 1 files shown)


arkodg (commented):
/approve

arkodg (commented):
/ok-to-test

sanjaypujare:
Failure seems unrelated to the change.

/retest-required

geps/gep-713/index.md
```diff
@@ -1228,7 +1227,7 @@ Thus its status as EXPERIMENTAL DO NOT USE YET.
 
 Implementations MAY supply a Validating Admission Webhook that will return a
 WARNING message when an applied object is affected by some Policy, which may be 
-an inherited or indirect one.
+an inherited one.
```
robscott, line 1230: I think this meant to say "or direct". Could probably remove everything after "by some Policy" and get a simpler result.

robscott (commented):
Thanks @sanjaypujare!

/ok-to-test
/approve

geps/gep-713/index.md
sanjaypujare: Okay, will do.

The "indirect" got me thinking of another scenario. A metaresource is a resource so it can be the target of another metaresource, right? So let's say I have a resource A which is an HTTPRoute. Then I have a metaresource (policy) B whose target is A. And I have another metaresource (policy?) C whose target is B. Is that possible? If yes, then C is indirectly or transitively attached to A. Let me know if my understanding is not correct.

robscott: I think that's theoretically possible, but it has not been defined anywhere. I'd have to see a specific use case, but on the surface it seems like it would be pretty difficult for a user to follow.

youngnick: Agreed with all of the above, I think that striking the text after `by some Policy` is better.

sanjaypujare: > Agreed with all of the above, I think that striking the text after `by some Policy` is better.

Done

youngnick:
/lgtm

I'll leave the hold for @robscott to check in his morning though.

robscott:
Thanks @sanjaypujare!

/hold cancel

PR #2691, opened 2023-12-22
Significant docs cleanup and refresh

robscott:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:
This PR is pretty massive, here's an attempt at a changelog:

**Logo and Design**
* Logo files have been added
* Logo replaces title in header
* Logo added to main page
* Header background/overall primary color changes to Kubernetes primary color
* Updated main diagram on front page and added link to file that can be used to edit

**Mesh**
* A new "Mesh" section has been added, including Overview, GAMMA, and Service Facets Pages. (None of this content is new, it's just all more neatly contained in a Mesh-specific section).
* Once this merges, `gateway-api.sigs.k8s.io/mesh` will get you to this section
* Redirects from old pages have been added

**Layout**
* Tabs now show up on small displays (previously they'd hide below ~1200px width).
* GEPs have been moved from "reference" to a new "enhancements" tab
* Blog tab/navigation has been hidden, but individual pages remain for the sake of any old links out there
* Sidebar navigation is now fully expanded by default (not sure about this one, but it mostly feels like an improvement)

**Docs Consistency**
* A new documentation style guide has been added to extend the Kubernetes upstream one
* Project name has been clarified as "Gateway API", references to "the Gateway API" have been corrected
* "Experimental" and "Standard" channel callouts have been standardized
* Absolute links have been replaced with relative ones wherever possible

**Misc**
* Style of FAQs has been cleaned up throughout, now questions are included in navigation 

**Which issue(s) this PR fixes**:
Fixes #2553

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #2704, opened 2024-01-05
Reordering GEPs from most to least stable

robscott:
**What type of PR is this?**
/kind documentation

This is a follow up to @youngnick's feedback in https://github.com/kubernetes-sigs/gateway-api/pull/2691#discussion_r1442498991. 

```release-note
NONE
```

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -109,13 +109,21 @@ nav:
     - Policy Attachment: reference/policy-attachment.md
   - Enhancements:
     - Overview: geps/overview.md
-    - Provisional:
-      - geps/gep-91/index.md
-      - geps/gep-1619/index.md
-      - geps/gep-1651/index.md
-      - geps/gep-1867/index.md
-    # - Implementable:
-    #   -
+    - Memorandum:
+      - geps/gep-917/index.md
+      - geps/gep-922/index.md
+      - geps/gep-1324/index.md
+      - geps/gep-2659/index.md
+    - Standard:
+      - geps/gep-709/index.md
+      - geps/gep-718/index.md
+      - geps/gep-724/index.md
+      - geps/gep-726/index.md
+      - geps/gep-746/index.md
+      - geps/gep-820/index.md
+      - geps/gep-851/index.md
+      - geps/gep-1323/index.md
+      - geps/gep-1364/index.md
     - Experimental:
       - geps/gep-713/index.md
       - geps/gep-957/index.md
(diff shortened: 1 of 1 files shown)


robscott:
I'm not actually sure about this change, but created this PR for the sake of discussion.

/hold for consensus
/assign @youngnick

mkdocs.yml
```diff
@@ -109,13 +109,21 @@ nav:
     - Policy Attachment: reference/policy-attachment.md
   - Enhancements:
     - Overview: geps/overview.md
-    - Provisional:
```
robscott, line 112: My thought process behind the previous ordering was that _most_ people that are viewing enhancements probably want to see the enhancements that are in progress and/or have recently changed, and those are most likely to be provisional or experimental. That led to ascending levels of stability with "declined" moved all the way to the end as that effectively means "end of life" for a GEP. 

This follows Nick's suggestion to change the order to descending levels of stability. Comparing the [deploy preview for this](https://deploy-preview-2704--kubernetes-sigs-gateway-api.netlify.app/geps/overview/) with the [ordering that's live](https://gateway-api.sigs.k8s.io/geps/overview/), I think I may have a very slight preference for the state before this PR, but I don't feel strongly enough to block a change here.

youngnick: I guess my feeling is that we should be emphasizing the things that are more stable by putting them first - right now the Memorandum and Standard sections are not that populated, but over time we expect these to get much more full.

I think that the fact that the API is GA now means that we will have an increasing number of readers of this page that are concerned with the stuff that's done, rather than the stuff that's being worked on.

So I think that putting the more stable things at the top prepares us for that future.

robscott: I think that leads to an interesting question about the fundamental purpose of GEPs. Should they be used as a reference point for how we made a decision or should they take the place of documentation/API Spec and be a source of truth? I tend to think that GEPs are more to show how we reached a conclusion but that we should consider the following order in terms of authoritativeness:

1. API Spec
2. Conformance Tests
3. Documentation
4. GEP

IE when something is unclear in one, go down the list until it is, but if there's a conflict between a GEP and the API Spec, always trust the API Spec. If we agree with that general idea, I'd say the primary reasons to look at GEPs would be either:

1. Read up on a new provisional or experimental feature that is not well documented yet 
2. Read historical rationale for making certain design decisions in existing API

I'm not sure which of those would be more common, but I'd tend to lean towards the first one, which would make me lean towards the existing ordering. This is not a strong opinion though.

In either case, if we agree on the order of authoritativeness that I just came up with, I can write a follow up PR to document that somewhere.

robscott:
Closing this because it's gone stale, but feel free to reopen and merge if there's still interest.

PR #2710, opened 2024-01-11
aligning GatewayCoreFeatures to be as the rest of the CoreFeatures

LiorLieberman:
I checked and we are not breaking anyone

https://github.com/search?q=GatewayExtendedFeatures&type=code

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/utils/suite/features.go
+++ b/conformance/utils/suite/features.go
@@ -61,7 +61,7 @@ const (
 var GatewayExtendedFeatures = sets.New(
 	SupportGatewayPort8080,
 	SupportGatewayStaticAddresses,
-).Insert(GatewayCoreFeatures.UnsortedList()...)
+)
 
 // -----------------------------------------------------------------------------
 // Features - ReferenceGrant Conformance (Core)
@@ -220,6 +220,7 @@ var MeshCoreFeatures = sets.New(
 //
 // NOTE: as new feature sets are added they should be inserted into this set.
 var AllFeatures = sets.New[SupportedFeature]().
+       Insert(GatewayCoreFeatures.UnsortedList()...).
 	Insert(GatewayExtendedFeatures.UnsortedList()...).
 	Insert(ReferenceGrantCoreFeatures.UnsortedList()...).
 	Insert(HTTPRouteCoreFeatures.UnsortedList()...).


conformance/utils/suite/features.go
```diff
@@ -61,7 +61,7 @@ const (
 var GatewayExtendedFeatures = sets.New(
 	SupportGatewayPort8080,
 	SupportGatewayStaticAddresses,
-).Insert(GatewayCoreFeatures.UnsortedList()...)
```
robscott, line 64: I'd always interpreted "Extended" here as everything that is "Extended" or greater, essentially "Extended" + "Core". I'm really hesitant to change that here even if no one on GitHub is currently using this variable. I'd rather just more clearly document the current meaning.

sunjayBhatia: Looks like `HTTPRouteExtendedFeatures` does not include the core feature list, so whichever way we go, would be good to standardize on one approach

robscott: Good find @sunjayBhatia! I agree that we should be consistent across our extended feature lists here, so I take back my objection. I think Lior's idea to keep "Core" features out of "Extended" lists probably is the most sensible option here, we just should probably add comments to clearly state that distinction.

LiorLieberman: yep that was the intention of the PR, to be consistent. 

So LGTM ? 🙂

conformance/utils/suite/features.go
```diff
@@ -61,7 +61,7 @@ const (
 var GatewayExtendedFeatures = sets.New(
```
robscott, line 61: Mind adding a comment above this and the corresponding HTTPRoute list to clarify that this is _only_ Extended features and intentionally does not include Core features. That probably seems obvious, but it's likely worth clarifying because it hasn't always been that way.

LiorLieberman (commented):
clarify extended features sets dont include core features

sunjayBhatia (approved):
/lgtm

robscott:
Thanks @LiorLieberman! Sorry I lost track of this one - do you mind adding a release note so we can merge this?

/approve
/retest
/hold for release note

PR #2712, opened 2024-01-11
 GEP-2627 DNS Configuration - Initial Provisional PR

### LINKED ISSUE #2627: GEP: DNS configuration as part of Gateway API
This is a GEP tracking issue

A GEP to propose a solution for configuring DNS requirements and configuration associated with a gateway

**Why this is needed**:

Based on discussion https://github.com/kubernetes-sigs/gateway-api/discussions/2541 there is interest is potentially solving DNS configuration within Gateway API. The discussion and initial GEP will expand on the What and WHY

maleck13:
/kind gep

Adds a draft GEP outlining more on the what and why for supporting DNS configuration as part of Gateway API

Fixes #2627

### DIFF
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -8,13 +8,16 @@
 For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards.  The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API.   Instead of leaving this as an exercise for the user to figure out, this proposal attempts to provide options to ease Gateway API operations.
 
 ## Goals
-* Allow cluster operators to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener.
-* Provide a mechanism to allow the DNS configuration to be delegated to a chosen controller.
-* Provide a standard CRD-based API with expressive status reporting and remove the need for "loose" APIs such as annotations.
+* Provide a way for Gateway API resource owners to mark their resources as relevant for external DNS provisioning
+* Ensure that the above method has a way for multiple providers to be present in the cluster and be able to actuate external DNS provisioning requests
+* Ensure that any method is based on structured fields and makes the most of `status` on whatever resources are relevant, whether they are existing Gateway API resources or new resources.
 * Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
+* Clarity on the scope of hostnames under management. (This should _not_ be able to be used to affect the standard in-cluster DNS configuration)
 
 ## Non-Goals
 
+* Anything to do with configuring in-cluster DNS. This support is for configuration outside the cluster only.
+* Ways to define if the Gateway API resources are allowed to request particular hostnames. These choices should be left to the implementations that actually actuate the requests for hostnames. However, `status` flows should be specified so as to make clear if a hostname provisioning request cannot be performed.
 * Cover more complex DNS routing strategies that come into play for multi-cluster topologies such as round robin, failover, health checks, weighted and geo location with this first pass. Supporting these types of use cases for distributed gateways (e.g., in different regions or multiple gateways for resilience within a region) and offering a form of global load balancing leveraging DNS is a potential future goal.
 
 ## Use Cases
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -25,7 +25,7 @@ As a cluster administrator, I would like to have the DNS names automatically pop
 
 As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
 
-As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway. 
+As a cluster administrator, I would like the DNS records to be updated automatically if the `spec` of assigned gateways changes, whether those changes are for IP address or hostname. 
 
 ## API
 
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -10,7 +10,7 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 ## Goals
 * Allow cluster operators to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener.
 * Provide a mechanism to allow the DNS configuration to be delegated to a chosen controller.
-* Provide a standard CRD-based API as an alternative to the need for "loose" APIs such as annotations.   Support an easily extended, standardized, versioned, and status-reporting API.
+* Provide a standard CRD-based API with expressive status reporting and remove the need for "loose" APIs such as annotations.
 * Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
 
 ## Non-Goals
@@ -19,7 +19,7 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 
 ## Use Cases
 
-As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used and limit which domains can be used.
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used.
 
 As a cluster administrator, I would like to have the DNS names automatically populated into my specified DNS zones as a set of records based on the assigned addresses of my gateways so that I do not have to undertake external automation or management of this essential task.
 
@@ -32,7 +32,7 @@ As a cluster administrator, I would like the DNS records to be setup automatical
 Initial draft will not offer an API yet until the use cases are agreed. Some thoughts worth thinking about: 
 - I think it is important that we try to move away from APIs based on annotations which, while convenient, are not a full API and suffer from several limitations. An example: I want to configure a listener with a domain I own that is in a different provider than the domains of the other listeners. I want to add a new option to configure a particular weighting and so on. Soon you end up with a large set of connected annotations that often grow in complexity that really should be expressed as an API.
 
-- It is also important that this API can be delegated to controllers other than the Gateway provider. This is because there are existing solutions that may want to support whatever API decided upon. It should not **have** to be a gateway provider that has to integrate with many DNS providers. 
+- It is also important that this API can be delegated to controllers other than the Gateway API provider/implementor. This is because there are existing solutions that may want to support whatever API decided upon. It should not **have** to be a gateway provider that has to integrate with many DNS providers. 
 
 ## Conformance Details
 
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -1,0 +1,50 @@
+# GEP-2627: DNS configuration within Gateway API
+
+* Issue: [#2627](https://github.com/kubernetes-sigs/gateway-api/issues/2627)
+* Status: Provisional
+
+## TLDR
+
+For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards.  The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API.   Instead of leaving this as an exercise for the user to figure out, this proposal attempts to provide options to ease Gateway API operations.
+
+## Goals
+* Allow cluster operators or infrastructure providers to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener and have that DNS service interacted with via a controller to provision and configure the defined listener hostnames as DNS names and records based on the addresses assigned to the gateway instances.
+* Provide a standard CRD-based API as an alternative to the need for "loose" APIs such as annotations.   Support an easily extended, standardized, versioned, and status-reporting API. 
+* Increase portability and supportability across gateway providers and third party implementors, for this type of key configuration
+
+## Non-Goals
+
+* Cover more complex DNS routing strategies that come into play for multi-cluster topologies such as round robin, failover, health checks, weighted and geo location with this first pass. Supporting these types of use cases for distributed gateways (e.g., in different regions or multiple gateways for resilience within a region) and offering a form of global load balancing leveraging DNS is a potential future goal.
+
+## Use Cases
+
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used and limit which domains can be used.
+
+A a cluster administrator, I would like to have the dns names automatically populated into my specified dns zones as a set of records based on the assigned addresses of my gateways and have the status of the DNS records reported back to me, so that I do not have to undertake external automation or management of this essential task and can leverage existing kube based monitoring tools to know the status of the integration.
+
+As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway. 
+
+## API
+
+Initial draft will not offer an API yet until the use cases are agreed. Some thoughts worth thinking about: 
+- I think it is important that we try to move away from APIs based on annotations which, while convenient, are not a full API and suffer from several limitations. An example: I want to configure a listener with a domain I own that is in a different provider than the domains of the other listeners. I want to add a new option to configure a particular weighting and so on. Soon you end up with a large set of connected annotations that often grow in complexity that really should be expressed as an API.
+
+- It is also important that this API can be delegated to controllers other than the Gateway provider. This is because there are existing solutions that may want to support whatever API decided upon. It should not **have** to be a gateway provider that has to integrate with many DNS providers. 
+
+## Conformance Details
+
+(This section describes the names to be used for the feature or
+features in conformance tests and profiles.
+
+These should be `CamelCase` names that specify the feature as
+precisely as possible, and are particularly important for
+Extended features, since they may be surfaced to users.)
+
+## Alternatives
+
+it is possible to use `external-dns` to manage dns based on HTTPRoutes and Gateways https://github.com/kubernetes-sigs/external-dns/blob/7f3c10d65297ec1c4bcc8dd6f88c189b7f3e80d0/docs/tutorials/gateway-api.md. The aim of this GEP is not remove this as an option, but instead provide a common API that could then be leveraged by something like external-dns. 
+
+## References
+
+(Add any additional document links. Again, we should try to avoid
+too much content not in version control to avoid broken links)
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -29,7 +29,7 @@ As a cluster administrator, I would like to have the DNS names automatically pop
 As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
 
 As a cluster administrator, I would like the DNS records to be updated automatically if the `spec` of assigned gateways changes, whether those changes are for IP address or hostname. 
-
+As a DNS administrator, I should be able to ensure that only approved External DNS controllers can make changes to DNS zone configuration. (This should in general be taken care of by DNS system <-> External DNS controller interactions like user credentials and operation status responses, but it is important to remember that it needs to happen).
 ## API
 
 Initial draft will not offer an API yet until the use cases are agreed. Some thoughts worth thinking about: 
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -22,7 +22,7 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 
 ## Use Cases
 
-As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used.
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which gateways should be used for provisioning DNS records, and, if necessary, which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used.
 
 As a cluster administrator, I would like to have the DNS names automatically populated into my specified DNS zones as a set of records based on the assigned addresses of my gateways so that I do not have to undertake external automation or management of this essential task.
 
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -8,9 +8,10 @@
 For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards.  The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API.   Instead of leaving this as an exercise for the user to figure out, this proposal attempts to provide options to ease Gateway API operations.
 
 ## Goals
-* Allow cluster operators or infrastructure providers to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener and have that DNS service interacted with via a controller to provision and configure the defined listener hostnames as DNS names and records based on the addresses assigned to the gateway instances.
-* Provide a standard CRD-based API as an alternative to the need for "loose" APIs such as annotations.   Support an easily extended, standardized, versioned, and status-reporting API. 
-* Increase portability and supportability across gateway providers and third party implementors, for this type of key configuration
+* Allow cluster operators to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener.
+* Provide a mechanism to allow the DNS configuration to be delegated to a chosen controller.
+* Provide a standard CRD-based API as an alternative to the need for "loose" APIs such as annotations.   Support an easily extended, standardized, versioned, and status-reporting API.
+* Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
 
 ## Non-Goals
 
@@ -20,7 +21,9 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 
 As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used and limit which domains can be used.
 
-A a cluster administrator, I would like to have the dns names automatically populated into my specified dns zones as a set of records based on the assigned addresses of my gateways and have the status of the DNS records reported back to me, so that I do not have to undertake external automation or management of this essential task and can leverage existing kube based monitoring tools to know the status of the integration.
+As a cluster administrator, I would like to have the DNS names automatically populated into my specified DNS zones as a set of records based on the assigned addresses of my gateways so that I do not have to undertake external automation or management of this essential task.
+
+As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
 
 As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway. 
 
@@ -33,12 +36,7 @@ Initial draft will not offer an API yet until the use cases are agreed. Some tho
 
 ## Conformance Details
 
-(This section describes the names to be used for the feature or
-features in conformance tests and profiles.
-
-These should be `CamelCase` names that specify the feature as
-precisely as possible, and are particularly important for
-Extended features, since they may be surfaced to users.)
+TBD
 
 ## Alternatives
 
@@ -46,5 +44,4 @@ it is possible to use `external-dns` to manage dns based on HTTPRoutes and Gatew
 
 ## References
 
-(Add any additional document links. Again, we should try to avoid
-too much content not in version control to avoid broken links)
+TBD
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -15,7 +15,6 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 
 ## Non-Goals
 
-* Anything to do with configuring in-cluster DNS. This support is for configuration outside the cluster only.
 * Providing any upstream hostname validation mechanisms. We can provide status for validation failure, but implementations are responsible for validation.
 * Multi-cluster DNS for multi-cluster ingress solutions
 
--- a/geps/gep-2627/index.md
+++ b/geps/gep-2627/index.md
@@ -5,7 +5,7 @@
 
 ## TLDR
 
-For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards. The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API. Instead of leaving this unspecified and having implementations do this in potentially wildly different ways, the purpose of this proposal is to provide a standard way to specify DNS for Gateways.
+For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API. Instead of leaving this unspecified and having implementations likely to do this in different ways, the purpose of this proposal is to provide a standard way to specify DNS for Gateways.
 
 ## Goals
 * Provide DNS specification for Gateway resources
@@ -16,7 +16,7 @@ For gateway infrastructure to be valuable we need to be able to connect clients 
 ## Non-Goals
 
 * Providing any upstream hostname validation mechanisms. We can provide status for validation failure, but implementations are responsible for validation.
-* Multi-cluster DNS for multi-cluster ingress solutions
+* Multi-cluster DNS for multi-cluster ingress solutions (at least not as part of the initial API)
 
 ## Use Cases
 
@@ -27,7 +27,9 @@ As a cluster administrator, I would like to have the DNS names automatically pop
 As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
 
 As a cluster administrator, I would like the DNS records to be updated automatically if the `spec` of assigned gateways changes, whether those changes are for IP address or hostname. 
+
 As a DNS administrator, I should be able to ensure that only approved External DNS controllers can make changes to DNS zone configuration. (This should in general be taken care of by DNS system <-> External DNS controller interactions like user credentials and operation status responses, but it is important to remember that it needs to happen).
+
 ## API
 
 Initial draft will not offer an API yet until the use cases are agreed. Some thoughts worth thinking about: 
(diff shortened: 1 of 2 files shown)


candita (commented):
Some grammar fixes needed but generally looks good, could use some clarifications on the goals and use cases.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Issue: [#2627](https://github.com/kubernetes-sigs/gateway-api/issues/2627)
+* Status: Provisional
+
+## TLDR
+
+For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames they are also are assigned addresses and ports to express where on the network the gateway can be reached. The correct configuration of DNS is critical piece but is currently not expressible as part of Gateway API instead it is left as an exercise of the user to figure out. Bringing these two concerns closer together will serve to provide a single plane to express and view not only the proxy configuration but also the DNS based connectivity configuration and status.
+
+## Goals
+* Allow cluster operators or infrastructure providers to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener and have that DNS service interacted with via a controller to provision and configure the defined listener hostnames as DNS names and records based on the addresses assigned to the gateway instances.
```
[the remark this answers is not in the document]
maleck13: Ok will re-work

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Use Cases
+
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used and limit which domains can be used.
+
+A a cluster administrator, I would like to have the dns names automatically populated into my specified dns zones as a set of records based on the assigned addresses of my gateways and have the status of the DNS records reported back to me, so that I do not have to undertake external automation or management of this essential task and can leverage existing kube based monitoring tools to know the status of the integration.
```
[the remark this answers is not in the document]
maleck13: will split it up

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway. 
+
+## API
+
+Initial draft will not offer an API yet until the use cases are agreed. Some thoughts worth thinking about: 
+- I think it is important that we try to move away from APIs based on annotations which, while convenient, are not a full API and suffer from several limitations. An example: I want to configure a listener with a domain I own that is in a different provider than the domains of the other listeners. I want to add a new option to configure a particular weighting and so on. Soon you end up with a large set of connected annotations that often grow in complexity that really should be expressed as an API.
+
+- It is also important that this API can be delegated to controllers other than the Gateway provider. This is because there are existing solutions that may want to support whatever API decided upon. It should not **have** to be a gateway provider that has to integrate with many DNS providers.
```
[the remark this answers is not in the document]
maleck13: A control plane like istiod or envoygateway, not sure if there is an official term for an Gateway API implementor.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Use Cases
+
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used and limit which domains can be used.
+
+A a cluster administrator, I would like to have the dns names automatically populated into my specified dns zones as a set of records based on the assigned addresses of my gateways and have the status of the DNS records reported back to me, so that I do not have to undertake external automation or management of this essential task and can leverage existing kube based monitoring tools to know the status of the integration.
+
+As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway.
```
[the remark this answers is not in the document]
maleck13: `This would have to be configurable, not automatic. Maybe not all hostnames will be in the same IP subnet.`  by this do you mean if a gateway is assigned multiple addresses in its spec, one public and one private for example? 

`I'm not clear on use cases where the IPaddr or hostname changes - can you give an example?`  This is really to indicate that if, the assigned address in the Gateway is changed or new listeners are added, the DNS will be updated appropriately

geps/gep-2627/index.md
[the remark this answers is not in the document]
maleck13: changed

costinm:
This needs a very serious security review. As a DNS admin - either public or private - I expect very strict controls on who can create DNS entries affecting the entire domain or VPC.  It is certainly not something to enable in all clusters or namespaces - or trust that users have the security expertise to use this safely.

 I think external-dns (alpha) CRD is the right approach for users who want to use KRM to manage DNS - the other options allowing random namespace owners to create arbitrary DNS (A and TXT) entries are a huge security risk. 

I would be very strongly against this feature - DNS management is a very different thing from Gateway management, the user controlling DNS decides if a specific name (including ability to get ACME certs, etc) can be delegated to a particular Gateway in a particular cluster or not - which in turn allows specific routes (like .well-known/ for getting certs) to be sent to specific namespaces. 

The argument that 'user just want an easy way to shoot themselves in  the foot and break their own security - because they know what they're doing' is not valid for this particular case, since the implications are far broader than a single cluster.

costinm:
I should mention Istio - and any mesh based on interception, or L4-based security mechanisms - is completely dependent on DNS security - both the DNS resolver ( bob.com resolving to evil.com IP ) but also control over creation of entries in private or public DNS. 

At the very least this option should  not be allowed if any GAMMA implementation is used in the same VPC ( not only cluster).

maleck13:
@costinm Thanks for your response.
```
This needs a very serious security review. As a DNS admin - either public or private - I expect very strict controls on who can create DNS entries affecting the entire domain or VPC. It is certainly not something to enable in all clusters or namespaces - or trust that users have the security expertise to use this safely. 

I think external-dns (alpha) CRD is the right approach for users who want to use KRM to manage DNS - the other options allowing random namespace owners to create arbitrary DNS (A and TXT) entries are a huge security risk.

```
Not sure if it is clear, but the proposal and [discussion](https://github.com/kubernetes-sigs/gateway-api/discussions/2541) was focused around a new API  and as such would be controlled by regular RBAC controls just as creating a Gateway would be or creating an external dns resource would be. So there should be no issue controlling who is allowed to create such an object and in turn control the DNS for a given Gateway's listeners.

costinm:
Btw - the CRD defined by external-dns project seems like a good start and has support for most DNS servers. Any reason users can't just use that ?

youngnick:
@costinm, this is an _initial_ GEP that lays down the scope and terms to _start the discussion_ of if, and if so, how, we can represent that Gateways should have external DNS configured in a standard, portable way. I agree that there are significant security considerations, and that those security considerations should be included in the GEP.

If you have _specific_ feedback about extra use cases or other concerns that you would like addressed, the correct way to indicate them is via comments on the file, rather than broad generalizations discouraging further work on this.

Once this PR merges and we have a basis for discussion, I'd love for you to shepherd some more changes to it to ensure that your concerns are captured.

costinm:
Also not aware that discussion can happen only after this PR merge - do you
have a doc listing those rules and processes ?

youngnick:
This work is to provide a standardized way to request DNS records are created for Gateway instances - as you said earlier, basically the same as what external-dns does today. The idea is to see if we can find some common ground and provide a relatively standardized way to communicate this config, that can cover common use cases across multiple providers or implementations.

Some implementations (external-dns for one) are already looking at the Gateway object to provision DNS config, the intent here is to ensure that relevant information has a structured, extensible, portable format that it can be stored in.

The process for GEPs is documented at https://gateway-api.sigs.k8s.io/geps/overview/, and as this is the first stage (moving from discussion into Provisional), the bar is intended to be pretty low. It's intended that we use this chance to agree:
- that the problem is one we should consider
- what terms we will use to describe the problem
- what use cases we are considering

There can be multiple PRs while a GEP is in Provisional, but the intent is to provide a place that is more durable than Github comments to raise concerns, answer questions, and so on.

The best type of feedback here is feedback on the specific file, yes, but I think that questions about if the problem is in scope for Gateway API should be addressed here, so that we can all be on the same page.

I believe the problem of ensuring that there is a reliable way for Gateway owners to request provisioning of DNS records to cover access to the "outside" of the Gateway is in scope for Gateway API. If we agree on that, then the discussion becomes about:
- what config is needed for the given use cases
- where that config should live

It could be that minimal extra configuration is needed, but from the fact that external-dns and other projects already have Policy objects to cover this, it seems likely that we will need additional structured configuration of some kind. Additionally, if there are at least three implementations of the same feature, then that is the level at which we will consider doing something not provider-specific, to leave room for other implementations to do the same task in a compatible, portable way.

costinm:
As I mentioned in my previous comment - I believe this is out of scope for
the Gateway WG - and
perfectly in scope for the 'external DNS WG' - which is focused on
representing multiple
K8S resources in DNS - as well as the 'Core API WG', which define the core
DNS.

It looks like the 'external DNS' WG also believes (implicitly) it is in
their scope - given that they
already implemented support for representing gateway resources among many
others.

The Gateway WG does have a larger security problem around ownership for
hostnames -
it does solve some of the 'confused deputy' issues and has some basic
delegation/grant model,
however the 'persona' who owns the DNS zone and domain ( and indirectly all
associated security - ability
to get certs for subdomains, etc) is different from the persona configuring
a Gateway in a namespace.
DNS may span multiple clusters and non-k8s environments - and operates on a
top-down
model, and needs a consistent security and delegation model across all
record types and uses.

How to configure DNS (securely) is clearly a very important problem - but
best handled in a WG focused
on DNS, not in 'leaf' APIs that interact with a single record and small
subset - and without a
security/IAM model matching DNS needs.

The proposal doesn't mention any 'prior art' or existing LB implementation
that programs DNS -
Istio does handle DNS interception but only for egress use cases. The
permissions needed for
an in-cluster load balancer to program a customer DNS zone - and the
complexity on integrating
with the many DNS APIs ( see the list of plugins in external DNS and ACME
programs ) makes it
unlikely that even 2-3 gateway implementations will be able to do that -
while external-dns has
a working implementation that can be used with any gateway ( and vendors
can provide
specialized ones as well - most of the code is related to the 100 DNS APIs).

In general: any GEP that is not based on existing art and features that a
significant number
of gateways supports ( I would go for 'majority'...) will lead to
fragmentation and user pain with
the proliferation of optional features - as well as poor APIs that are not
based on well established
and broadly implemented designs (and I would say in Istio we have a bit of
experience with that...)

youngnick:
@costinm, my responses below.

> As I mentioned in my previous comment - I believe this is out of scope for the Gateway WG - and perfectly in scope for the 'external DNS WG' - which is focused on representing multiple K8S resources in DNS - as well as the 'Core API WG', which define the core DNS. It looks like the 'external DNS' WG also believes (implicitly) it is in their scope - given that they already implemented support for representing gateway resources among many others.

I disagree. People _using_ this API need to have _an_ answer for how they ask some DNS management system to consider their Gateway in scope for management. Right now, in external-dns, the support only looks at Route resources, which misses the whole model we have for having hostnames match between Gateway and Route. I can see you logged kubernetes-sigs/external-dns#4402 to request adding support for Gateways to the API, which I agree should definitely be supported.

The reason here is that we (Gateway API) have not documented anywhere how _consumers_ of the DNS information _already present_ in the API should use that information. We don't have anything that says "Implementations wanting to do automatic configuration must consider both Gateway and Routes in generating information". It's _implied_ by the spec (since you can't get an address without looking at the Gateway), but we haven't actually _said_ that anywhere.

@maleck13 and the other folks who are working on Kuadrant have also _already built_ a solution for managing _the same config_, in a _different_ way to the external-dns team. So, _some_ discussion here is required, even if the outcome is that we end up picking one of those implementations as the canonical one. (I think that's unlikely.)

Nobody is suggesting that Gateway API own end-to-end DNS provisioning here. But, as a traffic routing specification, we _interact_ with DNS, and unless we define how that happens, and explicitly rule things out of scope, then we will end up owning it by default. I don't want that, and it seems that you don't either. But that requries us defining what _is_ in scope.

> The Gateway WG does have a larger security problem around ownership for hostnames - it does solve some of the 'confused deputy' issues and has some basic delegation/grant model, however the 'persona' who owns the DNS zone and domain ( and indirectly all associated security - ability to get certs for subdomains, etc) is different from the persona configuring a Gateway in a namespace.

Yes, and this is one of the _many_ reasons why we haven't addressed this at all up until now.

Again, this _provisional_ GEP is _not suggesting_ that we solve this problem. In fact, I believe that this should be the responsibility of the implementation that is actuating the creation of DNS records. If a Gateway is created somewhere, and that person shouldn't be creating hostnames there, that's up to the thing that handles DNS creation, _not_ the Gateway API implementation.

Again, this GEP is about defining how other controllers that _maybe don't_ implement routing configuration can manage DNS config _sourced from_ Gateway API resources. Doing this _does not mean_ that north-south Gateway implementations need to support it. Sure, they can if they _want to_, but the idea here is to ensure that everyone doing it (which, again, _multiple implementations already are_) does it in the same way.

> DNS may span multiple clusters and non-k8s environments - and operates on a top-down model, and needs a consistent security and delegation model across all record types and uses. How to configure DNS (securely) is clearly a very important problem - but best handled in a WG focused on DNS, not in 'leaf' APIs that interact with a single record and small subset - and without a security/IAM model matching DNS needs.

This is _not yet_ a proposal, to be clear. It's an attempt to _start talking about_ what a proposal will involve. The fact that certain questions have been asked and others haven't is exactly what this process is supposed to achieve.

Also, part of the job of a Gateway is to map _addresses_ to _routing config_. Part of the routing config is the hostname, so it's reasonable natural to assume that people may want to mark hostnames on a Gateway as "please configure these automatically".

I agree that it shouldn't be the job of Gateway API to decide how the _requests_ to map an address to a hostname should be actuated. But it's _entirely within our scope_ to define something that says "use these mappings and not these ones." You're asking us to boil the ocean of solving all DNS management problems, of course that's out of scope. But marking particular mappings is _well_ inside the scope of Gateway API.


> The proposal doesn't mention any 'prior art' or existing LB implementation that programs DNS - Istio does handle DNS interception but only for egress use cases. The permissions needed for an in-cluster load balancer to program a customer DNS zone - and the complexity on integrating with the many DNS APIs ( see the list of plugins in external DNS and ACME programs ) makes it unlikely that even 2-3 gateway implementations will be able to do that - while external-dns has a working implementation that can be used with any gateway ( and vendors can provide specialized ones as well - most of the code is related to the 100 DNS APIs).

Yes, because it is at the stage of _gathering questions_, not _proposing solutions_.

> In general: any GEP that is not based on existing art and features that a significant number of gateways supports ( I would go for 'majority'...) will lead to fragmentation and user pain with the proliferation of optional features - as well as poor APIs that are not based on well established and broadly implemented designs (and I would say in Istio we have a bit of experience with that...)

I don't understand what saying this is intended to achieve. Yes, the point of the GEP process is to ensure that the features we include can be well supported by multiple implementations. Specifications that we define _don't_ have to be implemented by _every_ implementation, and I think External DNS is a great example of functionality that will probably be mostly implemented by a _different_ set of controllers than the ones that provide core routing functions.

These functions may be optional in the spec, but if we don't define them, each implementation will define their own, probably using annotations, and we will be back where we were with Ingress all over again.

I'm about to go and add some suggested changes to capture some of what we've been discussing here. If you wish to discuss this further, I think that we should move it to the community meeting or a separate dedicated meeting to get some higher-bandwidth communication.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Goals
+* Allow cluster operators to declaratively express which DNS service they want to use with a particular Gateway or Gateway Listener.
+* Provide a mechanism to allow the DNS configuration to be delegated to a chosen controller.
+* Provide a standard CRD-based API with expressive status reporting and remove the need for "loose" APIs such as annotations.
+* Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
+
+## Non-Goals
+
+* Cover more complex DNS routing strategies that come into play for multi-cluster topologies such as round robin, failover, health checks, weighted and geo location with this first pass. Supporting these types of use cases for distributed gateways (e.g., in different regions or multiple gateways for resilience within a region) and offering a form of global load balancing leveraging DNS is a potential future goal.
```
youngnick, line 18: Based on the discussion with @costinm, I suggest the following changes to make our initial scope a bit more clear:

```suggestion
## Goals
* Provide a way for Gateway API resource owners to mark their resources as relevant for external DNS provisioning
* Ensure that the above method has a way for multiple providers to be present in the cluster and be able to actuate external DNS provisioning requests
* Ensure that any method is based on structured fields and makes the most of `status` on whatever resources are relevant, whether they are existing Gateway API resources or new resources.
* Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
* Clarity on the scope of hostnames under management. (This should _not_ be able to be used to affect the standard in-cluster DNS configuration)

## Non-Goals

* Anything to do with configuring in-cluster DNS. This support is for configuration outside the cluster only.
* Ways to define if the Gateway API resources are allowed to request particular hostnames. These choices should be left to the implementations that actually actuate the requests for hostnames. However, `status` flows should be specified so as to make clear if a hostname provisioning request cannot be performed.
* Cover more complex DNS routing strategies that come into play for multi-cluster topologies such as round robin, failover, health checks, weighted and geo location with this first pass. Supporting these types of use cases for distributed gateways (e.g., in different regions or multiple gateways for resilience within a region) and offering a form of global load balancing leveraging DNS is a potential future goal.
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Use Cases
+
+As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used.
```
youngnick, line 22:
```suggestion
As a cluster administrator, I manage a set of domains and a set of gateways. I would like to declaratively define which gateways should be used for provisioning DNS records, and, if necessary, which DNS provider to use to configure connectivity for clients accessing these domains and my gateway so that I can see and configure which DNS provider is being used.
```

geps/gep-2627/index.md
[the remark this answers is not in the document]
youngnick:
```suggestion
As a cluster administrator, I would like the DNS records to be updated automatically if the `spec` of assigned gateways changes, whether those changes are for IP address or hostname. 
```

Is this a bit clearer?

geps/gep-2627/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+As a cluster administrator, I would like to have the DNS names automatically populated into my specified DNS zones as a set of records based on the assigned addresses of my gateways so that I do not have to undertake external automation or management of this essential task.
+
+As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
+
+As a cluster administrator, I would like the DNS records to be setup automatically based on the assigned gateways address and if the IP or hostname changes, I would like for the DNS to update automatically to ensure traffic continues to reach my gateway. 
+
```
youngnick, line 29:
```suggestion
As a DNS administrator, I should be able to ensure that only approved External DNS controllers can make changes to DNS zone configuration. (This should in general be taken care of by DNS system <-> External DNS controller interactions like user credentials and operation status responses, but it is important to remember that it needs to happen).
```

Suggestion to capture some of @costinm's feedback.

youngnick (commented):
I added some suggestions based on my discussion with @costinm in the main comments of the PR. Hopefully these capture the flavor of our discussion.

abursavich:
> Right now, in external-dns, the support only looks at Route resources, which misses the whole model we have for having hostnames match between Gateway and Route. I can see you logged [kubernetes-sigs/external-dns#4402](https://github.com/kubernetes-sigs/external-dns/issues/4402) to request adding support for Gateways to the API, which I agree should definitely be supported.

Hey... I'm just now seeing this issue/PR for the first time.

I implemented Gateway API support in external-dns and this claim isn't true. It has always respected Listener AllowedRoutes and Hostnames. I'd be happy to hear about and fix any cases where this isn't working properly.

abursavich:
To better support DNS use cases, I'd be happy to see a change in Gateway API to allow hostnames in Routes and Listeners that don't _need_ hostnames (i.e. move Hostnames to CommonRouteSpec). It would be mostly backwards compatible because the empty hostname matches everything. The only wrinkle is that the Hostname field in the Listener is documented as "ignored for protocols that don't require hostname based matching." It wouldn't be backwards compatible from a Go package API standpoint, but I don't think that's been a problem before :)

HTTPRoutes, TLSRoutes, and GRPCRoutes and their associated Gateway Listeners can mutually specify hostnames which lets Cluster Operators and Application Developers collaborate to control which hostnames are used. I understand that hostnames would be unavailable in the actual routing path for TCPRoutes and UDPRoutes. However, it could be used in Listener attachment and/or DNS management (the part I care about).

External-DNS does all the things you would hope it would do to determine the correct hostnames for HTTP, TLS, and GRPC Routes. For TCP and UDP Routes it reuses the existing [external-dns.alpha.kubernetes.io/hostname](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/annotations/annotations.md#external-dnsalphakubernetesiohostname) annotation on the Routes without any control or validation from the Gateway. The caveat that makes this somewhat more tenable is that external-dns provides a lot of ways to filter (e.g. explicit lists or regular expressions) which domains are allowed and/or blocked.

As far as which WG owns or has claimed this as in their scope... I'm not on any WGs. I'm just some guy who wanted DNS automation for Gateway API to exist, so I implemented it in external-dns three years ago, and I've tried to keep it up to date since then. There was an [issue](https://github.com/kubernetes-sigs/external-dns/issues/2045) to add support for it, which I think was created by a member of GWAPI. I wrote the code then presented my design with a [detailed doc](https://docs.google.com/document/d/1o92u-OkoI0xWJwlaA5Rz9WqQ972HPzlPfEXpULKNx5k/edit#heading=h.v4uycgbxt1r0) at a GWAPI meeting, which went very smoothly. Then it was a pretty big pain in the ass to get the [pull request](https://github.com/kubernetes-sigs/external-dns/pull/2292) reviewed and merged. For my efforts I've mostly just gotten complaints from strangers on the internet who think external-dns is messed up, when really their GWs and Routes are misconfigured because they don't understand GWAPI concepts like the security model and cross-namespace route binding.

robscott:
>To better support DNS use cases, I'd be happy to see a change in Gateway API to allow hostnames in Routes and Listeners that don't need hostnames (i.e. move Hostnames to CommonRouteSpec)

@abursavich Listener Hostname is already optional:

https://github.com/kubernetes-sigs/gateway-api/blob/4cd0ab41932e776a77b1b40decae1a07998c2620/apis/v1/gateway_types.go#L270-L271

We do have a requirement that the combination of hostname, protocol, and port is unique among Listeners, is that what you'd like to remove?

abursavich:
> We do have a requirement that the combination of hostname, protocol, and port is unique among Listeners, is that what you'd like to remove?

@robscott,  my suggestion is to _add_ hostnames to Routes that don't directly use them: TCPRoute and UDPRoute. I can understand why you might not want to do that. It can't be enforced in the data path and would probably confuse some people, but it would be helpful for DNS.

youngnick:
One thing that we probably should make more clear is that, for TCPRoute and UDPRoute, you can't have more than one of the type attached to any specific Listener. So adding a `hostname` to the Routes probably wouldn't do much - for both of those Kinds you need to look at the Listener `hostname` anyway.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
+# GEP-2627: DNS configuration within Gateway API
+
+* Issue: [#2627](https://github.com/kubernetes-sigs/gateway-api/issues/2627)
+* Status: Provisional
+
+## TLDR
+
+For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards.  The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API.   Instead of leaving this as an exercise for the user to figure out, this proposal attempts to provide options to ease Gateway API operations.
```
shaneutt, line 8: I think this is great, but I would maybe consider a slightly different language at the end, if you're comfortable with it:

```suggestion
For gateway infrastructure to be valuable we need to be able to connect clients to these gateways. A common way to achieve this is to use domain names/hostnames and DNS. Gateways define listeners that can have assigned hostnames or wildcards. The guidelines for DNS configuration are a critical piece of service networking, but this is currently not expressible as part of Gateway API. Instead of leaving this unspecified and having implementations do this in potentially wildly different ways, the purpose of this proposal is to provide a standard way to specify DNS for Gateways.
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Goals
+* Provide a way for Gateway API resource owners to mark their resources as relevant for external DNS provisioning
```
shaneutt, line 11: I anticipate internal DNS provisioning being important as well. Here's another wording, for your considerations:

```suggestion
* Provide DNS specification for Gateway resources
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Ensure that the above method has a way for multiple providers to be present in the cluster and be able to actuate external DNS provisioning requests
```
shaneutt, line 12: What do you think about:

```suggestion
* Support multiple DNS providers and a selection mechanism for Gateways
```

?

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Ensure that any method is based on structured fields and makes the most of `status` on whatever resources are relevant, whether they are existing Gateway API resources or new resources.
```
shaneutt, line 13: I think this collides a bit with the first goal, perhaps:

```suggestion
* Provide Gateway status to communicate the state of provisioned DNS
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
```
shaneutt, line 14: To me this suggests to me some desire to go beyond just adding specification and status for which users can request DNS configuration for a `Gateway`, and the implementation can go deploy that and report status. This seems to suggest some greater involvement in the actual provisioning tooling, is that what was intended?

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Clarity on the scope of hostnames under management. (This should _not_ be able to be used to affect the standard in-cluster DNS configuration)
```
shaneutt, line 15: I was wondering if you might be able to elaborate on this goal a bit further please?

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Non-Goals
+
+* Anything to do with configuring in-cluster DNS. This support is for configuration outside the cluster only.
```
shaneutt, line 19: I'm not opposed to this as a non-goal, except I was wondering if you would please elaborate a bit more on what's driving you to leave it out?

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Ways to define if the Gateway API resources are allowed to request particular hostnames. These choices should be left to the implementations that actually actuate the requests for hostnames. However, `status` flows should be specified so as to make clear if a hostname provisioning request cannot be performed.
```
shaneutt, line 20:
```suggestion
* Providing any upstream hostname validation mechanisms. We can provide status for validation failure, but implementations are responsible for validation.
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+* Cover more complex DNS routing strategies that come into play for multi-cluster topologies such as round robin, failover, health checks, weighted and geo location with this first pass. Supporting these types of use cases for distributed gateways (e.g., in different regions or multiple gateways for resilience within a region) and offering a form of global load balancing leveraging DNS is a potential future goal.
```
shaneutt, line 21:
```suggestion
* Multi-cluster DNS for multi-cluster ingress solutions
```

geps/gep-2627/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+As a cluster administrator, I would like to have the DNS names automatically populated into my specified DNS zones as a set of records based on the assigned addresses of my gateways so that I do not have to undertake external automation or management of this essential task.
+
+As a cluster administrator I would have the status of the DNS records reported back to me, so that I can leverage existing kube based monitoring tools to know the status of the integration.
+
+As a cluster administrator, I would like the DNS records to be updated automatically if the `spec` of assigned gateways changes, whether those changes are for IP address or hostname. 
+As a DNS administrator, I should be able to ensure that only approved External DNS controllers can make changes to DNS zone configuration. (This should in general be taken care of by DNS system <-> External DNS controller interactions like user credentials and operation status responses, but it is important to remember that it needs to happen).
```
shaneutt, line 32: We don't cover any application developer user stories here, we seem pretty focused on the cluster administrator. I anticipate that application developers will in fact will have relevant use cases. What are your thoughts on application developer roles needs here?

shaneutt (changes requested):
I'm sorry that this one ended up not getting attention in a more timely manner, I'm glad you bumped it.

This is great, I too anticipate needs here and I really would prefer we provide a more consistent and standard way to do this rather than having users have potentially very different ways to handle this from implementation to implementation.

Most of my comments are minor, albeit I do think we need to consider the [roles and personas](https://gateway-api.sigs.k8s.io/concepts/roles-and-personas/) at play here a bit more and potentially expand our user stories?

Let me know your thoughts :thinking:

geps/gep-2627/index.md
maleck13: The intention is to suggest that with this API I as the user can choose from any provider that supports this API similar to how Gateway API itself works with the `controllerName` in the GatewayClass. So if this API was implemented by `controller A` and I wanted to move to a new implementation in controller B that would be possible and not need a different API to achieve that

geps/gep-2627/index.md
maleck13: So it was a while since I looked at this. I am trying to recall. But I believe this was focused around only using the hosts defined in the gateway and not getting involved with or conflicting with the exsting service name dns IE `data.test.svc.cluster.local`

geps/gep-2627/index.md
maleck13: As mentioned above. This could prob be clarified more as it isn't that we can't handle host names and IPs that are not for external use but rather that we wont interfer with things like the service discovery and existing DNS infrastructure in k8s

geps/gep-2627/index.md
shaneutt: Is it OK to pull it, just so there's no ambiguity?

```suggestion
```

OR do you prefer we keep it (maybe a rewording)?

geps/gep-2627/index.md
shaneutt: I was considering this overnight and I'm actually starting to think that we should in fact remove this non-goal. My thinking is that since we're pretty much just a spec and status where implementations can use whatever integration (cloud or otherwise) they want to fulfill it and report the status, then in-cluster really shouldn't matter too much from our perspective (except that we _might_ want to provide spec delineation so that a user can request "this DNS name, ONLY intra-cluster", or something like that).

```suggestion
```

Can you be persuaded? Or did you have some strong feelings that it would be a detriment to include in-cluster DNS now?

geps/gep-2627/index.md
maleck13: I am ok with removing this 👍

geps/gep-2627/index.md
maleck13: Yes. I can be persuaded. Coming back to this, I think you are correct and this may have been added based on some questions around security. As you say, this is proposing a spec and defined status and as such can be implemented by multiple providers. So ultimately it is down to the cluster admin and the RBAC system to decide what implementation to use and who they want to allow to create this API. 
One thought though is around the target ref, I guess I had initially only thought of it as targeting a gateway initially are you thinking beyond that ? such as a service ?

geps/gep-2627/index.md
maleck13: So while I like the change here, I wanted a way to communicate that it isn't that it will never be a goal. I guess ultimately this means not as of this GEP but doesn't rule it out for a future GEP

geps/gep-2627/index.md
maleck13: I am struggling a little to think of specific application developer use cases around this. What comes to mind is allowing the DNSPolicy to potentially target a HTTPRoute or possibly some form of health check based on the available endpoints behind a service. So as an attempt at a use case "As an application developer, I do not want my service to be exposed via the Gateway until there are pods that are healthy and ready to serve the traffic" ? Did you have anything specific in mind?

geps/gep-2627/index.md
shaneutt: Well, I think we need to consider the user stories (see https://github.com/kubernetes-sigs/gateway-api/pull/2712#discussion_r1951304911) and decide who needs this functionality, and if its safe to provide them.

I think if you're OK with committing the suggestion above, we can consider this comment resolved and then can follow up in that other comment.

mloiseleur:
I'm discovering this PR today. It seems really interesting :+1: . 

What I can share is that there is clearly a community interest on using Kubernetes as a registry. 

I'm not sure if it can be considered, but I'll try. 

Public DNS is a more global subject than just Gateway API. For various reasons, there are many records pointing to other or non-Kubernetes resources. So my humble suggestion to this GEP would be to extend this Goal:

> * Provide DNS specification for Gateway resources

to something like:

> * Provide DNS specification for Gateway resources as first citizen
> * Provide DNS specification for other dynamic sources like Ingress, IngressRoute or even manually set values

That would allow `external-dns` (and others software implementing it) to use this new API as the de-facto standard for all DNS use cases, without needing to create an other CRD for other sources.

PS: For those interested, a first attemp on `external-dns` was discussed [here](https://github.com/kubernetes-sigs/external-dns/pull/4598). The PR was not reviewed as the author wanted, so he created a standalone software [phonebook](https://github.com/pier-oliviert/phonebook).

geps/gep-2627/index.md
maleck13: will mark as resolved as we haven't had any input here. Things like a default health check based on endpoints are nice to haves but not integral.

maleck13:
@shaneutt Have I addressed your requested changes? I am getting some more time to focus around this area and would like to find the right way to move it forward.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,52 @@
...
+## Goals
+* Provide DNS specification for Gateway resources
+* Support multiple DNS providers and a selection mechanism for Gateways
+* Provide Gateway status to communicate the state of provisioned DNS
+* Increase portability and supportability between Gateway API implementations and third party controllers offering DNS integration.
+
+## Non-Goals
+
+* Providing any upstream hostname validation mechanisms. We can provide status for validation failure, but implementations are responsible for validation.
+* Multi-cluster DNS for multi-cluster ingress solutions (at least not as part of the initial API)
```
shaneutt, line 19: Make sure to add a little context explaining _why_ these are non-goals.

geps/gep-2627/index.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Conformance Details
+
+TBD
+
+## Alternatives
+
+it is possible to use `external-dns` to manage dns based on HTTPRoutes and Gateways https://github.com/kubernetes-sigs/external-dns/blob/7f3c10d65297ec1c4bcc8dd6f88c189b7f3e80d0/docs/tutorials/gateway-api.md. The aim of this GEP is not remove this as an option, but instead provide a common API that could then be leveraged by something like external-dns. 
+
+The Kuadrant project (full disclosure I work on this project), offers a [DNSPolicy API](https://docs.kuadrant.io/1.2.x/kuadrant-operator/doc/reference/dnspolicy/#dnspolicy) which in part was the basis and inspiration for opening this GEP. The DNSPolicy offered by Kuadrant goes beyond what is outlined here as it also handles multi-cluster ingress and offers common routing options such as GEO and Weighted responses.
```
shaneutt: You don't need to make a disclaimer for this CNCF project :shrug: 

```suggestion
The Kuadrant project offers a [DNSPolicy API](https://docs.kuadrant.io/1.2.x/kuadrant-operator/doc/reference/dnspolicy/#dnspolicy) which in part was the basis and inspiration for opening this GEP. The DNSPolicy offered by Kuadrant goes beyond what is outlined here as it also handles multi-cluster ingress and offers common routing options such as GEO and Weighted responses. 
```

However this might make more sense under `## References` than under `## Alternatives`?

shaneutt (approved):
Seems like a good start for `Provisional`, just a couple small comments to sort out and then :+1: from me.

/cc @mlavacca @robscott @youngnick 

@maleck13 do note that we're in the [v1.4 release cycle](https://github.com/kubernetes-sigs/gateway-api/discussions/3824), and since this isn't one of the items for that release continued reviews for this will have a low priority (but we will try). Would definitely recommend prepping to submit this for `v1.5` afterwards, and see if you can gather supporters to help you move it forward!

geps/gep-2627/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: internal.gateway.networking.k8s.io/v1alpha1
+kind: GEPDetails
+number: 2627
+name: DNS configuration for Gateway API
+status: Provisional
+authors:
+  - maleck13
+
```
shaneutt:
```suggestion
```

mlavacca (commented):
Thanks for the work on this, @maleck13!

/lgtm

PR #2721, opened 2024-01-18
Loosening TLS validation to enable indirect TLS config

### LINKED ISSUE #763: Document Self Service TLS Certificate Approaches
As a follow up to the discussion in https://github.com/kubernetes-sigs/gateway-api/pull/749#discussion_r686410424, it would be good to clarify exactly how the API can be used to allow app owners to provide their own TLS certs. 

**Why this is needed**:
This is a common request, and the existing guidance in GEP #746 is pretty broad and may not scale as much as needed.

### LINKED ISSUE #2668: CertificateRefs should NOT be required when TLS mode is "Terminate"
**What happened**:
CEL validation currently requires that CertificateRefs are specified when TLS mode is set to "Terminate":

https://github.com/kubernetes-sigs/gateway-api/blob/195d9882f816c9002eede96b2eda171ac454a6c1/apis/v1/gateway_types.go#L378

This validation should be removed as implementations should be able to support other ways to configure TLS certificates. (One example would be [GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/secure-gateway#secure-using-ssl-certificate), but I don't think it is a secure practice to _require_ all TLS certificates to be configured within a cluster).

Thanks to @gauravkghildiyal for catching this one!

### LINKED ISSUE #2713: TLS Config Should be Optional
**What happened**:
Our CEL validation currently requires that TLS config is specified on a Listener when the protocol is `HTTPS` or `TLS`:

https://github.com/kubernetes-sigs/gateway-api/blob/9a663ba09604fd0bfd364d50b1c78bd87c2305b0/apis/v1/gateway_types.go#L189

Similar to #2668, I believe this validation is unnecessarily strict and should be reverted. 

As I've been thinking about more TLS user stories for Gateway API, particularly self service and generated certs (https://github.com/kubernetes-sigs/gateway-api/issues/763#issuecomment-1887895742), I've realized that the strictness of our validation here has made it unnecessarily complex to handle some very real use cases. I think it is critical for our API to be flexible enough here to enable different personas to configure TLS. 

When [I pulled TLS Certs out of Routes](https://github.com/kubernetes-sigs/gateway-api/pull/749), I still had every intention of enabling app owners to configure TLS Certs (thus #763 was created). Unfortunately I think we lost track of that goal when writing this specific validation. 

Note the this is closely tied to #2668 and if we get consensus around this approach as well, I'll work on a change that incorporates reverting both validations + adding documentation and guidance for how implementations should handle the absence of these fields.

robscott:
**What type of PR is this?**
/kind documentation
/kind feature

This loosens TLS validation as requested by #2668 and #2713 and adds documentation for the self-service approaches and overall indirect TLS config this enables. This also adds a new `InvalidTLSConfig` reason for the `Programmed` condition on Gateways.

Fixes #763 
Fixes #2668
Fixes #2713

```release-note
TLS Configuration is no longer required on Gateway Listeners to enable more flexible TLS configuration.
```

This is a big PR, going to leave it on hold for a bit while we wait for consensus.

### DIFF
--- a/apis/v1/gateway_types.go
+++ b/apis/v1/gateway_types.go
@@ -186,7 +186,6 @@ type GatewaySpec struct {
 	// +listMapKey=name
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=64
-	// +kubebuilder:validation:XValidation:message="tls must be specified for protocols ['HTTPS', 'TLS']",rule="self.all(l, l.protocol in ['HTTPS', 'TLS'] ? has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls must not be specified for protocols ['HTTP', 'TCP', 'UDP']",rule="self.all(l, l.protocol in ['HTTP', 'TCP', 'UDP'] ? !has(l.tls) : true)"
 	// +kubebuilder:validation:XValidation:message="tls mode must be Terminate for protocol HTTPS",rule="self.all(l, (l.protocol == 'HTTPS' && has(l.tls)) ? (l.tls.mode == '' || l.tls.mode == 'Terminate') : true)"
 	// +kubebuilder:validation:XValidation:message="hostname must not be specified for protocols ['TCP', 'UDP']",rule="self.all(l, l.protocol in ['TCP', 'UDP']  ? (!has(l.hostname) || l.hostname == '') : true)"
@@ -375,19 +374,18 @@ const (
 )
 
 // GatewayTLSConfig describes a TLS configuration.
-//
-// +kubebuilder:validation:XValidation:message="certificateRefs must be specified when TLSModeType is Terminate",rule="self.mode == 'Terminate' ? size(self.certificateRefs) > 0 : true"
 type GatewayTLSConfig struct {
 	// Mode defines the TLS behavior for the TLS session initiated by the client.
 	// There are two possible modes:
 	//
-	// - Terminate: The TLS session between the downstream client
-	//   and the Gateway is terminated at the Gateway. This mode requires
-	//   certificateRefs to be set and contain at least one element.
+	// - Terminate: The TLS session between the downstream client and the
+	//   Gateway is terminated at the Gateway. This mode requires certificates
+	//   to be specified in some way, such as populating the certificateRefs
+	//   field.
 	// - Passthrough: The TLS session is NOT terminated by the Gateway. This
 	//   implies that the Gateway can't decipher the TLS stream except for
-	//   the ClientHello message of the TLS protocol.
-	//   CertificateRefs field is ignored in this mode.
+	//   the ClientHello message of the TLS protocol. The certificateRefs field
+	//   is ignored in this mode.
 	//
 	// Support: Core
 	//
@@ -701,8 +699,10 @@ const (
 	// true.
 	GatewayReasonProgrammed GatewayConditionReason = "Programmed"
 
-	// This reason is used with the "Programmed" and "Accepted" conditions when the Gateway is
-	// syntactically or semantically invalid.
+	// This reason is used with the "Programmed" and "Accepted" conditions when
+	// the Gateway is syntactically or semantically invalid. For example, this
+	// could include unspecified TLS configuration, or some unrecognized or
+	// invalid values in the TLS configuration.
 	GatewayReasonInvalid GatewayConditionReason = "Invalid"
 
 	// This reason is used with the "Programmed" condition when the
--- a/apis/v1/gateway_types.go
+++ b/apis/v1/gateway_types.go
@@ -374,6 +374,8 @@ const (
 )
 
 // GatewayTLSConfig describes a TLS configuration.
+//
+// +kubebuilder:validation:XValidation:message="certificateRefs or options must be specified when mode is Terminate",rule="self.mode == 'Terminate' ? size(self.certificateRefs) > 0 || size(self.options) > 0 : true"
 type GatewayTLSConfig struct {
 	// Mode defines the TLS behavior for the TLS session initiated by the client.
 	// There are two possible modes:
--- a/site-src/guides/implementers.md
+++ b/site-src/guides/implementers.md
@@ -211,7 +211,7 @@ certificate stored by the external `vendor.example.com` TLS Certificate
 provider.
 
 #### 2. Automatically generated TLS certs that are populated later
-Many users would prefer that TLS certs were automatically generated on their
+Many users would prefer that TLS certs will be automatically generated on their
 behalf. One potential implementation of that would involve a controller that
 watches Gateways and HTTPRoutes, generates TLS certs, and attaches them to the
 Gateway. Depending on the implementation details, Gateway owners may need to
--- a/site-src/guides/implementers.md
+++ b/site-src/guides/implementers.md
@@ -86,7 +86,7 @@ For information on how to manage Gateway API CRDs, including when it is
 acceptable to bundle CRD installation with your implementation, refer to our
 [CRD Management Guide](/guides/crd-management).
 
-### Conformance and Version compatibility
+### Conformance and Version Compatibility
 
 A conformant Gateway API implementation is one that passes the conformance tests
 that are included in each Gateway API bundle version release.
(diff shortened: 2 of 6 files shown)


apis/v1/gateway_types.go
```diff
@@ -701,10 +698,16 @@ const (
...
-	// syntactically or semantically invalid.
+	// This reason is used with the "Programmed" and "Accepted" conditions when
+	// the Gateway is syntactically or semantically invalid.
 	GatewayReasonInvalid GatewayConditionReason = "Invalid"
 
+	// This reason is used with the "Programmed" condition when the Gateway is
+	// not programmed because the TLS config is invalid. This could include
+	// unspecified TLS configuration, or some unrecognized or invalid values in
+	// the TLS configuration.
+	GatewayReasonInvalidTLSConfig GatewayConditionReason = "InvalidTLSConfig"
```
EItanya, line 709: This feels a bit too similar to `Invalid` to warrant its own `Reason`. `InvalidTLSConfig` feels like a string details extension of `Invalid`, rather than its own `Reason`. This is definitely more of a nit that anything, but creating a new Reason which is a superset of another `Reason` could lead to `Reason` bloat

site-src/guides/implementers.md
```diff
@@ -159,16 +159,118 @@ the `metadata.generation` field of the object at the time the status is generate
...
+      options:
+        vendor.example.com/certificate-name: store-example-com
+```
+
+In this example, the `store-example-com` name would refer to the name of a
+certificate stored by the external `vendor.example.com` TLS Certificate
+provider.
+
+#### 2. Automatically generated TLS certs that are populated later
+Many users would prefer that TLS certs were automatically generated on their
```
EItanya, line 212:
```suggestion
Many users would prefer that TLS certs be automatically generated on their
```

site-src/guides/implementers.md
```diff
@@ -159,16 +159,118 @@ the `metadata.generation` field of the object at the time the status is generate
...
+#### 3. Certs that are specified by other personas
+In some organizations, Application Developers are responsible for managing TLS
+Certificates (see [Roles and Personas](/concepts/roles-and-personas) for more on
+this and other roles).
+
+To enable this use case, a new controller and CRD would be created. This
+CRD would link hostnames to user-provided certs, and then the controller would
+populate the certs specified by that CRD on Gateway listeners that matched those
+hostnames. This would also likely benefit from a Listener or Gateway-level
+opt-in for the behavior.
```
EItanya, line 241: Would an option be a bad example of a way to offer opt-in behavior here? It would definitely be a bit of an abuse of the `map[string]string` structure, but would have the benefit of not requiring API changes.


For example:

`example.com/tls-allowed-namespace-selector: foo=bar`

EItanya (commented):
Looks great overall, I'm not super familiar with CEL so I'll just comment on the new intended behavior WRT documentation

sunjayBhatia (commented):
worth updating this page with some content as well? https://gateway-api.sigs.k8s.io/guides/tls

at least alluding to the fact that some implementations may decide to implement downstream TLS, specifying certs, other options, etc. in different ways

it strikes me that making the `tls` section no longer required for some protocols, it could make the table here confusing: https://gateway-api.sigs.k8s.io/guides/tls/#clientserver-and-tls

apis/v1/gateway_types.go
robscott: Good call, decided to drop this entirely.

site-src/guides/implementers.md
robscott: For some reason "be" doesn't feel quite right to me here, but I'm not exactly a grammar expert. Maybe there's a better way to phrase the whole sentence?

robscott:
>worth updating this page with some content as well? https://gateway-api.sigs.k8s.io/guides/tls

>at least alluding to the fact that some implementations may decide to implement downstream TLS, specifying certs, other options, etc. in different ways

Good point, added a note to start that page off.

site-src/guides/implementers.md
robscott: I'm generally against any form of structured content within these kind of annotation-like fields since it's so easy to mess up. Maybe I should add some guidelines for what doesn't belong here? Was that what you were suggesting here?

site-src/guides/implementers.md
```diff
@@ -159,16 +159,118 @@ the `metadata.generation` field of the object at the time the status is generate
...
+
+Requests SHOULD match at most one Listener. For example, if Listeners are
+defined for "foo.example.com" and "*.example.com", a request to
+"foo.example.com" SHOULD only be routed using routes attached to the
+"foo.example.com" Listener (and not the "*.example.com" Listener).
+
+Implementations that do not support Listener Isolation MUST clearly document
+this. In the future, we plan to add Listener Isolation conformance tests to
+ensure this behavior is consistent across implementations that claim support for
+the feature.
```
shaneutt, line 182: This text will likely bitrot, it might be helpful to link the relevant issues here, and then in the issues add a TODO list item to update the documentation here.

apis/v1/gateway_types.go
```diff
@@ -375,8 +374,6 @@ const (
 )
 
 // GatewayTLSConfig describes a TLS configuration.
-//
-// +kubebuilder:validation:XValidation:message="certificateRefs must be specified when TLSModeType is Terminate",rule="self.mode == 'Terminate' ? size(self.certificateRefs) > 0 : true"
```
frankbu, line 379: Is entirely removing this validation necessary? Maybe, instead of removing, it could be changed to at least one of either `certificateRefs` or `options` must be specified when mode is `Terminate`?

frankbu: @robscott note that the comment on line 385 (below) also needs to be updated

site-src/guides/implementers.md
robscott: Good call. Created https://github.com/kubernetes-sigs/gateway-api/issues/2803 and linked to it here.

apis/v1/gateway_types.go
robscott: >@robscott note that the comment on line 385 (below) also needs to be updated\

Thanks @frankbu! I updated the relevant comments in that area.

> Is entirely removing this validation necessary? Maybe, instead of removing, it could be changed to at least one of either certificateRefs or options must be specified when mode is Terminate?

Unfortunately I don't think that would work for one of the initial user stories that led to this - specifically a user wanting certs to be added automatically after they created a Gateway (https://github.com/kubernetes-sigs/gateway-api/issues/763#issuecomment-1887895742). I think the most likely way that would work would involve some kind of controller that automatically generates and populates certificates for Gateways in a given set of namespaces or maybe with a predefined label or annotation. I'm assuming that in most cases, this kind of config would apply to an entire Gateway, not just a listener, so TLS options wouldn't really fit.

frankbu: I was thinking that case could be handled by not having any user GatewayTLSConfig setting at all. In that case, the validation only applies when the user is explicitly setting the mode to Terminate using either certificateRefs or some other impl-specific option.

My concern is that if we remove all validation now, we won't be able to add it later. Alternatively, we can always take away more validation in the future without breaking anybody.

site-src/guides/implementers.md
youngnick: It feels a little more readable to me to use `will be` in place of `be`.

But, I think it depends on where we're talking about this behavior happening.

- `were`, in the past
- `are` in the present
- `will be` in the future.

Personally I think that `will be` is probably what we're going for.

apis/v1/gateway_types.go
robscott: Chatted on Slack about this and although there are some use cases that could benefit from setting mode and nothing else, they're not sufficient to justify dropping this validation altogether at this point. If we need to, we can revisit this in the future, but we can never retroactively tighten validation. Have updated this to require "options" or "certificateRefs" when mode is "Terminate" as @frankbu suggested.

site-src/guides/implementers.md
robscott: Thanks! Updated to "will be".

frankbu (approved):
Thanks @robscott! LGTM

shaneutt (approved):
/approve
/lgtm

PR #2741, opened 2024-01-26
Move BackendTLSPolicy WellKnownCACerts field to implementation-specific

### LINKED ISSUE #2726: BackendTLSPolicy wellKnownCACerts field "System" value should not be core
We're working on implementing BackendTLSPolicy in Contour and are a little confused about the `wellKnownCACerts` field.

It is marked as `Support: Core for “System”` in the spec but in the [GEP](https://gateway-api.sigs.k8s.io/geps/gep-1897/#api) the usage of system certificates is documented as implementation-specific: `The use and definition of system certificates is implementation-dependent, and the intent is that these certificates are obtained from the underlying operating system.`

It looks like the spec was updated here: https://github.com/kubernetes-sigs/gateway-api/pull/2448#discussion_r1342170962

While I don't disagree that implementations would not be able to support some form of fetching system certs to use, I would still argue the support for this field should still be implementation-specific.

If a field is marked core I would expect a conformance test to exist for it eventually and that all implementations should to be able to pass a test verifying the behavior it introduces, regardless of how exactly they make it work. I'm not sure how a conformance test for this would work against all implementations.
- Would this mean that we assert that a TLS connection to a backend app is verified on the Gateway implementation end of the connection with a specific root CA cert from the system?
- What happens if say the test expects a server certificate to be verified but an implementation has a newer/older version of an OS package providing the root certs?
- If an implementation decides to explicitly use a container image with no filesystem/minimal OS packages, how would this work?

This seems like this test is going to start mandating a specific runtime environment for implementations, which I don't think is desirable.

sunjayBhatia:
**What type of PR is this?**

/kind cleanup
/kind documentation

Changes WellKnownCACerts field of BackendTLSPolicy to Implementation-specific support

Also updates language around what to do if WellKnownCACerts is not supported

Fixes #2726

```release-note
BackendTLSPolicy WellKnownCACerts field has been updated to implementation-specific support
```

### DIFF
--- a/apis/v1alpha2/backendtlspolicy_types.go
+++ b/apis/v1alpha2/backendtlspolicy_types.go
@@ -85,7 +85,8 @@ type BackendTLSPolicyConfig struct {
 	// If CACertRefs is empty or unspecified, then WellKnownCACerts must be
 	// specified. Only one of CACertRefs or WellKnownCACerts may be specified,
 	// not both. If CACertRefs is empty or unspecified, the configuration for
-	// WellKnownCACerts MUST be honored instead.
+	// WellKnownCACerts MUST be honored instead if supported by the
+	// implementation.
 	//
 	// References to a resource in a different namespace are invalid for the
 	// moment, although we will revisit this in the future.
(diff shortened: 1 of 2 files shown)


youngnick:
This is a good change, thanks @sunjayBhatia.

/lgtm

robscott:
Not sure how I missed this one, thanks @sunjayBhatia!

/approve

PR #2754, opened 2024-02-05
add github workflow to check supportedFeatures freshness

LiorLieberman:
**What type of PR is this?**

/kind test
/area conformance

Kubebuilder validation gets out of date is someone is adding a supportedFeature. 
When a gateway controller publish supportedFeatures into the GWC, the validation blocks it if some features are not in that list.

Right now, suite.AllFeatures contains features that are not in the validation.

This enables the implementation of #2163 in gateway controllers by fixing the discrepancy between the validation list and the actual features.
It also keep freshness of this kubebuilder validation list

```release-note
NONE
```


.github/workflows/supported_features.yml
```diff
@@ -0,0 +1,26 @@
...
+          GREEN='\033[0;32m'
+          NC='\033[0m'
+          current_features=$(grep -B 1 'type SupportedFeature string' apis/v1/gatewayclass_types.go | grep -oP '// \+kubebuilder:validation:Enum=\K.*') 
+          expected_features=$(go run hack/supportedfeatures/main.go)
+          if [[ "$current_features" != "$expected_features" ]]; then
+            echo "found diff between the current and the expected supportedFeatures"
+            echo -e "expected: ${GREEN}${expected_features}${NC}"
+            echo -e "current: ${RED}${current_features}${NC}"
+            exit 1
+          fi
```
levikobi, line 26: Maybe after logging the expected and current, you can log a call to action, something like: "Please update kubebuilder validation list (...)" or ask to run a script that makes the update automagically (like a simpler version of hack/update-codegen.sh)

levikobi (commented):
I like the idea of this change, it's definitely going to remove the issue of the validation getting out of sync. I just have a small suggestion

.github/workflows/supported_features.yml
LiorLieberman: good call. I added a call to action.
Lets add some makefile command in a followup PR

SG?

LiorLieberman: I ended up adding this update script

robscott:
Thanks @LiorLieberman!

/lgtm
/approve

PR #2758, opened 2024-02-07
remove GWC status kubebuilder validation for supportedFeatures

LiorLieberman:
**What type of PR is this?**

/kind cleanup

Having this validation creates an unsafe situation;
If a user have a newer implementation than its installed CRD, the controller will not be able to accept/reconcile the GatewayClass. It will try to publish the supportedFeatures to the status but will get rejected because the installed, older, CRD might not have some newer features which the newer controller claims to support.

One way to overcome this would be to for implementations to maintain a mapping between supported features to GWC version. However, implementing this approach would require significant maintenance efforts across all implementations without equivalent added value.

Bottom line, its a status field, I dont see any reason to validate. Especially when implementation will move to run conformance tests from this field - the risk of publishing random strings would be even lower.

```release-note
NONE
```

/cc @robscott @howardjohn @levikobi @shaneutt @arkodg


arkodg (approved):
/approve

removing this allows implementations to publish additional features that may be supported as extensions

robscott:
Thanks @LiorLieberman!

/lgtm
/approve

LiorLieberman:
> /approve
> 
> removing this allows implementations to publish additional features that may be supported as extensions

LiorLieberman:
> removing this allows implementations to publish additional features that may be supported as extensions

@arkodg FWIW this field is not intended to publish those features. The supportedFeatures has a clear meaning and guidance what should be published. 

If you are interested in policy/extension representation, I have once opened a [gh discussion](https://github.com/kubernetes-sigs/gateway-api/discussions/2118) about it (but I know there were many other threads).  So feel free to loop in and bring it in the community meetings/slack.

PR #2766, opened 2024-02-12
Add contour conformance report for v1.0.0

sunjayBhatia:
**What type of PR is this?**

/area conformance

Contour v1.28.1 supports Gateway API v1.0.0, add conformance report for this version.

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/reports/v1.0.0/projectcontour-contour.yaml
+++ b/conformance/reports/v1.0.0/projectcontour-contour.yaml
@@ -1,0 +1,50 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
+date: "2024-02-12T16:01:26Z"
+gatewayAPIVersion: v1.0.0
+implementation:
+  contact:
+  - '@projectcontour/maintainers'
+  organization: projectcontour
+  project: contour
+  url: https://projectcontour.io/
+  version: v1.28.0
+kind: ConformanceReport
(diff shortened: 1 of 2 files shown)


skriss (approved):

robscott:
Thanks @sunjayBhatia! This LGTM, but I'm not sure where we are in the transition between report structures here.

@mlavacca @shaneutt Are we holding off on using the new reporting structure until the v1.1 release?

conformance/reports/v1.0.0/projectcontour-contour.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: gateway.networking.k8s.io/v1alpha1
```
robscott, line 1: Recommend updating implementations page to link to this report as well (https://gateway-api.sigs.k8s.io/implementations/#contour)

robscott (commented):
I think both @shaneutt and @mlavacca are OOO, and since no other reports have migrated to the new structure yet, I think it's fine to merge this as is. Just have one non-blocking suggestion.

/approve

conformance/reports/v1.0.0/projectcontour-contour.yaml
sunjayBhatia: good catch, done!

PR #2772, opened 2024-02-14
Add route port attachment doc

frankbu:
**What type of PR is this?**

/kind documentation

Add documentation for route port attachment needed to promote GEP-957 to standard.

Fixes #

```release-note
NONE
```

### DIFF
--- a/site-src/api-types/httproute.md
+++ b/site-src/api-types/httproute.md
@@ -83,7 +83,7 @@ spec:
     port: 8080
 ```
 
-Binding to a port also llows you to attach to multiple listeners at once.
+Binding to a port also allows you to attach to multiple listeners at once.
 For example, binding to port `8090` of the `acme-lb` Gateway would be more
 convenient than binding to the corresponding listeners by name:
 
--- a/site-src/api-types/httproute.md
+++ b/site-src/api-types/httproute.md
@@ -43,6 +43,64 @@ spec:
 Note that the target Gateway needs to allow HTTPRoutes from the route's
 namespace to be attached for the attachment to be successful.
 
+You can also attach routes to specific sections of the parent resource.
+For example, let's say that the `acme-lb` Gateway includes the following
+listeners:
+
+```yaml
+  listeners:
+  - name: foo
+    protocol: HTTP
+    port: 8080
+    ...
+  - name: bar
+    protocol: HTTP
+    port: 8090
+    ...
+  - name: baz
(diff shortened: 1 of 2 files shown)


site-src/api-types/httproute.md
```diff
@@ -43,6 +43,64 @@ spec:
...
+    sectionName: bar
+  - name: acme-lb
+    sectionName: baz
+```
+
+However, when binding Routes by port number, Gateway admins will no longer have
+the flexibility to switch ports on the Gateway without also updating the Routes.
+The approach should only be used when a Route should apply to a specific port
+number as opposed to listeners whose ports may be changed.
+
```
robscott, line 103: I'm not sure if we _need_ to cover this, but there are a couple more details that could potentially be useful:

1. You can set both port and sectionName at the same time
2. If you set port or sectionName for a reference to a parent, any other references to the same parent also need to specify the same field(s) (port and/or sectionName)

site-src/api-types/httproute.md
```diff
@@ -43,6 +43,64 @@ spec:
 Note that the target Gateway needs to allow HTTPRoutes from the route's
 namespace to be attached for the attachment to be successful.
 
+You can also attach routes to specific sections of the parent resource.
```
robscott, line 46: Not a blocker or required by any means, but it _might_ be helpful to have some section headings somewhere in here to make it a bit easier to link to a specific section. Not quite necessary yet, but if we end up adding more content here in response to the rest of this review, it might become more helpful.

robscott (commented):
Thanks @frankbu! This mostly LGTM, just a few nits.

site-src/api-types/httproute.md
frankbu: I specifically didn't mention these thing here, because 1) I didn't think it is helpful to get into to much of a deep-dive in this section, and 2) I thought that the API reference section did explain that all quite well.

If you feel strongly, I could add something here, maybe just something very short and sweet with a reference to the API docs for more details?

Let me know what you think.

site-src/api-types/httproute.md
frankbu: I could add a couple of subsections for sectionName and port attachment, but I agree that it might not be very helpful or needed unless the section gets longer.

robscott:
Thanks for writing this up @frankbu!

/lgtm
/approve

PR #2775, opened 2024-02-15
Add route port attachment conformance test

frankbu:
**What type of PR is this?**

/kind test
/area conformance

Conformance test needed to promote GEP-957 to standard.

Tested and passes with Istio implementation.

Fixes #

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-listener-port-matching.go
+++ b/conformance/tests/httproute-listener-port-matching.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"strings"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/gateway-api/conformance/utils/http"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteListenerPortMatching)
+}
+
+var HTTPRouteListenerPortMatching = suite.ConformanceTest{
+	ShortName:   "HTTPRouteListenerPortMatching",
+	Description: "Multiple HTTP listeners with different ports, each with a different HTTPRoute",
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportHTTPRoute,
+		suite.SupportHTTPRouteParentRefPort,
+	},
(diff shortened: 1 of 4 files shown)


conformance/tests/httproute-listener-port-matching.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: httproute-listener-port-matching
```
robscott, line 4: Any chance we can just add more listeners to one of the base Gateways instead of creating a new one?

conformance/tests/httproute-listener-port-matching.go
```diff
@@ -0,0 +1,107 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
robscott, line 2:
```suggestion
Copyright 2024 The Kubernetes Authors.
```

robscott (commented):
Thanks @frankbu! This mostly LGTM, just a couple nits.

robscott:
Since it looks like Istio is already passing these tests, we need a couple more implementations to also pass these tests to consider this feature ready to graduate to standard channel. Can a few others test this out?

/cc @arkodg @pleshakov @sunjayBhatia

conformance/tests/httproute-listener-port-matching.yaml
frankbu: I'm not sure it's a good idea to complicate the base gateways for this optional functionality that requires several listeners with a strange combination of ports and hostnames to test effectively. I followed the same pattern used in https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-listener-hostname-matching.yaml which also uses its own gateway. Maybe we could come up with some base gateway with lots of listeners that we could use for both tests, but not sure it's worth it.

sunjayBhatia (approved):
test passes with Contour 👍🏽

robscott:
So far we've gotten confirmation that these tests pass on Istio and Contour. We just need one more implementation to confirm that these tests pass for them for this feature to graduate to standard.

@arkodg, @mlavacca, or @pleshakov can any of you confirm if these tests pass on your implementation(s)?

conformance/tests/httproute-listener-port-matching.go
```diff
@@ -0,0 +1,105 @@
...
+	ConformanceTests = append(ConformanceTests, HTTPRouteListenerPortMatching)
+}
+
+var HTTPRouteListenerPortMatching = suite.ConformanceTest{
+	ShortName:   "HTTPRouteListenerPortMatching",
+	Description: "Multiple HTTP listeners with different ports, each with a different HTTPRoute",
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportHTTPRoute,
+		suite.SupportHTTPRouteParentRefPort,
```
sayboras, line 40: do we need SupportGatewayPort8080 here ?

youngnick:
Updates #957

conformance/tests/httproute-listener-port-matching.go
frankbu: Good question. It also needs to support port 8090 for this test. Maybe SupportHTTPRouteParentRefPort implicitly means supports arbitrary ports?

cc @robscott

frankbu:
Test passes with Envoy Gateway! 😃 

cc @arkodg @robscott

robscott:
Thanks @frankbu! Do you mind making a follow up PR to formally graduate this GEP to standard and remove the `experimental` tags around the fields?

/lgtm

frankbu:
> Thanks @frankbu! Do you mind making a follow up PR to formally graduate this GEP to standard and remove the experimental tags around the fields?

Already done: #2777

PR #2777, opened 2024-02-16
Promote GEP-957 to standard channel

### LINKED ISSUE #957: GEP: Route Port Matching

frankbu:
**What type of PR is this?**

/kind gep

GEP-957 should be moved from experimental to standard channel.

Depends on: #2775 

Fixes #957

```release-note
Attaching routes to gateways by port number (GEP-957) feature moved to standard channel.
```

### DIFF
--- a/apis/v1/shared_types.go
+++ b/apis/v1/shared_types.go
@@ -150,7 +150,6 @@ type ParentReference struct {
 	// Support: Extended
 	//
 	// +optional
-	// <gateway:experimental>
 	Port *PortNumber `json:"port,omitempty"`
 }
 
--- a/config/crd/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
+++ b/config/crd/experimental/gateway.networking.k8s.io_backendtlspolicies.yaml
@@ -387,7 +387,7 @@ spec:
                           pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
                           type: string
                         port:
-                          description: |+
+                          description: |-
                             Port is the network port this Route targets. It can be interpreted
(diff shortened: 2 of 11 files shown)


robscott:
Thanks @frankbu! With #2775 merged, this is good to go. Will defer to someone else for final LGTM.

/approve

kflynn:
Hit it! 🙂 
/approve

mlavacca (commented):
/lgtm

PR #2786, opened 2024-02-20
Enable dependabot for `/gwctl` and `/conformance/echo-basic/`

### LINKED ISSUE #2780: Dependabot no longer working well for go dependencies
**What happened**:
PRs opened by dependabot do not appear to be running `go mod tidy` and are failing our presubmits.

PR opened by dependabot to pass our presubmits.

Not sure what the solution is here, maybe it's just better dependabot config, or maybe we should look into an alternative like renovate that [appears to have a few more configuration options here](https://docs.renovatebot.com/configuration-options/#postupdateoptions). With that said, I can't find any prior art for running renovate with k8s projects, so it's probably safest to stick with dependabot if we can get it to work.

jongwooo:
**What type of PR is this?**

/kind bug

Enable dependabot for `/gwctl` and `/conformance/echo-basic/`.

/cc @youngnick @robscott 

Fixes #2780 

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,34 @@ updates:
       - dependencies
       - ok-to-test
 
+  # Dependencies listed in gwctl/go.mod
+  - package-ecosystem: "gomod"
+    directory: "/gwctl" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    groups:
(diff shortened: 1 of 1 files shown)


robscott:
Thanks @jongwooo!

/lgtm
/approve

Issue #2550, opened 2023-10-31
Display for conformance profiles results


xtineskim:
As discussed in the Gateway API meeting, post Kubecon NA 2023, there will be an implementations reorganization by profiles. Since it seems open ended, wanted to open an issue to see what folks would like to be captured.
Data displayed will be pulled from the [conformance reports](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance/reports), how to display it meaningfully for an end-user trying to decide on which implementation suits their requirements 

Some things that come to mind:

- Table that can be filtered by certain features (for ex  version 0.8.0, `HTTPRouteMethodMatching`), category of implementations (L7, L4, Mesh), maybe stats of passing/failings/skipped tests?
- Page headers based on use cases that a user can jump to to see implementations (use cases as in the supported features)
- All implementations listed out, and having filters for selecting use case scenarios (similar to online shopping haha)

Some questions:
- what else from the conformance reports should be pulled and displayed? 
- there seemed to be some small discussions on other displays besides badges for implementation projects? 

This is open for suggestions, please leave any thoughts/questions/feedback

Related issues:
https://github.com/kubernetes-sigs/gateway-api/issues/1709

Related discussion: 
https://github.com/kubernetes-sigs/gateway-api/discussions/2577

xtineskim:
seeing as there are no strong opinions from the meetings/on this issue, will work on a poc and present to the group  👍

shaneutt:
Thank you for taking this one on! Let us know if you run into any trouble, or need any help! :vulcan_salute:

xtineskim:
thanks @shaneutt! currently been exploring the use of `rmarkdown`. Not sure if other k8s sites use it, but the ability to plot make more "dynamic" code would be great - do you know if maintainers would be opposed to using a flavour of markdown?( ideally I don't want to overly complicate the build for the site)

shaneutt:
I for one am not opposed, my co-maintainers @youngnick and @robscott can chime in if they have thoughts :thought_balloon:

robscott:
Hey @xtineskim, I've never seen k8s use it, but that looks promising to me, worth a shot.

mikemorris:
> currently been exploring the use of `rmarkdown`

Would this necessitate switching to a different static site generator, or would it be possible to integrate with the existing [MkDocs](https://github.com/kubernetes-sigs/gateway-api/blob/main/mkdocs.yml) configuration somehow?

[...]

PR #2793, opened 2024-02-21
Adds HAProxy Kubernetes Ingress Controller to Gateway API Implementat…

### LINKED ISSUE #2779: Add HAProxy Kubernetes Ingress Controller to Gateway API Implementation page

On https://gateway-api.sigs.k8s.io/implementations/, I'd like to add the following implementation:

> ### HAProxy Kubernetes Ingress Controller
>
> HAProxy Kubernetes Ingress Controller is an open-source project maintained by HAProxy Technologies that provides fast and efficient traffic management, routing, and observability for Kubernetes. It has built-in support for the Gateway API since version 1.10. The same deployment of the ingress controller will allow you to use both the Ingress API and Gateway API. See the [documentation](https://www.haproxy.com/documentation/kubernetes-ingress/gateway-api/enable-gateway-api/) for more details. In the [GitHub repository](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/gateway-api.md), you will also find additional information about supported API resources.

**Why this is needed**:

To provide Gateway API implementation information for those looking in the K8s docs.

Relevant source file: https://github.com/kubernetes-sigs/gateway-api/blob/main/site-src/implementations.md

NickMRamirez:
**What type of PR is this?**

/kind documentation

Fixes #2779 

-->
```release-note
* DOCUMENTATION: Added HAProxy Kubernetes Ingress Controller to Gateway API Implementations page.
```

### DIFF
--- a/site-src/implementations.md
+++ b/site-src/implementations.md
@@ -24,6 +24,7 @@ cover, and documentation to help users get started.
 - [Gloo Gateway 2.0][5] (beta)
 - [Google Kubernetes Engine][6] (GA)
 - [HAProxy Ingress][7] (alpha)
+- [HAProxy Kubernetes Ingress Controller][32] (GA)
 - [HashiCorp Consul][8]
 - [Istio][9] (beta)
 - [Kong][10] (GA)
@@ -80,6 +81,7 @@ cover, and documentation to help users get started.
 [29]:#tyk
 [30]:#easegress
 [31]:#avi-kubernetes-operator
+[32]:#haproxy-kubernetes-ingress-controller
 
 [gamma]:/concepts/gamma/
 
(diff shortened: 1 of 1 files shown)


NickMRamirez:
@robscott To your question "have you tried running conformance tests with your implementation yet?" I am asking our developers about that.

NickMRamirez:
@robscott , 

> I'd recommend noting that in the description of your implementation here.

My concern here is that the implementation and Gateway API version it's based on are subject to change. My feeling is that it's better for the user to go to the docs for the project and see what's new there. What do you think? 

I'll pass along the idea to test against v1.0 to the developers.

You are correct about the two HAProxy implementations. Naming is tricky. It's certainly something we've contended with, since users do get confused by the two projects. But after a lot of discussion, we haven't come up with a solution, unfortunately.

robscott:
>My concern here is that the implementation and Gateway API version it's based on are subject to change. My feeling is that it's better for the user to go to the docs for the project and see what's new there. What do you think?

I think the vast majority of implementations in this list support HTTPRoute and do not support TCPRoute, so unless you're working to change that sometime soon, it could be useful to call out the difference and save prospective users some time. Certainly not a requirement though. 

This will all eventually be resolved when @xtineskim completes https://github.com/kubernetes-sigs/gateway-api/issues/2550, which should show the features that all implementations support, as long as they've submitted a conformance report - see https://gateway-api.sigs.k8s.io/geps/gep-1709/ for more. As you note though, TCP is definitely a gap in our conformance coverage right now, so until we resolve that, mentioning support directly is probably the best path.

With all that said, it's great to see another implementation here. Feel free to submit follow up PRs if you'd like to tweak anything, but I think this is good to go as is.

/lgtm
/approve

shaneutt:
We don't need to block this PR for it, but at some point we should go back to our discussions about orgs with multiple Gateway API projects, as there may be some better ways we can provide display and organization for users who find their way into the page.

NickMRamirez:
Thanks all!

PR #2797, opened 2024-02-22
Remove hidden dependency for client-go/plugin/pkg/client/auth

ingwonsong:
**What type of PR is this?**
/kind cleanup
/area conformance

There is unnecessary import of `client-go/plugin/pkg/client/auth` in `conformance/utils/kubernetes/certificates.go`.
So, this PR proposes to remove the import.

Since `conformance/conformance.go` already import the auth plugin package, it is not necessary to import. 
In addition, due to this dependency, if we want to make a code based on `conformance/utils/suite` or `conformance/utils/kubernetes` package, the code should unavoidably have the auth plugin dependencies comes from `client-go/plugin/pkg/client/auth`.

```release-note
NONE
```


robscott:
Thanks @ingwonsong!

/ok-to-test
/lgtm
/approve

PR #2800, opened 2024-02-22
feat: CRD version and channel added to conformance reports

### LINKED ISSUE #2738: Automatically discover and set the Gateway API version and channel during conformance suite runs

As described in [GEP-1709](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709/index.md), the conformance report includes the Gateway API version and the Gateway API channel. These two parameters need to be automatically discovered by the suite, instead of being manually set by a user. To do so, the [related paragraph](https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1709/index.md#gateway-api-version-and-channel) of the GEP has to be implemented.

/area conformance

mlavacca:
**What type of PR is this?**

/kind feature
/area conformance

Gateway API version and channel are properly discovered during experimental conformance suite runs and set in the conformance report.

Fixes #2738 

```release-note
Gateway API Channel has been added to the conformance report API. Gateway API version and channel are properly discovered during experimental conformance suite runs and set in the conformance report.
```

### DIFF
--- a/conformance/utils/suite/experimental_suite.go
+++ b/conformance/utils/suite/experimental_suite.go
@@ -97,9 +97,14 @@ type ExperimentalConformanceOptions struct {
 	Options
 
 	Mode                string
+	AllowCRDsMismatch   bool
 	Implementation      confv1a1.Implementation
 	ConformanceProfiles sets.Set[ConformanceProfileName]
 }
+
+const (
+	undefinedKeyword = "UNDEFINED"
+)
 
 // NewExperimentalConformanceTestSuite is a helper to use for creating a new ExperimentalConformanceTestSuite.
 func NewExperimentalConformanceTestSuite(options ExperimentalConformanceOptions) (*ExperimentalConformanceTestSuite, error) {
@@ -117,7 +122,14 @@ func NewExperimentalConformanceTestSuite(options ExperimentalConformanceOptions)
 	}
 	apiVersion, apiChannel, err := getAPIVersionAndChannel(installedCRDs.Items)
 	if err != nil {
-		return nil, err
+		// in case an error is returned and the AllowCRDsMismatch flag is false, the suite fails.
+		// This is the default behavior but can be customized in case one wants to experiment
+		// with mixed versions/channels of the API.
+		if !options.AllowCRDsMismatch {
+			return nil, err
+		}
+		apiVersion = undefinedKeyword
+		apiChannel = undefinedKeyword
 	}
 
 	mode := flags.DefaultMode
@@ -395,5 +407,9 @@ func getAPIVersionAndChannel(crds []apiextensionsv1.CustomResourceDefinition) (v
 	if version == "" || channel == "" {
 		return "", "", errors.New("no Gateway API CRDs with the proper annotations found in the cluster")
 	}
+	if version != consts.BundleVersion {
+		return "", "", errors.New("the installed CRDs version is different from the suite version")
+	}
+
 	return version, channel, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/consts/consts.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package consts
+
+const (
+	BundleVersionAnnotation = "gateway.networking.k8s.io/bundle-version"
+	ChannelAnnotation       = "gateway.networking.k8s.io/channel"
```
shaneutt, line 21: Public symbols should have `godoc` comments

shaneutt (commented):
Thanks @mlavacca this is great. I'm assuming this has already been tested out on an implementation?

mlavacca:
> Thanks @mlavacca this is great. I'm assuming this has already been tested out on an implementation?

Yep, this has been tested with the Kong Ingress Controller.

pkg/consts/consts.go
mlavacca: oh right 👍

mlavacca: Comments added

shaneutt (approved):
/approve
/cc @robscott @youngnick

conformance/utils/suite/experimental_suite.go
```diff
@@ -83,51 +96,63 @@ type ExperimentalConformanceOptions struct {
...
+	}
+
+	installedCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+	err := options.Client.List(context.TODO(), installedCRDs)
+	if err != nil {
+		return nil, err
+	}
+	apiVersion, apiChannel, err := getAPIVersionAndChannel(installedCRDs.Items)
+	if err != nil {
+		return nil, err
```
robscott, line 114: If I'm understanding this correctly, this means that it would be impossible to run conformance tests in a cluster with mismatched CRDs? I'm not sure we want to be so strict here, maybe we should present a warning and/or have some kind of output that means that CRDs from different bundle versions or release channels were used.

For example, imagine that someone wanted the GA version of HTTPRoute, Gateway, GatewayClass installed because they were using those in production, but also wanted to experiment with gRPCRoute for a different controller. It seems like this change would prevent anyone from running conformance tests in that cluster.

As an alternative, it may be viable to say that the only way you can print out a formal/verified conformance report is to have all CRDs match version and channel, but still allow anyone else to run tests with mismatched version/channel just to verify the results or do some form of spot testing.

Or one more option - this is really just metadata, if we can't find it, maybe we just don't populate it at all?

robscott (commented):
Thanks @mlavacca!

conformance/utils/suite/experimental_suite.go
mlavacca: I get your point @robscott, I've made the following change:
- default behavior: in case there is a mismatch between versions or channels, the suite fails.
- the option `AllowCRDsMismatch` allows to customize this behavior; in case it is set to false and there is a mismatch between versions and channels, the suite does not fail, and the value `UNDEFINED` is set in the `GatewayAPIVersion` and `GatewayAPIChannel` of the report. There is a corresponding flag to customize such options when using the CLI.
- I've added an additional check on the suite version. In case the CRDs version is different from the suite version, the suite fails (unless the `AllowCRDsMismatch` option is set).

robscott:
Thanks @mlavacca!

/lgtm

Issue #2649, opened 2023-12-06
GEP: Inherited Policy Attachment


youngnick:
Split out of #713.

This GEP covers work to define and implement an example Inherited Policy.

This includes at a minimum:
- defining what makes an Inherited Policy
- suggested patterns for ensuring discoverable status

Graduation criteria will be something like:
- At least one included Inherited Policy
- Included Policy must have conformance tests
- Included Policy may be Extended support (and so, optional).

Issue #2648, opened 2023-12-06
GEP: Direct Policy Attachment


youngnick:
Split out of #713.

This GEP covers work to define and implement an example Direct Policy (most likely BackendTLSPolicy, GEP #1897, since that's already existent and is a Direct Policy.)

This includes at a minimum:
- defining what makes a Direct Policy
- suggested patterns for ensuring discoverable status

Graduation criteria will be something like:
- At least one included Direct Policy
- Included Policy must have conformance tests
- Included Policy may be Extended support (and so, optional).

wpbeckwith:
I know this was split off from #713 but I'm confused as to where circuit breakers from the issue landed.  Is that here or in the inherited policy attachment issue?

youngnick:
Circuit breaking is a separate feature that we know we need, but yes, there is currently no GEP for that. I'll make an issue to be the place where people can register their interest now.

youngnick:
#3358 covers circuit breaking as a feature. Note that this is just to give folks a number, not a commitment to work on this yet. That's a discussion to have during the scoping phase, which will open after Gateway API 1.2 releases.

rikatz:
/lifecycle frozen

Adding frozen so it stops going to stale

PR #2813, opened 2024-02-26
Split GEP-713 into a Memorandum and separate GEPs for Direct and Inherited Policy

youngnick:
/kind gep

This PR splits up GEP-713 (Metaresources and Policy Attachment), changing GEP-713 into a Memorandum about the general idea of Polkicy Attachment, with GEP-2648 and GEP-2649 being Direct Attached and Inherited Policy designs respectively.

The idea is to allow us to iterate on and graduate each type of Policy separately.

This is a _big_ change, so I'll summarize the changes here:
- GEP-713 is now a Memorandum GEP
- GEP-2648 defines Direct Attached Policies and includes some changes to required and recommended status designs for that type
- GEP-2649 now holds the documentation about defaults and overrides, and what makes an Inherited Policy
- The definition of what makes a Policy Direct or Inherited has been updated:
  - Direct Policies now must _only_ affect the targeted object
  - Inherited Policies are any Policy that affects objects other than the targeted one

Updates #713
Updates #2648
Updates #2649 

```release-note
The recommendations for the Policy Attachment pattern have been substantially updated and clarified.
```

I'm pretty sure this makes sense, but I've read this over so many times now that I can't see it properly, and I need review. Please feel free to ping me about this to discuss, or bring it to #sig-network-gateway-api or the community meeting.


youngnick:
I've merged in #2442's changes to the Direct Attached Policy document.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,714 @@
...
+
+### Merging into existing `spec` fields
+
+It's possible (even likely) that configuration in a Policy may need to be merged
+into an existing object's fields somehow, particularly for Inherited policies.
+
+When merging into an existing fields inside an object, Policy objects should
+merge values at a scalar level, not at a struct or object level.
+
+For example, in the `CDNCachingPolicy` example above, the `cdn` struct contains
```
guicassolato, line 167: Is this the example provided in the _Hierarchy_ section?

```suggestion
For example, in the `CDNCachingPolicy` example below, the `cdn` struct contains
```

geps/gep-2649/index.md
```diff
@@ -0,0 +1,714 @@
...
+merging should take place. This mainly applies to `overrides`, since for
+`defaults`, there should be no value present in a field on the final object.
+
+This table shows how this works for various types:
+
+|Type|Object config|Override Policy config|Result|
+|----|-------------|----------------------|------|
+|string| `key: "foo"` | `key: "bar"`  | `key: "bar"` |
+|list| `key: ["a","b"]` | `key: ["c","d"]` | `key: ["c","d"]` |
+|`map[string]string`| `key: {"foo": "a", "bar": "b"}` | `key: {"foo": "c", "bar": "d"}` | `key: {"foo": "c", "bar": "d"}` |
```
guicassolato, line 195: It would be interesting to see an example with lists of structs being affected.

E.g.

|Type|Object config|Override Policy config|Result|
|----|-------------|----------------------|------|
|list| `key: [{"foo": "a", "bar": "b"}, {"foo": "c", "bar": "d"}]`| `key: [{"foo": "a", "baz": "e"}]` | ? |

What should the result be?

Note the second field of the struct in the policy says `baz`, not `bar`.

As for the value in the first field, should we expect `"foo": "c"` in the second struct to be replaced with `"foo": "a"`?

I wonder if it would make a difference if the field was called `name` instead of `foo`. Imagine `listType=map, listMapKey=name`.

gcs278 (commented):
+1 for the revised GEP organization. I couldn't find any comments other than superficial markdown rendering issues and some minor typos. Reviewing these GEP updates helps me prepare for getting the Session Persistence Policy Attachment GEP through. Hope this helps.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,714 @@
...
+resources such as Gateways or Namespaces that may apply to multiple child
+resources.
+
+As with most APIs, there are countless ways we could choose to expand this in
+the future. This includes supporting multiple targetRefs and/or label selectors.
+Although this would enable compelling functionality, it would increase the
+complexity of an already complex API and potentially result in more conflicts
+between policies. Although we may choose to expand the targeting capabilities
+in the future, at this point it is strongly preferred to start with a simpler
+pattern that still leaves room for future expansion.
```
guicassolato, line 259: This is an interesting decision, with which fundamentally I do not disagree of course. My 2 cents…

I can’t help thinking though on the implications of the `targetRef` structure not counting with the `sectionName` field as an option, unlike proposed for [Direct Policies](https://github.com/kubernetes-sigs/gateway-api/blob/a01cdd6cf0bb52dc2a9d967b11006b81a12b1239/geps/gep-2648/index.md?plain=1#L55-L58).

I understand the rationale behind this has to do with the cardinality of combinatory cases involving partial targeting of a resource on top of all other variables that come with Inherited Policies, and not because anybody would claim such use case is not needed. Otherwise it wouldn’t have been proposed for the other class of policies.

I believe that being able to target a section of a resource is as much needed for Inherited Policies as it is for Direct Policies, especially given complex objects such as Gateways (multiple listeners) and HTTPRoutes (multiple rules.) By not having such option in the `targetRef`, policy designers will sometimes have to find a way to specify that the fields declared in a metaresource have effect only at given point within the targeted object. I guess that will vary from implementation to implementation. Maybe it will be by means of merging values into those lists of complex structs using the _convention_ of the `name` field being the key of the structure, rather than as an identifier of a section _by design_.

I guess another option would be splitting the target resource into many. Of course, this approach would not pass the test where an implementation provides more than one kind of policy, after all one policy cannot target multiple objects. As an example, imagine xRoute _R_ is targeted by TimeoutPolicy _T_ and now one of its rules needs to be targeted by a new AuthPolicy _A_. If you split _R_ into _R1_ and _R2_, then you need 2 copies of TimeoutPolicy _T_ as well.

One could even argue moreover that the fact itself that an object would _have to split_ into many is a design flaw. Taken to the limit, you can end up into a situation of one route object per route rule, one gateway per gateway listener, etc.

Not having the `sectionName` in the `targetRef` is also inconsistent with the design for Direct Policies, which makes evolving policies from one class into the other not straightforward.

Again, I obviously won’t deny the cognitive complexity involved in computing and/or reasoning on the final effective policy while cascading throughout the hierarchy, with sets and subsets to work out along the way, and also that starting small and slowly expanding functionalities is a legit strategy. In this particular case however, for the sake of consistency between the APIs or simply because it’s needed in some cases and quite easily the lack of direction could lead to implementation chaos, I think that having the `sectionName` in the `targetRef` should be a choice of one who designs a metaresource and not a restriction imposed by the framework.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,714 @@
...
+
+## Policy Attachment examples and behavior
+
+This approach is building on concepts from all of the alternatives discussed
+below. This is very similar to the (now removed) BackendPolicy resource in the API,
+but also borrows some concepts from the [ServicePolicy
+proposal](https://github.com/kubernetes-sigs/gateway-api/issues/611).
+
+### Policy Attachment for Ingress
+Attaching a Directly Attached Policy to Gateway resources for ingress use cases
```
kate-osborn, line 210: Should this example be moved to the GEP on direct policy attachment?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,410 @@
...
+  things outside their own namespace should be avoided unless it uses a handshake
+  of some sort, where the things outside the namespace can opt–out of the behavior.
+  (Notably, this is the design that we used for ReferenceGrant).
+
+## Apply Policies to Sections of a Resource
+
+The `sectionName` field of `targetRef` can be used to target a specific section of other resources:
+
+* Service.Ports.Name
+* xRoute.Rules.Name
```
kate-osborn, line 102: I find this example confusing since xRoute Rules do not have a name field yet. I'm also curious why Gateway.Listeners.Name isn't listed here.

geps/gep-2648/index.md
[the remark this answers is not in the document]
youngnick: Ah, thanks for the reminder, yes, they need to be added to the table of contents file. I'll do that now.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,714 @@
...
+  Policies attached _lower down_ the hierarchy beat the same type of Policy
+  attached further _up_ the hierarchy.
+* For `defaults`, the _most specific_ value is the one _inside the object_ that
+  the Policy applies to; that is, if a Policy specifies a `default`, and an object
+  specifies a value, the _object's_ value wins.
+* Policies interact with the fields they are controlling in a "replace value"
+  fashion.
+  * For fields where the `value` is a scalar, (like a string or a number)
+    MUST have their value _replaced_ by the value in the Policy if it wins.
+    Notably, this means that a `default` will only ever replace an empty or unset
```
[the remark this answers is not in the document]
youngnick: Only a string with 0 length, updated.

geps/gep-2649/index.md
youngnick: Your points are fair, but I am _super_ worried about the complexity problem with `sectionName` for Inherited Policies.

Here I'm trying to go with "no is temporary, yes is forever", and not do it _yet_. If someone can come up with a way to manage the complexity that will result from having multiple Inherited policies, possibly with defaults and overrides, that also only target _part_ of a resource, and _still_ make it so that the owners of the affected resources can understand the system, without breaking the apiserver with constant status update load, then I'd love to say yes.

But that is a _lot_ of things that need to line up, so I'd prefer to leave this as is _for now_.

geps/gep-2649/index.md
youngnick: Interestingly, under the criteria as I've updated them, this is no longer a Direct Attached Policy, because it affects Routes and attaches to a Gateway. I'll update the wording.

geps/gep-2648/index.md
youngnick: I've added Gateway.Listeners.Name and put a caveat on the Named rules thing. Thanks!

youngnick:
I've got two TODOs left that I will get to early next week, both are relatively minor though, so please review away.

- Add changelog including the original PRs plus Flynn's one
- Update example to be BackendTLSPolicy

geps/gep-2649/index.md
guicassolato: Fair enough, @youngnick. I think your concerns are well-justified. At Kuadrant, we'll continue to try to find a solution for the problem.

API-wise, having the option to use the `sectionName` in the `targetRef` would open up this path to explore (without breaking apart from GEP-2649.) We were hoping to obtain simpler policies and simpler status blocks as well.

Our other option is to use the policy spec – I mean the part that is not `targetRef` obviously. Maybe users will write bigger policies, with bits that sometimes apply to the entire target, sometimes only to parts of it. The status block of the policy will probably be a bit messier.

Either way, targeting only a part of an object is sometimes unavoidable, I believe. Especially when the domain is auth (different authz rules for different route endpoints – e.g. admin, non-admin, anonymous access, etc) and rate-limit (again, different rate limits for different API endpoints – e.g. more expensive operation, less expensive operation), Kuadrant's kinds of Inherited Policies.

youngnick: Look, I can't stop you from using the `targetRef` variant that has `sectionName`. It's your policy. But I suppose I could make this "Inherited Policies SHOULD NOT use `sectionName` " instead if you really want to be covered.

But you could go and do this right now, and then we can change the document later. This is still Experimental, so that kind of change is okay.

youngnick: I reread this now, and I think I explained myself poorly.

What I'm saying is; it's _fine_ for you to go and build an Inherited Policy that uses `sectionName` to try things out. It will _currently_ be not compliant to the spec, but in the event you find that the discoverability and complexity problems are not as bad as we fear, then we can update this with your experience.

I would, however, prefer not to block this merging on this specific thing. Like I said earlier, this is "not supported _for now_". But I'm happy to work with you on what doing `sectionName` on Inherited Policies could look like.

geps/gep-2648/index.md
```diff
@@ -0,0 +1,412 @@
...
+
+See GEP-1897 for all the details of this Policy object.
+
+## Direct Policy design guidelines
+
+With this example in mind, here are some guidelines for when to consider
+using Direct Policy Attachment:
+
+* The number or scope of objects is exactly _one_ object.
+* The modifications to be made to the objects don’t have any transitive information -
```
howardjohn, line 71: I don't think its clear what it means to "only affect the single object". I don't feel like there are any definitions for what the scope of such objects are.

To make things more concrete, what if I apply a BackendTimeoutPolicy to a Gateway. What are the implications here?

geps/gep-2649/index.md
```diff
@@ -0,0 +1,733 @@
...
+      annotations will have its value _entirely replaced_ by an `override` setting.
+* In the case that two Policies of the same type specify different fields, then
+  _all_ of the specified fields MUST take effect on the affected object, using
+  the precedence rules given above.
+
+Examples to further illustrate these rules are given below.
+
+#### Supported Resources
+It is important to note that not every implementation will be able to support
+policy attachment to each resource described in the hierarchy above. When that
```
kate-osborn, line 146: I think this section should be nested under the Hierarchy header that comes later in the doc.

geps/gep-2648/index.md
youngnick: If the backend timeouts are set per-backend, then it's an Inherited Policy, because there's no backend config in Gateway.

The intent here is to tightly constrain what a Direct Attached Policy is so that we don't have to boil the ocean of finding a good way to tell everyone whose objects are affected by a policy about that in the status of all the resources.

It's also about ensuring that we don't have "Direct Attached" Policy objects that effectively set defaults, even if the `defaults` stanza isn't used.

youngnick: I added some more text here.

geps/gep-2649/index.md
youngnick: Good suggestion, thanks.

geps/gep-2648/index.md
howardjohn: The goals make sense,I am just trying to fully grok the meaning of this since it seems like this is an important aspect 

What does it mean to have a timeout set per-backend?

I can say logically that a timeout seems to apply to a backend in my mental model. And maybe in proxy implementations. But I could have an implementation that has one global value "timeout", and Gateway is the only logical place to put that policy in that implementation. Is that ok, or does that violate this?

What if my data plane implementation does happen to have per-backend config, but I don't want to deal with it so I only support gateway level config - is that ok?

More broadly, what exactly defined what is the scope of impact of an object? Is it the ability to attach a single type to multiple places? Is it the implementation details of how a policy is implemented? Is it a users mental model of how it works?

robscott: I think this question overlaps with the session persistence GEP @gcs278 has been working on (#2634). In that case we've found that some implementations (Envoy or GRPC-based in particular) can only attach this config at Route level. So even if a user attaches the config at the Backend layer, the underlying implementation is going to be implementing it at the Routing layer. Would that go against this guideline?

Similarly, that GEP currently proposes that Routes will include inline overrides that could override the values specified on a Service. That's not really policy at that point, but it is some config overriding a direct-attached policy which I believe would be a new concept for us. 

To be clear, I actually like the approach that GEP is proposing, I'm just trying to make sure that the corresponding changes we're making here can line up with that so we don't end up with conflicting GEPs.

youngnick: Good questions, thanks @howardjohn

> I can say logically that a timeout seems to apply to a backend in my mental model. And maybe in proxy implementations. But I could have an implementation that has one global value "timeout", and Gateway is the only logical place to put that policy in that implementation. Is that ok, or does that violate this?

I guess in this case it would be a Direct Attached Policy, yes. It's all about _which resources you need to know about to understand the state of the system_, at the end of the day.
 
> What if my data plane implementation does happen to have per-backend config, but I don't want to deal with it so I only support gateway level config - is that ok?

If you only need to know about the Policy and the Gateway to know everything you need to, then yes.

> More broadly, what exactly defined what is the scope of impact of an object? Is it the ability to attach a single type to multiple places? Is it the implementation details of how a policy is implemented? Is it a users mental model of how it works?

The best way I've found to think of the scope is which objects are _impacted_ by the Policy. That's going to depend on exactly what the Policy actually _does_, and so is going to be Policy-specific.

To use your examples:
- being able to attach a single type to multiple places is almost certainly going to end up as an Inherited Policy, because _probably_, the settings will effectively be defaults or overrides (if they can be attached at different levels). If a Policy can only be attached to sibling objects in a hierarchy (only to Routes, for example), _and_ only affects properties of those sibling objects (so, attaching to Routes can't change backend properties for Service), then it can be a Direct Attached Policy.
- Is the implementation details of how a policy is implemented? In a word, yes. Policy writers will need to be specific about describing what a Policy affects.
- Is it a user's mental model of how it works? Ideally, this model should be pretty similar if not identical to how it's implemented, and the Policy should be able to _clearly indicate_ what objects it is affecting. What "affecting" means depends on the settings defined by the Policy and the implementation.

My personal guideline is "If I'm a user, how will I know if my thing is affected by a Policy?" If there's _any_ level of action-at-a-distance, then the Policy is an Inherited one.

If you have suggestions on better ways to communicate this, happy to change things up more.

geps/gep-2649/index.md
youngnick: This is a good point, I'll add some rules around listMapType objects (since they are effectively maps).

geps/gep-2648/index.md
robscott: > If there's any level of action-at-a-distance, then the Policy is an Inherited one.

I think there's probably a category of policies that we know will exclusively be attached at one level and thus would never benefit from defaults and overrides, but would also still have some knock-on effects above or below where they're attached. 

It may be simpler to just say "direct policies can be attached at exactly one layer while inherited policies are used when a policy can be attached at more than one layer with support for defaults and overrides". I think that may still achieve the goal you're aiming for here while also leaving room for these inevitable policies that clearly only fit in one place but could indirectly affect other layers.

howardjohn: Do we need take in policies that are additive? They don't make sense to have defaults and overrides, but do target multiple layers. Authorization policies are likely to fall info this category

robscott: 🤔 that's a good point. Are you referring to policies that essentially just have implicit defaulting behavior all the way down the stack? For example:

policy attached to GW:
```
foo: 1
bar: 2
```

policy attached to Route:
```
bar: 4
baz : 3
```

computed result:
```
foo: 1
bar: 4
baz: 3
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+
+### Other Status designs
+
+This section contains other recommendations for status designs. Note that this
+is a SHOULD rather than a MUST, as this design is still not final.
+
+
+#### Standard status struct
+
+Status: Experimental
```
robscott, line 259: Why does this need a separate status?

geps/gep-2648/metadata.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: internal.gateway.networking.k8s.io/v1alpha1
+kind: GEPDetails
+number: 2648
+name: Direct Policy Attachment
+status: Experimental
+# Any authors who contribute to the GEP in any way should be listed here using
+# their Github handle.
+authors:
+  - robscott
+  - youngnick
```
robscott, line 10: Nit: I think you're the primary author of policy GEPs at this point
```suggestion
  - youngnick
  - robscott
```

geps/gep-2649/index.md
```diff
@@ -0,0 +1,744 @@
...
+in communicating the currently active setting to the owners of the modified
+objects. These are discussed in more detail below.
+
+## Inherited Policy design guidelines
+
+Here are some guidelines for when to consider using a Inherited Policy object:
+
+* The settings or configuration are bound to one containing object, but affect
+  other objects attached to that one (for example, affecting HTTPRoutes attached
+  to a single Gateway, or all HTTPRoutes in a GatewayClass).
```
robscott, line 85: Related to other longer comment thread, I think this may end up capturing more than intended. Often the GW API resources don't match closely with the underlying infra, so you maybe combining information from multiple layers in the underlying implementation. For example, GKE's implementation of the API takes information from Gateways and Routes when configuring [URLMaps](https://cloud.google.com/compute/docs/reference/rest/v1/urlMaps), so it would be ~impossible for _any_ potential policy we created to not meet this specific condition.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,744 @@
...
+  Policies attached _higher_ up the hierarchy beat the same type of Policy
+  attached further down the hierarchy.
+* `defaults` settings operate in a "more specific beats less specific" fashion -
+  Policies attached _lower down_ the hierarchy beat the same type of Policy
+  attached further _up_ the hierarchy.
+* For `defaults`, the _most specific_ value is the one _inside the object_ that
+  the Policy applies to; that is, if a Policy specifies a `default`, and an object
+  specifies a value, the _object's_ value wins.
+* Policies interact with the fields they are controlling in a "replace value"
+  fashion.
```
robscott, line 120: Not sure I'm following this - is this referring to an interaction between a policy and a built-in resource like a Route?

geps/gep-2649/index.md
```diff
@@ -0,0 +1,744 @@
...
+target. This ensures that only users with write access to a policy resource in a
+given scope will be able to modify policy at that level. Since GatewayClass is a
+cluster scoped resource, this means that any policy attached to it must also be
+cluster scoped.
+
+GatewayClass parameters provide an alternative to policy attachment that may be
+easier for some implementations to support. These parameters can similarly be
+used to set defaults and requirements for an entire GatewayClass.
+
+### Merging into existing `spec` fields
```
robscott, line 156: This logic is fairly complex. It would be great if we could eventually offer gwctl as a library that could help with implementing this logic + ensure consistency across implementations (cc @gauravkghildiyal).

geps/gep-2649/index.md
```diff
@@ -0,0 +1,744 @@
...
+#### Defaults interacting with other defaults for RetryOnPolicy, value in HTTPRoute
+||No default|Namespace default A|Gateway default A|HTTPRoute default A|
+|----|-----|-----|----|----|
+|No default|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|
+|Namespace default B|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|
+|Gateway default B|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|
+|HTTPRoute default B|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|Value in HTTPRoute|
+
+
+## User discoverability and status
```
robscott, line 663: Why doesn't this use the same status as BackendTLSPolicy/direct policy attachment as a baseline? I think we could add more on top of that for inherited policies, but that seems like a good baseline for all policies.

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+#### On targeted resources
+
+Implementations that use Direct Policy objects SHOULD put a Condition into
+`status.Conditions` of any objects affected by a Direct Policy, if that field
+is present.
+
+If they do, that Condition MUST have a `type` ending in `PolicyAffected` (like
+`gateway.networking.k8s.io/PolicyAffected`),
+and have the optional `observedGeneration` field kept up to date when the `spec`
+of the Policy-attached object changes.
```
robscott, line 231: How does this work for implementations like Kuadrant that are implementing policies alongside of the primary implementations of these APIs? I think the intent here is that in the case of a Gateway for example, only the controller implementing that Gateway directly should populate this status?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+
+Implementations SHOULD use their own unique domain prefix for this Condition
+`type` - it is recommended that implementations use the same domain as in the
+`controllerName` field on GatewayClass (or some other implementation-unique
+domain for implementations that do not use GatewayClass).
```
robscott, line 236: If we had multiple controllers populating the same conditions list, wouldn't we risk a loop where each controller was trying to populate the "correct" set of conditions?

robscott (commented):
Thanks for all the work on this @youngnick!

geps/gep-2649/index.md
guicassolato: Thanks, @youngnick! Awesome example, to leave no doubts!

geps/gep-2648/index.md
guicassolato: Kuadrant uses `kuadrant.io/DNSPolicyAffected` or `kuadrant.io/TLSPolicyAffected` (depending on the kind of policy), to set the condition on the Gateway object. It also keeps the `observedGeneration` field up to date.

To avoid arm wrestling against the primary gateway controller:
1. it's always additive, relying on [`meta.SetStatusCondition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition) and [`meta.RemoveStatusCondition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#RemoveStatusCondition) - i.e. it tries not to touch on the other conditions;
2. in case of race condition detected and therefore the API server failing to update due to the state of the object changed, it requeues to try again later clean.

geps/gep-2648/index.md
guicassolato: I can imagine maybe an API or protocol that the primary gateway controller would commit to, and thus take responsibility for all updates. That not being the case, we may depend on the multiple controllers just to be good citizens, I guess. But I could be oversimplifying. Not sure if this would still hold for – say, let's exaggerate – dozens of controllers. At the very least, it can get a bit noisy.

geps/gep-2648/index.md
howardjohn: I was thinking:
GatewaY:
```
rules: [a]
```

Route:
```
rules: [b]
```

Result: `rules: [a,b]`

youngnick: At the end of the day, what I'm trying to do is make it clear that Direct Policies have an easier time with status. If you do _any_ cross object inheritance as mentioned in the previous two comments, those count as inherited, _because you have to solve the Inherited status problem_ to have full status.

"Direct Policies can be attached at exactly one layer" is a good _start_, but if they have knock-on effects further down the hierarchy, then again, _they have the same status problem as Inherited Policies_.

Maybe I just make it more explicit here:
- Direct Policies can be attached at exactly one layer
- Direct Policies can have effects only on the layer they attach to
- Direct Policies can effects only on the object withing the layer that they attach to

This stops the "but my Direct Policy affected another Route attached to the same Gateway when attached to a Route" problem which could arise if we only had the first two items.

geps/gep-2649/index.md
youngnick: Yes. I'll see if I can update to make this clearer, but this is talking about the case where we have a _field_ in an object (like a Route), that is defaulted or overridden by a Policy.

geps/gep-2648/index.md
youngnick: Thanks @guicassolato, that's exactly what I intended with this, to not block multiple controllers from being able to say "hey, this object is affected by some policy, you should look into that".

geps/gep-2648/index.md
youngnick: As @guicassolato said on another comment, the way this is _supposed_ to work is that controllers are supposed to update Conditions lists _additively_, putting back any Condition they don't understand or update as they found it. And because Conditions are a listTypeMap, the ordering shouldn't matter as far as the apiserver is concerned (I think, would need to check with apimachinery if that's actually the case).

geps/gep-2648/index.md
youngnick: Oh, and @howardjohn, additive policies being Inherited is _definitely_ intended, that's one of the things I was worried that the original definition missed. That kind of implicit default still has all the status problems of a more explicit Inherited policy with a defined `defaults` section, so I wanted to make sure that we caught that in the definitions.

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+* Status: Experimental
+
+(See status definitions [here](/geps/overview/#gep-states)
+
+## TLDR
+
+Describe and specify a design pattern for a class of metaresource that can
+affect specific settings across a single target object.
+
+This is a design for a _pattern_, not an API field or new object.
```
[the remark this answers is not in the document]
youngnick: Yeah, this is a great idea. I'll update GEP-713 with something.

geps/gep-2648/index.md
youngnick: This is an attempt to codify what we're doing in BackendTLSPolicy into this GEP, that's all.

geps/gep-713/index.md
```diff
@@ -26,16 +20,25 @@ settings across either one object (this is "Direct Policy Attachment"), or objec
...
+  case, the Policy is an Inherited Policy. A common way to do this is to include
+  either a `defaults` section, an `overrides` section, or both.
+- Policy objects that affect _only_ the object specified in the `targetRef` are
+  Direct Attached Policies (or more simply, Direct Policies.)
+
+Direct Attached Policies are further specified in the addendum GEP GEP-2648,
+Direct Policy Attachment.
 
-For Inherited Policies, this GEP also describes a set of expected behaviors
+Inherited Policies are further specified in the addendem GEP-2649, Inherited
```
candita, line 40: nit: 
```suggestion
Inherited Policies are further specified in the addendum GEP-2649, Inherited
```

geps/gep-713/index.md
```diff
@@ -26,16 +20,25 @@ settings across either one object (this is "Direct Policy Attachment"), or objec
...
+Policy Attachment. This GEP also describes a set of expected behaviors
```
candita, line 41: Does "this GEP" refer to GEP-2649 or this document?
```suggestion
Policy Attachment. [GEP-2649 | This document] also describes a set of expected behaviors
```

geps/gep-713/index.md
```diff
@@ -109,395 +114,67 @@ additional fields that are _not_ in the `spec` of the object.
...
-  targetRef: # This struct is defined as part of Gateway API
-    group: "" # Empty string means core - this is a standard convention
-    kind: Service
-    name: fooService
-  tls:
-    certificateAuthorityRefs:
-      - name: CAcert
-    port: 8443
+The basic rule here is "Does the Policy affect _any_ other object aside from
+the one it targets?" If not, it's Direct. If so, it's inherited.
```
candita, line 128: nit: 
```suggestion
The basic rule here is "Does the Policy affect _any_ other object aside from
the one it targets?" If not, it's Direct. If so, it's Inherited.
```

geps/gep-713/index.md
```diff
@@ -109,395 +114,67 @@ additional fields that are _not_ in the `spec` of the object.
...
 
-```
+The reason for this is that Direct Attached Policies make it _much_ easier to
+understand the state of the system, and so can use a more simple `status` design.
+However, Inherited Policies require knowledge of more resources, and consequently
+a more complex status design.
```
candita, line 133: I think this belongs in the TLDR when explaining why there needs to be two different kind of policies.

geps/gep-713/index.md
```diff
@@ -109,395 +114,67 @@ additional fields that are _not_ in the `spec` of the object.
...
-Kind within a single namespace (or to an instance of a single Kind at cluster scope),
-and only modifies the behavior of the object that matches its binding.
+For more description of the details of Direct Policy Attachment,
+see [GEP-2648](https://gateway-api.sigs.k8s.io/geps/gep-2648/).
 
-As an example, one use case that Gateway API currently does not support is how
-to configure details of the TLS required to connect to a backend (in other words,
-if the process running inside the backend workload expects TLS, not that some
-automated infrastructure layer is provisioning TLS as in the Mesh case).
+### Inherited Policy Attachment
```
candita, line 120: I imagine this has already been discussed ad nauseam, but why not name `Inherited` Policy Attachment as `Multiple` or `Indirect`?

geps/gep-713/index.md
```diff
@@ -26,16 +20,25 @@ settings across either one object (this is "Direct Policy Attachment"), or objec
...
-- and must include a common `TargetRef` struct in their specification to identify
+- MUST include a common `TargetRef` struct in their specification to identify
   how and where to apply that policy.
-- _may_ include either a `defaults` section, an `overrides` section, or both. If
-  these are included, the Policy is an Inherited Policy, and should use the
-  inheritance rules defined in this document.
+- MAY affect more objects than the object specified in the `targetRef`. In this
+  case, the Policy is an Inherited Policy. A common way to do this is to include
+  either a `defaults` section, an `overrides` section, or both.
+- Policy objects that affect _only_ the object specified in the `targetRef` are
```
candita, line 34: Can there be more than one object specified in the targetRef, as long as they are not hierarchically related?
```suggestion
- Policy objects that affect _only_ the object/s specified in the `targetRef` are
```

geps/gep-713/index.md
```diff
@@ -109,395 +114,67 @@ additional fields that are _not_ in the `spec` of the object.
...
-GatewayClass parameters provide an alternative to policy attachment that may be
-easier for some implementations to support. These parameters can similarly be
-used to set defaults and requirements for an entire GatewayClass.
-
-### Targeting External Services
+### Targeting Virtual Types
 In some cases (likely limited to mesh) we may want to apply policies to requests
-to external services. To accomplish this, implementations can choose to support
+to external services. To accomplish this, implementations MAY choose to support
 a reference to a virtual resource type:
```
candita, line 178:
```suggestion
a reference to a virtual resource type.  For example:
```

geps/gep-713/index.md
```diff
@@ -568,8 +206,8 @@ ties:
 * Inside Inherited Policies, the same setting in `overrides` beats the one in
   `defaults`.
 * The oldest Policy based on creation timestamp. For example, a Policy with a
```
candita, line 208:
```suggestion
* The oldest Policy based on creation timestamp beats a newer one. For example, a Policy with a
```

geps/gep-713/index.md
```diff
@@ -1043,6 +683,9 @@ Support: Provisional
 
 This solution is IN PROGRESS and so is not binding yet.
 
+However, a version of this proposal is now included in the Direct Policy
+Attachment GEP.
+
 This solution requires definition in a GEP of its own to become binding.
```
candita, line 689: It would be good to start the issue for the PolicyAncestor GEP and include a reference to it here.

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+* Fully specify the entire design space for Direct Policy Attachment
+
+## Introduction
+
+GEP-713 defines two classes of Policy Attachment: Direct and Inherited.
+
+Direct Attached Policies (or Direct Policies) _only_ affect the object they are
+attached to; that is, the object specified in their `targetRef`.
+
+Note that as soon as the Policy affects more objects than the referenced object,
```
candita, line 32: I mentioned this in a review of the gep-713 changes, but again, shouldn't a Direct Policy be able to refer to a collection/set of objects, as long as they are all at the same hierarchical level?  E.g. a targetRef of a set of common rules or specific routes; or a struct that groups non-hierarchical objects from the same namespace.

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+  and MUST clearly indicate the objects that `sectionName` may be used on in its
+  documentation.
+
+## Direct Policy Attachment
+
+A Direct Policy Attachment is tightly bound to one instance of a particular
+Kind within a single namespace (or to an instance of a single Kind at cluster scope),
+and only modifies the behavior of the object that matches its binding.
+
+As an example, the BackendTLSPolicy is specified in GEP-1897, BackendTLSPolicy -
```
candita, line 58:
```suggestion
As an example, the BackendTLSPolicy is specified in [GEP-1897](https://gateway-api.sigs.k8s.io/geps/gep-1897/), BackendTLSPolicy -
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+* The modifications to be made to the objects don’t have any transitive information -
+  that is, the modifications MUST only affect the single object that the targeted
+  metaresource is bound to, and MUST NOT have ramifications that flow beyond that
+  object. No attaching a Policy to a Gateway and affecting settings in Routes or
+  backends. If a Direct Attached Policy attaches to an object, it can only affect
+  properties _of that object_.
+* In terms of status, it SHOULD be reasonably easy for a user to understand that
+  everything is working - basically, as long as the targeted object exists, and
+  the modifications are valid, the metaresource is valid, and this should be
+  straightforward to communicate in one or two Conditions. The `status` stanza
```
candita, line 81:
```suggestion
  everything is working - basically, as long as the targeted object exists, 
  the modifications are valid, the metaresource is valid, and the status is clearly expressed in one or two Conditions.  The `status` stanza
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+  significant security concerns, and/or difficulties about merging cross-namespace
+  policy objects. Notably, Mesh use cases may need to do something like this for
+  consumer policies, but in general, Policy objects that modify the behavior of
+  things outside their own namespace should be avoided unless it uses a handshake
+  of some sort, where the things outside the namespace can opt–out of the behavior.
+  (Notably, this is the design that we used for ReferenceGrant).
+
+## Apply Policies to Sections of a Resource
+
+The `sectionName` field of `targetRef` can be used to target a specific section of other resources:
```
candita, line 94:
```suggestion
The `sectionName` field of `targetRef` can be used to target a specific section of other resources, for example:
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+
+* Service.Ports.Name
+* Gateway.Listners.Name
```
candita, line 97:
```suggestion
* Gateway.Listeners.Name
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+  targetRef:
+    name: http-app-1
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    sectionName: bar
+```
+
+If a `sectionName` is specified, but does not exist on the targeted object, the
+Policy must fail to attach, and the policy implementation should record a
+`resolvedRefs` or similar Condition in the Policy's status.
```
candita, line 138:
```suggestion
`resolvedRefs` failure or similar Condition in the Policy's status.
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+kind: HTTPRoute
+metadata:
+  name: http-app-1
+  labels:
+    app: foo
+spec:
+  hostnames:
+  - "foo.com"
+  rules:
+  - name: bar
```
candita, line 114: This is the sectionName, but I believe the same name can be used in different stanzas that are different objects, which breaks the application of policy to section name.  For example, couldn't I have a rule and a backendRef with the same name?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+    name: http-app-1
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    sectionName: bar
+```
+
+If a `sectionName` is specified, but does not exist on the targeted object, the
+Policy must fail to attach, and the policy implementation should record a
+`resolvedRefs` or similar Condition in the Policy's status.
+
```
candita, line 139: Could you add a brief description of what happens when more than one object matches the sectionName in the targetRef?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+## User discoverability and status
+
+### Standard label on CRD objects
+
+Each CRD that defines a Direct Policy object MUST include a label that specifies that
+it is a Policy object, and that label MUST specify that the object is a `direct`
+one.
+
+The label is `gateway.networking.k8s.io/policy: direct`.
```
candita, line 156: If we have this label, why would we still need to follow a rule that place the suffix "Policy" on the name of the CRD?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+Condition format.
+
+Policy authors should consider namespacing the `conditions` stanza with a
+`controllerName`, as in Route status, if more than one implementation will be
+reconciling the Policy type.
+
+#### On `Policy` objects
+
+Each Direct Attached Policy MUST populate the `Accepted` condition and reasons
+as defined in the Go spec, a snapshot of which is included below. The canonical
```
candita, line 176:
```suggestion
as defined in the PolicyTypes API, a snapshot of which is included below. The canonical
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+and have the optional `observedGeneration` field kept up to date when the `spec`
+of the Policy-attached object changes.
+
+Implementations SHOULD use their own unique domain prefix for this Condition
+`type` - it is recommended that implementations use the same domain as in the
+`controllerName` field on GatewayClass (or some other implementation-unique
+domain for implementations that do not use GatewayClass).
+
+For objects that do _not_ have a `status.Conditions` field available (`Secret`
+is a good example), that object SHOULD instead have an annotation of
```
candita, line 239: I hope!
```suggestion
is a good example), that object SHOULD instead have a label specifiying
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+`gateway.networking.k8s.io/PolicyAffected: true` (or with an
+implementation-specific domain prefix) added instead.
+
+Because these Conditions or annotations are namespaced per-implementation,
```
candita, line 243: Annotation or label?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+  // or semantically invalid.
+  PolicyReasonInvalid PolicyConditionReason = "Invalid"
+
+  // PolicyReasonTargetNotFound is used with the "Accepted" condition when the policy is attached to
+  // an invalid target resource
+  PolicyReasonTargetNotFound PolicyConditionReason = "TargetNotFound"
+)
+```
+
+#### On targeted resources
```
candita, line 222: There isn't a section on this in gep-2649.  Is it not possible or not recommended?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+
+The following hypothetical Policy sets the minimum TLS version required on a
+Gateway Listener:
+
+```yaml
+apiVersion: networking.example.io/v1alpha1
+kind: TLSMinimumVersionPolicy
+metadata:
+  name: minimum12
+  namespace: appns
```
candita, line 391: Is this missing the label?
```suggestion
  namespace: appns
  label: `gateway.networking.k8s.io/policy:direct'
```

geps/gep-2648/index.md
```diff
@@ -0,0 +1,415 @@
...
+
+Policy objects SHOULD use the upstream `PolicyAncestorStatus` struct in their
+respective Status structs. Please see the included `PolicyAncestorStatus` struct,
+and its use in the `BackendTLSPolicy` object for detailed examples. Included here
+is a representative version.
+
+This pattern enables different conditions to be set for different "Ancestors"
+of the target resource. This is particularly helpful for policies that may be
+implemented by multiple controllers or attached to resources with different
+capabilities. This pattern also provides a clear view of what resources a
```
kate-osborn, line 271: What does "capabilities" mean here? Can you include an example?

geps/gep-2648/index.md
robscott: Interesting - that feels like something we should drop into implementation guidelines so we avoid unfortunate surprises. 

I know we put a lot of thought into using `controllerName` in our nested conditions on Routes and Policies, are we saying that in those cases an implementation like Kuadrant would create a separate `RouteParentStatus` entry with their controller name, even if they're not the ones directly implementing the parent? I think that idea generally makes sense to me, but it gets more confusing when you have resources that we historically thought of as having status that was owned by a single controller.

geps/gep-713/index.md
youngnick: As of right now, no. I'm open to discussing that, but I think that should be a separate follow-up discussion (pretty sure that @guicassolato and @arkodg want multiple `targetRef`s in one Policy to be given the thumbs-up).

geps/gep-713/index.md
youngnick: "Inherited" is a closer match to what's actually happening. I do find myself using "Indirect" sometimes though, so it may well end up as an accepted synonum.

geps/gep-713/index.md
youngnick: Great point, thanks. I added a paraphrase of this paragraph to the beginning of the document.

geps/gep-2648/index.md
youngnick: Originally, I thought no, as it will add another layer of complexity onto the status design (each targetRef will probably need a new status...) But I think I could be talked around on this - but I think that this dicussion needs to happen _after_ this PR merges - this has already been in progress for way too long.

geps/gep-2648/index.md
youngnick: At the present, this is left to the Policy creator to define. I think this is another good follow-up discussion though.

geps/gep-2648/index.md
youngnick: For ease of use of humans, only.

geps/gep-2648/index.md
youngnick: I didn't explicitly specify it because it's similar to everything else, I will add it.

geps/gep-2648/index.md
youngnick: I'll make a note to do an update to the implemntation guidelines later about this.

geps/gep-2648/index.md
youngnick: I've updated to label, although I originally chose annotation because these are machine-generated and machine-added. I updated because it seems useful to be able to search objects to find policy affected ones (which you can't do with annotations).

geps/gep-2648/index.md
youngnick: Clarified and added.

geps/gep-2648/index.md
youngnick: Yes, it's from an earlier version that didn't have the requirement, thanks.

geps/gep-2649/index.md
youngnick: This is only intended to capture resources _in the same hierarchy_, or that users _need to know about_ to understand the state of the system.

If, in GKE's Policy objects, the users _need to know about_ the URLMaps when looking at the Gateways and Routes that they own, then yes, they can't be Direct Attached Policy objects, even if they meet the other criteria.

geps/gep-2649/index.md
```diff
@@ -0,0 +1,744 @@
...
+
+  // PolicyReasonInvalid is used with the "Accepted" condition when the policy is syntactically
+  // or semantically invalid.
+  PolicyReasonInvalid PolicyConditionReason = "Invalid"
+
+  // PolicyReasonTargetNotFound is used with the "Accepted" condition when the policy is attached to
+  // an invalid target resource
+  PolicyReasonTargetNotFound PolicyConditionReason = "TargetNotFound"
+)
+```
```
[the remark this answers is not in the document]
youngnick: I've been trying to keep this PR as small as possible, since it's already enormous. But I'll add some updates to the status sections to roll your and @robscott's feedback in.

geps/gep-2649/index.md
youngnick: Added some extra words at the top to try and make this clearer.

geps/gep-2648/index.md
guicassolato: The example provided assumes GEP-995 (https://github.com/kubernetes-sigs/gateway-api/pull/2593), I understand.

geps/gep-2649/index.md
youngnick: I was leaving this for a future discussion, in the interest of getting this in. I will include the existing status suggestion from Direct Policy Attachment as a SHOULD here.

youngnick: I didn't include it because it seems very likely that it could create fanout problems in the Inherited case (since a Policy may have many ancestors).

I'd like to plan on following up here once @guicassolato and the Kuadrant folks have more information.

geps/gep-2649/index.md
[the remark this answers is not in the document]
youngnick: I added some information suggesting the PolicyAffected condition, as in Direct Attached Policy, but I want us all to spend some time considering apiserver fanout  load concerns before going ahead with any more recommendations here.

geps/gep-713/index.md
arkodg: yah would love guidance / nod reg multiple `targetRefs` from GatewayAPI
cc @guydc

mikemorris: Istio has policies which allow specifying either a `targetRef` or `selector` such as https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy - I could see plural `targetRefs` being useful too.

howardjohn: :+1: to `targetRefs`

youngnick:
As discussed in #2927, I've marked GEP-2658 and GEP-2649 as provisional, and included notices at the top of each file that the file may change soon.

Hopefully this lets us get this one across the line for now, PTAL @robscott @howardjohn @mikemorris @kflynn @guicassolato @kate-osborn.

geps/gep-2649/index.md
mikemorris: I honestly think that the change introduced in https://github.com/kubernetes-sigs/gateway-api/pull/2442 to allow some form of inheritance precedence logic on **Direct** policies is misplaced, and should instead be allowed **only** for **Inherited** policies (or a single consolidated policy type), while it could be reasonable for Direct policies to raise a `Conflicted` status if multiple instances of the same policy type were to target overlapping sections of the same resource.

Targeting separate _sibling_ sections of the same resource with Direct policies (e.g. different policies of the same type each targeting a different Listener on a Gateway) feels reasonable, but introducing a limited form of inheritance on direct policies where targeting both a Gateway and a specific Listener on it _without_ being able to tweak advanced details like merge strategy feels like a mistake.

IMO `sectionName` actually fits relatively neatly and clearly into the existing hierarchy as:
- GatewayClass
- Gateway Namespace
- Gateway
- Gateway Listener (`sectionName`)
- Route Namespace
- Route
- Route Rule (`sectionName`, with #995 merged)

geps/gep-2648/index.md
mikemorris: > For example, couldn't I have a rule and a backendRef with the same name?

Hrm, I've definitely never considered this - since the backendRef is itself an object so I've only ever thought of targeting it directly with an policy targetRef to the kind/namespace/name of the ref'd object rather than scoping the targeting to only via it's backend role within a specific HTTPRoute rule.

> I believe the same name can be used in different stanzas that are different objects

For backendRefs yea, you could have two rules targeting the same named backendRef, which I think makes the backendRef name inappropriate as a standalone selector - scoping it _under_ an HTTPRoute rule sectionName (which I believe https://github.com/kubernetes-sigs/gateway-api/pull/2593 suggests making unique when specified) I think would allow precise targeting though, like

```
apiVersion: networking.acme.io/v1alpha2
kind: RetryPolicy
metadata:
  name: foo
spec:
  targetRef:
    kind: HTTPRoute
    name: http-app-1
    sectionName: bar
  maxRetries: 5
  backendRefs:
  - "baz"
```

geps/gep-2648/index.md
mikemorris: I don't think this should be possible/allowed - I believe `targetRef` is only intended to select a single object as defined by a unique group/kind/namespace/name/sectionName.

I think https://github.com/kubernetes-sigs/gateway-api/pull/2813#discussion_r1552455260 may propose a possible path forward for your use case though?

geps/gep-2648/index.md
```diff
@@ -0,0 +1,465 @@
...
+`sectionName` and one without, the more specific policy (i.e., the one with a
+`sectionName`) will have its entire `spec` applied to the named section.
+The less specific policy will also have its `spec` applied to the target but
+MUST not affect the named section. The less specific policy will have its `spec`
+applied to all other sections of the target that are not targeted by any other
+more specific policies.
+
+When more than one object matches the same object _and_ `sectionName`, the usual
+conflict-resolution rules (as defined in GEP-713 should be used). These boil down
+to "oldest by creation date wins".
```
mikemorris, line 194: @candita I don't think this clarification actually solves for your use case, where you're explicitly trying to match multiple objects with a given backendRef `name`.

Refs https://github.com/kubernetes-sigs/gateway-api/pull/2813#discussion_r1520616890 and https://github.com/kubernetes-sigs/gateway-api/pull/2813#discussion_r1552455260

geps/gep-2648/index.md
robscott: > controllers are supposed to update Conditions lists additively, putting back any Condition they don't understand or update as they found it

This seems to require a controller to be aware of any condition it may have theoretically set in the past instead of assuming full ownership of the conditions and replacing with the current set of conditions it knows about. I generally would prefer some form of namespacing/nesting in status as broadly as possible that allows each controller to clearly own separate lists. This PR doesn't need to solve this, but it would be nice to have some kind of note somewhere in the GEP covering this as a follow up area and/or unresolved.

mkdocs.yml
```diff
@@ -113,6 +113,8 @@ nav:
       - geps/gep-1619/index.md
       - geps/gep-1651/index.md
       - geps/gep-1867/index.md
+      - geps/gep-2648/index.md
+      - geps/gep-2649/index.md
     # - Implementable:
     #   -
```
robscott, line 119: Can't comment on the right line for this, but don't forget to move 713 down to memorandum.

robscott (approved):
Thanks for all the work on this @youngnick! 

I think this is a great update that represents where we are now, but also really appreciate you starting the broader discussion in https://github.com/kubernetes-sigs/gateway-api/discussions/2927 to cover what's next (and linking it from all the GEPs here).

geps/gep-2648/index.md
youngnick: I added a "SHOULD NOT use `backendRef`" above, which should clear this up, I think.

geps/gep-2648/index.md
youngnick: It's actually part of the contract of upstream Conditions that controllers don't touch Conditions they don't know about - precisely to solve this sort of case. I agree that namespacing or nesting is better, but we can't always rely on that, especially if you're affecting some Core object that doesn't have a namespaced Conditions list.

geps/gep-2648/index.md
youngnick: I added some text above to make this clearer.

geps/gep-2648/index.md
guicassolato: +1 to not using `sectionName` for `backendRef`, and to "section of a HTTPRoute" == a HTTPRouteRule.

mikemorris: FWIW the need for "multi-dimensional targeting" isn't unique to BackendTLSPolicy - I hope that expecting all Gateways supporting BackendTLSPolicy to respect a policy applied to a Service for any backendRefs targeting it can be sufficient to start with, but scoping a policy to more limited `from` clauses is also a pattern we'll need to figure out if we want to have AuthorizationPolicy for mesh too (e.g. apply policy to targetRef `destination` and scope ALLOW rules as `from` potentially multiple `origin` sources, similar to the structure of https://istio.io/latest/docs/reference/config/security/authorization-policy/).

mikemorris (approved):
Appreciate all the work you've put into this @youngnick, and sorry I haven't been more involved earlier to help support! This feels like it's in adequate shape to get in as provisional while we continue to work on refining the structure over in https://github.com/kubernetes-sigs/gateway-api/discussions/2927

geps/gep-2649/index.md
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/discussions/2927#discussioncomment-9024156 to continue this discussion.

geps/gep-2648/index.md
candita: @robscott 

> I think sectionName would be used exclusively to refer to rule names, and that rule names would need to be unique within a Route when they were specified. 

Ok, I wasn't aware that we had clearly defined `sectionNames` for `HTTPRoute` or any other object.  Where do we document that information and more importantly, how is the constraint enforced?

geps/gep-2648/index.md
candita: @mikemorris I wanted information on what was considered a `SectionName` in the object.  I [heard from Rob that we have known `SectionNames` for each object.](https://github.com/kubernetes-sigs/gateway-api/pull/2813#discussion_r1554111330)  I do hope we have this documented somewhere.

geps/gep-2648/index.md
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/pull/2946 to improve documentation for `SectionName`

geps/gep-2648/index.md
mikemorris: Opened https://github.com/kubernetes-sigs/gateway-api/pull/2946 to improve documentation for SectionName

shaneutt (approved):
/lgtm

mkdocs.yml
```diff
@@ -115,10 +115,12 @@ nav:
...
     - Experimental:
-      - geps/gep-713/index.md
+      - geps/gep-957/index.md
```
gcs278, line 123: I think this was some sort of merge conflict with https://github.com/kubernetes-sigs/gateway-api/pull/2777?

I put up https://github.com/kubernetes-sigs/gateway-api/pull/2971 to fix.

robscott: Thanks for catching and fixing this!

PR #2814, opened 2024-02-26
Conformance test for HTTPRoute weights

dprotaso:
**What type of PR is this?**
/kind cleanup
/kind test
/area conformance

There currently isn't a conformance test for weighted routing. This is a core feature for HTTPRoute.

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-weight.go
+++ b/conformance/tests/httproute-weight.go
@@ -17,13 +17,15 @@ limitations under the License.
 package tests
 
 import (
+	"cmp"
+	"errors"
 	"fmt"
 	"math"
+	"slices"
 	"strings"
 	"sync"
 	"testing"
 
-	"github.com/stretchr/testify/require"
 	"golang.org/x/sync/errgroup"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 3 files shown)


conformance/tests/httproute-weight.go
```diff
@@ -0,0 +1,138 @@
...
+			gwAddr  = kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		)
+
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		t.Run("Requests should have a distribution that matches the weight", func(t *testing.T) {
+			const (
+				totalRequests       = 1000.0
+				concurrentRequests  = 10
+				tolerancePercentage = 5.0
```
alexsnaps, line 61: One possible way of somewhat addressing the flakiness, if I understood the problem right, would be to not necessarily expect that the expected outcome happens after 1000 requests, but instead keep on going for `n` amount of requests and consider the test passing, if the observed weights are within the tolerance level within those follow up request _at one point_. 

I understand this wouldn't completely eliminate any chances of failure, but it seems reasonable that e.g. between 1000 and 2000 requests, at one point the percentage is actually split somewhat correctly as per the weight between the different backends. Anyways thought of that during the meeting, so sharing it here.

robscott: I actually really like the idea @alexsnaps had. Maybe we could have a range like 100-2000 requests, and every 100 requests we check to see if the desired distribution has been achieved (within the configured tolerance). If it has, we consider the test passing and move on, if not, we wait until it has passed or we've hit the max number of requests.

I'd also be interested in potentially increasing the tolerance here to something like 10% just to further reduce the probability of flakes.

dprotaso: Can someone shed more light why/when an implementations distribution would change over time that would warrant such a windowing acceptance test?

Also a follow up what's the issue with having a larger sample size and tolerance instead?

alexsnaps: I can't point to a particular implementation, but based on my experience doing similar stuff, you generally avoid any coordination routing these streams, which such a perfectly accurate balancing would require. e.g. a shared counter across multiple threads is something you don't want to do, other than _absolutely_ required. The idea is to favor parallelism over concurrency. That point of coordination would end up being your bottleneck. And when you think about it, what does it even mean to _balance perfectly_ when doing things in parallel...? 
Such an approach, testing that you eventually observe a somewhat acceptable balance over a set time/request frame, is how I went about it in the past. I like the point of view you defended, from a user reading these tests... That being said, I also think that test that are reliable are equally important. Possibly more so for conformance test, excuse my ignorance here, but can we "loosen" the contract after the fact on these conformance tests? Also are they ever "frozen", possibly wrt to a specific spec version?

dprotaso: >  e.g. a shared counter across multiple threads is something you don't want to do, other than absolutely required

This sounds like a weird implementation decision - wouldn't it be simpler to use a reliable random number generator for traffic splitting and that would be more reliable than counters?

robscott: I think what we're suggesting is actually very similar to how most of our Gateway conformance tests work. For example, when checking for a condition to exist in status, we check every n seconds with a fixed timeout. We don't want to wait forever for implementations that immediately set conditions as that would unnecessarily slow down their test runs, but we also need conformance tests to work for implementations that take longer to set conditions. 

In my opinion this is just another example of the expectation we have in Gateway conformance that the results here will be eventually consistent.

dprotaso: > In my opinion this is just another example of the expectation we have in Gateway conformance that the results here will be eventually consistent.

Those expectations exists because there is propagation delay. eg. latency for resources to reconcile and network programming to roll out. 

I've accounted for that prior to starting the distribution tests by ensuring making the request returns a consistent response - here https://github.com/kubernetes-sigs/gateway-api/pull/2814/files#diff-c9bcca12b7c5753736b6f27c6a1ee546b3ea116f402da71f105564d586a8d990R81-R82

After that point I'd assume things are 'stable' to perform the sampling should return the desired results.

robscott: Sure, but I think what @alexsnaps and I are getting at is that most implementations would pass these tests after the very first check _most of the time_, but that a probability based LB is never going to be able to guarantee that it can pass 100% of the time. This kind of approach of starting with a small number of requests allows us to pass/complete the test early 99% of the time but also helps avoid the risk of flakes by supporting additional requests to increase the likelihood that even a probability based LB will meet the desired distribution.

dprotaso: But why not accomodate the long tail here by increasing the error tolerance from say 5% to 10%?

dprotaso:
rebased

conformance/tests/httproute-weight.go
robscott: We should probably do both. The thing I'm trying to get at is that most implementations are going to pass this test after a small number of requests _most_ of the time. We should pass early when that happens, but continue to retry if it doesn't, similar to other conformance tests.

dprotaso:
adding approvers and reviewers for input here on approach - see the discussion here - https://github.com/kubernetes-sigs/gateway-api/pull/2814#discussion_r1503473000

conformance/tests/httproute-weight.go
dprotaso: I updated the test - it will now attempt the distribution test 10 times before failing

dprotaso:
Conformance reviewers/approvers - Can I get review please @arkodg @LiorLieberman @michaelbeaumont @mlavacca @sunjayBhatia @Xunzhuo

conformance/tests/httproute-weight.go
```diff
@@ -0,0 +1,157 @@
...
+			// Assert request succeeds before doing our distribution check
+			http.MakeRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, expected)
+
+			for i := 0; i < 10; i++ {
+				if err := testDistribution(t, suite, gwAddr, expected); err != nil {
+					t.Logf("Traffic distribution test failed (%d/10): %s", i+1, err)
+				} else {
+					return
+				}
+			}
```
robscott, line 75: Nit: I think it would be better if this was a cumulative total. For example, if we were targeting an equal distribution between backends, and the first iteration send 40% to A and 60% to B while the second flipped those (60% to A and 40% to A), I think we'd want that test to pass.

LiorLieberman: Interesting. Why? Do implementations commonly implement weights in a cumulative way? 

Like I think if one iteration got 10% and 90% and the other one got 90% and 10% not sure we are fine with that.

LiorLieberman (commented):
Looks good to me, thanks Dave! 

Although it is not a user facing change, I think one sentence in the release note section could still fit in to represent this. It is an important contribution. But I am not sure whats the maintainers preference.

arkodg (approved):
/approve

thanks !

dprotaso:
There's a go.mod change so this requires top level maintainers to approve

/unassign @sunjayBhatia @arkodg @mlavacca

dprotaso:
error seems unrelated to PR

>  level=error msg="Running error: context loading failed: no go files to analyze"
> [0;31mTest FAILED: hack/../hack/verify-golint.sh 

/retest

dprotaso:
Need to merge/rebase `main`

dprotaso:
rebased and ran `go mod tidy`

mlavacca (approved):
Thanks @dprotaso!

/lgtm

PR #2819, opened 2024-02-26
HTTPBackendRef RequestMirror Conformance Test

dprotaso:
### What type of PR is this?

/kind cleanup
/kind test
/area conformance

### What this PR does / why we need it

There's currently no conformance test for this feature which is 'Extended'

### Additional notes
I dropped path logic in filter so we can re-use the same test for the backend

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-request-mirror.yaml
+++ b/conformance/tests/httproute-request-mirror.yaml
@@ -7,7 +7,38 @@ spec:
   parentRefs:
   - name: same-namespace
   rules:
-  - filters:
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /mirror
+    filters:
+    - type: RequestMirror
+      requestMirror:
+        backendRef:
+          name: infra-backend-v2
+          namespace: gateway-conformance-infra
+          port: 8080
+    backendRefs:
+    - name: infra-backend-v1
+      port: 8080
+      namespace: gateway-conformance-infra
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /mirror-and-modify-headers
+    filters:
+    - type: RequestHeaderModifier
+      requestHeaderModifier:
+        set:
+        - name: X-Header-Set
+          value: set-overwrites-values
+        add:
+        - name: X-Header-Add
+          value: header-val-1
+        - name: X-Header-Add-Append
+          value: header-val-2
+        remove:
+        - X-Header-Remove
     - type: RequestMirror
       requestMirror:
         backendRef:
--- a/conformance/tests/httproute-request-mirror-backend.yaml
+++ b/conformance/tests/httproute-request-mirror-backend.yaml
@@ -7,14 +7,14 @@ spec:
   parentRefs:
   - name: same-namespace
   rules:
-  - filters:
-    - type: RequestMirror
-      requestMirror:
-        backendRef:
-          name: infra-backend-v2
-          namespace: gateway-conformance-infra
-          port: 8080
-    backendRefs:
+  - backendRefs:
     - name: infra-backend-v1
-      port: 8080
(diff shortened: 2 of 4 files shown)


dprotaso:
adding approvers

adding reviewers

conformance/tests/httproute-request-mirror.yaml
```diff
@@ -7,38 +7,7 @@ spec:
...
-          port: 8080
-    backendRefs:
-    - name: infra-backend-v1
-      port: 8080
-      namespace: gateway-conformance-infra
-  - matches:
-    - path:
-        type: PathPrefix
-        value: /mirror-and-modify-headers
-    filters:
```
arkodg, line 29: why rm this case thats testing the filter combo ?

dprotaso: For BackendRef modifying headers and mirroring are two separate features

dprotaso:
I've added back all the test cases for the regular filter. The backend filter has a subset of these cases.

arkodg:
has this been tested against any implementation ? afaik this _easily_ possible on envoy based implementations

dprotaso:
> has this been tested against any implementation ? 

I ran these tests against Istio. The backendRef variant fails as expected because they don't support that feature.

arkodg:
imo lets add conformance tests if at least 1 implementation supports this case.

robscott:
> imo lets add conformance tests if at least 1 implementation supports this case.

+1, I think it's pretty difficult to add conformance tests before at least one implementation supports a feature. If we were to do that, I think we'd need to introduce some concept of conformance test stability and consider untested conformance tests as a non-blocking reference point.

dprotaso:
> imo lets add conformance tests if at least 1 implementation supports this case.

But then an implementation can't advertise the support this feature on the GatewayClass. They'll have to wait for a new gateway api release to do this.

> +1, I think it's pretty difficult to add conformance tests before at least one implementation supports a feature. If we were to do that, I think we'd need to introduce some concept of conformance test stability and consider untested conformance tests as a non-blocking reference point.

Then why is this functionality in the API?

dprotaso:
Also the test is identical to the pre-split filter - so for this case I wouldn't be concerned with stability of the conformance test.

mlavacca:
> > imo lets add conformance tests if at least 1 implementation supports this case.
> 
> But then an implementation can't advertise the support this feature on the GatewayClass. They'll have to wait for a new gateway api release to do this.
> 
> > +1, I think it's pretty difficult to add conformance tests before at least one implementation supports a feature. If we were to do that, I think we'd need to introduce some concept of conformance test stability and consider untested conformance tests as a non-blocking reference point.
> 
> Then why is this functionality in the API? If we have the shape of feature in our API we can write conformance for it

I +1 what @arkodg and @robscott said about waiting to have at least one implementation on board with this feature. In my opinion, conformance tests are not intended to shape the API, but instead to validate API implementations. Regardless of the nature of the test (this test could be straightforward, as it shares the same logic of the `requestMirror` feature), we should wait for at least one implementation to properly run the tests, otherwise, the risk of having tests which are just a bunch of untested code is pretty high (I'm not saying it's the case of this specific PR, my concern is about the approach we want to have).

mlavacca:
In a backlog grooming session, we went through this issue and discussed the need to define the status of conformance tests (experimental/GA). Shane is going to take a look into it.

/assign @shaneutt

shaneutt (commented):
@dprotaso Thanks for adding this!

In general, I think this looks good:

/approve

However, one thing that does give me some pause: We haven't really seen any support from other implementations saying that this test makes sense and works for them?

candita:
ping @dprotaso

PR #2823, opened 2024-02-27
Timeouts - add test case for disabling a timeout

dprotaso:
**What type of PR is this?**
/kind cleanup
/kind test
/area conformance

This tests includes tests cases for when the timeout is explicitly disabled 

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-timeout-backend-request.go
+++ b/conformance/tests/httproute-timeout-backend-request.go
@@ -59,6 +59,10 @@ var HTTPRouteTimeoutBackendRequest = suite.ConformanceTest{
 				Request:   http.Request{Path: "/backend-timeout?delay=1s"},
 				Response:  http.Response{StatusCode: 504},
 				Namespace: ns,
+			}, {
+				Request:   http.Request{Path: "/disable-backend-timeout?delay=2s"},
+				Response:  http.Response{StatusCode: 200},
+				Namespace: ns,
 			},
 		}
 
--- a/conformance/tests/httproute-timeout-backend-request.go
+++ b/conformance/tests/httproute-timeout-backend-request.go
@@ -60,7 +60,7 @@ var HTTPRouteTimeoutBackendRequest = suite.ConformanceTest{
 				Response:  http.Response{StatusCode: 504},
 				Namespace: ns,
 			}, {
-				Request:   http.Request{Path: "/disable-backend-timeout?delay=2s"},
+				Request:   http.Request{Path: "/disable-backend-timeout?delay=1s"},
 				Response:  http.Response{StatusCode: 200},
 				Namespace: ns,
 			},
--- a/conformance/tests/httproute-timeout-backend-request.yaml
+++ b/conformance/tests/httproute-timeout-backend-request.yaml
@@ -16,3 +16,12 @@ spec:
       port: 8080
     timeouts:
       backendRequest: 500ms
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /disabled-backend-timeout
+    backendRefs:
+    - name: infra-backend-v1
+      port: 8080
+    timeouts:
+      backendRequest: "0s"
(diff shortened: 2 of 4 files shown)


dprotaso:
cc @frankbu

conformance/tests/httproute-timeout-backend-request.go
```diff
@@ -59,6 +59,10 @@ var HTTPRouteTimeoutBackendRequest = suite.ConformanceTest{
 				Request:   http.Request{Path: "/backend-timeout?delay=1s"},
 				Response:  http.Response{StatusCode: 504},
 				Namespace: ns,
+			}, {
+				Request:   http.Request{Path: "/disable-backend-timeout?delay=2s"},
```
frankbu, line 63: nit: maybe make the delay just 1s (or even less)? That will also prove that "0s" == no timeout, but speeds up the tests.

frankbu (approved):
Thanks @dprotaso! One small nit, but otherwise LGTM.

robscott:
@frankbu do you know if Istio is already implementing this? If so, do you mind confirming if it actually passes this test?

conformance/tests/httproute-timeout-backend-request.go
frankbu: > I also wanted the timeout to be > than the delay from the other test case - to ensure there is isolation between the rules.

I don't see how 2s does that any better than 1s ... anything greater than 500ms will do that.

I still would prefer the timeout be reduced from 2s -> 1s, to eliminate an extra couple of seconds of unnecessary delay (2 tests x 1s) that doesn't prove anything extra. I think we should try to minimize unnecessary delays wherever we can because they do add up.

frankbu:
> @frankbu do you know if Istio is already implementing this? If so, do you mind confirming if it actually passes this test?

Yep, it's implemented in Istio and the test passes.

conformance/tests/httproute-timeout-backend-request.go
dprotaso: 1s sounds good

dprotaso:
@frankbu can you test again with the smaller timeout

frankbu:
> @frankbu can you test again with the smaller timeout

@dprotaso  Yes, I already did ... works fine. Thanks for changing it!

conformance/tests/httproute-timeout-request.go
```diff
@@ -59,6 +59,10 @@ var HTTPRouteTimeoutRequest = suite.ConformanceTest{
 				Request:   http.Request{Path: "/request-timeout?delay=1s"},
 				Response:  http.Response{StatusCode: 504},
 				Namespace: ns,
+			}, {
+				Request:   http.Request{Path: "/disable-request-timeout?delay=2s"},
```
frankbu, line 63: Change this one to 1s too?

```suggestion
				Request:   http.Request{Path: "/disable-request-timeout?delay=1s"},
```

dprotaso: oh whoops I changed the wrong one - do you want both of them to be 1s or just the new case I added?

frankbu: Both new tests with 1s delay are tested and working fine with Istio.

dprotaso:
adding approvers

adding reviewers

sunjayBhatia (approved):

dprotaso:
Conformance reviewers/approvers - can I get review please @arkodg @mlavacca @LiorLieberman @Xunzhuo

arkodg (approved):
/approve

thanks !

arkodg (commented):
/lgtm

conformance/tests/httproute-timeout-backend-request.yaml
```diff
@@ -16,3 +16,12 @@ spec:
       port: 8080
     timeouts:
       backendRequest: 500ms
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /disabled-backend-timeout
```
sayboras, line 22: The test request is with `/disable-backend-timeout?delay=1s`, but path prefix here is `/disabled-backend-timeout`

There is discrepancy between test request path `/disable-backend-timeout?delay=1s` and path prefix in http route (e.g. `/disabled-backend-timeout`), which will cause 404 response instead of 200.

PR #2824, opened 2024-02-27
Conformance tests for HTTPRoute+Gateway HTTPS Listener

dprotaso:
**What type of PR is this?**
/kind cleanup
/kind test
/area conformance

There's no 'Core' conformance tests that validates the combination of a HTTPRoute with a Gateway HTTPS Listener.

There are other tests but they're 'Extended' and behind various feature flags

N/A

```release-note
NONE
```


sunjayBhatia (approved):
test passes w/ contour 👍🏽 

looks like the gateway in use `same-namespace-with-https-listener` is only ever used otherwise in `HTTPRouteRedirectPortAndScheme` which as you say is blocked from being generally part of core since it has other extended features required

conformance/tests/httproute-tls-listener.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+	"sigs.k8s.io/gateway-api/conformance/utils/tls"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, HTTPRouteTLSListener)
+}
+
+var HTTPRouteTLSListener = suite.ConformanceTest{
+	ShortName:   "HTTPRouteTLSListener",
```
robscott, line 35: The name of this test is very confusing to me, I think this is actually a HTTPS listener right? Just trying to confirm because we don't actually support TLS -> HTTPRoute per https://gateway-api.sigs.k8s.io/v1alpha2/guides/tls/#clientserver-and-tls.

dprotaso:
Yeah it’s an HTTPS listener. Can you suggest a better name?

conformance/tests/httproute-tls-listener.go
dprotaso: got it - i didn't realize HTTPS was a distinct protocol on the listener

dprotaso: renamed the test

conformance/tests/httproute-https-listener.go
```diff
@@ -0,0 +1,64 @@
...
+
+		certNN := types.NamespacedName{Name: "tls-validity-checks-certificate", Namespace: ns}
+		cPem, keyPem, err := GetTLSSecret(suite.Client, certNN)
+		if err != nil {
+			t.Fatalf("unexpected error finding TLS secret: %v", err)
+		}
+
+		t.Run("Simple HTTPS request should reach infra-backend", func(t *testing.T) {
+			tls.MakeTLSRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, cPem, keyPem, "example.org", http.ExpectedResponse{
+				Request:   http.Request{Host: "example.org", Path: "/"},
```
michaelbeaumont, line 57: Should we add a negative test that uses a different SNI value here? And I guess the next step would be to add a Gateway HTTPS listener with an explicit `hostname`, which I _think_ we don't have yet.

dprotaso: Do you prefer that to be done this PR or a follow up?

michaelbeaumont: Whichever you prefer, it's not that important I suppose. I guess the negative test should be easy enough to add but the listener + hostname can definitely wait.

michaelbeaumont (approved):
LGTM, passes with Kuma

conformance/tests/httproute-https-listener.go
dprotaso: Added the additional test cases - please take a look again.

michaelbeaumont: Nice, passes with Kuma

michaelbeaumont (approved):

conformance/tests/httproute-https-listener.go
```diff
@@ -0,0 +1,91 @@
...
+				Response:  http.Response{StatusCode: 200},
+				Backend:   "infra-backend-v1",
+				Namespace: "gateway-conformance-infra",
+			})
+		})
+		t.Run("Unknown host should return an 404", func(t *testing.T) {
+			tls.MakeTLSRequestAndExpectEventuallyConsistentResponse(t, suite.RoundTripper, suite.TimeoutConfig, gwAddr, cPem, keyPem, "unknown-example.org", http.ExpectedResponse{
+				Request:  http.Request{Host: "second-example.org", Path: "/"},
```
frankbu, line 68: I don't think it checks the Host, in the 404 case, so it doesn't matter what the Host value is here (maybe Host is not needed here at all?), but it's kind of confusing that it's not the same as the server arg value (unknown-example.org) which is the host it's actually calling.
```suggestion
				Request:  http.Request{Host: "unknown-example.org", Path: "/"},
```

conformance/tests/httproute-https-listener.go
```diff
@@ -0,0 +1,91 @@
...
+				Response: http.Response{StatusCode: 404},
+			})
+		})
+
+		t.Run("Gateway with hostname", func(t *testing.T) {
+			t.Run("Simple HTTPS request should reach infra-backend", func(t *testing.T) {
```
michaelbeaumont, line 74: Again, feel free to handle in an additional PR but could we route to a different backend in the `second-example.org` case? To ensure that the first listener isn't being matched.

dprotaso: Sure

conformance/tests/httproute-https-listener.go
dprotaso: Request is verified when status code is 200. Otherwise this field is a noop.

I've simplified the case setup to be more clear.

conformance/tests/httproute-https-listener.go
dprotaso: done

dprotaso:
# approvers

# reviewers

michaelbeaumont (approved):

dprotaso:
Conformance reviewers/approvers - can I get review please @arkodg @mlavacca @LiorLieberman @Xunzhuo

arkodg (approved):
/approve

thanks !

dprotaso:
@sunjayBhatia @mlavacca can you add an `/lgtm`

conformance/tests/httproute-https-listener.go
```diff
@@ -0,0 +1,81 @@
...
+
+		cases := []struct {
+			host       string
+			statusCode int
+			backend    string
+		}{
+			{host: "example.org", statusCode: 200, backend: "infra-backend-v1"},
+			{host: "unknown-example.org", statusCode: 404},
+			{host: "second-example.org", statusCode: 200, backend: "infra-backend-v2"},
+			{host: "unknown-example.org", statusCode: 404},
```
sunjayBhatia, line 66: repeat of line 64?

Issue #1718, opened 2023-02-08
Test different variants of K8s Services in conformance testing


dprotaso:
**What would you like to be added**:

We should add conformance tests that ensure Gateway API works well with different variants of K8s Services, Endpoint, EndpointSlices

Short List Variants
- Service - with selector
- Service - with selector, [endpoint slice mirror disabled](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#endpointslice-mirroring)
- Headless Service - with manual endpoints (and their different types)
- Headless Service - with manual endpoint slices (and their different types - ie. FQDN)

**Why this is needed**:

Different platforms that build on-top of K8s will utilize Services in different ways to suit their needs - it would be great to know Gateway conformance  supports these variants. ie. some implementations currently don't track EndpointSlices which is pretty fundamental as they've been stable for numerous years

dprotaso:
Related - https://github.com/kubernetes-sigs/gateway-api/pull/1816

@shaneutt Given `ExternalName` support is marked as `SHOULD NOT` - does it still warrant having an opt-in test? The purpose being for implementations to do support that type they'll have a way to verify it

shaneutt:
> Related - #1816
> 
> @shaneutt Given `ExternalName` support is marked as `SHOULD NOT` - does it still warrant having an opt-in test? The purpose being for implementations (that support this type) having a common way to verify their support.

Because we consider it `Implementation-specific` it sorta suggests the _opposite_ of conformance: implementations can choose to implement that functionality in a way they deem appropriate and safe. I'm not saying that we should _never_ have conformance tests that cover `Implementation-specific` behaviors, but I'm hesitant to say we need one here and wonder if that would make it feel more like "extended". What specifically did you have in mind?

dprotaso:
@Xunzhuo are you still working on this?

dprotaso:
I've created a first pass of a test here - https://github.com/kubernetes-sigs/gateway-api/pull/2828

Currently the test exercises the following variants:

1. Service - Headless (spec.clusterIP=None) with Selector
1. Service - Headless no selector but with a manual Endpoint (addressType: IPv4)
1. Service - Headless no selector but with a manual EndpointSlice (addressType: IPv4)
1. Service - with no selector but with a manual Endpoint (addressType: IPv4)
1. Service - with no selector but with a manual EndpointSlice (addressType: IPv4)

dprotaso:
Circling back what's done is

[...]

PR #2828, opened 2024-02-28
HTTPRoute Conformance with various service backends

dprotaso:
/kind cleanup
/kind test
/area conformance

This adds an HTTPRoute conformance (Core) to test various Service backends. 

Currently the test exercises the following variants:
1. Service - Headless (`spec.clusterIP=None`) with Selector
1. Service - Headless no selector with a manual EndpointSlice
1. Service - No selector with a manual EndpointSlice

Part of https://github.com/kubernetes-sigs/gateway-api/issues/1718

```release-note
NONE
```

### DIFF
--- a/conformance/tests/httproute-service-types.go
+++ b/conformance/tests/httproute-service-types.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"net/netip"
-	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -60,13 +59,16 @@ var HTTPRouteServiceTypes = suite.ConformanceTest{
 				"headless",
 			}
 
-			serviceTypes = slices.Concat(typeManaged, typeManualEndpointSlices)
+			serviceTypes = make([]string, 0, len(typeManualEndpointSlices)+len(typeManaged))
 
 			ctx     = context.TODO()
 			ns      = "gateway-conformance-infra"
 			routeNN = types.NamespacedName{Name: "service-types", Namespace: ns}
 			gwNN    = types.NamespacedName{Name: "same-namespace", Namespace: ns}
 		)
+
+		serviceTypes = append(serviceTypes, typeManualEndpointSlices...)
+		serviceTypes = append(serviceTypes, typeManaged...)
 
 		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
 		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
--- a/conformance/tests/httproute-service-types.go
+++ b/conformance/tests/httproute-service-types.go
@@ -21,6 +21,7 @@ import (
 	"slices"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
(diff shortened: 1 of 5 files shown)


dprotaso:
Linter is unblocked by https://github.com/kubernetes-sigs/gateway-api/pull/2829

conformance/tests/httproute-service-types.go
```diff
@@ -0,0 +1,183 @@
...
+			typeManualEndpointSlices = []string{
+				"manual-endpointslices",
+				"headless-manual-endpointslices",
+			}
+
+			typeManaged = []string{
+				"headless",
+			}
+
+			serviceTypes = slices.Concat(typeManaged, typeManualEndpoints, typeManualEndpointSlices)
```
sunjayBhatia, line 65: i think we'll need to bump the go.mod go minimum version to 1.22 since this is a 1.22 feature

conformance/tests/httproute-service-types.go
```diff
@@ -0,0 +1,183 @@
...
+			routeNN = types.NamespacedName{Name: "service-types", Namespace: ns}
+			gwNN    = types.NamespacedName{Name: "same-namespace", Namespace: ns}
+		)
+
+		gwAddr := kubernetes.GatewayAndHTTPRoutesMustBeAccepted(t, suite.Client, suite.TimeoutConfig, suite.ControllerName, kubernetes.NewGatewayRef(gwNN), routeNN)
+		kubernetes.HTTPRouteMustHaveResolvedRefsConditionsTrue(t, suite.Client, suite.TimeoutConfig, routeNN, gwNN)
+
+		deployment := &appsv1.Deployment{}
+		if err := suite.Client.Get(ctx, client.ObjectKey{Namespace: ns, Name: "infra-backend-v1"}, deployment); err != nil {
+			t.Fatal("Failed to list Deployment 'infra-backend-v1':", err)
```
sunjayBhatia, line 78: nit: could use testify helpers to simplify these

sunjayBhatia (approved):
tests all pass with Contour in endpointslice mode, just some small nits in comments, LGTM overall

I think this makes total sense to have, though I wonder if other implementations that possibly have different strategies for routing to Services will agree?

conformance/tests/httproute-service-types.go
dprotaso: good catch

sunjayBhatia (approved):

sunjayBhatia:
lgtm, i guess just gotta get CI to pass 👍🏽

dprotaso:
Rebased and added an exclude option for the filter.

The linter has a fix in a newer version but it's not part of golangci-lint (https://github.com/Antonboom/testifylint/pull/71)

dprotaso:
Need to assign gateway maintainers because of the go.mod/linter config changes

/assign @shaneutt @youngnick @robscott

conformance/tests/httproute-service-types.yaml
```diff
@@ -0,0 +1,122 @@
...
+    targetPort: 3000
+---
+apiVersion: discovery.k8s.io/v1
+kind: EndpointSlice
+metadata:
+  name: manual-endpointslices
+  namespace: gateway-conformance-infra
+  labels:
+    kubernetes.io/service-name: manual-endpointslices
+addressType: IPv4
```
robscott, line 58: This is probably not the safest assumption, you probably want to have EndpointSlices for both IP families and fill them with any addresses you find for either.

conformance/tests/httproute-service-types.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - name: first-port
+    protocol: TCP
+    port: 8080
+    targetPort: 3000
+---
+apiVersion: v1
+kind: Endpoints
+metadata:
+  name: headless-manual-endpoints
+  namespace: gateway-conformance-infra
```
robscott, line 96: In most cases this is really just testing that the EndpointSlice mirroring controller is working. I'd personally prefer skipping this as it's not really something that's testing the GW API implementation and it adds complexity here and in the corresponding go test definition.

robscott (commented):
Thanks @dprotaso!

conformance/tests/httproute-service-types.yaml
dprotaso: This is a good catch - Is there a larger issue here about IPv6 support with Gateway API? That might warrant a separate conformance test/issue/discussion?

Reading up a bit I do see that `SingleStack` is the default for a K8s Services and the ip family used is the first family type from `service-cluster-ip-range` flag.

conformance/tests/httproute-service-types.yaml
dprotaso: > In most cases this is really just testing that the EndpointSlice mirroring controller is working.

Can you elaborate?  My intent for testing headless is to ensure that implementations aren't just sending traffic to the cluster IP.

conformance/tests/httproute-service-types.yaml
robscott: We don't actually have any requirement in Gateway API that controllers support dual stack or a specific IP family, but I think our conformance tests should at least be possible to run in an IPv6-only cluster. As far as I know, this would be the first conformance test that would prevent that if we don't add EndpointSlices for both IP families.

conformance/tests/httproute-service-types.yaml
robscott: Sorry for the confusion, I'm actually referring to any tests that cover custom Endpoints instead of custom EndpointSlices. The EndpointSlice mirroring controller exists to mirror custom Endpoints into custom EndpointSlices so controllers can just read from the EndpointSlice API. That's why I think we should avoid tests that set custom Endpoints because we're really just testing the mirroring controller at that point, not Gateway controllers.

dprotaso: Do you think it's worth adding the `endpointslice.kubernetes.io/skip-mirror: true` label to these Endpoints?

robscott: No, there's no reason for controllers to watch both Endpoints and EndpointSlice APIs. Gateway API is newer than the EndpointSlice API so hopefully all implementations are reading from EndpointSlices, but they definitely shouldn't be reading from both APIs.

dprotaso: dropped the `Endpoint` cases

conformance/tests/httproute-service-types.yaml
dprotaso: Added IPv6 support

dprotaso:
Conformance reviewers/approvers - can I get review please @arkodg @mlavacca @michaelbeaumont @LiorLieberman @Xunzhuo

dprotaso:
Also @robscott to confirm the changes he requested re IP4/IP6

gwctl/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/gateway-api/gwctl
 
-go 1.21
+go 1.22
```
arkodg, line 3: is the bump needed ?

dprotaso: Using golang `slices.Concat` requires 1.22.

dprotaso: ok reverted the bump to go1.21

arkodg (approved):
/approve

thanks !

dprotaso:
Because of the `.golangci.yaml` changes this requires top level maintainers to approve 

/unassign @arkodg @mlavacca 

PR #2829, opened 2024-02-28
bump golangci-lint to latest version

dprotaso:
**What type of PR is this?**
/kind cleanup

Linter is blocking https://github.com/kubernetes-sigs/gateway-api/pull/2828

N/A

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,7 +65,7 @@ issues:
         - gocyclo
         - errcheck
         - dupl
-     # To many false positives - for examples see: https://github.com/Antonboom/testifylint/issues/67
+     # Too many false positives - for examples see: https://github.com/Antonboom/testifylint/issues/67
     - linters:
       - testifylint
       text: "require must only be used in the goroutine running the test function"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,7 @@ linters:
   - whitespace
   disable:
   - scopelint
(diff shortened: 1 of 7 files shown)


.golangci.yml
```diff
@@ -27,6 +27,7 @@ linters:
   - whitespace
   disable:
   - scopelint
+  - testifylint # To many false positives https://github.com/Antonboom/testifylint/issues/67
```
shaneutt, line 30:
```suggestion
  - testifylint # Too many false positives https://github.com/Antonboom/testifylint/issues/67
```

shaneutt: For posterity, can we enumerate some examples of the false positives in this comment for future code readers?

sunjayBhatia, line 30: another way to disable this specific set of false positives would be something like:
```
- linters: ["testifylint"]
  text: "require must only be used in the goroutine running the test function"
```

dprotaso: oh interesting - is that stable?

dprotaso: I've updated the linked issue with the false positive examples - I'd prefer people refer to that

sunjayBhatia: I'm not sure, that string was probably more recently added since that check wasn't in previous testifylint versions, there might be other ways golangcilint allows ignoring lint errors, will have to look

sunjayBhatia (approved):

robscott:
Thanks @dprotaso!

/lgtm
/approve

PR #2835, opened 2024-03-01
test: when running a specific test don't require the feature to be enabled

dprotaso:
**What type of PR is this?**
/kind test
/kind gep
/area conformance

This is a small quality of life improvment. When running a single test you have to make sure the feature is enabled which isn't necessary. With this change you can now just pass the `gateway-class` and `run-test` flag

```
CONFORMANCE_FLAGS="-gateway-class=istio -run-test=GatewayWithAttachedRoutes" make conformance
```

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/utils/suite/suite.go
+++ b/conformance/utils/suite/suite.go
@@ -233,11 +233,14 @@ func (test *ConformanceTest) Run(t *testing.T, suite *ConformanceTestSuite) {
 		t.Parallel()
 	}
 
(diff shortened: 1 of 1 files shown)


sunjayBhatia (approved):
tried it out and it works, runs a single test only 👍🏽

dprotaso:
This change is getting pulled into https://github.com/kubernetes-sigs/gateway-api/pull/2835

Issue #2795, opened 2024-02-21
Implement `gwctl describe namespaces`


gauravkghildiyal:
We ideally want to have something like this as the output for `gwctl describe namespace my-namespace`

```
Name: my-namespace
Labels: 
- key1: value1
- key2: value2
Annotations: 
- key1: value1
- key2: value2
Status: Active
DirectlyAttachedPolicies:
  TYPE                   NAME
  ----                   ----
  TimeoutPolicy.foo.com  demo-timeout-policy-on-namespace
  RetryOnPolicy.baz.com  demo-retry-policy-1
```

If it makes things easy, to start with, you can ignore the `Label`, `Annotation` and `Status` fields and do that in a follow up PR (once you are somewhat comfortable with the codebase). FYI: Adding those fields involves [adding a Namespace resource](https://github.com/kubernetes-sigs/gateway-api/blob/faca85dee9c521d05ba2de537326fd7eb05808bf/gwctl/pkg/resourcediscovery/nodes.go#L250-L262) (instead of the just the name) to the Namespace node, and then populating it within the [discoverNamespaces](https://github.com/kubernetes-sigs/gateway-api/blob/faca85dee9c521d05ba2de537326fd7eb05808bf/gwctl/pkg/resourcediscovery/discoverer.go#L228-L241)

gauravkghildiyal:
@Devaansh-Kumar -- Please use "/assign" to assign this to yourself. Do reach out if you get stuck somewhere

Devaansh-Kumar:
@gauravkghildiyal I have some confusions regarding how to phrase the command. 
1. Ideally we want to use the command in the following way: `gwctl describe namespace my-namespace`. If the user types just `gwctl describe namespace` should be describe all the namespaces available or just return an error asking to specify a namespace? And what should the behavior be if `-n` flag is used? According me having `-n` flag does not make sense.
2. For just the namespace name and DirectlyAttachedPolicies, should I create a function like `DiscoverResourcesForNamespace` in `discover.go`?

gauravkghildiyal:
@Devaansh-Kumar Those are valid questions.
1. Yes you are right, for `gwctl describe namespace`, we will describe all namespaces. I think what will help you here is that we're trying to match the behaviour of `kubectl`. Similar to `kubectl`, `-n` flag is redundant for namespaces.
2. Yes you are on the right track.

Devaansh-Kumar:
[...]

PR #2836, opened 2024-03-02
Implement basic functionality of gwctl describe namespace

Devaansh-Kumar:
**What type of PR is this?**
Add one of the following kinds:
/area gwctl
/kind feature
/cc @gauravkghildiyal 

Implement `gwctl describe namespace` command

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2795

```release-note
Implemented command `gwctl describe namespace`
```


Devaansh-Kumar:
@gauravkghildiyal I have implemented the command as close as I could according to your guidlines. 
The output for a sample test case come out as:
Screenshot from 2024-03-04 22-47-28

However I have one question about the format of output: If `kubectl apply` is used instead of `kubectl create` to create a namespace from a yaml file, then the annotation `kubectl.kubernetes.io/last-applied-configuration` is set which might not look very good in the output:
image

Do you think it is alright to leave it as it is?

gwctl/pkg/resourcediscovery/nodes.go
```diff
@@ -246,11 +247,15 @@ func (b *BackendNode) ID() backendID {
 	)
 }
 
-// HTTPRouteNode models the relationships and dependencies of a Namespace.
+// NamespaceNode models the relationships and dependencies of a Namespace.
 type NamespaceNode struct {
 	// NamespaceName identifies the Namespace.
 	NamespaceName string
```
gauravkghildiyal, line 253: Lets replace the NamespaceName with the actual `corev1.Namespace` resource. This would make this node consistent with how the other nodes are defined -- where nodes are simply wrappers over the underlying base resource. This would also mean that we don't need to define each field of the Namepace individually in the node here.

This means that now:
1. We can modify the `ResourceModel.addNamespace()` to accept `namespaces ...corev1.Namespace` and not need a separate `addNamespaceWithLabelsAnnotationsStatus`
2. Within `Discoverer.discoverNamespaces()`, we would fetch all `corev1.Namespace` resources. Then, while iterating over the existing resources, once we figure out the namespace of the resource, we call `addNamespace()` and pass in the `corev1.Namespace` that we would have already fetched.

gwctl/pkg/printer/namespace.go
```diff
@@ -0,0 +1,60 @@
...
+
+type namespaceDescribeView struct {
+	Name                     string                 `json:",omitempty"`
+	Labels                   map[string]string      `json:",omitempty"`
+	Annotations              map[string]string      `json:",omitempty"`
+	Status                   string                 `json:",omitempty"`
+	DirectlyAttachedPolicies []policymanager.ObjRef `json:",omitempty"`
+}
+
+func (nsp *NamespacesPrinter) PrintDescribeView(resourceModel *resourcediscovery.ResourceModel) {
```
gauravkghildiyal, line 24: Lets add a test for this as well. (You can use some other tests like those for gateway and httproute for examples on how they can be written)

gauravkghildiyal (commented):
This is almost there Devaansh! Just a couple of suggestions and we should be good. Thanks you!

On the topic of whether to exclude `last-applied-configuration`, I think leaving that should be just fine. We will always have the possibility of some line spreading across multiple lines (based on the width of the output screen) but that should normally not be reason enough to remove that field entirely.

gwctl/pkg/cmd/describe/describe.go
```diff
@@ -132,6 +133,18 @@ func runDescribe(args []string, params *utils.CmdParams, flags *describeFlags) {
...
 
+	case "namespace":
+		filter := resourcediscovery.Filter{}
+		if len(args) > 1 {
+			filter.Name = args[1]
+		}
+		
+		resourceModel, err := discoverer.DiscoverResourcesForNamespace(filter)
+		if err != nil {
+			panic(err)
```
gauravkghildiyal, line 144: Forgot to call this out. @pottekkat is working on fixing the panics that we currently have in the codebase. Lets try to ensure that the new code we're adding tries to properly write to the correct output stream (instead of panic'ing)

Ref https://github.com/kubernetes-sigs/gateway-api/pull/2825 for relevant changes

gwctl/pkg/resourcediscovery/nodes.go
Devaansh-Kumar: The reason I created a seperate `addNamespaceWithLabelsAnnotationsStatus` function was because in `Discoverer.discoverNamespaces()` we have `resourceModel.addNamespace()` for gateway, httproute and backends which already return string to the function, so I were to modify `ResourceModel.addNamespace()` to accept `namespaces ...corev1.Namespace` then those function calls would break.
How can I handle those functions?

gwctl/pkg/printer/namespace.go
Devaansh-Kumar: Yes I was planning to add that in a seperate PR, but I will do it in this one PR itself now.

gwctl/pkg/cmd/describe/describe.go
Devaansh-Kumar: Understood, I need to just replace `panic(err)` with 
```golang
fmt.Fprintf(os.Stderr, "relevant error message: %v\n", err)
os.Exit(1) 
``` 
wherever I have made code changes right?

Devaansh-Kumar:
@gauravkghildiyal I have made the suggested changes

gauravkghildiyal:
Thanks @Devaansh-Kumar! 

/lgtm
/approve

gauravkghildiyal:
@Devaansh-Kumar 

Please use the provided template in the PR description to unblock the automatic merge

````
<!--
If no, just write "NONE" in the release-note block below.
If yes, please enter a release note below:
-->
```release-note

```
````

gauravkghildiyal:
Looks like there's a couple of tests as well that will need to be fixed as part of the change. :) (Shouldn't likely be very problematic)

Devaansh-Kumar:
@gauravkghildiyal I have fixed the failing test in `httproute_test.go`. The `pull-gateway-api-test` is now passing.

gauravkghildiyal:
Thanks!

Please take not of https://github.com/kubernetes-sigs/gateway-api/pull/2836#issuecomment-1985157432, and also rebase

Devaansh-Kumar:
@gauravkghildiyal I have rebased and updated the PR message

PR #2845, opened 2024-03-05
conformance: Gateway Infrastructure Metadata

dprotaso:
*What type of PR is this?*

/kind test
/area conformance

*What this PR does / why we need it:*
Include a conformance test and create a feature for this experimental functionality

N/A

```release-note
NONE
```

### DIFF
--- a/conformance/tests/gateway-infrastructure-metadata.go
+++ b/conformance/tests/gateway-infrastructure-metadata.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The Kubernetes Authors.
+Copyright 2024 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/conformance/tests/gateway-infrastructure-metadata.go
+++ b/conformance/tests/gateway-infrastructure-metadata.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"sigs.k8s.io/gateway-api/conformance/utils/kubernetes"
+	"sigs.k8s.io/gateway-api/conformance/utils/suite"
+)
+
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayInfraMetadata)
+}
+
+var GatewayInfraMetadata = suite.ConformanceTest{
+	ShortName:   "GatewayInfraMetadata",
+	Description: "A Gateway should accept infrastructure metadata",
+	Features: []suite.SupportedFeature{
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

conformance/tests/gateway-infrastructure-metadata.go
```diff
@@ -0,0 +1,68 @@
...
+func init() {
+	ConformanceTests = append(ConformanceTests, GatewayInfraMetadata)
+}
+
+var GatewayInfraMetadata = suite.ConformanceTest{
+	ShortName:   "GatewayInfraMetadata",
+	Description: "A Gateway should accept infrastructure metadata",
+	Features: []suite.SupportedFeature{
+		suite.SupportGateway,
+		suite.SupportGatewayInfrastructureMetadata,
```
robscott, line 39: I don't really know how you do more than this, but it seems like this isn't really testing much. If an implementation takes no action, they'll support this. The only thing that would fail this test is if an implementation went out of their way to mark this config as rejected?

dprotaso:
apologies - unsure why prow didn't assign - trying again without the `#` comment

/cc @LiorLieberman @michaelbeaumont

dprotaso:
Conformance reviewers/approvers - can I get review please @arkodg @mlavacca @sunjayBhatia @michaelbeaumont @LiorLieberman @Xunzhuo

conformance/tests/gateway-infrastructure-metadata.go
```diff
@@ -0,0 +1,68 @@
...
+	Manifests: []string{
+		"tests/gateway-infrastructure-metadata.yaml",
+	},
+	Test: func(t *testing.T, suite *suite.ConformanceTestSuite) {
+		gwNN := types.NamespacedName{
+			Name:      "gateway-infra-metadata",
+			Namespace: "gateway-conformance-infra",
+		}
+
+		conditions := []metav1.Condition{
```
arkodg, line 50: imo I dont know this test is very useful, we ideally need a way to retrieve the metatdata from the data plane but since its unique, we are add a conformance test for it

dprotaso: > way to retrieve the metatdata from the data plane

That's implementation specific.

The test really just asserts that we're able to set this property on the `infrastructure` stanza and it's accepted by the implementation - nothing else. 

Also I think it's important to have a `Feature` enum so end users know they can use this feature.

arkodg: so the real use case is to surface the feature in the `GatewayClass` :) ?

dprotaso: It's part of it.

Like as an end-user I want to know if a class supports this feature. Likewise if an implementation advertises that then I expect I can set these attributes and the gateway is accepted and programmed.

robscott: I'm not sure about this one, it's not necessarily wrong, just also not clear that it provides much value. Just being able to set a value doesn't really mean it's not supported, it may just mean the Gateway implementation hasn't gone out of it's way to block this from being set. I know we don't really have strong guidelines for what does or doesn't fit as a conformance test, but this feels like it's on the edge, and I'd lean to say that it doesn't really provide sufficient value.

It seems conceptually similar to confirming that an implementation supports setting TLS options to some arbitrary value, I'm not sure how useful a test like this is without also confirming the desired result.

dprotaso: I'm fine with dropping this conformance tests - I guess I'm trying to answer the question

> How do I know which implementations support setting this property and won't fail.

Is the conformance report a good place for that?

dprotaso: Maybe PR'ing the feature string makes sense and then we enable report generation to include additional features that don't have tests?

mlavacca (commented):
+1 on the doubts already posed by @robscott and @arkodg. I do not see this test very useful, and also I think we should not include it just for the sake of having the feature listed in the GatewayClass status, which is a matter of a completely different GEP and feature, still experimental. On the other hand, I do not see a way to improve the test to increase its meaningfulness, as the feature relies very much on the implementation details.

dprotaso:
Sounds good everyone

PR #2853, opened 2024-03-09
Add doc and httproute example for HTTP request mirroring

### LINKED ISSUE #2833: Documentation and examples needed for HTTP Request Mirroring
Although we have [conformance test coverage](https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/tests/httproute-request-mirror.yaml) for request mirroring, we don't have any corresponding documentation. This would fit well as a guide, similar to the existing [redirect one](https://gateway-api.sigs.k8s.io/guides/http-redirect-rewrite/). 

**Why this is needed**:
This is an important feature of the API and it's currently completely missing from our documentation.

sawsa307:
* Added example HTTPRoute yaml for configuring HTTP request mirroring.

**What type of PR is this?**
/kind documentation

Added examples for HTTP request mirroring.

Fixes #2833

```release-note
None
```

### DIFF
--- a/site-src/guides/http-request-mirroring.md
+++ b/site-src/guides/http-request-mirroring.md
@@ -1,0 +1,27 @@
+# HTTP request mirroring
+
+??? example "Extended Support Feature"
+
+    As of v1.0.0, the Request Mirroring feature is an Extended feature, and
+    requires implementations to support the `HTTPRouteRequestMirror` feature.
+
+The [HTTPRoute resource](/api-types/httproute) allows you to mirror HTTP
+requests to another backend using
+[filters](/api-types/httproute#filters-optional). This guide shows how to use
+this feature.
+
+Mirrored requests will must only be sent to one single destination endpoint
+within this backendRef, and responses from this backend MUST be ignored by
+the Gateway.
+
+Request mirroring is particularly useful in blue-green deployment. It can be
+used to assess the impact on application performance without impacting
+responses to clients in any way.
+
+```yaml
+{% include 'standard/http-request-mirroring/httproute-mirroring.yaml' %}
+```
+
+In this example, all requests are forwarded to service `foo-v1` on port `8080`,
+and they are also forwarded to service `foo-v2` on port `8080`, but responses
+are only generated from service `foo-v1`.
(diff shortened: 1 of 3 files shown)


site-src/guides/http-request-mirroring.md
```diff
@@ -0,0 +1,22 @@
+# HTTP request mirroring
```
youngnick, line 1: Would be able to put a notice at the top of this file that Request Mirroring currently has Extended support and requires the `HTTPRouteRequestMirror` to be supported on the implementation?

Maybe something like this:
```suggestion
# HTTP request mirroring

??? example "Extended Support Feature"

    As of v1.0.0, the Request Mirroring feature is an Extended feature, and requires implementations to support
    the `HTTPRouteRequestMirror` feature.
```

youngnick (commented):
This looks great @sawsa307, thanks! Just one small request and one question.

The request is in the below comment, but the question is: Have you validated this config against an implementation that passes the conformance test? Seems like Cilium, Contour, Envoy Gateway, Istio, Kong, Kuma, and Nginx all support it. (You can tell by searching the `conformance/reports/v1.0` directory for `HTTPRouteRequestMirror` )

site-src/guides/http-request-mirroring.md
sawsa307: Hi Nick thanks for the comment!
Yes I have tested that this works on istio and it worked. Let me know if you would like the result to the PR description!

youngnick: If you use the same formatting I did, it will appear as a neat expandable headline, like at the top of https://gateway-api.sigs.k8s.io/guides/backend-protocol/ . Not a blocker though.

youngnick:
This LGTM but I'll wait for @robscott or @howardjohn to give the final LGTM.

sawsa307:
> This LGTM but I'll wait for @robscott or @howardjohn to give the final LGTM.

Thanks for reviewing! I just added the route to the new page in mkdoc.yml and verified that it works via `mkdocs serve`.

robscott:
Thanks @sawsa307!

/lgtm

Issue #2851, opened 2024-03-08
Promote the experimental conformance test suite to standard


mlavacca:
**What would you like to be added**:

The experimental conformance test suite is going to be promoted to standard. In order to do so, we need to:
- remove all the `experimental` keywords used in the suite code
- replace the old standard suite with the new one

PR #2864, opened 2024-03-12
Add support for end-users to create their own conformance profiles

dprotaso:
**What type of PR is this?**

/kind test
/kind cleanup
/area conformance

Advanced end users would like to create their own conformance profiles out of tree. This makes it easier to validate an implementation supports a certain set of features without relying on manually reading conformance reports on the gateway API website.

With this change writing your own profile to run is as simple as creating a single `_test.go` file 

```go
package gatewayapi

import (
	"testing"

	"sigs.k8s.io/gateway-api/conformance"
	"sigs.k8s.io/gateway-api/conformance/utils/suite"

	"k8s.io/apimachinery/pkg/util/sets"
)

// KnativeConformanceProfile is a ConformanceProfile that covers testing Gateway API features
// that Knative require
var (
	KnativeConformanceProfile = suite.ConformanceProfile{
		Name: "KNATIVE",
		CoreFeatures: sets.New(
			// Core HTTPRoute Features
			suite.SupportGateway,
			suite.SupportReferenceGrant,
			suite.SupportHTTPRoute,

			// HTTPRoute Extended Features
			suite.SupportHTTPRouteHostRewrite,
			// suite.SupportHTTPRouteBackendRequestHeaderModification, // https://github.com/kubernetes-sigs/gateway-api/pull/2820

			// HTTPRoute Experimental Features
			suite.SupportHTTPRouteBackendProtocolH2C,
			suite.SupportHTTPRouteBackendProtocolWebSocket,

			// HTTPRoute - Optional Features Knative might care about
			suite.SupportHTTPRouteRequestTimeout,
			suite.SupportHTTPRouteSchemeRedirect,

			// suite.SupportGatewayInfrastructureMetadata, // https://github.com/kubernetes-sigs/gateway-api/pull/2845
		),
	}
)

func TestConformance(t *testing.T) {
	suite.RegisterConformanceProfile(KnativeConformanceProfile)
	conformance.RunConformance(t, nil)
}
```

N/A

```release-note
NONE
```


dprotaso:
Note I also refactored a bunch so the tests can be run downstream a bit more easily without a whole lot of boilerplate

/assign @shaneutt @mlavacca

dprotaso:
Also if we don't like this direction then I think it would be good to keep some of the refactor then it would be possible for downstream to run the test as follows

```go
func TestConformance(t *testing.T) {
	opts := conformance.DefaultOptions()
        opts.SupportedFeatures.Union(.... /* features I care about */)
	conformance.RunConformance(t, opts)
}
```

conformance/utils/suite/experimental_profiles.go
```diff
@@ -93,6 +93,14 @@ var (
 	}
 )
 
+func RegisterConformanceProfile(p ConformanceProfile) {
```
shaneutt, line 96: Public symbol needs `godoc` comment.

conformance/conformance.go
```diff
@@ -0,0 +1,97 @@
...
+
+func RunConformance(t *testing.T, opts *suite.ExperimentalConformanceOptions) {
+	if opts == nil {
+		opts = DefaultOptions(t)
+	}
+
+	if opts.ConformanceProfiles.Len() == 0 {
+		RunLegacyConformance(t, &opts.Options)
+		return
+	}
```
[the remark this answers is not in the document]
dprotaso: I'm good to error here

conformance/conformance.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func writeReport(logf func(string, ...any), report confv1a1.ConformanceReport, output string) error {
+	//nolint:musttag // the linter is complaining report doesn't have yaml tags - but it has json ones
+	rawReport, err := yaml.Marshal(report)
+	if err != nil {
+		return err
+	}
+
+	if output != "" {
```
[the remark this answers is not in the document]
dprotaso: I left this as was.

My 2cents here is I don't think we should force generating the report. For myself I only care about the `go test` output

conformance/utils/flags/experimental_flags.go
```diff
@@ -29,6 +29,7 @@ const (
 )
 
 var (
+	EndUser                    = flag.Bool("end-user", false, "Disable implementation validation checks allowing end users to run tests")
```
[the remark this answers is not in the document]
dprotaso: I'll call this `no-implementation`.

Any thoughts on just relaxing the validation to only error when generating the report?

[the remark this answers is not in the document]
dprotaso: >Any thoughts on just relaxing the validation to only error when generating the report?

Actually I'll probably just do that.

conformance/legacy.go
```diff
@@ -0,0 +1,94 @@
+/*
```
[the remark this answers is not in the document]
dprotaso: Let's do it in a follow up PR - the renaming will muddle the PR

conformance/conformance.go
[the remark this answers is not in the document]
dprotaso: Let's keep this in the interim and drop it when we remove the legacy support in a follow up

dprotaso:
Few of the comments mentioned dropping the legacy support in this PR. I suggest we do this in a follow up PR. It'll be easier to understand the changes that took place.

shaneutt (approved):
/approve
/cc @youngnick @robscott @mlavacca for `lgtm`

mlavacca (commented):
I've synced on Slack with @dprotaso, and we agreed to wait for the conformance suite promotion (https://github.com/kubernetes-sigs/gateway-api/issues/2851) to land before merging this one.

mlavacca:
/unhold

@dprotaso #2851 has been merged, hence we can move this one forward

dprotaso:
@mlavacca take a look now

conformance/conformance.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/stretchr/testify/require"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/yaml"
+)
+
+func DefaultOptions(t *testing.T) suite.ConformanceOptions {
```
mlavacca, line 41: I think DefaultOptions is a misleading name here: the options returned by this function are fully customized by the user's flags.

conformance/utils/suite/profiles.go
```diff
@@ -111,6 +111,16 @@ var (
 	}
 )
 
+// RegisterConformanceProfile allows downstream tests to register unique profiles that
+// define their own set of features
+func RegisterConformanceProfile(p ConformanceProfile) {
```
mlavacca, line 116: I do not see any reference to this function; is it actually used anywhere?

conformance/utils/suite/suite.go
```diff
@@ -126,6 +125,7 @@ type ConformanceOptions struct {
 	MeshManifests        string
 	NamespaceLabels      map[string]string
 	NamespaceAnnotations map[string]string
+	ReportOutputPath     string
```
mlavacca, line 128: This variable is neither assigned with the flag value nor used in the suite's proper field.

conformance/utils/suite/profiles.go
dprotaso: This is exported so it can be used downstream - see the PR message body you'll see the code snippet

```
suite.RegisterConformanceProfile(KnativeConformanceProfile)
```

conformance/conformance.go
dprotaso: I can add godoc to say this will parse the flags if supplied. Do you have a suggestions for a different name?

conformance/utils/suite/suite.go
dprotaso: It's populated from the flag here  - https://github.com/kubernetes-sigs/gateway-api/pull/2864/files/3ae59a7f8680f862a6e483789f8e78a2d0fd1f11#diff-eb3b4f37d6de9237b962307f425e6cf2ab994b5a107611d32c40fc5e27c7bdd4R87

And it's used in `conformance.go` where the report is written out optionally.

dprotaso: Since the suite doesn't write the report to disk I didn't think it necessary to have a property in the suite `struct`

mlavacca: 👍

conformance/conformance.go
mlavacca: What about just `SetOptions`?

dprotaso: `SetOptions` to me seems like we're setting something on a struct/object - but we're not doing that but actually creating a `suite.ConformanceOptions` struct

mlavacca: ok, let's keep `DefaultOptions`. I saw you added some comment on it, I'm good with it 👍

conformance/utils/suite/suite.go
```diff
@@ -199,10 +199,6 @@ func NewConformanceTestSuite(options ConformanceOptions) (*ConformanceTestSuite,
 		mode = options.Mode
 	}
 
-	if options.FS == nil {
-		options.FS = &conformance.Manifests
-	}
```
tao12345666333, line 204: I don't know why you removed this part, but it's a breaking change for downstream projects.

If the downstream project uses gwapi's conformance test as a library, it will cause panic after this change.

dprotaso: I moved the defaulting of the options here

https://github.com/kubernetes-sigs/gateway-api/pull/2864/files#diff-eb3b4f37d6de9237b962307f425e6cf2ab994b5a107611d32c40fc5e27c7bdd4R85

tao12345666333: @dprotaso I know, but if downstream project not use `DefaultOptions`, it will cause panic. like this one https://github.com/Kong/kubernetes-ingress-controller/actions/runs/8661628250/job/23752106390?pr=5863#step:5:290

```
--- FAIL: TestGatewayConformance (10.89s)
    --- FAIL: TestGatewayConformance/GatewayInvalidRouteKind (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12be5d0]

goroutine 1479 [running]:
testing.tRunner.func1.2({0x452c460, 0x64d1600})
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1545 +0x3f7
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1548 +0x716
panic({0x452c460?, 0x64d1600?})
	/opt/hostedtoolcache/go/1.21.6/x64/src/runtime/panic.go:920 +0x270
bytes.(*Buffer).Read(0x0, {0xc002d5a000, 0x1000, 0x7538b00?})
	/opt/hostedtoolcache/go/1.21.6/x64/src/bytes/buffer.go:319 +0x30
bufio.(*Reader).fill(0xc001565ec0)
	/opt/hostedtoolcache/go/1.21.6/x64/src/bufio/bufio.go:113 +0x29a
bufio.(*Reader).Peek(0xc001565ec0, 0x1000)
	/opt/hostedtoolcache/go/1.21.6/x64/src/bufio/bufio.go:151 +0xc7
k8s.io/apimachinery/pkg/util/yaml.GuessJSONStream({0x50173a0?, 0x0}, 0x1000)
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/util/yaml/decoder.go:374 +0x1cb
k8s.io/apimachinery/pkg/util/yaml.(*YAMLOrJSONDecoder).Decode(0xc002d63b10, {0x48a9ca0, 0xc002ba6cc0})
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.29.3/pkg/util/yaml/decoder.go:278 +0xae
sigs.k8s.io/gateway-api/conformance/utils/kubernetes.Applier.prepareResources({0x0, 0x0, {0xc0014b6570, 0x24}, {0xc000bb5080, 0x21}, {0x0, 0x0, 0x0}, {0x0, ...}, ...}, ...)
	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.1-0.20240409212030-c57c131b49f8/conformance/utils/kubernetes/apply.go:190 +0xc5
sigs.k8s.io/gateway-api/conformance/utils/kubernetes.Applier.MustApplyWithCleanup({0x0, 0x0, {0xc0014b6570, 0x24}, {0xc000bb5080, 0x21}, {0x0, 0x0, 0x0}, {0x0, ...}, ...}, ...)
	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.1-0.20240409212030-c57c131b49f8/conformance/utils/kubernetes/apply.go:254 +0x2a8
sigs.k8s.io/gateway-api/conformance/utils/suite.(*ConformanceTest).Run(0xc001565a40, 0xc001c4fd40, 0xc000eae000)
	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.1-0.20240409212030-c57c131b49f8/conformance/utils/suite/conformance.go:64 +0x706
sigs.k8s.io/gateway-api/conformance/utils/suite.(*ConformanceTestSuite).Run.func1(0x0?)
	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api@v1.0.1-0.20240409212030-c57c131b49f8/conformance/utils/suite/suite.go:390 +0x45
testing.tRunner(0xc001c4fd40, 0xc001404ee8)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1595 +0x262
created by testing.(*T).Run in goroutine 372
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1648 +0x846
FAIL	github.com/kong/kubernetes-ingress-controller/v3/test/conformance	86.039s
```

This is the configuration it uses.

https://github.com/Kong/kubernetes-ingress-controller/pull/5863/commits/150b1d4f8e5f2b07c243edb57391c2b4be95e000#diff-cd95a80b6232fe201246d9c47d7595ff74f315ba2ccd0f54eb0da46d0b9839c7

mlavacca: I agree with @tao12345666333 on this one. We cannot expect that implementation always use the `DefaultOptions` method, as the suite is often used as a library.  And in case FS is not set I think we should provide the default instead of expecting implementations to always set the default.

Issue #2870, opened 2024-03-13
Implement `gwctl describe policycrds`


gauravkghildiyal:
The output that we're expecting is something along the lines of:

```
Name:  {{Name}}
Namespace: {{Namespace}}
Labels:
  {{Labels}}
Annotations: 
  {{Annotations}}
API Version:  {{API Version}}
Kind:         {{Kind}}
Metadata:
  {{Metadata}}
Spec:
  {{Spec]}
Status:
  {{Status}}
```

Ref. https://gateway-api.sigs.k8s.io/geps/gep-2722/

PR #2822, opened 2024-02-27
Inject a fake clock during testing `gwctl get gateways`

jongwooo:
**What type of PR is this?**

**What this PR does / why we need it**:

/area gwctl
/kind test
/cc @gauravkghildiyal

Inject a fake clock during testing `gwctl get gateways`.

```diff
type GatewaysPrinter struct {
	Out   io.Writer
+	Clock clock.Clock
}
```

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubernetes-sigs/gateway-api/pull/2805#discussion_r1502192342

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2872, opened 2024-03-14
Implemented `gwctl describe policycrds`

Devaansh-Kumar:
**What type of PR is this?**
Add one of the following kinds:
/area gwctl
/kind feature
/cc @gauravkghildiyal 

Implement `gwctl describe policycrds` command

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2870

```release-note
Implemented command `gwctl describe policycrds`
```


gwctl/pkg/policymanager/manager.go
```diff
@@ -196,6 +201,36 @@ func (p PolicyCRD) IsClusterScoped() bool {
 	return p.crd.Spec.Scope == apiextensionsv1.ClusterScoped
 }
 
+func (p PolicyCRD) Spec() map[string]interface{} {
```
gauravkghildiyal, line 204: These helper functions seem to be purely concerned with printing at the moment. Lets keep them in the printer package.

gwctl/pkg/printer/policies.go
```diff
@@ -148,3 +150,65 @@ func (pp *PoliciesPrinter) PrintDescribeView(policies []policymanager.Policy) {
 		}
 	}
 }
+
+type policyCrdDescribeView struct {
+	// PolicyCrd   *apiextensionsv1.CustomResourceDefinition `json:",omitempty"`
```
gauravkghildiyal, line 155: Does this comment need to be cleaned up?

gwctl/pkg/printer/policies_test.go
```diff
@@ -354,3 +354,155 @@ timeoutpolicies.bar.com      Direct       Namespaced  5m
 		t.Errorf("Unexpected diff\ngot=\n%v\nwant=\n%v\ndiff (-want +got)=\n%v", got, want, diff)
 	}
 }
+
+func TestPolicyCrd_PrintDescribeView(t *testing.T) {
+	// fakeClock := testingclock.NewFakeClock(time.Now())
```
gauravkghildiyal, line 359: Can we clean this up if we don't need a fakeClock? Else, can we evaluate if we need a fakeClock and use it?

gwctl/pkg/printer/policies_test.go
```diff
@@ -354,3 +354,155 @@ timeoutpolicies.bar.com      Direct       Namespaced  5m
...
+					},
+				},
+			},
+		},
+	}
+
+	params := utils.MustParamsForTest(t, common.MustClientsForTest(t, objects...))
+	pp := &PoliciesPrinter{
+		Out:   &bytes.Buffer{},
+		// Clock: fakeClock,
```
gauravkghildiyal, line 442: Same comment as above.

gwctl/pkg/printer/policies_test.go
```diff
@@ -354,3 +354,155 @@ timeoutpolicies.bar.com      Direct       Namespaced  5m
...
+
+
+Name: timeoutpolicies.bar.com
+Labels:
+  gateway.networking.k8s.io/policy: direct
+APIVersion: apiextensions.k8s.io/v1
+Kind: CustomResourceDefinition
+Metadata:
+  creationTimestamp: null
+  labels:
```
gauravkghildiyal, line 484: For things like `label`, `annotations` etc which we're including outside the `metadata`, maybe we should not print them twice?

gwctl/cmd/describe.go
```diff
@@ -96,6 +96,23 @@ func runDescribe(cmd *cobra.Command, args []string, params *utils.CmdParams) {
...
 		}
 		policiesPrinter.PrintDescribeView(policyList)
+	
+	case "policycrd", "policycrds":
+		var policyCrdList []policymanager.PolicyCRD
+		if len(args) == 1 {
+			policyCrdList = params.PolicyManager.GetCRDs()
+		} else {
+			var found bool
+			policyCrd, found := params.PolicyManager.GetCRD(args[1])
```
gauravkghildiyal, line 106: For users, args[1] would usually mean the actual "name" of the CRD. It's worth noting here that the "name" of the CRD is not the `policyCrdID` which is being used as the "key" for the policyCRDs map within `manager.go`.

This means that if someone does `gwctl describe policycrds mycrd.foo.bar`, it might not work.

I think we need to modify the `GetCRD` function to accept a `crdName` and then manually search for the CRD in the map and return the right one.

gwctl/pkg/printer/policies.go
```diff
@@ -148,3 +150,65 @@ func (pp *PoliciesPrinter) PrintDescribeView(policies []policymanager.Policy) {
...
+	Name        string                 `json:",omitempty"`
+	Namespace   string                 `json:",omitempty"`
+	Labels      map[string]string      `json:",omitempty"`
+	Annotations map[string]string      `json:",omitempty"`
+	APIVersion  string                 `json:",omitempty"`
+	Kind        string                 `json:",omitempty"`
+	Metadata    map[string]interface{} `json:",omitempty"`
+	Spec        map[string]interface{} `json:",omitempty"`
```
gauravkghildiyal, line 163: Will using the concrete type here instead of `map[string]interface{}` make things simple? Any reason why we purposely want to keep this generic?

Same comment for Metadata and Status

Devaansh-Kumar: Yes, initially I used concrete types itself, but the output was not being parsed properly by `yaml` package, so I had to convert it to `map[string]interface{}`.
Should I leave it as it is or do you suggest any other alternative?

gwctl/cmd/describe.go
Devaansh-Kumar: Yes I agree with this. I will make the changes

gwctl/pkg/printer/policies_test.go
Devaansh-Kumar: Agreed, I will remove the `label` and `annotations` outside as its more difficult to remove from within `metadata`

Devaansh-Kumar:
@gauravkghildiyal I have resolved the above issues, just reconfirm this comment once: https://github.com/kubernetes-sigs/gateway-api/pull/2872#discussion_r1535330070

gwctl/pkg/printer/policies.go
gauravkghildiyal: Can you elaborate on what problem you run into if you use something like this:

```
	Metadata    *metav1.ObjectMeta                              `json:",omitempty"`
	Spec        *apiextensionsv1.CustomResourceDefinitionSpec   `json:",omitempty"`
	Status      *apiextensionsv1.CustomResourceDefinitionStatus `json:",omitempty"`
```

I think this should make things _very_ easy. Am I missing something?

gwctl/pkg/printer/policies_test.go
gauravkghildiyal: Lets try using concrete types and see if that helps ease this as well.

gwctl/pkg/printer/policies.go
```diff
@@ -148,3 +151,88 @@ func (pp *PoliciesPrinter) PrintDescribeView(policies []policymanager.Policy) {
...
+	Name        string                 `json:",omitempty"`
+	Namespace   string                 `json:",omitempty"`
+	APIVersion  string                 `json:",omitempty"`
+	Kind        string                 `json:",omitempty"`
+	Metadata    map[string]interface{} `json:",omitempty"`
+	Spec        map[string]interface{} `json:",omitempty"`
+	Status      map[string]interface{} `json:",omitempty"`
+}
+
+func (pp *PoliciesPrinter) PolicyCrd_PrintDescribeView(policyCrds []policymanager.PolicyCRD) {
```
gauravkghildiyal, line 165: This isn't a blocker, but it would be nice if we could adjust the overall evolving names of other functions as well:

```
PrintPoliciesGetView
PrintPoliciesDescribeView
PrintPolicyCRDsGetView
PrintPolicyCRDsDescribeView
```

If we do include this change, lets correspondingly change the Test function names as well :)

gwctl/pkg/printer/policies.go
Devaansh-Kumar: My mistake, I was not dereferencing the type, due to which it wasnt parsing properly. I works fine now

Devaansh-Kumar:
@gauravkghildiyal I have resolved the above issues, please review

gwctl/cmd/describe.go
```diff
@@ -95,7 +95,24 @@ func runDescribe(cmd *cobra.Command, args []string, params *utils.CmdParams) {
...
+			if !found {
+				fmt.Fprintf(os.Stderr, "failed to find PolicyCrd: %v\n", err)
+				os.Exit(1)
+			}
+			if found {
```
gauravkghildiyal, line 111: This conditional is redundant

gauravkghildiyal (commented):
Thanks @Devaansh-Kumar, almost there!

Can we try to address https://github.com/kubernetes-sigs/gateway-api/pull/2872/files#r1534775341 please?

gauravkghildiyal:
@Devaansh-Kumar I see you pushed some new changes...is https://github.com/kubernetes-sigs/gateway-api/pull/2872/files#r1534775341  still a work in progress?

/ok-to-test

Devaansh-Kumar:
> @Devaansh-Kumar I see you pushed some new changes...is https://github.com/kubernetes-sigs/gateway-api/pull/2872/files#r1534775341 still a work in progress?
> 
> /ok-to-test

@gauravkghildiyal  I have already left a comment on that thread before, I think I have already solved it but maybe you are asking for something else that I dont understand. 

You had told to not print out `Labels` and `Annotations` twice as they appear in `Metadata` as well. So I changed the code to only print `Metadata`. Am I missing something?

gauravkghildiyal:
> Am I missing something?

Can we try to get those outside Metadata please, as per the original plan. I'm hoping it won't be too challenging now (correct me if I'm wrong)

Devaansh-Kumar:
> > Am I missing something?
> 
> Can we try to get those outside Metadata please, as per the original plan. I'm hoping it won't be too challenging now (correct me if I'm wrong)

Ok, now I understood what you wanted. You want me to parse the entire `*metav1.ObjectMeta` struct and not include `label` and `annotations`, right?

Devaansh-Kumar:
@gauravkghildiyal I have tried out your suggested change here by creating a new function `ExcludeFieldsFromStruct` to remove fields from a struct. I put this in a seperate file so that we can use this in the future as well.

gwctl/pkg/printer/policies_test.go
```diff
@@ -354,3 +354,149 @@ timeoutpolicies.bar.com      Direct       Namespaced  5m
...
+
+	got := pp.Out.(*bytes.Buffer).String()
+	want := `
+Name: healthcheckpolicies.foo.com
+APIVersion: apiextensions.k8s.io/v1
+Kind: CustomResourceDefinition
+Labels:
+  gateway.networking.k8s.io/policy: inherited
+Metadata:
+  creationTimestamp: null
```
jongwooo, line 452: Consider using `fakeClock` to validate the output for `creationTimestamp`. Ref. #2822

```go
fakeClock := testingclock.NewFakeClock(time.Now())
```

gauravkghildiyal: +1

gauravkghildiyal:
@Devaansh-Kumar 

How about something like this:

```golang
type policyCrdDescribeView struct {
	Name        string                                          `json:",omitempty"`
	Namespace   string                                          `json:",omitempty"`
	APIVersion  string                                          `json:",omitempty"`
	Kind        string                                          `json:",omitempty"`
	Labels      map[string]string                               `json:",omitempty"`
	Annotations map[string]string                               `json:",omitempty"`
	Metadata    *metav1.ObjectMeta                              `json:",omitempty"`
	Spec        *apiextensionsv1.CustomResourceDefinitionSpec   `json:",omitempty"`
	Status      *apiextensionsv1.CustomResourceDefinitionStatus `json:",omitempty"`
}
.
.
.
		metadata := crd.ObjectMeta.DeepCopy()
		// Remove fields which are printed independently from the metadata.
		metadata.Name = ""
		metadata.Namespace = ""
		metadata.Labels = nil
		metadata.Annotations = nil

		views := []policyCrdDescribeView{
			{
				Name:      crd.Name,
				Namespace: crd.Namespace,
			},
			{
				APIVersion: crd.APIVersion,
				Kind:       crd.Kind,
			},
			{
				Labels:      crd.Labels,
				Annotations: crd.Annotations,
			},
			{
				Metadata: metadata,
			},
			{
				Spec: &crd.Spec,
			},
			{
				Status: &crd.Status,
			},
		}
```

I guess you won't need the `ExcludeFieldsFromStruct` then? The intent was/is:
- To avoid duplication
- And show important fields at the top level (like kubectl), instead of within metadata.

Devaansh-Kumar:
@gauravkghildiyal I have made the necessary changes, please review

gauravkghildiyal:
Thanks for getting this through @Devaansh-Kumar 

/lgtm
/approve

PR #2874, opened 2024-03-14
Conformance display profiles on gatewayapi docs

xtineskim:
**What type of PR is this?**
/kind documentation

Related to https://github.com/kubernetes-sigs/gateway-api/issues/2550. tldr, we lack a way to quickly show different implementations of the Gateway API and their features. After trying various plugins for `mkdocs`, the most convenient was a markdown table that could be sorted based on columns. Not the most ideal experience, but it's a start for showing conformance profile reports on the gateway api site
https://github.com/kubernetes-sigs/gateway-api/issues/2550 (although this will need more iteration in the future.It's a start)

Fixes #

```release-note
NONE
```


robscott:
Thanks @xtineskim!

/ok-to-test

hack/mkdocs-copy-geps.py
```diff
@@ -14,11 +14,79 @@
...
+        temp = temp.drop(["level_1"],axis=1)
+        table = table.merge(temp, how="left")
+
+    # dropping TLS supportedFeatures column since no implementation has listed any supported features
+    table = table.drop(["TLS: Supported Features"], axis=1)
+    table.rename(columns={"project":"Project", "name":"Protocol Profile","version":"Version" }, inplace=True)
+    table = table.fillna("N/A")
+    # Output markdown table
+    with open('site-src/implementation-table.md','w') as f:
+        f.write("This table is populated from the conformance reports uploaded by project implmentations.\n\n")
```
mlavacca, line 64:
```suggestion
        f.write("This table is populated from the conformance reports uploaded by project implementations.\n\n")
```

site-src/implementation-table.md
```diff
@@ -0,0 +1,13 @@
...
+| kubernetes-ingress-controller | HTTP HTTP          | v3.0.2        | ['HTTPRouteMethodMatching', 'HTTPRouteQueryParamMatching', 'HTTPRouteResponseHeaderModification']                                                                                                                                                                                                                                                                           | N/A                                                                                                                                                                                                                                                                                                                                               |
```
mlavacca, line 11: This one is kong-kubernetes-ingress-controller, Probably the logic removes the org name somehow. Other than that, the only profile (HTTP) is duplicated; my guess is that the kong ingress controller is the only implementation to have multiple reports so far, so it may be the profiles get duplicated somehow.

mlavacca (commented):
Thanks for this PR, @xtineskim!

site-src/implementation-table.md
xtineskim: good catch! the pr that formalized the reporting procedure was merged, making some changes to address them  😁

sunjayBhatia:
deploy preview table: https://deploy-preview-2874--kubernetes-sigs-gateway-api.netlify.app/implementation-table/

robscott:
Thanks for all the work on this @xtineskim! This is great to see! Would it be make sense to have a structure that looked something like this?

# Gateway Profile
## HTTPRoute

| | BackendTimeout | HostRewrite | MethodMatching | ParentRefPort | PathRedirect | PathRewrite | PortRedirect | ... | 
| - | - | - | - | - | - | - | - | - |
| Cilium | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ... |
| Kong  | ✅ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ... |
...

# Mesh Profile
## HTTPRoute

| | BackendTimeout | HostRewrite | MethodMatching | ParentRefPort | PathRedirect | PathRewrite | PortRedirect | ... | 
| - | - | - | - | - | - | - | - | - |
| Kuma | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ... |
| Istio  | ✅ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ... |
...

^ Those are not accurate, but hopefully they illustrate the general idea.

All of the above would unfortunately rely on a lot of horizontal scrolling for now, but maybe eventually we could further segment that out further to have different categories of features grouped together in different tables (even within HTTPRoute). 

As a bit of inspiration, there's also this [incredibly helpful spreadsheet for ingress controllers](https://docs.google.com/spreadsheets/d/191WWNpjJ2za6-nbG4ZoUMXMpUK8KlCIosvQB0f-oq3k/edit#gid=907731238) that had a lot of the same goals we're trying to accomplish here.

sunjayBhatia:
> As a bit of inspiration, there's also this [incredibly helpful spreadsheet for ingress controllers](https://docs.google.com/spreadsheets/d/191WWNpjJ2za6-nbG4ZoUMXMpUK8KlCIosvQB0f-oq3k/edit#gid=907731238) that had a lot of the same goals we're trying to accomplish here.

+1 to this approach, agreed its super helpful to see laid out like this

site-src/implementation-table.md
```diff
@@ -1,12 +1,19 @@
-This table is populated from the conformance reports uploaded by project implementations.
+The following tables are populated from the conformance reports uploaded by project implementations. They are separated into the extended features that each project supports listed in their reports.
+
```
xtineskim, line 2: opted for having the projects along the x-axis and the features on the y-axis, since the long names made the scrolling tougher

hack/mkdocs-copy-geps.py
```diff
@@ -15,10 +15,86 @@
...
+ 
+  projects = reports['organization']
+
+  http_reports = reports.loc[reports["name"]=='HTTP']
+  http_reports.set_index('organization')
+  http_reports.sort_values(['organization','version'], inplace=True)
+  http_reports.drop_duplicates(subset='organization', inplace=True, keep='last')
+  
+  table = pandas.DataFrame(columns=http_reports['organization'])
+  table.insert(loc=0, column='Features', value=httproute_extended_conformance_features_list)
```
candita, line 52: It might be better use of the space if we come up with list of shortened feature names.  For example, just removing the HttpRoute prefix will help.  Maybe there could be a link from the shortened feature name to its reference in the API or a similar option.  

HTTPRouteBackendRequestHeaderModification could be called Header Modification and point to https://github.com/kubernetes-sigs/gateway-api/blob/e81022784e13b041a91f73725e9f3e393fccdbfc/conformance/utils/suite/features.go#L101

candita: This also helps if readers aren't familiar with the terms we use for our features.

conformance/reports/v1.0.0/cilium/v1.15.0-pre.3-report.yaml
```diff
@@ -38,7 +38,7 @@ profiles:
         - HTTPRouteRequestTimeout
         - HTTPRouteResponseHeaderModification
         - HTTPRouteSchemeRedirect
-      name: HTTP
+    name: HTTP
```
candita, line 41: If  an implementation adds a broken-format report, will it mess up the display completely, or just for that implmentation?

hack/mkdocs-copy-geps.py
```diff
@@ -15,10 +15,86 @@
...
+    
+    generate_tables(yamlReports)
+
+
+
+# NOTE: will have to be updated if new features are added
+httproute_extended_conformance_features_list = ['HTTPRouteBackendRequestHeaderModification',"HTTPRouteQueryParamMatching",'HTTPRouteMethodMatching',"HTTPRouteResponseHeaderModification","HTTPRoutePortRedirect","HTTPRouteSchemeRedirect","HTTPRoutePathRedirect","HTTPRouteHostRewrite","HTTPRoutePathRewrite","HTTPRouteRequestMirror","HTTPRouteRequestMultipleMirrors","HTTPRouteRequestTimeout", "HTTPRouteBackendTimeout","HTTPRouteParentRefPort"]
+
+
+def generate_tables(reports):
```
candita, line 41: Maybe rename this to `generate_conformance_table`

hack/mkdocs-copy-geps.py
candita: But we can leave this for the next version, as we wouldn't want to hold the first iteration.

mlavacca (commented):
I like @robscott's proposal to group reports by macro-categories (Gateway and Mesh) and have implementations on the y-axis. I see the additional problems we need to tackle here:
1. for each Gateway API Version, an implementation can submit multiple reports for different implementation versions
2. for each Gateway API version and implementation version, there can be multiple reports related to the modes
3. an organization can have multiple projects (at Kong, for example, we have the [Gateway Operator](https://github.com/Kong/gateway-operator) we'll add conformance for soon)

For this reason, I'd like to iterate on Rob's proposal and add the following details:

## Gateway Profile

### HTTPRoute

| Org                                  | Project               | Version             | Mode         | HostRewrite              | MethodMatching             | MethodMatching           | PathRedirect     | ...         |
|:------------------------------------------|:-------------------|:-------------------|:-------------------|:-------------------|:-------------------|:-------------------|:-------------------|:-------------------|
| Kong | kubernetes-ingress-controller               | v3.1.2               | default                | :white_check_mark:                | :x:                | :x:                | :x:                |               |
| Kong               | project-2 | v1.2.3 | with-the-lot | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |  |
| EnvoyProxy                   | envoy-gateway| 0.6.0 | default | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | |

## Mesh Profile

### HTTPRoute

...

I understand this decreases the readability of the table, but I didn't find a better way to include all the needed information.

conformance/reports/v1.0.0/cilium/v1.15.0-pre.3-report.yaml
xtineskim: good question - it will be missing from the display. For example, currently the `generate_table` filters for the name "HTTP", and if that column is missing (the format results in the column being in an incorrect location in the dataframe), then it will be dropped. I guess it's somewhat of a check

xtineskim:
@mlavacca thanks for your input! I think with @candita's comment about having abbreviated names (maybe not in this pr), I think it would be better. I'll make the changes to switch the axis

Per your comments listed:
> 1. for each Gateway API Version, an implementation can submit multiple reports for different implementation versions

Currently I am just grabbing the latest report uploaded (nginx, contour both have 2 yaml reports, but i sort the versions and keep the latest one). I'm hoping that as new conformance reports are generated, all the profiles will be included in the latest report

>2.for each Gateway API version and implementation version, there can be multiple reports related to the modes

Ah good catch with the mode, I forgot that use case ([reference for others](https://gateway-api.sigs.k8s.io/geps/gep-1709/?h=confor#implementation-mode)). I haven't seen any project report use that field either (yet). To address this, I can add a filter for mode for the loaded dataframes to look for the mode type, but also add in the default mode for current projects. Thanks for the reminder on this!

>3.an organization can have multiple projects

I'm assuming though that this will have it's own folder, and a different project name? And if the project name is different, that shouldn't be a problem 👍

xtineskim:
@robscott thank you for the suggestions! 🤝  slowly addressing the above comments, and i like the proposal of having the Gateway Profile and the Mesh Profile. 
Just a few q's:
- The Mesh Profiles information are pulled from the name of `MESH` in the reports (for example [kuma's](https://github.com/kubernetes-sigs/gateway-api/blob/a101a237332dbb9216cf5d8b85acc238ba56d4cf/conformance/reports/v1.0.0/kumahq-kuma/2.6.0-report.yaml#L70). Not sure if naming of MESH will have to be changed in the future if other routes will be added under the MESH conformance? I guess just assume keeping it for HTTPRoute for now?
- Should I still have the Kuma listed under the Gateway Profile as well for HTTPRoute (pulling from [this part of the report](https://github.com/kubernetes-sigs/gateway-api/blob/a101a237332dbb9216cf5d8b85acc238ba56d4cf/conformance/reports/v1.0.0/kumahq-kuma/2.6.0-report.yaml#L41)).

robscott:
>The Mesh Profiles information are pulled from the name of MESH in the reports (for example [kuma's](https://github.com/kubernetes-sigs/gateway-api/blob/a101a237332dbb9216cf5d8b85acc238ba56d4cf/conformance/reports/v1.0.0/kumahq-kuma/2.6.0-report.yaml#L70). Not sure if naming of MESH will have to be changed in the future if other routes will be added under the MESH conformance? I guess just assume keeping it for HTTPRoute for now?

This is actually one of my concerns with the current names of the conformance profiles. I think we may want to just settle on "mesh" and "gateway" conformance profiles and then have per-route details covered by the supported features. Interested in what @shaneutt and @mlavacca think about that though.

>Should I still have the Kuma listed under the Gateway Profile as well for HTTPRoute (pulling from [this part of the report](https://github.com/kubernetes-sigs/gateway-api/blob/a101a237332dbb9216cf5d8b85acc238ba56d4cf/conformance/reports/v1.0.0/kumahq-kuma/2.6.0-report.yaml#L41)).

Yes, I think both Kuma and Istio support the Mesh and Gateway concepts in Gateway API, so they'd be covered under both sections.

michaelbeaumont:
> Should I still have the Kuma listed under the Gateway Profile as well for HTTPRoute (pulling from [this part of the report](https://github.com/kubernetes-sigs/gateway-api/blob/a101a237332dbb9216cf5d8b85acc238ba56d4cf/conformance/reports/v1.0.0/kumahq-kuma/2.6.0-report.yaml#L41)).

> Yes, I think both Kuma and Istio support the Mesh and Gateway concepts in Gateway API, so they'd be covered under both sections.

Yep, Kuma does indeed support HTTPRoute for both Mesh and Gateway!

mikemorris:
Confirming that Istio likewise supports HTTPRoute for both [gateway](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#configuring-a-gateway) and [mesh](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#mesh-traffic) use cases.

site-src/implementation-table.md
```diff
@@ -0,0 +1,25 @@
...
+| projectcontour | :x:                                         | :white_check_mark:            | :white_check_mark:        | :white_check_mark:                    | :white_check_mark:      | :white_check_mark:        | :white_check_mark:      | :white_check_mark:     | :white_check_mark:     | :white_check_mark:       | :white_check_mark:                | :white_check_mark:        | :white_check_mark:        | :x:                      |
+| solo.io        | :x:                                         | :white_check_mark:            | :white_check_mark:        | :white_check_mark:                    | :white_check_mark:      | :white_check_mark:        | :white_check_mark:      | :x:                    | :x:                    | :x:                      | :x:                               | :x:                       | :x:                       | :x:                      |
```
robscott, line 16: @xtineskim I think we might need to show both "Organization" and "Project" here for now. Otherwise we end up with incomplete information. For example, in this case, most people would want to know the product name ("Gloo Gateway"), while in the case of Kong above, the "Kubernetes Ingress Controller" product name is really not sufficient. Until we have a better solution in conformance reports, I think we'll just have to settle for reporting both.

@mlavacca Do you think it's worth adding a new field (or merging fields) in conformance reports to make this easier to handle? Maybe we should just have a single product name field that represents everything?

site-src/implementation-table.md
```diff
@@ -0,0 +1,25 @@
+The following tables are populated from the conformance reports uploaded by project implementations. They are separated into the extended features that each project supports listed in their reports.
+
+## Gateway Profile
```
robscott, line 3: I'm not sure where/how to fit this in, but we'll likely want to always support the current and previous version of Gateway API here. Maybe we need something like:

```
## v1.1
### Gateway
#### HTTPRoute

...

## v1.0
```

Admittedly a report for v1.1 hasn't merged yet, but that should happen relatively soon with https://github.com/kubernetes-sigs/gateway-api/pull/3025.

site-src/implementations.md
```diff
@@ -7,6 +7,8 @@ Implementors and integrators of Gateway API are encouraged to update this
 document with status information about their implementations, the versions they
 cover, and documentation to help users get started.
 
+[View a table to quickly compare supported features of projects](/implementation-table)
```
robscott, line 10: This might benefit from using some kind of callout so it's more noticeable (https://gateway-api.sigs.k8s.io/contributing/style-guide/#callouts)

hack/mkdocs-copy-geps.py
```diff
@@ -15,10 +15,109 @@
 import shutil
 import logging
 from mkdocs import plugins
+import yaml
```
robscott, line 18: Nit: I think this file should probably be renamed to represent the broader scope of the script.

mkdocs.yml
```diff
@@ -5,6 +5,9 @@ site_dir: site
 docs_dir: site-src
```
robscott, line 5: Recommend adding a link to the new implementation-tables page somewhere in the navigation so it doesn't feel quite as disconnected from the rest of the site.

Maybe we have an "Implementations" section with "List" (current page) and "Comparison" (new page) options.

hack/mkdocs-copy-geps.py
robscott: +1 on this being very nice to have but also not a blocker. If you run out of time to fit this into this PR, do you mind filing a follow up issue so we don't lose track of it?

robscott (commented):
Thanks for all the work on this @xtineskim! This is looking like it's really close, just a few more nits but otherwise LGTM.

site-src/implementation-table.md
youngnick: I think this could also be a followup item though if we can't get it in _this_ PR.

site-src/implementation-table.md
```diff
@@ -0,0 +1,25 @@
+The following tables are populated from the conformance reports uploaded by project implementations. They are separated into the extended features that each project supports listed in their reports.
```
youngnick, line 1: How about something like this? Please feel free to wordsmith here, but I think this covers the main things: only Core-conformant implementations are listed here, and this is still being worked on (although I should say again that I think that @xtineskim has done a great job here.)

```suggestion
The following tables are populated from the conformance reports uploaded by project implementations. They are separated into the extended features that each project supports listed in their reports.

Implementations only appear in this page if they pass Core conformance for the resource type, and the features listed should be Extended features.

!!! warning+ "Comparison under active development"
    This page is under active development and is not in its final form, especially for the project name and the names of the features.
    However, as it is based on submitted conformance reports, the information is correct.
```

site-src/implementation-table.md
mlavacca: in the reports we already have `organization` and `project`, which if merged provide a unique tuple. Acutally, this is very related to my comment above: https://github.com/kubernetes-sigs/gateway-api/pull/2874#pullrequestreview-2003274625. I agree we should address both org+project and implementation version.

xtineskim: gotcha, i'll add them in 👍

site-src/implementation-table.md
xtineskim: yeah i think i'd like to noodle on it a bit. I don't like the way i have the version handled, but maybe a way to have older pages potentially? 🤔

site-src/implementation-table.md
xtineskim: added this into the portion of the python file that spits out this markdown! Thank you for the suggestion @youngnick  ! 🖊️ 🙏

hack/mkdocs-copy-geps.py
xtineskim: im thinking of putting this conformance report generation into it's own `make-implementation-table.py` and have it as it's own hook. Thoughts??

robscott:
Thanks @xtineskim, this is a great addition!

/lgtm
/approve

PR #2892, opened 2024-03-22
Add label filtering (-l) for get commands

### LINKED ISSUE #2890: Add label filtering (-l) for get commands
Refer https://gateway-api.sigs.k8s.io/geps/gep-2722/ for details

yeedove:
**What type of PR is this?**

Fixes #2890

```release-note
Add label filtering (-l) for get commands
```

- Correct label format output
```
➜  gwctl get httproutes -l app=backend
NAMESPACE  NAME     HOSTNAMES        PARENT REFS  AGE
default    backend  www.example.com  1            30d
```

- Error label format output
```
➜  gwctl get httproutes -l aps=dsd,adsa,=ds
Unable to find resources that match the label selector "aps=dsd,adsa,=ds": found '=', expected: identifier after ','
```

### DIFF
--- a/gwctl/pkg/printer/gateways_test.go
+++ b/gwctl/pkg/printer/gateways_test.go
@@ -366,56 +366,23 @@ EffectivePolicies:
 	}
 }
 
+// TestGatewaysPrinter_LabelSelector tests label selector filtering for Gateways in 'get' command.
 func TestGatewaysPrinter_LabelSelector(t *testing.T) {
 	fakeClock := testingclock.NewFakeClock(time.Now())
-	objects := []runtime.Object{
-		&gatewayv1.GatewayClass{
+	gateway := func(name string, labels map[string]string) *gatewayv1.Gateway {
+		return &gatewayv1.Gateway{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "internal-class",
-			},
-			Spec: gatewayv1.GatewayClassSpec{
-				ControllerName: "example.net/gateway-controller",
-				Description:    common.PtrTo("random"),
-			},
-		},
-		&gatewayv1.GatewayClass{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "external-class",
-			},
-			Spec: gatewayv1.GatewayClassSpec{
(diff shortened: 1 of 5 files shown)


gwctl/pkg/printer/gateways_test.go
```diff
@@ -364,3 +365,181 @@ EffectivePolicies:
 		t.Errorf("Unexpected diff\ngot=\n%v\nwant=\n%v\ndiff (-want +got)=\n%v", got, want, diff)
 	}
 }
+
+func TestGatewaysPrinter_LabelSelector(t *testing.T) {
```
gauravkghildiyal, line 369: This suggestions applies to the rest of the test functions as well.

Lets aim to be very focused on the piece of functionality this function is testing. In this case, since we're mostly concerned with the fact that label based filtering is working, we can try and reduce some of the other pieces of the test. For example:
1. Maybe lets have only one GatewayClass. All gateways can be for the same GatewayClass.
2. Lets create a helper function, within the test function like the following:
    ```
    gateway := func(name string, labels map[string]string) gatewayv1.Gateway {...}
    ```
    This function will return the exact same gateway, with only the name and labels varying.
3. Within the function above, lets try to set the bare minimal set of values within the Gateway resource (of course, it need not be the strict minimal set, but whatever we can cut things down to without much trial and error)

This should really help with:
1. Readers focusing on what is being tested.
2. Reduce the impact of unrelated changes affecting this test function.
3. Some wise people have said that code is a liability, so we should try and minimize when possible.

gauravkghildiyal (commented):
Thanks @yeedove! This is neat :)

One minor request, can we please have consistent names of the test functions (like TestHTTPRoutesPrinter_LabelSelector) and add a small comment before the test function describing that this is focused only on checking if label based filtering is working fine.

gwctl/pkg/printer/gateways_test.go
yeedove: yup，done

yeedove: Yup, done!

PR #2896, opened 2024-03-23
docs: add dark-mode capability

meysam81:
**What type of PR is this?**

/kind documentation

Having the ability to choose between dark and light mode is crucial when reading documents online behind LED lights. This PR is an effort to make that a reality for people who prefer reading in dark mode, especially myself.

N/A

Just the docs.


meysam81:
ping @youngnick

robscott:
Hey @meysam81, thanks for adding this! Do you mind changing the default to still be light mode? We still have a variety of graphics that haven't been adapted to work well on dark mode, so I don't think we're ready to use that as a default in any case, but I think it's fine if people want to manually opt in.

meysam81:
@robscott 
Without any further change, the current default is the system default, which is more adaptable to the user's settings. Are you suggesting that it's better to have the light mode as default altogether? :thinking:

robscott:
> @robscott Without any further change, the current default is the system default, which is more adaptable to the user's settings. Are you suggesting that it's better to have the light mode as default altogether? 🤔

Yep, until we can update the graphics to adjust for dark mode, I think we need to default to the light mode. I default to dark mode on my computers but would still personally prefer the light mode on the website until we can also make corresponding updates to our graphics. For example, just looking at the main page, both the diagram and logo do not look ideal with dark mode. I think it's great to offer the choice, I just wouldn't want dark mode to be default for anyone quite yet.
Screenshot 2024-04-08 at 8 22 50 PM

meysam81:
@robscott 
Yes, having the option to choose the dark mode is great, and I don't mind having to do it manually once (and not touching the cookies ever :grin: ). For me, I would rather see less appealing graphics than having to endure light mode; it gets more frustrating after evenings.
But I'll push for the change shortly.
Thanks for taking the time to review this.

meysam81:
kindly take a look :pray: 
@youngnick @shaneutt @robscott

robscott:
@meysam81 I may be missing something, but at least for me on Chrome, it's still defaulting to dark mode (matching my system preferences).

meysam81:
@robscott 

Would you try again, please? This last change should do it.

robscott:
Hey @meysam81, sadly still seeing the same results on Chrome and Safari where it defaults to dark mode for me (matching my OS).

meysam81:
@robscott Does it by any chance have a different outcome when clearing cookies and/or trying from incognito I wonder?

robscott:
> @robscott Does it by any chance have a different outcome when clearing cookies and/or trying from incognito I wonder?

Yep, incognito and private modes respectively.

meysam81:
@robscott 

The behavior of the theme following the system preferences seems to be a design decision, as explained here: https://github.com/squidfunk/mkdocs-material/discussions/7093#discussioncomment-9220910

It seems like this PR can be accepted or rejected as is because, as far as my understanding goes, there can't be a configuration change that can force the light mode on a system with a dark-mode preference. :shrug:

robscott:
Sorry for the delay on this @meysam81! We discussed this at a community meeting and the overall consensus was that we should hold off until we can also update the images. I've created https://github.com/kubernetes-sigs/gateway-api/issues/3146 to track that effort.

shaneutt:
Blocked on #3146 then it seems:

/hold

sftim:
> It seems like this PR can be accepted or rejected as is because, as far as my understanding goes, there can't be a configuration change that can force the light mode on a system with a dark-mode preference. 🤷

You can use CSS to do something close
```
:root {
  color-scheme: only light; // dark mode support is incomplete
}
```

and then optionally let people opt in to something that overrides that

Issue #2905, opened 2024-03-26
Implementations are required to accept a Gateway with a UDP listener when running HTTPRoute conformance tests


pleshakov:
**What happened**:

https://github.com/kubernetes-sigs/gateway-api/pull/2661 added support for a UDP conformance test.

However, the corresponding Gateway is always installed https://github.com/kubernetes-sigs/gateway-api/blob/6865ac1ce5018695a60a180157c0644f03666af9/conformance/base/manifests.yaml#L721-L734

Implementations that don't support UDP will not Accept this Gateway:
- Because protocol is not supported https://github.com/kubernetes-sigs/gateway-api/blob/6865ac1ce5018695a60a180157c0644f03666af9/apis/v1/gateway_types.go#L327-L331
- Because no other non-UDP listeners are present - https://github.com/kubernetes-sigs/gateway-api/blob/6865ac1ce5018695a60a180157c0644f03666af9/apis/v1/gateway_types.go#L781

As a result, the tests cannot proceed and are stuck at:
```
helpers.go:222: gateway-conformance-infra/udp-gateway Gateway not Accepted yet
    helpers.go:221: Accepted condition set to Status False with Reason ListenersNotValid, expected Status True
    helpers.go:221: Accepted was not in conditions list [[{Accepted False 1 2024-03-26 21:20:04 +0000 UTC ListenersNotValid Gateway has no valid listeners} {Programmed False 1 2024-03-26 21:20:04 +0000 UTC Invali
d Gateway has no valid listeners}]]
. . .
```

**What you expected to happen**:

Supporting UDPRoute is not required for HTTPRoute conformance tests. 

**How to reproduce it (as minimally and precisely as possible)**:

Running the conformance test with supported features that don't have UDP. For example:

```
--supported-features=HTTPRouteQueryParamMatching,HTTPRouteMethodMatching,HTTPRoutePortRedirect,HTTPRouteSchemeRedirect,HTTPRouteHostRewrite,HTTPRoutePathRew
rite,GatewayPort8080
```

Here is an example output of the Gateway status set by [NGF](https://github.com/nginxinc/nginx-gateway-fabric) during a conformance tests run:
```yaml
status:
  addresses:
  - type: IPAddress
    value: 10.244.0.71
  conditions:
  - lastTransitionTime: "2024-03-26T21:20:04Z"
    message: Gateway has no valid listeners
    observedGeneration: 1
    reason: ListenersNotValid
    status: "False"
[...]

PR #2906, opened 2024-03-27
fix: move udp-gateway into specific test fixture

cnvergence:
**What type of PR is this?**

/kind bug

Fixes https://github.com/kubernetes-sigs/gateway-api/issues/2905

```release-note
NONE
```

### DIFF
--- a/conformance/base/manifests.yaml
+++ b/conformance/base/manifests.yaml
@@ -717,18 +717,3 @@ data:
     foo.bar.com:53 {
       whoami
     }
----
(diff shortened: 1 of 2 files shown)


pleshakov (approved):
👍 

I ran the updated tests against an implementation that doesn't support UDP ([NGF](https://github.com/nginxinc/nginx-gateway-fabric)) and, since the UDP Gateway wasn't created, the tests succeed.

youngnick:
Good catch, thanks.

/lgtm

PR #2909, opened 2024-03-28
Finish Route Rule API design and update edge case behaviors.

ginayeh:
**What type of PR is this?**
/kind gep

N/A

The main goal of this PR is to get https://github.com/kubernetes-sigs/gateway-api/issues/1619 ready for implementable by answering the open questions in [Graduation Criteria for Implementable Status](https://gateway-api.sigs.k8s.io/geps/gep-1619/#graduation-criteria-for-implementable-status):
  4. Finish designing the Route Rule API and document edge cases in Edge Case Behavior for configuring session persistence on both `BackendLBPolicy` and route rules.

Fixes #2634 

```release-note
NONE
```

### DIFF
--- a/geps/gep-1619/index.md
+++ b/geps/gep-1619/index.md
@@ -633,7 +633,7 @@ When using multiple backends in traffic splitting, all backend services should h
 Nonetheless, implementations MUST carefully consider how to manage traffic splitting scenarios in which one service has
 persistence enabled while the other does not. This includes scenarios where users are transitioning to or from an
 implementation version designed with or without persistence. For traffic splitting scenario within a single route rule,
-this GEP leaves the decision to the implementation. Implementations MAY choose to apply session persistence to all
+this GEP leaves the decision to the implementation. Implementations MUST choose to apply session persistence to all
 backends equally, reject the session persistence configuration entirely, or apply session persistence only for the
 backends with it configured.
 
@@ -959,6 +959,8 @@ TODO
 - How do implementations drain established sessions during backend upgrades without disruption?
     - Do we need a "session draining timeout" as documented by [A55: xDS-Based Stateful Session Affinity for Proxyless gRPC](https://github.com/grpc/proposal/blob/master/A55-xds-stateful-session-affinity.md#background)
       defined in this API?
+- How do we provide a standard way to communicate that an implementation does not support Route Rule API?
+    - Do we want something conceptually similar to the `IncompatibleFilters` reason?
 
 ## TODO
 
--- a/geps/gep-1619/index.md
+++ b/geps/gep-1619/index.md
@@ -921,7 +921,7 @@ spec:
 ```
 
 In this traffic splitting scenario within a single route rule, this GEP leaves the decision to the implementation. An
-implementation MAY choose to:
+implementation MUST choose one of the following:
 
 1. Apply session persistence configured in `BackendLBPolicy` to `servicev1` and `servicev2` equally
 2. Reject the session persistence configured in `BackendLBPolicy` so that `servicev1` does not have session persistence
--- a/geps/gep-1619/index.md
+++ b/geps/gep-1619/index.md
@@ -821,6 +821,44 @@ this by the following commands:
 2. Curl to `/b` routes MUST direct traffic to `servicev2` since the persistent session established earlier is not
    shared with this route path.
 
+#### Route Rules Referencing to a Session Persistent Enabled Service Must Not Share Sessions
+
+Consider the situation in which two different route paths are going to the same service, and session persistence is enabled with the service via `BackendLBPolicy`:
+
+```yaml
+kind: HTTPRoute
+metadata:
+  name: routeX
+spec:
+  rules:
+  - matches:
+    - path:
+      value: /a
+    backendRefs:
+    - name: servicev1
+  - matches:
+    - path:
+      value: /b
+    backendRefs:
+    - name: servicev1
+---
+kind: BackendLBPolicy
+metadata:
+  name: lbp
+spec:
+  targetRef:
+    kind: Service
+    Name: servicev1
+  sessionPersistence:
+    sessionName: service-cookie
+    type: Cookie
+```
+
+Route rules referencing the same service MUST NOT share persistent sessions (i.e. the same cookie), even if the session persistence is attached to the service via `BackendLBPolicy`, and each route rule should have different persistent sessions.
+
+1. Curl to `/a` which establishes a persistent session with `servicev1`
+2. Curl to `/b` which establishes another persistent session with `servicev1` since the previous session established earlier is not shared with this route path.
+
 #### Session Naming Collision
 
 Consider the situation in which two different services have cookie-based session persistence configured with the
--- a/geps/gep-1619/index.md
+++ b/geps/gep-1619/index.md
@@ -887,7 +887,7 @@ spec:
 
 In this scenario, session persistence is enabled at route rule level and all services in the traffic split have persistent session.
 
-That is to say, traffic routing to `servicev1` previously will continue to be routed to `servicev1` when cookie is present, and traffic routing to `servicev2` previously will continue to be routed to `servicev2` when cookie is present.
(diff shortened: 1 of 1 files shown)


robscott (commented):
Thanks @ginayeh!

geps/gep-1619/index.md
```diff
@@ -549,6 +549,20 @@ type HTTPRouteRule struct {
...
+type GRPCRouteRule struct {
+    [...]
+
+    // SessionPersistence defines and configures session persistence
+    // for the route rule.
+    //
+    // Support: Extended
+    //
+    // +optional
+    SessionPersistence *SessionPersistence `json:"sessionPersistence"`
```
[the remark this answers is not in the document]
ginayeh: We have two examples in the GEP so far.  Do you think we should add more?
1) https://github.com/kubernetes-sigs/gateway-api/blob/main/geps/gep-1619/index.md#two-routes-rules-have-session-persistence-to-the-same-service
2) added by PR at `Traffic Splitting with route rule inline sessionPersistence field`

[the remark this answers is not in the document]
robscott: Whoops, I missed that those existed before this PR, thanks!

geps/gep-1619/index.md
```diff
@@ -926,13 +920,14 @@ spec:
     type: Cookie
 ```
 
-A potentially unexpected situation occurs when:
+In this traffic splitting scenario within a single route rule, this GEP leaves the decision to the implementation. An
+implementation MAY choose to:
```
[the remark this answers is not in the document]
ginayeh: Updated.

geps/gep-1619/index.md
```diff
@@ -529,11 +530,10 @@ const (
 
 ### Route Rule API
 
-To support route-level configuration, this GEP also introduces an API as inline fields within HTTPRoute and GRPCRoute.
-Any configuration that is specified at Route level will override configuration that is attached at the backend level.
-This route-level API for enabling session persistence currently uses the same `SessionPersistence` struct from the
-`BackendLBPolicy` API. However, this API should be reevaluated in a future update to this GEP, considering the
-associated edge cases when configuring both routes and backends.
+To support route rule level configuration, this GEP also introduces an API as inline fields within HTTPRouteRule and GRPCRouteRule.
+Any configuration that is specified at Route Rule level MUST override configuration that is attached at the backend level because route rule have a more global view and responsibility for the overall traffic routing.
```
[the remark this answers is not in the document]
ginayeh: Great point.  I added this as an open question in the GEP.  Please take a look.

robscott:
Thanks @ginayeh! Will defer to @gcs278 or @youngnick for final LGTM.

/approve

gcs278:
Thanks @ginayeh! I will take a look at it early next week.

geps/gep-1619/index.md
```diff
@@ -867,52 +880,33 @@ spec:
...
 ```
 
-In this traffic splitting scenario within a single route rule, this GEP leaves the decision to the implementation. An
-implementation MAY choose to:
+In this scenario, session persistence is enabled at route rule level and all services in the traffic split have persistent session.
 
-1. Apply session persistence configured in `BackendLBPolicy` to `servicev1` and `servicev2` equally
-2. Reject the session persistence configured in `BackendLBPolicy` so that `servicev1` does not have session persistence
-3. Apply session persistence for only `servicev1`, potentially causing all traffic to eventually migrate to `servicev1`
+That is to say, traffic routing to `servicev1` previously will continue to be routed to `servicev1` when cookie is present, and traffic routing to `servicev2` previously will continue to be routed to `servicev2` when cookie is present.
```
gcs278, line 890: Is it clearer to call it out as a MUST?

```suggestion
That is to say, traffic routing to `servicev1` previously MUST continue to be routed to `servicev1` when cookie is present, and traffic routing to `servicev2` previously MUST continue to be routed to `servicev2` when cookie is present.
```

geps/gep-1619/index.md
gcs278, line 924: You updated the wording to `MUST` above. That makes sense, as they MUST choose an option. Should we do that here?

```suggestion
implementation MUST choose one of the following:
```

geps/gep-1619/index.md
```diff
@@ -604,8 +618,8 @@ and/or [GEP-2648](/geps/gep-2648) for more specific details on how to handle ove
 Edge cases will arise when implementing session persistence support for both backends and route rules through
 `BackendLBPolicy` and the route rule's `sessionPersistence` field. For guidance on addressing conflicting
 attachments, please consult the [Edge Case Behavior](#edge-case-behavior) section, which outlines API
-use cases. Only a subset of implementations have already designed their data plane to incorporate route-level session
-persistence, making it likely that route-level session persistence will be less widely implemented.
+use cases. Only a subset of implementations have already designed their data plane to incorporate route rule level session
+persistence, making it likely that route rule level session persistence will be less widely implemented.
 
 ### Traffic Splitting
```
gcs278, line 625: Since we dropped the `weight 0` scenario, I think we should still place a call out for `weight 0`, as it could be a ambiguous situation. I think the `Traffic Splitting` section makes the most sense. You can use own judgement if you disagree. I can't suggest for this line, so commenting this nearby:

```
When a
persistent session is established and traffic splitting is configured across services, the persistence to a single
backend MUST be maintained across services, **even if the weight is set to 0.**
```

gcs278 (commented):
Thanks! Looking good to me, just some minor nit picks.

geps/gep-1619/index.md
ginayeh: Thanks for pointing out.  I forgot to add more context about why I removed the weight 0 example and also want to confirm I'm not missing anything critical.

```
1. Curl to /a which establishes a persistent session with servicev1
2. Curl to /b routes to servicev1 due to route persistence despite weight: 0 configuration
```
The previous example we had is doing two route rules with path `/a` and `/b`, and path `/b' has traffic split with weight 0.  Since each route rule should have its own persistent session, so applying cookies from persistent session `/a` to servicev1 should not affact `/b` traffic routing with servicev1.  As a result, `/b` should be routed to servicev2 with `weight: 100`.  Does it make sense?

Then, I was thinking about how a persistent session can be established with a `weight: 0` backend?  Maybe servicev1 had a non-zero weight and a persistent session established, and then route resource is re-deployed and servicev1 has a new `weight: 0` configuration, but I am not sure if we expect session persistence to work after deployment, so I stopped adding an example with `weight: 0`.  Is there other example with `weight: 0` that we can include?

gcs278: > Since each route rule should have its own persistent session, so applying cookies from persistent session /a to servicev1 should not affact /b traffic routing with servicev1. 

So, to be clear, even if the session persistence is attached to the Service via BackendLBPolicy, each route rule should have different cookies / persistent sessions? Hmm. I was originally thinking that the cookie would be associated with the service when attached via BackendLBPolicy, so that it could be shared among route rules (in the weight 0 example).

If it isn't shared among route rules when using BackendLBPolicy, would that be problematic for setting the cookie name? For instance, ServiceA with cookie named "foo". There's two route rules for /x and /y, but both go just to ServiceA. The cookie "foo" will encode the backendID for ServiceA. Won't the cookie be re-used for /x and /y (two route paths)?

I think regardless, you bring up a good point that I think we should clarify in the GEP: does attaching to Service mean cookies are reused across route rules?

ginayeh: I think attaching BackendLBPolicy to a Service should NOT mean cookies are resued across route rules referenceing to the same service.  Let me update PR with a clarification on this.

ginayeh: Another commit was pushed for this.  PTAL @gcs278

gcs278: Thanks for the clarification @ginayeh. LGTM

robscott:
Thanks @ginayeh and @gcs278!

/hold cancel

PR #2910, opened 2024-03-29
Adding GEP 2907: TLS Configuration Placement and Terminology

### LINKED ISSUE #2907: GEP: TLS Configuration Placement and Terminology
A memorandum GEP that provides a high level vision for TLS configuration in Gateway API. This GEP will not include any specific API changes, but will propose naming patterns for top level fields/buckets that should be used to TLS configuration.

**Why this is needed**:
We've got two TLS GEPs in flight (#91 and #1897), and likely more on the way. If we're not careful, each will end up with distinct naming and structure that made sense in isolation but does not fit the broader picture.

robscott:
**What type of PR is this?**
/kind gep

This adds a new memorandum GEP that attempts to chart a vision for where our TLS configuration will live and what it will be called.

Fixes #2907

```release-note
NONE
```

/cc @arkodg @candita @youngnick

### DIFF
--- a/geps/gep-2907/index.md
+++ b/geps/gep-2907/index.md
@@ -1,0 +1,209 @@
+# GEP-2907: TLS Configuration Placement and Terminology
+
+* Issue: [#2907](https://github.com/kubernetes-sigs/gateway-api/issues/2907)
+* Status: Memorandum
+
+## TLDR
+This GEP aims to define high level TLS terminology and structure within Gateway
+API to ensure that all our independent proposals related to TLS result in a
+coherent set of APIs. This will result in some adjustments to provisional and
+experimental TLS-related GEPs, specifically [BackendTLSPolicy](/geps/gep-1897)
+and [Client Certificate Verification](/geps/gep-91).
+
+## Goals
+* Define high-level terminology for how we refer to TLS in Gateway API.
+* Define top level fields where TLS configuration can live.
+
+## Non-Goals
+* Add or change fields directly. (This may inspire changes in other GEPs
+  though).
+* Commit to including specific parts of TLS configuration in Gateway API. (This
+  is merely to provide space for future configuration, not a commitment that we
+  will add it to the API.)
+
+## Out of Initial Scope
+There are a variety of related TLS concepts in Gateway API that are not currently
+in scope for this GEP. In the future, this GEP may be expanded to include:
+
+1. Automatic mTLS (often associated with Service mesh)
+1. TLS Passthrough
+1. TLSRoute
+
+
+## Introduction
+
+### Where TLS could be configured
+We have three different places where we might want to configure TLS:
+
+#### A. Gateways
+Config attached at this layer will apply to everything behind a single address
+such as a virtual IP.
+
+#### B. Gateway Listeners
+Each Gateway contains multiple “Listeners”. Each HTTPS Listener in a Gateway
+must have a unique combination of protocol, port, and SNI (Server Name
+Indication). TLS configuration attached at this layer should be [isolated via
+SNI](/guides/implementers/#listener-isolation).
+
+#### C. BackendTLSPolicy
+This policy allows us to attach unique TLS configuration per Backend. Depending
+on the organization, this policy may be owned by the application owner or the
+cluster operator. Note that this configuration will be used by all Gateways
+(potentially from different implementations) that are connecting to the backend.
+
+### "Frontend" and "Backend"
+A guiding principle in this naming is to use consistent naming for “Downstream”
+(1+2) and “Upstream” (3+4), similar to Envoy. To avoid the confusion with what
+is upstream and downstream, Gateway API will use “Frontend” (1+2) and “Backend”
+(3+4).
+
+*  **Frontend:** The entity connecting to a Gateway, typically a client
+   application and/or web browser.
+*  **Backend:** The entity a Gateway is routing traffic to, typically the
+   endpoints behind a Service.
+
+There are essentially 4 different segments that could be relevant to TLS
+configuration in Gateway API:
+
+```mermaid
+flowchart LR
+    Frontend -.->|1. Client Cert| Gateway
+    Gateway -.->|2. Server Cert| Frontend
+    Gateway -.->|3. Client Cert| Backend
+    Backend -.->|4. Server Cert| Gateway
+```
+
+The above diagram depicts these four segments as edges in a graph.
+
+## Proposed Segments
+Note that this does not represent any form of commitment that any of these
+fields or concepts will exist within Gateway API. If or when they do, we propose
+the following naming structure:
+
+### 1. Validate Client Certificate provided by Frontend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway Listener | `Listener.TLS.FrontendValidation` | Proposed |
+
+#### Rationale
+Use FrontendValidation to leave room for concepts like trust anchors and trust
+domains. Anything not strictly tied to validation would belong to Listener.TLS
+which is now reserved exclusively for “Frontend” TLS configuration (1+2).
+
+#### Why Not Frontend.Validation?
+Part of the idea behind this GEP is that Listener.TLS is really entirely focused
+on 1+2 - the bits of TLS between frontend and Gateway. That means that if we
+were to add any additional TLS configuration in the Listener.TLS struct, it
+would tied to that limited scope, and thus we don't really need a separate
+Frontend struct.
+
+One could make an argument that Listener.TLS.Validation should be the field name
+here to avoid any ambiguity, but in this specific context, we think that it's
+probably helpful to specifically spell out frontend.
+
+#### Why Not FrontendTLSPolicy?
+It could be reasonable to try to mirror BackendTLSPolicy for Frontend TLS. As we
+considered that possibility, we evaluated why BackendTLSPolicy exists as a
+separate policy resource:
+
+1. Some parts of the config, such as the SNI that should be used to connect to
+   the Service/backend only make sense when attached per Service - it's very
+   unlikely that you'd want to use the same SNI to connect to all backends from
+   a Gateway.
+1. The Service API is overloaded and very slow to change, Gateway API has taken
+   the approach of exposing config at that level via policy resources that are
+   tied to specific personas (why Session Persistence config and Backend TLS are
+   in different resources).
+
+We don't think either of those reasons really apply to frontend TLS. Instead,
+frontend TLS could theoretically be configured either for an entire Gateway, or
+a specific Gateway listener. Given that many implementations already support
+distinguishing this config per SNI, it seems to make sense to start with
+listener level attachment. We think that the persona that would be responsible
+for a Gateway is not sufficiently different than the persona that would be
+responsible for frontend TLS, so the current proposal is likely the best option
+available to us.
+
+
+### 2. Configure TLS Termination, including Server Certificate
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway Listener | `Listener.TLS` | GA |
+
+#### Rationale
+This is already finalized in the API and so we're stuck with this name. In
+hindsight a name that was more clearly tied to frontend TLS would have been
+ideal here.
+
+
+### 3. Configure Client Certificate that Gateway should use to connect to Backend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.Spec.BackendTLS.ClientCertificateRef` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.Spec.ClientCertificateRef` | Not Proposed |
+
+#### Rationale
+It's not yet obvious which of the above options are preferable, but a case could
+be made for either or both. If we add a `BackendTLS` struct to Gateway it would
+leave room for concepts like TLS version, ALPN, cipher suites, etc.
+
+#### Why Not Listener level to match FrontendValidation?
+In general, we'd expect this identity to represent a Gateway as a whole, and
+thus per-Listener configuration probably does not make sense here. Having
+FrontendValidation at the Listener level allows a Gateway to accept different
+client certs per SNI (a matching attribute of the frontend -> Gateway
+connection). On the other hand, when determining the identity a Gateway should
+use when connecting to a backend, it should likely either be tied directly to
+the Gateway or Backend, but the Listener is not particularly relevant in this
+context.
+
+
+### 4. Validate Server Certificate that is provided by Backend
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.Spec.BackendTLS.Validation` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.Spec.Validation` | Experimental (Different Name) |
+
+#### Rationale
+Rename `BackendTLSPolicy.TLS` to `BackendTLSPolicy.Validation`. Because this is
+already clearly backend-focused, config like TLS Version, ALPN, and cipher
+suites, could live at the same level without confusion. It's also plausible that
+some Gateways may want to express Gateway-wide CAs that are trusted.
+
+#### Why BackendTLS on Gateways?
+Although this GEP is intentionally not committing to adding new fields or
+features, it's possible that at some point in the future we may want to have
+some kind of Backend TLS config at the Gateway level. For example, it may be
+useful to configure a set of CAs that a Gateway trusts or a client cert that a
+Gateway should use to connect to all backends. If this existed, there would
+likely be some overlap in config with BackendTLSPolicy, and if a future GEP
+proposed this, it would need to include how overlapping config should be
+handled.
+
+#### Why the inconsistency between Frontend and Backend TLS config on Gateways?
+If we were to populate all the possible fields described in this GEP, we'd end
+up with the following config on Gateways:
+
+```
+Gateway.Spec.Listeners.TLS.FrontendValidation
+Gateway.Spec.Listeners.TLS
+Gateway.Spec.BackendTLS.ClientCertificateRef
+Gateway.Spec.BackendTLS.Validation
+```
+
+This is all tied to the need to provide backwards compatibility with the TLS
+config that is already GA on Listeners (`Listener.TLS`). If we were naming that
+field today with the broader vision of TLS configuration throughout Gateway API
+we would almost certainly choose a more descriptive name that was more clearly
+tied to Frontend TLS. Unfortunately we're too late to change that name, but we
+can try to make the rest of the terminology clearly tied to frontend or backend
+TLS.
+
+One [suggestion by @candita](https://github.com/kubernetes-sigs/gateway-api/pull/2910#discussion_r1552534017)
+would be to introduce a Listener like resource for BackendTLS, resulting in a
+more consistent naming scheme within Gateway TLS configuration. Although it's
+not yet clear if we need this additional layer, we should reconsider it as we're
+further developing Backend TLS.
--- a/geps/gep-2907/metadata.yaml
+++ b/geps/gep-2907/metadata.yaml
@@ -1,0 +1,15 @@
+apiVersion: internal.gateway.networking.k8s.io/v1alpha1
+kind: GEPDetails
+number: 2907
+name: TLS Configuration Placement and Terminology
+status: Memorandum
+authors:
+  - robscott
+relationships:
+  seeAlso:
+    - number: 91
+      name: Client Certificate Verification for Gateway Listeners
+      description: Will use some of the terminology defined by this GEP.
+    - number: 1897
+      name: TLS from Gateway to Backend
+      description: Will use some of the terminology defined by this GEP.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -146,6 +146,7 @@ nav:
       - geps/gep-1324/index.md
       - geps/gep-2659/index.md
       - geps/gep-2722/index.md
+      - geps/gep-2907/index.md
     - Declined:
       - geps/gep-735/index.md
       - geps/gep-1282/index.md


robscott:
Deploy Preview: https://deploy-preview-2910--kubernetes-sigs-gateway-api.netlify.app/geps/gep-2907/

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+This is already finalized in the API and so we're stuck with this name. In
+hindsight a name that was more clearly tied to frontend TLS would have been
+ideal here.
+
+
+### 3. Configure Client Certificate that Gateway should use to connect to Backend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.BackendTLS.ClientCertificateRef` | Not Proposed |
```
arkodg, line 85: +1 to `Gateway.BackendTLS.ClientCertificateRef`

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+## Proposed Buckets
+Note that this does not represent any form of commitment that any of these
+fields or concepts will exist within Gateway API. If they did, they would follow
+the following naming structure though:
+
+### 1. Validate Client Certificate provided by Frontend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway Listener | `Listener.TLS.FrontendValidation` | Proposed |
```
arkodg, line 62: did you mean `Listener.TLS.Frontend.Validation` ?

robscott: Nope, this was intentional. Part of the idea behind this GEP is that `Listener.TLS` is really entirely focused on 1+2 - the bits of TLS between frontend and Gateway. That means that if we were to add any additional TLS configuration in the `Listerner.TLS` struct, it would tied to that limited scope, and thus we don't really need a separate `Frontend` struct.

One could make an argument that `Listener.TLS.Validation` should be the field name here to avoid any ambiguity, but in this specific context, I think that it's probably helpful to specifically spell out `frontend`.

youngnick: I think that if you include basically the above comment into the text of the GEP, we should be good here.

youngnick:
LGTM with one sadly blocking comment about explaining the naming things better.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 3. Configure Client Certificate that Gateway should use to connect to Backend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.BackendTLS.ClientCertificateRef` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.ClientCertificateRef` | Not Proposed |
+
+#### Rationale
+It's not yet obvious which of the above options are preferable, but a case could
```
candita, line 89: Could you add details on why Listener.TLS.BackendValidation wouldn't work?  Wouldn't it be analogous to Listener.TLS.FrontendValidation that was proposed?

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+This GEP aims to define high level TLS terminology and structure within Gateway
+API to ensure that all our independent proposals related to TLS result in a
+coherent set of APIs.
+
+## Goals
+* Define high-level terminology for how we refer to TLS in Gateway API.
+* Define top level fields where TLS configuration can live.
+
+## Non-Goals
+* Add or change fields directly. (This may inspire changes in other GEPs
```
candita, line 16: I see some proposed changes, so this is not really a non-goal.

candita (commented):
Remembering the discussions around why the contents of BackendTLSPolicy wasn't going to be added directly to Gateway, one reason was because BackendTLS should be accessible to the app developer role, and the burden of app or service-level certificate management shouldn't be placed on the cluster operator. How do we reconcile those discussions with this proposal? Add a new role, like `application operator` or similar?

If we kept both BackendTLSPolicy and Gateway.BackendTLS, then we'd need to define how and when one could override the other. 

How about considering a new policy attachment called `FrontendTLSPolicy`, which would be available to cluster operators, but maybe not app developers. It would be similar to BackendTLSPolicy in many ways, but perhaps targetRefs could be listeners on the gateway?  Would it work for @arkodg's use cases ?

geps/gep-2907/index.md
robscott: I guess what I'm trying to say here is that if we translate this into actual API changes, those will be tracked by the GEPs associated with those APIs. This GEP is a memorandum GEP and not meant to track or own any individual API changes, will try to clean this up.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
+# GEP-2907: TLS Configuration Placement and Terminology
+
+* Issue: [#2907](https://github.com/kubernetes-sigs/gateway-api/issues/2907)
+* Status: Memorandum
+
+## TLDR
+This GEP aims to define high level TLS terminology and structure within Gateway
+API to ensure that all our independent proposals related to TLS result in a
+coherent set of APIs.
```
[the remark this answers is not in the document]
robscott: Added a note linking to the GEPs that will be affected by this one, but wanting to keep the actual ownership and detail of those changes to those GEPs.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+* Add or change fields directly. (This may inspire changes in other GEPs
+  though).
+* Commit to including specific parts of TLS configuration in Gateway API. (This
+  is merely to provide space for future configuration, not a commitment that we
+  will add it to the API.)
+
+## Introduction
+
+There are essentially 4 different "buckets" that could be relevant to TLS
+configuration in Gateway API:
```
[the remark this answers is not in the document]
robscott: Definitely open to better names here, but I think "paths" might unfortunately get confused with HTTP paths.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+## Goals
+* Define high-level terminology for how we refer to TLS in Gateway API.
+* Define top level fields where TLS configuration can live.
+
+## Non-Goals
+* Add or change fields directly. (This may inspire changes in other GEPs
+  though).
+* Commit to including specific parts of TLS configuration in Gateway API. (This
+  is merely to provide space for future configuration, not a commitment that we
+  will add it to the API.)
```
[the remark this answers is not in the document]
robscott: Going to come back to this one as it requires more thought. They're not necessarily non-goals, just on the periphery of what I was focused on here.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+```mermaid
+flowchart LR
+    Frontend -.->|1. Client Cert| Gateway
+    Gateway -.->|2. Server Cert| Frontend
+    Gateway -.->|3. Client Cert| Backend
+    Backend -.->|4. Server Cert| Gateway
+```
+
+### Where it can fit
+We have three different places where this configuration could live:
```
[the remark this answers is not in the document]
robscott: Updated, but left "three" in for now since I don't think this number will change.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### A. Gateways
+We can add config that applies to an entire Gateway. Config attached at this
+layer will apply to everything behind a single VIP.
```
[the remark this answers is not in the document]
robscott: Good call, this was confusing. Reworked this a bit since I'm mostly trying to describe the concept of a Gateway when it comes to TLS to someone that might not be that familiar with the rest of the API.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### B. Gateway Listeners
+Each Gateway contains multiple “Listeners”. Each Listener in a Gateway must have
+a unique combination of protocol, port, and hostname (SNI in the case of HTTPS).
+TLS configuration attached at this layer needs to be isolated via SNI.
```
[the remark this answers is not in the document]
robscott: Good call, added a link to our Listener Isolation docs.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+#### C. BackendTLSPolicy
+This policy allows us to attach unique TLS configuration per Backend. Depending
+on the organization, this policy may be owned by the application owner or the LB
+admin. Note that this configuration will be used by all Gateways (potentially
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+* Commit to including specific parts of TLS configuration in Gateway API. (This
+  is merely to provide space for future configuration, not a commitment that we
+  will add it to the API.)
+
+## Introduction
+
+There are essentially 4 different "buckets" that could be relevant to TLS
+configuration in Gateway API:
+
+```mermaid
```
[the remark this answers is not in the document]
robscott: Good call, reworked this to include definitions.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+#### C. BackendTLSPolicy
+This policy allows us to attach unique TLS configuration per Backend. Depending
+on the organization, this policy may be owned by the application owner or the LB
+admin. Note that this configuration will be used by all Gateways (potentially
+from different implementations) that are connecting to the backend.
+
+## Proposed Buckets
+Note that this does not represent any form of commitment that any of these
+fields or concepts will exist within Gateway API. If they did, they would follow
+the following naming structure though:
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+hindsight a name that was more clearly tied to frontend TLS would have been
+ideal here.
+
+
+### 3. Configure Client Certificate that Gateway should use to connect to Backend
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.BackendTLS.ClientCertificateRef` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.ClientCertificateRef` | Not Proposed |
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

geps/gep-2907/index.md
robscott: Added a paragraph to cover this. In general, having FrontendValidation at the Listener level allows a Gateway to accept different client certs per SNI (a matching attribute of the frontend -> Gateway connection). On the other hand, when determining the identity a Gateway should use when connecting to a backend, it should likely either be tied directly to the Gateway or Backend, but the Listener is not particularly relevant in this context.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,109 @@
...
+### Where it can fit
+We have three different places where this configuration could live:
+
+#### A. Gateways
+We can add config that applies to an entire Gateway. Config attached at this
+layer will apply to everything behind a single VIP.
+
+#### B. Gateway Listeners
+Each Gateway contains multiple “Listeners”. Each Listener in a Gateway must have
+a unique combination of protocol, port, and hostname (SNI in the case of HTTPS).
```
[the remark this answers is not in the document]
robscott: Updated, thanks!

arkodg (approved):
Looks good, thanks for adding this GEP

geps/gep-2907/index.md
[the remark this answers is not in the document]
Miciah: How about "legs in a client session"?  Some alternative words: segments, links, components.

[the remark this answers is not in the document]
candita: +1 to segments

youngnick:
/lgtm

subject to @candita and @Miciah being satisfied with language updates. This is a 1.1 release blocker, since it blocks moving forward with GEP-91, which is also a release blocker for 1.1.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,127 @@
...
+the Gateway or Backend, but the Listener is not particularly relevant in this
+context
+
+
+### 4. Validate Server Certificate that is provided by Backend
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.Spec.BackendTLS.Validation` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.Spec.Validation` | Experimental (Different Name) |
+
```
dprotaso, line 122: Can we settle what to do hostnames/sans?

eg. in the client validation GEP we have

```go
  // SubjectAltNames contains one or more alternate names to verify
  // the subject identity in the certificate presented by the client.
  //
  // Support: Core
  //
  // +optional
  SubjectAltNames []string `json:"subjectAltNames,omitempty"`
```

but in BackendTLSPolicy we have

```go
// Hostname is used for two purposes in the connection between Gateways and
// backends:
//
// 1. Hostname MUST be used as the SNI to connect to the backend (RFC 6066).
// 2. Hostname MUST be used for authentication and MUST match the certificate
//    served by the matching backend.
//
// Support: Core
Hostname v1beta1.PreciseHostname `json:"hostname"`
```

It'd be great to one name these things the same and potentially settle on supporting alternate names in both frontend and backend use cases.

dprotaso: cc @arkodg

robscott: For now I believe @arkodg has removed SANs from the client validation GEP. I think we definitely need to consider both if/when there's overlap again, I'm particularly interested in adding SANs to BackendTLSPolicy for example, but I think both are out of scope for v1.1 so we can follow up when they're proposed in the future.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,127 @@
...
+There are essentially 4 different "buckets" that could be relevant to TLS
+configuration in Gateway API:
+
+```mermaid
+flowchart LR
+    Frontend -.->|1. Client Cert| Gateway
+    Gateway -.->|2. Server Cert| Frontend
+    Gateway -.->|3. Client Cert| Backend
+    Backend -.->|4. Server Cert| Gateway
+```
```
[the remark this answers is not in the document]
robscott: Good call, added, thanks! I also reordered the sections here so they're hopefully a bit less confusing.

geps/gep-2907/index.md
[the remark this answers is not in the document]
robscott: Added an "Out of initial scope" section below this to cover this topic for now.

geps/gep-2907/index.md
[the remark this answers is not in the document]
robscott: Updated, thanks!

geps/gep-2907/index.md
robscott: Copied nearly verbatim, thanks!

robscott:
Thanks everyone for the feedback! I think I've addressed everything now, so PTAL. I'm going to remove the hold so the next LGTM merges this. I think we've got pretty good consensus around this GEP at this point and I really want to unblock @arkodg's GEP that depends on this one.

/hold cancel

geps/gep-2907/index.md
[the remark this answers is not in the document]
candita: I called it "Longer Term Goals" in GEP-1897, but they mean the same thing.

candita:
Sorry, I feel like we are rushing this.

/hold

geps/gep-2907/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+#### C. BackendTLSPolicy
+This policy allows us to attach unique TLS configuration per Backend. Depending
+on the organization, this policy may be owned by the application owner or the
+cluster operator. Note that this configuration will be used by all Gateways
+(potentially from different implementations) that are connecting to the backend.
+
+### "Frontend" and "Backend"
+A guiding principle in this naming is to use consistent naming for “Upstream”
+(1+2) and “Downstream” (3+4), similar to Envoy. To avoid the confusion with what
```
frankbu, line 56: For `similar to Envoy` terminology, 3+4 are the upstream, 1+2 is downstream. I see what you mean by "avoid confusion" 😃 
```suggestion
A guiding principle in this naming is to use consistent naming for “Upstream”
(3+4) and “Downstream” (1+2), similar to Envoy. To avoid the confusion with what
```

geps/gep-2907/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+| Proposed Placement | Name | Status |
+|-|-|-|
+| Gateway | `Gateway.Spec.BackendTLS.ClientCertificateRef` | Not Proposed |
+| BackendTLSPolicy | `BackendTLSPolicy.Spec.ClientCertificateRef` | Not Proposed |
+
+#### Rationale
+It's not yet obvious which of the above options are preferable, but a case could
+be made for either or both. If we add a `BackendTLS` struct to Gateway it would
+leave room for concepts like TLS version, ALPN, cipher suites, etc.
```
candita, line 127: @robscott I feel we shouldn't add certain backend configuration that the app developer should have access to.  This part should be left TBD if it's not needed to unblock GEP-91.  We should not have a proposal that includes the indecision of "either or both" right from the beginning.

I mentioned earlier:

> Remembering the discussions around why the content of BackendTLSPolicy wasn't going to be added directly to Gateway, one reason was because BackendTLS should be accessible to the app developer role, and the burden of app or service-level certificate management shouldn't be placed on the cluster operator. How do we reconcile those discussions with this proposal? Add a new role, like application operator or similar?


>How about considering a new policy attachment called FrontendTLSPolicy, which would be available to cluster operators, but maybe not app developers. It would be similar to BackendTLSPolicy in many ways, but perhaps targetRefs could be listeners on the gateway? Would it work for @arkodg's use cases ?

geps/gep-2907/index.md
robscott: 🤦  updated, thanks!

geps/gep-2907/index.md
robscott: Good question, I think there are several reasons BackendTLSPolicy exists as a separate policy resource:

1. Some parts of the config, such as the SNI that should be used to connect to the Service/backend only make sense when attached per Service - it's very unlikely that you'd want to use the same SNI to connect to all backends from a Gateway.
2. The Service API is overloaded and very slow to change, Gateway API has taken the approach of exposing config at that level via policy resources that are tied to specific personas (why Session Persistence config and Backend TLS are in different resources).

I don't think either of those reasons really apply to frontend TLS. Instead, frontend TLS could theoretically be configured either for an entire Gateway, or a specific Gateway listener. Given that many implementations already support distinguishing this config per SNI, it seems to make sense to start with listener level attachment. I think that the persona that would be responsible for a Gateway is not sufficiently different than the persona that would be responsible for frontend TLS, so I tend to think that the current proposal is the best option available to us.

I'll work on clarifying all of that in the GEP.

geps/gep-2907/index.md
```diff
@@ -0,0 +1,183 @@
...
+#### C. BackendTLSPolicy
+This policy allows us to attach unique TLS configuration per Backend. Depending
+on the organization, this policy may be owned by the application owner or the
+cluster operator. Note that this configuration will be used by all Gateways
+(potentially from different implementations) that are connecting to the backend.
+
+### "Frontend" and "Backend"
+A guiding principle in this naming is to use consistent naming for “Downstream”
+(1+2) and “Upstream” (3+4), similar to Envoy. To avoid the confusion with what
+is upstream and downstream, Gateway API will use “Frontend” and “Backend”.
```
candita, line 57:
```suggestion
is upstream and downstream, Gateway API will use “Frontend” (1+2) and “Backend” (3+4).
```

geps/gep-2907/index.md
```diff
@@ -0,0 +1,183 @@
...
+    Gateway -.->|3. Client Cert| Backend
+    Backend -.->|4. Server Cert| Gateway
+```
+
+The above diagram depicts these four segments as edges in a graph.
+
+## Proposed Segments
+Note that this does not represent any form of commitment that any of these
+fields or concepts will exist within Gateway API. If or when they do, we propose
+the following naming structure:
```
candita, line 80: I'd like to summarize the proposed segments, only because when I look at them together they don't seem as unified as they could be:

> Gateway.Listener.TLS.FrontendValidation
> Gateway.Listener.TLS
> Gateway.Spec.BackendTLS.ClientCertificateRef
> Gateway.Spec.BackendTLS.Validation

To me the problem of dealing with Listener is better solved by doing something like this:

> Gateway.Listener.TLS.FrontendValidation
> Gateway.Listener.TLS
> Gateway.Transmitter.TLS.BackendValidation
> Gateway.Transmitter.TLS
 
or even

> Gateway.Listener.TLS.Validation
> Gateway.Listener.TLS
> Gateway.Transmitter.TLS.Validation
> Gateway.Transmitter.TLS

We don't have to use the term `Transmitter`, just something other than `BackendTLS`, which clashes with `BackendTLSPolicy`.  I feel a move toward this pattern presents an easier learning curve for people with memory challenges (like me).

geps/gep-2907/index.md
robscott: Added a new section "Why Not FrontendTLSPolicy?" to address this.

geps/gep-2907/index.md
robscott: Good call, updated, thanks!

geps/gep-2907/index.md
robscott: Good questions, I've added two separate but related sections to address this:

* Why the inconsistency between Frontend and Backend TLS config on Gateways?
* Why Not Listener level to match FrontendValidation?

I've also added a note that we should reconsider these answers before developing any kind of backend TLS config at a Gateway level. Hope that all works for you.

robscott:
Thanks for all the great feedback @candita, I think I've addressed everything now. Would appreciate an additional round of review and/or LGTM from anyone that has time so we can get this one in and unblock the dependent updates.

/hold cancel

arkodg (approved):
/lgtm

